PR #116, opened 2017-06-21
Sort ResourceID.Path keys for consistent output

jbardin:
While the API may not care what order the key-value path parts are in,
sort the keys so that we can maintain a consistent order for tests and
debugging.

Add more key-values to the test so that it fails more frequently than
not.


stack72 (commented):
LGTM :)

PR #139, opened 2017-06-27
[MS] Fixes #127 Improve docs for delete_os_disk_on_termination

### LINKED ISSUE #127: Delete managed OS disk when destroying VM
Given an `azurerm_virtual_machine` using a managed OS disk is created then destroyed. The same configuration can not be reused because the managed OS disk resource still exists. When a VM is destroyed and has a managed OS disk, the OS disk should be destroyed too to prevent leaking of resources.

Data disks have the problem too, but it can be mitigated using the `azurerm_managed_disk` resource separately and linking to that created instance in the `storage_data_disk` block. The OS block can not do this due to `azurerm_managed_disk` not supporting creation `FromImage`.

StephenWeatherford:
What do you think?  The other possibility is to add a new example example "Usage (Managed Disks, Deleted When VM Deleted)", or maybe just add a note.

### DIFF
--- a/website/docs/r/virtual_machine.html.markdown
+++ b/website/docs/r/virtual_machine.html.markdown
@@ -10,6 +10,110 @@ description: |-
 
 Create a virtual machine.
 
+## Example Usage (Managed Disks, recommended)
+
+```hcl
+resource "azurerm_resource_group" "test" {
+  name     = "acctestrg"
+  location = "West US 2"
+}
+
+resource "azurerm_virtual_network" "test" {
+  name                = "acctvn"
+  address_space       = ["10.0.0.0/16"]
+  location            = "West US 2"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+}
+
+resource "azurerm_subnet" "test" {
+  name                 = "acctsub"
+  resource_group_name  = "${azurerm_resource_group.test.name}"
+  virtual_network_name = "${azurerm_virtual_network.test.name}"
+  address_prefix       = "10.0.2.0/24"
+}
+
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
Hey @StephenWeatherford

Thanks for this PR - this LGTM. Thinking about it, I don't think this warrants a separate example - however it'd be good to also integrate this into the Unmanaged Disks example too.

Looking at the examples - given Managed Disks are the recommended approach going forward, is it worth moving this to be the first example on the page? In addition - given the Managed Disks example contains multiple disks, we could _probably_ do the same for the Unmanaged disks and combine the two examples - what do you think?

StephenWeatherford:
Great suggestions, Tom!  Implemented.

website/docs/r/virtual_machine.html.markdown
```diff
@@ -77,10 +74,27 @@ resource "azurerm_virtual_machine" "test" {
...
-    vhd_uri       = "${azurerm_storage_account.test.primary_blob_endpoint}${azurerm_storage_container.test.name}/myosdisk1.vhd"
-    caching       = "ReadWrite"
-    create_option = "FromImage"
+    name              = "myosdisk1"
+    caching           = "ReadWrite"
+    create_option     = "FromImage"
+    managed_disk_type = "Standard_LRS"
+  }
+
+  // Optional data disks
```
tombuildsstuff: checking our other examples we use `#` for comments in HCL rather than `//` - I'm going to push a commit to update this :)

tombuildsstuff:
Awesome - thanks for this @StephenWeatherford :)

PR #172, opened 2017-07-17
Adding Import support to Search Service

tombuildsstuff:
Adding Import support to `azurerm_search_service`

Tests pass: 

```
$ TF_ACC=1 envchain azurerm go test ./azurerm -v -timeout 120m -run TestAccAzureRMSearchService
=== RUN   TestAccAzureRMSearchService_importBasic
--- PASS: TestAccAzureRMSearchService_importBasic (88.71s)
=== RUN   TestAccAzureRMSearchService_basic
--- PASS: TestAccAzureRMSearchService_basic (310.06s)
=== RUN   TestAccAzureRMSearchService_updateReplicaCountAndTags

--- PASS: TestAccAzureRMSearchService_updateReplicaCountAndTags (1997.81s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	2396.616s
```

### DIFF
--- a/website/docs/r/search_service.html.markdown
+++ b/website/docs/r/search_service.html.markdown
@@ -21,7 +21,7 @@ resource "azurerm_resource_group" "test" {
 resource "azurerm_search_service" "test" {
   name                = "acceptanceTestSearchService1"
   resource_group_name = "${azurerm_resource_group.test.name}"
-  location            = "West US"
+  location            = "${azurerm_resource_group.test.location}"
   sku                 = "standard"
 
   tags {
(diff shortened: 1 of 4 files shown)


website/docs/r/search_service.html.markdown
```diff
@@ -54,3 +54,11 @@ The following arguments are supported:
 The following attributes are exported:
 
 * `id` - The Search Service ID.
+
+## Import
+
+Search Services can be imported using the `resource id`, e.g.
```
radeksimko: I guess the intention here was to put back ticks just around `id`?

radeksimko (approved):
This looks functionally good to me, but I'm not sure the `%+v` is a good idea.

I understand the motivation is probably to get all the details about the error (including type, code etc.), so we can then easily catch it if necessary, but I think this is developer's concern, not user's.
User should see a user-friendly error, which is hopefully what `%s` provides - i.e. the `String()` or default string representation of that error is user-friendly.

That said I'm not familiar with error handling conventions in Azure, so I might not know the full background.

website/docs/r/search_service.html.markdown
tombuildsstuff: We can do, but we should update all of them to match this, as this was C&P'd from an existing page for consistency ([example](https://www.terraform.io/docs/providers/azurerm/r/eventhub_namespace.html#import))

PR #173, opened 2017-07-17
Removing Location from the traffic manager docs

### LINKED ISSUE #169: Remove the *location* from the azurerm_traffic_manager_profile sample on documentation
**Observed:**
Trying Azure Traffic Manager Profile sample available at https://www.terraform.io/docs/providers/azurerm/r/traffic_manager_profile.html leads to the following error:
* azurerm_traffic_manager_profile.xxxxx: : invalid or unknown key: location

**Fix**
Remove the *location* from the azurerm_traffic_manager_profile sample on documentation available at https://www.terraform.io/docs/providers/azurerm/r/traffic_manager_profile.html.

tombuildsstuff:
Fixes #169

### DIFF
--- a/website/docs/r/traffic_manager_profile.html.markdown
+++ b/website/docs/r/traffic_manager_profile.html.markdown
@@ -14,10 +14,8 @@ Creates a Traffic Manager Profile to which multiple endpoints can be attached.
 
 ```hcl
 resource "azurerm_traffic_manager_profile" "test" {
-  name                = "profile1"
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
@radeksimko 

> Just out of interest is this the only way to create a TM in Azure or is location something we plan to open up to the user eventually?

They're a global-only resource, so AFAIK there's no intention to make that a user-facing option

PR #175, opened 2017-07-19
Adds service_principal_object_id attribute to data_source_arm_client_config

whiskeyjay:
The service principal ID is needed for granting a SP access to certain resources. For example, adding secrets to KeyVault and give access to the SP in the configuration (although currently the TF KeyVault resource doesn't support adding secrets yet). With this new attribute uses won't have to manually copy the object ID of the SP from portal and hard code it in the configuration file.


azurerm/data_source_arm_client_config_test.go
```diff
@@ -43,6 +46,22 @@ func testAzureRMClientConfigAttr(name, key, value string) resource.TestCheckFunc
 	}
 }
 
+func testAzureRMClientConfigGUIDAttr(name, key string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		r, err := regexp.Compile("^[A-Fa-f0-9]{8}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{12}$")
```
tombuildsstuff: would it make more sense to use [this method in the `go.uuid` library](http://godoc.org/github.com/satori/go.uuid#FromString) (which is already vendored) for this?

tombuildsstuff (approved):
Hey @whiskeyjay 

Thanks for this - one small question but this otherwise LGTM :)

azurerm/data_source_arm_client_config_test.go
whiskeyjay: I thought it is easier this way because I don't have to extract the attribute value from state myself?

tombuildsstuff: fair point - it's not a blocker by any means - just seeing if we can reuse the UUID parser if it made sense :)

tombuildsstuff:
Tests pass:
```
$ TF_ACC=1 envchain azurerm go test ./azurerm -v -timeout 120m -run TestAccAzureRMClientConfig
=== RUN   TestAccAzureRMClientConfig_basic
--- PASS: TestAccAzureRMClientConfig_basic (11.59s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	11.620s
```

PR #179, opened 2017-07-20
resource_arm_dns_ns_record: switch dependency from riviera to azure-sdk-for-go

sebastus:


azurerm/resource_arm_dns_ns_record_test.go
```diff
@@ -99,42 +100,49 @@ func testCheckAzureRMDnsNsRecordExists(name string) resource.TestCheckFunc {
...
-		readResponse, err := readRequest.Execute()
 		if err != nil {
-			return fmt.Errorf("Bad: GetNSRecordSet: %s", err)
+			return nil
 		}
 
-		if readResponse.IsSuccessful() {
-			return fmt.Errorf("Bad: DNS NS Record still exists: %s", readResponse.Error)
+		if resp.StatusCode != http.StatusNotFound {
+			return fmt.Errorf("DNS MSNSecord still exists:\n%#v", resp.RecordSetProperties)
```
tombuildsstuff: _minor_ I think this should be `NSRecord`? :)

azurerm/resource_arm_dns_ns_record_test.go
```diff
@@ -99,42 +100,49 @@ func testCheckAzureRMDnsNsRecordExists(name string) resource.TestCheckFunc {
...
+		nsName := rs.Primary.Attributes["name"]
+		zoneName := rs.Primary.Attributes["zone_name"]
+		resourceGroup := rs.Primary.Attributes["resource_group_name"]
+
+		resp, err := conn.Get(resourceGroup, zoneName, nsName, dns.NS)
 
-		readResponse, err := readRequest.Execute()
 		if err != nil {
-			return fmt.Errorf("Bad: GetNSRecordSet: %s", err)
+			return nil
```
tombuildsstuff: same as the other PR's - I think we should be checking a 404 above this, and then returning if there's an error?

azurerm/resource_arm_dns_ns_record.go
```diff
@@ -59,147 +59,128 @@ func resourceArmDnsNsRecord() *schema.Resource {
...
-		ZoneName:          d.Get("zone_name").(string),
+	parameters := dns.RecordSet{
+		Name:                &name,
+		RecordSetProperties: &props,
 	}
 
-	readResponse, err := readRequest.Execute()
+	//last parameter is set to empty to allow updates to records after creation
+	// (per SDK, set it to '*' to prevent updates, all other values are ignored)
+	resp, err := dnsClient.CreateOrUpdate(resGroup, zoneName, name, dns.NS, parameters, "", "")
```
tombuildsstuff: what's the penultimate option for here? generally when there's more than one empty argument we'd tend to separate these out into variables above, i.e.:
```
foo := "bar"
bar := "foo"
...CreateOrUpdate(...., foo, bar)
```

azurerm/resource_arm_dns_ns_record.go
```diff
@@ -59,147 +59,128 @@ func resourceArmDnsNsRecord() *schema.Resource {
...
-		return fmt.Errorf("Error creating DNS NS Record: %s", err)
-	}
-	if !createResponse.IsSuccessful() {
-		return fmt.Errorf("Error creating DNS NS Record: %s", createResponse.Error)
+	records, err := expandAzureRmDnsNsRecords(d)
+	props := dns.RecordSetProperties{
+		Metadata:  metadata,
+		TTL:       &ttl,
+		NsRecords: &records,
 	}
```
tombuildsstuff: can we move this into the `parameters` assignment below?

PR #180, opened 2017-07-20
resource_arm_dns_a_record: switch dependency from riviera to azure-sdk-for-go

sebastus:


azurerm/resource_arm_dns_a_record.go
```diff
@@ -53,124 +53,121 @@ func resourceArmDnsARecord() *schema.Resource {
...
-	records := make([]dns.ARecord, len(recordStrings))
-	for i, v := range recordStrings {
-		records[i] = dns.ARecord{
-			IPv4Address: v.(string),
-		}
+	records, err := expandAzureRmDnsARecords(d)
+	props := dns.RecordSetProperties{
+		Metadata: metadata,
+		TTL:      &ttl,
+		ARecords: &records,
```
tombuildsstuff: _minor_ we might as well combine this with the `parameters` assignment below?

azurerm/resource_arm_dns_a_record.go
```diff
@@ -53,124 +53,121 @@ func resourceArmDnsARecord() *schema.Resource {
...
-	readRequest.Command = &dns.GetARecordSet{
-		Name:              d.Get("name").(string),
-		ResourceGroupName: d.Get("resource_group_name").(string),
-		ZoneName:          d.Get("zone_name").(string),
-	}
-
-	readResponse, err := readRequest.Execute()
+	//last parameter is set to empty to allow updates to records after creation
+	// (per SDK, set it to '*' to prevent updates, all other values are ignored)
+	resp, err := dnsClient.CreateOrUpdate(resGroup, zoneName, name, dns.A, parameters, "", "")
```
tombuildsstuff: what's the penultimate argument for here? when there's more than one empty string we tend to separate these out into variables i.e.:
```
foo := "bar"
bar := "foo"
...CreateOrUpdate(..., foo, bar)
```

azurerm/resource_arm_dns_a_record_test.go
```diff
@@ -99,42 +100,49 @@ func testCheckAzureRMDnsARecordExists(name string) resource.TestCheckFunc {
...
+		aName := rs.Primary.Attributes["name"]
+		zoneName := rs.Primary.Attributes["zone_name"]
+		resourceGroup := rs.Primary.Attributes["resource_group_name"]
+
+		resp, err := conn.Get(resourceGroup, zoneName, aName, dns.A)
 
-		readResponse, err := readRequest.Execute()
 		if err != nil {
-			return fmt.Errorf("Bad: GetARecordSet: %s", err)
+			return nil
```
tombuildsstuff: I think we should be checking for a 404 above this, and then returning the error otherwise, rather than ignoring this error?

PR #182, opened 2017-07-20
Resource arm dns cname record

sebastus:


tombuildsstuff:
Tests pass:

```
$ acctests azurerm TestAccAzureRMDnsCNameRecord_
=== RUN   TestAccAzureRMDnsCNameRecord_importBasic
--- PASS: TestAccAzureRMDnsCNameRecord_importBasic (101.52s)
=== RUN   TestAccAzureRMDnsCNameRecord_basic
--- PASS: TestAccAzureRMDnsCNameRecord_basic (106.86s)
=== RUN   TestAccAzureRMDnsCNameRecord_subdomain
--- PASS: TestAccAzureRMDnsCNameRecord_subdomain (88.39s)
=== RUN   TestAccAzureRMDnsCNameRecord_updateRecords
--- PASS: TestAccAzureRMDnsCNameRecord_updateRecords (104.18s)
=== RUN   TestAccAzureRMDnsCNameRecord_withTags
--- PASS: TestAccAzureRMDnsCNameRecord_withTags (111.62s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	512.600s
```

azurerm/resource_arm_dns_cname_record.go
```diff
@@ -54,112 +54,112 @@ func resourceArmDnsCNameRecord() *schema.Resource {
...
-	d.Set("name", resp.Name)
-	d.Set("resource_group_name", id.ResourceGroup)
-	d.Set("zone_name", id.Path["dnszones"])
-	d.Set("ttl", resp.TTL)
-	d.Set("record", resp.CNAMERecord.CNAME)
+	d.Set("name", name)
+	d.Set("resource_group_name", resGroup)
+	d.Set("zone_name", zoneName)
+	d.Set("ttl", *result.TTL)
+	d.Set("etag", *result.Etag)
```
tombuildsstuff: we should be able to remove the `*` for these two sets as `d.Set()` should take care of that for us automatically

azurerm/resource_arm_dns_cname_record.go
```diff
@@ -54,112 +54,112 @@ func resourceArmDnsCNameRecord() *schema.Resource {
...
+	metadata := expandTags(tags)
 
-	createResponse, err := createRequest.Execute()
-	if err != nil {
-		return fmt.Errorf("Error creating DNS CName Record: %s", err)
-	}
-	if !createResponse.IsSuccessful() {
-		return fmt.Errorf("Error creating DNS CName Record: %s", createResponse.Error)
+	props := dns.RecordSetProperties{
+		Metadata:    metadata,
```
tombuildsstuff: from what I can see, these are the same thing, but just to confirm: is `metadata` the same thing as `tags`? [The API documentation](https://docs.microsoft.com/en-us/rest/api/dns/recordsets#RecordSets_CreateOrUpdate) doesn't make it clear - but [we currently submit these as `tags`](https://github.com/jen20/riviera/blob/master/dns/create_dns_cname_recordset.go#L23)?

azurerm/resource_arm_dns_cname_record.go
```diff
@@ -54,112 +54,112 @@ func resourceArmDnsCNameRecord() *schema.Resource {
...
-	d.Set("resource_group_name", id.ResourceGroup)
-	d.Set("zone_name", id.Path["dnszones"])
-	d.Set("ttl", resp.TTL)
-	d.Set("record", resp.CNAMERecord.CNAME)
+	d.Set("name", name)
+	d.Set("resource_group_name", resGroup)
+	d.Set("zone_name", zoneName)
+	d.Set("ttl", *result.TTL)
+	d.Set("etag", *result.Etag)
+	d.Set("record", cName)
```
tombuildsstuff: can we check to see that `result.RecordSetProperties.CnameRecord` isn't nil here - unfortunately swagger/server-side changes could lead to Terraform crashing if we don't check - so it'd be good to change this too:

```
if props := result.RecordSetProperties; props != nil {
  if record := props.CnameRecord; record != nil {
    d.Set("record", record.Cname)
  }
}
```

(again, we shouldn't need the `*` here because `d.Set` will handle that automatically)

azurerm/resource_arm_dns_cname_record.go
```diff
@@ -54,112 +54,112 @@ func resourceArmDnsCNameRecord() *schema.Resource {
...
 
-	flattenAndSetTags(d, &resp.Tags)
+	flattenAndSetTags(d, result.Metadata)
```
tombuildsstuff: (same question  about Metadata+Tags)

azurerm/resource_arm_dns_cname_record.go
```diff
@@ -54,112 +54,112 @@ func resourceArmDnsCNameRecord() *schema.Resource {
 				Required: true,
 			},
 
+			"etag": {
```
tombuildsstuff: can we add this to the documentation too?

sebastus: Tom - I'm new here so still learning. eTag can be left out - I'll remove it.

azurerm/resource_arm_dns_cname_record.go
sebastus: In the portal and in the sdk that property is called metadata, but it's otherwise identical to tags on other resources. and yes, in the tf file they are tags. I'll add a test to cover it.

azurerm/resource_arm_dns_cname_record.go
sebastus: done

azurerm/resource_arm_dns_cname_record.go
sebastus: got it.

azurerm/resource_arm_dns_cname_record.go
sebastus: test was already present.

azurerm/resource_arm_dns_cname_record.go
sebastus: same answer.

sebastus: I guess the question back to you is: do you prefer the provider match the nomenclature in the tf file or that in the sdk?

tombuildsstuff: >  do you prefer the provider match the nomenclature in the tf file or that in the sdk?

so we should be aiming for consistency with the schema (i.e. what's used in the `*.tf` file) - rather than renaming this to match the SDK; providing the field we send to/from the API is actually tags (albeit renamed), then that's fine :) i.e. users should be able to upgrade from resources made with the Riviera SDK to resources made with the Azure SDK without any code changes :)

(from what I can see, if this is tags then I'd consider this inconsistency some design feedback for the Azure API's, as the rest of the Azure API's refer to this field as `tags`)

azurerm/resource_arm_dns_cname_record.go
```diff
@@ -59,107 +59,101 @@ func resourceArmDnsCNameRecord() *schema.Resource {
...
-	createResponse, err := createRequest.Execute()
-	if err != nil {
-		return fmt.Errorf("Error creating DNS CName Record: %s", err)
-	}
-	if !createResponse.IsSuccessful() {
-		return fmt.Errorf("Error creating DNS CName Record: %s", createResponse.Error)
+	props := dns.RecordSetProperties{
+		Metadata:    metadata,
+		TTL:         &ttl,
+		CnameRecord: &cnameRecord,
```
tombuildsstuff: _minor_: can we combine this into the `properties` object below?

azurerm/resource_arm_dns_cname_record.go
```diff
@@ -59,107 +59,101 @@ func resourceArmDnsCNameRecord() *schema.Resource {
...
-		ZoneName:          d.Get("zone_name").(string),
+	parameters := dns.RecordSet{
+		Name:                &name,
+		RecordSetProperties: &props,
 	}
 
-	readResponse, err := readRequest.Execute()
+	//last parameter is set to empty to allow updates to records after creation
+	// (per SDK, set it to '*' to prevent updates, all other values are ignored)
+	resp, err := dnsClient.CreateOrUpdate(resGroup, zoneName, name, dns.CNAME, parameters, "", "")
```
tombuildsstuff: what's the penultimate argument for here? in general we'd tend to split these out into assigned variables if there's multiple

i.e.
```
foo := "bar"
bar := "foo"
...CreateOrUpdate(..., foo, bar)
```

azurerm/resource_arm_dns_cname_record.go
```diff
@@ -59,107 +59,101 @@ func resourceArmDnsCNameRecord() *schema.Resource {
...
-	if !deleteResponse.IsSuccessful() {
-		return fmt.Errorf("Error deleting DNS CName Record: %s", deleteResponse.Error)
+
+	resGroup := id.ResourceGroup
+	name := id.Path["CNAME"]
+	zoneName := id.Path["dnszones"]
+
+	resp, error := dnsClient.Delete(resGroup, zoneName, name, dns.CNAME, "")
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("Error deleting DNS CNAME Record %s: %s", name, error)
```
tombuildsstuff: can we make the second argument `%+v` to output the full error here?

azurerm/resource_arm_dns_cname_record_test.go
```diff
@@ -116,42 +117,49 @@ func testCheckAzureRMDnsCNameRecordExists(name string) resource.TestCheckFunc {
...
+		cnameName := rs.Primary.Attributes["name"]
+		zoneName := rs.Primary.Attributes["zone_name"]
+		resourceGroup := rs.Primary.Attributes["resource_group_name"]
+
+		resp, err := conn.Get(resourceGroup, zoneName, cnameName, dns.CNAME)
 
-		readResponse, err := readRequest.Execute()
 		if err != nil {
-			return fmt.Errorf("Bad: GetCNAMERecordSet: %s", err)
+			return nil
```
tombuildsstuff: I think we should be checking for the 404 above this, and then returning the error if it's present - rather than hiding this?

PR #185, opened 2017-07-21
resource_arm_dns_mx_record: switch dependency from riviera to azure-sdk-for-go

sebastus:


azurerm/resource_arm_dns_mx_record.go
```diff
@@ -67,150 +68,140 @@ func resourceArmDnsMxRecord() *schema.Resource {
...
-	if err != nil {
-		return fmt.Errorf("Error creating DNS MX Record: %s", err)
-	}
-	if !createResponse.IsSuccessful() {
-		return fmt.Errorf("Error creating DNS MX Record: %s", createResponse.Error)
+	records, err := expandAzureRmDnsMxRecords(d)
+	props := dns.RecordSetProperties{
+		Metadata:  metadata,
+		TTL:       &ttl,
+		MxRecords: &records,
```
tombuildsstuff: _minor_ can we combine this into the `properties` assignment below?

azurerm/resource_arm_dns_mx_record.go
```diff
@@ -67,150 +68,140 @@ func resourceArmDnsMxRecord() *schema.Resource {
...
-		ZoneName:          d.Get("zone_name").(string),
+	parameters := dns.RecordSet{
+		Name:                &name,
+		RecordSetProperties: &props,
 	}
 
-	readResponse, err := readRequest.Execute()
+	//last parameter is set to empty to allow updates to records after creation
+	// (per SDK, set it to '*' to prevent updates, all other values are ignored)
+	resp, err := dnsClient.CreateOrUpdate(resGroup, zoneName, name, dns.MX, parameters, "", "")
```
tombuildsstuff: what's the penultimate argument for here? in general we'd tend to assign the values to variables when there's multiple default values, so that it's clear what the values are

azurerm/resource_arm_dns_mx_record.go
```diff
@@ -67,150 +68,140 @@ func resourceArmDnsMxRecord() *schema.Resource {
...
+// by azure-sdk-for-go to manipulate azure resources, hence Preference
+// is an int32
+func expandAzureRmDnsMxRecords(d *schema.ResourceData) ([]dns.MxRecord, error) {
+	recordStrings := d.Get("record").(*schema.Set).List()
+	records := make([]dns.MxRecord, len(recordStrings))
+
+	for i, v := range recordStrings {
+		mxrecord := v.(map[string]interface{})
+		preference := mxrecord["preference"].(string)
+		i64, _ := strconv.ParseInt(preference, 10, 32)
```
tombuildsstuff: I think we should be able to update the schema and the set to be an `int` here - and then remove the `strconv` parse here?

azurerm/resource_arm_dns_mx_record_test.go
```diff
@@ -99,42 +100,49 @@ func testCheckAzureRMDnsMxRecordExists(name string) resource.TestCheckFunc {
...
+		mxName := rs.Primary.Attributes["name"]
+		zoneName := rs.Primary.Attributes["zone_name"]
+		resourceGroup := rs.Primary.Attributes["resource_group_name"]
+
+		resp, err := conn.Get(resourceGroup, zoneName, mxName, dns.MX)
 
-		readResponse, err := readRequest.Execute()
 		if err != nil {
-			return fmt.Errorf("Bad: GetMXRecordSet: %s", err)
+			return nil
```
tombuildsstuff: I think we should be checking for the 404 above this, and then returning the error if it's present here?

azurerm/resource_arm_dns_mx_record.go
tombuildsstuff: added a TODO for this

PR #186, opened 2017-07-21
resource_arm_dns_srv_record: switch dependency from riviera to azure-sdk-for-go

sebastus:


azurerm/resource_arm_dns_srv_record.go
```diff
@@ -77,154 +77,138 @@ func resourceArmDnsSrvRecord() *schema.Resource {
...
-		return fmt.Errorf("Error creating DNS SRV Record: %s", err)
-	}
-	if !createResponse.IsSuccessful() {
-		return fmt.Errorf("Error creating DNS SRV Record: %s", createResponse.Error)
+	records, err := expandAzureRmDnsSrvRecords(d)
+	props := dns.RecordSetProperties{
+		Metadata:   metadata,
+		TTL:        &ttl,
+		SrvRecords: &records,
 	}
```
tombuildsstuff: _minor_ could we combine this into the `properties` object below?

azurerm/resource_arm_dns_srv_record.go
```diff
@@ -77,154 +77,138 @@ func resourceArmDnsSrvRecord() *schema.Resource {
...
-		ZoneName:          d.Get("zone_name").(string),
+	parameters := dns.RecordSet{
+		Name:                &name,
+		RecordSetProperties: &props,
 	}
 
-	readResponse, err := readRequest.Execute()
+	//last parameter is set to empty to allow updates to records after creation
+	// (per SDK, set it to '*' to prevent updates, all other values are ignored)
+	resp, err := dnsClient.CreateOrUpdate(resGroup, zoneName, name, dns.SRV, parameters, "", "")
```
tombuildsstuff: what's the penultimate option for here? generally when there's more than one empty argument we'd tend to separate these out into variables above, i.e.:
```
foo := "bar"
bar := "foo"
...CreateOrUpdate(...., foo, bar)
```

azurerm/resource_arm_dns_srv_record.go
```diff
@@ -77,154 +77,138 @@ func resourceArmDnsSrvRecord() *schema.Resource {
...
+func flattenAzureRmDnsSrvRecords(records *[]dns.SrvRecord) []map[string]interface{} {
+	results := make([]map[string]interface{}, 0, len(*records))
+
+	if records != nil {
+		for _, record := range *records {
+			results = append(results, map[string]interface{}{
+				"priority": *record.Priority,
+				"weight":   *record.Weight,
+				"port":     *record.Port,
+				"target":   *record.Target,
```
tombuildsstuff: can we use the `azure.String()` wrapper to safely set these values?

azurerm/resource_arm_dns_srv_record_test.go
```diff
@@ -99,42 +100,49 @@ func testCheckAzureRMDnsSrvRecordExists(name string) resource.TestCheckFunc {
...
+		srvName := rs.Primary.Attributes["name"]
+		zoneName := rs.Primary.Attributes["zone_name"]
+		resourceGroup := rs.Primary.Attributes["resource_group_name"]
+
+		resp, err := conn.Get(resourceGroup, zoneName, srvName, dns.SRV)
 
-		readResponse, err := readRequest.Execute()
 		if err != nil {
-			return fmt.Errorf("Bad: GetSRVRecordSet: %s", err)
+			return nil
```
tombuildsstuff: I think we should be checking for the 404 above this, and then returning the error if it's present, rather than masking the fact there's an error here?

PR #187, opened 2017-07-21
resource_arm_dns_txt_record: switch dependency from riviera to azure-sdk-for-go

sebastus:


azurerm/resource_arm_dns_txt_record.go
```diff
@@ -54,152 +54,143 @@ func resourceArmDnsTxtRecord() *schema.Resource {
 				Required: true,
 			},
 
+			"etag": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
tombuildsstuff: as in the other PR - do we need to expose this field?

azurerm/resource_arm_dns_txt_record.go
```diff
@@ -54,152 +54,143 @@ func resourceArmDnsTxtRecord() *schema.Resource {
...
+	metadata := expandTags(tags)
 
-	txtRecords, recordErr := expandAzureRmDnsTxtRecords(d)
-	if recordErr != nil {
-		return fmt.Errorf("Error Building list of Azure RM Txt Records: %s", recordErr)
+	records, err := expandAzureRmDnsTxtRecords(d)
+	props := dns.RecordSetProperties{
+		Metadata:   metadata,
+		TTL:        &ttl,
+		TxtRecords: &records,
```
tombuildsstuff: _minor_ could we merge this into the `properties` object below?

azurerm/resource_arm_dns_txt_record.go
```diff
@@ -54,152 +54,143 @@ func resourceArmDnsTxtRecord() *schema.Resource {
...
+	if records != nil {
+		for _, record := range *records {
+			txtRecord := make(map[string]interface{})
 
-		txtRecord := dns.TXTRecord{
-			Value: data["value"].(string),
-		}
-
-		txtRecords = append(txtRecords, txtRecord)
+			value := (*record.Value)[0]
```
tombuildsstuff: can there be more than one value here?

azurerm/resource_arm_dns_txt_record.go
```diff
@@ -54,152 +54,143 @@ func resourceArmDnsTxtRecord() *schema.Resource {
...
-		txtRecord := make(map[string]interface{})
-		txtRecord["value"] = record.Value
+func expandAzureRmDnsTxtRecords(d *schema.ResourceData) ([]dns.TxtRecord, error) {
+	recordStrings := d.Get("record").(*schema.Set).List()
+	records := make([]dns.TxtRecord, len(recordStrings))
+
+	for i, v := range recordStrings {
+		value := make([]string, 1)
+		record := v.(map[string]interface{})
+		value[0] = record["value"].(string)
```
tombuildsstuff: is this is a single item, we can make this:
```
value := []string { record["value"].(string) }
```

azurerm/resource_arm_dns_txt_record_test.go
```diff
@@ -99,42 +100,49 @@ func testCheckAzureRMDnsTxtRecordExists(name string) resource.TestCheckFunc {
...
+		txtName := rs.Primary.Attributes["name"]
+		zoneName := rs.Primary.Attributes["zone_name"]
+		resourceGroup := rs.Primary.Attributes["resource_group_name"]
+
+		resp, err := conn.Get(resourceGroup, zoneName, txtName, dns.TXT)
 
-		readResponse, err := readRequest.Execute()
 		if err != nil {
-			return fmt.Errorf("Bad: GetTXTRecordSet: %s", err)
+			return nil
```
tombuildsstuff: I think we should be checking for a 404 above this, and then returning this error, rather than masking it?

PR #188, opened 2017-07-21
resource_arm_dns_zone: switch dependency from riviera to azure-sdk-for-go

sebastus:


azurerm/resource_arm_dns_zone.go
```diff
@@ -51,87 +51,78 @@ func resourceArmDnsZone() *schema.Resource {
 				Set:      schema.HashString,
 			},
 
+			"etag": {
```
tombuildsstuff: can we add this to the documentation?

azurerm/resource_arm_dns_zone.go
```diff
@@ -51,87 +51,78 @@ func resourceArmDnsZone() *schema.Resource {
...
+	metadata := expandTags(tags)
 
-	readRequest := rivieraClient.NewRequest()
-	readRequest.Command = &dns.GetDNSZone{
-		Name:              d.Get("name").(string),
-		ResourceGroupName: d.Get("resource_group_name").(string),
+	parameters := dns.Zone{
+		Name:     &zoneName,
+		Location: &location,
+		Tags:     metadata,
```
tombuildsstuff: same question as in the other PR - is `metadata` the same as `tags` in the previous API? the documentation is unclear

azurerm/resource_arm_dns_zone.go
```diff
@@ -143,18 +134,21 @@ func resourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
...
-		return fmt.Errorf("Error deleting DNS Zone: %s", deleteResponse.Error)
+
+	resGroup := id.ResourceGroup
+	zoneName := id.Path["dnszones"]
+
+	resultChan, errorChan := zonesClient.Delete(resGroup, zoneName, "", make(chan struct{}))
+	result := <-resultChan
+	error := <-errorChan
+	if result.Response.StatusCode != http.StatusOK {
+		return fmt.Errorf("Error deleting DNS zone %s: %s", zoneName, error)
```
tombuildsstuff: can we make this second argument `%+v`?

azurerm/resource_arm_dns_zone_test.go
```diff
@@ -68,42 +68,47 @@ func testCheckAzureRMDnsZoneExists(name string) resource.TestCheckFunc {
...
-		readRequest.Command = &dns.GetDNSZone{}
+		zoneName := rs.Primary.Attributes["name"]
+		resourceGroup := rs.Primary.Attributes["resource_group_name"]
+
+		resp, err := conn.Get(resourceGroup, zoneName)
 
-		readResponse, err := readRequest.Execute()
 		if err != nil {
-			return fmt.Errorf("Bad: GetDNSZone: %s", err)
+			return nil
```
tombuildsstuff: I think this should be returning the error and checking for the 404 above it?

azurerm/resource_arm_dns_zone_test.go
```diff
@@ -68,42 +68,47 @@ func testCheckAzureRMDnsZoneExists(name string) resource.TestCheckFunc {
...
 	}
 }
 
 func testCheckAzureRMDnsZoneDestroy(s *terraform.State) error {
-	conn := testAccProvider.Meta().(*ArmClient).rivieraClient
+	conn := testAccProvider.Meta().(*ArmClient).zonesClient
 
 	for _, rs := range s.RootModule().Resources {
-		if rs.Type != "azurerm_dns_zone" {
+		if rs.Type != "azurerm_dns_zone_record" {
```
tombuildsstuff: this should be `azurerm_dns_zone`

azurerm/resource_arm_dns_zone_test.go
```diff
@@ -68,42 +68,47 @@ func testCheckAzureRMDnsZoneExists(name string) resource.TestCheckFunc {
...
+		if !hasResourceGroup {
+			return fmt.Errorf("Bad: no resource group found in state for DNS zone: %s", zoneName)
+		}
 
-		readResponse, err := readRequest.Execute()
+		conn := testAccProvider.Meta().(*ArmClient).zonesClient
+		resp, err := conn.Get(resourceGroup, zoneName)
 		if err != nil {
-			return fmt.Errorf("Bad: GetDNSZone: %s", err)
+			return fmt.Errorf("Bad: Get zone: %v", err)
```
tombuildsstuff: can we make this `Get DNS Zone` to be more descriptive with the error?

azurerm/resource_arm_dns_zone.go
sebastus: removed eTag

azurerm/resource_arm_dns_zone.go
sebastus: done.

azurerm/resource_arm_dns_zone_test.go
sebastus: thanks

azurerm/resource_arm_dns_zone_test.go
sebastus: the pattern we follow is to first check error, then statusCode, throughout the provider. If this is wrong, we should take it up with the team. I added an error return.

sebastus: when I added the error return, it failed 100% of the time. which I guess you could see coming since you suggested checking the 404 first. I really think we need to take this up with the team. I checked a bit and this is the pattern we follow in that routine.

azurerm/resource_arm_dns_zone.go
sebastus: yes, metadata == tags.

azurerm/resource_arm_dns_zone_test.go
tombuildsstuff: indeed - in some other resources we check the 404 inside the error condition - but as long as we do it before returning the error we should be fine :)

> I really think we need to take this up with the team. I checked a bit and this is the pattern we follow in that routine.

I might be mis-reading this - but with regards to the 404 I'd consider the SDK returning an error for a 404 response to be the expected behaviour, given it's not a successful operation, or were you raising something else?

PR #189, opened 2017-07-21
resource_arm_sql_server: switch dependency from riviera to azure-sdk-for-go

sebastus:


azurerm/resource_arm_sql_server.go
```diff
@@ -60,101 +59,111 @@ func resourceArmSqlServer() *schema.Resource {
...
+		serverVersion = sql.OneTwoFullStopZero
+	}
+	if version == string(sql.TwoFullStopZero) {
+		serverVersion = sql.TwoFullStopZero
 	}
-	if !createResponse.IsSuccessful() {
-		return fmt.Errorf("Error creating SQL Server: %s", createResponse.Error)
+	if serverVersion == "" {
+		return fmt.Errorf("Invalid server version provided. It must be one of 12.0 or 2.0, passed as string: %s", version)
+	}
```
tombuildsstuff: [we can remove this in favour of adding validation here](https://github.com/terraform-providers/terraform-provider-azurerm/pull/189/files#diff-73132f55101d4243316253ba03f30742R36)

i.e.

```
ValidateFunc: validation.StringInSlice([]string {
  string(sql.OneTwoFullStopZero),
  string(sql.TwoFullStopZero),
}, true)
```

azurerm/resource_arm_sql_server.go
```diff
@@ -60,101 +59,111 @@ func resourceArmSqlServer() *schema.Resource {
...
-	if err != nil {
-		return fmt.Errorf("Error creating SQL Server: %s", err)
+	version := d.Get("version").(string)
+	var serverVersion sql.ServerVersion
+	if version == string(sql.OneTwoFullStopZero) {
+		serverVersion = sql.OneTwoFullStopZero
+	}
+	if version == string(sql.TwoFullStopZero) {
+		serverVersion = sql.TwoFullStopZero
 	}
```
tombuildsstuff: we can replace all of this with:
```
version := sql.ServerVersion(version)
```

tombuildsstuff: (and we might as well move this cast into the properties assignment below tbh)

azurerm/resource_arm_sql_server.go
```diff
@@ -60,101 +59,111 @@ func resourceArmSqlServer() *schema.Resource {
...
+		Name:             &name,
+		ServerProperties: &props,
+		Tags:             metadata,
+		Location:         &location,
 	}
 
-	readResponse, err := readRequest.Execute()
+	//last parameter is set to empty to allow updates to records after creation
+	// (per SDK, set it to '*' to prevent updates, all other values are ignored)
+	result, err := sqlServersClient.CreateOrUpdate(resGroup, name, parameters)
```
tombuildsstuff: can we remove these comments as they're not valid in this context? :)

azurerm/resource_arm_sql_server.go
```diff
@@ -60,101 +59,111 @@ func resourceArmSqlServer() *schema.Resource {
...
-	d.Set("fully_qualified_domain_name", resp.FullyQualifiedDomainName)
-	d.Set("administrator_login", resp.AdministratorLogin)
-	d.Set("version", resp.Version)
+	serverVersion := serverProperties.Version
 
-	flattenAndSetTags(d, resp.Tags)
+	d.Set("name", name)
+	d.Set("resource_group_name", resGroup)
+	d.Set("version", string(serverVersion))
+	d.Set("location", *result.Location)
```
tombuildsstuff: can we update `location` to use the `azureRMNormalizeLocation(*resp.Location)` method? this helps ensure it's stored consistently in the statefile

azurerm/resource_arm_sql_server.go
```diff
@@ -60,101 +59,111 @@ func resourceArmSqlServer() *schema.Resource {
...
 	}
-	if !deleteResponse.IsSuccessful() {
-		return fmt.Errorf("Error deleting SQL Server: %s", deleteResponse.Error)
+
+	resGroup := id.ResourceGroup
+	name := id.Path["servers"]
+
+	result, error := sqlServersClient.Delete(resGroup, name)
+	if result.Response.StatusCode != http.StatusOK {
+		return fmt.Errorf("Error deleting SQL Server %s: %s", name, error)
```
tombuildsstuff: can we update the second argument to be `%+v`?

azurerm/resource_arm_sql_server_test.go
```diff
@@ -62,48 +62,53 @@ func TestAccAzureRMSqlServer_withTags(t *testing.T) {
...
-		readRequest.Command = &sql.GetServer{}
+		sqlServerName := rs.Primary.Attributes["name"]
+		resourceGroup := rs.Primary.Attributes["resource_group_name"]
+
+		resp, err := conn.Get(resourceGroup, sqlServerName)
 
-		readResponse, err := readRequest.Execute()
 		if err != nil {
-			return fmt.Errorf("Bad: GetServer: %s", err)
+			return nil
```
tombuildsstuff: shouldn't we be returning this error here, and checking to see if it's a 404 above?

tombuildsstuff (changes requested):
Hey @sebastus 

Thanks for this PR - I've reviewed and left some comments in-line - but this is off to a good start :)

Thanks!

azurerm/resource_arm_sql_server.go
sebastus: done.

azurerm/resource_arm_sql_server.go
sebastus: done.

azurerm/resource_arm_sql_server.go
sebastus: done.

azurerm/resource_arm_sql_server.go
sebastus: done.

azurerm/resource_arm_sql_server.go
sebastus: done.

azurerm/resource_arm_sql_server_test.go
sebastus: we should take this up with the team. this pattern is throughout AzureRM

tombuildsstuff: I might be mis-reading your point - but I believe returning an error from the SDK this is the correct use-case, given it's an unsuccessful (i.e. non-200) state?

azurerm/resource_arm_sql_server_test.go
```diff
@@ -99,14 +100,16 @@ func testCheckAzureRMSqlServerDestroy(s *terraform.State) error {
...
+		if autorestErr, ok := err.(autorest.DetailedError); ok {
+			switch autorestErr.StatusCode {
+			case http.StatusNotFound:
+				return nil
+			default:
+				return err
+			}
+		} else {
+			return fmt.Errorf("Bad type assertion for error returned from sqlServersClient.")
 		}
```
tombuildsstuff: As far as I know there's no guarantees given on Error types returned by the SDK - so I think this might be safer + simpler to instead read the status code from the Response object?  [Here's an example of a recently refactored test](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_dns_zone_test.go#L102) which shows what I mean - what do you think?

(an aside: I think this would read cleaner as an if statement anyway, given we're only concerned with a 404 here?)

sebastus:
@tombuildsstuff  Made the changes you requested.

azurerm/resource_arm_sql_server.go
```diff
@@ -60,101 +65,101 @@ func resourceArmSqlServer() *schema.Resource {
...
-		Location:                   d.Get("location").(string),
-		ResourceGroupName:          d.Get("resource_group_name").(string),
-		AdministratorLogin:         azure.String(d.Get("administrator_login").(string)),
-		AdministratorLoginPassword: azure.String(d.Get("administrator_login_password").(string)),
-		Version:                    azure.String(d.Get("version").(string)),
-		Tags:                       *expandedTags,
+	metadata := expandTags(tags)
+
+	parameters := sql.Server{
+		Location: &location,
```
tombuildsstuff: @sebastus I was doing some edge-case testing on this - we can't submit the FQDN since it's only returned as a Computed property (as such I've removed this)

azurerm/resource_arm_sql_server.go
```diff
@@ -60,101 +65,101 @@ func resourceArmSqlServer() *schema.Resource {
...
 	if err != nil {
-		return fmt.Errorf("Error creating SQL Server: %s", err)
-	}
-	if !createResponse.IsSuccessful() {
-		return fmt.Errorf("Error creating SQL Server: %s", createResponse.Error)
-	}
+		// if the name is in-use, Azure returns a 409 "Unknown Service Error" which is a bad UX
+		if responseWasConflict(response.Response) {
+			return fmt.Errorf("SQL Server names need to be globally unique and '%s' is already in use.", name)
+		}
```
tombuildsstuff: @sebastus heads up that doing some edge-case testing I realised Azure returns a "409 Unknown Service Error" which isn't such a great UX, so I've added this to be more descriptive (appears the name needs to be globally unique as it's used in the FQDN)

tombuildsstuff (approved):
Hey @sebastus 

Thanks for pushing those updates - doing some testing on this I'd noticed some minor things/edge-cases and have pushed a couple of commits to address those.

The tests pass and this LGTM:

```
$ TF_ACC=1 envchain azurerm go test ./azurerm -v -timeout 120m -run TestAccAzureRMSqlServer
=== RUN   TestAccAzureRMSqlServer_importBasic
--- PASS: TestAccAzureRMSqlServer_importBasic (126.10s)
=== RUN   TestAccAzureRMSqlServer_basic
--- PASS: TestAccAzureRMSqlServer_basic (104.16s)
=== RUN   TestAccAzureRMSqlServer_withTags
--- PASS: TestAccAzureRMSqlServer_withTags (130.70s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	360.980s
```

Thanks for this!

PR #178, opened 2017-07-20
resource groups - switch dependency from riviera to azure-sdk-for-go

sebastus:

PR #190, opened 2017-07-21
resource_arm_sql_firewall_rule: switch dependency from riviera to azure-sdk-for-go

sebastus:


azurerm/resource_arm_sql_firewall_rule.go
```diff
@@ -51,96 +50,88 @@ func resourceArmSqlFirewallRule() *schema.Resource {
...
-	}
-	if !createResponse.IsSuccessful() {
-		return fmt.Errorf("Error creating SQL Server Firewall Rule: %s", createResponse.Error)
+	name := d.Get("name").(string)
+	resGroup := d.Get("resource_group_name").(string)
+	serverName := d.Get("server_name").(string)
+	startIpAddress := d.Get("start_ip_address").(string)
+	endIpAddress := d.Get("end_ip_address").(string)
+
+	props := sql.FirewallRuleProperties{
```
tombuildsstuff: we might as well combine this into the `parameters` block below?

azurerm/resource_arm_sql_firewall_rule.go
```diff
@@ -51,96 +50,88 @@ func resourceArmSqlFirewallRule() *schema.Resource {
...
-	readRequest.Command = &sql.GetFirewallRule{}
+	resGroup := id.ResourceGroup
+	name := id.Path["firewallRules"]
+	serverName := id.Path["servers"]
 
-	readResponse, err := readRequest.Execute()
+	result, err := sqlFirewallRulesClient.Get(resGroup, serverName, name)
 	if err != nil {
-		return fmt.Errorf("Error reading SQL Server Firewall Rule: %s", err)
+		return fmt.Errorf("Error reading DNS PTR record %s: %v", name, err)
```
tombuildsstuff: this should be `SQL Server Firewall Rule`

azurerm/resource_arm_sql_firewall_rule.go
```diff
@@ -51,96 +50,88 @@ func resourceArmSqlFirewallRule() *schema.Resource {
...
+	d.Set("name", name)
 	d.Set("resource_group_name", resGroup)
-	d.Set("location", azureRMNormalizeLocation(*resp.Location))
-	d.Set("name", resp.Name)
-	d.Set("server_name", id.Path["servers"])
-	d.Set("start_ip_address", resp.StartIPAddress)
-	d.Set("end_ip_address", resp.EndIPAddress)
+	d.Set("server_name", serverName)
+	d.Set("start_ip_address", *props.StartIPAddress)
+	d.Set("end_ip_address", *props.EndIPAddress)
```
tombuildsstuff: can we add a check to ensure props isn't `nil` around these properties (for invalid Swagger/to cover a bad api responses) i.e.

```
if props := result.FirewallRuleProperties; props != nil {
  d.Set("start_ip_address", props.StartIPAddress)
 ...
}
```

(as mentioned in the other PR, we shouldn't need the `*` here)

azurerm/resource_arm_sql_firewall_rule.go
```diff
@@ -51,96 +50,88 @@ func resourceArmSqlFirewallRule() *schema.Resource {
...
-	if !deleteResponse.IsSuccessful() {
-		return fmt.Errorf("Error deleting SQL Server Firewall Rule: %s", deleteResponse.Error)
+
+	resGroup := id.ResourceGroup
+	name := id.Path["firewallRules"]
+	serverName := id.Path["servers"]
+
+	result, error := sqlFirewallRulesClient.Delete(resGroup, serverName, name)
+	if result.Response.StatusCode != http.StatusOK {
+		return fmt.Errorf("Error deleting sql firewall rule %s: %s", name, error)
```
tombuildsstuff: this second argument should be `%+v`

azurerm/resource_arm_sql_firewall_rule_test.go
```diff
@@ -43,48 +43,55 @@ func TestAccAzureRMSqlFirewallRule_basic(t *testing.T) {
...
+		name := rs.Primary.Attributes["name"]
+		serverName := rs.Primary.Attributes["server_name"]
+		resourceGroup := rs.Primary.Attributes["resource_group_name"]
+
+		resp, err := conn.Get(resourceGroup, serverName, name)
 
-		readResponse, err := readRequest.Execute()
 		if err != nil {
-			return fmt.Errorf("Bad: GetFirewallRule: %s", err)
+			return nil
```
tombuildsstuff: we should be returning the error here, but checking for a not found result above it?

azurerm/resource_arm_sql_firewall_rule.go
```diff
@@ -51,96 +50,88 @@ func resourceArmSqlFirewallRule() *schema.Resource {
...
-		ServerName:        d.Get("server_name").(string),
+	parameters := sql.FirewallRule{
+		Name: &name,
+		FirewallRuleProperties: &props,
 	}
 
-	readResponse, err := readRequest.Execute()
+	//last parameter is set to empty to allow updates to records after creation
+	// (per SDK, set it to '*' to prevent updates, all other values are ignored)
+	result, err := sqlFirewallRulesClient.CreateOrUpdate(resGroup, serverName, name, parameters)
```
sebastus: done

azurerm/resource_arm_sql_firewall_rule.go
sebastus: done

azurerm/resource_arm_sql_firewall_rule.go
sebastus: done

azurerm/resource_arm_sql_firewall_rule.go
sebastus: done

azurerm/resource_arm_sql_firewall_rule_test.go
sebastus: we should talk to the team about this - this approach is used throughout AzureRM

azurerm/resource_arm_sql_firewall_rule.go
sebastus: did something - is it what you were thinking? If not, please give an example.

azurerm/resource_arm_sql_firewall_rule_test.go
tombuildsstuff: (I've commented in the other PR about this)

azurerm/resource_arm_sql_firewall_rule.go
```diff
@@ -51,96 +50,86 @@ func resourceArmSqlFirewallRule() *schema.Resource {
...
 
-	readRequest := rivieraClient.NewRequest()
-	readRequest.Command = &sql.GetFirewallRule{
-		Name:              d.Get("name").(string),
-		ResourceGroupName: d.Get("resource_group_name").(string),
-		ServerName:        d.Get("server_name").(string),
+	props := sql.FirewallRuleProperties{
+		StartIPAddress: &startIpAddress,
+		EndIPAddress:   &endIpAddress,
+	}
```
tombuildsstuff: _minor_ can we combine this within the `parameters` object below?

azurerm/resource_arm_sql_firewall_rule_test.go
tombuildsstuff: (I've commented on the other PR about this) - but I think this should become:

```
if resp.Response.StatusCode != http.StatusNotFound {
    return fmt.Errorf("SQL Firewall Rule still exists:\n%#v", resp.FirewallRuleProperties)
}

return err
```

azurerm/resource_arm_sql_firewall_rule.go
sebastus: can you give me an example of what you mean? I did combine the lines according to this suggestion, but then vet had a problem with what I did.

tombuildsstuff: sorry, missed this - [here's an example from the SQL Server Resource](https://github.com/terraform-providers/terraform-provider-azurerm/blob/2c73509a1b276a4439c9da5d722ffd59bb011a9e/azurerm/resource_arm_sql_server.go#L84)

azurerm/resource_arm_sql_firewall_rule.go
```diff
@@ -51,96 +50,86 @@ func resourceArmSqlFirewallRule() *schema.Resource {
...
 
-	readResponse, err := readRequest.Execute()
+	result, err := sqlFirewallRulesClient.Get(resGroup, serverName, name)
 	if err != nil {
-		return fmt.Errorf("Error reading SQL Server Firewall Rule: %s", err)
+		return fmt.Errorf("Error reading SQL firewall rule %s: %v", name, err)
 	}
-	if !readResponse.IsSuccessful() {
-		log.Printf("[INFO] Error reading SQL Server Firewall Rule %q - removing from state", d.Id())
+	if result.Response.StatusCode == http.StatusNotFound {
```
tombuildsstuff: can we move this check inside the `if err != nil` statement? in addition it's better to use [this wrapper](https://github.com/terraform-providers/terraform-provider-azurerm/blob/2c73509a1b276a4439c9da5d722ffd59bb011a9e/azurerm/resource_arm_sql_server.go#L123) because if response.Response is nil, this'll currently crash

azurerm/resource_arm_sql_firewall_rule.go
```diff
@@ -51,96 +50,86 @@ func resourceArmSqlFirewallRule() *schema.Resource {
...
+		return err
 	}
-	if !deleteResponse.IsSuccessful() {
-		return fmt.Errorf("Error deleting SQL Server Firewall Rule: %s", deleteResponse.Error)
+
+	resGroup := id.ResourceGroup
+	name := id.Path["firewallRules"]
+	serverName := id.Path["servers"]
+
+	result, error := sqlFirewallRulesClient.Delete(resGroup, serverName, name)
```
tombuildsstuff: can we switch `error` -> `err` here?

azurerm/resource_arm_sql_firewall_rule.go
```diff
@@ -51,96 +50,86 @@ func resourceArmSqlFirewallRule() *schema.Resource {
...
+	if result.Response.StatusCode != http.StatusOK {
+		return fmt.Errorf("Error deleting sql firewall rule %s: %+v", name, error)
```
tombuildsstuff: given the swagger status codes can change over time (200 -> 201 etc) - can we change this to check for an error and within that handle the not found case using the wrapper method? [Here's an example from the SQL Server resource](https://github.com/terraform-providers/terraform-provider-azurerm/blob/2c73509a1b276a4439c9da5d722ffd59bb011a9e/azurerm/resource_arm_sql_server.go#L157)

tombuildsstuff:
Hey @sebastus

As mentioned in #191, I've been working on a similar task which requires the Riviera SDK to be removed - as such I've rolled your PR's (#191 #190 and #178) into a new PR - which also removes the bulk of the Riviera SDK; I hope you don't mind (and we really appreciate your efforts here) - and this is visible in #289.

As such I'm going to close this PR for the moment in favour of #289.

Thanks!

PR #191, opened 2017-07-22
resource_arm_sql_database: switch dependency from riviera to azure-sdk-for-go

sebastus:


azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
 
-	if v, ok := d.GetOk("edition"); ok {
-		command.Edition = azure.String(v.(string))
-	}
+	maxSizeBytes := d.Get("max_size_bytes").(string)
 
-	if v, ok := d.GetOk("collation"); ok {
-		command.Collation = azure.String(v.(string))
-	}
+	edition := sql.DatabaseEdition(d.Get("edition").(string))
```
tombuildsstuff: given this field is optional, I think we want to optionally assign it to `props` below? i.e.

```
if v, ok := d.GetOk("edition"); ok {
  props.Edition = sql.DatabaseEdition(v.(string))
}
```

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+	requestedServiceObjectiveIDString := d.Get("requested_service_objective_id").(string)
+	var requestedServiceObjectiveID uuid.UUID
 
-	if v, ok := d.GetOk("source_database_deletion_date"); ok {
-		command.SourceDatabaseDeletionDate = azure.String(v.(string))
-	}
+	requestedServiceObjectiveNameString := d.Get("requested_service_objective_name").(string)
+	var requestedServiceObjectiveName sql.ServiceObjectiveName
+
+	if requestedServiceObjectiveIDString != "" && requestedServiceObjectiveNameString == "" {
```
tombuildsstuff: we can use the `conflicts_with` option in the Schema to handle this for us automatically, i.e.:

```
"requested_service_objective_id": {
    Type:     schema.TypeString,
    Optional: true,
    Computed: true,
    ConflictsWith: []string{ "requested_service_objective_name" },
},
 
"requested_service_objective_name": {
    Type:     schema.TypeString,
    Optional: true,
    Computed: true,
    ConflictsWith: []string{ "requested_service_objective_id" },
},
```

and then just have if statements for each field, and assign the value directly to props as needed, rather than checking the state of both:
```
if v, ok := d.Get("requested_service_objective_id").(string); ok {
  rsoid, err := uuid.FromString(v.(string))
   if err != nil {
     return err
  }
  props.RequestedServiceObjectiveID  = &rsoid
}

if v, ok := d.Get("requested_service_objective_name").(string); ok {
  props.RequestedServiceObjectiveID  = sql.ServiceObjectiveName(v.(string))
}
```

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+	if requestedServiceObjectiveName != "" {
+		props.RequestedServiceObjectiveName = requestedServiceObjectiveName
+	} else {
+		props.RequestedServiceObjectiveID = &requestedServiceObjectiveID
+	}
 
-	createResponse, err := createRequest.Execute()
-	if err != nil {
-		return fmt.Errorf("Error creating SQL Database: %s", err)
+	if requestedServiceObjectiveNameString == "ElasticPool" {
```
tombuildsstuff: is there an enum/constant we can use for this?

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+		props.ElasticPoolName = &elasticPoolName
 	}
-	if !createResponse.IsSuccessful() {
-		return fmt.Errorf("Error creating SQL Database: %s", createResponse.Error)
+
+	parameters := sql.Database{
+		Name:               &databaseName,
```
tombuildsstuff: can we remove this, as it's specified in the URL generated from the CreateOrUpdate method below

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
-	readRequest := rivieraClient.NewRequest()
-	readRequest.Command = &sql.GetDatabase{
-		Name:              d.Get("name").(string),
-		ResourceGroupName: d.Get("resource_group_name").(string),
-		ServerName:        d.Get("server_name").(string),
+	resultChan, errChan := sqlDatabasesClient.CreateOrUpdate(resGroup, serverName, databaseName, parameters, make(chan struct{}))
+	resultServer := <-resultChan
+	errServer := <-errChan
+	if errServer != nil {
+		return errServer
```
tombuildsstuff: I think we should be able to reuse `err` here rather than assigning a new error object?

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
-	readResponse, err := readRequest.Execute()
-	if err != nil {
-		return fmt.Errorf("Error reading SQL Database: %s", err)
+	if resultServer.StatusCode != http.StatusOK {
+		return fmt.Errorf("Cannot create Sql database %s (resource group %s) ID", databaseName, resGroup)
 	}
-	if !readResponse.IsSuccessful() {
-		return fmt.Errorf("Error reading SQL Database: %s", readResponse.Error)
+
+	resultDb, errDb := sqlDatabasesClient.Get(resGroup, serverName, databaseName, "")
```
tombuildsstuff: I think we should be able to reuse `err` here rather than assigning a new error object?

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+	resultServer := <-resultChan
+	errServer := <-errChan
+	if errServer != nil {
+		return errServer
 	}
 
-	readResponse, err := readRequest.Execute()
-	if err != nil {
-		return fmt.Errorf("Error reading SQL Database: %s", err)
+	if resultServer.StatusCode != http.StatusOK {
```
tombuildsstuff: I think we can remove this and set `resultChan` above to be a `_` as any errors will be returned as an error object?

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+	databaseName := id.Path["databases"]
+
+	result, err := sqlDatabasesClient.Get(resGroup, serverName, databaseName, "")
 	if err != nil {
-		return fmt.Errorf("Error reading SQL Database: %s", err)
+		return fmt.Errorf("Error reading SQL database %s: %v", databaseName, err)
 	}
-	if !readResponse.IsSuccessful() {
-		log.Printf("[INFO] Error reading SQL Database %q - removing from state", d.Id())
+	if result.Response.StatusCode == http.StatusNotFound {
```
tombuildsstuff: can we move this inside of the `if err != nil` statement above and use the wrapper function, [as shown here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/1dbc1e7fde08c1047a2a6df4c1c33e8e142c6901/azurerm/resource_arm_sql_server.go#L122) as this handles the response being nil

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
-	resp := readResponse.Parsed.(*sql.GetDatabaseResponse)
+	databaseProperties := *result.DatabaseProperties
 
-	d.Set("name", resp.Name)
-	d.Set("creation_date", resp.CreationDate)
-	d.Set("default_secondary_location", resp.DefaultSecondaryLocation)
-	d.Set("elastic_pool_name", resp.ElasticPoolName)
+	d.Set("name", databaseName)
+	d.Set("resource_group_name", resGroup)
+	d.Set("location", *result.Location)
```
tombuildsstuff: can we change this to be `azureRMNormalizeLocation(*resp.Location)` to store it in the state correctly?

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+	d.Set("server_name", serverName)
+	d.Set("creation_date", *result.CreationDate)
+	d.Set("default_secondary_location", *result.DefaultSecondaryLocation)
```
tombuildsstuff: can we add some error checking for if result is nil here? i.e.

```
if props := result.DatabaseProperties; props != nil {
  d.Set("creation_date", result.CreationDate)
  d.Set("default_secondary_location", result.DefaultSecondaryLocation)
}
```

also FWIW I don't think we need the `*` here as the `d.Set` should take care of that for us?

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
 
-	flattenAndSetTags(d, resp.Tags)
+	if databaseProperties.ElasticPoolName != nil {
```
tombuildsstuff: (same here)

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+		return err
 	}
-	if !deleteResponse.IsSuccessful() {
-		return fmt.Errorf("Error deleting SQL Database: %s", deleteResponse.Error)
+
+	resGroup := id.ResourceGroup
+	serverName := id.Path["servers"]
+	databaseName := id.Path["databases"]
+
+	result, error := sqlDatabasesClient.Delete(resGroup, serverName, databaseName)
```
tombuildsstuff: can we reuse the `err` variable above here?

azurerm/resource_arm_sql_database.go
```diff
@@ -118,128 +119,157 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+	if result.Response.StatusCode != http.StatusOK {
```
tombuildsstuff: can we update this [to check for the error and use the wrapper function as outlined in the SQL Server resource](https://github.com/terraform-providers/terraform-provider-azurerm/blob/1dbc1e7fde08c1047a2a6df4c1c33e8e142c6901/azurerm/resource_arm_sql_server.go#L157)?

azurerm/resource_arm_sql_database.go
```diff
@@ -257,3 +287,20 @@ func validateArmSqlDatabaseEdition(v interface{}, k string) (ws []string, errors
 	}
 	return
 }
+
+func validateArmSqlDatabaseCreateMode(v interface{}, k string) (ws []string, errors []error) {
```
tombuildsstuff: (if we switch to using the `validation.StringInStrings` and the enum's/constants above - I think we can remove this?)

azurerm/resource_arm_sql_database_test.go
```diff
@@ -136,48 +136,55 @@ func TestAccAzureRMSqlDatabase_datawarehouse(t *testing.T) {
...
+		conn := testAccProvider.Meta().(*ArmClient).sqlDatabasesClient
+		resp, err := conn.Get(resourceGroup, serverName, name, "")
 		if err != nil {
-			return fmt.Errorf("Bad: GetDatabase: %s", err)
+			return fmt.Errorf("Bad: Get SQL Database: %v", err)
 		}
-		if !readResponse.IsSuccessful() {
-			return fmt.Errorf("Bad: GetDatabase: %s", readResponse.Error)
+
+		if resp.StatusCode == http.StatusNotFound {
```
tombuildsstuff: can we update this to be within the err statement above, as 404 should be being returned as an error

azurerm/resource_arm_sql_database_test.go
```diff
@@ -136,48 +136,55 @@ func TestAccAzureRMSqlDatabase_datawarehouse(t *testing.T) {
...
+		name := rs.Primary.Attributes["name"]
+		serverName := rs.Primary.Attributes["server_name"]
+		resourceGroup := rs.Primary.Attributes["resource_group_name"]
+
+		resp, err := conn.Get(resourceGroup, serverName, name, "")
 
-		readResponse, err := readRequest.Execute()
 		if err != nil {
-			return fmt.Errorf("Bad: GetDatabase: %s", err)
+			return nil
```
tombuildsstuff: can we update this to return the error, and add a check for `if 404 { return nil } ` - given a 404 is an error type?

azurerm/resource_arm_sql_database_test.go
```diff
@@ -136,48 +136,55 @@ func TestAccAzureRMSqlDatabase_datawarehouse(t *testing.T) {
...
 	}
 }
 
 func testCheckAzureRMSqlDatabaseDestroy(s *terraform.State) error {
-	conn := testAccProvider.Meta().(*ArmClient).rivieraClient
+	conn := testAccProvider.Meta().(*ArmClient).sqlDatabasesClient
 
 	for _, rs := range s.RootModule().Resources {
-		if rs.Type != "azurerm_sql_database" {
+		if rs.Type != "azurerm_sql_database_record" {
```
tombuildsstuff: can we change this back to `azurerm_sql_database`, given that's the resource name?

tombuildsstuff:
Hey @sebastus

I've been working on a similar task which requires the Riviera SDK to be removed - as such I've rolled your PR's (#191 #190 and #178) into a new PR - which also removes the bulk of the Riviera SDK; I hope you don't mind (and we really appreciate your efforts here) - and this is visible in #289.

As such I'm going to close this PR for the moment in favour of #289.

Thanks!

PR #205, opened 2017-08-01
Adding Sweepers to the CDN's

tombuildsstuff:
Adding Sweeper to the CDN Profile resource, given there's a tight usage quota.

Whilst we're currently matching only on `acctest` - there's more to come for other resources, as such I've set it up for that now.


grubernaut:
Tests failing `go fmt` check

Issue #207, opened 2017-08-02
Terraform crashes when Azure returns a azurerm_virtual_machine with empty Network interface properties


pather87:
Hi there,

I am experiencing a crash when using Terraform v0.9.2+
The crash.log shows that a response from the Azure API returns a azurerm_virtual_machine configuration with a network Interface and an empty properties field.

The Bug was introduced with [Commit](https://github.com/hashicorp/terraform/pull/11290/commits/e514dc2da050dd11758552c7497a5b3d7586cd70
)

On Line 659 it is checked if NetworkInterfaceReferenceProperties is nil, which is not when a empty properties field is returned. The following attempt to access the "Primary" key is resulting in a nil pointer dereference.

I'm not sure if it is a Bug at Azure site that an empty properties field is returned. If not, it must be further checked if the properties field is empty before trying to access a key.

### Affected Resource(s)
- azurerm_virtual_machine

### Panic Output
[crash.log](https://gist.github.com/pather87/ce399367e1b8af1baa285de3b2949061)

pather87:
#208

tombuildsstuff:
Fixed via #208

PR #208, opened 2017-08-02
check if nic.NetworkInterfaceReferenceProperties.Primary field exists…

sschne:
… before trying to dereference it #207

### DIFF
--- a/azurerm/resource_arm_virtual_machine.go
+++ b/azurerm/resource_arm_virtual_machine.go
@@ -701,7 +701,7 @@ func resourceArmVirtualMachineRead(d *schema.ResourceData, meta interface{}) err
 
 		if resp.VirtualMachineProperties.NetworkProfile.NetworkInterfaces != nil {
 			for _, nic := range *resp.VirtualMachineProperties.NetworkProfile.NetworkInterfaces {
-				if nic.NetworkInterfaceReferenceProperties != nil && *nic.NetworkInterfaceReferenceProperties.Primary {
+				if nic.NetworkInterfaceReferenceProperties != nil && nic.NetworkInterfaceReferenceProperties.Primary != nil && *nic.NetworkInterfaceReferenceProperties.Primary {
 					d.Set("primary_network_interface_id", nic.ID)
 					break
 				}


tombuildsstuff (commented):
Hey @pather87 

Thanks for opening this PR + issue :)

I've spent a little while trying to reproduce this machine configuration within Azure and I'm having a hard-time doing so (although, I can see this response is valid by the API documentation). Out of interest would it be possible to know how this machine was created? It'd be great if we could get a reproducible configuration for this so we could add an acceptance test to ensure we don't cause this error again - however this otherwise LGTM :)

Thanks!

pather87:
Hey @tombuildsstuff 

I have checked the Debug output of the terraform run on a working version. We have around 30 machines running in Azure and it appears that only 2 specific machines show this behaviour. Our terraform configuration of all these machines is quite the same, so i doubt that it has something to do with that.
What i can see from tfstate history is, that these machines have been setup last year with terraform version 0.7.0, but again also have been other machines, so i also doubt that it is related to a specific terraform version. 

These were one of the first machines that have been setup, so i am uncertain that this can be reproduced. IMO it might be related to a special behaviour of Azure at the time of installation.

Regards

tombuildsstuff:
Hey @pather87 

> Our terraform configuration of all these machines is quite the same, so i doubt that it has something to do with that. [..] These were one of the first machines that have been setup, so i am uncertain that this can be reproduced. IMO it might be related to a special behaviour of Azure at the time of installation.

Got you - so from my understanding Azure appears to have some kind of schema for machines - given those two machines were provisioned a while ago, they're likely to be using an older schema. When I've seen this in the past I believe `redeploying` these machines (either in the Portal or the CLI) - or altering a property of these machines (e.g. attaching an extra data disk) will force an update to the latest schema (and Azure VM Agent).

All that said - this PR LGTM and I'm just running the tests now :)

Thanks!

tombuildsstuff (approved):
Hey @pather87 

Thanks for this change - this LGTM! :)

tombuildsstuff:
Ignoring a subscription-specific test failure - the tests pass: 
screen shot 2017-08-04 at 10 59 04

PR #215, opened 2017-08-05
Making the Test Location an Environment Variable: Part 1

tombuildsstuff:

Need to update the CI server to pass in this new parameter before these will run too

Resources:
- [x] Virtual Machines
- [x] Virtual Machine Extensions
- [x] Virtual Machine Scale Sets
- [x] Virtual Networks
- [x] Virtual Network Peerings

Part of #216


tombuildsstuff:
@radeksimko 

Thanks for reviewing.

> This LGTM, but I have not thoroughly checked the code which was just moved due to high LOC - I trust you that you didn't break anything while moving the code.

I'm fairly happy with it, and the tests pass (ignoring a subscription related failure):

screen shot 2017-08-08 at 13 00 51

> Also - having the experience w/ AWS - doesn't Azure also have limited support for some services in some locations? i.e. are you sure those tests are going to pass in all locations?

Indeed, [however there appear to be Primary regions which support everything within each continent](https://azure.microsoft.com/en-gb/regions/services/). As such, we'll still end up running the majority of the tests in `West US`, `East US` or `West Europe` - however this at least allows us to get the lot running against Azure Stack and Germany and go from there :)

Issue #216, opened 2017-08-05
Location used for Tests needs to be configurable


tombuildsstuff:
So that we're able to run the AzureRM Acceptance Tests against different clouds, we need to make the `Location` field, used to specify where to deploy the resource configurable through some means. This is simplest as an environment variable - however this will need to be added to every resource.

This issue tracks the progress rather than having one giant PR:

Resources:

Final Set:

- [x] Data Sources
- [x] Application Insights
- [x] Availability Sets
- [x] CDN Endpoints
- [x] CDN Profile
- [x] Container Registry
- [x] Container Service
- [x] CosmosDB Account

#229:
- [x] DNS A Record
- [x] DNS AAAA Record
- [x] DNS CNAME Record
- [x] DNS MX Record
- [x] DNS PTR Record
- [x] DNS SRV Record
- [x] DNS TXT Record
- [x] DNS Zone
- [x] EventHub
- [x] EventHub Authorization Rule
- [x] EventHub Consumer Group
- [x] EventHub Namespace
- [x] Express Route Circuit

#227:
- [x] Image
- [x] Key Vault
- [x] Load Balancer
- [x] Load Balancer Backend Address Pool
- [x] Load Balancer NAT Pool
- [x] Load Balancer NAT Rule
- [x] Load Balancer Probe
- [x] Load Balancer Rule
- [x] Local Network Gateway

#225:
- [x] Managed Disk
- [x] Network Interface
- [x] Network Security Group
- [x] Network Security Rule
- [x] Public IP
- [x] Redis Cache
- [x] Resource Group
- [x] Route
- [x] Route Table

#221

- [x] Search Service
- [x] ServiceBus Namespace
- [x] ServiceBus Queue
- [x] ServiceBus Subscription
- [x] ServiceBus Topic
- [x] SQL Database
- [x] SQL ElasticPool
- [x] SQL Firewall Rule
- [x] SQL Server
- [x] Storage Account
- [x] Storage Blob
- [x] Storage Container
- [x] Storage Queue
- [x] Storage Share
- [x] Storage Tables

#217
- [x] Subnets
- [x] Template Deployments
- [x] Traffic Manager Endpoints
- [x] Traffic Manager Profile

#215
- [x] Virtual Machines
- [x] Virtual Machine Extensions
- [x] Virtual Machine Scale Sets
- [x] Virtual Networks
- [x] Virtual Network Peerings

PR #217, opened 2017-08-07
Making the Test Location an Environment Variable: Part 2

tombuildsstuff:

Resources:
- [x] Subnets
- [x] Template Deployments
- [x] Traffic Manager Endpoints
- [x] Traffic Manager Profile

Part of #216


tombuildsstuff:
Tests pass - Traffic Manager:
screen shot 2017-08-08 at 15 22 50

Subnets:

screen shot 2017-08-08 at 15 23 02

PR #221, opened 2017-08-08
Making the Test Location an Environment Variable: Part 3

tombuildsstuff:
Continuing on #216

Resources:

- [x] Search Service
- [x] ServiceBus Namespace
- [x] ServiceBus Queue
- [x] ServiceBus Subscription
- [x] ServiceBus Topic
- [x] SQL Database
- [x] SQL ElasticPool
- [x] SQL Firewall Rule
- [x] SQL Server
- [x] Storage Account
- [x] Storage Blob
- [x] Storage Container
- [x] Storage Queue
- [x] Storage Share
- [x] Storage Tables


tombuildsstuff:
Tests pass:

screen shot 2017-08-09 at 11 55 45

PR #227, opened 2017-08-09
Making the Test Location an Environment Variable: Part 5

tombuildsstuff:
Continuing on #216

Resources:

- [x] Image
- [x] Key Vault
- [x] Load Balancer
- [x] Load Balancer Backend Address Pool
- [x] Load Balancer NAT Pool
- [x] Load Balancer NAT Rule
- [x] Load Balancer Probe
- [x] Load Balancer Rule
- [x] Local Network Gateway

### DIFF
--- a/azurerm/import_arm_image_test.go
+++ b/azurerm/import_arm_image_test.go
@@ -10,35 +10,36 @@ import (
 
 func TestAccAzureRMImage_importStandalone(t *testing.T) {
 	ri := acctest.RandInt()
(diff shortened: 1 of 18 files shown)


tombuildsstuff:
Tests pass:

Image:

screen shot 2017-08-09 at 16 18 24

Key Vault:

screen shot 2017-08-09 at 16 18 29

Load Balancer:
screen shot 2017-08-09 at 16 18 12

PR #229, opened 2017-08-09
Making the Test Location an Environment Variable: Part 6

tombuildsstuff:
Continuing on #216

Resources:

- [x] DNS A Record
- [x] DNS AAAA Record
- [x] DNS CNAME Record
- [x] DNS MX Record
- [x] DNS PTR Record
- [x] DNS SRV Record
- [x] DNS TXT Record
- [x] DNS Zone
- [x] EventHub
- [x] EventHub Authorization Rule
- [x] EventHub Consumer Group
- [x] EventHub Namespace
- [x] Express Route Circuit


tombuildsstuff:
Tests pass.

`D` prefix tests:

screen shot 2017-08-09 at 23 41 48

`E` prefix tests:

screen shot 2017-08-09 at 23 41 07

PR #245, opened 2017-08-15
`azurerm_network_interface` - support for multiple IP Configurations / setting the primary

### LINKED ISSUE #26: Enhancement: Allow more than one IP Configuration on azurerm_network_interface
_This issue was originally opened by @carinadigital as hashicorp/terraform#9874. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Allow more one or more IP configurations to be attached to an azurerm_network_interface.

### Terraform Version
0.7.8

### Affected Resource(s)
azurerm_network_interface

### Important Factoids
Some challenges that I've seen.
1. TODO notice in code for the ip configuration loop https://github.com/hashicorp/terraform/blob/master/builtin/providers/azurerm/resource_arm_network_interface_card.go#L266
2. Schema for top level private_ip_address can only store one value.
https://github.com/hashicorp/terraform/blob/master/builtin/providers/azurerm/resource_arm_network_interface_card.go#L54

### LINKED ISSUE #60: Terraform does not change security group of network interface on Azure
_This issue was originally opened by @vchan2002 as hashicorp/terraform#13474. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
v0.9.2

### Affected Resource(s)
- azurerm_network_interface

### Expected Behavior
The security group of the Azure Network interface should have changed.

### Actual Behavior
The security group of the Azure Network interface did not change.

### Steps to Reproduce
a.  Setup an azure_network_interface and terraform it so that it exists in Azure.
b.  Change the network_security_group_id of the network interface in terraform.
c.  terraform plan -target will show that it is planned to be changed.
d.  terraform apply -target will show that nothing has changed.
e.  Actual resource on Azure did not change.

tombuildsstuff:
- [x] Swapping `ip_configuration` from a Set -> List
- [x] Switching over to using constants within the SDK
- [x] Adding support for setting the Primary `ip_configuration`. Migrated across hashicorp/terraform#12277
- [x] Detecting changes in the `network_security_group_id`. Fixes #60 
- [x] Tests covering multiple `ip_configurations`. Fixes #26 & a deadlock issue
- [x] Refactoring the tests to use Dynamic ID's
- [x] Update Documentation
- [x] Confirm no state migration needed between Master -> this branch

### DIFF
--- a/azurerm/resource_arm_network_interface.go
+++ b/azurerm/resource_arm_network_interface.go
@@ -531,6 +531,21 @@ func expandAzureRmNetworkInterfaceIpConfigurations(d *schema.ResourceData) ([]ne
 		ipConfigs = append(ipConfigs, ipConfig)
 	}
 
+	// if we've got multiple IP Configurations - one must be designated Primary
+	if len(ipConfigs) > 1 {
+		hasPrimary := false
+		for _, config := range ipConfigs {
+			if config.Primary != nil && *config.Primary {
+				hasPrimary = true
+				break
+			}
+		}
+
+		if !hasPrimary {
+			return nil, nil, nil, fmt.Errorf("If multiple `ip_configurations` are specified - one must be designated as `primary`.")
+		}
+	}
+
 	return ipConfigs, &subnetNamesToLock, &virtualNetworkNamesToLock, nil
 }
 
--- a/azurerm/resource_arm_network_interface.go
+++ b/azurerm/resource_arm_network_interface.go
@@ -9,6 +9,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/arm/network"
 	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceArmNetworkInterface() *schema.Resource {
@@ -81,9 +82,12 @@ func resourceArmNetworkInterface() *schema.Resource {
 						},
 
 						"private_ip_address_allocation": {
-							Type:             schema.TypeString,
-							Required:         true,
-							ValidateFunc:     validateNetworkInterfacePrivateIpAddressAllocation,
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(network.Dynamic),
+								string(network.Static),
+							}, true),
 							StateFunc:        ignoreCaseStateFunc,
 							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
 						},
@@ -415,19 +419,6 @@ func resourceArmNetworkInterfaceIpConfigurationHash(v interface{}) int {
 	return hashcode.String(buf.String())
 }
 
-func validateNetworkInterfacePrivateIpAddressAllocation(v interface{}, k string) (ws []string, errors []error) {
-	value := strings.ToLower(v.(string))
-	allocations := map[string]bool{
-		"static":  true,
-		"dynamic": true,
-	}
-
-	if !allocations[value] {
-		errors = append(errors, fmt.Errorf("Network Interface Allocations can only be Static or Dynamic"))
-	}
-	return
-}
-
 func flattenNetworkInterfaceIPConfigurations(ipConfigs *[]network.InterfaceIPConfiguration) []interface{} {
 	result := make([]interface{}, 0, len(*ipConfigs))
 	for _, ipConfig := range *ipConfigs {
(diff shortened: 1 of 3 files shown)


dominik-lekse:
Great work, I have been looking forward for the primary ip_configuration feature.

azurerm/resource_arm_network_interface.go
```diff
@@ -288,13 +294,11 @@ func resourceArmNetworkInterfaceRead(d *schema.ResourceData, meta interface{}) e
...
 	if iface.IPConfigurations != nil {
-		d.Set("ip_configuration", schema.NewSet(resourceArmNetworkInterfaceIpConfigurationHash, flattenNetworkInterfaceIPConfigurations(iface.IPConfigurations)))
+		d.Set("ip_configuration", flattenNetworkInterfaceIPConfigurations(iface.IPConfigurations))
 	}
 
 	if iface.VirtualMachine != nil {
-		if *iface.VirtualMachine.ID != "" {
-			d.Set("virtual_machine_id", *iface.VirtualMachine.ID)
-		}
+		d.Set("virtual_machine_id", *iface.VirtualMachine.ID)
```
radeksimko: I think you can omit the `*` here to prevent potential crashes?

azurerm/resource_arm_network_interface.go
```diff
@@ -356,7 +362,7 @@ func resourceArmNetworkInterfaceDelete(d *schema.ResourceData, meta interface{})
 		defer azureRMUnlockByName(networkSecurityGroupName, networkSecurityGroupResourceName)
 	}
 
-	configs := d.Get("ip_configuration").(*schema.Set).List()
+	configs := d.Get("ip_configuration").([]interface{})
```
grubernaut: could this cast panic?

azurerm/resource_arm_network_interface.go
```diff
@@ -48,19 +47,14 @@ func resourceArmNetworkInterface() *schema.Resource {
...
-
 			"virtual_machine_id": {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
 			},
 
 			"ip_configuration": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
```
radeksimko: Are you sure this doesn't require state migration? It's the most common example for migration, IMO 🤔

azurerm/resource_arm_network_interface.go
radeksimko: Correct me if I'm wrong, but I think this will crash with the old state (with Set) without migration.

grubernaut (approved):
LGTM, minor question

azurerm/resource_arm_network_interface_test.go
```diff
@@ -37,16 +38,133 @@ func TestAccAzureRMNetworkInterface_disappears(t *testing.T) {
...
+		},
+	})
+}
+
+func TestAccAzureRMNetworkInterface_removeNetworkSecurityGroupId(t *testing.T) {
+	resourceName := "azurerm_network_interface.test"
+	rInt := acctest.RandInt()
+	location := testLocation()
+	config := testAccAzureRMNetworkInterface_basicWithNetworkSecurityGroup(rInt, location)
+	updatedConfig := testAccAzureRMNetworkInterface_basic(rInt, location)
```
radeksimko: Nitpick, but I think `config` & `updatedConfig` don't need to be variables for that single purpose a few lines below. 🙂

website/docs/r/network_interface.html.markdown
```diff
@@ -91,11 +92,13 @@ The `ip_configuration` block supports:
 
 * `load_balancer_inbound_nat_rules_ids` - (Optional) List of Load Balancer Inbound Nat Rules IDs involving this NIC
 
+* `primary` - (Optional) Is this the Primary Network Interface? If set to `true` this should be the first `ip_configuration` in the array.
```
radeksimko: > If set to `true` this should be the first `ip_configuration` in the array

This feels a little bit unintuitive for the user - is it because the API requires it? Can't we just sort it for the user prior to sending the request to the API?

radeksimko (changes requested):
I left you some comments/questions there - the most important one is regarding state migration.

azurerm/resource_arm_network_interface.go
tombuildsstuff: I tested with a resource/state created on master -> this branch and apparently not ¯\_(ツ)_/¯

website/docs/r/network_interface.html.markdown
tombuildsstuff: so, the user can submit the second (or later) ip_configuration as `primary` and that'll update but lead to a perpetual diff, as it's returned first in the response; we could _probably_ do something there, but this mirrors the UI

The alternative I considered here was just assuming the first would be the primary, but that may not necessarily be optimal? WDYT?

azurerm/resource_arm_network_interface.go
tombuildsstuff: I may have tested something wrong, but this appeared to be fine when testing the migration from a resource/state created with `master`?

website/docs/r/network_interface.html.markdown
radeksimko: > The alternative I considered here was just assuming the first would be the primary

Hm, I think making assumptions is _probably_ even worse UX. This is at the very least explicit. So if you think it's ok then go ahead - I just thought as a user I'd be little bit annoyed or confused 🙂

radeksimko (approved):
As discussed via Slack - this doesn't require migration. (TIL)

tombuildsstuff:
Tests pass: 
screen shot 2017-08-17 at 13 55 15

PR #253, opened 2017-08-17
Module for Computegroup based on Azure vmss

dcaro:


examples/tf-azurerm-computegroup/main.tf
```diff
@@ -0,0 +1,122 @@
...
+
+resource "azurerm_subnet" "subnet1" {
+  name                 = "acctsub"
+  resource_group_name  = "${azurerm_resource_group.rg.name}"
+  virtual_network_name = "${azurerm_virtual_network.vnet.name}"
+  address_prefix       = "10.0.1.0/24"
+}
+
+resource "azurerm_public_ip" "publicip" {
+  name                         = "test"
```
tombuildsstuff: could I suggest this becomes `${var.resource_group_name}-pip` to make this more obvious?

examples/tf-azurerm-computegroup/main.tf
```diff
@@ -0,0 +1,122 @@
...
+  location                     = "${var.location}"
+  resource_group_name          = "${azurerm_resource_group.rg.name}"
+  public_ip_address_allocation = "static"
+  domain_name_label            = "${azurerm_resource_group.rg.name}"
+  tags                         = "${var.tags}"
+}
+
+resource "azurerm_lb" "loadbalancer" {
+  name                = "loadbalancer"
```
tombuildsstuff: same here, would this make more sense as `${var.resource_group_name}-lb`?

examples/tf-azurerm-computegroup/README.md
```diff
@@ -0,0 +1,70 @@
...
+    location            = "westus"
+    vm_size             = "Standard_A0"
+    admin_username      = "azureuser"
+    admin_password      = "ComplexPassword"
+    ssh_key             = "~/.ssh/demo_key.pub"
+    nb_instance         = 2
+    protocol            = {
+                            ssh = ["22", "Tcp", "50000", "50119"]
+                            ftp = ["21", "Tcp", "51000", "51119"]
+                          }
```
tombuildsstuff: I think it'd make more sense to hard-code some rules (i.e. SSH/FTP or HTTP/HTTPS) for examples, rather than making this generic?

examples/tf-azurerm-computegroup/README.md
```diff
@@ -0,0 +1,70 @@
...
+                            publisher = "Canonical"
+                            offer = "UbuntuServer"
+                            sku = "14.04.2-LTS"
+                            version = "latest"
+                          }
+    tags                = {
+                            environment = "dev"
+                            costcenter  = "it"
+                          }
+}
```
tombuildsstuff: could we run this through `terraform fmt` to format it to match the other examples? i.e.

```
module "computegroup" {
  source              = "./path/to/module"
  resource_group_name = "my-resource-group"
  location            = "westus"
  vm_size             = "Standard_A0"
  admin_username      = "azureuser"
  admin_password      = "ComplexPassword"
  ssh_key             = "~/.ssh/demo_key.pub"
  nb_instance         = 2

  protocol = {
    ssh = ["22", "Tcp", "50000", "50119"]
    ftp = ["21", "Tcp", "51000", "51119"]
  }

  os = {
    id        = ""
    publisher = "Canonical"
    offer     = "UbuntuServer"
    sku       = "14.04.2-LTS"
    version   = "latest"
  }

  tags = {
    environment = "dev"
    costcenter  = "it"
  }
}
```

examples/tf-azurerm-computegroup/README.md
```diff
@@ -0,0 +1,70 @@
...
+- `vmss_name` - Name of the virtual machine scale set
+- `vnet_id` - Id of the Virtual Network deployed as part of the VM scale set
+- `subnet_id` - Id of the Subnet deployed as part of the VM scale set
+
+Authors
+=======
+Originally created by Microsoft
+
+License
+=======
```
tombuildsstuff: I may be wrong, but I don't think these need a separate licence section, since they should be covered by the one in the repo root?

examples/tf-azurerm-computegroup/main.tf
```diff
@@ -0,0 +1,122 @@
...
+  location            = "${var.location}"
+  resource_group_name = "${azurerm_resource_group.rg.name}"
+  tags                = "${var.tags}"
+
+  frontend_ip_configuration {
+    name                 = "PublicIPAddress"
+    public_ip_address_id = "${azurerm_public_ip.publicip.id}"
```
tombuildsstuff: _minor/not a blocker_ for the Terraform resource name (in the context of the PIP and the RG) - you may wish to use something more descriptive, i.e. "azurerm_resource_group.vmss" rather than `rg` and `publicip`, as it can help make the console output clearer :)

examples/tf-azurerm-computegroup/main.tf
```diff
@@ -0,0 +1,122 @@
...
+    tier     = "Standard"
+    capacity = "${var.nb_instance}"
+  }
+
+  storage_profile_image_reference {
+    id        = "${var.os["id"]}"
+    publisher = "${var.os["publisher"]}"
+    offer     = "${var.os["offer"]}"
+    sku       = "${var.os["sku"]}"
+    version   = "${var.os["version"]}"
```
tombuildsstuff: I'd argue these should be separate variables, rather than in the hash - as it makes it possible for Terraform to validate them and prompt for them if they're not specified

examples/tf-azurerm-computegroup/main.tf
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  storage_profile_os_disk {
+    name              = ""
```
tombuildsstuff: from memory, this shouldn't need to be specified with a managed disk, as it'll be given a random name, which will show up as a change in the `terraform plan` after a `terraform apply`?

examples/tf-azurerm-computegroup/main.tf
```diff
@@ -0,0 +1,122 @@
...
+    create_option     = "FromImage"
+    managed_disk_type = "Standard_LRS"
+  }
+
+  storage_profile_data_disk {
+      lun          = 0
+    caching        = "ReadWrite"
+    create_option  = "Empty"
+    disk_size_gb   = 10
+  }
```
tombuildsstuff: I'd argue you could _probably_ remove this - but I'm guessing it's been added to encourage users not to store things on the OS disk?

examples/tf-azurerm-computegroup/variables.tf
```diff
@@ -0,0 +1,59 @@
+variable "resource_group_name" {
+    description = "The name of the resource group in which the resources will be created"
+    default = "vmssrg"
+}
+
+variable "location" { 
+    description = "The location where the resources will be created"
+    default = "West US"
```
tombuildsstuff: I'd argue this shouldn't be defaulted, so that the user has to specify it (for instance, this sample wouldn't work in Germany/China or Azure Stack in the future with this location)

examples/tf-azurerm-computegroup/variables.tf
```diff
@@ -0,0 +1,59 @@
...
+    default = "azureuser"
+ }
+
+variable "admin_password" {
+    default = ""
+}
+
+variable "ssh_key" {
+    description = "Path to the public key to be used for ssh access to the VM"
+    default = "~/.ssh/demo_key.pub"
```
tombuildsstuff: might be worth defaulting this to `~/.ssh/id_rsa.pub` given that's the default ssh key path?

examples/tf-azurerm-computegroup/main.tf
```diff
@@ -0,0 +1,122 @@
...
+
+  os_profile {
+    computer_name_prefix = "vmss"
+    admin_username       = "${var.admin_username}"
+    admin_password       = "${var.admin_password}"
+  }
+
+  os_profile_linux_config {
+    disable_password_authentication = true
```
tombuildsstuff: from memory if `disable_password_authentication` is enabled, you should be able to remove the `admin_password` assignment above (this change was recent)

examples/tf-azurerm-computegroup/variables.tf
```diff
@@ -0,0 +1,59 @@
...
+variable "os" { 
+    type = "map"
+    default = {  
+        id = ""
+        publisher = "Canonical"
+        offer = "UbuntuServer"
+        sku = "14.04.2-LTS"
+        version = "latest"
+    }
+}
```
tombuildsstuff: as mentioned above, I think these variables would make more sense with prefixes, i.e. `vm_os_version` - that way Terraform is able to both provide validation for them, but prompt for them if they're not specified

examples/tf-azurerm-computegroup/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Module Input Variables 
+----------------------
+
+- `resource_group_name` - The name of the resource group in which the resources will be created.
+- `location` - The Azure location where the resources will be created.
+- `vn_size` - The initial size of the virtual machine that will be used in the VM Scale Set.
+- `admin_username` - The name of the administrator to access the machines part of the virtual machine scale set. 
+- `admin_password` - The password of the administrator account. The password must comply with the complexity requirements for Azure virtual machines.
+- `ssh_key` - The path on the local machine of the ssh public key in the case of a Linux deployment.
```
dcaro: I've been thinking if we should default to the existing one or not, I have the answer :)

examples/tf-azurerm-computegroup/README.md
```diff
@@ -0,0 +1,70 @@
...
+- Creates a virtual network and a subnet using the 10.0.0.0/16 and 10.0.1.0/24 address space.
+- Creates a load balancer for the scale set of virtual machines
+- Exposes through NAT one or several ports of the VMs on the load balancer
+
+Module Input Variables 
+----------------------
+
+- `resource_group_name` - The name of the resource group in which the resources will be created.
+- `location` - The Azure location where the resources will be created.
+- `vn_size` - The initial size of the virtual machine that will be used in the VM Scale Set.
```
dcaro: Good point.

examples/tf-azurerm-computegroup/main.tf
dcaro: Great suggestion !

examples/tf-azurerm-computegroup/main.tf
dcaro: Not sure I understand, can you give an example ?

dcaro:
I have included the comments and changed the load balancing piece to leverage the azurerm load balancer module.

dcaro:
Done some updates to have a secure by default module.

tombuildsstuff:
After chatting with @echuvyrov yesterday, we're going to close this in favour of the split out repository

PR #254, opened 2017-08-17
Added NSGs to the basic vNet Module

echuvyrov:
NSGs should be required for the basic vNet module

### DIFF
--- a/examples/tf-azurerm-vnet/outputs.tf
+++ b/examples/tf-azurerm-vnet/outputs.tf
@@ -18,7 +18,10 @@ output "vnet_dns_servers" {
   value = "${azurerm_virtual_network.vnet.dns_servers}"
 }
 
-output "vnet_subnet" {
+output "vnet_subnets" {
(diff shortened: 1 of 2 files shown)


examples/tf-azurerm-vnet/outputs.tf
```diff
@@ -18,7 +18,10 @@ output "vnet_dns_servers" {
   value = "${azurerm_virtual_network.vnet.dns_servers}"
 }
 
-output "vnet_subnet" {
+output "vnet_subnets" {
   value = "${azurerm_virtual_network.vnet.subnet}"
```
tombuildsstuff: I don't think this will work in it's current form, instead I think this wants to return a list of subnet ID's via:

```
value = "${azurerm_subnet.subnet.*.id}"
```

What do you think?

echuvyrov: Changes made and pushed, thanks - this works better!

PR #256, opened 2017-08-18
Azure Load balancer module

dtzar:


tombuildsstuff:
After chatting with @echuvyrov yesterday, we're going to close this in favour of the split out repository containing this example

Issue #224, opened 2017-08-09
[v0.10.0] azurerm_image permanent recreation bug


donestefan:
### Terraform Version
Terraform v0.10.0
terraform-provider-azurerm_v0.1.4_x4

### Affected Resource(s)
azurerm_image

### Terraform Configuration Files
```hcl
resource "azurerm_image" "new-managed-image" {
  name                  = "my-new-managed-image.vhd"
  location              = "${azurerm_resource_group.some-rg.location}"
  resource_group_name   = "${azurerm_resource_group.some-rg.name}"
  os_disk {
    os_type  = "Linux"
    os_state = "Generalized"
    blob_uri = "https://somestoreaccount.blob.core.windows.net/some-storage-container/some-linux.vhd"
  }
}
```

### Expected Behavior
image is created on first "terrafrom apply"
image will be deleted and recreated when "blob_uri" changes and "terraform apply" is run again
image will not be deleted when blob_uri is unchanged and "terraform apply" is run again

### Actual Behavior
image is deleted and recreated in every "terraform apply" run - even if blob_uri is unchanged

### Steps to Reproduce
1. create a config like shown above
2. `terraform apply` (image will be created)
3. `terraform plan` (shows unnecessary recreation)
```hcl
$ terraform plan
-/+ azurerm_image.new-managed-image (new resource required)
....
      os_disk.9999999999.blob_uri:        "" => "https://somestoreaccount.blob.core.windows.net/some-storage-container/some-linux.vhd" (forces new resource)
....
```

donestefan:
Currently I'm using `ignore_changes` to work around the problem.

```
resource "azurerm_image" "new-managed-image" {
.....
  os_disk {
    ...
    blob_uri =
  }
  lifecycle {
    ignore_changes = ["os_disk"]
  }
}
```

Funny enough, the changes to `os_disk` will not be ignored completely afterwards.
Changing the `blob_uri` in the `os_disk` block will still cause the image to be deleted and recreated.
While i find this behaviour quite confusing - it luckily makes everything work just as i would expected it to work. The image is only recreated when blob_uri was changed.

echuvyrov:
hey @donestefan I am trying to recreate the scenario above and not having much luck. I pasted my full config file below the output I am seeing - can you paste a small repro config for me to investigate further?

```
azurerm_image.testdestination: Still creating... (10s elapsed)
azurerm_image.testdestination: Still creating... (20s elapsed)
[...]

PR #259, opened 2017-08-19
[MS] Added default caching value as expected by API - fixing #224

echuvyrov:
The managed disk was recreated on every run if "caching" parameter was missing. This PR should address that (bug documented here https://github.com/terraform-providers/terraform-provider-azurerm/issues/224)

### DIFF
--- a/azurerm/resource_arm_image.go
+++ b/azurerm/resource_arm_image.go
@@ -83,7 +83,7 @@ func resourceArmImage() *schema.Resource {
 						"caching": {
 							Type:     schema.TypeString,
 							Optional: true,
-							Default:  compute.None,
+							Default:  string(compute.None),
 							ValidateFunc: validation.StringInSlice([]string{
 								string(compute.None),
 								string(compute.ReadOnly),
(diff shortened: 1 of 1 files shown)


azurerm/resource_arm_image.go
```diff
@@ -83,7 +83,7 @@ func resourceArmImage() *schema.Resource {
 						"caching": {
 							Type:     schema.TypeString,
 							Optional: true,
-							Computed: true,
+							Default:  compute.None,
```
tombuildsstuff: I think this needs to be `string(compute.None)` or compilation will fail?

tombuildsstuff (commented):
Hey @echuvyrov 

Thanks for this PR - this is a good catch, if we can make the default value `string(compute.None)` to make the compiler happy then this otherwise LGTM :)

Thanks!

echuvyrov:
hey @tombuildsstuff changes made :)

tombuildsstuff (approved):
Hey @echuvyrov 

Thanks for pushing that update - this LGTM :)

tombuildsstuff:
Fixes #224

PR #258, opened 2017-08-19
Upgrading to v10.3 of the Azure SDK for Go

tombuildsstuff:
Upgrading to `v10.3.0-beta` of the Azure SDK for Go so that we can get access to the SDK's for Container Instance and EventGrid ([SDK ChangeLog](https://github.com/Azure/azure-sdk-for-go/releases))

PR #260, opened 2017-08-19
New Resource: `azurerm_eventgrid_topic`

tombuildsstuff:
**Note:** Needs to be merged after #258 - since it's based on top of it

---

Adds support for EventGrid Topics which were announced recently. Given this is a new service it's currently limited to 2 of the Azure Regions (as such, I've mentioned this on the documentation page) - but expanded support will come in time.

I also attempted to add support for EventGrid Subscriptions at the same time, but they need some 🤔 around how's best to implement support for it; given they take a Scope (which can be a Subscription ID, Resource Group ID / Event Hub Namespace ID / EventGrid Topic ID) - which needs some extensive changes to the way Resource ID's are mapped. As such I've reversed that out for the moment.

Tests pass:

```
$ TF_ACC=1 envchain azurerm go test ./azurerm -v -timeout 120m -run TestAccAzureRMEventGridTopic
=== RUN   TestAccAzureRMEventGridTopic_importBasic
--- PASS: TestAccAzureRMEventGridTopic_importBasic (158.39s)
=== RUN   TestAccAzureRMEventGridTopic_importBasicWithTags
--- PASS: TestAccAzureRMEventGridTopic_importBasicWithTags (158.91s)
=== RUN   TestAccAzureRMEventGridTopic_basic
--- PASS: TestAccAzureRMEventGridTopic_basic (173.57s)
=== RUN   TestAccAzureRMEventGridTopic_basicWithTags
--- PASS: TestAccAzureRMEventGridTopic_basicWithTags (162.44s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	653.339s
```


azurerm/config.go
```diff
@@ -293,6 +295,12 @@ func (c *Config) getArmClient() (*ArmClient, error) {
 	img.Sender = autorest.CreateSender(withRequestLogging())
 	client.imageClient = img
 
+	ehtc := eventgrid.NewTopicsClientWithBaseURI(endpoint, c.SubscriptionID)
```
radeksimko: Nit: What does the `ehtc` mean here in this context? Shouldn't it be `egt` (Event Grid Topics) or something like that?

azurerm/provider.go
```diff
@@ -88,13 +88,15 @@ func Provider() terraform.ResourceProvider {
...
 			"azurerm_eventhub":                    resourceArmEventHub(),
 			"azurerm_eventhub_authorization_rule": resourceArmEventHubAuthorizationRule(),
 			"azurerm_eventhub_consumer_group":     resourceArmEventHubConsumerGroup(),
 			"azurerm_eventhub_namespace":          resourceArmEventHubNamespace(),
 			"azurerm_express_route_circuit":       resourceArmExpressRouteCircuit(),
-			"azurerm_image":                       resourceArmImage(),
-			"azurerm_key_vault":                   resourceArmKeyVault(),
+
+			"azurerm_image":     resourceArmImage(),
+			"azurerm_key_vault": resourceArmKeyVault(),
```
radeksimko: Nit: Is there any particular reason why these 2 resources deserve to be separated from others, and not `azurerm_express_route_circuit`?

azurerm/resource_arm_eventgrid_topic.go
```diff
@@ -0,0 +1,154 @@
...
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceArmEventGridTopicCreateUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).eventGridTopicsClient
+
+	log.Printf("[INFO] preparing arguments for AzureRM EventGrid Topic creation.")
```
radeksimko: I think this log message could be a lot more useful if it was moved down to line 71 and also displayed contents of `properties` before passing it to the SDK.

azurerm/resource_arm_eventgrid_topic.go
```diff
@@ -0,0 +1,154 @@
...
+		Tags:            expandTags(tags),
+	}
+
+	_, createErr := client.CreateOrUpdate(resourceGroup, name, properties, make(chan struct{}))
+	err := <-createErr
+	if err != nil {
+		return err
+	}
+
+	read, err := client.Get(resourceGroup, name)
```
radeksimko: Is there any reason we cannot use `name` as ID and avoid making that extra API call? Otherwise this is susceptible to dataloss (and dangling resource) in case the second call fails for any reason.

azurerm/resource_arm_eventgrid_topic.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		return err
+	}
+	resourceGroup := id.ResourceGroup
+	name := id.Path["topics"]
+
+	resp, err := client.Get(resourceGroup, name)
+	if err != nil {
+		if responseWasNotFound(resp.Response) {
+			d.SetId("")
```
radeksimko: Nitpick: We tend to log this as `WARN` to make such cases easier to spot in the debug log.

azurerm/resource_arm_eventgrid_topic.go
```diff
@@ -0,0 +1,154 @@
...
+	d.Set("resource_group_name", resourceGroup)
+	d.Set("location", azureRMNormalizeLocation(*resp.Location))
+	d.Set("name", resp.Name)
+
+	if props := resp.TopicProperties; props != nil {
+		d.Set("endpoint", props.Endpoint)
+	}
+
+	d.Set("primary_access_key", keys.Key1)
+	d.Set("secondary_access_key", keys.Key2)
```
radeksimko: Are those access keys sensitive in any way - i.e. would it cause problem if they were exposed in the plan/apply output? If yes, then they should be also marked as `Sensitive` in the schema.

azurerm/resource_arm_eventgrid_topic.go
```diff
@@ -0,0 +1,154 @@
...
+	deleteResp, deleteErr := client.Delete(resGroup, name, make(chan struct{}))
+	resp := <-deleteResp
+	err = <-deleteErr
+
+	if err != nil {
+		if responseWasNotFound(resp) {
+			return nil
+		}
+
+		return err
```
radeksimko: I think you can leave out this return here as the one below is equivalent given the circumstances.

azurerm/resource_arm_eventgrid_topic_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMEventGridTopicDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMEventGridTopicExists("azurerm_eventgrid_topic.test"),
```
radeksimko: It would be nice to add more checks here, at least basic state checks with `resource.TestCheckResourceAttr` and/or `resource.TestCheckResourceAttrSet`.

radeksimko (commented):
I left you some comments there - the most important one is about ID/name and possibly sensitivity of access keys, other are mostly nitpicks.

azurerm/provider.go
tombuildsstuff: Given a lot of the resources share the same initial character, the resources are _vaguely_ grouped by letter so it was easier to skim-read, but happy to move them up?

azurerm/config.go
tombuildsstuff: fixed.

azurerm/resource_arm_eventgrid_topic.go
tombuildsstuff: Consistency with the Azure API's (and existing resources) - which use ID's in the format:

```
/subscriptions/foo/resourceGroups/bar/provider/Microsoft.Blah/eventGrid/name
```

There's an issue open in the SDK to expose a method of obtaining the Resource ID without making this Get Request; once that's done we can switch this out (but that's a future enhancement).

radeksimko: I see, that's sad 😢

azurerm/resource_arm_eventgrid_topic.go
tombuildsstuff: Fixed. (FWIW these should be visible as 404's from the SDK anyway which tends to be the easiest way to grep for them, given the debug logs are _very_ chatty currently)

azurerm/provider.go
radeksimko: Nah, that's ok - I was just curious what's the logic behind it 🙂

azurerm/resource_arm_eventgrid_topic_test.go
tombuildsstuff: As discussed on Slack, we mostly lean on Import tests here - but I'd added checks for the optional/computed fields

tombuildsstuff:
Updated, here's the result of the updated tests:

```
$ TF_ACC=1 envchain azurerm go test ./azurerm -v -timeout 120m -run TestAccAzureRMEventGridTopic_basic
=== RUN   TestAccAzureRMEventGridTopic_basic
--- PASS: TestAccAzureRMEventGridTopic_basic (101.45s)
=== RUN   TestAccAzureRMEventGridTopic_basicWithTags
--- PASS: TestAccAzureRMEventGridTopic_basicWithTags (96.81s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	198.277s
```

and the import tests:
```
$ TF_ACC=1 envchain azurerm go test ./azurerm -v -timeout 120m -run TestAccAzureRMEventGridTopic_import
=== RUN   TestAccAzureRMEventGridTopic_importBasic
--- PASS: TestAccAzureRMEventGridTopic_importBasic (108.05s)
=== RUN   TestAccAzureRMEventGridTopic_importBasicWithTags
--- PASS: TestAccAzureRMEventGridTopic_importBasicWithTags (113.95s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	222.029s
```

radeksimko (approved):
LGTM :shipit:

PR #263, opened 2017-08-22
CosmosDB: fixing the validation on the name field

tombuildsstuff:
Following on from #262 - this fixes the validation to ensure upper-case characters are rejected in the `plan` phase rather than the `apply` phase.

### DIFF
--- a/azurerm/resource_arm_cosmos_db_account.go
+++ b/azurerm/resource_arm_cosmos_db_account.go
@@ -398,7 +398,7 @@ func resourceAzureRMCosmosDBAccountFailoverPolicyHash(v interface{}) int {
 func validateAzureRmCosmosDBAccountName(v interface{}, k string) (ws []string, errors []error) {
 	value := v.(string)
 
-	r, _ := regexp.Compile("^([a-z0-9\\-]+)$")
+	r, _ := regexp.Compile("^[a-z0-9\\-]+$")
 	if !r.MatchString(value) {
 		errors = append(errors, fmt.Errorf("CosmosDB Account Name can only contain lower-case characters, numbers and the `-` character."))
 	}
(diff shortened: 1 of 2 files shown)


azurerm/resource_arm_cosmos_db_account.go
```diff
@@ -398,7 +398,7 @@ func resourceAzureRMCosmosDBAccountFailoverPolicyHash(v interface{}) int {
 func validateAzureRmCosmosDBAccountName(v interface{}, k string) (ws []string, errors []error) {
 	value := v.(string)
 
-	r, _ := regexp.Compile("[a-z0-9-]")
+	r, _ := regexp.Compile("^([a-z0-9\\-]+)$")
```
radeksimko: I can believe the escaping & quantifier (`+`) is necessary, but are the control characters like brackets & `^` & `$` too?
I mean, it applies to the whole string anyway and you don't need to capture anything via subgroup here?

radeksimko (approved):
The expression does no harm and does what it's supposed to do, so LGTM, but can be less verbose, IMO.

azurerm/resource_arm_cosmos_db_account.go
tombuildsstuff: Running the tests it looks like we can remove the brackets but not the `^`&`$` - updated

tombuildsstuff:
Tests pass:
```
$ acctests azurerm TestAccAzureRMCosmosDBAccountName_validation
=== RUN   TestAccAzureRMCosmosDBAccountName_validation
--- PASS: TestAccAzureRMCosmosDBAccountName_validation (0.00s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	0.017s
```

azurerm/resource_arm_cosmos_db_account.go
radeksimko: Apparently I have still a lot to learn about regular expressions! 😅

PR #285, opened 2017-08-29
New data source: azurerm_subscription

### LINKED ISSUE #284: Data source to get information about a specified Azure subscription
Add a data source to return information (most usefully the subscription's display name) for a specified Azure subscription.
If not subscription ID is specified use the subscription ID of the current Azure Resource Manager provider.

ewbankkit:
New data source to return information about a specific Azure subscription.
If not subscription ID is specified, the subscription ID of the current Azure Resource Manager provider is used.
Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/284.


ewbankkit:
The _resources/subscriptions_ package was added via
```
govendor fetch github.com/Azure/azure-sdk-for-go/arm/resources/subscriptions@v10.3.0-beta
```.

**Comment by ewbankkit** [comment]
Acceptance tests:
```
make testacc TEST=./azurerm TESTARGS='-run=TestAccDataSourceAzureRMSubscription_'
TF_ACC=1 go test ./azurerm -v -run=TestAccDataSourceAzureRMSubscription_ -timeout 120m
=== RUN   TestAccDataSourceAzureRMSubscription_current
--- PASS: TestAccDataSourceAzureRMSubscription_current (9.59s)
=== RUN   TestAccDataSourceAzureRMSubscription_specific
--- PASS: TestAccDataSourceAzureRMSubscription_specific (8.23s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	18.380s
```

File: azurerm/data_source_subscription.go [hunk 1]
```diff
@@ -0,0 +1,47 @@
...
+			"subscription_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"display_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
tombuildsstuff: are there any other fields (e.g. Tenant ID?) that would make sense to be exposed here?

azurerm/data_source_subscription.go
```diff
@@ -0,0 +1,47 @@
...
+	groupClient := client.subscriptionsGroupClient
+
+	subscriptionId := d.Get("subscription_id").(string)
+	if subscriptionId == "" {
+		subscriptionId = client.subscriptionId
+	}
+
+	resp, err := groupClient.Get(subscriptionId)
+	if err != nil {
+		return fmt.Errorf("Error reading subscription: %s", err)
```
tombuildsstuff: given the SDK's currently masking some error messages - can we change this the formatted string to `%+v` to return the full error message?

azurerm/data_source_subscription_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceAzureRMSubscription_specificConfig(os.Getenv("ARM_SUBSCRIPTION_ID")),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(resourceName, "subscription_id"),
+					testCheckAzureRMSubscriptionId(resourceName),
```
tombuildsstuff: I think we can remove this in favour of checking via:

```
resource.TestCheckResourceAttr(resourceName, "subscription_id", "{Subscription ID Value}")
```

tombuildsstuff (approved):
Hey @ewbankkit 

Thanks for this PR - I've taken a look and left a few minor comments but this generally looks good to me :)

Out of interest - would it be possible to give an example of when this Data Source would be used? Presumably this is to access information about multiple subscriptions within the same Tenant?

Thanks!

ewbankkit:
I'll work on the changes you suggested.
My use case is for obtaining the subscription's display name and then using it for naming various other resources but I can certainly see value in returning other information.
The [underlying API](https://docs.microsoft.com/en-us/rest/api/resources/Subscriptions/Get) returns some additional fields.

azurerm/data_source_subscription.go
ewbankkit: I have added the subscription state and possible subscription policies.
I'm not adding the tenant as the [tenant API](https://docs.microsoft.com/en-us/rest/api/resources/tenants) only works for the "current" account.

azurerm/data_source_subscription.go
ewbankkit: Done.

azurerm/data_source_subscription_test.go
ewbankkit: The reason we can't do as suggested is that `{Subscription ID Value}` is obtained from the `ArmClient` for the running test step and this can't be done when the steps are declared (`testAccProvider.Meta().(*ArmClient)` is `nil` at this point).

ewbankkit:
Made changes after review and rebased.

azurerm/data_source_subscription_test.go
tombuildsstuff: true, but we could instead check this against the `ARM_SUBSCRIPTION_ID` environment variable? It's not a blocker by any means and this otherwise LGTM :)

tombuildsstuff (approved):
Hey @ewbankkit 

Thanks for pushing those updates - I've taken another look and this LGTM :)

Tests Pass:
```
$ acctests azurerm TestAccDataSourceAzureRMSubscription
=== RUN   TestAccDataSourceAzureRMSubscription_current
--- PASS: TestAccDataSourceAzureRMSubscription_current (8.29s)
=== RUN   TestAccDataSourceAzureRMSubscription_specific
--- PASS: TestAccDataSourceAzureRMSubscription_specific (7.36s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	15.681s
```

Thanks!

PR #288, opened 2017-08-30
Fix invalid parameter

changeworld:
The Azure CDN Profile SKU can only be Premium_Verizon, Standard_Verizon or Standard_Akamai.

If you try to keep sku = "Standard" and `terraform plan` run, it throws following error.

```
1 error(s) occurred:

* azurerm_cdn_profile.test: CDN Profile SKU can only be Premium_Verizon, Standard_Verizon or Standard_Akamai
```

`http_port` and` https_port` are written as `Optional`, and no error occurs in `terraform plan`, but `terraform apply` run, it throws following error.

```
1 error(s) occurred:

* azurerm_cdn_endpoint.test: 1 error(s) occurred:

* azurerm_cdn_endpoint.test: cdn.EndpointsClient#Create: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="BadRequest" Message="Invalid port \"0\". Port value must be a number between 1 and 65535."
```

As a result of Try and Error, at least this sample works.

### DIFF
--- a/website/docs/r/cdn_endpoint.html.markdown
+++ b/website/docs/r/cdn_endpoint.html.markdown
@@ -22,7 +22,7 @@ resource "azurerm_cdn_profile" "test" {
   name                = "acceptanceTestCdnProfile1"
   location            = "West US"
   resource_group_name = "${azurerm_resource_group.test.name}"
-  sku                 = "Standard"
+  sku                 = "Standard_Verizon"
 }
 
 resource "azurerm_cdn_endpoint" "test" {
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
Hey @changeworld 

Thanks for opening this PR :)

I've taken a look into this and discovered that we could instead just default the values to `80` and `443`. As such I've opened PR #301 which includes the fix you've made to the `azurerm_cdn_profile` SKU but also defaults the ports to `80` and `443` respectively.

As such I hope you don't mind but I'm going to close this PR in favour of that one, but thanks a lot for finding and reporting this issue! :)

Thanks!

PR #296, opened 2017-08-31
Cleanup: Adding sweeper for Resource Groups

stack72:


tombuildsstuff:
Hey @stack72 

Thanks for this PR - apologies for the time it's spent waiting.

I spent yesterday sorting out the subscription so that we should be able to merge this - previously this would have caused all of the acceptance tests to fail due to the sweepers failing due to a few locked resources; but I've isolated the resources which were causing the issues and we should be able to merge this now :)

Thanks!

PR #299, opened 2017-09-01
`azurerm_cosmos_db_account`: allow setting the Kind to MongoDB/GlobalDocumentDB

### LINKED ISSUE #290: Feature Request: extend CosmosDB Account to allow the kind of DB to be specified
Hey, 

It would be great it cosmosdb resource was extend to enable a user to specify which kind of database they would like to use within cosmos. 

This looks to be possible via the REST API: https://docs.microsoft.com/en-us/rest/api/documentdbresourceprovider/DatabaseAccounts/List#definitions_databaseaccountslistresult_databaseaccount_properties

I looked into the [code](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_cosmos_db_account.go) and think this would require a addition field in the schema call **Kind** that can be one of the following values: 
- GlobalDocumentDB
- MongoDB
- {Unknown Value for GraphDB}  
- {Unknown Value for KeyValue}  

Perhaps the default would be GlobalDocumentDB as that's the current behaviour.

tombuildsstuff:
Allow setting the `kind` field to either `MongoDB` or `GlobalDocumentDB` (which it's defaulted too for existing users)

Fixes #290

New test passes - just running the lot:
```
$ acctests azurerm TestAccAzureRMCosmosDBAccount_mongoDB
=== RUN   TestAccAzureRMCosmosDBAccount_mongoDB
--- PASS: TestAccAzureRMCosmosDBAccount_mongoDB (670.91s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm    670.928s
```

### DIFF
--- a/azurerm/resource_arm_cosmos_db_account.go
+++ b/azurerm/resource_arm_cosmos_db_account.go
@@ -53,6 +53,18 @@ func resourceArmCosmosDBAccount() *schema.Resource {
 				}, true),
 			},
 
+			"kind": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Default:  string(cosmosdb.GlobalDocumentDB),
+				ValidateFunc: validation.StringInSlice([]string{
+					string(cosmosdb.GlobalDocumentDB),
+					string(cosmosdb.MongoDB),
+				}, true),
+				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+			},
+
 			"ip_range_filter": {
 				Type:     schema.TypeString,
 				Optional: true,
@@ -152,6 +164,7 @@ func resourceArmCosmosDBAccountCreateUpdate(d *schema.ResourceData, meta interfa
 	name := d.Get("name").(string)
 	location := d.Get("location").(string)
 	resGroup := d.Get("resource_group_name").(string)
+	kind := d.Get("kind").(string)
 	offerType := d.Get("offer_type").(string)
 	ipRangeFilter := d.Get("ip_range_filter").(string)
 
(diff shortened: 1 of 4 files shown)


azurerm/resource_arm_cosmos_db_account.go
```diff
@@ -215,6 +229,7 @@ func resourceArmCosmosDBAccountRead(d *schema.ResourceData, meta interface{}) er
 	d.Set("name", resp.Name)
 	d.Set("location", azureRMNormalizeLocation(*resp.Location))
 	d.Set("resource_group_name", resGroup)
+	d.Set("kind", string(resp.Kind))
```
radeksimko: I would have thought `d.Set` can do the casting, but maybe not... 🤔

azurerm/resource_arm_cosmos_db_account_test.go
```diff
@@ -110,6 +110,27 @@ func TestAccAzureRMCosmosDBAccount_eventualConsistency(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMCosmosDBAccountDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMCosmosDBAccountExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "kind", "MongoDB"),
```
radeksimko: It would be nice to add a similar check to existing tests - just to verify that we expose the default kind.

azurerm/resource_arm_cosmos_db_account.go
```diff
@@ -53,6 +53,18 @@ func resourceArmCosmosDBAccount() *schema.Resource {
 				}, true),
 			},
 
+			"kind": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
```
radeksimko: As this is a ForceNew field, are you sure we don't need a state migration? Technically the old state will have empty `kind` (or no `kind` to be precise) and we're setting default here, so `"" => "global-document-db"` would cause recreation, unless I'm mistaken?

tombuildsstuff: Just verified - seems good since there's a default value and this field was missing previously:

```
$ envchain azurerm terraform plan

azurerm_resource_group.test: Refreshing state... (ID: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/tharvey-dev)
azurerm_cosmosdb_account.test: Refreshing state... (ID: /subscriptions/00000000-0000-0000-0000-...ntDB/databaseAccounts/tharvey-cosmosdb)
No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, Terraform
doesn't need to do anything.
```

azurerm/resource_arm_cosmos_db_account_test.go
tombuildsstuff: Done

azurerm/resource_arm_cosmos_db_account.go
tombuildsstuff: we could, however I think it's worth being explicit that this value is coming from an external type?

tombuildsstuff:
Tests pass: 
screen shot 2017-09-01 at 11 36 04

radeksimko (approved):

PR #304, opened 2017-09-04
New Data Source: `azurerm_public_ips`

hbuckle:
Adds a new data source to retrieve the IDs of unassigned public IP addresses from a resource group. We have a pool of public IP addresses that have been pre-whitelisted by third parties so this will allow us to pick them up dynamically instead of needing to hardcode the IDs into our configuration.
This is my first time working with go and terraform so any feedback is appreciated!


tombuildsstuff:
Hey @hbuckle 

Thanks for opening this PR :)

Taking a look at this, I think this Data Source might be more useful were it to return a list of Public IP's (and most of their properties) rather than just their ID's. We can apply the same filter logic to this so that it's possible to return only Public IP's which aren't attached to a VM/LB etc (and thus your workflow should still work 😄) - but the advantage here would be that the Data Source would work for multiple use-cases, here's an example of what I mean:

```
data "azurerm_public_ips" "test" {
  resource_group_name = "foo"
  attached = false
}

output "public_ip_address_ids" {
  value = "${data.azurerm_public_ips.*.id}"
}
```

What do you think? :)

Thanks!

hbuckle:
I have started looking at adding more properties, the only issue is that DNS settings are optional for public IPs, and for dynamic IPs they will only have an IP address if they are assigned. What's the right way to handle this, just add empty values to the results?

tombuildsstuff:
Hey @hbuckle 

Sorry for the delayed response here - I'd missed we'd not replied to this!

> I have started looking at adding more properties, the only issue is that DNS settings are optional for public IPs, and for dynamic IPs they will only have an IP address if they are assigned. What's the right way to handle this, just add empty values to the results?

So if there's no value we can either not set the object (in the case of a simple type like a String/Integer) - or set an empty value (in the case of a complex object). For instance, [in the `azurerm_public_ip` resource](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_public_ip.go#L182) we only set the `fqdn` if there's a value - otherwise (given it's a string) where it's not being set, it means that it'll get a default value of an empty string.

Hopefully that helps? :)

Thanks!

hbuckle:
I've updated the data source so that it returns a list of public IPs. Each public IP is represented by a map with the IP's properties.

hbuckle:
Hi - does the updated version fit in with the suggested changes? Is it possible to merge?

tombuildsstuff:
hey @hbuckle - sorry for the delayed review here - I'm taking another look at this now :)

azurerm/data_source_arm_public_ips.go
```diff
@@ -0,0 +1,89 @@
...
+
+	resGroup := d.Get("resource_group_name").(string)
+	minimumCount, minimumCountOk := d.GetOk("minimum_count")
+	attachedOnly := d.Get("attached").(bool)
+	resp, err := publicIPClient.List(resGroup)
+	if err != nil {
+		if resp.StatusCode == http.StatusNotFound {
+			d.SetId("")
+		}
+		return fmt.Errorf("Error making Read request on Azure resource group %s: %s", resGroup, err)
```
tombuildsstuff: could we make these formatting arguments`%q` and `%+v`?

azurerm/data_source_arm_public_ips.go
```diff
@@ -0,0 +1,89 @@
...
+
+func dataSourceArmPublicIPsRead(d *schema.ResourceData, meta interface{}) error {
+	publicIPClient := meta.(*ArmClient).publicIPClient
+
+	resGroup := d.Get("resource_group_name").(string)
+	minimumCount, minimumCountOk := d.GetOk("minimum_count")
+	attachedOnly := d.Get("attached").(bool)
+	resp, err := publicIPClient.List(resGroup)
+	if err != nil {
+		if resp.StatusCode == http.StatusNotFound {
```
tombuildsstuff: can we update this to use the helper method `utils.ResponseWasNotFound(resp)` - which also handles connection drops?

azurerm/data_source_arm_public_ips.go
```diff
@@ -0,0 +1,89 @@
...
+			},
+		},
+	}
+}
+
+func dataSourceArmPublicIPsRead(d *schema.ResourceData, meta interface{}) error {
+	publicIPClient := meta.(*ArmClient).publicIPClient
+
+	resGroup := d.Get("resource_group_name").(string)
+	minimumCount, minimumCountOk := d.GetOk("minimum_count")
```
tombuildsstuff: I think we can remove this field?

azurerm/data_source_arm_public_ips.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		return fmt.Errorf("Error making Read request on Azure resource group %s: %s", resGroup, err)
+	}
+	var filteredIps []network.PublicIPAddress
+	for _, element := range *resp.Value {
+		if (element.IPConfiguration != nil) == attachedOnly {
+			filteredIps = append(filteredIps, element)
+		}
+	}
+	if minimumCountOk && len(filteredIps) < minimumCount.(int) {
```
tombuildsstuff: I think now this is a more generic Data Source - that we can remove this?

azurerm/data_source_arm_public_ips.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, element := range filteredIps {
+		m := make(map[string]string)
+		m["public_ip_address_id"] = *element.ID
+		m["name"] = *element.Name
+		if element.PublicIPAddressPropertiesFormat.DNSSettings != nil {
+			if element.PublicIPAddressPropertiesFormat.DNSSettings.Fqdn != nil && *element.PublicIPAddressPropertiesFormat.DNSSettings.Fqdn != "" {
+				m["fqdn"] = *element.PublicIPAddressPropertiesFormat.DNSSettings.Fqdn
+			} else {
+				m["fqdn"] = ""
+			}
```
tombuildsstuff: I think we can remove the else statement here - since it'll be set to an empty string automatically?

azurerm/data_source_arm_public_ips.go
```diff
@@ -0,0 +1,89 @@
...
+			if element.PublicIPAddressPropertiesFormat.DNSSettings.DomainNameLabel != nil && *element.PublicIPAddressPropertiesFormat.DNSSettings.DomainNameLabel != "" {
+				m["domain_name_label"] = *element.PublicIPAddressPropertiesFormat.DNSSettings.DomainNameLabel
+			} else {
+				m["domain_name_label"] = ""
```
tombuildsstuff: I think we can remove the else statement here - since it'll be set to an empty string automatically?

azurerm/data_source_arm_public_ips.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+	if minimumCountOk && len(filteredIps) < minimumCount.(int) {
+		return fmt.Errorf("Not enough unassigned public IP addresses in resource group %s", resGroup)
+	}
+	var results []map[string]string
+	for _, element := range filteredIps {
+		m := make(map[string]string)
+		m["public_ip_address_id"] = *element.ID
+		m["name"] = *element.Name
```
tombuildsstuff: can we add an `if` check around this, in-case it's nil?

azurerm/data_source_arm_public_ips.go
```diff
@@ -0,0 +1,89 @@
...
+			filteredIps = append(filteredIps, element)
+		}
+	}
+	if minimumCountOk && len(filteredIps) < minimumCount.(int) {
+		return fmt.Errorf("Not enough unassigned public IP addresses in resource group %s", resGroup)
+	}
+	var results []map[string]string
+	for _, element := range filteredIps {
+		m := make(map[string]string)
+		m["public_ip_address_id"] = *element.ID
```
tombuildsstuff: can we add an `if` check around this, in-case it's nil?

website/docs/d/public_ips.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+    public_ip_address_id = "${lookup(data.azurerm_public_ips.test.public_ips[count.index], "public_ip_address_id")}"
+  }
+}
+```
+
+## Argument Reference
+
+* `resource_group_name` - (Required) Specifies the name of the resource group.
+* `attached` - (Required) Whether to return public IPs that are attached or not.
+* `minimum_count` - (Optional) Specifies the minimum number of IP addresses that
```
tombuildsstuff: (as above) can we remove this field?

hbuckle:
Sorry for the delay, finally got around to updating this

hbuckle:
@tombuildsstuff any chance of going over this again?

tombuildsstuff:
@hbuckle sorry, we're still trying to work through the backlog of pending PR reviews; I'll take a look into this shortly

tombuildsstuff:
@hbuckle heads up that I've just rebased this on top of master so it merges cleanly, taking a look now ;0

tombuildsstuff:
@hbuckle I hope you don't mind, but I've pushed some commits to add support for also filtering by the prefix of the `name` and the `allocation_type` for the Public IP Address in order to expand the use-case of this data source

website/docs/d/public_ips.html.markdown
```diff
@@ -0,0 +1,38 @@
...
+data "azurerm_public_ips" "test" {
+  resource_group_name = "pip-test"
+  attached            = false
+}
+```
+
+## Argument Reference
+
+* `resource_group_name` - (Required) Specifies the name of the resource group.
+* `attached` - (Optional) Should we only return IP Addresses which are attached to a device (VM/LB) etc?
```
mbfrahry: I wouldn't frame this as a question and instead have it read like `Returns IP Address that are attached to a device or not attached to a device` or something to that effect.

Issue #287, opened 2017-08-30
Add support for ACI with public container registries


abhijeetgaiha:
Adding support for Azure Container Instance, with support for public container registries.

abhijeetgaiha:
Working on this currently. Support for private registries to come later.

dcaro:
I'm also looking at adding the provider for ACI

abhijeetgaiha:
Still to come:
- Command lines
- Volume Mounts
- Environment Variables
- Private container registries

ruzickap:
Support for Registry Arguments like it's in ```az``` command would be nice :-)

```
Image Registry Arguments
    --registry-login-server       : The container image registry login server.
    --registry-password           : The password to log in container image registry server.
    --registry-username           : The username to log in container image registry server.
```

Thanks for a great job...

abhijeetgaiha:
@ruzickap It's on the way.

jyapurv:
Hi, any news? :(

jincod:
Hi @abhijeetgaiha 

Do you have any updates?

achandmsft:
Hi @abhijeetgaiha let us know if you are working on this?

tombuildsstuff:
👋 

Support for Image Registries in ACI was merged in #1529 which was released in v1.10 of the AzureRM Provider. You can upgrade to this version by updating the version number specified in the Provider block (as shown below) and then running `terraform init -upgrade`

```
provider "azurerm" {
  version = "=1.18.0"
}
```

Thanks!

PR #311, opened 2017-09-07
Implementation of Azure container groups

abhijeetgaiha:
This has a basic implementation for #287 .

The following features are TBD:
- Command lines
- Volume Mounts
- Environment Variables
- Private container registries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 IMPROVEMENTS:
 
+* `azurerm_app_service_plan` - support for Linux App Services [GH-332]
 * `azurerm_postgresql_server` - supporting additional storage sizes [GH-239]
 * `azurerm_public_ip` - verifying the ID is valid before importing [GH-320]
 * `azurerm_sql_server` - verifying the name is valid before creating [GH-323]


azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,294 @@
...
+
+			"resource_group_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"ip_address_type": {
+				Type:             schema.TypeString,
+				Required:         true,
+				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
```
tombuildsstuff: I'm assuming there's a limited set of options here? if so - could we add some validation to this, e.g.
```
  "foo",
  "bar",
}, true)
```

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,294 @@
...
+	return &schema.Resource{
+		Create: resourceArmContainerGroupCreate,
+		Read:   resourceArmContainerGroupRead,
+		Update: resourceArmContainerGroupCreate,
+		Delete: resourceArmContainerGroupDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
```
tombuildsstuff: I think this also needs to be `ForceNew`?

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,294 @@
...
+			},
+
+			"location": locationSchema(),
+
+			"resource_group_name": {
+				Type:     schema.TypeString,
+				Required: true,
```
tombuildsstuff: I think this also needs to be `ForceNew`?

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,294 @@
...
+			"ip_address_type": {
+				Type:             schema.TypeString,
+				Required:         true,
+				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+			},
+
+			"os_type": {
+				Type:             schema.TypeString,
+				Required:         true,
+				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
```
tombuildsstuff: I'm assuming there's a limited set of options here? if so - could we add some validation to this, e.g.
```
  "foo",
  "bar",
}, true)
```

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,294 @@
...
+			IPAddress: &containerinstance.IPAddress{
+				Type:  &IPAddressType,
+				Ports: &containerGroupPorts,
+			},
+			OsType: containerinstance.OperatingSystemTypes(OSType),
+		},
+	}
+
+	_, error := containerGroupsClient.CreateOrUpdate(resGroup, name, containerGroup)
+	if error != nil {
```
tombuildsstuff: _minor_ can we swap `error` out for `err` to match the other resources?

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,294 @@
...
+		return error
+	}
+
+	read, readErr := containerGroupsClient.Get(resGroup, name)
+	if readErr != nil {
+		return readErr
+	}
```
tombuildsstuff: we can reuse the existing `err` object here, rather than creating a new variable here

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,294 @@
...
+
+	d.Set("name", name)
+	d.Set("resource_group_name", resGroup)
+	d.Set("location", azureRMNormalizeLocation(*resp.Location))
+	flattenAndSetTags(d, resp.Tags)
+
+	d.Set("os_type", string(resp.OsType))
+	d.Set("ip_address_type", *resp.IPAddress.Type)
+	d.Set("ip_address", *resp.IPAddress.IP)
```
tombuildsstuff: can we add a check around the use of `resp.IPAddress`? i.e.

```
if address := resp.IPAddress; address != nil {
 d.Set("ip_address_type", address.Type)
 d.Set("ip_address", address.IP)
}

File: azurerm/resource_arm_container_group.go [hunk 8]
```diff
@@ -0,0 +1,294 @@
...
+
+	containers := *resp.Containers
+
+	containerConfigs := make([]interface{}, 0, len(containers))
```
tombuildsstuff: rather than putting this directly in here, can we pull this out into a separate `flattenContainerInstanceContainers` method which returns the `[]interface{}` object to set?

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,294 @@
...
+
+	return nil
+}
+func resourceArmContainerGroupDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient)
+	containterGroupsClient := client.containerGroupsClient
+
+	// container group properties
+	resGroup := d.Get("resource_group_name").(string)
+	name := d.Get("name").(string)
```
tombuildsstuff: rather than using the state - can we parse these from the URL as in the Get method? i.e.

```
if err != nil {
  return err
}

resGroup := id.ResourceGroup
name := id.Path["containerGroups"]
```

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,294 @@
...
+
+	_, error := containterGroupsClient.Delete(resGroup, name)
+
+	if error != nil {
+		return error
```
tombuildsstuff: _minor_ can we rename this to `err` to match the other resources (`error` conflicts with the class name)

azurerm/resource_arm_container_group_test.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAzureRMContainerGroup_basicLinux(t *testing.T) {
+	ri := acctest.RandInt()
+
+	name := fmt.Sprintf("acctestcontainergroup-%d", ri)
+	resourceGroupName := fmt.Sprintf("acctestRG-%d", ri)
+
+	config := testAccAzureRMContainerGroupBasicLinux(name, resourceGroupName, testLocation())
```
tombuildsstuff: can we pass `ri` into this method and format the name and the resource group name in there, as we do in the other test cases? I can't see any usages of the `name` and `resourceGroupName` variables in this function

azurerm/resource_arm_container_group_test.go
```diff
@@ -0,0 +1,173 @@
...
+	})
+}
+
+func TestAccAzureRMContainerGroup_basicWindows(t *testing.T) {
+	ri := acctest.RandInt()
+
+	name := fmt.Sprintf("acctestcontainergroup-%d", ri)
+	resourceGroupName := fmt.Sprintf("acctestRG-%d", ri)
+
+	config := testAccAzureRMContainerGroupBasicWindows(name, resourceGroupName, testLocation())
```
tombuildsstuff: (same here)

azurerm/resource_arm_container_group_test.go
```diff
@@ -0,0 +1,173 @@
...
+	ip_address_type="public"
+	os_type = "windows"
+  
+	container {
+	  name = "%s"
+	  image = "winappimage:latest"
+	  cpu ="2.0"
+	  memory = "3.5"
+	  port = "80"
+#	  protocol = "TCP"
```
tombuildsstuff: is this commented out as it's not returned?

azurerm/resource_arm_container_group_test.go
```diff
@@ -0,0 +1,173 @@
...
+	ip_address_type="public"
+	os_type = "linux"
+  
+	container {
+	  name = "%s"
+	  image = "nginx:latest"
+	  cpu ="1"
+	  memory = "1.0"
+	  port = "80"
+#	  protocol = "TCP"
```
tombuildsstuff: is this commented out as it's not returned?

azurerm/resource_arm_container_group_test.go
```diff
@@ -0,0 +1,173 @@
...
+		conn := testAccProvider.Meta().(*ArmClient).containerGroupsClient
+
+		resp, err := conn.Get(resourceGroup, name)
+		if err != nil {
+			return fmt.Errorf("Bad: Get on containerGroupsClient: %+v", err)
+		}
+
+		if resp.StatusCode == http.StatusNotFound {
+			return fmt.Errorf("Bad: Container Group %q (resource group: %q) does not exist", name, resourceGroup)
+		}
```
tombuildsstuff: can we move this inside the `if err != nil` check above and use the `utils.ResponseWasNotFound` method to handle network failures? i.e.

```
if err != nil {
    return return fmt.Errorf("Bad: Container Group %q (resource group: %q) does not exist", name, resourceGroup)
  }

  return fmt.Errorf("Bad: Get on containerGroupsClient: %+v", err)
}
```

azurerm/resource_arm_container_group_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+		resp, err := conn.Get(resourceGroup, name)
+
+		if err != nil {
+			return nil
+		}
+
+		if resp.StatusCode != http.StatusNotFound {
+			return fmt.Errorf("Container Group still exists:\n%#v", resp)
+		}
```
tombuildsstuff: (as above) can we move this inside the `if err != nil` check?

website/docs/r/container_group.html.markdown
```diff
@@ -0,0 +1,105 @@
...
+    port = "80"
+  }
+
+  tags {
+    environment = "testing"
+  }
+}
+```
+
+## Example Usage (Windows)
```
tombuildsstuff: given the only difference between these two is the OS field - I think we can just use the Linux example (and just rename that to `Example Usage`)?

website/docs/r/container_group.html.markdown
```diff
@@ -0,0 +1,105 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) Specifies the name of the Container Group. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the resource group in which to create the Container Group. Changing this forces a new resource to be created.
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created. Changing this forces a new resource to be created.
+
+* `ip_address_type` - (Required) Specifies the ip address type of the container. `public` is the only acceptable value at this time. Changing this forces a new resource to be created.
```
tombuildsstuff: If this is the only acceptable value at this time, we can make this field `Optional` and `Default` it to `Public` for the moment - so users don't have to specify it (in the schema)

website/docs/r/container_group.html.markdown
```diff
@@ -0,0 +1,105 @@
...
+
+* `memory` - (Required) The required memory of the containers in GB. Changing this forces a new resource to be created.
+
+* `port` - (Optional) A public port for the container. Changing this forces a new resource to be created.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `ip_address` - The IP address allocated to the container group.
```
tombuildsstuff: there's also the ID of the container group as `id`

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,294 @@
...
+
+			"ip_address": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"container": {
+				Type:     schema.TypeList,
+				Required: true,
+				Elem: &schema.Resource{
```
tombuildsstuff: given only a single container's currently supported - can we add `MaxItems: 1` here?

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,294 @@
...
+					Schema: map[string]*schema.Schema{
+
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
```
tombuildsstuff: these are marked as `ForceNew` in the documentation below - if so, can we add `ForceNew: true` to all of these fields in the `container` group?

azurerm/resource_arm_container_group.go
tombuildsstuff: (as below) we should be able to make this `Optional` and `Default: "Public"` given `Public" is the only option at this time, which would mean users don't need to specify it :)

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Actually, more than 1 container is supported for Linux. Updated docs.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Fixed.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done

azurerm/resource_arm_container_group_test.go
abhijeetgaiha: Yes, if it's set, it causes a non-empty apply everytime. I think this is a bug in the API currently.

azurerm/resource_arm_container_group_test.go
abhijeetgaiha: Yes, if it's set, it causes a non-empty apply everytime. I think this is a bug in the API currently.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

website/docs/r/container_group.html.markdown
abhijeetgaiha: Done.

website/docs/r/container_group.html.markdown
abhijeetgaiha: Done.

website/docs/r/container_group.html.markdown
abhijeetgaiha: I've updated the linux example to use two containers.

azurerm/resource_arm_container_group_test.go
abhijeetgaiha: I actually did it this way for other tests (data_source_arm_public_ip_test.go). I'll change it anyway.

azurerm/resource_arm_container_group_test.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group_test.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group_test.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
tombuildsstuff: I might have missed it, but I can't see these changes?

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,317 @@
...
+							ValidateFunc: validation.IntBetween(1, 65535),
+						},
+
+						"protocol": {
+							Type:     schema.TypeString,
+							Optional: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"tcp",
+								"udp",
+							}, true),
```
tombuildsstuff: given the validation function allows for differences in case (the `true`) can we add a `DiffSuppressFunc` to this to ignore the case?

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,317 @@
...
+
+	// container group properties
+	resGroup := d.Get("resource_group_name").(string)
+	name := d.Get("name").(string)
+	location := d.Get("location").(string)
+	OSType := d.Get("os_type").(string)
+	IPAddressType := d.Get("ip_address_type").(string)
+	tags := d.Get("tags").(map[string]interface{})
+
+	containersConfig := d.Get("container").([]interface{})
```
tombuildsstuff: can we pull all of this out into a separate function `expandContainerGroupContainers` which returns the deserialized object?

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,317 @@
...
+				Resources: &containerinstance.ResourceRequirements{
+					Requests: &containerinstance.ResourceRequests{
+						MemoryInGB: &memory,
+						CPU:        &cpu,
+					},
+				},
+			},
+		}
+
+		if port != 0 {
```
tombuildsstuff: given this is an `Optional` field, we should be able to do this instead:

```
if v, ok := data["port"]; ok {
  ...
}
```

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,317 @@
...
+			}
+
+			container.Ports = &[]containerinstance.ContainerPort{containerPort}
+
+			// container group port (port number + protocol)
+			containerGroupPort := containerinstance.Port{
+				Port: &port,
+			}
+
+			if protocol != "" {
```
tombuildsstuff: (same as for port) for the Protocol field

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,317 @@
...
+
+		// 	for _, envVarRaw := range envVars {
+		// 		envVar := containerinstance.EnvironmentVariable{
+		// 			Name:  &envVarRaw.get("name").(string),
+		// 			Value: &envVarRaw.get("value").(string),
+		// 		}
+		// 		envVarsList = append(envVarsList, envVar)
+		// 	}
+		// 	container.EnvironmentVariables = &envVarsList
+		// }
```
tombuildsstuff: can we remove this, since it's not being used?

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	// container group properties
+	resGroup := id.ResourceGroup
+	name := id.Path["containerGroups"]
+
+	_, err = containterGroupsClient.Delete(resGroup, name)
+
+	if err != nil {
+		return err
```
tombuildsstuff: can we check for both a 404 and a dropped connection before throwing an error? we can do this via:

```
resp, err = containterGroupsClient.Delete(resGroup, name)
if err != nil && !utils.ResponseWasNotFound(resp) {
  return nil
}

return nil
```

azurerm/resource_arm_container_group_test.go
```diff
@@ -0,0 +1,172 @@
...
+	return fmt.Sprintf(`
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-%d"
+  location = "%s"
+}
+
+resource "azurerm_container_group" "test" {
+	
+	name = "acctestcontainergroup-%d"
+	location = "%s"
```
tombuildsstuff: rather than threading this through in multiple places, can we update this to `"${azurerm_resource_group.test.location}"`?

azurerm/resource_arm_container_group_test.go
```diff
@@ -0,0 +1,172 @@
...
```
tombuildsstuff: rather than threading this through in multiple places, can we update this to `"${azurerm_resource_group.test.location}"`?

examples/aci-linux-multi/main.tf
```diff
@@ -0,0 +1,30 @@
+resource "azurerm_resource_group" "aci-rg" {
+    name="aci-test"
+    location="west us"
+}
+
+resource "azurerm_container_group" "aci-test" {
+    name = "mynginx"
+    location = "west us"
```
tombuildsstuff: _minor_ since Location is available on the Resource Group resource, could we update this to `"${azurerm_resource_group.aci-rg.location}"` to keep it a little DRY-er?, rather than in each resource? (I'm slowly working through the documentation for each resource to show this)

website/docs/r/container_group.html.markdown
```diff
@@ -0,0 +1,113 @@
...
+```hcl
+resource "azurerm_resource_group" "aci-rg" {
+  name     = "aci-test"
+  location = "west us"
+}
+
+resource "azurerm_container_group" "winapp" {
+  
+  name = "mywinapp"
+  location = "west us"
```
tombuildsstuff: _minor_ can we update this to be `"${azurerm_resource_group.aci-rg.location}"`

website/docs/r/container_group.html.markdown
tombuildsstuff: could we consolidate this into a single example, as they're fundamentally the same (compared to say Container Service in which some fields aren't valid for a non-Kubernetes cluster)?

website/docs/r/container_group.html.markdown
```diff
@@ -0,0 +1,113 @@
...
+```hcl
+resource "azurerm_resource_group" "aci-rg" {
+  name     = "aci-test"
+  location = "west us"
+}
+
+resource "azurerm_container_group" "aci-helloworld" {
+  
+  name = "aci-hw"
+  location = "west us"
```
tombuildsstuff: _minor_ can we update this to be `"${azurerm_resource_group.aci-rg.location}"`

azurerm/resource_arm_container_group.go
abhijeetgaiha: I fixed the docs. This shouldn't force the whole group to be updated.

website/docs/r/container_group.html.markdown
```diff
@@ -0,0 +1,113 @@
...
+
+* `resource_group_name` - (Required) The name of the resource group in which to create the Container Group. Changing this forces a new resource to be created.
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `ip_address_type` - (Optional) Specifies the ip address type of the container. `Public` is the only acceptable value at this time.
+
+* `os_type` - (Required) The OS for the container group. Allowed values are `linux` and `windows` Changing this forces a new resource to be created.
+
+* `container` - (Required) The definition of a container that is part of the group. Currently, only single containers are supported for Windows OS and multiple containers are supported for Linux OS.
```
tombuildsstuff: We'd tend to phrase this as:

```
A `container` block as defined below.

~> **Note:** if `os_type` is set to `Windows` currently only a single container block is supported
```

website/docs/r/container_group.html.markdown
```diff
@@ -0,0 +1,113 @@
...
+
+The `container` block supports:
+
+* `name` - (Required) Specifies the name of the Container.
```
tombuildsstuff: Can we suffix this with "Changing this value forces a new resource to be created"?

website/docs/r/container_group.html.markdown
```diff
@@ -0,0 +1,113 @@
...
+
+* `image` - (Required) The container image name.
```
tombuildsstuff: Can we suffix this with "Changing this value forces a new resource to be created"?

website/docs/r/container_group.html.markdown
```diff
@@ -0,0 +1,113 @@
...
+
+* `cpu` - (Required) The required number of CPU cores of the containers.
```
tombuildsstuff: Can we suffix this with "Changing this value forces a new resource to be created"?

website/docs/r/container_group.html.markdown
abhijeetgaiha: you mean two container groups in one file?

website/docs/r/container_group.html.markdown
```diff
@@ -0,0 +1,113 @@
...
+
+* `memory` - (Required) The required memory of the containers in GB.
+
+* `port` - (Optional) A public port for the container.
```
tombuildsstuff: Can we suffix this with "Changing this value forces a new resource to be created"?

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

website/docs/r/container_group.html.markdown
tombuildsstuff: I was referring to picking either the Windows or the Linux example and removing the other (since they're virtually the same)

azurerm/resource_arm_container_group_test.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group_test.go
abhijeetgaiha: Done.

examples/aci-linux-multi/main.tf
abhijeetgaiha: Done.

website/docs/r/container_group.html.markdown
abhijeetgaiha: Done.

abhijeetgaiha: Removed the windows example.

website/docs/r/container_group.html.markdown
abhijeetgaiha: Done.

website/docs/r/container_group.html.markdown
abhijeetgaiha: Done.

website/docs/r/container_group.html.markdown
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

website/docs/r/container_group.html.markdown
abhijeetgaiha: Done.

website/docs/r/container_group.html.markdown
```diff
@@ -0,0 +1,113 @@
...
+
+The `container` block supports:
+
+* `name` - (Required) Specifies the name of the Container.
+
+* `image` - (Required) The container image name.
+
+* `cpu` - (Required) The required number of CPU cores of the containers.
+
+* `memory` - (Required) The required memory of the containers in GB.
```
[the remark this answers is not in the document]
abhijeetgaiha: Done.

website/docs/r/container_group.html.markdown
abhijeetgaiha: Done.

website/docs/r/container_group.html.markdown
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,323 @@
...
+					"linux",
+				}, true),
+			},
+
+			"tags": tagsForceNewSchema(),
+
+			"ip_address": {
+				Type:     schema.TypeString,
+				Computed: true,
+				ForceNew: true,
```
tombuildsstuff: given this is Computed, it doesn't need to be ForceNew - so we can remove this

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,323 @@
...
+	flattenAndSetTags(d, resp.Tags)
+
+	d.Set("os_type", string(resp.OsType))
+	if address := resp.IPAddress; address != nil {
+		d.Set("ip_address_type", address.Type)
+		d.Set("ip_address", address.IP)
+	}
+
+	containerConfigs := flattenContainerGroupContainers(resp.Containers)
+	d.Set("container", containerConfigs)
```
tombuildsstuff: _minor_ given this is setting a complex object, could we update this too:

```
containerConfigs := flattenContainerGroupContainers(resp.Containers)
if err = d.Set("container", containerConfigs) {
  return fmt.Errorf("Error setting `container`: %+v", err)
}
```

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,323 @@
...
+
+func flattenContainerGroupContainers(containers *[]containerinstance.Container) []interface{} {
+
+	containerConfigs := make([]interface{}, 0, len(*containers))
+	for _, container := range *containers {
+		containerConfig := make(map[string]interface{})
+		containerConfig["name"] = *container.Name
+		containerConfig["image"] = *container.Image
+
+		resourceRequests := *(*container.Resources).Requests
```
tombuildsstuff: could we add an `if`check around this to avoid Swagger/API issues? i.e.

```
if resources := container.Resources; resources != nil {
  if resourceRequests := resources.Requests; resourceRequests != nil {
    # ...
  }
}
```

azurerm/resource_arm_container_group.go
```diff
@@ -0,0 +1,323 @@
...
+			container.Ports = &[]containerinstance.ContainerPort{containerPort}
+
+			// container group port (port number + protocol)
+			containerGroupPort := containerinstance.Port{
+				Port: &port,
+			}
+
+			if v, ok := data["protocol"]; ok {
+				protocol := v.(string)
+				containerGroupPort.Protocol = containerinstance.ContainerGroupNetworkProtocol(strings.ToUpper(protocol))
```
tombuildsstuff: does this actually need to be `strings.ToUpper`?

abhijeetgaiha: It's needed as the ContainerGroupNetworkProtocol enum is composed of only all uppercase strings.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: The API will always return these values. Changing just in case.

tombuildsstuff:
Hey @abhijeetgaiha 

Sorry, I've unintentionally closed this by force-pushing the changes we discussed (to add an update test).. :( I'll re-open this in a separate PT (which will keep your commits)

PR #316, opened 2017-09-08
Support for Active Directory User Credentials

### LINKED ISSUE #42: AzureRM delegated user access?
_This issue was originally opened by @glenjamin as hashicorp/terraform#12208. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
terraform 0.8.7

### Affected Resource(s)
Azure Provider

I've been using terraform to configure a few things with the azurerm provider, which is broadly working well.

The part I'm not entirely keen on is how I have to create an Azure AD application that has appropriate permissions to perform resource operations per user who I want to have that ability - and annoyingly that azure apps can't be added to groups for permissions.

Clearly this isn't really terraform's fault. Azure AD applications are also allowed to run in a delegated resource mode, where they perform commands as the user in question - this is how the NodeJS, Python CLI apps and the Powershell Cmdlets work. The python code samples are relatively easy to follow here: https://github.com/AzureAD/azure-activedirectory-library-for-python 

Currently terraform is effectively using the "Acquire Token with Client Credentials" method whereas the other CLI tools use "Acquire Token with device code".

Is this something that terraform could be extended to support? 

Annoyingly there's no existing go code in the AzureAD org, but the pieces it uses all seem to be standard parts of OAuth2. The other downside is that the token only lasts for 1 hour, but does provide a refresh token that can be used to get a fresh one.

Perhaps an "easy" extension would be to allow terraform to accept the fully resolved azure bearer token, and then it would be possible for users to use their own mechanism to get and keep a valid bearer token if they wanted?

tombuildsstuff:
- [x] Adding support for authenticating using the Azure CLI access tokens
- [x] Rebase from `master`
- [x] Update vendor to include https://github.com/Azure/go-autorest/pull/169 once merged
- [x] Extensive testing
  - [x] Service Principal
    - [x] Azure Public
    - [x] Azure Germany
  - [x] Azure CLI (Python)
    - [x] Azure Public
    - [x] Azure Germany
    - [x] Long Running Tasks
  - [x] Azure CLI (Node)
    - [x] Azure Public
    - [x] Azure Germany
  - [x] Azure CloudShell
    - [x] Azure Public
    - [x] Long Running Tasks
- [x] Documentation

~This is dependent on https://github.com/Azure/go-autorest/pull/169~. Merged

Fixes #42


azurerm/provider.go
```diff
@@ -182,6 +192,118 @@ func (c *Config) validate() error {
...
+		if err != nil {
+			return fmt.Errorf("Error parsing expiration date: %q", accessToken.ExpiresOn)
+		}
+
+		if expirationDate.UTC().Before(time.Now().UTC()) {
+			log.Printf("[DEBUG] Token '%s' has expired", token.AccessToken)
+			continue
+		}
+
+		// TODO: replace this with something that's not terrible
```
mbfrahry: Will this TODO happen in this PR?

mbfrahry (approved):
LGTM! Just checking on a TODO

azurerm/provider.go
tombuildsstuff: Nope - I can't find a better way of doing this (as such I've removed it for the moment)

PR #322, opened 2017-09-11
Improve the provider initialization

svanharmelen:
The `registerAzureResourceProvidersWithSubscription` function is being called exactly once while executing the `providerConfigure` func.

In turn the `providerConfigure` func is also called exactly once while executing a Terraform command (e.g. plan, apply).

So when executing TF commands from the commandline there is absolutely no added value for using a `sync.Once` as the registration call is only being executed once anyway.

If on the other hand you use the TF packages in a long running microservice process (like we do), having this `sync.Once` actually causes issues when using multiple different (and regularly changing) provider credentials (as only the first used credentials will register the providers).

So since it adds nothing, but does cause issues in specific setups, we refactored the provider to not use the `sync.Once` anymore.

Additionally we added an option to skip the credential validation call. This is nice if you make multiple validation or plan calls in a row from a microservice so you can skip the expensive HTTP call (as you already know the credentials are valid).

### DIFF
--- a/azurerm/provider.go
+++ b/azurerm/provider.go
@@ -376,8 +376,6 @@ func registerProviderWithSubscription(providerName string, client resources.Prov
 	return nil
 }
 
-var providerRegistrationOnce sync.Once
-
 func determineAzureResourceProvidersToRegister(providerList []resources.Provider) map[string]struct{} {
 	providers := map[string]struct{}{
 		"Microsoft.Automation":          {},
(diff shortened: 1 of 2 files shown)


svanharmelen:
@tombuildsstuff would be great if you could review this PR so we can try to land it before the next release. Thx!

azurerm/provider.go
```diff
@@ -424,24 +430,23 @@ func determineAzureResourceProvidersToRegister(providerList []resources.Provider
 // whether they are actually used by the configuration or not). It was confirmed by Microsoft
 // that this is the approach their own internal tools also take.
 func registerAzureResourceProvidersWithSubscription(providerList []resources.Provider, client resources.ProvidersClient) error {
+	providers := determineAzureResourceProvidersToRegister(providerList)
+
 	var err error
-	providerRegistrationOnce.Do(func() {
```
tombuildsstuff: I'm not 100% sure of the original need for this - I'm assuming it's needed for the acceptance tests (which run in parallel) - to verify this I'll run the acceptance tests against this branch shortly to confirm

tombuildsstuff (approved):
Hey @svanharmelen 

Thanks for this PR - apologies for the delay in reviewing this!

I've taken a look through and I'm going to run the acceptance tests against this branch to confirm this won't affect those running in parallel - but this otherwise LGTM :)

Thanks!

azurerm/provider.go
tombuildsstuff: Acceptance tests passed - so looks like we're good here :)

svanharmelen: Cool, good the hear that! And thanks for the review :+1:

svanharmelen:
Thanks @tombuildsstuff!

PR #330, opened 2017-09-12
Adding validation to all resource_group_name attributes

### LINKED ISSUE #228: Add validation to `resource_group_name` field
At the moment we're not applying any validation the `resource_group_name` field apart from in the `azurerm_resource_group` resource.

Due to the way URL's are constructed in the Azure API's (by prefixing `/subscriptions/{subscription_id}/resourceGroups/{name}` to each request) - should the ID of a Resource Group be specified instead of it's Name - Azure will return an error about an Invalid API version (unless we're lucky and both API's have the same version).

This results in the following error:
```
disk.DisksClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidApiVersionParameter" Message="The api-version '2016-04-30-preview' is invalid. The supported versions are '2017-06-01,2017-05-10,2017-05-01,2017-03-01,2016-09-01,2016-07-01,2016-06-01,2016-02-01,2015-11-01,2015-01-01,2014-04-01-preview,2014-04-01,2014-01-01,2013-03-01,2014-02-26,2014-04'."
```

We should update the `resource_group_name` field across all resources to have validation to check this is a Name and not an ID.

Related bugs: #223 https://github.com/Azure/azure-sdk-for-go/issues/706

mbfrahry:
This PR makes a schema for resource group name and adds validation to every resource using resource_group_name. Fixes #228

### DIFF
--- a/azurerm/resource_arm_dns_zone.go
+++ b/azurerm/resource_arm_dns_zone.go
@@ -25,12 +25,7 @@ func resourceArmDnsZone() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"resource_group_name": {
-				Type:             schema.TypeString,
-				Required:         true,
-				ForceNew:         true,
-				DiffSuppressFunc: resourceAzurermResourceGroupNameDiffSuppress,
-			},
+			"resource_group_name": resourceGroupNameSchema(),
 
 			"number_of_record_sets": {
 				Type:     schema.TypeString,
--- a/azurerm/resource_arm_dns_zone.go
+++ b/azurerm/resource_arm_dns_zone.go
@@ -25,7 +25,7 @@ func resourceArmDnsZone() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"resource_group_name": resourceGroupNameDiffSupressSchema(),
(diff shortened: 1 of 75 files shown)


azurerm/resource_arm_dns_zone.go
```diff
@@ -25,12 +25,7 @@ func resourceArmDnsZone() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"resource_group_name": {
-				Type:             schema.TypeString,
-				Required:         true,
-				ForceNew:         true,
-				DiffSuppressFunc: resourceAzurermResourceGroupNameDiffSuppress,
-			},
+			"resource_group_name": resourceGroupNameSchema(),
```
tombuildsstuff: shouldn't this be `resourceGroupNameDiffSupressSchema()`?

tombuildsstuff (approved):
one nit, but otherwise LGTM :)

azurerm/resource_arm_dns_zone.go
mbfrahry: Good find! Changing it

PR #336, opened 2017-09-15
Container groups: Add support for env vars and command line

abhijeetgaiha:
More progress on #287

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ FEATURES:
 
 IMPROVEMENTS:
 
-* `azurerm_app_service_plan` - support for Linux App Services [GH-332]
+* `azurerm_app_service_plan` - support for Linux App Service Plans [GH-332]
 * `azurerm_postgresql_server` - supporting additional storage sizes [GH-239]
 * `azurerm_public_ip` - verifying the ID is valid before importing [GH-320]
 * `azurerm_sql_server` - verifying the name is valid before creating [GH-323]


azurerm/resource_arm_container_group.go
```diff
@@ -197,6 +188,9 @@ func resourceArmContainerGroupRead(d *schema.ResourceData, meta interface{}) err
 	resp, err := containterGroupsClient.Get(resGroup, name)
 
 	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return nil
```
tombuildsstuff: given the Read method is used for several things (both checking that a resource exists at Plan time, and for parsing data from the API) - can we add `d.SetId("")` before the `return nil` here to ensure that deleted resources are detected?

azurerm/resource_arm_container_group.go
```diff
@@ -265,12 +259,33 @@ func flattenContainerGroupContainers(containers *[]containerinstance.Container)
 		}
 		// protocol isn't returned in container config
 
+		if len(*container.EnvironmentVariables) > 0 {
```
tombuildsstuff: this'll crash if `container.EnvironmentVariables` is nil - could we add an if statement around this?

azurerm/resource_arm_container_group_test.go
```diff
@@ -25,6 +26,20 @@ func TestAccAzureRMContainerGroup_basicLinux(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, "container.0.env_vars.foo", "bar"),
+					resource.TestCheckResourceAttr(resourceName, "container.0.env_vars.foo1", "bar1"),
+					resource.TestCheckResourceAttr(resourceName, "container.0.image", "microsoft/aci-helloworld:latest"),
+					resource.TestCheckResourceAttr(resourceName, "container.0.memory", "0.5"),
+					resource.TestCheckResourceAttr(resourceName, "container.0.name", "hw"),
+					resource.TestCheckResourceAttr(resourceName, "container.0.port", "80"),
+					resource.TestCheckResourceAttr(resourceName, "ip_address_type", "Public"),
+					resource.TestCheckResourceAttr(resourceName, "os_type", "Linux"),
+					resource.TestCheckResourceAttr(resourceName, "tags.%", "1"),
+					resource.TestCheckResourceAttr(resourceName, "tags.environment", "testing"),
```
tombuildsstuff: I'd suggest that checking all of these is probably overkill, since the `Import` tests will do this for us (by deploying the input configuration, and then comparing the remote state against it - without specifying each fields values)

Instead - we could probably just check the OS Type / Environment Variables here?

website/docs/r/container_group.html.markdown
```diff
@@ -32,6 +32,10 @@ resource "azurerm_container_group" "aci-helloworld" {
         cpu ="0.5"
         memory =  "1.5"
         port = "80"
+        env_vars {
+            "env"="staging"
```
tombuildsstuff: it might sound odd, but in this current example the Environment Variable looks very similar to Tags (which we normally use `"Environment" = "Staging"` as an example) - as such would it make more sense to make this `"NODE_ENV" = "Staging"`?

azurerm/resource_arm_container_group.go
```diff
@@ -108,25 +108,16 @@ func resourceArmContainerGroup() *schema.Resource {
 							}, true),
 						},
 
-						"env_var": {
-							Type:     schema.TypeList,
+						"env_vars": {
```
tombuildsstuff: is it worth making this `environment_variables`?

azurerm/resource_arm_container_group.go
```diff
@@ -108,25 +108,16 @@ func resourceArmContainerGroup() *schema.Resource {
...
+							Type:     schema.TypeMap,
+							Optional: true,
```
tombuildsstuff: be aware, if we'd done a release of this, we'd have to include a State Migration to handle converting between the two fields going from a TypeList -> TypeMap - but I think we can probably overlook that this time since this hasn't been released yet

abhijeetgaiha: 👍

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done.

azurerm/resource_arm_container_group.go
abhijeetgaiha: It's never nil, but a check is warranted to be safe. Done.

azurerm/resource_arm_container_group_test.go
abhijeetgaiha: In addition, I'll leave command in there too, remove the others.

website/docs/r/container_group.html.markdown
abhijeetgaiha: I made them the same intentionally, but this change is fine.

azurerm/resource_arm_container_group.go
abhijeetgaiha: Done. Also matches with the CLI.

azurerm/resource_arm_container_group.go
tombuildsstuff: To add some context to why this check is needed: if the Swagger is incorrect (which historically has happened a lot) this would be mapped incorrectly and crash

tombuildsstuff:
@abhijeetgaiha sorry, messed up git again (I think it's pushing to a remote `master` branch which is breaking things) - I've opened #338 which includes this.

PR #393, opened 2017-10-04
Making the `azurerm_client_config` data source work with AzureCLI auth

### LINKED ISSUE #392: azurerm_client_config error listing Service Principals
Error on getting data from azurerm_client_config
```* data.azurerm_client_config.current: data.azurerm_client_config.current: Error listing Service Principals: autorest.DetailedError{Original:(*azure.RequestError)(0xc420619ef0), PackageType:"graphrbac.ServicePrincipalsClient", Method:"List", StatusCode:401, Message:"Failure responding to request", ServiceError:[]uint8(nil), Response:(*http.Response)(0xc420619e60)}```

### Terraform Version
```
Terraform v0.10.7
.
└── provider.azurerm 0.2.2
```
### Affected Resource(s)
- data.azurerm_client_config

```hcl
provider "azurerm" {}

data "azurerm_client_config" "current" {}

output "account_id" {
  value = "${data.azurerm_client_config.current.account_id}"
}
```

### Debug Output
[REMOVED]

### Actual Behavior
failed to get any data 

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
Tried both with az cli auth and service principal
i'm not an admin of whole account but have subscription owner role
list service principals from az cli successful with same credentials
`az ad sp list`

tombuildsstuff:
Fixes #392

Unfortunately testing this is tricky, since we rely on a Service Principal to run the acceptance tests (and not AzureCLI auth) - however I've verified this works as expected in both cases

Tests pass:

```
$ acctests azurerm TestAccDataSourceAzureRMClientConfig_basic
=== RUN   TestAccDataSourceAzureRMClientConfig_basic
--- PASS: TestAccDataSourceAzureRMClientConfig_basic (12.65s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	12.673s
```

### DIFF
--- a/azurerm/config.go
+++ b/azurerm/config.go
@@ -46,10 +46,11 @@ import (
 // ArmClient contains the handles to all the specific Azure Resource Manager
 // resource classes' respective clients.
 type ArmClient struct {
-	clientId       string
-	tenantId       string
-	subscriptionId string
-	environment    azure.Environment
+	clientId              string
+	tenantId              string
+	subscriptionId        string
+	usingServicePrincipal bool
+	environment           azure.Environment
 
 	StopContext context.Context
 
@@ -233,10 +234,11 @@ func (c *Config) getArmClient() (*ArmClient, error) {
 
 	// client declarations:
 	client := ArmClient{
-		clientId:       c.ClientID,
-		tenantId:       c.TenantID,
-		subscriptionId: c.SubscriptionID,
-		environment:    env,
+		clientId:              c.ClientID,
+		tenantId:              c.TenantID,
+		subscriptionId:        c.SubscriptionID,
+		environment:           env,
+		usingServicePrincipal: c.ClientSecret != "",
 	}
 
 	oauthConfig, err := adal.NewOAuthConfig(env.ActiveDirectoryEndpoint, c.TenantID)
--- a/azurerm/data_source_arm_client_config.go
+++ b/azurerm/data_source_arm_client_config.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/arm/graphrbac"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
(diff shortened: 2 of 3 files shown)


azurerm/data_source_arm_client_config.go
```diff
@@ -38,28 +39,35 @@ func dataSourceArmClientConfig() *schema.Resource {
...
+		if listErr != nil {
+			return fmt.Errorf("Error listing Service Principals: %#v", listErr)
+		}
+
+		if listResult.Value == nil || len(*listResult.Value) != 1 {
+			return fmt.Errorf("Unexpected Service Principal query result: %#v", listResult.Value)
+		}
 
-	servicePrincipal := (*listResult.Value)[0]
+		servicePrincipal = &(*listResult.Value)[0]
```
radeksimko: 🙀

azurerm/config.go
```diff
@@ -46,10 +46,11 @@ import (
...
 // resource classes' respective clients.
 type ArmClient struct {
-	clientId       string
-	tenantId       string
-	subscriptionId string
-	environment    azure.Environment
+	clientId              string
+	tenantId              string
+	subscriptionId        string
+	usingServicePrincipal bool
```
radeksimko: Nitpick, but `using*` to me sounds like a state that may be managed from somewhere outside, how do you feel about calling it `useServicePrinciple` instead?

radeksimko (approved):
LGTM, although I don't know much about service principles nor Azure, so take that with a pinch of salt.

azurerm/config.go
tombuildsstuff: it will be managed from somewhere outside (it's determined in the Config class, and then cast to an `ArmConfig` once the authentication tokens have been obtained) - as such I think `using` makes more sense in this context than `use`? That said, I'm happy to change this - but I'm a little stumped for a better name ¯\_(ツ)_/¯

tombuildsstuff: (merging this for the moment - we can rename this later if needed)

PR #397, opened 2017-10-05
Ensuring we register the MySQL Resource Provider

tombuildsstuff:
Tests pass:

```
$ acctests azurerm TestAccAzureRMResourceProviderRegistration
=== RUN   TestAccAzureRMResourceProviderRegistration
# ...
--- PASS: TestAccAzureRMResourceProviderRegistration (1.93s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	1.955s
```

### DIFF
--- a/azurerm/provider.go
+++ b/azurerm/provider.go
@@ -404,6 +404,7 @@ func determineAzureResourceProvidersToRegister(providerList []resources.Provider
 		"Microsoft.ContainerInstance":   {},
 		"Microsoft.ContainerRegistry":   {},
 		"Microsoft.ContainerService":    {},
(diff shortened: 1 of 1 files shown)


radeksimko (commented):
Were these tests not passing before? i.e. how did you discover it's not being registered?

tombuildsstuff:
The tests were passing before - but they're verifying that the values in the array are of the correct case - not that resources can be provisioned within that Resource Provider (which is outside the scope of the test).

I noticed it when doing the SDK upgrade - the Resource Provider gets automatically registered a resource is provisioned in the Portal (which we do to confirm the SKU's) - and wasn't added to the list.

PR #157, opened 2017-07-07
[WIP] Add encryption_settings to VMs and computed disks

whiskeyjay:
Moving the original PR (https://github.com/hashicorp/terraform/pull/14261) from core to here. Will add new tests soon.

PR #399, opened 2017-10-05
`azurerm_managed_disk` additions

### LINKED ISSUE #327: azurerm_managed_disk does not support FromImage
The Azure REST API supports creating a separate managed disk instance from a standard offering image, as opposed to within a VM OS disk definition block.

https://docs.microsoft.com/en-us/rest/api/compute/manageddisks/disks/disks-create-or-update

There are a number of terraform use-cases where separating the disk definition from the VM definition is extremely useful. 
Particularly: Certain changes to a VM require a destroy/rebuild and terraform can't reattach an existing OS disk without code changes (FromImage -> Attach).

This means that all your VM redeploys that don't set `delete os disk on termination` will fail.

This means otherwise generic and maintainable IaaS code is now operation specific to that one change.

### Terraform Version
0.10.0

### Affected Resource(s)
- azurerm_managed_disk

### Expected Behavior
Terraform creates the managed disk and then attaches it to the VM.

### Actual Behavior
Terraform throws the following error:
`module.dcos.azurerm_managed_disk.dcosBootstrapOSDisk: expected create_option to be one of [Import Empty Copy], got FromImage`

tombuildsstuff:
- Enables creating Managed Disks from Platform Images by supporting "FromImage"
- Enables creating Encrypted Managed Disks as started in #157 

Fixes #327

### DIFF
--- a/azurerm/resource_arm_managed_disk.go
+++ b/azurerm/resource_arm_managed_disk.go
@@ -47,9 +47,10 @@ func resourceArmManagedDisk() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 				ValidateFunc: validation.StringInSlice([]string{
-					string(disk.Import),
-					string(disk.Empty),
 					string(disk.Copy),
+					string(disk.Empty),
+					string(disk.FromImage),
+					string(disk.Import),
 				}, true),
 			},
 
@@ -66,6 +67,12 @@ func resourceArmManagedDisk() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"image_reference_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
 			"os_type": {
 				Type:     schema.TypeString,
 				Optional: true,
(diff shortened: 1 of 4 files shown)


azurerm/resource_arm_managed_disk.go
```diff
@@ -77,10 +84,12 @@ func resourceArmManagedDisk() *schema.Resource {
 
 			"disk_size_gb": {
```
mbfrahry: Should we change how we get `disk_size_gb` now that it's not required? Swap to `GetOk`?

mbfrahry (approved):
LGTM with a minor comment

azurerm/resource_arm_managed_disk.go
tombuildsstuff: we could - however given this is an Integer it wouldn't help (as the default value is still `0`) - as such I think the current `Get` with a check > 0 should work (from below):

```
if v := d.Get("disk_size_gb"); v != 0 {
```

PR #404, opened 2017-10-06
`azurerm_template_deployment` - support for specifying parameters via `parameters_body`

civascu:
Currently, the resource azurerm_template_deployment only supports passing in parameters as a key, value pair. This does not allow the use of parameters that reference a KeyVault that is enabled for template deployment, which is desired in some use-cases. 

One such use-case is executing a bootstrap script when a virtual machine (or scale set) comes online - via the Azure extension mechanism - and inject secrets into the script being executed, without the operator that runs terraform to know their value upfront.

This is enabled similar to the template_body parameter, and it passes the JSON string to the Azure Rest API.

### DIFF
--- a/azurerm/resource_arm_template_deployment.go
+++ b/azurerm/resource_arm_template_deployment.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2017-05-10/resources"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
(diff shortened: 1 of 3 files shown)


civascu:
One similar attempt I ran into was here: https://github.com/hashicorp/terraform/issues/8331

azurerm/resource_arm_template_deployment.go
```diff
@@ -42,6 +42,11 @@ func resourceArmTemplateDeployment() *schema.Resource {
 				Optional: true,
 			},
 
+			"parameters_body": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
```
paultyng: Can you set this as mutually exclusive with `parameters` (see `ConflictsWith`) and validate that it is valid JSON?

paultyng:
Looks like there is a merge conflict as well, so you should probably rebase.

azurerm/resource_arm_template_deployment.go
paultyng: You may also want a `StateFunc` to normalize the JSON so that things like whitespace don't cause diffs.

tombuildsstuff:
hey @civascu 

Apologies for the delayed movement on this PR - I'm going to rebase this and push a commit to fix the issues that @paultyng identified (I hope you don't mind!); then we should be good to run the tests with this.

Thanks!

azurerm/resource_arm_template_deployment.go
tombuildsstuff: pushed a commit to add these

tombuildsstuff:
Tests pass:

```
$ acctests azurerm TestAccAzureRMTemplateDeployment_
=== RUN   TestAccAzureRMTemplateDeployment_basic
--- PASS: TestAccAzureRMTemplateDeployment_basic (218.88s)
=== RUN   TestAccAzureRMTemplateDeployment_disappears
--- PASS: TestAccAzureRMTemplateDeployment_disappears (210.92s)
=== RUN   TestAccAzureRMTemplateDeployment_withParams
--- PASS: TestAccAzureRMTemplateDeployment_withParams (230.95s)
=== RUN   TestAccAzureRMTemplateDeployment_withOutputs
--- PASS: TestAccAzureRMTemplateDeployment_withOutputs (221.63s)
=== RUN   TestAccAzureRMTemplateDeployment_withError
--- PASS: TestAccAzureRMTemplateDeployment_withError (354.04s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	1236.456s
```

PR #407, opened 2017-10-09
Provision sample for ASP.NET on azure_rm_app_service

pjmolina:
Hi!
Added an example for provisioning an **ASP.NET** website into **Azure App Service** with terraform.


examples/app-service-asp-net/app.tf
```diff
@@ -0,0 +1,65 @@
...
+output "service" {
+  value = "${azurerm_app_service.common_service.name}"
+}
+
+output "serviceUrl" {
+  value = "https://${azurerm_app_service.common_service.name}.azurewebsites.net"
+}
+
+output "adminUrl" {
+  value = "https://${azurerm_app_service.common_service.name}.scm.azurewebsites.net"
```
tombuildsstuff: _note:_ these two values (`serviceUrl` and `adminUrl`) assume we're running in Azure Public (and not Germany/Government/China etc) - I took a quick look into if we could expose these variables so this example could also work in those environments, but unfortunately those values aren't immediately available - so this should be fine :)

examples/app-service-asp-net/app.tf
```diff
@@ -0,0 +1,65 @@
+# Configure the Microsoft Azure Provider
+provider "azurerm" {
+  subscription_id = "${var.subscription_id}"
+  client_id       = "${var.client_id}"
+  client_secret   = "${var.client_secret}"
+  tenant_id       = "${var.tenant_id}"
+}
+
+resource "azurerm_resource_group" "g1" {
+  name     = "{$var.groupName}"
```
tombuildsstuff: this should be `${var.groupName}` (I'll push a commit to fix this)

examples/app-service-asp-net/app.tf
```diff
@@ -0,0 +1,65 @@
+# Configure the Microsoft Azure Provider
+provider "azurerm" {
+  subscription_id = "${var.subscription_id}"
+  client_id       = "${var.client_id}"
+  client_secret   = "${var.client_secret}"
+  tenant_id       = "${var.tenant_id}"
+}
```
tombuildsstuff: we tend to leave this commented out in case people opt to use Azure CLI auth (I'll push a commit to amend this)

examples/app-service-asp-net/variables.tf
```diff
@@ -0,0 +1,45 @@
+variable "groupName" {
```
tombuildsstuff: _minor_ I might suggest renaming this `resource_group_name` to be clearer (and underscores for consistency) -  (I'll push a commit for this)

examples/app-service-asp-net/variables.tf
```diff
@@ -0,0 +1,45 @@
...
+  type    = "string"
+  default = "yourGroupName"
+}
+
+variable "location" {
+  type    = "string"
+  default = "westeurope"
+}
+
+variable "webName" {
```
tombuildsstuff: (same here) `webName` -> `web_name` (I'll push a commit for this)

examples/app-service-asp-net/app.tf
```diff
@@ -0,0 +1,65 @@
...
+  #   "SOME_KEY" = "some-value"
+  # }
+  # connection_string {
+  #   name  = "Database"
+  #   type  = "SQLServer"
+  #   value = "Server=some-server.mydomain.com;Integrated Security=SSPI"
+  # }
+
+  provisioner "local-exec" {
+    command = "curl -k -u ${var.deploy_user}:${var.deploy_pass} -X PUT --data-binary @${var.deployZipFile} https://${azurerm_app_service.common_service.name}.scm.azurewebsites.net/api/zip/site/wwwroot/"
```
tombuildsstuff: out of interest, how are these values set? From memory these need to be set either in the portal/through the CLI before use, given Terraform currently doesn't support them and they're disabled by default?

pjmolina: Thanks for the review Tom.
Yes, I didn't find a way yet to setup such deploy Azure credentials in a CLI/programmatic way.
As far as I know, them should be setup once in the Azure Portal per Azure Account.
Probably MS Azure guys can help with it.

tombuildsstuff: hey @pjmolina 

In order for us to be able to ship this PR (without making the user having to navigate to the portal to enter the credentials) - I've pushed a commit to remove the publishing credentials snippet from the sample for the moment. Once support for configuring the App Service Publish Credentials is available through Terraform we can look into adding this back in - however for the moment this feels like the best compromise to ship this.

I've also opened #612 which exports the Default Hostname for the App Service - such that we can determine what it is (since `azurewebsites.net` is only valid for Azure Public and not China/Germany/Government) - so once that's merged this PR should be good to merge :)

Hope that makes sense and you don't mind?

Thanks!

pjmolina: Looks great @tombuildsstuff !

tombuildsstuff: @pjmolina totally get that - we'll look to add it back in to the example once support for configuring the credentials is added to the Provider :)

PR #408, opened 2017-10-10
New Resource: `azurerm_key_vault_certificate`

tombuildsstuff:
- Support for Importing and Generating Key Vault Certificates
- Updating the Key Vault SKU to be a List rather than a Set
- Refactoring all the Key Vault child resources to use the same url parser given it's the same

The documentation for Key Vault Certificates is pretty non-existant, so I'm unable to determine which fields can be updated - as such I've made everything ForceNew for the moment and will implement Update in a future update.


tombuildsstuff:
Tests pass:

screen shot 2017-10-10 at 12 11 23

mbfrahry (approved):
LGTM!

PR #409, opened 2017-10-10
refactoring of `azurerm_network_interface` and `azurerm_availability_set`

tombuildsstuff:
- Exporting the list of Private IP Addresses assigned to the VM
- Refactoring the `azurerm_network_interface` and `azurerm_availability_set` resources

### DIFF
--- a/azurerm/resource_arm_network_interface.go
+++ b/azurerm/resource_arm_network_interface.go
@@ -151,6 +151,13 @@ func resourceArmNetworkInterface() *schema.Resource {
 				Computed: true,
 			},
 
+			"private_ip_addresses": {
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
+
 			"tags": tagsSchema(),
 		},
 	}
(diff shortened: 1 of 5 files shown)


azurerm/resource_arm_network_interface.go
```diff
@@ -151,6 +151,13 @@ func resourceArmNetworkInterface() *schema.Resource {
 				Computed: true,
 			},
 
+			"private_ip_addresses": {
+				Type:     schema.TypeSet,
```
radeksimko: I'm thinking this _may_ theoretically cause some troubles when referencing the field as set has computed indexes? 🤔

radeksimko (approved):
Just one minor question, otherwise LGTM. I'm assuming all related acceptance tests are passing.

azurerm/resource_arm_network_interface.go
tombuildsstuff: Good point.. I'll look into switching this over to a list

tombuildsstuff: Updated

tombuildsstuff:
Tests pass:

screen shot 2017-10-11 at 15 12 07

PR #419, opened 2017-10-11
Typo fix

tmonk42:
Replace Postgres with MySQL.


tmonk42:
Looks like the MySQL documentation might have been copied from Postgres. A few updates here for places where the search and replace missed and updating the examples to include a valid version (examples had 9.5, which is a valid postgres version...)

PR #421, opened 2017-10-12
`azurerm_storage_account` - fixes #302

### LINKED ISSUE #302: azurerm_storage_account invalid parameter Standard_ZRS for BlobStorage
When creating a storage account which is blob storage, Standard_ZRS is not a valid type. terraform plan and apply don't warn on this so the user is left to decipher a 400 message from Azure

### Terraform Version
Terraform v0.10.3

### Affected Resource(s)
azurerm_storage_account

```hcl
resource "azurerm_storage_account" "anyname123" {
    name = "anyname123"
    resource_group_name = "anyresourcegroup"
    location = "UK South"
    account_type = "Standard_ZRS"
    account_kind = "BlobStorage"
    access_tier = "hot"
}
```

### Debug Output
N/A

### Panic Output
N/A

### Expected Behavior
It should have told me that Standard_ZRS as account_type is not supported for account_kind of BlobStorage

### Actual Behavior
A plan succeeds

An apply fails with below message
```
* azurerm_storage_account.anyname123: 1 error(s) occurred:

* azurerm_storage_account.anyname123: Error creating Azure Storage Account 'anyname123': storage.AccountsClient#Create: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidInput" Message="One of the request inputs is not valid."
```

### Steps to Reproduce
 `terraform apply`

### Important Factoids
N/A

### References
N/A

tombuildsstuff:
Returning a user friendly error when trying to provision a Blob Storage Account with `ZRS` redundancy

Fixes #302

### DIFF
--- a/azurerm/resource_arm_storage_account.go
+++ b/azurerm/resource_arm_storage_account.go
@@ -253,6 +253,10 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
 
 	// AccessTier is only valid for BlobStorage accounts
 	if accountKind == string(storage.BlobStorage) {
+		if string(parameters.Sku.Name) == string(storage.StandardZRS) {
+			return fmt.Errorf("A `account_replication_type` of `ZRS` isn't supported for Blob Storage accounts.")
+		}
+
 		accessTier, ok := d.GetOk("access_tier")
 		if !ok {
 			// default to "Hot"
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
Tests still pass:

screen shot 2017-10-12 at 16 11 06

azurerm/resource_arm_storage_account.go
```diff
@@ -253,6 +253,10 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
 
 	// AccessTier is only valid for BlobStorage accounts
 	if accountKind == string(storage.BlobStorage) {
+		if string(parameters.Sku.Name) == string(storage.StandardZRS) {
```
vancluever: Would this check be needed in `Update` too as `account_replication_type` is not `ForceNew`?

tombuildsstuff: good point, most of the Azure resources are `CreateOrUpdate` - so I'd assumed this was the same - I'll update it now..

PR #367, opened 2017-09-29
Upgrading to v11 of the Azure SDK for Go

tombuildsstuff:
Upgrading to v11 of the Azure SDK for Go

- [x] Vendoring `v11.0.0-beta` of the Azure SDK
- [x] State Migrations for:
  - [x] Container Registry
      - [x] `Basic` SKU -> `Classic`
      - [x] State Migration Storage Account Name & Key -> Storage Account ID
  - [x] Key Vault  
    - [x] `certificate_permissions` All -> every option
    - [x] `key_permissions` All -> every option
    - [x] `secret_permissions` All -> every option
  - [x] ServiceBus Namespace
    - [x] Remove `capacity` if not a `premium` sku
- [x] Pending issues:
  - [x] ~[Traffic Manager API version used in the SDK isn't available](https://github.com/Azure/azure-sdk-for-go/issues/794)~ fixed
  - [x] ~[Container Registry API version used in the SDK isn't available](https://github.com/Azure/azure-sdk-for-go/issues/795)~ fixed
 - [ ] Ensure all breaking changes are updated in the Documentation

### Changes and Deprecation Notices:

  - `azurerm_automation_account` - the SKU `Free` has been replaced with `Basic`.
  - `azurerm_container_registry` - Azure has updated the SKU from `Basic` to `Classic`, with new `Basic`, `Standard` and `Premium` SKU's introduced.
  - `azurerm_container_registry` - the `storage_account` block is now `storage_account_id` and is only required for `Classic` SKU's
  - `azurerm_key_vault` - `certificate_permissions`, `key_permissions` and `secret_permissions` have all had the `All` option removed by Azure. Each permission now needs to be specified manually.
  - `azurerm_servicebus_namespace` - The `capacity` field can only be set for `Premium` SKU's
  - `azurerm_servicebus_queue` - The `enable_batched_operations` and `support_ordering` fields have been deprecated by Azure.
  - `azurerm_servicebus_subscription` - The `dead_lettering_on_filter_evaluation_exceptions` has been removed by Azure.
  - `azurerm_servicebus_topic` - The `enable_filtering_messages_before_publishing` field has been removed by Azure.

PR #422, opened 2017-10-12
Fixing broken link in documentation

piotrgo:


piotrgo:
Also, I can't find any info on how often is documentation updated with new changes, but this site (https://www.terraform.io/docs/providers/azurerm/r/storage_account.html) doesn't look right according to the code (ie. missing information about file encryption)

tombuildsstuff:
Hey @piotrgo 

Thanks for this PR :)

> Also, I can't find any info on how often is documentation updated with new changes, but this site (https://www.terraform.io/docs/providers/azurerm/r/storage_account.html) doesn't look right according to the code (ie. missing information about file encryption)

We update the website after each release for new functionality (as you've identified) or for changes which don't need to wait for a release we try and push them out sooner. In this case we've been trying to ship PR #367 (which upgrades us to the latest version of the Azure SDK for Go) as part of the `0.3` release so that we can try to limit the number of breaking changes needed - so this functionality should go out in the near future :)

Thanks!

PR #430, opened 2017-10-16
Adding Key Vault Access Policy to the sidebar

tombuildsstuff:


website/azurerm.erb
```diff
@@ -35,6 +35,10 @@
                     <a href="/docs/providers/azurerm/d/image.html">azurerm_image</a>
                 </li>
 
+                <li<%= sidebar_current("docs-azurerm-datasource-key_vault_access_policy") %>>
```
radeksimko: Nitpick, but I think the convention here is to prefer hyphens.

tombuildsstuff: Good point: updated

PR #445, opened 2017-10-20
only use the default subscription if not set explicitly

lamdor:
I found this after I was setting the the subscription and tenant Id in the provider config via:

```hcl
provider "azurerm" {
  subscription_id = "...."
  tenant_id       = "...."
}

```
 and then logging in with `az login`. However, it the terraform provider was still trying to use what ever my configured default subscription is. I expected it to use what I had it configured explicitly to use. This fixes that.

### DIFF
--- a/azurerm/provider.go
+++ b/azurerm/provider.go
@@ -253,13 +253,15 @@ func (c *Config) LoadTokensFromAzureCLI() error {
 		return fmt.Errorf("Azure CLI Authorization Profile was not found. Please ensure the Azure CLI is installed and then log-in with `az login`.")
 	}
 
-	// pull out the TenantID and Subscription ID from the Azure Profile
-	for _, subscription := range profile.Subscriptions {
-		if subscription.IsDefault {
-			c.SubscriptionID = subscription.ID
-			c.TenantID = subscription.TenantID
-			c.Environment = normalizeEnvironmentName(subscription.EnvironmentName)
-			break
+	// pull out the TenantID and Subscription ID from the Azure Profile, if not set
+	if c.SubscriptionID == "" && c.TenantID == "" {
+		for _, subscription := range profile.Subscriptions {
+			if subscription.IsDefault {
+				c.SubscriptionID = subscription.ID
+				c.TenantID = subscription.TenantID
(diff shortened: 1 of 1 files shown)


tombuildsstuff (approved):
Hey @rubbish 

Thanks for this PR - apologies for the delayed review! I've taken a look and done some testing and this LGTM :)

Thanks!

pmarques:
Thanks @rubbish and @tombuildsstuff, this fix is really nice, just got into the same problem.

Just a quick thing, this fix broke my "configurations" which keeps throwing just a weird message:

```                                                                                                                                              
Error: Error refreshing state: 1 error(s) occurred:                                                                      
                                                                                                                 
* provider.azurerm: No valid (unexpired) Azure CLI Auth Tokens found. Please run `az login`.     
```

You can do what the message tells you to do but after running again you get the same error.
My provider is only configured with the **subscription_id** (since per documentation everything is optional):
```
provider "azurerm" {
  subscription_id = "${var.subscription_id}"
}
```

This happens because **tenant_id** is now required, this condition [condition](https://github.com/Banno/terraform-provider-azurerm/blob/d63c3703e3075371ef3ae33a2c9fb2c93ff16fc2/azurerm/provider.go#L269) fails and then **validateBearerAuth** [here](https://github.com/Banno/terraform-provider-azurerm/blob/d63c3703e3075371ef3ae33a2c9fb2c93ff16fc2/azurerm/provider.go#L238) will also fail.

I can think at least 2 ways of addressing this issue:
 * we can change the log message [here](https://github.com/Banno/terraform-provider-azurerm/blob/d63c3703e3075371ef3ae33a2c9fb2c93ff16fc2/azurerm/provider.go#L359) to print the actual cause
 * We can review the provider config validation to make **tenant_id** *optional*, dropping the TenantID validation [here](https://github.com/Banno/terraform-provider-azurerm/blob/d63c3703e3075371ef3ae33a2c9fb2c93ff16fc2/azurerm/provider.go#L238),  as far I was able to check it works if we remove that validation but it can be a good practice anyway.

tombuildsstuff:
@pmarques thanks for raising this - I'm going to look into a fix for this now - and given issue #562 is specific to this I'll post an update there

Thanks!

Issue #340, opened 2017-09-17
Please add public_ip_address capability to azurerm_virtual_machine_scale_set


samirageb:
According to this [blog post](https://azure.microsoft.com/en-us/blog/new-networking-features-in-azure-scale-sets/) Azure scale sets can now get IPv4 addresses assigned directly to the VM. For our particular workload, this would help us tremendously. 

### Terraform Version
Terrraform 0.10.5
AzureRM TF Provider 0.2.0

### Affected Resource(s)
- azurerm_virtual_machine_scale_set

tombuildsstuff:
Hey @samirageb 

Thanks for opening this issue :)

From what I can see this functionality requires the new API version - [which is requested in this issue](https://github.com/Azure/azure-sdk-for-go/issues/744) - and should be available to us once [the v11 SDK](https://github.com/Azure/azure-sdk-for-go/pull/768) has been merged :)

Thanks!

katbyte:
I believe this was fixed by #448 as can be seen https://github.com/terraform-providers/terraform-provider-azurerm/pull/448/files#diff-35d6282e7fd373522705ded934097ca0R298

Issue #339, opened 2017-09-17
Please add accelerated networking to azurerm_virtual_machine_scale_set


samirageb:
I assume this [issue](https://github.com/Azure/azure-sdk-for-go/issues/744) is also a pre-req before this capability can be added, but wanted to get this tracked.

### Terraform Version
- Terraform 0.10.5
- AzureRM TF Provider 0.2.0

### Affected Resource(s)
- azurerm_virtual_machine_scale_set

### References
- #272

tombuildsstuff:
Hey @samirageb 

Thanks for opening this issue :)

Your assumption is correct (as both VM's and VMSS's are part of the `Compute` SDK) - so once [the v11 SDK PR](https://github.com/Azure/azure-sdk-for-go/pull/768) has been merged we'll look into upgrading to that and then investigating supporting this functionality :)

Thanks!

tombuildsstuff:
hi @samirageb 

Thanks for opening this issue - apologies for the delayed response here!

Support for Accelerated Networking was added in #448 which has been available in the AzureRM Provider since v0.3.1: https://github.com/terraform-providers/terraform-provider-azurerm/blob/v0.3.1/CHANGELOG.md#031-october-21-2017

Thanks!

PR #448, opened 2017-10-20
Tackling various vmss issues

### LINKED ISSUE #160: VMSS: support for specifying the OS Disk Size
### Terraform Version
Terraform v0.9.11

### Affected Resource(s)
- azurerm_virtual_machine_scale_set

There are several things that aren't available for azurerm_virtual_machine_scale_set but are available for individual VMs.  I have done all of them using ARM templates so I know it's possible in Azure, but terraform doesn't allow them (I tried :)

Please let me know if I should split this out into separate requests for each.

1. ~Would it be possible to support a boot_diagnostics block in the VMSS configuration, just like the one in azurerm_virtual_machine?  See the "How do I turn on boot diagnostics" section here: https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-faq~ (@tombuildsstuff: this is now available)

2. ~Could you please add support for network security groups in a VMSS?  This is already there for azurerm_network_interface but not for the scale set.  It seems like it would just be a network_security_group_id field inside the VMSS network_profile.~ (@tombuildsstuff: this is now available)

3. Could you please add the ability to specify a size for OS Disks?  In azurerm_virtual_machine there is a storage_os_disk block where you can specify disk_size_gb, but with the VMSS the storage_profile_**os**_disk block does not support the disk_size_gb field (the VMSS storage_profile_data_disk does support disk_size_gb but I need the OS disk).

Thank you!

mbfrahry:
This PR adds the new sdk features to virtual machine scale sets and fixes #160, #339, and #340 

```
=== RUN   TestAccAzureRMVirtualMachineScaleSet_basicPublicIP
--- PASS: TestAccAzureRMVirtualMachineScaleSet_basicPublicIP (402.10s)
=== RUN   TestAccAzureRMVirtualMachineScaleSet_bootDiagnostic
--- PASS: TestAccAzureRMVirtualMachineScaleSet_bootDiagnostic (400.54s)
=== RUN   TestAccAzureRMVirtualMachineScaleSet_networkSecurityGroup
--- PASS: TestAccAzureRMVirtualMachineScaleSet_networkSecurityGroup (400.65s)
=== RUN   TestAccAzureRMVirtualMachineScaleSet_basicAcceleratedNetworking
--- PASS: TestAccAzureRMVirtualMachineScaleSet_basicAcceleratedNetworking (339.38s)
```

### DIFF
--- a/azurerm/resource_arm_virtual_machine_scale_set.go
+++ b/azurerm/resource_arm_virtual_machine_scale_set.go
@@ -628,14 +628,6 @@ func resourceArmVirtualMachineScaleSetCreate(d *schema.ResourceData, meta interf
 		scaleSetProps.VirtualMachineProfile.DiagnosticsProfile = &diagnosticProfile
 	}
 
-	/* TODO remove this
-	networkConfigurations := scaleSetProps.VirtualMachineProfile.NetworkProfile.NetworkInterfaceConfigurations
-	for _, netConfig := range *networkConfigurations {
-		for _, ipConfig := range *netConfig.VirtualMachineScaleSetNetworkConfigurationProperties.IPConfigurations {
-			return fmt.Errorf("Here: %+v", *ipConfig.VirtualMachineScaleSetIPConfigurationProperties.Primary)
-		}
-	}*/
-
 	scaleSetParams := compute.VirtualMachineScaleSet{
 		Name:     &name,
 		Location: &location,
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -250,6 +250,17 @@ func resourceArmVirtualMachineScaleSet() *schema.Resource {
 							Required: true,
 						},
 
+						"accelerated_networking": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
```
vancluever: Maybe omit zero value here?

PR #482, opened 2017-11-01
Support for managed service identity

hbuckle:
Adds the identity property to virtual machines so that managed service identity can be enabled:
[https://docs.microsoft.com/en-us/azure/active-directory/msi-overview](https://docs.microsoft.com/en-us/azure/active-directory/msi-overview)

### DIFF
--- a/azurerm/resource_arm_virtual_machine.go
+++ b/azurerm/resource_arm_virtual_machine.go
@@ -675,9 +675,7 @@ func resourceArmVirtualMachineRead(d *schema.ResourceData, meta interface{}) err
 		}
 	}
 
-	if resp.Identity != nil {
-		d.Set("identity", flattenAzureRmVirtualMachineIdentity(resp.Identity))
-	}
+	d.Set("identity", flattenAzureRmVirtualMachineIdentity(resp.Identity))
 
 	if resp.VirtualMachineProperties.AvailabilitySet != nil {
 		d.Set("availability_set_id", strings.ToLower(*resp.VirtualMachineProperties.AvailabilitySet.ID))
@@ -908,6 +906,10 @@ func flattenAzureRmVirtualMachineImageReference(image *compute.ImageReference) [
 }
 
 func flattenAzureRmVirtualMachineIdentity(identity *compute.VirtualMachineIdentity) []interface{} {
+	if identity == nil {
+		return make([]interface{}, 0)
+	}
+
 	result := make(map[string]interface{})
 	result["type"] = string(identity.Type)
 	if identity.PrincipalID != nil {
--- a/azurerm/resource_arm_virtual_machine.go
+++ b/azurerm/resource_arm_virtual_machine.go
@@ -72,7 +72,7 @@ func resourceArmVirtualMachine() *schema.Resource {
 			},
 
 			"identity": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Optional: true,
 				Computed: true,
 				MaxItems: 1,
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_virtual_machine.go
```diff
@@ -71,6 +71,25 @@ func resourceArmVirtualMachine() *schema.Resource {
...
+			"identity": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:     schema.TypeString,
+							Required: true,
```
tombuildsstuff: it'd be good to add some validation (and a case-insensitive check) for this field, even if there's only one allowable value at this time:

```
ValidateFunc: validation.StringInSlice([]string { "SystemAssigned", true),
```

website/docs/r/virtual_machine.html.markdown
```diff
@@ -323,6 +324,36 @@ resource "azurerm_virtual_machine" "test" {
 3. Contains a numeric digit
 4. Contains a special character
 
+`identity` supports the following:
+
+* `type` - (Required) Specifies the identity type of the virtual machine. The only allowable value is `SystemAssigned`. To enable Managed Service Identity the ManagedIdentityExtension must also be added to the virtual machine. The Principal ID can be retrieved after the virtual machine has been created, e.g.
```
tombuildsstuff: _minor_ I think it'd be worth quoting the `ManagedIdentityExtension` extension name here

hbuckle:
tweaks added

gunzip:
Does MSI works for App Service as well ? (other than virtual machines).
What's the equivalent "virtual_machine_name" in this case ? Thank you for the hard work !

hbuckle:
@gunzip This is only for virtual machines - I hadn't noticed it was available for app service as well. Once this gets merged I can open a new PR for app service as well

azurerm/resource_arm_virtual_machine.go
```diff
@@ -647,6 +675,10 @@ func resourceArmVirtualMachineRead(d *schema.ResourceData, meta interface{}) err
 		}
 	}
 
+	if resp.Identity != nil {
+		d.Set("identity", flattenAzureRmVirtualMachineIdentity(resp.Identity))
+	}
```
tombuildsstuff: apologies, I've just noticed I didn't post the output of the acceptance tests in here. After running the VM acceptance tests - the acceptance tests without this field specified currently fail:

```
------- Stdout: -------
=== RUN   TestAccAzureRMVirtualMachine_tags
--- FAIL: TestAccAzureRMVirtualMachine_tags (416.59s)
    testing.go:459: Step 0 error: After applying this step, the plan was not empty:
        
        DIFF:
        
        UPDATE: azurerm_virtual_machine.test
          identity.#: "" => "<computed>"
        
        STATE:
...
```

In order to work around this - the field needs to be set to an empty object when there's no Identity returned - as such I've pushed a commit to fix this

tombuildsstuff: I've re-run all the VM tests and this now looks good - apologies for the delayed re-review here!

tombuildsstuff:
I've re-run the VM acceptance tests and these now pass:

screen shot 2017-12-04 at 11 43 01

tombuildsstuff:
@hbuckle apologies for the delay in re-reviewing this - but thanks for this PR this LGTM :)

pixelicous:
can we add this support for keyvaults as well?

PR #485, opened 2017-11-02
Clarify the error message for the validation of public IP domain name label

ricoli:
As per the regex, only lowercase characters are allowed.


tombuildsstuff (approved):
LGTM - thanks for this :)

Issue #464, opened 2017-10-27
Crash in terraform plan when comparing to nsg rule with IP list


clippermadness:
### Terraform Version
0.10.7

### Affected Resource(s)
- azurerm_network_security_group

### Expected Behavior
Should have tried to delete the rule, since it's in Azure but not in my config.
When I remove the rule from Azure, terraform plan does not crash.

### Actual Behavior
Crash:
```
2017-10-27T10:28:47.048-0700 [DEBUG] plugin.terraform-provider-azurerm_v0.3.1_x4:       {
2017-10-27T10:28:47.048-0700 [DEBUG] plugin.terraform-provider-azurerm_v0.3.1_x4:         "name": "AllowSpaceShackInboundToSSH",
2017-10-27T10:28:47.048-0700 [DEBUG] plugin.terraform-provider-azurerm_v0.3.1_x4:         "id": "/subscriptions/5e384bd8-c8ba-4040-b29d-4e37a8921981/resourceGroups/development/providers/Microsoft.Network/networkSecurityGroups/pipeline_pipelinetier_dev/securityRules/AllowSpaceShackInboundToSSH",
2017-10-27T10:28:47.048-0700 [DEBUG] plugin.terraform-provider-azurerm_v0.3.1_x4:         "etag": "W/\"473bd9ba-d56b-4d26-80bf-8984563c7042\"",
2017-10-27T10:28:47.048-0700 [DEBUG] plugin.terraform-provider-azurerm_v0.3.1_x4:         "properties": {
2017-10-27T10:28:47.048-0700 [DEBUG] plugin.terraform-provider-azurerm_v0.3.1_x4:           "provisioningState": "Succeeded",
2017-10-27T10:28:47.048-0700 [DEBUG] plugin.terraform-provider-azurerm_v0.3.1_x4:           "protocol": "TCP",
2017-10-27T10:28:47.048-0700 [DEBUG] plugin.terraform-provider-azurerm_v0.3.1_x4:           "sourcePortRange": "*",
2017-10-27T10:28:47.048-0700 [DEBUG] plugin.terraform-provider-azurerm_v0.3.1_x4:           "destinationPortRange": "22",
2017-10-27T10:28:47.048-0700 [DEBUG] plugin.terraform-provider-azurerm_v0.3.1_x4:           "destinationAddressPrefix": "*",
[...]

PR #492, opened 2017-11-03
Nil refence fix for issue #464

madenwala:
This nil check is needed to prevent a terraform crash that occurs when a user tries to update a network rule that was manually changed through the Azure Portal.   Fix is to check if properties on the props object is nil before you try to assign those string values to the  sgRule dictionary.

### DIFF
--- a/azurerm/resource_arm_network_security_group.go
+++ b/azurerm/resource_arm_network_security_group.go
@@ -226,10 +226,18 @@ func flattenNetworkSecurityRules(rules *[]network.SecurityRule) []interface{} {
 			sgRule["name"] = *rule.Name
 
 			if props := rule.SecurityRulePropertiesFormat; props != nil {
-				sgRule["destination_address_prefix"] = *props.DestinationAddressPrefix
-				sgRule["destination_port_range"] = *props.DestinationPortRange
-				sgRule["source_address_prefix"] = *props.SourceAddressPrefix
-				sgRule["source_port_range"] = *props.SourcePortRange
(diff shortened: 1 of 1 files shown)


thiagocaiubi:
Hi @madenwala . Thanks for the work!

Why this PR was self-approved and why it lacks unit tests?

madenwala:
Sorry about that, I'm new to this. I didn't know that I shouldn't submit the merge.  Not exactly sure how to create a test case around this as the crash only occurs if you manually edit the fields through the Azure portal and then try to run your TF script again.   The fix is very simple, just adding a nil check before you try to ready the property value.

tombuildsstuff (commented):
Taking a look at this retroactively, this change LGTM - however as discussed on Slack this shouldn't have been merged without a review and an acceptance test to cover it. I've enabled Branch Protection for the moment which requires a 👍 before merging - which should mean this can't happen in the future.

In addition - as discussed it'd be good to add an acceptance test for this in a follow-up PR - if you can reach out we can talk through the best way to achieve this :)

Thanks!

tombuildsstuff:
Hey @thiagocaiubi 

> Why this PR was self-approved and why it lacks tests?

Thanks for raising this - in addition to @madenwala's comments I've followed up internally and have enabled Branch Protection on the `master` branch (which requires a 👍 from an admin) which should avoid this issue going forward.

Regarding the missing tests - I've reached out to @madenwala and we're going to discuss the options here and add these in a follow up PR.

Thanks!

PR #493, opened 2017-11-03
remove illegal character from names

joshgav:
`%s` and `%d` were included in name strings, but these are not acceptable by Azure


tombuildsstuff:
Fixes #494

PR #508, opened 2017-11-03
simplifying azurerm_public_ip datasource sample

dcaro:
Suggesting a more simple example of use of the azurerm_public_ip data provider in the docs.

### DIFF
--- a/website/docs/d/public_ip.html.markdown
+++ b/website/docs/d/public_ip.html.markdown
@@ -15,134 +15,16 @@ Use this data source to access the properties of an existing Azure Public IP Add
 
 ```hcl
 data "azurerm_public_ip" "datasourceip" {
-    name = "testPublicIp"
-    resource_group_name = "acctestRG"
+    name = "name_of_public_ip"
+    resource_group_name = "name_of_resource_group"
 }
 
-resource "azurerm_virtual_network" "helloterraformnetwork" {
-    name = "acctvn"
-    address_space = ["10.0.0.0/16"]
-    location = "West US 2"
-    resource_group_name = "acctestRG"
+output "public_ip_address_id" {
+    domain_name_label = "${data.azurerm_public_ip.datasourceip.id}"
 }
 
-resource "azurerm_subnet" "helloterraformsubnet" {
-    name = "acctsub"
-    resource_group_name = "acctestRG"
-    virtual_network_name = "${azurerm_virtual_network.helloterraformnetwork.name}"
-    address_prefix = "10.0.2.0/24"
-}
-
-resource "azurerm_network_interface" "helloterraformnic" {
-    name = "tfni"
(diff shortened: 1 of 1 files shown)


website/docs/d/public_ip.html.markdown
```diff
@@ -15,134 +15,16 @@ Use this data source to access the properties of an existing Azure Public IP Add
...
-        name = "testconfiguration1"
-        subnet_id = "${azurerm_subnet.helloterraformsubnet.id}"
-        private_ip_address_allocation = "static"
-        private_ip_address = "10.0.2.5"
-        public_ip_address_id = "${data.azurerm_public_ip.datasourceip.id}"
-    }
-}
-```
-
-## Example Usage (Retrieve the Dynamic Public IP of a new VM)
```
tombuildsstuff: this is a topic which comes up fairly frequently - as such could we add this back in?

website/docs/d/public_ip.html.markdown
```diff
@@ -15,134 +15,16 @@ Use this data source to access the properties of an existing Azure Public IP Add
...
+    resource_group_name = "name_of_resource_group"
 }
 
-resource "azurerm_virtual_network" "helloterraformnetwork" {
-    name = "acctvn"
-    address_space = ["10.0.0.0/16"]
-    location = "West US 2"
-    resource_group_name = "acctestRG"
+output "public_ip_address_id" {
+    domain_name_label = "${data.azurerm_public_ip.datasourceip.id}"
```
tombuildsstuff: I may be wrong, but I think you're trying to output the domain name label rather than the ID (which isn't directly useful in an output IMO) - as such I think this needs to be:

```
output "domain_name_label" {
  value = "${data.azurerm_public_ip.datasourceip.domain_name_label}"
}
```

website/docs/d/public_ip.html.markdown
dcaro: Agree, we are planning to make a full article on docs on this topic to guide customers on how to use it.

tombuildsstuff: could we restore this sample for the moment - and then when the guide is published we should be able to replace this sample with a link to the guide?

dcaro: Fully agree, I am working on the docs this week and will update this PR when done. Will notify you.

tombuildsstuff: @dcaro any chance we could restore this second example? this otherwise looks good to me :)

tombuildsstuff:
hey @dcaro 

Thanks for pushing those updates, apologies for the delayed response here.

In order to get this merge I've incorporated these changes into #926 (which keeps, but simplifies the second example) - which I'm going to close this PR in favour of, I hope you don't mind!

Thanks!

PR #511, opened 2017-11-03
Editing the sample to add resourcegroup

VaijanathB:
Editing the sample to add resource group to make the example standalone.


VaijanathB:
Will send updated PR for this file as this requires unique name for traffic manager

PR #587, opened 2017-11-29
Updated the way user agent string gets assigned.

WodansSon:
Changed the way the user agent string gets assigned because it was over writing the Go SDK user agent information.

### DIFF
--- a/azurerm/config.go
+++ b/azurerm/config.go
@@ -198,9 +198,11 @@ func withRequestLogging() autorest.SendDecorator {
 }
 
 func setUserAgent(client *autorest.Client) {
-	// if the user agent already has a value append Terraform user agent string
-	if tfVersion := fmt.Sprintf("HashiCorp-Terraform-v%s", terraform.VersionString()); client.UserAgent != "" {
-		client.UserAgent = fmt.Sprintf("%s;%s", client.UserAgent, tfVersion)
+	tfVersion := fmt.Sprintf("HashiCorp-Terraform-v%s", terraform.VersionString())
+
+	// if the user agent already has a value append the Terraform user agent string
+	if curUserAgent := client.UserAgent; curUserAgent != "" {
+		client.UserAgent = fmt.Sprintf("%s;%s", curUserAgent, tfVersion)
 	} else {
 		client.UserAgent = tfVersion
 	}
--- a/azurerm/config.go
+++ b/azurerm/config.go
@@ -198,10 +198,14 @@ func withRequestLogging() autorest.SendDecorator {
 }
 
 func setUserAgent(client *autorest.Client) {
-	version := terraform.VersionString()
-	client.UserAgent = fmt.Sprintf("HashiCorp-Terraform-v%s", version)
+	// if the user agent already has a value append Terraform user agent string
+	if tfVersion := fmt.Sprintf("HashiCorp-Terraform-v%s", terraform.VersionString()); client.UserAgent != "" {
+		client.UserAgent = fmt.Sprintf("%s;%s", client.UserAgent, tfVersion)
+	} else {
+		client.UserAgent = tfVersion
+	}
 
-	// append the CloudShell version to the user agent
+	// append the CloudShell version to the user agent if it exists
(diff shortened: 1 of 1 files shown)


azurerm/config.go
```diff
@@ -198,10 +198,14 @@ func withRequestLogging() autorest.SendDecorator {
 }
 
 func setUserAgent(client *autorest.Client) {
-	version := terraform.VersionString()
-	client.UserAgent = fmt.Sprintf("HashiCorp-Terraform-v%s", version)
+	// if the user agent already has a value append Terraform user agent string
```
metacpp: This function seems to be a bit harder to read now, suggestions:

1. Put assignment statement in the if-clause, will it be harder for debugging coz one line of code contains too much statements ?
2. No matter tfversion or azureagent, you're appending it into user-agent, can we merge the code to avoid some kind of duplications ? Consider: what if we need to append even more fields here?
3. "HashiCorp-Terraform-v%s" seems to be a template/format string, shall we put it in constants ?
4. I can understand the second if-clause that you use assignment then checking coz they're related. But for the first if-clause, the assigment of tfVersion has nothing to do with the later checking, shall we move it to a separated line ?

metacpp (dismissed):

mcardosos:
LGTM

azurerm/config.go
jeffreyCline: 1. I did it that way because of comments from my previous PR where it was suggested to put the assignment of the variable in the if block.

2. I don't really see where it can get duplications, unless this code is called more than once, and if that happens I would consider that a bug. If you wanted to make sure there will not be duplications in the user agent string you should then also do a sub-string check before appending the text to the user-agent.

3. Yes it is a template string but it is only ever used once in the code, doesn't seem to be a good reason to make this a constant.

4. Yes, it can be it's own line, but as I stated above I did it this way because of my previous PR comments.

metacpp: #1/4, if a = xxx; a != "" is very good pattern coz it really saves one line and these 2 lines are contextual. While if b = xxx; a != "", i feel it's very weird. Who gave the suggestion ? tom ? We probably can sync with him...

#2, the duplication means the logic to append the string with ; as delimiter. If you're asked to append 10 more strings in the user-agent, you will not like to copy and paste. Anyway, it's just a suggestion.

#3 read-only and scope are different concepts, even its scope is function level, you still can use constants.

metacpp (approved):

metacpp (approved):

azurerm/config.go
```diff
@@ -198,10 +198,16 @@ func withRequestLogging() autorest.SendDecorator {
...
 
 func setUserAgent(client *autorest.Client) {
-	version := terraform.VersionString()
-	client.UserAgent = fmt.Sprintf("HashiCorp-Terraform-v%s", version)
+	tfVersion := fmt.Sprintf("HashiCorp-Terraform-v%s", terraform.VersionString())
 
-	// append the CloudShell version to the user agent
+	// if the user agent already has a value append the Terraform user agent string
+	if curUserAgent := client.UserAgent; curUserAgent != "" {
+		client.UserAgent = fmt.Sprintf("%s;%s", curUserAgent, tfVersion)
```
tombuildsstuff: Given this function is called in multiple places - have you tested this behaviour and checked the outputs?

tombuildsstuff (approved):
This LGTM - I'm assuming you've confirmed this behaviour via an HTTP Proxy of some sort? This methods called multiple times (which is my only concern)

azurerm/config.go
jeffreyCline: Yes I have, you can see the results [here](https://gist.github.com/jeffreyCline/ba277c2779cae20a72199f5cb6bf5f07). It works as expected without any duplicate strings added. However, I question why it gets called 164 times when provisioning a single object.

tombuildsstuff: 👍 cool - AFAIK that's due to us setting the user agent for each API Client (such as a SQLDBClient/CosmosDBClient/AppServiceClient etc) that we instantiate when the Provider Plugin's initialised

PR #591, opened 2017-11-29
Add "kind" field to app service.

JunyiYi:
Enable "kind" field in app service.
Now you are able to create FunctionsApp (but you still need to care about other configurations).

### DIFF
--- a/azurerm/resource_arm_app_service.go
+++ b/azurerm/resource_arm_app_service.go
@@ -33,6 +33,16 @@ func resourceArmAppService() *schema.Resource {
 
 			"location": locationSchema(),
 
+			"kind": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"functionapp",
+				}, true),
+				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+			},
+
 			"app_service_plan_id": {
 				Type:     schema.TypeString,
 				Required: true,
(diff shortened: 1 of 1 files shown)


azurerm/resource_arm_app_service.go
```diff
@@ -33,6 +33,16 @@ func resourceArmAppService() *schema.Resource {
 
 			"location": locationSchema(),
 
+			"kind": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"functionapp",
```
tombuildsstuff: we should add all the other values in here too (from memory there's 4 values - `functionapp`, `mobileapi` and `app` and something else?) - there should be constants for these in the Azure SDK.

azurerm/resource_arm_app_service.go
```diff
@@ -33,6 +33,16 @@ func resourceArmAppService() *schema.Resource {
 
 			"location": locationSchema(),
 
+			"kind": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
```
tombuildsstuff: we should add a Default value for this for existing users, e.g. "app" (which appears to be the default for a Web App/Website/App Service)

JunyiYi: Sure. And where did your find those available values? My knowledge about "functionapp" is provided by the Azure Functions team.

azurerm/resource_arm_app_service.go
JunyiYi: Will do if I could confirm the available values.

azurerm/resource_arm_app_service.go
tombuildsstuff: from memory creating an App Service / Function App / Mobile API in the portal and either seeing it in the output of resources.azure.com / API responses / the export template function - although they may also be in the Go SDK

JunyiYi:
Thanks @tombuildsstuff ,

This is the way recommended by Azure Functions team: "Set kind to functionapp, and remember to configure the required storage as well as connection string app settings." And there are no dedicated SDK calls or API's for Azure Functions right now.

I agree that users should not care about the detailed difference between different "app"s, so based on this, we would also like to add a new module maybe called azure functions, and users should use that module to provision an azure functions app. And the provider here will just provide the essential building blocks.

Regards.

PR #594, opened 2017-12-01
Upgrading to v11.2.2-beta of the Azure SDK for Go

tombuildsstuff:


vendor/github.com/Azure/azure-sdk-for-go/arm/appinsights/components.go
```diff
@@ -98,7 +98,9 @@ func (client ComponentsClient) CreateOrUpdatePreparer(resourceGroupName string,
 // CreateOrUpdateSender sends the CreateOrUpdate request. The method will close the
 // http.Response Body if it receives an error.
 func (client ComponentsClient) CreateOrUpdateSender(req *http.Request) (*http.Response, error) {
-	return autorest.SendWithSender(client, req)
+	return autorest.SendWithSender(client,
+		req,
+		azure.DoRetryWithRegistration(client.Client))
```
tombuildsstuff: we've got a variable [to opt-out of Resource Provider registration](https://www.terraform.io/docs/providers/azurerm/index.html#skip_provider_registration) - which we use to skip checking for the Resource Provider Registration.

This method will attempt to register the Resource Provider in the case that it's unregistered, which means it has a different behaviour to the variable listed above - however I'm going to open an issue on Go-AutoRest asking for a means of failing when Resource Providers aren't registered, rather than registering - so that we can still support this use-case.

I've opened an issue on Go-AutoRest about this: https://github.com/Azure/go-autorest/issues/210

mbfrahry (approved):
LGTM!

vendor/github.com/Azure/azure-sdk-for-go/arm/appinsights/components.go
tombuildsstuff: There's a feature toggle for this coming in v12 - but there's no harm in shipping this as-is.

Running the test suite now..

tombuildsstuff: Tests seem good 👍

justaugustus:
Hey @tombuildsstuff, do we have an ETA for this making it into the next release?

tombuildsstuff:
@justaugustus I'm hoping to do a release sometime this week - stay tuned :)

PR #597, opened 2017-12-02
Adding Azure Functions Support

TsuyoshiUshio:
Hello!

I add an Azure Functions  support for App Service Resource. 

I tested in my local machine and execute `make build` and `make test`. 
I add some code, tests, and add some documentation.

I wrote whole tests, However, I haven't tested `make testacc` since it said that it costs a lot. 
If this pull request is valuable, could you try the test, please. 

If it has any problem, I'd happy to fix the issue. 

I hope this helps.


tombuildsstuff:
Hey @TsuyoshiUshio 

Thanks for this PR - both yourself and @JunyiYi submitted support for this functionality in PR #591.

[In that PR](https://github.com/terraform-providers/terraform-provider-azurerm/pull/591#pullrequestreview-80135269) we discussed making Azure Functions a separate resource, rather than just adding the `kind` field to App Service and loosening the validation. The main benefit for doing this is that it allows us to use validation specific to both Functions and App Services - rather than making the resources generic (and optional) - which makes the resources more useful to end-users.

For this reason I'm going to close this PR - however this PR otherwise looks good and I'd like to thank you for this contribution :)

Thanks!

PR #598, opened 2017-12-04
Support global parallelism testing through environment variable.

metacpp:
1. Support parallel testing for all acceptance tests through TF_PARL.
2. Update the makefile to support run acc test cases with 20 threads.


GNUmakefile
```diff
@@ -1,18 +1,31 @@
...
 		xargs -t -n4 go test $(TESTARGS) -timeout=30s -parallel=4
 
-testacc: fmtcheck
-	TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 180m
+# example usage:
+#  1. Enable parallel run for all acceptance tests with default number(20) of
+#  threads:
+#    TF_PARA=1 make testacc
+#  2. Enable parallel run for acceptance tests with matching pattern.
+#    TF_PARA=1 TESTARGS="-run TestAccAzureRMStorage" make testacc
```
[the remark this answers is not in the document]
metacpp: Agree, will update it soon.

GNUmakefile
```diff
@@ -1,18 +1,31 @@
...
+testacc: fmtcheck test-install
```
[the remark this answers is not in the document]
metacpp: I am not the original author of test/testacc here. From their original definitions:
test: fmtcheck, test-install, unit-test
testacc: fmtcheck, acc-test.

What I changed here is just add an extra step, test-install, into testacc. Do you want me to include the unit test into acc test ? I checked the resource.Test, it will run unit tests by default, but will skip acc tests if TF_ACC is not set.

azurerm/provider_test.go
```diff
@@ -12,6 +12,9 @@ import (
 	"github.com/hashicorp/terraform/terraform"
 )
 
+// ParaTestEnvVar is used to enable parallelism in testing.
+const ParaTestEnvVar = "TF_PARA"
```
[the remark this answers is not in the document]
metacpp: I was struggling the naming principles here:
1. TF_PARALLEL: clear but a bit longer while using it the command line.
2. TF_PARA: short to type in command line, but as you said, not quite clear, might confuse people.

Then I took a look at TF_ACC, I decided to use #2 to keep just one naming principle in the same source file.

@jeffreyCline , @JunyiYi what do you think ? I am open to use either of them.

azurerm/provider_test.go
```diff
@@ -48,6 +51,17 @@ func testAccPreCheck(t *testing.T) {
...
+	// If "TF_PARA" is set to some non-empty value, all the parallel test cases
+	// will be run in parallel. It requires that the test cases are all
+	// independent ones without any shared resource before turn it on.
+	// example usage:
+	//  TF_ACC=1 TF_PARA=1 go test [TEST] [TESTARGS] -v -parallel=n
+	// To run specified cases in parallel, please refer to below PR:
+	// 	https://github.com/hashicorp/terraform/pull/16807
+	if os.Getenv(ParaTestEnvVar) != "" {
+		t.Parallel()
+	}
```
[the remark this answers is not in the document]
metacpp: I agree with you, that's why I added similar flag in this [PR](https://github.com/hashicorp/terraform/pull/16807), but since resource's test suite is widely used by many different providers, I set its default value be false since I don't want to break any legacy code. The current test cases in azurerm provider have very strong dependencies in resource.Test and resource.TestCase, if we want to introduce another different flag(RunInParallel) on this provider only, we probably need to introduce a middle layer to handle this.

For the acc tests in "resource_arm_network_watcher_test.go", I am curious that how you handle it today in TeamCity CI coz it's also parallel running environment. Can we use subtests to group the current test cases for network watcher ? By this way, we can still keep it running with other tests in parallel, but its subtests run in sequentially.

azurerm/provider_test.go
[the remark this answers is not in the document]
metacpp: https://www.allacronyms.com/parallel/abbreviated

We decided to use "TF_PARL" instead according to above link.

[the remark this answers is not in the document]
tombuildsstuff: @metacpp whilst I understand that - we tend to be more descriptive in the environment variables we use - such as:

- `TF_WORKSPACE`
- `TF_IN_AUTOMATION`
- `TF_SKIP_PROVIDER_VERIFY`
- `TF_PLUGIN_CACHE_DIR`

As such I think this is best as `TF_PARALLEL` for consistency with the rest of Terraform - given this is an opt-in setting; when setting this as an environment variable this can be set once and forgotten (like the Provider environment variables are - as such, I don't think adding a couple of extra characters to be readable is a bad thing).

README.md
```diff
@@ -95,7 +95,11 @@ $ make test
...
 
 *Note:* Acceptance tests create real resources, and often cost money to run.
+TF_PARL is used to run all tests in parallel, TF_PARL_NUM is used to define the 
+number of paralleled tests to run, and TESTARGS is used to pass extra
+testing arguments like below.
+
 
 ```sh
-$ make testacc
+$ [TF_PARL=1] [TF_PARL_NUM=20] [TESTARGS="-run TestAccAzureRMStorage"] make testacc
```
tombuildsstuff: I think it's worth splitting this into two separate examples - one as it was (which is the default/recommended) and a second which shows an example usage in parallel (as above) - given I think most contributors will end up running these individually for cost/debugging reasons?

azurerm/resource_arm_network_watcher_test.go
```diff
@@ -14,7 +14,7 @@ func TestAccAzureRMNetworkWatcher_basic(t *testing.T) {
 	resourceGroup := "azurerm_network_watcher.test"
 	rInt := acctest.RandInt()
 	resource.Test(t, resource.TestCase{
```
tombuildsstuff: rather than shoe-horning this into the `testAccPreCheck` method - I think it might be worth overriding/sub-classing `resource.TestCase` to expose a `RunInParallel` field as per the original PR - to match the design of the API? e.g.

```
resource.Test(t, resource.TestCase{
  PreCheck: func() {},
  Providers: testAccProviders,
  RunInParallel: false,
  CheckDestroy: #...
})
```

GNUmakefile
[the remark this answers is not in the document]
tombuildsstuff: > What I changed here is just add an extra step, test-install, into testacc.

I was thinking more of removing `test-install` from `testacc` - as it's not needed since we should be running `make test` first anyway (which will catch this). The benefit to having these split out is it's easier to expose what's failed (i.e. unit/acc tests), rather than just "some tests failed" - if that makes sense?

```
make test
make acctest
```

> Do you want me to include the unit test into acc test ? I checked the resource.Test, it will run unit tests by default, but will skip acc tests if TF_ACC is not set.

No that's fine - if we can just remove `test-install` this should be fine :)

azurerm/provider_test.go
```diff
@@ -49,6 +52,21 @@ func testAccPreCheck(t *testing.T) {
 			t.Fatalf("`%s` must be set for acceptance tests!", variable)
 		}
 	}
+
+	// If "TF_PARL" is set to some non-empty value, all the parallel test cases
+	// will be run in parallel. It requires that the test cases are all
+	// independent ones without any shared resource before turn it on.
```
tombuildsstuff: I'd argue the second sentence in this comment is incorrect since it's not the reason why we can't run these tests in Parallel - which are mostly due to limitations in Azure (either Rate Limiting, or Usage Limits [such as only a single instance in a region w/Network Watcher])

azurerm/provider_test.go
```diff
@@ -49,6 +52,21 @@ func testAccPreCheck(t *testing.T) {
...
+	// example usage:
+	//  TF_ACC=1 TF_PARL=1 go test [TEST] [TESTARGS] -v -parallel=n
+	// To run specified cases in parallel, please refer to below PR:
+	// 	https://github.com/hashicorp/terraform/pull/16807
```
tombuildsstuff: I don't think we need these 4 lines, since they're included in the README?

README.md
metacpp: Parallel running will not make user pay more, the size of test package per run is the key factor, and user can use TESTARGS to filter out necessary cases to run.

I updated the README with 3 common usage cases, hope it's helpful to end user.

azurerm/provider_test.go
metacpp: Agree, updated the comments and the code for better reading.

azurerm/provider_test.go
metacpp: Different levels of details. README is based on Makefile while this is based on go test.

azurerm/resource_arm_network_watcher_test.go
metacpp: @tombuildsstuff We don't want to change 600+ test cases now, that's one of the biggest concern since the beginning of this PR.

In another [PR](https://github.com/hashicorp/terraform/pull/16807/files), we discussed the pain point of 600+ test cases, and @vancluever suggested me to change testAccPreCheck to inject parallelism logic, since this is the smallest change but achieving the goal in a very short time.

We do have plan to add a middle layer, aztesting, to extend functionalities of resource.Test, especially if Terraform core runtime team don't want to add some features into core. That will take even more time to finish the work, I suggest we finish this PR asap considering it's been pending for almost 2 weeks and we're blocked by it to enable Jenkins CI on our side as monitoring, and at the same time, we start another thread to do the code refactoring.

README.md
tombuildsstuff: my point was more the user has the option to cancel them out; TESTARGS is a prefix match; and most tests follow the pattern:
 - `foo_basic`
 - `foo_basicUpdated`

meaning in order to run `foo_basic` using the `TESTARGS` of `-run=foo_basic` - the test `foo_basicUpdate` will also be run (leading to extra resources/costs) when in parallel, rather than being able to be cancelled via ctrl+c - it's not a big deal, but it's one to be aware of

tombuildsstuff:
After some investigation and discussing this internally we've discovered this approach won't work without some major refactoring due to the way that logging works (for both HTTP Requests/Responses and `log.Printf`'s). Instead for the moment we're opting to build a runner [similar to the one we use for TeamCity](http://github.com/jen20/teamcity-go-test) but for JUnit to be able to output the logs as needed in both Jenkins and VSTS/TFS.

Whilst hashicorp/terraform#16356 goes some way towards solving this - it's only a band-aid over the problem until Parallelisation is supported natively (since it's writing them to external files per test, rather than within each test blocks) - once this is supported we can look to using the native parallelisation support (once logging is tracked on a per-test basis).

As such I'm going to close this PR - but thanks for this contribution!

PR #630, opened 2017-12-14
Skipping Resource Provider Registration if specified

tombuildsstuff:
AutoRest 9.5.2 exposes a new property to skip Resource Provider Registration if the Resource Providers isn't registered. This hooks up the existing property to each client to ensure the variable is configured as needed (meaning that if a user doesn't have permission to register the Resource Providers - we won't attempt to do so)

Also includes some continued refactoring of the `client.go` file to organise things into groups

I'd probably recommend viewing this by commit


paultyng (approved):

tombuildsstuff:
Tests look good - merging this in and re-running the suite..

PR #638, opened 2017-12-16
Fixes #637 - adds tags to azurerm_local_network_gateway

### LINKED ISSUE #637: azurerm_local_network_gateway does not support tags
The azurerm_local_network_gateway resource does not support tags even though the Azure SDK supports tags for Local Network Gatways. I will be submitting a pull request shortly that will address this issue.

russellds:
This pull request adds tags to azurerm_local_network_gateway. This addresses issue #637.


russellds:
@tombuildsstuff Thank you! I'll take a look at what you changed to update the documentation so in the future I can make sure I do that as well!

PR #639, opened 2017-12-18
Allow authentication via managed service identity

hbuckle:
Adds a provider option to allow authentication using managed service identity
e.g.
```
provider "azurerm" {
  subscription_id = "xxx"
  tenant_id       = "xxx"
  use_msi         = true
}
```


paultyng:
@hbuckle Code wise this looks good to me. I have no experience with MSI though. We have some additional authentication changes coming targeted at 1.0.3 so grouping this up with those.

paultyng:
It looks like this has a minor merge conflict, could maybe use a rebase.

VaijanathB:
@hbuckle , 

Am I understanding correctly that this only supports MSI with ServicePrincipal and does not do it with azure cli ? Also I pulled your branch and tried to verify using the construct specified 

provider "azurerm" {
  subscription_id = "xxx"
  tenant_id       = "xxx"
  use_msi         = true
}
and was not able to authenticate. Can you please elaborate how to do it ? 

Thanks,
VJ

hbuckle:
@VaijanathB MSI works by creating a service principal to act on behalf of an Azure resource (such as a VM). For it to work you need to be running Terraform from a VM in Azure that has MSI configured - see authenticating_via_msi.html.markdown for details

azurerm/provider.go
```diff
@@ -196,11 +207,27 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
...
 			TenantID:                  d.Get("tenant_id").(string),
 			Environment:               d.Get("environment").(string),
+			UseMsi:                    d.Get("use_msi").(bool),
+			MsiEndpoint:               d.Get("msi_endpoint").(string),
 			SkipCredentialsValidation: d.Get("skip_credentials_validation").(bool),
 			SkipProviderRegistration:  d.Get("skip_provider_registration").(bool),
 		}
 
-		if config.ClientSecret != "" {
+		if config.UseMsi {
```
metacpp: Shall we use the same if-else logic as in the azurerm/config.go ?

website/docs/authenticating_via_msi.html.markdown
```diff
@@ -0,0 +1,97 @@
...
+Managed Service Identity allows an Azure virtual machine to retrieve a token to access the Azure API without needing to pass in credentials. This works by creating a service principal in Azure Active Directory that is associated to a virtual machine. This service principal can then be granted permissions to Azure resources.
+There are various ways to configure managed service identity - see the [Microsoft documentation](https://docs.microsoft.com/en-us/azure/active-directory/msi-overview) for details.
+You can then run Terraform from the MSI enabled virtual machine by setting the use_msi provider option to true.
+
+### Configuring Managed Service Identity using Terraform
+
+Managed service identity can also be configured using Terraform. The following template shows how. Note that for a Linux VM you must use the ManagedIdentityExtensionForLinux.
+
+```hcl
+resource "azurerm_virtual_machine" "virtual_machine" {
```
VaijanathB: Can you put a Linux example ? This example creates WindowsServer and the documentation above says about Linux.

azurerm/provider.go
hbuckle: Do you mean from getAuthorizationToken ? I think the logic is slightly different in each

website/docs/authenticating_via_msi.html.markdown
hbuckle: As noted, for Linux it would be exactly the same except you use ManagedIdentityExtensionForLinux instead of ManagedIdentityExtensionForWindows extension

azurerm/provider.go
metacpp: Yes, I mean why we don't check if ClientSecret exists firstly.

azurerm/provider.go
```diff
@@ -63,6 +64,16 @@ func Provider() terraform.ResourceProvider {
...
 			},
+			"use_msi": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("ARM_USE_MSI", false),
+			},
+			"msi_endpoint": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("ARM_MSI_ENDPOINT", ""),
```
metacpp: Can we change the default environment variable to "AZURE_MSI_ENDPOINT", coz this endpoint is not just for getting tokens for ARM, but also for graph/datalake/keyvault.

azurerm/provider.go
```diff
@@ -63,6 +64,16 @@ func Provider() terraform.ResourceProvider {
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("ARM_SKIP_PROVIDER_REGISTRATION", false),
 			},
+			"use_msi": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("ARM_USE_MSI", false),
```
metacpp: Change to "USE_MSI".

tombuildsstuff: given all the other Environment Variables are prefixed with `ARM` - we should leave this as `ARM_USE_MSI` for consistency (we may change this for `AZURE_USE_MSI` in future, but it should be consistent for now)

azurerm/provider.go
tombuildsstuff: (same here) - given the prefix for the other Environment Variables is `ARM` - we'd be best to leave this as `ARM_MSI_ENDPOINT` rather than making this `AZURE_MSI_ENDPOINT` - in future we can migrate across to use `AZURE` as a prefix if needed)

metacpp: Agree.

azurerm/provider.go
metacpp: Agree.

website/docs/authenticating_via_msi.html.markdown
```diff
@@ -0,0 +1,97 @@
...
+
+Terraform supports authenticating to Azure through Managed Service Identity, Service Principal or the Azure CLI.
+
+We recommend using Managed Service Identity when running in a Shared Environment (such as within a CI server/automation) that you do not wish to configure credentials for - and [authenticating via the Azure CLI](authenticating_via_azure_cli.html) when you're running Terraform locally. Note that managed service identity is only available for virtual machines within Azure.
+
+## Configuring Managed Service Identity
+
+Managed Service Identity allows an Azure virtual machine to retrieve a token to access the Azure API without needing to pass in credentials. This works by creating a service principal in Azure Active Directory that is associated to a virtual machine. This service principal can then be granted permissions to Azure resources.
+There are various ways to configure managed service identity - see the [Microsoft documentation](https://docs.microsoft.com/en-us/azure/active-directory/msi-overview) for details.
+You can then run Terraform from the MSI enabled virtual machine by setting the use_msi provider option to true.
```
tombuildsstuff: can we quote `use_msi` and `true` here?

metacpp:
@hbuckle any update ?

azurerm/provider.go
hbuckle: @tombuildsstuff by leaving it blank then the endpoint should be auto-discovered using GetMSIVMEndpoint(). I only put the option there in case the endpoint can't be discovered (such as when running in a container)

tombuildsstuff:
Results of testing this so far:

- ❌ fails in CloudShell
- ✅ MSI on a Windows VM

Still to test:

- ❔ MSI on a Linux VM

tombuildsstuff:
hey @hbuckle 

Just to give an update here - I've been working through testing this, and this looks good on the VM's (both Linux and Windows), however there's an issue when running in CloudShell (where MSI's recently been enabled). In CloudShell the API doesn't return the same schema as the API available when running in a VM (an integer value is returned, rather than a string) for the `expiresOn` field, which means deserialization fails.

Since MSI is enabled by default in CloudShell - I've reached out to Microsoft to see how if they can patch the API to return the correct data type - if that's not going to be possible for a while the other option would be to error when running in CloudShell with MSI enabled, until we've confirmed the API's been patched (which would allow us to ship this for use in VM's).

Apologies for the delay, but bear with us here - I'm hoping we'll have an answer about this later this week :)

Thanks!

azurerm/helpers/authentication/validation.go
```diff
@@ -49,3 +49,22 @@ func (c *Config) ValidateServicePrincipal() error {
 
 	return err.ErrorOrNil()
 }
+
+func (c *Config) ValidateMsi() error {
+	var err *multierror.Error
+
+	if c.SubscriptionID == "" {
```
metacpp: How can we get the default subscription info here like populateSubscriptionFromCLIProfile in MSI ?

metacpp:
@tombuildsstuff @hbuckle if subscription_id and tenant_id are not provided in the provider configuration, we need to get them from environment: variable and profile, especially the cloud shell environment.

tombuildsstuff:
:wave: hey @hbuckle 

Just to let you know that support for MSI has just been released in v1.2.0 of the AzureRM Provider - full details of what's included are available here: https://github.com/terraform-providers/terraform-provider-azurerm/blob/v1.2.0/CHANGELOG.md#120-march-02-2018

Thanks!

PR #642, opened 2017-12-19
Add name_regex to azurerm_image data source

### LINKED ISSUE #454: Extend azurerm_image to support searching for an image
It would be very useful to modify azurerm_image to work like aws_ami and be able to at least do some basic filtering to return a particular image.  Even just a name prefix search with support for 'most_recent' would allow e.g. maintaining a series of images with Packer and always deploying with the latest one.

Thanks!

paultyng:
Fixes #454

### DIFF
--- a/azurerm/data_source_image.go
+++ b/azurerm/data_source_image.go
@@ -2,8 +2,13 @@ package azurerm
 
 import (
 	"fmt"
+	"log"
+	"regexp"
+	"sort"
 
+	"github.com/Azure/azure-sdk-for-go/arm/compute"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
@@ -12,9 +17,23 @@ func dataSourceArmImage() *schema.Resource {
 		Read: dataSourceArmImageRead,
 		Schema: map[string]*schema.Schema{
 
+			"name_regex": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ValidateFunc:  validation.ValidateRegexp,
+				ConflictsWith: []string{"name"},
+			},
+			"sort_descending": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
 			"name": {
-				Type:     schema.TypeString,
-				Required: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"name_regex"},
 			},
 
 			"resource_group_name": resourceGroupNameForDataSourceSchema(),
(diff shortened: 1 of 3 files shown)


azurerm/data_source_image_test.go
```diff
@@ -169,3 +246,163 @@ output "location" {
...
+
+data "azurerm_image" "test1" {
+  name_regex          = "^def-acctest-\\d+"
+  resource_group_name = "${azurerm_resource_group.test.name}${substr(azurerm_image.abc.name, 0, 0)}${substr(azurerm_image.def.name, 0, 0)}"
+}
+
+data "azurerm_image" "test2" {
+  name_regex          = "^[a-z]+-acctest-\\d+"
+  name_descending     = true
+  resource_group_name = "${azurerm_resource_group.test.name}${substr(azurerm_image.abc.name, 0, 0)}${substr(azurerm_image.def.name, 0, 0)}"
```
paultyng: This was a hack to deal with `depends_on` and data source issues, is there a better way to do this?

paultyng:
image

website/docs/d/image.html.markdown
```diff
@@ -26,11 +26,14 @@ output "image_id" {
 
 ## Argument Reference
 
-* `name` - (Required) The name of the Image.
+* `name` - (Optional) The name of the Image.
+* `name_regex` - (Optional) Regex pattern of the image to match.
+* `name_descending` - (Optional) By default when matching by regex, images are sorted by name in ascending order and the first match is chosen, to sort descending, set this flag.
```
tombuildsstuff: would it be instead worth making this a `most_recent` flag as in the AWS data source?

azurerm/data_source_image_test.go
tombuildsstuff: alas not really - some tests create the config in one step then apply the data source in another; I think there's a core bug about that?

azurerm/data_source_image_test.go
```diff
@@ -169,3 +246,163 @@ output "location" {
...
+
+resource "azurerm_image" "def" {
+  name                = "def-acctest-%d"
+  location            = "${azurerm_resource_group.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+
+  os_disk {
+    os_type  = "Linux"
+    os_state = "Generalized"
+    blob_uri = "${azurerm_storage_account.test.primary_blob_endpoint}${azurerm_storage_container.test.name}/myosdisk1.vhd"
```
tombuildsstuff: I think we should be able to make this `${azurerm_virtual_machine.storage_os_disk.0.blob_uri}`?

azurerm/data_source_image_test.go
```diff
@@ -169,3 +246,163 @@ output "location" {
...
+
+resource "azurerm_image" "abc" {
+  name                = "abc-acctest-%d"
+  location            = "${azurerm_resource_group.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+
+  os_disk {
+    os_type  = "Linux"
+    os_state = "Generalized"
+    blob_uri = "${azurerm_storage_account.test.primary_blob_endpoint}${azurerm_storage_container.test.name}/myosdisk1.vhd"
```
tombuildsstuff: I think we should be able to make this `${azurerm_virtual_machine.storage_os_disk.0.blob_uri}` - which means we could remove the `depends_on`?

azurerm/data_source_image.go
```diff
@@ -12,9 +16,21 @@ func dataSourceArmImage() *schema.Resource {
 		Read: dataSourceArmImageRead,
 		Schema: map[string]*schema.Schema{
 
+			"name_regex": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateNameRegex,
```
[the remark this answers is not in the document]
tombuildsstuff: I might be wrong - but I think there's a helper that does this in the `validation` namespace?

azurerm/data_source_image.go
```diff
@@ -92,37 +108,103 @@ func dataSourceArmImageRead(d *schema.ResourceData, meta interface{}) error {
...
+
+		if len(list) < 1 {
 			d.SetId("")
 			return nil
 		}
-		return fmt.Errorf("[ERROR] Error making Read request on Azure Image %q (resource group %q): %+v", name, resGroup, err)
+
+		if len(list) > 1 {
+			desc := d.Get("name_descending").(bool)
+			log.Printf("[DEBUG] arm_image - multiple results found and `name_descending` is set to: %t", desc)
```
tombuildsstuff: for the AWS provider we fail in this case unless `most_recent` is specified - is it worth mirroring that behaviour?

azurerm/data_source_image.go
```diff
@@ -92,37 +108,103 @@ func dataSourceArmImageRead(d *schema.ResourceData, meta interface{}) error {
...
+
+			sort.Slice(list, func(i, j int) bool {
```
tombuildsstuff: it'd be good to add an acceptance test for this behaviour too?

paultyng:
image

azurerm/data_source_image.go
```diff
@@ -12,9 +17,23 @@ func dataSourceArmImage() *schema.Resource {
...
 		Schema: map[string]*schema.Schema{
 
+			"name_regex": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ValidateFunc:  validation.ValidateRegexp,
+				ConflictsWith: []string{"name"},
+			},
+			"name_descending": {
```
tombuildsstuff: after chatting with @radeksimko - can we rename this `sort_descending`? it's possible we'll filter by additional fields in the future

paultyng:
```
=== RUN   TestAccDataSourceAzureRMImage_basic
--- PASS: TestAccDataSourceAzureRMImage_basic (492.36s)
=== RUN   TestAccDataSourceAzureRMImage_localFilter
--- PASS: TestAccDataSourceAzureRMImage_localFilter (528.13s)
```

PR #644, opened 2017-12-20
Prepending `Data Source: ` to the Data Source docs

tombuildsstuff:
Prepending `Data Source: ` to the Data Source docs for consistency with the changes made in terraform-providers/terraform-provider-aws#2713


paultyng:
LGTM, I assume markdown doesn't need those escapes any longer or something? Or maybe never needed them?

tombuildsstuff:
@paultyng indeed, it was done for consistency with the other providers - but they're not needed so this is gradually being removed in different providers

Issue #579, opened 2017-11-21
Feature Request: Add support for new Preview features (LB Standard, Public IP Standard, Availability Zones, High Availability Ports)


justaugustus:
This is a bucket issue for adding support for the following upcoming Azure Preview features:
- [x] Azure LB / Public IP Standard (https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-standard-overview): https://github.com/terraform-providers/terraform-provider-azurerm/pull/665
- [ ] Availability Zones (https://docs.microsoft.com/en-us/azure/availability-zones/az-overview)
- [ ] High Availability Ports (https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-ha-ports-overview)
- [ ] Remove Preview notices on associated docs

We (CoreOS) would like to add support for this feature set in Tectonic, but it will be dependent on formal upstream support for the features in the [Azure SDK for Go](https://github.com/Azure/azure-sdk-for-go) and the [azurerm TF provider](https://github.com/terraform-providers/terraform-provider-azurerm).

x-ref: coreos/tectonic-installer#2443

Upstream refs:
* Azure/azure-sdk-for-go#871

cc: @robszumski / @metral

tombuildsstuff:
hey @justaugustus 

Taking another look through this issue it appears the remaining feature requests are duplicates of #370 (for Availability Zones) and #372 (High Availability Ports) - as such I'm going to close this combined issue in favour of the specific ones - just so it's easier to track, I hope you don't mind!

Thanks!

justaugustus:
Sounds good, @tombuildsstuff. Thanks for the heads-up!

PR #665, opened 2018-01-05
Add support for Azure LB Standard and Public IP Standard

justaugustus:
Ref: #579

### DIFF
--- a/azurerm/resource_arm_loadbalancer.go
+++ b/azurerm/resource_arm_loadbalancer.go
@@ -100,6 +100,11 @@ func resourceArmLoadBalancer() *schema.Resource {
 				},
 			},
 
+			"sku": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
 			"tags": tagsSchema(),
 		},
 	}
@@ -114,6 +119,7 @@ func resourceArmLoadBalancerCreate(d *schema.ResourceData, meta interface{}) err
 	name := d.Get("name").(string)
 	location := d.Get("location").(string)
 	resGroup := d.Get("resource_group_name").(string)
+	sku := convertSkuStringToLoadBalancerSku(d.Get("sku").(string))
 	tags := d.Get("tags").(map[string]interface{})
 	expandedTags := expandTags(tags)
 
@@ -127,6 +133,7 @@ func resourceArmLoadBalancerCreate(d *schema.ResourceData, meta interface{}) err
 		Name:     utils.String(name),
 		Location: utils.String(location),
 		Tags:     expandedTags,
+		Sku:      &sku,
 		LoadBalancerPropertiesFormat: &properties,
 	}
 
@@ -238,6 +245,17 @@ func resourceArmLoadBalancerDelete(d *schema.ResourceData, meta interface{}) err
 	}
 
 	return nil
+}
+
+func convertSkuStringToLoadBalancerSku(s string) network.LoadBalancerSku {
+	skuNames := map[string]network.LoadBalancerSkuName{
+		"basic":    network.LoadBalancerSkuNameBasic,
+		"standard": network.LoadBalancerSkuNameStandard,
+	}
+
+	return network.LoadBalancerSku{
+		Name: skuNames[s],
+	}
 }
 
 func expandAzureRmLoadBalancerFrontendIpConfigurations(d *schema.ResourceData) *[]network.FrontendIPConfiguration {
--- a/azurerm/resource_arm_loadbalancer.go
+++ b/azurerm/resource_arm_loadbalancer.go
@@ -205,6 +205,10 @@ func resourecArmLoadBalancerRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("location", azureRMNormalizeLocation(*location))
 	}
 
+	if sku := loadBalancer.Sku; sku != nil {
+		d.Set("sku", string(sku.Name))
+	}
+
 	if props := loadBalancer.LoadBalancerPropertiesFormat; props != nil {
 		if feipConfigs := props.FrontendIPConfigurations; feipConfigs != nil {
 			d.Set("frontend_ip_configuration", flattenLoadBalancerFrontendIpConfiguration(feipConfigs))
--- a/azurerm/resource_arm_loadbalancer.go
+++ b/azurerm/resource_arm_loadbalancer.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
@@ -102,7 +103,14 @@ func resourceArmLoadBalancer() *schema.Resource {
 
 			"sku": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
+				Default:  string(network.LoadBalancerSkuNameBasic),
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(network.LoadBalancerSkuNameBasic),
+					string(network.LoadBalancerSkuNameStandard),
+				}, true),
+				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
 			},
 
 			"tags": tagsSchema(),
@@ -119,7 +127,9 @@ func resourceArmLoadBalancerCreate(d *schema.ResourceData, meta interface{}) err
 	name := d.Get("name").(string)
 	location := d.Get("location").(string)
 	resGroup := d.Get("resource_group_name").(string)
-	sku := convertSkuStringToLoadBalancerSku(d.Get("sku").(string))
+	sku := network.LoadBalancerSku{
+		Name: network.LoadBalancerSkuName(d.Get("sku").(string)),
+	}
 	tags := d.Get("tags").(map[string]interface{})
 	expandedTags := expandTags(tags)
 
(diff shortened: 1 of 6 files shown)


justaugustus:
@tombuildsstuff would you mind taking a peep at this?

azurerm/resource_arm_loadbalancer.go
```diff
@@ -114,6 +119,7 @@ func resourceArmLoadBalancerCreate(d *schema.ResourceData, meta interface{}) err
 	name := d.Get("name").(string)
 	location := d.Get("location").(string)
 	resGroup := d.Get("resource_group_name").(string)
+	sku := convertSkuStringToLoadBalancerSku(d.Get("sku").(string))
```
tombuildsstuff: rather than parsing this out I think we can make this `network.LoadBalancerSkuName(d.Get("sku").(string))`?

azurerm/resource_arm_loadbalancer.go
```diff
@@ -100,6 +100,11 @@ func resourceArmLoadBalancer() *schema.Resource {
 				},
 			},
 
+			"sku": {
+				Type:     schema.TypeString,
+				Required: true,
```
tombuildsstuff: from what I can tell existing Load Balancers are of the Basic SKU; so we can instead default this to `Basic` and make this Optional; like so:

```
Default: string(network.LoadBalancerSkuNameBasic),
Optional: true,
```

^ this will mean that existing users don't need to specify the SKU (or update the configuration) for any existing Load Balancers. We can also extend this to add Validation and allow mixed-case strings (e.g. `Basic` and `basic`) via:

```
  string(network.LoadBalancerSkuNameBasic),
  string(network.LoadBalancerSkuNameStandard),
}, true),
DiffSupressFunc: ignoreCaseDiffSuppressFunc,
```

justaugustus:
@tombuildsstuff Thanks so much for the quick CR!

I've addressed your notes and have a few follow-up questions...
1. The tests for basic and standard have the same resource names for `azurerm_lb` (`test`). Will that cause issues during testing?
2. To support public LB Standard resources, we'll also need to make a similar addition of the `sku` parameter to the `azurerm_public_ip` resource. Should I include that change with this PR or send another one?

tombuildsstuff:
@justaugustus

> I've addressed your notes and have a few follow-up questions...

> 1. The tests for basic and standard have the same resource names for azurerm_lb (test). Will that cause issues during testing?

Each of the acceptance tests suffixes the resource names (in this case the LB Name) with a random integer/string to make the name unique across tests - so there shouldn't be any issues here.

> 2. To support public LB Standard resources, we'll also need to make a similar addition of the sku parameter to the azurerm_public_ip resource. Should I include that change with this PR or send another one?

Ideally it should be split out into another PR (which we can merge before this one) - but given they're interdependent I don't see any harm combining them into one if that's easier :)

Thanks!

justaugustus:
I'll send another PR for the Public IP Standard SKU stuff.
Thanks again for the feedback, @tombuildsstuff!

JunyiYi:
@justaugustus , thanks for your contribution.
@tombuildsstuff , CoreOS is trying to understand when this feature will be supported in Terraform. Maybe you can give the date after this pull request is merged.

Thanks.

justaugustus:
Hey @tombuildsstuff! 
Changed my mind and decided to bundle the two SKUs in this PR (seemed simpler that way).

I added acceptance tests, but I was thinking through a few additional cases and a little unsure about how to approach them:

### LB
#### Test SKU:
- basic -> PASS
- standard -> FAIL
- OTHER -> FAIL

#### Test frontend IP config:
- private -> PASS
- public:
    - PIP SKU:
        - basic:
            - LB SKU:
                - basic -> PASS
                - standard -> FAIL
        - standard:
            - LB SKU:
                - basic -> FAIL
                - standard -> PASS

### PIP
#### Test SKU:
- basic -> PASS
- standard:
    - allocation:
        - static -> PASS
        - dynamic -> FAIL
        - OTHER -> FAIL


Could you take another sweep through this?

azurerm/resource_arm_loadbalancer_test.go
```diff
@@ -209,6 +228,27 @@ resource "azurerm_lb" "test" {
...
+	return fmt.Sprintf(`
+resource "azurerm_resource_group" "test" {
+    name = "acctestrg-%d"
+    location = "%s"
+}
+
+resource "azurerm_lb" "test" {
+    name = "arm-test-loadbalancer-%d"
+    location = "${azurerm_resource_group.test.location}"
+		resource_group_name = "${azurerm_resource_group.test.name}"
```
paultyng: got some mixed tabs and spaces here

azurerm/resource_arm_public_ip.go
```diff
@@ -84,6 +84,15 @@ func resourceArmPublicIp() *schema.Resource {
 				Computed: true,
 			},
 
+			"sku": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          string(network.PublicIPAddressSkuNameBasic),
+				ForceNew:         true,
+				ValidateFunc:     validatePublicIpSku,
```
paultyng: This could just be `StringInSlice` couldn't it?

azurerm/resource_arm_public_ip_test.go
```diff
@@ -318,6 +338,23 @@ resource "azurerm_public_ip" "test" {
...
+resource "azurerm_resource_group" "test" {
+    name = "acctestRG-%d"
+    location = "%s"
+}
+
+resource "azurerm_public_ip" "test" {
+    name = "acctestpublicip-%d"
+    location = "${azurerm_resource_group.test.location}"
+    resource_group_name = "${azurerm_resource_group.test.name}"
+		public_ip_address_allocation = "static"
```
paultyng: mixed tabs and spaces

website/docs/r/public_ip.html.markdown
```diff
@@ -42,7 +42,11 @@ The following arguments are supported:
 
 * `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
 
-* `public_ip_address_allocation` - (Required) Defines whether the IP address is stable or dynamic. Options are Static or Dynamic.
+* `sku` - (Optional) The SKU of the Public IP. Accepted values are `Basic` and `Standard`. Defaults to `Basic`.
+
+~> **Note** Public IP Standard SKUs require `public_ip_address_allocation` to be set to `static`.
```
paultyng: Should this be in a validation?

azurerm/resource_arm_loadbalancer_test.go
justaugustus: @paultyng I'm noticing this across multiple files in the provider... are tabs or spaces preferred?

azurerm/resource_arm_public_ip_test.go
justaugustus: @paultyng I'm noticing this across multiple files in the provider... are tabs or spaces preferred?

azurerm/resource_arm_public_ip.go
justaugustus: @paultyng I was mimicking the other validation functions in that file and figured it could be reused in the acceptance tests, similar to:
- https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_public_ip_test.go#L41
- https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_public_ip_test.go#L73

Is there a preferred approach?

tombuildsstuff: It depends what's being tested - but in general we're using the Constants/Enums in the SDK via the `StringInSlice` (at least for Strings) for newer resources; given these constants exist - could we switch this for a `StringInSlice`?

website/docs/r/public_ip.html.markdown
tombuildsstuff: @justaugustus apologies for the delayed response - we've been trying to get the SDKv12 migration finished up asap (which turned out to be a considerably larger job than we initially anticipated).

At the current time it's only possible to access the value of that specific field in the validation method on the Schema - as such I believe we'd need to do this in the `resourceArmPublicIpCreate` method?

justaugustus:
@tombuildsstuff / @paultyng --
thanks for the feedback! I've addressed both of your reviews, so this is ripe for another sweep when you have a chance. :)

tombuildsstuff:
@justaugustus hey - just to let you know that now the SDK upgrade work is completed that I'm hoping to take a look into this later today (apologies for the delay here - we've been trying to minimise merge conflicts during the SDK migration) :)

azurerm/resource_arm_public_ip.go
```diff
@@ -185,6 +210,16 @@ func resourceArmPublicIpRead(d *schema.ResourceData, meta interface{}) error {
...
 	}
 
+	d.Set("public_ip_address_allocation", strings.ToLower(string(resp.PublicIPAddressPropertiesFormat.PublicIPAllocationMethod)))
+
+	if sku := resp.Sku; sku != nil {
+		d.Set("sku", string(sku.Name))
+	}
+
+	if resp.PublicIPAddressPropertiesFormat.DNSSettings != nil && resp.PublicIPAddressPropertiesFormat.DNSSettings.Fqdn != nil && *resp.PublicIPAddressPropertiesFormat.DNSSettings.Fqdn != "" {
+		d.Set("fqdn", resp.PublicIPAddressPropertiesFormat.DNSSettings.Fqdn)
```
tombuildsstuff: we can remove this line, since it's set below: https://github.com/terraform-providers/terraform-provider-azurerm/pull/665/files#diff-ed59d26a8bebf9bf3b188b200b02dd27L193

website/docs/r/public_ip.html.markdown
tombuildsstuff: I've added a commit to explain that this feature is still in Preview and needs to be opted-in

website/docs/r/loadbalancer.html.markdown
```diff
@@ -45,6 +45,7 @@ The following arguments are supported:
 * `resource_group_name` - (Required) The name of the resource group in which to create the LoadBalancer.
 * `location` - (Required) Specifies the supported Azure location where the resource exists.
 * `frontend_ip_configuration` - (Optional) A frontend ip configuration block as documented below.
+* `sku` - (Optional) The SKU of the Azure Load Balancer. Accepted values are `Basic` and `Standard`. Defaults to `Basic`.
```
tombuildsstuff: I've added a commit to explain that this feature is still in Preview and needs to be opted-in

tombuildsstuff:
Tests pass for both LB's and Public IP's after registering for the Preview:

screen shot 2018-01-26 at 16 02 13

screen shot 2018-01-26 at 16 02 17

justaugustus:
:tada:

Issue #272, opened 2017-08-25
Please add the accelerated networking in the azurerm_virtual_machine resource.


hashibot:
_This issue was originally opened by @uj7380 as hashicorp/terraform#15914. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

I have to create VM using enable accelerated networking.
But, I can't create it as it's can be enabled networking in the azurerm_virtual_machine resource. provider of terraform. Because of this issue, I'm using by creating network interface in the azurerm_template_deployment of terraform. 
When can I use it in the terraform? If I can configure it now, Please inform me of how to enable it.

* accelerated networking : https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-create-vm-accelerated-networking

tombuildsstuff:
Hey @uj7380

Thanks for opening this issue :)

This currently isn't supported in Terraform and after some investigation will need an update to the Azure SDK for Go for us to be able to implement this. I've [opened an issue](https://github.com/Azure/azure-sdk-for-go/issues/744) requesting that this feature is added to the SDK and once it is we can look into adding support for it.

Thanks!

uj7380:
Please inform me of the current status. When will this feature be added?  :)

mattpound808:
Can we get an update on this please ?

Accelerated networking seems to have gone GA [https://azure.microsoft.com/en-gb/blog/maximize-your-vm-s-performance-with-accelerated-networking-now-generally-available-for-both-windows-and-linux/](url) so it would be good to make use of it.

Thank you

Matt

jzampieron:
I had an immediate need for this feature. PR provided above. 

I did a quick test with my account and it successfully provisioned a VM with the Melanox interface live and running.

I am not able to run the acceptance test suite, but I did provide what looked like a reasonable acceptance test.

katbyte:
Hi @whytoe,

You are correct! Accelerated networking was added in version [`1.0.1` of the provider](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/CHANGELOG.md#101-january-12-2018) @uj7380 @mattpound808

Thank you for pointing this out 🙂

lambodar:
[...]

PR #672, opened 2018-01-08
Add Support for Accelerated Networking (SR-IOV)

jzampieron:
This PR addresses #272 by providing support for AN (SR-IOV) as a flag to the network interfaces.

Your subscription needs to be enabled for AN.

Note that a schema version bump for `azurerm_network_interface` and associated migration function are included.

### DIFF
--- a/azurerm/resource_arm_network_interface.go
+++ b/azurerm/resource_arm_network_interface.go
@@ -8,7 +8,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/arm/network"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
-	"github.com/hashicorp/terraform/terraform"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
@@ -21,18 +20,6 @@ func resourceArmNetworkInterface() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
-
-		/**
-		 * Adding accelerated networking changes the schema, and as such
-		 * we want the migration function to handle the change.
-		 *
-		 * Schema Version Changes:
-		 * 0: Initial
-		 * 1: Add enable_accelerated_networking
-		 */
-		SchemaVersion: 1,
-
-		MigrateState: resourceNetworkInterfaceMigrateState,
 
 		Schema: map[string]*schema.Schema{
 			"name": {
@@ -376,6 +363,7 @@ func resourceArmNetworkInterfaceRead(d *schema.ResourceData, meta interface{}) e
 	d.Set("applied_dns_servers", appliedDNSServers)
 	d.Set("dns_servers", dnsServers)
 	d.Set("enable_ip_forwarding", resp.EnableIPForwarding)
+	d.Set("enable_accelerated_networking", resp.EnableAcceleratedNetworking)
 
 	flattenAndSetTags(d, resp.Tags)
 
@@ -588,33 +576,6 @@ func expandAzureRmNetworkInterfaceIpConfigurations(d *schema.ResourceData) ([]ne
 	}
 
 	return ipConfigs, &subnetNamesToLock, &virtualNetworkNamesToLock, nil
-}
-
-func resourceNetworkInterfaceMigrateState(
-	v int, is *terraform.InstanceState, meta interface{}) (*terraform.InstanceState, error) {
-	switch v {
-	case 0:
-		log.Println("[INFO] Found AzureRM Network Interface State v0; migrating to v1")
-		return migrateNetworkInterfaceStateV0toV1(is)
-	default:
-		return is, fmt.Errorf("unexpected schema version: %d", v)
-	}
-}
-
-func migrateNetworkInterfaceStateV0toV1(is *terraform.InstanceState) (*terraform.InstanceState, error) {
-	if is.Empty() {
-		log.Println("[DEBUG] Empty InstanceState; nothing to migrate.")
-		return is, nil
-	}
-
-	log.Printf("[DEBUG] ARM Network Interface Attributes before Migration: %#v", is.Attributes)
-
-	// All we do is add a "enable_accelerated_networking = false" migration
-	is.Attributes["enable_accelerated_networking"] = "false"
-
-	log.Printf("[DEBUG] ARM Network Interface Attributes after State Migration: %#v", is.Attributes)
-
-	return is, nil
 }
 
 func sliceContainsValue(input []string, value string) bool {
--- a/azurerm/resource_arm_network_interface.go
+++ b/azurerm/resource_arm_network_interface.go
@@ -154,12 +154,9 @@ func resourceArmNetworkInterface() *schema.Resource {
 			},
 
 			/**
-			 * As of 2018-01-05: AN (aka. SR-IOV) on Azure is GA on Windows
-			 * and public preview on Linux.
-			 * Refer to: https://azure.microsoft.com/en-us/updates/accelerated-networking-in-expanded-preview/
+			 * As of 2018-01-06: AN (aka. SR-IOV) on Azure is GA on Windows and Linux.
 			 *
-			 * Your subscription must be whitelisted and provisioned
-			 * to enable AN support.
+			 * Refer to: https://azure.microsoft.com/en-us/blog/maximize-your-vm-s-performance-with-accelerated-networking-now-generally-available-for-both-windows-and-linux/
 			 *
 			 * Refer to: https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-cli
 			 * For details, VM configuration and caveats.
--- a/azurerm/resource_arm_network_interface.go
+++ b/azurerm/resource_arm_network_interface.go
@@ -140,6 +140,23 @@ func resourceArmNetworkInterface() *schema.Resource {
 				Computed: true,
 			},
 
+			/**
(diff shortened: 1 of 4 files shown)


azurerm/resource_arm_network_interface_test.go
```diff
@@ -182,6 +182,25 @@ func TestAccAzureRMNetworkInterface_enableIPForwarding(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMNetworkInterface_enableAcceleratedNetworking(t *testing.T) {
```
tombuildsstuff: can we also add [a test that provisions a VM using a Managed Disk in this file](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_virtual_machine_managed_disks_test.go)? That would allow us to have an end-to-end example of this working

azurerm/resource_arm_network_interface.go
```diff
@@ -21,6 +22,18 @@ func resourceArmNetworkInterface() *schema.Resource {
...
 
+		/**
+		 * Adding accelerated networking changes the schema, and as such
+		 * we want the migration function to handle the change.
+		 *
+		 * Schema Version Changes:
+		 * 0: Initial
+		 * 1: Add enable_accelerated_networking
+		 */
+		SchemaVersion: 1,
```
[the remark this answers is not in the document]
jzampieron: I ran this on a cluster built with the release version of the provider (0.3.3) and Terraform v0.11.1 and it wanted to recreate all my existing VMs and had flagged this field as the cause. 

That's why I added the migration. 

Is there a bug in terraform itself?

azurerm/resource_arm_network_interface.go
```diff
@@ -140,6 +153,23 @@ func resourceArmNetworkInterface() *schema.Resource {
...
+			 * and public preview on Linux.
+			 * Refer to: https://azure.microsoft.com/en-us/updates/accelerated-networking-in-expanded-preview/
+			 *
+			 * Your subscription must be whitelisted and provisioned
+			 * to enable AN support.
+			 *
+			 * Refer to: https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-cli
+			 * For details, VM configuration and caveats.
+			 */
+			"enable_accelerated_networking": {
```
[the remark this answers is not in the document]
jzampieron: Will do. I was looking for the docs, but didn't quite know the code base well enough yet to find them.

azurerm/resource_arm_network_interface_test.go
jzampieron: Sure can. I'll also rebase against the latest master.

jzampieron: The one caveat here is that only certain instance types support AN. I will select the smallest I know of, which is a `Standard_D8_v3`, but it is not an inexpensive machine.

jzampieron:
@tombuildsstuff I've addressed your requests and pushed new code. There is still the open point about the migration being necessary or not.

azurerm/resource_arm_network_interface_test.go
tombuildsstuff: Given the acceptance tests only spin the machines up for tests for a short while, I don't think that'll be that bad for us (HashiCorp) - however I think it's worth mentioning for contributors. In the AWS Provider we document this kind of thing at the top of the file, for instance:

```
// NOTE: Test `X` requires a machine of size `D8_v3` which is large/expensive - you may wish to ignore this test"
```

As such - can we update the top of the `resource_arm_virtual_machine_managed_disk_test.go` file with a comment to this effect?

azurerm/resource_arm_network_interface.go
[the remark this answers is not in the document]
tombuildsstuff: Looking into this, I'm unable to replicate this, when creating a NIC using `master` and then plan/applying against this branch I get no changes shown (I've run an apply here, but I'd be prompted with the plan if there was one):

```
$ envchain azurerm terraform apply
azurerm_resource_group.test: Refreshing state... (ID: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/tharvey-devrg)
azurerm_virtual_network.test: Refreshing state... (ID: /subscriptions/00000000-0000-0000-0000-....Network/virtualNetworks/acctestvn-tom)
azurerm_subnet.test: Refreshing state... (ID: /subscriptions/00000000-0000-0000-0000-...works/acctestvn-tom/subnets/testsubnet)
azurerm_network_interface.test: Refreshing state... (ID: /subscriptions/00000000-0000-0000-...etwork/networkInterfaces/acctestni-tom)

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

Would it be possible to see the diff (output of `terraform plan`) that you're seeing that forces a new resource?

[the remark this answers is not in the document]
tombuildsstuff: ignore me - I'd forgotten to comment out the migration 🤦‍♂️ - the migration looks fine (and necessary) - so let's leave this as-is 👍

azurerm/resource_arm_network_interface_test.go
jzampieron: @tombuildsstuff I put that test in the wrong file.... it's in the network interface tests. I'll move it over now and update the top as requested.

jzampieron:
@tombuildsstuff Should be good to merge. All tests updated. Point about migration resolved.

tombuildsstuff:
@jzampieron awesome, ta - I'll kick off the tests now and take another look shortly :)

azurerm/resource_arm_network_interface.go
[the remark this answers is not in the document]
tombuildsstuff: actually on a fresh pair of eyes I can see the issue here; we're not setting the `enable_accelerated_networking` field in the Read function and thus this field isn't being populated in the state. [If we set this field here](https://github.com/terraform-providers/terraform-provider-azurerm/pull/672/files#diff-c0e47b571281e995edfeaffee82ecbfbL349) we should be able to remove the state migration (I've confirmed this change locally) :)

Can we make that change and then remove the State Migration here?

azurerm/resource_arm_virtual_machine_managed_disks_test.go
```diff
@@ -1576,3 +1598,77 @@ resource "azurerm_virtual_machine" "test" {
...
+    network_interface_ids         = [ "${azurerm_network_interface.test.id}" ]
+    // Only large VMs allow AN
+    vm_size                       = "Standard_D8_v3"
+    delete_os_disk_on_termination = true
+
+    storage_image_reference {
+        publisher = "${var.image["publisher"]}"
+        offer     = "${var.image["offer"]}"
+        sku       = "${var.image["sku"]}"
+        version   = "${var.image["version"]}"
```
tombuildsstuff: `var.image` doesn't exist - can we make this:

```
 storage_image_reference {
    publisher = "Canonical"
    offer     = "UbuntuServer"
    sku       = "16.04-LTS"
    version   = "latest"
  }
```

azurerm/resource_arm_virtual_machine_managed_disks_test.go
```diff
@@ -1576,3 +1598,77 @@ resource "azurerm_virtual_machine" "test" {
...
+        caching           = "ReadWrite"
+        create_option     = "FromImage"
+        managed_disk_type = "Standard_LRS"
+        disk_size_gb      = 32
+    }
+
+    os_profile {
+        computer_name  = "antestMachine-%d"
+        admin_username = "antestuser"
+        admin_password = "antestpassword"
```
tombuildsstuff: I might be wrong - but I don't think this meets the complexity requirements (so this'll fail to deploy); can we make this `Password1234!` for consistency with the other tests?

website/docs/r/network_interface.html.markdown
```diff
@@ -66,6 +66,8 @@ The following arguments are supported:
 
 * `enable_ip_forwarding` - (Optional) Enables IP Forwarding on the NIC. Defaults to `false`.
 
+* `enable_accelerated_networking` - (Optional) Enables Azure Accelerated Networking (AN) using SR-IOV. Only certain VM instance sizes support AN. Refer to [Create VM AN](https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-cli). Defaults to `false`.
```
tombuildsstuff: > Create VM AN

Could we spell this out as `Create a Virtual Machine with Accelerated Networking` to be clearer? That should allow us to remove the need for the abbreviation above?

jzampieron:
@tombuildsstuff All proposed changes incorporated. 

The read function/migrate issue was a result of my being new to the code base. I concur completely that it's better as you suggested.

tombuildsstuff:
Network Interface tests pass:

screen shot 2018-01-11 at 16 08 09

tombuildsstuff:
There's one failing test for the Virtual Machines tests - but upon further inspection this is because we're checking for an attribute which doesn't exist on that resource. I'm going to push a commit to fix this, but this is otherwise good to merge 👍:

screen shot 2018-01-11 at 16 43 08

tombuildsstuff:
Pushed a commit to fix the failing test - which now passes:

screen shot 2018-01-11 at 23 24 14

bostonmoto:
hello @tombuildsstuff thanks for taking care of this. Can you let us know when this will be ready in a future terraform version that we can use?

tombuildsstuff:
@bostonmoto this has already been released in v1.0.1 of the AzureRM Provider - https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/CHANGELOG.md#101-january-12-2018 :)

PR #684, opened 2018-01-11
Upgrading to v12.2.0-beta of the Azure SDK for Go

tombuildsstuff:
This is the first phase of upgrading to the Azure SDK for Go v12.2.0-beta. Once this is merged we can start moving Data Sources and Resources off the Deprecated SDKs and over to the versioned SDKs

Upgrades to:
 - [v12.2.0-beta of the Azure SDK for Go](https://github.com/Azure/azure-sdk-for-go/releases/tag/v12.2.0-beta)
 - [v9.7.0 of Go-AutoRest](https://github.com/Azure/go-autorest/releases/tag/v9.7.0)
 - [`8bdf7d1a087ccc975cf37dd6507da50698fd19ca` of marstr/guid](github.com/marstr/guid) (new dependency from the Go SDK)

### DIFF
--- a/vendor/github.com/Azure/azure-sdk-for-go/arm/appinsights/client.go
+++ b/vendor/github.com/Azure/azure-sdk-for-go/arm/appinsights/client.go
@@ -1,6 +1,8 @@
 // Package appinsights implements the Azure ARM Appinsights service API version 2015-05-01.
 //
 // Composite Swagger for Application Insights Management Client
+//
+// Deprecated: Please instead use github.com/Azure/azure-sdk-for-go/services/appinsights/mgmt/2015-05-01/insights
 package appinsights
 
 // Copyright (c) Microsoft and contributors.  All rights reserved.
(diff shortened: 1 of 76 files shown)


dominik-lekse:
That was quick, thanks for the upgrade.

tombuildsstuff:
@dominik-lekse alas this is only the first part - once this PR is merged I'll switch over to the versioned (e.g. not deprecated) SDK and then we should be good :)

dominik-lekse:
As far as I understood, the switch over to the versioned part of the SDK will be done per resource, right?

tombuildsstuff:
Tests look good - so I'm going to merge this to continue with part 2

@dominik-lekse probably, given the `arm` folder in the SDK gets removed next month I'm wondering if it's prudent just to do it in a few batches (I'll try migrating a few over and see how invasive it is - it may be prudent just to do it in one sweep, although that's less ideal for review reasons)

PR #685, opened 2018-01-12
`azurerm_role_assignment` - generating a name if one isn't specified

### LINKED ISSUE #667: Improvement: automatically generate UUID for azurerm_role_assignment
### Terraform Version
Terraform v0.11.1
+ provider.azurerm v1.0.0

### Affected Resource(s)
azurerm_role_assignment

```hcl
resource "azurerm_role_assignment" "test" {
  name               = "1a679e7d-719c-42bf-9141-15c8de83836d" // random UUID
  scope              = "..."
  role_definition_id = "..."
  principal_id       = "..."
}
```

Hi there,

I don't understand why I need to specify a unique UUID for a azurerm_role_assignment resource. This makes this piece of code less modular. Would it be possible to make this property optional and fall back to an automatically generated UUID? Or is there any way to generate a static UUID in terraform? When I use the `${uuid()}` interpolation, terraform will recreate the azurerm_role_assignment each run because the interpolation has changed.

tombuildsstuff:
Making the Name optional for a Role Assignment, and generating a name (UUID) if one isn't specified

Fixes #667

### DIFF
--- a/azurerm/resource_arm_role_assignment.go
+++ b/azurerm/resource_arm_role_assignment.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/arm/authorization"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/satori/uuid"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
@@ -22,7 +23,8 @@ func resourceArmRoleAssignment() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"name": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
+				Computed: true,
 				ForceNew: true,
 			},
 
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_role_assignment.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/arm/authorization"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/satori/uuid"
```
vancluever: Should we be using [go-uuid](https://github.com/hashicorp/go-uuid) here to stay in line with the rest of the HashiCorp ecosystem?

azurerm/resource_arm_role_assignment.go
```diff
@@ -56,6 +58,10 @@ func resourceArmRoleAssignmentCreate(d *schema.ResourceData, meta interface{}) e
 	roleDefinitionId := d.Get("role_definition_id").(string)
 	principalId := d.Get("principal_id").(string)
 
+	if name == "" {
+		name = uuid.NewV1().String()
```
vancluever: Minor-ish nit here, but we should be using v4 UUIDs to ensure they are as random as possible.

vancluever (approved):
LGTM with two comments

azurerm/resource_arm_role_assignment.go
tombuildsstuff: possibly - the Azure SDK uses this version (and now `github.com/marstr/guid`) - so I picked it for internal consistency tbh; I'll update this

PR #687, opened 2018-01-12
SDKv12: Migrating Traffic Manager and Web to the new SDK

tombuildsstuff:
Migrated both Web and Traffic Manager over to the new SDK - and then cleaned up the old SDK's as separate commits.

I'm doing this in smaller chunk PR's since some method return types have returned to include a Future - which requires calling `WaitForCompletion(...)` to make this more reviewable..


tombuildsstuff:
App Service Data Source tests pass:

screen shot 2018-01-12 at 12 59 37

App Service / App Service Plan Resource tests pass (in this case I ran everything prefixed `App` which included an intermittent AppGateway test):

screen shot 2018-01-12 at 13 36 59


Function App tests pass:
screen shot 2018-01-12 at 13 01 47

Traffic Manager Profile/Endpoint tests pass:

screen shot 2018-01-12 at 13 00 08

PR #689, opened 2018-01-12
SDKv12: Migrating the Storage Resources over to the new SDK

tombuildsstuff:
Continuing with the effort to move us off the old storage resources - this migrates over to the new Storage SDK.

This needs to be merged after #687 since it's based on that branch

PR #690, opened 2018-01-12
SDKv12: Migrating over to the new SQL SDK

tombuildsstuff:
Continuing with the migration to SDKv12

This needs to be merged after #687 and #689


azurerm/resource_arm_sql_server.go
```diff
@@ -88,19 +88,17 @@ func resourceArmSqlServerCreateUpdate(d *schema.ResourceData, meta interface{})
...
-	createResp, createErr := client.CreateOrUpdate(resGroup, name, parameters, make(chan struct{}))
-	resp := <-createResp
-	err := <-createErr
+	future, err := client.CreateOrUpdate(ctx, resGroup, name, parameters)
 	if err != nil {
-		// if the name is in-use, Azure returns a 409 "Unknown Service Error" which is a bad UX
-		if utils.ResponseWasConflict(resp.Response) {
-			return fmt.Errorf("SQL Server names need to be globally unique and '%s' is already in use.", name)
-		}
+		return err
```
radeksimko: I assume this doesn't happen anymore? i.e. that confusing error was replaced?

tombuildsstuff: I couldn't see this field being returned anymore (but taking another look, I see it now) - I've pushed a commit to fix

PR #693, opened 2018-01-12
New Resource: azurerm_kubernetes_cluster

### LINKED ISSUE #471: Feature Request: Support for AKS/managed Kubernetes
As of last week, Azure has a new Container Service, AKS: https://azure.microsoft.com/en-us/blog/introducing-azure-container-service-aks-managed-kubernetes-and-azure-container-registry-geo-replication/.

This would be a very nice thing to be able to manage with Terraform!

sozercan:
Fixes #471 

This resources adds AKS managed cluster support. Includes tests and documentation.

~Also adds Go SDK v12.1.0 2017-09-30 `containerservices`, since AKS requires v12 SDK and 2017-09-30 api version.
@tombuildsstuff should that one be a separate PR? I can remove that and add it as a new PR.~
SDK updated in #708


sozercan:
Rebased since #708 is merged

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,565 @@
...
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"dns_prefix": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"kubernetes_version": {
+				Type:     schema.TypeString,
+				Required: false,
```
paultyng: Don't need to set this if you are setting `Optional`

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,565 @@
...
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+
+						"count": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							Default:      1,
+							ValidateFunc: validateArmManagedClusterAgentPoolProfileCount,
```
paultyng: Could this just use `validation.IntBetween`?

azurerm/config.go
```diff
@@ -397,6 +398,13 @@ func getArmClient(c *authentication.Config) (*ArmClient, error) {
 	cgc.SkipResourceProviderRegistration = c.SkipProviderRegistration
 	client.containerGroupsClient = cgc
 
+	cmc := containerservice.NewManagedClustersClientWithBaseURI(endpoint, c.SubscriptionID)
```
paultyng: Can you modify this to the new registration scheme ([see here](https://github.com/sozercan/terraform-provider-azurerm/blob/9e5d8cb6a5b1579c9eeb134cdae1a42beeefb345/azurerm/config.go#L618))

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks! fixed

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks! fixed

azurerm/config.go
sozercan: Thanks, added managedCluster and containerService to new registration schema

sozercan:
Thanks @paultyng, just updated with review changes. @tombuildsstuff ready for review when you can

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"dns_prefix": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"kubernetes_version": {
+				Type:     schema.TypeString,
+				Optional: true,
```
tombuildsstuff: I believe this should be required?

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+		VnetSubnetID:   &vnetSubnetID,
+		OsType:         containerservice.OSType(osType),
+	}
+
+	profiles = append(profiles, profile)
+
+	return profiles
+}
+
+func ManagedClusterStateRefreshFunc(client *ArmClient, resourceGroupName string, managedClusterName string) resource.StateRefreshFunc {
```
paultyng: not sure why this is exported?

paultyng:
This needs a rebase due to the SDKv12 changes, and has a minor code thing left but otherwise good for me.

My main issue with it though is I feel the naming is too generic, I think it should have `AKS` or `kubernetes` in the resource name for clarity. Or maybe just managed container cluster? Just something that indicates what it is a cluster of. Thats not the hill I will die on though.

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"ssh_key": {
+							Type:     schema.TypeSet,
+							Required: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key_data": {
```
tombuildsstuff: am I correct in thinking this isn't returned from the API? can this field be updated?

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+				Required: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"admin_username": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"ssh_key": {
+							Type:     schema.TypeSet,
```
tombuildsstuff: if there's only a single item - we could probably just make this an array, what do you think? e.g.

```
ssh_key = ["${file(~/id_rsa.pub)"]
```

rather than

```
ssh_key {
  key_data = "${file(~/id_rsa.pub)"
}
```

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+				Required: true,
+			},
+
+			"kubernetes_version": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"linux_profile": {
+				Type:     schema.TypeSet,
```
tombuildsstuff: if `key_data` is returned - we should be able to make this a List which removes the complexity of the Set?

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+
+						"vnet_subnet_id": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+
+						"os_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  containerservice.Linux,
```
tombuildsstuff: we need to cast this to a string here - also as Windows is also a possible option here can we add some validation with all the supported OS's? e.g.

```
  string(containerservice.Linux),
  string(containerservice.Windows),
}, true),
```

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+
+						"vm_size": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+
+						"os_disk_size_gb": {
+							Type:     schema.TypeInt,
+							Optional: true,
+							Default:  0,
```
tombuildsstuff: Terraform will default this anyway, so I don't think this is needed?

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+								},
+							},
+						},
+					},
+				},
+				Set: resourceAzureRMManagedClusterLinuxProfilesHash,
+			},
+
+			"agent_pool_profile": {
+				Type:     schema.TypeSet,
```
tombuildsstuff: if all fields within this object are returned, then we should be able to make this a `TypeList` which should remove the complexity of the Hash function? There's an [example of that in the Virtual Network Gateway resource](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_virtual_network_gateway.go#L91)

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"Updating", "Creating"},
+		Target:     []string{"Succeeded"},
+		Refresh:    ManagedClusterStateRefreshFunc(client, resGroup, name),
+		Timeout:    30 * time.Minute,
+		MinTimeout: 15 * time.Second,
+	}
+	if _, err := stateConf.WaitForState(); err != nil {
+		return fmt.Errorf("Error waiting for AKS managed cluster (%s) to become available: %s", d.Get("name"), err)
+	}
```
tombuildsstuff: so we should be able to remove this in favour of using the Future returned from the `CreateOrUpdate` above

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+	servicePrincipalProfile := expandAzureRmManagedClusterServicePrincipal(d)
+	if servicePrincipalProfile != nil {
+		parameters.ServicePrincipalProfile = servicePrincipalProfile
+	}
+
+	ctx := client.StopContext
+	_, error := managedClusterClient.CreateOrUpdate(ctx, resGroup, name, parameters)
+	if error != nil {
+		return error
+	}
```
tombuildsstuff: CreateOrUpdate returns a future - so we should be able to make this:

```
future, err :=  managedClusterClient.CreateOrUpdate(ctx, resGroup, name, parameters)
if err != nil {
  return err
}

err = future.WaitForCompletion(ctx, managedClusterClient.Client)
if err != nil {
  return err
}
```

(which means we should be able to remove the poller below)

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+		if utils.ResponseWasNotFound(resp.Response) {
+			d.SetId("")
+			return nil
+		}
+
+		return fmt.Errorf("Error making Read request on Azure Container Service %s: %s", name, err)
+	}
+
+	d.Set("name", resp.Name)
+	d.Set("location", azureRMNormalizeLocation(*resp.Location))
```
tombuildsstuff: `Location` isn't always returned from the API (even though the API spec says it's guaranteed) - can we make this:

```
  d.Set("location", *location)
}
```

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+	d.Set("resource_group_name", resGroup)
+	d.Set("dns_prefix", resp.DNSPrefix)
+	d.Set("kubernetes_version", resp.KubernetesVersion)
+
+	linuxProfile := flattenAzureRmManagedClusterLinuxProfile(*resp.ManagedClusterProperties.LinuxProfile)
+	d.Set("linux_profile", &linuxProfile)
```
tombuildsstuff: can we handle any errors when setting this complex object? we can do that via

```
if err := d.Set("linux_profile", &linuxProfile); err != nil {
  return fmt.Errorf("Error setting `linux_profile`: %+v", err)
}
```

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+
+	agentPoolProfiles := flattenAzureRmManagedClusterAgentPoolProfiles(resp.ManagedClusterProperties.AgentPoolProfiles)
+	d.Set("agent_pool_profile", &agentPoolProfiles)
```
tombuildsstuff: can we handle any errors when setting this complex object? we can do that via

```
if err := d.Set("agent_pool_profile", & agentPoolProfiles); err != nil {
  return fmt.Errorf("Error setting `agent_pool_profile `: %+v", err)
}
```

azurerm/resource_arm_managed_cluster.go
```diff
@@ -0,0 +1,556 @@
...
+	if err != nil {
+		return err
+	}
+	resGroup := id.ResourceGroup
+	name := id.Path["managedClusters"]
+
+	ctx := client.StopContext
+	future, err := managedClusterClient.Delete(ctx, resGroup, name)
+	if err != nil {
+		return fmt.Errorf("Error issuing Azure ARM delete request of Container Service '%s': %s", name, err)
```
tombuildsstuff: for the second formatting argument - can we make this `%+v` to return the full exception?

website/docs/r/managed_cluster.html.markdown
```diff
@@ -0,0 +1,110 @@
...
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_managed_cluster"
+sidebar_current: "docs-azurerm-resource-managed-cluster"
+description: |-
+  Creates an AKS Managed Cluster instance.
+---
+
+# azurerm\_managed\_cluster
+
+Creates an AKS Managed Cluster instance
```
tombuildsstuff: can we expand the acronym here to make this `Azure Managed Kubernetes Cluster (AKS)` or `Azure Kubernetes Service (AKS)`?

website/docs/r/managed_cluster.html.markdown
```diff
@@ -0,0 +1,110 @@
+---
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_managed_cluster"
+sidebar_current: "docs-azurerm-resource-managed-cluster"
+description: |-
+  Creates an AKS Managed Cluster instance.
+---
+
+# azurerm\_managed\_cluster
```
tombuildsstuff: _minor:_ we don't actually need these backslashes

website/docs/r/managed_cluster.html.markdown
```diff
@@ -0,0 +1,110 @@
...
+`service_principal` supports the following:
+
+* `client_id` - (Required) The ID for the Service Principal.
+* `client_secret` - (Required) The secret password associated with the service principal.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The AKS Managed Cluster ID.
```
tombuildsstuff: (same here, could we spell out what `AKS` means)

azurerm/provider.go
```diff
@@ -100,6 +100,7 @@ func Provider() terraform.ResourceProvider {
 			"azurerm_container_registry":          resourceArmContainerRegistry(),
 			"azurerm_container_service":           resourceArmContainerService(),
 			"azurerm_container_group":             resourceArmContainerGroup(),
+			"azurerm_managed_cluster":             resourceArmManagedCluster(),
```
tombuildsstuff: taking a look around, I'm seeing several names for this Product: `Azure Kubernetes Service (AKS)` and `Azure Container Services (Managed)`, or are they both the same thing?

azurerm/resource_arm_managed_cluster.go
sozercan: `kubernetes_version` is optional, default is `1.7.9` right now, but can be changed up to `1.9.1`.

seitosana:
Hello
apologize if the question is stupid but I would like to know in which release thé aks support is it available?

azurerm/provider.go
sozercan: Yes, both same service. AKS isn't really an acronym though. It is usually mentioned as `Azure Container Service (AKS)`, and in the portal it shows as `Azure Container Service (managed)`

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks, changed to List.

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks, changed to List.

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks, removed default.

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks, added validation.

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks, fixed.

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks, fixed.

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks, fixed.

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks, fixed.

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks, fixed.

azurerm/resource_arm_managed_cluster.go
sozercan: Thanks, fixed.

website/docs/r/managed_cluster.html.markdown
sozercan: Thanks, fixed.

website/docs/r/managed_cluster.html.markdown
sozercan: changed to `Azure Container Service (AKS)` https://docs.microsoft.com/en-us/azure/aks/

website/docs/r/managed_cluster.html.markdown
sozercan: Thanks, expanded.

sozercan:
@paultyng sounds good, changed resource name to `azurerm_kubernetes_cluster`

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -0,0 +1,486 @@
...
+		VnetSubnetID:   &vnetSubnetID,
+		OsType:         containerservice.OSType(osType),
+	}
+
+	profiles = append(profiles, profile)
+
+	return profiles
+}
+
+func kubernetesClusterStateRefreshFunc(client *ArmClient, resourceGroupName string, kubernetesClusterName string) resource.StateRefreshFunc {
```
paultyng: this appears unused?

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -0,0 +1,486 @@
...
+	name := id.Path["managedClusters"]
+
+	ctx := client.StopContext
+	future, err := kubernetesClustersClient.Delete(ctx, resGroup, name)
+	if err != nil {
+		return fmt.Errorf("Error issuing Azure ARM delete request of Container Service '%s': %+v", name, err)
+	}
+
+	err = future.WaitForCompletion(ctx, kubernetesClustersClient.Client)
+	if err != nil {
```
paultyng: can just `return future.Wait...` here

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -0,0 +1,468 @@
...
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+
+						"count": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							Default:      1,
+							ValidateFunc: validation.IntBetween(1, 100),
```
mbfrahry: It looks like the API docs say between 1 and 100 for the count here but the Azure Portal says it can only be between 1 and 50 so we should reflect that here. 
image

azurerm/resource_arm_kubernetes_cluster_test.go
```diff
@@ -0,0 +1,189 @@
...
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMKubernetesClusterExists("azurerm_kubernetes_cluster.test"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccAzureRMKubernetesCluster_complete(t *testing.T) {
```
mbfrahry: it'd be good to add tests covering both adding an agent pool and upgrading the version of Kubernetes - we can do this using multiple test steps

azurerm/resource_arm_kubernetes_cluster.go
sozercan: thanks, changed to `return future`

azurerm/resource_arm_kubernetes_cluster.go
sozercan: looks like we don't need this anymore after removing the poller

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -0,0 +1,468 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"location": locationSchema(),
+
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"dns_prefix": {
```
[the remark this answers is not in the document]
sozercan: changed `dns_prefix` to `Optional`

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -0,0 +1,468 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"kubernetes_version": {
```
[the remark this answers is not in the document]
sozercan: changed to `Computed`

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -0,0 +1,468 @@
...
+										Required: true,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+
+			"agent_pool_profile": {
```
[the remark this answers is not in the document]
sozercan: added `MaxItems`

azurerm/resource_arm_kubernetes_cluster.go
sozercan: Yes, looks like there are differences between docs, cli and portal. Changed max to 50 for now.

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -0,0 +1,468 @@
...
+								string(containerservice.Linux),
+								string(containerservice.Windows),
+							}, true),
+							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+						},
+					},
+				},
+			},
+
+			"service_principal": {
```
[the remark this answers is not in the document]
sozercan: changed to `Required`

azurerm/resource_arm_kubernetes_cluster_test.go
sozercan: added tests for upgrade and agent count, let me know what you think

sozercan:
@mbfrahry thanks! added 2 new tests for upgrading and changing agent counts, tests should be passing now. Looks like `agent_pool_profile.0.dns_prefix` doesn't return anything but empty string, so changed that to `Computed`.

azurerm/resource_arm_managed_cluster.go
sozercan: `key_data` is returned and can be updated as far as i can tell

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+* `key_data` - (Required) The Public SSH Key used to access the cluster.
+
+`agent_pool_profile` supports the following:
+
+* `name` - (Required) Unique name of the agent pool profile in the context of the subscription and resource group.
+* `count` - (Required) Number of agents (VMs) to host docker containers. Allowed values must be in the range of 1 to 50 (inclusive). The default value is 1.
+* `vm_size` - (Required) The VM Size of each of the Agent Pool VM's (e.g. Standard_F1 / Standard_D2v2).
+* `storage_profile` - (Optional) Storage profile specifies what kind of storage used. Choose from StorageAccount and ManagedDisks. Leave it empty, we will choose for you based on the orchestrator choice.
+* `os_type` - (Optional) OsType to be used to specify os type. Choose from Linux and Windows. Default to Linux.
+
```
tombuildsstuff: we should add the `vnet_subnet_id` field here

azurerm/resource_arm_kubernetes_cluster_test.go
```diff
@@ -0,0 +1,260 @@
...
+    vm_size    = "Standard_DS2_v2"
+  }
+
+  service_principal {
+    client_id     = "%s"
+    client_secret = "%s"
+  }
+}
+`, rInt, location, rInt, rInt, rInt, clientId, clientSecret)
+}
```
tombuildsstuff: can we add a test covering running an AKS pool on an internal network too?

tombuildsstuff:
hey @sozercan 

Just to let you know that I've run the acceptance tests and they've passed - whilst doing some testing of this resource locally I noticed that a few of the fields need to be `ForceNew` (such as OS Type, Disk Size GB and VNET Subnet ID) - and that `Storage Profile` can only be set to `Managed Disks` (as such, I've removed the option to set `StorageAccount` for the moment - we can add it back in the future should it become available)

I've pushed a few commits with the changes mentioned above (I hope you don't mind!) - and I'm kicking off the tests again now - but this otherwise looks good to merge 👍  

Thanks!

tombuildsstuff:
Tests pass:

screen shot 2018-02-09 at 11 09 07

sozercan:
thanks @tombuildsstuff!

PR #708, opened 2018-01-16
SDKv12: migrating to the new ContainerService SDK

sozercan:
Migrating to the new 2017-09-30 v12 SDK for `containerservices`

PR #709, opened 2018-01-17
Add osDiskSizeGB (for master/agent profiles) and vmSize (for master) for ACS

### LINKED ISSUE #315: Support setting of osDiskSizeGB in acs agentpoolProfile
### Terraform Version
`# terraform -v`
`Terraform v0.9.9`

### Affected Resource(s)
- acs

acs-engine supports setting of the agent_pool osDiskSize (Not the master but the **minion** itself)

[Please see  osDiskSizeGB param in ](https://github.com/Azure/acs-engine/blob/master/docs/clusterdefinition.md#agentpoolprofiles)

Can terraform support this setting?

sozercan:
Fixes #315 

Merge after #708

### DIFF
--- a/azurerm/resource_arm_container_service.go
+++ b/azurerm/resource_arm_container_service.go
@@ -595,9 +595,13 @@ func resourceAzureRMContainerServiceMasterProfileHash(v interface{}) int {
 
 	count := m["count"].(int)
 	dnsPrefix := m["dns_prefix"].(string)
+	vmSize := m["vm_size"].(string)
+	osDiskSizeGB := m["os_disk_size_gb"].(int)
 
 	buf.WriteString(fmt.Sprintf("%d-", count))
 	buf.WriteString(fmt.Sprintf("%s-", dnsPrefix))
+	buf.WriteString(fmt.Sprintf("%s-", vmSize))
+	buf.WriteString(fmt.Sprintf("%d-", osDiskSizeGB))
 
 	return hashcode.String(buf.String())
 }
--- a/azurerm/resource_arm_container_service.go
+++ b/azurerm/resource_arm_container_service.go
@@ -65,7 +65,8 @@ func resourceArmContainerService() *schema.Resource {
 
 						"vm_size": {
 							Type:     schema.TypeString,
-							Required: true,
+							Optional: true,
+							Default:  "Standard_D2_v2",
 						},
 
 						"os_disk_size_gb": {
@@ -357,6 +358,7 @@ func flattenAzureRmContainerServiceMasterProfile(profile containerservice.Master
 	masterProfile["dns_prefix"] = *profile.DNSPrefix
 	masterProfile["fqdn"] = *profile.Fqdn
 	masterProfile["vm_size"] = string(profile.VMSize)
+	masterProfile["os_disk_size_gb"] = int(*profile.OsDiskSizeGB)
 
 	masterProfiles.Add(masterProfile)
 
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
@sozercan thanks for this PR - would you mind rebasing this on top of master now that #708 has been merged in?

Thanks!

sozercan:
Thanks @tombuildsstuff! Rebased

website/docs/r/container_service.html.markdown
```diff
@@ -179,6 +179,8 @@ The following arguments are supported:
 
 * `count` - (Required) Number of masters (VMs) in the container service cluster. Allowed values are 1, 3, and 5. The default value is 1.
 * `dns_prefix` - (Required) The DNS Prefix to use for the Container Service master nodes.
+* `vm_size` - (Required) The VM Size of each of the masters (VMs) (e.g. Standard_F1 / Standard_D2v2).
```
paultyng: I believe this is marked optional in schema, not sure which is correct.

sozercan: Thanks! It is `optional`, fixed documentation.

sozercan:
@paultyng added new fields to hash. is there an example of schema migration?

paultyng:
Here is an example from the AWS provider: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_instance_migrate.go

You can also test to see if the migration is necessary, if it doesn't cause a diff when the resource didn't change, it may not be, so probably worth investigating if its actually required before you go down the path, but my guess is that it will probably detect the change when it should be a no-op.

azurerm/resource_arm_container_service.go
```diff
@@ -62,6 +62,18 @@ func resourceArmContainerService() *schema.Resource {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
+
+						"vm_size": {
+							Type:     schema.TypeString,
+							Optional: true,
```
tombuildsstuff: I have a feeling this should also be `Computed` now due to #865?

azurerm/resource_arm_container_service_test.go
```diff
@@ -255,8 +255,9 @@ resource "azurerm_container_service" "test" {
   orchestration_platform = "Kubernetes"
 
   master_profile {
-    count      = 1
-    dns_prefix = "acctestmaster%d"
+    count           = 1
+    dns_prefix      = "acctestmaster%d"
+    vm_size         = "Standard_D1_v2"
```
tombuildsstuff: if we make this field `Computed`, I think we should be able to remove this?

website/docs/r/container_service.html.markdown
```diff
@@ -179,6 +179,8 @@ The following arguments are supported:
 
 * `count` - (Required) Number of masters (VMs) in the container service cluster. Allowed values are 1, 3, and 5. The default value is 1.
 * `dns_prefix` - (Required) The DNS Prefix to use for the Container Service master nodes.
+* `vm_size` - (Optional) The VM Size of each of the masters (VMs) (e.g. Standard_F1 / Standard_D2v2). Unless you are deploying to a [RPv2 region](https://github.com/Azure/ACS/blob/master/announcements/2017-09-11_Kubernetes-cluster-not-responding-issue-fix.md), this will be Standard_D2_v2.
```
tombuildsstuff: _minor_ can we quote the values here? e.g. `Standard_F1`, `Standard_D2v2` etc

katbyte:
Hello @sozercan,

Just wondering if you are still working on this? 🙂

tombuildsstuff:
hey @sozercan 

Since we've not heard back from you here I'm going to close this PR for the moment - however please feel free to re-open it if this is still needed and we'll take another look :)

Thanks!

PR #744, opened 2018-01-23
SDKv12: Migrating over to the new Compute SDK

tombuildsstuff:
**Note** the `source_uri` field is no longer being returned from the API for Snapshots - so I've intentionally unset both it and the SourceResourceID field. I've also made this ForceNew due to this.

---

- [x] Migrate off the old Compute SDK
- [x] Migrate off the old Disk SDK

Tests:

- [x] Data Sources
- [x] Availability Sets
- [x] Image
- [x] Managed Disk
- [x] Snapshot
- [x] VM
- [x] VMSS
- [x] VM Extension
- [x] Virtual Networking


tombuildsstuff:
Network Interfaces test:

screen shot 2018-01-23 at 15 45 55

tombuildsstuff:
Availability Sets tests pass:

screen shot 2018-01-23 at 16 09 46

tombuildsstuff:
Managed Disk tests pass:

screen shot 2018-01-23 at 16 55 19

tombuildsstuff:
Virtual Networking tests pass:

screen shot 2018-01-23 at 16 59 19

tombuildsstuff:
VM tests pass, ignoring a transient service error:

screen shot 2018-01-24 at 08 38 41


Image tests pass:

screen shot 2018-01-24 at 08 38 37

tombuildsstuff:
Data Source tests pass:

screen shot 2018-01-24 at 14 00 05

tombuildsstuff:
Snapshot tests now pass:

screen shot 2018-01-24 at 14 41 57

Re-running the Data Source tests given this will impact them too

azurerm/data_source_image.go
```diff
@@ -134,28 +135,24 @@ func dataSourceArmImageRead(d *schema.ResourceData, meta interface{}) error {
...
 				d.SetId("")
 				return nil
 			}
 			return fmt.Errorf("[ERROR] Error getting list of images (resource group %q): %+v", resGroup, err)
 		}
-		for _, ri := range *resp.Value {
-			if r.Match(([]byte)(*ri.Name)) {
-				list = append(list, ri)
+
+		for resp.NotDone() {
```
paultyng: Do you know if the `resp` object will abort if the cancel on the `StopContext` is invoked? If not, we may want a `select` in here on the `StopContext.Done`, but not a big deal, I don't think it should hold up this pr as its an edge case to have enough images where it would be an issue I think.

tombuildsstuff: not from what I can see, whilst it's a valid concern it's such an edge case (and minor issue in terms of iterations) I think we can safely ignore it for the moment

PR #755, opened 2018-01-25
added missing built-in roles. open issues 670 and 671 will be fixed.

agnelantony2:
Working with Arun Chandrasekhar from Microsoft on this issue. We are trying use Terraform modules for Azure Role Definitions and Assignments. Looks like current code supports only 4 built-in roles. Hence added missing built-in roles and VirtualMachineOperator had an incorrect GUID. Updated that GUID as well.

Thanks,
Agnel


paultyng:
Where did the data come from? Is it somewhere we could `curl`? Maybe write a little script to generate this?

tombuildsstuff:
hey @agnelantony2 

As mentioned above I've added a commit to fix the tests - however unfortunately I'm unable to push to your fork (so that we could get this into the release which is happening shortly).

I hope you don't mind, but I've re-opened this PR as #762 which includes your original commit and the follow up that I've done - but this will allow us to merge this in as part of the current release. As such I'm going to close this PR for the moment - but I'd like to thank you for this contribution - and it's worth noting that the original commit is still authored by you :)

Thanks!

Issue #724, opened 2018-01-19
Feature Request: azurerm_storage_account add name_prefix


jansepke:
Hi,

because the name of a azurerm_storage_account must be unique, it would be great to let terraform generate parts of it using a prefix. this would be the same behavior as in [aws_s3_bucket]
(https://www.terraform.io/docs/providers/aws/r/s3_bucket.html) with bucket and bucket_prefix.

### Terraform Version
Terraform v0.11.1
+ provider.azurerm v1.0.0

### Affected Resource(s)
- azurerm_storage_account

tombuildsstuff:
hey @jansepke 

Apologies for the delayed response here!

As discussed in #767 - I'm going to close this in favour of hashicorp/terraform#17389 since this is a Core issue; chatting about this internally, I believe that @katbyte's suggestion of an interpolation function could be a nice solution to this.

Thanks!

PR #767, opened 2018-01-29
[WIP, Question] add name_prefix to storage_account

jansepke:
this should solve #724

this uses the pattern from aws_s3_bucket to easily create unique storage_account names. This could be also implemented for sql_server or cdn_endpoint.

My current question / problem is that the name can only be 24 chars long and the generated Id from the PrefixedUniqueId function is quite long. I could cut the generated name after 24 chars but then most of the randomness would get lost.

A possible solution could be a random hash generator function in this repo or in the terraform main repo. What do you think?

Todo:
* Tests
* Docs

### DIFF
--- a/azurerm/resource_arm_storage_account.go
+++ b/azurerm/resource_arm_storage_account.go
@@ -30,8 +30,16 @@ func resourceArmStorageAccount() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"name": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"name_prefix"},
+				ValidateFunc:  validateArmStorageAccountName,
+			},
+			"name_prefix": {
 				Type:         schema.TypeString,
-				Required:     true,
+				Optional:     true,
 				ForceNew:     true,
 				ValidateFunc: validateArmStorageAccountName,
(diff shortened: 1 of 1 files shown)


radeksimko:
> I'd be interested to hear your thoughts here on how this prefix has worked out in the AWS Provider?

I was originally hesitant, but generally speaking we keep accepting PRs with such fields in the AWS land. I can see how this approach looks cleaner and shorter. 🤷‍♂️ 

That said we should have a helper in the schema perhaps, so that no provider/resource has to duplicate the code we see here and come up with their own unique ways of appending random string.

I just opened https://github.com/hashicorp/terraform/issues/17389 to track this.

jansepke:
Should I close this one and open a new one when hashicorp/terraform#17389 is solved? or will this take too long so we do this one here as an intermediate work around?

tombuildsstuff:
hey @jansepke 

After chatting about this internally we've decided against taking this approach in the Azure Provider - whilst we support this in the other providers (since one resource added support for it and then others followed); I have a feeling this'd be better supported as an extension of Terraform Core (either as an Interpolation Function [as suggested by @katbyte in hashicorp/terraform#17389] or as a Schema property [as was previously discussed above]).

Given this is a core enhancement, I'm going to close this PR in favour of the issue being tracked in hashicorp/terraform#17389.

Thanks!

PR #770, opened 2018-01-29
d/`azurerm_builtin_role_definition` - loading available role definitions from Azure

jansepke:
and fix a small typo.

should we remove the validation list so we do not need to add new roles?


paultyng:
Are the API names the same as the ones already listed? In all cases? We may need some mapping for legacy names if not. I'll have to run some acceptance testing.

jansepke:
I made a [diff check](https://www.diffchecker.com/CbfmjERg) with this result:

new in the API:
- Lab Creator
- Resource Policy Contributor (Preview)
- Storage Blob Data Contributor (Preview)
- Storage Blob Data Reader (Preview)
- Storage Queue Data Contributor (Preview)
- Storage Queue Data Reader (Preview)
- Virtual Machine Contributor

removed in the API:
- Auditor
- CloudAware Collector Storage Account Keys Access
- DB Admin
- Lab Accounts User
- Network Security Operator
- VirtualMachineContributor

for the VirtualMachineContributor role we could do a mapping to Virtual Machine Contributor.
for all other removed roles: could this depend on the subscription or something user or region specific?

paultyng:
I'll look in to this and reply back here when I track down an answer

tombuildsstuff:
@jansepke @paultyng 

> for all other removed roles: could this depend on the subscription or something user or region specific?

My guess would be that's more likely to be subscription specific (given resources are enabled by Resource Provider)

> `Storage Queue Data Contributor (Preview)`

We need to proceed carefully here, since the ` (Preview)` suffix will be renamed/removed during the Preview cycle - and users code will stop working since Terraform won't be able to find the value in the API. It may be able to work around this by placing a (yellow) info block in the documentation (see an example below) - but we may also want to see if it's possible to return roles which are stable?

```
~> **NOTE:** Azure roles with the suffix `(Preview)` are not necessarily stable and may change at any time.
```

jansepke:
@tombuildsstuff I added a validation that will error on preview roles

azurerm/data_source_builtin_role_definition.go
```diff
@@ -135,85 +135,22 @@ func dataSourceArmBuiltInRoleDefinitionRead(d *schema.ResourceData, meta interfa
...
-		"Storage Account Contributor":               "/providers/Microsoft.Authorization/roleDefinitions/17d1049b-9a84-46fb-8f53-869881c3d3ab",
-		"Storage Account Key Operator Service Role": "/providers/Microsoft.Authorization/roleDefinitions/81a9662b-bebf-436f-a333-f67b29880f12",
-		"Support Request Contributor":               "/providers/Microsoft.Authorization/roleDefinitions/cfd33db0-3dd1-45e3-aa9d-cdbdf3b6f24e",
-		"Traffic Manager Contributor":               "/providers/Microsoft.Authorization/roleDefinitions/a4b10055-b0c7-44c2-b00f-c7b5b3550cf7",
-		"User Access Administrator":                 "/providers/Microsoft.Authorization/roleDefinitions/18d7d88d-d35e-4fb5-a5c3-7773c20a72d9",
-		"VirtualMachineContributor":                 "/providers/Microsoft.Authorization/roleDefinitions/9980e02c-c2be-4d73-94e8-173b1dc7cf3c",
-		"Web Plan Contributor":                      "/providers/Microsoft.Authorization/roleDefinitions/2cc479cb-7b4d-49a8-b449-8c00fd0f0a4b",
-		"Website Contributor":                       "/providers/Microsoft.Authorization/roleDefinitions/de139f84-1756-47ae-9be6-808fbbe84772",
+	filter := fmt.Sprintf("roleName eq '%s'", name)
```
tombuildsstuff: in order to keep compatibility with folks using the `VirtualMachineContributor` key (e.g. so that we don't need to bump the major version number of the Provider) - can we add an if statement above this, which we can remove in a future version of Terraform? e.g.

```
if name == "VirtualMachineContributor" {
  name = "Virtual Machine Contributor"
}
filter := fmt.Sprintf("roleName eq '%s'", name)
```

jansepke:
@tombuildsstuff I added the if statement and also removed the validation, because this will now happen on the azure API side

tombuildsstuff:
Tests pass:

screen shot 2018-03-12 at 14 08 46

PR #772, opened 2018-01-29
Set ip address on public_ip regardless of dns settings.

genevieve:
https://github.com/terraform-providers/terraform-provider-azurerm/issues/771#issuecomment-361359112


tombuildsstuff:
hey @genevieve 

Thanks again for this PR - I've tried pushing a commit to add an acceptance test to ensure this issue is caught in future (as suggested by @TheLinuxNinja in https://github.com/terraform-providers/terraform-provider-azurerm/issues/771#issuecomment-362008303), however I don't have permissions to push this to your fork.

So that we can get this in - I've opened #789 which includes the acceptance test, which we'll merge after this PR has been merged in.

Thanks!

PR #778, opened 2018-01-30
Fixes issue when we need to attach/detach data disks 

VaijanathB:
If we wanted to attach and detach a dataDisk after the VM creation, this was causing reprovisioning of the VM. By removing these two flags, now we can create new data disks and attach them to the existing vm.

Test cases run
1) Create a VM. Then attach a Data Disk
2) Detach a data Disk from the VM.
3) Delete VM to delete the resources too. 

This is a blocking issue for Sapient who are relying on Terraform for their third party tool.  
The fix should fix this issue. https://github.com/terraform-providers/terraform-provider-azurerm/issues/85


tombuildsstuff:
the VM tests pass:

screen shot 2018-02-08 at 09 49 08

tombuildsstuff:
@VaijanathB would you be able to rebase this PR?

VaijanathB:
This is being merged as part of "Fix reprovisioning of VM when we attach and detach data disk. #813"

Issue #582, opened 2017-11-22
Please allow data disks to be added to existing machines in inventory in-line without destroy/recreate


jstewart612:
The Azure Resource Manager control panel lets you attach a data disk without destroying the machine, so why can't Terraform?

nbering:
I was able to do this smoothly in past releases, though I admit I probably haven't tried since Terraform 0.9.11. What version are you using? Can you give and example of the configuration, the specifics of how you changed it, and the terraform plan output (with secrets removed)? This not only helps with diagnostics, but also helps other users looking at this later to determine if the issue they're discussing is the same one you're facing.

jstewart612:
main.tf

```
# Subscription-wide values
variable "client_id"                    {}
variable "client_secret"                {}
variable "subscription_id"              {}
variable "tenant_id"                    {}

# Terraform Remote State values
variable "storage_account_name"         {}
variable "container_name"               {}
variable "key"                          {}

# Data Center and Environment Options
variable "location"                     {}
variable "resource_group_name"          {}

# Availability Set Options
variable "platform_update_domain_count" {}
variable "platform_fault_domain_count"  {}
variable "managed"                      {}

# Configure the Microsoft Azure Provider
provider "azurerm" {
  subscription_id = "${var.subscription_id}"
  client_id       = "${var.client_id}"
  client_secret   = "${var.client_secret}"
  tenant_id       = "${var.tenant_id}"
}

module "rentpath-appgw" {
  source = "modules/rentpath-appgw"

  # Subscription-wide values
  subscription_id     = "${var.subscription_id}"
  client_id           = "${var.client_id}"
  client_secret       = "${var.client_secret}"
  tenant_id           = "${var.tenant_id}"
  resource_group_name = "${var.resource_group_name}"
  location            = "${var.location}"

  # Data Center and Environment Options
  location            = "${var.location}"
  resource_group_name = "${var.resource_group_name}"

  # Availability Set options
  platform_update_domain_count = "${var.platform_update_domain_count}"
  platform_fault_domain_count  = "${var.platform_fault_domain_count}"
  managed                      = "${var.managed}"
}

#module "haproxy-int" {
#  source = "modules/haproxy-int"

  # Subscription-wide values
#  subscription_id     = "${var.subscription_id}"
#  client_id           = "${var.client_id}"
#  client_secret       = "${var.client_secret}"
#  tenant_id           = "${var.tenant_id}"
[...]

PR #792, opened 2018-02-04
Set ignore case function for diffs on the managed data disk ids.

jzampieron:
This provides a way to work around #582 by using the Azure Portal
to attach disks.

### DIFF
--- a/azurerm/resource_arm_virtual_machine.go
+++ b/azurerm/resource_arm_virtual_machine.go
@@ -252,10 +252,11 @@ func resourceArmVirtualMachine() *schema.Resource {
 						},
 
 						"managed_disk_id": {
-							Type:     schema.TypeString,
-							Optional: true,
-							ForceNew: true,
-							Computed: true,
+							Type:             schema.TypeString,
+							Optional:         true,
+							ForceNew:         true,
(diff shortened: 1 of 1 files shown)


jzampieron:
Just as a follow up. I do have an open ticket with Azure support to discuss the expected behavior here. 

At the very least the Azure documentation needs to reflect the case behavior of these REST values.

tombuildsstuff:
Ignoring some temporary test failures in Azure - the tests LGTM:

screen shot 2018-02-06 at 11 12 26

tombuildsstuff:
> Just as a follow up. I do have an open ticket with Azure support to discuss the expected behavior here.
> At the very least the Azure documentation needs to reflect the case behavior of these REST values.

@jzampieron yeah, that sounds sensible - @metacpp should be able to raise the issue with the Portal team

tombuildsstuff:
@jzampieron just to let you know that this has been released in v1.1.1 :)

jzampieron:
As an update. Azure support was able to reproduce the behavior and is discussing further with the Portal engineering team to understand if this is expected, desired, etc and generate appropriate follow-up actions.

I will track it here as I remember to.

Issue #370, opened 2017-09-29
Feature Request - Availability Zones


tomasquith:
Hi, Following the announcements at Ignite it would be nice to explore the Availability Zones functionality which is now in preview via terraform. It's in a limited preview at the moment, only being in the one US region. It would be good to trial this as it means we can closely replicate AWS in terms of availability.

What is the stance on supporting preview features via terraform? 

https://azure.microsoft.com/en-gb/blog/introducing-azure-availability-zones-for-resiliency-and-high-availability/

justaugustus:
@metacpp pinging you on this one, as you assigned yourself to #579.

tombuildsstuff:
Support for this was added in #811 - closing.

PR #811, opened 2018-02-08
Add availability zone support

nordbergm:
Hey all,

This PR adds availability zone support to resources that are included in the AZ public preview. I'm unclear on what the policy is on preview features in Terraform, but it would be great if we could get this in.

It should also address feature request #370.

#### Updated Resources
* `azurerm_virtual_machine`
* `azurerm_virtual_machine_scale_set`
* `azurerm_public_ip`
* `azurerm_managed_disk `
* `azurerm_lb`

#### Updated Data Sources
* `azurerm_managed_disk `

### DIFF
--- a/azurerm/resource_arm_loadbalancer.go
+++ b/azurerm/resource_arm_loadbalancer.go
@@ -292,7 +292,7 @@ func expandAzureRmLoadBalancerFrontendIpConfigurations(d *schema.ResourceData) *
 		}
 
 		name := data["name"].(string)
-		zones := zoneValuesToStrings(data["zones"].([]interface{}))
+		zones := expandZones(data["zones"].([]interface{}))
 		frontEndConfig := network.FrontendIPConfiguration{
 			Name: &name,
 			FrontendIPConfigurationPropertiesFormat: &properties,
--- a/azurerm/resource_arm_loadbalancer.go
+++ b/azurerm/resource_arm_loadbalancer.go
@@ -86,7 +86,7 @@ func resourceArmLoadBalancer() *schema.Resource {
 							Set:      schema.HashString,
 						},
 
-						"zones": zonesSchema(),
+						"zones": singleZonesSchema(),
 					},
 				},
 			},
@@ -217,7 +217,6 @@ func resourecArmLoadBalancerRead(d *schema.ResourceData, meta interface{}) error
 
 			privateIpAddress := ""
 			privateIpAddresses := make([]string, 0, len(*feipConfigs))
-			zones := make([]string, 0, len(*feipConfigs))
 			for _, config := range *feipConfigs {
 				if feipProps := config.FrontendIPConfigurationPropertiesFormat; feipProps != nil {
 					if ip := feipProps.PrivateIPAddress; ip != nil {
@@ -228,15 +227,10 @@ func resourecArmLoadBalancerRead(d *schema.ResourceData, meta interface{}) error
 						privateIpAddresses = append(privateIpAddresses, *feipProps.PrivateIPAddress)
 					}
 				}
-				configZones := config.Zones
-				for _, zone := range *configZones {
-					zones = append(zones, zone)
-				}
 			}
 
 			d.Set("private_ip_address", privateIpAddress)
 			d.Set("private_ip_addresses", privateIpAddresses)
-			d.Set("zones", zones)
 		}
 	}
 
@@ -316,6 +310,7 @@ func flattenLoadBalancerFrontendIpConfiguration(ipConfigs *[]network.FrontendIPC
 	for _, config := range *ipConfigs {
 		ipConfig := make(map[string]interface{})
 		ipConfig["name"] = *config.Name
+		ipConfig["zones"] = *config.Zones
 
 		if props := config.FrontendIPConfigurationPropertiesFormat; props != nil {
 			ipConfig["private_ip_address_allocation"] = props.PrivateIPAllocationMethod
--- a/azurerm/resource_arm_loadbalancer.go
+++ b/azurerm/resource_arm_loadbalancer.go
@@ -85,6 +85,8 @@ func resourceArmLoadBalancer() *schema.Resource {
 							Elem:     &schema.Schema{Type: schema.TypeString},
 							Set:      schema.HashString,
 						},
+
+						"zones": zonesSchema(),
 					},
 				},
 			},
@@ -215,6 +217,7 @@ func resourecArmLoadBalancerRead(d *schema.ResourceData, meta interface{}) error
 
 			privateIpAddress := ""
 			privateIpAddresses := make([]string, 0, len(*feipConfigs))
+			zones := make([]string, 0, len(*feipConfigs))
 			for _, config := range *feipConfigs {
 				if feipProps := config.FrontendIPConfigurationPropertiesFormat; feipProps != nil {
 					if ip := feipProps.PrivateIPAddress; ip != nil {
@@ -225,10 +228,15 @@ func resourecArmLoadBalancerRead(d *schema.ResourceData, meta interface{}) error
 						privateIpAddresses = append(privateIpAddresses, *feipProps.PrivateIPAddress)
 					}
 				}
+				configZones := config.Zones
+				for _, zone := range *configZones {
+					zones = append(zones, zone)
+				}
 			}
 
 			d.Set("private_ip_address", privateIpAddress)
 			d.Set("private_ip_addresses", privateIpAddresses)
+			d.Set("zones", zones)
 		}
 	}
 
@@ -290,9 +298,11 @@ func expandAzureRmLoadBalancerFrontendIpConfigurations(d *schema.ResourceData) *
 		}
 
 		name := data["name"].(string)
+		zones := zoneValuesToStrings(data["zones"].([]interface{}))
 		frontEndConfig := network.FrontendIPConfiguration{
 			Name: &name,
 			FrontendIPConfigurationPropertiesFormat: &properties,
+			Zones: zones,
 		}
 
 		frontEndConfigs = append(frontEndConfigs, frontEndConfig)
--- a/azurerm/resource_arm_loadbalancer.go
+++ b/azurerm/resource_arm_loadbalancer.go
@@ -310,7 +310,14 @@ func flattenLoadBalancerFrontendIpConfiguration(ipConfigs *[]network.FrontendIPC
 	for _, config := range *ipConfigs {
 		ipConfig := make(map[string]interface{})
 		ipConfig["name"] = *config.Name
-		ipConfig["zones"] = *config.Zones
+
+		zones := make([]string, 0)
+		if zs := config.Zones; zs != nil {
+			for _, zone := range *zs {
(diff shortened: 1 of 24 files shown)


tombuildsstuff:
Fixes #370

nordbergm:
hey @tombuildsstuff,

Thanks for the review! Your pointers should be sorted now, but let me know if there's anything else.

azurerm/resource_arm_managed_disk.go
```diff
@@ -135,7 +138,8 @@ func resourceArmManagedDiskCreate(d *schema.ResourceData, meta interface{}) erro
 		Sku: &compute.DiskSku{
 			Name: (skuName),
 		},
-		Tags: expandedTags,
+		Tags:  expandedTags,
+		Zones: zones,
```
tombuildsstuff: Given we set this value always - if a user isn't opted into the Preview most deployments will fail:

```
* azurerm_managed_disk.test: compute.DisksClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="ResourceTypeNotSupportAvailabilityZones" Message="The resource type 'disks' does not support availability zones at location 'westeurope' and api-version '2017-03-30'."
```

Can we instead change this to assign Zones only if the field is set? e.g.

```
if v, ok := d.Get("zones"); ok {
  zones := expandZones(v.([]interface{}))
  createDisk.Zones = zones
}
```

which should remove this error

azurerm/resource_arm_loadbalancer.go
```diff
@@ -290,9 +292,11 @@ func expandAzureRmLoadBalancerFrontendIpConfigurations(d *schema.ResourceData) *
 		}
 
 		name := data["name"].(string)
+		zones := expandZones(data["zones"].([]interface{}))
 		frontEndConfig := network.FrontendIPConfiguration{
 			Name: &name,
 			FrontendIPConfigurationPropertiesFormat: &properties,
+			Zones: zones,
```
tombuildsstuff: from what I can see, this will be the same issue as below (where a user isn't opted into the Preview, this fails) - can we update it to use the pattern identified below?

azurerm/resource_arm_public_ip.go
```diff
@@ -157,7 +160,8 @@ func resourceArmPublicIpCreate(d *schema.ResourceData, meta interface{}) error {
 		Location: &location,
 		Sku:      &sku,
 		PublicIPAddressPropertiesFormat: &properties,
-		Tags: expandTags(tags),
+		Tags:  expandTags(tags),
+		Zones: zones,
```
tombuildsstuff: from what I can see, this will be the same issue as above (where a user isn't opted into the Preview, this fails) - can we update it to use the pattern identified below?

azurerm/resource_arm_virtual_machine.go
```diff
@@ -577,7 +580,8 @@ func resourceArmVirtualMachineCreate(d *schema.ResourceData, meta interface{}) e
 		Name:                     &name,
 		Location:                 &location,
 		VirtualMachineProperties: &properties,
-		Tags: expandedTags,
+		Tags:  expandedTags,
+		Zones: zones,
```
tombuildsstuff: from what I can see, this will be the same issue as above (where a user isn't opted into the Preview, this fails) - can we update it to use the pattern identified below?

azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -564,6 +566,7 @@ func resourceArmVirtualMachineScaleSetCreate(d *schema.ResourceData, meta interf
 	location := d.Get("location").(string)
 	resGroup := d.Get("resource_group_name").(string)
 	tags := d.Get("tags").(map[string]interface{})
+	zones := expandZones(d.Get("zones").([]interface{}))
```
tombuildsstuff: from what I can see, this will be the same issue as above (where a user isn't opted into the Preview, this fails) - can we update it to use the pattern identified below?

azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -632,6 +635,7 @@ func resourceArmVirtualMachineScaleSetCreate(d *schema.ResourceData, meta interf
 		Tags:     expandTags(tags),
 		Sku:      sku,
 		VirtualMachineScaleSetProperties: &scaleSetProps,
+		Zones: zones,
```
tombuildsstuff: from what I can see, this will be the same issue as above (where a user isn't opted into the Preview, this fails) - can we update it to use the pattern identified below?

tombuildsstuff:
hey @nordbergm 

Thanks for pushing those updates - I kicked off the test suite here without the Preview enabled (to confirm the existing tests work) - however the existing tests fail as we're setting the Zones field regardless:

```
* azurerm_managed_disk.test: compute.DisksClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="ResourceTypeNotSupportAvailabilityZones" Message="The resource type 'disks' does not support availability zones at location 'westeurope' and api-version '2017-03-30'."
```

Can we update the assignment of the `Zones` field to check if there's any items before setting it, [as shown in this comment](https://github.com/terraform-providers/terraform-provider-azurerm/pull/811#discussion_r167732620) which would fix the issue. Once that's done - both the existing and the new tests should pass as expected :)

Thanks!

nordbergm:
hey @tombuildsstuff,

Sorry about those test failures. I initially looked at the Azure Go client and how `Zones` marshalling is defined as `omitempty`.

```go
Zones           *[]string `json:"zones,omitempty"`
```

My expectation was that it would omit an empty array from the request. After some testing it seems this is generally the case, but since the field is defined as a pointer, the pointer has to be `nil` to be omitted.

A perhaps simpler and more central change, would be to have `expandZones` return `nil` in case the array is empty. How do you feel about that approach? I'm rather new with Go, so maybe this is some big no no.

nordbergm:
@tombuildsstuff I pushed an example of my suggested approach. Let me know what you think, and if you still experience any issues with the tests.

agup006:
@tombuildsstuff is the approach @nordbergm ok? Any updates are appreciated

tombuildsstuff:
@agup006 @nordbergm sorry for the delay in re-reviewing this - we're trying to work through the review backlog - I'm taking a look into this now.

azurerm/resource_arm_loadbalancer.go
```diff
@@ -306,6 +310,7 @@ func flattenLoadBalancerFrontendIpConfiguration(ipConfigs *[]network.FrontendIPC
 	for _, config := range *ipConfigs {
 		ipConfig := make(map[string]interface{})
 		ipConfig["name"] = *config.Name
+		ipConfig["zones"] = *config.Zones
```
tombuildsstuff: so there's a crash here if `config.Zones` is nil - can we make this:

```
zones := make([]string, 0)
if zs := config.Zones; zs != nil {
  for _, zone := range *zs {
    zones = append(zones, zone)
  }
}

ipConfig["zones"] = zones
```

tombuildsstuff:
hey @nordbergm @agup006 

Thanks for pushing the latest updates - apologies about the delayed review here! The latest updates looks good, however running the tests there appears to be a crash in the Load Balancer resource (which I've commented in-line about)

If we can sort that out, we'll take another look through/kick of the tests - but this should be good :)

Thanks!

---

Tests:

Data Source tests pass:

screenshot-1-data-source

Managed Disk resource tests pass:

screenshot-2-managed-disks

Public Ip tests pass:

screenshot-3-public-ip

Reminder of the test prefixes for me, for later:

```
refs/pull/811/merge

TestAccDataSourceAzureRMManagedDisk_
TestAccAzureRMLoadBalancer_
TestAccAzureRMManagedDisk_
TestAccAzureRMPublicIp
TestAccAzureRMVirtualMachineScaleSet_
TestAccAzureRMVirtualMachine_
```

azurerm/resource_arm_loadbalancer.go
tombuildsstuff: I hope you don't mind, but I've pushed a commit to fix this

tombuildsstuff:
Load Balancer tests pass:

screen shot 2018-03-06 at 09 41 54

tombuildsstuff:
VM/VMSS tests pass, ignoring a transient Azure failure:

screen shot 2018-03-06 at 10 50 59

nordbergm:
Thanks @tombuildsstuff!

lubars:
Availability zones are working for me in combination with a load balancer (even when everything is in the same zone, "1").  The first error I'm getting is:

`azurerm_lb.default: Error Creating/Updating Load Balancer "lbtest" (Resource Group "lbtest"): network.LoadBalancersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="LoadBalancerFrontendIPConfigCannotHaveZoneWhenReferencingPublicIPAddress" Message="Load balancer frontendIPConfiguration /subscriptions/56364a0d-3867-4679-8171-e141f4cc87fc/resourceGroups/lbtest/providers/Microsoft.Network/loadBalancers/lbtest/frontendIPConfigurations/PublicIPAddress has 1 zone specified and is referencing a publicIPAddress /subscriptions/56364a0d-3867-4679-8171-e141f4cc87fc/resourceGroups/lbtest/providers/Microsoft.Network/publicIPAddresses/lbtest. Networking supports zones only for frontendIpconfigurations which reference a subnet." Details=[]`

Is it actually the case that zonal load balancers may not have a public IP address?  I then did a second test, replacing the public_ip_address_id with the subnet_id (in the frontend_ip_configuration of the azurerm_lb resource); this worked for a single VM, but with two or more VMs, it looked like only the first VM got added to the backend address pool, and the rest were rejected:

`azurerm_virtual_machine.default.1: Code="VmIsNotInSameAvailabilitySetAsLb" Message="Virtual Machine /subscriptions/56364a0d-3867-4679-8171-e141f4cc87fc/resourceGroups/lbtest/providers/Microsoft.Compute/virtualMachines/lbtest2 is using different Availability Set than other Virtual Machines connected to the Load Balancer(s) lbtest."
azurerm_virtual_machine.default.2: Code="VmIsNotInSameAvailabilitySetAsLb" Message="Virtual Machine /subscriptions/56364a0d-3867-4679-8171-e141f4cc87fc/resourceGroups/lbtest/providers/Microsoft.Compute/virtualMachines/lbtest3 is using different Availability Set than other Virtual Machines connected to the Load Balancer(s) lbtest."
`

This message is somewhat mysterious because it refers to availability sets, all references to which had been removed after receiving this error:

`azurerm_virtual_machine.default: compute.VirtualMachinesClient#CreateOrUpdate: 
Failure sending request: StatusCode=400 -- Original Error: Code="BadRequest" 
Message="Virtual Machine cannot be created because both Availability Zone and 
Availability Set were specified. Deploying an Availability Set to an Availability 
Zone isn’t supported."
`

Thanks for any insight you can provide.

PR #826, opened 2018-02-13
Enable source control management as option for user to deploy web apps

metacpp:
This PR includes:

1. New argument, scm_type, as deployment option for web app. Supported value list include "None" and "LocalGit", will add more options in a later time.
2. New site_source_control_props with necessary git repo and branch information. 

[TestAccAzureRMAppService](https://gist.github.com/metacpp/027499f521813dc1476bb1fa92b1ff06)


azurerm/resource_arm_app_service.go
```diff
@@ -160,6 +160,15 @@ func resourceArmAppService() *schema.Resource {
...
 						},
+						"scm_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  string(web.ScmTypeNone),
+							ValidateFunc: validation.StringInSlice([]string{
+								string(web.ScmTypeNone),
+								string(web.ScmTypeLocalGit),
+							}, false),
+						},
```
tombuildsstuff: so it'd also be good to expose the value of the SCM URL as a computed field - which I believe should be returned from [the `GetSourceControl` API call when this is set to `ScmTypeLocalGit`](https://docs.microsoft.com/en-us/rest/api/appservice/webapps/getsourcecontrol)?

website/docs/r/app_service.html.markdown
```diff
@@ -160,6 +162,8 @@ The following arguments are supported:
 
 * `websockets_enabled` - (Optional) Should WebSockets be enabled?
 
+* `scm_type` - (Optional) The Source Control Management Type. Possible values are `None` and `LocalGit`. Defaults to `None`.
```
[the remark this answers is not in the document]
metacpp: Good suggestion.

metacpp:
@tombuildsstuff I've updated the PR to include the repo url and branch name.

azurerm/resource_arm_app_service.go
```diff
@@ -239,6 +248,23 @@ func resourceArmAppService() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"source_control": {
+				Type:     schema.TypeList,
+				Optional: true,
```
tombuildsstuff: can we remove this `Optional` assignment? it's only needed for user specifiable fields (and not readonly ones)

website/docs/r/app_service.html.markdown
```diff
@@ -170,6 +176,15 @@ The following attributes are exported:
...
 * `outbound_ip_addresses` - A comma separated list of outbound IP addresses - such as `52.23.25.3,52.143.43.12`
 
+* `source_control` - (Optional) The default local Git source control information if deployment option is set to `LocalGit`.
+
+---
+
+`source_control` supports the following:
+
+* `repo_url` - URL of the Git repo.
+* `branch` - Branch name of the Git repo.
```
tombuildsstuff: can we suffix both of these with `.. the Git repository for this App Service`

azurerm/resource_arm_app_service_test.go
```diff
@@ -1259,3 +1281,34 @@ resource "azurerm_app_service" "test" {
 }
 `, rInt, location, rInt, rInt)
 }
+
+func testAccAzureRMAppService_scmType(rInt int, location string) string {
+	return fmt.Sprintf(`
+resource "azurerm_resource_group" "test" {
+  name = "acctestRG-%d"
+  location = "%s"
```
JunyiYi: Aligning at '='?

azurerm/resource_arm_app_service_test.go
```diff
@@ -1259,3 +1281,34 @@ resource "azurerm_app_service" "test" {
...
+}
+
+resource "azurerm_app_service" "test" {
+  name                = "acctestAS-%d"
+  location            = "${azurerm_resource_group.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  app_service_plan_id = "${azurerm_app_service_plan.test.id}"
+
+  site_config {
+    scm_type          = "LocalGit"
```
JunyiYi: Reduce spaces here?

azurerm/resource_arm_app_service.go
metacpp: Good catch.

azurerm/resource_arm_app_service_test.go
metacpp: Good catch.

azurerm/resource_arm_app_service_test.go
metacpp: Good catch.

azurerm/resource_arm_app_service.go
```diff
@@ -239,6 +248,23 @@ func resourceArmAppService() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"site_source_control_props": {
```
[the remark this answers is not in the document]
metacpp: Sure.

azurerm/resource_arm_app_service.go
```diff
@@ -239,6 +248,23 @@ func resourceArmAppService() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
```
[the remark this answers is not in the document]
metacpp: Yes.

website/docs/r/app_service.html.markdown
metacpp: Done.

metacpp:
@tombuildsstuff the PR has been updated, please check.

metacpp:
@tombuildsstuff Thanks for the reviewing.

tombuildsstuff:
@metacpp given App Service Slots shares the same `Set` function - it appears we need to add the field to that schema too:

```
------- Stdout: -------
=== RUN   TestAccAzureRMAppServiceSlot_importBasic
--- FAIL: TestAccAzureRMAppServiceSlot_importBasic (51.57s)
    testing.go:459: Step 0 error: Error applying: 1 error(s) occurred:
        
        * azurerm_app_service_slot.test: 1 error(s) occurred:
        
        * azurerm_app_service_slot.test: Invalid address to set: []string{"site_config", "0", "scm_type"}
    testing.go:519: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.
        
        Error: Error refreshing: 1 error(s) occurred:
        
        * azurerm_app_service_slot.test: azurerm_app_service_slot.test: Invalid address to set: []string{"site_config", "0", "scm_type"}
        
        State: <nil>
FAIL
```

I'll push a commit for this today

tombuildsstuff:
hey @metacpp 

So I've been looking into this this afternoon - it appears the branch isn't returned until it's actually configured in Kudu. Here's the contents of the Statefile when the `LocalSCM` field is set to `LocalGit`:

```
"site_config.#": "1",
"site_config.0.always_on": "false",
"site_config.0.default_documents.#": "0",
"site_config.0.dotnet_framework_version": "v4.0",
"site_config.0.java_container": "",
"site_config.0.java_container_version": "",
"site_config.0.java_version": "",
"site_config.0.local_mysql_enabled": "false",
"site_config.0.managed_pipeline_mode": "Integrated",
"site_config.0.php_version": "",
"site_config.0.python_version": "",
"site_config.0.remote_debugging_enabled": "false",
"site_config.0.remote_debugging_version": "VS2012",
"site_config.0.scm_type": "LocalGit",
"site_config.0.use_32_bit_worker_process": "false",
"site_config.0.websockets_enabled": "false",
"source_control.#": "1",
"source_control.0.branch": "",
"source_control.0.repo_url": "https://tom-slot.scm.azurewebsites.net",
```

Given this PR isn't really usable without the settings configured in Kudu - I think we're going to need to park this PR until Kudu support is added to the Azure SDK for Go [which is being tracked in this issue](https://github.com/Azure/azure-sdk-for-go/issues/1065). When support is available in the Azure SDK for configuring the SCM Slot (via Kudu) we should be able to include this again.

Apologies for the bad news - unfortunately Kudu's the source of a lot of this information, and it's currently missing from the SDK.

Thanks!

metacpp:
@tombuildsstuff I also noticed that the branch is empty while writing the test case, that's why I only validated the source_control.#. Based on my previous experience while playing with LocalGit on Azure Web App, the branch is actually optional, coz by default it's always master. I don't remember that Azure provide the option for user to create other branches in that repo.

Besides the empty branch name, I'm more concerned on the deployment credential: user name and password. How can we get/set it today through Terraform ?

metacpp:
@tombuildsstuff the schema has been updated for slot. I think if that schema was completely the same as app service, we should consider making it reusable. @jeffreyCline any idea on it ?

tombuildsstuff:
@metacpp 

> Besides the empty branch name, I'm more concerned on the deployment credential: user name and password. How can we get/set it today through Terraform ?

From the investigation we did last week it appears at this point we can't - given it's all configured through Kudu and there's no Swagger/SDK for Kudu at this time; I don't believe we can't merge this PR until Kudu support exists in the SDK, unfortunately.

On that note, the `repo_url` being returned to us through the `GetSourceControl` method (e.g. `https://tom-slot.scm.azurewebsites.net`) appears to actually be the Kudu endpoint, rather than a Git endpoint?

> I think if that schema was completely the same as app service, we should consider making it reusable

It is for the fields which currently exist, but we've not built out the rest of the Schema yet so there may be differences; whilst this would be good to do at some point we should probably hold off until we support more of the schema (so we can safely say the two are identical)

---

Given this PR appears to be blocked on Kudu - I'm going to suggest we close it for the moment, and [once Kudu support has been added to the Azure SDK for Go](https://github.com/Azure/azure-sdk-for-go/issues/1065) we can re-open it and take advantage of the SCM information available from the Kudu site.

Thanks!

tombuildsstuff:
@metacpp 

> Very similar if we want to work around it.

whilst that's true for Azure Public, it's a different suffix for the other regions - whilst we could probably code around it, we really need the extra information that Kudu provides for this data to be useful.

> Let me involve me into that thread.

👍 sounds good

metacpp:
@tombuildsstuff I reopened this PR coz I worked with App Service team have the LocalGit deployment supported.

azurerm/resource_arm_app_service.go
```diff
@@ -226,6 +236,25 @@ func resourceArmAppService() *schema.Resource {
 			// https://github.com/Azure/azure-rest-api-specs/issues/1697
 			"tags": tagsForceNewSchema(),
 
+			"site_credential": {
+				Type:     schema.TypeList,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"user_name": {
```
tombuildsstuff: we'd tend to write this as `username` - I'll push a commit to make this consistent

tombuildsstuff:
App Service tests pass:

screen shot 2018-03-28 at 10 00 13

azurerm/resource_arm_app_service.go
metacpp: Thanks for the updating.

PR #833, opened 2018-02-13
Correcting line endings in Windows

Nepomuceno:
Just correcting the way that line endings are done in windows avoid git magic on line endings.

Similar to the way that the golang project uses.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,10 @@
+# Treat all files in the repo as binary, with no git magic updating
+# line endings. Windows users contributing will need to use a
+# modern version of git and editors capable of LF line endings.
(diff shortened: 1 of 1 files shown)


metacpp:
@tombuildsstuff I added one commit, in general, I think it's ready to be merged.

tombuildsstuff:
Hi @Nepomuceno 

After chatting about this internally, unfortunately we're not going to be able to accept this PR. To give some additional context here: we've previously accepted a similar change in the main Terraform repository that's since been reverted, due to issues/conflicts with line endings for vendored files (with `govendor`) when running on different platforms.

Thanks!

PR #834, opened 2018-02-14
Add support for custom IPsec/IKE policies for VPN connections

dominik-lekse:
This pull requests adds support for custom IPsec/IKE policies for VPN connections. Also, policy based traffic selectors can now be enabled explicitly for a connection.

### Tests

I have added a new test which covers a custom `ipsec_policy` and the `use_policy_based_traffic_selectors` switch.

```
TF_ACC=1 go test ./azurerm -v -run=TestAccAzureRMVirtualNetworkGatewayConnection_ipsecpolicy -timeout 180m
=== RUN   TestAccAzureRMVirtualNetworkGatewayConnection_ipsecpolicy
--- PASS: TestAccAzureRMVirtualNetworkGatewayConnection_ipsecpolicy (2401.33s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	2401.357s
```

### References

- Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/759

### DIFF
--- a/azurerm/resource_arm_virtual_network_gateway_connection.go
+++ b/azurerm/resource_arm_virtual_network_gateway_connection.go
@@ -79,19 +79,128 @@ func resourceArmVirtualNetworkGatewayConnection() *schema.Resource {
 			"enable_bgp": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
+				Computed: true,
+			},
+
+			"use_policy_based_traffic_selectors": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
 			},
 
 			"routing_weight": {
 				Type:     schema.TypeInt,
 				Optional: true,
-				Default:  10,
+				Computed: true,
 			},
 
 			"shared_key": {
 				Type:      schema.TypeString,
 				Optional:  true,
 				Sensitive: true,
+			},
+
+			"ipsec_policy": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"dh_group": {
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(network.DHGroup1),
+								string(network.DHGroup14),
+								string(network.DHGroup2),
+								string(network.DHGroup2048),
+								string(network.DHGroup24),
+								string(network.ECP256),
+								string(network.ECP384),
+								string(network.None),
+							}, true),
+						},
+						"ike_encryption": {
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(network.AES128),
+								string(network.AES192),
+								string(network.AES256),
+								string(network.DES),
+								string(network.DES3),
+							}, true),
+						},
+						"ike_integrity": {
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(network.MD5),
+								string(network.SHA1),
+								string(network.SHA256),
+								string(network.SHA384),
+							}, true),
+						},
+						"ipsec_encryption": {
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(network.IpsecEncryptionAES128),
+								string(network.IpsecEncryptionAES192),
+								string(network.IpsecEncryptionAES256),
+								string(network.IpsecEncryptionDES),
+								string(network.IpsecEncryptionDES3),
+								string(network.IpsecEncryptionGCMAES128),
+								string(network.IpsecEncryptionGCMAES192),
+								string(network.IpsecEncryptionGCMAES256),
+								string(network.IpsecEncryptionNone),
+							}, true),
+						},
+						"ipsec_integrity": {
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(network.IpsecIntegrityGCMAES128),
+								string(network.IpsecIntegrityGCMAES192),
+								string(network.IpsecIntegrityGCMAES256),
+								string(network.IpsecIntegrityMD5),
+								string(network.IpsecIntegritySHA1),
+								string(network.IpsecIntegritySHA256),
+							}, true),
+						},
+						"pfs_group": {
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(network.PfsGroupECP256),
+								string(network.PfsGroupECP384),
+								string(network.PfsGroupNone),
+								string(network.PfsGroupPFS1),
+								string(network.PfsGroupPFS2),
+								string(network.PfsGroupPFS2048),
+								string(network.PfsGroupPFS24),
+							}, true),
+						},
+						"sa_datasize": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							Computed:     true,
+							ValidateFunc: validation.IntAtLeast(1024),
+						},
+						"sa_lifetime": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							Computed:     true,
+							ValidateFunc: validation.IntAtLeast(300),
+						},
+					},
+				},
 			},
 
 			"tags": tagsSchema(),
@@ -180,7 +289,9 @@ func resourceArmVirtualNetworkGatewayConnectionRead(d *schema.ResourceData, meta
 		d.Set("virtual_network_gateway_id", conn.VirtualNetworkGateway1.ID)
 	}
 
-	d.Set("authorization_key", conn.AuthorizationKey)
+	if conn.AuthorizationKey != nil {
+		d.Set("authorization_key", conn.AuthorizationKey)
+	}
 
 	if conn.Peer != nil {
 		d.Set("express_route_circuit_id", conn.Peer.ID)
@@ -194,9 +305,26 @@ func resourceArmVirtualNetworkGatewayConnectionRead(d *schema.ResourceData, meta
 		d.Set("local_network_gateway_id", conn.LocalNetworkGateway2.ID)
 	}
 
-	d.Set("enable_bgp", conn.EnableBgp)
-	d.Set("routing_weight", conn.RoutingWeight)
-	d.Set("shared_key", conn.SharedKey)
+	if conn.EnableBgp != nil {
+		d.Set("enable_bgp", conn.EnableBgp)
+	}
+
+	if conn.UsePolicyBasedTrafficSelectors != nil {
+		d.Set("use_policy_based_traffic_selectors", conn.UsePolicyBasedTrafficSelectors)
+	}
+
+	if conn.RoutingWeight != nil {
+		d.Set("routing_weight", conn.RoutingWeight)
+	}
+
+	if conn.SharedKey != nil {
+		d.Set("shared_key", conn.SharedKey)
+	}
+
+	if conn.IpsecPolicies != nil {
+		ipsec_policies := flattenArmVirtualNetworkGatewayConnectionIpsecPolicies(conn.IpsecPolicies)
+		d.Set("ipsec_policy", ipsec_policies)
+	}
 
 	flattenAndSetTags(d, resp.Tags)
 
@@ -227,11 +355,9 @@ func resourceArmVirtualNetworkGatewayConnectionDelete(d *schema.ResourceData, me
 
 func getArmVirtualNetworkGatewayConnectionProperties(d *schema.ResourceData) (*network.VirtualNetworkGatewayConnectionPropertiesFormat, error) {
 	connectionType := network.VirtualNetworkGatewayConnectionType(d.Get("type").(string))
-	enableBgp := d.Get("enable_bgp").(bool)
 
 	props := &network.VirtualNetworkGatewayConnectionPropertiesFormat{
 		ConnectionType: connectionType,
-		EnableBgp:      &enableBgp,
 	}
 
 	if v, ok := d.GetOk("virtual_network_gateway_id"); ok {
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -79,13 +79,19 @@ func resourceArmVirtualNetworkGatewayConnection() *schema.Resource {
 			"enable_bgp": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
+				Computed: true,
```
tombuildsstuff: is this same default returned from Azure?

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -79,13 +79,19 @@ func resourceArmVirtualNetworkGatewayConnection() *schema.Resource {
...
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
 			},
 
 			"routing_weight": {
 				Type:     schema.TypeInt,
 				Optional: true,
-				Default:  10,
+				Computed: true,
```
tombuildsstuff: is this same default returned from Azure?

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -194,9 +305,26 @@ func resourceArmVirtualNetworkGatewayConnectionRead(d *schema.ResourceData, meta
 		d.Set("local_network_gateway_id", conn.LocalNetworkGateway2.ID)
 	}
 
-	d.Set("enable_bgp", conn.EnableBgp)
-	d.Set("routing_weight", conn.RoutingWeight)
-	d.Set("shared_key", conn.SharedKey)
+	if conn.EnableBgp != nil {
+		d.Set("enable_bgp", conn.EnableBgp)
```
tombuildsstuff: do we want to dereference this field?

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -180,7 +289,9 @@ func resourceArmVirtualNetworkGatewayConnectionRead(d *schema.ResourceData, meta
 		d.Set("virtual_network_gateway_id", conn.VirtualNetworkGateway1.ID)
 	}
 
-	d.Set("authorization_key", conn.AuthorizationKey)
+	if conn.AuthorizationKey != nil {
+		d.Set("authorization_key", conn.AuthorizationKey)
```
tombuildsstuff: providing this value is a string, we shouldn't need this check?

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -194,9 +305,26 @@ func resourceArmVirtualNetworkGatewayConnectionRead(d *schema.ResourceData, meta
...
+		d.Set("routing_weight", conn.RoutingWeight)
+	}
+
+	if conn.SharedKey != nil {
+		d.Set("shared_key", conn.SharedKey)
+	}
+
+	if conn.IpsecPolicies != nil {
+		ipsec_policies := flattenArmVirtualNetworkGatewayConnectionIpsecPolicies(conn.IpsecPolicies)
+		d.Set("ipsec_policy", ipsec_policies)
```
tombuildsstuff: given this is a complex object, can we check for errors here via:

```
if err := d.Set("ipsec_policy", ipsec_policies); err != nil {
  return fmt.Errorf("Error setting `ipsec_policy`: %+v", err)
}
```

_minor_ could we also rename the variable `ipsec_policies` to `ipsecPolicies` to match the other variables?

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -294,14 +420,27 @@ func getArmVirtualNetworkGatewayConnectionProperties(d *schema.ResourceData) (*n
...
+
+	if v, ok := d.GetOk("use_policy_based_traffic_selectors"); ok {
+		props.UsePolicyBasedTrafficSelectors = utils.Bool(v.(bool))
+	}
+
 	if v, ok := d.GetOk("routing_weight"); ok {
-		routingWeight := int32(v.(int))
-		props.RoutingWeight = &routingWeight
+		routing_weight := int32(v.(int))
+		props.RoutingWeight = &routing_weight
```
tombuildsstuff: _minor_ could we revert this variable change?

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -294,14 +420,27 @@ func getArmVirtualNetworkGatewayConnectionProperties(d *schema.ResourceData) (*n
 		}
 	}
 
+	if v, ok := d.GetOk("enable_bgp"); ok {
```
tombuildsstuff: if this is a bool, it'll always have a value (since the default is `false`) - as such we may as well just assign this above?

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -294,14 +420,27 @@ func getArmVirtualNetworkGatewayConnectionProperties(d *schema.ResourceData) (*n
...
+		props.EnableBgp = utils.Bool(v.(bool))
+	}
+
+	if v, ok := d.GetOk("use_policy_based_traffic_selectors"); ok {
```
tombuildsstuff: (same here)

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -339,3 +478,77 @@ func resourceGroupAndVirtualNetworkGatewayConnectionFromId(virtualNetworkGateway
...
+
+		if v, ok := ipsec_policy["sa_datasize"].(int); ok {
+			sa_datasize := int32(v)
+			ipsecPolicy.SaDataSizeKilobytes = &sa_datasize
+		}
+
+		if v, ok := ipsec_policy["sa_lifetime"].(int); ok {
+			sa_lifetime := int32(v)
+			ipsecPolicy.SaLifeTimeSeconds = &sa_lifetime
+		}
```
tombuildsstuff: _minor_ could we update all of these variables to use camelCase rather than snake_case?

azurerm/resource_arm_virtual_network_gateway_connection_test.go
```diff
@@ -341,3 +360,89 @@ resource "azurerm_virtual_network_gateway_connection" "test_2" {
...
+  default = "%d"
+}
+
+resource "azurerm_resource_group" "test" {
+  name = "acctestRG-${var.random}"
+  location = "%s"
+}
+
+resource "azurerm_virtual_network" "test" {
+  name = "test-${var.random}"
```
tombuildsstuff: can we make this `acctestvn` to be picked up by the acceptance testing sweepers?

azurerm/resource_arm_virtual_network_gateway_connection_test.go
```diff
@@ -341,3 +360,89 @@ resource "azurerm_virtual_network_gateway_connection" "test_2" {
...
+
+resource "azurerm_subnet" "test" {
+  name = "GatewaySubnet"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  virtual_network_name = "${azurerm_virtual_network.test.name}"
+  address_prefix = "10.0.1.0/24"
+}
+
+resource "azurerm_public_ip" "test" {
+  name = "test-${var.random}"
```
tombuildsstuff: can we change `test` -> `acctest`?

azurerm/resource_arm_virtual_network_gateway_connection_test.go
```diff
@@ -341,3 +360,89 @@ resource "azurerm_virtual_network_gateway_connection" "test_2" {
...
+  location = "${azurerm_resource_group.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  public_ip_address_allocation = "Dynamic"
+}
+
+resource "azurerm_virtual_network_gateway" "test" {
+  name = "test-${var.random}"
```
tombuildsstuff: (same here) can we change `test` -> `acctest`?

azurerm/resource_arm_virtual_network_gateway_connection_test.go
```diff
@@ -341,3 +360,89 @@ resource "azurerm_virtual_network_gateway_connection" "test_2" {
...
+  ip_configuration {
+    name = "vnetGatewayConfig"
+    public_ip_address_id = "${azurerm_public_ip.test.id}"
+    private_ip_address_allocation = "Dynamic"
+    subnet_id = "${azurerm_subnet.test.id}"
+  }
+}
+
+resource "azurerm_local_network_gateway" "test" {
+  name = "test-${var.random}"
```
tombuildsstuff: (same here) can we change `test` -> `acctest`?

azurerm/resource_arm_virtual_network_gateway_connection_test.go
```diff
@@ -341,3 +360,89 @@ resource "azurerm_virtual_network_gateway_connection" "test_2" {
...
+  location = "${azurerm_resource_group.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+
+  gateway_address = "168.62.225.23"
+  address_space = ["10.1.1.0/24"]
+}
+
+resource "azurerm_virtual_network_gateway_connection" "test" {
+  name = "test-${var.random}"
```
tombuildsstuff: (same here) can we change `test` -> `acctest`?

website/docs/r/virtual_network_gateway_connection.html.markdown
```diff
@@ -248,8 +247,44 @@ The following arguments are supported:
...
+
+* `ipsec_integrity` - (Required) The IPSec integrity algorithm. Valid
+    options are `GCMAES128`, `GCMAES192`, `GCMAES256`, `MD5`, `SHA1`, or `SHA256`.
+    
+* `pfs_group` - (Required) The DH group used in IKE phase 2 for new child SA.
+    Valid options are `ECP256`, `ECP384`, `PFS1`, `PFS2`, `PFS2048`, `PFS24`,
+    or `None`.
+
+* `sa_datasize` - (Optional) The IPSec SA payload size in KB. Must be at least
+    `1024` KB. Defaults to `102400000` KB.
```
tombuildsstuff: given the value can stretch so high, would it be worth converting this value to/from MB?

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -294,14 +420,27 @@ func getArmVirtualNetworkGatewayConnectionProperties(d *schema.ResourceData) (*n
...
 	if v, ok := d.GetOk("shared_key"); ok {
-		sharedKey := v.(string)
-		props.SharedKey = &sharedKey
+		props.SharedKey = utils.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("ipsec_policy"); ok {
+		ipsec_policies := v.([]interface{})
+		props.IpsecPolicies = expandArmVirtualNetworkGatewayConnectionIpsecPolicies(ipsec_policies)
+		log.Printf("Ipsec policies: %+q", props.IpsecPolicies)
```
tombuildsstuff: can we remove this print statement?

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -339,3 +478,77 @@ func resourceGroupAndVirtualNetworkGatewayConnectionFromId(virtualNetworkGateway
...
+		ipsecPolicies = append(ipsecPolicies, *ipsecPolicy)
+	}
+
+	return &ipsecPolicies
+}
+
+func flattenArmVirtualNetworkGatewayConnectionIpsecPolicies(ipsecPolicies *[]network.IpsecPolicy) []interface{} {
+	ipsec_policies := make([]interface{}, 0, len(*ipsecPolicies))
+
+	for _, ipsecPolicy := range *ipsecPolicies {
```
tombuildsstuff: there's a potential crash here if `ipsecPolicies` is nil - could we make this:

```
ipsec_policies := make([]interface{}, 0)
if inputPolicies := ipsecPolicies; inputPolicies != nil {
  for _, ipsecPolicy := range *inputPolicies {
    # ..
  }
}
```

azurerm/resource_arm_virtual_network_gateway_connection.go
dominik-lekse: Yes, false is the API default for enable_bgp. The intention for this change was to increase the transparency of the provider towards the API, i.e. allow the API to decide for the default. Unfortunately, the API docs are not quite elaborate on the default values (https://docs.microsoft.com/en-us/rest/api/network-gateway/virtualnetworkgatewayconnections/createorupdate).

azurerm/resource_arm_virtual_network_gateway_connection.go
dominik-lekse: Same reasoning as above, 10 is the default value, but it should not be fixed in the provider.

azurerm/resource_arm_virtual_network_gateway_connection.go
dominik-lekse: Regarding the variable naming, I found it helpful to quickly identify the source of which the value originates. I.e. I chose CamelCase for values from the Azure Go SDK whereas snake_case for value from the Terraform schema.

However, to match the other resources, I will revert it.

azurerm/resource_arm_virtual_network_gateway_connection.go
dominik-lekse: In particular in a mapping function like flattenArmVirtualNetworkGatewayConnectionIpsecPolicies, the two casing makes sense to differentiate between Terraform schema and struct of Azure Go SDK.

azurerm/resource_arm_virtual_network_gateway_connection_test.go
dominik-lekse: Sure, this functionality was new to me.

website/docs/r/virtual_network_gateway_connection.html.markdown
dominik-lekse: I would recommend to keep KBytes as a unit here for two reasons: 1) In the Azure docs this values is consistently described in this unit (https://docs.microsoft.com/en-us/azure/vpn-gateway/vpn-gateway-vpn-faq#ipsecike) and 2) it is quite common among VPN vendors to configure this in Kbytes (e.g https://www.cisco.com/c/en/us/td/docs/security/security_management/cisco_security_manager/security_manager/4-1/user/guide/CSMUserGuide_wrapper/vpipsec.html#pgfId-959357)

azurerm/resource_arm_virtual_network_gateway_connection.go
dominik-lekse: Sure, this is a true left over 😄

azurerm/resource_arm_virtual_network_gateway_connection.go
dominik-lekse: Yes, I shouldn't rely on the caller here

azurerm/resource_arm_virtual_network_gateway_connection.go
dominik-lekse: The idea was to just not set the field in the API struct, if it is not defined in the Terraform schema instance. Is this overly safe?

azurerm/resource_arm_virtual_network_gateway_connection.go
dominik-lekse: It is an optional field in the API, is it then also safe to avoid this check?

azurerm/resource_arm_virtual_network_gateway_connection.go
dominik-lekse: This is currently done in ResourceData.Set

website/docs/r/virtual_network_gateway_connection.html.markdown
tombuildsstuff: makes sense 👍

azurerm/resource_arm_virtual_network_gateway_connection.go
tombuildsstuff: It's more this won't work as it reads unfortunately due to the way the defaults work in HCL (there's changes coming in this area in the future); given the field in the schema is a Boolean, there'll always be a default value of `false`, so the `ok` in this if statement will always be true.

We _could_ instead update this to use `GetOkExists` which returns the raw value instead of the defaulted value if nothings specified (e.g. an empty string in this case, iirc) - however I don't think there's any harm setting a default to `false` tbh - what do you think?

azurerm/resource_arm_virtual_network_gateway_connection.go
tombuildsstuff: Since `d.Set` handles setting pointer values (even if they're nil - to an empty value) - it should be fine :)

azurerm/resource_arm_virtual_network_gateway_connection.go
dominik-lekse: (same here)

stintel:
Tested this PR and it works well for me. Would be nice to have this in the next release. Without a custom IPsec policy, an Azure VPN GW uses security algorithms that are no longer considered secure (3DES, SHA1, MODP1024 - see https://wiki.strongswan.org/projects/strongswan/wiki/SecurityRecommendations)

tombuildsstuff:
Tests pass:

screen shot 2018-02-28 at 17 24 32

dominik-lekse:
@tombuildsstuff Thanks for merging 👍

stintel:
@tombuildsstuff thanks for merging
@dominik-lekse thanks for your contribution

tombuildsstuff:
:wave: hey folks!  

Just to let you know that support for this has just been released in v1.2.0 of the AzureRM Provider - full details of what's included are available here: https://github.com/terraform-providers/terraform-provider-azurerm/blob/v1.2.0/CHANGELOG.md#120-march-02-2018

Thanks!

PR #846, opened 2018-02-16
Retry to create storage container.

genevieve:
https://github.com/terraform-providers/terraform-provider-azurerm/issues/841

### DIFF
--- a/azurerm/resource_arm_storage_container.go
+++ b/azurerm/resource_arm_storage_container.go
@@ -4,10 +4,12 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"time"
 
 	"regexp"
 
 	"github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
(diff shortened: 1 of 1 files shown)


azurerm/resource_arm_storage_container.go
```diff
@@ -124,6 +125,18 @@ func resourceArmStorageContainerCreate(d *schema.ResourceData, meta interface{})
 	return resourceArmStorageContainerRead(d, meta)
 }
 
+func checkContainerIsCreated(reference *storage.Container) func() *resource.RetryError {
+	return func() *resource.RetryError {
+		createOptions := &storage.CreateContainerOptions{}
```
paultyng: Should we explicitly set timeout here so we can effectively use the timeout duration of the `Retry` call? I'm not sure on the default, maybe its ok?

genevieve: There’s sin inyl one other instance of retrying in the codebase that’s I found and the timeout was not configurable. We can add a create timeout to the scheme if we think it should be adjustable?

tombuildsstuff: we lean on the SDK's default timeout through the codebase, so I don't think we need to override this value? (2m should be enough time for Azure to become consistent)

tombuildsstuff:
Tests pass:

screen shot 2018-02-16 at 09 45 39

vchauhan1:
Still it's failing with the following error. azurerm_storage_container.be-sc: Error creating container "XXXXXXX" in storage account "XXXXXXXX": storage: service returned error: StatusCode=403, ErrorCode=AuthenticationFailed, ErrorMessage=Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.

dariustehrani:
I can confirm this is still failing.

PR #632, opened 2017-12-14
`azurerm_application_gateway` - handling components being nil

tombuildsstuff:
Refactors the Application Gateway code to handle things being nil (such as in older Application Gateway instances).

Need to double-check the upgrade path - but this fixes #618

PR #848, opened 2018-02-16
`azurerm_application_gateway` - removing crash points

### LINKED ISSUE #816: AppGateway issue after changing SSLPolicy
Hi there,

We recently faced the issue when Terraform starts crashing every time after we have updated SSLPolicy in Azure Application Gateway via Powershell with this commands:

Set-AzureRmApplicationGatewaySslPolicy -PolicyType Predefined -PolicyName AppGwSslPolicy20170401S -ApplicationGateway $gw
Set-AzureRmApplicationGateway -ApplicationGateway $gw

After that Terraform crashes every time with "unexpected EOF" messages.

We tried to remove the resource from the state and got the same error again on import.
I attached Gist with full error log.

Can anyone suggest any workaround?

### Terraform Version
v 0.11.3

### Affected Resource(s)

azurerm_application_gateway

### Debug Output
https://gist.github.com/chugunkov-devpro/49c77edd65dc73ef26ac943980bbbf82

tombuildsstuff:
The Azure API returns the latest version of a resource at the last time it was changed, which for an existing resource could be from 2 years ago not conforming to the latest schema; such that some fields which are Required in the API are missing from the API response.

As such - this PR goes through and removes all potential crash points that I could see in the App Gateway resource's flatten functions - which should fix #816 and a few other issues

Unfortunately this is painful to test, since each test takes 30m - so this is a WIP until I can confirm everything works as expected. This is kind of a rework of #632 but without all the additional tests (which we can't add due to an API bug)

Fixes #816

### DIFF
--- a/azurerm/resource_arm_application_gateway.go
+++ b/azurerm/resource_arm_application_gateway.go
@@ -1385,298 +1385,376 @@ func flattenApplicationGatewayIPConfigurations(ipConfigs *[]network.ApplicationG
 }
 
 func flattenApplicationGatewayFrontendPorts(portConfigs *[]network.ApplicationGatewayFrontendPort) []interface{} {
-	result := make([]interface{}, 0, len(*portConfigs))
+	result := make([]interface{}, 0)
 
-	for _, config := range *portConfigs {
-		port := map[string]interface{}{
-			"id":   *config.ID,
-			"name": *config.Name,
-			"port": int(*config.ApplicationGatewayFrontendPortPropertiesFormat.Port),
+	if configs := portConfigs; configs != nil {
+		for _, config := range *configs {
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
Ignoring [the known Azure API deletion bug](https://github.com/Azure/azure-rest-api-specs/issues/2187), which happens a large percentage of the time (which we can't work around 😞):

> * azurerm_subnet.test: Error waiting for completion for Subnet "subnet-3251487714471370920" (VN "acctest-vnet-3251487714471370920" / Resource Group "acctestrg-3251487714471370920"): Long running operation terminated with status 'Failed': Code="InternalServerError" Message="An error occurred."

The tests pass:

screen shot 2018-02-16 at 15 00 38

PR #849, opened 2018-02-16
d/App Service Plan: handling a 404 not being returned as an error

### LINKED ISSUE #766: Crash while handling Not Found response
### Terraform Version

Terraform v0.11.2
+ provider.azurerm v1.1.0

### Affected Resource(s)

It seams that all resources are affected, as all could receive Not Found response.

```
data "azurerm_app_service_plan" "ServicePlanb" {
  name = "ServicePlan"
  resource_group_name = "A"
  # while ServicePlan actually exists under resource group B
}
```

### Panic Output

https://gist.github.com/irnc/0b84ce980d6af2c5861d7499c38b12b8 contains only extract with panic stack trace.

### Expected Behavior

Not Found error should be handled.

### Actual Behavior

Terraform crashes on a code branch, which was not supposed to be run, i.e. `err != nil` branch was expected to be run instead of https://github.com/terraform-providers/terraform-provider-azurerm/blob/v1.1.0/azurerm/data_source_app_service_plan.go#L97

### Steps to Reproduce

- `terraform plan`

tombuildsstuff:
Fixes #766

### DIFF
--- a/azurerm/data_source_app_service_plan.go
+++ b/azurerm/data_source_app_service_plan.go
@@ -86,20 +86,23 @@ func dataSourceAppServicePlanRead(d *schema.ResourceData, meta interface{}) erro
 	ctx := meta.(*ArmClient).StopContext
 	resp, err := client.Get(ctx, resourceGroup, name)
 	if err != nil {
-		if utils.ResponseWasNotFound(resp.Response) {
-			d.SetId("")
-			return nil
-		}
-
 		return fmt.Errorf("Error making Read request on App Service Plan %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	if utils.ResponseWasNotFound(resp.Response) {
+		d.SetId("")
+		return nil
 	}
 
 	d.SetId(*resp.ID)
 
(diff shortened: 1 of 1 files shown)


azurerm/data_source_app_service_plan.go
```diff
@@ -86,21 +86,24 @@ func dataSourceAppServicePlanRead(d *schema.ResourceData, meta interface{}) erro
...
 	if err != nil {
-		if utils.ResponseWasNotFound(resp.Response) {
-			d.SetId("")
-			return nil
-		}
-
 		return fmt.Errorf("Error making Read request on App Service Plan %q (Resource Group %q): %+v", name, resourceGroup, err)
 	}
 
+	if utils.ResponseWasNotFound(resp.Response) {
```
mbfrahry: Does this change need to be rolled out to all of the resources? Or is this just app service plan specific?

tombuildsstuff: it appears to be resource dependent, unfortunately - so we'll need to go through the lot (although, that's outside the scope of this PR)

PR #884, opened 2018-02-23
Add support for vmss application gateway backend pools

### LINKED ISSUE #857: Feature Request: Application Gateway Backend Pool in VMSS
### Terraform Version
Terraform v0.11.3
+ provider.azurerm v1.1.1

### Affected Resource(s)
 azurerm_virtual_machine_scale_set (When used with azurerm_application_gateway)

I'm deploying **azurerm_virtual_machine_scale_set** with **azurerm_application_gateway**, but I can't see a way to associate the VMSS nodes, with the backend pool.

It can be done successfully using **azurerm_lb**...

azurerm_virtual_machine_scale_set.network_profile.ip_configuration.load_balancer_backend_address_pool_ids = ["${azurerm_lb_backend_address_pool.bpepool.id}"]

But if I swap out the backend pool id for one used by **azurerm_application_gateway** I get the following error:

ag-beap is referencing resource of a wrong type. The Id is expected to reference resources of type loadBalancers/backendAddressPools. 

Should there be another attribute under azurerm_virtual_machine_scale_set.network_profile.ip_configuration - Like application_gateway_backend_pools?

...or maybe there is another way to do this that I'm missing?

agolomoodysaada:
Closes #857

### DIFF
--- a/azurerm/resource_arm_virtual_machine_scale_set.go
+++ b/azurerm/resource_arm_virtual_machine_scale_set.go
@@ -274,6 +274,13 @@ func resourceArmVirtualMachineScaleSet() *schema.Resource {
 										Required: true,
 									},
 
+									"application_gateway_backend_address_pool_ids": {
+										Type:     schema.TypeSet,
+										Optional: true,
+										Elem:     &schema.Schema{Type: schema.TypeString},
+										Set:      schema.HashString,
+									},
+
 									"load_balancer_backend_address_pool_ids": {
 										Type:     schema.TypeSet,
 										Optional: true,
@@ -933,6 +940,14 @@ func flattenAzureRmVirtualMachineScaleSetNetworkProfile(profile *compute.Virtual
 				if ipConfig.VirtualMachineScaleSetIPConfigurationProperties.Subnet != nil {
 					config["subnet_id"] = *properties.Subnet.ID
 				}
+
+				addressPools := make([]interface{}, 0)
+				if properties.ApplicationGatewayBackendAddressPools != nil {
+					for _, pool := range *properties.ApplicationGatewayBackendAddressPools {
+						addressPools = append(addressPools, *pool.ID)
+					}
+				}
+				config["application_gateway_backend_address_pool_ids"] = schema.NewSet(schema.HashString, addressPools)
 
 				if properties.LoadBalancerBackendAddressPools != nil {
 					addressPools := make([]interface{}, 0, len(*properties.LoadBalancerBackendAddressPools))
--- a/azurerm/resource_arm_virtual_machine_scale_set_test.go
+++ b/azurerm/resource_arm_virtual_machine_scale_set_test.go
@@ -361,6 +361,25 @@ func TestAccAzureRMVirtualMachineScaleSet_customImage(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMVirtualMachineScaleSet_applicationGateway(t *testing.T) {
+	ri := acctest.RandInt()
+	config := testAccAzureRMVirtualMachineScaleSetApplicationGatewayTemplate(ri, testLocation())
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMVirtualMachineScaleSetDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVirtualMachineScaleSetExists("azurerm_virtual_machine_scale_set.test"),
+					testCheckAzureRMVirtualMachineScaleSetHasApplicationGateway("azurerm_virtual_machine_scale_set.test"),
+				),
+			},
+		},
+	})
+}
+
 func TestAccAzureRMVirtualMachineScaleSet_loadBalancer(t *testing.T) {
 	ri := acctest.RandInt()
 	config := testAccAzureRMVirtualMachineScaleSetLoadBalancerTemplate(ri, testLocation())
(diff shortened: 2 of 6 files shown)


azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -934,6 +941,14 @@ func flattenAzureRmVirtualMachineScaleSetNetworkProfile(profile *compute.Virtual
 					config["subnet_id"] = *properties.Subnet.ID
 				}
 
+				if properties.ApplicationGatewayBackendAddressPools != nil {
+					addressPools := make([]interface{}, 0, len(*properties.ApplicationGatewayBackendAddressPools))
+					for _, pool := range *properties.ApplicationGatewayBackendAddressPools {
+						addressPools = append(addressPools, *pool.ID)
+					}
+					config["application_gateway_backend_address_pool_ids"] = schema.NewSet(schema.HashString, addressPools)
```
tombuildsstuff: so we probably want to set a Set regardless here (e.g. to an empty list if it's not specified) - as such could we make this:

```
addressPools := make([]interface{}, 0)
if properties.ApplicationGatewayBackendAddressPools != nil {
  for _, pool := range *properties.ApplicationGatewayBackendAddressPools {
    addressPools = append(addressPools, *pool.ID)
  }
}
config["application_gateway_backend_address_pool_ids"] = schema.NewSet(schema.HashString, addressPools)
```

website/docs/r/virtual_machine_scale_set.html.markdown
```diff
@@ -129,10 +129,11 @@ resource "azurerm_virtual_machine_scale_set" "test" {
     primary = true
 
     ip_configuration {
-      name                                   = "TestIPConfiguration"
-      subnet_id                              = "${azurerm_subnet.test.id}"
-      load_balancer_backend_address_pool_ids = ["${azurerm_lb_backend_address_pool.bpepool.id}"]
-      load_balancer_inbound_nat_rules_ids    = ["${element(azurerm_lb_nat_pool.lbnatpool.*.id, count.index)}"]
+      name                                         = "TestIPConfiguration"
+      subnet_id                                    = "${azurerm_subnet.test.id}"
+      application_gateway_backend_address_pool_ids = ["${azurerm_application_gateway.mygateway.backend_address_pool.id}"]
```
tombuildsstuff: given this isn't present in the example (and I believe it conflicts with the LB fields?), it might be prudent to comment this out as an example?

agolomoodysaada: I see your point. While it's not common, they're actually not mutually exclusive. You could techinically wire your VMSS with both a load balancer (for NAT rules or internal LBs) and an application gateway (for public HTTP traffic).

agolomoodysaada: I commented it out anyway.

tombuildsstuff: thanks for confirming that - I guess we can leave it commented out for the moment either way :)

azurerm/resource_arm_virtual_machine_scale_set_test.go
```diff
@@ -374,6 +374,7 @@ func TestAccAzureRMVirtualMachineScaleSet_loadBalancer(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testCheckAzureRMVirtualMachineScaleSetExists("azurerm_virtual_machine_scale_set.test"),
 					testCheckAzureRMVirtualMachineScaleSetHasLoadbalancer("azurerm_virtual_machine_scale_set.test"),
+					testCheckAzureRMVirtualMachineScaleSetHasApplicationGateway("azurerm_virtual_machine_scale_set.test"),
```
tombuildsstuff: So at the moment this test will fail since there's no Application Gateway's defined in [the config used in this test](https://github.com/terraform-providers/terraform-provider-azurerm/pull/884/files#diff-62e6c44ba2b5a3617bb9c78744b7a84bR366). Could we create a separate config (as linked above) + test for creating a VMSS with an Application Gateway configured, and revert this test as-is? This'll mean we have test cases for both VMSS with an LB and an App Gateway - both of which are valid

agolomoodysaada: Makes sense. Thank you for the feedback. 
For some reason the tests were passing locally for me. Do I have to configure something locally to make the test fail?

agolomoodysaada: Oh I see. I have to run `make testacc`. Is there a way to specify an individual acceptance test to run?

agolomoodysaada:
@tombuildsstuff I made the changes you requested and tried running the acceptance test. Unfortunately, after +20mins (it takes a long time for appgateway to provision), I get the following error

```
=== RUN   TestAccAzureRMVirtualMachineScaleSet_applicationGateway
--- FAIL: TestAccAzureRMVirtualMachineScaleSet_applicationGateway (1279.82s)
	testing.go:459: Step 0 error: Error applying: 1 error(s) occurred:

		* azurerm_virtual_machine_scale_set.test: Resource 'azurerm_application_gateway.test' does not have attribute 'backend_address_pool.*.id' for variable 'azurerm_application_gateway.test.backend_address_pool.*.id'
```

What am I doing wrong?

azurerm/resource_arm_virtual_machine_scale_set_test.go
tombuildsstuff: there is:

```
TF_ACC=1 go test ./azurerm -v -timeout 120m -run=TestPrefixYouWantToRun
```

where `TestPrefixYouWantToRun` is the prefix of the test you want to run (which can be a full test name) - hope that helps :)

abn:
@agolomoodysaada I have opened #1027 to replace this one and re-used your commit (most of it). There were couple of test failures, one was due to the issue you have mentioned above; I have worked around it (which seems to be the only way) by referencing the element directly. The second failure was because the backend pools were not loaded correctly into state and hence test check was failing.

Both issues were resolved and included in #1027

tombuildsstuff:
hey @agolomoodysaada 

Apologies for the delayed response here - I'd missed that I'd not replied to your comment.

I hope you don't mind, but I'm going to close this PR in favour of #1027 which includes your original commits/attributions that's been recently opened by @abn.

Thanks!

agolomoodysaada:
Thank you @abn and @tombuildsstuff ... Glad to see this merged

Issue #745, opened 2018-01-23
Feature Request: please add the ability to specify site_config in azurerm_function_app


dcherniv:
azurerm_app_service supports site_config options. Since functions is essentially an app_service, please add the ability to specify site_config for it.
Without this, there's no way to control things like 32/64 bit workers and others.
![image](https://user-images.githubusercontent.com/19437418/35285727-249adeb0-002c-11e8-9d29-5a415fe0c777.png)

tombuildsstuff:
hey @dcherniv 

Thanks for opening this issue :)

So that we can look into this (and that there's differences between the App Service site config and a Function App site config) - would it be possible to know which fields are you looking to use within the `site_config` block?

Thanks!

dcherniv:
@tombuildsstuff 
From: https://www.terraform.io/docs/providers/azurerm/r/app_service.html
always_on - (Optional) Should the app be loaded at all times? Defaults to false.
use_32_bit_worker_process - (Optional) Should the App Service run in 32 bit mode, rather than 64 bit mode?

use_32_bit_worker_process is probably more important because for me when function_app was provisioned with the following, it defaulted to 32 bit workers for some odd reason.

```
resource "azurerm_function_app" "default" {
  name                      = "${var.name}"
  location                  = "${var.region}"
  resource_group_name       = "${var.rg_name}"
  app_service_plan_id       = "${var.as_plan_id}"
  storage_connection_string = "${var.storage_connection}"
  app_settings              = "${var.app_settings}"
  tags {
     Environment    = "${var.tag_env}"
     Name           = "${var.tag_name}"
     Owner          = "${var.tag_owner}"
     Contact        = "${var.tag_contact}"
     CostProject    = "${var.tag_cost}"
     "Resource Group" = "${var.rg_name}"
  }
}
```

gunzip:
I'd add [client_affinity_enabled as well](https://github.com/terraform-providers/terraform-provider-azurerm/blob/9257fc3233e5650620cc1562e9ea81d1bbf0784b/azurerm/resource_arm_app_service.go#L167):
https://www.terraform.io/docs/providers/azurerm/r/app_service.html#client_affinity_enabled

This is very important in the context of Functions which are served behind an API management resource:
when you have client_affinity_enabled the Function backend sends a cookie with a domain which is not the same as the one of the API gateway.

gunzip:
always_on is now supported thanks of
[...]

PR #886, opened 2018-02-23
Issue #745, adds more options to azurerm_function_app

cloudify:
@tombuildsstuff how do you suggest I test this patch? should I create new tests or extend the current ones?

### DIFF
--- a/azurerm/resource_arm_function_app.go
+++ b/azurerm/resource_arm_function_app.go
@@ -122,6 +122,16 @@ func resourceArmFunctionApp() *schema.Resource {
 				Computed: true,
 			},
 
+			"client_affinity_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+
+				// TODO: (tombuildsstuff) support Update once the API is fixed:
+				// https://github.com/Azure/azure-rest-api-specs/issues/1697
+				ForceNew: true,
+			},
+
 			"site_config": {
 				Type:     schema.TypeList,
 				Optional: true,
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_function_app.go
```diff
@@ -134,6 +144,16 @@ func resourceArmFunctionApp() *schema.Resource {
 							Optional: true,
 							Default:  false,
 						},
+						"use_32_bit_worker_process": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
```
tombuildsstuff: Based on [this comment](https://github.com/terraform-providers/terraform-provider-azurerm/issues/745#issuecomment-360933129) I believe this needs to be defaulted to `true` for Function Apps?

website/docs/r/function_app.html.markdown
```diff
@@ -87,7 +89,12 @@ The following arguments are supported:
 
 `site_config` supports the following:
 
-* `always_on` - (Optional) Should the app be loaded at all times? Defaults to `false`.
+* `always_on` - (Optional) Should the Function App be loaded at all times? Defaults to `false`.
+* `use_32_bit_worker_process` - (Optional) Should the Function App run in 32 bit mode, rather than 64 bit mode?
```
tombuildsstuff: We should add "Defaults to `value`" (which should be true, based on above)

azurerm/resource_arm_function_app.go
```diff
@@ -410,6 +441,14 @@ func flattenFunctionAppSiteConfig(input *web.SiteConfig) []interface{} {
 		result["always_on"] = *input.AlwaysOn
 	}
 
+	if input.Use32BitWorkerProcess != nil {
+		result["use_32_bit_worker_process"] = *input.Use32BitWorkerProcess
+	}
```
tombuildsstuff: could we add an acceptance test with the 64 bit worker process enabled? here's [an example of how we're doing that](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_function_app_test.go#L14)

tombuildsstuff: so that we can get this merged, I've pushed an acceptance test for this - I hope you don't mind!

cloudify: not at all! thanks for pushing this, I got caught in higher priority tasks and couldn't focus on this now...

tombuildsstuff:
Tests pass:

screen shot 2018-03-06 at 09 25 55

PR #887, opened 2018-02-23
New Resource: `azurerm_iothub`

shelleybess:
Add IotHub resource to provider.


shelleybess:
Created using SDK-for-go version v11.0.0-beta. Needs update.

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,261 @@
...
+							}, true),
+						},
+
+						"tier": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(devices.Free),
+								string(devices.Standard),
+							}, true),
```
tombuildsstuff: since we're allowing this to be case insensitive, can we add:

```
```

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,261 @@
...
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(devices.F1),
+								string(devices.S1),
+								string(devices.S2),
+								string(devices.S3),
+							}, true),
```
tombuildsstuff: since we're allowing this to be case insensitive, can we add:

```
```

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,261 @@
...
+					},
+				},
+			},
+
+			"type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"etag": {
```
tombuildsstuff: I think it might be better not to expose this field (since, it's handled internally within Azure?) and could cause user confusion - could we remove it?

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func expandAzureRmIotHubSku(d *schema.ResourceData) devices.IotHubSkuInfo {
+	skuList := d.Get("sku").([]interface{})
+	skuMap := skuList[0].(map[string]interface{})
+	cap := int64(skuMap["capacity"].(int))
+
+	return devices.IotHubSkuInfo{
+		Name:     devices.IotHubSku(skuMap["name"].(string)),
+		Tier:     devices.IotHubSkuTier(skuMap["tier"].(string)),
```
tombuildsstuff: _minor_ could we pull `name` and `tier` out above and cast them to a string there, since it'll make a crash log easier to understand, e.g.:

```
name := skuMap["name"].(string)
tier  := skuMap["tier"].(string)
# ...

return devices.IotHubSkuInfo{
  Name:     devices.IotHubSku(name),
  Tier:     devices.IotHubSkuTier(tier),
}

```

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	iothubName := id.Path["IotHubs"]
+	desc, err := iothubClient.Get(ctx, id.ResourceGroup, iothubName)
+
+	if err != nil {
+		return err
+	}
+
+	properties := desc.Properties
```
tombuildsstuff: if the Swagger is invalid / the API response varies (such as when reading an old resource) - this isn't guaranteed to be returned, unfortunately. as such could we check for it having a value:

```
if props := desc.Properties; props != nil {
  # ..
}
```

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,261 @@
...
+		keyMap := make(map[string]interface{})
+		keyMap["key_name"] = *key.KeyName
+		keyMap["primary_key"] = *key.PrimaryKey
+		keyMap["secondary_key"] = *key.SecondaryKey
+		keyMap["permissions"] = string(key.Rights)
+		keys = append(keys, keyMap)
+	}
+
+	if err != nil {
+		return err
```
tombuildsstuff: can we detail this error, e.g.

```
return fmt.Errorf("Error flattening `shared_access_policy`: %+v", err)

File: azurerm/resource_arm_iothub_resource_provider.go [hunk 7]
```diff
@@ -0,0 +1,261 @@
...
+
+	keysResp, err := iothubClient.ListKeys(ctx, id.ResourceGroup, iothubName)
+	keyList := keysResp.Response()
+
+	var keys []map[string]interface{}
+	for _, key := range *keyList.Value {
+		keyMap := make(map[string]interface{})
+		keyMap["key_name"] = *key.KeyName
+		keyMap["primary_key"] = *key.PrimaryKey
+		keyMap["secondary_key"] = *key.SecondaryKey
```
tombuildsstuff: can we add if statements around these three, since there's a potential crash if they're not returned

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	if err != nil {
+		return err
+	}
+
+	d.Set("shared_access_policy", keys)
+	d.Set("hostname", *properties.HostName)
+	d.Set("etag", *desc.Etag)
+	d.Set("type", *desc.Type)
```
tombuildsstuff: since these are simple data types `d.Set` should take care of dereferencing these for us - so we can remove the `*`'s here :)

azurerm/provider.go
```diff
@@ -291,6 +292,7 @@ func determineAzureResourceProvidersToRegister(providerList []resources.Provider
 		"Microsoft.ServiceBus":          {},
 		"Microsoft.Sql":                 {},
 		"Microsoft.Storage":             {},
+		"Microsoft.Devices":             {},
```
tombuildsstuff: _really minor_ could we sort this alphabetically?

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,261 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	armClient := meta.(*ArmClient)
+	iothubClient := armClient.iothubResourceClient
+	ctx := armClient.StopContext
+
+	future, err := iothubClient.Delete(ctx, id.ResourceGroup, id.Path["IotHubs"])
```
tombuildsstuff: can we make a variable `name` here, rather than pulling it out in multiple places?

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return err
+	}
+
+	status := future.Status()
+	if status == "unkown" {
+		return fmt.Errorf("Error Waiting for the deletion of IoTHub %q (Resource Group %q): %+v", id.Path["IotHubs"], id.ResourceGroup, err)
+	}
```
tombuildsstuff: we should be able to make this:

```
err = future.WaitForCompletion(ctx, client.Client)
if err != nil {
  if response.WasNotFound(future.Response()) {
    return nil
  }

  return fmt.Errorf("Error waiting for the deletion...")
}

return nil
```

azurerm/resource_arm_iothub_resource_provider_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+			return fmt.Errorf("Bad: Get on iothubResourceClient: %+v", err)
+		}
+
+		return nil
+
+	}
+}
+
+func testAccAzureRMIotHub_basicStandard(name, rgname string) string {
```
tombuildsstuff: could we update this to match the other tests by passing in `rInt` and `location`? (we use `location` since we run tests against different Azure regions)

vendor/vendor.json
```diff
@@ -313,11 +325,11 @@
...
-			"checksumSHA1": "HfqZyKllcHQDvTwgCaYL1jUPmW0=",
+			"checksumSHA1": "5UH4IFIB/98iowPCzzVs4M4MXiQ=",
 			"comment": "v9.7.0",
 			"path": "github.com/Azure/go-autorest/autorest/validation",
-			"revision": "8c58b4788dedd95779efe0ac2055bb6a1b9b8e01",
-			"revisionTime": "2018-01-06T16:29:27Z",
+			"revision": "1e3943ee722538420f21945f4bc820347abe433d",
+			"revisionTime": "2018-02-15T17:42:55Z",
 			"version": "v9.7.0",
 			"versionExact": "v9.7.0"
```
tombuildsstuff: from what I can see this is from `Azure/go-autorest` v10 - could we downgrade this for the moment, since I don't think it's needed?

website/docs/r/iothub.html.markdown
```diff
@@ -0,0 +1,66 @@
+---
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_iothub"
+sidebar_current: "docs-azurerm-resource-iothub"
+description: |-
+  Creates a new IotHub resource 
+---
+
+# azurerm\_iothub
```
tombuildsstuff: _minor_ we can remove the backslash here, it's not needed

website/docs/r/iothub.html.markdown
```diff
@@ -0,0 +1,66 @@
...
+resource "azurerm_resource_group" "test" {
+  name     = "resourceGroup1"
+  location = "West US"
+}
+
+
+resource "azurerm_iothub" "test" {
+	name                             = "test"
+	resource_group_name              = "${azurerm_resource_group.test.name}"
+	location                         = "westus"
```
tombuildsstuff: _minor_ could we make this reference the location field on the resource group, e.g.

```
location = "${azurerm_resource_group.test.location}"
```

website/docs/r/iothub.html.markdown
```diff
@@ -0,0 +1,66 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) Specifies the name of the IotHub resource. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the resource group under which the IotHub resource has to be created. Changing this forces a new resource to be created.
+
+* `location` - (Required) Specifies the supported Azure location where the resource has to be createc. Changing this forces a new resource to be created.
+
+* `sku` - (Required) Specifies the capacity and tier of the IotHub.
```
tombuildsstuff: could we update the documentation to match the other resources here? e.g.

```
* `sku` - (Required) A `sku` block as defined below

---

A `sku` block contains:

* `name` - (Required) Defines the name of the SKU used for this IoTHub. Possible values are `F1`, `S1`, `S2` and `S3`.
* `tier` - (Required) Defines the tier of the SKU used for this IoTHub. Possible values are `Free` and `Standard`.
* `capacity` - (Required) Defines the capacity of the SKU used for this IoTHub.

```

website/docs/r/iothub.html.markdown
```diff
@@ -0,0 +1,66 @@
...
+
+The following attributes are exported:
+
+* `id` - The IotHub ID.
+
+* `etag` - The etag of IotHub Resource.
+
+* `hostname` - The hostname of the IotHub Resource.
+
+* `shared_access_policy` - The list of access policy which contain access keys and permissions.
```
tombuildsstuff: can we define this, e.g.

```
* `shared_access_policy` - A list of `shared_access_policy` blocks which are defined below.

---

`shared_access_policy` contains the following:

* `key_name` - Lorem Dolor Ipsum
* `primary_key ` - Lorem Dolor Ipsum
* `secondary_key ` - Lorem Dolor Ipsum
* `permissions ` - Lorem Dolor Ipsum
```

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,261 @@
...
+
+func resourceArmIotHub() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceArmIotHubCreateAndUpdate,
+		Read:   resourceArmIotHubRead,
+		Update: resourceArmIotHubCreateAndUpdate,
+		Delete: resourceArmIotHubDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
```
tombuildsstuff: given we're supporting Import here - could we add a test for it? [Here's an example of how we're doing this for Subnets](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/import_arm_subnet_test.go#L10)

website/docs/r/iothub.html.markdown
```diff
@@ -0,0 +1,66 @@
...
+
+
```
tombuildsstuff: can we add a section here with the information needed for Import, e.g.

"
## Import

IoTHubs can be imported using the `resource id`, e.g.

```shell
terraform import azurerm_iothub.hub1 /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/mygroup1/providers/Microsoft.Devices/IotHubs/hub1
```
"

vendor/vendor.json
shelleybess: I'm encountering an issue here. v10 includes the feature `Expose the polling URL for futures.` which is used by the sdk-for-go IoTHub models.go.

azurerm/resource_arm_iothub_resource_provider.go
shelleybess: This is added :) I'll add more import and resource tests given the time.

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,264 @@
...
+		return err
+	}
+
+	err = future.WaitForCompletion(ctx, iothubClient.Client)
+	if err != nil {
+		if response.WasNotFound(future.Response()) {
+			return nil
+		}
+
+		return fmt.Errorf("Error waiting for the deletion of IoTHub %q", iotHubName)
```
shelleybess: There is a timeout issue here. IoTHub is deleted in Azure, but terraform times out with an error destroying resource.

tombuildsstuff: so this timeout is configured in the client initialisation above: https://github.com/terraform-providers/terraform-provider-azurerm/pull/887/files/3e927a6d96d0c91b3dc132d9b9e83f4b1fc96650#diff-848235249455ee48aaf3ea5d431960aeR597

we can actually swap that out for this function:

https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/config.go#L435
[which sets a 60m polling timeout](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/config.go#L190), which should solve this issue. There's a separate task to move everything over to that at some point, but any changes to `azure/config.go` cause merge conflicts, so we're trying to do it piecemeal rather than bit by bit

Hope that helps :)

azurerm/import_arm_iothub_test.go
```diff
@@ -0,0 +1,31 @@
...
+func TestAccAzureRMIotHub_importBasic(t *testing.T) {
+	resourceName := "azurerm_iothub.test"
+
+	ri := acctest.RandInt()
+	config := testAccAzureRMIotHub_basicStandard(ri, testLocation())
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMSubnetDestroy,
```
tombuildsstuff: can we update this to be `testCheckAzureRMIotHubDestroy`?

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,264 @@
...
+
+	res, err := iothubClient.CheckNameAvailability(ctx, devices.OperationInputs{
+		Name: &name,
+	})
+
+	if err != nil {
+		return err
+	}
+
+	if !*res.NameAvailable {
```
tombuildsstuff: will this fail when updating a resource?

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,264 @@
...
+		Subscriptionid: &subscriptionID,
+		Sku:            &skuInfo,
+	}
+
+	if tagsI, ok := d.GetOk("tags"); ok {
+		tags := tagsI.(map[string]interface{})
+		desc.Tags = *expandTags(tags)
+	}
+
+	future, err := iothubClient.CreateOrUpdate(ctx, rg, name, desc, updateIoT)
```
tombuildsstuff: so we should be able to submit an empty string rather than retrieving the eTag here (since Terraform's plan will show the user what will be done)

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,264 @@
...
+		keyMap["secondary_key"] = *key.SecondaryKey
+		keyMap["permissions"] = string(key.Rights)
+		keys = append(keys, keyMap)
+	}
+
+	if err != nil {
+		return fmt.Errorf("Error flattening `shared_access_policy` in IoTHub %q: %+v", iothubName, err)
+	}
+
+	d.Set("shared_access_policy", keys)
```
tombuildsstuff: can we sort the ordering on these four lines, I think this should be:

```
if err := d.Set("shared_access_policy", keys); err != nil {
  return fmt.Errorf("...")
}
```

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,264 @@
...
+	d.Set("hostname", properties.HostName)
+	d.Set("type", desc.Type)
+	flattenAndSetTags(d, &desc.Tags)
```
tombuildsstuff: can we also set the `sku` block? This should be returned as part of the properties object

azurerm/resource_arm_iothub_resource_provider_test.go
```diff
@@ -0,0 +1,109 @@
...
+			},
+		},
+	})
+
+}
+
+func testCheckAzureRMIotHubDestroy(s *terraform.State) error {
+	ctx := testAccProvider.Meta().(*ArmClient).StopContext
+
+	conn := testAccProvider.Meta().(*ArmClient).expressRouteCircuitClient
```
tombuildsstuff: I think this wants to be using `iothubResourceClient`?

azurerm/resource_arm_iothub_resource_provider_test.go
```diff
@@ -0,0 +1,109 @@
...
+			name = "S1"
+			tier = "Standard"
+			capacity = "1"
+		}
+
+		tags {
+			"purpose" = "testing"
+		}
+}
+`, rInt, location, rInt)
```
tombuildsstuff: _really minor_ could we sort the spacing out here?

website/azurerm.erb
```diff
@@ -359,6 +359,17 @@
                 </ul>
             </li>
 
+            <li<%= sidebar_current("docs-azurerm-resource-iothub") %>>
+              <a href="#">IotHub Resources</a>
```
tombuildsstuff: 🤔 do you think it'd make sense for this to live under the `Messaging Resources` block?

website/docs/r/iothub.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+* `capacity` - (Required) The number of provisioned IoT Hub units. 
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The IotHub ID.
+
+* `etag` - The etag of IotHub Resource.
```
tombuildsstuff: _minor_ could we remove this since it's now no longer being exposed

azurerm/resource_arm_iothub_resource_provider.go
shelleybess: It updates the resource correctly. I want to add an accTest for this as well.

website/azurerm.erb
shelleybess: I agree. It shouldn't have its own block.

azurerm/resource_arm_iothub_resource_provider.go
shelleybess: This is very helpful, thank you!

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	if tagsI, ok := d.GetOk("tags"); ok {
+		tags := tagsI.(map[string]interface{})
+		desc.Tags = *expandTags(tags)
+	}
+
+	future, err := iothubClient.CreateOrUpdate(ctx, rg, name, desc, "")
+	if err != nil {
+		return err
```
tombuildsstuff: can we wrap this error to be more descriptive with what went wrong:

```

return fmt.Errorf("Error creating IoTHub %q (Resource Group %q): %+v", err)
```

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,271 @@
...
+			keyMap["primary_key"] = *primaryKey
+		}
+
+		if secondaryKey := key.SecondaryKey; secondaryKey != nil {
+			keyMap["secondary_key"] = *secondaryKey
+		}
+
+		keyMap["permissions"] = string(key.Rights)
+		keys = append(keys, keyMap)
+	}
```
tombuildsstuff: can we pull this out into a separate function `flattenAzureIotSharedAccessPolicy`?

azurerm/resource_arm_iothub_resource_provider.go
```diff
@@ -0,0 +1,271 @@
...
+	if err := d.Set("shared_access_policy", keys); err != nil {
+		return fmt.Errorf("Error flattening `shared_access_policy` in IoTHub %q: %+v", iothubName, err)
+	}
+
+	if properties := desc.Properties; properties != nil {
+		d.Set("hostname", properties.HostName)
+	}
+
+	d.Set("type", desc.Type)
+	flattenAndSetTags(d, &desc.Tags)
```
tombuildsstuff: so the Import test (`TestAccAzureRMIotHub_importBasic `) currently fails due to the following fields not being set - can we set them:

- `location`
- `name`
- `resource_group_name`
- `sku`
  - `capacity`
  - `name`
  - `tier`

on that note - could we pull this out into a new `flatten` function? I believe this should be what's needed to flatten the SKU block:

```
func flattenArmIoTHubSku(input *devices.IotHubSkuInfo) []interface {
  sku := make(map[string]interface{}, 0)

  if capacity := input.Capacity; capacity != nil {
    sku["capacity"] = *capacity
  }

  if name := input.Name; name != nil {
    sku["name"] = *name
  }

  if tier := input.Tier; tier != nil {
    sku["tier"] = *tier
  }

  return []interface{}{result}
}
```

azurerm/resource_arm_iothub_resource_provider_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAzureRMIotHub_basicStandard(t *testing.T) {
```
tombuildsstuff: _minor_ can we split this into two tests - one `_basic` without the SKU block, and one `_complete` with it?

azurerm/config.go
```diff
@@ -585,6 +590,15 @@ func (c *ArmClient) registerDatabases(endpoint, subscriptionId string, auth auto
 	c.sqlServersClient = sqlSrvClient
 }
 
+func (c *ArmClient) registerDeviceClients(endpoint, subscriptionId string, auth autorest.Authorizer, sender autorest.Sender) {
+	ihrc := devices.NewIotHubResourceClientWithBaseURI(endpoint, subscriptionId)
+	setUserAgent(&ihrc.Client)
+	ihrc.Authorizer = auth
+	ihrc.Sender = sender
+	ihrc.SkipResourceProviderRegistration = c.skipProviderRegistration
```
tombuildsstuff: rather than these 4 lines:

```
setUserAgent(&ihrc.Client)
ihrc.Authorizer = auth
ihrc.Sender = sender
ihrc.SkipResourceProviderRegistration = c.skipProviderRegistration
```

can we swap this over to using the newer style registration function? e.g.

```
c.configureClient(&ihrc.Client, auth)
```

vendor/vendor.json
tombuildsstuff: taking a look into this - this is because SDKv14 is vendored in, downgrading to `v12.5` should fix this for the moment (we'll upgrade in the near future)

vendor/vendor.json
```diff
@@ -106,6 +106,12 @@
 			"version": "v12.5.0-beta",
 			"versionExact": "v12.5.0-beta"
 		},
+		{
+			"checksumSHA1": "l4BaDOG5showWuLiGBjsY4eVhtk=",
+			"path": "github.com/Azure/azure-sdk-for-go/services/iothub/mgmt/2017-07-01/devices",
+			"revision": "1e334c402ea1460704b0263e5d188f28ad946ce1",
+			"revisionTime": "2018-02-16T17:41:56Z"
```
tombuildsstuff: can we vendor this using the `v12.5.0-beta` tag? I believe that should mean we can downgrade to `v10` of AutoRest too - you can do this via:

```
govendor update github.com/Azure/azure-sdk-for-go/services/iothub/mgmt/2017-07-01/devices@=v12.5.0-beta
```

website/docs/r/iothub.html.markdown
```diff
@@ -0,0 +1,92 @@
...
+		name = "S1"
+		tier = "Standard"
+		capacity = "1"    
+	}
+
+	tags {
+		"purpose" = "testing"
+	    }
+	}
+
```
tombuildsstuff: _minor_ could we make the spacing consistent here by switching to use spaces instead of tabs?

website/docs/r/iothub.html.markdown
```diff
@@ -0,0 +1,92 @@
...
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+---
+
+A `sku` block supports the following:
+
+* `name` - (Required) The name of the sku. Supports `F1`, `S1`, `S2`, and `S3`.
+
+* `tier` - (Required) The billing tier for the IoT Hub. `Free` or `Standard`.
```
tombuildsstuff: can we make this "Possible values are `free` or `Standard`"

tombuildsstuff:
hey @shelleybess 

Thanks for pushing the latest set of updates - just to let you know that I'm going to push a few commits to fix the issues outlined above so that we can ship this as part of the v1.3.0 release of the AzureRM Provider; I hope you don't mind!

Thanks!

shelleybess:
@tombuildsstuff 

That's great, thank you! I don't mind at all. I'm sorry I haven't had time to wrap that up. The deletion of IoTHub is still timing out. I have not made any changes to that yet.

Thank you :)

tombuildsstuff:
hey @shelleybess 

> That's great, thank you! I don't mind at all. I'm sorry I haven't had time to wrap that up.

No worries at all :)

> The deletion of IoTHub is still timing out. I have not made any changes to that yet.

Thanks for confirming that - taking a look into this this appears to be a bug in the Azure SDK, which is coming from a badly defined Swagger document (since the Go SDK is generated) - as such I've opened this issue to get that fixed: https://github.com/Azure/azure-rest-api-specs/issues/2661

In the interim we should be able to work around this by replacing the Delete waiter with a custom waiter (here's an example in the Template Deployment resource](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_template_deployment.go#L266)) - I'll push a commit to add one (since I've got the branch checked out)

Thanks!

tombuildsstuff:
Tests pass:

screen shot 2018-03-14 at 21 44 14

Issue #455, opened 2017-10-25
azurerm_network_security_group in 0.3.0+ wants to continually reapply


clangaxon:
Upgrading from 0.2.2 to 0.3.1, Terraform always shows a pending modification on all azurerm_network_security_group objects. It appears to be that it now sorts the security rules by priority. This seems like a good thing going forward, but it disagrees with the existing state.  Apply won't fix this as there are no actual changes made.

Is there a possible fix or simple work around?

Thanks.

dominik-lekse:
Thanks @clangaxon for opening this issue.

I can confirm this behavior appeared after upgrading from a 0.2.x to the recent version 0.3.1. Without too much investigation, I suspect a state migration is needed to fix this.

As a manual fix, you could try to reorder the corresponding items of the nsg objects in the Terraform state.

clangaxon:
I suspect you are correct in that we could manually reorder them - however this would be hand editing thousands of lines, as we have many very large configs. :)

amosdavis:
I have tried re-ordering as you suggested and it doesn't solve the problem.  I am still prompted to re-apply the same nsg.

clangaxon:
I tried removing all the rules from an NSG and Terraform fails to do anything on apply. Seems like azurerm_network_security_group is completely broken in 0.3+.

dominik-lekse:
@clangaxon Thanks for your research on this. I confirm that I also had no success when reordering the items in the Terraform state.

privatwolke:
I had a look at the code responsible for handling network security groups and I believe that the commit that broke this was https://github.com/terraform-providers/terraform-provider-azurerm/commit/fbdcf746c8d8bf5b99b48764720432ac645c7bab#diff-02a62c4182918d899d123e535c80ac12L40 by @tombuildsstuff where the container type for handling security rules was changed from a set to a list. The desire might have been to simplify the handling and not having to compute hash values for security rules.

dominik-lekse:
Might also be the case that the Azure API returns the list of rules in a nsg in a non-deterministic order. When the rules in the provider have been handled as a set this might not have mattered, but as far as I know, a list in a Terraform schema is order-aware.

This is an unverified guess, but I have learned that the Azure API is full of surprises.

luisamador:
I'm hitting the same issue. We are on Terraform v0.10.8 and using the azurerm 0.3 provider. It is annoying because it's breaking Terraform's idempotency.

[...]

PR #893, opened 2018-02-26
Reverting the change to make Nested security rules from List to Set. …

VaijanathB:
This had caused regression #455.  I have verified that resources created after 0.3.1 also work properly once we revert this changes. 
I have locally run all the tests using the following command and all the tests pass. 
make testacc TEST=./azurerm TESTARGS="-run=TestAccAzureRMNetwork"

Also I have verified that Issue #455 is resolved when we make this change.


azurerm/resource_arm_network_security_group.go
```diff
@@ -216,8 +219,23 @@ func resourceArmNetworkSecurityGroupDelete(d *schema.ResourceData, meta interfac
...
+	var buf bytes.Buffer
+	m := v.(map[string]interface{})
+	buf.WriteString(fmt.Sprintf("%s-", m["protocol"].(string)))
+	buf.WriteString(fmt.Sprintf("%s-", m["source_port_range"].(string)))
+	buf.WriteString(fmt.Sprintf("%s-", m["destination_port_range"].(string)))
+	buf.WriteString(fmt.Sprintf("%s-", m["source_address_prefix"].(string)))
+	buf.WriteString(fmt.Sprintf("%s-", m["destination_address_prefix"].(string)))
+	buf.WriteString(fmt.Sprintf("%s-", m["access"].(string)))
+	buf.WriteString(fmt.Sprintf("%d-", m["priority"].(int)))
+	buf.WriteString(fmt.Sprintf("%s-", m["direction"].(string)))
```
tombuildsstuff: since all items are contained within the Set's hash function, we actually don't need to specify this function at all - so can we remove this method? the default set hash function will include all values in the set

azurerm/resource_arm_network_security_group.go
```diff
@@ -216,8 +219,23 @@ func resourceArmNetworkSecurityGroupDelete(d *schema.ResourceData, meta interfac
...
+
+	return hashcode.String(buf.String())
+}
+
+func flattenNetworkSecurityRules(rules *[]network.SecurityRule) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0, len(*rules))
```
tombuildsstuff: there's a crash here is `rules` is nil - can we make this:

```
result := make([]map[string]interface{}, 0)
```

azurerm/resource_arm_network_security_group.go
VaijanathB: Agreed. removed this function.

azurerm/resource_arm_network_security_group.go
VaijanathB: Fixed it.

VaijanathB:
I have addressed both of these comments. I also reran all the Network Acceptance test locally and they pass with this change.

tombuildsstuff:
I've rebased this - tests pass:

screen shot 2018-03-01 at 14 33 24

tombuildsstuff:
@VaijanathB has confirmed the upgrade path from both 0.2.x -> this branch & 1.1.2 -> this branch

franciscocabral:
I don't know where I should comment this, but the [docs](https://www.terraform.io/docs/providers/azurerm/d/storage_account.html#account_kind) need to be updated.

tombuildsstuff:
@franciscocabral would you mind opening a new issue about that with some more information about what needs to be updated? Thanks! :)

PR #905, opened 2018-02-28
New Resource: `azurerm_application_security_group`

### LINKED ISSUE #369: Feature Request: Application Security Groups
Hi, Following the announcements at Ignite it would be nice to explore the Application Security Groups functionality which is now in preview via terraform. It seems to make security groups more like AWS and infinitely more friendly than they currently are.

https://docs.microsoft.com/en-us/azure/virtual-network/create-network-security-group-preview

tombuildsstuff:
Tasks:

- [x] Upgrade to v12 of the Azure SDK for Go
- [x] Document that this is still in Public Preview
- [x] Fix the rebase issue in `azurerm/config.go`

Fixes #369

### DIFF
--- a/azurerm/config.go
+++ b/azurerm/config.go
@@ -324,67 +324,15 @@ func getArmClient(c *authentication.Config) (*ArmClient, error) {
 		return keyVaultSpt, nil
 	})
 
-	// NOTE: these declarations should be left separate for clarity should the
-	// clients be wished to be configured with custom Responders/PollingModes etc...
-	asc := compute.NewAvailabilitySetsClientWithBaseURI(endpoint, c.SubscriptionID)
-	setUserAgent(&asc.Client)
-	asc.Authorizer = auth
-	asc.Sender = sender
-	client.availSetClient = asc
-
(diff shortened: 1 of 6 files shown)


tombuildsstuff:
Tests pass:

screen shot 2018-02-28 at 11 47 37

azurerm/config.go
```diff
@@ -323,22 +324,15 @@ func getArmClient(c *authentication.Config) (*ArmClient, error) {
 		return keyVaultSpt, nil
 	})
 
-	csc := containerservice.NewContainerServicesClientWithBaseURI(endpoint, c.SubscriptionID)
```
katbyte: I'm at a loss as to why this was removed as it seems unrelated?

tombuildsstuff: So this is actually being registered twice in master at the moment (due to a previous merge conflict):

https://github.com/terraform-providers/terraform-provider-azurerm/blob/f76c05bc2cc04e0a1974d0182497abc07abeba18/azurerm/config.go#L326

https://github.com/terraform-providers/terraform-provider-azurerm/blob/f76c05bc2cc04e0a1974d0182497abc07abeba18/azurerm/config.go#L503

this consolidates this into one

PR #904, opened 2018-02-28
Changed admin_username and key_data to require a new resource. 

katbyte:
Also fixed a terraform crash when key_data is empty. Added Kubernetes cluster example. Fixes #903, #895

PR #907, opened 2018-02-28
Export FQDN for managed Kubernetes cluster

### LINKED ISSUE #906: FQDN missing from managed Kubernetes cluster
### Terraform Version
Terraform v0.11.3
Azure provider v1.1.2

### Affected Resource(s)
- azurerm_kubernetes_cluster

```hcl
resource "azurerm_kubernetes_cluster" "aks_container" {
    name        = "akc-${random_integer.random_int.result}"
    location    = "${var.resource_group_location}"
    dns_prefix  = "akc-${random_integer.random_int.result}"

    resource_group_name = "${azurerm_resource_group.akc-rg.name}"
    kubernetes_version  = "1.8.7"


    linux_profile {
        admin_username = "${var.linux_admin_username}"
        ssh_key {
            key_data = ""
        }
    }

    agent_pool_profile {
        name        = "agentpool"
        count       = "2"
        vm_size     = "Standard_DS2_v2"
        os_type     = "Linux"
    }

    service_principal {
        client_id     = "${var.client_id}"
        client_secret = "${var.client_secret}"
    }
}

output "fqdn" {
  value = "${azurerm_kubernetes_cluster.aks_container.fqdn}"
}
```

### Expected Behavior
FQDN should be set

### Actual Behavior
`output.fqdn: Resource 'azurerm_kubernetes_cluster.aks_container' does not have attribute 'fqdn' for variable 'azurerm_kubernetes_cluster.aks_container.fqdn`

### Steps to Reproduce
1. `terraform apply`

### References
- GH-907

phekmat:
- Set the FQDN for the Kubernetes cluster. Fixes #906 
- Remove the FQDN from the agent_pool_profile as it's not populated by
Azure for managed Kubernetes

### DIFF
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -39,6 +39,11 @@ func resourceArmKubernetesCluster() *schema.Resource {
 				Computed: true,
 			},
 
+			"fqdn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
 			"kubernetes_version": {
 				Type:     schema.TypeString,
 				Optional: true,
@@ -99,8 +104,9 @@ func resourceArmKubernetesCluster() *schema.Resource {
 						},
 
 						"fqdn": {
-							Type:     schema.TypeString,
-							Computed: true,
+							Type:       schema.TypeString,
+							Computed:   true,
+							Deprecated: "This field has been deprecated. Use the parent `fqdn` instead",
 						},
 
 						"vm_size": {
@@ -246,6 +252,7 @@ func resourceArmKubernetesClusterRead(d *schema.ResourceData, meta interface{}) 
 	}
 	d.Set("resource_group_name", resGroup)
 	d.Set("dns_prefix", resp.DNSPrefix)
+	d.Set("fqdn", resp.Fqdn)
 	d.Set("kubernetes_version", resp.KubernetesVersion)
 
 	linuxProfile := flattenAzureRmKubernetesClusterLinuxProfile(*resp.ManagedClusterProperties.LinuxProfile)
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -260,7 +260,7 @@ func resourceArmKubernetesClusterRead(d *schema.ResourceData, meta interface{}) 
 		return fmt.Errorf("Error setting `linux_profile`: %+v", err)
 	}
 
-	agentPoolProfiles := flattenAzureRmKubernetesClusterAgentPoolProfiles(resp.ManagedClusterProperties.AgentPoolProfiles)
+	agentPoolProfiles := flattenAzureRmKubernetesClusterAgentPoolProfiles(resp.ManagedClusterProperties.AgentPoolProfiles, resp.Fqdn)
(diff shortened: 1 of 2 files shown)


phekmat:
I can rebase this on #904 if that makes it easier to get into 1.1.3 as well

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -98,11 +103,6 @@ func resourceArmKubernetesCluster() *schema.Resource {
 							Computed: true,
 						},
 
-						"fqdn": {
```
katbyte: By moving this out we will break existing terraform code. Could we mark this as deprecated to be removed in a future release and properly set it?

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -321,10 +322,6 @@ func flattenAzureRmKubernetesClusterAgentPoolProfiles(profiles *[]containerservi
 			agentPoolProfile["dns_prefix"] = *profile.DNSPrefix
 		}
 
-		if profile.Fqdn != nil {
```
katbyte: Can just be set from profile.Fqdn for now.

phekmat: I'd be surprised that it would break existing code (`profile.Fqdn` has always been empty, so the field hasn't been useful), but I can make that change and deprecate it.

phekmat:
I updated the code to deprecate the FQDN in the agent_pool_profile, but overlooked your latest comment. Should I go ahead and update the code to use the parent FQDN in the `agent_pool_profile.#.fqdn`? 

It feels counterproductive to me since the field will eventually be removed (and any current usages never worked to begin with), but I'm ok with making that change too.

katbyte:
Yes please if you don't mind :)

This was we can merge the new field in sooner and removing the old field later, likely grouping with multiple breaking changes in a larger release.

phekmat:
Updated. Thanks for the review and feedback!

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -260,7 +260,7 @@ func resourceArmKubernetesClusterRead(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf("Error setting `linux_profile`: %+v", err)
 	}
 
-	agentPoolProfiles := flattenAzureRmKubernetesClusterAgentPoolProfiles(resp.ManagedClusterProperties.AgentPoolProfiles)
+	agentPoolProfiles := flattenAzureRmKubernetesClusterAgentPoolProfiles(resp.ManagedClusterProperties.AgentPoolProfiles, *resp.Fqdn)
```
katbyte: In the unlikely situation where `resp.Fqdn` is null this will crash. I suggest passing in the pointer and then checking it for null like all the other values in `agent_pool_profile` and dereferencing it there.

katbyte:
Other then one last minor change this LGTM, once that is done we can get this merged :)

azurerm/resource_arm_kubernetes_cluster.go
phekmat: Good catch. Updated

tombuildsstuff:
Tests pass:

screen shot 2018-03-02 at 12 46 41

PR #918, opened 2018-03-01
New Example: joining a virtual machine to an Active Directory Domain

tombuildsstuff:
This PR adds an end-to-end example for joining a Windows Server (2016) Virtual Machine to a newly created Active Directory Domain. This question comes up a lot - so I believe it's helpful to have an example demonstrating how to use the VM Extension to bind a machine to an Active Directory domain.

For the purposes of this example we spin up an Active Directory Domain on demand (which, whilst on a single node isn't exactly production ready, it's suitable for an example/demo). This also provisions the VM Client with a Public IP, so users can confirm the machine is bound to the domain (it also ensures the binding process works as expected, since machines don't get a hostname/ip configuration from the AD server, but from Azure since that's booted first; given most users will have an Active Directory Domain provisioned ahead of time, I feel this is suitable for demonstration/example purposes).

### DIFF
--- a/examples/vm-joined-to-active-directory/README.md
+++ b/examples/vm-joined-to-active-directory/README.md
@@ -1,0 +1,87 @@
+# Example: Joining a Windows Machine to an Active Directory Domain
+
+This example creates an Active Directory Domain, a Windows Client; to demonstrate how to bind a Windows Client to an Active Directory Domain using a Virtual Machine Extension in Terraform (using [the `azurerm_virtual_machine_extension` resource](https://www.terraform.io/docs/providers/azurerm/r/virtual_machine_extension.html)).
+
+This example is built around [the Virtual Machine Extension](https://www.terraform.io/docs/providers/azurerm/r/virtual_machine_extension.html) found in the `windows-client` module (documented below) - which demonstrates binding a Windows Virtual Machine to an Active Directory Domain. For the purposes of this example we create an Active Directory Domain, since it's easier to demonstrate - however you can achieve the same thing with an existing Active Directory Domain.
+
+## Notes
+
+- This is intended as an example of binding machines to an Active Directory Domain, and **we'd not recommend using it in production as-is** as the configuration has been simplified for example purposes, e.g.:
+  - The Active Directory Forest has a single node, for demonstration purposes
+  - There's no security rules configured on the network, so everything's open internally etc.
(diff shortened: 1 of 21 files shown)


cgriggs01:
👍

examples/vm-joined-to-active-directory/README.md
```diff
@@ -0,0 +1,87 @@
+# Example: Joining a Windows Machine to an Active Directory Domain
+
+This example creates an Active Directory Domain, a Windows Client; to demonstrate how to bind a Windows Client to an Active Directory Domain using a Virtual Machine Extension in Terraform (using [the `azurerm_virtual_machine_extension` resource](https://www.terraform.io/docs/providers/azurerm/r/virtual_machine_extension.html)).
+
+This example is built around [the Virtual Machine Extension](https://www.terraform.io/docs/providers/azurerm/r/virtual_machine_extension.html) found in the `windows-client` module (documented below) - which demonstrates binding a Windows Virtual Machine to an Active Directory Domain. For the purposes of this example we create an Active Directory Domain, since it's easier to demonstrate - however you can achieve the same thing with an existing Active Directory Domain.
+
+## Notes
+
+- This is intended as an example of binding machines to an Active Directory Domain, and **we'd not recommend using it in production as-is** as the configuration has been simplified for example purposes, e.g.:
```
paddycarver: nitpick:

> it is not recommended for production use

feels a bit stronger to me than "we'd not recommend"

examples/vm-joined-to-active-directory/README.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+This will take around 20m to provision - once completed you should see the Public IP Address of the Windows Client machine (which is bound to the Active Directory Domain):
+
+```
+```
+
+## Variables
+
+ * `prefix` - The prefix used for all resources in this example. Needs to be a short (6 characters) alphanumeric string. Example: `myprefix`.
```
paddycarver: nitpick: you say it needs to be 6 characters, and the example is 8.

tombuildsstuff: 🤦‍♂️ good spot

examples/vm-joined-to-active-directory/README.md
tombuildsstuff: fixed

PR #935, opened 2018-03-05
`azurerm_application_gateway` - fixes a crash where ssl_policy is nil when importing

### LINKED ISSUE #933: terraform crashed when trying to import azurerm application gateway
_This issue was originally opened by @avnsiva as hashicorp/terraform#17500. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
Terraform v0.11.2
+ provider.azurerm v1.2.0
[https://gist.github.com/avnsiva/4b1f9763736f5bbfcbc35831e9cef4d5](https://gist.github.com/avnsiva/4b1f9763736f5bbfcbc35831e9cef4d5)

tombuildsstuff:
This isn't possible to write a test for since it relies on an older version of the schema being returned from the API - but I've verified the response [in this crash log](https://gist.github.com/avnsiva/4b1f9763736f5bbfcbc35831e9cef4d5) which confirms this isn't returned (and [that it'll crash here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_application_gateway.go#L1362)

Fixes #933

### DIFF
--- a/azurerm/resource_arm_application_gateway.go
+++ b/azurerm/resource_arm_application_gateway.go
@@ -1359,9 +1359,11 @@ func flattenApplicationGatewayWafConfig(waf *network.ApplicationGatewayWebApplic
 func flattenApplicationGatewaySslPolicy(policy *network.ApplicationGatewaySslPolicy) []interface{} {
 	result := make([]interface{}, 0)
 
-	if protocols := policy.DisabledSslProtocols; protocols != nil {
-		for _, proto := range *protocols {
-			result = append(result, string(proto))
+	if pol := policy; policy != nil {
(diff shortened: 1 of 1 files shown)


avnsiva:
Hi, Great to know that this is fixed. Will I be able to get a binary of terraform with this fix?

tombuildsstuff:
@avnsiva this'll be released in v1.3 in the near future (we'll comment on the issue when that's released), if you need it before then [you can build a custom binary using the instructions in the readme](https://github.com/terraform-providers/terraform-provider-azurerm#building-the-provider)

Hope that helps :)

PR #940, opened 2018-03-06
New data source: azurerm_subscriptions

katbyte:


azurerm/helpers/subscription/subscription.go
```diff
@@ -0,0 +1,40 @@
...
+
+		"quota_id": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
+
+		"spending_limit": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
```
tombuildsstuff: do you think it's also worth exposing the TenantID here?

azurerm/data_source_subscriptions.go
```diff
@@ -0,0 +1,62 @@
...
+
+		s := make(map[string]interface{})
+
+		s["subscription_id"] = *val.SubscriptionID
+		s["display_name"] = *val.DisplayName
+		s["state"] = val.State
+		if policies := val.SubscriptionPolicies; policies != nil {
+			s["location_placement_id"] = *policies.LocationPlacementID
+			s["quota_id"] = *policies.QuotaID
+			s["spending_limit"] = policies.SpendingLimit
```
tombuildsstuff: we should add nil checks around these, otherwise there's a potential crash

website/docs/d/subscriptions.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+---
+
+# Data Source: azurerm_subscription
+
+Use this data source to access a list of all Azure subscription currently available.
+
+## Example Usage
+
+```hcl
+data "azurerm_subscriptions" "availible" {}
```
tombuildsstuff: _minor_ `availible` -> `available`

website/docs/d/subscriptions.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+
+output "availible_subscriptions" {
```
tombuildsstuff: (same here)

website/docs/d/subscriptions.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+  value = "${data.azurerm_subscriptions.current.subscriptions}"
+}
+
+output "first_availible_subscription_display_name" {
```
tombuildsstuff: (same here)

azurerm/helpers/subscription/subscription.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func SubscriptionSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"subscription_id": {
+			Type:     schema.TypeString,
+			Optional: true,
```
tombuildsstuff: Given this is only optional for the `Subscription` Data Source - I believe we don't want this set for the `Subscriptions` Data Source? We can override the schema for the Subscription data source to add this, there's examples in the AWS Provider

azurerm/helpers/subscription/subscription.go
katbyte: The tenant will always be what the provider is using and accessible from `client_config` data source.

azurerm/helpers/subscription/subscription.go
```diff
@@ -4,11 +4,11 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
-func SubscriptionSchema() map[string]*schema.Schema {
-	return map[string]*schema.Schema{
+func SubscriptionSchema(subscriptionIdOptional bool) map[string]*schema.Schema {
```
paultyng: super minor, but use `ID` (see https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

katbyte:
Updated with requested changes.

PR #946, opened 2018-03-07
Add vpn client protocol support for virtual network gateway

dcaro:
Added support for vpn client protocol and radius server with the following fields: 
vpn_client_protocol
radius_server_address
radius_server_secret


azurerm/resource_arm_virtual_network_gateway.go
```diff
@@ -660,3 +728,35 @@ func validateArmVirtualNetworkGatewayExpressRouteSku() schema.SchemaValidateFunc
...
 	}, true)
 }
+
+func resourceArmVirtualNetworkGatewayCustomizeDiff(diff *schema.ResourceDiff, v interface{}) error {
+
+	if vpnClient, ok := diff.GetOk("vpn_client_configuration"); ok {
+		if vpnClientConfig, ok := vpnClient.([]interface{})[0].(map[string]interface{}); ok {
+			if vpnClientConfig["radius_server_address"] != "" {
+				if !validIP4(vpnClientConfig["radius_server_address"].(string)) {
+					return fmt.Errorf("radius_server_address is not an IPv4 address")
```
katbyte: This validation can be done in the schema, for example:

```go
				ValidateFunc: validation.StringMatch(
					regexp.MustCompile("^[-a-z0-9]{3,50}$"),
					"Cosmos DB Account name must be 3 - 50 characters long, contain only letters, numbers and hyphens.",
				),
```

azurerm/resource_arm_virtual_network_gateway.go
```diff
@@ -660,3 +728,35 @@ func validateArmVirtualNetworkGatewayExpressRouteSku() schema.SchemaValidateFunc
...
+				}
+				if vpnClientConfig["radius_server_secret"] == "" {
```
katbyte: as above, could be validated by schema & regex

azurerm/resource_arm_virtual_network_gateway.go
```diff
@@ -660,3 +728,35 @@ func validateArmVirtualNetworkGatewayExpressRouteSku() schema.SchemaValidateFunc
 		string(network.VirtualNetworkGatewaySkuTierUltraPerformance),
 	}, true)
 }
+
+func resourceArmVirtualNetworkGatewayCustomizeDiff(diff *schema.ResourceDiff, v interface{}) error {
+
+	if vpnClient, ok := diff.GetOk("vpn_client_configuration"); ok {
+		if vpnClientConfig, ok := vpnClient.([]interface{})[0].(map[string]interface{}); ok {
+			if vpnClientConfig["radius_server_address"] != "" {
```
katbyte: I think what you want to do here is check to make sure one or the other is set as mentioned above?

as in:
```go
_, hasRadius := vpnClientConfig["radius_server_address"]
_, hasRootCert := vpnClientConfig["root_certificate"]

if !hasRadius && !hasRootCert {
    #error
}

if hasRadius {
    if _, hasRadiusSecret := vpnClientConfig["radius_server_secret"]; !hasRadiusSecret {
        #error
    }
}

```

azurerm/resource_arm_virtual_network_gateway.go
```diff
@@ -660,3 +728,35 @@ func validateArmVirtualNetworkGatewayExpressRouteSku() schema.SchemaValidateFunc
...
+				if vpnClientConfig["radius_server_secret"] != "" {
+					return fmt.Errorf("radius_server_secret cannot be set when radius_server_address is empty")
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func validIP4(ipAddress string) bool {
```
katbyte: could we rename this to `validateIP4`?

azurerm/resource_arm_virtual_network_gateway_test.go
```diff
@@ -269,11 +276,20 @@ resource "azurerm_virtual_network_gateway" "test" {
     private_ip_address_allocation = "Dynamic"
     subnet_id = "${azurerm_subnet.test.id}"
   }
+
+  vpn_client_configuration {
```
katbyte: Could we add an entirely new test for these properties instead of adding them to an existing test?

azurerm/resource_arm_virtual_network_gateway.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+	"regexp"
```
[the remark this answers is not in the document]
dcaro: sure, updated.

azurerm/resource_arm_virtual_network_gateway.go
dcaro: The schema is indeed a better place to do this check.

azurerm/resource_arm_virtual_network_gateway.go
dcaro: With this hasRadius and hasRootCert are set to true. I have changed a bit the logic to check if both radius server address and value are set. If this neither of them are set, we can assume that we are using the certificate method and nil is accepted.

azurerm/resource_arm_virtual_network_gateway_test.go
dcaro: sure, done.

dcaro:
I have just pushed an update with the changes that you have suggested. Let me know if there are any other points. Thanks for the review.

katbyte:
Tests pass aside from the usual flakiness:

screen shot 2018-06-14 at 23 48 23
screen shot 2018-06-14 at 23 48 43

Phydeauxman:
I am trying to use the new `vpn_client_protocol` argument and it is not working.  I am running Terraform v0.11.7 and provider.azurerm v1.8.0.  The output I am getting from plan is:

`Error: azurerm_virtual_network_gateway.sdvnet: vpn_client_configuration.0: invalid or unknown key: vpn_client_protocol`

PR #962, opened 2018-03-11
New Resource: `azurerm_cognitive_account`

### LINKED ISSUE #908: Feature Request: Cognitive Service Account support
Hi there,

There is no support for azure [cognitive service Accounts](https://azure.microsoft.com/en-us/services/cognitive-services/) there is support for it on the [Go SDK](https://github.com/Azure/azure-sdk-for-go/tree/master/services/cognitiveservices) docs for the usage [here](https://docs.microsoft.com/en-gb/azure/#pivot=products&panel=ai)

### Affected Resource(s)

- cognitive_account

Nepomuceno:
Note: This is my first full resource so this might take more time to implement than desired

To Do: 
 

- [x]    Resource Implementation:
    - [x]  Create
    - [x]  Update
    - [x]  Read
    - [x]  Delete
    - [x]  Verify the Schema design is correct
    - [x]  Add Validation
- [x]  Tests
    - [x]  Basic Tests
    - [x]  Validation Tests
- [x]  Documentation

Fixes #908


azurerm/resource_arm_cognitive_account.go
```diff
@@ -0,0 +1,258 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"kind": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateCognitiveAccountKind,
```
tombuildsstuff: we can actually in-line these by doing:

```
	"Academic",
	"Bing.Autosuggest",
	"Bing.Autosuggest.v7",
	"Bing.CustomSearch",
	"Bing.Search",
	"Bing.Search.v7",
	"Bing.Speech",
	"Bing.SpellCheck",
	"Bing.SpellCheck.v7",
	"ComputerVision",
	"ContentModerator",
	"CustomSpeech",
	"Emotion",
	"Face",
	"LUIS",
	"Recommendations",
	"SpeakerRecognition",
	"Speech",
	"SpeechTranslation",
	"TextAnalytics"
	"TextTranslation",
	"WebLM",
}, false),
```

azurerm/resource_arm_cognitive_account.go
```diff
@@ -0,0 +1,258 @@
...
+			},
+
+			"sku": {
+				Type:     schema.TypeSet,
```
tombuildsstuff: we should be able to make this a `TypeList` rather than a `TypeSet`- which means it can be expanded via:

```
inputs := d.Get("sku").([]interface{})
input := inputs[0].(map[string]interface{})
# key into input to get the values
```

and flattened via:

```
func flatten... (input *cognitiveservuces.Sku)[]interface{} {
  output := make(map[string]interface{}, 0)
  output["foo"] = input.Bar
  return []interface{}{output}
}
```

azurerm/resource_arm_cognitive_account.go
```diff
@@ -0,0 +1,258 @@
...
+
+						"tier": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ForceNew:     false,
+							ValidateFunc: validateCognitiveAccountSkuTier,
+						},
+					},
+				},
+				Set: resourceArmCognitiveAccountSkuHash,
```
tombuildsstuff: (which would mean we're able to remove this)

azurerm/resource_arm_cognitive_account.go
```diff
@@ -0,0 +1,258 @@
...
+							Required:     true,
+							ForceNew:     false,
+							ValidateFunc: validateCognitiveAccountSkuName,
+						},
+
+						"tier": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ForceNew:     false,
+							ValidateFunc: validateCognitiveAccountSkuTier,
```
tombuildsstuff: can we remove the validation for the SKU Name & Tier? In general it's simpler to let Azure handle that as it can vary by account/region

azurerm/resource_arm_cognitive_account.go
```diff
@@ -0,0 +1,258 @@
...
+
+	return hashcode.String(buf.String())
+}
+
+type dummy struct{}
+
+func resourceArmCognitiveAccountCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).cognitiveAccountsClient
+	ctx := meta.(*ArmClient).StopContext
+	fmt.Printf("[INFO] preparing arguments for Azure ARM  Cognitive Servies Account creation.\n")
```
tombuildsstuff: we should be able to make this:

```
log.Printf("[INFO] preparing arguments for Azure ARM  Cognitive Servies Account creation")
```

azurerm/resource_arm_cognitive_account.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	account, err := client.Create(ctx, resGroup, name, createParameter)
+	if err != nil {
+		fmt.Printf("Error: %s \n", err.Error())
+		return err
+	}
+	if account.AccountProperties.ProvisioningState != cognitiveservices.Succeeded {
+		return fmt.Errorf("Error provisioning Account")
+	}
```
tombuildsstuff: I think we're safe to remove this check (since the SDK will return an error if it's not a `200 OK`)

azurerm/resource_arm_cognitive_account.go
```diff
@@ -0,0 +1,258 @@
...
+		Sku:        &sku,
+		Properties: &prop,
+		Tags:       *expandTags(tags),
+	}
+
+	account, err := client.Create(ctx, resGroup, name, createParameter)
+	if err != nil {
+		fmt.Printf("Error: %s \n", err.Error())
+		return err
+	}
```
tombuildsstuff: can we change this to return the error:

```
return fmt.Errorf("Error creating Cognitive Services Account %q (Resource Group %q): %+v", name, resourceGroup, err)
```

azurerm/resource_arm_cognitive_account.go
```diff
@@ -0,0 +1,258 @@
...
+
+	resGroup := id.ResourceGroup
+	name := id.Path["accounts"]
+	resp, err := client.GetProperties(ctx, resGroup, name)
+	if err != nil {
+		return err
+	}
+	d.Set("name", resp.Name)
+	d.Set("location", azureRMNormalizeLocation(*resp.Location))
```
tombuildsstuff: for older resources the Azure API doesn't always match the response schema, can we make this:

```
}
```

to avoid a crash in these circumstances?

azurerm/resource_arm_cognitive_account.go
```diff
@@ -0,0 +1,258 @@
...
+	d.Set("resource_group_name", resGroup)
+	d.Set("kind", resp.Kind)
+	d.Set("sku", flattenCognitiveSku(*resp.Sku))
```
tombuildsstuff: given this is a complex object - can we set this via:

```
  flattenedSku := flattenCognitiveSku(*sku)
  if err := d.Set("sku", flattenedSku"); err != nil {
   return fmt.Errorf("Error flattening `sku`: %+v", err)
  }
}
```

azurerm/resource_arm_cognitive_account.go
```diff
@@ -0,0 +1,258 @@
...
+	return nil
+}
+
+func resourceArmCognitiveAccountDelete(d *schema.ResourceData, meta interface{}) error {
+	fmt.Printf("[INFO] resource Arm Cognitive account Delete.\n")
+	client := meta.(*ArmClient).cognitiveAccountsClient
+	ctx := meta.(*ArmClient).StopContext
+	name := d.Get("name").(string)
+	resGroup := d.Get("resource_group_name").(string)
+	_, err := client.Delete(ctx, resGroup, name)
```
tombuildsstuff: rather than pulling these out of the state - can we parse them from the URI (as in the read function above):

```
if err != nil {
  return err
}

resGroup := id.ResourceGroup
name := id.Path["accounts"]
```

azurerm/resource_arm_cognitive_account_test.go
```diff
@@ -0,0 +1,115 @@
...
+			resp, err := client.Get(ctx, resourceGroup, name)
+
+			if err != nil {
+				if utils.ResponseWasNotFound(resp.Response) {
+					return nil
+				}
+
+				return err
+			}
+		*/
```
tombuildsstuff: can we add this check back in? This is used to ensure the Cognitive Services account has been deleted

azurerm/resource_arm_cognitive_account_test.go
```diff
@@ -0,0 +1,115 @@
...
+		client := testAccProvider.Meta().(*ArmClient).appServiceEnvironmentsClient
+		ctx := testAccProvider.Meta().(*ArmClient).StopContext
+		resp, err := client.Get(ctx, resourceGroup, appServiceEnvironmentName)
+		if err != nil {
+			if utils.ResponseWasNotFound(resp.Response) {
+				return fmt.Errorf("Bad: App Service Environment %q (resource group: %q) does not exist", appServiceEnvironmentName, resourceGroup)
+			}
+
+			return fmt.Errorf("Bad: Get on appServiceEnvironmentsClient: %+v", err)
+		}*/
```
tombuildsstuff: can we add this back in? This is used to ensure that the Cognitive Services account exists

azurerm/resource_arm_cognitive_account_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	// "github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+// TODO: an update test going from External -> Internal -> External
+
+func TestAccAzureRMCognitiveAccount_basic(t *testing.T) {
```
tombuildsstuff: can we add a secondary `TestAccAzureRMCognitiveAccount_complete` test featuring some Tags?

tombuildsstuff: can we also add an `_import` test for both of these in `azurerm/import_arm_cognitive_account_test.go`? there's an example of what this looks like here: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/import_arm_key_vault_test.go

azurerm/resource_arm_cognitive_account_test.go
```diff
@@ -0,0 +1,115 @@
...
+	log.SetOutput(os.Stdout)
```
tombuildsstuff: can we remove this line?

tombuildsstuff:
ping @Nepomuceno :)

Nepomuceno:
Sorry getting back to it today 😄

tombuildsstuff:
hey @Nepomuceno 

Just to let you know that since this required a rebase I've gone ahead and done that (and at the same time made the pending changes needed so that we can get this merged - I hope you don't mind!)

Thanks!

tombuildsstuff:
Tests pass:

```
$ acctests azurerm TestAccAzureRMCognitiveAccount_
=== RUN   TestAccAzureRMCognitiveAccount_basic
--- PASS: TestAccAzureRMCognitiveAccount_basic (96.87s)
=== RUN   TestAccAzureRMCognitiveAccount_complete
--- PASS: TestAccAzureRMCognitiveAccount_complete (93.81s)
=== RUN   TestAccAzureRMCognitiveAccount_update
--- PASS: TestAccAzureRMCognitiveAccount_update (100.22s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	291.778s
```

azurerm/provider.go
```diff
@@ -144,6 +144,7 @@ func Provider() terraform.ResourceProvider {
 			"azurerm_availability_set":                        resourceArmAvailabilitySet(),
 			"azurerm_cdn_endpoint":                            resourceArmCdnEndpoint(),
 			"azurerm_cdn_profile":                             resourceArmCdnProfile(),
+			"azurerm_cognitive_account":                       resourceArmCognitiveAccount(),
```
tombuildsstuff: @katbyte I'm not sure if this should be `azurerm_cognitive_services_account` - WDYT?

katbyte: Unless there could be a `azurerm_cognitive_services_account` and a `azurerm_cognitive_account` i think the services part noise and would prefer it as is.

tombuildsstuff:
Tests pass:

screenshot 2018-10-01 at 20 56 07

PR #965, opened 2018-03-12
azurerm_virtual_machine storage_os_disk and storage_data_disk force a new resource

### LINKED ISSUE #956: Cannot change osDisk.name, resource azurerm_virtual_machine should be destroyed an re-created
Hi there,

I'm trying to change the name of the `storage_os_disk` of my instance and plan says it'll only be an update in-place but it ends up as an error since it's not permitted to change it, even in the console. Consequently, I think it should be destroy/recreate when that change occurs.

### Terraform Version
Tested on 0.10.8 and 0.11.3

### Affected Resource(s)
azurerm_virtual_machine

```hcl
resource "azurerm_virtual_machine" "my_instance" {
...
  storage_os_disk {
    name              = "${var.custom_disk_name}"
    caching           = "ReadWrite"
    create_option     = "FromImage"
    managed_disk_type = "Standard_LRS"
  }
...
}
```

### Debug Output
Here is the result of plan and apply : 
https://gist.github.com/jnancel-claranet/c5f7f986d829214402c28fcb017e7437

### Steps to Reproduce
1. Change name attribute in `os_storage_disk` in `azurerm_virtual_machine`
2. `terraform plan`
3. `terraform apply`

katbyte:
fixes #956.

Also some minor fixes to the documentation.

### DIFF
--- a/azurerm/resource_arm_virtual_machine.go
+++ b/azurerm/resource_arm_virtual_machine.go
@@ -174,6 +174,7 @@ func resourceArmVirtualMachine() *schema.Resource {
 						"name": {
 							Type:     schema.TypeString,
 							Required: true,
+							ForceNew: true,
 						},
 
 						"vhd_uri": {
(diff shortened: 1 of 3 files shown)


katbyte:
No suspicious failing tests.
screen shot 2018-03-13 at 11 08 45

PR #972, opened 2018-03-13
`azurerm_sql_database` - support for importing from bacpac

hbuckle:
Add ability to import a database from a bacpac (https://docs.microsoft.com/en-us/azure/sql-database/sql-database-import)

I can't see any way to add a test for this one because it needs a bacpac file in a storage account somewhere.

I've tested it using an AdventureWorks bacpac (https://s3-eu-west-1.amazonaws.com/resourcescrono/AdventureWorks2014FullForAzure.bacpac)


azurerm/resource_arm_sql_database.go
```diff
@@ -237,6 +285,27 @@ func resourceArmSqlDatabaseCreateUpdate(d *schema.ResourceData, meta interface{}
 		return err
 	}
 
+	if _, ok := d.GetOk("database_import"); ok {
+		if !strings.EqualFold(createMode, "default") {
+			return fmt.Errorf("database_import can only be used when create_mode is Default")
```
vlouwagie: @hbuckle Should this error be a validation on the `database_import` property instead? Otherwise, you could be running your scripts and end up with the error during a deployment instead of when generating the plan? Or possibly use `ConflictsWith: ["create_mode"]` on the `database_import` since the default for `create_mode` is `default` (and its required to be default).

tombuildsstuff: @vlouwagie ideally yes, however unfortunately the validation API's only expose a single property at this point; as such unfortunately I don't think that's feasible at this time

azurerm/resource_arm_sql_database.go
```diff
@@ -237,6 +285,27 @@ func resourceArmSqlDatabaseCreateUpdate(d *schema.ResourceData, meta interface{}
...
+		importParameters := expandAzureRmSqlDatabaseImport(d)
+		importFuture, err := client.CreateImportOperation(ctx, resourceGroup, serverName, name, "terraform", importParameters)
+		if err != nil {
+			return err
+		}
+
+		// this is set in config.go, but something sets
+		// it back to 15 minutes, which isn't long enough
+		// for most imports
+		client.Client.PollingDuration = 60 * time.Minute
```
tombuildsstuff: from what I recall the 15 minute reset is pulled from the API's retry interval recommendation * the number of retries allowed - I've been meaning to investigate this further

tombuildsstuff (changes requested):
hey @hbuckle 

Thanks for this PR :) apologies for the delayed review here!

I've taken a look through and left some comments in-line, but the main thing missing here is an acceptance test covering this functionality. I believe it should be possible to spin up a DB, take a backup and them import that into a new DB all via the API to create the test rather than caching a file somewhere perminantly - what do you think?

Thanks!

hbuckle:
@tombuildsstuff RE the acceptance test, using the Azure portal I created a new blank database, exported it to a bacpac in a storage account and then imported it into a second new database. However the import fails with the error "Archive file cannot be size 0" so I think we need a way to get some data into the database. Do you know a way to do this using the API ?

tombuildsstuff:
@hbuckle sorry I'd missed your comment about the AdventureWorks bacpac, I guess we can place that somewhere and restore it? Let me reach out internally to see where's best to do that, but we should be able to upload it to blob storage from wherever it's hosted and then restore it as needed :)

hbuckle:
@tombuildsstuff  Yes I think we basically just need a long lived storage account to do the restore from. Let me know if you have one available.

hbuckle:
Actually I have just seen the azurerm_storage_blob resource can upload a local file to storage, so I can include a bacpac in the testdata folder. I'll go down that route

tombuildsstuff:
@hbuckle rather than storing a large binary file in the repository, it may be better to use a SQL Server Driver such as [this one](https://github.com/denisenkom/go-mssqldb) to create a table and insert some random test data (so we have this in code) - what do you think?

hbuckle:
@tombuildsstuff my only concern is that requires connecting directly to the DB, which probably won't work if you are behind a corporate firewall. I can make a bacpac that is much smaller than adventureworks, only a few kb and include that?

tombuildsstuff:
@hbuckle after chatting internally as long as it's a small file (e.g. a couple hundred KB) this should be fine - we were mostly concerned about large binary files :)

hbuckle:
@tombuildsstuff Acceptance test added

```
go test github.com/terraform-providers/terraform-provider-azurerm/azurerm -v -timeout 600s -run TestAccAzureRMSqlDatabase_bacpac
=== RUN   TestAccAzureRMSqlDatabase_bacpac
--- PASS: TestAccAzureRMSqlDatabase_bacpac (312.25s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm       312.288s
```

hbuckle:
@tombuildsstuff anything more you need from me on this one?

tombuildsstuff:
@hbuckle apologies for the delay, I've been out for a few weeks - I'm taking a look at this now :)

azurerm/resource_arm_sql_database.go
```diff
@@ -57,6 +58,56 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+							Type:      schema.TypeString,
+							Required:  true,
+							Sensitive: true,
+						},
+						"storage_key_type": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"StorageAccessKey", "SharedAccessKey",
+							}, true),
```
tombuildsstuff: this `true` represents the strings being case-insensitive; as such can we also add `DiffSuppressFunc: ignoreCaseDiffSuppressFunc` here to handle the casing being different?

azurerm/resource_arm_sql_database.go
```diff
@@ -57,6 +58,56 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+								"SQL", "ADPassword",
+							}, true),
+						},
+						"operation_mode": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "Import",
+							ValidateFunc: validation.StringInSlice([]string{
+								"Import",
+							}, true),
```
tombuildsstuff: (same here)

tombuildsstuff (approved):
hey @hbuckle 

Thanks for pushing those updates - apologies on the delayed re-review of this. I've taken a look through and left a couple of minor comments in-line for which I'll push a commit to fix, but this otherwise LGTM - thanks for this!

Thanks!

tombuildsstuff:
Tests pass:

screen shot 2018-06-19 at 12 55 32

PR #975, opened 2018-03-13
Revert "azurerm_virtual_machine storage_os_disk and storage_data_disk force a new resource"

katbyte:
Reverts terraform-providers/terraform-provider-azurerm#965.

Having ForceNew here causes an entire new resource when a DataDisk is added. A CustomizeDiff is required to correctly solve this.


katbyte:
I'm going to add the `CustomizeDiff` function onto that branch,

PR #976, opened 2018-03-13
The number of Fault/Update domains changes per-region

### LINKED ISSUE #944: Availability set and PlatformFaultDomainCount
Trying to deploy on Azure within the wrong region could have some severe consequences. 
For example, using the West Central US, will lead to errors when using the availability set defaults

### Terraform Version
`Terraform v0.11.3`

### Affected Resource(s)
- azurerm_availability_set

```hcl
  name     = "resourceGroup1"
  location = "West US"
}

resource "azurerm_availability_set" "test" {
  name                = "acceptanceTestAvailabilitySet1"

  tags {
    environment = "Production"
  }
}
```

### Debug Output
```
[ERROR] root: eval: *terraform.EvalApplyPost, err: 1 error(s) occurred: azurerm_availability_set.vm: compute.AvailabilitySetsClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidParameter" Message="The specified fault domain count 3 must fall in the range 1 to 2."
```

### Expected Behavior
The azure plugin should be smart enough to figure out what the fault domain count is. Or an API endpoint should be available for the plugin to determine what the proper `platform_update_domain_count` and `platform_fault_domain_count` should be. 

### Actual Behavior
Your code failed, because you didn't know that Azure has limitations, that are semi-poorly referenced here:
https://github.com/MicrosoftDocs/azure-docs/blob/master/includes/managed-disks-common-fault-domain-region-list.md

### Steps to Reproduce
1. `terraform apply` (using one of the regions listed above with only 2 Fault Domain

tombuildsstuff:
Fixes #944

### DIFF
--- a/website/docs/r/availability_set.html.markdown
+++ b/website/docs/r/availability_set.html.markdown
@@ -42,7 +42,11 @@ The following arguments are supported:
 
 * `platform_update_domain_count` - (Optional) Specifies the number of update domains that are used. Defaults to 5.
 
+~> **NOTE:** The number of Update Domains varies depending on which Azure Region you're using - [a list can be found here](https://github.com/MicrosoftDocs/azure-docs/blob/04809e3072e0061ca436b0986cf2222c472651ba/includes/managed-disks-common-fault-domain-region-list.md).
+
 * `platform_fault_domain_count` - (Optional) Specifies the number of fault domains that are used. Defaults to 3.
+
+~> **NOTE:** The number of Fault Domains varies depending on which Azure Region you're using - [a list can be found here](https://github.com/MicrosoftDocs/azure-docs/blob/04809e3072e0061ca436b0986cf2222c472651ba/includes/managed-disks-common-fault-domain-region-list.md).
 
 * `managed` - (Optional) Specifies whether the availability set is managed or not. Possible values are `true` (to specify aligned) or `false` (to specify classic). Default is `false`.
 
--- a/website/docs/r/availability_set.html.markdown
+++ b/website/docs/r/availability_set.html.markdown
@@ -42,11 +42,11 @@ The following arguments are supported:
 
 * `platform_update_domain_count` - (Optional) Specifies the number of update domains that are used. Defaults to 5.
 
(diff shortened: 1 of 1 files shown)


website/docs/r/availability_set.html.markdown
```diff
@@ -42,8 +42,12 @@ The following arguments are supported:
 
 * `platform_update_domain_count` - (Optional) Specifies the number of update domains that are used. Defaults to 5.
 
+~> **NOTE:** The number of Update Domains varies depending on which Azure Region you're using - [a list can be found here](https://github.com/MicrosoftDocs/azure-docs/blob/04809e3072e0061ca436b0986cf2222c472651ba/includes/managed-disks-common-fault-domain-region-list.md).
```
FrenchBen: Can we refer to the master branch instead?
https://github.com/MicrosoftDocs/azure-docs/blob/master/includes/managed-disks-common-fault-domain-region-list.md

website/docs/r/availability_set.html.markdown
```diff
@@ -42,8 +42,12 @@ The following arguments are supported:
...
+
 * `platform_fault_domain_count` - (Optional) Specifies the number of fault domains that are used. Defaults to 3.
 
+~> **NOTE:** The number of Fault Domains varies depending on which Azure Region you're using - [a list can be found here](https://github.com/MicrosoftDocs/azure-docs/blob/04809e3072e0061ca436b0986cf2222c472651ba/includes/managed-disks-common-fault-domain-region-list.md).
```
FrenchBen: same here...

website/docs/r/availability_set.html.markdown
tombuildsstuff: So the reason not to do that is in case the file gets renamed / removed, but yeah as you’ve mentioned this may make sense too. Given it’s available in the git history, we could probably update this however

FrenchBen: If the file happens to disappear, then an update would make sense. Using master IMHO prevents this doc from being updated as often, when new regions appear or get supported.

tombuildsstuff: fixed :)

Issue #955, opened 2018-03-08
Create virtual network rule for Azure SQL service


bostonmoto:
Hi there,

A new feature just came out I believe a couple of months ago in azure to create a virtual network rule which allows subets in a virtual network to communicate with Azure SQL.

https://docs.microsoft.com/en-us/azure/sql-database/sql-database-vnet-service-endpoint-rule-overview

You can also add a "virtual network" to the SQL server through the portal or powershell.

https://docs.microsoft.com/en-us/azure/sql-database/sql-database-vnet-service-endpoint-rule-powershell

I do not see any support yet in terraform for adding a vnet rule to an azure SQL server - Can we add support for this by any chance? Right now I have to manually add the virtual network rule from the console to allow my servers to talk to the Azure SQL in a secure manner. Is this something we can get added to terraform? Please let me know if any other details needed.

cc: @tombuildsstuff 

Thanks!

tombuildsstuff:
hey @bostonmoto 

Thanks for opening this issue :)

Just to let you know that support for SQL Virtual Network Rules was added in #978 which [has been released as part of v1.3.1](https://github.com/terraform-providers/terraform-provider-azurerm/blob/v1.3.1/CHANGELOG.md) and is now available [on the Terraform Website](https://www.terraform.io/docs/providers/azurerm/r/sql_virtual_network_rule.html) - as such I'm going to close this issue :)

Thanks!

PR #978, opened 2018-03-14
New Resource: `azurerm_sql_virtual_network_rule `

vlouwagie:
### Summary
New resource that adds a SQL server to a subnet. ([Azure Docs](https://docs.microsoft.com/en-us/azure/templates/microsoft.sql/servers/virtualnetworkrules), [relevant Azure GO SDK client](https://github.com/Azure/azure-sdk-for-go/blob/master/services/sql/mgmt/2015-05-01-preview/sql/virtualnetworkrules.go))

### Issue
#955 

### TODO
(From the main terraform's [contrib guide](https://github.com/hashicorp/terraform/blob/master/.github/CONTRIBUTING.md#new-resource))

- [x] **Minimal LOC**
- [x]  **Resource**
    - [x] Read
    - [x] Create
    - [x] Update
    - [x] Delete
- [x]  **Resource Acceptance Test** - SQL Endpoint Validation Enabled and Disabled
- [x]  **Resource Acceptance Test** - Basic Provisioning and Teardown
- [x]  **Resource Acceptance Test** - Multiple Subnets
- [x] **Resource Acceptance Test** - Update Subnets (without forcing new resource)
- [x] **Validation Function Test** - (Barely) Valid Rule Names
- [x] **Validation Function Test** - Invalid Rule Names
- [x] **Import Acceptance Test**
- [x]  **Documentation**
- [x]  **Well-Formed Code** - used other clients as a reference, let me know if it isn't well-formed! :) 
- [x] Requested Changes
    - [x] [Update to New Registration Syntax](https://github.com/terraform-providers/terraform-provider-azurerm/pull/978/files/8a15347e01666f14f84ac4b2ef850249f8b71eb9#diff-848235249455ee48aaf3ea5d431960ae)
    - [x] [Rule Name Validation](https://github.com/terraform-providers/terraform-provider-azurerm/pull/978/files/8a15347e01666f14f84ac4b2ef850249f8b71eb9#diff-79d80ecbfd98132cca850f3fe6f7b615)
    - [x] [Create test case for updating subnets](https://github.com/terraform-providers/terraform-provider-azurerm/pull/978/files/8a15347e01666f14f84ac4b2ef850249f8b71eb9#diff-79d80ecbfd98132cca850f3fe6f7b615)
    - [x] [Check for nil in resp object](https://github.com/terraform-providers/terraform-provider-azurerm/pull/978/files/8a15347e01666f14f84ac4b2ef850249f8b71eb9#diff-79d80ecbfd98132cca850f3fe6f7b615)
    - [x] [Change virtual_network_subnet_id to subnet_id](https://github.com/terraform-providers/terraform-provider-azurerm/pull/978/files/8a15347e01666f14f84ac4b2ef850249f8b71eb9#diff-79d80ecbfd98132cca850f3fe6f7b615)
    - [x] [Delete: Check for 404 from future objects](https://github.com/terraform-providers/terraform-provider-azurerm/pull/978/files/8a15347e01666f14f84ac4b2ef850249f8b71eb9#diff-79d80ecbfd98132cca850f3fe6f7b615)
    - [x] [Remove Commented-Out Code](https://github.com/terraform-providers/terraform-provider-azurerm/pull/978/files/8a15347e01666f14f84ac4b2ef850249f8b71eb9#diff-25353a6f65c03e21e0123f16c7857755)
    - [x] [Fix Spacing in Test Case Configurations](https://github.com/terraform-providers/terraform-provider-azurerm/pull/978/files/8a15347e01666f14f84ac4b2ef850249f8b71eb9#diff-25353a6f65c03e21e0123f16c7857755)
    - [x] Add Check/Waiter function for Create

### DIFF
--- a/azurerm/config.go
+++ b/azurerm/config.go
@@ -598,10 +598,7 @@ func (c *ArmClient) registerDatabases(endpoint, subscriptionId string, auth auto
 	c.sqlServerAzureADAdministratorsClient = sqlADClient
 
 	sqlVNRClient := sql.NewVirtualNetworkRulesClientWithBaseURI(endpoint, subscriptionId)
-	setUserAgent(&sqlVNRClient.Client)
-	sqlVNRClient.Authorizer = auth
-	sqlVNRClient.Sender = sender
-	sqlVNRClient.SkipResourceProviderRegistration = c.skipProviderRegistration
+	c.configureClient(&sqlVNRClient.Client, auth)
 	c.sqlVirtualNetworkRulesClient = sqlVNRClient
 }
 
--- a/azurerm/config.go
+++ b/azurerm/config.go
@@ -128,6 +128,7 @@ type ArmClient struct {
 	sqlFirewallRulesClient               sql.FirewallRulesClient
 	sqlServersClient                     sql.ServersClient
 	sqlServerAzureADAdministratorsClient sql.ServerAzureADAdministratorsClient
+	sqlVirtualNetworkRulesClient         sql.VirtualNetworkRulesClient
 
 	// KeyVault
 	keyVaultClient           keyvault.VaultsClient
@@ -595,6 +596,13 @@ func (c *ArmClient) registerDatabases(endpoint, subscriptionId string, auth auto
 	sqlADClient.Sender = sender
 	sqlADClient.SkipResourceProviderRegistration = c.skipProviderRegistration
 	c.sqlServerAzureADAdministratorsClient = sqlADClient
+
+	sqlVNRClient := sql.NewVirtualNetworkRulesClientWithBaseURI(endpoint, subscriptionId)
+	setUserAgent(&sqlVNRClient.Client)
+	sqlVNRClient.Authorizer = auth
+	sqlVNRClient.Sender = sender
+	sqlVNRClient.SkipResourceProviderRegistration = c.skipProviderRegistration
+	c.sqlVirtualNetworkRulesClient = sqlVNRClient
 }
 
 func (c *ArmClient) registerDNSClients(endpoint, subscriptionId string, auth autorest.Authorizer, sender autorest.Sender) {
--- a/azurerm/resource_arm_sql_virtual_network_rule.go
+++ b/azurerm/resource_arm_sql_virtual_network_rule.go
@@ -35,7 +35,7 @@ func resourceArmSqlVirtualNetworkRule() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"virtual_network_subnet_id": {
+			"subnet_id": {
 				Type:     schema.TypeString,
 				Required: true,
 			},
@@ -56,7 +56,7 @@ func resourceArmSqlVirtualNetworkRuleCreateUpdate(d *schema.ResourceData, meta i
 	name := d.Get("name").(string)
 	serverName := d.Get("server_name").(string)
 	resourceGroup := d.Get("resource_group_name").(string)
-	virtualNetworkSubnetId := d.Get("virtual_network_subnet_id").(string)
+	virtualNetworkSubnetId := d.Get("subnet_id").(string)
 	ignoreMissingVnetServiceEndpoint := d.Get("ignore_missing_vnet_service_endpoint").(bool)
 
 	parameters := sql.VirtualNetworkRule{
@@ -108,7 +108,7 @@ func resourceArmSqlVirtualNetworkRuleRead(d *schema.ResourceData, meta interface
 	d.Set("name", resp.Name)
 	d.Set("resource_group_name", resourceGroup)
 	d.Set("server_name", serverName)
-	d.Set("virtual_network_subnet_id", resp.VirtualNetworkSubnetID)
+	d.Set("subnet_id", resp.VirtualNetworkSubnetID)
 	d.Set("ignore_missing_vnet_service_endpoint", resp.IgnoreMissingVnetServiceEndpoint)
 
 	return nil
--- a/azurerm/resource_arm_sql_virtual_network_rule.go
+++ b/azurerm/resource_arm_sql_virtual_network_rule.go
@@ -4,12 +4,10 @@ import (
 	"fmt"
 	"log"
 	"regexp"
-	//"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/sql/mgmt/2015-05-01-preview/sql"
 	"github.com/hashicorp/terraform/helper/schema"
-	//"github.com/hashicorp/terraform/helper/validation"
-
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/response"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
@@ -133,17 +131,33 @@ func resourceArmSqlVirtualNetworkRuleDelete(d *schema.ResourceData, meta interfa
 
 	future, err := client.Delete(ctx, resourceGroup, serverName, name)
 	if err != nil {
+		//If the error is that the resource we want to delete does not exist in the first
+		//place (404), then just return with no error.
+		if response.WasNotFound(future.Response()) {
+			return nil
+		}
+
 		return fmt.Errorf("Error deleting SQL Virtual Network Rule: %+v", err)
 	}
 
 	err = future.WaitForCompletion(ctx, client.Client)
 	if err != nil {
+
+		//Same deal as before. Just in case.
+		if response.WasNotFound(future.Response()) {
+			return nil
+		}
+
 		return err
 	}
 
 	return nil
 }
 
+/*
+	This function checks the format of the SQL Virtual Network Rule Name to make sure that
+	it does not contain any potentially invalid values.
+*/
 func validateSqlVirtualNetworkRuleName(v interface{}, k string) (ws []string, errors []error) {
 	value := v.(string)
 
--- a/azurerm/resource_arm_sql_virtual_network_rule.go
+++ b/azurerm/resource_arm_sql_virtual_network_rule.go
@@ -89,7 +89,7 @@ func resourceArmSqlVirtualNetworkRuleCreateUpdate(d *schema.ResourceData, meta i
 		Refresh:                   sqlVirtualNetworkStateStatusCodeRefreshFunc(ctx, client, resourceGroup, serverName, name),
 		Timeout:                   10 * time.Minute,
 		MinTimeout:                1 * time.Minute,
-		ContinuousTargetOccurence: 2,
+		ContinuousTargetOccurence: 5,
 	}
 
 	if _, err := stateConf.WaitForState(); err != nil {
--- a/azurerm/resource_arm_sql_virtual_network_rule.go
+++ b/azurerm/resource_arm_sql_virtual_network_rule.go
@@ -110,8 +110,11 @@ func resourceArmSqlVirtualNetworkRuleRead(d *schema.ResourceData, meta interface
 	d.Set("name", resp.Name)
 	d.Set("resource_group_name", resourceGroup)
 	d.Set("server_name", serverName)
-	d.Set("subnet_id", resp.VirtualNetworkSubnetID)
-	d.Set("ignore_missing_vnet_service_endpoint", resp.IgnoreMissingVnetServiceEndpoint)
+
+	if props := resp.VirtualNetworkRuleProperties; props != nil {
+		d.Set("subnet_id", props.VirtualNetworkSubnetID)
+		d.Set("ignore_missing_vnet_service_endpoint", props.IgnoreMissingVnetServiceEndpoint)
+	}
 
 	return nil
 }
--- a/azurerm/resource_arm_sql_virtual_network_rule.go
+++ b/azurerm/resource_arm_sql_virtual_network_rule.go
@@ -47,6 +47,11 @@ func resourceArmSqlVirtualNetworkRule() *schema.Resource {
 				Optional: true,
 				Default:  false, //When not provided, Azure defaults to false
 			},
+
+			"endpoint_state": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -114,6 +119,7 @@ func resourceArmSqlVirtualNetworkRuleRead(d *schema.ResourceData, meta interface
 	if props := resp.VirtualNetworkRuleProperties; props != nil {
 		d.Set("subnet_id", props.VirtualNetworkSubnetID)
 		d.Set("ignore_missing_vnet_service_endpoint", props.IgnoreMissingVnetServiceEndpoint)
+		d.Set("endpoint_state", props.State)
 	}
 
 	return nil
--- a/azurerm/resource_arm_sql_virtual_network_rule.go
+++ b/azurerm/resource_arm_sql_virtual_network_rule.go
@@ -9,12 +9,12 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
-func resourceArmSqlVnetRule() *schema.Resource {
+func resourceArmSqlVirtualNetworkRule() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceArmSqlVnetRuleCreateUpdate,
-		Read:   resourceArmSqlVnetRuleRead,
-		Update: resourceArmSqlVnetRuleCreateUpdate,
-		Delete: resourceArmSqlVnetRuleDelete,
+		Create: resourceArmSqlVirtualNetworkRuleCreateUpdate,
+		Read:   resourceArmSqlVirtualNetworkRuleRead,
+		Update: resourceArmSqlVirtualNetworkRuleCreateUpdate,
+		Delete: resourceArmSqlVirtualNetworkRuleDelete,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
@@ -49,8 +49,8 @@ func resourceArmSqlVnetRule() *schema.Resource {
 	}
 }
 
-func resourceArmSqlVnetRuleCreateUpdate(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*ArmClient).sqlVnetRulesClient
+func resourceArmSqlVirtualNetworkRuleCreateUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).sqlVirtualNetworkRulesClient
 	ctx := meta.(*ArmClient).StopContext
 
 	name := d.Get("name").(string)
@@ -59,10 +59,10 @@ func resourceArmSqlVnetRuleCreateUpdate(d *schema.ResourceData, meta interface{}
 	virtualNetworkSubnetId := d.Get("virtual_network_subnet_id").(string)
 	ignoreMissingVnetServiceEndpoint := d.Get("ignore_missing_vnet_service_endpoint").(bool)
 
-	parameters := sql.VnetRule{
-		VnetRuleProperties: &sql.VnetRuleProperties{
-			virtualNetworkSubnetId: utils.String(virtualNetworkSubnetId),
-			ignoreMissingVnetServiceEndpoint: utils.String(ignoreMissingVnetServiceEndpoint),
+	parameters := sql.VirtualNetworkRule{
+		VirtualNetworkRuleProperties: &sql.VirtualNetworkRuleProperties{
+			VirtualNetworkSubnetID:           utils.String(virtualNetworkSubnetId),
+			IgnoreMissingVnetServiceEndpoint: utils.Bool(ignoreMissingVnetServiceEndpoint),
 		},
 	}
 
@@ -78,11 +78,11 @@ func resourceArmSqlVnetRuleCreateUpdate(d *schema.ResourceData, meta interface{}
 
 	d.SetId(*resp.ID)
 
-	return resourceArmSqlVnetRuleRead(d, meta)
+	return resourceArmSqlVirtualNetworkRuleRead(d, meta)
 }
 
-func resourceArmSqlVnetRuleRead(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*ArmClient).sqlVnetRulesClient
+func resourceArmSqlVirtualNetworkRuleRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).sqlVirtualNetworkRulesClient
 	ctx := meta.(*ArmClient).StopContext
 
 	id, err := parseAzureResourceID(d.Id())
@@ -108,14 +108,14 @@ func resourceArmSqlVnetRuleRead(d *schema.ResourceData, meta interface{}) error 
 	d.Set("name", resp.Name)
 	d.Set("resource_group_name", resourceGroup)
 	d.Set("server_name", serverName)
-	d.Set("virtual_network_subnet_id", resp.virtualNetworkSubnetId)
-	d.Set("ignore_missing_vnet_service_endpoint", resp.ignoreMissingVnetServiceEndpoint)
+	d.Set("virtual_network_subnet_id", resp.VirtualNetworkSubnetID)
+	d.Set("ignore_missing_vnet_service_endpoint", resp.IgnoreMissingVnetServiceEndpoint)
 
 	return nil
 }
 
-func resourceArmSqlVnetRuleDelete(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*ArmClient).sqlVnetRulesClient
+func resourceArmSqlVirtualNetworkRuleDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).sqlVirtualNetworkRulesClient
 	ctx := meta.(*ArmClient).StopContext
 
 	id, err := parseAzureResourceID(d.Id())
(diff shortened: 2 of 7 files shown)


tombuildsstuff:
hey @vlouwagie 

Thanks for this PR :)

I've taken a look through and this is looking pretty good; however I'm wondering is it possible to have a SQL Server connected to multiple Virtual Networks? If it's not - I'm wondering if it'd make sense to add a `virtual_network` block to the existing `azurerm_sql_server` resource and limit it to a single connection (so that it's possible to create/update the connection within that resource) - what do you think?

> Questions
> > 1. Should I include the import resource with this pull request, or as a separate feature/pull request?

If you can add it to this PR that'd be great - but I'd suggest holding off for the moment until we've worked out if this'd be better located within the `azurerm_sql_server` resource

Thanks!

vlouwagie:
@tombuildsstuff Yep, it is [possible](https://i.imgur.com/5kxqar5.png) for a single SQL server connected to one or more subnets from one or more virtual networks. This resource has a lot in common with the SQL firewall rule resource, so I originally modelled it after that.


### Template for multiple rules

```
    name = "acctestRG_1"
    location = "westus"
}
#Vnet 1
resource "azurerm_virtual_network" "vnet1" {
  name                = "acctestvnet1"
  address_space       = ["10.7.29.0/24"]
}
#Vnet 2
resource "azurerm_virtual_network" "vnet2" {
  name                = "acctestvnet2"
  address_space       = ["10.1.29.0/29"]
}
#Subnet 1
resource "azurerm_subnet" "vnet1subnet1" {
  name = "acctestsubnet1"
  virtual_network_name = "${azurerm_virtual_network.vnet1.name}"
  address_prefix = "10.7.29.0/29"
  service_endpoints = ["Microsoft.Sql"]
}
#Subnet 2 (same vnet)
resource "azurerm_subnet" "vnet1subnet2" {
  name = "acctestsubnet2"
  virtual_network_name = "${azurerm_virtual_network.vnet1.name}"
  address_prefix = "10.7.29.128/29"
  service_endpoints = ["Microsoft.Sql"]
}
#Subnet 3 (different vnet)
resource "azurerm_subnet" "vnet2subnet1" {
  name = "acctestsubnet3"
  virtual_network_name = "${azurerm_virtual_network.vnet2.name}"
  address_prefix = "10.1.29.0/29"
  service_endpoints = ["Microsoft.Sql"]
}
#Sql Server
    name = "acctestsqlserver1"
    location = "${azurerm_resource_group.test.location}"
    version = "12.0"
    administrator_login = "missadmin"
    administrator_login_password = "${md5(1)}!"
}
#Rules
resource "azurerm_sql_virtual_network_rule" "test" {
    name = "acctestsqlvnetrule1"
    server_name = "${azurerm_sql_server.test.name}"
    virtual_network_subnet_id = "${azurerm_subnet.vnet1subnet1.id}"
    ignore_missing_vnet_service_endpoint = true
}
resource "azurerm_sql_virtual_network_rule" "test2" {
    name = "acctestsqlvnetrule2"
    server_name = "${azurerm_sql_server.test.name}"
    virtual_network_subnet_id = "${azurerm_subnet.vnet1subnet2.id}"
    ignore_missing_vnet_service_endpoint = false
}
resource "azurerm_sql_virtual_network_rule" "test3" {
    name = "acctestsqlvnetrule3"
    server_name = "${azurerm_sql_server.test.name}"
    virtual_network_subnet_id = "${azurerm_subnet.vnet2subnet1.id}"
    ignore_missing_vnet_service_endpoint = false
}
```

tombuildsstuff:
> @tombuildsstuff Yep, it is possible for a single SQL server to be connected to one or more subnets from one or more virtual networks. This resource has a lot in common with the SQL firewall rule resource, so I originally modelled it after that.

@vlouwagie cool, thanks for confirming that. Could we add a test to cover that scenario, perhaps using the configuration you've posted above?

Thanks!

vlouwagie:
> Could we add a test to cover that scenario, perhaps using the configuration you've posted above?

@tombuildsstuff Sure! I will add a task for that in the pull request description.

azurerm/config.go
```diff
@@ -595,6 +596,13 @@ func (c *ArmClient) registerDatabases(endpoint, subscriptionId string, auth auto
 	sqlADClient.Sender = sender
 	sqlADClient.SkipResourceProviderRegistration = c.skipProviderRegistration
 	c.sqlServerAzureADAdministratorsClient = sqlADClient
+
+	sqlVNRClient := sql.NewVirtualNetworkRulesClientWithBaseURI(endpoint, subscriptionId)
+	setUserAgent(&sqlVNRClient.Client)
+	sqlVNRClient.Authorizer = auth
+	sqlVNRClient.Sender = sender
+	sqlVNRClient.SkipResourceProviderRegistration = c.skipProviderRegistration
+	c.sqlVirtualNetworkRulesClient = sqlVNRClient
```
tombuildsstuff: _minor_ can we switch this over to using the new registration syntax? e.g. `c.configureClient(&sqlVNRClient.Client, auth)

azurerm/resource_arm_sql_virtual_network_rule.go
```diff
@@ -0,0 +1,141 @@
...
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				//TODO: Validation (invalid if non-numeric characters)
```
tombuildsstuff: 👍

azurerm/resource_arm_sql_virtual_network_rule.go
```diff
@@ -0,0 +1,141 @@
...
+
+			"server_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"virtual_network_subnet_id": {
+				Type:     schema.TypeString,
+				Required: true,
```
tombuildsstuff: should this be `ForceNew` or is it possible to migrate between Virtual Networks? Can we add a test for this if that's the case

azurerm/resource_arm_sql_virtual_network_rule.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		return fmt.Errorf("Error reading SQL Virtual Network Rule: %+v", err)
+	}
+
+	d.Set("name", resp.Name)
+	d.Set("resource_group_name", resourceGroup)
+	d.Set("server_name", serverName)
+	d.Set("virtual_network_subnet_id", resp.VirtualNetworkSubnetID)
+	d.Set("ignore_missing_vnet_service_endpoint", resp.IgnoreMissingVnetServiceEndpoint)
```
tombuildsstuff: for these two fields can we use the field in `resp.[XXX]Properties` and nil-check it (e.g. to handle bad API responses); since that's what's recommended by Azure e.g.

```
if props := resp.[XXX]Properties; props != nil {
  d.Set("virtual_network_subnet_id", props.VirtualNetworkSubnetID)
  d.Set("ignore_missing_vnet_service_endpoint", props.IgnoreMissingVnetServiceEndpoint)
}
```

azurerm/resource_arm_sql_virtual_network_rule.go
```diff
@@ -0,0 +1,141 @@
...
+
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"server_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"virtual_network_subnet_id": {
```
tombuildsstuff: 🤔 do you think it'd be worth making this `subnet_id` to be more consistent with the other Terraform resources?

azurerm/resource_arm_sql_virtual_network_rule.go
```diff
@@ -0,0 +1,141 @@
...
+	name := id.Path["virtualNetworkRules"]
+
+	future, err := client.Delete(ctx, resourceGroup, serverName, name)
+	if err != nil {
+		return fmt.Errorf("Error deleting SQL Virtual Network Rule: %+v", err)
+	}
+
+	err = future.WaitForCompletion(ctx, client.Client)
+	if err != nil {
+		return err
```
tombuildsstuff: checking the SDK - it appears that 404 will return an error here; do we want to check for that? e.g.

```
if err != nil {
  if response.WasNotFound(future.Response()) {
    return nil
  }
  return err
}
```

azurerm/resource_arm_sql_virtual_network_rule.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	resourceGroup := id.ResourceGroup
+	serverName := id.Path["servers"]
+	name := id.Path["virtualNetworkRules"]
+
+	future, err := client.Delete(ctx, resourceGroup, serverName, name)
+	if err != nil {
```
tombuildsstuff: checking the SDK - it appears that 404 will return an error here; do we want to check for that? e.g.

```
if err != nil {
  if response.WasNotFound(future.Response()) {
    return nil
  }
  return err
}
```

azurerm/resource_arm_sql_virtual_network_rule_test.go
```diff
@@ -0,0 +1,473 @@
...
+	})
+}
+
+/*
+	--Testing for Failure--
+	Test if we are able to create a vnet with out the SQL endpoint, but SQL rule
+	is still applied since the endpoint validation will be set to false.
+*/
+func TestAccAzureRMSqlVirtualNetworkRule_IgnoreEndpointInvalid(t *testing.T) {
+	//resourceName := "azurerm_sql_virtual_network_rule.test"
```
tombuildsstuff: can we remove this?

azurerm/resource_arm_sql_virtual_network_rule_test.go
```diff
@@ -0,0 +1,473 @@
...
+    administrator_login_password = "${md5(%d)}!"
+}
+resource "azurerm_sql_virtual_network_rule" "test" {
+    name = "acctestsqlvnetrule%d"
+    resource_group_name = "${azurerm_resource_group.test.name}"
+    server_name = "${azurerm_sql_server.test.name}"
+    virtual_network_subnet_id = "${azurerm_subnet.test.id}"
+    ignore_missing_vnet_service_endpoint = false
+}
+`, rInt, location, rInt, rInt, rInt, rInt, rInt)
```
tombuildsstuff: _minor_ could we make the spacing consistent here with 2 spaces for indents?

azurerm/resource_arm_sql_virtual_network_rule_test.go
```diff
@@ -0,0 +1,473 @@
...
+    administrator_login = "missadmin"
+    administrator_login_password = "${md5(%d)}!"
+}
+resource "azurerm_sql_virtual_network_rule" "test" {
+    name = "acctestsqlvnetrule%d"
+    resource_group_name = "${azurerm_resource_group.test.name}"
+    server_name = "${azurerm_sql_server.test.name}"
+    virtual_network_subnet_id = "${azurerm_subnet.test.id}"
+    ignore_missing_vnet_service_endpoint = true
+}
```
tombuildsstuff: _minor_ could we make the spacing consistent here with 2 spaces for indents?

azurerm/resource_arm_sql_virtual_network_rule_test.go
```diff
@@ -0,0 +1,473 @@
...
```
tombuildsstuff: _minor_ could we make the spacing consistent here with 2 spaces for indents?

azurerm/resource_arm_sql_virtual_network_rule_test.go
```diff
@@ -0,0 +1,473 @@
...
+    administrator_login = "missadmin"
+    administrator_login_password = "${md5(%d)}!"
+}
+resource "azurerm_sql_virtual_network_rule" "test" {
+    name = "acctestsqlvnetrule%d"
+    resource_group_name = "${azurerm_resource_group.test.name}"
+    server_name = "${azurerm_sql_server.test.name}"
+    virtual_network_subnet_id = "${azurerm_subnet.test.id}"
+    ignore_missing_vnet_service_endpoint = false
+}
```
tombuildsstuff: _minor_ could we make the spacing consistent here with 2 spaces for indents?

azurerm/resource_arm_sql_virtual_network_rule_test.go
```diff
@@ -0,0 +1,473 @@
...
+    virtual_network_subnet_id = "${azurerm_subnet.vnet1_subnet2.id}"
+    ignore_missing_vnet_service_endpoint = false
+}
+resource "azurerm_sql_virtual_network_rule" "rule3" {
+    name = "acctestsqlvnetrule3%d"
+    resource_group_name = "${azurerm_resource_group.test.name}"
+    server_name = "${azurerm_sql_server.test.name}"
+    virtual_network_subnet_id = "${azurerm_subnet.vnet2_subnet1.id}"
+    ignore_missing_vnet_service_endpoint = false
+}
```
tombuildsstuff: _minor_ could we make the spacing consistent here with 2 spaces for indents?

vlouwagie:
Thanks for the feedback! I will investigate each one and get back to you. Also added the requested changes to the pull request description. 😄

azurerm/resource_arm_sql_virtual_network_rule_test.go
vlouwagie: 👍

azurerm/resource_arm_sql_virtual_network_rule_test.go
vlouwagie: 👍

azurerm/resource_arm_sql_virtual_network_rule_test.go
vlouwagie: 👍

azurerm/resource_arm_sql_virtual_network_rule_test.go
vlouwagie: 👍

azurerm/resource_arm_sql_virtual_network_rule_test.go
vlouwagie: 👍

azurerm/resource_arm_sql_virtual_network_rule_test.go
vlouwagie: 👍

azurerm/resource_arm_sql_virtual_network_rule.go
vlouwagie: @tombuildsstuff Yes, that definitely makes sense. 

On a similar note, what is your opinion on `ignore_missing_vnet_service_endpoint`? Do you think it would be better if it was shorted a bit as well to `ignore_missing_service_endpoint` or even `ignore_missing_endpoint`?

azurerm/config.go
vlouwagie: 👍

azurerm/resource_arm_sql_virtual_network_rule.go
tombuildsstuff: given the description ([in the API Docs](https://docs.microsoft.com/en-us/rest/api/sql/virtualnetworkrules/createorupdate)): `Create firewall rule before the virtual network has vnet service endpoint enabled.` - I guess leaving it as `ignore_missing_vnet_service_endpoint ` would probably make the most sense? Whilst I don't feel that name's particularly descriptive, it's how the Azure API's chosen to label it, so it probably makes sense to match it incase it does extra stuff in the future

azurerm/resource_arm_sql_virtual_network_rule.go
vlouwagie: 👍

azurerm/resource_arm_sql_virtual_network_rule.go
vlouwagie: @tombuildsstuff If the `client.Delete` function errors while returning, the `future` variable will still contain a valid `future` then, correct? ([SDK Delete](https://github.com/Azure/azure-sdk-for-go/blob/master/services/sql/mgmt/2015-05-01-preview/sql/virtualnetworkrules.go#L130))

azurerm/resource_arm_sql_virtual_network_rule.go
vlouwagie: 👍

azurerm/resource_arm_sql_virtual_network_rule.go
vlouwagie: It does indeed look like you can change the subnets of a rule, and it does apply correctly. I will make a task to create a test case for that. 😄 

On a side note, I also tested migrating from one SQL server to another without `ForceNew` (just in case) , and it creates a rule on the 2nd target SQL server, but does not delete it on the first server, so the `ForceNew` attribute is needed on the `server_name`.

azurerm/resource_arm_sql_virtual_network_rule_test.go
```diff
@@ -0,0 +1,658 @@
...
+
+func TestAccAzureRMSqlVirtualNetworkRule_switchSubnets(t *testing.T) {
+	resourceName := "azurerm_sql_virtual_network_rule.test"
+	ri := acctest.RandInt()
+
+	preConfig := testAccAzureRMSqlVirtualNetworkRule_subnetSwitchPre(ri, testLocation())
+	postConfig := testAccAzureRMSqlVirtualNetworkRule_subnetSwitchPost(ri, testLocation())
+
+	// Create regex strings that will ensure that one subnet name exists, but not the other
+	preConfigRegex := regexp.MustCompile(fmt.Sprintf("(subnet1%d)$|(subnet[^2]%d)$", ri, ri))  //subnet 1 but not 2
```
vlouwagie: @tombuildsstuff I ended up making a [regex match test](https://regexr.com/3mcse) that checks if the subnet ID contains one subnet name but not the other in the `preConfig` setup, and then check for the inverse on the `postConfig` setup. I wasn't sure if there was a better way to check if a computed item was a different value from the previous step.

It still works, but let me know if there actually is a better way. 😄

azurerm/resource_arm_sql_virtual_network_rule.go
tombuildsstuff: it'll contain the HTTP response, but not necessarily a valid future (since a 500 for example won't contain the headers needed for a future, at least from my understanding)

azurerm/resource_arm_sql_virtual_network_rule.go
```diff
@@ -0,0 +1,206 @@
...
+			return nil
+		}
+
+		return fmt.Errorf("Error deleting SQL Virtual Network Rule: %+v", err)
+	}
+
+	err = future.WaitForCompletion(ctx, client.Client)
+	if err != nil {
+
+		//Same deal as before. Just in case.
```
tombuildsstuff: to clarify this, this is the main code path - as the deletion of the SQL Network Rule's should be a long running operation (LRO); afaik :)

azurerm/resource_arm_sql_virtual_network_rule.go
vlouwagie: 👍

azurerm/resource_arm_sql_virtual_network_rule.go
vlouwagie: ✅

azurerm/resource_arm_sql_virtual_network_rule.go
vlouwagie: ✅

azurerm/resource_arm_sql_virtual_network_rule.go
vlouwagie: @tombuildsstuff Could you elaborate a bit on this comment?

azurerm/resource_arm_sql_virtual_network_rule.go
vlouwagie: ✅

vlouwagie:
I added the endpoint state as a computed fields since that is also available from the API. I also added the documentation. I will be removing the WIP as all tasks (as of now) are complete and ready to be reviewed! 😄

azurerm/resource_arm_sql_virtual_network_rule_test.go
tombuildsstuff: that works, the other options would be building up the ID by hand (e.g. `/subscriptions/0000../foo/bar/subnets/mySubnet`) or just check the suffix is `mySubnet` - but this LGTM :)

vlouwagie: > or just check the suffix is `mySubnet`

Sounds good. This is basically what I'm doing anyways, but with a little extra!

tombuildsstuff:
hey @vlouwagie 

Thanks for pushing the updates - this now looks good to me. In running the tests I've noticed there appears to be an eventual consistency issue when creating this, where the API returns this as created but it doesn't exist until a few seconds later, which is causing the tests to fail intermittently:

```
------- Stdout: -------
=== RUN   TestAccAzureRMSqlVirtualNetworkRule_multipleSubnets
--- FAIL: TestAccAzureRMSqlVirtualNetworkRule_multipleSubnets (162.67s)
    testing.go:513: Step 0 error: Error applying: 1 error(s) occurred:
        
        * azurerm_sql_virtual_network_rule.rule3: 1 error(s) occurred:
        
        * azurerm_sql_virtual_network_rule.rule3: Error retrieving SQL Virtual Network Rule: sql.VirtualNetworkRulesClient#Get: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="ResourceNotFound" Message="The requested resource of type 'Microsoft.Sql/servers/virtualNetworkRules' with name 'acctestsqlvnetrule34331446545651011683' was not found."
FAIL
```

We should be able to add a check/waiter function once the Network Rule's been created before retrieving it to ensure that this actually exists - something like this (modified from the IoTHub resource to add the `ContinuousTargetOccurence` attribute, which ensures the successful state is returned N times before marking it as successful):

```
func waitForIotHubToBeDeleted(ctx context.Context, client devices.IotHubResourceClient, resourceGroup, name string) error {
	stateConf := &resource.StateChangeConf{
		Pending: []string{"200"},
		Target:  []string{"404"},
		Refresh: iothubStateStatusCodeRefreshFunc(ctx, client, resourceGroup, name),
		Timeout: 40 * time.Minute,
		ContinuousTargetOccurence: 5,
	}
	if _, err := stateConf.WaitForState(); err != nil {
		return fmt.Errorf("Error waiting for IotHub (%q in Resource Group %q) to be deleted: %+v", name, resourceGroup, err)
	}

	return nil
}
```

If we can add this check function that should solve the eventual consistency issue and then this should be good to merge :)

Thanks!

vlouwagie:
@tombuildsstuff I've been able to fail the tests myself as well. I think I might go one step further with your example, and check the provisioning state of the create/updated rule, similar to how the `resource_arm_redis_cache` does it, since it's also checking state on create, rather than delete.

**resource_arm_redis_cache.go** (Lines [252](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_redis_cache.go#L252), [463](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_redis_cache.go#L463))
```go
log.Printf("[DEBUG] Waiting for Redis Instance (%s) to become available", d.Get("name"))
stateConf := &resource.StateChangeConf{
	Pending:    []string{"Updating", "Creating"},
	Target:     []string{"Succeeded"},
	Refresh:    redisStateRefreshFunc(ctx, client, resGroup, name),
	Timeout:    60 * time.Minute,
	MinTimeout: 15 * time.Second,
}
if _, err := stateConf.WaitForState(); err != nil {
	return fmt.Errorf("Error waiting for Redis Instance (%s) to become available: %s", d.Get("name"), err)
}
```

```go
func redisStateRefreshFunc(ctx context.Context, client redis.Client, resourceGroupName string, sgName string) resource.StateRefreshFunc {
	return func() (interface{}, string, error) {
		res, err := client.Get(ctx, resourceGroupName, sgName)
		if err != nil {
			return nil, "", fmt.Errorf("Error issuing read request in redisStateRefreshFunc to Azure ARM for Redis Cache Instance '%s' (RG: '%s'): %s", sgName, resourceGroupName, err)
		}

		return res, *res.ProvisioningState, nil
	}
}
```

What do you think of this instead?

tombuildsstuff:
@vlouwagie yeah, sorry that's what I meant :)

The only thing I'd suggest adding to that would be a `ContinuousTargetOccurence: 5` (as in my example above) to ensure that the Virtual Network Rule is returned 5 times in a row before returning; otherwise the function will return as soon as it exists a single time (which won't be the case if it's eventually consistent)

Thanks!

vlouwagie:
I increased the `ContinuousTargetOccurence ` attribute to 5.

Let me know if there is anything else I can do for this. 😄

website/docs/r/sql_virtual_network_rule.html.markdown
```diff
@@ -0,0 +1,96 @@
...
+
+* `endpoint_state` - The provisioning status of the endpoint.
+
+
+=> Possible values for `endpoint_state` are:
+* Ready - Indicates that the endpoint was provisioned successfully.
+* Initializing - Indicates that the endpoint provisioning is in the initialization stage.
+* InProgress - Indicates that the endpoint is still bring provisioned.
+* Deleting - Indicates that the endpoint is currently being deleted.
+* Unknown - Indicates that the state of the endpoint is unknown.
```
tombuildsstuff: I'm going to push a commit to remove this field, since it shouldn't be being used by end-users

tombuildsstuff: (done)

vlouwagie:
@tombuildsstuff I suppose that field is not necessary, even more now since we added a check/wait function for the ready status.

tombuildsstuff:
@vlouwagie yeah - we can always expose it if there's a use-case for it, but in general Terraform should only be returning from the Create method when it's ready/replicated (as in this case 😄). Thanks for this PR - this'll go out in v1.3.1 later today :)

PR #991, opened 2018-03-16
Replace DNS NS record objects with records array

phekmat:
- Replace and deprecate the `azurerm_dns_ns_record.record` objects with
a single `records` array to simplify working with DNS zone resources

The motivation behind this was primarily to be able to use `azurerm_dns_zone.name_servers` directly, but it's equally useful with plain `variable`s. TXT records have the same issue and it would be nice to give it the same treatment.


azurerm/resource_arm_dns_ns_record.go
```diff
@@ -152,31 +166,41 @@ func resourceArmDnsNsRecordDelete(d *schema.ResourceData, meta interface{}) erro
 	return nil
 }
 
-func flattenAzureRmDnsNsRecords(records *[]dns.NsRecord) []map[string]interface{} {
+func flattenAzureRmDnsNsRecordsSet(records *[]dns.NsRecord) []map[string]interface{} {
```
katbyte: Could we just get a quick //todo remove this when "record" is removed here?

azurerm/resource_arm_dns_ns_record.go
```diff
@@ -122,7 +132,11 @@ func resourceArmDnsNsRecordRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("zone_name", zoneName)
 	d.Set("ttl", resp.TTL)
 
-	if err := d.Set("record", flattenAzureRmDnsNsRecords(resp.NsRecords)); err != nil {
+	if err := d.Set("records", flattenAzureRmDnsNsRecords(resp.NsRecords)); err != nil {
+		return err
```
katbyte: could we wrap this error with a bit more detail, such as:
```go
if err := d.Set("records", flattenAzureRmDnsNsRecords(resp.NsRecords)); err != nil {
    return fmt.Errorf("Error setting `records `: %+v", err)
    }
```

azurerm/resource_arm_dns_ns_record.go
```diff
@@ -122,7 +132,11 @@ func resourceArmDnsNsRecordRead(d *schema.ResourceData, meta interface{}) error
...
+	}
+
+	if err := d.Set("record", flattenAzureRmDnsNsRecordsSet(resp.NsRecords)); err != nil {
```
katbyte: same as above

azurerm/resource_arm_dns_ns_record.go
```diff
@@ -33,9 +35,21 @@ func resourceArmDnsNsRecord() *schema.Resource {
 				Required: true,
 			},
 
+			"records": {
+				Type: schema.TypeList,
+				//TODO: add `MinItems: 1`` and `Required: true` once we remove the `record` attribute
```
tombuildsstuff: _minor_ `MinItems: 1` is implied by the fact that this would be required, so we only actually need `Required: true` :)

azurerm/resource_arm_dns_ns_record.go
```diff
@@ -152,35 +171,64 @@ func resourceArmDnsNsRecordDelete(d *schema.ResourceData, meta interface{}) erro
...
+	return results
+}
 
-		nsRecord := dns.NsRecord{
-			Nsdname: &nsdName,
+func expandAzureRmDnsNsRecords(d *schema.ResourceData) ([]dns.NsRecord, error) {
+	var records []dns.NsRecord
+
+	//TODO: remove this once we remove the `record` attribute
+	if d.HasChange("record") {
```
tombuildsstuff: IMO we should flip this around to check for `records` then `record` (as the else statement) since `records` is now the recommended field

azurerm/resource_arm_dns_ns_record_migration.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func resourceArmDNSNsMigrateState(
```
tombuildsstuff: (as above) I think we can remove this, since this should be handled by a `refresh` which happens during a plan/apply?

azurerm/resource_arm_dns_ns_record.go
```diff
@@ -33,9 +35,21 @@ func resourceArmDnsNsRecord() *schema.Resource {
 				Required: true,
 			},
 
+			"records": {
```
tombuildsstuff: can we also update the documentation for this field? this is hosted in the `website/docs/r/dns_ns_record.html.markdown` file

azurerm/resource_arm_dns_ns_record.go
```diff
@@ -33,9 +35,17 @@ func resourceArmDnsNsRecord() *schema.Resource {
 				Required: true,
 			},
 
-			"record": {
-				Type:     schema.TypeSet,
+			"records": {
+				Type:     schema.TypeList,
```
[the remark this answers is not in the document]
phekmat: I had the same concern. I couldn't find an explicit guarantee that Azure won't change the order someday, but it does respect the order in the portal and the REST API/CLI for both reads and writes.

azurerm/resource_arm_dns_ns_record.go
phekmat: Done. I added a warning that one or the other is required, but I'm open to changing the verbiage.

azurerm/resource_arm_dns_ns_record_migration.go
phekmat: Done

azurerm/resource_arm_dns_ns_record.go
phekmat: Works for me. Done

azurerm/resource_arm_dns_ns_record.go
phekmat: Ah, good to know. I updated the TODO

azurerm/resource_arm_dns_ns_record.go
```diff
@@ -18,6 +18,8 @@ func resourceArmDnsNsRecord() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
+		MigrateState:  resourceArmDNSNsMigrateState,
+		SchemaVersion: 1,
```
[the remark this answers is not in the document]
phekmat: Done

phekmat:
@katbyte That's actually a bug I introduced with the last commit in inverting the `if` condition in `expandAzureRmDnsNsRecords`. I'll fix that and re-run the acceptance tests locally to verify

azurerm/resource_arm_dns_ns_record.go
```diff
@@ -18,6 +18,8 @@ func resourceArmDnsNsRecord() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
+		// MigrateState:  resourceArmDNSNsMigrateState,
```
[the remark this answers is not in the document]
phekmat: Done

phekmat:
I've fixed the issue causing the test failure (though I think it can be simplified; I'm not 100% sure). I've also added a test for changing from `record` to `records`.

PR #993, opened 2018-03-17
`azurerm_app_service` - allow changing `client_affinity_enabled` without requiring a resource recreation

sabbox:
This PR allows to change ARR Affinity setting without destroying app service.
I've added test case but not sure how to verify properly app service not re-created


tombuildsstuff:
App Service tests pass:

screen shot 2018-03-26 at 12 02 18

PR #1000, opened 2018-03-19
Increase max staleness prefix for multiple regions and fix documentation

azmelanar:
Hi,

In this PR two fixes:

- fix in documentation, because key is unknown during planning
`Error: azurerm_cosmosdb_account.cosmosdb: consistency_policy.0: invalid or unknown key: max_staleness`
- increase max_staleness_prefix, because
`The maximum lag must be between 5 seconds and 1 day for accounts with a single region, and between 5 minutes and 1 day for accounts with multiple regions.`

Please let me know, if there are any questions.

Thanks

### DIFF
--- a/azurerm/resource_arm_cosmos_db_account.go
+++ b/azurerm/resource_arm_cosmos_db_account.go
@@ -87,7 +87,7 @@ func resourceArmCosmosDBAccount() *schema.Resource {
 							Type:         schema.TypeInt,
 							Optional:     true,
 							Default:      5,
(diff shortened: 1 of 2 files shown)


katbyte:
Tests pass:
screen shot 2018-03-19 at 15 19 40

lorenyu:
Does it make sense to increate the default values, or to change the example to use a higher value for max staleness? Since the example is across multiple regions, the default of 5 seconds / 100 operations isn't allowed in Azure. People who try out the example would end up seeing this error:

image

PR #1011, opened 2018-03-20
New Data Source: `azurerm_storage_account_sas `

### LINKED ISSUE #59: Feature Request: support for Shared Access Signatures on Storage Objects [AzureRM]
_This issue was originally opened by @amcguign as hashicorp/terraform#13451. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

In order to fufill our audit  requirements we need to ensure that all Azure Blob storage logs are being ingested into Splunk. 

After some investigation we believe SAS tokens are the most suitable way to do this. "Using a shared access signature (SAS) is a powerful way to grant limited access to objects in your storage account to other clients, without having to expose your account key." -- https://docs.microsoft.com/en-us/azure/storage/storage-dotnet-shared-access-signature-part-1

jzampieron:
This PR is a WIP support for programmatically generating Shared Access Signatures for Storage Accounts.

The signature computation and url query string generation from the connection strings generated by terraform `azurerm_storage_account` resources seem to be working.

Much additional validation and checking logic and more test cases are necessary.

(fixes #59)


jzampieron:
@tombuildsstuff This is now working all the way through as a data source. I still need to write more/better validation functions, documentation and an acceptance test or two. 

I'd appreciate a quick once-over to make sure I'm still on-track. The definition now looks like:
```
data "azurerm_storage_account_sas" "sas_for_metrics" {
    connection_string = "${azurerm_storage_account.dcos.primary_connection_string}"
    https_only        = true
    resource_types {
        service   = true
        container = false
        object    = false
    }
    services {
        blob  = true
        queue = false
        table = false
        file  = false
    }
    start   = "2018-03-21"
    expiry  = "2020-03-21"
    permissions {
        read    = true
        write   = true
        delete  = false
        list    = false
        add     = true
        create  = true
        update  = false
        process = false
    }
}
```
And gathering the output is this style:
```
resource "local_file" "sas_output" {
    content  = "${data.azurerm_storage_account_sas.sas_for_metrics.sas}"
    filename = "${path.cwd}/outputs/sas_output.txt"
}
```

jzampieron:
@tombuildsstuff This should be now worth a review. I added better unit tests, an acceptance test and the documentation. LMK next steps. Thanks!

tombuildsstuff:
Fixes #59

azurerm/data_source_storage_account_sas.go
```diff
@@ -0,0 +1,390 @@
...
+						},
+
+						"table": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+
+						"file": {
+							Type:     schema.TypeBool,
+							Required: true,
```
tombuildsstuff: I think each of these properties also needs `ForceNew: true` on them?`

azurerm/data_source_storage_account_sas.go
```diff
@@ -0,0 +1,390 @@
...
+
+						"update": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+
+						"process": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
```
tombuildsstuff: I think each of these properties also needs `ForceNew: true` on them?

azurerm/data_source_storage_account_sas.go
```diff
@@ -0,0 +1,390 @@
...
+
+						"container": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+
+						"object": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
```
tombuildsstuff: I think each of these properties also needs `ForceNew: true` on them?

azurerm/data_source_storage_account_sas.go
```diff
@@ -0,0 +1,390 @@
...
+
+						"process": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+					},
+				},
+			},
+
+			"sas": {
```
tombuildsstuff: 🤔 is it worth making this `sas_token` or `storage_access_token`?

azurerm/data_source_storage_account_sas.go
```diff
@@ -0,0 +1,390 @@
...
+	start := d.Get("start").(string)
+	expiry := d.Get("expiry").(string)
+	permissionsIface := d.Get("permissions").([]interface{})
+
+	resourceTypes := buildResourceTypesString(resourceTypesIface[0].(map[string]interface{}))
+	services := buildServicesString(servicesIface[0].(map[string]interface{}))
+	permissions := buildPermissionsString(permissionsIface[0].(map[string]interface{}))
+
+	_, svcErr := validateArmStorageAccountSasResourceTypes(resourceTypes, "")
+	if svcErr != nil {
```
tombuildsstuff: _minor_ we can actually just reuse the same error object here all the way down (`err`)

azurerm/data_source_storage_account_sas.go
```diff
@@ -0,0 +1,390 @@
...
+
+	// Details on how to do this are here:
+	// https://docs.microsoft.com/en-us/rest/api/storageservices/Constructing-an-Account-SAS
+	accountName := kvp[connStringAccountNameKey]
+	accountKey := kvp[connStringAccountKeyKey]
+	var signedProtocol = "https,http"
+	if httpsOnly {
+		signedProtocol = "https"
+	}
+	signedIp := ""
```
tombuildsstuff: should we also expose this value?

azurerm/data_source_storage_account_sas_test.go
```diff
@@ -0,0 +1,277 @@
...
+  expiry  = "%s"
+  permissions {
+	read    = true
+	write   = true
+	delete  = false
+    list    = false
+    add     = true
+	create  = true
+	update  = false
+	process = false
```
tombuildsstuff: _minor_ can we make the spacing consistent here?

website/docs/d/storage_account_sas.html.markdown
```diff
@@ -0,0 +1,87 @@
+---
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_storage_account_sas"
+sidebar_current: "docs-azurerm-datasource-storage-account-sas"
+description: |-
+  Create a Shared Access Signature (SAS) for an Azure Storage Account.
+
+---
+
+# Data Source: azurerm_storage_account_sas
```
tombuildsstuff: 🤔 given `sas is storage account secret, it is worth naming this `azurerm_storage_account_secret/token`?

website/docs/d/storage_account_sas.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+* `start` - (Required) The starting time and date of validity of this SAS. Must be a valid ISO-8601 format time/date string.
+* `expiry` - (Required) The expiration time and date of this SAS. Must be a valid ISO-8601 format time/date string.
+* `permissions` - (Required) A set of true/false flags which define the granted permissions.
+
+Refer to the [SAS creation reference from Azure](https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-an-account-sas)
+for additional details on the fields above.
+
+## Attributes Reference
+
+* `sas` - (Computed) The computed Account Shared Access Signature (SAS).
```
tombuildsstuff: can we remove `(Computed)` here? that should be implied as it's in the Attributes reference

website/docs/d/storage_account_sas.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+
+## Argument Reference
+
+* `connection_string` - (Required) The connection string for the storage account to which this SAS applies. Typically directly from the `primary_connection_string` attribute of a terraform created `azurerm_storage_account` resource.
+* `https_only` - (Optional) Only permit `https` access. If `false`, both `http` and `https` are permitted. (Default `true`)
+* `resouce_types` - (Required) A set of true/false flags which define the storage account resource types that are granted access by this SAS. 
+* `services` - (Required) A set of true/false flags which define the storage account services that are granted access by this SAS.
+* `start` - (Required) The starting time and date of validity of this SAS. Must be a valid ISO-8601 format time/date string.
+* `expiry` - (Required) The expiration time and date of this SAS. Must be a valid ISO-8601 format time/date string.
+* `permissions` - (Required) A set of true/false flags which define the granted permissions.
```
tombuildsstuff: _minor_ can we quote `true` and `false` here?

website/docs/d/storage_account_sas.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+output "sas_url_query_string" {
+  value = "${data.azurerm_storage_account_sas.sas}"
+}
+```
+
+## Argument Reference
+
+* `connection_string` - (Required) The connection string for the storage account to which this SAS applies. Typically directly from the `primary_connection_string` attribute of a terraform created `azurerm_storage_account` resource.
+* `https_only` - (Optional) Only permit `https` access. If `false`, both `http` and `https` are permitted. (Default `true`)
+* `resouce_types` - (Required) A set of true/false flags which define the storage account resource types that are granted access by this SAS.
```
tombuildsstuff: _minor_ can we quote `true` and `false` here?

website/docs/d/storage_account_sas.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+  value = "${data.azurerm_storage_account_sas.sas}"
+}
+```
+
+## Argument Reference
+
+* `connection_string` - (Required) The connection string for the storage account to which this SAS applies. Typically directly from the `primary_connection_string` attribute of a terraform created `azurerm_storage_account` resource.
+* `https_only` - (Optional) Only permit `https` access. If `false`, both `http` and `https` are permitted. (Default `true`)
+* `resouce_types` - (Required) A set of true/false flags which define the storage account resource types that are granted access by this SAS. 
+* `services` - (Required) A set of true/false flags which define the storage account services that are granted access by this SAS.
```
tombuildsstuff: _minor_ can we quote `true` and `false` here?

website/docs/d/storage_account_sas.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+
+output "sas_url_query_string" {
+  value = "${data.azurerm_storage_account_sas.sas}"
+}
+```
+
+## Argument Reference
+
+* `connection_string` - (Required) The connection string for the storage account to which this SAS applies. Typically directly from the `primary_connection_string` attribute of a terraform created `azurerm_storage_account` resource.
+* `https_only` - (Optional) Only permit `https` access. If `false`, both `http` and `https` are permitted. (Default `true`)
```
tombuildsstuff: can we update "(Default `true`)" -> "Defaults to `true`" for consistency

azurerm/data_source_storage_account_sas.go
```diff
@@ -0,0 +1,390 @@
...
+			},
+
+			"resource_types": {
+				Type:     schema.TypeList,
+				Required: true,
+				ForceNew: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"service": {
```
phekmat: Would it be possible to default all of these (the `resource_types`, `permissions`, and `services` parameters) to `false` and make them optional?

jzampieron: @tombuildsstuff Agree with all of the above and will incorporate those changes.
@phekmat I debated that myself and came to the conclusion to be explicitly verbose here since there are security vs function issues here. 

If we default all to false, then there's an implicit no-access (all broken) case. 

IHMO it's better here to be verbose and demand that folks specify exactly what they want... 

This avoids an oops by granting too much access and an oops by granting no access both of which, IMHO, are a worse UX than some extra typing.

azurerm/data_source_storage_account_sas.go
```diff
@@ -0,0 +1,390 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			// Always in UTC and must be ISO-8601 format
+			"expiry": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
```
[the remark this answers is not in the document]
jzampieron: I'd prefer to go the route you specify with the validation function. In my research there are some slight differences between ISO-8601 and RFC3339. MSFT specifies that these are ISO-8601 and since the signatures are based on the actual text characters, the subtle differences may result in breakage/no-working. 

Given the fragile nature, by design, of the signatures, I think leaving as-is is preferable rather than constraining the input to possibly broken validation.

azurerm/data_source_storage_account_sas.go
```diff
@@ -0,0 +1,390 @@
...
+						},
+					},
+				},
+			},
+
+			// Always in UTC and must be ISO-8601 format
+			"start": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
```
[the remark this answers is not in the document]
jzampieron: Same notes as above.

azurerm/data_source_storage_account_sas.go
jzampieron: This is actually a result of my lack of understanding of the semantics of that flag on the outer `permissions` list object vs the inner schema. I'll add the flag.

azurerm/data_source_storage_account_sas.go
jzampieron: It's not really a token in the uuid/JWT sense... its actually presented as a URL query string... with inconsistent URL escaping in the various parts. Notice only some of the fields are passed to `url.QueryEscape` to be 100% consistent with how MSFT generates and expects these.

I debated including the `?`, but again, the `?` is consistent with the output from the Azure portal... and how most folks will use this, which is to simply concat it to the primary blob end point.

azurerm/data_source_storage_account_sas.go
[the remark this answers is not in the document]
tombuildsstuff: makes sense from my side - we can always add it in later based on feedback 👍

website/docs/d/storage_account_sas.html.markdown
jzampieron: For posterity: SAS is `shared access signature` in the context of this data source. It's not actually even really a signature by itself, but a whole URL query string, which is how the Azure tools and even Azure storage explorer use these.

azurerm/data_source_storage_account_sas_test.go
jzampieron: I see what happened. Go fmt didn't change the tabs/spaces in the string literal.

Will fix this in a minute.

azurerm/data_source_storage_account_sas.go
jzampieron: I debated this. The format is a little odd of the field and it's not all that functionally useful.

I say we omit it for now and if anyone asks for this added feature we can add it later.

jzampieron:
@tombuildsstuff Any further improvements required here? Thanks.

tombuildsstuff:
@jzampieron sorry, I started a review but didn't hit submit - I'll take another look now!

jzampieron:
@tombuildsstuff Checking in. I don't see any other actions required. Should be good to merge.

tombuildsstuff:
@jzampieron apologies for the delayed response here - I started this review then got distracted, I'll post that shortly and re-review this next week.

Chatting with Microsoft last week I discovered there's a couple of endpoints which provide this information - [`List Account SAS`](https://docs.microsoft.com/en-us/rest/api/storagerp/storageaccounts/listaccountsas) and [`List Storage SAS`](https://docs.microsoft.com/en-us/rest/api/storagerp/storageaccounts/listservicesas) - from what I can see it appears the `List Account SAS` endpoint would allow us to be consistent by leaning on Azure here instead; what do you think about switching over to use that?

Thanks!

jzampieron:
I'll work through the review in detail as above. w.r.t. the larger question of generating vs using the API. 

Short term I think we should just go with this to have an answer to the immediate feature needs. 

Longer term moving to the API will shift some of the burden to MSFT for compatibility, although it shouldn't really matter either way since the generation algorithm really shouldn't change at this point.

website/docs/d/storage_account_sas.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+
+## Argument Reference
+
+* `connection_string` - (Required) The connection string for the storage account to which this SAS applies. Typically directly from the `primary_connection_string` attribute of a terraform created `azurerm_storage_account` resource.
+* `https_only` - (Optional) Only permit `https` access. If `false`, both `http` and `https` are permitted. Defaults to `true`.
+* `resouce_types` - (Required) A set of `true`/`false` flags which define the storage account resource types that are granted access by this SAS. 
+* `services` - (Required) A set of `true`/`false` flags which define the storage account services that are granted access by this SAS.
+* `start` - (Required) The starting time and date of validity of this SAS. Must be a valid ISO-8601 format time/date string.
+* `expiry` - (Required) The expiration time and date of this SAS. Must be a valid ISO-8601 format time/date string.
+* `permissions` - (Required) A set of `true`/`false` flags which define the granted permissions.
```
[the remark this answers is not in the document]
jzampieron: Done.

azurerm/data_source_storage_account_sas.go
```diff
@@ -0,0 +1,405 @@
...
+		retVal += "o"
+	}
+
+	return retVal
+}
+
+func validateArmStorageAccountSasResourceTypes(v interface{}, _ string) (ws []string, es []error) {
+	input := v.(string)
+
+	if !regexp.MustCompile(`\A([cos]{1,3})\z`).MatchString(input) {
```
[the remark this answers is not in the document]
jzampieron: Done.

website/docs/d/storage_account_sas.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+  value = "${data.azurerm_storage_account_sas.sas}"
+}
+```
+
+## Argument Reference
+
+* `connection_string` - (Required) The connection string for the storage account to which this SAS applies. Typically directly from the `primary_connection_string` attribute of a terraform created `azurerm_storage_account` resource.
+* `https_only` - (Optional) Only permit `https` access. If `false`, both `http` and `https` are permitted. Defaults to `true`.
+* `resouce_types` - (Required) A set of `true`/`false` flags which define the storage account resource types that are granted access by this SAS. 
+* `services` - (Required) A set of `true`/`false` flags which define the storage account services that are granted access by this SAS.
```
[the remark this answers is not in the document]
jzampieron: Done.

website/docs/d/storage_account_sas.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+output "sas_url_query_string" {
+  value = "${data.azurerm_storage_account_sas.sas}"
+}
+```
+
+## Argument Reference
+
+* `connection_string` - (Required) The connection string for the storage account to which this SAS applies. Typically directly from the `primary_connection_string` attribute of a terraform created `azurerm_storage_account` resource.
+* `https_only` - (Optional) Only permit `https` access. If `false`, both `http` and `https` are permitted. Defaults to `true`.
+* `resouce_types` - (Required) A set of `true`/`false` flags which define the storage account resource types that are granted access by this SAS.
```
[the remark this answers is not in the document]
jzampieron: Done

PR #1033, opened 2018-03-27
New Resource: `azurerm_express_route_circuit_peering`

tombuildsstuff:
This PR adds support for Express Route Circuit Peering - in this PR we're only spinning up the connection, rather than physically connecting to it so we don't need to get some fibre hooked up - but this seems to suffice :)


tombuildsstuff:
Tests pass:

screen shot 2018-03-28 at 08 13 47

PR #1035, opened 2018-03-28
Combining the Network Watcher tests to avoid Azure provisioning issues

tombuildsstuff:
Currently it's only possible to provision one Network Watcher per region, meaning these tests are currently flapping. This PR updates the test to one sequentially run test, which should avoid this issue.


tombuildsstuff:
The new combined test passes:

screen shot 2018-03-28 at 08 54 25

PR #1055, opened 2018-04-02
Updates and bug fixes to azurerm_cosmosdb_account

katbyte:
reviewed and improved azurerm_cosmosdb_account:
- Added new field `enable_automatic_failover`
- Now exports `endpoint`, `read_enpoints` and `write_endpoints`
- `failover_policy` has been renamed to `geo_locations` to better match the API & portal terms
- replication locations can now be renamed and have their `failover_priority` change (except for the primary location)
- properties can now be changed at the same time replication locations are added
- replication location ID's can now be set via the `prefix` attribute customizing their endpoints
- `consistency_policy` is now a list so its attributes can be referenced 
- added additional input validation
- now supports `ConsistentPrefix` consistency level

Also added more comprehensive tests and an `cosmos-db` example

PR #1056, opened 2018-04-02
Added new datasource azurerm_cosmosdb_account

katbyte:
Review and merge after #1055


GNUmakefile
```diff
@@ -14,6 +14,9 @@ test: fmtcheck
 testacc: fmtcheck
 	TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 180m
 
+debugacc: fmtcheck
+	TF_ACC=1 dlv test $(TEST) --headless --listen=:2345 --api-version=2 -- -test.v $(TESTARGS)
```
tombuildsstuff: 👍  this is a helpful one to expose; we should probably document how to use it at some point

azurerm/data_source_cosmos_db_account.go
```diff
@@ -0,0 +1,218 @@
...
+	d.SetId(*resp.ID)
+	d.Set("name", resp.Name)
+	d.Set("location", azureRMNormalizeLocation(*resp.Location))
+	d.Set("resource_group_name", resourceGroup)
+	flattenAndSetTags(d, resp.Tags)
+
+	d.Set("kind", string(resp.Kind))
+	d.Set("offer_type", string(resp.DatabaseAccountOfferType))
+	d.Set("ip_range_filter", resp.IPRangeFilter)
+	d.Set("endpoint", resp.DocumentEndpoint)
```
tombuildsstuff: so we should be accessing all of these properties via the `Properties` sub-object, since this matches the actual API response (e.g. `resp.Properties.DocumentEndpoint`); and also check for this being nil via:

```
if props := resp.[XX]Properties; props != nil {
  d.Set("ip_range_filter", props.IPRangeFilter)
}
```

I've been meaning to open a bug on the Azure SDK repository about removing the top-level fields, since they're duplicating the other values and not consistent with the API Response (with the exception of Name, Location, Kind which need to be at the top-level)

azurerm/data_source_cosmos_db_account.go
```diff
@@ -0,0 +1,218 @@
...
+	if v := resp.EnableAutomaticFailover; v != nil {
+		d.Set("enable_automatic_failover", resp.EnableAutomaticFailover)
+	}
+
+	if err := d.Set("consistency_policy", flattenAzureRmCosmosDBAccountConsistencyPolicy(resp.ConsistencyPolicy)); err != nil {
+		return fmt.Errorf("Error setting CosmosDB Account %q `consistency_policy` (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	//sort `geo_locations` by fail over priority
+	locations := make([]map[string]interface{}, len(*resp.FailoverPolicies))
```
tombuildsstuff: there's a potential crash here if FailoverPolicies is nil (which is possible from a bad API Response) - can we access this on `.XXProperties` and nil-check it?

website/docs/d/cosmosdb_account.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+
+* `offer_type` - The Offer Type to used by this CosmosDB Account.
+
+* `kind` - The Kind of the CosmosDB account.
+
+* `ip_range_filter` - The current IP Filter for this CosmosDB account
+
+* `enable_automatic_failover` - If automatic failover is enabled for this CosmosDB Account.
+
+`consistency_policy` The current consistency Settings for this CosmosDB account with the follow properties:
```
tombuildsstuff: `follow` -> `following`

azurerm/data_source_cosmos_db_account.go
```diff
@@ -0,0 +1,218 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"location": locationForDataSourceSchema(),
+
+			"resource_group_name": resourceGroupNameForDataSourceSchema(),
```
tombuildsstuff: _really minor_ we order fields by Required -> Optional -> Computed; since `name` and `resource_group_name` are both required here and everything else (including `location`) is Computed - it'd be consistent to move these up - but it's not a blocker / an FYI for the future

azurerm/data_source_cosmos_db_account.go
```diff
@@ -0,0 +1,218 @@
...
+	d.Set("location", azureRMNormalizeLocation(*resp.Location))
+	d.Set("resource_group_name", resourceGroup)
+	flattenAndSetTags(d, resp.Tags)
+
+	d.Set("kind", string(resp.Kind))
+	d.Set("offer_type", string(resp.DatabaseAccountOfferType))
+	d.Set("ip_range_filter", resp.IPRangeFilter)
+	d.Set("endpoint", resp.DocumentEndpoint)
+
+	if v := resp.EnableAutomaticFailover; v != nil {
```
tombuildsstuff: since this is a Bool, we shouldn't need to nil-check this

azurerm/data_source_cosmos_db_account.go
```diff
@@ -0,0 +1,218 @@
...
+	//sort `geo_locations` by fail over priority
+	locations := make([]map[string]interface{}, len(*resp.FailoverPolicies))
+	for _, l := range *resp.FailoverPolicies {
+		locations[*l.FailoverPriority] = map[string]interface{}{
+			"id":                *l.ID,
+			"location":          azureRMNormalizeLocation(*l.LocationName),
+			"failover_priority": int(*l.FailoverPriority),
+		}
+	}
+	d.Set("geo_location", &locations)
```
tombuildsstuff: we shouldn't need the `&` here since `d.Set` will take care of the referencing for us

tombuildsstuff:
Tests pass:

screen shot 2018-05-06 at 08 34 01

PR #1062, opened 2018-04-03
App Service Plan: Documenting the requirement on an Isolated SKU for ASEv2

### LINKED ISSUE #1009: App Service Plan resource documentation error
How do I report an error in documentation?  I am working on deploying an App Service Plan into and App Service Environment and the the documentation on the terraform.io website incorrectly states that if you are attaching an App Service Plan to an App Service Environment you must use the Premium SKU.  It should read that it is required to use the Isolated SKU.

tombuildsstuff:
Fixes #1009

### DIFF
--- a/website/docs/r/app_service_plan.html.markdown
+++ b/website/docs/r/app_service_plan.html.markdown
@@ -83,7 +83,7 @@ The following arguments are supported:
 
 * `app_service_environment_id` - (Optional) The ID of the App Service Environment where the App Service Plan should be located. Changing forces a new resource to be created.
 
-~> **NOTE:** Attaching to an App Service Environment requires the App Service Plan use a `Premium` SKU.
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
> Theoretically it's possible to add validation for this (that's assuming that there's an easy way to find the ASE version).

It is by making an API call to Azure to retrieve the App Service Environment; however given SKU requirements can change (such as until yesterday ExpressRoute Circuits needed a Premium SKU for MicrosoftPeerings, but [now work with Standard SKU's](https://azure.microsoft.com/en-gb/updates/microsoft-peering-on-expressroute-standard/), but only in Azure Public) - we tend to push SKU validation onto Azure (since it can also vary by region). If it were possible to determine those requirements through an API call then this'd be possible, but I don't believe it is at this time :(

PR #1074, opened 2018-04-05
Wrap d.Set("location") with a nil check call and normalize d.Get("location") 

katbyte:


tombuildsstuff:
Context for this PR: Azure appears to return the state of a resource at the time that it was last modified, rather than as right now. This means for older resources the `location` field can be omited (which will cause a crash) - this PR fixes this.

PR #1086, opened 2018-04-06
Redis Cache: support for clusters on the internal network

### LINKED ISSUE #82: Support for setting subnetId and staticIP for Premium sku in Azure Redis Cache
_This issue was originally opened by @mcfin as hashicorp/terraform#14187. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

**Terraform Version**
0.9.4

**Affected Resource(s)**
azurerm_redis_cache

**Expected Behavior**
Should be able to specify the subnetId and staticIP of the Redis cache.

**Actual Behavior**
There is no documentation for supplying these arguments and it does not appear that resource_arm_redis_cache.go provides for supplying them either.

**References**
It looks like the Microsoft Azure SDK for Go provides support for the arguments.
https://github.com/Azure/azure-sdk-for-go/blob/master/arm/redis/models.go

tombuildsstuff:
This PR adds support for Redis Caches on the Internal Network

@metacpp / @JunyiYi / @paultyng / @katbyte I've named the field for the Private Static IP Address `private_static_ip_address` ([the API/Docs expose this as `StaticIP`](https://docs.microsoft.com/en-us/rest/api/redis/redis/create) - "Static IP address. Required when deploying a Redis cache inside an existing Azure Virtual Network.") to avoid any confusion that a Public Static IP address could be used - but perhaps this should match the API - what do you think?

Tasks:
 - [x] Support for Redis Clusters on the Internal Network
 - [x] Upgrade to the new API version which fixes the Subnet deletion issue

Fixes #82


tombuildsstuff:
Tests pass:

screen shot 2018-04-06 at 15 57 33

azurerm/resource_arm_redis_cache.go
```diff
@@ -406,19 +426,22 @@ func resourceArmRedisCacheRead(d *schema.ResourceData, meta interface{}) error {
...
 	if sku := resp.Sku; sku != nil {
 		d.Set("capacity", sku.Capacity)
 		d.Set("family", sku.Family)
 		d.Set("sku_name", sku.Name)
 	}
 
-	if resp.ShardCount != nil {
-		d.Set("shard_count", resp.ShardCount)
+	if props := resp.Properties; props != nil {
+		d.Set("ssl_port", props.SslPort)
```
metacpp: Shall we build a helper method to flatten it ?

metacpp:
@tombuildsstuff LGTM and I resolved the conflict for you, please double check if there is anything wrong.

azurerm/resource_arm_redis_cache.go
tombuildsstuff: nope - we only do that for complex (e.g. nested) field, a single field isn't necessary

hafizullah:
Great work guys. Could this be used for all tiers or just premium tier?

tombuildsstuff:
@hafizullah Terraform works with the same SKU's as available in Azure - from memory support for this is limited to the Premium SKU, but I may be wrong :)

PR #1094, opened 2018-04-09
Continuing to fix test failures

tombuildsstuff:
- Ensuring the `https_only` field is set for `azurerm_function_app`
- Removing the `AzurePublicPeering` tests since this is no longer available for new connections


tombuildsstuff:
Function App tests pass too:

screen shot 2018-04-10 at 12 05 22

Issue #555, opened 2017-11-14
stopped virtual machine in terraform


hashibot:
_This issue was originally opened by @retheshnair as hashicorp/terraform#16629. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

 Currently when I stop VM terraform doesn't change the state in azure

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please try upgrading because your issue may have already been fixed.

[root@puppettest PUPPETPREPROD]# terraform -v
Terraform v0.10.8

[root@puppettest PUPPETPREPROD]#

tombuildsstuff:
hey @retheshnair

Thanks for opening this issue.

The AzureRM Terraform Provider doesn't currently support managing the State of Virtual Machines (e.g. Running/Stopped) - so that we can understand the issue better, would it be possible to get some more information about the issue you're facing?

Thanks!

clippermadness:
@tombuildsstuff I may be seeing the same or a similar issue with stopped (deallocated) VMs in Azure and terraform. I have three machines that I keep stopped and launch programatically based on data availability in my application. If they are stopped, terraform plan reports:

`An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  ~ azurerm_virtual_machine.pipeline_audprocessor_dev2
      storage_os_disk.0.disk_size_gb:      "0" => "32"
      storage_os_disk.0.managed_disk_type: "" => "Standard_LRS"

  ~ azurerm_virtual_machine.pipeline_vidprocessor_faces_dev2
      storage_os_disk.0.disk_size_gb:      "0" => "32"
      storage_os_disk.0.managed_disk_type: "" => "Standard_LRS"

  ~ azurerm_virtual_machine.pipeline_vidprocessor_frames_dev2
      storage_os_disk.0.disk_size_gb:      "0" => "32"
      storage_os_disk.0.managed_disk_type: "" => "Standard_LRS"


Plan: 0 to add, 3 to change, 0 to destroy.`

When I run terraform apply (just for giggles), I get errors:

`Error: Error applying plan:

3 error(s) occurred:

* azurerm_virtual_machine.pipeline_audprocessor_dev2: 1 error(s) occurred:

[...]

PR #1100, opened 2018-04-10
Use disk API to load managed disk info when VM is stopped.

JunyiYi:
When VM is stopped, `vmClient` won't return disks information.
In this pull request, I've done the following steps to resolve such issues similar to #555 :
* Check the status of VM using `InstanceView`
* When it is stopped, use `diskClient` to query the managed disk information
* Use the information to update the VM schema like `"disk_size_gb"` and `"managed_disk_type"`

### DIFF
--- a/azurerm/resource_arm_virtual_machine.go
+++ b/azurerm/resource_arm_virtual_machine.go
@@ -632,13 +632,27 @@ func resourceArmVirtualMachineRead(d *schema.ResourceData, meta interface{}) err
 	name := id.Path["virtualMachines"]
 
 	resp, err := vmClient.Get(ctx, resGroup, name, "")
-
 	if err != nil {
 		if utils.ResponseWasNotFound(resp.Response) {
 			d.SetId("")
 			return nil
 		}
 		return fmt.Errorf("Error making Read request on Azure Virtual Machine %s: %+v", name, err)
+	}
+
+	instance, err := vmClient.InstanceView(ctx, resGroup, name)
+	if err != nil {
+		return fmt.Errorf("Error making InstanceView request on Azure Virtual Machine %s: %+v", name, err)
+	}
+
+	stopped := false
+	if instance.Statuses != nil {
+		for _, status := range *instance.Statuses {
+			if status.Code != nil && *status.Code == "PowerState/deallocated" {
+				stopped = true
+				break
+			}
+		}
 	}
 
 	d.Set("name", resp.Name)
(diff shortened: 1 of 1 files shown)


azurerm/resource_arm_virtual_machine.go
```diff
@@ -1064,7 +1120,7 @@ func flattenAzureRmVirtualMachineOsProfileLinuxConfiguration(config *compute.Lin
 	return []interface{}{result}
 }
 
-func flattenAzureRmVirtualMachineOsDisk(disk *compute.OSDisk) []interface{} {
+func flattenAzureRmVirtualMachineOsDisk(disk *compute.OSDisk, stopped bool, meta interface{}) ([]interface{}, error) {
```
metacpp: OS ?

azurerm/resource_arm_virtual_machine.go
```diff
@@ -641,6 +640,21 @@ func resourceArmVirtualMachineRead(d *schema.ResourceData, meta interface{}) err
...
 
+	instance, err := vmClient.InstanceView(ctx, resGroup, name)
+	if err != nil {
+		return fmt.Errorf("Error making InstanceView request on Azure Virtual Machine %s: %+v", name, err)
+	}
+
+	stopped := false
+	if instance.Statuses != nil {
+		for _, status := range *instance.Statuses {
+			if status.Code != nil && *status.Code == "PowerState/deallocated" {
```
metacpp: Only deallocated status will get empty info ?

azurerm/resource_arm_virtual_machine.go
```diff
@@ -957,7 +979,27 @@ func flattenAzureRmVirtualMachineOsProfileSecrets(secrets *[]compute.VaultSecret
...
+
+	resp, err := client.Get(ctx, resGroup, name)
+	if err != nil {
+		return nil, fmt.Errorf("[ERROR] Error making Read request on stopped VM Azure Managed Disk %s (resource group %s): %s", name, resGroup, err)
+	}
+
+	return &resp, nil
+}
+
+func flattenAzureRmVirtualMachineDataDisk(disks *[]compute.DataDisk, stopped bool, meta interface{}) (interface{}, error) {
```
metacpp: This function is taking more responsibility, we may make it simpler and only do flatten work.

azurerm/resource_arm_virtual_machine.go
metacpp: same suggestion as above.

azurerm/resource_arm_virtual_machine.go
```diff
@@ -1598,3 +1597,91 @@ func resourceArmVirtualMachineStorageImageReferenceHash(v interface{}) int {
...
+		if osDisks, ok := d.GetOk("storage_os_disk"); ok {
+			if err := resourceArmVirtualMachineUpdateManagedDisk(osDisks.([]interface{}), meta); err != nil {
+				return err
+			}
+			if err := d.Set("storage_os_disk", osDisks); err != nil {
+				return fmt.Errorf("[DEBUG] Error setting virtual machine revised storage OS disk error: %#v", err)
+			}
+		}
+		if dataDisks, ok := d.GetOk("storage_data_disk"); ok {
+			if err := resourceArmVirtualMachineUpdateManagedDisk(dataDisks.([]interface{}), meta); err != nil {
```
tombuildsstuff: updating the state twice isn't ideal here - we should assign this to the objects used prior to the flatten methods e.g. https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_virtual_machine.go#L960

JunyiYi:
@tombuildsstuff , I agree with adding test cases. But are there any clean ways to shut down the virtual machine (without raw Azure REST API calls) in a multi-step test case?

tombuildsstuff:
@JunyiYi we do this in many tests where we call the Azure API from within a Test Case - see the `TestAccAzureRMResourceGroup_disappears` test for instance where we delete the Resource Group and ensure that Terraform detects it; so we should be able to achieve the same thing by calling the `Stop` API and `Start` API's in turn :)

JunyiYi:
@tombuildsstuff , sure, I have added the test cases.

azurerm/resource_arm_virtual_machine.go
```diff
@@ -1598,3 +1626,23 @@ func resourceArmVirtualMachineStorageImageReferenceHash(v interface{}) int {
...
+
+	var diskInfo *compute.Disk
+	if disk != nil && disk.ID != nil {
+		diskID, err := parseAzureResourceID(*disk.ID)
+		if err != nil {
+			return nil, err
+		}
+		diskResp, err := client.Get(ctx, diskID.ResourceGroup, diskID.Path["disks"])
+		if err != nil {
+			return nil, err
```
tombuildsstuff: _minor_ for consistency purposes and in order to return a more useful error message - can we update this to be:

```
id, err := parseAzureResourceID(*disk.ID)
if err != nil {
  return nil, fmt.Errorf("Error parsing Disk ID %q: %+v", *disk.ID, err)
}

name := id.Path["disks"]
diskResp, err := client.Get(ctx, resourceGroup, name)
if err != nil {
  return nil, fmt.Errorf("Error retrieving Disk %q (Resource Group %q): %+v", name, resourceGroup, err)
}
```

PR #1103, opened 2018-04-10
Attempting to add further SCM types to app_services

istairbn:
https://github.com/terraform-providers/terraform-provider-azurerm/issues/1104
Trying to add an app service with GitHub SCM, it keeps failing out. Not entirely sure if this is the whole fix, but I think this is where it is validating against available SCMs?


metacpp:
@istairbn thanks for the contribution. Expose deployment integration with 3rd party like GitHub, Bitbucket, will require 3 things:

1. Set OAuth token to list the organization and repo under the user name.
2. Set SCMType.
3. Set repo and branch info in source control properties.

tombuildsstuff:
@metacpp 

> 1. Create OAuth token and set it through web management API, since for continuous deployment, Azure Web App require authorization to create WebHook on that repo. Details please visit: https://github.com/projectkudu/kudu/wiki/Investigating-continuous-deployment

So I believe that's just an Azure Portal limitation - it should be possible to achieve the same thing in Terraform by using [the `github_repository_webhook` resource](https://www.terraform.io/docs/providers/github/r/repository_webhook.html) to create the webhook and a Github deployment key for pulling the code from the repository - which can be specified as part of the clone URI (which is all the Portal's going to be doing internally), e.g.

```
git clone https://username:password@github.com/tombuildsstuff/example.git
```

It's worth noting that for Public repositories the Username and Password shouldn't be needed, so this would support both use-cases. (Internally Github supports using a Deployment Key as the `password` rather than an actual username and password) - as does Bitbucket afaik (since it's part of Git).

As such - I believe once the broken (set credentials) API endpoint is fixed it should be possible to implement these:

> 2. Set SCMType to predefined type.
> 3. Set repo and branch info in source control properties.

metacpp:
@tombuildsstuff thanks for the sharing, I was not aware of the resources provided by GitHub provider.

I did a quick glance on GitHub provider usage, it will require:
1. Personal access token with appropriate permission granted.
2. Organization name. It seems that personal might not be supported.

For web-hook resource, it does not support personal repositories, which will miss some of user scenarios. And if you take a look at the format web-hook URL set by Azure:
`https://$siteName:site-level-credential@siteName.scm.azurewebsites.net/deploy?scmType=GitHub`
This URL will need attributes(site_credential and source_control) of provisioned app-service resource. And since the web-hook is triggered by event(push), which means that you need to make a push to trigger the deployment.

For public repo under organization, with

metacpp:
@istairbn I am going to close this PR, please refer to #1121 for further work.

PR #1107, opened 2018-04-10
Filter out  from properties of tags object.

### LINKED ISSUE #945: Azurerm_metric_alertrules tag $type is treated as changed after apply
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.11.3
+ provider.azurerm v1.1.2
+ provider.template v1.0.0

### Affected Resource(s)
- azurerm_metric_alertrule

```hcl
resource "azurerm_metric_alertrule" "critical_cpu" {
  count               = "${length(local.cpu_critical_rule) > 0 ? length(var.resource_names) : 0}"

  name        = "${var.resource_names[count.index]}-cpu-critical"
  description = "CPU is high (>${local.cpu_critical_rule[0]}), please check for a potential auto-scaling issue"
  enabled     = true

  resource_id = "${var.resource_ids[count.index]}"
  metric_name = "Percentage CPU"
  operator    = "GreaterThan"
  threshold   = "${local.cpu_critical_rule[0]}"
  aggregation = "${local.cpu_critical_rule[1]}"
  period      = "${local.cpu_critical_rule[2]}"

  email_action {
    send_to_service_owners = false

    custom_emails = "${local.critical_notification_emails}"
  }

  tags = "${merge(var.tags, map("metric", "CPU", "level", "critical"))}"
}
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Expected Behavior
When plan metric alert, the tags should be exactly the number of tags I set via tags property. Tag $type should be treated as unchanged.

### Actual Behavior
Tags $type been treated as changed. 

      tags.$type: "Microsoft.WindowsAzure.Management.Common.Storage.CasePreservedDictionary, Microsoft.WindowsAzure.Management.Common.Storage" => ""
      tags.%:     "9" => "8"

### Steps to Reproduce
1. `terraform plan`
2. `terraform apply`
3. `terraform plan`

metacpp:
This PR is used to fix #945 while running `terraform plan` after `terraform apply` due to "$type" in the tags property map.

### DIFF
--- a/azurerm/resource_arm_metric_alertrule.go
+++ b/azurerm/resource_arm_metric_alertrule.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"log"
 
+	"strings"
+
 	"github.com/Azure/azure-sdk-for-go/services/monitor/mgmt/2017-05-01-preview/insights"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
@@ -138,7 +140,12 @@ func resourceArmMetricAlertRule() *schema.Resource {
 				},
 			},
 
-			"tags": tagsForMetricAlertRuleSchema(),
+			"tags": {
+				Type:         schema.TypeMap,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validateMetricAlertRuleTags,
+			},
 		},
 	}
 }
@@ -413,4 +420,19 @@ func resourceGroupAndAlertRuleNameFromId(alertRuleId string) (string, string, er
 	resourceGroup := id.ResourceGroup
 
 	return resourceGroup, name, nil
+}
+
+func validateMetricAlertRuleTags(v interface{}, f string) (ws []string, es []error) {
+	// Normal validation required by any AzureRM resource.
+	ws, es = validateAzureRMTags(v, f)
+
+	tagsMap := v.(map[string]interface{})
+
+	for k := range tagsMap {
+		if strings.EqualFold(k, "$type") {
+			es = append(es, fmt.Errorf("the %q is not allowed as tag name", k))
+		}
+	}
+
+	return ws, es
 }
--- a/azurerm/resource_arm_metric_alertrule.go
+++ b/azurerm/resource_arm_metric_alertrule.go
@@ -138,7 +138,7 @@ func resourceArmMetricAlertRule() *schema.Resource {
 				},
 			},
 
-			"tags": tagsSchema(),
+			"tags": tagsForMetricAlertRuleSchema(),
 		},
 	}
 }
@@ -276,7 +276,10 @@ func resourceArmMetricAlertRuleRead(d *schema.ResourceData, meta interface{}) er
 		d.Set("webhook_action", webhook_actions)
 	}
 
-	flattenAndSetTags(d, resp.Tags, "$type")
+	// Return a new tag map filtered by the specified tag names.
+	tagMap := filterTags(resp.Tags, "$type")
+
+	flattenAndSetTags(d, tagMap)
 
 	return nil
 }
--- a/azurerm/resource_arm_metric_alertrule.go
+++ b/azurerm/resource_arm_metric_alertrule.go
@@ -276,7 +276,7 @@ func resourceArmMetricAlertRuleRead(d *schema.ResourceData, meta interface{}) er
 		d.Set("webhook_action", webhook_actions)
 	}
 
-	flattenAndSetTags(d, resp.Tags)
+	flattenAndSetTags(d, resp.Tags, "$type")
 
 	return nil
 }
--- a/azurerm/resource_arm_metric_alertrule_test.go
+++ b/azurerm/resource_arm_metric_alertrule_test.go
@@ -27,6 +27,7 @@ func TestAccAzureRMMetricAlertRule_virtualMachineCpu(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testCheckAzureRMMetricAlertRuleExists(resourceName),
 					resource.TestCheckResourceAttr(resourceName, "enabled", "true"),
+					resource.TestCheckNoResourceAttr(resourceName, "$type"),
 				),
 			},
 			{
--- a/azurerm/resource_arm_metric_alertrule_test.go
+++ b/azurerm/resource_arm_metric_alertrule_test.go
@@ -11,6 +11,60 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
+func TestValidateMetricAlertRuleTags(t *testing.T) {
+	cases := []struct {
+		Name     string
+		Value    map[string]interface{}
+		ErrCount int
+	}{
+		{
+			Name: "Single Valid",
+			Value: map[string]interface{}{
+				"hello": "world",
+			},
+			ErrCount: 0,
+		},
+		{
+			Name: "Single Invalid",
+			Value: map[string]interface{}{
+				"$Type": "hello/world",
+			},
+			ErrCount: 1,
+		},
+		{
+			Name: "Single Invalid lowercase",
+			Value: map[string]interface{}{
+				"$type": "hello/world",
+			},
+			ErrCount: 1,
+		},
+		{
+			Name: "Multiple Valid",
+			Value: map[string]interface{}{
+				"hello": "world",
+				"foo":   "bar",
+			},
+			ErrCount: 0,
+		},
+		{
+			Name: "Multiple Invalid",
+			Value: map[string]interface{}{
+				"hello": "world",
+				"$type": "Microsoft.Foo/Bar",
+			},
+			ErrCount: 1,
+		},
+	}
+
+	for _, tc := range cases {
+		_, errors := validateMetricAlertRuleTags(tc.Value, "azurerm_metric_alert_rule")
+
+		if len(errors) != tc.ErrCount {
+			t.Fatalf("Expected %q to return %d errors but returned %d", tc.Name, tc.ErrCount, len(errors))
+		}
+	}
+}
+
 func TestAccAzureRMMetricAlertRule_virtualMachineCpu(t *testing.T) {
 	resourceName := "azurerm_metric_alertrule.test"
 	ri := acctest.RandInt()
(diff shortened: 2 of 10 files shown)


azurerm/tags.go
```diff
@@ -88,7 +88,10 @@ func flattenAndSetTags(d *schema.ResourceData, tagsMap map[string]*string) {
 	output := make(map[string]interface{}, len(tagsMap))
 
 	for i, v := range tagsMap {
-		output[i] = *v
+		// Filter out $type from tags object to avoid unexpected change on plan.
+		if i != "$type" {
+			output[i] = *v
+		}
```
tombuildsstuff: I think it's best to filter this on the individual resource where this needs to be filtered out (as people may have a legitimate use for the `$type` tag) - can we instead filter this out on the resource before calling this function? e.g.

```
tagsToOutput := make([]string, 0)
for i, v := range resp.Tags {
  tag = *v
  // Filter out $type from tags object to avoid unexpected change on plan.
  if tag != "$type" {
    tagsToOutput[i] = tag
  }
}
```

in addition - can we add some validation to ensure that the `$type` tag can't be added on that resource? We can do this using a `ValidateFunc` on the `tags` schema item for this resource :)

metacpp: What about the webhookAction?

metacpp: I've updated the code and tests.

metacpp:
@tombuildsstuff The PR is updated and I will share test results later.

metacpp:
```
=== RUN   TestAccAzureRMMetricAlertRule_sqlDatabaseStorage
--- PASS: TestAccAzureRMMetricAlertRule_sqlDatabaseStorage (261.07s)
PASS
=== RUN   TestAccAzureRMMetricAlertRule_importVirtualMachineCpu
--- PASS: TestAccAzureRMMetricAlertRule_importVirtualMachineCpu (512.51s)
PASS
=== RUN   TestAccAzureRMMetricAlertRule_virtualMachineCpu
--- PASS: TestAccAzureRMMetricAlertRule_virtualMachineCpu (522.19s)
PASS
ok    ../azurerm/test-azurerm    522.200s
```

azurerm/tags.go
```diff
@@ -79,16 +84,24 @@ func expandTags(tagsMap map[string]interface{}) map[string]*string {
...
 
+	// Only first optional parameter will be processed.
+	skipPropName := ""
+	if len(skipPropNames) > 0 && len(skipPropNames[0]) > 0 {
+		skipPropName = skipPropNames[0]
+	}
+
 	for i, v := range tagsMap {
-		output[i] = *v
+		if !strings.EqualFold(i, skipPropName) {
```
JunyiYi: Why using case insensitive comparison instead of "==" operator here?

azurerm/tags.go
```diff
@@ -79,16 +84,24 @@ func expandTags(tagsMap map[string]interface{}) map[string]*string {
...
-func flattenAndSetTags(d *schema.ResourceData, tagsMap map[string]*string) {
+func flattenAndSetTags(d *schema.ResourceData, tagsMap map[string]*string, skipPropNames ...string) {
 	if tagsMap == nil {
 		d.Set("tags", make(map[string]interface{}))
 		return
 	}
 
 	output := make(map[string]interface{}, len(tagsMap))
 
+	// Only first optional parameter will be processed.
```
JunyiYi: What about the rest of the parameters?

azurerm/tags.go
metacpp: "$type" seems to be reserved usage in PLAN, it should be case insensitive. Or, to be safer, we should use case sensitive instead.

azurerm/tags.go
metacpp: This is going to support optional parameter. A better way is to build a map instead. Let me refactor the code.

azurerm/tags.go
tombuildsstuff: 👍 for it being case insensitive

azurerm/resource_arm_metric_alertrule.go
```diff
@@ -276,7 +276,7 @@ func resourceArmMetricAlertRuleRead(d *schema.ResourceData, meta interface{}) er
 		d.Set("webhook_action", webhook_actions)
 	}
 
-	flattenAndSetTags(d, resp.Tags)
+	flattenAndSetTags(d, resp.Tags, "$type")
```
tombuildsstuff: this function's already doing too much as it is - can we filter out the tags to remove before passing this in? e.g.

```
tagsToOutput := make([]string, 0)
for k, v := range resp.Tags {
  if strings.EqualFold(k, "$type") {
   continue
  }
  tagsToOutput[k] = v
}
flattenAndSetTags(d, tagsToOutput, "$type")
```

azurerm/resource_arm_metric_alertrule_test.go
```diff
@@ -34,6 +34,7 @@ func TestAccAzureRMMetricAlertRule_virtualMachineCpu(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testCheckAzureRMMetricAlertRuleExists(resourceName),
 					resource.TestCheckResourceAttr(resourceName, "enabled", "false"),
+					resource.TestCheckNoResourceAttr(resourceName, "$type"),
```
tombuildsstuff: this is looking for the field `$type` rather than the Tag `$type` - as such this won't work. Instead this wants to check for `tags.$type`

azurerm/resource_arm_metric_alertrule_test.go
```diff
@@ -54,6 +55,7 @@ func TestAccAzureRMMetricAlertRule_sqlDatabaseStorage(t *testing.T) {
 				Config: config,
 				Check: resource.ComposeTestCheckFunc(
 					testCheckAzureRMMetricAlertRuleExists(resourceName),
+					resource.TestCheckNoResourceAttr(resourceName, "$type"),
```
tombuildsstuff: this is looking for the field `$type` rather than the Tag `$type` - as such this won't work. Instead this wants to check for `tags.$type`

azurerm/tags.go
tombuildsstuff: let's instead revert the changes to this method entirely and leave it in the single resource where this change is needed - this method's already doing too much

azurerm/resource_arm_metric_alertrule_test.go
metacpp: Good catch.

azurerm/resource_arm_metric_alertrule_test.go
metacpp: Good catch.

azurerm/resource_arm_metric_alertrule.go
metacpp: The reason that I refactored the function is to avoid future copy & paste for other resources:
1. The effort to add filtering logic is pretty small(few lines of code).
2. It does not break the KISS principle but emphasize the DRY principle, it's very common that having a customized filtering before doing any real work.

metacpp:
Updated the PR and reran the related tests without any regression:

=== RUN   TestAccAzureRMMetricAlertRule_sqlDatabaseStorage
--- PASS: TestAccAzureRMMetricAlertRule_sqlDatabaseStorage (204.23s)
PASS
=== RUN   TestAccAzureRMMetricAlertRule_importVirtualMachineCpu
--- PASS: TestAccAzureRMMetricAlertRule_importVirtualMachineCpu (500.87s)
PASS
=== RUN   TestAccAzureRMMetricAlertRule_virtualMachineCpu
--- PASS: TestAccAzureRMMetricAlertRule_virtualMachineCpu (560.33s)
PASS
ok    ../azurerm/test-azurerm    560.355s

Kicked off another comprehensive testing on Teamcity.

azurerm/tags.go
```diff
@@ -56,6 +57,10 @@ func validateAzureRMTags(v interface{}, k string) (ws []string, es []error) {
 			es = append(es, fmt.Errorf("the maximum length for a tag key is 512 characters: %q is %d characters", k, len(k)))
 		}
 
+		if strings.EqualFold(k, "$type") {
+			es = append(es, fmt.Errorf("the %q is not allowed as tag name", k))
+		}
```
tombuildsstuff: given this method is shared across all tags (and this validation is only specific to the `metrics_alertrule` resource) - can we make a specific validation method for the that resource? we can call the main validation function, and then check that `$type` isn't included?

metacpp: Interesting, it seems that this file is inconsistent between local and remote. Let me check it and refresh it.

metacpp: I did some investigation for $type while using it in tag naming for metric_alertrule resource:

1. You can't directly use $type as naming of key in TF, you need to use "$type".
2. "$type" is not acceptable by API:
```
Error: Error applying plan:

1 error(s) occurred:

* azurerm_metric_alertrule.critical_cpu: 1 error(s) occurred:

* azurerm_metric_alertrule.critical_cpu: insights.AlertRulesClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="UnsupportedRequestContent" Message="Request content is not well formed or supported."
```

metacpp:
Added one unit test and ran it successfully.
=== RUN   TestExpandARMTags
--- PASS: TestExpandARMTags (0.00s)

metacpp:
@tombuildsstuff does it look good to you for this PR?

azurerm/tags.go
tombuildsstuff: That's true for _this specific API_ but isn't the case across the whole of Azure, for instance I'm able to create a resource with `$type` as the key (note: this is double-escaped for HCL):

```
  name     = "tharvey-dev3"
  location = "West US"
  tags {
    "$$type" = "hi"
  }
}
```

in the case where this value isn't allowed we'll add validation logic to Terraform to the specific resources to handle this - but in general the key `$type` is a permitted value and as such we shouldn't obstruct that across the board.

metacpp: Please review my latest change, it seems that even if the PR is updated, the comment is still shown on old commit.

metacpp:
screen shot 2018-04-19 at 9 45 08 am

PR #1120, opened 2018-04-13
Improve the README for the preparation work before run acceptance tests

seanknox:
This provides feedback to the user with all of the ENV vars needed to run acceptance tests. Also updates README to mention the ENV vars.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,14 @@ $ make test
 
 In order to run the full suite of Acceptance tests, run `make testacc`.
 
+The following ENV variables must be set in your shell prior to running acceptance tests:
+- ARM_CLIENT_ID
+- ARM_CLIENT_SECRET
+- ARM_SUBSCRIPTION_ID
+- ARM_TENANT_ID
+- ARM_TEST_LOCATION
+- ARM_TEST_LOCATION_ALT
+
 *Note:* Acceptance tests create real resources, and often cost money to run.
 
 ```sh


GNUmakefile
```diff
@@ -11,7 +25,16 @@ test: fmtcheck
 	echo $(TEST) | \
 		xargs -t -n4 go test $(TESTARGS) -timeout=30s -parallel=4
 
-testacc: fmtcheck
+testacc:
+	@:$(call check_defined, \
+						ARM_CLIENT_ID \
```
metacpp: Environment VAR is not the only way to authorize terraform to provision resource.

README.md
```diff
@@ -103,6 +103,14 @@ $ make test
 
 In order to run the full suite of Acceptance tests, run `make testacc`.
 
+The following ENV variables must be set in your shell prior to running acceptance tests:
```
metacpp: ENV VAR is not the only way, but you can suggest user to use environment variables of SP. And refer to:
https://docs.microsoft.com/en-us/azure/virtual-machines/linux/terraform-install-configure

GNUmakefile
tombuildsstuff: @metacpp this is to run the Acceptance Tests - which are only supported by Service Principals

README.md
tombuildsstuff: striking out the above comment since it's incorrect, Service Principals are the only supported authentication method for the Acceptance Tests

GNUmakefile
```diff
@@ -11,7 +25,16 @@ test: fmtcheck
...
+						ARM_CLIENT_SECRET \
+						ARM_SUBSCRIPTION_ID \
+						ARM_TENANT_ID \
+						ARM_TEST_LOCATION \
+						ARM_TEST_LOCATION_ALT, \
+				"ARM_CLIENT_ID ARM_CLIENT_SECRET ARM_SUBSCRIPTION_ID ARM_TENANT_ID ARM_TEST_LOCATION and ARM_TEST_LOCATION_ALT must be set")
+	@$(MAKE) fmtcheck
```
tombuildsstuff: so there's already a check in the codebase for these variables, which should fail when these aren't specified (but only when running the Acceptance Tests, not any unit tests): https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/provider_test.go#L37

```
$ TF_ACC=1 go test ./azurerm -v -timeout 120m -run=TestAccAzureRMResourceGroup
=== RUN   TestAccAzureRMResourceGroup_importBasic
--- FAIL: TestAccAzureRMResourceGroup_importBasic (0.00s)
	provider_test.go:49: `ARM_CLIENT_ID` must be set for acceptance tests!
--- FAIL: TestAccAzureRMResourceGroup_basic (0.00s)
	provider_test.go:49: `ARM_CLIENT_ID` must be set for acceptance tests!
=== RUN   TestAccAzureRMResourceGroup_disappears
--- FAIL: TestAccAzureRMResourceGroup_disappears (0.00s)
	provider_test.go:49: `ARM_CLIENT_ID` must be set for acceptance tests!
--- FAIL: TestAccAzureRMResourceGroup_withTags (0.00s)
	provider_test.go:49: `ARM_CLIENT_ID` must be set for acceptance tests!
FAIL
exit status 1
FAIL	github.com/terraform-providers/terraform-provider-azurerm/azurerm	0.027s
```

seanknox:
> As mentioned in one of the comments we've already got a helper function for this within the tests that gets called during any Acceptance Test run - as such I'm not sure we need to duplicate this in the Makefile? That said, the readme enhancements look good to me - if we can revert the changes to the Makefile this otherwise LGTM :)

Thanks for reviewing! I saw (and used) the in-code checks, but if multiple are unspecified it required multiple runs to learn what they were. Happy to remove the Makefile checks in favor of the doc updates.

tombuildsstuff:
@seanknox thanks for pushing those updates :)

> Thanks for reviewing! I saw (and used) the in-code checks, but if multiple are unspecified it required multiple runs to learn what they were. Happy to remove the Makefile checks in favor of the doc updates.

That's a valid point - in which case I'd suggest it may be worth updating the validation in the Provider to return all of the field names which are missing? I'm going to merge this PR for the moment - but this logic exists in one place in the test framework, so it should be easy enough to update this to return an array of missing fields in the future: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/provider_test.go#L49

Thanks!

Issue #1104, opened 2018-04-10
Add further ScmTypes for azurerm_app_service


istairbn:
I have an app_service that uses GitHub as SCM. The only two options available currently are None and LocalGit
https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_app_service.go#L153 

However, it looks like the Azure SDK for Go now accepts many more SCM types. 

https://github.com/Azure/azure-sdk-for-go/blob/0b745831a99da5f43aa3cc641cfcb9ed323e1f9f/profiles/latest/web/mgmt/web/models.go#L732 

I had a go at adding it in the validation, but I'm not entirely sure if I got it in the right place. 
https://github.com/terraform-providers/terraform-provider-azurerm/pull/1103

tombuildsstuff:
Link to the latest SDK: https://github.com/Azure/azure-sdk-for-go/blob/v17.3.1/profiles/latest/web/mgmt/web/models.go#L479

whytoe:
@tombuildsstuff Is there any movement on this?

thoregilk:
Any progress on this @tombuildsstuff ? I`ve been working around it by invoking AZ CLI and local-exec.  But when I run plan or apply for the second time, it tries to modify the settings 

Terraform will perform the following actions:

  ~ module.webapp_monitor.azurerm_app_service.default
      site_config.0.scm_type: "ExternalHg" => "None"

And since scm_type is currently only expected to be None or LocalGit I`m not able to prevent Terraform to modify the setting.
expected site_config.0.scm_type to be one of [None LocalGit], got ExternalHg

benc-uk:
I have written an example using the Go SDK for Azure, and using the web resource provider and the CreateOrUpdateSourceControl function to allow you to add an external Git repo URL to your Azure App Service and have it automatically built & deployed by Kudu (part of App Service)

https://github.com/benc-uk/go-azure/blob/2484d00ab5293a28f77eff073c0bb6ed9214d7d1/src/webapp/main.go#L142

I've tried looking at your source and adding this support in myself with a mind to submitting a PR, but I couldn't get past it telling me that "source_control" was ready only

thoregilk:
We had to implement a work-around to add ExternalGit as scm_type while waiting. It is done by calling a powershell script with a null_resource. The script uses az account get-access-token to get a access token for Powershell and login-azurermaccount. 

[...]

PR #1121, opened 2018-04-14
[WIP]Add support of GitHub as deployment option

metacpp:
***Still work in progress***

Support GitHub as deployment option for azure web app, the workflow will be:

1. Generate Personal Access Token (read permission in both org and repo) to authorize Web App to access repo.
2. Set SCM Type as GitHub.
3. Set URL of repo and branch.

Will add example usage and tests in this PR very soon.


metacpp:
This PR is used to resolve #1104

azurerm/resource_arm_app_service.go
```diff
@@ -794,7 +882,7 @@ func validateAppServiceName(v interface{}, k string) (ws []string, es []error) {
 	return
 }
 
-func flattenAppServiceSiteCredential(input *web.UserProperties) []interface{} {
+func flattenAppServiceSiteCredential(input *web.User) []interface{} {
```
tombuildsstuff: we should be pulling from the `UserProperties ` object since the non-Properties object isn't actually valid in Azure (e.g. it doesn't actually match the schema and can be nil in older configurations)

azurerm/resource_arm_app_service.go
```diff
@@ -709,12 +755,33 @@ func flattenAppServiceSiteConfig(input *web.SiteConfig) []interface{} {
...
+	}
+
+	if v, ok := scm["branch"]; ok {
+		scmRet.Branch = utils.String(v.(string))
+	}
+
+	return scmRet
+}
+
+func flattenAppServiceSourceControl(input *web.SiteSourceControl) []interface{} {
```
tombuildsstuff: we should be pulling from the `SiteSourceControlProperties` object since the non-Properties object isn't actually valid in Azure (e.g. it doesn't actually match the schema and can be nil in older configurations)

metacpp:
@tombuildsstuff external_scm_credential is personal access token used by azure behind the scene to access org and repo information. It's not like the azure its own repo, there's no chance for any CI pipeline to directly access the repo hosted on GitHub. It's completely different scenario.

azurerm/resource_arm_app_service.go
```diff
@@ -275,8 +276,10 @@ func resourceArmAppService() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
 			"source_control": {
 				Type:     schema.TypeList,
+				Optional: true,
```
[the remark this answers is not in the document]
metacpp: In GitHub scenario, you need to specify the repo URL and branch name for Azure Web App to consume. It's not the same scenario as LocalGit, which will automatically generate the SCM URL hosted on Azure web app. I will update the example usage for GitHub scenario, then you will know what's the workflow.

azurerm/resource_arm_app_service.go
metacpp: User is using composition with UserProperties, which means that UserProperties object is sub-object of User object.

azurerm/resource_arm_app_service.go
metacpp: Internal Go SDK is using SiteSourceControl as returned type and SiteSourceControlProperties is embedded class, which means that you can access all the fields in SiteSourceControl. It's meaningless to do any kind of de-referencing before flattening.

I don't understand what's your concern here, maybe a more concrete example ?

tombuildsstuff: at this point in time, it's changed in the past - that's not how it's represented in the JSON Response so we should match that where possible (which we do throughout the provider)

metacpp: There's no any API return SiteSourceControlProperties, instead, it's SiteSourceControl.
If SiteSourceControl is nil, then you will not be able to get SiteSourceControlProperties, since they're the same object but different layer.

metacpp:
This PR does have dependencies on [#2910](https://github.com/Azure/azure-rest-api-specs/pull/2910) and [#1661](https://github.com/Azure/azure-sdk-for-go/pull/1661).

metacpp:
And for the personal token type, we have weak dependency on [#2946](https://github.com/Azure/azure-rest-api-specs/issues/2946)

tombuildsstuff:
@metacpp what's the status of this PR - it's been a `WIP` for a while and has merge conflicts?

tombuildsstuff:
@metacpp since we've not heard back from you here (and there's merge conflicts) - I'm going to close this PR for the moment; please feel free to re-open it if this is being actively worked on.

Thanks!

PR #1124, opened 2018-04-15
Add AzureRM Servicebus rule module

jpovey:
Add AzureRM Servicebus rule module, tests and documentation.  Applies to feature request https://github.com/terraform-providers/terraform-provider-azurerm/issues/145

PR #1150, opened 2018-04-23
Refactoring ServiceBus Subscription Rule

tombuildsstuff:
Refactoring the resource to make it consistent with the other resources by fixing some issues raised during PR review in #1124

cc @jpovey

### DIFF
--- a/azurerm/resource_arm_servicebus_subscription_rule.go
+++ b/azurerm/resource_arm_servicebus_subscription_rule.go
@@ -248,7 +248,7 @@ func resourceArmServiceBusSubscriptionRuleDelete(d *schema.ResourceData, meta in
 func expandAzureRmServiceBusCorrelationFilter(d *schema.ResourceData) (*servicebus.CorrelationFilter, error) {
 	configs := d.Get("correlation_filter").([]interface{})
 	if len(configs) == 0 {
-		return nil, fmt.Errorf("`correlation_filter` is required when `` is set to `CorrelationFilter`")
+		return nil, fmt.Errorf("`correlation_filter` is required when `filter_type` is set to `CorrelationFilter`")
 	}
 
 	config := configs[0].(map[string]interface{})
--- a/azurerm/resource_arm_servicebus_subscription_rule.go
+++ b/azurerm/resource_arm_servicebus_subscription_rule.go
@@ -7,14 +7,15 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/servicebus/mgmt/2017-04-01/servicebus"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/response"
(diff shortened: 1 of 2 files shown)


jpovey:
Hey Tom

So I know for any future contributions, is there a reason why it is preferred to inline a lot of the code into the CreateAndUpdate function rather than abstract it away into self contained functions?

cheers
Jon

tombuildsstuff:
@jpovey 

> So I know for any future contributions, is there a reason why it is preferred to inline a lot of the code into the CreateAndUpdate function rather than abstract it away into self contained functions?

In this case both of these functions were only expanding a single item - in general in the provider we inline an expand function unless it's of a significant size (e.g. the Disks resource on a Virtual Machine) - so consistency with the rest of the provider for the most part

Hope that helps - thanks again for the PR :)

azurerm/resource_arm_servicebus_subscription_rule.go
```diff
@@ -7,14 +7,15 @@ import (
...
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/response"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
 func resourceArmServiceBusSubscriptionRule() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceArmServiceBusSubscriptionRuleCreate,
+		Create: resourceArmServiceBusSubscriptionRuleCreateUpdate,
```
vancluever: I don't know if dual-purpose create/update functions are a good idea. An edge case I could see in the fact that ID is set in both cases is that if the ID changes for some reason (that is not expected), it will result in an orphaned resource. I definitely understand that there is a lot of common functionality between both workflows (possibly nearly all of it), but I think it's important to differentiate - the shared functionality could be wrapped, for example.

tombuildsstuff: the Azure API's are (almost) all upserts - thus we're calling the same SDK method/API endpoint with the same data, as such splitting this out into separate Update calls is probably unnecessary imo.

Regarding the ID potentially changing, since we're retrieving the ID from the `Get` endpoint, as opposed to the Create/Update endpoint, which could be different - these should be consistent; given this is a low likelihood based on the way the Azure API design works - I'd suggest instead we can patch any relevant bugs with, should this affect any endpoints:

```
if d.IsNewResource() {
  resp, err := client.Get(resourceGroup, name)
  if err != nil {
    return fmt.Errorf("Error retrieving ID for [Name] %q (Resource Group %q): %+v)", resourceGroup, name, err)
  }
  d.SetId(*resp.ID)
}
```

WDYT?

vancluever: This all sounds sane. My comment was not implying that this should be held back :slightly_smiling_face: I did also notice though that other Azure resources were not sharing the same kind of upserting convention in the CRUD functions, at least in naming. Still LGTM!

PR #1152, opened 2018-04-23
WIP: Try to make NSG plans less verbose

sylr:
Attempt to solve https://github.com/terraform-providers/terraform-provider-azurerm/issues/858

This currently does not work. After a local change in a rule `plan` says there is no diff.

@tombuildsstuff I need your help here.

### DIFF
--- a/azurerm/data_source_network_security_group.go
+++ b/azurerm/data_source_network_security_group.go
@@ -1,8 +1,11 @@
 package azurerm
 
 import (
+	"bytes"
 	"fmt"
+	"log"
 
+	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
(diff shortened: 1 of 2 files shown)


azurerm/data_source_network_security_group.go
```diff
@@ -21,8 +24,17 @@ func dataSourceArmNetworkSecurityGroup() *schema.Resource {
 			"location": locationForDataSourceSchema(),
 
 			"security_rule": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
```
katbyte: The datasource should be a `TypeList` because a set is difficult to address:
list `security_rule.0.name ` vs set `security_rule.1400690338.name `

azurerm/resource_arm_network_security_group.go
```diff
@@ -37,6 +40,14 @@ func resourceArmNetworkSecurityGroup() *schema.Resource {
 				Type:     schema.TypeSet,
 				Optional: true,
 				Computed: true,
+				Set: func(v interface{}) int {
```
katbyte: Because the hash function only looks at `direction` and `priority` any other changes will be ignored by terraform.

PR #1164, opened 2018-04-25
r/keyvault: increasing the availability checks

tombuildsstuff:
During some edge case testing I noticed that the Key Vault itself is available, however some of the external tests fail as the DNS wasn't fully available. This PR adds a 30s delay before checking, and then bumps the number of continuous target occurrences to correct this.

```
$ acctests azurerm TestAccAzureRMKeyVaultSecret_importComplete
=== RUN   TestAccAzureRMKeyVaultSecret_importComplete
go --ver--- PASS: TestAccAzureRMKeyVaultSecret_importComplete (166.51s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	166.533s
```


tombuildsstuff:
All the Key Vault + Key Vault Certificate/Key/Secret tests pass:

screen shot 2018-04-25 at 12 01 24

PR #1180, opened 2018-04-27
DNS Zone data source can be retrieved by name only.

### LINKED ISSUE #1178: Using Data DNS Zone to find a zone without knowing it's resource group.
Hey y'all.

We want to use the dns zone data source to find a dns zone with a specified name in order to add records to that zone instead of forcing the creation of a new dns zone in our templates and adding records to that zone. We can do this with the gcp and aws providers now and would like to add that functionality to the azure provider.

If a resource group name is [not specified](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/data_source_dns_zone.go#L49-L51), the dns zone data source will use the [zones client to list](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/vendor/github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2016-04-01/dns/zones.go#L264-L267) zones in the resource groups of your subscription and return the first that matches the provided zone name. Would y'all accept a PR for that functionality?

genevieve:
- Use resource group name if provided, otherwise, use findZone.
- findZone will get the resource groups in your subscription and then
check the zones in each resource group till it finds the first with
a matching dns zone name.
- Could not use the zones client ListComplete api call because
a Zone does not have any properties related to the Resource Group
it is a part of. We have to loop over the list of resource groups
and then their zones so that we can have both the zone and the resource
group name to return and set on the data source.
- Docs update.

Fixes #1178.


katbyte:
On further inspection wrapping wasn't needed. Merged master and tests pass:
screen shot 2018-05-19 at 00 25 58

katbyte:
Hey @genevieve ,

Just wanted to let you know we have released [v1.6.0](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/CHANGELOG.md#160-may-24-2018) of the provider allowing dns zones to be looked up by name only 🙂

PR #1187, opened 2018-05-01
Update loadbalancer.html.markdown

piotrgo:
Removed the note about availability of Standard SKU for LB, as it's now in GA as described in docs https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-standard-overview


piotrgo:
@tombuildsstuff I am not sure where to take this discussion, but I'd be tempted to set "Standard" SKU as the default as obviously it's newer and better, although not free anymore.

Issue #1004, opened 2018-03-20
Update Provider resources for MySql and PostgreSQL to support new pricing tier


jeffreyCline:
For more information on the change see the [announcement](https://azure.microsoft.com/en-gb/blog/updates-to-azure-database-for-mysql-and-azure-database-for-postgresql/) on the Microsoft Azure Blog.

metacpp:
@jeffreyCline 

The new SDK contains the GA APIs now:
https://github.com/Azure/azure-sdk-for-go/releases/tag/v14.6.0

metacpp:
@jeffreyCline I updated the title of this issue a common one.

jeffreyCline:
Currently waiting on Hotfix version of the v12.5.0-Beta Go SDK, because upgrading to v14.6.0 breaks the Provider code.

PR #1190, opened 2018-05-02
Update PostgreSQL to GA 2017-12-01 API

jeffreyCline:
Upgrade the PostgreSQL resource the adopt the GA API 2017-12-01. This PR will resolve issue #1004 .


jeffreyCline:
[Test Pass Results](https://gist.github.com/jeffreyCline/715466e2f9b3716d80c6e971ac4cf202)

azurerm/resource_arm_postgresql_server.go
```diff
@@ -70,7 +81,18 @@ func resourceArmPostgreSQLServer() *schema.Resource {
...
 							ValidateFunc: validation.StringInSlice([]string{
 								string(postgresql.Basic),
-								string(postgresql.Standard),
+								string(postgresql.GeneralPurpose),
+								string(postgresql.MemoryOptimized),
+							}, true),
+							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+						},
+
+						"family": {
```
JunyiYi: What is the relationship between this and "sku.name"? Because I found that "sku.name" also contains "Gen4" or "Gen5".

tombuildsstuff:
@jeffreyCline FYI I'm going to rebase this PR on top of master, since the Diff is showing changes which are already in master (but aren't in this branch)

tombuildsstuff:
@jeffreyCline actually in testing this I've found a couple of bugs in this and MySQL regarding SKU's - I'll push a couple of commits to enable SKU changes without forcing a new resource - which is supported for some options:

screen shot 2018-05-22 at 20 08 32

tombuildsstuff:
Re-tested the upgrade path from both MySql & Postgresql -> new version and they work as expected. It's also possible to update the Capacity as in the Portal - and there's tests for this.

tombuildsstuff:
MySQL tests pass:

screen shot 2018-05-22 at 21 03 31

tombuildsstuff:
PostgreSQL tests pass:

screen shot 2018-05-23 at 10 55 06

PR #1191, opened 2018-05-02
scheduler_job: Added storage queue support

katbyte:


azurerm/resource_arm_scheduler_job.go
```diff
@@ -829,6 +912,27 @@ func flattenAzureArmSchedulerJobActionRequest(d *schema.ResourceData, blockName
...
+		block["storage_queue_name"] = *v
+	}
+	if v := qm.Message; v != nil {
+		block["message"] = *v
+	}
+
+	//sas_token is not returned by the API
+	if v, ok := d.GetOk(fmt.Sprintf("%s.0.sas_token", blockName)); ok {
+		block["sas_token"] = v.(string)
+	}
```
tombuildsstuff: it should be fine to not set this field?

katbyte: Its the same as with web, if i don't pull the value forward there is always a diff:

```
action_storage_queue.0.sas_token: "" => "$SASTOKEN"
```

katbyte:
builds on #1172

PR #1197, opened 2018-05-03
r/kubernetes_cluster: info on unexpected errors

tombuildsstuff:
This PR cleans up a few rough edges in the Kubernetes resource:

- Whilst using the Kubernetes Cluster resource, I noticed there's an undocumented regex on the Agent Pool names - as such this PR adds validation to that affect (needs to start with a lower-case char, and be up to 12 chars of only a-z0-9. At the moment this fails with the unhelpful error message:

```
* azurerm_kubernetes_cluster.test: containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="" Message=""
```

- Documenting there's a known issue with routing on nodes attached to an internal subnet
- Documenting how to use the outputs with the Kubernetes Provider

### DIFF
--- a/website/docs/r/kubernetes_cluster.html.markdown
+++ b/website/docs/r/kubernetes_cluster.html.markdown
@@ -119,7 +119,7 @@ The following arguments are supported:
 * `os_type` - (Optional) The Operating System used for the Agents. Possible values are `Linux` and `Windows`.  Changing this forces a new resource to be created. Defaults to `Linux`.
 * `vnet_subnet_id` - (Optional) The ID of the Subnet where the Agents in the Pool should be provisioned. Changing this forces a new resource to be created.
 
-~> **NOTE:** There's a known issue where Agents connected to an Internal Network (e.g. on a Subnet) have their network routing configured incorrectly; such that Pods cannot be located across nodes. This is a bug in the Azure API - and will be fixed there in the future.
+~> **NOTE:** There's a known issue where Agents connected to an Internal Network (e.g. on a Subnet) have their network routing configured incorrectly; such that Pods cannot communicate across nodes. This is a bug in the Azure API - and will be fixed there in the future.
 
 `service_principal` supports the following:
 
--- a/website/docs/r/kubernetes_cluster.html.markdown
+++ b/website/docs/r/kubernetes_cluster.html.markdown
@@ -119,6 +119,8 @@ The following arguments are supported:
 * `os_type` - (Optional) The Operating System used for the Agents. Possible values are `Linux` and `Windows`.  Changing this forces a new resource to be created. Defaults to `Linux`.
 * `vnet_subnet_id` - (Optional) The ID of the Subnet where the Agents in the Pool should be provisioned. Changing this forces a new resource to be created.
 
+~> **NOTE:** There's a known issue where Agents connected to an Internal Network (e.g. on a Subnet) have their network routing configured incorrectly; such that Pods cannot be located across nodes. This is a bug in the Azure API - and will be fixed there in the future.
+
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
@dcaro as discussed yesterday, would you mind confirming the statement about Subnets is correct?

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -148,6 +150,19 @@ The following attributes are exported:
 
   * `password` - A password or token used to authenticate to the Kubernetes cluster.
 
+-> **NOTE:** You can use this It's possible to use this information with [the Kubernetes Provider](/docs/providers/kubernetes/index.html) like so:
```
bflad: Minor typo here 😄  `You can use this It's possible to use`

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -119,6 +119,8 @@ The following arguments are supported:
 * `os_type` - (Optional) The Operating System used for the Agents. Possible values are `Linux` and `Windows`.  Changing this forces a new resource to be created. Defaults to `Linux`.
 * `vnet_subnet_id` - (Optional) The ID of the Subnet where the Agents in the Pool should be provisioned. Changing this forces a new resource to be created.
 
+~> **NOTE:** There's a known issue where Agents connected to an Internal Network (e.g. on a Subnet) have their network routing configured incorrectly; such that Pods cannot be located across nodes. This is a bug in the Azure API - and will be fixed there in the future.
```
dcaro: There's a known issue where Agents connected to a Network (e.g. on a Subnet) not provisioned as part of the creation of the AKS cluster have their network routing configured incorrectly; such that Pods cannot communicate across nodes.

tombuildsstuff: so does the issue only manifest when attaching an existing cluster to an internal subnet? at this time the resource requires a recreation for this kind of change

tombuildsstuff: Since the `vnet_subnet_id` is ForceNew, this can only affect New clusters from within Terraform; so I think we're covered on that front. I'll update to include the comment about communication 👍

exceptorr:
@tombuildsstuff Hello!
> Documenting there's a known issue with routing on nodes attached to an internal subnet

Could you please share more info (maybe link to AKS issue tracker?) We've run into issue that looks similar - just want to ensure this is the same as issue described in Terraform's documentation.

Thanks!

tombuildsstuff:
👋 @exceptorr 

>> Documenting there's a known issue with routing on nodes attached to an internal subnet

> Could you please share more info (maybe link to AKS issue tracker?) We've run into issue that looks similar - just want to ensure this is the same as issue described in Terraform's documentation.

This came up in a conversation with @dcaro who should be able to give more information about it; I believe this may be fixed in the new non-Preview API version but I may be wrong, Damien should be able to give more information.

Thanks!

dcaro:
Thanks @exceptorr It was the case during the preview of AKS, with the GA of AKS last week using custom VNET is now supported as per https://docs.microsoft.com/en-us/azure/aks/networking-overview.
@tombuildsstuff we should update the documentation accordingly.

tombuildsstuff:
@dcaro I’m assuming that’s still going to be an issue for us until we upgrade to the GA API? I’ve asked the SDK team if they can ship this in the next release, which we can upgrade to when that’s out

dcaro:
Yes, we'll have to wait until the next release of the SDK.

PR #1199, opened 2018-05-04
Introduce the new HDInsight resource

### LINKED ISSUE #37: Feature Request: HDinsight Cluster Deployment
_This issue was originally opened by @retheshnair as hashicorp/terraform#11599. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi there,
I was trying to automate the deployment the HDinsight Cluster Deployment . I could not find terraform configuration to natively support HDinsight  Cluster  Deployment .Currently We are getting it deploy via ARM template + terraform based automation, which is not clean. It will nice to have HDinsight  Cluster  Deployment native support on terraform

JunyiYi:
- This code change is based on a auto-generated code, so it is a lot different than other resources, especially when expand and flatten fields.
- Some of the fields are manually tuned to be more logical.
- The update is also different from other resources, it requires invocations to other APIs.

This PR also resolves #37 .


azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+					"storm",
+					"spark",
+				}, true),
+				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+			},
+			"cluster_version": {
+				Required: true,
+				ForceNew: true,
+				Type:     schema.TypeString,
+			},
```
tombuildsstuff: I think it'd make sense to combine these into a `cluster` block:

```
cluster {
  type = "hadoop"
  version = "x.y.z"
}
```

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
+package azurerm
+
+// Code based on the terraform.provider plugin by Microsoft (R) AutoRest Code Generator.
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/services/hdinsight/mgmt/2015-03-01-preview/hdinsight"
```
tombuildsstuff: is there a stable version of this API available? This API is over 3 years old and still in Preview

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+			},
+			"component_version": {
+				Optional: true,
+				ForceNew: true,
+				Type:     schema.TypeMap,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"connectivity_endpoints": {
```
tombuildsstuff: can we move all of the computed fields to the end of the schema, we do it in the other resources to make it simpler to read

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+							Type:     schema.TypeString,
+						},
+					},
+				},
+			},
+			"location": locationSchema(),
+			"name": {
+				Required: true,
+				ForceNew: true,
+				Type:     schema.TypeString,
```
tombuildsstuff: in general we order this by Required fields, then Optional and then Computed - can we update to match?

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+			},
+			"os_type": {
+				Required: true,
+				ForceNew: true,
+				Type:     schema.TypeString,
+				ValidateFunc: validation.StringInSlice([]string{
+					"Linux",
+					"Windows",
```
tombuildsstuff: Windows support is EOL'd and being fully removed in July, so let's just remove it for the moment

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"password": {
+										Optional: true,
+										ForceNew: true,
+										Type:     schema.TypeString,
+									},
+									"ssh_keys": {
+										Required: true,
+										ForceNew: true,
```
tombuildsstuff: both Password and SSH keys are required? in general one of these is, but not both?

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+								Schema: map[string]*schema.Schema{
+									"name": {
+										Required: true,
+										ForceNew: true,
+										Type:     schema.TypeString,
+									},
+									"parameters": {
+										Required: true,
+										ForceNew: true,
+										Type:     schema.TypeString,
```
tombuildsstuff: shouldn't this be a TypeMap which gets stitched together?

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+						"linux_os_profile": {
+							Optional: true,
+							ForceNew: true,
+							Type:     schema.TypeList,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"password": {
+										Optional: true,
+										ForceNew: true,
```
tombuildsstuff: this won't be returned by the API - as such `linux_os_profile` will need to be a Set rather than a List with a hash function excluding this field; in addition this needs to be `Sensitive: true`

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+							Type:     schema.TypeList,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"id": {
+										Required: true,
+										ForceNew: true,
+										Type:     schema.TypeString,
+									},
+									"subnet": {
```
tombuildsstuff: rather than requiring the Subnet Name and the Virtual Network ID - let's instead accept a Subnet ID and parse this information out of the URI

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+					Schema: map[string]*schema.Schema{
+						"linux_os_profile": {
+							Optional: true,
+							ForceNew: true,
+							Type:     schema.TypeList,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"password": {
+										Optional: true,
```
tombuildsstuff: (as above) this needs to be `Sensitive: true` and `worker_node` needs to be a Set

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+									"id": {
+										Required: true,
+										ForceNew: true,
+										Type:     schema.TypeString,
+									},
+									"subnet": {
+										Required: true,
+										ForceNew: true,
+										Type:     schema.TypeString,
+									},
```
tombuildsstuff: as above - let's instead pass in the `subnet_id` and parse the components out that we need. this means we can also fail with a more helpful error message should the resources not be found

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+							},
+						},
+						"min_instance_count": {
+							Optional:     true,
+							Default:      1,
+							ForceNew:     true,
+							Type:         schema.TypeInt,
+							ValidateFunc: validation.IntBetween(1, 32),
+						},
+						"script_actions": {
```
tombuildsstuff: is this just webhooks? if not does `name` want to be called `command`?

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+										ForceNew: true,
+										Type:     schema.TypeString,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			"security_profile": {
```
tombuildsstuff: I believe this is Windows specific?

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+			},
+			"tags": tagsSchema(),
+			"tier": {
+				Optional: true,
+				ForceNew: true,
+				Type:     schema.TypeString,
+				Default:  "standard",
+				ValidateFunc: validation.StringInSlice([]string{
+					"standard",
+					"premium",
```
tombuildsstuff: are there Constants/Enums in the SDK we can use for this?

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+	if paramValue, paramExists := d.GetOk("os_type"); paramExists {
+		parameters.Properties.OsType = hdinsight.OSType(paramValue.(string))
+	}
+	if paramValue, paramExists := d.GetOk("tier"); paramExists {
+		parameters.Properties.Tier = hdinsight.Tier(paramValue.(string))
+	}
+	parameters.Properties.ClusterDefinition = &hdinsight.ClusterDefinition{}
+	if paramValue, paramExists := d.GetOk("cluster_type"); paramExists {
+		parameters.Properties.ClusterDefinition.Kind = utils.String(paramValue.(string))
+	}
```
tombuildsstuff: can we update this to match the convention:

```
# ..
osType := d.Get("os_type").(string)
parameters := hdinsight.ClusterCreateParametersExtended{
  Properties:  &hdinsight.ClusterCreateProperties{
    OsType: hdinsight.OSType(osType.(string)),
    # ..
  },
  Tags: expandTags(tags),
}
```

azurerm/resource_arm_hd_insight_clusters.go
```diff
@@ -0,0 +1,918 @@
...
+			parametersPropertiesClusterDefinitionComponentVersion := utils.String(tmpParamItemOfComponentVersion.(string))
+			tmpParamOfComponentVersion[tmpParamKeyOfComponentVersion] = parametersPropertiesClusterDefinitionComponentVersion
+		}
+		parameters.Properties.ClusterDefinition.ComponentVersion = tmpParamOfComponentVersion
+	}
+	tmpParamOfConfigurations := make(map[string]interface{})
+	tmpParamOfGatewayConfigurations := make(map[string]interface{})
+	tmpParamOfGatewayConfigurations["restAuthCredential.isEnabled"] = true
+	tmpParamOfGatewayConfigurations["restAuthCredential.username"] = d.Get("restauth_username")
+	tmpParamOfGatewayConfigurations["restAuthCredential.password"] = d.Get("restauth_password")
```
tombuildsstuff: can we parse this out in a separate expand and flatten function? Redis has good examples of this with the configuration

azurerm/resource_arm_hd_insight_clusters.go
JunyiYi: Taking a look at the [go sdk](https://github.com/Azure/azure-sdk-for-go/tree/master/services/hdinsight), this seems to be the only API.

azurerm/resource_arm_hd_insight_clusters.go
JunyiYi: This is base on an auto-generation tool which handles all the marshalling and unmarshalling stuffs.

azurerm/resource_arm_hd_insight_clusters.go
JunyiYi: This is base on an auto-generation tool which handles all the set and get method. And I don't see any benefits to extract to a lot of local methods just to flatten or expand them. It is hard to see the overall structure.

azurerm/resource_arm_hd_insight_clusters.go
JunyiYi: I think only SSH is required, because I tried once without it, and the API call failed.

azurerm/resource_arm_hd_insight_clusters.go
JunyiYi: Probably, I will remove it because I cannot find any configurations in the portal when creating a new HDInsight.

azurerm/resource_arm_hd_insight_clusters.go
JunyiYi: No, this is used to execute scripts on a specific node. For example, 
```
{
  name = "Install Git"
  uri = "http://shared-position/installgit.sh"
  parameters = "--arg1 val1 --arg2 val2"
}
```

vendor/vendor.json
```diff
@@ -184,6 +178,12 @@
 			"version": "v15.1.0",
 			"versionExact": "v15.1.0"
 		},
+		{
+			"checksumSHA1": "lm0tvCajScn5lWy5FJMw4V98BHQ=",
+			"path": "github.com/Azure/azure-sdk-for-go/services/preview/hdinsight/mgmt/2015-03-01-preview/hdinsight",
+			"revision": "4650843026a7fdec254a8d9cf893693a254edd0b",
+			"revisionTime": "2018-05-04T19:14:26Z"
```
tombuildsstuff: can we also pull this from the same Git tag as everything else?

azurerm/resource_arm_hd_insight_clusters.go
tombuildsstuff: Yes in general we group related fields; such that it provides a better user-experience; which is why we don't 1:1 map the SDK

azurerm/resource_arm_hd_insight_clusters.go
tombuildsstuff: Consistency with every other Provider - it also has the benefit of grouping this such that it's easier to understand

azurerm/resource_arm_hd_insight_clusters.go
JunyiYi: I still don't think it makes sense to group `cluster` related attributes into the `cluster` sub-field which duplicates itself and will cause confusion to the customer.

azurerm/resource_arm_hd_insight_clusters.go
JunyiYi: I don't think "keep with every other provider" persuade me:
1. The other provider code doesn't look good, it is hard to locate the figure out the structure in such a spaghetti-like code structure.
2. We will never make any steps forward if we always keep consistency with the legacy code.
3. Abstract to functions means reusing, and obviously it is not the case in other providers.

azurerm/resource_arm_hd_insight_clusters.go
JunyiYi: There's no difference for a user to pass it in a format of map or just a string. And sometimes a string would seems to be more intuitive (like `-DartifactName="Abcde" --enable-debug true`).

paultyng:
Is your intention that subsequent updates to this are by humans or by regeneration?

azurerm/resource_arm_hd_insight_clusters.go
paultyng: If it doesn't make sense to have the group, it doesn't make sense to have the prefix, so either it should be grouped due to the prefixing creating an implicit group or it should have the prefixes stripped if that doesn't make it lose its meaning.

azurerm/resource_arm_hd_insight_clusters.go
paultyng: > 1. The other provider code doesn't look good, it is hard to locate the figure out the structure in such a spaghetti-like code structure.

This is just an opinion and also not persuasive, probably many people on the team and a number of community contributors would beg to differ.

> 2. We will never make any steps forward if we always keep consistency with the legacy code.

What step forward are you making here? Code generation? Because other providers with code generation adhere to the standards, so this just seems to be a deficiency in your generation.

> 3. Abstract to functions means reusing, and obviously it is not the case in other providers.

Not just reusing, it offers the option of [unit testing](https://github.com/terraform-providers/terraform-provider-azurerm/blob/150e426d6d4c7ede6319b38761de5166fb3708ac/azurerm/tags_test.go#L71-L96), it also is good organization when you have multiple concerns.

An extremely large, complex, function is harder to digest for a human that a smaller one. Breaking it in to smaller, purpose built functions, is more clear. This function has multiple related but different things it is doing:

* Reading state and creating the SDK types
* Calling the SDK and handling the result

Calling the SDK and handling the SDK result seem like a reasonable thing for this function to do, converting resource state to an SDK type using all the associated type assertions is an unnecessary addition complicating the SDK interaction logic with a bunch of unrelated lines that the next person in this code will have to read through until they can figure out where the real action is starting. By separating out to the flatten/expand model, interaction and struct assignment is easily separated mentally.

paultyng: A lot of this comes down to the question of is this solely for auto-generated maintenance or do you intend to fix bugs or flow idiosyncrasies manually in these files?

tombuildsstuff:
Closing in favour of #1267

PR #1201, opened 2018-05-04
azurerm_cosmosdb_account: ip_range_filter input validation

### LINKED ISSUE #1192: Feature Request: Add Missing Options for azurerm_cosmosdb_account Firewall
The azurerm_cosmosdb_account resource currently has a parameter called `ip_range_filter` that allows you to restrict CosmosDB access to specific IPs or CIDR blocks but there are also 2 check box options in the UI that do not appear to be supported in the current resource:

`Allow access to Azure Services`

and

`Allow access to Azure Portal`

Is this functionality not exposed in the current version of the SDK?

katbyte:
- moved import tests to regular testing for better coverage & shorter test time
- added additional input validation and example for `ip_range_filter` property

fixes #1192


examples/cosmos-db/main.tf
```diff
@@ -18,10 +18,13 @@ resource "azurerm_cosmosdb_account" "db" {
 
     enable_automatic_failover = true
 
+    //set ip_range_filter to allow azure services (0.0.0.0) and azure portal.
+    ip_range_filter = "0.0.0.0,104.42.195.92,40.76.54.131,52.176.6.30,52.169.50.45,52.187.184.26"
```
tombuildsstuff: AFAIK this should only need to be `0.0.0.0`, as the Azure Portal IP Addresses will change depending on where you're located?

azurerm/resource_arm_cosmos_db_account.go
```diff
@@ -496,6 +500,8 @@ func resourceArmCosmosDBAccountRead(d *schema.ResourceData, meta interface{}) er
 		d.Set("write_endpoints", writeEndpoints)
 	}
 
+	// ListKeys returns a data structure containing a DatabaseAccountListReadOnlyKeysResult pointer
+	// implying that it also returns the read only keys, however this appears to not be the case
```
tombuildsstuff: Is this consistent across all database types? If it's there, it may be DB specific ¯\_(ツ)_/¯

katbyte: Seems to be the case, I added it quick and it failed for every test.

examples/cosmos-db/main.tf
katbyte: These are the IPs I got when I selected the option in the azure portal.

PR #1202, opened 2018-05-06
New Data Sources: `azurerm_key_vault` & `azurerm_key_vault_secret`

### LINKED ISSUE #606: azurerm_key_vault_secret as a data source?
We'd like to use secrets managed in an Azure Key Vault to provision the rest of our infrastructure. For example, database connection info. My instinct is that the "most correct" way to do that is to retrieve the secrets via a data source so they can be passed around within the HCL. 

As this data source doesn't exist yet, we currently work around it by using `az` commands to retrieve the secrets and pass them in as command line args, all in a wrapper script. This means that we have extra code to maintain and keep secure, and as our hierarchy of modules grows this will get in the way.

Does this idea break any terraform conventions or philosophies? Or would this secrets data source be valuable to others?

tombuildsstuff:
Key Vault Tests Pass:

```
$ acctests azurerm TestAccDataSourceAzureRMKeyVault_
=== RUN   TestAccDataSourceAzureRMKeyVault_basic
--- PASS: TestAccDataSourceAzureRMKeyVault_basic (220.18s)
=== RUN   TestAccDataSourceAzureRMKeyVault_complete
--- PASS: TestAccDataSourceAzureRMKeyVault_complete (194.67s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	414.874s
```

Key Vault Secret Tests Pass:

```
$ acctests azurerm TestAccDataSourceAzureRMKeyVaultSecret_
=== RUN   TestAccDataSourceAzureRMKeyVaultSecret_basic
--- PASS: TestAccDataSourceAzureRMKeyVaultSecret_basic (231.85s)
=== RUN   TestAccDataSourceAzureRMKeyVaultSecret_complete
--- PASS: TestAccDataSourceAzureRMKeyVaultSecret_complete (216.68s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	448.561s
```

Fixes #606


azurerm/data_source_key_vault.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+		d.Set("vault_uri", props.VaultURI)
+	}
+
+	flattenAndSetTags(d, resp.Tags)
+
+	return nil
+}
+
+func flattenKeyVaultDataSourceSku(sku *keyvault.Sku) []interface{} {
```
katbyte: Out of curiosity, is there a reason this is a list/block instead of a string value?

website/docs/d/key_vault.html.markdown
```diff
@@ -0,0 +1,70 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The Vault ID.
+
+* `vault_uri` - The URI of the vault for performing operations on keys and secrets.
+
+* `location` - The Azure Region in which the Key Vault exists
```
katbyte: The only line that doesn't end with a `.`

azurerm/data_source_key_vault.go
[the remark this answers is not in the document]
tombuildsstuff: It's also possible to set some additional fields here, but we don't expose them

mikaelkrief:
Hi,
I will be very interrested by this feature on my client.
Do you have an idea where it will be released ?
Thanks

tombuildsstuff:
@mikaelkrief it'll be released as a part of v1.5.0 very soon :)

MikeRosTX:
Looks like a larger audience might be need this release.
https://www.facebook.com/groups/devops.engineers/permalink/850242665167530/

PR #1207, opened 2018-05-06
New Resource: `azurerm_virtual_machine_data_disk_attachment`

### LINKED ISSUE #795: Feature Request: new resource to attach disks to virtual machines
An issue that I have run into recently is the inability to use a count function when adding data disks to a VM storage configuration.  I can create the data disks using a count function outside of the resource, but I have to specify multiple storage_data_disk items in the azurerm_virtual_machine resource to get them attached.  Hashicorp is considering ways to added a count function within a resource (using the foreach terminology in this [thread](https://github.com/hashicorp/terraform/issues/7034)).  

Rather than waiting for that approach to shake out, I would like to request that resources like azurerm_managed_disk support a mapping function to attach to a azurerm_virtual_machine resource, or that a new mapping resource be created that will map the attachment of the disk to the virtual machine, something like azurerm_managed_disk_attach.  Here's an example of how that would work:

```hcl
resource azurerm_managed_disk_attach {
  name  =  "disk-attach-${count.index}"
  count  =  "${var.number-of-disks}"
  disk_id  =  "${element(azurerm_managed_disk.mydisk.*.id, count.index)}"
  vm_id  =  "${azurerm_virtual_machine.mymachine.id}"
}
```

tombuildsstuff:
This PR creates a new resource which can be used for attaching disks to Virtual Machine. The acceptance tests for this need some more thought and this needs some extensive testing, in general.

After spending _a lot of time_ investigating/thinking about this - I've ended up adapting this resource such that it only supports attaching existing Managed Disks. This makes supporting this resource considerably simpler (since supporting creating new Managed Disks involves manipulating the `create_option` field); unfortunately it appears this approach isn't possible for Unmanaged Disks [due to this bug](https://github.com/Azure/azure-sdk-for-go/issues/1829#issuecomment-393282772) - as such I've dropped support for Unmanaged Disks at this time.

Fixes #795


azurerm/provider.go
```diff
@@ -108,118 +108,119 @@ func Provider() terraform.ResourceProvider {
...
-			"azurerm_traffic_manager_endpoint":            resourceArmTrafficManagerEndpoint(),
-			"azurerm_traffic_manager_profile":             resourceArmTrafficManagerProfile(),
-			"azurerm_virtual_machine_extension":           resourceArmVirtualMachineExtensions(),
-			"azurerm_virtual_machine":                     resourceArmVirtualMachine(),
-			"azurerm_virtual_machine_scale_set":           resourceArmVirtualMachineScaleSet(),
-			"azurerm_virtual_network":                     resourceArmVirtualNetwork(),
-			"azurerm_virtual_network_gateway":             resourceArmVirtualNetworkGateway(),
-			"azurerm_virtual_network_gateway_connection":  resourceArmVirtualNetworkGatewayConnection(),
-			"azurerm_virtual_network_peering":             resourceArmVirtualNetworkPeering(),
+			"azurerm_application_gateway":                  resourceArmApplicationGateway(),
```
metacpp: why did you change so many lines?

azurerm/resource_arm_virtual_machine_data_disk_attachment.go
```diff
@@ -0,0 +1,287 @@
...
+		expandedDisk.DiskSizeGB = utils.Int32(int32(v.(int)))
+	}
+
+	disks := *virtualMachine.StorageProfile.DataDisks
+	if !d.IsNewResource() {
+		// find the existing disk, remove it from the array
+		dataDisks := make([]compute.DataDisk, 0)
+		for _, dataDisk := range disks {
+			if !strings.EqualFold(*dataDisk.Name, *expandedDisk.Name) {
+				disks = append(disks, dataDisk)
```
metacpp: It seems that dataDisks is never used since initialization. I guess you were using it store the left disks. Then I think you should do:

dataDisks = append(dataDisks, dataDisk)

azurerm/provider.go
JunyiYi: Because one space is added to align with each other.

tombuildsstuff:
> I have some concerns of introducing multiples ways to do one thing. It might cause updating issues similar to azurerm_network_security_rule.

@JunyiYi indeed, this is why this is still WIP whilst I've been thinking about how's best to implement this/chatting internally about it. We should be keeping both resources actually - the `nested` resource is beneficial as it allows the user to explicitly say "I only want these disks attached" vs the `attachment` resource for "I'm not concerned with what other disks are attached, just attach this given disk".

After mulling on this for a while, I'm coming around to the idea of forcing users to create a separate managed disk ahead of time, which allows us to remove the `create_option` field altogether (which is a *huge* pain-point), and also handles disk cleanup (by the `azurerm_managed_disk` resource being deleted); which would allow us to infer the details from the disk, making the attachment resource look something like this:

```
resource "azurerm_managed_disk" "test" {
  name                 = "tomdev-disk25"
  location             = "${azurerm_resource_group.test.location}"
  resource_group_name  = "${azurerm_resource_group.test.name}"
  create_option        = "Empty"
  disk_size_gb         = 10
}

resource "azurerm_virtual_machine_data_disk_attachment" "second" {
  name               = "some-name" # perhaps we could infer this from the Disk Name?
  virtual_machine_id = "${azurerm_virtual_machine.windows.id}"
  managed_disk_id    = "${azurerm_managed_disk.test.id}"
  lun                = 33
  caching            = "ReadOnly"
}
```

rather than this:

```
resource "azurerm_managed_disk" "test" {
  name                 = "tomdev-disk25"
  location             = "${azurerm_resource_group.test.location}"
  resource_group_name  = "${azurerm_resource_group.test.name}"
  create_option        = "Empty"
  disk_size_gb         = 10
}

resource "azurerm_virtual_machine_data_disk_attachment" "existing_disk" {
  name               = "existing-disk"
  virtual_machine_id = "${azurerm_virtual_machine.windows.id}"
  managed_disk_id    = "${azurerm_managed_disk.test.id}"
  managed_disk_type  = "Standard_LRS"
  disk_size_gb       = "${azurerm_managed_disk.test.disk_size_gb}"
  lun                = 33
}
```

The outstanding question is how we'd achieve the same thing with unmanaged disks, which I'm looking into at the moment :)

azurerm/provider.go
tombuildsstuff: on that note: you can use `?w=1` to ignore whitespace changes in reviews in Github, it's really handy for things like this :)

azurerm/disks.go
```diff
@@ -0,0 +1,40 @@
+package azurerm
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+type UnmanagedDiskMetadata struct {
```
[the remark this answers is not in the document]
tombuildsstuff: they should be private, however this is a WIP before refactoring (and placing this somewhere in `helpers/`)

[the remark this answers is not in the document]
tombuildsstuff: (also, we tend to rely directly on the SDK Structs rather than composing them ourselves, since this is information which should be provided by the SDK but isn't at this point)

rohrerb:
Excited to see this merged. :D

tombuildsstuff:
I've been doing some initial exploratory testing against this and it seems ok, but it needs some more (in particular verifying disks created inline -> split out are handled as expected, which I'll pick up when I'm back). I've also pushed a WIP commit which should prevent disk attachments created using the independent resource from being impacted when changes are made to the parent VM resource which uses the value of the computed fields, but this needs additional testing.

rohrerb:
Hi Tom,

Any chance this will get merged soon?  Need any help testing?

-Brandon

tombuildsstuff:
👋 @rohrerb I've been out for a couple of weeks, so I've not been working on this; but I'm working through the edge-case testing at the moment and hope to finish that this week. That said - we're definitely interested to hear your feedback if you're able to test this :)

Thanks!

azurerm/resource_arm_virtual_machine_data_disk_attachment.go
```diff
@@ -0,0 +1,286 @@
...
+		if utils.ResponseWasNotFound(virtualMachine.Response) {
+			return fmt.Errorf("Virtual Machine %q (Resource Group %q) was not found", virtualMachineName, resourceGroup)
+		}
+
+		return fmt.Errorf("Error loading Virtual Machine %q (Resource Group %q): %+v", virtualMachineName, resourceGroup, err)
+	}
+
+	dataDisks := make([]compute.DataDisk, 0)
+	for _, dataDisk := range *virtualMachine.StorageProfile.DataDisks {
+		// since this field isn't (and shouldn't be) case-sensitive; we're deliberately not using `strings.Equals`
```
katbyte: Did you mean `strings.EqualFold()`

website/docs/r/virtual_machine.html.markdown
```diff
@@ -1,7 +1,7 @@
 ---
 layout: "azurerm"
 page_title: "Azure Resource Manager: azurerm_virtual_machine"
-sidebar_current: "docs-azurerm-resource-compute-virtual-machine"
+sidebar_current: "docs-azurerm-resource-compute-virtual-machine-x"
```
katbyte: Was the `-x` supposed to be added?

tombuildsstuff: Yep, there's an incorrect highlight on the sidebar at the moment

PR #1211, opened 2018-05-07
Added Import support to Redis Cache/Firewall Rule

tombuildsstuff:
This was a feature request that came up IRL at Microsoft Build

This also makes the `maxclients` field Computed, since it's not set-able anymore; and adds a note about the storage connection string bug in https://github.com/Azure/azure-rest-api-specs/issues/3037


tombuildsstuff:
Tests pass:

screen shot 2018-05-11 at 10 50 12

PR #1215, opened 2018-05-08
App Service Plan: updating the Linux examples

### LINKED ISSUE #1213: azurerm_app_service_plan: docs for webapp on linux
The docs under https://www.terraform.io/docs/providers/azurerm/r/app_service_plan.html state the following example:

      name     = "api-rg-pro"
      location = "West Europe"
    }

    resource "azurerm_app_service_plan" "test" {
      name                = "api-appserviceplan-pro"
      kind                = "Linux"
      reserved            = true
    
      sku {
        tier = "Standard"
        size = "S1"
      }
    }

In the same time on the same page we find the description, which states:

> properties supports the following:
> reserved - (Optional) Is this App Service Plan Reserved. Defaults to false.

For me the example does not work, instead it shows the following message:
> Error: azurerm_app_service_plan.appserviceplan-alpha: : invalid or unknown key: reserved

The correct way to specify the reserved would be:

    resource "azurerm_app_service_plan" "test" {
        ...
        properties {
          reserved = true
        }

tombuildsstuff:
Fixes #1213

### DIFF
--- a/azurerm/resource_arm_app_service_plan_test.go
+++ b/azurerm/resource_arm_app_service_plan_test.go
@@ -293,6 +293,10 @@ resource "azurerm_app_service_plan" "test" {
     tier = "Basic"
     size = "B1"
   }
+
+  properties {
+    reserved = true
+  }
 }
 `, rInt, location, rInt)
 }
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
Tests pass:

screen shot 2018-05-08 at 09 04 48

PR #1232, opened 2018-05-14
redis firewall rule: fix max_clients in tests

katbyte:


katbyte:
tests pass:
screen shot 2018-05-14 at 11 35 47

PR #1246, opened 2018-05-16
App Service: moving `ip_restrictions` within `site_config` block

tombuildsstuff:
- Combining the Site Config from App Service (Data Sources & Resources) & App Service Slots
- Added support for `ip_restriction` to App Service Slots
- Moved `ip_restriction` within the `site_config` block - since it's Site specific

### DIFF
--- a/azurerm/data_source_app_service.go
+++ b/azurerm/data_source_app_service.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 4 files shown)


tombuildsstuff:
Tests pass:

screen shot 2018-05-23 at 13 54 02

tombuildsstuff:
@radeksimko support for `ip_restrictions` was only merged recently (and hasn't made it out to a release yet) - as such I don't feel we need to do a migration for that?

Issue #1249, opened 2018-05-17
Low priority VM Scale Set support


CoRfr:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

This feature is still in preview but it would be quite useful to have it supported in terraform.
https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-use-low-priority

### New or Affected Resource(s)

* azurerm_virtual_machine_scale_set

### Potential Terraform Configuration

```hcl
resource "azurerm_virtual_machine_scale_set" "scaleset" {
  name                = "autoscalewad"
  location            = "${azurerm_resource_group.rg.location}"
  resource_group_name = "${azurerm_resource_group.rg.name}"
  upgrade_policy_mode = "Manual"
  overprovision       = true
  depends_on          = ["azurerm_lb.lb", "azurerm_virtual_network.vnet"]
  priority            = "Low"
```

### References

* https://azure.microsoft.com/en-us/updates/public-preview-low-priority-vms-on-vm-scale-sets/
* https://github.com/ecalder6/azure-gaming/blob/master/LowPri.json

katbyte:
Hey @CoRfr ,

Just wanted to let you know we have released [v1.6.0](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/CHANGELOG.md#160-may-24-2018) of the provider containing your PR for this issue 🙂

PR #1250, opened 2018-05-17
Add 'priority' property for VM scale set

CoRfr:
Purpose is to support low-priority VMs, which should partially
address #1249.

This is simply adding a 'priority' parameter that can be set
to Low or Regular (=default).

The evictionPolicy is not supported yet as it was introduced
in a more recent version of the compute API (2018-04-01).

Cf https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-use-low-priority

### DIFF
--- a/azurerm/resource_arm_virtual_machine_scale_set.go
+++ b/azurerm/resource_arm_virtual_machine_scale_set.go
@@ -115,6 +115,16 @@ func resourceArmVirtualMachineScaleSet() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"priority": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(compute.Low),
+					string(compute.Regular),
+				}, true),
+			},
+
 			"os_profile": {
 				Type:     schema.TypeList,
 				Required: true,
@@ -651,6 +661,7 @@ func resourceArmVirtualMachineScaleSetCreate(d *schema.ResourceData, meta interf
 	updatePolicy := d.Get("upgrade_policy_mode").(string)
 	overprovision := d.Get("overprovision").(bool)
 	singlePlacementGroup := d.Get("single_placement_group").(bool)
+	priority := d.Get("priority").(string)
 
 	scaleSetProps := compute.VirtualMachineScaleSetProperties{
 		UpgradePolicy: &compute.UpgradePolicy{
@@ -661,6 +672,7 @@ func resourceArmVirtualMachineScaleSetCreate(d *schema.ResourceData, meta interf
 			StorageProfile:   &storageProfile,
 			OsProfile:        osProfile,
 			ExtensionProfile: extensions,
+			Priority:         compute.VirtualMachinePriorityTypes(priority),
 		},
 		Overprovision:        &overprovision,
 		SinglePlacementGroup: &singlePlacementGroup,
@@ -768,6 +780,7 @@ func resourceArmVirtualMachineScaleSetRead(d *schema.ResourceData, meta interfac
 
 		if profile := properties.VirtualMachineProfile; profile != nil {
 			d.Set("license_type", profile.LicenseType)
+			d.Set("priority", profile.Priority)
 
 			osProfile := flattenAzureRMVirtualMachineScaleSetOsProfile(d, profile.OsProfile)
 			if err := d.Set("os_profile", osProfile); err != nil {
--- a/azurerm/resource_arm_virtual_machine_scale_set_test.go
+++ b/azurerm/resource_arm_virtual_machine_scale_set_test.go
@@ -437,6 +437,26 @@ func TestAccAzureRMVirtualMachineScaleSet_overprovision(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMVirtualMachineScaleSet_priority(t *testing.T) {
+	resourceName := "azurerm_virtual_machine_scale_set.test"
+	ri := acctest.RandInt()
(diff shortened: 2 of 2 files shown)


azurerm/resource_arm_virtual_machine_scale_set_test.go
```diff
@@ -820,7 +854,7 @@ func testCheckAzureRMVirtualMachineScaleSetSinglePlacementGroup(name string, exp
 		}
 
 		if *resp.SinglePlacementGroup != expectedSinglePlacementGroup {
-			return fmt.Errorf("Bad: Overprovision should have been %t for scale set %v", expectedSinglePlacementGroup, name)
+			return fmt.Errorf("Bad: SinglePlacementGroup should have been %t for scale set %v", expectedSinglePlacementGroup, name)
```
katbyte: Thank you for fixing this too!

azurerm/resource_arm_virtual_machine_scale_set_test.go
```diff
@@ -437,6 +437,25 @@ func TestAccAzureRMVirtualMachineScaleSet_overprovision(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMVirtualMachineScaleSetDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVirtualMachineScaleSetExists("azurerm_virtual_machine_scale_set.test"),
+					testCheckAzureRMVirtualMachineScaleSetPriority("azurerm_virtual_machine_scale_set.test", "Low"),
```
katbyte: Unless there is a reason for us to query Azure directly could we change this to check the state? like:
```go
resource.TestCheckResourceAttr(resourceName, "azurerm_virtual_machine_scale_set.test. priority", "low"),
```

I know this isn't like many of the tests in the resource, but without good reason this is how we should be verifying the new property as it also checks the state.

CoRfr:
@katbyte I updated the test and also rebased.

azurerm/resource_arm_virtual_machine_scale_set_test.go
CoRfr: NP!

azurerm/resource_arm_virtual_machine_scale_set_test.go
```diff
@@ -812,6 +829,21 @@ func testCheckAzureRMVirtualMachineScaleSetOverprovision(name string) resource.T
 	}
 }
 
+func testCheckAzureRMVirtualMachineScaleSetPriority(name string, expectedPriority string) resource.TestCheckFunc {
```
katbyte: I think you forgot to remove this.

azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -115,6 +115,11 @@ func resourceArmVirtualMachineScaleSet() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"priority": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
```
katbyte: Is this possible to validate against the `VirtualMachinePriorityTypes` enum?

azurerm/resource_arm_virtual_machine_scale_set_test.go
```diff
@@ -437,6 +437,23 @@ func TestAccAzureRMVirtualMachineScaleSet_overprovision(t *testing.T) {
...
+	ri := acctest.RandInt()
+	config := testAccAzureRMVirtualMachineScaleSetPriorityTemplate(ri, testLocation())
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMVirtualMachineScaleSetDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+				Check:  resource.TestCheckResourceAttr(resourceName, "priority", "Low"),
```
katbyte: could we also check that the resource exists:
```go
					testCheckAzureRMVirtualMachineExists(resourceName, &afterCreate),
					resource.TestCheckResourceAttr(resourceName, "priority", "Low"),
				),
```

azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -115,6 +115,11 @@ func resourceArmVirtualMachineScaleSet() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"priority": {
+				Type:     schema.TypeString,
+				Optional: true,
```
djsly: could we add "ForceNew" here since a scale set cannot have a mix of Low and Regular VM. Unless Azure will update all pre-existing VM if you change your scaleset from `Regular` to `Low`?

azurerm/resource_arm_virtual_machine_scale_set_test.go
```diff
@@ -2577,6 +2609,90 @@ resource "azurerm_virtual_machine_scale_set" "test" {
...
+
+  storage_profile_image_reference {
+    publisher = "Canonical"
+    offer     = "UbuntuServer"
+    sku       = "16.04-LTS"
+    version   = "latest"
+  }
+}
+
+`, rInt, location, rInt, rInt, rInt, rInt, rInt)
```
djsly: can we reuse the first `rInt` using `%[1]d`, I cleaned up all the previous tests in my pending PR

katbyte:
tests pass (with 2 known failures)
screen shot 2018-05-24 at 09 34 57

Thank you for the contribution @CoRfr 🙂

Issue #760, opened 2018-01-26
Azure DB for PostgreSQL collation changes automatically (see #396)


ghost:
Hi there,

I've placed a note on Issue #396 noting that the proposed work-around is no longer effective, apparently due to Azure performing better validation.

Therefore, please re-visting this issue.

Cheers!

stuart-shelton-hmcts:
(To confirm, we're still using Terraform 0.10.5...)

stuart-shelton-hmcts:
Additionally, if a configuration block defines `pgsql_srv_name` (a field which the platform rejects if it contains any upper-case characters) as, say, `pg-localenv`, then having applied this configuration, a subsequent plan will show:
```
-/+ module.pgsql-repo.azurerm_postgresql_database.pgsql_db (new resource required)
<...>
      name:                         "pg_localenv_db" => "pg_localenv_DB" (forces new resource)
```
... noting both that the hyphens in the configuration and now transformed into underscores, that an extra three characters have been added, and that the capitalisation of these characters is inconsistent!

stuart-shelton-hmcts:
If the `srv_name` is updated to the apparently required value (with a '_DB' suffix) the error is still:

```
* azurerm_postgresql_server.pgsql_srv: postgresql.ServersClient#Create: Failure sending request: StatusCode=200 -- Original Error: Long running operation terminated with status 'Failed': Code="InvalidServerName" Message="Server name 'pg_localenv_DB' cannot be empty or null. It can only be made up of lowercase letters 'a'-'z', the numbers 0-9 and the hyphen. The hyphen may not lead or trail in the name."
```

stuart-shelton-hmcts:
Okay, ignore me - the '_DB'/'_db' thing was a local problem with a custom module we're using.

That just leaves the collation issue where 'en_US' becomes 'en-US', resulting in the original value appearing to be a change.

This is still present in terraform `0.11.2` with azurerm provider `1.1.0`.

cpressland:
We're seeing the same thing, when provisioning with this block:

```
resource "azurerm_postgresql_database" "devExample" {
  name = "example"
  resource_group_name = "${azurerm_resource_group.devWestEurope.name}"
  server_name = "${azurerm_postgresql_server.devExample.name}"
  charset = "UTF8"
  collation = "en-GB"
}
```

We recieve the following error:

```
[...]

Issue #396, opened 2017-10-04
Azure DB for PostgreSQL collation changes automatically.


piotrgo:
Hi, 

We've run into an issue with newly added Azure DB for Postgres. I am not sure whether it's just a misinformation in documentation, or an actual problem with Terraform implementing Azure DB.
DB's collation changes from en_US to en-US and Terraform want's to change it.

### Terraform Version
0.10.5

### Affected Resource(s)
azurerm_postgresql_database

### Terraform Configuration Files
```hcl
resource "azurerm_postgresql_database" "pgsql_db" {
  name                             = "${replace(var.pgsql_srv_name, "-", "_")}_db"
  resource_group_name = "${var.resource_group_name}"
  server_name                 = "${azurerm_postgresql_server.pgsql_srv.name}"
  charset                          = "${var.charset}"
  collation                        = "en_US"
}
```

### Initial Plan Output
```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  + module.pgsql-artifactory.azurerm_postgresql_database.pgsql_db
      id:                             <computed>
      charset:                        "UTF8"
      collation:                      "en_US"
      name:                           "dbname"
      resource_group_name:            "rgname"
      server_name:                    "srvname"

Plan: 1 to add, 0 to change, 0 to destroy.

------------------------------------------------------------------------
```

Apply runs without any problems, immediately after, another plan results with:

```
Terraform will perform the following actions:

-/+ module.pgsql-artifactory.azurerm_postgresql_database.pgsql_db (new resource required)
      id:                             "/subscriptions/ID" => <computed> (forces new resource)
      charset:                        "UTF8" => "UTF8"
      collation:                      "en-US" => "en_US" (forces new resource)
      name:                           "dbname" => "dbname"
      resource_group_name:"rgname" => "rgname"
      server_name:                "srvname" => "srvname"

Plan: 1 to add, 0 to change, 1 to destroy.

------------------------------------------------------------------------


```

### Expected Behavior
DB should be created with specified collation as per PGSQL documentation (underscore in collation type) and no further changes should be planned by Terraform

### Actual Behavior
DB is being created fine, but on subsequent plan it is being marked for a change from en_US to en-US

### Steps to Reproduce
Follow an example from the documentation for setting up Azure DB for PostgreSQL and Azure PostgreSQL DB with collation type set as per PGSQL documentation.

[...]

PR #1255, opened 2018-05-18
[Bug Fix] Fixes PostgreSQL collation issues #396 and #760

jeffreyCline:
**NOTE:** This PR is based on the `u-postgreslq` branch, once the `u-postgreslq` branch PR [#1090](https://github.com/terraform-providers/terraform-provider-azurerm/pull/1190) is merged into `master` this PR will only affect the following 4 files to fix the collation issues per #396 and #760 

* resource_arm_postgresql_database.go
* validators.go
* validators_test.go
* utils.go


katbyte:
Hi @jeffreyCline,

Thanks for fixing this 🙂 I am going to wait review this until the other 2 PRs are merged so the diff is only the 4 file to review.

tombuildsstuff:
@jeffreyCline I've rebased (well, merged since this was done previously) this PR so it now only contains the original changes :)

azurerm/resource_arm_postgresql_server.go
```diff
@@ -372,6 +372,8 @@ func expandAzureRmPostgreSQLServerSku(d *schema.ResourceData) *postgresql.Sku {
 	skus := d.Get("sku").([]interface{})
 	sku := skus[0].(map[string]interface{})
 
+	fmt.Printf("[SKU]: %s", sku)
```
tombuildsstuff: _minor_ can we remove this?

azurerm/resource_arm_postgresql_database.go
```diff
@@ -119,8 +120,7 @@ func resourceArmPostgreSQLDatabaseRead(d *schema.ResourceData, meta interface{})
 	d.Set("resource_group_name", resGroup)
 	d.Set("server_name", serverName)
 	d.Set("charset", resp.Charset)
-	d.Set("collation", resp.Collation)
-
+	d.Set("collation", utils.AzureRMNormalizeCollation(d.Get("collation").(string), "-", "_", -1))
```
tombuildsstuff: providing we split this out to a separate line (to make it more readable) - we should be able to just use a string replacement directly? e.g.

```
collation := strings.Replace(d.Get(resp.Collation, "-", "_", -1)
d.Set("collation", collation)
```

note: in general we shouldn't be retrieving any properties from the schema in the Read function - since they won't exist during Import and can lead to crashes.

azurerm/utils/utils.go
```diff
@@ -15,3 +17,8 @@ func Int64(input int64) *int64 {
 func String(input string) *string {
 	return &input
 }
+
+func AzureRMNormalizeCollation(input string, find string, replace string, count int) *string {
+	collation := strings.Replace(input, find, replace, count)
+	return &collation
+}
```
tombuildsstuff: (which means we don't need this method)

azurerm/resource_arm_postgresql_server.go
jeffreyCline: DOH! Sorry, that was part of my debug code... I forgot to remove it thanks for spotting it! :)

jeffreyCline:
I made all the requested changes, I also snuck a fix in for the charset issue... PostgreSQL had the same charset casing issues that MySQL did, except opposite.... PostgreSQL always returns uppercase, while MySQL always returns lowercase… go figure. I also added two test cases to validate the casing behavior, you can see the test results below.

[Test Results](https://gist.github.com/jeffreyCline/b295fe72ff4bb2c52c2faa4e33a59d4f)

tombuildsstuff:
Tests pass after adding the crash fix / switching to use the `.Properties` fields (which matches the Azure response):

```
$ acctests azurerm TestAccAzureRMPostgreSQLDatabase_
=== RUN   TestAccAzureRMPostgreSQLDatabase_importBasic
--- PASS: TestAccAzureRMPostgreSQLDatabase_importBasic (402.40s)
=== RUN   TestAccAzureRMPostgreSQLDatabase_basic
--- PASS: TestAccAzureRMPostgreSQLDatabase_basic (486.39s)
=== RUN   TestAccAzureRMPostgreSQLDatabase_charsetLowercase
--- PASS: TestAccAzureRMPostgreSQLDatabase_charsetLowercase (424.71s)
=== RUN   TestAccAzureRMPostgreSQLDatabase_charsetMixedcase
--- PASS: TestAccAzureRMPostgreSQLDatabase_charsetMixedcase (424.97s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	1738.507s
```

PR #1261, opened 2018-05-18
Change terraform version in requirements.

JunyiYi:


tombuildsstuff:
hi @JunyiYi 

Due to the way the Terraform Plugin system works the AzureRM Provider will currently work on any version of Terraform from v0.10 and above (although, this may/will change in future releases) - as such this change isn't correct and since it's a one-line change I'm going to close this PR.

Thanks!

PR #1262, opened 2018-05-18
Adjust the log messages to align with each other.

JunyiYi:


azurerm/resource_arm_network_security_rule.go
```diff
@@ -343,5 +343,5 @@ func resourceArmNetworkSecurityRuleDelete(d *schema.ResourceData, meta interface
 		return fmt.Errorf("Error waiting for the deletion of Network Security Rule %q (NSG %q / Resource Group %q): %+v", sgRuleName, nsgName, resGroup, err)
 	}
 
-	return err
+	return nil
```
tombuildsstuff: _minor_ this change is superfluous since err will be nil at this point (due to the if condition above)

PR #1265, opened 2018-05-19
Add `timezone` field to Virtual Machine (Windows Configuration)

### LINKED ISSUE #554: Feature Request: Missing timeZone from os_profile_windows_config in azurerm_virtual_machine
modifying the timezone is available when using ARM Template, however there is no argument for this in terraform. This should be included in os_profile_windows_config.

JunyiYi:
This pull request adds the `timezone` field to the `os_profile_windows_config` of `azurerm_virtual_machine`. It also introduces a validation method to verify Azure supported timezone. The timezone list is retrieved from [this blog](http://jackstromberg.com/2017/01/list-of-time-zones-consumed-by-azure/).

Issue #554 will be resolved after this PR is merged.

(fixes #554)


azurerm/validators.go
```diff
@@ -74,3 +75,114 @@ func validateIso8601Duration() schema.SchemaValidateFunc {
...
+		"Russia Time Zone 10",
+		"Central Pacific Standard Time",
+		"Russia Time Zone 11",
+		"New Zealand Standard Time",
+		"UTC+12",
+		"Fiji Standard Time",
+		"Tonga Standard Time",
+		"Samoa Standard Time",
+		"Line Islands Standard Time",
+	}, false)
```
katbyte: Are you sure the API is case sensitive?

azurerm/validators_test.go
```diff
@@ -128,3 +128,42 @@ func TestValidateIso8601Duration(t *testing.T) {
...
+			Errors: 0,
+		},
+		{
+			// Wrong casing
+			Value:  "china standard time",
+			Errors: 1,
+		},
+		{
+			// Invalid UTC time zone
+			Value:  "UTC-10",
```
katbyte: Even if azure won't accept it this we probably should not use it as a failure case because it could be confusing.

azurerm/validators_test.go
```diff
@@ -128,3 +128,42 @@ func TestValidateIso8601Duration(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateAzureTimeZone(t *testing.T) {
```
katbyte: _minor_ I would question the need for this validation as we are just doing a "is the string in this array" check using the built in validator.

azurerm/validators.go
```diff
@@ -74,3 +75,114 @@ func validateIso8601Duration() schema.SchemaValidateFunc {
 		return
 	}
 }
+
+func validateAzureTimeZone() schema.SchemaValidateFunc {
```
katbyte: Can we rename this to `validateAzureVmTimeZone()`? other APIs might accept different values (the automation one accepts UTC+10 for instance)

azurerm/validators.go
```diff
@@ -74,3 +75,114 @@ func validateIso8601Duration() schema.SchemaValidateFunc {
...
+	return validation.StringInSlice([]string{
```
katbyte: Could we please get a comment here linking to where this list is from? Are you sure it is accurate?

azurerm/resource_arm_virtual_machine.go
```diff
@@ -374,6 +374,11 @@ func resourceArmVirtualMachine() *schema.Resource {
 							Optional: true,
 							Default:  false,
 						},
+						"timezone": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validateAzureTimeZone(),
+						},
```
tombuildsstuff: this also needs a state migration to update the `resourceArmVirtualMachineStorageOsProfileWindowsConfigHash` function and a default value (UTC)

azurerm/validators.go
```diff
@@ -74,3 +75,114 @@ func validateIso8601Duration() schema.SchemaValidateFunc {
...
+		"Dateline Standard Time",
```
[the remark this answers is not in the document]
JunyiYi: This is sorted by time offset (From UTC-12 to UTC+14). For your reference: [http://jackstromberg.com/2017/01/list-of-time-zones-consumed-by-azure/](http://jackstromberg.com/2017/01/list-of-time-zones-consumed-by-azure/)

azurerm/resource_arm_virtual_machine.go
```diff
@@ -1291,6 +1300,11 @@ func expandAzureRmVirtualMachineOsProfileWindowsConfig(d *schema.ResourceData) (
 		config.EnableAutomaticUpdates = &update
 	}
 
+	if v := osProfileConfig["timezone"]; v != nil {
+		provision := v.(string)
```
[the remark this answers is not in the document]
metacpp: 2 lines of code is better for debugging, while 1 line of code is better for simplicity. @JunyiYi you can make decision.

azurerm/validators_test.go
```diff
@@ -128,3 +128,42 @@ func TestValidateIso8601Duration(t *testing.T) {
...
+			Errors: 0,
+		},
+		{
+			// Valid UTC time zone
+			Value:  "UTC-11",
+			Errors: 0,
+		},
+		{
+			// Wrong casing
+			Value:  "china standard time",
```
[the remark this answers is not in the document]
JunyiYi: I think a more restricted rule will benefit us.

azurerm/validators.go
JunyiYi: Explained below.

azurerm/validators.go
JunyiYi: I'm trying to make it a general validation rule which would also be used in other timezones (like [the one here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/94d287bcb11c3f6cb1ce238d95f9d6dc67168577/azurerm/resource_arm_automation_schedule.go#L72))

azurerm/resource_arm_virtual_machine.go
JunyiYi: Sure. Is there any samples? I found [one here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/83ecdac05b2479499a564988b4dbe2e3cd7706de/azurerm/resource_arm_storage_account_migration.go).

azurerm/validators_test.go
JunyiYi: Let me see whether Azure actually accepts it or not. If so, I will update the validation rule.

azurerm/resource_arm_virtual_machine.go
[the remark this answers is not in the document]
tombuildsstuff: there's no need to split this out - we can make this a single line; there's no value in debugging a pointer reference/dereference

azurerm/validators_test.go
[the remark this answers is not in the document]
tombuildsstuff: whilst in principal I'd agree here - the Azure CLI / ARM Templates / PowerShell are case insensitive here - thus we should match the users expectations and make this case insensitive

azurerm/validators_test.go
tombuildsstuff: given Azure could possibly support this as a value - it seems odd to use it as  invalid data for test purposes? can we update this to `UTC-48` which should never be a valid timezone

azurerm/validators.go
tombuildsstuff: given the Azure API's vary so much there's no need to make this generic; in Go the general convention is to value duplication over a completely DRY codebase with extra complexities for switch statements

azurerm/resource_arm_virtual_machine.go
tombuildsstuff: that'll do as an example, there's a few in the codebase - this value also needs to be added to the Hash function, otherwise this won't be detected as ForceNew for existing Virtual Machines (which is why the state migrations needed)

azurerm/resource_arm_virtual_machine.go
```diff
@@ -374,6 +374,11 @@ func resourceArmVirtualMachine() *schema.Resource {
 							Optional: true,
 							Default:  false,
 						},
+						"timezone": {
+							Type:         schema.TypeString,
+							Optional:     true,
```
tombuildsstuff: in the Azure API IIRC this defaults to `UTC` - we should default that here too?

tombuildsstuff: does this also need to be ForceNew, since IIRC this value is used as part of the initial post-sysprep?

tombuildsstuff: if not - can we add an example for updating a VM's TimeZone from UTC -> PST?

azurerm/validators.go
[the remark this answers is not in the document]
katbyte: I realize that, but can we sort it alphabetically? Most things in the provider are sorted alphabetically. Sorting by UTC adds a level of unnecessary complexity for maintainers and users alike.

azurerm/validators.go
katbyte: The property you linked to accepts UTC+10 in it API as well as the portal sets it to an entirely different timezone format. For example PST is "America/Vancouver"

azurerm/validators_test.go
JunyiYi: Tried it with Azure, and this value is not supported. Only the UTC time offsets listed in my list is accepted by Azure.

azurerm/resource_arm_virtual_machine.go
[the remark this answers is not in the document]
JunyiYi: Fixed.

azurerm/validators_test.go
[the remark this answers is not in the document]
JunyiYi: Fixed.

azurerm/validators.go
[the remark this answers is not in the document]
JunyiYi: Fixed.

azurerm/validators.go
JunyiYi: Fixed.

azurerm/resource_arm_virtual_machine.go
JunyiYi: Tried with Azure, it has to be `ForceNew`.

azurerm/validators.go
JunyiYi: Changed.

azurerm/resource_arm_virtual_machine.go
JunyiYi: Added.

JunyiYi:
Hi @katbyte , the test case has been added now.

azurerm/validators_test.go
```diff
@@ -182,3 +182,37 @@ func TestValidateIntBetweenDivisibleBy(t *testing.T) {
...
+			Errors: 0,
+		},
+		{
+			// Valid UTC time zone
+			Value:  "utc-11",
+			Errors: 0,
+		},
+		{
+			// Invalid UTC time zone
+			Value:  "UTC-10",
```
katbyte: Could we get an obviously wrong value like `UTC-42` here?

JunyiYi: Done.

JunyiYi:
hi @katbyte , the test case has been updated.

azurerm/resource_arm_virtual_machine.go
```diff
@@ -1595,6 +1612,7 @@ func resourceArmVirtualMachineStorageOsProfileWindowsConfigHash(v interface{}) i
 		if v, ok := m["enable_automatic_upgrades"]; ok {
 			buf.WriteString(fmt.Sprintf("%t-", v.(bool)))
 		}
+		buf.WriteString(fmt.Sprintf("%s-", m["timezone"].(string)))
```
tombuildsstuff: there's a crash where where this field doesn't exist - as such we need to nil-check this.

```
panic: interface conversion: interface {} is nil, not string
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm:
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: goroutine 32 [running]:
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: github.com/terraform-providers/terraform-provider-azurerm/azurerm.resourceArmVirtualMachineStorageOsProfileWindowsConfigHash(0x25b6c00, 0xc4201a73e0, 0xc420530ab8)
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/Users/tharvey/code/src/github.com/terraform-providers/terraform-provider-azurerm/azurerm/resource_arm_virtual_machine.go:1615 +0x3d5
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema.(*Set).hash(0xc420530aa0, 0x25b6c00, 0xc4201a73e0, 0xc420530aa0, 0x0)
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/Users/tharvey/code/src/github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema/set.go:205 +0x3d
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema.(*Set).add(0xc420530aa0, 0x25b6c00, 0xc4201a73e0, 0x268dc00, 0x1, 0xc420530aa0)
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/Users/tharvey/code/src/github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema/set.go:192 +0x72
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema.(*Set).Add(0xc420530aa0, 0x25b6c00, 0xc4201a73e0)
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/Users/tharvey/code/src/github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema/set.go:69 +0x44
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema.NewSet(0x28650d8, 0xc4206f2d40, 0x1, 0x1, 0xc4205309e0)
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/Users/tharvey/code/src/github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema/set.go:56 +0x7a
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: github.com/terraform-providers/terraform-provider-azurerm/azurerm.resourceArmVirtualMachineRead(0xc4206b1ea0, 0x27afe20, 0xc420360000, 0xc4206b1ea0, 0x0)
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/Users/tharvey/code/src/github.com/terraform-providers/terraform-provider-azurerm/azurerm/resource_arm_virtual_machine.go:712 +0x1166
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema.(*Resource).Refresh(0xc4202113b0, 0xc4203ac370, 0x27afe20, 0xc420360000, 0xc4203ea4a8, 0x10bc701, 0x24e1500)
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/Users/tharvey/code/src/github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema/resource.go:354 +0x167
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema.(*Provider).Refresh(0xc4202205b0, 0xc4203ac320, 0xc4203ac370, 0x1, 0x10623c3, 0x27f0001)
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/Users/tharvey/code/src/github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/helper/schema/provider.go:308 +0x9a
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/plugin.(*ResourceProviderServer).Refresh(0xc4205644c0, 0xc4205348e0, 0xc4205349e0, 0x0, 0x0)
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/Users/tharvey/code/src/github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/hashicorp/terraform/plugin/resource_provider.go:549 +0x4e
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: reflect.Value.call(0xc4200924e0, 0xc4201c2238, 0x13, 0x27ed6f9, 0x4, 0xc42069ef18, 0x3, 0x3, 0xc4200b3480, 0x0, ...)
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/usr/local/Cellar/go/1.10.1/libexec/src/reflect/value.go:447 +0x969
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: reflect.Value.Call(0xc4200924e0, 0xc4201c2238, 0x13, 0xc420050718, 0x3, 0x3, 0x0, 0x0, 0x0)
2018-05-25T15:52:08.840+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/usr/local/Cellar/go/1.10.1/libexec/src/reflect/value.go:308 +0xa4
2018-05-25T15:52:08.841+0100 [DEBUG] plugin.terraform-provider-azurerm: net/rpc.(*service).call(0xc4203ae2c0, 0xc4206ea370, 0xc4200b0bf0, 0xc4200b0c00, 0xc4201be780, 0xc420564a60, 0x24e14c0, 0xc4205348e0, 0x16, 0x24e1500, ...)
2018-05-25T15:52:08.841+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/usr/local/Cellar/go/1.10.1/libexec/src/net/rpc/server.go:384 +0x14e
2018-05-25T15:52:08.841+0100 [DEBUG] plugin.terraform-provider-azurerm: created by net/rpc.(*Server).ServeCodec
2018-05-25T15:52:08.841+0100 [DEBUG] plugin.terraform-provider-azurerm: 	/usr/local/Cellar/go/1.10.1/libexec/src/net/rpc/server.go:480 +0x43a
```

azurerm/resource_arm_virtual_machine_migration.go
```diff
@@ -0,0 +1,85 @@
...
+		if len(paths) == 3 && (paths[2] == "enable_automatic_upgrades" || paths[2] == "provision_vm_agent") {
+			b, err := strconv.ParseBool(v)
+			if err != nil {
+				return 0, err
+			}
+			data[paths[2]] = b
+		}
+	}
+	data["timezone"] = timezone
+	return resourceArmVirtualMachineStorageOsProfileWindowsConfigHash(data), nil
```
tombuildsstuff: The underlying data structures will diverge significantly in a future version of Terraform - as such direct manipulation of the state will cause hard to diagnose crashes later. Instead we need to use the schema parser ([example](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_container_registry_migrate.go#L61)) rather than assuming the format for keys, (as here)

This means we can simply assign the new value for the field, and allow Terraform Core to recompute the hash value; meaning the tests _should_ stay the same, but all of this can disappear

azurerm/resource_arm_virtual_machine_migration.go
```diff
@@ -0,0 +1,85 @@
...
+			toAdd[strings.Join(paths, ".")] = v
+		}
+	}
+	for _, k := range toDel {
+		delete(is.Attributes, k)
+	}
+	for k, v := range toAdd {
+		is.Attributes[k] = v
+	}
+}
```
tombuildsstuff: (which means this function can be removed)

website/docs/r/virtual_machine.html.markdown
```diff
@@ -473,6 +473,7 @@ output "principal_id" {
 
 * `provision_vm_agent` - (Optional) This value defaults to false.
 * `enable_automatic_upgrades` - (Optional) This value defaults to false.
+* `timezone` - (Optional) Specifies the time zone of the virtual machine. e.g. "Pacific Standard Time".
```
tombuildsstuff: We can suffix this with ". Defaults to `UTC`"

azurerm/validators_test.go
```diff
@@ -182,3 +182,37 @@ func TestValidateIntBetweenDivisibleBy(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateAzureVMTimeZone(t *testing.T) {
```
tombuildsstuff: `VM` -> `VirtualMachine` (we spell out the resource name for tests)

azurerm/validators.go
```diff
@@ -76,6 +77,118 @@ func validateIso8601Duration() schema.SchemaValidateFunc {
 	}
 }
 
+func validateAzureVMTimeZone() schema.SchemaValidateFunc {
+	// Candidates are listed here: http://jackstromberg.com/2017/01/list-of-time-zones-consumed-by-azure/
```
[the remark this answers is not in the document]
JunyiYi: Nope. Azure doesn't provide any list.

azurerm/resource_arm_virtual_machine.go
JunyiYi: But it contains a default value. If the user didn't provide anything, will the default value be populated to the field?

JunyiYi: It is caused by [`NewSet`](https://github.com/terraform-providers/terraform-provider-azurerm/blob/fd5e54bc1deef0603ba5317c111cb84b52dfa376/azurerm/resource_arm_virtual_machine.go#L712) rather than any `Get`.

So how do we handle default values in `Read`? Do we just leave it empty or set it to the field?

azurerm/resource_arm_virtual_machine_test.go
```diff
@@ -68,3 +70,110 @@ func testCheckAzureRMVirtualMachineDestroy(s *terraform.State) error {
...
+    os_profile {
+		computer_name = "winhost01"
+		admin_username = "testadmin"
+		admin_password = "Password1234!"
+    }
+
+    os_profile_windows_config {
+        timezone = "Pacific Standard Time"
+    }
+}
```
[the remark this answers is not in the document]
JunyiYi: Fixed.

website/docs/r/virtual_machine.html.markdown
JunyiYi: Done.

azurerm/validators_test.go
JunyiYi: Fixed.

azurerm/resource_arm_virtual_machine.go
tombuildsstuff: > But it contains a default value. If the user didn't provide anything, will the default value be populated to the field?

That's not true for existing values in the state (or, older VM's which didn't expose this field) which won't have this value (and thus will crash) - but we can work around that by nil-checking it

azurerm/validators.go
```diff
@@ -76,6 +77,118 @@ func validateIso8601Duration() schema.SchemaValidateFunc {
 	}
 }
 
+func validateAzureVirtualMachineTimeZone() schema.SchemaValidateFunc {
+	// Candidates are listed here: http://jackstromberg.com/2017/01/list-of-time-zones-consumed-by-azure/
+	return validation.StringInSlice([]string{
```
tombuildsstuff: appears this comes from the .net method `TimeZoneInfo.FindSystemTimeZoneById()` - with this appearing to be a [source from Microsoft for the information](https://support.microsoft.com/en-us/help/973627/microsoft-time-zone-index-values)?

azurerm/resource_arm_virtual_machine_migration.go
tombuildsstuff: (heads up that this needs to be resolved before we can merge this, which is why this is still "requested changes" 😄)

Just to add some additional context to testing this; you'll want to compare the state of a VM using the current provider release (currently v1.6.0) to the current development build you're working on; for instance:


```hcl
provider "azurerm" {
  version = "=1.6.0"
}

  # ..
}

# ...
```

```bash
$ rm -rf .terraform/
$ terraform init
$ terraform apply
```

then un-pinning the provider version and re-initing/applying should show the state migration:

```hcl
provider "azurerm" {
}

  # ..
}

# ...
```

```bash
$ rm -rf .terraform/
$ terraform init # to use the development version you've built via `make`
$ terraform plan|apply
```

In order to test this effectively unfortunately you'll want to destroy & recreate the VM for each iteration, which the `taint` command is quite helpful for:

```
$ terraform taint azurerm_virtual_machine.test
$ # switch back to "1.6.0"
$ terraform apply
```

Hope that helps :)

azurerm/resource_arm_virtual_machine.go
tombuildsstuff: given [this field is case insensitive in the schema](https://github.com/terraform-providers/terraform-provider-azurerm/pull/1265/files#diff-b5d4f83f88d9bfb743ea77939d1394dfR385) you'll also want to make this value consistent by either upper/lower-casing it; otherwise you'll get spurious diff's here

website/docs/r/virtual_machine.html.markdown
```diff
@@ -473,6 +473,7 @@ output "principal_id" {
 
 * `provision_vm_agent` - (Optional) This value defaults to false.
 * `enable_automatic_upgrades` - (Optional) This value defaults to false.
+* `timezone` - (Optional) Specifies the time zone of the virtual machine. e.g. `Pacific Standard Time`. Defaults to `UTC`.
```
tombuildsstuff: can we link to the MS page defined above here? e.g.

```
* `timezone` - (Optional) Specifies the time zone of the Virtual Machine, [the possible values of which are defined here](XXX). Defaults to `UTC`.
```

azurerm/resource_arm_virtual_machine_migration.go
JunyiYi: The test steps is really helpful here. So AFAIK, we can only do the manual test here for this kind of issues? Are there any unit-test infrastructure support it?

azurerm/validators.go
JunyiYi: According to my test this is not the full list. For example, Azure also accepts "UTC-02" which is not in the list. And the weird thing is, Azure rejects "UTC-03".

website/docs/r/virtual_machine.html.markdown
JunyiYi: The problem is I only found a 3-rd party blog which seems to be the correct value (I randomly tried the values in the list, and all are accepted. I also tried several reasonable item which is not list, and all are rejected). Do I still suppose to put it into the doc?

azurerm/validators.go
tombuildsstuff: honestly that feels like a bug in the API - as if I saw `UTC-02` on the list, as an operator I'd assume that `UTC-3` or `UTC-03` would work?

azurerm/resource_arm_virtual_machine_migration.go
tombuildsstuff: > So AFAIK, we can only do the manual test here for this kind of issues? Are there any unit-test infrastructure support it?

The associated unit test is checking the logic from a given version - this is more exploratory testing

JunyiYi:
Hi @tombuildsstuff , I changed the default value of `timezone` from `"UTC"` to `""` because the value is empty in Azure (not `UTC`) when you create a virtual machine using the old plugin. In that case, I also removed the state migration code because AFAIK there should be no need to do the state migration any more.

azurerm/resource_arm_virtual_machine.go
JunyiYi: Fixed.

jeffreyCline:
Sound delicious!

azurerm/validators.go
```diff
@@ -76,6 +77,122 @@ func validateIso8601Duration() schema.SchemaValidateFunc {
 	}
 }
 
+func validateAzureVirtualMachineTimeZone(acceptEmpty bool) schema.SchemaValidateFunc {
```
tombuildsstuff: since this is only used in a single place, we can assume this is the case - I'll push a commit to remove this.

tombuildsstuff:
@JunyiYi running the tests for this it appears there's a test failure in one of the new tests:

screenshot 2018-06-22 at 15 27 14

Would you be able to take a look?

Thanks!

JunyiYi:
@tombuildsstuff yes I forgot to update the test case after changing the Hash Function. It is fixed in my last commit.

tombuildsstuff:
Ignoring a transitive test failure, the tests pass:

screenshot 2018-06-26 at 16 02 09

PR #1266, opened 2018-05-21
New Resource: `azurerm_logic_app_workflow`

### LINKED ISSUE #610: Feature Request: Logic App Resource
Hi there!

It would be great to see support for Azure Logic Apps (https://azure.microsoft.com/en-gb/services/logic-apps/) via a new resource.

tombuildsstuff:
This PR adds support for Logic Apps. After initially making this a huge self-contained resource, I've pivoted to adding split-out resources for the Actions and Triggers. This allows us to provide better Diff's for specific actions, whilst still allowing for custom Actions and Triggers via the `_custom` suffixed resources.

Adds the following resources:

```
* **New Data Source:** `azurerm_logic_app_workflow` [GH-1266]
* **New Resource:** `azurerm_logic_app_action_custom` [GH-1266]
* **New Resource:** `azurerm_logic_app_action_http` [GH-1266]
* **New Resource:** `azurerm_logic_app_trigger_custom` [GH-1266]
* **New Resource:** `azurerm_logic_app_trigger_http_request` [GH-1266]
* **New Resource:** `azurerm_logic_app_trigger_recurrence` [GH-1266]
* **New Resource:** `azurerm_logic_app_workflow` [GH-1266]
```

Fixes #610


azurerm/resource_arm_logic_app_workflow.go
```diff
@@ -0,0 +1,469 @@
...
+			},
+
+			"trigger": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"recurrence": {
```
JunyiYi: There are so many other triggers than recurrence trigger. [https://docs.microsoft.com/en-us/azure/logic-apps/logic-apps-workflow-actions-triggers](https://docs.microsoft.com/en-us/azure/logic-apps/logic-apps-workflow-actions-triggers)

azurerm/resource_arm_logic_app_workflow.go
```diff
@@ -0,0 +1,469 @@
...
+						},
+					},
+				},
+			},
+
+			"action": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
```
JunyiYi: There could be more than 1 actions.

azurerm/resource_arm_logic_app_workflow.go
```diff
@@ -0,0 +1,469 @@
...
+				Type:     schema.TypeMap,
+				Optional: true,
+				Computed: true,
+			},
+
+			"trigger": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
```
JunyiYi: 1 is not enough.

azurerm/resource_arm_logic_app_workflow.go
```diff
@@ -0,0 +1,469 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"location": locationSchema(),
+
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"parameters": {
```
JunyiYi: There are two different parameters here. One is the workflow parameter, the other is the definition parameter.

azurerm/resource_arm_logic_app_workflow.go
```diff
@@ -0,0 +1,469 @@
...
+		WorkflowProperties: &logic.WorkflowProperties{
+			Definition: definition,
+			Parameters: parameters,
+		},
+		Tags: expandTags(tags),
+	}
+
+	_, err := client.CreateOrUpdate(ctx, resourceGroup, name, properties)
+	if err != nil {
+		return err
```
JunyiYi: Error message is not consistent with other `return` statements. Provide more detailed info.

azurerm/resource_arm_logic_app_workflow.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+
+	read, err := client.Get(ctx, resourceGroup, name)
+	if err != nil {
+		return err
```
JunyiYi: Refine error message here.

azurerm/resource_arm_logic_app_workflow.go
```diff
@@ -0,0 +1,469 @@
...
+	return resourceArmLogicAppWorkflowRead(d, meta)
+}
+
+func resourceArmLogicAppWorkflowRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).logicWorkflowsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	id, err := parseAzureResourceID(d.Id())
+	if err != nil {
+		return err
```
JunyiYi: Error message too simple.

azurerm/resource_arm_logic_app_workflow.go
```diff
@@ -0,0 +1,469 @@
...
+	return nil
+}
+
+func resourceArmLogicAppWorkflowDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).logicWorkflowsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	id, err := parseAzureResourceID(d.Id())
+	if err != nil {
+		return err
```
JunyiYi: `return fmt.Errorf()`

tombuildsstuff:
Tests pass:

screenshot 2018-07-10 at 16 05 12

azurerm/resource_arm_logic_app_workflow.go
```diff
@@ -0,0 +1,264 @@
...
+		},
+		Tags: expandTags(tags),
+	}
+
+	_, err := client.CreateOrUpdate(ctx, resourceGroup, name, properties)
+	if err != nil {
+		return fmt.Errorf("[ERROR] Error creating Logic App Workflow %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	read, err := client.Get(ctx, resourceGroup, name)
```
bflad: I'm guessing this SDK read here is necessary because the SDK create function does not provide the ID -- it feels extraneous to do this and should probably be avoided if possible (e.g. generate the ID from given arguments). The resource read function can catch non-existent IDs.

azurerm/resource_arm_logic_app_workflow.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		return err
+	}
+	resourceGroup := id.ResourceGroup
+	name := id.Path["workflows"]
+
+	resp, err := client.Get(ctx, resourceGroup, name)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			d.SetId("")
```
bflad: Same issue here where we should provide logging message when removing resource from state.

azurerm/resource_arm_logic_app_workflow.go
tombuildsstuff: we do this across the provider since the URI used for Create doesn't always match the URI used for Get, unfortunately

azurerm/logic_apps.go
```diff
@@ -0,0 +1,186 @@
...
+		if utils.ResponseWasNotFound(read.Response) {
+			d.SetId("")
+			return nil
+		}
+
+		return fmt.Errorf("[ERROR] Error making Read request on Logic App Workflow %q (Resource Group %q): %+v", logicAppName, resourceGroup, err)
+	}
+
+	if read.WorkflowProperties == nil {
+		return fmt.Errorf("[ERROR] Error parsing Logic App Workflow - `WorkflowProperties` is nil")
```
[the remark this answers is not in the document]
tombuildsstuff: nope, if this happens the API's returning data that doesn't confirm to the schema - so this is to prevent crashes

azurerm/resource_arm_logic_app_trigger_http_request.go
```diff
@@ -0,0 +1,188 @@
...
+
+	err = resourceLogicAppTriggerRemove(d, meta, resourceGroup, logicAppName, name)
+	if err != nil {
+		return fmt.Errorf("Error removing Trigger %q from Logic App %q (Resource Group %q): %+v", name, logicAppName, resourceGroup, err)
+	}
+
+	return nil
+}
+
+func validateLogicAppTriggerHttpRequestRelativePath(v interface{}, k string) (ws []string, errors []error) {
```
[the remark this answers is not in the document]
tombuildsstuff: so that function checks the string itself is a valid regex, rather than testing for a specific regex; whilst I thought we had a validate method for that I couldn't seem to find it (although I may have missed it)?

azurerm/import_arm_logic_app_action_custom_test.go
```diff
@@ -0,0 +1,29 @@
+package azurerm
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccAzureRMLogicAppActionCustom_importBasic(t *testing.T) {
```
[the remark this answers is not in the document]
tombuildsstuff: I'm not opposed to that approach (we've started doing it for some resources) - but I think that's a larger project outside the scope of this PR - so I'm going to skip this for now.

azurerm/logic_apps.go
```diff
@@ -0,0 +1,186 @@
...
+	log.Printf("[DEBUG] Preparing arguments for Logic App Workspace %q (Resource Group %q) %s %q", logicAppName, resourceGroup, kind, name)
+
+	// lock to prevent against Actions or Triggers conflicting
+	azureRMLockByName(logicAppName, logicAppResourceName)
+	defer azureRMUnlockByName(logicAppName, logicAppResourceName)
+
+	read, err := client.Get(ctx, resourceGroup, logicAppName)
+	if err != nil {
+		if utils.ResponseWasNotFound(read.Response) {
+			d.SetId("")
```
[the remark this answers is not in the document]
tombuildsstuff: on reflection this should be returning an error, since this is the Create method for a sub-resource (updated)

tombuildsstuff:
Tests pass:

screenshot 2018-07-17 at 13 29 54

azurerm/data_source_logic_app_workflow_test.go
```diff
@@ -0,0 +1,80 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMLogicAppWorkflowDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMLogicAppWorkflowExists(dataSourceName),
+					resource.TestCheckResourceAttr(dataSourceName, "parameters.%", "0"),
```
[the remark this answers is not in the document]
tombuildsstuff: that's an enhancement we should make, but across all resources; going to skip this for now.

PR #1269, opened 2018-05-21
Add an AD Application provider.

Stavrus:
Split from PR #961.


azurerm/resource_arm_ad_application.go
```diff
@@ -0,0 +1,325 @@
...
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"reply_urls": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MinItems: 1,
```
tombuildsstuff: given this has to be specified at create time - isn't this Required?

azurerm/resource_arm_ad_application.go
```diff
@@ -0,0 +1,325 @@
...
+
+			"homepage": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"identifier_uris": {
+				Type:     schema.TypeList,
+				Optional: true,
```
tombuildsstuff: is this field user-specifyable in the portal?

azurerm/resource_arm_ad_application.go
```diff
@@ -0,0 +1,325 @@
...
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[DEBUG] Application ID %q was not found - removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+
+		return fmt.Errorf("Error loading Application %q: %+v", d.Id(), err)
+	}
+
+	d.Set("display_name", resp.DisplayName)
```
tombuildsstuff: can we access all of these properties on `resp.[XX]Properties`? it's how they're returned from the API - these fields aren't necessarily guaranteed to be populated. This also means we can do a nil-check e.g.

```
if props := resp.[XX]Properties; props != nil {
  d.Set("name", resp.DisplayName)
  # ...
}
```

Stavrus:
@tombuildsstuff Yes, it totally makes sense to split the KeyCredentials and PasswordCredentials out like that and I'll pull them out into new resources with new PRs\* as I make your requested fixes to this one.

I do want to clarify that it is possible to assign those credentials to ServicePrincipals as well. I believe this is to support multi-tenanted applications, allowing individual credentials per tenant. It's not a feature that appears in the portal but the endpoints are there in the SDK. Given the above, this will leave us with 4 credentials resources in total: `azurerm_azuread_application_password` / `azurerm_azuread_application_certificate` / `azurerm_azuread_serviceprincipal_password` / `azurerm_azuread_serviceprincipal_certificate`.

\* Since their tests are dependent on this application resource I'll have to wait on actually creating them until this one's merged to simplify the PR process.

azurerm/key_credentials.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		// Azure only allows an in-place update of the Key Credentials list.
+		// Existing keys, matched by their KeyID, must be sent back with their
+		// Value attribute set to nil. New keys need to provide a Value.
+		// By referencing the existing schema (o), we can determine which
+		// entries in the list are existing keys.
+		if o != nil && o.Contains(v) {
+			cred.Value = nil
+		}
```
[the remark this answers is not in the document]
Stavrus: I apologize, I'm not quite sure what you mean by the question so just in-case I'll go a bit overboard in my explanation.

When we update the list, either to change properties on existing keys, to add new keys, or to delete keys, we need to send a list to Azure that contains every single entry we want in the list after everything is said and done. There are no endpoints for creation and/or deletion of individual items. The Value field for these objects is special in that once the item has been created, the field cannot be modified. This is because it contains sensitive data and is thus write-only once. It also always returns nil on reads. To update the list with existing items, we must nil the Value field to indicate we are not changing it. All other properties on the item work the normal way for read/writes.

Example: if we have a list of items A, B, and C, and I want to delete A, leave B alone, edit C, and add D, then I must send to Azure a list with items B, C, and D. Since B and C already exist, I must nil their Value field because I cannot change it after it's been created. I can still change the other properties on C, as long as Value is left nil when I send the item back to Azure. If I don't send B back, with its Value field nil, it will get deleted even though I don't need to change any other property on it.

azurerm/password_credentials.go
```diff
@@ -0,0 +1,177 @@
...
+	"fmt"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
+	"github.com/Azure/go-autorest/autorest/date"
+	"github.com/hashicorp/terraform/helper/hashcode"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func passwordCredentialsSchema() *schema.Schema {
```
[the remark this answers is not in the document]
Stavrus: They're also settable against a Service Principal, which is why they're currently split out in their own files. As mentioned in a previous comment I'll move these to their own resource per your suggestion.

azurerm/resource_arm_ad_application.go
Stavrus: I believe this is the "App ID URI" under settings->properties. You can't specify it during creation in the portal but can modify it afterwards as long as it's globally unique. The manifest section, which allows you to edit all the application settings as a JSON document, is another way of changing it through the portal but that's probably cheating :)

azurerm/resource_arm_ad_application.go
Stavrus: You can edit these at any time to add/remove as needed, and they're not required for creation. I believe the bare minimum properties required for creating an application are the display name, 1 identifier URI, and the homepage. Those last two properties attempt to set default values for you based on the display name you give in their expand functions as a convenience, but the expand for reply_urls doesn't as it's not required. I modeled the default functionality based on how Azure CLI's [az ad app create](https://docs.microsoft.com/en-us/cli/azure/ad/app?view=azure-cli-latest) works.

azurerm/resource_arm_ad_application.go
```diff
@@ -0,0 +1,325 @@
...
+				MinItems: 1,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"available_to_other_tenants": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
```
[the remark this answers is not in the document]
Stavrus: Sure, not a problem. To give some context when picking what to set as default I followed the Azure CLI's [az ad app create](https://docs.microsoft.com/en-us/cli/azure/ad/app?view=azure-cli-latest) which only has the display name as a required parameter.

I think there's also some extra AD permissions required in setting it to true (default is false), but I would have to look into which permissions these are.

azurerm/resource_arm_ad_application.go
```diff
@@ -0,0 +1,325 @@
...
+			},
+
+			"key_credential": keyCredentialsSchema(),
+
+			"password_credential": passwordCredentialsSchema(),
+
+			"app_id": {
```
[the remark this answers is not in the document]
Stavrus: Yes, the `app_id` is different from `id`. It's `object_id` that's redundant so I'll be removing that.

tombuildsstuff:
> I do want to clarify that it is possible to assign those credentials to ServicePrincipals as well. I believe this is to support multi-tenanted applications, allowing individual credentials per tenant. It's not a feature that appears in the portal but the endpoints are there in the SDK. Given the above, this will leave us with 4 credentials resources in total: azurerm_azuread_application_password / azurerm_azuread_application_certificate / azurerm_azuread_serviceprincipal_password / azurerm_azuread_serviceprincipal_certificate.

@Stavrus that's fair - my understanding is that these endpoints are for the `v2` / new authorization endpoints which are still being rolled out; for the moment (e.g. until these are more widely available) I think we only need to support the Application, Service Principal, Service Principal Certificate and Service Principal Key resources (e.g. we can ignore the `azurerm_azuread_application_password` / `azurerm_azuread_application_certificate` resources)?

Stavrus:
@tombuildsstuff Hey, sorry for the delay in updating this PR. I've made most of the requested changes and it should be ready for re-review. I'll be working on the Password/Key resources in the meantime.

What I left out/need further comment on:
- "available_to_other_tenants" is still optional. I'm leaning towards keeping the smallest use-case of the resource -- that is, the minimum required parameters -- to just the name as this is how it is for the Azure CLI `az ad app create` command.
- `resp.[XX]Properties` - I believe you're referring to something like [this](https://github.com/Azure/azure-sdk-for-go/blob/master/services/compute/mgmt/2018-04-01/compute/models.go#L1262)? It doesn't exist for an [AD object](https://github.com/Azure/azure-sdk-for-go/blob/master/services/graphrbac/1.6/graphrbac/models.go#L63). The closest I can get is the AdditionalProperties map, but I don't think that's what you're intending.
- Password/Key resources should be for Applications and not Service Principals, according to [this](https://github.com/Azure/azure-sdk-for-go/issues/699).

azurerm/resource_arm_azuread_application.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		return fmt.Errorf("Error loading Azure AD Application %q: %+v", d.Id(), err)
+	}
+
+	d.Set("name", resp.DisplayName)
+	d.Set("application_id", resp.AppID)
+	d.Set("homepage", resp.Homepage)
+	d.Set("identifier_uris", resp.IdentifierUris)
+	d.Set("reply_urls", resp.ReplyUrls)
```
tombuildsstuff: given these are complex objects - I'm going to push a commit to add a flatten function for each of these and check for errors when setting them

azurerm/resource_arm_azuread_application.go
```diff
@@ -0,0 +1,214 @@
...
+	d.Set("available_to_other_tenants", resp.AvailableToOtherTenants)
```
tombuildsstuff: in some manual testing of this resource - it appears this property needs to be `false` upon deletion; I'll push a commit to ensure this is set before deleting the application

azurerm/resource_arm_azuread_application.go
```diff
@@ -0,0 +1,214 @@
...
+
+func expandAzureRmActiveDirectoryApplicationIdentifierUris(d *schema.ResourceData, name string) *[]string {
+	identifierUris := d.Get("identifier_uris").([]interface{})
+	identifiers := []string{}
+	for _, id := range identifierUris {
+		identifiers = append(identifiers, id.(string))
+	}
+	if len(identifiers) == 0 {
+		identifiers = append(identifiers, fmt.Sprintf("http://%s", name))
+	}
```
tombuildsstuff: looks like we don't need this from testing - so I'll push a commit to remove it

website/docs/r/azuread_application.html.markdown
```diff
@@ -0,0 +1,66 @@
+---
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_azuread_application"
+sidebar_current: "docs-azurerm-resource-active-directory-application"
```
tombuildsstuff: _minor_ let's update this to `docs-azurerm-resource-azuread-application`

tombuildsstuff:
Tests pass:

```
 $ acctests azurerm TestAccAzureRMActiveDirectoryApplication_
=== RUN   TestAccAzureRMActiveDirectoryApplication_importBasic
--- PASS: TestAccAzureRMActiveDirectoryApplication_importBasic (38.38s)
=== RUN   TestAccAzureRMActiveDirectoryApplication_importComplete
--- PASS: TestAccAzureRMActiveDirectoryApplication_importComplete (25.42s)
=== RUN   TestAccAzureRMActiveDirectoryApplication_basic
--- PASS: TestAccAzureRMActiveDirectoryApplication_basic (20.18s)
=== RUN   TestAccAzureRMActiveDirectoryApplication_availableToOtherTenants
--- PASS: TestAccAzureRMActiveDirectoryApplication_availableToOtherTenants (21.60s)
=== RUN   TestAccAzureRMActiveDirectoryApplication_complete
--- PASS: TestAccAzureRMActiveDirectoryApplication_complete (20.19s)
=== RUN   TestAccAzureRMActiveDirectoryApplication_update
--- PASS: TestAccAzureRMActiveDirectoryApplication_update (39.92s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	165.735s
```

PR #1285, opened 2018-05-23
azurerm_sql_firewall_rule doc: fix Name description

kjhosein:
Typo in the `name` attribute's description.  (Was set to the same as the `server_name`)


katbyte:
Hi @kjhosein,

Thank you for noticing and opening a PR! LGTM 👍

PR #1291, opened 2018-05-24
New Resource: `azurerm_monitor_diagnostic_setting` / New Data Source: `azurerm_monitor_diagnostic_categories`

### LINKED ISSUE #657: New Resource: `azurerm_monitor_diagnostic_setting`
_This issue was originally opened by @rohitBaravkar as hashicorp/terraform#16998. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

hi,

I have to enable diagnostic setting on with the help of terraform script, there is support for creating a workspace with log analytics, but I didn't get how to connect resource like service bus to the same workspace to send logs.

thanks,

janboll:
Hi, 

I implemented a new resource to be able to configure diagnostic settings on any given resource (fixes #657). I am opening this PR to request feedback as soon as possible. 
There are still open points
- Documentation
- Testing 
  - Update / delete
  - Complete feature 
- Import
- Datasource 

Please have a look at the PR and check if I am steering into the right direction.
Thanks,
Jan


azurerm/resource_arm_monitor_diagnostics.go
```diff
@@ -0,0 +1,309 @@
...
+	if err != nil {
+		return err
+	}
+
+	read, err := client.Get(ctx, resourceId, name)
+	if err != nil {
+		return err
+	}
+	if read.ID == nil {
+		return fmt.Errorf("Cannot read Diagnostic Settings")
```
tombuildsstuff: can we make this error message clearer to match the other resources? e.g. `fmt.Errorf("Cannot read ID for Monitor Diagnostics %q", name)`

azurerm/resource_arm_monitor_diagnostics.go
```diff
@@ -0,0 +1,309 @@
...
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+			},
+
+			"storage_account_id": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
```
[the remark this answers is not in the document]
janboll: I put this here cause the casing in the URL changes, depending on the input. Did you ever experience this? For example: when getting it from terraform as parameter the IDs are camel cased, when importing the state it is all lowercase.

azurerm/resource_arm_monitor_diagnostics.go
```diff
@@ -0,0 +1,309 @@
...
+			},
+
+			"event_hub_authorization_rule_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"workspace_id": {
+				Type:     schema.TypeString,
+				Optional: true,
```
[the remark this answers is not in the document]
janboll: Does Terraform provide something like: require at least one of the following parameters? Cause you have to set at least storage_account_id, workspace_id or (event_hub_name and event_hub_authorization_rule_id)

azurerm/resource_arm_monitor_diagnostics.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	if len(workspaceId) > 0 {
+		diagnosticSettings.WorkspaceID = &workspaceId
+	}
+
+	if len(eventHubAuthorizationRuleId) > 0 && len(eventHubName) > 0 {
+		diagnosticSettings.EventHubAuthorizationRuleID = &eventHubAuthorizationRuleId
+		diagnosticSettings.EventHubName = &eventHubName
+	}
```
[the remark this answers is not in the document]
janboll: I tried this, this does not work with the DiagnosticSetting API :-( 
`azurerm_monitor_diagnostics.test: insights.DiagnosticSettingsClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="LinkedInvalidPropertyId" Message="Property id '' at path 'properties.eventHubAuthorizationRuleId' is invalid. Expect fully qualified resource Id that start with '/subscriptions/{subscriptionId}' or '/providers/{resourceProviderNamespace}/'."`

janboll:
I had to change the schema cause of the way the Azure API behaves. If you omit one of _MetricSettings_ or _LogSettings_ it will just create these loggings with default settings instead of skipping them. 
Therefore I built the schema so that it matches this behaviour, it will always enable all diagnostic settings unless you disable them using _disabled_settings_

laughtonsm:
Any news on this PR being merged?  Really need this feature 👍

azurerm/resource_arm_monitor_diagnostics.go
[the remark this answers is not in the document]
katbyte: You can accomplish that via a `CustomizeDiff` as [seen here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/4518e2bf0a02fcc7f4dd4e66e5fe7440538cb841/azurerm/resource_arm_automation_schedule.go#L113)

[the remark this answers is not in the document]
janboll: Thank you for the hint. I tried to implement a CustomizeDiff function, but hit a problem with state references. Following code only works if I provide a storage_account_id as a literal value. It does not work using HCL interpolation:
```GO
CustomizeDiff: func(diff *schema.ResourceDiff, v interface{}) error {
...
_, hasStorageAccountId := diff.GetOk("storage_account_id")
...
if !hasStorageAccountId ... {
   return fmt.Errorf("You must specify at least one data sink for diagnostic settings")
}
...
```
So above works using following HCL:
```HCL
storage_account_id = "/subscriptions/xxx/resourceGroups/xxx/providers/Microsoft.Storage/storageAccounts/xxx"
```
But it breaks for following:
```HCL
storage_account_id = "${azurerm_storage_account.test_storage_logs.id}"
```

twinkler:
:+1: Any updates?

BeardedBarbarian:
👍 thumbs up for this!

tombuildsstuff:
@janboll apologies for the really delayed re-review here - I'm going to take another look through this tomorrow :)

tombuildsstuff:
@janboll hey - just to let you know that I'm taking a look at this at the moment. As a part of this I've pulled the code locally to read through properly, and have rebased this on top of master/switched to using the same API as the rest of the resources which use the `insights` SDK - I hope you don't mind!

azurerm/resource_arm_monitor_diagnostics.go
```diff
@@ -0,0 +1,284 @@
...
+			},
+
+			"event_hub_name": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"event_hub_authorization_rule_id": {
+				Type:     schema.TypeString,
+				Optional: true,
```
tombuildsstuff: since this is an (optional) ID, we should be able to validate this too using `ValidateFunc: azure.ValidateResourceIDOrEmpty`

azurerm/resource_arm_monitor_diagnostics.go
```diff
@@ -0,0 +1,284 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"target_resource_id": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
```
tombuildsstuff: can we remove the `DiffSuppressFunc` from this? in addition we can validate this using `ValidateFunc: azure.ValidateResourceID` (rather than the `azure.ValidateResourceIDOrEmpty` which is intended for optional fields)

azurerm/monitor_diagnostics.go
```diff
@@ -0,0 +1,48 @@
...
+
+func retrieveResourceId(diagnosticSettingId string) string {
+	substring := "/providers/microsoft.insights/diagnosticSettings/"
+	return diagnosticSettingId[0:strings.Index(diagnosticSettingId, substring)]
+}
+
+func retrieveDiagnosticName(diagnosticSettingId string) string {
+	substring := "/providers/microsoft.insights/diagnosticSettings/"
+	return diagnosticSettingId[len(substring)+strings.Index(diagnosticSettingId, substring):]
+}
```
tombuildsstuff: on reflection I feel like a less brittle way of doing this would be to compose our own resource ID of `{targetResourceID}|{name}` since some of the Azure API's are case sensitive - I'm going to push a commit to fix this.

azurerm/monitor_diagnostics.go
```diff
@@ -0,0 +1,48 @@
...
+	categoryList, err := client.List(ctx, targetResourceId)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	for _, item := range *categoryList.Value {
+		if item.DiagnosticSettingsCategory.CategoryType == "Metrics" {
+			returnMetricSettings = append(returnMetricSettings, *item.Name)
+		}
+		if item.DiagnosticSettingsCategory.CategoryType == "Logs" {
```
tombuildsstuff: after pulling down the code and looking into this, I think this belongs as a data source; something like:

```
data "azurerm_diagnostic_setting_categories" "test" {
  target_resource_id = "/path/to/resource"
}

outputs "log_names" {
  value = "${data.azurerm_diagnostic_setting_categories.test.log_names}" # or log_types, or something?
}

outputs "metric_names" {
  value = "${data.azurerm_diagnostic_setting_categories.test.metric_names}" # or metric_types, or something?
}
```

these could then be used [with some of the changes coming in Terraform 0.12 in the near future](https://www.hashicorp.com/blog/hashicorp-terraform-0-12-preview-for-and-for-each) to do something like:

```
resource "azurerm_diagnostics_setting" "example" {
  # ...

  dynamic "metric_rule" {
    for_each = data.azurerm_diagnostic_setting_categories.test.metric_names

    content {
      name = metric.name
      # ...
    }
  }
}
```

azurerm/resource_arm_monitor_diagnostics.go
```diff
@@ -0,0 +1,284 @@
...
+		return fmt.Errorf("Invalid value for disabled settings provided, use one or multiple of: %q", allSettings)
+	}
+
+	if len(allSettings) == len(disabledSettings) {
+		return fmt.Errorf("You can not disable all settings, rather delete diagnostic logging")
+	}
+
+	diagnosticSettings := &insights.DiagnosticSettings{}
+	diagnosticSettings.Metrics = expandMetricsConfiguration(*allMetricSettings, disabledSettings, retentionDays)
+	diagnosticSettings.Logs = expandLogConfiguration(*allLogSettings, disabledSettings, retentionDays)
```
tombuildsstuff: so I need to look into this further, but I think `metrics` and `logs` want to be nested elements here, rather than having one setting to configure all of the values. Whilst I appreciate this makes it simpler in some circumstances (e.g. when you want to enable everything) - this means the values could change between Terraform runs which would be unexpected, should a new log become available. Admittedly the same thing is true for a Data Source, but they're intended for Dynamic Data (so this wouldn't be out of place)

azurerm/resource_arm_monitor_diagnostics.go
```diff
@@ -0,0 +1,284 @@
...
+	resp, err := client.Get(ctx, monitoringId.ResourceID, monitoringId.Name)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error making Read request on Diagnostic Setting %s: %+v", monitoringId.Name, err)
+	}
+
+	d.Set("name", *resp.Name)
```
tombuildsstuff: FYI `d.Set` will take care of dereferencing the pointer as needed; so we can remove the `*` here (I'll push a commit with this)

azurerm/resource_arm_monitor_diagnostics.go
```diff
@@ -0,0 +1,284 @@
...
+		d.Set("event_hub_name", *resp.EventHubName)
+	}
+
+	if resp.EventHubAuthorizationRuleID != nil {
+		d.Set("event_hub_authorization_rule_id", *resp.EventHubAuthorizationRuleID)
+	}
+
+	if resp.WorkspaceID != nil {
+		d.Set("workspace_id", *resp.WorkspaceID)
+	}
```
tombuildsstuff: since we don't need to dereference the pointer - we can just set these directly (and Terraform will handle setting empty strings as appropriate)

azurerm/resource_arm_monitor_diagnostics.go
```diff
@@ -0,0 +1,284 @@
...
+
+	return returnSetting
+}
+
+func resourceArmMonitorDiagnosticsDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).monitorDiagnosticSettingsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	name := d.Get("name").(string)
+	targetResoureId := d.Get("target_resource_id").(string)
```
tombuildsstuff: we need to parse the Resource ID out of the URI (since these values don't necessarily exist in the config/state, although they should) - I'll push a commit for this

tombuildsstuff:
@janboll apologies this has sat for a little while - I hope you don't mind but I've pushed a commit to rebase this on top of master so we can proceed with this :)

azurerm/data_source_monitor_diagnostic_categories.go
```diff
@@ -0,0 +1,89 @@
...
+			case insights.Metrics:
+				metrics = append(metrics, *v.Name)
+			default:
+				return fmt.Errorf("Unsupported category type %q", string(category.CategoryType))
+			}
+		}
+	}
+
+	if err := d.Set("logs", logs); err != nil {
+		return fmt.Errorf("Error flattening `logs`: %+v", err)
```
katbyte: Should this be:
```suggestion
		return fmt.Errorf("Error setting `logs`: %+v", err)
```

azurerm/data_source_monitor_diagnostic_categories.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	if err := d.Set("metrics", metrics); err != nil {
+		return fmt.Errorf("Error flattening `metrics`: %+v", err)
```
katbyte: Should this be:
```suggestion
		return fmt.Errorf("Error setting `metrics`: %+v", err)
```

azurerm/data_source_monitor_diagnostic_categories.go
```diff
@@ -0,0 +1,89 @@
...
+
+	for _, v := range val {
+		if v.Name == nil {
+			continue
+		}
+
+		if category := v.DiagnosticSettingsCategory; category != nil {
+			switch category.CategoryType {
+			case insights.Logs:
+				logs = append(logs, *v.Name)
```
katbyte: Should we be nil checking here?

azurerm/data_source_monitor_diagnostic_categories.go
```diff
@@ -0,0 +1,89 @@
...
+			case insights.Metrics:
+				metrics = append(metrics, *v.Name)
```
katbyte: Should we be nil checking here?

azurerm/resource_arm_monitor_diagnostic_setting.go
```diff
@@ -0,0 +1,441 @@
...
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				// NOTE: there's no validation requirements listed for this
```
katbyte: Could we maybe do a `validation.NoZeroValues?`

azurerm/resource_arm_monitor_diagnostic_setting.go
```diff
@@ -0,0 +1,441 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: azure.ValidateResourceID,
+			},
+
+			"eventhub_name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
```
katbyte: I think we have a validation function for this:
```suggestion
				ValidateFunc: azure.ValidateEventHubName(),
			},
```

azurerm/resource_arm_monitor_diagnostic_setting.go
```diff
@@ -0,0 +1,441 @@
...
+	}
+
+	storageAccountId := d.Get("storage_account_id").(string)
+	if storageAccountId != "" {
+		properties.DiagnosticSettings.StorageAccountID = utils.String(storageAccountId)
+		valid = true
+	}
+
+	if !valid {
+		return fmt.Errorf("Either a `eventhub_authorization_rule_id`, `log_analytics_workspace_id` or `storage_account_id` must be set")
```
katbyte: We could move this to a [CustomizeDiff](https://github.com/terraform-providers/terraform-provider-azurerm/blob/cadce839c18ec0d88c9152c029108e350d4892b7/azurerm/resource_arm_servicebus_namespace.go#L94)

Also, is it valid to have all of these set at the same time? or can only one be used?

azurerm/resource_arm_monitor_diagnostic_setting.go
```diff
@@ -0,0 +1,441 @@
...
+	}
+
+	// the Azure SDK prefixes the URI with a `/` such this makes a bad request if we don't trim the `/`
+	targetResourceId := strings.TrimPrefix(actualResourceId, "/")
+	_, err := client.CreateOrUpdate(ctx, targetResourceId, properties, name)
+	if err != nil {
```
katbyte: Very minor, but this and the above line could be joined:
```go
if _, err := client.CreateOrUpdate(ctx, targetResourceId, properties, name); err != nil {
```

azurerm/resource_arm_monitor_diagnostic_setting.go
```diff
@@ -0,0 +1,441 @@
...
+	d.Set("name", id.name)
+	d.Set("target_resource_id", id.resourceID)
+
+	d.Set("eventhub_name", resp.EventHubName)
+	d.Set("eventhub_authorization_rule_id", resp.EventHubAuthorizationRuleID)
+	d.Set("log_analytics_workspace_id", resp.WorkspaceID)
+	d.Set("storage_account_id", resp.StorageAccountID)
+
+	if err := d.Set("log", flattenMonitorDiagnosticLogs(resp.Logs)); err != nil {
+		return fmt.Errorf("Error flattening `log`: %+v", err)
```
katbyte:
```suggestion
		return fmt.Errorf("Error setting `log`: %+v", err)
```

azurerm/resource_arm_monitor_diagnostic_setting.go
```diff
@@ -0,0 +1,441 @@
...
+	}
+
+	if err := d.Set("metric", flattenMonitorDiagnosticMetrics(resp.Metrics)); err != nil {
+		return fmt.Errorf("Error flattening `metric`: %+v", err)
```
katbyte:
```suggestion
		return fmt.Errorf("Error setting `metric`: %+v", err)
```

azurerm/resource_arm_monitor_diagnostic_setting.go
```diff
@@ -0,0 +1,441 @@
...
+	targetResourceId := id.resourceID
+	actualResourceId := strings.TrimPrefix(targetResourceId, "/")
+	resp, err := client.Delete(ctx, actualResourceId, id.name)
+	if err != nil {
+		if !response.WasNotFound(resp.Response) {
+			return fmt.Errorf("Error deleting Monitor Diagnostics Setting %q for Resource %q: %+v", id.name, targetResourceId, err)
+		}
+	}
+
+	// TODO: we need to poll to ensure this is actually gone
```
katbyte: Is the comment still valid?

azurerm/resource_arm_monitor_diagnostic_setting.go
```diff
@@ -0,0 +1,441 @@
...
+
+	return results
+}
+
+type monitorDiagnosticId struct {
+	resourceID string
+	name       string
+}
+
+func parseMonitorDiagnosticId(monitorId string) (*monitorDiagnosticId, error) {
```
katbyte: Might it be better to return two strings rather than a bespoke struck here? This seems the less common method in the provider

azurerm/utils/utils.go
```diff
@@ -19,3 +19,21 @@ func Float(input float64) *float64 {
...
+func SliceContainsString(haystack []interface{}, needle string) bool {
+	for _, s := range haystack {
+		if s == needle {
+			return true
+		}
+	}
+	return false
+}
+
+func StringSliceContainsStringSlice(leftSide, rightSide []interface{}) bool {
```
katbyte: I don't see this used anywhere?

azurerm/utils/utils.go
```diff
@@ -19,3 +19,21 @@ func Float(input float64) *float64 {
 func String(input string) *string {
 	return &input
 }
+
+func SliceContainsString(haystack []interface{}, needle string) bool {
```
katbyte: I don't see this used anywhere?

azurerm/utils/utils.go
janboll: You are right, should I remove it again?

azurerm/utils/utils.go
janboll: You are right, should I remove it again?

azurerm/data_source_monitor_diagnostic_categories.go
janboll: Is the check in line 64 sufficient?

azurerm/data_source_monitor_diagnostic_categories.go
janboll: Is the check in line 64 sufficient?

azurerm/resource_arm_monitor_diagnostic_setting.go
tombuildsstuff: I think we'd be better to move towards structs in the longer term instead vs returning `(string, string, error)` makes this less clear imo?

tombuildsstuff:
@janboll thanks for pushing those changes 👍

Apologies (as I'd meant to post this yesterday) - I hope you don't mind, but I pushed some commits to update this resource (and added a Data Source) to better match the representation in Azure so that we can get this merged :)

Thanks!

azurerm/resource_arm_monitor_diagnostic_setting.go
tombuildsstuff: yep - in some edge-case testing there's some eventual consistency going on here such that we'll want to use a Wait Check Function to ensure it's actually gone imo (sorry thought I'd fixed this prior to pushing)

azurerm/data_source_monitor_diagnostic_categories.go
tombuildsstuff: this is caught above - so it should be fine 👍

azurerm/data_source_monitor_diagnostic_categories.go
tombuildsstuff: this is caught above - so it should be fine 👍

azurerm/resource_arm_monitor_diagnostic_setting.go
tombuildsstuff: pushed a commit for this

azurerm/resource_arm_monitor_diagnostic_setting.go
tombuildsstuff: with some additional testing - there is an edge-case I missed first time around - which is that at least one setting needs to be enabled for this to provision successfully (else the API returns a 404 on creation). I've pushed a commit with logic for this

azurerm/utils/utils.go
tombuildsstuff: 👍 pushed a commit to remove this (I hope you don't mind!)

azurerm/utils/utils.go
tombuildsstuff: 👍 pushed a commit to remove this (I hope you don't mind!)

janboll:
@tombuildsstuff Thanks, really appreciate help on this PR! 👍

azurerm/resource_arm_monitor_diagnostic_setting.go
katbyte: 👍, def should get a better mechanism for parsing IDs into structs sooner then later

twem:
The examples appear to be missing from the documentation. Has this been logged as an issue?

tombuildsstuff:
@twem thanks for the heads up - that's been reported/fixed in #2391 and will be updated as a part of the next release :)

Issue #999, opened 2018-03-19
Feature Request: Allow setting properties on `azurerm_storage_blob` objects


dcolebatch:
When creating blobs for serving static sites or content, setting the `Content-Type` property is important. The default `octet-stream` causes all files, even `index.html`, to download in the browser, not render.  If we set `Content-Type: text/html` file is served correctly.

A powershell example of setting the property to enable static file hosting in Azure:
https://stackoverflow.com/a/42893335

Here's an example of a blob where I'd like to set the Content-Type: https://github.com/dcolebatch/azure_serverless_example/blob/master/setup/main.tf#L51

katbyte:
Closed via #1304 and set to be released in `1.7.0`

katbyte:
@dcolebatch,

Just wanted to let you know we have released 1.7 and it should resolve this issue.

PR #1304, opened 2018-05-26
Introduce `content_type` to `azurerm_storage_blob`

JunyiYi:
* user can specify the blob MIME type (Azure accepts all kinds of string not only MIME, so I didn't add validation rule here)
* the MIME type will be returned in HTTP header when user access the output `url` of the blob
* the MIME type is updatable and unlike other properties in SDK, it is a plain string rather than a pointer

This resolves issue #999 .

### DIFF
--- a/azurerm/resource_arm_storage_blob.go
+++ b/azurerm/resource_arm_storage_blob.go
@@ -547,7 +547,7 @@ func resourceArmStorageBlobUpdate(d *schema.ResourceData, meta interface{}) erro
 		return fmt.Errorf("Error getting storage account %s: %+v", storageAccountName, err)
 	}
 	if !accountExists {
-		return fmt.Errorf("Storage account %s not found", storageAccountName)
+		return fmt.Errorf("Storage account %s not found in resource group %s", storageAccountName, resourceGroupName)
 	}
 
 	name := d.Get("name").(string)
@@ -560,7 +560,8 @@ func resourceArmStorageBlobUpdate(d *schema.ResourceData, meta interface{}) erro
 		blob.Properties.ContentType = d.Get("content_type").(string)
 	}
 
-	err = blob.SetProperties(nil)
+	options := &storage.SetBlobPropertiesOptions{}
+	err = blob.SetProperties(options)
 	if err != nil {
 		return fmt.Errorf("Error setting properties of blob %s (container %s, storage account %s): %+v", name, storageContainerName, storageAccountName, err)
 	}
@@ -601,7 +602,8 @@ func resourceArmStorageBlobRead(d *schema.ResourceData, meta interface{}) error 
 	container := blobClient.GetContainerReference(storageContainerName)
 	blob := container.GetBlobReference(name)
 
-	err = blob.GetProperties(nil)
+	options := &storage.GetBlobPropertiesOptions{}
+	err = blob.GetProperties(options)
 	if err != nil {
 		return fmt.Errorf("Error getting properties of blob %s (container %s, storage account %s): %+v", name, storageContainerName, storageAccountName, err)
 	}
--- a/azurerm/resource_arm_storage_blob.go
+++ b/azurerm/resource_arm_storage_blob.go
@@ -605,9 +605,7 @@ func resourceArmStorageBlobRead(d *schema.ResourceData, meta interface{}) error 
 	if err != nil {
 		return fmt.Errorf("Error getting properties of blob %s (container %s, storage account %s): %+v", name, storageContainerName, storageAccountName, err)
 	}
-	if blob.Properties.ContentType != "" {
-		d.Set("content_type", blob.Properties.ContentType)
-	}
+	d.Set("content_type", blob.Properties.ContentType)
 
 	url := blob.GetURL()
 	if url == "" {
--- a/azurerm/resource_arm_storage_blob.go
+++ b/azurerm/resource_arm_storage_blob.go
@@ -20,6 +20,7 @@ func resourceArmStorageBlob() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceArmStorageBlobCreate,
 		Read:   resourceArmStorageBlobRead,
+		Update: resourceArmStorageBlobUpdate,
 		Exists: resourceArmStorageBlobExists,
 		Delete: resourceArmStorageBlobDelete,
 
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_storage_blob.go
```diff
@@ -149,6 +156,7 @@ func resourceArmStorageBlobCreate(d *schema.ResourceData, meta interface{}) erro
 	blobType := d.Get("type").(string)
 	cont := d.Get("storage_container_name").(string)
 	sourceUri := d.Get("source_uri").(string)
+	contentType := d.Get("content_type").(string)
 
 	log.Printf("[INFO] Creating blob %q in storage account %q", name, storageAccountName)
 	if sourceUri != "" {
```
metacpp: Refer to this article to simplify the `happy-path` code block:
https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

tombuildsstuff: we can pull this out into separate methods - but I think that's outside of the scope of this PR tbh; so let's do that in a separate PR

JunyiYi: Agree with @tombuildsstuff . The blob storage code is a messy and I also think the fields are not good and make people confused. That will be a big refactoring, so let's do it in another PR if possible.

azurerm/resource_arm_storage_blob.go
```diff
@@ -53,6 +54,12 @@ func resourceArmStorageBlob() *schema.Resource {
 				Default:      0,
 				ValidateFunc: validateArmStorageBlobSize,
 			},
+			"content_type": {
+				Type:          schema.TypeString,
```
jeffreyCline: Should do a state migration test here? Just to make sure it doesn't break anything downstream?

JunyiYi: I have run the new `terraform plan` after the old `terraform apply`, and it turns out nothing has been changed. I think there is not needs to do state migration here. @katbyte @tombuildsstuff @paultyng what do you think?

azurerm/resource_arm_storage_blob.go
```diff
@@ -53,6 +54,12 @@ func resourceArmStorageBlob() *schema.Resource {
...
+				Optional:      true,
+				Default:       "application/octet-stream",
```
tombuildsstuff: where does this default come from?

azurerm/resource_arm_storage_blob.go
tombuildsstuff: related question below: where does this default value come from?

azurerm/resource_arm_storage_blob.go
tombuildsstuff: also, at _some point in the future_ (read: when the SDK's are more stable) we'll be switching over to the new [Storage SDKs](http://github.com/Azure/azure-storage-blob-go/issues/48); which would probably be a sensible time to do this refactoring

azurerm/resource_arm_storage_blob_test.go
```diff
@@ -666,3 +711,45 @@ resource "azurerm_storage_blob" "destination" {
 }
 `, rInt, location, rString, sourceBlobName)
 }
+
+func testAccAzureRMStorageBlobPage_blockContentType(rInt int, rString, sourceBlobName, location, contentType string) string {
```
tombuildsstuff: _minor_ we've got a _vague_ convention here of making these `(rInt, rString, location, [others])`

azurerm/resource_arm_storage_blob.go
tombuildsstuff: can we also document this new field in the docs (`website/docs/r/storage_blob.html.markdown`)

azurerm/resource_arm_storage_blob.go
```diff
@@ -556,6 +600,13 @@ func resourceArmStorageBlobRead(d *schema.ResourceData, meta interface{}) error
 
 	container := blobClient.GetContainerReference(storageContainerName)
 	blob := container.GetBlobReference(name)
+
+	err = blob.GetProperties(nil)
+	if err != nil {
+		return fmt.Errorf("Error getting properties of blob %s (container %s, storage account %s): %+v", name, storageContainerName, storageAccountName, err)
+	}
+	d.Set("content_type", blob.Properties.ContentType)
```
tombuildsstuff: can we nil-check if `Properties != nil` here?

azurerm/resource_arm_storage_blob.go
JunyiYi: Sure. The default value comes from `Content-Type` section in https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob#request-headers-all-blob-types.

azurerm/resource_arm_storage_blob.go
JunyiYi: The default value comes from `Content-Type` section in https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob#request-headers-all-blob-types.

azurerm/resource_arm_storage_blob.go
```diff
@@ -524,6 +535,39 @@ func resourceArmStorageBlobBlockUploadWorker(ctx resourceArmStorageBlobBlockUplo
...
+	storageContainerName := d.Get("storage_container_name").(string)
+
+	container := blobClient.GetContainerReference(storageContainerName)
+	blob := container.GetBlobReference(name)
+
+	if d.HasChange("content_type") {
+		blob.Properties.ContentType = d.Get("content_type").(string)
+	}
+
+	err = blob.SetProperties(nil)
```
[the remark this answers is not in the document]
JunyiYi: The default timeout 30s should be just fine: https://docs.microsoft.com/en-us/rest/api/storageservices/setting-timeouts-for-blob-service-operations.

azurerm/resource_arm_storage_blob.go
```diff
@@ -556,6 +600,13 @@ func resourceArmStorageBlobRead(d *schema.ResourceData, meta interface{}) error
 
 	container := blobClient.GetContainerReference(storageContainerName)
 	blob := container.GetBlobReference(name)
+
+	err = blob.GetProperties(nil)
```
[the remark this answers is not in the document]
JunyiYi: The default timeout 30s should be just fine: https://docs.microsoft.com/en-us/rest/api/storageservices/setting-timeouts-for-blob-service-operations.

azurerm/resource_arm_storage_blob.go
JunyiYi: `Properties` is not a pointer type: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/vendor/github.com/Azure/azure-sdk-for-go/storage/blob.go#L34. Actually both `Properties` and `ContentType` are not pointer types.

katbyte:
Tests pass: 
screen shot 2018-06-11 at 23 15 53

Issue #901, opened 2018-02-27
VMSS properties enhancement 


djsly:
### Affected Resource(s)
- azurerm_virtual_machine_scale_set

https://www.terraform.io/docs/providers/azurerm/r/virtual_machine_scale_set.html#id-1

it would be beneficial to augment the attribute reference list of the scale_set with the VM name and IPs to allow us to perform remote provisioner on them. 

We are currently migrating from Availability Set to ScaleSet (auto-scaling disabled) and want to still use our post-provisioning logic. 

From what I see, the azure-go-sdk allows us to fetch that information. https://github.com/Azure/azure-sdk-for-go/blob/master/services/compute/mgmt/2017-12-01/compute/virtualmachinescalesetvms.go

Would you be open to augment the list of attributes? I'm looking on a general agreement before starting any PR work.

tombuildsstuff:
hey @djsly 

Thanks for opening this issue :)

So we'd certainly be interested in exposing that information, presumably as a Computed List of some sort?

I have a separate question since I'm unfamiliar with the lifecycle of Failed instances within a VM Scale Set with auto scaling disabled: I'm wondering if it'd be better to use a VM Extension in this instance, since if a machine is recycled/replaced due to entering the Failed state then it'll come up without the Provisioner having run?

Thanks!

djsly:
@tombuildsstuff great question indeed! I was thinking the same and my understanding is that the VMSS will return once all of the initial VM count is populated. 

The computed list of VM info should be the same size of the requested scale count. To scale up the number of VM, terraform would still be the source of truth, therefore, on the next `apply` the ScaleSet size would increase and the computed list be updated to reflect the new addition. 

In case a VM fails, it will be recreated using the same OS/Data Disk, therefore the pre-provisioning stage would still be intact. To be honest, I'm not a big fan or remote-provisioner with ScaleSet... I think that leveraging `cloud-init` is the best mechanisms for this use case once you start using the auto-scaling feature. 

Regarding the expected PR:

as a first stage, I would like to see a computed list of VMs with their IP and status being exposes as ScaleSet properties. I'm sure that there are a lot more nuts and bolt we could add to enhance the VMSS, but I'm aiming for a minimal first iteration :)

bpoland:
If you are intending to manage the lifecycle with terraform, why use a scale set at all?  Certainly if you start using auto-scaling, you wouldn't want to use terraform provisioners or any other per-VM actions would you?

[...]

PR #1321, opened 2018-05-30
initial datasource for vmss

djsly:
fixes: #901 

This is an initial implementation of the VM specific properties using a Data Source

### DIFF
--- a/azurerm/data_source_virtual_machine_scale_set.go
+++ b/azurerm/data_source_virtual_machine_scale_set.go
@@ -1,0 +1,147 @@
+package azurerm
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2017-12-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2017-09-01/network"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmVirtualMachineScaleSet() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmVirtualMachineScaleSetRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"resource_group_name": resourceGroupNameForDataSourceSchema(),
+
+			"location": locationForDataSourceSchema(),
+
+			"vm_hostnames": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"vm_statuses": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"vm_primary_private_ip_addresses": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+		},
+	}
+}
+
+func dataSourceArmVirtualMachineScaleSetRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).vmScaleSetClient
+	vmsClient := meta.(*ArmClient).vmScaleSetVMsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	name := d.Get("name").(string)
+	resourceGroupName := d.Get("resource_group_name").(string)
+
+	vmssResp, err := client.Get(ctx, resourceGroupName, name)
+	if err != nil {
+		return fmt.Errorf("Error reading VM Scale Set: %+v", err)
+	}
+
+	d.SetId(*vmssResp.ID)
+	d.Set("name", name)
+	d.Set("resource_group_name", resourceGroupName)
+	if location := vmssResp.Location; location != nil {
+		d.Set("location", azureRMNormalizeLocation(*location))
+	}
+
+	vmsResp, err := vmsClient.ListComplete(ctx, resourceGroupName, name, "", "", "")
+	if err != nil {
+		if utils.ResponseWasNotFound(vmsResp.Response().Response) {
+			log.Printf("[INFO] VMs for AzureRM Virtual Machine Scale Set (%s) were Not Found.", name)
+		}
+		return fmt.Errorf("Error making Read request for VMs on Azure Virtual Machine Scale Set %s: %+v", name, err)
+	}
+
+	vm_hostnames, vm_statuses, vm_primary_private_ip_addresses := flattenAzureRmVirtualMachineScaleSetVMs(vmsResp, meta, resourceGroupName, name)
+	if err := d.Set("vm_hostnames", vm_hostnames); err != nil {
+		return fmt.Errorf("[DEBUG] Error setting Virtual Machine Scale Set hostname array error: %#v", err)
+	}
+
+	if err := d.Set("vm_statuses", vm_statuses); err != nil {
+		return fmt.Errorf("[DEBUG] Error setting Virtual Machine Scale Set vm hostnames array error: %#v", err)
+	}
+
+	if err := d.Set("vm_primary_private_ip_addresses", vm_primary_private_ip_addresses); err != nil {
+		return fmt.Errorf("[DEBUG] Error setting Virtual Machine Scale Set vm primary private ip array error: %#v", err)
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


djsly:
Hi @seanknox, np, we are using remote exec with Bastion Host within a `null_resource`

```
resource "null_resource" "kubenode_postinstall" {
  count = "${var.enabled == "true" ? var.nb_nodes : 0}"

  triggers {
    kubenode               = "${data.azurerm_virtual_machine_scale_set.kubenode.vm_hostnames[count.index]}"
  }

  connection {
    type        = "ssh"
    host        = "${data.azurerm_virtual_machine_scale_set.kubenode.vm_primary_private_ip_addresses[count.index]}"
    user        = "${var.sshuser}"
    private_key = "${file(var.private_key_path)}"

    bastion_host = "${var.bastion_host}"
  }

  # BOOTSTRAP
  provisioner "file" {
    content     = "${data.template_file.kubenode_postinstall.rendered}"
    destination = "/tmp/postinstall.sh"
  }

  provisioner "remote-exec" {
    inline = [
      "sudo chmod +x /tmp/postinstall.sh",
      "set -o pipefail; sudo /tmp/postinstall.sh 2>&1 | tee /tmp/postinstall.log",
    ]
  }

  depends_on = ["azurerm_virtual_machine_scale_set.kubenode"]
```

seanknox:
@djsly Sorry to hijack your PR, but is`data.azurerm_virtual_machine_scale_set`  a defined data source in the azure provider? I didn't see it referenced in https://www.terraform.io/docs/providers/azurerm/index.html.

djsly:
it comes with this PR.

djsly:
@katbyte @tombuildsstuff who can have a look at this

djsly:
@katbyte have you had a chance to forward this PR to some internal folks per our discussion in the other PR. 

Thanks!

azurerm/data_source_virtual_machine_scale_set.go
```diff
@@ -0,0 +1,147 @@
...
+
+			"location": locationForDataSourceSchema(),
+
+			"vm_hostnames": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"vm_statuses": {
```
jeffreyCline: Shouldn't this be `vm_status`, since it's a list multiple values can already be assumed. `vm_statuses` feels a bit odd.

azurerm/data_source_virtual_machine_scale_set.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmVirtualMachineScaleSet() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmVirtualMachineScaleSetRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
```
jeffreyCline: Add `ForceNew; true,` if the name changes this should force a new resource.

azurerm/data_source_virtual_machine_scale_set.go
```diff
@@ -0,0 +1,147 @@
...
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"vm_statuses": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"vm_primary_private_ip_addresses": {
```
jeffreyCline: Same as above, `vm_primary_private_ip_address`, it's a list so multiple values are going to be assumed.

azurerm/data_source_virtual_machine_scale_set.go
```diff
@@ -0,0 +1,147 @@
...
+func dataSourceArmVirtualMachineScaleSetRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).vmScaleSetClient
+	vmsClient := meta.(*ArmClient).vmScaleSetVMsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	name := d.Get("name").(string)
+	resourceGroupName := d.Get("resource_group_name").(string)
+
+	vmssResp, err := client.Get(ctx, resourceGroupName, name)
+	if err != nil {
```
jeffreyCline: Add a not found check...
`if utils.ResponseWasNotFound(vmssResp.Response) {
    return fmt.Errorf("Error: Virtual Machine Scale Set %q (Resource Group %q) was not found", name, resourceGroupName)}`

azurerm/data_source_virtual_machine_scale_set.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		return fmt.Errorf("Error making Read request for VMs on Azure Virtual Machine Scale Set %s: %+v", name, err)
+	}
+
+	vm_hostnames, vm_statuses, vm_primary_private_ip_addresses := flattenAzureRmVirtualMachineScaleSetVMs(vmsResp, meta, resourceGroupName, name)
+	if err := d.Set("vm_hostnames", vm_hostnames); err != nil {
+		return fmt.Errorf("[DEBUG] Error setting Virtual Machine Scale Set hostname array error: %#v", err)
+	}
+
+	if err := d.Set("vm_statuses", vm_statuses); err != nil {
```
jeffreyCline: update to `vm_status`

azurerm/data_source_virtual_machine_scale_set.go
```diff
@@ -0,0 +1,147 @@
...
+		return fmt.Errorf("[DEBUG] Error setting Virtual Machine Scale Set vm hostnames array error: %#v", err)
+	}
+
+	if err := d.Set("vm_primary_private_ip_addresses", vm_primary_private_ip_addresses); err != nil {
```
jeffreyCline: Update to `vm_primary_private_ip_address`

azurerm/data_source_virtual_machine_scale_set.go
```diff
@@ -0,0 +1,147 @@
...
+	for vmsItt.NotDone() {
+		vmssVM := vmsItt.Value()
+		s := map[string]interface{}{
+			"name":     *vmssVM.Name,
+			"type":     *vmssVM.Type,
+			"hostname": *vmssVM.OsProfile.ComputerName,
+			"status":   *vmssVM.ProvisioningState,
+		}
+		primary_private_ip_address := flattenAzureRmVirtualMachineScaleSetVMNetworkInterfaces(meta, resGroup, name, *vmssVM.InstanceID)
+		vm_statuses = append(vm_statuses, s["status"].(string))
```
jeffreyCline: Update to `vm_status`

azurerm/data_source_virtual_machine_scale_set.go
```diff
@@ -0,0 +1,147 @@
...
+		vm_hostnames = append(vm_hostnames, s["hostname"].(string))
+		vm_primary_private_ip_addresses = append(vm_primary_private_ip_addresses, primary_private_ip_address)
```
jeffreyCline: Update to `vm_primary_private_ip_address`

azurerm/data_source_virtual_machine_scale_set.go
```diff
@@ -0,0 +1,147 @@
...
+	d.Set("name", name)
+	d.Set("resource_group_name", resourceGroupName)
+	if location := vmssResp.Location; location != nil {
+		d.Set("location", azureRMNormalizeLocation(*location))
+	}
+
+	vmsResp, err := vmsClient.ListComplete(ctx, resourceGroupName, name, "", "", "")
+	if err != nil {
+		if utils.ResponseWasNotFound(vmsResp.Response().Response) {
+			log.Printf("[INFO] VMs for AzureRM Virtual Machine Scale Set (%s) were Not Found.", name)
```
JunyiYi: Is it by intention not returning the error message here?

tombuildsstuff:
hey @djsly 

We've been thinking about how to proceed with this PR for a while - since the information being exposed here (Virtual Machine Names / Statuses / IP Addresses) could change between Terraform runs there's a question over whether this information belongs within Terraform.

After some time thinking about this, I've come to the conclusion that as this information is primarily operational (and thus could change between runs) - it probably doesn't belong as on the Resource/Data Source within Terraform (although I could definitely see the use-case for a Data Source without this information). Instead I believe a better approach would be to shell out to the Azure CLI / JQ using [the External provider](https://www.terraform.io/docs/providers/external/data_source.html) to obtain this information. Whilst I appreciate this is different stance to previously - this approach also mirrors our other providers which don't expose this information since it can change out-of-band. As such I'd like to thank you for this contribution however I'm going to close this PR for the moment.

Thanks!

Issue #658, opened 2018-01-03
Encryption Settings - Storage Account


hashibot:
_This issue was originally opened by @raghavanr as hashicorp/terraform#17022. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

how to encrypt VM and Storage Account with Microsoft.Keyvault rather than Microsoft.Storage using terraform.
getting the error when trying with the below code..
``
resource "azurerm_storage_account" "sa" {
  name                     = "sa"
  resource_group_name      = "${azurerm_resource_group.myrg.name}"
  location                 = "${azurerm_resource_group.myrg.location}"
  account_tier             = "Standard"
  account_replication_type = "LRS"
  enable_blob_encryption = "true"
  enable_file_encryption = "true"
  account_encryption_source = "Microsoft.Keyvault"
}

<img width="946" alt="keyvaultpropertieserror" src="https://user-images.githubusercontent.com/33410658/34509386-64a4f6d4-f06f-11e7-96c0-2f0b28143e75.png">

scottrangerio:
Having done a little digging into this, I've managed to track down what I believe the issue is.

https://github.com/terraform-providers/terraform-provider-azurerm/blob/9733b43b2b0b439c5b46c3b68cd3e4c3f85e9e8e/azurerm/resource_arm_storage_account.go#L243-L249

doesn't set the required KeyVaultProperties as defined in 

https://github.com/Azure/azure-sdk-for-go/blob/934e2462aeb6e0c14186dcfeedd73a026d1b8eeb/arm/storage/models.go#L402-L406

https://github.com/Azure/azure-sdk-for-go/blob/934e2462aeb6e0c14186dcfeedd73a026d1b8eeb/arm/storage/models.go#L444-L448

bg2132332:
Hi, this was linked to issue #1323, which is now released, but I don't see KeyVaultProperties added.  Is that something planned?

LaurentLesle:
After reviewing the issue, looks like the "KeyVaultProperties" are missing.

this is what I have to do to setup storage encryption with Keyvault from ARM template:

"properties": {
                        "supportsHttpsTrafficOnly": true,
                        "encryption": {
                            **"keyvaultproperties": {
                              "keyvaulturi": "https://kv242342.vault.azure.net",
                              "keyname": "storage-key-rsa-hsm-2048",
[...]

PR #1323, opened 2018-05-31
Storage Account: Add identity property

lstolyarov:
Hey @tombuildsstuff,

This PR adds the object id as an output of a storage account. At the moment `resp.Identity` is returning `nil`. Any ideas why?


azurerm/resource_arm_storage_account.go
```diff
@@ -495,6 +500,11 @@ func resourceArmStorageAccountRead(d *schema.ResourceData, meta interface{}) err
 	}
 	d.Set("account_kind", resp.Kind)
 
+	log.Printf("[INFO] Identity is %q", resp.Identity)
+	if identity := resp.Identity; identity != nil {
```
tombuildsstuff: @lstolyarov you'll want to use `resp.[XX]Properties.Identity` here - since the top level fields `resp.Identity` isn't guaranteed to have a value, unfortunately. These fields exist primarily for the older API's where the responses aren't guaranteed in the `properties` block in the JSON

tombuildsstuff:
Related issue: #658

azurerm/resource_arm_storage_account.go
lstolyarov: @tombuildsstuff what do you mean by `[XX]Properties`? There is an `AccountProperties` filed in `resp` but that does not have an `Identity` field

tombuildsstuff: >  @tombuildsstuff what do you mean by `[XX]Properties`? There is an `AccountProperties` filed in `resp` but that does not have an `Identity` field 

The `properties` field is prefixed with the name of the struct, so it's different (hence `[XX]Properties`, sorry I should have looked this up).

I don't see any option to enable this in the Portal - is this feature in Preview?

lstolyarov: Well it is not really a feature. This PR is to be able to extract the Object ID (or Principal ID as it is referred to in the SDK). I don't think there is any anyway to get this information out in the portal - it could be extracted with powershell.

tombuildsstuff: sorry, I was referring to pulling this out via the API/SDK - I'm not sure if you've seen it but [the Azure Resources Explorer](https://resources.azure.com) can be really helpful here.

In other API's (e.g. App Service) the `identity` block won't be returned from the API until it's enabled, which can be done by sending the following Request (in the Create/Update):

```
Identity: &storage.Identity{
  Type: utils.String("SystemAssigned"),
},
```

Once the Identity's assigned, the `identity` block returned from the API SDK as Principal ID/Object ID. That said, given this isn't available in the Portal yet - and I can't see any reference to it online; I have a feeling this may be in Preview?

lstolyarov: Ok, that makes more sense now. 

Thanks for the Azure Resources Explorer - its really useful.

I am not sure if it would technically count as a preview feature or not. My understanding that behind the scenes all resources have object ids and this is required as input for key vault. It is not really useful in the portal as when you enable storage account encryption and select the key  vault you want to use that object id gets populated automatically.

azurerm/resource_arm_storage_account.go
```diff
@@ -223,6 +223,33 @@ func resourceArmStorageAccount() *schema.Resource {
...
 			},
 
+			"identity": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
```
katbyte: As there is only one possible value for now should we just default it to `SystemAssigned`?

azurerm/resource_arm_storage_account.go
```diff
@@ -651,3 +700,30 @@ func validateArmStorageAccountType(v interface{}, k string) (ws []string, es []e
...
+	}
+}
+
+func flattenAzureRmStorageAccountIdentity(identity *storage.Identity) []interface{} {
+	if identity == nil {
+		return make([]interface{}, 0)
+	}
+
+	result := make(map[string]interface{})
+	result["type"] = *identity.Type
```
katbyte: Could we also check identity.Type for `nil`?

website/docs/r/storage_account.html.markdown
```diff
@@ -101,6 +111,17 @@ The following attributes are exported in addition to the arguments listed above:
...
+
+---
+
+`identity` exports the following:
+
+* `principal_id` - The Principal ID for the Service Principal associated with the Identity of this Storage Account.
+
+* `tenant_id` - The Tenant ID for the Service Principal associated with the Identity of this Storage Account.
+
+-> You can access the Principal ID via `${azurerm_storage_account.test.identity.0.principal_id}` and the Tenant ID via `${azurerm_storage_account.test.identity.0.tenant_id}`
```
katbyte: I don't think we need this?

katbyte (changes requested):
Hello @lstolyarov,

Thank you for opening this PR. It mostly LGTM and I've let a couple of comments inline. Once those are sorted I look forward to merging this for you 🙂

website/docs/r/storage_account.html.markdown
```diff
@@ -81,6 +83,14 @@ The following arguments are supported:
 
 ~> **Note:** [More information on Validation is available here](https://docs.microsoft.com/en-gb/azure/storage/blobs/storage-custom-domain-name)
 
+---
+
+`identity` supports the following:
+
+* `type` - (Required) Specifies the identity type of the Storage Account. At this time the only allowed value is `SystemAssigned`.
+
+~> The assigned `principal_id` and `tenant_id` can be retrieved after the Storage Account has been created. More details are available below.
```
katbyte: Might it be worth explaining these properties won't be assigned until the identity type value is set to `SystemAssigned` ?

virtualbubble:
Any idea on when this will be released?

website/docs/r/storage_account.html.markdown
lstolyarov: I think it is still useful for the user to be able to see an example on how to access the property. However, if you still think it's not required - I can remove it.

katbyte (changes requested):
Hi @lstolyarov,

Thank you for the updates, this LGTM now however running the tests I am consistently getting this failure:

```
------- Stdout: -------
=== RUN   TestAccAzureRMStorageAccount_updateResourceByEnablingIdentity
--- FAIL: TestAccAzureRMStorageAccount_updateResourceByEnablingIdentity (93.41s)
    testing.go:513: Step 1 error: Check failed: Check 3/4 error: azurerm_storage_account.testsa: Attribute 'identity.0.principal_id' didn't match "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$", got ""
FAIL
```

Additionally, could you please merge/rebase to resolve the merge conflicts?

website/docs/r/storage_account.html.markdown
katbyte: Personally I think it should be removed but its very minor and not a blocker

lstolyarov:
Hey @katbyte,

This should be ready for another review now. Tests are passing & merge conflicts are resolved.

katbyte (approved):
Hi @lstolyarov,

Thank you for the updates, LGTM 👍

katbyte:
tests pass (failing ones are encryption and too many networks)
screen shot 2018-06-13 at 20 01 21

PR #1349, opened 2018-06-05
[WIP] Data sources for azure applications

sophos-jeff:
Hi this is really just a proof of concept of a feature.  I know i still need to write tests and documentation but before i went to far down that road I wanted to get some feedback on this feature.  This would be the first support to the AAD apis in azure to get a list of Azure Directory Objects.  The use case I have put forth is getting application names from a filter.

This is quite rough but wondering if I should continue down the road of improving and getting some direction before I go to far in a direction that might not be the way you guys would want to go with this.

Here is some example code and its outputs based upon the code I wrote here.

Code example
```hcl
data "azurerm_aad_application" "selected" {
  filter {
    name = "displayName"
    values = ["sample-ms"]
  }
}

data "azurerm_aad_application_names" "selected" {
  filter {
    name = "displayName"
    prefix = true
    values = ["sample-ms","DoesThisWork"]
  }
}
```

result
```bash
application_id = 2be2cc44-0627-48ee-b52f-3f53fa6ff5f1
applications = [
    sample-ms,
    sample-ms.test.account,
    sample-ms2,
    DoesThisWork,
    DoesThisWork2,
    DoesThisWork3
]
homepage = http://sample-ms
object_id = 2b66442c-52b9-4cf8-995a-dbabc0f16574
```

Let me know if this is something you guys would be interested in.  I know I have another PR open right now and will address those comments but thought I would get some feedback on this potential feature as well.

Also note: There seems to have been a bug in the auth code that was only an issue if not authenticating via service principal.  I added the following code to get a new token that would be good for microsoft graph

```golang
	err = spt.Refresh()

	if err != nil {
		return nil, err
	}
```

According to the go sdk documentation when using adal.NewServicePrincipalTokenFromManualToken you need to refresh the token to get a new one.  Otherwise you get a 401 on the graph resource.

Thank you for your time and consideration!

-Jeff


sophos-jeff:
@tombuildsstuff Now that the other PR was merged I will continue on this as well.  Any suggestions on the path forward here or am I going in the right direction and should just continue with it?

azurerm/config.go
```diff
@@ -300,6 +303,13 @@ func getAuthorizationToken(c *authentication.Config, oauthConfig *adal.OAuthConf
 	}
 
 	spt, err := adal.NewServicePrincipalTokenFromManualToken(*oauthConfig, c.ClientID, endpoint, *c.AccessToken)
+
+	if err != nil {
+		return nil, err
+	}
+
+	err = spt.Refresh()
```
tombuildsstuff: this is a great spot - so that we can ship this fix faster I've split this out into #1544 (and credited you in that PR) - I hope you don't mind

tombuildsstuff:
hey @sophos-jeff 

Thanks for this PR - apologies for the delayed review here, we needed to get the AzureAD Application Resource in first so that we could work out how's best to proceed with this.

I've spent a little time looking into how we'd add an AzureAD Application Data Source and I think there's two Data Sources needed here:

- `azurerm_azuread_application` - which returns the AD Application based on it's ID or Name
- `azurerm_azuread_applications` - which returns a list of Object ID's, which can then be used in conjunction with the `azurerm_azuread_application` data source, something like this:

```
data "azurerm_azuread_applications" "examples" {
  # could also be name_prefixes, or name_regex
  name_prefix = "sample-ms"
}


data "azurerm_azuread_application" "example" {
  count = "${len(data.azurerm_azuread_applications.examples.object_ids)}"
  object_id = "${data.azurerm_azuread_applications.examples.object_ids[count.index]}"
}

output "application_names" {
  value = "${azurerm_azuread_application.example.*.name}"
}
```

Whilst I've been working through this, I've ended up prototyping enough of `azurerm_azuread_application` that I'm going to open a PR for it - in which case can we focus this PR on supporting the `azurerm_azuread_applications` resource. What do you think of the above proposal?

Thanks!

sophos-jeff:
That sounds good @tombuildsstuff 

If you can please tag me in your PR so I am aware of it and I will pivot this to be azurerm_azuread_applications (Or actualyl probably close this one and open a new one for applications.  Ill let you make the first step here though and will follow your lead.

tombuildsstuff:
@sophos-jeff sorry I'd missed I'd not replied to this - support for the AD Application Data Source was added in #1552

tombuildsstuff:
hey @sophos-jeff 

Support for AzureAD Applications and Service Principals is now available in Terraform (and there's Data Sources which allow filtering for these by name) - as such I believe the purpose of this PR has changed and I'm going to close this PR for the moment. If you're still looking to have support for filtering Azure AD Applications based on the prefix of the name - would you mind opening a new Issue about this / opening a PR with a new Data Source specifically for that?

Thanks!

PR #1350, opened 2018-06-06
Fix storage_account data source documentation

### LINKED ISSUE #1316: Storage account kind documentation
The documentation don't mention StorageV2 kind. I needed to go through the issues/prs and the code to know how to use StorageV2 as a kind for my storage account.

In the [docs](https://www.terraform.io/docs/providers/azurerm/d/storage_account.html#account_kind) we have 

>account_kind - Defines the Kind of account, either BlobStorage or Storage.

Where is missing the StorageV2 option.

katbyte:
Fixes #1316

### DIFF
--- a/website/docs/d/storage_account.html.markdown
+++ b/website/docs/d/storage_account.html.markdown
@@ -35,13 +35,13 @@ output "storage_account_tier" {
 
 * `location` - The Azure location where the Storage Account exists
 
-* `account_kind` - Defines the Kind of account, either `BlobStorage` or `Storage`.
+* `account_kind` - The Kind of account.
 
-* `account_tier` - Defines the Tier of this storage account.
+* `account_tier` - The Tier of this storage account.
 
-* `account_replication_type` - Defines the type of replication used for this storage account.
+* `account_replication_type` - The type of replication used for this storage account.
 
-* `access_tier` - Defines the access tier for `BlobStorage` accounts.
+* `access_tier` - The access tier for `BlobStorage` accounts.
 
 * `enable_blob_encryption` - Are Encryption Services are enabled for Blob storage? See [here](https://azure.microsoft.com/en-us/documentation/articles/storage-service-encryption/)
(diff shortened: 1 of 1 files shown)


website/docs/d/storage_account.html.markdown
```diff
@@ -35,13 +35,13 @@ output "storage_account_tier" {
 
 * `location` - The Azure location where the Storage Account exists
 
-* `account_kind` - Defines the Kind of account, either `BlobStorage` or `Storage`.
```
metacpp: #1316 raised the concern on missed option of StorageV2. Is it correct that we directly delete all the options in the documentation? Then where will user find these options? Shall we provide a link ?

website/docs/d/storage_account.html.markdown
```diff
@@ -35,13 +35,13 @@ output "storage_account_tier" {
...
+* `account_kind` - The Kind of account.
 
-* `account_tier` - Defines the Tier of this storage account.
+* `account_tier` - The Tier of this storage account.
 
-* `account_replication_type` - Defines the type of replication used for this storage account.
+* `account_replication_type` - The type of replication used for this storage account.
 
-* `access_tier` - Defines the access tier for `BlobStorage` accounts.
```
metacpp: We should remove "Defines" in all fields' description, is it a documentation style? Do we need to update the other documents?

website/docs/d/storage_account.html.markdown
katbyte: This is a datasource, not a resource so we don't to specify valid values. 

There is an argument to be said for listing possible values, however currently most of the documentation for datasources don't so for consistencies sake I have removed them.

website/docs/d/storage_account.html.markdown
katbyte: As mentioned above this is a datasource, so it doesn't define the access tier, just returns what the access tier is.

PR #1380, opened 2018-06-12
azurerm_storage_account encryption now on by default

### LINKED ISSUE #1028: Default for undefined booleans without default (azurerm_storage_account/enable_blob_encryption)
### Terraform Version
Terraform v0.11.5
+ provider.azurerm v1.3.0

### Affected Resource(s)
- azurerm_storage_account

```hcl
resource "azurerm_storage_account" "backend" {
  name                     = "mybackend"
  resource_group_name      = "${azurerm_resource_group.backend.name}"
  location                 = "${azurerm_resource_group.backend.name.location}"
  account_kind             = "StorageV2"
  account_replication_type = "GRS"
}
```

I was wondering how TF handles boolean values that are set to `computed` if there is no `default` present.
The one I am especially interested in is `enable_blob_encryption` [option](https://www.terraform.io/docs/providers/azurerm/r/storage_account.html#enable_blob_encryption). I assume since the [code](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/data_source_storage_account.go#L64) does not specify a `default: true` it will go to [0 -> false](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L518).

However since Azure enforces encryption on all storage accounts per default I wonder how Terraform behaves in this scenario.

katbyte:
Fixes #1028

### DIFF
--- a/azurerm/resource_arm_storage_account.go
+++ b/azurerm/resource_arm_storage_account.go
@@ -127,13 +127,13 @@ func resourceArmStorageAccount() *schema.Resource {
 			"enable_blob_encryption": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Computed: true,
+				Default:  true,
 			},
 
 			"enable_file_encryption": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Computed: true,
+				Default:  true,
 			},
 
 			"enable_https_traffic_only": {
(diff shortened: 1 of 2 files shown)


azurerm/resource_arm_storage_account.go
```diff
@@ -296,20 +297,17 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
 				Services: &storage.EncryptionServices{
 					Blob: &storage.EncryptionService{
 						Enabled: utils.Bool(enableBlobEncryption),
-					}},
+					},
+					File: &storage.EncryptionService{
+					Enabled: utils.Bool(enableFileEncryption),
```
mbfrahry: Fmting issue here

PR #1391, opened 2018-06-14
enable_partitioning not supported servicebus Premium Sku.

lrxtom2:
According to azure-doc last update, enable_partitioning is not supported in a Premium tier namespace.

reference:
In a Premium tier namespace, partitioning entities is not supported. However, you can still create Service Bus queues and topics in 1, 2, 3, 4, 5, 10, 20, 40, or 80-GB sizes (the default is 1 GB). You can see the size of your queue or topic by looking at its entry on the Azure portal, in the Overview blade for that entity.
https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning

### DIFF
--- a/azurerm/resource_arm_servicebus_queue.go
+++ b/azurerm/resource_arm_servicebus_queue.go
@@ -166,10 +166,9 @@ func resourceArmServiceBusQueueCreateUpdate(d *schema.ResourceData, meta interfa
 		return nsErr
 	}
 
-	// Enforce Premium namespace to have partitioning enabled in Terraform. It is always enabled in Azure for
-	// Premium SKU.
-	if namespace.Sku.Name == servicebus.Premium && !d.Get("enable_partitioning").(bool) {
-		return fmt.Errorf("ServiceBus Queue (%s) must have Partitioning enabled for Premium SKU", name)
+	// In a Premium tier namespace, partitioning entities is not supported.
+	if namespace.Sku.Name == servicebus.Premium && d.Get("enable_partitioning").(bool) {
+		return fmt.Errorf("ServiceBus Queue (%s) does not support Partitioning enabled for Premium SKU", name)
 	}
 
 	// Enforce Premium namespace to have Express Entities disabled in Terraform since they are not supported for
--- a/azurerm/resource_arm_servicebus_queue.go
+++ b/azurerm/resource_arm_servicebus_queue.go
@@ -172,7 +172,7 @@ func resourceArmServiceBusQueueCreateUpdate(d *schema.ResourceData, meta interfa
 		return fmt.Errorf("ServiceBus Queue (%s) does not support Express Entities in Premium SKU and must be disabled", name)
 	}
 
-	_, err := client.CreateOrUpdate(ctx, resourceGroup, namespaceName, name, parameters)
+	_, err = client.CreateOrUpdate(ctx, resourceGroup, namespaceName, name, parameters)
 	if err != nil {
 		return err
 	}
--- a/azurerm/resource_arm_servicebus_queue.go
+++ b/azurerm/resource_arm_servicebus_queue.go
@@ -161,14 +161,9 @@ func resourceArmServiceBusQueueCreateUpdate(d *schema.ResourceData, meta interfa
 	// We need to retrieve the namespace because Premium namespace works differently from Basic and Standard,
 	// so it needs different rules applied to it.
 	namespacesClient := meta.(*ArmClient).serviceBusNamespacesClient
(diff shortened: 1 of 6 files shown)


azurerm/resource_arm_servicebus_topic.go
```diff
@@ -151,6 +151,19 @@ func resourceArmServiceBusTopicCreate(d *schema.ResourceData, meta interface{})
 		parameters.SBTopicProperties.DuplicateDetectionHistoryTimeWindow = utils.String(duplicateWindow)
 	}
 
+	// We need to retrieve the namespace because Premium namespace works differently from Basic and Standard,
+	// so it needs different rules applied to it.
+	namespacesClient := meta.(*ArmClient).serviceBusNamespacesClient
+	namespace, nsErr := namespacesClient.Get(ctx, resourceGroup, namespaceName)
```
jeffreyCline: Super minor, but can we make nsErr just err to keep it consistent in the codebase?

azurerm/resource_arm_servicebus_queue.go
```diff
@@ -166,10 +166,9 @@ func resourceArmServiceBusQueueCreateUpdate(d *schema.ResourceData, meta interfa
 		return nsErr
 	}
 
-	// Enforce Premium namespace to have partitioning enabled in Terraform. It is always enabled in Azure for
-	// Premium SKU.
-	if namespace.Sku.Name == servicebus.Premium && !d.Get("enable_partitioning").(bool) {
-		return fmt.Errorf("ServiceBus Queue (%s) must have Partitioning enabled for Premium SKU", name)
+	// In a Premium tier namespace, partitioning entities is not supported.
+	if namespace.Sku.Name == servicebus.Premium && d.Get("enable_partitioning").(bool) {
+		return fmt.Errorf("ServiceBus Queue (%s) does not support Partitioning enabled for Premium SKU", name)
```
jeffreyCline: Since we are changing the behavior can we update the documentation too with this change?

tombuildsstuff:
> I'm not sure of the best way forward, but we need to make sure we don't force a change upon users who don't want it yet

@katbyte FWIW we'd tend to normally leave this as a supported property and add a `~> **NOTE:** this is only available for existing accounts`

azurerm/resource_arm_servicebus_topic_test.go
```diff
@@ -165,7 +165,7 @@ func TestAccAzureRMServiceBusTopic_enablePartitioningPremium(t *testing.T) {
 			{
 				Config: postConfig,
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(resourceName, "enable_partitioning", "true"),
+					resource.TestCheckResourceAttr(resourceName, "enable_partitioning", "false"),
```
metacpp: Yo'd better validate the behavior  if `enable_partitioning` is true.

metacpp:
@lrxtom2 what's user impact if we don't make the fix? Is enable_partitioning still valid for Premium at API level?

katbyte:
@metacpp, @lrxtom2,

An easy fix would be to just allow it for all SKUs, document the limitation and let the API handle the validation.

azurerm/resource_arm_servicebus_topic_test.go
tombuildsstuff: This isn't possible due to the API behaviour change/limitation

azurerm/resource_arm_servicebus_topic.go
```diff
@@ -151,6 +151,19 @@ func resourceArmServiceBusTopicCreate(d *schema.ResourceData, meta interface{})
...
+	if nsErr != nil {
+		return nsErr
+	}
+
+	// In a Premium tier namespace, partitioning entities is not supported.
+	if namespace.Sku.Name == servicebus.Premium && d.Get("enable_partitioning").(bool) {
+		return fmt.Errorf("ServiceBus Queue (%s) does not support Partitioning enabled for Premium SKU", name)
+	}
```
tombuildsstuff: gonna push a commit to remove this, since existing Namespaces should continue working

azurerm/resource_arm_servicebus_queue.go
tombuildsstuff: gonna push a commit to remove this, since existing Namespaces should continue working

tombuildsstuff:
@lrxtom2 so that we can get this merged I've pushed a commit with the comments raised by @katbyte 
and @jeffreyCline - I hope you don't mind.

PR #1393, opened 2018-06-14
Moving from net.dial to http.get so environment proxy settings can be picked up

lstolyarov:
This fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/604

Tested with and without a proxy.


katbyte:
test pass: 
screen shot 2018-06-14 at 10 29 23

(fail test is not new)

PR #1399, opened 2018-06-14
normalize acctest resource group name: acctestrg -> acctestRG

katbyte:
104 instances of acctestrg, over 400 of acctestRG


azurerm/resource_arm_policy_assignment_test.go
```diff
@@ -123,7 +123,7 @@ POLICY_RULE
 }
 
 resource "azurerm_resource_group" "test" {
-  name = "acctestrgpol-%d"
+  name = "acctestRGpol-%d"
```
jeffreyCline: Is there a reason why we are using `acctestRGpol` here and not `acctestRG`?

azurerm/resource_arm_policy_assignment_test.go
```diff
@@ -171,7 +171,7 @@ PARAMETERS
...
```
jeffreyCline: Same question here.

PR #1415, opened 2018-06-20
Add Standard_Microsoft for CDN profile SKU type

IronCore864:
Adding support to "Standard Microsoft" pricing tier.

As stated [HERE](https://azure.microsoft.com/en-us/pricing/details/cdn/) in MS doc.

cdn.StandardMicrosoft added in this [PR HERE](https://github.com/Azure/azure-sdk-for-go/pull/2082)

CI failing because of this, and I can't update the model in this repo because of govendor status check in makefile.

### DIFF
--- a/azurerm/resource_arm_cdn_profile.go
+++ b/azurerm/resource_arm_cdn_profile.go
@@ -39,6 +39,7 @@ func resourceArmCdnProfile() *schema.Resource {
 				ValidateFunc: validation.StringInSlice([]string{
 					string(cdn.StandardAkamai),
 					string(cdn.StandardVerizon),
+					string(cdn.StandardMicrosoft),
 					string(cdn.PremiumVerizon),
 				}, true),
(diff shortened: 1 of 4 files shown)


tombuildsstuff:
hey @IronCore864 

Thanks for this PR :)

Since this value isn't available in the SDK at present - we should be able to add this as a string directly; e.g.

```
  string(cdn.StandardAkamai),
  string(cdn.StandardVerizon),
  // TODO: replace with a constant in the SDK
  "Standard_Microsoft",
  string(cdn.PremiumVerizon),
}, true),
```

Once that's available in the SDK - we can then switch over to using the newly available Constant/Enum. This PR otherwise LGTM, in order to get this merged can we revert the changes to the vendored SDK and add an acceptance test for this new CDN type?

Thanks!

tombuildsstuff:
hey @IronCore864 

I hope you don't mind - but I've made the above changes in #1465 which I'm going to close this PR in favour of so that we can get this merged.

Thanks!

PR #1457, opened 2018-06-28
allow postgres version 10.0

featheredtoast:
We're attempting to get a Postgres 10 instance up through Terraform, but are unable to due to restrictive version checks.

This is a fix that adds support "10.0" as a valid version string until the azure go api adds a proper "10.0" enum.


featheredtoast:
OK done - Updated the docs, added the requested TODO comment.

featheredtoast:
No worries, just happy to have gotten this in quickly - thanks for fixing it up! ❤️

Issue #1434, opened 2018-06-24
Feature Request : addonProfiles and networkProfile for Azure Kubernetes Service (AKS)


idelix:
AKS now allows to place cluster in custom vnet and also to integrate monitoring with log analytics workspace, this is now available in ARM template but seems to be missing from terraform resource: azurerm_kubernetes_cluster.

Just for reference here is working example of ARM Template with both addonProfiles and networkProfile:

```
        {
            "name": "[variables('aksClusterName')]",
            "type": "Microsoft.ContainerService/managedClusters",
            "apiVersion": "2018-03-31",
            "location": "[resourceGroup().location]",
            "tags": {
                "Owner": "[variables('tagName')]"
            },
            "dependsOn": [
                "[concat('Microsoft.OperationalInsights/workspaces/', variables('laName'))]",
                "[concat('Microsoft.Network/virtualNetworks/', variables('vnetName'))]"
            ],
            "properties": {
                "mode": "Incremental",
                "dnsPrefix": "[variables('aksDnsPrefix')]",
                "kubernetesVersion": "[variables('aksVersion')]",
                "agentPoolProfiles": [
                    {
                        "name": "[variables('aksPoolName')]",
                        "osDiskSizeGB": "[variables('aksOsDiskSizeGB')]",
                        "count": "[variables('aksVmCount')]",
                        "vmSize": "[variables('aksVmSize')]",
                        "osType": "Linux",
                        "vnetSubnetID": "[variables('subnetId')]",
                        "maxPods": 30,
                        "storageProfile": "ManagedDisks"
                    }
                ],
                "linuxProfile": {
                    "adminUsername": "[parameters('aksUserName')]",
                    "ssh": {
                        "publicKeys": [
                            {
                                "keyData": "[variables('aksKeyData')]"
                            }
                        ]
                    }
                },
                "servicePrincipalProfile": {
                    "clientId": "[parameters('aksServicePrincipalClientId')]",
                    "secret": "[parameters('aksServicePrincipalSecret')]"
                },
                "addonProfiles": {
                    "omsagent": {
                        "enabled": true,
                        "config": {
                            "logAnalyticsWorkspaceResourceID": "[resourceId('Microsoft.OperationalInsights/workspaces', variables('laName'))]"
                        }
                    }
                },
                "networkProfile": {
                    "networkPlugin": "[variables('aksNetworkPlugin')]",
                    "serviceCidr": "[variables('aksServiceCidr')]",
[...]

Issue #1389, opened 2018-06-13
Update azurerm_kubernetes_cluster for GA


holmesb:
AKS has gone [GA](https://azure.microsoft.com/en-us/blog/azure-kubernetes-service-aks-ga-new-regions-new-features-new-productivity/).  Main functionality now lacking is the [basic\advanced network setting](https://docs.microsoft.com/en-us/azure/aks/networking-overview) and the various parameters if advanced is chosen:
- VNet
- Subnet
- Service address range
- DNS IP
- Docker bridge IP

torresdal:
I would gladly contribute to this, but looks like we are dependant on [azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go/tree/master/services/containerservice/mgmt) to include the new package from [autorest containerservices](https://github.com/Azure/azure-rest-api-specs/tree/master/specification/containerservices/resource-manager) (package-2018-03)?

Currently `azure-sdk-for-go` only has the 2017-09-30 package.

I haven't contributed to this project before, so I'm new to how the dependency to the `azure-sdk-for-go`-project and `azure-rest-api-specs`-project works. Anyone know how/when a new autorest package will be generated?

robinsk:
Right, this seems relevant for us. We require advanced networking, and face problems when creating an AKS cluster from Terraform.

When creating the cluster from the Azure Portal, there are 31 "connected devices" on the vnet per NIC (i.e. per host/vm in the cluster), which [makes sense with maximum 30 pods per host](https://docs.microsoft.com/en-us/azure/aks/networking-overview).

When we create the cluster from Terraform, specifing `vnet_subnet_id`, there's only one connected device per host, so none of our pods have any networking.

twem:
Also agent_pool_profile, os_type=windows has a dependency on windowsProfile but that is not documented or in the code.

``
Error code  
InvalidParameter  
Message  
Required parameter windowsProfile is missing (null).
``

torresdal:
@twem I think Windows agent is still in private preview. At least according to this: https://docs.microsoft.com/en-us/azure/aks/faq

> To run Windows Server containers, you need to run Windows Server-based nodes. Windows Server-based nodes are currently in private preview. If you need to run Windows Server containers on Kubernetes in Azure outside of the preview, please see the documentation for acs-engine.

[...]

Issue #1276, opened 2018-05-22
Support for AKS HTTP Application Routing


rlucioni:
The [HTTP Application Routing](https://docs.microsoft.com/en-us/azure/aks/http-application-routing) add-on can be enabled through the Azure portal when deploying an AKS cluster. It would be great if enabling this was an option when creating new `azurerm_kubernetes_cluster` resources (e.g., via a Boolean argument like `enable_http_application_routing`).

tombuildsstuff:
hey @rlucioni 

Thanks for opening this issue :)

We took a look into this recently and this is available in the new version of the AKS API - however there's no matching Swagger for that at this time. @dcaro was following up internally about getting this available - once that's done and a new version of [the Azure SDK for Go](https://github.com/Azure/azure-sdk-for-go) is available - we can take a look at support this.

Thanks!

rlucioni:
Great, thanks @tombuildsstuff. Looking forward to seeing this supported.

tombuildsstuff:
To provide an update here, the Swagger for this has been merged: https://github.com/Azure/azure-rest-api-specs/pull/3131

lfshr:
Not quite as simple as the boolean. This is an AddonProfile. - Related to #1434

@mb290 nodeResourceGroup addition is a much simpler one. @tombuildsstuff I can bundle up the nodeResourceGroup with #1479 if you like, it's a relatively small change.

lfshr:
@tombuildsstuff this functionality is provided by addonProfiles. I think we should move this alongside #1434

tombuildsstuff:
@lfshr 👍 closing this in favour of #1434

PR #1479, opened 2018-07-03
AKS: NetworkProfile / Advanced Networking

lfshr:
Added `network_profile` to `kubernetes_cluster` resource.
#1389 #1434

### DIFF
--- a/azurerm/data_source_kubernetes_cluster.go
+++ b/azurerm/data_source_kubernetes_cluster.go
@@ -161,6 +161,39 @@ func dataSourceArmKubernetesCluster() *schema.Resource {
 				},
 			},
 
+			"network_profile": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"network_plugin": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"service_cidr": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"dns_service_ip": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"docker_bridge_cidr": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"pod_cidr": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+
 			"tags": tagsForDataSourceSchema(),
 		},
 	}
@@ -215,6 +248,10 @@ func dataSourceArmKubernetesClusterRead(d *schema.ResourceData, meta interface{}
 		if err := d.Set("service_principal", servicePrincipal); err != nil {
 			return fmt.Errorf("Error setting `service_principal`: %+v", err)
 		}
+	}
+
+	if networkProfile := resp.NetworkProfile; networkProfile != nil {
+		d.Set("network_profile", flattenKubernetesClusterDataSourceNetworkProfile(resp.NetworkProfile))
 	}
 
 	kubeConfigRaw, kubeConfig := flattenKubernetesClusterDataSourceAccessProfile(&profile)
@@ -347,6 +384,18 @@ func flattenKubernetesClusterDataSourceKubeConfig(config kubernetes.KubeConfig) 
 	values["client_certificate"] = user.ClientCertificteData
 	values["client_key"] = user.ClientKeyData
 	values["cluster_ca_certificate"] = cluster.ClusterAuthorityData
+
+	return []interface{}{values}
+}
+
+func flattenKubernetesClusterDataSourceNetworkProfile(profile *containerservice.NetworkProfile) []interface{} {
+	values := make(map[string]interface{})
+
+	values["network_plugin"] = profile.NetworkPlugin
+	values["service_cidr"] = profile.ServiceCidr
+	values["dns_service_ip"] = profile.DNSServiceIP
+	values["docker_bridge_cidr"] = profile.DockerBridgeCidr
+	values["pod_cidr"] = profile.PodCidr
 
 	return []interface{}{values}
 }
--- a/azurerm/data_source_kubernetes_cluster.go
+++ b/azurerm/data_source_kubernetes_cluster.go
@@ -250,8 +250,10 @@ func dataSourceArmKubernetesClusterRead(d *schema.ResourceData, meta interface{}
 		}
 	}
 
-	if networkProfile := resp.NetworkProfile; networkProfile != nil {
-		d.Set("network_profile", flattenKubernetesClusterDataSourceNetworkProfile(resp.NetworkProfile))
+	networkProfile := flattenKubernetesClusterDataSourceNetworkProfile(resp.NetworkProfile)
+
+	if err := d.Set("network_profile", networkProfile); err != nil {
+		return fmt.Errorf("Error setting `network_profile`: %+v", err)
 	}
 
 	kubeConfigRaw, kubeConfig := flattenKubernetesClusterDataSourceAccessProfile(&profile)
@@ -392,10 +394,22 @@ func flattenKubernetesClusterDataSourceNetworkProfile(profile *containerservice.
 	values := make(map[string]interface{})
 
 	values["network_plugin"] = profile.NetworkPlugin
-	values["service_cidr"] = profile.ServiceCidr
-	values["dns_service_ip"] = profile.DNSServiceIP
-	values["docker_bridge_cidr"] = profile.DockerBridgeCidr
-	values["pod_cidr"] = profile.PodCidr
+
+	if profile.ServiceCidr != nil {
+		values["service_cidr"] = *profile.ServiceCidr
+	}
+
+	if profile.DNSServiceIP != nil {
+		values["dns_service_ip"] = *profile.DNSServiceIP
+	}
+
+	if profile.DockerBridgeCidr != nil {
+		values["docker_bridge_cidr"] = *profile.DockerBridgeCidr
+	}
+
+	if profile.PodCidr != nil {
+		values["pod_cidr"] = *profile.PodCidr
+	}
 
 	return []interface{}{values}
 }
--- a/azurerm/data_source_kubernetes_cluster.go
+++ b/azurerm/data_source_kubernetes_cluster.go
@@ -171,11 +171,6 @@ func dataSourceArmKubernetesCluster() *schema.Resource {
 							Computed: true,
 						},
 
-						"network_policy": {
-							Type:     schema.TypeString,
-							Computed: true,
-						},
-
 						"service_cidr": {
 							Type:     schema.TypeString,
 							Computed: true,
@@ -399,7 +394,6 @@ func flattenKubernetesClusterDataSourceNetworkProfile(profile *containerservice.
 	values := make(map[string]interface{})
 
 	values["network_plugin"] = profile.NetworkPlugin
-	values["network_policy"] = profile.NetworkPolicy
 
 	if profile.ServiceCidr != nil {
 		values["service_cidr"] = *profile.ServiceCidr
--- a/azurerm/data_source_kubernetes_cluster.go
+++ b/azurerm/data_source_kubernetes_cluster.go
@@ -171,6 +171,11 @@ func dataSourceArmKubernetesCluster() *schema.Resource {
 							Computed: true,
 						},
 
+						"network_policy": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
 						"service_cidr": {
 							Type:     schema.TypeString,
 							Computed: true,
@@ -394,6 +399,7 @@ func flattenKubernetesClusterDataSourceNetworkProfile(profile *containerservice.
 	values := make(map[string]interface{})
 
 	values["network_plugin"] = profile.NetworkPlugin
+	values["network_policy"] = profile.NetworkPolicy
 
 	if profile.ServiceCidr != nil {
 		values["service_cidr"] = *profile.ServiceCidr
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -220,21 +220,25 @@ func resourceArmKubernetesCluster() *schema.Resource {
 							Type:     schema.TypeString,
 							Default:  "azure",
 							Optional: true,
-						},
-
-						"dns_service_ip": {
-							Type:     schema.TypeString,
-							Required: true,
-						},
-
-						"docker_bridge_cidr": {
-							Type:     schema.TypeString,
-							Required: true,
+							ForceNew: true,
 						},
 
 						"service_cidr": {
 							Type:     schema.TypeString,
 							Required: true,
+							ForceNew: true,
+						},
+
+						"dns_service_ip": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+
+						"docker_bridge_cidr": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
 						},
 
 						"pod_cidr": {
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -30,22 +30,6 @@ func resourceArmKubernetesCluster() *schema.Resource {
 					return nil
 				}
 
-				// ensure all Agent Pool Profiles are attached to a network
-				if rawAgentPools, exists := diff.GetOk("agent_pool_profile"); exists {
-					attachedToASubnet := false
-					for _, ap := range rawAgentPools.([]interface{}) {
-						agentPool := ap.(map[string]interface{})
-						if subnetId, ok := agentPool["vnet_subnet_id"]; ok && subnetId != "" {
-							attachedToASubnet = true
-							break
-						}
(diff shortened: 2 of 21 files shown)


lfshr:
@tombuildsstuff I couldn't get the tests running properly (wouldn't fail when it should have), may need some assistance here. Also think I should merge the two tests into one.

lfshr:
@tombuildsstuff the resource and data source is good to review. Just about to add an example into the examples

lfshr:
Result from pinging node agent from another vm within the same vnet: 
```
liam@test-networking:~$ ping 10.1.0.35
PING 10.1.0.35 (10.1.0.35) 56(84) bytes of data.
64 bytes from 10.1.0.35: icmp_seq=1 ttl=64 time=1.50 ms
64 bytes from 10.1.0.35: icmp_seq=2 ttl=64 time=0.853 ms
64 bytes from 10.1.0.35: icmp_seq=3 ttl=64 time=0.787 ms
64 bytes from 10.1.0.35: icmp_seq=4 ttl=64 time=2.14 ms
64 bytes from 10.1.0.35: icmp_seq=5 ttl=64 time=0.884 ms
64 bytes from 10.1.0.35: icmp_seq=6 ttl=64 time=0.752 ms
64 bytes from 10.1.0.35: icmp_seq=7 ttl=64 time=0.730 ms
64 bytes from 10.1.0.35: icmp_seq=8 ttl=64 time=0.692 ms
64 bytes from 10.1.0.35: icmp_seq=9 ttl=64 time=0.682 ms
64 bytes from 10.1.0.35: icmp_seq=10 ttl=64 time=0.893 ms
64 bytes from 10.1.0.35: icmp_seq=11 ttl=64 time=0.745 ms
```

Nothing more satisfying :smile:

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -209,6 +209,48 @@ func resourceArmKubernetesCluster() *schema.Resource {
 				Set: resourceAzureRMKubernetesClusterServicePrincipalProfileHash,
 			},
 
+			"network_profile": {
```
tombuildsstuff: can [we document this block/these new fields here](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/kubernetes_cluster.html.markdown#L127)? (I've had to link to the Blame since you can't link to a regular line in a Markdown file in Github ¯\_(ツ)_/¯)

azurerm/data_source_kubernetes_cluster.go
```diff
@@ -161,6 +161,39 @@ func dataSourceArmKubernetesCluster() *schema.Resource {
 				},
 			},
 
+			"network_profile": {
```
tombuildsstuff: can [we document this block/these new fields here](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/d/kubernetes_cluster.html.markdown#L52)? (I've had to link to the Blame since you can't link to a regular line in a Markdown file in Github ¯\_(ツ)_/¯)

tombuildsstuff:
@lfshr hope you don't mind but I've pushed a commit to fix the build so we can run the tests

lfshr:
@tombuildsstuff Decided to run a few more tests. I seem to still have the communication error when using kubenet (default plugin when a network_profile block isn't specified) and a subnet specified in agent_pool_profile. This may be down to my own incompetencies though.

There's also a non-descriptive error that comes back from the sdk when using azure as the network_plugin and the agent_pool_profile subnet is not specified:
> azurerm_kubernetes_cluster.test: containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="CidrStringParseError" Message="Cidr  is invalid" Target="networkProfile.serviceCIDR"

This isn't actually a parse error, this error should say something along the lines of  "agent_pool_profile.0.vnet_subnet_id is required when using azure network_profile". I'm going to open up an issue with MS to fix the error message. Do we also want some sort of validation in place on the terraform side that will spit out a meaningful message?

lfshr:
@tombuildsstuff I have that commit locally. Forgot to push it. 😞

lfshr:
@tombuildsstuff this is good to go - minus the above mentioned notes.

mb290:
hi @lfshr did nodeResourceGroup get included in this in the end? I know #1276 was closed in favor of #1434 but I can't see the new output in either of the updated files.

lfshr:
@mb290 it's not included at the moment. @tombuildsstuff are you happy for node ResourceGroup to go in with this PR? I'm happy to do it in another PR.

mb290:
@lfshr no problem if not, I just don't want it to get lost :). I'm also more than happy to do a PR myself and finally put my mark on the project.

lfshr:
@tombuildsstuff what's the update on this PR?

lfshr:
@mb290 probably open this up as a separate PR. This one seems good to go IMO.
It's a fairly small change, so feel free! I'll take it on in a few days if not. General rule of thumb for myself is no more then 2 PRs open at one time, lest I get lost in a sea of PRs.

tombuildsstuff:
@lfshr

> @tombuildsstuff are you happy for node ResourceGroup to go in with this PR? I'm happy to do it in another PR.

Sorry this got missed in the sea of Github notifications :(
I'd probably agree that belongs in a separate PR - but as it's minor we should be able to merge it pretty easily :)

tombuildsstuff:
> This isn't actually a parse error, this error should say something along the lines of "agent_pool_profile.0.vnet_subnet_id is required when using azure network_profile". I'm going to open up an issue with MS to fix the error message. Do we also want some sort of validation in place on the terraform side that will spit out a meaningful message?

I'm going to suggest we document this for the moment; given this is open-ended

tombuildsstuff:
hey @lfshr 

I've kicked off the tests for this but there's a failure in the `TestAccAzureRMKubernetesCluster_advancedNetworking ` test - we're seeing this:

```
network_profile.0.pod_cidr:                 "" => "10.244.0.0/24" (forces new resource)
```

It appears that no response is being returned from the API for the `pod_cidr` field - do you happen to know if that's the case only for the `azure` networking, or if this is for the `kubeconfig` networking too?

Thanks!

lfshr:
@tombuildsstuff pod_cidr should not be required when the network_profile is set to "azure" but it is. I've opened an issue here: https://github.com/Azure/azure-sdk-for-go/issues/2178

Sorry, I should have picked up on this issue. (and probably should not have included my workaround)

lfshr:
@tombuildsstuff I don't really know how to deal with this one. The API will return nothing in this field (because the network_plugin is azure) but the field is required when running the CreateOrUpdate command :confused:

lfshr:
A possible workaround for this could be to set the default value of the pod_cidr on the read function. eg.

```hcl
if profile.PodCidr != nil {
	values["pod_cidr"] = *profile.PodCidr
} else {
	values["pod_cidr"] = "10.244.0.0/24"
}
```

@tombuildsstuff what do you think? Would adding a Computed: true onto the field fix this? :confused:

jrudley:
Hi, Any update on this PR? Thanks

tombuildsstuff:
@lfshr thanks for opening the SDK issue, so one option in the interim would be to make it optional and default it to an empty string, which I think that should allow us to work around the above issue?

lfshr:
@tombuildsstuff would passing an empty string to the API get round the issue? I think it does a regex. I can validate tonight.

tombuildsstuff:
@lfshr that's a good point - taking a look you're right:

```
{Target: "parameters.ManagedClusterProperties.NetworkProfile", Name: validation.Null, Rule: false,
						Chain: []validation.Constraint{{Target: "parameters.ManagedClusterProperties.NetworkProfile.PodCidr", Name: validation.Null, Rule: false,
							Chain: []validation.Constraint{{Target: "parameters.ManagedClusterProperties.NetworkProfile.PodCidr", Name: validation.Pattern, Rule: `^([0-9]{1,3}\.){3}[0-9]{1,3}(\/([0-9]|[1-2][0-9]|3[0-2]))?$`, Chain: nil}}},
							{Target: "parameters.ManagedClusterProperties.NetworkProfile.ServiceCidr", Name: validation.Null, Rule: false,
```

in which case we need a Swagger/SDK fix to be able to ship this, unfortunately. I'll follow up in the other issue :)

joshgav:
Hi @tombuildsstuff, is it possible to not specify "podCidr" at all when the `azure` network type is chosen? An empty string fails the regex, but a non-existent field shouldn't (right @marstr?).

azurerm/resource_arm_kubernetes_cluster_test.go
```diff
@@ -337,6 +359,67 @@ resource "azurerm_kubernetes_cluster" "test" {
...
+    vnet_subnet_id = "${azurerm_subnet.test.id}"
+  }
+
+  service_principal {
+    client_id     = "%s"
+    client_secret = "%s"
+	}
+	
+  network_profile {
+		network_plugin     = "azure"
```
metacpp: Use whitespaces instead of tabs here.

metacpp:
@tombuildsstuff a quick question here:
```
network_profile.0.pod_cidr:                 "" => "10.244.0.0/24" (forces new resource)
```
Why it was triggered here? An upgrade?

lfshr:
@metacpp it was triggered due to pod_cidr having a default value (workaround for issue Azure/azure-sdk-for-go#2178).

pod_cidr is not used with "Azure" network_plugin, therefor the SDK does not return a value for this field when it performs the read function. So the next time terraform is run, it thinks it's upgrading from nothing to the default value.

tombuildsstuff:
@lfshr sorry for the delay getting back to this, I'm chatting with @marstr at the moment to confirm if the workaround discussed in the SDK works :)

lfshr:
@tombuildsstuff no worries at all. I understand the annoyingness of the situation. 😄

tombuildsstuff:
@lfshr I've spent some time looking into this and it appears there's an issue in the API too: https://gist.github.com/tombuildsstuff/621488831c31edc7d5a413148c646234

From what I can see the API doesn't allow missing values for the `ServiceCidr` field (since the API defaults them to an empty string, see the link above) - however this field can't be set for the `azure` profile:

```
{
	"code": "CidrStringParseError",
	"message": "Cidr  is invalid",
	"target": "networkProfile.serviceCIDR"
}
```

when it is set, this requires that DNSServiceIP and DockerBridgeCidr are also set, which means I don't believe it's possible to provision an Azure network profile at this time via the API?

cc @marstr

lfshr:
@tombuildsstuff regarding your above issue please see the following: https://github.com/Azure/AKS/issues/506

metacpp:
@lfshr I did a small experimentation, even you pass `nil` to Azure Go SDK to bypass all client validation, successfully created AKS cluster, you will still meet the same issue as above due to default values returned in response

```
network_profile.#:                          "1" => "1"
network_profile.0.dns_service_ip:           "10.0.0.10" => "" (forces new resource)
network_profile.0.docker_bridge_cidr:       "172.17.0.1/16" => "" (forces new resource)
network_profile.0.network_plugin:           "azure" => "azure"
network_profile.0.service_cidr:             "10.0.0.0/16" => "" (forces newresource)
```

lfshr:
@metacpp do you think this solution will work if we set the default value to nil? Seems hacky, but may work. Unfortunately I can't test this until later tonight.

tombuildsstuff:
@metacpp that's not the result I saw last night with the same code unfortunately - and repeating it this morning it appears the API returns the error message I posted above:

```
PUT /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/tom-aksdev/providers/Microsoft.ContainerService/managedClusters/tom-akscluster?api-version=2018-03-31 HTTP/1.1
Host: management.azure.com
User-Agent: Go/go1.10.2 (amd64-darwin) go-autorest/v10.11.4 Azure-SDK-For-Go/v17.4.0 containerservice/2018-03-31;HashiCorp-Terraform-v0.11.3
Content-Length: 1001
Content-Type: application/json; charset=utf-8
Accept-Encoding: gzip

{
	"location": "westeurope",
	"name": "tom-akscluster",
	"properties": {
		"agentPoolProfiles": [{
			"name": "ohhai21",
			"count": 1,
			"vmSize": "Standard_F2",
			"osDiskSizeGB": 30,
			"dnsPrefix": "",
			"storageProfile": "ManagedDisks",
			"osType": "Linux"
		}],
		"dnsPrefix": "tomakscluster",
		"kubernetesVersion": "",
		"linuxProfile": {
			"adminUsername": "acctestuser1",
			"ssh": {
				"publicKeys": [{
					"keyData": "ssh-key-here"
				}]
			}
		},
		"networkProfile": {
			"networkPlugin": "azure",
			"serviceCidr": "10.10.0.0/16"
		},
		"servicePrincipalProfile": {
			"clientId": "00000000-0000-0000-0000-000000000000",
			"secret": "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
		}
	},
	"tags": {
		"Environment": "Production"
	}
}
```

returns the response:

```
{
	"code": "InternalOperationError",
	"message": "An error has occurred in subscription 00000000-0000-0000-0000-000000000000, resourceGroup: tom-aksdev request: ServiceCidr, DNSServiceIP, DockerBridgeCidr should all be empty or neither should be empty."
}
```

Are you sure you used the compiled version of the Provider?

lfshr:
@tombuildsstuff did you look at my above comments? The error you're receiving on the serviceCidr is a non-descriptive error. See Azure/AKS#506

tombuildsstuff:
@metacpp unfortunately that's not the result I saw last night with the same code - and repeating it this morning on a freshly created cluster it appears the API returns the error message I posted above is correct:

```
PUT /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/tom-aksdev/providers/Microsoft.ContainerService/managedClusters/tom-akscluster?api-version=2018-03-31 HTTP/1.1
Host: management.azure.com
User-Agent: Go/go1.10.2 (amd64-darwin) go-autorest/v10.11.4 Azure-SDK-For-Go/v17.4.0 containerservice/2018-03-31;HashiCorp-Terraform-v0.11.3
Content-Length: 972
Content-Type: application/json; charset=utf-8
Accept-Encoding: gzip

{
	"location": "westeurope",
	"name": "tom-akscluster",
	"properties": {
		"agentPoolProfiles": [{
			"name": "ohhai21",
			"count": 1,
			"vmSize": "Standard_F2",
			"osDiskSizeGB": 30,
			"dnsPrefix": "",
			"storageProfile": "ManagedDisks",
			"osType": "Linux"
		}],
		"dnsPrefix": "tomakscluster",
		"kubernetesVersion": "",
		"linuxProfile": {
			"adminUsername": "acctestuser1",
			"ssh": {
				"publicKeys": [{
					"keyData": "ssh-rsa KEY CONTENT GOES HERE"
				}]
			}
		},
		"networkProfile": {
			"networkPlugin": "azure"
		},
		"servicePrincipalProfile": {
			"clientId": "00000000-0000-0000-0000-000000000000",
			"secret": "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
		}
	},
	"tags": {
		"Environment": "Production"
	}
}
```

returns the response:

```
{
	"code": "InternalOperationError",
	"message": "An error has occurred in subscription 00000000-0000-0000-0000-000000000000, resourceGroup: tom-aksdev request: ServiceCidr, DNSServiceIP, DockerBridgeCidr should all be empty or neither should be empty."
}
```

Are you sure you used the compiled version of the Provider?

lfshr:
@tombuildsstuff did you look at my above comments? The error you're receiving on the serviceCidr is a non-descriptive error. See Azure/AKS#506
The above ARM template will not work. You need to supply the subnetId inside agentPoolProfile and the serviceCidr must be within this subnet.

tombuildsstuff:
@lfshr I read the first two links but missed the third, apologies. I'd suggest we reply within those specific issues to work around the lack of threading here?

> You need to supply the subnetId inside agentPoolProfile and the serviceCidr must be within this subnet.

🤦‍♂️ good spot - I'll re-test that now

tombuildsstuff:
@lfshr cool, so I can confirm setting that works. Given I've made those changes locally, I'm going to commit them (and then also add a couple more for acceptance tests for both the Azure network profile & the Kubenet network profile) - so that we can run the tests and ship this (I hope you don't mind!)

@metacpp thanks for the heads up :)

lfshr:
@tombuildsstuff not at all. Do what you need to do. Anything you need from my end?

metacpp:
@lfshr @tombuildsstuff You can see all my changes on top of @lfshr 's branch:
https://github.com/lfshr/terraform-provider-azurerm/pull/2/files

1. It bypasses the validation error.
2. It resolves the rerun of `terraform plan`.

It's just a reference. :-)

tombuildsstuff:
Resource Tests pass:

screenshot 2018-07-25 at 21 26 12

Data Source tests pass:

screenshot 2018-07-25 at 21 26 06

tombuildsstuff:
@lfshr nope, all good - the tests pass :) Thanks again for this PR - apologies for the confusion here.

metacpp:
@tombuildsstuff what're the conclusions here?

giggio:
Do we know when is 1.12 expected?

tombuildsstuff:
@giggio we don't give exact dates since things can change - but I have a feeling it'll probably be next week at this point.

PR #1487, opened 2018-07-04
 Updating to v18.0.0 of Azure/azure-sdk-for-go / v10.12.0 of Azure/go-autorest

tombuildsstuff:
Need to upgrade the SDK to use the new version of the HDInsights API


lfshr:
#1488 depends on this

PR #1498, opened 2018-07-05
new resource: azurerm_servicebus_namespace_authorization_rule

katbyte:
- New resource azurerm_servicebus_namespace_authorization_rule
- fixed service bus example
- added namespace name validation everywhere


azurerm/resource_arm_servicebus_namespace_authorization_rule.go
```diff
@@ -0,0 +1,217 @@
...
+	d.Set("resource_group_name", resGroup)
+
+	if err := d.Set("rights", flattenServiceBusAuthorizationRuleRights(resp.Rights)); err != nil {
+		return fmt.Errorf("Error flattening rights: %+v", err)
+	}
+
+	d.Set("primary_key", keysResp.PrimaryKey)
+	d.Set("primary_connection_string", keysResp.PrimaryConnectionString)
+	d.Set("secondary_key", keysResp.SecondaryKey)
+	d.Set("secondary_connection_string", keysResp.SecondaryConnectionString)
```
tombuildsstuff: can we access rights -> secondary_connection_string via the `[XX]Properties` object & nil-check them?

azurerm/resource_arm_servicebus_namespace_authorization_rule.go
```diff
@@ -0,0 +1,217 @@
...
+
+	if err != nil {
+		return fmt.Errorf("Error issuing Azure ARM delete request of ServiceBus Namespace Authorization Rule %q (Resource Group %q): %+v", name, resGroup, err)
+	}
+
+	return nil
+}
+
+func flattenServiceBusAuthorizationRuleRights(rights *[]servicebus.AccessRights) *schema.Set {
+	slice := make([]interface{}, 0, 0)
```
tombuildsstuff: _minor_ we don't actually need this last index

azurerm/resource_arm_servicebus_namespace_authorization_rule_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/services/servicebus/mgmt/2017-04-01/servicebus"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+	"strconv"
```
tombuildsstuff: _minor_ can we fix the imports here?

azurerm/resource_arm_servicebus_namespace_authorization_rule_test.go
```diff
@@ -0,0 +1,190 @@
...
+		},
+		{
+			Name:   "manage",
+			Rights: []string{string(servicebus.Listen), string(servicebus.Send), string(servicebus.Manage)},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.Name, func(t *testing.T) {
+			testAccAzureRMServiceBusNamespaceAuthorizationRule(t, tc.Rights)
```
tombuildsstuff: whilst this makes it more readable - this means it's not possible to run these tests individually; can we split them out as regular tests?

azurerm/resource_arm_servicebus_namespace_authorization_rule.go
```diff
@@ -0,0 +1,217 @@
...
+				MinItems: 1,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+					ValidateFunc: validation.StringInSlice([]string{
+						string(servicebus.Listen),
+						string(servicebus.Send),
+						string(servicebus.Manage),
+					}, true),
+				},
+				Set: set.HashStringIgnoreCase,
```
tombuildsstuff: can we switch this over to using Bools rather than a Set, since this makes the diff clearer (and then these can be inferred)?

azurerm/resource_arm_servicebus_namespace_authorization_rule_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+func testCheckAzureRMServiceBusNamespaceAuthorizationRuleExists(name string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		// Ensure we have enough information in state to look up in API
+		rs, ok := s.RootModule().Resources[name]
+		if !ok {
+			return fmt.Errorf("Not found: %s", name)
+		}
+
+		name := rs.Primary.Attributes["name"]
```
tombuildsstuff: _minor_ recently we've been renaming this field to avoid conflicting with the `name` parameter - it may be worth doing that here for consistency?

website/docs/r/servicebus_namespace_authorization_rule.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_servicebus_namespace_authorization_rule"
+sidebar_current: "docs-azurerm-resource-servicebus-namespace-authorization-rule"
+description: |-
+  Manages a ServiceBus Namespace authorization Rule within a ServiceBus Topic.
+---
+
+# azurerm_servicebus_namespace_authorization_rule
+
+Manages a ServiceBus Namespace authorization Rule within a ServiceBus Topic.
```
tombuildsstuff: would this make more sense as `Manages a ServiceBus Authorization Rule scoped at the Namespace level`?

website/docs/r/servicebus_namespace_authorization_rule.html.markdown
```diff
@@ -0,0 +1,73 @@
+---
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_servicebus_namespace_authorization_rule"
+sidebar_current: "docs-azurerm-resource-servicebus-namespace-authorization-rule"
+description: |-
+  Manages a ServiceBus Namespace authorization Rule within a ServiceBus Topic.
```
tombuildsstuff: (same here)

azurerm/resource_arm_servicebus_namespace_authorization_rule.go
katbyte: There are no properties object returned by list keys.

katbyte:
Requested updates made, mind giving it another quick review @tombuildsstuff ?

azurerm/resource_arm_servicebus_namespace_authorization_rule_test.go
tombuildsstuff: the variable, but it's not a big deal/blocker

azurerm/resource_arm_servicebus_namespace_authorization_rule.go
```diff
@@ -0,0 +1,192 @@
...
+			"listen": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true, //because we set this to true if managed is chosen
+			},
+
+			"send": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true, //because we set this to true if managed is chosen
```
tombuildsstuff: `send` and `listen` want default values of `false` since otherwise it's not possible to disable send/listen by simply removing the code / explicitly setting these values to false, which is misleading (as is an option in other providers, but we're having to work around due to the SDK)

website/docs/r/servicebus_namespace_authorization_rule.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+```hcl
+
+resource "azurerm_resource_group" "example" {
+  name     = "terraform-servicebus"
+  location = "West US"
+}
+
+resource "azurerm_servicebus_namespace" "example" {
+  name                = "${var.servicebus_name}"
+  location            = "${var.location}"
```
tombuildsstuff: neither of these variables exists in this context - can we hard-code example values here? e.g. `myservicebusnamespace` / `${azurerm_resource_group.example.location}` respectively

PR #1508, opened 2018-07-06
resource/container_group: parse command as sh

tharvik:
[container_group.command](https://www.terraform.io/docs/providers/azurerm/r/container_group.html#command) is a `schema.TypeString` but on the [azure's doc](https://docs.microsoft.com/en-us/rest/api/container-instances/containergroups/createorupdate#container), it is a `string[]`.
The current implementation take the command and split it on " ", and so, disallows having an argument containing a space.
In the long run, it would be best to move to a `schema.TypeList`, but this PR allows to keep the old schema and for complex command by taking it as a sh expression.

### DIFF
--- a/azurerm/resource_arm_container_group.go
+++ b/azurerm/resource_arm_container_group.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/flynn-archive/go-shlex"
+
 	"github.com/Azure/azure-sdk-for-go/services/containerinstance/mgmt/2018-04-01/containerinstance"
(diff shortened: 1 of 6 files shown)


tombuildsstuff:
hey @tharvik 

We've spent a little while mulling over how's best to proceed here - as this PR highlights the `command` field should really be an array rather than a string which we parse. Given the repository being used to do the parsing of the command had been archived, I've opened #1740 which deprecates the `command` field in favour of a new `commands` field (which takes an array to match the API, whilst allowing both to continue working for the moment) - which I'm going to close this PR in favour of, I hope you don't mind!

Thanks!

PR #1509, opened 2018-07-06
resource/container_group: add registry credential

tharvik:
Found a serie of commits adding the registry credentials to a container group, made by @abhijeetgaiha, adding [this feature](https://docs.microsoft.com/en-us/rest/api/container-instances/containergroups/createorupdate#imageregistrycredential). Reverting 0dd53df3274a5fd297d638ec4f513f204044a367 and c658442bdfd0fac9607bcff007e8e977dbf5f512 makes it work.

Maybe @abhijeetgaiha wants to comments why he removed it two days after adding it, it's not really explained in the commits messages

### DIFF
--- a/azurerm/resource_arm_container_group.go
+++ b/azurerm/resource_arm_container_group.go
@@ -80,6 +80,31 @@ func resourceArmContainerGroup() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"image_registry_credential": {
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"server": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+						"username": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+						"password": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+					},
+				},
+			},
+
 			"container": {
 				Type:     schema.TypeList,
 				Required: true,
@@ -207,6 +232,8 @@ func resourceArmContainerGroupCreate(d *schema.ResourceData, meta interface{}) e
 	restartPolicy := d.Get("restart_policy").(string)
 
 	containers, containerGroupPorts, containerGroupVolumes := expandContainerGroupContainers(d)
+	credentials := expandContainerGroupImageRegistryCredentials(d)
+
 	containerGroup := containerinstance.ContainerGroup{
 		Name:     &name,
 		Location: &location,
@@ -218,8 +245,9 @@ func resourceArmContainerGroupCreate(d *schema.ResourceData, meta interface{}) e
 				Type:  &IPAddressType,
 				Ports: containerGroupPorts,
 			},
-			OsType:  containerinstance.OperatingSystemTypes(OSType),
-			Volumes: containerGroupVolumes,
+			OsType:                   containerinstance.OperatingSystemTypes(OSType),
+			Volumes:                  containerGroupVolumes,
+			ImageRegistryCredentials: credentials,
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


azurerm/resource_arm_container_group.go
```diff
@@ -80,6 +80,31 @@ func resourceArmContainerGroup() *schema.Resource {
...
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"server": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+						"username": {
```
katbyte: Could we get a blank line between these properties like the rest of the resource?

azurerm/resource_arm_container_group.go
```diff
@@ -80,6 +80,31 @@ func resourceArmContainerGroup() *schema.Resource {
...
+			"image_registry_credential": {
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"server": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
```
katbyte: Could we get some simple validation here (and on username/password)?

```go
ValidateFunc: validation.NoZeroValues
```

Would suffice

azurerm/resource_arm_container_group.go
```diff
@@ -218,8 +245,9 @@ func resourceArmContainerGroupCreate(d *schema.ResourceData, meta interface{}) e
 				Type:  &IPAddressType,
 				Ports: containerGroupPorts,
 			},
-			OsType:  containerinstance.OperatingSystemTypes(OSType),
-			Volumes: containerGroupVolumes,
+			OsType:                   containerinstance.OperatingSystemTypes(OSType),
+			Volumes:                  containerGroupVolumes,
+			ImageRegistryCredentials: credentials,
```
katbyte: very minor but we could just directly assign this here

```go
ImageRegistryCredentials: expandContainerGroupImageRegistryCredentials(d)
```

azurerm/resource_arm_container_group.go
```diff
@@ -286,6 +314,12 @@ func resourceArmContainerGroupRead(d *schema.ResourceData, meta interface{}) err
 	}
 	d.Set("restart_policy", string(resp.RestartPolicy))
 
+	imageRegistryCredentials := flattenContainerGroupImageRegistryCredentials(d, resp.ImageRegistryCredentials)
+	err = d.Set("image_registry_credential", imageRegistryCredentials)
+	if err != nil {
```
katbyte: Also minor, this could be combined into one line:

```go
	if err = d.Set("image_registry_credential", imageRegistryCredentials); err != nil {
```

azurerm/resource_arm_container_group.go
```diff
@@ -323,6 +357,29 @@ func resourceArmContainerGroupDelete(d *schema.ResourceData, meta interface{}) e
 	return nil
 }
 
+func flattenContainerGroupImageRegistryCredentials(d *schema.ResourceData, credentials *[]containerinstance.ImageRegistryCredential) *[]interface{} {
+	ircConfigs := make([]interface{}, 0, len(*credentials))
+	ircConfigsOld := d.Get("image_registry_credential").([]interface{})
+
+	for _, irc := range *credentials {
+		ircConfig := make(map[string]interface{})
+		ircConfig["server"] = *irc.Server
```
katbyte: Could we wrap these assignments in nil checks?

```go
if v := ircConfig["server"] = *irc.Server; v != nil {
    ircConfig["server"] = *v
}
```

azurerm/resource_arm_container_group.go
```diff
@@ -323,6 +357,29 @@ func resourceArmContainerGroupDelete(d *schema.ResourceData, meta interface{}) e
...
+		ircConfig["username"] = *irc.Username
+		// search for password in the old configs
+		for _, oldIrcConfig := range ircConfigsOld {
```
katbyte: I'm not entirely sure, but i think the index will always be the same (otherwise each plan/apply we would attempt to update) so you might beable to simplify this to:

```go
for i, oldIrcConfig := range ircConfigsOld {
...

    if v, ok := d.GetOk(fmt.Sprintf("image_registry_credential.%d.password", i)); ok {
        block["password"] = v.(string)
    }
}
```

abhijeetgaiha:
I was unable to write a test for the feature and had to revert it. It's functionally complete but lacks the test. The test requires setting up a private docker registry which I wasn't able to figure out in time.

abhijeetgaiha:
@tharvik I would recommend adding the test.

katbyte:
Hi @tharvik,

As you pulled these out of repo limbo, i hope you don't mind but i going to close this PR in favour of #1529 as it as both documentation and tests.

thanks again!

PR #1515, opened 2018-07-07
Function app consumption plan bug

### LINKED ISSUE #1453: App Service Plan based Function App has erroneous application settings
We are experiencing an issue when deploying Azure Function app's via Terraform.

When deploying an Azure Function to an App Service Plan, there are two erroneous application settings being added by default.

They are:

- WEBSITE_CONTENTSHARE
- WEBSITE_CONTENTAZUREFILECONNECTIONSTRING

According to [Microsoft Docs](https://docs.microsoft.com/en-us/azure/azure-functions/functions-app-settings#websitecontentazurefileconnectionstring), these should only be set on Consumption Plan function apps.

Weirdly, those app settings have been there since the resource was included, but it's now causing major issues in Azure. If those settings are removed (say on a deployment that sets other app settings), it causes the backend function runtime to explode and render the entire function app useless.

### Community Note

### Terraform Version
Terraform v0.11.7

### Affected Resource(s)

* azurerm_function_app

```hcl
resource "azurerm_app_service_plan" "app_plan" {
  name                = "${var.name}"
  tags                = "${var.tags}"

  sku {
    tier = "Standard"
    size = "S2"
  }
}

resource "azurerm_function_app" "function_apps" {
  name                      = "${var.name}"
  location                  = "${var.location}"
  resource_group_name       = "${var.resource_group_name}"
  app_service_plan_id       = "${azurerm_app_service_plan.app_plan.id}"
  storage_connection_string = "${var.storage_account_connection_string}"
  version                   = "~1"
  tags                      = "${var.tags}"

  site_config {
    always_on = "${var.always_on}"
  }
}

```

Note: This is not our exact code as I've got it split into modules etc, but the correct properties are set

### Expected Behavior

Ideally, these two properties would only be set when creating a consumption app, although I'm not 100% sure that would be checked considering it's the app service plan that dictates if it's consumption...

### Actual Behavior
A app service plan function app is created with those two application settings set

### Steps to Reproduce

1. `terraform apply`

ranieuwe:
This PR fixes #1453 by disabling the `WEBSITE_CONTENT` app setting when a dynamic App service plan is used. All `TestAccAzureRMFunctionApp`* tests are successful. 

It is very hard to write a test for this scenario because the default app settings variables get deleted on resource read and are not transported to the terraform schema. This means it is not possible not to do a `TestCheckResourceAttr` or similar. Have manually inspected the resources and the settings are applied correctly with these fixes.

As an aside, the function app logic will need refactoring as it is fairly inefficient and not DRY. Create calls update, likely to update the app_settings, but update basically reruns the same ARM call that Create just made, slowing down the progress by approx 20 to 30 seconds. 

Because of this double logic you'll see the app service plan needing to be retrieved twice and introduced throughout. There is no cleaner way without refactoring extensively.

### DIFF
--- a/azurerm/resource_arm_function_app.go
+++ b/azurerm/resource_arm_function_app.go
@@ -505,7 +505,7 @@ func getBasicFunctionAppAppSettings(d *schema.ResourceData, appServiceTier strin
 func getFunctionAppServiceTier(ctx context.Context, appServicePlanId string, meta interface{}) (string, error) {
 	id, err := parseAzureResourceID(appServicePlanId)
 	if err != nil {
-		return "", fmt.Errorf("[ERROR] Unable to parse App Service Plan ID '%s': %+v", appServicePlanId, err)
+		return "", fmt.Errorf("[ERROR] Unable to parse App Service Plan ID %q: %+v", appServicePlanId, err)
 	}
 
 	log.Printf("[DEBUG] Retrieving App Server Plan %s", id.Path["serverfarms"])
@@ -513,7 +513,7 @@ func getFunctionAppServiceTier(ctx context.Context, appServicePlanId string, met
 	appServicePlansClient := meta.(*ArmClient).appServicePlansClient
 	appServicePlan, err := appServicePlansClient.Get(ctx, id.ResourceGroup, id.Path["serverfarms"])
 	if err != nil {
-		return "", fmt.Errorf("[ERROR] Could not retrieve App Service Plan ID '%s': %+v", appServicePlanId, err)
+		return "", fmt.Errorf("[ERROR] Could not retrieve App Service Plan ID %q: %+v", appServicePlanId, err)
 	}
 
 	return *appServicePlan.Sku.Tier, nil
--- a/azurerm/resource_arm_function_app.go
+++ b/azurerm/resource_arm_function_app.go
@@ -516,7 +516,12 @@ func getFunctionAppServiceTier(ctx context.Context, appServicePlanId string, met
 		return "", fmt.Errorf("[ERROR] Could not retrieve App Service Plan ID %q: %+v", appServicePlanId, err)
 	}
 
-	return *appServicePlan.Sku.Tier, nil
+	if sku := appServicePlan.Sku; sku != nil {
+		if tier := sku.Tier; tier != nil {
+			return *tier, nil
+		}
+	}
+	return "", fmt.Errorf("No `sku` block was returned for App Service Plan ID %q", appServicePlanId)
 }
 
 func expandFunctionAppAppSettings(d *schema.ResourceData, appServiceTier string) map[string]*string {
--- a/azurerm/resource_arm_function_app.go
+++ b/azurerm/resource_arm_function_app.go
@@ -495,8 +495,8 @@ func getBasicFunctionAppAppSettings(d *schema.ResourceData, appServiceTier strin
 		{Name: &contentFileConnStringPropName, Value: &storageConnection},
 	}
 
-	// If the application plan is dynamic (consumption), we do not want to include WEBSITE_CONTENT components
-	if strings.EqualFold(appServiceTier, "dynamic") {
+	// If the application plan is NOT dynamic (consumption plan), we do NOT want to include WEBSITE_CONTENT components
+	if !strings.EqualFold(appServiceTier, "dynamic") {
 		return basicSettings
 	}
 	return append(basicSettings, consumptionSettings...)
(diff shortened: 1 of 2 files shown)


azurerm/resource_arm_function_app.go
```diff
@@ -469,19 +484,45 @@ func getBasicFunctionAppAppSettings(d *schema.ResourceData) []web.NameValuePair
...
 		{Name: &functionVersionPropName, Value: &functionVersion},
+	}
+
+	consumptionSettings := []web.NameValuePair{
 		{Name: &contentSharePropName, Value: &contentShare},
 		{Name: &contentFileConnStringPropName, Value: &storageConnection},
 	}
+
+	// If the application plan is dynamic (consumption), we do not want to include WEBSITE_CONTENT components
+	if strings.EqualFold(appServiceTier, "dynamic") {
```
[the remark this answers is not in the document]
ranieuwe: You are absolute right. D'oh. Thanks for spotting this. Will fix

azurerm/resource_arm_function_app.go
```diff
@@ -469,19 +484,45 @@ func getBasicFunctionAppAppSettings(d *schema.ResourceData) []web.NameValuePair
...
+
+	log.Printf("[DEBUG] Retrieving App Server Plan %s", id.Path["serverfarms"])
+
+	appServicePlansClient := meta.(*ArmClient).appServicePlansClient
+	appServicePlan, err := appServicePlansClient.Get(ctx, id.ResourceGroup, id.Path["serverfarms"])
+	if err != nil {
+		return "", fmt.Errorf("[ERROR] Could not retrieve App Service Plan ID '%s': %+v", appServicePlanId, err)
+	}
+
+	return *appServicePlan.Sku.Tier, nil
```
tombuildsstuff: there's a crash here if `appServicePlan.Sku` or `appServicePlan.Sku.Tier` is nil (which happens for older resources); can we make this:

```
if sku := appServicePlan.Sku; sku != nil {
 if tier := sku.Tier; tier != nil {
    return *tier, nil
  }
}
return fmt.Errorf("No `sku` block was returned for App Service Plan %q (Resource Group %q)", resourceGroup, name)
```

azurerm/resource_arm_function_app.go
```diff
@@ -469,19 +484,45 @@ func getBasicFunctionAppAppSettings(d *schema.ResourceData) []web.NameValuePair
...
+	if err != nil {
+		return "", fmt.Errorf("[ERROR] Unable to parse App Service Plan ID '%s': %+v", appServicePlanId, err)
+	}
+
+	log.Printf("[DEBUG] Retrieving App Server Plan %s", id.Path["serverfarms"])
+
+	appServicePlansClient := meta.(*ArmClient).appServicePlansClient
+	appServicePlan, err := appServicePlansClient.Get(ctx, id.ResourceGroup, id.Path["serverfarms"])
+	if err != nil {
+		return "", fmt.Errorf("[ERROR] Could not retrieve App Service Plan ID '%s': %+v", appServicePlanId, err)
```
tombuildsstuff: _minor_ these two `'%s'` can become `%q`

tombuildsstuff (changes requested):
hey @ranieuwe 

Thanks for this PR :)

I've taken a look through and aside from the logic bug this mostly LGTM; if we can fix up the potential crash then we should be good to run the tests/get this merged

Thanks!

tombuildsstuff:
> As an aside, the function app logic will need refactoring as it is fairly inefficient and not DRY. Create calls update, likely to update the app_settings, but update basically reruns the same ARM call that Create just made, slowing down the progress by approx 20 to 30 seconds.

Some context here: for both App Service and Function Apps the App Settings weren't persisted on the initial creation (at least when we first worked on this, there's a new API version so this may have changed) - so this was the only way to ensure the settings were actually set (and that users wouldn't see a perpetual diff)

ranieuwe:
Changes done, nil check added and logic inversed. Tests all pass.

tombuildsstuff:
hey @ranieuwe 

I hope you don't mind - but I've taken a look into this and have pushed a commit which verifies the settings aren't set at the server end :)

Thanks!

tombuildsstuff (approved):
hey @ranieuwe 

Thanks for pushing those changes - I've pushed a commit to add a test for this but this now LGTM 👍 

Thanks!

tombuildsstuff:
Tests pass:

screenshot 2018-07-16 at 14 28 53

Issue #1177, opened 2018-04-27
Feature Request: API Management support


pietro:
I couldn't find anything on the documentation about API Gateways, policies, etc. for Azure API Management[0][1].

[0] https://docs.microsoft.com/en-us/azure/api-management/
[1] https://docs.microsoft.com/en-us/rest/api/apimanagement/

woosley:
I am looking for the same thing, is this in any plan?

torresdal:
I've started to look at the other relevant `api-management` resources to implement. I would like your input on where to draw the line between IaC (Infrastructure as Code) and deployment. The `api-management`-service is clearly IaC in my opinion and I didn't hesitate to implement  it. Other relevant resources of `api-management` is not necessarily as clear.

I've listed possible `api-management` Terraform resources extracted from the [Azure API Management REST API](https://docs.microsoft.com/en-us/rest/api/apimanagement/). After this exercise I was left with the following questions:

1. Which of these are clearly IaC and should be resources in Terraform? Or which ones are clearly not IaC and does not belong in Terraform?
2. In which priority? *"Without resource x, y, z, we can't configure API Management in Terraform"* 
2. Are some of these resources better handled elsewhere (other tooling?)
3. Should all be separate resources, and does some belong together/merged?

**Note:** The ones with ~~strikethrough~~ are resources I personally think is irrelevant for Terraform.

* `api-management-api`
* `api-management-api-diagnostic`
* ~~`api-management-api-issue`~~
* `api-management-api-logger`
* `api-management-api-operation`
* `api-management-api-operation-policy`
* `api-management-api-policy`
* `api-management-api-release`
* `api-management-api-schema`
* `api-management-api-version-set`
* `api-management-authorization-server`
* `api-management-backend`
* `api-management-certificate`
* `api-management-delegation`
* `api-management-diagnostic`
* `api-management-email-template`
* `api-management-group`
* `api-management-identity-provider`
* ~~`api-management-issue`~~
* `api-management-logger` 
* ~~`api-management-notification`~~
* `api-management-oidc-provider`
* `api-management-policy`
* `api-management-product`
* `api-management-product-group`
* `api-management-product-policy`
* `api-management-product-subscription`
* `api-management-property`
* `api-management-tag`
* `api-management-user`

torresdal:
Hi all! Quick status. Picking this up again after a rather long summer vacation :-) 

[...]

PR #1516, opened 2018-07-08
New Resource: `azurerm_api_management`

torresdal:
Added a new resource for API Management as requested in #1177. Contains the resource, tests and documentation.

ISSUE: It's marked as WIP since there is currently an issue with how the azure-sdk-for-go (or more likely Azure itself) handles deletion of the service when there are associated subnets. Basically the subnets fail to delete because it's still in use by the API Management service, even though it is deleted. I will create a separate issue for this and link back, but any pointers would be appreciated 👍 

Because of the issue mentioned above, the integration test `TestAccAzureRMApiManagement_complete` currently fail.


torresdal:
Regarding the issue (#1520) where subnet used by `azurerm_api_management` cannot be deleted, when `azurerm_api_management` is in fact destroyed and should no longer have a dependency to the subnet.

Has this happend with other Azure Terraform resources? It's strange that it takes 30-50 minutes to create the API Mangement resource in Azure, but only seconds to delete it...

torresdal:
As requested by @tombuildsstuff I've registered an issue for the subnet issue at https://github.com/Azure/azure-sdk-for-go/issues/2199

tombuildsstuff:
@torresdal awesome, thanks - I'm taking a look through this at the moment :)

tombuildsstuff:
@torresdal 

> Regarding the issue (#1520) where subnet used by azurerm_api_management cannot be deleted, when azurerm_api_management is in fact destroyed and should no longer have a dependency to the subnet.
> Has this happend with other Azure Terraform resources? It's strange that it takes 30-50 minutes to create the API Mangement resource in Azure, but only seconds to delete it...

We've had this in a few cases, [Virtual Network Gateway's](https://github.com/Azure/azure-rest-api-specs/issues/1233) and [Application Gateway's](https://github.com/Azure/azure-rest-api-specs/issues/2187) - in both of these cases the API affects the entire resource and as such we were blocked from shipping them until they were fixed*. However in this instance this is an optional field, as such I think we should be able to ship the rest of the resource and then add the subnet support back in when the API is fixed; what do you think?

[* Unfortunately we didn't realise the scale of the Application Gateway bug until after we'd shipped it - and now we've shipped it so it's hard to pull from the provider (it's causing us a lot of pain internally, since it's breaking our acceptance tests nightly / leaving test resources kicking around which fill our quota's).]

azurerm/data_source_api_management.go
```diff
@@ -0,0 +1,288 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"publisher_email": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"sku": apiManagementDataSourceSkuSchema(),
```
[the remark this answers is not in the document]
torresdal: Sure, but just to make sure I understand your preference, since I see you have not mentioned `apiManagementDataSourceCertificateInfoSchema`. What I've done is create funcs for schemas that are used multiple locations. Both `apiManagementDataSourceSkuSchema` and `apiManagementDataSourceCertificateInfoSchema` is used twice in the overall data source schema.
  
If this is not a good practice here (or in general 😃 ) I'll inline them.

website/docs/r/api_management.html.markdown
```diff
@@ -0,0 +1,262 @@
...
+sidebar_current: "docs-azurerm-resource-api-management"
+description: |-
+  Create a Api Management service.
+---
+
+# azurerm_api_management
+
+Create a Api Management component.
+
+-> **Note:** When `azurerm_api_management` service has dependencies on one or more `azurerm_subnet` resources, referenced using `vnet_subnet_id` or `additional_location.*.vnet_subnet_id` - the `azurerm_subnet`s will fail to be destroyed because of an issue in Azure where a reference to the `azurerm_api_management` service is left behind after its destruction. The `azurerm_api_management` service will eventually be released from `azurerm_subnet`, but it could take hours.
```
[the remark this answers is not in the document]
torresdal: Should we replace this with an explanation for why Virtual Networks is not supported at the moment?

torresdal:
Thanks for pointing out lots of useful improvements in your review @tombuildsstuff! I believe everything should now be up to date.

In addition: 

* I removed everything related to Virtual Networks as you suggested and retested to make sure `destroy` now works - and it does
* Removed `sku` from `additional_location` after verifying that is was not needed as you pointed out

I left you a few comments on a couple of your suggestions (inlining schema function and if we should document why we currently don't support Virtual Networks).

azurerm/data_source_api_management.go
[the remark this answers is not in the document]
tombuildsstuff: tbh in this case I think we could probably remove the second SKU in the data source, since we've also removed it from the resource (which is why I originally posted this)

(in general we tend to follow the Go guidelines of duplicating code the first time, and then refactoring it into a method - but it's mostly a gut feeling based on where something's likely to diverge, in this case I'd argue it probably wouldn't since it's the same object in both cases)

torresdal:
Anything else I can do on this before it get merged?

Also, what is your preference regarding changes in master that needs to be resolved in the pull-request branch - merge or rebase?

tombuildsstuff:
hey @torresdal 

Sorry for the delay in re-reviewing this - I'd started reviewing this and forgot to hit submit; let me check if the comments are still valid and re-submit this; apologies for the delay here!

> Also, what is your preference regarding changes in master that needs to be resolved in the pull-request branch - merge or rebase?

Rebasing is generally preferred from our side if possible, but we're fairly flexible here :)

Thanks!

tombuildsstuff:
@torresdal actually I've just noticed the vendoring here, so I'm going to rebase this & push a commit to resolve that; hope you don't mind!

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,672 @@
...
+			return nil
+		}
+
+		return err
+	}
+
+	return nil
+}
+
+func expandAzureRmApiManagementProperties(d *schema.ResourceData, sku *apimanagement.ServiceSkuProperties) *apimanagement.ServiceProperties {
```
tombuildsstuff: can we in-line this within the Create method?

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,672 @@
...
+
+		d.Set("notification_sender_email", props.NotificationSenderEmail)
+		d.Set("created", props.CreatedAtUtc.Format(time.RFC3339))
+		d.Set("gateway_url", props.GatewayURL)
+		d.Set("gateway_regional_url", props.GatewayRegionalURL)
+		d.Set("portal_url", props.PortalURL)
+		d.Set("management_api_url", props.ManagementAPIURL)
+		d.Set("scm_url", props.ScmURL)
+		d.Set("static_ips", props.StaticIps)
+		d.Set("custom_properties", &props.CustomProperties)
```
tombuildsstuff: rather than setting these directly (which will result in a set error) - could we add a flatten function for this (as below) and convert this list to a `map[string]interface{}` rather than setting `map[string]string` here?

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,672 @@
...
+
+		additionalLocations := flattenApiManagementAdditionalLocations(props.AdditionalLocations)
+		if err := d.Set("additional_location", additionalLocations); err != nil {
+			return fmt.Errorf("Error setting `additional_location`: %+v", err)
+		}
+
+		certificates := flattenApiManagementCertificates(d, props.Certificates)
+
+		if err != nil {
+			return err
```
tombuildsstuff: could we wrap this to explain what failed, e.g. 
```
return fmt.Errorf("Error flattening `certificate`: %+v", err)
```

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,672 @@
...
+		d.Set("portal_url", props.PortalURL)
+		d.Set("management_api_url", props.ManagementAPIURL)
+		d.Set("scm_url", props.ScmURL)
+		d.Set("static_ips", props.StaticIps)
+		d.Set("custom_properties", &props.CustomProperties)
+
+		hostnameConfigurations, err := flattenApiManagementHostnameConfigurations(d, props.HostnameConfigurations)
+
+		if err != nil {
+			return err
```
tombuildsstuff: could we wrap this to explain what failed, e.g. 
```
return fmt.Errorf("Error flattening `hostname_configuration `: %+v", err)
```

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,672 @@
...
+	if location := resp.Location; location != nil {
+		d.Set("location", azureRMNormalizeLocation(*location))
+	}
+
+	if props := resp.ServiceProperties; props != nil {
+		d.Set("publisher_email", props.PublisherEmail)
+		d.Set("publisher_name", props.PublisherName)
+
+		d.Set("notification_sender_email", props.NotificationSenderEmail)
+		d.Set("created", props.CreatedAtUtc.Format(time.RFC3339))
```
tombuildsstuff: can we remove the creation date, I can't think when this would be likely to be used within Terraform?

azurerm/data_source_api_management.go
```diff
@@ -0,0 +1,403 @@
...
+			},
+
+			"notification_sender_email": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"created": {
+				Type:     schema.TypeString,
+				Computed: true,
```
tombuildsstuff: can we remove the creation date, I can't think when this would be likely to be used within Terraform?

torresdal:
I believe I've addressed all the requested changes by @tombuildsstuff now (thanks for a thorough review btw!), so hopefully this should be ready to move on.

torresdal:
@jeffreyCline Any chance you could you do a final review? Thanks.

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateApiManagementName,
```
katbyte: Could we add this to the azure package in a file api_management.go? and then use it for the data source too

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+			},
+
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"location": locationSchema(),
+
+			"publisher_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validateApiManagementPublisherName,
```
katbyte: Could we add this to the azure package in a file api_management.go?

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+			},
+
+			"publisher_email": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validateApiManagementPublisherEmail,
```
katbyte: Could we add this to the azure package in a file api_management.go?

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+								string(apimanagement.SkuTypeStandard),
+								string(apimanagement.SkuTypePremium),
+							}, true),
+							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+						},
+						"capacity": {
+							Type:     schema.TypeInt,
+							Optional: true,
+							Default:  1,
+						},
```
katbyte: minor, could we validate this? I presume `IntAtLeast(0)` would be a good starting point

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+						},
+					},
+				},
+			},
+
+			"security": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Resource{
```
katbyte: Should this have a `MaxItems: 1` or can there be multiple?

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+								string(apimanagement.Proxy),
+								string(apimanagement.Scm),
+							}, true),
+							DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+						},
+
+						"host_name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
```
katbyte: A `validation.NoZeroValues` would be nice here

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+
+						"certificate": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
```
katbyte: A `validation.NoZeroValues` would be nice here

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,672 @@
...
+
+	resGroup := d.Get("resource_group_name").(string)
+	name := d.Get("name").(string)
+	location := azureRMNormalizeLocation(d.Get("location").(string))
+	tags := d.Get("tags").(map[string]interface{})
+
+	var sku *apimanagement.ServiceSkuProperties
+	if skuConfig := d.Get("sku").([]interface{}); skuConfig != nil {
+		sku = expandAzureRmApiManagementSku(skuConfig)
+	}
```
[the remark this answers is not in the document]
katbyte: seconded

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+	publisher_email := d.Get("publisher_email").(string)
+	notification_sender_email := d.Get("notification_sender_email").(string)
+
+	custom_properties := expandApiManagementCustomProperties(d)
+
+	additional_locations := expandAzureRmApiManagementAdditionalLocations(d, sku)
+	certificates := expandAzureRmApiManagementCertificates(d)
+	hostname_configurations := expandAzureRmApiManagementHostnameConfigurations(d)
+
+	properties := apimanagement.ServiceProperties{
```
katbyte: We could join the next two object initializations:
```go
apiManagement := apimanagement.ServiceResource{
		Location:          &location,
		ServiceProperties: &apimanagement.ServiceProperties{
             PublisherName:          utils.String(publisher_name),
     		PublisherEmail:         utils.String(publisher_email),
	     	CustomProperties:       custom_properties,
     		AdditionalLocations:    additional_locations,
	     	Certificates:           certificates,
     		HostnameConfigurations: hostname_configurations,
	     },
		Tags:              expandTags(tags),
		Sku:               sku,
	}
```

azurerm/data_source_api_management_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccDataSourceAzureRMApiManagement_basic(t *testing.T) {
```
katbyte: This test is currently failing:

```
azurerm_api_management.test: Error creating API Management Service "acctestAM-1444076724406126485" (Resource Group "amtestRG-1444076724406126485"): apimanagement.ServiceClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidParameters" Message="Invalid parameter: Setting up Additional Locations is not supported for Sku Type 'Developer'."
FAIL
```

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+		Sku:               sku,
+	}
+
+	future, err := client.CreateOrUpdate(ctx, resGroup, name, apiManagement)
+	if err != nil {
+		return fmt.Errorf("Error creating API Management Service %q (Resource Group %q): %+v", name, resGroup, err)
+	}
+
+	err = future.WaitForCompletion(ctx, client.Client)
+	if err != nil {
```
katbyte: Very minor but this could be:
```go
if err := future.WaitForCompletion(ctx, client.Client); err != nil {
```

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+		}
+
+		if err := d.Set("certificate", flattenApiManagementCertificates(d, props.Certificates)); err != nil {
+			return fmt.Errorf("Error setting `certificate`: %+v", err)
+		}
+	}
+
+	if sku := resp.Sku; sku != nil {
+		if err := d.Set("sku", flattenApiManagementServiceSku(sku)); err != nil {
+			return fmt.Errorf("Error flattening `sku`: %+v", err)
```
katbyte: I think this should be `Error setting 'sku'`?

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+
+	if v := input["Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Protocols.Tls10"]; v != nil {
+		val, err := strconv.ParseBool(*v)
+		if err != nil {
+			return nil, fmt.Errorf("Error parsing `disable_frontend_tls10` %q: %+v", *v, err)
+		}
+		output["disable_frontend_tls10"] = val
+	}
+
+	if v := input["Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Protocols.Tls11"]; v != nil {
```
katbyte: Could this changed to use a local helper?

```go
if err := setCustomPropertyFrom(input, "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Protocols.Tls11", output, "disable_frontend_tls11"); err != nil {
    return nil, err
}

setCustomPropertyFrom(input, path, output, key) {
  if v := input[path]; v != nil {
    val, err := strconv.ParseBool(*v)
    if err != nil {
      return nil, fmt.Errorf("Error parsing `%s` %q: %+v", key, *v, err)
    }
    output[key] = val
  }

  return nil 
}
```

not blocking the pr, just a idea

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+	}
+
+	return &additionalLocations
+}
+
+func expandAzureRmApiManagementSku(configs []interface{}) *apimanagement.ServiceSkuProperties {
+	config := configs[0].(map[string]interface{})
+
+	nameConfig := config["name"].(string)
+	name := apimanagement.SkuType(nameConfig)
```
katbyte: these two lines could be combined:
```go
name := apimanagement.SkuType(config["name"].(string))
```

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+	certificates := make([]interface{}, 0)
+
+	if props != nil {
+		for i, prop := range *props {
+			certificate := make(map[string]interface{}, 0)
+
+			certificate["store_name"] = string(prop.StoreName)
+
+			// certificate password isn't returned, so let's look it up
+			passwKey := fmt.Sprintf("certificate.%d.certificate_password", i)
```
katbyte: You will have to look these up by name, if someone edits this on the portal the index's may no longer match.

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+			if v, ok := d.GetOk(passwKey); ok {
+				password := v.(string)
+				certificate["certificate_password"] = password
+			}
+
+			// encoded certificate isn't returned, so let's look it up
+			certKey := fmt.Sprintf("certificate.%d.encoded_certificate", i)
```
katbyte: Same as above comment

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+			if v, ok := d.GetOk(certKey); ok {
+				cert := v.(string)
+				certificate["encoded_certificate"] = cert
```
katbyte: These could be combined: 
```go
certificate["encoded_certificate"] = v.(string)
```

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+}
+
+func flattenApiManagementHostnameConfigurations(d *schema.ResourceData, configs *[]apimanagement.HostnameConfiguration) []interface{} {
+	host_configs := make([]interface{}, 0)
+
+	if configs != nil {
+		for i, config := range *configs {
+			host_config := make(map[string]interface{}, 0)
+
+			if config.Type != "" {
```
katbyte: as this is an enum i'm not sure if you need to check for "" ?

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+			if config.DefaultSslBinding != nil {
+				host_config["default_ssl_binding"] = *config.DefaultSslBinding
+			}
+
+			if config.NegotiateClientCertificate != nil {
+				host_config["negotiate_client_certificate"] = bool(*config.NegotiateClientCertificate)
+			}
+
+			// certificate password isn't returned, so let's look it up
+			passKey := fmt.Sprintf("hostname_configuration.%d.certificate_password", i)
```
katbyte: You will have to look these up by name, if someone edits this on the portal the index's may no longer match.

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+			if v, ok := d.GetOk(passKey); ok {
+				password := v.(string)
+				host_config["certificate_password"] = password
+			}
+
+			// encoded certificate isn't returned, so let's look it up
+			certKey := fmt.Sprintf("hostname_configuration.%d.certificate", i)
```
katbyte: You will have to look these up by name, if someone edits this on the portal the index's may no longer match.

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+		}
+
+		if profile.Capacity != nil {
+			sku["capacity"] = *profile.Capacity
+		}
+
+		skus = append(skus, sku)
+	}
+
+	return skus
```
katbyte: I think this could be simplified: `return []interface{}{sku}`

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+}
+
+func validateApiManagementName(v interface{}, k string) (ws []string, es []error) {
```
katbyte: This should be moved to the azure helper in an api_management.go file

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+	value := v.(string)
+
+	if matched := regexp.MustCompile(`^[0-9a-zA-Z-]{1,50}$`).Match([]byte(value)); !matched {
+		es = append(es, fmt.Errorf("%q may only contain alphanumeric characters and dashes up to 50 characters in length", k))
+	}
+
+	return
+}
+
+func validateApiManagementPublisherName(v interface{}, k string) (ws []string, es []error) {
```
katbyte: This should be moved to the azure helper in an api_management.go file

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,773 @@
...
+	value := v.(string)
+
+	if matched := regexp.MustCompile(`^[\S*]{1,100}$`).Match([]byte(value)); !matched {
+		es = append(es, fmt.Errorf("%q may only be up to 100 characters in length", k))
+	}
+
+	return
+}
+
+func validateApiManagementPublisherEmail(v interface{}, k string) (ws []string, es []error) {
```
katbyte: This should be moved to the azure helper in an api_management.go file

azurerm/resource_arm_api_management_test.go
```diff
@@ -0,0 +1,241 @@
...
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMApiManagementDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMApiManagementExists("azurerm_api_management.test"),
+				),
+			},
+		},
```
katbyte: Could we add an import check step here?

azurerm/resource_arm_api_management_test.go
```diff
@@ -0,0 +1,241 @@
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMApiManagementDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMApiManagementExists("azurerm_api_management.test"),
+				),
+			},
```
katbyte: Could we add an import check step here?

website/docs/r/api_management.html.markdown
```diff
@@ -0,0 +1,223 @@
...
+  location = "North Europe"
+}
+
+resource "azurerm_api_management" "test" {
+  name                          = "api-mngmnt"
+  publisher_name                = "My Company"
+  publisher_email               = "company1@terraform.io"
+  notification_sender_email     = "api@terraform.io"
+
+  additional_location {
```
katbyte: could we take this example and run it through terraform fmt to get the correct indentation?

website/docs/r/api_management.html.markdown
```diff
@@ -0,0 +1,223 @@
...
+
+* `hostname_configuration` - (Optional) Custom hostname configuration of the API Management service. The `hostname_configuration` block is documented below.
+
+* `tags` - (Optional) A mapping of tags assigned to the resource.
+
+---
+
+`sku` block supports the following:
+
+* `name` - (Required) Specifies the plan's pricing tier.
```
katbyte: Could we list the possible values here?

azurerm/config.go
```diff
@@ -442,6 +447,16 @@ func getArmClient(c *authentication.Config) (*ArmClient, error) {
 	return &client, nil
 }
 
+func (c *ArmClient) registerApiManagementServiceClients(endpoint, subscriptionId string, auth autorest.Authorizer, sender autorest.Sender) {
+	ams := apimanagement.NewServiceClientWithBaseURI(endpoint, subscriptionId)
+	setUserAgent(&ams.Client)
```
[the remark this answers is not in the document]
torresdal: Not really :-) I followed the same convention as it were for other resources at the time, but I see that has now been refactored to use `configureClient`. So did I.

torresdal:
@katbyte I've addressed most of the things you pointed out. What I'm still a bit unsure about, is how to handle sensitive data not returned by the API. As you pointed out in https://github.com/terraform-providers/terraform-provider-azurerm/blob/97f1cfc6c96f60f314d804425ac76a38811d6aec/azurerm/resource_arm_api_management.go#L649 and https://github.com/terraform-providers/terraform-provider-azurerm/blob/97f1cfc6c96f60f314d804425ac76a38811d6aec/azurerm/resource_arm_api_management.go#L721 there is a potential issue with ordering when I extract data from the state using index.

The problem with these two cases is there is nothing that really uniquely identify the certificates. In the first case I can use a combination of `thumbprint` and `store_name`, but that is also problematic since `thumbprint` is Computed. Right now I don't see any other option than to leave it as is. I do hope there is a better solution though...

As for testing these cases, I've used `ImportStateVerifyIgnore` to work around the issue, which you can see here: https://github.com/terraform-providers/terraform-provider-azurerm/blob/97f1cfc6c96f60f314d804425ac76a38811d6aec/azurerm/resource_arm_api_management_test.go#L124 Is this the correct way to handle this?

I've also changed the reference to the API from `github.com/Azure/azure-sdk-for-go/services/apimanagement/mgmt/2017-03-01/apimanagement` to `github.com/Azure/azure-sdk-for-go/services/preview/apimanagement/mgmt/2018-06-01-preview/apimanagemen`, which I posted a qustion about on Slack: https://terraform-azure.slack.com/archives/CB9RVKPDL/p1537003012000200

katbyte:
@torresdal,

In the first instance, the list of certs in the resource, as azure is not returning any real way to link what is configured in terraform to azure I am going to suggest we don't even attempt to read it back into state. Drop the thumbprint and have it as a one way 'set and then update when the tf changes' for now.

As for the 2nd situation you can use the `hostname_configuration` to look up the details by looking it up in the old state: `stateOld := d.Get("hostname_configuration").([]interface{})`

And you are correct wrt `ImportStateVerifyIgnore`! that is the exact way to handle this.

katbyte:
@torresdal,

If you do no set the property on read, whatever the TF is set to stays in state. So removing 
```
if err := d.Set("certificate", flattenApiManagementCertificates(d, props.Certificates)); err != nil {
			return fmt.Errorf("Error setting `certificate`: %+v", err)
		}
```

means the only time tf will see a change, is if you change the config (i believe, its been some time since i did this)

In the other instance i mean you can iterate across the old one looking for the same hostname:

```
    stateOld := d.Get("hostname_configuration").([]interface{})
    if configs != nil {
		for i, config := range *configs {
            ...
			if config.HostName != nil {
				host_config["host_name"] = *config.HostName

                var old
                for _,o := range stateOld {
                    if o["host_name"] == host_config["host_name] {
                         old = o
                    }
                }

                host_config["certificate_password"] = old["certificate_password"]
			}


```

Does that clarify it?

torresdal:
@katbyte I believe I've now fixed everything. 

For case 1, after doing quite a bit of testing, it is my understanding that in order to avoid diffs I need to return the old state during read, so I did this:

```go
// Azure is not returning identifiable certs (no id), so there is no way to link
// what is configured in terraform to Azure. Because of this, we just return the old state.
if err := d.Set("certificate", d.Get("certificate").([]interface{})); err != nil {
	return fmt.Errorf("Error setting `certificate`: %+v", err)
}
```

If I don't, all certificate values will be empty strings, and TF will detect that as a diff.

For case 2 I made a few changes to the schema structure, which I hope will make it better, but in essence the solution is according to your suggestion. Here's an example of the new schema:

```tf
hostname_configurations {
  proxy {
    host_name                    = "api.terraform.io"
    certificate                  = "${base64encode(file("testdata/api_management_api_test.pfx"))}"
    certificate_password         = "terraform"
    default_ssl_binding          = true
    negotiate_client_certificate = false
  }

  proxy {
    host_name                    = "api2.terraform.io"
    certificate                  = "${base64encode(file("testdata/api_management_api2_test.pfx"))}"
    certificate_password         = "terraform"
    negotiate_client_certificate = true
  }

  portal {
    host_name            = "portal.terraform.io"
    certificate          = "${base64encode(file("testdata/api_management_portal_test.pfx"))}"
    certificate_password = "terraform"
  }
}
```

torresdal:
@katbyte (cc @tombuildsstuff ) Have you had time to look at this? Eager to get this out in the wild and start adding the other resources...

azurerm/helpers/azure/api_management.go
```diff
@@ -0,0 +1,116 @@
...
+
+		if val {
+			output[key] = val
+		}
+	}
+
+	return nil
+}
+
+func ResourceArmApiManagementCustomizeDiff(diff *schema.ResourceDiff, v interface{}) error {
```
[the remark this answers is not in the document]
tombuildsstuff: we actually need to remove this function - so I'm going to resolve this.

azurerm/helpers/azure/api_management.go
```diff
@@ -0,0 +1,116 @@
...
+	if hasHostnameScm {
+		if err := validateHostnameConfig(hostnameScm.([]interface{})); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func validateHostnameConfig(hostnames []interface{}) error {
```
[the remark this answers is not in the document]
tombuildsstuff: (as above) we're removing this function - so we can remove this

tombuildsstuff:
Data Source tests pass:

screenshot 2018-09-28 at 19 15 16

azurerm/data_source_api_management.go
```diff
@@ -0,0 +1,336 @@
...
+
+	for _, prop := range *input {
+		output := make(map[string]interface{}, 0)
+
+		if prop.Location != nil {
+			output["location"] = azureRMNormalizeLocation(*prop.Location)
+		}
+
+		if prop.PublicIPAddresses != nil {
+			output["static_ips"] = *prop.PublicIPAddresses
```
tombuildsstuff: on reflection, I guess this should probably be `public_ip_addresses` (since `private_ip_addresses` can also be returned for API Management Services on a network)?

azurerm/resource_arm_api_management.go
```diff
@@ -0,0 +1,847 @@
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"location": locationSchema(),
+
+						"gateway_regional_url": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"static_ips": {
```
tombuildsstuff: (as below) perhaps this should become `public_ip_addresses`?

azurerm/data_source_api_management.go
```diff
@@ -0,0 +1,336 @@
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"location": locationForDataSourceSchema(),
+
+						"gateway_regional_url": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"static_ips": {
```
tombuildsstuff: (as below) perhaps this should become `public_ip_addresses` since there can also be Private IP Addresses?

azurerm/data_source_api_management.go
torresdal: Renamed to `public_ip_addresses`. Just need to remember to add `private_ip_addresses` when/if we add support for internal vnets. I will retest this (internal vnets) now that we use the preview API, but I expect the same problem to be present.

website/docs/r/api_management.html.markdown
```diff
@@ -0,0 +1,223 @@
...
+
+* `certificate` - (Optional) One or more (up to 10) `certificate` blocks as defined below.
+
+* `identity` - (Optional) An `identity` block is documented below.
+
+* `hostname_configuration` - (Optional) A `hostname_configuration` block as defined below.
+
+* `notification_sender_email` - (Optional) Email address from which the notification will be sent.
+
+* `security` - A `security` block as defined below.
```
tombuildsstuff: this wants `(Optional)`

website/docs/r/api_management.html.markdown
```diff
@@ -0,0 +1,223 @@
...
+
+-> **info:** This maps to the `Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Protocols.Tls11` field
+
+* `disable_triple_des_chipers` - (Optional) Should the `TLS_RSA_WITH_3DES_EDE_CBC_SHA` cipher be disabled for alL TLS versions (1.0, 1.1 and 1.2)? Defaults to `false`.
+
+-> **info:** This maps to the `Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Ciphers.TripleDes168` field
+
+---
+
+`sku` block supports the following:
```
tombuildsstuff: This is missing an `A ` at the start

website/docs/r/api_management.html.markdown
tombuildsstuff: (fixed)

website/docs/r/api_management.html.markdown
tombuildsstuff: (fixed)

azurerm/provider.go
```diff
@@ -78,32 +78,34 @@ func Provider() terraform.ResourceProvider {
...
+			"azurerm_key_vault_secret":           dataSourceArmKeyVaultSecret(),
+			"azurerm_kubernetes_cluster":         dataSourceArmKubernetesCluster(),
+			"azurerm_log_analytics_workspace":    dataSourceLogAnalyticsWorkspace(),
+			"azurerm_logic_app_workflow":         dataSourceArmLogicAppWorkflow(),
+			"azurerm_managed_disk":               dataSourceArmManagedDisk(),
+			"azurerm_management_group":           dataSourceArmManagementGroup(),
+			"azurerm_network_interface":          dataSourceArmNetworkInterface(),
+			"azurerm_network_security_group":     dataSourceArmNetworkSecurityGroup(),
+			"azurerm_notification_hub":           dataSourceNotificationHub(),
+
```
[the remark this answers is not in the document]
tombuildsstuff: (fixed)

tombuildsstuff:
Data Source tests pass:

screenshot 2018-10-01 at 20 34 21

tombuildsstuff:
Resource Tests pass:

screenshot 2018-10-01 at 22 39 08

kumlien:
Nice work! Is this one already available for us 'normal users'? I dont' see it on  [the terraform site](https://www.terraform.io/docs/providers/azurerm/index.html) and my idea tf-plugin doesn't recognize it... (running  terraform-provider-azurerm_v1.16.0_x4).

torresdal:
@kumlien Looks like its scheduled to be in version 1.17.0.

PR #1517, opened 2018-07-09
Key vault add soft delete 

### LINKED ISSUE #1329: Missing soft delete option in azurerm_key_vault
### Community Note

### Description

Please add option for soft delete

* azurerm_key_vault

```
resource "azurerm_key_vault" "test" {
  name                = "testvault"
  location            = "West US"

  sku {
    name = "standard"
  }

  tenant_id = "d6e396d0-5584-41dc-9fc0-268df99bc610"

  access_policy {
    tenant_id = "d6e396d0-5584-41dc-9fc0-268df99bc610"
    object_id = "d746815a-0433-4a21-b95d-fc437d2d475b"

    key_permissions = [
      "get",
    ]

    secret_permissions = [
      "get",
    ]
  }
  
  soft_delete = true
  enabled_for_disk_encryption = true

  tags {
    environment = "Production"
  }
}
```

janboll:
Make it possible to enable soft delete via Terraform. 
This would fix #1329.


azurerm/resource_arm_key_vault.go
```diff
@@ -181,6 +182,18 @@ func resourceArmKeyVault() *schema.Resource {
 				Optional: true,
 			},
 
+			"enabled_for_soft_delete": {
+				Type:         schema.TypeBool,
+				Optional:     true,
+				ValidateFunc: validate.BoolIsTrue(),
```
tombuildsstuff: given the code below ignores this if it's not set to `true` - I think we can safely remove this validation function? This allows users to define this variable optionally using a variable e.g.

```
resource "azurerm_key_vault" "test" {
  count = 2
  enabled_for_soft_delete = "${count.index == 0 ? true : false}"
}

File: azurerm/resource_arm_key_vault.go [hunk 2]
```diff
@@ -181,6 +182,18 @@ func resourceArmKeyVault() *schema.Resource {
...
+			},
+
+			"enabled_for_purge_protection": {
+				Type:         schema.TypeBool,
+				Optional:     true,
+				ValidateFunc: validate.BoolIsTrue(),
```
tombuildsstuff: given the code below ignores this if it's not set to `true` - I think we can safely remove this validation function?

azurerm/resource_arm_key_vault_test.go
```diff
@@ -216,7 +220,7 @@ resource "azurerm_key_vault" "test" {
 
   sku {
     name = "premium"
-  }
+	}
```
tombuildsstuff: _minor_ can we revert this?

azurerm/resource_arm_key_vault_test.go
```diff
@@ -295,6 +305,9 @@ resource "azurerm_key_vault" "test" {
   sku {
     name = "premium"
   }
+	
+	enabled_for_soft_delete = "True"
+	enabled_for_purge_protection = "True"
```
tombuildsstuff: _minor_ can we remove the quotes around these values?

azurerm/resource_arm_key_vault_test.go
```diff
@@ -251,7 +259,9 @@ resource "azurerm_key_vault" "test" {
 
   sku {
     name = "premium"
-  }
+	}
+
+	enabled_for_soft_delete = "True"
```
tombuildsstuff: _minor_ can we remove the quotes around this value?

azurerm/resource_arm_key_vault.go
tombuildsstuff: given the comment in the docs `Once enabled you can not disable this setting anymore!` can we add `ForceNew: true` here?

azurerm/resource_arm_key_vault.go
tombuildsstuff: given the comment in the docs `Once enabled you can not disable this setting anymore!` can we add `ForceNew: true` here?

website/docs/r/key_vault.html.markdown
```diff
@@ -83,6 +83,13 @@ The following arguments are supported:
     Azure Resource Manager is permitted to retrieve secrets from the key vault.
     Defaults to false.
 
+* `enabled_for_soft_delete` - (Optional) Boolean flag to specify whether
+    the key vault is enabled for soft delete. Once enabled you can not disable this setting anymore!
+
+* `enabled_for_purge_protection` - (Optional) Boolean flag to specify whether
+    the key vault is enabled for purge protection. Once enabled you can not disable this setting anymore!
```
tombuildsstuff: can we replace `Once enabled you can not disable this setting anymore!` with `Changing this forces a new resource to be created.`

azurerm/resource_arm_key_vault_test.go
```diff
@@ -136,6 +138,8 @@ func TestAccAzureRMKeyVault_update(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "enabled_for_deployment", "true"),
 					resource.TestCheckResourceAttr(resourceName, "enabled_for_disk_encryption", "true"),
 					resource.TestCheckResourceAttr(resourceName, "enabled_for_template_deployment", "true"),
+					resource.TestCheckResourceAttr(resourceName, "enabled_for_soft_delete", "true"),
+					resource.TestCheckResourceAttr(resourceName, "enabled_for_purge_protection", "false"),
```
tombuildsstuff: given changing this value will force a new resource, it'll break this test - can we check for this value in a new test?

azurerm/resource_arm_key_vault_test.go
```diff
@@ -102,6 +102,8 @@ func TestAccAzureRMKeyVault_complete(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testCheckAzureRMKeyVaultExists(resourceName),
 					resource.TestCheckResourceAttrSet(resourceName, "access_policy.0.application_id"),
+					resource.TestCheckResourceAttr(resourceName, "enabled_for_soft_delete", "true"),
+					resource.TestCheckResourceAttr(resourceName, "enabled_for_purge_protection", "true"),
```
tombuildsstuff: rather than modifying this test - can we check these values in a new test?

azurerm/resource_arm_key_vault.go
janboll: I tried to use ForceNew, but changing from soft delete true to soft delete false yields following error: 
```azurerm_key_vault.test: Error updating Key Vault "vault5353625459524355837" (Resource Group "acctestRG-5353625459524355837"): keyvault.VaultsClient#CreateOrUpdate: Failure sending request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="ConflictError" Message="Exist soft deleted vault with the same name. "```
On the other hand, switching soft delete on works in place without issue. 

I thought about adding the purge to the delete function, but the PurgeDeleted function also has some limitations: https://docs.microsoft.com/en-us/azure/key-vault/key-vault-ovw-soft-delete. So this would potentially break as well.

I suspect that removing the validation function and adding `ForceNew: true` could not work due to the way soft deleted key vaults are handled on Azure side.

azurerm/resource_arm_key_vault_test.go
janboll: I created a seperate test for this: `TestAccAzureRMKeyVault_enable_soft_delete`

azurerm/resource_arm_key_vault.go
tombuildsstuff: > I thought about adding the purge to the delete function, but the PurgeDeleted function also has some limitations: https://docs.microsoft.com/en-us/azure/key-vault/key-vault-ovw-soft-delete. So this would potentially break as well.

in which case I think we'd want to purge the Key Vault if it exists at deletion time, via an opt-in flag? If users don't opt-into that and delete a soft-deleted KeyVault then they'll get this error, at which point it's up to them to recover the vault and import it into Terraform's Statefile to be able to proceed. What do you think?

We have a similar opt-in property in the VM resource where we optionally delete the OS and Data Disks when the VM is deleted.

janboll: This sounds like a good idea 👍 I added an option to toggle the purge behaviour and also updated the tests support this.

janboll:
Opened a new PR, I am sorry for the mess :-(

PR #1529, opened 2018-07-10
 resource/container_group: add registry credential

lawrencegripper:
When I noticed this feature was missing I checked the issues and didn't see anything so set about working on it. When I came to raise the PR I noticed that #1509 was already in progress.

As I'd already included docs, test and an example (not in the current PR work) I thought it's still worth raising this PR as these can be merged into the existing PR and don't go to waste.  

cc/ @tharvik and @katbyte 

Ps. will check open PR's next time :(


katbyte:
Hi @lawrencegripper,

Regardless thank you for your contribution! As #1509 was just pulling some old commits out of the repo and lacks documentation/tests I am going to close it in favour of yours.

I will review it shortly.

azurerm/resource_arm_container_group.go
```diff
@@ -277,6 +309,10 @@ func resourceArmContainerGroupRead(d *schema.ResourceData, meta interface{}) err
 	}
 	flattenAndSetTags(d, resp.Tags)
 
+	if imageRegCreds := flattenContainerImageRegistryCredentials(d, resp.ImageRegistryCredentials); imageRegCreds != nil {
+		d.Set("image_registry_credential", imageRegCreds)
```
katbyte: Could we check for an error here? like:

```go
if err := d.Set("capabilities", flattenAzureRmCosmosDBAccountCapabilities(resp.Capabilities)); err != nil {
		return fmt.Errorf("Error setting `capabilities`: %+v", err)
	}
```

`d.Set()` sill handle a nil value correctly

azurerm/resource_arm_container_group.go
```diff
@@ -531,6 +567,63 @@ func expandContainerEnvironmentVariables(input interface{}) *[]containerinstance
...
+
+	for _, c := range credsRaw {
+		credConfig := c.(map[string]interface{})
+
+		server := credConfig["server"].(string)
+		username := credConfig["username"].(string)
+		password := credConfig["password"].(string)
+
+		output = append(output, containerinstance.ImageRegistryCredential{
+			Server:   &server,
```
katbyte: Very minor but these could be assigned inline via `utils.String()`

```go
Server: utils.String(credConfig["username"].(string)),
```

azurerm/resource_arm_container_group.go
```diff
@@ -531,6 +567,63 @@ func expandContainerEnvironmentVariables(input interface{}) *[]containerinstance
...
+	for _, cred := range creds {
+		credConfig := make(map[string]interface{})
+		if cred.Server != nil {
+			credConfig["server"] = *cred.Server
+		}
+		if cred.Username != nil {
+			credConfig["username"] = *cred.Username
+		}
+
+		for i, configsOld := range configsOld {
```
katbyte: I'm not sure, but i presume the ordering is consistent so you could get the old values this way:

```go
for i, oldIrcConfig := range ircConfigsOld {
...

    if v, ok := d.GetOk(fmt.Sprintf("image_registry_credential.%d.password", i)); ok {
        block["password"] = v.(string)
    }
}
```

azurerm/resource_arm_container_group_test.go
```diff
@@ -176,6 +204,54 @@ resource "azurerm_container_group" "test" {
...
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  ip_address_type     = "public"
+  os_type             = "linux"
+
+  container {
+    name   = "hw"
+    image  = "microsoft/aci-helloworld:latest"
+    cpu    = "0.5"
+    memory = "0.5"
+	port   = "80"
```
katbyte: could we fix the alignment here?

azurerm/resource_arm_container_group.go
lawrencegripper: Sure, make sense I'll fix this up

azurerm/resource_arm_container_group.go
lawrencegripper: Ah much nicer, I thought about bringing in the `to` package from Autorest as has lots of these funcs too didn't realize you already had one.

azurerm/resource_arm_container_group.go
lawrencegripper: I'm a little uncertain on this one so was being careful. Does the ordering of elements relate to the ordering of the elements in HCL? So for example if the user update the `password` element in their template but also moved it a line up ahead of `server` would that affect your approach?

azurerm/resource_arm_container_group_test.go
lawrencegripper: Woop will do

azurerm/resource_arm_container_group.go
katbyte: Yes, the order of the list directly relates to the order of the elements in HCL. So if the user does as you describe, the Apply would be sending a new list anyways, and the subsequent read would be post Update. unless azure returns things in a different order I think it should be fine. You could always write a test to confirm 🙂

lawrencegripper: Cool, makes sense to me. Will also check with the test just to be on the safe side.

lawrencegripper:
Tests look good not and hopefully all feedback is sorted. 

```
2018/07/13 16:56:10 debugger.go:518: continuing
=== RUN   TestAccAzureRMContainerGroup_imageRegistryCredentials
--- PASS: TestAccAzureRMContainerGroup_imageRegistryCredentials (62.86s)
=== RUN   TestAccAzureRMContainerGroup_imageRegistryCredentialsUpdate
--- PASS: TestAccAzureRMContainerGroup_imageRegistryCredentialsUpdate (75.83s)
PASS
```

azurerm/resource_arm_container_group_test.go
```diff
@@ -11,11 +11,78 @@ import (
...
+		},
+	})
+}
+
 func TestAccAzureRMContainerGroup_linuxBasic(t *testing.T) {
 	resourceName := "azurerm_container_group.test"
 	ri := acctest.RandInt()
 
-	config := testAccAzureRMContainerGroup_linuxBasic(ri, testLocation())
+	config := testAccAzureRMContainerGroup_imageRegistryCredentials(ri, testLocation())
```
tombuildsstuff: the tests are failing since this has been updated but the counts are incorrect; going to push a test to revert this to `TestAccAzureRMContainerGroup_linuxBasic`

tombuildsstuff:
Tests pass with the test fix:

screenshot 2018-07-16 at 13 04 01

tombuildsstuff:
@lawrencegripper no worries - easily done, thanks again for this PR :)

PR #1532, opened 2018-07-10
Fix to not drop MySQL and PostgreSQL DB when storage_mb value is changed

### LINKED ISSUE #1522: MySQL and PG should be inplace update when updating storage etc.
### Community Note

### Affected Resource(s)

* azurerm_mysql_server
* azurerm_postgresql_server

### Description

Current Terraform version, if any of the following parameters updated: server Tier, Family, Admin Login, Storage_mb, it resulted in deleting and recreating the server. This causes data lose. 

The expected behavior is to call the backend as stated in this doc: https://docs.microsoft.com/en-us/rest/api/mysql/servers/update. Backend will perform correct in-place updates.

jeffreyCline:
Fixes issue #1522
[Test Results](https://gist.github.com/jeffreyCline/706f3d3a59b27abf28142e17006c2e54)

(fixes #1522 )

### DIFF
--- a/azurerm/resource_arm_postgresql_server.go
+++ b/azurerm/resource_arm_postgresql_server.go
@@ -138,7 +138,6 @@ func resourceArmPostgreSQLServer() *schema.Resource {
 						"storage_mb": {
 							Type:         schema.TypeInt,
 							Required:     true,
-							ForceNew:     true,
 							ValidateFunc: validateIntBetweenDivisibleBy(5120, 4194304, 1024),
 						},
 
--- a/azurerm/resource_arm_mysql_server.go
+++ b/azurerm/resource_arm_mysql_server.go
@@ -135,7 +135,6 @@ func resourceArmMySqlServer() *schema.Resource {
 						"storage_mb": {
 							Type:         schema.TypeInt,
 							Required:     true,
-							ForceNew:     true,
 							ValidateFunc: validateIntBetweenDivisibleBy(5120, 4194304, 1024),
 						},
 						"backup_retention_days": {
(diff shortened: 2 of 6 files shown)


tombuildsstuff:
@jeffreyCline can you run all of the Postgres/MySQL tests for this?

azurerm/resource_arm_postgresql_server.go
```diff
@@ -138,7 +138,6 @@ func resourceArmPostgreSQLServer() *schema.Resource {
 						"storage_mb": {
 							Type:         schema.TypeInt,
 							Required:     true,
-							ForceNew:     true,
```
tombuildsstuff: Can we update the docs to mention that this is no longer ForceNew? (And above)

katbyte:
@tombuildsstuff I am running them now

azurerm/resource_arm_postgresql_server.go
jeffreyCline: Yeah, I thought about that after I did the initial push. I'll update those now.

jeffreyCline: I just looked at the documentation and the original text does not mention that it will ForceNew so I believe this is a no op:

```Markdown
* `storage_mb` - (Required) Max storage allowed for a server. Possible values are between `5120` MB(5GB) and `1048576` MB(1TB) for the Basic SKU and between `5120` MB(5GB) and `4194304` MB(4TB) for General Purpose/Memory Optimized SKUs. For more information see the [product documentation](https://docs.microsoft.com/en-us/rest/api/mysql/servers/create#StorageProfile).
```

PR #1534, opened 2018-07-11
Remove tags validate in preConfig of TestAccAzureRMKeyVault_update

lrxtom2:
There are no tags block in preConfig method, so we should remove validate tags in the fitst step of this test: TestAccAzureRMKeyVault_update


azurerm/resource_arm_key_vault_test.go
```diff
@@ -125,7 +125,6 @@ func TestAccAzureRMKeyVault_update(t *testing.T) {
 					testCheckAzureRMKeyVaultExists(resourceName),
 					resource.TestCheckResourceAttr(resourceName, "access_policy.0.key_permissions.0", "create"),
 					resource.TestCheckResourceAttr(resourceName, "access_policy.0.secret_permissions.0", "set"),
-					resource.TestCheckResourceAttr(resourceName, "tags.environment", "Production"),
```
tombuildsstuff: can we add this back in but instead check that `tags.%` is equal to `0` which will ensure there's no tags in the Pre phase

lrxtom2:
Thanks for @tombuildsstuff 
it's a good idea. i would like to follow if meet the same issue.

PR #1564, opened 2018-07-13
Azure Active Directory Service Principals

### LINKED ISSUE #16: Feature Request: `azurerm_service_principal`
_This issue was originally opened by @colemickens as hashicorp/terraform#8094. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Feature request: Support creating AAD users and assigning RBAC roles.

Scenario: I want to automate creating Kubernetes clusters for Azure. We're doing this currently with Terraform + Ansible. It currently requires the user to provision a service account for the cluster manually and then feed those credentials into a config file that is used in a template_file data resource and then ultimately pushed into the VM and consumed by Ansible.

If I could have an `azurerm_aad_serviceprincipal` then I could automatically provision new service accounts, without needing to do it out of band. (Currently I'm considering using a local-exec provisioner to run `azure-cli` to create the SP and assign it permissions to the resource group that Terraform is deploying/creating).

Depends on: https://github.com/Azure/azure-sdk-for-go/issues/378

tombuildsstuff:
This PR adds support for creating Service Principals associated with an Azure Active Directory Application.

- [x] Resource for Service Principal
- [x] Data Source for Service Principal
- [x] Resource for a Service Principal Key

There's an eventual consistency issue in the Active Directory Application resource, but I can't immediately see a way around that - so that'll be fixed in a follow up PR

Fixes #16 

```
* **New Data Source:** `azurerm_service_principal` [GH-1564]
* **New Resource:** `azurerm_service_principal` [GH-1564]
* **New Resource:** `azurerm_service_principal_password` [GH-1564]
```

### DIFF
--- a/azurerm/data_source_azuread_application.go
+++ b/azurerm/data_source_azuread_application.go
@@ -14,7 +14,6 @@ func dataSourceArmAzureADApplication() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
-		// TODO: customizeDiff for validation of either name or object_id.
 
 		Schema: map[string]*schema.Schema{
 			"object_id": {
--- a/azurerm/data_source_azuread_application_test.go
+++ b/azurerm/data_source_azuread_application_test.go
@@ -19,6 +19,9 @@ func TestAccDataSourceAzureRMAzureADApplication_byObjectId(t *testing.T) {
 		CheckDestroy: testCheckAzureRMActiveDirectoryApplicationDestroy,
 		Steps: []resource.TestStep{
 			{
+				Config: testAccAzureRMActiveDirectoryApplication_basic(id),
+			},
+			{
 				Config: config,
 				Check: resource.ComposeTestCheckFunc(
 					testCheckAzureRMActiveDirectoryApplicationExists(dataSourceName),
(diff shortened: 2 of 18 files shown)


tiwood:
Hi @tombuildsstuff, I've just created a similar PR (#1570 ). Just want to check how we go ahead?

Additionally I want to discuss what you think is the best way to add access credentials to the service principal object. Initially I've added a access_credential block to my implementation, but maybe a separate resource (azurerm_azuread_service_principal_access_credential) would be more fitting?

tombuildsstuff:
hey @tiwood

> Hi @tombuildsstuff, I've just created a similar PR (#1570 ). Just want to check how we go ahead?

Thanks for the PR #1570 :)

Taking a look through both PR's, it appears this one's slightly further along - as such I'm going to close #1570 in favour of this PR for the moment - I hope you don't mind!

> Additionally I want to discuss what you think is the best way to add access credentials to the service principal object. Initially I've added a access_credential block to my implementation, but maybe a separate resource (azurerm_azuread_service_principal_access_credential) would be more fitting?

I've spent a while prototyping both approaches and have come to a similar conclusion that this is better as a separate resource (since this allows for them to be more easily generated using the `count` syntax, for one thing); as such I've pushed a commit to this PR which adds the separate resource. I also wondered if it'd be worth integrating the `azurerm_azurerad_service_principal` resource into the `azurerm_azuread_application` resource, but on reflection I feel that's probably the wrong approach (and ended up leaving it as a separate resource).

One thing that's not included in either of our PR's is support for Service Principal Certificates (although yours contains a "credential_type" field, from what I can see this needs to be a separate resource specifically for Credentials). Since you've been looking into this area anyway - would you be interested in adding support for that instead?

Thanks!

tombuildsstuff:
Ignoring a transitory failure - the Data Source tests pass:

screenshot 2018-07-19 at 13 38 43

tombuildsstuff:
Resource tests pass:

screenshot 2018-07-19 at 13 42 22

PR #1574, opened 2018-07-14
Allow - in disk snapshot names

CoRfr:
It appears that - was omitted from the possible character in snapshot
names.

It doesn't seem like there is any issue when using - in the name, as
I was able to create snapshots with dashes both from the portal and
through this plugin (with this patch applied).

### DIFF
--- a/azurerm/resource_arm_snapshot.go
+++ b/azurerm/resource_arm_snapshot.go
@@ -216,10 +216,10 @@ func resourceArmSnapshotDelete(d *schema.ResourceData, meta interface{}) error {
 }
 
 func validateSnapshotName(v interface{}, k string) (ws []string, errors []error) {
-	// a-z, A-Z, 0-9 and _. The max name length is 80
+	// a-z, A-Z, 0-9, _ and -. The max name length is 80
 	value := v.(string)
 
-	r, _ := regexp.Compile("^[A-Za-z0-9_]+$")
+	r, _ := regexp.Compile("^[A-Za-z0-9_-]+$")
 	if !r.MatchString(value) {
 		errors = append(errors, fmt.Errorf("Snapshot Names can only contain alphanumeric characters and underscores."))
 	}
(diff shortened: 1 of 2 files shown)


CoRfr:
Oh I see where the format is coming from:

https://docs.microsoft.com/en-us/rest/api/compute/snapshots/update
> The name of the snapshot that is being created. The name can't be changed after the snapshot is created. Supported characters for the name are a-z, A-Z, 0-9 and _. The max name length is 80 characters.

In practice though, it appears that '-' is supported. I'm not sure how to request REST API updates.



The name of the snapshot that is being created. The name can't be changed after the snapshot is created. Supported characters for the name are a-z, A-Z, 0-9 and _. The max name length is 80 characters.
--


The name of the snapshot that is being created. The name can't be changed after the snapshot is created. Supported characters for the name are a-z, A-Z, 0-9 and _. The max name length is 80 characters.
--





The name of the snapshot that is being created. The name can't be changed after the snapshot is created. Supported characters for the name are a-z, A-Z, 0-9 and _. The max name length is 80 characters.
--




The name of the snapshot that is being created. The name can't be changed after the snapshot is created. Supported characters for the name are a-z, A-Z, 0-9 and _. The max name length is 80 characters.
--

CoRfr:
Also it seems that disks have the same requirements for names, which is not enforced in terraform:
https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate

The documentation is also outdated since - are clearly accepted for disk names as well.

tombuildsstuff:
hi @CoRfr 

Just to let you know that is now available in [v1.10 of the AzureRM Provider(https://github.com/terraform-providers/terraform-provider-azurerm/blob/v1.10.0/CHANGELOG.md#1100-july-21-2018) - which you can update to this version by specifying it in the Provider block, like so:

```
provider "azurerm" {
  version = "=1.10.0"
}
```

Thanks!

PR #1610, opened 2018-07-19
Added support for AKS network profile

cznewt:
Currently, it is not possible to set advanced networking plugin (azure) instead of the basic one (kubenet) for AKS.

### DIFF
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -123,6 +123,58 @@ func resourceArmKubernetesCluster() *schema.Resource {
 				},
 			},
 
+			"network_profile": {
+				Type:     schema.TypeList,
+				Optional: true,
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hey @cznewt 

Thanks for this PR :)

@lfshr has also been taking a look at adding support for Network Profiles in the `azurerm_kubernetes_cluster` resource in 
https://github.com/terraform-providers/terraform-provider-azurerm/pull/1479 ; however it's blocked on an upstream issue in the Go SDK (where it's not possible to provision the `Azure` network profile due to a validation bug, which given this PR is exposing the same functionality, I believe it'll be subject to the same issue.

Given these two PR's are the same (in that these changes are incorporated into that one) - I'm going to suggest we close this PR in favour of that one for the moment? Apologies for the bad timing here, but in either case I'd like to thank you for the contribution :)

Thanks!

Issue #1412, opened 2018-06-19
Unable to detect storage account network rules differences between configuration and actual Azure resource


liemnotliam:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version
Terraform v0.11.7
+ provider.azurerm v1.7.0

### Affected Resource(s)

* azurerm_storage_account

### Terraform Configuration Files

The bug happens when either `ip_rules` or `virtual_network_subnet_ids` are empty. `ip_rules` are set as empty for this example.

```hcl
resource "azurerm_resource_group" "test" {
  name     = "test-rg"
  location = "eastus"
}

resource "azurerm_virtual_network" "test" {
  name                = "test-vnet"
  address_space       = ["10.0.0.0/16"]
  location            = "${azurerm_resource_group.test.location}"
  resource_group_name = "${azurerm_resource_group.test.name}"
}

resource "azurerm_subnet" "test" {
  name                 = "subnet"
  resource_group_name  = "${azurerm_resource_group.test.name}"
  virtual_network_name = "${azurerm_virtual_network.test.name}"
  address_prefix       = "10.0.2.0/24"
  service_endpoints    = ["Microsoft.Storage"]
}

resource "azurerm_storage_account" "test" {
  name                = "luewabsw12"
  resource_group_name = "${azurerm_resource_group.test.name}"

  location                 = "${azurerm_resource_group.test.location}"
  account_tier             = "Standard"
  account_replication_type = "LRS"

  network_rules {
    ip_rules                   = []
    virtual_network_subnet_ids = ["${azurerm_subnet.test.id}"]
  }
}

```

### Expected Behavior
Given the same Terraform configuration, if there are changes in storage account network rules, then Terraform should detect them. Given the example configuration above, if the IP rules remain empty and the virtual network rules differ from the Terraform configuration then Terraform should detect the change. Currently it doesn't due to the checks [here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_storage_account.go#L674).

### Actual Behavior
No changes found between resource configuration and the actual Azure storage account.
```
[...]

PR #1613, opened 2018-07-19
Additional checks for network_rules during read

mbfrahry:
This PR address #1412. 

The plan from that users config will no return 
```
  ~ azurerm_storage_account.test
      network_rules.#:                                       "0" => "1"
      network_rules.0.bypass.#:                              "" => <computed>
      network_rules.0.virtual_network_subnet_ids.#:          "0" => "1"
      network_rules.0.virtual_network_subnet_ids.3418614067: "" => "/subscriptions/c0a607b2-6372-4ef3-abdb-dbe52a7b56ba/resourceGroups/test-rg1412/providers/Microsoft.Network/virtualNetworks/test-vnet/subnets/subnet"
```

Tests still pass 
```
=== RUN   TestAccAzureRMStorageAccount_networkRules
--- PASS: TestAccAzureRMStorageAccount_networkRules (123.25s)
=== RUN   TestAccAzureRMStorageAccount_networkRulesDeleted
--- PASS: TestAccAzureRMStorageAccount_networkRulesDeleted (109.85s)
```

### DIFF
--- a/azurerm/resource_arm_storage_account.go
+++ b/azurerm/resource_arm_storage_account.go
@@ -701,7 +701,7 @@ func resourceArmStorageAccountRead(d *schema.ResourceData, meta interface{}) err
 		}
 
 		networkRules := props.NetworkRuleSet
-		if networkRules != nil && len(*networkRules.IPRules) > 0 && len(*networkRules.VirtualNetworkRules) > 0 {
+		if networkRules != nil {
 			if err := d.Set("network_rules", flattenStorageAccountNetworkRules(networkRules)); err != nil {
 				return fmt.Errorf("Error flattening `network_rules`: %+v", err)
 			}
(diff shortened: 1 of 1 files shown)


azurerm/resource_arm_storage_account.go
```diff
@@ -701,7 +701,7 @@ func resourceArmStorageAccountRead(d *schema.ResourceData, meta interface{}) err
 		}
 
 		networkRules := props.NetworkRuleSet
-		if networkRules != nil && len(*networkRules.IPRules) > 0 && len(*networkRules.VirtualNetworkRules) > 0 {
+		if networkRules != nil {
```
tombuildsstuff: we should be able to remove this `if` statement entirely and replace it with ones inside the flatten functions? this means we'll set the fields to an empty value if nil is returned from the api?

azurerm/resource_arm_storage_account.go
```diff
@@ -829,6 +829,9 @@ func expandStorageAccountBypass(networkRule map[string]interface{}) storage.Bypa
 }
 
 func flattenStorageAccountNetworkRules(input *storage.NetworkRuleSet) []interface{} {
+	if len(*input.IPRules) == 0 && len(*input.VirtualNetworkRules) == 0 {
```
tombuildsstuff: which means this could become

```
if input == nil {
  return []interface{}{}
}
```

tombuildsstuff: and we could make this: https://github.com/terraform-providers/terraform-provider-azurerm/pull/1613/files#diff-f4d2502d01b7fb77deb7bb221025fb51R845

```
ipRules := make([]interface{}, 0)

if input != nil {
  for i, ipRule := range *input {
    if range := ipRule.IPAddressOrRange; range != nil {
      ipRules[i] = *range
    }
  }
}

return ipRules
```

(and same for the other two sub-functions)

Issue #1392, opened 2018-06-14
azurerm_automation_schedule is missing properties: week_days, month_days, month_week_day


katbyte:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The resource `azurerm_automation_schedule` is missing the following properties: `week_days`, `month_days`, `month_week_day`

### New or Affected Resource(s)

* azurerm_automation_schedule

katbyte:
Shouldn't be done till after #1384 is merged

katbyte:
Was fixed by #1626

PR #1626, opened 2018-07-22
azurerm_automation_schedule - Add missing properties for the advanced schedule

TFaga:
This pull request adds the missing properties in the `azurerm_automation_schedule` resource to configure the advanced schedule. Addresses issue #1392

### DIFF
--- a/azurerm/resource_arm_automation_schedule.go
+++ b/azurerm/resource_arm_automation_schedule.go
@@ -108,68 +108,62 @@ func resourceArmAutomationSchedule() *schema.Resource {
 				//todo figure out how to validate this properly
 			},
 
-			"advanced_schedule": {
+			"week_days": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+					ValidateFunc: validation.StringInSlice([]string{
+						string(automation.Monday),
+						string(automation.Tuesday),
+						string(automation.Wednesday),
+						string(automation.Thursday),
+						string(automation.Friday),
+						string(automation.Saturday),
+						string(automation.Sunday),
+					}, true),
+				},
+				Set:           set.HashStringIgnoreCase,
+				ConflictsWith: []string{"month_days", "monthly_occurrence"},
+			},
+			"month_days": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeInt,
+					ValidateFunc: validate.IntBetweenAndNot(-1, 31, 0),
+				},
+				Set:           set.HashInt,
+				ConflictsWith: []string{"week_days", "monthly_occurrence"},
+			},
+
+			"monthly_occurrence": {
 				Type:     schema.TypeList,
 				Optional: true,
-				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"week_days": {
-							Type:     schema.TypeSet,
-							Optional: true,
-							Elem: &schema.Schema{
-								Type: schema.TypeString,
-								ValidateFunc: validation.StringInSlice([]string{
-									string(automation.Monday),
-									string(automation.Tuesday),
-									string(automation.Wednesday),
-									string(automation.Thursday),
-									string(automation.Friday),
-									string(automation.Saturday),
-									string(automation.Sunday),
-								}, true),
-							},
-							Set: set.HashStringIgnoreCase,
+						"day": {
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: suppress.CaseDifference,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(automation.Monday),
+								string(automation.Tuesday),
+								string(automation.Wednesday),
+								string(automation.Thursday),
+								string(automation.Friday),
+								string(automation.Saturday),
+								string(automation.Sunday),
+							}, true),
 						},
-						"month_days": {
-							Type:     schema.TypeSet,
-							Optional: true,
-							Elem: &schema.Schema{
-								Type:         schema.TypeInt,
-								ValidateFunc: validate.IntBetweenAndNot(-1, 31, 0),
-							},
-							Set: set.HashInt,
-						},
-
-						"monthly_occurrence": {
-							Type:     schema.TypeList,
-							Optional: true,
-							Elem: &schema.Resource{
-								Schema: map[string]*schema.Schema{
-									"day": {
-										Type:             schema.TypeString,
-										Required:         true,
-										DiffSuppressFunc: suppress.CaseDifference,
-										ValidateFunc: validation.StringInSlice([]string{
-											string(automation.Monday),
-											string(automation.Tuesday),
-											string(automation.Wednesday),
-											string(automation.Thursday),
-											string(automation.Friday),
-											string(automation.Saturday),
-											string(automation.Sunday),
-										}, true),
-									},
-									"occurrence": {
-										Type:         schema.TypeInt,
-										Required:     true,
-										ValidateFunc: validation.IntBetween(1, 5),
-									},
-								},
-							},
+						"occurrence": {
+							Type:         schema.TypeInt,
+							Required:     true,
+							ValidateFunc: validate.IntBetweenAndNot(-1, 5, 0),
 						},
 					},
 				},
+				ConflictsWith: []string{"week_days", "month_days"},
 			},
 		},
 
@@ -178,24 +172,22 @@ func resourceArmAutomationSchedule() *schema.Resource {
 			frequency := strings.ToLower(diff.Get("frequency").(string))
 			interval, _ := diff.GetOk("interval")
 			if frequency == "onetime" && interval.(int) > 0 {
-				return fmt.Errorf("`interval` cannot be set when frequency is not OneTime")
+				return fmt.Errorf("`interval` cannot be set when frequency is `OneTime`")
 			}
 
-			advancedSchedules, hasAdvancedSchedule := diff.GetOk("advanced_schedule")
-			if hasAdvancedSchedule {
-				if asl := advancedSchedules.([]interface{}); len(asl) > 0 {
-					if frequency != "week" && frequency != "month" {
-						return fmt.Errorf("`advanced_schedule` can only be set when frequency is `Week` or `Month`")
-					}
+			_, hasWeekDays := diff.GetOk("week_days")
+			if hasWeekDays && frequency != "week" {
+				return fmt.Errorf("`week_days` can only be set when frequency is `Week`")
+			}
 
-					as := asl[0].(map[string]interface{})
-					if frequency == "week" && as["week_days"].(*schema.Set).Len() == 0 {
-						return fmt.Errorf("`week_days` must be set when frequency is `Week`")
-					}
-					if frequency == "month" && as["month_days"].(*schema.Set).Len() == 0 && len(as["monthly_occurrence"].([]interface{})) == 0 {
-						return fmt.Errorf("Either `month_days` or `monthly_occurrence` must be set when frequency is `Month`")
-					}
-				}
+			_, hasMonthDays := diff.GetOk("month_days")
+			if hasMonthDays && frequency != "month" {
+				return fmt.Errorf("`month_days` can only be set when frequency is `Month`")
+			}
+
+			_, hasMonthlyOccurances := diff.GetOk("monthly_occurrence")
+			if hasMonthlyOccurances && frequency != "month" {
+				return fmt.Errorf("`monthly_occurrence` can only be set when frequency is `Month`")
 			}
 
 			_, hasAccount := diff.GetOk("automation_account_name")
@@ -275,25 +267,13 @@ func resourceArmAutomationScheduleCreateUpdate(d *schema.ResourceData, meta inte
 		}
 	}
 
-	//only pay attention to the advanced schedule if frequency is either Week or Month
+	//only pay attention to the advanced schedule fields if frequency is either Week or Month
 	if properties.Frequency == automation.Week || properties.Frequency == automation.Month {
-		isUpdate := d.Id() != ""
-		if v, ok := d.GetOk("advanced_schedule"); ok {
-			if vl := v.([]interface{}); len(vl) > 0 {
-				advancedRef, err := expandArmAutomationScheduleAdvanced(vl, isUpdate)
-				if err != nil {
-					return err
-				}
-				properties.AdvancedSchedule = advancedRef
-			}
-		} else if isUpdate {
-			// To make sure all the properties are unset in the event of removal update, pass in an empty skeleton object
-			properties.AdvancedSchedule = &automation.AdvancedSchedule{
-				WeekDays:           &[]string{},
-				MonthDays:          &[]int32{},
-				MonthlyOccurrences: &[]automation.AdvancedScheduleMonthlyOccurrence{},
-			}
+		advancedRef, err := expandArmAutomationScheduleAdvanced(d, d.Id() != "")
+		if err != nil {
+			return err
 		}
+		properties.AdvancedSchedule = advancedRef
 	}
 
 	_, err := client.CreateOrUpdate(ctx, resGroup, accountName, name, parameters)
@@ -360,8 +340,17 @@ func resourceArmAutomationScheduleRead(d *schema.ResourceData, meta interface{})
 	if v := resp.TimeZone; v != nil {
 		d.Set("timezone", v)
 	}
-	if err := d.Set("advanced_schedule", flattenArmAutomationScheduleAdvanced(resp.AdvancedSchedule)); err != nil {
-		return fmt.Errorf("Error setting `advanced_schedule`: %+v", err)
+
+	if v := resp.AdvancedSchedule; v != nil {
+		if err := d.Set("week_days", flattenArmAutomationScheduleAdvancedWeekDays(v)); err != nil {
+			return fmt.Errorf("Error setting `week_days`: %+v", err)
+		}
+		if err := d.Set("month_days", flattenArmAutomationScheduleAdvancedMonthDays(v)); err != nil {
+			return fmt.Errorf("Error setting `month_days`: %+v", err)
+		}
+		if err := d.Set("monthly_occurrence", flattenArmAutomationScheduleAdvancedMonthlyOccurrences(v)); err != nil {
+			return fmt.Errorf("Error setting `monthly_occurrence`: %+v", err)
+		}
 	}
 	return nil
 }
@@ -389,35 +378,36 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
 	return nil
 }
 
-func expandArmAutomationScheduleAdvanced(v []interface{}, isUpdate bool) (*automation.AdvancedSchedule, error) {
-
-	// Get the one and only advance schedule configuration
-	advancedSchedule := v[0].(map[string]interface{})
-	weekDays := advancedSchedule["week_days"].(*schema.Set).List()
-	monthDays := advancedSchedule["month_days"].(*schema.Set).List()
-	monthlyOccurrences := advancedSchedule["monthly_occurrence"].([]interface{})
+func expandArmAutomationScheduleAdvanced(d *schema.ResourceData, isUpdate bool) (*automation.AdvancedSchedule, error) {
 
 	expandedAdvancedSchedule := automation.AdvancedSchedule{}
 
 	// If frequency is set to `Month` the `week_days` array cannot be set (even empty), otherwise the API returns an error.
 	// During update it can be set and it will not return an error. Workaround for the APIs behavior
-	if len(weekDays) > 0 || isUpdate {
+	if v, ok := d.GetOk("week_days"); ok {
+		weekDays := v.(*schema.Set).List()
 		expandedWeekDays := make([]string, len(weekDays))
 		for i := range weekDays {
 			expandedWeekDays[i] = weekDays[i].(string)
 		}
 		expandedAdvancedSchedule.WeekDays = &expandedWeekDays
+	} else if isUpdate {
+		expandedAdvancedSchedule.WeekDays = &[]string{}
 	}
 
 	// Same as above with `week_days`
-	if len(monthDays) > 0 || isUpdate {
+	if v, ok := d.GetOk("month_days"); ok {
+		monthDays := v.(*schema.Set).List()
 		expandedMonthDays := make([]int32, len(monthDays))
 		for i := range monthDays {
 			expandedMonthDays[i] = int32(monthDays[i].(int))
 		}
 		expandedAdvancedSchedule.MonthDays = &expandedMonthDays
+	} else if isUpdate {
+		expandedAdvancedSchedule.MonthDays = &[]int32{}
 	}
 
+	monthlyOccurrences := d.Get("monthly_occurrence").([]interface{})
 	expandedMonthlyOccurrences := make([]automation.AdvancedScheduleMonthlyOccurrence, len(monthlyOccurrences))
 	for i := range monthlyOccurrences {
 		m := monthlyOccurrences[i].(map[string]interface{})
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_automation_schedule.go
```diff
@@ -107,14 +108,78 @@ func resourceArmAutomationSchedule() *schema.Resource {
...
 			interval, _ := diff.GetOk("interval")
-			if strings.ToLower(diff.Get("frequency").(string)) == "onetime" && interval.(int) > 0 {
-				return fmt.Errorf("interval canot be set when frequency is not OneTime")
+			if frequency == "onetime" && interval.(int) > 0 {
+				return fmt.Errorf("`interval` cannot be set when frequency is not OneTime")
+			}
+
+			advancedSchedules, hasAdvancedSchedule := diff.GetOk("advanced_schedule")
+			if hasAdvancedSchedule {
+				if asl := advancedSchedules.([]interface{}); len(asl) > 0 && asl[0] != nil {
```
tombuildsstuff: I think we can remove `asl[0] != nil` since that'll always be true if `len(asl) > 0`?

azurerm/resource_arm_automation_schedule.go
```diff
@@ -193,6 +258,20 @@ func resourceArmAutomationScheduleCreateUpdate(d *schema.ResourceData, meta inte
 			properties.Interval = 1
 		}
 	}
+
+	//only pay attention to the advanced schedule if frequency is either Week or Month
+	if properties.Frequency == automation.Week || properties.Frequency == automation.Month {
+		if v, ok := d.GetOk("advanced_schedule"); ok {
+			if vl := v.([]interface{}); len(vl) > 0 && vl[0] != nil {
```
tombuildsstuff: same here - I think we can remove `vl[0] != nil` since that's implied?

azurerm/resource_arm_automation_schedule.go
```diff
@@ -257,6 +336,9 @@ func resourceArmAutomationScheduleRead(d *schema.ResourceData, meta interface{})
 	if v := resp.TimeZone; v != nil {
 		d.Set("timezone", v)
 	}
+	if v := resp.AdvancedSchedule; v != nil {
+		d.Set("advanced_schedule", flattenArmAutomationScheduleAdvanced(v))
+	}
```
tombuildsstuff: can we ensure we set this all the time and move the nil-check within the flatten function? that ensures changes to this are detected. in addition - given this is a complex object can we check for errors when setting it? e.g.

```
if err := d.Set("advanced_schedule", flattenArmAutomationScheduleAdvanced(v)); err != nil {
  return fmt.Errorf("Error setting `advanced_schedule`: %+v", err)
}
```

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +364,90 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+
+	// Get the one and only advance schedule configuration
+	advancedSchedule := v[0].(map[string]interface{})
+	weekDays := advancedSchedule["week_days"].(*schema.Set).List()
+	monthDays := advancedSchedule["month_days"].(*schema.Set).List()
+	monthlyOccurrences := advancedSchedule["monthly_occurrence"].([]interface{})
+
+	expandedAdvancedSchedule := automation.AdvancedSchedule{}
+
+	if len(weekDays) > 0 {
```
tombuildsstuff: can we remove this if statement - since it means it's not possible to remove these values once they've been set.

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +364,90 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+		expandedWeekDays := make([]string, len(weekDays))
+		for i := range weekDays {
+			expandedWeekDays[i] = weekDays[i].(string)
+		}
+		expandedAdvancedSchedule.WeekDays = &expandedWeekDays
+	}
+
+	if len(monthDays) > 0 {
```
tombuildsstuff: can we remove this if statement - since it means it's not possible to remove these values once they've been set.

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +364,90 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+		expandedMonthDays := make([]int32, len(monthDays))
+		for i := range monthDays {
+			expandedMonthDays[i] = int32(monthDays[i].(int))
+		}
+		expandedAdvancedSchedule.MonthDays = &expandedMonthDays
+	}
+
+	if len(monthlyOccurrences) > 0 {
```
tombuildsstuff: can we remove this if statement - since it means it's not possible to remove these values once they've been set.

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +364,90 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+func flattenArmAutomationScheduleAdvanced(v *automation.AdvancedSchedule) []interface{} {
+	result := make(map[string]interface{})
+
+	if v.WeekDays != nil {
+		flattenedWeekDays := schema.NewSet(set.HashStringIgnoreCase, []interface{}{})
+		for i := range *v.WeekDays {
+			flattenedWeekDays.Add((*v.WeekDays)[i])
+		}
+		result["week_days"] = flattenedWeekDays
+	}
```
tombuildsstuff: can we update this to be:

```
flattenedWeekDays := schema.NewSet(set.HashStringIgnoreCase, []interface{}{})
if v.WeekDays != nil {
  for _, v := range *v.WeekDays {
    flattenedWeekDays.Add(v)
  }
}
result["week_days"] = flattenedWeekDays
```

this allows this to be set even if nil's returned from the API - which means diff's will be shown

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +364,90 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+	if v.MonthDays != nil {
+		flattenedMonthDays := schema.NewSet(set.HashInt, []interface{}{})
+		for i := range *v.MonthDays {
+			flattenedMonthDays.Add(int(((*v.MonthDays)[i])))
+		}
+		result["month_days"] = flattenedMonthDays
+	}
```
tombuildsstuff: (as above)

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +364,90 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+	if v.MonthlyOccurrences != nil {
+		flattenedMonthlyOccurrences := make([]map[string]interface{}, len(*v.MonthlyOccurrences))
+		for i := range *v.MonthlyOccurrences {
+			f := make(map[string]interface{})
+			f["day"] = (*v.MonthlyOccurrences)[i].Day
+			f["occurrence"] = int(*(*v.MonthlyOccurrences)[i].Occurrence)
+			flattenedMonthlyOccurrences[i] = f
+		}
+		result["monthly_occurrence"] = flattenedMonthlyOccurrences
+	}
```
tombuildsstuff: (as above)

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +364,90 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+				Day:        automation.ScheduleDay(m["day"].(string)),
+			}
+		}
+		expandedAdvancedSchedule.MonthlyOccurrences = &expandedMonthlyOccurrences
+	}
+
+	return &expandedAdvancedSchedule, nil
+}
+
+func flattenArmAutomationScheduleAdvanced(v *automation.AdvancedSchedule) []interface{} {
```
tombuildsstuff: given we're now passing in a nil-value here, can we return an empty list e.g.

```
if v == nil {
  return []interface{}{}
}
```

azurerm/resource_arm_automation_schedule.go
```diff
@@ -107,14 +108,78 @@ func resourceArmAutomationSchedule() *schema.Resource {
...
+						"monthly_occurrence": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"day": {
+										Type:             schema.TypeString,
+										Required:         true,
+										DiffSuppressFunc: suppress.CaseDifference,
+										ValidateFunc:     validateScheduleDay(),
```
tombuildsstuff: can we in-line this validation method for the moment? I have a feeling we may end up extracting out a `listOfWeekDays` method in time

website/docs/r/automation_schedule.html.markdown
```diff
@@ -61,6 +65,22 @@ The following arguments are supported:
 
 * `timezone` - (Optional) The timezone of the start time. Defaults to `UTC`. For possible values see: https://msdn.microsoft.com/en-us/library/ms912391(v=winembedded.11).aspx
 
+* `advanced_schedule` - (Optional) Advanced fine-grained schedule frequency configuration. The `advanced_schedule` block supports fields documented below.
+
+The `advanced_schedule` block supports:
```
tombuildsstuff: _minor_ can we add `---` above this line (and then a blank line) to add a divider

website/docs/r/automation_schedule.html.markdown
```diff
@@ -61,6 +65,22 @@ The following arguments are supported:
...
+
+* `week_days` - (Optional) List of days of the week that the job should execute on.
+
+* `month_days` - (Optional) List of days of the month that the job should execute on. Must be between 1 and 31. 0 for last day of the month.
+
+* `monthly_occurrence` - (Optional) List of occurrences of days within a month. The `monthly_occurrence` block supports fields documented below.
+
+The `monthly_occurrence` block supports:
```
tombuildsstuff: (as above) can we add a divider here by adding `---` and then a new line?

TFaga:
Hey!

Thanks for the quick response and check. I've went through the comments and made the appropriate changes to the code. 

I've however noticed one behavior that may be problematic. If you specify the frequency of the schedule as either `Week` or `Month` and don't add any advance scheduling properties, the API will still return a present (though empty) `advancedSchedule` object in the response, like so:

```json
{
    "properties": {
        ...
        "interval": 1,
        "frequency": "Month",
        "advancedSchedule": {
            "monthDays": null,
            "monthlyOccurrences": null,
            "weekDays": null
        }
    }
}
```
This means that it gets saved in the state as having one item in the `advanced_schedule` while there are none specified in the `.tf` file. In conclusion this means that on each `apply` it will try to remove this item, which effectively won't do anything and making next `apply` do the same. Any thoughts on what best to do here?

Thanks!

azurerm/resource_arm_automation_schedule.go
TFaga: I've noticed an issue with this particular property. When creating the resource, if frequency is set to `Month` the `week_days` array cannot be set (even empty), otherwise the API returns an error. However during update it can and should be in order to reset it if it was removed from state. I've added a check for whether the current operation is a Create or Update and act accordingly. Perhaps separating the create and update methods would be a more clean solution?

azurerm/resource_arm_automation_schedule.go
TFaga: Same as with `weekDays`.

azurerm/resource_arm_automation_schedule.go
TFaga: The problem above is not present with `monthlyOccurrences` so I removed any checks here as advised.

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +387,84 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+	weekDays := advancedSchedule["week_days"].(*schema.Set).List()
+	monthDays := advancedSchedule["month_days"].(*schema.Set).List()
+	monthlyOccurrences := advancedSchedule["monthly_occurrence"].([]interface{})
+
+	expandedAdvancedSchedule := automation.AdvancedSchedule{}
+
+	// If frequency is set to `Month` the `week_days` array cannot be set (even empty), otherwise the API returns an error.
+	// During update it can be set and it will not return an error. Workaround for the APIs behavior
+	if len(weekDays) > 0 || update {
+		expandedWeekDays := make([]string, len(weekDays))
```
tombuildsstuff: so this second argument is the starting position in the array, if we make this `0` there's no empty element

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +387,84 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+		for i := range weekDays {
+			expandedWeekDays[i] = weekDays[i].(string)
+		}
+		expandedAdvancedSchedule.WeekDays = &expandedWeekDays
+	}
+
+	// Same as above with `week_days`
+	if len(monthDays) > 0 || update {
+		expandedMonthDays := make([]int32, len(monthDays))
```
tombuildsstuff: so this second argument is the starting position in the array, if we make this `0` there's no empty element

TFaga:
@tombuildsstuff Hey, thanks for the quick response and info. I'm not sure I understand though, or rather I think I didn't phrase it very well. The problem that arises is that the schedule client/API is inconsistent with what properties of the advance schedule it deems valid on input. 

When creating a new resource and for instance setting the `frequency` as `Month` the `week_days` property must not be set (not even array with length 0; it needs to be `nil`) or the call will return a validation error. Same deal for frequency `Week` and `month_days` property. However during an update it can be set to an array with length 0 and it will work as expected. That is why that `if` statement is there; during the initial creation of the resource, if the `week_days` or `month_days` are not set, the property for the client must not be set as well (they can and should be during an update).

To clarify with API examples.

This will error:

```js
{
    "properties": {
        "frequency": "Month",
        "advancedSchedule": {
            "monthDays": [2],
            "monthlyOccurrences": [],
            "weekDays": [] // <- this
        }
    }
}
```

While this will succeed:


```js
{
    "properties": {
        "frequency": "Month",
        "advancedSchedule": {
            "monthDays": [2],
            "monthlyOccurrences": [],
            "weekDays": null // <- this
        }
    }
}
```

Only on create, however. On update the first example will work without issue.

I've pushed a commit addressing the same problem if the whole `advanced_schedule` block is not present in the config or if it's removed from an existing resource and updated.

katbyte:
Hi @TFaga,

You can always add a `MinItems: 1` to `weekDays` property to prevent an empty array being passed in.

Additionally, might i suggest pulling `monthDays`, `monthlyOccurrences` and `weekDays` out of a block to the top level for simplicities sake? that would also help with the empty `advancedSchedule` response.

What do you think @tombuildsstuff ?

website/docs/r/automation_schedule.html.markdown
```diff
@@ -61,6 +65,26 @@ The following arguments are supported:
...
 
+* `advanced_schedule` - (Optional) Advanced fine-grained schedule frequency configuration. The `advanced_schedule` block supports fields documented below.
+
+---
+
+The `advanced_schedule` block supports:
+
+* `week_days` - (Optional) List of days of the week that the job should execute on.
+
+* `month_days` - (Optional) List of days of the month that the job should execute on. Must be between 1 and 31. 0 for last day of the month.
```
JunyiYi: >0 [](start = 112, length = 1)

*minor* can we quote the numbers in backtick such that " Must be between `1` and `31`".
*question* are you sure `0` represents the last day of the month, or `-1`?

website/docs/r/automation_schedule.html.markdown
```diff
@@ -61,6 +65,26 @@ The following arguments are supported:
...
+
+* `monthly_occurrence` - (Optional) List of occurrences of days within a month. The `monthly_occurrence` block supports fields documented below.
+
+---
+
+The `monthly_occurrence` block supports:
+
+* `day` - (Required) Day of the occurrence. Must be one of Monday, Tuesday, Wednesday, Thursday, Friday, Saturday, Sunday.
+
+* `occurrence` - (Required) Occurrence of the week within the month. Must be between 1 and 5.
```
JunyiYi: *question* tried with the Azure Rest API site, I also see `-1` is available here (e.g. `"Monday" -1` means the last Monday of a month). Can you confirm whether it is supported or not?

azurerm/resource_arm_automation_schedule_test.go
```diff
@@ -338,3 +405,113 @@ func checkAccAzureRMAutomationSchedule_recurring_basic(resourceName string, freq
 		resource.TestCheckResourceAttr(resourceName, "timezone", "UTC"),
 	)
 }
+
+func testAccAzureRMAutomationSchedule_recurring_advanced_week(rInt int, location, frequency string, interval int, weekDay string) string {
```
JunyiYi: >interval int [](start = 100, length = 12)

*suggestion* The only possible value used in your test cases is `1`, so you may hardcode the `1` in the HCL below and remove this parameter. Similar two suggestions below.

azurerm/resource_arm_automation_schedule.go
```diff
@@ -107,14 +108,94 @@ func resourceArmAutomationSchedule() *schema.Resource {
 				//todo figure out how to validate this properly
 			},
 
-			//todo missing properties: week_days, month_days, month_week_day from advanced automation section
+			"advanced_schedule": {
```
JunyiYi: >"advanced_schedule" [](start = 3, length = 19)

*suggestion* We don't need to introduce wrapper `advanced_schedule` object here. What I mean is to move the mutual exclusive objects `week_days`, `month_days` and `monthly_occurrence` to the top level so we can reduce the depth of the hierarchy.

azurerm/resource_arm_automation_schedule.go
```diff
@@ -107,14 +108,94 @@ func resourceArmAutomationSchedule() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"week_days": {
```
JunyiYi: >"week_days" [](start = 6, length = 11)

Since `week_days`, `month_days` and `monthly_occurrence` are mutually exclusive, why not declare `ConflictsWith` attribute in these three objects?

azurerm/resource_arm_automation_schedule.go
```diff
@@ -107,14 +108,94 @@ func resourceArmAutomationSchedule() *schema.Resource {
...
+											string(automation.Thursday),
+											string(automation.Friday),
+											string(automation.Saturday),
+											string(automation.Sunday),
+										}, true),
+									},
+									"occurrence": {
+										Type:         schema.TypeInt,
+										Required:     true,
+										ValidateFunc: validation.IntBetween(1, 5),
```
JunyiYi: >validation.IntBetween(1, 5) [](start = 24, length = 27)

Can you confirm whether Go SDK supports `-1` or not?

azurerm/resource_arm_automation_schedule.go
```diff
@@ -107,14 +108,94 @@ func resourceArmAutomationSchedule() *schema.Resource {
...
 		},
 
 		CustomizeDiff: func(diff *schema.ResourceDiff, v interface{}) error {
 
+			frequency := strings.ToLower(diff.Get("frequency").(string))
 			interval, _ := diff.GetOk("interval")
-			if strings.ToLower(diff.Get("frequency").(string)) == "onetime" && interval.(int) > 0 {
-				return fmt.Errorf("interval canot be set when frequency is not OneTime")
+			if frequency == "onetime" && interval.(int) > 0 {
+				return fmt.Errorf("`interval` cannot be set when frequency is not OneTime")
```
JunyiYi: >"`interval` cannot be set when frequency is not OneTime" [](start = 22, length = 56)

`interval` cannot be set when frequency is ~~not~~ `OneTime`.

JunyiYi:
Hi @TFaga , thanks for the contribution. I have looked through your code and provided some comments. Feel free to ask if some of them confused your.

website/docs/r/automation_schedule.html.markdown
TFaga: It's indeed `-1`. Thank you for the correction.

website/docs/r/automation_schedule.html.markdown
TFaga: It's supported. Confirmed and added.

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +377,81 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+		}
+	}
+	expandedAdvancedSchedule.MonthlyOccurrences = &expandedMonthlyOccurrences
+
+	return &expandedAdvancedSchedule, nil
+}
+
+func flattenArmAutomationScheduleAdvancedWeekDays(v *automation.AdvancedSchedule) *schema.Set {
+	flattenedWeekDays := schema.NewSet(set.HashStringIgnoreCase, []interface{}{})
+	if v.WeekDays != nil {
```
katbyte: I believe this could be written as:
```go
    if weekDays := v.WeekDays; weekdays != nil {
        for i, v := range *weekDays {
			flattenedWeekDays.Add(v)
		}
	}
```

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +377,81 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+		for i := range *v.WeekDays {
+			flattenedWeekDays.Add((*v.WeekDays)[i])
+		}
+	}
+	return flattenedWeekDays
+}
+
+func flattenArmAutomationScheduleAdvancedMonthDays(v *automation.AdvancedSchedule) *schema.Set {
+	flattenedMonthDays := schema.NewSet(set.HashInt, []interface{}{})
+	if v.MonthDays != nil {
```
katbyte: The same as above, could definitely be simplified.

azurerm/resource_arm_automation_schedule.go
```diff
@@ -282,3 +377,81 @@ func resourceArmAutomationScheduleDelete(d *schema.ResourceData, meta interface{
...
+			flattenedMonthDays.Add(int(((*v.MonthDays)[i])))
+		}
+	}
+	return flattenedMonthDays
+}
+
+func flattenArmAutomationScheduleAdvancedMonthlyOccurrences(v *automation.AdvancedSchedule) []map[string]interface{} {
+	flattenedMonthlyOccurrences := make([]map[string]interface{}, 0)
+	if v.MonthlyOccurrences != nil {
+		for i := range *v.MonthlyOccurrences {
```
katbyte: Again, could use the syntax: `for _, v := range monthlyOccurrences {` and not have to deref and index so much.

azurerm/resource_arm_automation_schedule_test.go
```diff
@@ -338,3 +405,99 @@ func checkAccAzureRMAutomationSchedule_recurring_basic(resourceName string, freq
...
+	)
+}
+
+func testAccAzureRMAutomationSchedule_recurring_advanced_month_week_day(rInt int, location string, weekDay string, weekDayOccurrence int) string {
+
+	return fmt.Sprintf(`
+%s
+
+resource "azurerm_automation_schedule" "test" {
+  name	 	              = "acctestAS-%d"
```
katbyte: Could we make sure the =s are aligned here?

azurerm/resource_arm_automation_schedule_test.go
```diff
@@ -338,3 +405,99 @@ func checkAccAzureRMAutomationSchedule_recurring_basic(resourceName string, freq
...
+		resource.TestCheckResourceAttr(resourceName, "week_days.#", "1"),
+	)
+}
+
+func testAccAzureRMAutomationSchedule_recurring_advanced_month(rInt int, location string, monthDay int) string {
+	return fmt.Sprintf(`
+%s
+
+resource "azurerm_automation_schedule" "test" {
+  name	 	              = "acctestAS-%d"
```
katbyte: Could we make sure the =s are aligned here?

azurerm/resource_arm_automation_schedule_test.go
```diff
@@ -338,3 +405,99 @@ func checkAccAzureRMAutomationSchedule_recurring_basic(resourceName string, freq
 		resource.TestCheckResourceAttr(resourceName, "timezone", "UTC"),
 	)
 }
+
+func testAccAzureRMAutomationSchedule_recurring_advanced_week(rInt int, location string, weekDay string) string {
+	return fmt.Sprintf(`
+%s
+
+resource "azurerm_automation_schedule" "test" {
+  name	 	              = "acctestAS-%d"
```
katbyte: Could we make sure the =s are aligned here?

website/docs/r/automation_schedule.html.markdown
```diff
@@ -61,6 +65,20 @@ The following arguments are supported:
 
 * `timezone` - (Optional) The timezone of the start time. Defaults to `UTC`. For possible values see: https://msdn.microsoft.com/en-us/library/ms912391(v=winembedded.11).aspx
 
+* `week_days` - (Optional) List of days of the week that the job should execute on. Only valid for `Week`.
```
katbyte: Could we be a bit clearer by phrasing `Only valid for 'Week'` as `Only valid when frequency is 'Week'`? Applies to the new two properties to

website/docs/r/automation_schedule.html.markdown
```diff
@@ -61,6 +65,20 @@ The following arguments are supported:
...
+
+* `month_days` - (Optional) List of days of the month that the job should execute on. Must be between `1` and `31`. `-1` for last day of the month. Only valid for `Month`.
+
+* `monthly_occurrence` - (Optional) List of occurrences of days within a month. Only valid for `Month`. The `monthly_occurrence` block supports fields documented below.
+
+---
+
+The `monthly_occurrence` block supports:
+
+* `day` - (Required) Day of the occurrence. Must be one of Monday, Tuesday, Wednesday, Thursday, Friday, Saturday, Sunday.
```
katbyte: Could we quote these values as such: `Monday`

TFaga:
Hi @katbyte 

Thanks for the quick review. I've made and pushed the changes in the comments.

katbyte:
Tests pass:
screen shot 2018-08-13 at 12 54 04

PR #1627, opened 2018-07-22
New Resources: `azurerm_firewall` & `azurerm_firewall_network_rule_collection`

hbuckle:
I'm opening this to get some feedback on the design - so far I've split network rules into a separate resource from the firewall to make them easier to manage.
This does have a couple of drawbacks, first it makes the code more complicated because there isn't a separate client for rules, everything has to be done through the AzureFirewalls client. Second it's slow, because all the rules have to be added or updated individually and one at a time.
The alternative is to implement the whole thing as one resource including rules as a property, but this could be harder to manage from the user's perspective.
Thoughts?


azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,355 @@
...
+				Required: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(network.AzureFirewallRCActionTypeAllow),
+					string(network.AzureFirewallRCActionTypeDeny),
+				}, true),
+				StateFunc:        ignoreCaseStateFunc,
+				DiffSuppressFunc: suppress.CaseDifference,
+			},
+
+			"rule": {
```
tombuildsstuff: 🤔 would it make sense for Rules to be split out into their own sub-resource? Looking at the API Docs and [the SDK](https://github.com/Azure/azure-sdk-for-go/blob/v18.0.0/services/network/mgmt/2018-04-01/network/azurefirewalls.go#L47) I can't see any validation/limitation for this?

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,355 @@
...
+	client := meta.(*ArmClient).azureFirewallsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	resourceGroup := d.Get("resource_group_name").(string)
+	name := d.Get("name").(string)
+	fwName := d.Get("azure_firewall_name").(string)
+
+	firewall, err := client.Get(ctx, resourceGroup, fwName)
+	if err != nil {
+		return fmt.Errorf("Error retrieving Azure Firewall %q (Resource Group %q): %+v", name, resourceGroup, err)
```
tombuildsstuff: if the firewall returns a 404 (there's a helper you can use) we should remove this from the state, which would allow a re-creation here (whilst this seems odd, this is a nested resource so it's a little different) e.g.

```
if err != nil {
  if utils.ResponseWasNotFound(firewall.Response()) {
    d.SetID("")
    return nil
  }
}
```

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,355 @@
...
+	firewall.AzureFirewallPropertiesFormat.IPConfigurations = &ipConfigs
+
+	future, err := client.CreateOrUpdate(ctx, resourceGroup, firewallName, firewall)
+	if err != nil {
+		return fmt.Errorf("Error creating/updating Azure Firewall %q (Resource Group %q): %+v", firewallName, resourceGroup, err)
+	}
+
+	err = future.WaitForCompletion(ctx, client.Client)
+	if err != nil {
+		return fmt.Errorf("Error waiting for creation/update of Azure Firewall %q (Resource Group %q): %+v", firewallName, resourceGroup, err)
```
tombuildsstuff: _minor_ `creation/update` -> `deletion`

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+	updatedCollections := removeArmAzureFirewallNetworkRuleCollectionByName(&firewall, name)
+	firewall.AzureFirewallPropertiesFormat.NetworkRuleCollections = updatedCollections
+
+	ipConfigs := fixArmAzureFirewallIPConfiguration(&firewall)
+	firewall.AzureFirewallPropertiesFormat.IPConfigurations = &ipConfigs
+
+	future, err := client.CreateOrUpdate(ctx, resourceGroup, firewallName, firewall)
+	if err != nil {
+		return fmt.Errorf("Error creating/updating Azure Firewall %q (Resource Group %q): %+v", firewallName, resourceGroup, err)
```
tombuildsstuff: _minor_ `creation/update` -> `deleting`

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,355 @@
...
+
+	d.Set("name", collection.Name)
+	d.Set("action", string(collection.AzureFirewallNetworkRuleCollectionPropertiesFormat.Action.Type))
+	d.Set("priority", collection.AzureFirewallNetworkRuleCollectionPropertiesFormat.Priority)
+	if rules := collection.AzureFirewallNetworkRuleCollectionPropertiesFormat.Rules; rules != nil {
+		flattenedRules := flattenArmAzureFirewallNetworkRules(rules)
+		if err := d.Set("rule", flattenedRules); err != nil {
+			return fmt.Errorf("Error flattening `rules`: %+v", err)
+		}
+	}
```
tombuildsstuff: (as above) can we set this in all cases / remove the crashes here?

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,355 @@
...
+	return nil, false
+}
+
+func removeArmAzureFirewallNetworkRuleCollectionByName(firewall *network.AzureFirewall, name string) *[]network.AzureFirewallNetworkRuleCollection {
+	var index int
+	collections := *firewall.AzureFirewallPropertiesFormat.NetworkRuleCollections
+	for i, collection := range collections {
+		if collection.Name != nil && *collection.Name == name {
+			index = i
```
tombuildsstuff: I think it may also be possible to call `delete(collections, collection)` here, since we're breaking anyway?

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,355 @@
...
+	rules := make([]network.AzureFirewallNetworkRule, 0)
+
+	for _, nwRule := range nwRules {
+		rule := nwRule.(map[string]interface{})
+
+		name := rule["name"].(string)
+		ruleToAdd := network.AzureFirewallNetworkRule{
+			Name: &name,
+		}
+		if description := rule["description"].(string); description != "" {
```
tombuildsstuff: I think we're safe to assign this regardless (since if the key's nil in TF, we should reset the description?)

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,355 @@
...
+		if r, ok := rule["protocols"].(*schema.Set); ok && r.Len() > 0 {
+			var protocols []network.AzureFirewallNetworkRuleProtocol
+			for _, v := range r.List() {
+				s := network.AzureFirewallNetworkRuleProtocol(v.(string))
+				protocols = append(protocols, s)
+			}
+			ruleToAdd.Protocols = &protocols
+		}
+		rules = append(rules, ruleToAdd)
+	}
```
tombuildsstuff: (same for all of these) if the lists are nil can we set them anyway, since this'll reset the values to an empty list as needed

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,355 @@
...
+	return rules
+}
+
+func flattenArmAzureFirewallNetworkRules(rules *[]network.AzureFirewallNetworkRule) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0)
+
+	if rules != nil {
+		for _, rule := range *rules {
+			fwRule := make(map[string]interface{})
+			fwRule["name"] = *rule.Name
```
tombuildsstuff: can we nil-check this?

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,355 @@
...
+		rules = append(rules, ruleToAdd)
+	}
+
+	return rules
+}
+
+func flattenArmAzureFirewallNetworkRules(rules *[]network.AzureFirewallNetworkRule) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0)
+
+	if rules != nil {
```
tombuildsstuff: _minor_ we could probably invert this to make it more readable, but it's not a blocker by any means :)

hbuckle:
Thanks for the review so far - I'll carry on and add the application rules as a third resource.
I did originally want to make rule collections and rules separate resources, unfortunately the API doesn't allow rule collections to be created without rules in them, so I don't think it will be possible.

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,342 @@
...
+		return fmt.Errorf("Cannot read Azure Firewall %q (Resource Group %q) ID", firewallName, resourceGroup)
+	}
+
+	var collectionID string
+	for _, collection := range *read.AzureFirewallPropertiesFormat.NetworkRuleCollections {
+		if *collection.Name == name {
+			collectionID = *collection.ID
+		}
+	}
+	d.SetId(collectionID)
```
tombuildsstuff: so the ID needs to contain all the components required to import it, as such we'll need to be a little creative here. In general for nested resources we follow the pattern `{parentId}/somename/{name}` - so in this case I'd suggest we use `fmt.Sprintf("%s/networkRuleCollection/%s", *read.ID, *collection.Name)` which will allow us to parse it out below as `id.Path["networkRuleCollection"]`

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,342 @@
...
+	return resourceArmAzureFirewallNetworkRuleCollectionRead(d, meta)
+}
+
+func resourceArmAzureFirewallNetworkRuleCollectionRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).azureFirewallsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	resourceGroup := d.Get("resource_group_name").(string)
+	name := d.Get("name").(string)
+	fwName := d.Get("azure_firewall_name").(string)
```
tombuildsstuff: we need to parse these values from the ID here, so once the change to the ID is made above, can we make this:

```
if err != nil {
  return err
}

name := id.Path["networkRuleCollection"]
firewallName := id.Path["azureFirewalls"]
```

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,342 @@
...
+
+	err = future.WaitForCompletion(ctx, client.Client)
+	if err != nil {
+		return fmt.Errorf("Error waiting for creation/update of Azure Firewall %q (Resource Group %q): %+v", firewallName, resourceGroup, err)
+	}
+
+	return nil
+}
+
+func expandArmAzureFirewallNetworkRuleCollection(d *schema.ResourceData) network.AzureFirewallNetworkRuleCollection {
```
tombuildsstuff: can we in-line this in the Create function - we tend to only split out the nested items

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,342 @@
...
+		Rules:    &rules,
+	}
+	col := network.AzureFirewallNetworkRuleCollection{
+		Name: &name,
+		AzureFirewallNetworkRuleCollectionPropertiesFormat: &properties,
+	}
+	return col
+}
+
+func findArmAzureFirewallNetworkRuleCollectionByName(firewall *network.AzureFirewall, name string) (*network.AzureFirewallNetworkRuleCollection, int, bool) {
```
tombuildsstuff: we should be able to make the int a pointer here, and return nil when it doesn't exist?

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,342 @@
...
+	for i, collection := range *firewall.AzureFirewallPropertiesFormat.NetworkRuleCollections {
```
tombuildsstuff: there's a few potential crashes here, can we make this:

```
if firewall == nil || firewall.AzureFirewallPropertiesFormat == nil || firewall.AzureFirewallPropertiesFormat.NetworkRuleCollections == nil {
  return nil, nil, false
}
```

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,342 @@
...
+		if collection.Name != nil && *collection.Name == name {
+			return &collection, i, true
+		}
+	}
+	return nil, -1, false
+}
+
+func removeArmAzureFirewallNetworkRuleCollectionByName(firewall *network.AzureFirewall, name string) *[]network.AzureFirewallNetworkRuleCollection {
+	collections := *firewall.AzureFirewallPropertiesFormat.NetworkRuleCollections
```
tombuildsstuff: (as above) can we add some nil-checks here?

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,342 @@
...
+	for i, collection := range collections {
+		if collection.Name != nil && *collection.Name == name {
+			collections = append(collections[:i], collections[i+1:]...)
+			continue
+		}
+	}
```
tombuildsstuff: can we update this to be:

```
output := make([]network.AzureFirewallNetworkRuleCollection, 0)

for _, collection := range collections {
  if collection.Name != nil && *collection.Name != name {
    output = append(output, collection)
  }
}

return &output
```

primarily as this'll make the code more readable

azurerm/resource_arm_azure_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,342 @@
...
+		destinationAddresses := rule["destination_addresses"].(*schema.Set)
+		destinationPorts := rule["destination_ports"].(*schema.Set)
+		protocols := rule["protocols"].(*schema.Set)
+		ruleToAdd := network.AzureFirewallNetworkRule{
+			Name: &name,
+		}
+		ruleToAdd.Description = &description
+		ruleToAdd.SourceAddresses = expandArmAzureFirewallSet(sourceAddresses)
+		ruleToAdd.DestinationAddresses = expandArmAzureFirewallSet(destinationAddresses)
+		ruleToAdd.DestinationPorts = expandArmAzureFirewallSet(destinationPorts)
```
tombuildsstuff: _minor_ can we in-line these? (also we can use `utils.String(val)` rather than `&` for consistency with the other resources) e.g.

```
ruleToAdd := network.AzureFirewallNetworkRule{
  Name: &name,
  Description: utils.String(description),
  SourceAddresses: expandArmAzureFirewallSet(sourceAddresses),
  DestinationAddresses: expandArmAzureFirewallSet(destinationAddresses),
  DestinationPorts: expandArmAzureFirewallSet(destinationPorts),
}
```

website/docs/r/azure_firewall.html.markdown
```diff
@@ -0,0 +1,76 @@
+---
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_azure_firewall"
+sidebar_current: "docs-azurerm-resource-azurefirewall-x"
+description: |-
+  Create an Azure Firewall.
```
tombuildsstuff: _minor_ can we make this `Manages an Azure Firewall`

website/docs/r/azure_firewall.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+---
+
+# azurerm_azure_firewall
+
+Create an Azure Firewall.
+
+~> **NOTE** This resource is currently in public preview.
```
tombuildsstuff: is this in invite-only preview, or can customers opt-in?

website/docs/r/azure_firewall.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+* `resource_group_name` - (Required) The name of the resource group in which to create the resource.
+* `location` - (Required) Specifies the supported Azure location where the resource exists.
+* `ip_configuration` - (Required) An ip configuration block as documented below.
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+`ip_configuration` supports the following:
+
+* `name` - (Required) Specifies the name of the ip configuration.
+* `subnet_id` - (Required) Reference to the subnet associated with the ip configuration.
+* `internal_public_ip_address_id` - (Required) Reference to the public IP address associated with the firewall.
```
tombuildsstuff: can we add the above note below this resource?

```
~> **NOTE** The public IP must have a `Static` allocation and `Standard` sku
```

website/docs/r/azure_firewall.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+* `name` - (Required) Specifies the name of the firewall.
+* `resource_group_name` - (Required) The name of the resource group in which to create the resource.
+* `location` - (Required) Specifies the supported Azure location where the resource exists.
+* `ip_configuration` - (Required) An ip configuration block as documented below.
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+`ip_configuration` supports the following:
+
+* `name` - (Required) Specifies the name of the ip configuration.
+* `subnet_id` - (Required) Reference to the subnet associated with the ip configuration.
```
tombuildsstuff: can we add the above note below this resource?

```
~> **NOTE** The firewall subnet must be called `AzureFirewallSubnet` and the subnet mask must be at least `/25`
```

azurerm/resource_arm_azure_firewall.go
```diff
@@ -0,0 +1,272 @@
...
+						"private_ip_address": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"subnet_id": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: azure.ValidateResourceID,
+						},
+						"internal_public_ip_address_id": {
```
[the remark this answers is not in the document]
hbuckle: The API has both publicIpAddress and internalPublicIpAddress - I've asked a question about the difference - https://github.com/Azure/azure-rest-api-specs/issues/3474

azurerm/provider.go
```diff
@@ -138,6 +138,8 @@ func Provider() terraform.ResourceProvider {
 			"azurerm_automation_schedule":                     resourceArmAutomationSchedule(),
 			"azurerm_autoscale_setting":                       resourceArmAutoScaleSetting(),
 			"azurerm_availability_set":                        resourceArmAvailabilitySet(),
+			"azurerm_azure_firewall":                          resourceArmAzureFirewall(),
+			"azurerm_azure_firewall_network_rule_collection":  resourceArmAzureFirewallNetworkRuleCollection(),
```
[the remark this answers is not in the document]
hbuckle: Although it looks a little unwieldy Azure Firewall is the name of the product, so the naming is consistent. Lots of things have firewalls in Azure so I think this makes it clear exactly which product is being managed.

hbuckle:
@tombuildsstuff Thanks, think I have fixed everything up. I'll review all the tests today then look at a separate PR to add application rule collections

azurerm/provider.go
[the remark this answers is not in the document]
tombuildsstuff: that's a fair point - but in general we ditch the word `Azure` from the front (e.g. Azure Kubernetes Service is `azurerm_kubernetes_cluster` / Azure Managed Database for MySQL becomes `azurerm_mysql_server`); so I think we should ditch it to be consistent with the other resources (the sole exception to this is `azurerm_azuread`, because that's how it's commonly referred too).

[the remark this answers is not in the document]
hbuckle: @tombuildsstuff OK, i've renamed everything to azurerm_firewall

tombuildsstuff:
@hbuckle hey sorry for the delayed re-review here - I'll take another look through this today

tombuildsstuff:
@hbuckle I thought I'd posted this message yesterday - apologies.

I ended up rebasing this yesterday and taking a look into what's needed to get this merged - and I've pushed a commit with some changes to finish this off (I hope you don't mind!) - in particular I've updated the Read and Delete functions to parse the ID of the Network Rule Collection rather than reading from the config (since this isn't always present during Import/Delete's) and updated some of the tests.

Since I've pushed some larger changes here (since this is a couple of resources) - I'm going to as @katbyte to take a look into this - but I'm going to kick off the tests now.

Thanks!

azurerm/import_arm_azure_firewall_test.go
```diff
@@ -0,0 +1,30 @@
+package azurerm
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccAzureRMFirewall_importBasic(t *testing.T) {
```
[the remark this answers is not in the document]
tombuildsstuff: 👍

azurerm/resource_arm_firewall.go
```diff
@@ -0,0 +1,316 @@
...
+			"ip_configuration": {
+				Type:     schema.TypeList,
+				Required: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
```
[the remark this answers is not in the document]
tombuildsstuff: 👍

azurerm/resource_arm_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,433 @@
...
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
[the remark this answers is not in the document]
tombuildsstuff: 👍

azurerm/resource_arm_firewall.go
```diff
@@ -0,0 +1,316 @@
...
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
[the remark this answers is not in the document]
tombuildsstuff: 👍

azurerm/resource_arm_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,433 @@
...
+
+			"azure_firewall_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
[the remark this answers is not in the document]
tombuildsstuff: 👍

hbuckle:
@tombuildsstuff Thanks for the update, I'm sure my go skills still leave something to be desired. I'll have another look through this today. I've started the application rule collection on another branch as well.
The publicipaddress vs internalpublicipaddress has been resolved in the [specs](https://github.com/Azure/azure-rest-api-specs/pull/3743) as well so we should be able to remove the workaround for that once it gets into the SDK

hbuckle:
@tombuildsstuff Thanks for the update, I'm sure my go skills still leave something to be desired. I'll have another look through this today. I've started the application rule collection on another branch as well.
The publicipaddress vs internalpublicipaddress has been resolved in the [specs](https://github.com/Azure/azure-rest-api-specs/pull/3743) as well so we should be able to remove the workaround for that once it gets into the SDK

tombuildsstuff:
> @tombuildsstuff Thanks for the update, I'm sure my go skills still leave something to be desired. I'll have another look through this today.

No worries, not at all - apologies this sat for a while, we're still trying to work through the open PR's.

> I've started the application rule collection on another branch as well.

That's awesome 👍 - thanks! 

> The publicipaddress vs internalpublicipaddress has been resolved in the specs as well so we should be able to remove the workaround for that once it gets into the SDK

Cool, that's good to know - I'll keep an eye on the SDK release and add a TODO for that

Thanks!

azurerm/resource_arm_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,433 @@
...
+			networkRules = append(networkRules, rule)
+		}
+	}
+	firewall.AzureFirewallPropertiesFormat.NetworkRuleCollections = &networkRules
+
+	ipConfigs, err := azure.FixFirewallIPConfiguration(props.IPConfigurations)
+	if err != nil {
+		return fmt.Errorf("Error fixing IP Configuration for Firewall %q (Resource Group %q): %+v", firewallName, resourceGroup, err)
+	}
+	firewall.AzureFirewallPropertiesFormat.IPConfigurations = ipConfigs
```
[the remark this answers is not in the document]
tombuildsstuff: Agreed since we're not dereferencing the pointer 👍

azurerm/resource_arm_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,433 @@
...
+	for _, rule := range *props.NetworkRuleCollections {
+		if rule.Name == nil {
+			continue
+		}
+
+		if *rule.Name != name {
+			networkRules = append(networkRules, rule)
+		}
+	}
+	firewall.AzureFirewallPropertiesFormat.NetworkRuleCollections = &networkRules
```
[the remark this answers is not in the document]
tombuildsstuff: Agreed since we're not dereferencing the pointer 👍

azurerm/resource_arm_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,433 @@
...
+		if index == -1 {
+			return fmt.Errorf("Error locating Network Rule Collection %q (Firewall %q / Resource Group %q)", name, firewallName, resourceGroup)
+		}
+
+		rules[index] = rule
+	} else {
+		rules = append(rules, rule)
+	}
+
+	firewall.AzureFirewallPropertiesFormat.NetworkRuleCollections = &rules
```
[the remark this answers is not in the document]
tombuildsstuff: nope - since the pointer's de-referenced above this'd be a new (non-nilable) object

azurerm/resource_arm_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,433 @@
...
+
+	if firewall.AzureFirewallPropertiesFormat == nil {
+		return fmt.Errorf("Error expanding Firewall %q (Resource Group %q): `properties` was nil.", firewallName, resourceGroup)
+	}
+	props := *firewall.AzureFirewallPropertiesFormat
+
+	if props.NetworkRuleCollections == nil {
+		return fmt.Errorf("Error expanding Firewall %q (Resource Group %q): `properties.NetworkRuleCollections` was nil.", firewallName, resourceGroup)
+	}
+	rules := *props.NetworkRuleCollections
```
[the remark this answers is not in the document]
tombuildsstuff: 👍

azurerm/resource_arm_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,433 @@
...
+
+	ipConfigurations, err := azure.FixFirewallIPConfiguration(props.IPConfigurations)
+	if err != nil {
+		return fmt.Errorf("Error fixing IP Configurations for Firewall %q (Resource Group %q): %+v", firewallName, resourceGroup, err)
+	}
+	firewall.AzureFirewallPropertiesFormat.IPConfigurations = ipConfigurations
```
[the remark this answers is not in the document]
tombuildsstuff: nope - since the pointer's de-referenced above this'd be a new (non-nilable) object

azurerm/resource_arm_firewall_network_rule_collection.go
```diff
@@ -0,0 +1,433 @@
...
+	}
+	props := *read.AzureFirewallPropertiesFormat
+
+	if props.NetworkRuleCollections == nil {
+		return fmt.Errorf("Error retrieving Network Rule Collection %q (Firewall %q / Resource Group %q): `props.NetworkRuleCollections` was nil", name, firewallName, resourceGroup)
+	}
+
+	rules := *props.NetworkRuleCollections
+	var rule *network.AzureFirewallNetworkRuleCollection
+	for _, r := range rules {
```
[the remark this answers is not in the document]
tombuildsstuff: 👍

tombuildsstuff:
Tests pass:

screenshot 2018-09-13 at 22 46 34

Issue #1484, opened 2018-07-03
New-Feature-Request SQL Threat Detection Policy


whytoe:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Enable the SQL Threat Detection feature when deploying SQL Servers in Platform as a Service

### New or Affected Resource(s)

* azurerm_sql_server

### Potential Terraform Configuration

sql_threat_detection : "true"

### References

The Azure API documentation
https://docs.microsoft.com/en-us/rest/api/sql/databasethreatdetectionpolicies/createorupdate

Go SDK does not appear to support it ( i couldnt find it)

* #0000

tombuildsstuff:
hey @whytoe 

Thanks for opening this issue :)

Taking a quick look at this / some notes for whoever picks this up: this appears to be available in the Go SDK here: https://github.com/Azure/azure-sdk-for-go/blob/v17.4.0/services/sql/mgmt/2014-04-01/sql/databasethreatdetectionpolicies.go#L52 - and this would probably make sense nested under the `azurerm_sql_server` resource rather than as an independent resource.

Thanks!

Lucretius:
@tombuildsstuff just going through the issue backlog, looks like this one got resolved a while back and can be closed.

andreas-schilling:
It's available on DB-Level, but not on server-level (which is recommended by Microsoft)

SirWayne:
Is there a time plan for this feature?

techjacker:
Is there any plan to add this for Azure storage?

The go SDK doesn't seem to support it currently.

tombuildsstuff:
@techjacker I may be wrong, but that appears to be a separate (unrelated) service to the one in this issue? If so, would you mind opening a separate feature request for that? Thanks!

techjacker:
@tombuildsstuff Sure I'll do this and add the link back to this thread when I'm done.

techjacker:
> @tombuildsstuff Sure I'll do this and add the link back to this thread when I'm done.

Created feature request here: https://github.com/terraform-providers/terraform-provider-azurerm/issues/3670

[...]

PR #1628, opened 2018-07-23
azurerm_sql_database - Add sql threat detection policy configuration

TFaga:
This pull request adds the ability to configure sql threat detection policy for an `azurerm_sql_database`. Addresses issue #1484


website/docs/r/sql_database.html.markdown
```diff
@@ -85,6 +87,17 @@ The following arguments are supported:
 * `authentication_type` - (Required) Specifies the type of authentication used to access the server. Valid values are `SQL` or `ADPassword`.
 * `operation_mode` - (Optional) Specifies the type of import operation being performed. The only allowable value is `Import`.
 
+`threat_detection_policy` supports  the following:
+
+* `state` - (Required) Specifies the state of the policy. If state is Enabled, storageEndpoint and storageAccountAccessKey are required.
+* `disabled_alerts` - (Optional) Specifies the semicolon-separated list of alerts that are disabled, or empty string to disable no alerts. Possible values: Sql_Injection; Sql_Injection_Vulnerability; Access_Anomaly; Usage_Anomaly.
+* `email_account_admins` - (Optional) Specifies that the alert is sent to the account administrators.
+* `email_addresses` - (Optional) Specifies the semicolon-separated list of e-mail addresses to which the alert is sent.
```
tombuildsstuff: Once we've updated this to a list (above) we should be able to make this:

```
A list of email addresses which alerts should be sent to.
```

website/docs/r/sql_database.html.markdown
```diff
@@ -85,6 +87,17 @@ The following arguments are supported:
...
+* `retention_days` - (Optional) Specifies the number of days to keep in the Threat Detection audit logs.
+* `storage_account_access_key` - (Optional) Specifies the identifier key of the Threat Detection audit storage account. If state is Enabled, storageAccountAccessKey is required.
+* `storage_endpoint` - (Optional) Specifies the blob storage endpoint (e.g. https://MyAccount.blob.core.windows.net). This blob storage will hold all Threat Detection audit logs. If state is Enabled, storageEndpoint is required.
+* `use_server_default` - (Optional) Specifies whether to use the default server policy.
```
tombuildsstuff: `Specifies whether to use the default server policy.` ->
```
Should the default server policy be used? Defaults to `true`
```

azurerm/resource_arm_sql_database_test.go
```diff
@@ -215,6 +215,44 @@ func TestAccAzureRMSqlDatabase_requestedServiceObjectiveName(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMSqlDatabase_threatDetectionPolicy(t *testing.T) {
+	resourceName := "azurerm_sql_database.test"
+	ri := acctest.RandInt()
+	location := testLocation()
+	preConfig := testAccAzureRMSqlDatabase_threatDetectionPolicy(ri, location, "Enabled", "Sql_Injection", "Enabled", "", "Disabled", 15)
+	postConfig := testAccAzureRMSqlDatabase_threatDetectionPolicy(ri, location, "Enabled", "Sql_Injection", "Enabled", "", "Disabled", 15)
```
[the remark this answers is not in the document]
tombuildsstuff: could we also add a new test to check removing a threat detection policy, it should be possible to test this by using the `testAccAzureRMSqlDatabase_basic` config to ensure it's removed in a secondary step (as is done in this test)

azurerm/resource_arm_sql_database.go
```diff
@@ -196,8 +197,98 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+						"storage_endpoint": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+
+						"use_server_default": {
+							Type:             schema.TypeString,
+							Optional:         true,
+							DiffSuppressFunc: suppress.CaseDifference,
+							Default:          string(sql.SecurityAlertPolicyUseServerDefaultDisabled),
```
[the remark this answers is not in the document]
TFaga: The reason I left it `Disabled` is because that's the default value returned from the API if no policy is set.

Default values from the API:

```js
{
    "properties": {
        "useServerDefault": "Disabled",
        "state": "Disabled",
        "disabledAlerts": "",
        "emailAddresses": "",
        "emailAccountAdmins": "Disabled",
        "storageEndpoint": "",
        "storageAccountAccessKey": "",
        "retentionDays": 0
    }
}
```

TFaga:
@tombuildsstuff Hey, thank you for the quick review. I've pushed a commit to address the changes.

website/docs/r/sql_database.html.markdown
```diff
@@ -85,6 +87,19 @@ The following arguments are supported:
...
+---
+
+`threat_detection_policy` supports the following:
+
+* `state` - (Required) The State of the Policy. Possible values are `Enabled`, `Disabled` or `New`.
+* `disabled_alerts` - (Optional) Specifies a list of alerts which should be disabled. Possible values include `Access_Anomaly`, `Sql_Injection` and `Sql_Injection_Vulnerability`.
+* `email_account_admins` - (Optional) Should the account administrators be emailed when this alert is triggered?
+* `email_addresses` - (Optional) A list of email addresses which alerts should be sent to.
+* `retention_days` - (Optional) Specifies the number of days to keep in the Threat Detection audit logs.
+* `storage_account_access_key` - (Optional) Specifies the identifier key of the Threat Detection audit storage account. If state is Enabled, storageAccountAccessKey is required.
```
katbyte: Echoing toms comment, could 

> If state is Enabled, storageAccountAccessKey is required. 

become: 

> Required if `state` is `enabled`.

azurerm/resource_arm_sql_database_test.go
```diff
@@ -215,6 +215,40 @@ func TestAccAzureRMSqlDatabase_requestedServiceObjectiveName(t *testing.T) {
...
+					testCheckAzureRMSqlDatabaseExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "threat_detection_policy.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "threat_detection_policy.0.state", "Enabled"),
+					resource.TestCheckResourceAttr(resourceName, "threat_detection_policy.0.retention_days", "15"),
+					resource.TestCheckResourceAttr(resourceName, "threat_detection_policy.0.disabled_alerts.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "threat_detection_policy.0.email_account_admins", "Enabled"),
+				),
+			},
+			{
+				Config: postConfig,
```
katbyte: Could we get an import test step to verify that import works?

```go
            {
				ImportState:       true,
				ImportStateVerify: true,
			},
```

TFaga:
Hi @katbyte ,

Thanks for the review. Apologies for the late response. I've pushed the changes in the comments.

azurerm/resource_arm_sql_database.go
```diff
@@ -456,3 +623,56 @@ func expandAzureRmSqlDatabaseImport(d *schema.ResourceData) sql.ImportExtensionR
...
+		threadDetection := tdl[0].(map[string]interface{})
+
+		properties.State = sql.SecurityAlertPolicyState(threadDetection["state"].(string))
+		properties.EmailAccountAdmins = sql.SecurityAlertPolicyEmailAccountAdmins(threadDetection["email_account_admins"].(string))
+		properties.UseServerDefault = sql.SecurityAlertPolicyUseServerDefault(threadDetection["use_server_default"].(string))
+
+		if v, ok := threadDetection["disabled_alerts"]; ok {
+			alerts := v.(*schema.Set).List()
+			expandedAlerts := make([]string, len(alerts))
+			for i := range alerts {
```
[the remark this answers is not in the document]
TFaga: I've updated to use the value from the current element of the loop, though still needed the index   for the `expandedAlerts` variable.

azurerm/resource_arm_sql_database.go
```diff
@@ -354,6 +468,19 @@ func resourceArmSqlDatabaseRead(d *schema.ResourceData, meta interface{}) error
 		return fmt.Errorf("Error making Read request on Sql Database %s: %+v", name, err)
 	}
 
+	oldThreatDetection, hasOldThreatDetection := d.GetOk("threat_detection_policy")
+	if hasOldThreatDetection {
```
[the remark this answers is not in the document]
TFaga: The problem was that the thread detection problem is a separate API call which always returns the default values, causing every `plan` operation to be the following every time, if the `threat_detection_policy` was not set in the manifest as it filled out the list in the state with all default values:

```
  ~ azurerm_sql_database.test
      threat_detection_policy.#: "1" => "0"
```

I've changed this so that the list element is marked `computed`, which alleviates the problem. Though I'm not sure if this is ok or if there is a better option to handle this case.

azurerm/resource_arm_sql_database.go
```diff
@@ -196,8 +197,111 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+								string(sql.SecurityAlertPolicyStateDisabled),
+								string(sql.SecurityAlertPolicyStateEnabled),
+								string(sql.SecurityAlertPolicyStateNew),
+							}, true),
+						},
+
+						"storage_account_access_key": {
+							Type:      schema.TypeString,
+							Optional:  true,
+							Sensitive: true,
```
jeffreyCline: Can we add some validation here to ensure something is entered?
```
ValidateFunc: validation.NoZeroValues,
```

azurerm/resource_arm_sql_database.go
```diff
@@ -196,8 +197,111 @@ func resourceArmSqlDatabase() *schema.Resource {
...
+						},
+
+						"storage_endpoint": {
+							Type:     schema.TypeString,
+							Optional: true,
```
jeffreyCline: Same here with the validation.

TFaga:
Hi @jeffreyCline, 

I've pushed the changes in the comments.

PR #1636, opened 2018-07-24
r/application_gateway: clarify default value of require_sni in an http_listener

Supermathie:
Is there any reason we can't just make the default for `require_sni` to be `true` when the listener is HTTPS?


tombuildsstuff:
@Supermathie 

Thanks for this PR :)

> Is there any reason we can't just make the default for require_sni to be true when the listener is HTTPS?

In general we either have a default value or make it required, rather than changing the value dynamically, but possibly in the future - unfortunately either way we're currently blocked on any enhancements to Application Gateway because of https://github.com/terraform-providers/terraform-provider-azurerm/issues/1576.

In the interim we're planning on rewriting the documentation for the Application Gateway resource in the near future to bring it into line with the other resources too, so that we can make this clearer.

Thanks!

PR #1641, opened 2018-07-24
Insert attribute reference sku

mvtiengo:
To use the standard LB it is necessary to pass the sku as an argument in azurerm_public_ip.

Example:
resource "azurerm_public_ip" "this" {
  name = "${var.prefix}-${terraform.workspace}-${var.module}-pip"
  resource_group_name = "${var.network_basic}"
  location = "${data.azurerm_resource_group.this.location}"
  resource_group_name = "${data.azurerm_resource_group.this.name}"
  public_ip_address_allocation = "static"
  sku = "standard"
}

### DIFF
--- a/website/docs/d/public_ip.html.markdown
+++ b/website/docs/d/public_ip.html.markdown
@@ -99,6 +99,7 @@ output "public_ip_address" {
 
 * `name` - (Required) Specifies the name of the public IP address.
 * `resource_group_name` - (Required) Specifies the name of the resource group.
+* `sku` - (Optional) The SKU of the Azure Public IP. Accepted values are Basic and Standard. Defaults to Basic.
 
 
 ## Attributes Reference


tombuildsstuff:
hi @mvtiengo 

Thanks for this PR - taking a quick look at this it appears this PR is for the Data Source, which obtains information about the Public IP - whereas the code sample in the description references the Public IP Resource (which is used to Provision this resource - [which already has this property listed](https://www.terraform.io/docs/providers/azurerm/r/public_ip.html#sku)). Out of interest - which of the two Public IP's was this intended for? :)

Taking a look at the code I can see this field is missing from the Data Source (insofar as it's not exposed in the code, so the docs appear to be right at this point) - but we should look to fix that so it's exposed for the Data Source too :)

Thanks!

Issue #1635, opened 2018-07-23
azurerm_azuread_service_principal delay before being usable


mb290:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

terraform -v
Terraform v0.11.7
+ provider.azurerm v1.10.0

### Affected Resource(s)

* terraform-provider-azurerm_v1.10.0_x4

### Terraform Configuration Files

```hcl
provider "azurerm" {
    version = "1.10.0"
 }

resource "azurerm_azuread_application" "test" {
  name                       = "exampleTFapplication"
  available_to_other_tenants = false
  oauth2_allow_implicit_flow = false
}

resource "azurerm_azuread_service_principal" "test" {
  application_id = "${azurerm_azuread_application.test.application_id}"
}

resource "azurerm_azuread_service_principal_password" "test" {
  service_principal_id = "${azurerm_azuread_service_principal.test.id}"
  value                = "BVcKK237/&&)hyz@%nsadasdsa(*&^CC#Nd3"
  end_date             = "2020-01-01T01:02:03Z"
}

resource "azurerm_resource_group" "test" {
  name     = "testResourceGroup1"
  location = "West US"
}

resource "azurerm_role_assignment" "test" {
    depends_on = ["azurerm_azuread_service_principal.test"]
  scope                = "${azurerm_resource_group.test.id}"
  role_definition_name = "Reader"
  principal_id         = "${azurerm_azuread_service_principal.test.id}"
}
```

### Panic Output

Error: Error applying plan:

1 error(s) occurred:

* azurerm_role_assignment.test: 1 error(s) occurred:

* azurerm_role_assignment.test: authorization.RoleAssignmentsClient#Create: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="PrincipalNotFound" Message="Principal ######################## does not exist in the directory #######-#####-######-#########."

(sensitive details have been hashed out).

### Expected Behavior

[...]

PR #1644, opened 2018-07-25
retry role assignment creation

joakimhellum-in:
#1635 
Would like to start discussion how to possibly implement workaround for this issue.

### DIFF
--- a/azurerm/resource_arm_role_assignment.go
+++ b/azurerm/resource_arm_role_assignment.go
@@ -4,9 +4,11 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-01-01-preview/authorization"
 	"github.com/hashicorp/go-uuid"
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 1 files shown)


mb290:
Hey, I implemented a new statement as follows in resourceArmRoleAssignmentCreate, however it doesn't look to wait because the check passes and it gets the ID back, which I kinda expected as it's returning when the creation is down. I think the retry is going to need to be placed on the following:

	_, err = roleAssignmentsClient.Create(ctx, scope, name, properties)

Here is my hack about so far which is similar to @joakimhellum-in has committed so far:

```
	err :=  resource.Retry(120*time.Second, validatePrincipalIdExists(principalId, meta))
	if err != nil {
		return err
	}
```

```
func validatePrincipalIdExists(principalId string, meta interface{}) func() *resource.RetryError {
	
	return func() *resource.RetryError {
		servicePrincipalsClient := meta.(*ArmClient).servicePrincipalsClient
		
		_, err := servicePrincipalsClient.Get(ctx, principalId)

		if err != nil {
			return resource.RetryableError(err)
		}
		return nil
		
	}
}

```

mb290:
I submitted PR #1647 with a tested solution.

joakimhellum-in:
thanks alot. i will close this one in favor of #1647  
@mb290 @tombuildsstuff  FYI did create some hundred service principals past week and found that 2 minutes might be enough (as opposed to 3min). not sure if you got the same results.

PR #1647, opened 2018-07-25
Adding retry functionality to role assignment

mb290:
This is a fix for #1635. It adds the retry functionality to the azurerm_role_assignment resource to take into account the time it takes for Azure to replicate the Principal ID to be usable by roles. Very similar to #1644


mb290:
@tombuildsstuff thanks for the feedback. The reason I chose 180 was to mimic the functionality of the az cli which was referenced in https://github.com/Azure/azure-cli/blob/master/src/command_modules/azure-cli-role/azure/cli/command_modules/role/custom.py#L959

tombuildsstuff:
@mb290 there's no issue blocking this from a code perspective - unfortunately our acceptance test account filled up, so we've needed to clear that out before we could run the tests; I've done that this morning and we should be able to merge this once the tests pass; apologies for the delay!

PR #1650, opened 2018-07-26
Storage Share Quota and Import Support

mbfrahry:
This PR allows quota to be updated. Import is now supported. General cleanup has also been done with moving the validation functions to their respective validation and validation test files.

### DIFF
--- a/azurerm/resource_arm_storage_share.go
+++ b/azurerm/resource_arm_storage_share.go
@@ -16,7 +16,6 @@ func resourceArmStorageShare() *schema.Resource {
 		Create: resourceArmStorageShareCreate,
 		Read:   resourceArmStorageShareRead,
 		Update: resourceArmStorageShareUpdate,
-		// Exists: resourceArmStorageShareExists,
 		Delete: resourceArmStorageShareDelete,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
--- a/azurerm/resource_arm_storage_share.go
+++ b/azurerm/resource_arm_storage_share.go
@@ -3,6 +3,7 @@ package azurerm
 import (
 	"fmt"
 	"log"
+	"regexp"
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/validation"
@@ -91,6 +92,9 @@ func resourceArmStorageShareRead(d *schema.ResourceData, meta interface{}) error
 	ctx := armClient.StopContext
 
 	id := strings.Split(d.Id(), "/")
+	if len(id) != 3 {
+		return fmt.Errorf("ID was not in the expected format - expected `{name}/{resourceGroup}/{storageAccountName}` got %q", id)
+	}
 	name := id[0]
 	resourceGroupName := id[1]
 	storageAccountName := id[2]
@@ -114,6 +118,7 @@ func resourceArmStorageShareRead(d *schema.ResourceData, meta interface{}) error
 	if !exists {
 		log.Printf("[INFO] Share %q no longer exists, removing from state...", name)
 		d.SetId("")
+		return nil
 	}
 
 	url := reference.URL()
@@ -136,6 +141,9 @@ func resourceArmStorageShareUpdate(d *schema.ResourceData, meta interface{}) err
 	ctx := armClient.StopContext
 
 	id := strings.Split(d.Id(), "/")
+	if len(id) != 3 {
+		return fmt.Errorf("ID was not in the expected format - expected `{name}/{resourceGroup}/{storageAccountName}` got %q", id)
+	}
 	name := id[0]
 	resourceGroupName := id[1]
 	storageAccountName := id[2]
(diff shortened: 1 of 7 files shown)


mbfrahry:
Tests passing 
```
=== RUN   TestAccAzureRMStorageShare_basic
--- PASS: TestAccAzureRMStorageShare_basic (80.86s)
=== RUN   TestAccAzureRMStorageShare_disappears
--- PASS: TestAccAzureRMStorageShare_disappears (86.22s)
=== RUN   TestAccAzureRMStorageShare_updateQuota
--- PASS: TestAccAzureRMStorageShare_updateQuota (98.15s)
```

azurerm/resource_arm_storage_share.go
```diff
@@ -75,16 +81,18 @@ func resourceArmStorageShareCreate(d *schema.ResourceData, meta interface{}) err
...
 
 func resourceArmStorageShareRead(d *schema.ResourceData, meta interface{}) error {
 	armClient := meta.(*ArmClient)
 	ctx := armClient.StopContext
 
-	resourceGroupName := d.Get("resource_group_name").(string)
-	storageAccountName := d.Get("storage_account_name").(string)
+	id := strings.Split(d.Id(), "/")
+	name := id[0]
+	resourceGroupName := id[1]
```
tombuildsstuff: as discussed on Slack - there's a crash here until we have a state migration for the ID switch

azurerm/resource_arm_storage_share.go
```diff
@@ -96,34 +105,70 @@ func resourceArmStorageShareRead(d *schema.ResourceData, meta interface{}) error
...
 	if err != nil {
-		return err
+		return fmt.Errorf("Error testing existence of share %q: %s", name, err)
 	}
 
 	if !exists {
-		// Exists already removed this from state
-		return nil
+		log.Printf("[INFO] Share %q no longer exists, removing from state...", name)
+		d.SetId("")
```
tombuildsstuff: this wants a `return nil` here

azurerm/resource_arm_storage_share.go
```diff
@@ -96,34 +105,70 @@ func resourceArmStorageShareRead(d *schema.ResourceData, meta interface{}) error
...
+	log.Printf("[INFO] Setting share %q properties in storage account %q", name, storageAccountName)
+	reference.Properties = storage.ShareProperties{
+		Quota: d.Get("quota").(int),
+	}
+	reference.SetProperties(options)
+
+	return resourceArmStorageShareRead(d, meta)
+}
+
 func resourceArmStorageShareExists(d *schema.ResourceData, meta interface{}) (bool, error) {
```
tombuildsstuff: I think this method can be removed?

website/docs/r/storage_share.html.markdown
```diff
@@ -57,3 +57,11 @@ The following attributes are exported in addition to the arguments listed above:
...
 * `id` - The storage share Resource ID.
 * `url` - The URL of the share
+
+## Import
+
+Storage Shares can be imported using the `resource id`, e.g.
+
+```shell
+terraform import azurerm_storage_share.testShare storageShareName/resourceGroupName/storageAccoutName
+```
```
tombuildsstuff: can we add `-> **NOTE:** This identifier is unique to Terraform` below this

azurerm/validators.go
```diff
@@ -246,3 +246,26 @@ func validateFilePath() schema.SchemaValidateFunc {
 		return
 	}
 }
+
+//Following the naming convention as laid out in the docs https://msdn.microsoft.com/library/azure/dn167011.aspx
+func validateArmStorageShareName(v interface{}, k string) (ws []string, errors []error) {
```
tombuildsstuff: since this is the only place it's used, can we move this back into the share resource?

azurerm/resource_arm_storage_share_test.go
```diff
@@ -49,15 +56,51 @@ func TestAccAzureRMStorageShare_disappears(t *testing.T) {
...
+				Config: config2,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMStorageShareExists(resourceName, &sS),
+					resource.TestCheckResourceAttr(resourceName, "quota", "5"),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
```
tombuildsstuff: we can remove the Imports from this test (since they're covered by the Basic test above; in general we don't Import test updates since they're covered elsewhere)

azurerm/resource_arm_storage_share.go
```diff
@@ -96,34 +105,70 @@ func resourceArmStorageShareRead(d *schema.ResourceData, meta interface{}) error
...
+	d.Set("quota", reference.Properties.Quota)
+
 	return nil
 }
 
+func resourceArmStorageShareUpdate(d *schema.ResourceData, meta interface{}) error {
+	armClient := meta.(*ArmClient)
+	ctx := armClient.StopContext
+
+	id := strings.Split(d.Id(), "/")
```
tombuildsstuff: there's a potential crash in these few lines if the segments aren't correct - as such we  probably want to add below this:

```
if len(id) != 3 {
  return fmt.Errorf("ID was not in the expected format - expected `{name}/{resourceGroup}/{storageAccountName}` got %q")
}
```

azurerm/resource_arm_storage_share.go
mbfrahry: Done

azurerm/resource_arm_storage_share.go
mbfrahry: Done

azurerm/validators.go
mbfrahry: Done

azurerm/resource_arm_storage_share_test.go
mbfrahry: Done

azurerm/resource_arm_storage_share.go
mbfrahry: Done

PR #1678, opened 2018-07-30
[WIP] New Resource: Virtual Machine State

mb290:
Here is a new resource which allows the control of the virtual machine state through Terraform.  It allows for a VM to be running/stopped/deallocated without the need to completely destroy the resource.  Where we leverage Azure IaaS components it isn't always realistic to destroy and recreate all the time and I don't want to have to grant permissions in the portal just so users can deallocate manually when the machine isn't being used.

I still need to finish writing the test cases but wanted to get an initial commit it. I've tested it locally and the scenarios so far seem to be working for me.

(It's my first full resource so be gentle :))

### DIFF
--- a/azurerm/provider.go
+++ b/azurerm/provider.go
@@ -257,6 +257,7 @@ func Provider() terraform.ResourceProvider {
 			"azurerm_virtual_machine_data_disk_attachment":    resourceArmVirtualMachineDataDiskAttachment(),
 			"azurerm_virtual_machine_extension":               resourceArmVirtualMachineExtensions(),
 			"azurerm_virtual_machine_scale_set":               resourceArmVirtualMachineScaleSet(),
+			"azurerm_virtual_machine_state":                   resourceArmVirtualMachineState(),
 			"azurerm_virtual_network":                         resourceArmVirtualNetwork(),
 			"azurerm_virtual_network_gateway":                 resourceArmVirtualNetworkGateway(),
 			"azurerm_virtual_network_gateway_connection":      resourceArmVirtualNetworkGatewayConnection(),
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
hey @mb290 

Thanks for opening this PR :)

There's a grey line here for infrastructure tooling - where Terraform ends and another tool begins. In general managing the state of a Virtual Machine (e.g. Running/Stopped/Deallocated) is something that's managed outside of Terraform (generally using the cloud providers CLI) - rather than as a part of the native resource.

As such - at this time I'm unsure whether Terraform's the right tool to be managing the state of a Virtual Machine - and (whilst the code looks good) - unfortunately I don't think this is something we could support within Terraform at this time. It's worth noting from a design perspective I'd suggest this kind of thing would be better scoped within the VM Resource itself (e.g. as a field), rather than being a new resource in of itself, since there's a 1:1 mapping between a VM and it's running state.

Instead I'd recommend using the Azure CLI to achieve this - you could do this from within Terraform by using a `null_resource` and the local exec provisioner (and possibly turn this into a module), like so:

```
variable "virtual_machine_name" {}
variable "resource_group_name" {}

resource "null_resource" "machine_state" {
  provisioner "local-exec" {
    command = "az vm stop --name ${var.virtual_machine_name} --resource-group ${var.resource_group_name}"
  }
}
```

Whilst I appreciate this solution may be a little different to how you envisioned it - the running state of a Virtual Machine is slightly outside Terraform's bounds here. All that said - I believe this is something we'll need to do in the long term it's just how we approach this requires some more thought across multiple providers (for instance, should we start a stopped VM to be able to run a provisioner and then stop it again?).

Thanks!

mb290:
@tombuildsstuff thanks for the feedback and I'm not overly surprised by the response :). The challenge with using the null_resource is this is fine for stopping the machine, but to try and build the logic in through this type of resource to be able to start etc is really not practical. 

I also agree with your perspective when you're looking at running terraform locally or using just the binary through a CI/CD workflow, however we leverage TFE and as such are really looking to use TFE to add governance to the entire state and life-cycle of the infrastructure, of which is a machine on/off is a valid state. From an enterprise perspective where you're using TFE to manage this, having to give user's access to the portal to start/stop makes RBAC significantly more complicated. We also leverage Sentinel as part of TFE and so a could use here would be to never let a user put a machine in a state that it's stopped but not deallocated.

PR #1713, opened 2018-08-02
update VMSS docs to include ip_forwarding (fixes #1709)

### LINKED ISSUE #1709: ip_forwarding property for azurerm_virtual_machine_scale_set is not documented in terraform.io
**Terraform Version**
terraform -v
Terraform v0.11.7

provider.azurerm v1.11.0

**Affected Resource(s)**
ls

azurerm_virtual_machine_scale_set

Issue
The page https://www.terraform.io/docs/providers/azurerm/r/virtual_machine_scale_set.html does not include any documentation on the ip_forwarding property for the scale set network interface which was added here: https://github.com/terraform-providers/terraform-provider-azurerm/issues/1226

katbyte:
fixes #1709


tombuildsstuff:
@bflad 

> should we also include that it defaults to false?

Yes we should - will push a commit to fix 👍

PR #1728, opened 2018-08-06
[WIP] Documentation Refactoring

tombuildsstuff:
This PR is still a WIP, but this refactors the Documentation and Examples quite significantly.

The primary purpose of this PR is to make the examples clearer to read, since the "Example Usage" section is intended to be how to use the singular resource, rather than a complete example (and we should be pointing to the `examples` directory in the repository for other samples).

Resources with multiple examples have been moved into the `examples/` folder with a cross-link between the docs page and the repo examples folder. This PR also starts the task of grouping examples by type (e.g. App Service, VM's etc) - which is of a larger-scope than this PR, but should provide a foundation.

There's still a bunch of work needed for this:

 - [ ] Finish working through the outstanding resources
 - [ ] Re-splitting this into multiple commits to make it reviewable
 - [ ] Testing each of the Examples


metacpp:
@tombuildsstuff you need to split this PR into smaller, iterative PRs to avoid conflicts with other contributors working on the same MD files.

tombuildsstuff:
@metacpp given this is a cross-cutting PR this makes sense to do as a single PR rather than incrementally - as such splitting this out will only cause more overhead - we'll rebase this when we come back to it (once #1746 has been merged)

PR #1737, opened 2018-08-08
resource/kubernetes_cluster: Fix missing forceNew on service principal fields

lawrencegripper:
Repro using the current provider:

1. Create a cluster
2. Change the Service Principal used 
3. See error:

> * azurerm_kubernetes_cluster.aks: containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="PropertyChangeNotAllowed" Message="Changing property 'servicePrincipalProfile.clientId' is not allowed." Target="servicePrincipalProfile.clientId"

This PR moves the SP properties to ForceNew to fix this issue and force a new cluster when changes are made.

### DIFF
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -226,10 +226,12 @@ func resourceArmKubernetesCluster() *schema.Resource {
 						"client_id": {
 							Type:     schema.TypeString,
 							Required: true,
+							ForceNew: true,
 						},
 
 						"client_secret": {
 							Type:      schema.TypeString,
+							ForceNew:  true,
 							Required:  true,
 							Sensitive: true,
 						},


tombuildsstuff:
hey @lawrencegripper 

Just to let you know that we've just shipped support for this as a part of v1.13.0 of the AzureRM Provider :)

Thanks!

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -226,10 +226,12 @@ func resourceArmKubernetesCluster() *schema.Resource {
 						"client_id": {
 							Type:     schema.TypeString,
 							Required: true,
+							ForceNew: true,
 						},
 
 						"client_secret": {
 							Type:      schema.TypeString,
+							ForceNew:  true,
```
hazzik: This does not have any effect because the secret is not retrieved from the environment nor stored in a state file.

PR #1742, opened 2018-08-08
azurerm_lb_probe: backport azurestack review comments

katbyte:
backport of issues found in [this pr](https://github.com/terraform-providers/terraform-provider-azurestack/pull/23)

also explicitly adds `Https` support to `protocol` property.

### DIFF
--- a/azurerm/resource_arm_loadbalancer_probe.go
+++ b/azurerm/resource_arm_loadbalancer_probe.go
@@ -217,7 +217,9 @@ func resourceArmLoadBalancerProbeRead(d *schema.ResourceData, meta interface{}) 
 				}
 			}
 		}
-		d.Set("load_balancer_rules", load_balancer_rules)
+		if err := d.Set("load_balancer_rules", load_balancer_rules); err != nil {
+			return fmt.Errorf("Error setting `load_balancer_rules` (Load Balancer Probe %q): %+v", name, err)
+		}
 	}
 
 	return nil
--- a/azurerm/resource_arm_loadbalancer_probe.go
+++ b/azurerm/resource_arm_loadbalancer_probe.go
@@ -9,14 +9,18 @@ import (
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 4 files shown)


azurerm/resource_arm_loadbalancer_probe.go
```diff
@@ -189,19 +201,24 @@ func resourceArmLoadBalancerProbeRead(d *schema.ResourceData, meta interface{})
...
+
+		var load_balancer_rules []string
+		if rules := properties.LoadBalancingRules; rules != nil {
+			for _, ruleConfig := range *rules {
+				if id := ruleConfig.ID; id != nil {
+					load_balancer_rules = append(load_balancer_rules, *id)
+				}
+			}
 		}
+		d.Set("load_balancer_rules", load_balancer_rules)
```
bflad: Nitpick: This is a trivial case (currently) but best practice when using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```go
if err := d.Set("load_balancer_rules", load_balancer_rules); err != nil {
  return fmt.Errorf("error setting load_balancer_rules: %s", err)
}
```

azurerm/resource_arm_loadbalancer_probe.go
```diff
@@ -9,47 +9,59 @@ import (
...
 			"protocol": {
 				Type:             schema.TypeString,
 				Computed:         true,
 				Optional:         true,
 				StateFunc:        ignoreCaseStateFunc,
-				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+				DiffSuppressFunc: suppress.CaseDifference,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(network.ProbeProtocolHTTP),
+					string(network.ProbeProtocolHTTPS),
```
bflad: Nitpick: We may want to consider adding an acceptance test to cover this new value

katbyte:
Tests pass:
screen shot 2018-08-09 at 12 16 48

PR #1746, opened 2018-08-09
Requiring existing resources to be imported / support for timeouts

tombuildsstuff:
This PR fixes a couple of cross-resource bugs in the Provider:

- Firstly this works around the fact most of the Azure API's are Upserts by requiring that resources are present in the state before creating them
- This adds support for Timeouts, which fixes #171

This is being done in one hit so it's consistent, rather than being inconsistent across resources.

## Resource Support

- [x] App Service
- [x] Application Gateway
- [x] Application Insights
- [x] Application Security Group
- [x] Automation
- [x] Auto Scale
- [x] Availability Set
- [x] AzureAD
- [x] CDN
- [x] Container Group
- [x] Container Registry
- [x] Container Service
- [x] CosmosDB
- [x] Data Lake
- [x] DNS
- [x] EventGrid
- [x] EventHub
- [x] Express Route
- [x] Function App
- [x] Image
- [x] IoTHub
- [x] Key Vault
- [x] Kubernetes
- [x] Local Network Gateway
- [x] Log Analytics
- [x] Managed Disks
- [x] Management Locks
- [x] Monitor
- [x] MySQL
- [x] Networks
- [x] Notification Hub
- [x] Packet Capture
- [x] Policy
- [x] Postgres
- [x] Public IP
- [x] Recovery Services
- [x] Redis
- [x] Relay
- [x] Roles
- [x] Routes
- [x] Scheduler Job
- [x] Search
- [x] Service Fabric
- [x] ServiceBus
- [x] Snapshot
- [x] SQL
- [x] Subnets
- [x] Traffic Manager
- [x] User Assigned Identity
- [x] Virtual Network Gateways

---

virtual resources are slightly different:

- [x] Load Balancer
- [x] Logic Apps

requires more extensive testing:

- [x] Storage
- [x] Template Deployments
- [x] Virtual Machines
- [x] Virtual Machine Disk Attachments
- [x] Virtual Machine Extentions
- [x] VM Scale Sets

needs to be done last / in this order:
- [x] Networks
- [x] Resource Groups

---

other

- [x] remove the default timeout in `config.go`
- [x] removes a bunch of crash points in ACS
- [x] Data Lake Store File ID dependent on #1856
- [x] `azurerm_dns_cname_record` - removing the removed `records` field
- [x] ~Dependent on #1816~ merged
- [ ] Upgrade to AutoRest v11
- [x] Update to include latest PR's once merged:
  - [ ] API Management
  - [ ] Automation DSC Configuration
[...]

PR #1752, opened 2018-08-10
Fix #502 - Terraform loses access token and requires az login

### LINKED ISSUE #502: Terraform loses access token and requires az login
Terraform requires az login after 30 minutes of in activity, even though az account list still works.

### Terraform Version
Terraform v0.10.8

### Affected Resource(s)
All

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

```hcl
resource "random_id" "server" {
  keepers = {
    azi_id = 1
  }

  byte_length = 8
}

    name = "resourceGroup1"
    location = "West Europe"
}

resource "azurerm_cosmosdb_account" "test" {
  name                = "${random_id.server.hex}"
  offer_type          = "Standard"
  consistency_policy {
    consistency_level = "BoundedStaleness"
  }

  failover_policy {
    location = "West Europe"
    priority = 0
  }

  failover_policy {
    location = "East US"
    priority = 1
  }

  tags {
    hello = "world"
  }
}
```

### Debug Output

Error: Error running plan: 1 error(s) occurred:

* provider.azurerm: No valid (unexpired) Azure CLI Auth Tokens found. Please run `az login`.

### Expected Behavior
az account list => works
terraform plan => works

### Actual Behavior
az account list => works
terraform plan => gives the message above
(this happens after a bit of time from when I did az login, the command works for a while, but go to lunch and it shows above error).

### Steps to Reproduce
1. `terraform plan`

lawrencegripper:
This PR uses the callbacks present in the ADAL library along with a 'find and replace in file' function. These changes allow TF to use an expired token and persist the updated refresh token back to the AzureCLI file.

### DIFF
--- a/azurerm/config.go
+++ b/azurerm/config.go
@@ -334,6 +334,13 @@ func getAuthorizationToken(c *authentication.Config, oauthConfig *adal.OAuthConf
 		return nil, err
 	}
 
+	// Set a callback which updates the azure cli refresh token on disk
+	if c.IsAzureCLI {
+		spt.SetRefreshCallbacks([]adal.TokenRefreshCallback{
+			c.AzureCLIRefreshCallback,
+		})
+	}
+
 	err = spt.Refresh()
 
 	if err != nil {
--- a/azurerm/helpers/authentication/access_token.go
+++ b/azurerm/helpers/authentication/access_token.go
@@ -29,8 +29,7 @@ func findValidAccessTokenForTenant(tokens []cli.Token, tenantId string) (*Access
 		}
 
 		if expirationDate.UTC().Before(time.Now().UTC()) {
-			log.Printf("[DEBUG] Token %q has expired", token.AccessToken)
-			continue
+			log.Printf("[DEBUG] Token %q has expired but will be renewed", token.AccessToken)
 		}
 
 		if !strings.Contains(accessToken.Resource, "management") {
--- a/azurerm/helpers/authentication/config.go
+++ b/azurerm/helpers/authentication/config.go
@@ -2,6 +2,9 @@ package authentication
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
 
 	"log"
 
@@ -26,10 +29,12 @@ type Config struct {
 	ClientSecret string
 
 	// Bearer Auth
-	AccessToken  *adal.Token
-	IsCloudShell bool
-	UseMsi       bool
-	MsiEndpoint  string
+	AccessToken             *adal.Token
+	IsCloudShell            bool
+	IsAzureCLI              bool
+	AzureCLIRefreshCallback adal.TokenRefreshCallback
+	UseMsi                  bool
+	MsiEndpoint             string
 }
 
 func (c *Config) LoadTokensFromAzureCLI() error {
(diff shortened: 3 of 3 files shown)


tombuildsstuff:
hey @lawrencegripper 

Thanks for this PR.

Given #1746 is going to be making some pretty major changes to the Azure API clients across the provider - I'm going to suggest we hold off reviewing this PR until #1746 is merged. I believe the general approach should work - however I'm now wondering if we should just parse and persist the file regardless (rather than replacing the auth token in it) given we're reliant on it / parsing it at the moment anyway 🤔

We're working through #1746 at the moment and should have this finished in the coming days; but once that's merged we'll circle back around to this :)

Thanks!

joshgav:
FWIW the file format is an internal detail and could change (or the file be removed) in the future. But we'll offer more robust ways to reuse the CLI tokens before that happens and keep you informed.

cc @yugangw-msft

lawrencegripper:
> however I'm now wondering if we should just parse and persist the file regardless (rather than replacing the auth token in it) given we're reliant on it / parsing it at the moment anyway

Yeah I think this is another option, I went down the find and replace route as it means that the TF provider can weather certain changes to the file without causing issues. For example, if the CLI team added an additional property it would be lost if the code unmarshaled to the TF struct (missing the new property) and wrote out the update back to the file.

tombuildsstuff:
hey @lawrencegripper 

Sorry for the delayed response here - whilst the approach taken in this PR is valid, since this PR was opened there's a new mechanism become available in the Azure SDK for Go where access tokens can be retrieved via shelling out to the Azure CLI (which is then responsible for refreshing the token and persisting it - which means it's officially supported).

In the next couple of releases (currently scheduled for v1.20) - we're planning on refactoring the authentication framework out into it's own package/library - such that it can be reused across [the AzureRM Backend](https://www.terraform.io/docs/backends/types/azurerm.html) too; at which time it'd make sense to switch over to using this new authentication method for the Azure CLI (it'd also be a good time to support authenticating [using a Service Principal and a Certificate](https://github.com/terraform-providers/terraform-provider-azurerm/issues/1573)).

I spent a little bit of time recently prototyping this and it seems to work fairly well - such that I believe it should be possible to reuse the authentication logic across the AzureRM Provider, the Azure Stack Provider and the AzureRM Backend - with feature toggles for which authentication modes are supported in which location (e.g. there's no CloudShell in Azure Stack). As mentioned above - this is currently scheduled for a couple of releases time, so whilst I appreciate this may be a little longer to wait - I think it's worth the extra time to make this consistent across the Backend too.

As such - I hope you don't mind but I'm going to close this PR for the moment in favour of the upcoming refactoring detailed above which is scheduled for the near future - that said I'd like to thank you for this contribution :)

Thanks!

lawrencegripper:
@tombuildsstuff Sure happy for a different approach, I do have one concern about using the `az account get-access-token` method which caused me to discount it when I was doing the PR. I'm guessing your way ahead of me on this one but just noting down in case it's useful. 

The CLI method doesn't handle an operation which runs longer than the life of the token, without some extra work. Calling `get-access-token` returns just a token, no refresh token like so..

```json 
az account get-access-token
{
  "accessToken": "tokenhere",
  "expiresOn": "2018-11-01 11:38:17.942332",
  "subscription": "subhere",
  "tenant": "tenanthere",
  "tokenType": "Bearer"
}

``` 

When that is used by the client, if the token has expired, the sdk client will return a 403 and error. Whereas in the this PR the client will see that a refresh_token is available and use this to get a new token then repeat the request (writing the update refresh token back to disk). 

It looks like the [default lifetime for a token is 1hour](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-configurable-token-lifetimes) but this is configurable by the tenant so could be shorter (minimum 10mins max 1day).

So if a `terraform apply` process runs longer than the `expiresOn` then it would fail and this *could* be 10mins after starting.

One workaround would be to hook into each request made by the SDK and inject some retry logic with shells out to get a new token if the first attempt returns a 403

yugangw-msft:
To use `az account get-access-token`, 2 suggestions in case they help,
1. Leverage the `expiresOn` field, and only re-invoke the command when the expiration is  <5 minutes away. JFYI, 5 minutes buffer is to handle clock skew.
2. Do the check right before send out every request payload to ARM. Since the token is good for 60 minutes, so most of such checks will be a no-op.

PR #1755, opened 2018-08-11
New data source: azurerm_log_analytics_workspace

TechyMatt:


azurerm/data_source_log_analytics_workspace.go
```diff
@@ -0,0 +1,102 @@
...
+		Read: dataSourceLogAnalyticsWorkspaceRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"location": locationForDataSourceSchema(),
+
+			"resource_group_name": resourceGroupNameDiffSuppressSchema(),
```
tombuildsstuff: can we update this to be `resourceGroupNameForDataSourceSchema`?

website/docs/d/log_analytics_workspace.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+
+* `id` - The Log Analytics Workspace ID.
+
+* `primary_shared_key` - The Primary shared key for the Log Analytics Workspace.
+
+* `secondary_shared_key` - The Secondary shared key for the Log Analytics Workspace.
+
+* `workspace_id` - The Workspace (or Customer) ID for the Log Analytics Workspace.
+
+* `portal_url` - The Portal URL for the Log Analytics Workspace.
```
tombuildsstuff: can we also add `retention_in_days`, `sku` and `tags` here?

azurerm/data_source_log_analytics_workspace_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+  tags {
+    environment = "production"
+  }
+}
+`, rInt, location, rString)
+}
+
+func testAccDataSourceAzureRMLogAnalyticsWorkspace_basicWithDataSource(rInt int, rString string, location string) string {
+	config := testAccDataSourceAzureRMStorageAccount_basic(rInt, rString, location)
```
tombuildsstuff: _minor_ we could probably switch to using the basic resource from the Resource instead here? e.g. `testAccAzureRMLogAnalyticsWorkspace_requiredOnly`

tombuildsstuff (changes requested):
hey @mb290 

Thanks for this PR - I've left a few minor comments inline but if we can fix those up this otherwise LGTM 👍 

Thanks!

mb290:
Hi @tombuildsstuff! Changes have been made. Didn't think/know about being able to leverage the resource from the resource test case so have updated it. Let me know if it's not the right way.

mb290:
@tombuildsstuff after a couple of failed builds and understanding the test harness much more I ended up using

`testAccAzureRMLogAnalyticsWorkspace_retentionInDaysComplete`

One of my tests was for the number of days retention so I wanted to make sure we aren't expecting MS to keep the same value, instead we can just use what we set. I also updated the expected Sku based on the new test case.

katbyte (approved):
@mb290, 

Thank you for the requested updates, this LGTM now 👍

katbyte:
Tests pass:

```
gofmt -w $(find . -name '*.go' |grep -v vendor)
TF_ACC=1 go test ./azurerm -v -test.run=TestAccDataSourceAzureRMLogAnalyticsWorkspace -timeout 180m
=== RUN   TestAccDataSourceAzureRMLogAnalyticsWorkspace_basic
--- PASS: TestAccDataSourceAzureRMLogAnalyticsWorkspace_basic (91.97s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	92.005s

```

PR #1773, opened 2018-08-15
[New Resource] azure_data_factory_v2

hbuckle:
Support for creating Data Factories. I don't have a GitHub enterprise account so I haven't been able to test the GitHub repo configuration. It's supposed to support public GitHub repos but the API doesn't seem to allow that currently (https://github.com/Azure/azure-rest-api-specs/issues/3643)


azurerm/provider.go
```diff
@@ -146,6 +146,7 @@ func Provider() terraform.ResourceProvider {
 			"azurerm_container_service":                       resourceArmContainerService(),
 			"azurerm_container_group":                         resourceArmContainerGroup(),
 			"azurerm_cosmosdb_account":                        resourceArmCosmosDBAccount(),
+			"azurerm_data_factory_v2":                         resourceArmDataFactoryV2(),
```
metacpp: >			"azurerm_data_factory_v2": [](start = 0, length = 29)

V2 seems to be redundant information here, coz it's the current version and we never release V1 before. We can explain it in the documentation, but we don't need to add "_v2" here. That's my 2 cents.

azurerm/resource_arm_data_factory_v2.go
```diff
@@ -0,0 +1,343 @@
...
+							Required: true,
+						},
+						"repository_name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"root_folder": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
```
metacpp: Where is `last_commit_id` ?

azurerm/provider.go
metacpp: You may take a look at Azure portal, it allows end user to select v1 or v2. 

---
In reply to: [210760621](https://github.com/terraform-providers/terraform-provider-azurerm/pull/1773#discussion_r210760621) [](ancestors = 210760621)

azurerm/resource_arm_data_factory_v2.go
```diff
@@ -0,0 +1,343 @@
...
+						"tenant_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+
+			"github_configuration": {
```
metacpp: You need to set personal_access_token for Data Factory to install web hooks on the repo.

azurerm/resource_arm_data_factory_v2.go
```diff
@@ -0,0 +1,343 @@
...
+						"root_folder": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"tenant_id": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+					},
+				},
```
metacpp: You need to figure out how to authenticate and authorize data factory to access your repo in vsts. With above information is not enough.

hbuckle: I tested with VSTS and it worked fine as it all just uses Azure AD for authentication. I can update the documentation to state that the VSTS account needs to be under the same Azure AD tenant as the user running Terraform.

azurerm/resource_arm_data_factory_v2.go
hbuckle: It's not required to create the repo - it could be added as a computed property but I'm not really sure what use it would be within Terraform?

azurerm/resource_arm_data_factory_v2.go
hbuckle: As noted on the linked rest-api-specs issue, there's no option to pass in an access token, it may be that GitHub configuration using the API doesn't work. I don't have access to GitHub enterprise to be able to test.

metacpp: image001

@hbuckle see the snapshot. From the UI portal of Data Factory, you will need to provide the token to guarantee it to access repo and install web hooks there.

azurerm/resource_arm_data_factory_v2.go
metacpp: If you set up the integration through UI portal, there's no such option. Is it computed-only ?

azurerm/resource_arm_data_factory_v2.go
hbuckle: Interesting, when I did it through the UI I was prompted to allow the AzureDataFactory OAuth application access to my account and then it worked - I was hoping that as long as that application can access the GitHub account then adding the config through the API would work but it doesn't seem to be the case.

metacpp:
@LijuanZ please help with taking a look at the schema.

azurerm/resource_arm_data_factory_v2.go
```diff
@@ -0,0 +1,343 @@
...
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"collaboration_branch": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"host_name": {
+							Type:     schema.TypeString,
+							Required: true,
```
LijuanZ: host_name is optional, it's needed only for GitHub Enterprise repo

azurerm/resource_arm_data_factory_v2.go
```diff
@@ -0,0 +1,343 @@
...
+						},
+						"root_folder": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"tenant_id": {
+							Type:     schema.TypeString,
+							Required: true,
```
LijuanZ: tenant_id is optional. If it's not set, the vsts repo should be in the same AAD tenant with the data factory

azurerm/resource_arm_data_factory_v2.go
```diff
@@ -0,0 +1,343 @@
...
+						"root_folder": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+					},
+				},
+				ConflictsWith: []string{"vsts_configuration"},
+			},
+
+			"vsts_configuration": {
```
LijuanZ: lastCommitId property is missing in both vsts_configuration and github_configuration and I'm not sure if it's needed in the use case in terraform. Basically this is an optional property used to track changes in which commit has been applied to the data factory. The property is populated if you associate a repo with the factory and go to UI portal and publish some changes

azurerm/resource_arm_data_factory_v2.go
metacpp: @hbuckle After sync up with ADF team, you need to use below API to set up configuration:
https://docs.microsoft.com/en-us/rest/api/datafactory/factories/configurefactoryrepo

And they don't support any authentication through API, UI portal did some extra work.

hbuckle: I am thinking of just removing the github_configuration as it doens't look like it will work using the API

hbuckle: I've updated the code to use configurefactoryrepo, but we've also discovered a casing issue in the specs so github configuration won't work until that filters down to the SDK

hbuckle:
Sorry for the long delay on this one - I no longer have access to the branch this PR is from so I have resubmitted the PR [here](https://github.com/terraform-providers/terraform-provider-azurerm/pull/2478)

Issue #1571, opened 2018-07-13
Feature Request: Virtual Network Rules for Postgresql


spacekitty76:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

Terraform v0.11.7
+ provider.azurerm v1.7.0

### Affected Resource(s)

azurerm_sql_virtual_network_rule

### Terraform Configuration Files

```hcl
provider "azurerm" {
  version = "~> 1.7.0"
}

locals {
  environment    = "${var.environment}"
  mmc_region     = "${var.mmc_region}"
  location_short = "${var.location_short}"
  subnet_number  = "${var.subnet_number}"
}

resource "azurerm_postgresql_server" "postgresql" {
  administrator_login          = "${var.postresql_admin_username}"
  administrator_login_password = "${var.postresql_admin_password}"
  location                     = "${var.location}"
  name                         = "${var.postgresql_server_name}"
  resource_group_name         = "${local.environment}_${local.mmc_region}_${local.location_short}_tier2_rg"

  sku {
    name     = "${var.sku_name}"
    capacity = "${var.sku_capacity}"
    tier     = "${var.sku_tier}"
    family   = "${var.sku_family}"
  }

  storage_profile {
    storage_mb            = "${var.storage_mb}"
    backup_retention_days = "${var.storage_retention}"
    geo_redundant_backup  = "${var.storage_geo_redundant}"
  }

  administrator_login          = "${var.administrator_login}"
  administrator_login_password = "${var.administrator_password}"
  ssl_enforcement              = "${var.ssl_enforcement}"
  version                      = "${var.posgresql_version}"

  tags {}
}

resource "azurerm_postgresql_database" "postgresql" {
  name                = "${var.postgresql_database_name}"
  resource_group_name = "${azurerm_postgresql_server.postgresql.resource_group_name}"
  server_name         = "${azurerm_postgresql_server.postgresql.name}"
  charset             = "${var.postgresql_database_charset}"
  collation           = "${var.posgresql_database_collation}"
}

data "azurerm_subnet" "existing-subnet02" {
[...]

PR #1774, opened 2018-08-15
New Resource: `azurerm_postgresql_virtual_network_rule`

ac-astuartkregor:
This PR allows users to create vnet rules akin to `azurerm_sql_virtual_network_rule` so that service endpoints can be used to access PostgreSQL instances.

The code was copied and adapted from the `azurerm_sql_virtual_network_rule`.

I'd be very grateful for any feedback especially if there are any styling or convention considerations.

My thanks to @lfshr and @tombuildsstuff for their advice and input.

(Fixes: #1571 )

### DIFF
--- a/azurerm/resource_arm_postgresql_virtual_network_rule.go
+++ b/azurerm/resource_arm_postgresql_virtual_network_rule.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/postgresql/mgmt/2017-12-01/postgresql"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/response"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
@@ -38,13 +39,13 @@ func resourceArmPostgreSQLVirtualNetworkRule() *schema.Resource {
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     true,
-				ValidateFunc: validation.NonZeroValues,
+				ValidateFunc: validation.NoZeroValues,
 			},
 
 			"subnet_id": {
 				Type:         schema.TypeString,
 				Required:     true,
-				ValidateFunc: validation.NonZeroValues,
+				ValidateFunc: validation.NoZeroValues,
 			},
 
 			"ignore_missing_vnet_service_endpoint": {
--- a/azurerm/resource_arm_postgresql_virtual_network_rule.go
+++ b/azurerm/resource_arm_postgresql_virtual_network_rule.go
@@ -29,7 +29,7 @@ func resourceArmPostgreSQLVirtualNetworkRule() *schema.Resource {
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     true,
-				ValidateFunc: validateSqlVirtualNetworkRuleName,
+				ValidateFunc: validatePostgreSQLVirtualNetworkRuleName,
 			},
 
 			"resource_group_name": resourceGroupNameSchema(),
--- a/azurerm/resource_arm_postgresql_virtual_network_rule.go
+++ b/azurerm/resource_arm_postgresql_virtual_network_rule.go
@@ -64,7 +64,7 @@ func resourceArmPostgreSQLVirtualNetworkRuleCreateUpdate(d *schema.ResourceData,
 	parameters := postgresql.VirtualNetworkRule{
 		VirtualNetworkRuleProperties: &postgresql.VirtualNetworkRuleProperties{
 			VirtualNetworkSubnetID:           utils.String(virtualNetworkSubnetId),
-			IgnoreMissingVnetServiceEndpoint: utils.Bool(true),
+			IgnoreMissingVnetServiceEndpoint: utils.Bool(false),
 		},
 	}
 
@@ -128,7 +128,6 @@ func resourceArmPostgreSQLVirtualNetworkRuleRead(d *schema.ResourceData, meta in
 
 	if props := resp.VirtualNetworkRuleProperties; props != nil {
 		d.Set("subnet_id", props.VirtualNetworkSubnetID)
-		d.Set("ignore_missing_vnet_service_endpoint", props.IgnoreMissingVnetServiceEndpoint)
 	}
 
 	return nil
--- a/azurerm/resource_arm_postgresql_virtual_network_rule.go
+++ b/azurerm/resource_arm_postgresql_virtual_network_rule.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/postgresql/mgmt/2017-12-01/postgresql"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/response"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
@@ -35,14 +36,16 @@ func resourceArmPostgreSQLVirtualNetworkRule() *schema.Resource {
 			"resource_group_name": resourceGroupNameSchema(),
 
 			"server_name": {
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.NonZeroValues,
 			},
 
 			"subnet_id": {
-				Type:     schema.TypeString,
-				Required: true,
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.NonZeroValues,
 			},
 
 			"ignore_missing_vnet_service_endpoint": {
--- a/azurerm/resource_arm_postgresql_virtual_network_rule.go
+++ b/azurerm/resource_arm_postgresql_virtual_network_rule.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/postgresql/mgmt/2017-12-01/postgresql"
@@ -59,16 +60,53 @@ func resourceArmPostgreSQLVirtualNetworkRuleCreateUpdate(d *schema.ResourceData,
 	name := d.Get("name").(string)
 	serverName := d.Get("server_name").(string)
 	resourceGroup := d.Get("resource_group_name").(string)
-	virtualNetworkSubnetId := d.Get("subnet_id").(string)
+	subnetId := d.Get("subnet_id").(string)
+
+	// due to a bug in the API we have to ensure the Subnet's configured correctly or the API call will timeout
+	// BUG: https://github.com/Azure/azure-rest-api-specs/issues/3719
+	subnetsClient := meta.(*ArmClient).subnetClient
+	subnetParsedId, err := parseAzureResourceID(subnetId)
+
+	subnetResourceGroup := subnetParsedId.ResourceGroup
+	virtualNetwork := subnetParsedId.Path["virtualNetworks"]
+	subnetName := subnetParsedId.Path["subnets"]
+	subnet, err := subnetsClient.Get(ctx, subnetResourceGroup, virtualNetwork, subnetName, "")
+	if err != nil {
+		if utils.ResponseWasNotFound(subnet.Response) {
+			return fmt.Errorf("Subnet with ID %q was not found: %+v", subnetId, err)
+		}
+
+		return fmt.Errorf("Error obtaining Subnet %q (Virtual Network %q / Resource Group %q: %+v", subnetName, virtualNetwork, subnetResourceGroup, err)
+	}
+
+	containsEndpoint := false
+	if props := subnet.SubnetPropertiesFormat; props != nil {
+		if endpoints := props.ServiceEndpoints; endpoints != nil {
+			for _, e := range *endpoints {
+				if e.Service == nil {
+					continue
+				}
+
+				if strings.EqualFold(*e.Service, "Microsoft.Sql") {
+					containsEndpoint = true
+					break
+				}
+			}
+		}
+	}
+
+	if !containsEndpoint {
+		return fmt.Errorf("Error creating PostgreSQL Virtual Network Rule: Subnet %q (Virtual Network %q / Resource Group %q) must contain a Service Endpoint for `Microsoft.Sql`", subnetName, virtualNetwork, subnetResourceGroup)
+	}
 
 	parameters := postgresql.VirtualNetworkRule{
 		VirtualNetworkRuleProperties: &postgresql.VirtualNetworkRuleProperties{
-			VirtualNetworkSubnetID:           utils.String(virtualNetworkSubnetId),
+			VirtualNetworkSubnetID:           utils.String(subnetId),
 			IgnoreMissingVnetServiceEndpoint: utils.Bool(false),
 		},
 	}
 
-	_, err := client.CreateOrUpdate(ctx, resourceGroup, serverName, name, parameters)
+	_, err = client.CreateOrUpdate(ctx, resourceGroup, serverName, name, parameters)
 	if err != nil {
 		return fmt.Errorf("Error creating PostgreSQL Virtual Network Rule %q (PostgreSQL Server: %q, Resource Group: %q): %+v", name, serverName, resourceGroup, err)
 	}
--- a/azurerm/resource_arm_postgresql_virtual_network_rule.go
+++ b/azurerm/resource_arm_postgresql_virtual_network_rule.go
@@ -11,6 +11,7 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/response"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
@@ -45,7 +46,7 @@ func resourceArmPostgreSQLVirtualNetworkRule() *schema.Resource {
 			"subnet_id": {
 				Type:         schema.TypeString,
 				Required:     true,
-				ValidateFunc: validation.NoZeroValues,
+				ValidateFunc: azure.ValidateResourceID,
 			},
 
 			"ignore_missing_vnet_service_endpoint": {
--- a/azurerm/resource_arm_postgresql_virtual_network_rule.go
+++ b/azurerm/resource_arm_postgresql_virtual_network_rule.go
@@ -10,7 +10,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/postgresql/mgmt/2017-12-01/postgresql"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
-	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/response"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
--- a/azurerm/resource_arm_postgresql_virtual_network_rule.go
+++ b/azurerm/resource_arm_postgresql_virtual_network_rule.go
@@ -1,0 +1,250 @@
+package azurerm
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"regexp"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/postgresql/mgmt/2017-12-01/postgresql"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/response"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func resourceArmPostgreSQLVirtualNetworkRule() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceArmPostgreSQLVirtualNetworkRuleCreateUpdate,
+		Read:   resourceArmPostgreSQLVirtualNetworkRuleRead,
+		Update: resourceArmPostgreSQLVirtualNetworkRuleCreateUpdate,
+		Delete: resourceArmPostgreSQLVirtualNetworkRuleDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateSqlVirtualNetworkRuleName,
+			},
+
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"server_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"subnet_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"ignore_missing_vnet_service_endpoint": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false, //When not provided, Azure defaults to false
+			},
+		},
+	}
+}
+
+func resourceArmPostgreSQLVirtualNetworkRuleCreateUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).postgresqlVirtualNetworkRulesClient
+	ctx := meta.(*ArmClient).StopContext
+
+	name := d.Get("name").(string)
+	serverName := d.Get("server_name").(string)
+	resourceGroup := d.Get("resource_group_name").(string)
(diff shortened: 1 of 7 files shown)


lfshr:
This fixes #1571 (title didn't seem to link the issue)

ac-astuartkregor:
Good spot, although the PR title should mean the issue is closed when the PR gets merged.

azurerm/resource_arm_postgresql_virtual_network_rule.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	return nil
+}
+
+/*
+	This function checks the format of the PostgreSQL Virtual Network Rule Name to make sure that
+	it does not contain any potentially invalid values.
+*/
+func validatePostgreSQLVirtualNetworkRuleName(v interface{}, k string) (ws []string, errors []error) {
```
jeffreyCline: Where did you find the definition for these constraints? Because according to the general [naming rules and restrictions](https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions#naming-rules-and-restrictions) it states:

Name | Length | Case | Allowable characters | Suggested Pattern | Example
-- | -- | -- | -- | -- | --
Network Security Group Rule | 1 - 80 | Case insensitive | Alphanumeric, hyphen, underscore, and period | `<descriptive context>` | `sql-allow`

* **Note:** They did not have an example for this exact resource so I picked the next best one that seemed reasonable/similar.

In general, avoid having any special characters (`-` or `_`) as the first or last character in any name. These characters will cause most validation rules to fail. I am also wondering if we could abstract this out to a general validation method in the azure package and use it on other resources as well?

azurerm/resource_arm_postgresql_virtual_network_rule.go
```diff
@@ -0,0 +1,250 @@
...
+
+	parameters := postgresql.VirtualNetworkRule{
+		VirtualNetworkRuleProperties: &postgresql.VirtualNetworkRuleProperties{
+			VirtualNetworkSubnetID:           utils.String(virtualNetworkSubnetId),
+			IgnoreMissingVnetServiceEndpoint: utils.Bool(ignoreMissingVnetServiceEndpoint),
+		},
+	}
+
+	_, err := client.CreateOrUpdate(ctx, resourceGroup, serverName, name, parameters)
+	if err != nil {
```
jeffreyCline: Can we combine these into a single line?
```
if _, err := client.CreateOrUpdate(ctx, resourceGroup, serverName, name, parameters); err != nil {
```

website/docs/r/postgresql_virtual_network_rule.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+  server_name         = "${azurerm_postgresql_server.postgresql_server.name}"
+  subnet_id           = "${azurerm_subnet.subnet.id}"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the SQL virtual network rule. Changing this forces a new resource to be created. Cannot be empty and must only contain alphanumeric characters and hyphens. Cannot start with a number, and cannot start or end with a hyphen.
```
jeffreyCline: Can we move the `Changing this forces a new resource to be created.` to the end of the sentence?

*  `name` - (Required) The name of the SQL virtual network rule. Cannot be empty and must only contain alphanumeric characters and hyphens. Cannot start with a number, and cannot start or end with a hyphen. Changing this forces a new resource to be created.

ac-astuartkregor:
Hi @jeffreyCline, many thanks for your feedback. I'll look at addressing this today. I should point out that since this is almost a carbon copy of sql virtual network rule resource, you may wish to also apply the same refactorings to that as well.

azurerm/resource_arm_postgresql_virtual_network_rule.go
```diff
@@ -0,0 +1,250 @@
...
+				ForceNew:     true,
+				ValidateFunc: validatePostgreSQLVirtualNetworkRuleName,
+			},
+
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"server_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
```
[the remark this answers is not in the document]
ac-astuartkregor: @jeffreyCline Do you have any suggestions for testing that this validation has been applied to this/these attribute(s)? Is it worth testing such a thing? Sorry if this is a daft question, however I'm not sure how it could be tested given the style of the existing unit and acceptance tests.

azurerm/resource_arm_postgresql_virtual_network_rule.go
ac-astuartkregor: These rules are present in the `resource_arm_sql_virtual_network_rule` which this resource has been copied from. They overall seem to be quite sensible, although I would be happy to adjust them to be a bit more stringent, especially around length or beginning/ending characters.

In terms of abstracting out this validation, this might be worth doing although it'd be worth looking at the existing resources where the validation might apply to assess how valuable that would be.

azurerm/resource_arm_postgresql_virtual_network_rule.go
ac-astuartkregor: I'm not sure I fully understand the motivation for combining these lines, could you please advise?

azurerm/resource_arm_postgresql_virtual_network_rule.go
```diff
@@ -0,0 +1,250 @@
...
+	"fmt"
+	"log"
+	"regexp"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/postgresql/mgmt/2017-12-01/postgresql"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/response"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
```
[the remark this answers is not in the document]
ac-astuartkregor: Would this be under the circumstances where the validation function is pulled out of this resource and made more generally available?

azurerm/resource_arm_postgresql_virtual_network_rule.go
jeffreyCline: The err variable is scoped only to the if block that handles it, and I believe that the code is a little easier to read this way as well. Since the err variable is not needed beyond the scope of the check it makes sense to combine the lines into a single line if check.

azurerm/resource_arm_postgresql_virtual_network_rule.go
jeffreyCline: This is actually fine for now, In the near future I will implement the validation functions in the azure helper package based on Microsoft's Naming conventions best practices. Once I have completed that work I will update this resource to use the new validation rules.

azurerm/resource_arm_postgresql_virtual_network_rule.go
ac-astuartkregor: Thanks for clarifying. I suspect I do not feel as strongly on this issue as you do, so I'm happy to change it. I'm not a seasoned Go programmer, so I am not particularly used to its syntax or conventions and as such, the current code (split across 2 lines) is a lot easier for me to read as opposed to combining the lines. However, because of that, I don't know how go is 'supposed' to look either. I don't mind changing it, however I suspect less experienced people will find it a bit more awkward (I'm biased, I know).

In terms of scoping, that's not something I'd have thought of. Does Go give particularly strong emphasis to scope beyond the expectation of dealing with it? Are there expected patterns for scopes and so on? Sorry for all these questions but you've piqued my curiosity. Am I correct in guessing that if the `if` comes before the variable declaration, that the variable is then scoped to that and only that block/line?

jeffreyCline: I'm not going to push this issue because this is the way the original resource was coded and it's not your fault. However, the use of `err` on line 84 is pointless because if the code gets to this point the `err` will always be `nil`. I would put the `if` check in a single line and remove the `err` variable from the log output completely. 

```
log.Printf("[DEBUG] Waiting for PostgreSQL Virtual Network Rule %q (PostgreSQL Server: %q, Resource Group: %q) to become ready.", name, serverName, resourceGroup)
``` 

Yes Go is big on scope and making the code as short and compact as possible. Go doesn't like to have variables hanging around any longer than they absolutely have too.

jeffreyCline:
image

azurerm/resource_arm_postgresql_virtual_network_rule.go
ac-astuartkregor: 🤦‍♂️ Good point, I totally hadn't considered that, my bad.

Ok, thanks for the input, I'll keep that in mind.

ac-astuartkregor:
When attempting to create an `azurerm_postgresql_virtual_network_rule` with the `Microsoft.Sql` endpoint missing, even though the request is invalid (by the same logic as `azurerm_sql_virtual_network_rule`), the ARM request receives an HTTP 202 and terraform continues to check for the creation of the resource, instead receiving an HTTP 404. The debug output is as follows:

```
2018-08-21T17:38:17.284+0100 [DEBUG] plugin.terraform-provider-azurerm: 2018/08/21 17:38:17 [DEBUG] AzureRM Request:
2018-08-21T17:38:17.284+0100 [DEBUG] plugin.terraform-provider-azurerm: PUT /subscriptions/<redacted>/resourceGroups/acctestRG-1/providers/Microsoft.DBforPostgreSQL/servers/acctestpostgresqlsvr-1/virtualNetworkRules/acctestpostgresqlvnetrule1?api-version=2017-12-01 HTTP/1.1
2018-08-21T17:38:17.284+0100 [DEBUG] plugin.terraform-provider-azurerm: Host: management.azure.com
2018-08-21T17:38:17.284+0100 [DEBUG] plugin.terraform-provider-azurerm: User-Agent: Go/go1.10.3 (amd64-darwin) go-autorest/v10.12.0 Azure-SDK-For-Go/v18.0.0 postgresql/2017-12-01;HashiCorp-Terraform-v0.11.3
2018-08-21T17:38:17.284+0100 [DEBUG] plugin.terraform-provider-azurerm: Content-Length: 243
2018-08-21T17:38:17.284+0100 [DEBUG] plugin.terraform-provider-azurerm: Authorization: Bearer <redacted>
  subnet_id:                            "" => "/subscriptions/<redacted>/resourceGroups/acctestRG-1/providers/Microsoft.Network/virtualNetworks/acctestvnet1/subnets/acctestsubnet1"
2018-08-21T17:38:17.284+0100 [DEBUG] plugin.terraform-provider-azurerm: Content-Type: application/json; charset=utf-8
2018-08-21T17:38:17.284+0100 [DEBUG] plugin.terraform-provider-azurerm: Accept-Encoding: gzip
2018-08-21T17:38:17.284+0100 [DEBUG] plugin.terraform-provider-azurerm: 
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: 2018/08/21 17:38:17 [DEBUG] AzureRM Response for https://management.azure.com/subscriptions/<redacted>/resourceGroups/acctestRG-1/providers/Microsoft.DBforPostgreSQL/servers/acctestpostgresqlsvr-1/virtualNetworkRules/acctestpostgresqlvnetrule1?api-version=2017-12-01:
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: HTTP/1.1 202 Accepted
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: Content-Length: 88
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: Azure-Asyncoperation: https://management.azure.com/subscriptions/<redacted>/resourceGroups/acctestRG-1/providers/Microsoft.DBforPostgreSQL/locations/canadacentral/azureAsyncOperation/8d3081ed-2509-4465-9c93-5dc6b26ee686?api-version=2017-12-01
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: Cache-Control: no-cache
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: Content-Type: application/json; charset=utf-8
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: Date: Tue, 21 Aug 2018 16:39:05 GMT
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: Expires: -1
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: Location: https://management.azure.com/subscriptions/<redacted>/resourceGroups/acctestRG-1/providers/Microsoft.DBforPostgreSQL/locations/canadacentral/operationResults/8d3081ed-2509-4465-9c93-5dc6b26ee686?api-version=2017-12-01
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: Pragma: no-cache
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: Retry-After: 30
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: Server: Microsoft-HTTPAPI/2.0
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: Strict-Transport-Security: max-age=31536000; includeSubDomains
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: X-Content-Type-Options: nosniff
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: X-Ms-Correlation-Request-Id: b66ac427-9f24-47fb-991a-80aeab0f6609
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: X-Ms-Ratelimit-Remaining-Subscription-Writes: 1197
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: X-Ms-Request-Id: 8d3081ed-2509-4465-9c93-5dc6b26ee686
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: X-Ms-Routing-Request-Id: CANADAEAST:20180821T163906Z:b66ac427-9f24-47fb-991a-80aeab0f6609
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: 
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: 2018/08/21 17:38:17 [DEBUG] Waiting for PostgreSQL Virtual Network Rule "acctestpostgresqlvnetrule1" (PostgreSQL Server: "acctestpostgresqlsvr-1", Resource Group: "acctestRG-1") to become ready: <nil>
2018-08-21T17:38:17.677+0100 [DEBUG] plugin.terraform-provider-azurerm: 2018/08/21 17:38:17 [DEBUG] Waiting for state to become: [Ready]
```

For comparison, here's the debug output for creating an `azurerm_sql_virtual_network_rule`:
```
2018-08-22T09:48:23.728+0100 [DEBUG] plugin.terraform-provider-azurerm: 2018/08/22 09:48:23 [DEBUG] AzureRM Request:
2018-08-22T09:48:23.728+0100 [DEBUG] plugin.terraform-provider-azurerm: PUT /subscriptions/<redacted>/resourceGroups/example-sql-server-vnet-rule/providers/Microsoft.Sql/servers/unqiueazuresqlserver/virtualNetworkRules/sql-vnet-rule?api-version=2015-05-01-preview HTTP/1.1
  resource_group_name:                  "" => "example-sql-server-vnet-rule"
2018-08-22T09:48:23.728+0100 [DEBUG] plugin.terraform-provider-azurerm: Host: management.azure.com
2018-08-22T09:48:23.728+0100 [DEBUG] plugin.terraform-provider-azurerm: User-Agent: Go/go1.10.3 (amd64-darwin) go-autorest/v10.12.0 Azure-SDK-For-Go/v18.0.0 sql/2015-05-01-preview;HashiCorp-Terraform-v0.11.3
2018-08-22T09:48:23.728+0100 [DEBUG] plugin.terraform-provider-azurerm: Content-Length: 260
2018-08-22T09:48:23.728+0100 [DEBUG] plugin.terraform-provider-azurerm: Authorization: <redacted>
server_name:                          "" => "unqiueazuresqlserver"
2018-08-22T09:48:23.728+0100 [DEBUG] plugin.terraform-provider-azurerm: Content-Type: application/json; charset=utf-8
2018-08-22T09:48:23.728+0100 [DEBUG] plugin.terraform-provider-azurerm: Accept-Encoding: gzip
2018-08-22T09:48:23.728+0100 [DEBUG] plugin.terraform-provider-azurerm: 
  subnet_id:                            "" => "/subscriptions/<redacted>/resourceGroups/example-sql-server-vnet-rule/providers/Microsoft.Network/virtualNetworks/example-vnet/subnets/example-subnet"
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: 2018/08/22 09:48:24 [DEBUG] AzureRM Response for https://management.azure.com/subscriptions/<redacted>/resourceGroups/example-sql-server-vnet-rule/providers/Microsoft.Sql/servers/unqiueazuresqlserver/virtualNetworkRules/sql-vnet-rule?api-version=2015-05-01-preview:
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: HTTP/1.1 400 Bad Request
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: Content-Length: 511
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: Cache-Control: no-cache
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: Content-Type: application/json; charset=utf-8
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: Date: Wed, 22 Aug 2018 08:48:24 GMT
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: Expires: -1
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: Pragma: no-cache
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: Server: Microsoft-HTTPAPI/2.0
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: Strict-Transport-Security: max-age=31536000; includeSubDomains
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: X-Content-Type-Options: nosniff
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: X-Ms-Correlation-Request-Id: 97da2c8d-682e-49e9-80bc-ca616387df5f
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: X-Ms-Ratelimit-Remaining-Subscription-Writes: 1198
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: X-Ms-Request-Id: 6fe2f017-f29f-4060-a896-274ba437fd18
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: X-Ms-Routing-Request-Id: CANADAEAST:20180822T084824Z:97da2c8d-682e-49e9-80bc-ca616387df5f
2018-08-22T09:48:24.908+0100 [DEBUG] plugin.terraform-provider-azurerm: 
```

Looking at the documentation for [PostgreSQL virtual network rules](https://docs.microsoft.com/en-us/REST/API/postgresql/virtualnetworkrules/createorupdate) and [SQL virtual network rules](https://docs.microsoft.com/en-us/REST/API/sql/virtualnetworkrules/createorupdate), this error response are not documented for the SQL virtual network rule. There isn't any documentation for the virtual network rules around the behaviour to do with what happens when `IgnoreMissingVnetServiceEndpoint` is set to `false` but the service endpoint is missing.

From this point, I'm not sure what to do. There is a difference in behaviour in the Azure API between these two resources. Should the test in Terraform be removed? Should the parameter be removed and defaulted?

jeffreyCline:
image

jeffreyCline:
@ac-astuartkregor What I would do in this case, since this is an API issue, is implement a t.Skip for this one test case.

```
func TestAccAzureRMPostgreSQLVirtualNetworkRule_IgnoreEndpointInvalid(t *testing.T) {
    t.Skip("Skipping this test because the ARM request receives status code HTTP 202 Accepted, instead of HTTP 404 Bad Request.") 
```

ac-astuartkregor:
@jeffreyCline Thanks for the advice, I'll also try and raise an issue on github somewhere for that API discrepancy.

tombuildsstuff:
@ac-astuartkregor sorry just seen this.

> From this point, I'm not sure what to do. There is a difference in behaviour in the Azure API between these two resources. Should the test in Terraform be removed? Should the parameter be removed and defaulted?

I'd suggest we'd be better to remove this test for the moment rather than skipping it - since the API doesn't support this behaviour at present (and may not) - and instead document this behaviour. It would however be worth raising an issue on [the Azure SDK for Go repository](https://github.com/Azure/azure-sdk-for-go) about this since the API should return an error rather than a 202 when the data's invalid?

ac-astuartkregor:
Hi @tombuildsstuff :) No worries
> It would however be worth raising an issue on the Azure SDK for Go repository about this since the API should return an error rather than a 202 when the data's invalid?

I was thinking about putting the issue there but was also wondering whether there is a more related repo on Github for such issues. I'll take a look.

In terms of documenting the behaviour, would you also suggest straight-up removing the parameter to the resource and putting it in the resource docs? I can go about removing the test shortly.

jeffreyCline:
@ac-astuartkregor Yes we need to remove the `ignore_vnet_service_endpoint` attribute completely. I am following up with the internal Microsoft service team to get this fixed, but I agree with Tom that there is no reason to block this from shipping because of this one issue. If you can remove that attribute from the resource and the documentation I think we will be good to go.

ac-astuartkregor:
Cheers @jeffreyCline, working on it.

azurerm/resource_arm_postgresql_virtual_network_rule.go
```diff
@@ -0,0 +1,247 @@
...
+		return fmt.Errorf("Error reading PostgreSQL Virtual Network Rule: %q (PostgreSQL Server: %q, Resource Group: %q): %+v", name, serverName, resourceGroup, err)
+	}
+
+	d.Set("name", resp.Name)
+	d.Set("resource_group_name", resourceGroup)
+	d.Set("server_name", serverName)
+
+	if props := resp.VirtualNetworkRuleProperties; props != nil {
+		d.Set("subnet_id", props.VirtualNetworkSubnetID)
+		d.Set("ignore_missing_vnet_service_endpoint", props.IgnoreMissingVnetServiceEndpoint)
```
tombuildsstuff: since this field's not currently present in the schema, can we remove it for the moment?

website/docs/r/postgresql_virtual_network_rule.html.markdown
```diff
@@ -0,0 +1,99 @@
...
+2. Cannot start with a number or hyphen
+3. Cannot end with a hyphen
+
+* `resource_group_name` - (Required) The name of the resource group where the PostgreSQL server resides. Changing this forces a new resource to be created.
+
+* `server_name` - (Required) The name of the SQL Server to which this PostgreSQL virtual network rule will be applied to. Changing this forces a new resource to be created.
+
+* `subnet_id` - (Required) The ID of the subnet that the PostgreSQL server will be connected to.
+
+~> **NOTE:** The resource is configured with `ignore_missing_vnet_service_endpoint` set to `true`, meaning the deployment will succeed even if the target subnet does not contain the `Microsoft.Sql` endpoint in the `service_endpoints` array. This attribute will be introduced once the API behaviour is consistent.
```
tombuildsstuff: given this defaults to `false` in the other resources, I think it'd be sensible to make this `false` by default for the moment (since otherwise this'll cause conflicts for users when we introduce this field with a default of `false` when it's fixed) - what do you think?

azurerm/resource_arm_postgresql_virtual_network_rule.go
ac-astuartkregor: That should be fine, aye. I'll take a look presently.

website/docs/r/postgresql_virtual_network_rule.html.markdown
ac-astuartkregor: I had a brief debate about this with myself and what happens when the endpoint is missing.

I see where you're coming from and perhaps it would be best to do that. I was going to ask what you think is the best approach from a user experience perspective.

* If `IgnoreMissingVnetServiceEndpoint` is false and the endpoint is missing, the resource creation will timeout.
* If `IgnoreMissingVnetServiceEndpoint` is true and the endpoing is missing, the resource creation will succeed but the functionality will be broken.

I suspect your suggestion is probably the better choice but I'm curious as to what you think.

ac-astuartkregor: Oh.

jeffreyCline:
@tombuildsstuff I removed the `waiting-response` tag since @ac-astuartkregor had committed the requested change to the PR.

website/docs/r/postgresql_virtual_network_rule.html.markdown
tombuildsstuff: whilst we've removed this, to ensure this fails gracefully I've pushed a commit (I hope you don't mind!) to add the above code to the Create method, to ensure the Subnet's in a usable state prior to creating the subnet rule

website/docs/r/postgresql_virtual_network_rule.html.markdown
```diff
@@ -88,7 +88,7 @@ The following arguments are supported:
 
 The following attributes are exported:
 
-* `id` - The ID of the SQL virtual network rule.
+* `id` - The ID of the PostgreSQL Virtual Network Rule.
```
ac-astuartkregor: Whoops, sorry @tombuildsstuff my bad.

tombuildsstuff: all good - I missed it on the first pass through :)

tombuildsstuff:
Tests pass:

screenshot 2018-09-04 at 08 28 13

ac-astuartkregor:
Hi @tombuildsstuff,
You're most welcome! Sorry that things were a bit sloppy in places and thank you for implementing the check on the subnet that you mentioned. Probably the best approach to be honest. Hopefully it won't be long before it can be removed.

Thanks for merging too :)

PR #1781, opened 2018-08-15
cleaning up some package tech debt 

katbyte:
- removed `subscription` and `[as]schema` packages
- removed extra suppress case diff function


azurerm/data_source_app_service.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/terraform/helper/schema"
-	azSchema "github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
```
metacpp: Based on the files you changed in this PR due to package name change, keep using package alias might be a better idea.

tombuildsstuff: that's not needed in this instance as nothing else in these files is using a package named `azure` - we should only use aliases where necessary

metacpp: 1. To avoid conflict with other packages.
2. Wrapping to ease future refactoring.

PR #1805, opened 2018-08-21
Key Vault: support for soft delete

janboll:
Adding the features from closed #1517

### DIFF
--- a/azurerm/data_source_key_vault.go
+++ b/azurerm/data_source_key_vault.go
@@ -105,6 +105,16 @@ func dataSourceArmKeyVault() *schema.Resource {
 				Computed: true,
 			},
 
+			"enabled_for_soft_delete": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+
+			"enabled_for_purge_protection": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+
 			"tags": tagsForDataSourceSchema(),
 		},
 	}
@@ -138,6 +148,9 @@ func dataSourceArmKeyVaultRead(d *schema.ResourceData, meta interface{}) error {
 		d.Set("enabled_for_deployment", props.EnabledForDeployment)
 		d.Set("enabled_for_disk_encryption", props.EnabledForDiskEncryption)
 		d.Set("enabled_for_template_deployment", props.EnabledForTemplateDeployment)
+		d.Set("enabled_for_soft_delete", props.EnableSoftDelete)
+		d.Set("enabled_for_purge_protection", props.EnablePurgeProtection)
+
 		if err := d.Set("sku", flattenKeyVaultDataSourceSku(props.Sku)); err != nil {
 			return fmt.Errorf("Error flattening `sku` for KeyVault %q: %+v", *resp.Name, err)
 		}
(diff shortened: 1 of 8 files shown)


leewilson86:
@tombuildsstuff Are the changes in this PR due to be released as part of 1.21.0? Or will it be released in a later version?

We have hit a scenario where the soft-delete and purge-protection features are becoming critical for us. 

Our workaround at the moment is by firing Azure PowerShell/Azure CLI commands to apply soft-delete and purge-protection which is messy in my opinion, and it also adds dependencies for developers to have either Azure Powershell or Azure CLI installed on their machines in order to run the TF for Key Vault. 

We have considered having PowerShell scripts accessed as an Azure Function to apply soft-protection and purge-protection but this maintaining our IaC more complex than it needs to be.

Setting the following:
```
enabled_for_soft_delete: true
enabled_for_purge_protection: true
```
... is way better than something like this:

```
resource "null_resource" "akv-protection" {
    provisioner "local-exec" {
        command = "az keyvault update --name MyKeyVault --enable-soft-delete true --enable-purge-protection true --subscription ********"
    }
}
```

This PR would really benefit us, and we would really like to see it released as part of this 1.21.0 or an upcoming version not in the too distant future.

leewilson86:
@tombuildsstuff @katbyte is there any update on when this PR will be released? Is it still targetted for v1.21.0 as per the milestones?

Thanks

tombuildsstuff:
hey @janboll 

Sorry for the delayed comment here - we've discussed internally why this is blocked but I've just realised we haven't commented to explain why here 🤦‍♂️.

This PR's complicated due to the `purge_on_delete` flag: whilst it's possible to retrieve the value of this field in the delete if the Terraform Configuration is present (e.g. if someone's running `terraform destroy`) it's not present (and thus will always return the default value for a bool [`false`]) if somebody's removed the Terraform Configuration for the resource (for example if the resource has been deleted in the local file, but exists in Azure, so a `terraform plan` will show this requiring deletion).

Our current thinking is that it could be possible for Terraform to enable purge protection on the Resource, but that running `terraform destroy` should actually completely destroy (e.g. purge) this resource. This would allow accidental deletions outside of Terraform to be recoverable, but allows Terraform to fulfil its promise of actually destroying the infrastructure when you run `terraform destroy` (e.g. so that the Resource Group can also be deleted). Whilst we have precedent for Resources in Terraform which exist and do soft deletions, these apply only to the keys (e.g. in AWS [the `aws_kms_key` resource](https://www.terraform.io/docs/providers/aws/r/kms_key.html#deletion_window_in_days)) - which are top-level resources whereas in Azure the presence of the Key Vault and the Resource Group makes this more complicated, since this will prevent them from being destroyed.

Thinking some more about this, originally I thought that this may need to go in 2.0 (since it was a breaking behavioural change, which is why this has been blocked) - but on reflection I believe if we lookup the state of the resource in the delete and purge it if soft-delete is enabled we should be able to ship this in a 1.x release - what do you think?

Thanks!

tombuildsstuff:
hey @janboll 

Just to give an update here - we'll be taking a look at the work required for 2.0 in the not-too-distant future; however since this is blocked on that (since this'll need to be rolled out as a part of 2.0). Rather than leaving this PR open stagnant I'm going to temporarily close this PR for the moment, but we'll re-open this and take another look when we start work on the 2.0 milestone in the not-too-distant future.

Thanks!

WodansSon:
Porting changes to new code base with PR #5344

PR #1819, opened 2018-08-23
docs/r/management_lock: Fix typo: note -> notes

phy1729:
Cf. https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_management_lock.go#L48


metacpp:
@phy1729 Thanks for the fix.

PR #1824, opened 2018-08-24
Adding azurerm example - 2-tier-iis-sql-vm with AD

ghostinthewires:
Adding an example that builds the infrastructure for to host a 2 tier application with IIS & MS SQL joined to Active Directory


ghostinthewires:
Hi, Do I need to do anything else on this?

examples/2-tier-iis-sql-vm/modules/sql-vm/variables.tf
```diff
@@ -0,0 +1,39 @@
...
+
+variable "admin_username" {
+  description = "The username associated with the local administrator account on the virtual machine"
+}
+
+variable "admin_password" {
+  description = "The password associated with the local administrator account on the virtual machine"
+}
+
+variable sqlvmcount {}
```
tombuildsstuff: this syntax (using a variable without the quotes) will break in Terraform 0.12 - it'd be worth adding quotes around this to ensure it's compatible in the future

examples/2-tier-iis-sql-vm/modules/network/variables.tf
```diff
@@ -0,0 +1,21 @@
...
+variable wafsubnet_name {}
+variable wafsubnet_prefix {}
+variable rpsubnet_name {}
+variable rpsubnet_prefix {}
+variable issubnet_name {}
+variable issubnet_prefix {}
+variable dbsubnet_name {}
+variable dbsubnet_prefix {}
+variable dcsubnet_name {}
+variable dcsubnet_prefix {}
```
tombuildsstuff: this syntax (defining a variable without the name being in quotes) will break in Terraform 0.12 - it'd be worth updating these to be quoted to ensure this remains compatible in the future

examples/2-tier-iis-sql-vm/modules/iis-vm/outputs.tf
```diff
@@ -0,0 +1,4 @@
+###################################################################################################
+# Outputs
+####################################################################################################
+
```
tombuildsstuff: we could remove this file since there's no outputs here

examples/2-tier-iis-sql-vm/modules/dc2-vm/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+variable "admin_username" {
+  description = "The username associated with the local administrator account on the virtual machine"
+}
+
+variable "admin_password" {
+  description = "The password associated with the local administrator account on the virtual machine"
+}
+
+variable "domainadmin_username" {
+  description = "The password associated with the local administrator account on the virtual machine"
```
tombuildsstuff: isn't this variable reflecting the username of the Domain Administrator?

examples/2-tier-iis-sql-vm/example/variables.tf
```diff
@@ -0,0 +1,52 @@
...
+variable vmcount {}
+
+# Domain Controller 2
+variable "dc2private_ip_address" {}
+variable "domainadmin_username" {}
+
+# SQL LB
+variable "lbprivate_ip_address" {}
+# SQL DB Servers
+variable sqlvmcount {}
```
tombuildsstuff: (as below) this syntax will break in Terraform 0.12 - could we quote these to ensure this remains compatible with future versions of Terraform?

examples/2-tier-iis-sql-vm/example/outputs.tf
```diff
@@ -0,0 +1,4 @@
+###################################################################################################
+# Outputs
+####################################################################################################
```
tombuildsstuff: since there's nothing in this file - could we remove it for the moment?

examples/2-tier-iis-sql-vm/example/main.tf
```diff
@@ -0,0 +1,102 @@
...
+  admin_username                = "${var.admin_username}"
+  admin_password                = "${var.admin_password}"
+  domainadmin_username          = "${var.domainadmin_username}"
+}
+
+##########################################################
+## Create SQL Server VM Join domain
+##########################################################
+module "sql-vm" {
+  source                    = "..\\modules\\sql-vm"
```
tombuildsstuff: this syntax will only work on Windows - I believe using `./modules/sql-vm` should work across Linux, macOS and Windows?

examples/2-tier-iis-sql-vm/example/main.tf
```diff
@@ -0,0 +1,102 @@
+# Configure the Microsoft Azure Provider
+provider "azurerm" {
+  subscription_id = "${var.subscription_id}"
+  client_id       = "${var.client_id}"
+  client_secret   = "${var.client_secret}"
+  tenant_id       = "${var.tenant_id}"
```
tombuildsstuff: Rather than specifying these variables in-line (which hard-codes this to a Service Principal) - if you leave these values unassigned Terraform will try and pull them either from the Environment Variables or from the Azure CLI - which means that these variables aren't as  likely to be accidentally checked-in to source control.

One thing that we'd strongly recommend doing in the Provider block however is pinning to a version, for example:

```
provider "azurerm" {
  version = "=1.17.0"
}
```

ghostinthewires:
Hi Tom,

Thanks for all the excellent feedback. Much appreciated.

I have contributed a Module to the Registry today. This will create a VM and create an Active Directory domain. I will add a Module for joining a VM to an existing domain next.

How can I get this Verified on the registry?

https://registry.terraform.io/modules/ghostinthewires/ad-create/azurerm/1.0.0

PR #1724, opened 2018-08-03
Modify Update and Delete azurerm_role_definition

shelleybess:
This pull request addresses the issue described in https://github.com/terraform-providers/terraform-provider-azurerm/issues/1540 where the azurerm_role_definition was unable to be updated or deleted.

Changes to Delete:
Changes have been made to get the path of the resource and pull the id from the end.

Changes to CreateOrUpdate:
Now retrieves the id from the end of the path if it exists or generates one if the resource path is empty.

(fixes #1540)

PR #1827, opened 2018-08-24
role_definition: Fix deletion when role_definition_id is not defined

phy1729:
`azurerm_role_definiton` cannot delete the resource if the `role_definition_id` is not provided. Parse the ID from the full ID that Terraform stores instead of relying on the user defining `role_definition_id`.

Given a main.tf
```
data "azurerm_subscription" "example" {}

resource "azurerm_role_definition" "example" {
  name  = "example"
  scope = "${data.azurerm_subscription.example.id}"

  permissions {
    not_actions = ["*"]
  }

  assignable_scopes = ["${data.azurerm_subscription.example.id}"]
}
```

After a `terraform apply`, I expect `terraform destroy` to delete the resource; however, it returns following error.
```
Error: Error applying plan:

1 error(s) occurred:

* azurerm_role_definition.example (destroy): 1 error(s) occurred:

* azurerm_role_definition.example: Error deleting Role Definition "": authorization.RoleDefinitionsClient#Delete: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidRoleDefinitionId" Message="The role definition ID '' is not valid."
```

### DIFF
--- a/azurerm/resource_arm_role_definition.go
+++ b/azurerm/resource_arm_role_definition.go
@@ -3,6 +3,7 @@ package azurerm
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-01-01-preview/authorization"
 	"github.com/hashicorp/go-uuid"
@@ -163,7 +164,8 @@ func resourceArmRoleDefinitionDelete(d *schema.ResourceData, meta interface{}) e
 	client := meta.(*ArmClient).roleDefinitionsClient
 	ctx := meta.(*ArmClient).StopContext
 
-	roleDefinitionId := d.Get("role_definition_id").(string)
+	parts := strings.Split(d.Id(), "/")
+	roleDefinitionId := parts[len(parts)-1]
 	scope := d.Get("scope").(string)
 
 	resp, err := client.Delete(ctx, scope, roleDefinitionId)


azurerm/resource_arm_role_definition.go
```diff
@@ -163,7 +164,8 @@ func resourceArmRoleDefinitionDelete(d *schema.ResourceData, meta interface{}) e
 	client := meta.(*ArmClient).roleDefinitionsClient
 	ctx := meta.(*ArmClient).StopContext
 
-	roleDefinitionId := d.Get("role_definition_id").(string)
+	parts := strings.Split(d.Id(), "/")
```
JunyiYi: @katbyte @tombuildsstuff , what is the point of introducing the `role_definition_id` if we already has the globally `id` attribute?

I would like to suggest deprecating it using the [similar syntax here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/670b1b93238dbf646dab956955fc708159327659/azurerm/resource_arm_dns_cname_record.go#L41). But if we have to keep it, I would also suggest the following logic:

```go
roleDefId := d.Get("role_definition_id").(string)
if roleDefId == "" {
  if err != nil { ... }
  roleDefId = id.Path[...]
}
```

JunyiYi:
This PR seems to be a subset of #1724 .

azurerm/resource_arm_role_definition.go
tombuildsstuff: @JunyiYi in this case we're going to need to update the ID to contain all of the required information - [see here](https://github.com/terraform-providers/terraform-provider-azurerm/pull/1816/commits/12b06b6f0140cb8e385131af47d85989e698d3aa) for an example of how we're doing this with the storage resources. In general we shouldn't be accessing the config in the Read and Delete functions since it can be present (at Import time during the Read function, and then the associated Terraform Config's been deleted in the Delete function)

tombuildsstuff:
hey @phy1729 

Thanks for this PR - apologies for the delayed movement here. I hope you don't mind but we've ended up taking a (similar) but slightly different approach in #1887 which I'm going to close this PR in favour of.

Thanks!

Issue #1812, opened 2018-08-22
VNET Peering - allow_virtual_network_access = false by default can we change it to true?


josephglaspie:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

This setting is optional but took me some time to figure out and get past.

azurerm_virtual_network_peering VNET PEERING `allow_virtual_network_access = false` by default and not needed in the module but is true by default in portal. 

for example I would build the following without issues but could not talk to vms in the VNET: 
  
 ```
resource "azurerm_virtual_network_peering" "resources-to-er" {
  name                      = "resources-to-er"
  resource_group_name       = "${azurerm_resource_group.er-rg.name}"
  virtual_network_name      = "${azurerm_virtual_network.resource-vnet.name}"
  remote_virtual_network_id = "${azurerm_virtual_network.er-vnet.id}"
  use_remote_gateways       = true
  depends_on                = ["azurerm_virtual_network_peering.er-to-resources"]
}
```

I would delete both peers and build them in the portal and everything worked. However, I noticed this was set as true in the portal but missing completely in my module. After investigating it turns out this setting is false by default in terraform. Not a show stopper but definitely adds some serious confusion to the mix.

```
resource "azurerm_virtual_network_peering" "resources-to-er" {
  name                      = "resources-to-er"
  resource_group_name       = "${azurerm_resource_group.er-rg.name}"
  virtual_network_name      = "${azurerm_virtual_network.resource-vnet.name}"
  remote_virtual_network_id = "${azurerm_virtual_network.er-vnet.id}"
  use_remote_gateways       = true
  allow_virtual_network_access = true
  depends_on                = ["azurerm_virtual_network_peering.er-to-resources"]
}
```

Can we have terraform default to true just like the portal does?

THANKS!!!! :)

metacpp:
@VaijanathB will take a look at this issue.

mocofound:
See PR #1829

PR #1829, opened 2018-08-26
Vnet Peering allow virtual network access false instead of true

mocofound:
It was identified that when creating a vnet peering, allow_virtual_network_access = false was the result while =true was the expected result in the portal.   #1812


azurerm/resource_arm_virtual_network_peering.go
```diff
@@ -48,7 +48,7 @@ func resourceArmVirtualNetworkPeering() *schema.Resource {
 			"allow_virtual_network_access": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Computed: true,
+				Default:  false,
```
JunyiYi: The default value is already `false` according to the issue.

azurerm/resource_arm_virtual_network_peering.go
```diff
@@ -48,7 +48,7 @@ func resourceArmVirtualNetworkPeering() *schema.Resource {
 			"allow_virtual_network_access": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Computed: true,
```
JunyiYi: Why did you remove `Computed`?

mocofound:
Closing;  Needs rework.

PR #1835, opened 2018-08-28
Allow azurerm_function_app to use upper case names in consumption plan

JunyiYi:
This fixes BUG #1765 , which will only happen on functions app with consumption plan using `"Dynamic"` `"Y1"` sku. The root cause is that content share name is not allowed to use upper case letters.

New API request: https://github.com/Azure/azure-rest-api-specs/issues/3750.

### DIFF
--- a/azurerm/resource_arm_function_app.go
+++ b/azurerm/resource_arm_function_app.go
@@ -518,7 +518,7 @@ func getBasicFunctionAppAppSettings(d *schema.ResourceData, appServiceTier strin
 
 	storageConnection := d.Get("storage_connection_string").(string)
 	functionVersion := d.Get("version").(string)
-	contentShare := d.Get("name").(string) + "-content"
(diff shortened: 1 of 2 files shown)


azurerm/resource_arm_function_app.go
```diff
@@ -518,7 +518,7 @@ func getBasicFunctionAppAppSettings(d *schema.ResourceData, appServiceTier strin
 
 	storageConnection := d.Get("storage_connection_string").(string)
 	functionVersion := d.Get("version").(string)
-	contentShare := d.Get("name").(string) + "-content"
+	contentShare := strings.ToLower(d.Get("name").(string)) + "-content"
```
tombuildsstuff: can we add a TODO with a link to the public bug report for this?

tombuildsstuff:
Tests pass:

screenshot 2018-08-30 at 12 01 29

Issue #1848, opened 2018-08-30
Add Support for EnableAggregationPipeline Capability in Cosmos DB (Mongo)


jamesggraf:
### Description

Cosmos DB has added a preview feature to enable aggregation pipelines to its implementation of Mongo. In Cosmos parlance this feature is called `EnableAggregationPipeline`

### New or Affected Resource(s)

* azurerm_cosmosdb_account

### Potential Terraform Configuration

```hcl
resource "azurerm_cosmosdb_account" "db" {
  name                = "my-cosmos-db"
  location            = "${azurerm_resource_group.rg.location}"
  resource_group_name = "${azurerm_resource_group.rg.name}"
  offer_type          = "Standard"
  kind                    = "MongoDB"

  enable_automatic_failover = false
  capabilities        = [{
    name = "EnableAggregationPipeline"
  }]
}
```

### References

More information about this capability can be found here:

* https://azure.microsoft.com/en-us/updates/azure-documentdb-now-supports-aggregate-functions/

jamesggraf:
I've created a PR for this feature. I've never written anything in Go and I'm not sure how to test this so please be gentle :)

https://github.com/terraform-providers/terraform-provider-azurerm/pull/1849

katbyte:
Blocked on https://github.com/Azure/azure-sdk-for-go/issues/2864

brunofurmon:
I actually had to come to this issue to find out how to configure the **capabilities** parameter of the azurerm module. Documentation lacks the correct format it needs to accept the parameter.

PR #1849, opened 2018-08-30
Added EnableAggregationPipeline to Cosmos DB capabilities #1848

jamesggraf:
I've never written anything in Go and I'm not 100% certain on how to test this so please be gentle :)


azurerm/resource_arm_cosmos_db_account.go
```diff
@@ -198,6 +198,7 @@ func resourceArmCosmosDBAccount() *schema.Resource {
 								"EnableTable",
 								"EnableGremlin",
 								`EnableCassandra`,
+								"EnableAggregationPipeline",
```
tombuildsstuff: can we add this value to [the list of acceptable values in the documentation](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/cosmosdb_account.html.markdown#L78) too?

jamesggraf: Done :)

tombuildsstuff:
hey @jamesggraf 

So I've run the test suite for this and it appears this fails because the feature isn't available:

```
------- Stdout: -------
=== RUN   TestAccAzureRMCosmosDBAccount_aggregation
--- FAIL: TestAccAzureRMCosmosDBAccount_aggregation (1092.50s)
	testing.go:513: Step 0 error: After applying this step, the plan was not empty:
		
		DIFF:
		
		UPDATE: azurerm_cosmosdb_account.test
		  capabilities.#:               "0" => "1"
		  capabilities.2476034231.name: "" => "EnableAggregationPipeline"
```

Out of interest - do you know if this feature is in an invite-only Beta/Preview?

Thanks!

jamesggraf:
Not as far as I know? If I enable that feature through the Preview Features pane in the Cosmos UI I see that capability listed as being enabled. Can you confirm that you see the same thing when you set it? Is this flag just not available through their API?

dmartosp:
Great!!! i need this new feature, when i could use it?

katbyte:
Hi @jamesggraf,

I can confirm the test is using mongo, you can see [it here](https://github.com/terraform-providers/terraform-provider-azurerm/pull/1849/files#diff-f169635c5b59a9e9b132e769c6479c34R584)

jamesggraf:
Thanks katbyte, is there documentation for how these tests are run? I'd like to better understand why Terraform is having trouble enabling this capability even though I've seen it try to unset it after it's enabled manually through Azure's portal.

Issue #1758, opened 2018-08-13
Data source for Azure AD principals


tomasaschan:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When creating resources, one often wants to also create role bindings for IAM, but I can't find a way to reference Azure AD principals with the currently available data sources. It would be nice to be able to query for e.g. users by email in the same way I can query for applications by name, and get e.g. their object ids.

### New or Affected Resource(s)

* `data azurerm_azuread_user`

### Potential Terraform Configuration
```hcl
provider "azurerm" {}

data azurerm_azuread_user "tomas" {
    username = "tomas@example.com"
}

output "tomas_object_id" {
    id = "${data.azurerm_azuread_user.tomas.id}"
}
```

tiwood:
I'm going to work on this in the next days.

tombuildsstuff:
hey @tomasaschan 

Thanks for opening this issue - apologies for the delayed response here!

Support for a Data Source for AzureAD Principals (Users / Groups) has previously been requested in #1272; so that we don't have multiple issues open tracking the same thing I'm going to close this issue in favour of #1272.

Thanks!

PR #1865, opened 2018-09-04
[WIP] New data source "azurerm_azuread_user"

tiwood:
This PR adds a new data source for Azure Active Directory users.

- [x] New data source 'azurerm_azuread_user'
- [ ] Add markdown documentation
- [ ] Add tests

## Example Usage

```hcl
// This queries an Azure AD user by ID
data "azurerm_azuread_user" "my_user_by_id" {
  object_id = "b2c800cc-b059-11e8-8d7f-47eeee7848b0"
}

// This queries an Azure AD user by userPrincipalName
data "azurerm_azuread_user" "my_user_by_upn" {
  user_principal_name = "john.doe@terraform.io"
}
```

* `id` - The Object ID of the Azure AD User.
* `object_id` - The Object ID of the Azure AD User.
* `user_principal_name` - The User Principal Name of the Azure AD User.

### DIFF
--- a/azurerm/config.go
+++ b/azurerm/config.go
@@ -109,6 +109,7 @@ type ArmClient struct {
 	roleDefinitionsClient   authorization.RoleDefinitionsClient
 	applicationsClient      graphrbac.ApplicationsClient
 	servicePrincipalsClient graphrbac.ServicePrincipalsClient
+	usersClient             graphrbac.UsersClient
 
 	// Autoscale Settings
 	autoscaleSettingsClient insights.AutoscaleSettingsClient
@@ -483,6 +484,10 @@ func (c *ArmClient) registerAuthentication(endpoint, graphEndpoint, subscription
 	servicePrincipalsClient := graphrbac.NewServicePrincipalsClientWithBaseURI(graphEndpoint, tenantId)
 	c.configureClient(&servicePrincipalsClient.Client, graphAuth)
 	c.servicePrincipalsClient = servicePrincipalsClient
+
+	usersClient := graphrbac.NewUsersClientWithBaseURI(graphEndpoint, tenantId)
+	c.configureClient(&usersClient.Client, graphAuth)
+	c.usersClient = usersClient
 }
 
 func (c *ArmClient) registerCDNClients(endpoint, subscriptionId string, auth autorest.Authorizer, sender autorest.Sender) {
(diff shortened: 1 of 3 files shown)


tiwood:
This closes #1758

tomasaschan:
@tiwood What's the status here? Anything I can do to help this becoming ready to merge?

katbyte:
@tomasaschan,

This PR still requires tests & documentation. It might also might be nice to expose more of the user's attributes.

tiwood:
I'm going to update the PR in the next few days.

@katbyte, any ideas which additional attributes should be exposed?

tombuildsstuff:
hey @tiwood @tomasaschan

Thanks for this PR - apologies for the delayed review of this!

Just to give some context behind the delayed review of this: as a part of the upcoming v2.0 release of the AzureRM Provider we're thinking through some of the longer-term changes that we'd like to make to the Provider.

At this point we're trying to determine where the Azure Active Directory resources belong, since there's now enough edge-cases/feature enhancements that this may make sense in it's own Provider; which would allow us to more easily ship support for the AzureAD resources which are harder to test (such as Users).

The proposal to split the AzureAD resources out into its own provider is being tracked in #2322 - would you be able to take a look and let us know your thoughts?

Thanks!

tomasaschan:
I'd say that even though I think the approach in #2322 is sound, there is value in adding something simple to the 1.x line anyway (at least if there is at least one 1.x release planned), to make this functionality available for people who are held back due to unrelated breaking changes between 1.x and 2.0.

I know I want this sooner rather than later, and I don't want to have to wait for 2.0 if I don't have to ;)

tombuildsstuff:
@tomasaschan as mentioned in #2322 - we're planning on splitting AzureAD out into it's own provider; such that this PR will need to be migrated over to that repository once it's available.

Given there's no (non trivial) way of configuring acceptance tests for the User resource in particular (since it requires confirming in an email and then completing a Captcha) - unfortunately I don't believe we'll be able to ship this in a release of the AzureRM Provider (since we'd have no means of running regressions testing short of doing this manually, which is complicated given the rate of changes in this codebase). However as mentioned in #2322 - once the AzureAD Provider has been split out into it's own codebase it should be possible to support this resource there (since it's going to have a slower release cadence, such that we're able to manually test this resource).

As there's no immediate path forward for this (until the AzureAD Provider has been split out, which should be happening soon-ish) - rather than leaving this PR in a holding position I'm going to temporarily close this PR for the moment until we're able to proceed with this (I hope you don't mind @tiwood)

Thanks!

PR #1877, opened 2018-09-06
New Data Source: `azurerm_management_group`

tombuildsstuff:
Tests pass:

```
$ acctests azurerm TestAccDataSourceArmManagementGroup_basic
=== RUN   TestAccDataSourceArmManagementGroup_basic
--- PASS: TestAccDataSourceArmManagementGroup_basic (57.37s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	57.713s
```

### DIFF
--- a/azurerm/data_source_management_group.go
+++ b/azurerm/data_source_management_group.go
@@ -1,0 +1,109 @@
+package azurerm
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/Azure/azure-sdk-for-go/services/preview/resources/mgmt/2018-03-01-preview/management"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmManagementGroup() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmManagementGroupRead,
+
+		Schema: map[string]*schema.Schema{
+			"group_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"display_name": {
+				Type:     schema.TypeString,
+				Computed: true,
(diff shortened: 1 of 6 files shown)


azurerm/data_source_management_group.go
```diff
@@ -0,0 +1,109 @@
...
+	client := meta.(*ArmClient).managementGroupsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	groupId := d.Get("group_id").(string)
+
+	recurse := true
+	resp, err := client.Get(ctx, groupId, "children", &recurse, "", managementGroupCacheControl)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[INFO] Management Group %q doesn't exist - removing from state", d.Id())
```
radeksimko: Not a huge deal, but I thought we used to log these as `WARN` in some other providers I worked with? 🤔

azurerm/data_source_management_group.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		d.Set("parent_management_group_id", parentId)
+
+	}
+
+	return nil
+}
+
+func flattenArmManagementGroupDataSourceSubscriptionIds(input *[]managementgroups.ChildInfo) (*schema.Set, error) {
+	subscriptionIds := &schema.Set{F: schema.HashString}
```
radeksimko: I may be wrong, but I _think_ that for _root_ (1st level) `TypeSet` fields the helper schema can convert `[]interface{}` to `Set` automatically.

azurerm/data_source_management_group.go
tombuildsstuff: actually this should be a failure since this is a data source 🤦‍♂️

PR #1879, opened 2018-09-06
New Resource: `azurerm_mysql_virtual_network_rule`

### LINKED ISSUE #1340: Virtual network rules for Azure MySQL service
Hi,
Azure MySQL service recently went GA with support from terraform, there is a feature called virtual network rules which is present for Azure SQL, can we get this added to Azure MySQL also.

https://azure.microsoft.com/en-us/blog/vnet-service-endpoints-for-azure-database-services-for-mysql-and-postgresql-in-preview/

Here is the feature for Azure SQL, hopefully it's straightforward enough to re-use the code
https://www.terraform.io/docs/providers/azurerm/r/sql_virtual_network_rule.html

Thanks!

### Community Note

* azurerm_mysql_server

```hcl
```

### References

* #0000

tombuildsstuff:
This is so similar to the PostgreSQL Virtual Network Rule I've refactored that too so we have a common validation function

Fixes #1340


tombuildsstuff:
PostgreSQL Tests pass:

screenshot 2018-09-06 at 17 11 29

tombuildsstuff:
Pulled the PostgreSQL refactor out into #1882

azurerm/resource_arm_mysql_virtual_network_rule_test.go
```diff
@@ -0,0 +1,476 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMMySqlVirtualNetworkRuleExists(resourceName1),
+					testCheckAzureRMMySqlVirtualNetworkRuleExists(resourceName2),
+					testCheckAzureRMMySqlVirtualNetworkRuleExists(resourceName3),
+				),
+			},
+		},
+	})
+}
+
```
jeffreyCline: Since they allow for in place SKU update should we add a test that:

**Test Steps:**
1. Create the basic MySQL server as in `TestAccAzureRMMySqlVirtualNetworkRule_basic`
2. Updates the SKU and validates that the rules also remain after the update?

This would have to be a `General Purpose` SKU that is upgraded to an `Optimized` SKU, so it the test would not be able to re-use the `TestAccAzureRMMySqlVirtualNetworkRule_basic` configuration file.

tombuildsstuff: IMO that check (that the MySQL Server isn't created when the SKU changes) belongs on the Server resource, rather than on this one? In other providers where the ID's are unique we check the ID of the resource before and after making the change to ensure they're the same, but where Azure uses the name field as the ID we'd need to look into using a different field to verify this is the same resource?

jeffreyCline:
@tombuildsstuff I updated the test SKU's to fix the test cases, hope you don't mind.

jeffreyCline:
@tombuildsstuff I also resolved the conflict with master, hope that's ok.

jeffreyCline:
image

tombuildsstuff:
Tests still pass:

screenshot 2018-09-12 at 16 40 50

PR #1911, opened 2018-09-12
Add unique client request id to pass to ARM API

metacpp:
This PR generates the UUID per provider process, and pass it through `x-ms-client-request-id` to ARM API.

With single `azurerm` provider process running locally, it works perfectly.

### DIFF
--- a/azurerm/config.go
+++ b/azurerm/config.go
@@ -60,9 +60,19 @@ import (
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/hashicorp/terraform/terraform"
+	uuid "github.com/hashicorp/go-uuid"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/authentication"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
+
+var MSClientRequestId = ""
+
+func init() {
+	// Initialize UUID to pass through `x-ms-client-request-id` header.
+	MSClientRequestId, _ = uuid.GenerateUUID()
+
+	log.Printf("[DEBUG] AzureRM Client Request Id: %s\n", MSClientRequestId)
+}
 
 // ArmClient contains the handles to all the specific Azure Resource Manager
 // resource classes' respective clients.
@@ -259,6 +269,7 @@ type ArmClient struct {
 func (c *ArmClient) configureClient(client *autorest.Client, auth autorest.Authorizer) {
 	setUserAgent(client)
 	client.Authorizer = auth
+	client.RequestInspector = azure.WithClientID(MSClientRequestId)
 	client.Sender = autorest.CreateSender(withRequestLogging())
 	client.SkipResourceProviderRegistration = c.skipProviderRegistration
 	client.PollingDuration = 60 * time.Minute
@@ -366,6 +377,9 @@ func getArmClient(c *authentication.Config) (*ArmClient, error) {
 			return nil, envErr
 		}
 	}
+
+
+	log.Printf("AzureRM Client Request Id: %s\n", )
 
 	// client declarations:
 	client := ArmClient{


azurerm/config.go
```diff
@@ -60,10 +60,20 @@ import (
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/hashicorp/terraform/terraform"
+	uuid "github.com/hashicorp/go-uuid"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/authentication"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
+var MSClientRequestId = ""
```
paultyng: no need to export this, would call it `msClientRequestID` (also note capital ID, per Go [initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) convention)

metacpp: Made the change.

azurerm/config.go
```diff
@@ -59,11 +59,21 @@ import (
...
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/authentication"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
+var msClientRequestID = ""
+
+func init() {
+	// Initialize UUID to pass through `x-ms-client-request-id` header.
+	msClientRequestID, _ = uuid.GenerateUUID()
```
tombuildsstuff: rather than having this in the init (and thus not being able to handle the error), I believe this'd be better within the `getArmClient` function in `./azurerm/config.go`? in which case this could become `if msClientRequestID = "" {/*generate*/}`

tombuildsstuff: is there any reason we couldn't use the Request ID returned from the Authorization API, rather than generating our own?

paultyng: re: error handling, i don't think we should honestly, its not worth crashing the run to track this.

metacpp: > is there any reason we couldn't use the Request ID returned from the Authorization API, rather than generating our own?

We could if that Request ID is uniquely generated each time when running `terraform plan` or `terraform apply`. Which Authorization API ? If it's from ARM API, it also needs to be correlated with other APIs within same session.

metacpp: > re: error handling, i don't think we should honestly, its not worth crashing the run to track this.

Agree.

metacpp:
@tombuildsstuff just made the changes per your suggestions. Passing the same client request ID generated  per provider process session is what Azure team like to have to correlate all API logs from the backend.

metacpp:
@paultyng @tombuildsstuff any further comment for this PR? :-)

paultyng:
@metacpp we were wrong about the multiple processes issue, its a single process per provider type (not block), so the global var should cover all aliases, but we may want to move this back to `init` so that all aliases of the provider share the same client identifier.

metacpp:
@paultyng I've made the change, please review it.

azurerm/config.go
```diff
@@ -264,9 +265,30 @@ type ArmClient struct {
...
+
+// clientRequestID generates a UUID to pass through `x-ms-client-request-id` header.
+func clientRequestID() string {
+	msClientRequestIDOnce.Do(func() {
+		var err error
+		msClientRequestID, err = uuid.GenerateUUID()
+
+		if err != nil {
+			log.Printf("[WARN] Fail to generate uuid for msClientRequestID: %+v", err)
+		}
```
jeffreyCline: Should we return nil here since we hit an error and not let it fall through?

metacpp: > Should we return nil here since we hit an error and not let it fall through?

This client request id is not blocking step to call ARM API, so we don't need to fail fast.

metacpp:
@paultyng Thanks for the merge!

PR #1913, opened 2018-09-12
Expose ID of an embedded subnet defined in azurerm_virtual_network

### LINKED ISSUE #666: Nested subnet declarations in azurerm_virtual_network should be have their IDs exposed as Attributes
It seems that when creating a subnet as part of an `azurerm_virtual_network`, the subnet is not actually referencable in any fashion from other resources, such as `azurerm_network_interface` which is **required** to have a `subnet_id`.

Unfortunately this means it's impossible to add a NIC (as far as I can tell) to a VNet which has been created with an embedded `subnet{}` declaration, e.g.:

```hcl
resource "azurerm_virtual_network" "public_prod" {
  name                = "${var.prefix}-jenkins-public-prod"
  resource_group_name = "${azurerm_resource_group.public_prod.name}"
  address_space       = ["10.0.0.0/16"]

  subnet {
    name           = "app-tier"
    address_prefix = "10.0.1.0/24"
    security_group = "${azurerm_network_security_group.public_app_tier.id}"
  }

```

### Terraform Version

Terraform v0.11.0
+ provider.azurerm v0.3.2
+ provider.random v1.1.0

### Affected Resource(s)

- `azurerm_virtual_network`

If Attributes were exposed by the `azurerm_virtual_network` regarding its `subnet`, then this would be easier to work with 😄

JunyiYi:
* Add test cases to verify ID is actually set
* Add documentation and sample code to explain how to reference the subnet ID
* Expose the subnet ID in azurerm_virtual_network resource

Actually it is hard to reference a specific subnet element in the current implementation because of the hash function. But that is a different story and requires state migration as well. So I just keep it as a small change in this PR.

The PR enables user referencing the subnet ID defined in `azurerm_virtual_network` resource mentioned in issue #666 .

Fixes #666

### DIFF
--- a/website/docs/r/virtual_network.html.markdown
+++ b/website/docs/r/virtual_network.html.markdown
@@ -56,10 +56,6 @@ resource "azurerm_virtual_network" "test" {
     environment = "Production"
   }
 }
-
-output "id_of_subnet2" {
-  value = "${azurerm_virtual_network.test.subnet.2796830261.id}"
-}
 ```
 
 ## Argument Reference
@@ -85,6 +81,8 @@ The following arguments are supported:
     subnets. Each `subnet` block supports fields documented below.
 
 * `tags` - (Optional) A mapping of tags to assign to the resource.
+
+---
 
 The `subnet` block supports:
 
--- a/website/docs/r/virtual_network.html.markdown
+++ b/website/docs/r/virtual_network.html.markdown
@@ -31,9 +31,9 @@ resource "azurerm_network_security_group" "test" {
 
 resource "azurerm_virtual_network" "test" {
   name                = "virtualNetwork1"
+  location            = "${azurerm_resource_group.test.location}"
   resource_group_name = "${azurerm_resource_group.test.name}"
   address_space       = ["10.0.0.0/16"]
-  location            = "West US"
   dns_servers         = ["10.0.0.4", "10.0.0.5"]
 
   subnet {
@@ -55,6 +55,10 @@ resource "azurerm_virtual_network" "test" {
   tags {
     environment = "Production"
   }
+}
+
+output "id_of_subnet2" {
+  value = "${azurerm_virtual_network.test.subnet.2796830261.id}"
 }
 ```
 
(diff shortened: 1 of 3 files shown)


website/docs/r/virtual_network.html.markdown
```diff
@@ -56,6 +56,10 @@ resource "azurerm_virtual_network" "test" {
     environment = "Production"
   }
 }
+
+output "id_of_subnet2" {
+  value = "${azurerm_virtual_network.test.subnet.2796830261.id}"
```
tombuildsstuff: rather than using the hash value, it should be possible to access this as `${azurerm_virtual_network.test.subnet[1].id}` or `${azurerm_virtual_network.test.subnet.*.id[1]}`

website/docs/r/virtual_network.html.markdown
```diff
@@ -105,6 +109,12 @@ The following attributes are exported:
 
 * `address_space` - The address space that is used the virtual network.
 
+* `subnet` - Set of multiple subnets defined in this virtual network. Each `subnet` block supports the fields documented below.
+
+The `subnet` block supports:
```
tombuildsstuff: can we add `---` above this to add a divider as we're doing for the new blocks, and also change `supports` to `exports`(since the values aren't setable)

website/docs/r/virtual_network.html.markdown
```diff
@@ -105,6 +109,12 @@ The following attributes are exported:
 
 * `address_space` - The address space that is used the virtual network.
 
+* `subnet` - Set of multiple subnets defined in this virtual network. Each `subnet` block supports the fields documented below.
```
tombuildsstuff: can we make this `* `subnet` - One or more `subnet` blocks as defined below.`

website/docs/r/virtual_network.html.markdown
JunyiYi: it seems do not work with set. `${azurerm_virtual_network.test.subnet[1].id}` says `Error reading config for output id_of_subnet2: parse error at 1:41: expected "}" but found "."`.

And `${azurerm_virtual_network.test.subnet.*.id[1]}` says `output.id_of_subnet2: At column 43, line 1: invalid index operation into non-indexable type: TypeString in: ${azurerm_virtual_network.test.subnet.*.id[1]}`.

tombuildsstuff: going to remove this example for the moment - once HCL2/TF 0.12 has landed this should be more accessible

tombuildsstuff:
Tests pass (now that #1985 has been temporarily reverted):

screenshot 2018-09-29 at 08 13 08

Issue #646, opened 2017-12-21
time parameter


hashibot:
_This issue was originally opened by @LaneEnglot as hashicorp/terraform#16950. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

```
.terraform --version
Terraform v0.11.1
+ provider.azurerm v0.3.1
```

### Terraform Configuration Files

```
resource "azurerm_servicebus_queue" "k8s_dev1_sb_que" {
  name                = "k8s-dev1-sb-que"
  resource_group_name = "${azurerm_resource_group.k8s-dev1-rg.name}"
  namespace_name      = "${azurerm_servicebus_namespace.k8s_dev1_sb.name}"
  requires_duplicate_detection = "False"
  default_message_ttl = "00:30:00"
  duplicate_detection_history_time_window = "00:15:00"
  enable_partitioning = true
}

resource "azurerm_servicebus_topic" "k8s_dev1_sb_topic" {
  name                = "k8s-dev1-sb-topic"
  resource_group_name = "${azurerm_resource_group.k8s-dev1-rg.name}"
  namespace_name      = "${azurerm_servicebus_namespace.k8s_dev1_sb.name}"
  enable_partitioning = true
  default_message_ttl = "00:30:00"
  duplicate_detection_history_time_window = "00:15:00"
}
```

### Output
```
* azurerm_servicebus_queue.k8s_dev1_sb_que: servicebus.QueuesClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="BadRequest" Message="The string '00:30:00' is not a valid TimeSpan value. CorrelationId: 35d09bb5-cca0-4d3f-a456-cd99a2e3db73"
* azurerm_servicebus_topic.k8s_ukbeta_sb_topic: 1 error(s) occurred:

* azurerm_servicebus_topic.k8s_dev1_sb_topic: servicebus.TopicsClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="BadRequest" Message="The string '00:30:00' is not a valid TimeSpan value. CorrelationId: 45dc663c-d549-47fb-8c09-10083abf7e2d"
[...]

PR #1921, opened 2018-09-13
Update TimeSpan documentation in Service Bus Queue/Topic

JunyiYi:
The original documentation regarding `auto_delete_on_idle`, `default_message_ttl` and `duplicate_detection_history_time_window` is not correct any more. Azure REST APIs have replaced all [C# TimeSpan format](https://docs.microsoft.com/en-us/dotnet/api/system.timespan.parse?redirectedfrom=MSDN&view=netframework-4.7.2#System_TimeSpan_Parse_System_String_) with the new [ISO8601 duration format](https://en.wikipedia.org/wiki/ISO_8601#Durations). This change is also caught in #646 .

In this PR:
- Update the documentation to use the latest ISO8601 duration format
- Add test cases in order to catch such problems as soon as possible in the future
- Add `ValidateFunc` so user could see the error before applying

### DIFF
--- a/azurerm/resource_arm_servicebus_queue.go
+++ b/azurerm/resource_arm_servicebus_queue.go
@@ -40,21 +40,24 @@ func resourceArmServiceBusQueue() *schema.Resource {
 			"resource_group_name": resourceGroupNameSchema(),
 
 			"auto_delete_on_idle": {
-				Type:     schema.TypeString,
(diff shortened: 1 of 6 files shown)


tombuildsstuff:
Tests pass:

screenshot 2018-09-13 at 16 48 39

PR #1926, opened 2018-09-14
App Service Plan: refactoring / handling errors in flatten functions

tombuildsstuff:
Noticed a couple of test failures we need more info to fix:

```
* azurerm_app_service_plan.test: web.AppServicePlansClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> <nil>
```

Ended up doing some crash fixes/set error handling when passing through too


tombuildsstuff:
Tests pass:

screenshot 2018-09-24 at 06 54 07

Issue #1860, opened 2018-09-04
azure_container_group: Support Additional Volume types


lawrencegripper:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

ACI support specifying different volume types. Currently only one of these is supported under the existing provider.  

Name | Type | Description | Supported in Provider
-- | -- | -- | --
azureFile | AzureFileVolume | The Azure File volume. | Yes
emptyDir | EmptyDirVolume | The empty directory volume. | No
gitRepo | GitRepoVolume | The git repo volume.| No
secret | object | The secret volume. | No

I want to add support (happy to submit as PR) for the secret volume type and am about to start looking at implementing it. Given the current code structure I'm not clear on the best approach, currently the provider has one `volume` type: 

``` golang 
"volume": {
	Type:     schema.TypeList,
	Optional: true,
	ForceNew: true,
	Elem: &schema.Resource{
		Schema: map[string]*schema.Schema{
			"name": {
				Type:     schema.TypeString,
				Required: true,
				ForceNew: true,
			},
			"mount_path": {
				Type:     schema.TypeString,
				Required: true,
				ForceNew: true,
			},
			"read_only": {
				Type:     schema.TypeBool,
				Optional: true,
				ForceNew: true,
				Default:  false,
			},
			"share_name": {
				Type:     schema.TypeString,
				Required: true,
				ForceNew: true,
			},
			"storage_account_name": {
				Type:     schema.TypeString,
				Required: true,
				ForceNew: true,
			},
			"storage_account_key": {
				Type:     schema.TypeString,
				Required: true,
				ForceNew: true,
			},
		},
	},          
```
One option is to add lots of option params to the existing volume structure. Adding fields for `secret_data`, `git_url` and a bool to enable empty dir but I think this would make the object quite confusing to use and be a tough one to validate. 

I think my preference would be to create a different, optional, item for each of the volume types, for example `secret_volume`. 

```golang 
"secret_volume": {
	Type:     schema.TypeList,
	Optional: true,
	ForceNew: true,
	Elem: &schema.Resource{
		Schema: map[string]*schema.Schema{
[...]

PR #1929, opened 2018-09-14
ACI resource: Add additional volume support. Refactor volume structure

lawrencegripper:
Resolves: #1860 

Currently working on refactoring the existing provider to support volume mounts and allow volumes to be shared between containers. 

Goal: Add missing volume types

### DIFF
--- a/azurerm/resource_arm_container_group.go
+++ b/azurerm/resource_arm_container_group.go
@@ -134,7 +134,6 @@ func resourceArmContainerGroup() *schema.Resource {
 							Type:     schema.TypeList,
 							Optional: true,
 							ForceNew: true,
-							MaxItems: 1,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"name": {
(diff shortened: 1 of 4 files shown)


lawrencegripper:
Hi, 

I think I've got the basics of how I want to tackle this now in the PR - would someone be able to take a quick look and let me know their thoughts? 

If everything looks good I'll finish things off with the additional volume types and do docs, tests etc

lawrencegripper:
Hopefully this is in a good state for an initial review now. 

This ended up larger than I expected as the API doesn't return 'secrets' or azure storage keys in the read requests so I've added quite a bit of code to match the names of the existing volumes and retrieve the data from the input state. Let me know if there is a better way to approach this, integration tests are passing, docs are added and a test is added for this along with an example.

lawrencegripper:
Is there anything outstanding on me on this one atm?

azurerm/resource_arm_container_group.go
```diff
@@ -185,13 +279,13 @@ func resourceArmContainerGroup() *schema.Resource {
 							Elem:     &schema.Schema{Type: schema.TypeString},
 						},
 
-						"volume": {
```
katbyte: As this is a breaking change we are going to have to support both methods here until 2.0

azurerm/resource_arm_container_group_test.go
```diff
@@ -512,6 +686,17 @@ resource "azurerm_container_group" "test" {
 	os_type             = "linux"
 	restart_policy      = "OnFailure"
 
+	volume {
+    name = "logs"
```
katbyte: Could we property format this?

lawrencegripper:
@katbyte thanks for the review, really appreciated. 

I'll try and find some time to fix up these bits, was originally just looking for an initial review of the approach before tidying things up so sorry about the messy bits!

Do you have any examples in the codebase of where an existing breaking change has been handled? That way I can review the approach and hopefully get it right.

PR #1934, opened 2018-09-17
Check if error is retryable before returning that it is

### LINKED ISSUE #1757: Bad Request errors should not be retried for 5 minutes before being reported to the user
### Community Note

### Terraform Version
```
> terraform -v
Terraform v0.11.7
+ provider.azurerm v1.12.0
```
### Affected Resource(s)

* azurerm_role_assignment (but probably many, possibly all, others)

```hcl
provider "azurerm" {}

resource "azurerm_resource_group" "rg" {
  name      = "rg_foo"
  location  = "West Europe"
}

data "azurerm_builtin_role_definition" "owner" {
  name = "Owner"
}

resource "azurerm_role_assignment" "owners" {
  scope               = "${azurerm_resource_group.rg.id}"
  role_definition_id  = "${data.azurerm_builtin_role_definition.owner.id}"
  principal_id        = "foo@bar.com" // invalid; this should be a guid
}
```

### Debug Output

Debug log and full console output: https://gist.github.com/tlycken/e18e55edb42c84f5d3043e823140b9d9

### Expected Behavior

After the request to create the resource failed with `400 Bad Request` the first time, the error should have been reported to the user, and the process should have been aborted. (Possibly, this specific error could have been validated already before sending the request, but nonetheless `Bad Request` errors should be handled better.)

### Actual Behavior

The process kept waiting for a resource that would never be created, retrying every now and then (see the debug log for details). Not until the role had failed to be created for 5 minutes was an error message presented to the user.

### Steps to Reproduce

1. `terraform apply`

carlpett:
Fixes #1757.

This PR checks if errors returned in `resource_arm_role_assignment` are actually retryable, rather than always retrying. If the approach is ok, it should also be applied to `resource_arm_storage_container`, which uses the same "always retry" logic.
Currently, I just check if the error (or nested error) is a `*url.Error` and if so if it is a temporary or timeout error. Anything else is deemed non-retryable. Are there other cases that should be retried?

### DIFF
--- a/azurerm/resource_arm_role_assignment.go
+++ b/azurerm/resource_arm_role_assignment.go
@@ -3,10 +3,12 @@ package azurerm
 import (
 	"fmt"
 	"log"
+	"net/url"
 	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-01-01-preview/authorization"
+	"github.com/Azure/go-autorest/autorest"
 	"github.com/hashicorp/go-uuid"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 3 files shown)


carlpett:
Also, this probably needs a test. Is the accepted way there to just write an acceptance test and push it and see if it works? I don't have an Azure account that I can/am allowed to run them on.

carlpett:
Sure! I mostly threw it together without having too much insights into the repo at large, your suggestions look very good. Will push an update soon!
Regarding the acceptance tests, do you trigger them for PRs that look good?

katbyte:
@carlpett,

Once the PR LGTM (or almost does) I will run the acceptance test internally for the entire resource. If you write an acceptance test it will be verified at that time.

Issue #1938, opened 2018-09-18
azurerm_service_fabric_cluster support for both admin and reader client certificates


steve-hawkins:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

You should be able to add both an admin and reader client certificate for Service Fabric clusters, currently there is a limit on the attribute which only allows one client certificate to be specified

I've raised this as a feature, but it could be argued that this is a bug

### New or Affected Resource(s)

* azurerm_service_fabric_cluster

### Potential Terraform Configuration

```hcl
  client_certificate_thumbprint {
    thumbprint = "${azurerm_key_vault_certificate.service_fabric_client_admin.thumbprint}"
    is_admin   = true
  }

  client_certificate_thumbprint {
    thumbprint = "${azurerm_key_vault_certificate.service_fabric_client.thumbprint}"
    is_admin   = false
  }
```

tombuildsstuff:
hey @steve-hawkins 

Just to let you know that support for this has now landed in v1.16 of the AzureRM Provider (here's the full changelog](https://github.com/terraform-providers/terraform-provider-azurerm/blob/v1.16.0/CHANGELOG.md)). You can upgrade to to this version by specifying it in your provider block:

```
provider "azurerm" {
  version = "=1.16.0"
}
```

... and then running `terraform init -upgrade` :)

Thanks!

PR #1939, opened 2018-09-18
allow up-to-two client certs

steve-hawkins:
This should address #1938


steve-hawkins:
@katbyte thanks for the feedback, think that test should cover it, but please let me know

katbyte:
@steve-hawkins,

You can run just one test with this command:
```
make fmt; make testacc TEST=./azurestack TESTARGS=-test.run=TestAccAzureStackStorageAccount_basic
```

or all tests starting with a pattern with this command:
```
make fmt; make testacc TEST=./azurestack TESTARGS=-test.run=TestAccAzureStackStorageAccount
```

katbyte:
Tests pass:
screen shot 2018-09-19 at 10 39 06

steve-hawkins:
thanks @katbyte for getting back so quickly

PR #1958, opened 2018-09-20
azurerm_virtual_machine_scale_set: make os_profile.0.admin_password property optional

katbyte:


katbyte:
Tests pass:
screen shot 2018-09-20 at 15 39 42

katbyte:
fixed in azure stack [here](https://github.com/terraform-providers/terraform-provider-azurestack/pull/49)

azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -658,6 +658,24 @@ func resourceArmVirtualMachineScaleSet() *schema.Resource {
...
+					disablePassword = v.(bool)
+				}
+
+				if !disablePassword {
+					return fmt.Errorf(" `os_profile.0.admin_password` must be set unless `os_profile_linux_config.0.disable_password_authentication` is true")
+				}
+			}
+
+			return nil
+		},
```
tombuildsstuff: Admin Password is a field which (IME) is almost always interpolated outside of simple scenarios - in which case this won't work, as such can we remove the CustomizeDiff here?

azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -973,22 +989,25 @@ func flattenAzureRmVirtualMachineScaleSetIdentity(identity *compute.VirtualMachi
...
-			key := make(map[string]interface{})
-			key["path"] = *i.Path
+	if config != nil {
+		result["disable_password_authentication"] = *config.DisablePasswordAuthentication
+
+		if config.SSH != nil && len(*config.SSH.PublicKeys) > 0 {
+			ssh_keys := make([]map[string]interface{}, 0, len(*config.SSH.PublicKeys))
+			for _, i := range *config.SSH.PublicKeys {
+				key := make(map[string]interface{})
+				key["path"] = *i.Path
```
tombuildsstuff: could we nil-check this?

azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -973,22 +989,25 @@ func flattenAzureRmVirtualMachineScaleSetIdentity(identity *compute.VirtualMachi
...
 
-	if config.SSH != nil && len(*config.SSH.PublicKeys) > 0 {
-		ssh_keys := make([]map[string]interface{}, 0, len(*config.SSH.PublicKeys))
-		for _, i := range *config.SSH.PublicKeys {
-			key := make(map[string]interface{})
-			key["path"] = *i.Path
+	if config != nil {
+		result["disable_password_authentication"] = *config.DisablePasswordAuthentication
+
+		if config.SSH != nil && len(*config.SSH.PublicKeys) > 0 {
```
tombuildsstuff: could we also nil-check `config.SSH.PublicKeys` here?

azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -270,7 +271,7 @@ func resourceArmVirtualMachineScaleSet() *schema.Resource {
 			},
 
 			"os_profile_linux_config": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
```
tombuildsstuff: this could be a bigger scope than we anticipated, given how the ARM API's return the latest snapshot of a resource; I think it'd be worth reverting this back to a Set for the moment (with a separate test function, and the hash function; and then spending some time on both VM/VMSS for 2.0?)

katbyte:
Address comments @tombuildsstuff

vijayakrishnarg1:
@tombuildsstuff Thanks for fix.

PR #1961, opened 2018-09-20
Add support for vnet rules to CosmosDB accounts.

### LINKED ISSUE #1342: Virtual Network Service Endpoints for Azure CosmosDB
### Community Note

### Description

Please add the ability to set Virtual Network Service Endpoints for Azure CosmosDB, in addition to the ip_range_filter

also to be able to configure it as a resource (external to the cosmos_db_account)

https://azure.microsoft.com/en-us/blog/virtual-network-service-endpoints-for-azure-cosmos-db-is-now-generally-available/

* azurerm_cosmos_db_account

```hcl
resource "azurerm_cosmosdb_account" "db" {
     ....
      vitual_networks = ""
}

resource "azurerm_cosmosdb_firewall_rule" "db_fw" {

}
```

pgavlin:
These changes add support for virtual network filters to CosmosDB
accounts. Using these filters requires two new properties:
- `is_virtual_network_filter_enabled`, which enables virutal network
  filters for the account
- `virtual_network_rule`, which specifies the subnets that are allowed
  to access the account

Fixes #1342.

### DIFF
--- a/azurerm/data_source_cosmos_db_account.go
+++ b/azurerm/data_source_cosmos_db_account.go
@@ -225,13 +225,7 @@ func dataSourceArmCosmosDBAccountRead(d *schema.ResourceData, meta interface{}) 
 			return fmt.Errorf("Error setting `capabilities`: %+v", err)
 		}
 
-		virtualNetworkRules := make([]map[string]interface{}, len(*props.VirtualNetworkRules))
-		for i, r := range *props.VirtualNetworkRules {
-			virtualNetworkRules[i] = map[string]interface{}{
-				"id": *r.ID,
-			}
-		}
-		if err := d.Set("virtual_network_rule", virtualNetworkRules); err != nil {
+		if err := d.Set("virtual_network_rule", flattenAzureRmCosmosDBAccountVirtualNetworkRulesAsList(props.VirtualNetworkRules)); err != nil {
 			return fmt.Errorf("Error setting `virtual_network_rule`: %+v", err)
 		}
 
@@ -284,4 +278,18 @@ func flattenAzureRmCosmosDBAccountCapabilitiesAsList(capabilities *[]documentdb.
 	}
 
 	return &slice
+}
+
+func flattenAzureRmCosmosDBAccountVirtualNetworkRulesAsList(rules *[]documentdb.VirtualNetworkRule) []map[string]interface{} {
+	if rules == nil {
+		return []map[string]interface{}{}
+	}
+
+	virtualNetworkRules := make([]map[string]interface{}, len(*rules))
+	for i, r := range *rules {
+		virtualNetworkRules[i] = map[string]interface{}{
+			"id": *r.ID,
+		}
+	}
+	return virtualNetworkRules
 }
--- a/azurerm/data_source_cosmos_db_account.go
+++ b/azurerm/data_source_cosmos_db_account.go
@@ -105,6 +105,24 @@ func dataSourceArmCosmosDBAccount() *schema.Resource {
 				},
 			},
 
+			"is_virtual_network_filter_enabled": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+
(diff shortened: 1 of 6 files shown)


azurerm/data_source_cosmos_db_account.go
```diff
@@ -206,6 +225,16 @@ func dataSourceArmCosmosDBAccountRead(d *schema.ResourceData, meta interface{})
 			return fmt.Errorf("Error setting `capabilities`: %+v", err)
 		}
 
+		virtualNetworkRules := make([]map[string]interface{}, len(*props.VirtualNetworkRules))
```
katbyte: ideally we would break this out into a flatten function.

Also we should be checking if `props.VirtualNetworkRules` is nil before dereferencing it.

azurerm/resource_arm_cosmos_db_account.go
```diff
@@ -205,6 +206,26 @@ func resourceArmCosmosDBAccount() *schema.Resource {
 				Set: resourceAzureRMCosmosDBAccountCapabilitiesHash,
 			},
 
+			"is_virtual_network_filter_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
+			"virtual_network_rule": {
```
katbyte: Would this make more sense as a simple set of strings `virtual_network_rule_subnet_ids` rather then a set of blocks containing id?

```
virtual_network_rule_subnet_ids  =[
"${azurerm_subnet.subnet1.id}",
"${azurerm_subnet.subnet2.id}",
]
```

azurerm/resource_arm_cosmos_db_account.go
```diff
@@ -205,6 +206,26 @@ func resourceArmCosmosDBAccount() *schema.Resource {
...
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
```
katbyte: Could we validate the id via `azure.ValidateResourcdId`?

azurerm/resource_arm_cosmos_db_account.go
```diff
@@ -873,6 +919,21 @@ func flattenAzureRmCosmosDBAccountCapabilities(capabilities *[]documentdb.Capabi
 	return &s
 }
 
+func flattenAzureRmCosmosDBAccountVirtualNetworkRules(rules *[]documentdb.VirtualNetworkRule) *schema.Set {
+	results := schema.Set{
+		F: resourceAzureRMCosmosDBAccountVirtualNetworkRuleHash,
+	}
+
+	for _, r := range *rules {
```
katbyte: We should probably be checking if rules is nil before dereferencing it

azurerm/data_source_cosmos_db_account.go
pgavlin: Done.

azurerm/resource_arm_cosmos_db_account.go
pgavlin: Done.

azurerm/resource_arm_cosmos_db_account.go
pgavlin: Done.

azurerm/resource_arm_cosmos_db_account.go
pgavlin: I think that we should err on the side of leaving this as-is: if the Azure API adds additional fields to these rules in the future, having a resource here rather than a simple ID makes such a change easier to accommodate. This would also align better with a future `cosmosdb_account_virtual_network_rule` resource as suggested in #1342.

katbyte: Good points 👍

PR #1973, opened 2018-09-24
.gitattributes for developers on Windows

steve-hawkins:
this change should make it easier for developers on Windows as it will ensure .go files remain LF rather than auto-converted to CRLF


steve-hawkins:
@tombuildsstuff I didn't feel this really fitted into the description of a feature or bug

please let me know if you need me to associate an issue with this PR

tombuildsstuff:
hey @steve-hawkins 

Thanks for this PR :)

Whilst at first glance this seems like a good proposal to make developing on Windows easier - the way Go (and specifically Govendor) does vendoring means this is a more complicated problem unfortunately.

Currently we use Govendor for dependency management, which checks repositories out into its own local cache (located at `$GOROOT/.cache`) before copying the relevant files into the current working directory - which then stores a checksum of the vendored files to ensure these haven't been modified. The issue with adding a `.gitattributes` file is this leads to changes between the upstream dependencies (which could be using either LF or CRLF) and the vendored files, such that `govendor status` would believe that local changes have been made to the vendored files (since the checksum of the checked out files wouldn't match the checksum in the cached folder) across all platforms.

As such for the moment (at least, until there's a way around that) unfortunately I'm going to have to close this PR for the moment - but I'd like to thank you for this contribution :)

Thanks!

steve-hawkins:
@tombuildsstuff thanks for the quick response

the `.gitattributes` was set to target `*.go` files only and simple to make sure that they didn't get changed from LF to CRLF when pulled down from the Git repo to a Windows machine

without this change on a Windows machine when you execute `go fmt` it see the CRLF as a violation and converts the `*.go` files back to LF this in turn is shown as a change in your local Git workspace

I understand your concern, but I may have missed something as the `.gitattributes` change in this PR is to honour the correct upstream line endings which would hopefully ensure that `govendor status` would not then report any changes as from the `go fmt` standards all `*.go` files need to be LF?

tombuildsstuff:
@steve-hawkins the issue's more around how the other repositories are checked out on disk, for instance if you're using `check out CRLF commit LF` (which I believe is the default behaviour on Windows?) - you'll see differences when this repository is using `LF` line-endings for Go files - such that the checksum between the local checkout of the dependency and the vendored files would be different.

Specifically whilst this PR only targets `*.go` files - that includes the `vendor/` directory (which includes the source code of our upstream dependencies) - such that we'd need a way of filtering this out such that this wasn't the case (which I don't believe is possible via a `.gitignore` at this time?)

> without this change on a Windows machine when you execute go fmt it will see the CRLF as a violation and converts the *.go files back to LF this in turn is shown as a change in your local Git workspace

If you have make installed and run `make fmt` - I believe this should exclude the vendor directory? It should also be possible to prefix go fmt e.g. `go fmt -w ./azurerm`

Whilst I appreciate this isn't ideal - I don't believe there's any way to work around the checkout issue mentioned above?

PR #2019, opened 2018-10-04
Feature: Add 'ip_version' to Public IP Address

### LINKED ISSUE #412: Feature Req - Ability to create an IPv6 address
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

### Affected Resource(s)
- opc_instance
- opc_storage_volume

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

```hcl
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
What should have happened?

### Actual Behavior
What actually happened?

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- GH-1234

Krueladin:
Fixes #412

### DIFF
--- a/azurerm/resource_arm_public_ip_test.go
+++ b/azurerm/resource_arm_public_ip_test.go
@@ -71,6 +71,34 @@ func TestAccAzureRMPublicIpStatic_basic(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMPublicIpStatic_zones(t *testing.T) {
+	resourceName := "azurerm_public_ip.test"
+	ri := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMPublicIpDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMPublicIPStatic_withZone(ri, testLocation()),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMPublicIpExists(resourceName),
+					resource.TestCheckResourceAttrSet(resourceName, "ip_address"),
+					resource.TestCheckResourceAttr(resourceName, "public_ip_address_allocation", "static"),
+					resource.TestCheckResourceAttr(resourceName, "zones.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "zones.0", "1"),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
 func TestAccAzureRMPublicIpStatic_basic_withDNSLabel(t *testing.T) {
 	resourceName := "azurerm_public_ip.test"
 	ri := acctest.RandInt()
@@ -489,7 +517,7 @@ resource "azurerm_public_ip" "test" {
 `, rInt, location, rInt)
 }
 
-func testAccAzureRMPublicIPStatic_basic_withZone(rInt int, location string) string {
+func testAccAzureRMPublicIPStatic_withZone(rInt int, location string) string {
 	return fmt.Sprintf(`
 resource "azurerm_resource_group" "test" {
     name     = "acctestRG-%d"


Krueladin:
Currently added the ip_version as an optional argument to avoid breaking changes. Could be changed to be required.

azurerm/resource_arm_public_ip.go
```diff
@@ -122,6 +135,13 @@ func resourceArmPublicIpCreate(d *schema.ResourceData, meta interface{}) error {
 
 	idleTimeout := d.Get("idle_timeout_in_minutes").(int)
 	ipAllocationMethod := network.IPAllocationMethod(d.Get("public_ip_address_allocation").(string))
+	ipVersion := network.IPVersion(d.Get("ip_version").(string))
+
+	if strings.ToLower(string(ipVersion)) == "ipv6" {
+		if strings.ToLower(string(ipAllocationMethod)) == "static" {
+			return fmt.Errorf("Cannot specify publicIpAllocationMethod as Static for IPv6 PublicIp")
```
katbyte: This should be moved [to a CustomizeDiff](https://github.com/terraform-providers/terraform-provider-azurerm/blob/cadce839c18ec0d88c9152c029108e350d4892b7/azurerm/resource_arm_servicebus_namespace.go#L94)

azurerm/resource_arm_public_ip.go
```diff
@@ -209,6 +230,7 @@ func resourceArmPublicIpRead(d *schema.ResourceData, meta interface{}) error {
 	if location := resp.Location; location != nil {
 		d.Set("location", azureRMNormalizeLocation(*location))
 	}
+	d.Set("ip_version", strings.ToLower(string(resp.PublicIPAddressPropertiesFormat.PublicIPAddressVersion)))
```
katbyte: it appears you are setting this twice, the one below is the way to go

azurerm/resource_arm_public_ip.go
```diff
@@ -218,6 +240,7 @@ func resourceArmPublicIpRead(d *schema.ResourceData, meta interface{}) error {
 
 	if props := resp.PublicIPAddressPropertiesFormat; props != nil {
 		d.Set("public_ip_address_allocation", strings.ToLower(string(props.PublicIPAllocationMethod)))
+		d.Set("ip_version", strings.ToLower(string(props.PublicIPAddressVersion)))
```
katbyte: You appear to be setting this twice, this is the one in the correct place

azurerm/resource_arm_public_ip.go
tombuildsstuff: since this field could be interpolated - I'd suggest we don't do that in this instance

azurerm/data_source_public_ip.go
```diff
@@ -20,6 +20,11 @@ func dataSourceArmPublicIP() *schema.Resource {
 
 			"resource_group_name": resourceGroupNameForDataSourceSchema(),
 
+			"ip_version": {
+				Type:     schema.TypeString,
+				Optional: true,
```
tombuildsstuff: since this is a Data Source this needs to be `Computed` rather than `Optional` (since the information is retrieved but not setable)

azurerm/resource_arm_public_ip.go
```diff
@@ -60,6 +60,19 @@ func resourceArmPublicIp() *schema.Resource {
...
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  string(network.IPv4),
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(network.IPv4),
+					string(network.IPv6),
+				}, true),
+				DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+				StateFunc:        ignoreCaseStateFunc,
```
tombuildsstuff: since this is a new field, can we make this case-sensitive? (e.g. `true` -> `false` in the validate func, and remove the DiffSuppressFunc/StateFunc)

azurerm/resource_arm_public_ip.go
```diff
@@ -122,6 +135,13 @@ func resourceArmPublicIpCreate(d *schema.ResourceData, meta interface{}) error {
 
 	idleTimeout := d.Get("idle_timeout_in_minutes").(int)
 	ipAllocationMethod := network.IPAllocationMethod(d.Get("public_ip_address_allocation").(string))
+	ipVersion := network.IPVersion(d.Get("ip_version").(string))
+
+	if strings.ToLower(string(ipVersion)) == "ipv6" {
+		if strings.ToLower(string(ipAllocationMethod)) == "static" {
```
tombuildsstuff: _minor_ rather than lower-casing it we can use `strings.EqualsFold(string(ipAllocationMethod), "static")` here to use a case-invarient check

website/docs/d/public_ip.html.markdown
```diff
@@ -107,4 +107,5 @@ output "public_ip_address" {
 * `idle_timeout_in_minutes` - Specifies the timeout for the TCP idle connection.
 * `fqdn` - Fully qualified domain name of the A DNS record associated with the public IP. This is the concatenation of the domainNameLabel and the regionalized DNS zone.
 * `ip_address` - The IP address value that was allocated.
+* `ip_version` - The IP Version to use, IPv6 or IPv4. Must be used with a 'dynamic' allocation type.
```
tombuildsstuff: _minor_ since this is only retrieving information about the Public IP - can we update this to a slightly different tense (and remove the information about `dynamic` being required, since this is readonly):

```
* `ip_version` - The IP version being used, for example `IPv4` or `IPv6`
```

azurerm/resource_arm_public_ip.go
Krueladin: Am making the change. But I am curious, is there a particular reason for this preference?

tombuildsstuff: heh sorry I missed out using the Enum here - in general we'd tend to do something like:

```
strings.EqualsFold(string(ipAllocationMethod), network.Static)
```

although in this specific case (if we remove the case sensitivity) - we should actually be able to do a strict equality comparison I guess? 🤔

tombuildsstuff: @Krueladin sorry, that's a typo - I've updated the example

azurerm/resource_arm_public_ip.go
tombuildsstuff: the issue's to do with casting the string value to an Enum - if the casing doesn't match, Go defaults to the last value in the Enum, meaning the wrong value could be selected; as such we need to ensure the correct casing is used here (e.g. `IPv4` rather than `ipv4`)

Krueladin: I'm making the change, but I haven't seen this as an issue yet. wouldn't the only fear of this be in reading the data? And with the data object IPv6 is the last value, but the test still is able to read IPv4.  And there were many purposefully strangely cased examples in my tests. Where in particular could you see this coming from?

tombuildsstuff: indeed - it's possible for the casing to be different for example where resources are provisioned via an ARM Template/the CLI and then imported

azurerm/resource_arm_public_ip.go
```diff
@@ -60,6 +60,17 @@ func resourceArmPublicIp() *schema.Resource {
...
 
+			"ip_version": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  string(network.IPv4),
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(network.IPv4),
+					string(network.IPv6),
+				}, false),
```
tombuildsstuff: @Krueladin I hope you don't mind - since this PR otherwise LGTM I've pushed a commit to make this case-sensitive

tombuildsstuff:
hey @Krueladin 

I've run the tests and pushed a couple of commits (to revert my earlier change & to add back in a missing test, I hope you don't mind) - and this now LGTM:

```
acctests azurerm TestAccAzureRMPublicIpStatic_basic
=== RUN   TestAccAzureRMPublicIpStatic_basic
--- PASS: TestAccAzureRMPublicIpStatic_basic (119.63s)
=== RUN   TestAccAzureRMPublicIpStatic_basic_withDNSLabel
--- PASS: TestAccAzureRMPublicIpStatic_basic_withDNSLabel (110.90s)
=== RUN   TestAccAzureRMPublicIpStatic_basic_defaultsToIPv4
--- PASS: TestAccAzureRMPublicIpStatic_basic_defaultsToIPv4 (98.85s)
=== RUN   TestAccAzureRMPublicIpStatic_basic_withIPv4
--- PASS: TestAccAzureRMPublicIpStatic_basic_withIPv4 (111.90s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	442.395s
```

New Test:
```
$ acctests azurerm TestAccAzureRMPublicIpStatic_zones
=== RUN   TestAccAzureRMPublicIpStatic_zones
--- PASS: TestAccAzureRMPublicIpStatic_zones (111.60s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	113.048s
```

Since this LGTM I'll kick off the data source tests and merge this shortly 👍 

Thanks!

PR #2020, opened 2018-10-04
Upgrade API version to `2018-09-01` for ACI.

metacpp:
Upgrade API version for ACI to `2018-09-01` to support `Private` ip_address_type.


metacpp:
Close this PR temporarily coz it does have dependency on networkprofile.

Issue #2007, opened 2018-10-02
New Resource: `azurerm_monitor_metric_alert`


mmclane:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I can currently use terraform to create an azurerm_metric_alertrule and azurerm_monitor_action_group  separately.  But I can't create an azurerm_metric_alertrule  and point it at an action group (either the one created by terraform or an existing one).  This seems like a huge oversight.

### New or Affected Resource(s)

* azurerm_metric_alertrule 
* azurerm_monitor_action_group 

### Potential Terraform Configuration

```hcl
resource "azurerm_metric_alertrule" "test" {
  name = "${azurerm_virtual_machine.test.name}-cpu"
  resource_group_name = "${azurerm_resource_group.test.name}"
  location = "${azurerm_resource_group.test.location}"

  description = "An alert rule to watch the metric Percentage CPU"

  enabled = true

  resource_id = "${azurerm_virtual_machine.test.id}"
  metric_name = "Percentage CPU"
  operator = "GreaterThan"
  threshold = 75
  aggregation = "Average"
  period = "PT5M"

  email_action {
    send_to_service_owners = false
    custom_emails = [
      "some.user@example.com",
    ]
  }

  webhook_action {
    service_uri = "https://example.com/some-url"
      properties = {
        severity = "incredible"
        acceptance_test = "true"
      }
  }
}
resource "azurerm_resource_group" "test" {
  name     = "monitoring-resources"
  location = "West US"
}

resource "azurerm_monitor_action_group" "test" {
  name                = "CriticalAlertsAction"
  resource_group_name = "${azurerm_resource_group.test.name}"
  short_name          = "p0action"

  email_receiver {
    name          = "sendtoadmin"
    email_address = "admin@contoso.com"
  }

  email_receiver {
    name          = "sendtodevops"
    email_address = "devops@contoso.com"
  }

  sms_receiver {
    name         = "oncallmsg"
    country_code = "1"
    phone_number = "1231231234"
  }

  webhook_receiver {
    name        = "callmyapiaswell"
    service_uri = "http://example.com/alert"
[...]

PR #2026, opened 2018-10-05
New Resource: azurerm_monitor_metric_alert

JunyiYi:
#### Description

In this pull request, I implemented a new Azure monitoring resource in M4 called metric alerts. It will monitor the metrics (such as transactions count, CPU percentage, etc.) of one or more resources. Together with the [`azurerm_monitor_activitylog_alert`](https://github.com/terraform-providers/terraform-provider-azurerm/pull/1989), they are called the [unified alerts](https://docs.microsoft.com/en-us/azure/monitoring-and-diagnostics/monitoring-overview-unified-alerts) in Azure monitor.

#### Related Issues

This resource will realize the requirements specified in #2007  .

#### References

Resource Description: https://docs.microsoft.com/en-us/azure/monitoring-and-diagnostics/alert-metric-overview?toc=/azure/azure-monitor/toc.json
REST API Reference: https://docs.microsoft.com/en-us/rest/api/monitor/metricalerts/createorupdate

#### Remarks

This resource is different than [`azurerm_metric_alertrule`](https://www.terraform.io/docs/providers/azurerm/r/metric_alertrule.html) (which is called [classic alerts](https://docs.microsoft.com/en-us/azure/monitoring-and-diagnostics/monitoring-overview-alerts-classic?toc=/azure/azure-monitor/toc.json) in Azure). Actually this resource deprecates the old one.


azurerm/resource_arm_monitor_metric_alert.go
```diff
@@ -0,0 +1,450 @@
...
+		},
+	}
+}
+
+func resourceArmMonitorMetricAlertCreateOrUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).monitorMetricAlertsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	name := d.Get("name").(string)
+	resGroup := d.Get("resource_group_name").(string)
```
katbyte: Could we use `resourceGroup` rather then `resGroup` here?

azurerm/resource_arm_monitor_metric_alert.go
```diff
@@ -0,0 +1,450 @@
...
+
+func resourceArmMonitorMetricAlertRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).monitorMetricAlertsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	id, err := parseAzureResourceID(d.Id())
+	if err != nil {
+		return err
+	}
+	resGroup := id.ResourceGroup
```
katbyte: Could we use `resourceGroup` rather then `resGroup` here?

azurerm/resource_arm_monitor_metric_alert.go
```diff
@@ -0,0 +1,450 @@
...
+
+func resourceArmMonitorMetricAlertDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).monitorMetricAlertsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	id, err := parseAzureResourceID(d.Id())
+	if err != nil {
+		return err
+	}
+	resGroup := id.ResourceGroup
```
katbyte: same here

azurerm/resource_arm_monitor_metric_alert.go
```diff
@@ -0,0 +1,450 @@
...
+	if resp, err := client.Delete(ctx, resGroup, name); err != nil {
+		if !response.WasNotFound(resp.Response) {
+			return fmt.Errorf("Error deleting metric alert %q (resource group %q): %+v", name, resGroup, err)
+		}
+	}
+
+	return nil
+}
+
+func expandMonitorMetricAlertStringArray(v []interface{}) *[]string {
```
katbyte: This is a pretty common piece of code, might be work moving into utils?

azurerm/resource_arm_monitor_metric_alert.go
```diff
@@ -0,0 +1,450 @@
...
+		AllOf:     &criterias,
+		OdataType: insights.OdataTypeMicrosoftAzureMonitorSingleResourceMultipleMetricCriteria,
+	}
+}
+
+func expandMonitorMetricAlertAction(v []interface{}) *[]insights.MetricAlertAction {
+	actions := make([]insights.MetricAlertAction, 0)
+	for _, actionValue := range v {
+		val := actionValue.(map[string]interface{})
+
```
katbyte: might i suggest the following whitespace:

```go
val := actionValue.(map[string]interface{})
agID := val["action_group_id"].(string)

props := make(map[string]*string)
if propsValue, ok := val["webhook_properties"]; ok {
```

azurerm/resource_arm_monitor_metric_alert.go
```diff
@@ -0,0 +1,450 @@
...
+		actions = append(actions, insights.MetricAlertAction{
+			ActionGroupID:     utils.String(agID),
+			WebhookProperties: props,
+		})
+	}
+	return &actions
+}
+
+func flattenMonitorMetricAlertStringArray(v *[]string) []interface{} {
```
katbyte: Same comment as above. Might be useful to have this in `utils`

azurerm/resource_arm_monitor_metric_alert.go
```diff
@@ -0,0 +1,450 @@
...
+		for _, item := range *v {
+			result = append(result, item)
+		}
+	}
+	return result
+}
+
+func flattenMonitorMetricAlertCriteria(v insights.BasicMetricAlertCriteria) []interface{} {
+	result := make([]interface{}, 0)
+	if v != nil {
```
katbyte: how about we invert this?

```go
if v == nil {
    return result
}
```

azurerm/resource_arm_monitor_metric_alert.go
```diff
@@ -0,0 +1,450 @@
...
+		if val, ok := v.AsMetricAlertSingleResourceMultipleMetricCriteria(); ok && val != nil && val.AllOf != nil {
```
katbyte: same here:

how about we invert this?

```go
if val, ok := v.AsMetricAlertSingleResourceMultipleMetricCriteria(); !ok || val == nil || val.AllOf == nil {
    return result
}
```

azurerm/resource_arm_monitor_metric_alert.go
```diff
@@ -0,0 +1,450 @@
...
+				result = append(result, item)
+			}
+		}
+	}
+	return result
+}
+
+func flattenMonitorMetricAlertAction(v *[]insights.MetricAlertAction) []interface{} {
+	result := make([]interface{}, 0)
+	if v != nil {
```
katbyte: Could we invert this condition too?

katbyte (changes requested):
Hi @JunyiYi,

Thanks for the new resource, i've given it a quick look and left some minor comments inline. However it appears both the tests and documentation is missing? Once we have those added can do a more in-depth review and get this merged 🙂

JunyiYi:
Thanks @katbyte for the reviewing. Yes it is still in progress (so the title contains `WIP`). Considering this is a similar resource to [`azurerm_monitor_activitylog_alert`](https://github.com/terraform-providers/terraform-provider-azurerm/pull/1989), you might also provide comments in that PR as well. I will combine both to update the code.

azurerm/resource_arm_monitor_metric_alert.go
JunyiYi: I agree. But I'm not sure whether to put it into `utils` or `helpers`?

azurerm/resource_arm_monitor_metric_alert.go
JunyiYi: This comment is out of date.

katbyte (approved):
@JunyiYi,

LGTM 👍

Issue #1977, opened 2018-09-25
Ability to set max_delivery_count on azurerm_servicebus_queue


SLStinson:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

Terraform v0.11.8
+ provider.azurerm v1.15.0

### Affected Resource(s)

* azurerm_servicebus_queue

### Terraform Configuration Files

```hcl
resource "azurerm_servicebus_queue" "test_queue" {
  name                = "test_queue"
  resource_group_name = "${azurerm_resource_group.test.name}"
  namespace_name      = "${azurerm_servicebus_namespace.test.name}"

  enable_partitioning = true
  requires_duplicate_detection = true
  max_delivery_count = 10
}
```

### Expected Behavior

The maxDeliveryCount property is set to 10

### Actual Behavior

Error: azurerm_servicebus_queue.test_queue: : invalid or unknown key: max_delivery_count

### Steps to Reproduce

1. `terraform apply`

### References

https://docs.microsoft.com/en-us/azure/templates/microsoft.servicebus/namespaces/queues

tombuildsstuff:
hi @tkbky 

Since support for this has been merged and now released in v1.17 I'm going to close this issue for the moment.

Thanks!

PR #2028, opened 2018-10-05
service_bus_queue: support for `max_delivery_count`

tkbky:
This change adds `max_delivery_count` to the service bus queue's schema so that it is configurable.

See #1977

### DIFF
--- a/azurerm/resource_arm_servicebus_queue.go
+++ b/azurerm/resource_arm_servicebus_queue.go
@@ -241,6 +241,7 @@ func resourceArmServiceBusQueueRead(d *schema.ResourceData, meta interface{}) er
 		d.Set("requires_duplicate_detection", props.RequiresDuplicateDetection)
 		d.Set("requires_session", props.RequiresSession)
 		d.Set("dead_lettering_on_message_expiration", props.DeadLetteringOnMessageExpiration)
+		d.Set("max_delivery_count", props.MaxDeliveryCount)
 
 		if maxSizeMB := props.MaxSizeInMegabytes; maxSizeMB != nil {
 			maxSize := int(*maxSizeMB)
--- a/azurerm/resource_arm_servicebus_queue.go
+++ b/azurerm/resource_arm_servicebus_queue.go
@@ -119,9 +119,10 @@ func resourceArmServiceBusQueue() *schema.Resource {
 			},
 
 			"max_delivery_count": {
-				Type:     schema.TypeInt,
-				Optional: true,
-				Default:  10,
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      10,
+				ValidateFunc: validation.IntAtLeast(1),
 			},
 		},
 	}
--- a/azurerm/resource_arm_servicebus_queue.go
+++ b/azurerm/resource_arm_servicebus_queue.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/services/servicebus/mgmt/2017-04-01/servicebus"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_servicebus_queue_test.go
```diff
@@ -551,3 +578,26 @@ resource "azurerm_servicebus_queue" "test" {
...
+    resource_group_name = "${azurerm_resource_group.test.name}"
+    location = "${azurerm_resource_group.test.location}"
+    sku = "standard"
+}
+
+resource "azurerm_servicebus_queue" "test" {
+    name = "acctestservicebusqueue-%d"
+    resource_group_name = "${azurerm_resource_group.test.name}"
+    namespace_name = "${azurerm_servicebus_namespace.test.name}"
+    maxDeliveryCount = 20
```
tombuildsstuff: can we update this to `max_delivery_count` to match the schema field?

azurerm/resource_arm_servicebus_queue.go
```diff
@@ -117,6 +117,12 @@ func resourceArmServiceBusQueue() *schema.Resource {
 				Optional:   true,
 				Deprecated: "This field has been removed by Azure.",
 			},
+
+			"max_delivery_count": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  10,
```
tombuildsstuff: out of interest where does this default value come from? (and if so, are there any validation requirements for this field?)

azurerm/resource_arm_servicebus_queue_test.go
tkbky: Sure.

tkbky:
@tombuildsstuff Thanks for reviewing the PR. I've made the requested changes.

azurerm/resource_arm_servicebus_queue_test.go
```diff
@@ -254,6 +254,33 @@ func TestAccAzureRMServiceBusQueue_isoTimeSpanAttributes(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, "max_delivery_count", "10"),
+				),
+			},
+			{
+				Config: testAccAzureRMServiceBusQueue_maxDeliveryCount(ri, location),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "max_delivery_count", "20"),
+				),
+			},
+		},
```
katbyte: Could we add an import check step here?

```go
{
				ImportState:       true,
				ImportStateVerify: true,
			},
```

tkbky: Sure.

tkbky:
@katbyte It's my miss. Thanks for the catch. I've updated the PR to include the import check step, and include the max delivery count when read.

azurerm/resource_arm_servicebus_queue.go
katbyte: I was able to set it to `999999999` but it does have to be at least `1` so we could go with  `validation.IntAtleast(1)`?

tkbky: @katbyte yes, having it to be at least `1` makes sense to me.

azurerm/resource_arm_servicebus_queue_test.go
```diff
@@ -551,3 +583,26 @@ resource "azurerm_servicebus_queue" "test" {
 }
 `, rInt, location, rInt, rInt)
 }
+
+func testAccAzureRMServiceBusQueue_maxDeliveryCount(rInt int, location string) string {
+	return fmt.Sprintf(`
+resource "azurerm_resource_group" "test" {
+    name = "acctestRG-%d"
```
katbyte: Could we align all the assignments in the TF code to match the other tests?

tkbky: Definitely. Sorry for not noticing the inconsistency.

tombuildsstuff:
Tests pass:

screenshot 2018-10-10 at 14 28 31

PR #2032, opened 2018-10-05
Send provider version in user agent

paultyng:
This fixes user agents to conform to the RFC, it also takes advantage
of a new build flag populating the binary version of the provider.

The version sent in the user agent without the flag will be
`terraform-provider-azurerm/dev` and for acceptance tests through
`make testacc` `terraform-provider-azurerm/acc`. The binary should
send a product in the user agent in the form of
`terraform-provider-azurerm/1.2.3`.

### DIFF
--- a/vendor/github.com/hashicorp/terraform/httpclient/client.go
+++ b/vendor/github.com/hashicorp/terraform/httpclient/client.go
@@ -1,0 +1,18 @@
+package httpclient
+
+import (
+	"net/http"
+
+	cleanhttp "github.com/hashicorp/go-cleanhttp"
+)
+
+// New returns the DefaultPooledClient from the cleanhttp
+// package that will also send a Terraform User-Agent string.
+func New() *http.Client {
+	cli := cleanhttp.DefaultPooledClient()
+	cli.Transport = &userAgentRoundTripper{
+		userAgent: UserAgentString(),
+		inner:     cli.Transport,
+	}
+	return cli
+}
--- a/version/version.go
+++ b/version/version.go
@@ -1,0 +1,6 @@
+package version
+
+var (
+	// ProviderVersion is set during the release process to the release version of the binary
+	ProviderVersion = "dev"
+)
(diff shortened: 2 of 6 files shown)


paultyng:
Previous user agent strings would look like:

```
Go/go1.10.3 (amd64-windows) go-autorest/v10.15.4 Azure-SDK-For-Go/v20.1.0 scheduler/2016-03-01;HashiCorp-Terraform-v0.11.3
```

They will now look like:
```
Go/go1.10.3 (amd64-windows) go-autorest/v10.15.4 Azure-SDK-For-Go/v20.1.0 scheduler/2016-03-01 Terraform/v1.2.3 terraform-provider-azurerm/1.2.3
```

version/version.go
```diff
@@ -0,0 +1,6 @@
+package version
+
+var (
+	// ProviderVersion is set during the release process to the release version of the binary
+	ProviderVersion = "dev"
+)
```
metacpp: You may consider to use below popular way to popular version.go (it can be moved out of source control) during building time:
[https://github.com/git/git/blob/master/GIT-VERSION-GEN](https://github.com/git/git/blob/master/GIT-VERSION-GEN)

Its format looks like git-tag.lastcommitID, which contains enough information for us to track.

vendor/github.com/hashicorp/terraform/httpclient/client.go
```diff
@@ -0,0 +1,18 @@
...
+// New returns the DefaultPooledClient from the cleanhttp
+// package that will also send a Terraform User-Agent string.
+func New() *http.Client {
+	cli := cleanhttp.DefaultPooledClient()
+	cli.Transport = &userAgentRoundTripper{
+		userAgent: UserAgentString(),
+		inner:     cli.Transport,
+	}
+	return cli
+}
```
metacpp: Do we need this file? I don't see any reference to it.

PR #2056, opened 2018-10-11
Exposed ignore_missing_vnet_service_endpoint attribute

jeffreyCline:
Fixed the issue for the ignore_missing_vnet_service_endpoint attribute, turns out it's an async operation for Postgre SQL. Currently have many customers asking for this functionality.


azurerm/resource_arm_postgresql_virtual_network_rule.go
```diff
@@ -61,52 +65,21 @@ func resourceArmPostgreSQLVirtualNetworkRuleCreateUpdate(d *schema.ResourceData,
...
 		},
 	}
 
-	_, err = client.CreateOrUpdate(ctx, resourceGroup, serverName, name, parameters)
+	future, err := client.CreateOrUpdate(ctx, resourceGroup, serverName, name, parameters)
+	if err != nil {
+		return fmt.Errorf("Error submitting PostgreSQL Virtual Network Rule %q (PostgreSQL Server: %q, Resource Group: %q): %+v", name, serverName, resourceGroup, err)
+	}
+
+	err = future.WaitForCompletionRef(ctx, client.Client)
```
tombuildsstuff: where you referenced this is an async operation - does this call wait for the connection to be provisioned?

website/docs/r/postgresql_virtual_network_rule.html.markdown
```diff
@@ -84,7 +85,7 @@ The following arguments are supported:
 
 * `subnet_id` - (Required) The ID of the subnet that the PostgreSQL server will be connected to.
 
-~> **NOTE:** Due to [a bug in the Azure API](https://github.com/Azure/azure-rest-api-specs/issues/3719) this resource currently doesn't expose the `ignore_missing_vnet_service_endpoint` field and defaults this to `false`. Terraform will check during the provisioning of the Virtual Network Rule that the Subnet contains the Service Rule to verify that the Virtual Network Rule can be created.
+* `ignore_missing_vnet_service_endpoint` - (Optional) Create the virtual network rule before the subnet has the virtual network service endpoint enabled. The default value is false.
```
tombuildsstuff:
```
Create the virtual network rule before the subnet has the virtual network service endpoint enabled. The default value is false.
```

I think we could phrase this better to match the syntax of the other fields:

```
Should the Virtual Network Rule be created before the Subnet has the Virtual Network Service Endpoint enabled? Defaults to `false`
```

jeffreyCline: How about I copy it from the API docs directly?

```
Create firewall rule before the virtual network has vnet service endpoint enabled.
```

azurerm/resource_arm_postgresql_virtual_network_rule.go
jeffreyCline: No, it does not... it just makes sure the subnet exists else it throws the below error:

> * azurerm_postgresql_virtual_network_rule.test: Error creating PostgreSQL Virtual Network Rule "acctestpostgresqlvnetrule" (PostgreSQL Server: "acctestpostgresqlsvr", Resource Group: "testRG-vnet-rule"): Code="VirtualNetworkRuleResourceNotFound" Message="Resource not found: 'The Resource 'Microsoft.Network/virtualNetworks/acctestvnet' under resource group 'testRG-vnet-rule' was not found.'."

azurerm/resource_arm_postgresql_virtual_network_rule.go
```diff
@@ -44,6 +43,11 @@ func resourceArmPostgreSQLVirtualNetworkRule() *schema.Resource {
 				ValidateFunc: validation.NoZeroValues,
 			},
 
+			"ignore_missing_vnet_service_endpoint": {
```
JunyiYi: Please put all `Optional` attributes after `Required` attributes.

azurerm/resource_arm_postgresql_virtual_network_rule.go
```diff
@@ -166,6 +139,7 @@ func resourceArmPostgreSQLVirtualNetworkRuleRead(d *schema.ResourceData, meta in
 
 	if props := resp.VirtualNetworkRuleProperties; props != nil {
 		d.Set("subnet_id", props.VirtualNetworkSubnetID)
+		d.Set("ignore_missing_vnet_service_endpoints", props.IgnoreMissingVnetServiceEndpoint)
```
JunyiYi: `nil` check?

azurerm/resource_arm_postgresql_virtual_network_rule_test.go
```diff
@@ -457,21 +477,80 @@ resource "azurerm_postgresql_virtual_network_rule" "rule1" {
   name                = "acctestpostgresqlvnetrule1%d"
   resource_group_name = "${azurerm_resource_group.test.name}"
   server_name         = "${azurerm_postgresql_server.test.name}"
-  subnet_id           = "${azurerm_subnet.vnet1_subnet1.id}"
+	subnet_id           = "${azurerm_subnet.vnet1_subnet1.id}"
```
JunyiYi: Wrong indentation. Please use space instead of tab.

azurerm/resource_arm_postgresql_virtual_network_rule_test.go
```diff
@@ -111,6 +111,26 @@ func TestAccAzureRMPostgreSQLVirtualNetworkRule_multipleSubnets(t *testing.T) {
...
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMPostgreSQLVirtualNetworkRuleDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMPostgreSQLVirtualNetworkRuleExists(resourceName),
+				),
+			},
+		},
```
JunyiYi: Maybe Import test as well?

azurerm/resource_arm_postgresql_virtual_network_rule.go
```diff
@@ -165,7 +138,14 @@ func resourceArmPostgreSQLVirtualNetworkRuleRead(d *schema.ResourceData, meta in
...
 	if props := resp.VirtualNetworkRuleProperties; props != nil {
-		d.Set("subnet_id", props.VirtualNetworkSubnetID)
+
+		if virtualNetworkSubnetId := props.VirtualNetworkSubnetID; virtualNetworkSubnetId != nil {
+			d.Set("subnet_id", virtualNetworkSubnetId)
+		}
+
+		if ignoreMissingVnetServiceEndpoint := props.IgnoreMissingVnetServiceEndpoint; ignoreMissingVnetServiceEndpoint != nil {
+			d.Set("ignore_missing_vnet_service_endpoint", ignoreMissingVnetServiceEndpoint)
+		}
```
tombuildsstuff: we don't need to nil-check either of these - d.Set will handle that for us (since this is a simple type) - can we revert this?

website/docs/r/postgresql_virtual_network_rule.html.markdown
tombuildsstuff: Sure - but we should still match the language used across the other resources, which prefixes things with `Should [xxx] be applied/enabled` etc?

azurerm/resource_arm_postgresql_virtual_network_rule.go
jeffreyCline: Yes, I can revert it no problem, but I did it cause in the SDK struct they are pointers and I was afraid that they may come back `nil` if something went wrong on the `read`

```
// VirtualNetworkSubnetID - The ARM resource id of the virtual network subnet.
VirtualNetworkSubnetID *string `json:"virtualNetworkSubnetId,omitempty"`

// IgnoreMissingVnetServiceEndpoint - Create firewall rule before the virtual network has vnet service endpoint enabled.
IgnoreMissingVnetServiceEndpoint *bool `json:"ignoreMissingVnetServiceEndpoint,omitempty"`
```

jeffreyCline:
image

azurerm/resource_arm_postgresql_virtual_network_rule.go
katbyte: `d.Set` will do the nil check itself

PR #2073, opened 2018-10-13
New Resource: api_management_api

torresdal:
Added next resource in the API Management family as requested in #1177 

For now this resource is quite basic with main focus on importing API's through swagger/wsdl. Remaining future options for this resource would typically depend on other not yet developed resources which is in the pipeline (like Authorization and Product).

**Note:** In API Management API the concepts of Revisions and Versions are supported, but I'm not sure those concepts belong in this Terraform resource, since they are more deployment concepts and not infrastructure. Because of this I've omitted them for now, except as computed values. If you agree with this we should probably say something about that in the docs.


website/docs/r/api_management_api.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+  location            = "${azurerm_resource_group.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the API Management API.
```
tombuildsstuff: since this is ForceNew, can we suffix this with: `Changing this forces a new resource to be created.`

website/docs/r/api_management_api.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+
+* `resource_group_name` - (Required) The Name of the Resource Group where the API Management API exists.
```
tombuildsstuff: since this is ForceNew, can we suffix this with: `Changing this forces a new resource to be created.`

website/docs/r/api_management_api.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+
+* `location` - (Required) The Azure location where the API Management API exists.
+
+* `service_name` - (Required) The Name of the API Management Service where the API Management API exists
```
tombuildsstuff: since this is ForceNew, can we suffix this with: `Changing this forces a new resource to be created.`

website/docs/r/api_management_api.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the API Management API.
+
+* `resource_group_name` - (Required) The Name of the Resource Group where the API Management API exists.
+
+* `location` - (Required) The Azure location where the API Management API exists.
```
tombuildsstuff: I think this can be removed?

website/docs/r/api_management_api.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+
+* `content_format` - (Required) Format of the Content in which the API is getting imported. Possible values include: 'swagger-json', 'swagger-link-json', 'wadl-link-json', 'wadl-xml', 'wsdl', 'wsdl-link'.
+
+* `content_value` - (Required) Content value when Importing an API. When a `*-link-*` `content_format` is used, the `content_value` must be a URL. If not, `content_value` is defined inline.
+
+* `wsdl_selector` - (Optional) Criteria to limit import of WSDL to a subset of the document. Only applicable to content with format `wsdl` or `wsdl-link`. The `wsdl_selector` block is documented below.
+
+---
+
+`wsdl_selector` block supports the following:
```
tombuildsstuff: _minor_ can we prefix this with `A`:

```suggestion
A `wsdl_selector` block supports the following:
```

tombuildsstuff: _minor_ can we prefix this with `A`
```suggestion
A `wsdl_selector` block supports the following:
```

website/docs/r/api_management_api.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+
+A `subscription_key_parameter_names` block exports the following:
+
+* `header` - (Optional) Subscription key header name. Default is `Ocp-Apim-Subscription-Key`.
+
+* `query` - (Optional) Subscription key query string parameter name. Default is `subscription-key`.
+
+---
+
+`import` block supports the following:
```
tombuildsstuff: _minor_ can we prefix this with `A`:

```suggestion
A `import` block supports the following:
```

website/docs/r/api_management_api.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+
+* `subscription_key_parameter_names` - (Optional) A `subscription_key` block as documented below.
+
+* `import` - (Optional) A `import` block as documented below.
+
+* `soap_pass_through` - Make API Management expose a SOAP front end, instead of a HTTP front end.
+
+---
+
+A `subscription_key_parameter_names` block exports the following:
```
tombuildsstuff: whilst this is Computed (and thus technically exported) - it's optional/computed, as such can we update this to match the other blocks:

```suggestion
A `subscription_key_parameter_names` block supports the following:
```

tombuildsstuff: whilst this is Computed (and thus technically exported) - it's optional/computed, as such can we update this to match the other blocks:

```suggestion
A `subscription_key_parameter_names` block supports the following:
```

website/docs/r/api_management_api.html.markdown
```diff
@@ -0,0 +1,124 @@
...
+
+* `header` - (Optional) Subscription key header name. Default is `Ocp-Apim-Subscription-Key`.
+
+* `query` - (Optional) Subscription key query string parameter name. Default is `subscription-key`.
```
tombuildsstuff: both of these defaults aren't set in Terraform - so could we remove them for the moment?

azurerm/resource_arm_api_management_api.go
```diff
@@ -0,0 +1,432 @@
...
+		d.Set("service_url", props.ServiceURL)
+		d.Set("path", props.Path)
+		d.Set("description", props.Description)
+		d.Set("revision", props.APIRevision)
+		d.Set("version", props.APIVersion)
+		d.Set("version_set_id", props.APIVersionSetID)
+		d.Set("is_current", props.IsCurrent)
+		d.Set("is_online", props.IsOnline)
+		d.Set("protocols", props.Protocols)
+		d.Set("soap_pass_through", string(props.APIType) == string(apimanagement.SoapPassThrough))
```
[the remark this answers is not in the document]
torresdal: To me the APIType in the Azure API is very confusing. The way I understand it, it IS actually only applicable to SOAP. Setting APIType for a REST API is not valid. If you check out the docs here: https://docs.microsoft.com/en-us/rest/api/apimanagement/api/createorupdate#soapapitype

...there are only two options: 

* `http` creates a SOAP to REST API
* `soap` creates a SOAP pass-through API

If you try to import a SOAP API and don't provide APIType, it believe it defaults to `http`. That's why I opted for a simpler and more descriptive option, where we use `soap_pass_through` instead (which is more accurate in my opinion). However, if this is just more confusing for the end user, I agree we should change it back to `api_type` and explain the details in the docs. My preference is to keep it though.

website/azurerm.erb
```diff
@@ -240,6 +244,9 @@
                   <a href="/docs/providers/azurerm/r/api_management.html">azurerm_api_management</a>
                 </li>
 
+                <li<%= sidebar_current("docs-azurerm-datasource-api-management-api-x") %>>
```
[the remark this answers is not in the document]
torresdal: Ahh.... That's what the `x` is for :-) But there will be resources below `-api-`, like `api-management-api-policy`, but using the `-x` was pure luck in this case :-)

website/docs/d/api_management_api.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+
+* `is_online` - Indicates if API revision is accessible via the gateway.
+
+---
+
+A `subscription_key_parameter_names` block exports the following:
+
+* `header` - Subscription key header name. Default is `Ocp-Apim-Subscription-Key`.
+
+* `query` - Subscription key query string parameter name. Default is `subscription-key`.
```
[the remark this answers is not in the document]
torresdal: The default is set on the Azure side. So a unset value will return these values on read. How do you usually handle/document that?

website/docs/r/api_management_api.html.markdown
torresdal: The default is set on the Azure side. So a unset value will return these values on read. How do you usually handle/document that?

azurerm/data_source_api_management_api.go
```diff
@@ -0,0 +1,149 @@
...
+		d.Set("service_url", props.ServiceURL)
+		d.Set("path", props.Path)
+		d.Set("description", props.Description)
+		d.Set("revision", props.APIRevision)
+		d.Set("version", props.APIVersion)
+		d.Set("version_set_id", props.APIVersionSetID)
+		d.Set("is_current", props.IsCurrent)
+		d.Set("is_online", props.IsOnline)
+		d.Set("protocols", props.Protocols)
+		d.Set("soap_pass_through", string(props.APIType) == string(apimanagement.SoapPassThrough))
```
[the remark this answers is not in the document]
torresdal: see my question/description below for the same field in the resource

azurerm/resource_arm_api_management_api.go
```diff
@@ -0,0 +1,432 @@
...
+	client := meta.(*ArmClient).apiManagementApiClient
+	ctx := meta.(*ArmClient).StopContext
+
+	log.Printf("[INFO] preparing arguments for AzureRM API Management API creation.")
+
+	resGroup := d.Get("resource_group_name").(string)
+	serviceName := d.Get("service_name").(string)
+	name := d.Get("name").(string)
+
+	//Currently we don't support revisions, so we use 1 as default
```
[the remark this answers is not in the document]
torresdal: Agree

azurerm/resource_arm_api_management_api.go
```diff
@@ -0,0 +1,432 @@
...
+	ctx := meta.(*ArmClient).StopContext
+
+	id, err := parseAzureResourceID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	resGroup := id.ResourceGroup
+	serviceName := id.Path["service"]
+	apiid := id.Path["apis"]
```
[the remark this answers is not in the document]
torresdal: I don't think so. The revision should be part of the id. According to the azure docs (https://docs.microsoft.com/en-us/rest/api/apimanagement/api/delete) the apiId is described as:

> Must be unique in the current API Management service instance. Non-current revision has ;rev=n as a suffix where n is the revision number.

azurerm/resource_arm_api_management_api.go
```diff
@@ -0,0 +1,432 @@
...
+		Description:                   &description,
+		DisplayName:                   &displayName,
+		Path:                          &path,
+		Protocols:                     expandApiManagementApiProtocols(d),
+		ServiceURL:                    &serviceUrl,
+		SubscriptionKeyParameterNames: expandApiManagementApiSubscriptionKeyParamNames(d),
+	}
+
+	if soapPassThrough {
+		props.APIType = apimanagement.APIType(apimanagement.SoapPassThrough)
```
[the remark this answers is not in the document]
torresdal: No :-) 

Would this be better?

```go
if v, ok := d.GetOk("soap_pass_through"); ok {
  soapPassThrough := v.(bool)

  if soapPassThrough {
    props.APIType = apimanagement.APIType(apimanagement.SoapPassThrough)
  } else {
    props.APIType = apimanagement.APIType(apimanagement.SoapToRest)
  }
}
```

azurerm/resource_arm_api_management_api.go
```diff
@@ -0,0 +1,432 @@
...
+				Elem: &schema.Schema{
+					Type:             schema.TypeString,
+					DiffSuppressFunc: ignoreCaseDiffSuppressFunc,
+					ValidateFunc: validation.StringInSlice([]string{
+						string(apimanagement.ProtocolHTTP),
+						string(apimanagement.ProtocolHTTPS),
+					}, true),
+				},
+				Optional: true,
+				Computed: true, // Azure API sets protocols to https by default
```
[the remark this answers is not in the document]
torresdal: Agree

azurerm/resource_arm_api_management_api.go
```diff
@@ -0,0 +1,432 @@
...
+	//Currently we don't support revisions, so we use 1 as default
+	apiId := fmt.Sprintf("%s;rev=%d", name, 1)
+	d.Set("api_id", apiId)
+
+	var properties *apimanagement.APICreateOrUpdateProperties
+
+	_, hasImport := d.GetOk("import")
+
+	// If import is used, we need to send properties to Azure API in two operations.
+	// First we execute import and then updated the other props.
```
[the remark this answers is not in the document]
torresdal: Agree with the simplification, not sure about `IsNewResource`... It could be my understanding of how TF manages this, but at least in Azure the resource (API) can change schema without recreating the resource (.e.g. you can import a new Swagger file to an existing API). My understanding is then that this should be able to import multiple times.

azurerm/resource_arm_api_management_api.go
```diff
@@ -0,0 +1,432 @@
...
+
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"import": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
```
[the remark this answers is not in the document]
torresdal: I believe in Azure the resource (API) can change schema without recreating the resource. E.g. you can import a new Swagger file to an existing API, without the resource being recreated. I believe the answer is no.

torresdal:
@tombuildsstuff I've resolved the ones that I've made code changes to, but left the ones I've commented on un-resolved. Could you please have a look at my comments, so we can find a solution to those, and then I'll push my changes?

torresdal:
Just pushed the changes we did so far as a result of the review. Only thing missing now is to have a look at the unresolved items in the review and agree on a solution for each.

website/docs/r/api_management_api.html.markdown
tombuildsstuff: since we're not setting them (and the API could change it's default result in time, as has happened in the past) - we tend not to set/document these

azurerm/resource_arm_api_management_api.go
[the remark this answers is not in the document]
tombuildsstuff: in which case the `IsNewResource` isn't needed (I was assuming that the schema for a given API versions was immutable/required a version bump)

torresdal:
@tombuildsstuff Anything else I should do or we should agree on before moving this on?

Issue #1873, opened 2018-09-05
export primary and secondary keys of azurerm_search_service


DimitriTchapmi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When you build an azurerm_search_service with terraform, it just export you the id attribute. It will be great if we could get primary and secondary key in addition. This will be so usefull in many cases, for example when you need to automate tasks with Rest API. I'm trying to output those keys and use them in ansible tasks.

### New or Affected Resource(s)

* azurerm_search_service

### Potential Terraform Configuration

```hcl
output "search_pri_key" {
value = "${azurerm_search_service.mysearch.primary_key}"
}
```

DimitriTchapmi:
For example
output "search_key" {
  value = "${azurerm_search_service.mysearch.primary_key}"
}

tombuildsstuff:
hi @DimitriTchapmi 

Just to let you know that support for this shipped in v1.17 - you can upgrade to this by updating the version in the Provider block (as shown below) and then running `terraform init -upgrade`:

```
provider "azurerm" {
  version = "=1.17.0"
}
```

Thanks!

DimitriTchapmi:
Thanks a lot

PR #2074, opened 2018-10-13
Azure Search: Add primary and secondary keys as exported attributes

cmpxchg16b:
to address issue #1873

### DIFF
--- a/azurerm/resource_arm_search_service.go
+++ b/azurerm/resource_arm_search_service.go
@@ -59,6 +59,16 @@ func resourceArmSearchService() *schema.Resource {
 			},
 
 			"tags": tagsForceNewSchema(),
+
+			"primary_key": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"secondary_key": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_search_service.go
```diff
@@ -59,6 +59,16 @@ func resourceArmSearchService() *schema.Resource {
 			},
 
 			"tags": tagsForceNewSchema(),
```
katbyte: Could we please put the `tags` Schema at the bottom after these new values?

azurerm/resource_arm_search_service.go
```diff
@@ -149,6 +159,18 @@ func resourceArmSearchServiceRead(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
+	adminKeysClient := meta.(*ArmClient).searchAdminKeysClient
+	adminKeysResp, err := adminKeysClient.Get(ctx, resourceGroup, name, nil)
+	if err == nil {
+		if primaryKey := adminKeysResp.PrimaryKey; primaryKey != nil {
+			d.Set("primary_key", primaryKey)
```
katbyte: `d.Set()` will handle nil values just fine so we can drop the nil tests here.

cmpxchg16b:
@katbyte thanks for review!

PR #2077, opened 2018-10-14
Convert all names entered for Azure DNS resources to lowercase by default. 

### LINKED ISSUE #1229: DNS Zone and Records are always recreated
### Community Note

When i define a DNS Zone or a Record Terraform always tries to recreate it. There seems to be a problem with case sensitive in the Names of the Resources.
image

draggeta:
This PR converts all names entered for Azure DNS resources to lowercase by default. 

I've encountered a few colleagues with the same issue as the OP. While telling them to use only lowercase is possible, the fact remains that Azure is case insensitive and stores everything in lowercase.

Fixes #1229

### DIFF
--- a/azurerm/resource_arm_dns_a_record.go
+++ b/azurerm/resource_arm_dns_a_record.go
@@ -3,6 +3,7 @@ package azurerm
 import (
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/preview/dns/mgmt/2018-03-01-preview/dns"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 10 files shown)


tombuildsstuff:
hi @draggeta 

Thanks for this PR :)

As I mentioned in #1229 - whilst I appreciate this PR would allow existing DNS Resources to be used in Terraform without matching the case, this is actually a bug in the AzureRM Provider where the Azure API's are (mostly) Upserts which allows existing resources to be modified without importing them into the state file. We're planning on fixing this in #1746 which means that going forward existing resources would need to be imported via `terraform import` before being used (and which requires the config matches the state of the resource in Azure).

Whilst the `name` field isn't /necessarily/ case-sensitive at this point in time, that's not true across all the Azure API's (or within a version of an API) - such that I believe it's best this field remains case-sensitive for the moment (and that instead the Terraform Configurations should be updated to ensure the casings are correct, which matches the behaviour of the other Terraform Providers).

One thing of note: in the next major version of the AzureRM Provider (2.0) - we're planning on making existing fields (e.g. Enum's/Constants) case-sensitive where possible to match the behaviour of the other Terraform Providers - and as such I believe we'd end up reverting this PR in the not-too-distant future to match that behaviour unfortunately.

Given this - unfortunately we're unable to merge this PR at this point in time and as such I'm going to close this PR for the moment - but I'd like to thank you for this contribution :)

Thanks!

PR #2079, opened 2018-10-15
New Resources: `azurerm_network_interface_(backend_address_pool|nat_rule)_association`

### LINKED ISSUE #880: Feature request: Support independent NIC ↔ load balancer backend pool association resource
This is an aspirational feature request; I totally understand that this is probably non-trivial to implement and might not be a high priority. I just wanted to record idea here so that other people with a similar need can refer to and subscribe to a particular GitHub issue.

I would love to be able to associate a network interface to a load balancer's backend pool in a (synthetic?) resource separate from the main resource defining the network interface.

Our setup involves running a cluster of larger VMs and colocating different services on different subsets of those VMs. I'd like to be able to define the VMs in one module dedicated to setting up the basic infrastructure of the underlying VMs, but define a load balancer passthrough in a different service-specific module (which might also stand up other resources needed by a specific service, like storage containers).

Right now, if I define the load balancer in the main VMs module, I can independently create backend pools, probes, and load balancer rules in the service-specific module. However, I can't associate network interfaces with the backend pool without editing load_balancer_backend_address_pools_ids directly on the azurerm_network_interface resource.

It would be really great if there were something like a azurerm_lb_backend_address_pool_association resource that could make that association between an existing backend pool and an existing network interface without having to edit either Terraform resource.

Note that the Azure portal does allow making this kind of association as a separate operation (accessible from the editing interface of the load balancer's backend address pool).

tombuildsstuff:
This PR adds a few Virtual Resources for managing the relationship between an Application Gateway/Load Balancer and a Network Interface:

* `azurerm_network_interface_application_gateway_backend_address_pool_association`
* `azurerm_network_interface_backend_address_pool_association`
* `azurerm_network_interface_nat_rule_association` 

Since this is a Virtual Resource (and there's an existing field for managing this on the Network Interface resource) - that'll need to be deprecated as a part of 2.0

Fixes #880


azurerm/provider.go
```diff
@@ -129,168 +129,170 @@ func Provider() terraform.ResourceProvider {
...
+			"azurerm_metric_alertrule":                                   resourceArmMetricAlertRule(),
+			"azurerm_monitor_action_group":                               resourceArmMonitorActionGroup(),
+			"azurerm_mysql_configuration":                                resourceArmMySQLConfiguration(),
+			"azurerm_mysql_database":                                     resourceArmMySqlDatabase(),
+			"azurerm_mysql_firewall_rule":                                resourceArmMySqlFirewallRule(),
+			"azurerm_mysql_server":                                       resourceArmMySqlServer(),
+			"azurerm_mysql_virtual_network_rule":                         resourceArmMySqlVirtualNetworkRule(),
+			"azurerm_network_interface":                                  resourceArmNetworkInterface(),
+			"azurerm_network_interface_backend_address_pool_association": resourceArmNetworkInterfaceBackendAddressPoolAssociation(),
+			"azurerm_network_interface_nat_rule_association":             resourceArmNetworkInterfaceNatRuleAssociation(),
```
tombuildsstuff: @katbyte given we're doing a resource for application gateway associations too, I think this needs `lb_` between `network_interface` and `backend/nat`- WDYT?

tombuildsstuff:
Tests pass for this (without the new Application Gateway resource) - ignoring a known test failure:

screenshot 2018-10-14 at 23 17 46

PR #2091, opened 2018-10-16
Update VMSS configs to reflect network_profile.*.ip_configuration.*.primary now being required

katbyte:
That change in #2035 was not reflected in some tests configurations, this fixes them.


katbyte:
Tests pass:
screen shot 2018-10-16 at 03 52 17

PR #2099, opened 2018-10-16
Use enum string instead and update documentation.

metacpp:
1. Use enum strings for `protocol` and `os_type`.
2. Add missed documentation for `protocol`.
3. Update test case for `os_type`.


metacpp:
ACI tests passed:
image

PR #2121, opened 2018-10-20
vendor house keeping

katbyte:
- remove unused azure DNS package
- remove =s from all package versions
- set httpclient version to 0.11.9 (first version it is available"


metacpp:
@katbyte  LGTM.

vendor/vendor.json
```diff
@@ -984,227 +984,231 @@
...
+			"version": "v0.11.3",
 			"versionExact": "v0.11.3"
 		},
 		{
 			"checksumSHA1": "R43En+SNnXiSxTUecrW58Ohprmc=",
 			"path": "github.com/hashicorp/terraform/version",
 			"revision": "3802b14260603f90c7a1faf55994dcc8933e2069",
 			"revisionTime": "2018-01-31T20:48:39Z",
-			"version": "=v0.11.3",
+			"version": "v0.11.3",
```
tombuildsstuff: we should pin these to the same version of TFCore (0.11.9)

PR #2128, opened 2018-10-22
Add eku to azurerm_key_vault_certificate

### LINKED ISSUE #1319: Feature Request:  Support enhanced key usage (eku) for keyvault certificate creation
### Community Note

### Description

When creating a new certificate with the azurerm_keyvault_certificate we would like to specify enhanced key usage (eku) properties for the certificate.  

* azurerm_key_vault_certificate

### References

* https://docs.microsoft.com/en-us/rest/api/keyvault/createcertificate/createcertificate#x509certificateproperties

draggeta:
This PR adds the Extended Key Usage argument to the azurerm_key_vault_certificate resource. 

This is the current behaviour:
- If the block isn't specified, it automatically adds the server and client authentication EKUs.
- If something is specified (or an empty list), it changes the resource to match what is in the block.
- If the block is removed, it doesn't revert to the default EKUs.

Two things I'm not sure about:
- I'm not sure if and how the last item in the list it should be fixed.
- Also, the standard says it should be Extended Key Usage [(RFC 5280)](https://tools.ietf.org/html/rfc5280#section-4.2.1.12) but Microsoft uses Enhanced Key Usage. For now I used Extended, but a case could be made to use the MS terminology as that is what can be found in the documentation for the API.

If anyone has feedback, it would be very welcome.

Test results:
```
PS> go test -timeout 180m -v -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.Prers/terraform-provider-azurerm/version.ProviderVersion=acc" github.com\terraform-providers\terraform-provider-azurerm\azurerm -run ^TestAccAzureRMKeyVaultCertificate_basicExtendedKeyUsage$
=== RUN   TestAccAzureRMKeyVaultCertificate_basicExtendedKeyUsage
--- PASS: TestAccAzureRMKeyVaultCertificate_basicExtendedKeyUsage (232.15s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm       236.905s    
```

Closes #1319


azurerm/resource_arm_key_vault_certificate.go
```diff
@@ -456,6 +465,14 @@ func expandKeyVaultCertificatePolicy(d *schema.ResourceData) keyvault.Certificat
 	for _, v := range certificateProperties {
 		cert := v.(map[string]interface{})
 
+		extendedKeyUsage := make([]string, 0)
```
katbyte: There is a util function `utils.ExpandStringArray` that you could use here instead

azurerm/resource_arm_key_vault_certificate.go
```diff
@@ -537,6 +555,7 @@ func flattenKeyVaultCertificatePolicy(input *keyvault.CertificatePolicy) []inter
 			usages = append(usages, string(usage))
 		}
 
+		certProps["extended_key_usage"] = props.Ekus
```
katbyte: We should probably check if this is nil

azurerm/resource_arm_key_vault_certificate.go
draggeta: Ah, didn't see that one. I had seen the ones for strings, ints etc. It is fixed now :)

azurerm/resource_arm_key_vault_certificate.go
draggeta: Done!

draggeta:
@katbyte I've added the acceptance test and tested these.

tombuildsstuff:
hi @draggeta 

Just to let you know that this has been released as a part of v1.18 of the AzureRM Provider ([the full changelog is available here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/v1.18.0/CHANGELOG.md)). You can upgrade to this by specifying the version in the provider block (as shown below) and then running `terraform init -upgrade`

```
provider "azurerm" {
  version = "=1.18.0"
}
```

Thanks!

PR #2131, opened 2018-10-22
Stripping the Authorization Header from the Logs

tombuildsstuff:
This PR ensures that the active Authorization header isn't printed out to the local debug logs (when `TF_LOG` is specified) by stripping it from the request prior to printing and then adding it back in when printing's completed


tombuildsstuff:
Travis CI isn't triggering as Github is still partially down - since this builds locally I'm going to merge this anyway

PR #2171, opened 2018-10-29
Pass `nil` to `os_disk_size_gb` of AKS

metacpp:
AKS team requires to pass `nil` instead of 0 if user does not define the size of os disk.

### DIFF
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -759,23 +759,21 @@ func expandAzureRmKubernetesClusterAgentProfiles(d *schema.ResourceData) []conta
 	name := config["name"].(string)
 	count := int32(config["count"].(int))
 	vmSize := config["vm_size"].(string)
-	osType := config["os_type"].(string)
-
-	// If this value is not provided, the default value read from configuration will be 0.
-	// AKS API prefers to accept nil pointer instead of 0.
 	osDiskSizeGB := int32(config["os_disk_size_gb"].(int))
-	var osDiskSizeGBPtr *int32 = nil
-	if osDiskSizeGB > 0 {
-		osDiskSizeGBPtr = utils.Int32(osDiskSizeGB)
-	}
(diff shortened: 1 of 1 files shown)


azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -759,14 +759,21 @@ func expandAzureRmKubernetesClusterAgentProfiles(d *schema.ResourceData) []conta
...
 	count := int32(config["count"].(int))
 	vmSize := config["vm_size"].(string)
-	osDiskSizeGB := int32(config["os_disk_size_gb"].(int))
 	osType := config["os_type"].(string)
 
+	// If this value is not provided, the default value read from configuration will be 0.
+	// AKS API prefers to accept nil pointer instead of 0.
+	osDiskSizeGB := int32(config["os_disk_size_gb"].(int))
+	var osDiskSizeGBPtr *int32 = nil
+	if osDiskSizeGB > 0 {
```
katbyte: A simpler (and more common pattern in the provider) way of doing this is after the profile def:

```go
if osDiskSizeGB > 0 {
    profile.OsDiskSizeGB = osDiskSizeGB
}
```

as profile.OsDiskSizeGB will be initialized to `nil`

metacpp: Very good suggestion and I've updated it.

metacpp:
All AKS tests passed:
screen shot 2018-10-28 at 11 49 25 pm

tombuildsstuff:
hi @metacpp 

Unfortunately this PR appears to be enabling a breaking behavioural change in the AKS API, which will in turn potentially break versions of Terraform (and other API clients) which have already shipped. Rather than taking this approach - a new version of the API should be created with this breaking change - which will allow the existing API to continue functioning as it does today, which means that existing users of this API will be able to continue using it.

As such rather than merging this and risk breaking AKS for all existing Terraform users - I'm going to close this PR for the moment - however once a new version of the API is out we can look towards moving to that.

Thanks!

PR #2086, opened 2018-10-16
New resource: DevSpace Controller for AKS cluster

metacpp:
This PR is a follow up on #1488 to add DevSpace controller resource for AKS cluster.

PR #2182, opened 2018-10-29
[Refactoring] Devspace Controller

metacpp:
Follow up the suggestions in #2086 by @tombuildsstuff .


azurerm/resource_arm_devspace_controller.go
```diff
@@ -171,21 +171,19 @@ func resourceArmDevSpaceControllerRead(d *schema.ResourceData, meta interface{})
 		d.Set("location", azureRMNormalizeLocation(*location))
 	}
 
-	d.Set("sku", flattenDevSpaceControllerSku(result.Sku))
+	if err := d.Set("sku", flattenDevSpaceControllerSku(result.Sku)); err != nil {
+		return fmt.Errorf("Error flattenning Sku for DevSpace Controller %q (Resource Group %q): %+v", name, resGroupName, err)
```
tombuildsstuff: this would be failing to set the schema field, so we should make this:

```suggestion
		return fmt.Errorf("Error flattening`sku`: %+v", err)
```

azurerm/resource_arm_devspace_controller.go
```diff
@@ -171,21 +171,19 @@ func resourceArmDevSpaceControllerRead(d *schema.ResourceData, meta interface{})
...
-		if props.TargetContainerHostResourceID != nil {
-			d.Set("target_container_host_resource_id", props.TargetContainerHostResourceID)
-		}
+		hostSuffix = props.HostSuffix
+		dataPlaneFqdn = props.DataPlaneFqdn
+		targetContainerHostResourceID = props.TargetContainerHostResourceID
 	}
+	d.Set("host_suffix", hostSuffix)
+	d.Set("data_plane_fqdn", dataPlaneFqdn)
+	d.Set("target_container_host_resource_id", targetContainerHostResourceID)
```
tombuildsstuff: we can move these inside the `properties` block above rather than conditionally setting these; since these'll get caught by the import tests if they're unset

website/docs/r/devspace_controller.html.markdown
```diff
@@ -80,20 +80,20 @@ The following arguments are supported:
 
 A `sku` block supports the following:
 
-* `name` - (Required) The name of the SKU for Azure DevSpace Controller. Currently supported value is `S1`. Changing this forces a new resource to be created.
-* `tier` - (Required) The tier of the SKU for Azure DevSpace Controller. Currently supported value is `Standard`. Changing this forces a new resource to be created.
+* `name` - (Required) The name of the SKU for DevSpace Controller. Currently supported value is `S1`. Changing this forces a new resource to be created.
```
tombuildsstuff:
```suggestion
* `name` - (Required) The name of the SKU for DevSpace Controller. Currently the only supported value is `S1`. Changing this forces a new resource to be created.
```

website/docs/r/devspace_controller.html.markdown
```diff
@@ -80,20 +80,20 @@ The following arguments are supported:
...
+* `tier` - (Required) The tier of the SKU for DevSpace Controller. Currently supported value is `Standard`. Changing this forces a new resource to be created.
```
tombuildsstuff:
```suggestion
* `tier` - (Required) The tier of the SKU for DevSpace Controller. Currently the only supported value is `Standard`. Changing this forces a new resource to be created.
```

website/docs/r/devspace_controller.html.markdown
```diff
@@ -80,20 +80,20 @@ The following arguments are supported:
...
 
-* `id` - The ID of the Azure DevSpace Controller.
+* `id` - The ID of the DevSpace Controller.
 
 * `data_plane_fqdn` - DNS name for accessing DataPlane services.
 
 ## Import
 
-Azure DevSpace Controller can be imported using the `resource id`, e.g.
+DevSpace Controller can be imported using the `resource id`, e.g.
```
tombuildsstuff:
```suggestion
DevSpace Controller's can be imported using the `resource id`, e.g.
```

azurerm/resource_arm_devspace_controller.go
metacpp: Are you saying that `props` will never be `nil` in this case? If this is the case, can I just remove the `nil` check for props ?

website/docs/r/devspace_controller.html.markdown
metacpp: Why `'s` here ?

azurerm/resource_arm_devspace_controller.go
tombuildsstuff: not quite - if `props` is `nil` then these properties won't be set as the code stands; which is the right approach, since these'll appear as an error on the Import tests - rather than showing as constant changes on users `terraform plan`'s (where there's not a whole bunch they can do to fix them)

tombuildsstuff: (this looks good with the changes - thanks 👍)

PR #2190, opened 2018-10-30
Fixing linting errors

tombuildsstuff:
These come from PR's merged after https://github.com/terraform-providers/terraform-provider-azurerm/pull/2173 was merged


katbyte:
@metacpp,

That is because each PR's linting ran before either was merged, so no errors where reported and PR's do not get CI'd upon every change to master afaik.

PR #2199, opened 2018-10-31
refactor: feature-toggled authentication methods

tombuildsstuff:
This PR refactors the authentication package such that it's capable of supporting multiple authentication methods which may be feature toggled on or off. This is the second step in refactoring the authentication package out into it's own library such that it can be consumed in Terraform Core(for the AzureRM Backend), the Azure Stack Provider and this Provider.

Due to the nature of the fact this is authentication related; this may require some additional unit testing and some manual UX testing to ensure we're supporting all the known scenarios:

- [x] Azure CLI
  - [x] Parsing the ClientID, Subscription ID and Tenant ID when nothings specified
  - [x] Parsing the Tenant ID when the Subscription ID is specified
  - [x] Parsing the Client ID when the Subscription ID & Tenant ID are specified

---

- [x] CloudShell
  - [x] Parsing the ClientID, Subscription ID and Tenant ID when nothings specified
  - [x] Parsing the Tenant ID when the Subscription Id is specified
  - [x] Parsing the Client ID when the Subscription ID & Tenant ID are specified

---

- [x] Managed Service Identity
- [x] Service Principal Client Certificate (this is intentionally feature-toggled off)
- [x] Service Principal Client Secret


tombuildsstuff:
Acceptance Tests still pass - so that works too 🙌 

```
 $ acctests azurerm TestAccAzureRMResourceGroup_withTags
--- PASS: TestAccAzureRMResourceGroup_withTags (88.29s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	90.068s
```

azurerm/helpers/authentication/auth_method_msi_test.go
```diff
@@ -0,0 +1,50 @@
...
+		{
+			Description: "Empty Configuration",
+			Config:      managedServiceIdentityAuth{},
+			ExpectError: true,
+		},
+		{
+			Description: "Missing MSI Endpoint",
+			Config:      managedServiceIdentityAuth{},
+			ExpectError: true,
+		},
```
tombuildsstuff: this wants a valid test too

azurerm/helpers/authentication/builder.go
```diff
@@ -0,0 +1,87 @@
...
+		log.Printf("[DEBUG] Using Managed Service Identity for Authentication")
+		method, err := newManagedServiceIdentityAuth(b)
+		if err != nil {
+			return nil, err
+		}
+		config.authMethod = method
+		return config.validate()
+	}
+
+	// note: this includes CloudShell
```
tombuildsstuff: this comment is unnecessary now the fields been renamed

azurerm/helpers/authentication/config.go
```diff
@@ -1,153 +1,34 @@
...
-	SkipCredentialsValidation bool
-	SkipProviderRegistration  bool
+	ClientID       string
+	SubscriptionID string
+	TenantID       string
+	Environment    string
 
-	// Service Principal Auth
-	ClientSecret string
+	// temporarily public feature flags
```
tombuildsstuff: comment probably wants adjusting

azurerm/helpers/authentication/builder.go
```diff
@@ -0,0 +1,87 @@
...
+	MsiEndpoint                    string
+
+	// Service Principal (Client Cert) Auth
+	SupportsClientCertAuth bool
+	ClientCertPath         string
+	ClientCertPassword     string
+
+	// Service Principal (Client Secret) Auth
+	SupportsClientSecretAuth bool
+	ClientSecret             string
```
tombuildsstuff: each of these probably wants some docs for what they actually do

azurerm/helpers/authentication/builder.go
```diff
@@ -0,0 +1,87 @@
+package authentication
+
+import (
+	"fmt"
+	"log"
+)
+
+type Builder struct {
```
tombuildsstuff: docs

azurerm/helpers/authentication/config.go
```diff
@@ -1,153 +1,34 @@
...
-
-	// always pull the Environment from the CLI
-	err = c.populateEnvironmentFromCLIProfile(cliProfile)
-	if err != nil {
-		// we want to expose a more friendly error to the user, but this is useful for debug purposes
-		log.Printf("Error Populating the Environment from the CLI Profile: %s", err)
-	}
-
-	return nil
+func (c Config) GetAuthorizationToken(oauthConfig *adal.OAuthConfig, endpoint string) (*autorest.BearerAuthorizer, error) {
```
tombuildsstuff: docs

azurerm/helpers/authentication/builder.go
tombuildsstuff: on reflection - maybe not

azurerm/helpers/authentication/auth_method.go
```diff
@@ -0,0 +1,12 @@
+package authentication
```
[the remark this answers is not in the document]
tombuildsstuff: this is an existing package so I don't think this PR would be the place to rename that - that said I think we could think about this when we split this out into it's own lib?

azurerm/helpers/authentication/auth_method_client_secret.go
```diff
@@ -0,0 +1,61 @@
...
+	if a.clientId == "" {
+		err = multierror.Append(err, fmt.Errorf(fmtErrorMessage, "Client ID"))
+	}
+	if a.clientSecret == "" {
+		err = multierror.Append(err, fmt.Errorf(fmtErrorMessage, "Client Secret"))
+	}
+	if a.tenantId == "" {
+		err = multierror.Append(err, fmt.Errorf(fmtErrorMessage, "Tenant ID"))
+	}
+	if a.environment == "" {
```
paultyng: seems like some of these fields aren't used, do they have to be required?

azurerm/helpers/authentication/auth_method.go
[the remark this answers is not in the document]
paultyng: Oh this package already exists and will be split out I think?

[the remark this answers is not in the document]
tombuildsstuff: yep - this package has been around for a while & will go into the new split-out helpers repo 👍

azurerm/helpers/authentication/auth_method_client_cert.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	if a.clientId == "" {
+		err = multierror.Append(err, fmt.Errorf(fmtErrorMessage, "Client ID"))
+	}
+
+	if a.clientCertPath == "" {
+		err = multierror.Append(err, fmt.Errorf(fmtErrorMessage, "Client Certificate Path"))
+	} else {
+		if strings.HasSuffix(strings.ToLower(a.clientCertPath), ".pfx") {
```
katbyte: Is there a reason to enforce that the file ends with .pfx? I've in the past used things like `cert.pfx.old`

azurerm/helpers/authentication/auth_method_client_cert_test.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	if builder.TenantID != servicePrincipal.tenantId {
+		t.Fatalf("Expected Tenant ID to be %q but got %q", builder.TenantID, servicePrincipal.tenantId)
+	}
+}
+
+func TestServicePrincipalClientCertAuth_validate(t *testing.T) {
+	data := []byte("client-cert-auth")
+	filePath := "./example.pfx"
```
katbyte: Should we use `ioutil .TempFile` here to ensure no possible collisions or pollution of the project dir if something crashes?

azurerm/helpers/authentication/auth_method_client_cert.go
tombuildsstuff: this is intentional since we require a PFX file here (whereas other providers require `cer` files etc) - so I'd suggest we start out strict and we can loosen this if required

azurerm/helpers/authentication/auth_method_client_cert_test.go
tombuildsstuff: started out with that approach - however ultimately ended up reverting it (since the TempFile is only a prefix not a suffix) - for the reasons outlined above

Issue #1990, opened 2018-09-28
Have azurerm_storage_blob updates on etag/hash changes


arichiardi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The AWS [s3_bucket_object](https://www.terraform.io/docs/providers/aws/r/s3_bucket_object.html) has a wonderful feature for forcing a new resource - etag.
It would be awesome if `azurerm_storage_blob` could do the same, at the moment the only way to have a new resource is to have different file names.

### New or Affected Resource(s)

* azurerm_storage_blob

### Potential Terraform Configuration

```hcl
resource "azurerm_storage_blob" "artifact_blob" {
  name = "${basename(var.artifact_path)}"
  etag = "${md5(file("path/to/file"))}"  <- desiderata
 
  ...

  type   = "block"
  source = "${var.artifact_path}"
}

mb290:
Can we just leverage the additional field Content-MD5 [https://github.com/Azure/azure-sdk-for-go/blob/master/storage/blob.go#L387]("https://github.com/Azure/azure-sdk-for-go/blob/master/storage/blob.go#L387")

mb290:
@tombuildsstuff "should" and "reality" have turned into different things! I'm working through the uploads and struggling through the following error:

> * azurerm_storage_blob.testsb: Error creating storage blob on Azure: Error updating block list for source file "text.txt": storage: service returned error: StatusCode=400, ErrorCode=InvalidMd5, ErrorMessage=The MD5 value specified in the request is invalid. MD5 value must be 128 bits and base64 encoded.

I'll report back with an update as I either succeed or fail miserably.

tombuildsstuff:
@mb290 👍 cool - it may be worth opening an issue on [the Azure SDK for Go repository](https://github.com/Azure/azure-sdk-for-go) for more clarification? Whilst that SDK is deprecated, unfortunately we're still blocked from migrating to [the other Go Storage SDK](https://github.com/Azure/azure-storage-blob-go)

mb290:
[...]

PR #2209, opened 2018-11-01
azurerm_storage_blob content change tracking

mb290:
This PR adds functionality to the Azure Storage Blob resource to track local file changes - as per [#1990](https://github.com/terraform-providers/terraform-provider-azurerm/issues/1990). This does not track updates performed directly in Azure though, which I was unable to work out how to achieve without the proposed implementation of [#19245](https://github.com/hashicorp/terraform/issues/19245) which there is some reluctance to do.


mb290:
@tombuildsstuff can I get some feedback on this? We're currently using a custom fork with the solution but I'd like to get something formal before we open it up too much. Thanks!

katbyte:
@mb290, 

I am going to close this as it is a partial solution. I have left comment with more details on #1990, and once we have a complete solution this PR can be reopened.

PR #2219, opened 2018-11-02
Linting: include sub packages

katbyte:


GNUmakefile
```diff
@@ -39,7 +39,7 @@ goimport:
 
 lint:
 	@echo "==> Checking source code against linters..."
-	@gometalinter ./$(PKG_NAME)
+	@gometalinter ./$(PKG_NAME)/...
```
paultyng: Does this just work if you do `@gometalinters ./...` like for `go test`, etc.?

azurerm/helpers/authentication/access_token_test.go
```diff
@@ -163,7 +163,7 @@ func TestAzureFindValidAccessTokenForTenant_ValidFromCloudShell(t *testing.T) {
 		t.Fatalf("Expected the Client ID to be %q but got %q", expectedToken.ClientID, token.ClientID)
 	}
 
-	if token.IsCloudShell != true {
+	if token.IsCloudShell {
```
paultyng: I think this inverted the boolean?

azurerm/helpers/authentication/config.go
```diff
@@ -73,21 +73,21 @@ func (c *Config) LoadTokensFromAzureCLI() error {
 	foundToken := false
 	if c.TenantID != "" {
 		// pull out the ClientID and the AccessToken from the Azure Access Token
-		tokensPath, err := cli.AccessTokensPath()
-		if err != nil {
-			return fmt.Errorf("Error loading the Tokens Path from the Azure CLI: %+v", err)
+		tokensPath, err2 := cli.AccessTokensPath()
```
paultyng: I think its fair to exclude the shadowing rule for `err` as its just so common, so you don't have to do this. I think using some other name for an error than `err` should be avoided if at all possible.

azurerm/helpers/authentication/access_token_test.go
katbyte: It sure did

tombuildsstuff: 🤔 I'd like to hold off on this PR until #2199 is merged; since otherwise there's going to be a ton of merge conflicts which'll be hard to resolve - WDYT?

katbyte: 👍

tombuildsstuff: @katbyte I think this should be safe to `git reset HEAD~1 && [lint]` now?

katbyte: @tombuildsstuff, 

updated and ready to go

azurerm/utils/response.go
```diff
@@ -26,7 +26,7 @@ func ResponseErrorIsRetryable(err error) bool {
 	return false
 }
 
-func responseWasStatusCode(resp autorest.Response, statusCode int) bool {
+func responseWasStatusCode(resp autorest.Response, statusCode int) bool { // nolint: unparam
```
tombuildsstuff: which parameter is unused here?

katbyte:
```
==> Checking source code against linters...
azurerm/utils/response.go:29:52:warning: parameter statusCode always receives 404 (unparam)
make: *** [lint] Error 1
```

PR #2221, opened 2018-11-02
Fixed value length check

jeffreyCline:


tombuildsstuff:
hey @jeffreyCline 

Thanks for this PR :)

Taking a quick look at the portal it appears the maximum supported value is 30 (and not 128):

screenshot 2018-11-02 at 23 54 12

As such could we invert this change (to fix the comments/error message/tests rather than the limit?) - but in general I'm 👍 with this change :)

Thanks!

katbyte:
Might I suggest actually hitting the API? multiple times i've found the value in the portal to be vastly different then what the API will allow.

azurerm/resource_arm_databricks_workspace.go
```diff
@@ -187,7 +187,7 @@ func validateDatabricksWorkspaceName(v interface{}, k string) (ws []string, erro
 	}
 
 	// Cannot be more than 128 characters
-	if len(value) > 30 {
+	if len(value) > 128 {
```
metacpp: According to API spec, it should be between 3 and 64:
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/databricks/resource-manager/Microsoft.Databricks/stable/2018-04-01/databricks.json#L684

I would suggest follow the API swagger.

metacpp:
> hey @jeffreyCline
> 
> Thanks for this PR :)
> 
> Taking a quick look at the portal it appears the maximum supported value is 30 (and not 128):
> 
> screenshot 2018-11-02 at 23 54 12
> 
> As such could we invert this change (to fix the comments/error message/tests rather than the limit?) - but in general I'm 👍 with this change :)
> 
> Thanks!

And its length definition is between 3 and 64 in API Swagger:
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/databricks/resource-manager/Microsoft.Databricks/stable/2018-04-01/databricks.json#L684

jeffreyCline:
According to the azure-sdk-for-go they too have implemented the length as 3 to 64...
https://github.com/Azure/azure-sdk-for-go/blob/ff2671f6653891a6025a1ee25ce033100a9f090a/services/databricks/mgmt/2018-04-01/databricks/workspaces.go#L161

metacpp:
> According to the azure-sdk-for-go they too have implemented the length as 3 to 64...

@jeffreyCline this is because of SDK was generated from API swagger.

PR #2232, opened 2018-11-04
Formatted acceptance test tf for resources A to K & data sources

katbyte:


katbyte:
@metacpp,

It is a [small script](https://github.com/katbyte/tfformatter) I've been toying with while on vacation, its pretty rough right now with a few bugs but eventually it could be a target & lint TF blocks in PRs. 

Either way it sure has made these acc tf formatting PR's quick and easy 🙂

Issue #2235, opened 2018-11-05
azurerm_monitor_log_profile is unfunctional


lrxtom2:
Resource 'azurerm_monitor_log_profile' is unfunctional, please have a check.

Thanks,
Chang

1 error(s) occurred:

* azurerm_monitor_log_profile.test: 1 error(s) occurred:

* azurerm_monitor_log_profile.test: Error Creating/Updating Log Profile "default": insights.LogProfilesClient#CreateOrUpdate: Failure sending request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=<nil> <nil>

JunyiYi:
This error is caused by the limitation of at most one log-profile in one subscription. The detailed error message is `The limit of 1 log profiles was reached. To create new log profile 'default', delete an existing one.`

tombuildsstuff:
@JunyiYi sounds like this could probably do with a blue info box on the documentation page (see below) to explain about this? e.g.

```
-> **NOTE:** It's only possible to configure one Log Profile per Subscription
```

JunyiYi:
@tombuildsstuff Yes I agree, let me simply add that.

tombuildsstuff:
Closing since this has been documented in #2265

PR #2265, opened 2018-11-07
Add Log Profile limitation note to the document

JunyiYi:
Follow-up for issue #2235 .

### DIFF
--- a/website/docs/r/monitor_log_profile.html.markdown
+++ b/website/docs/r/monitor_log_profile.html.markdown
@@ -10,6 +10,8 @@ description: |-
 
 Manages a [Log Profile](https://docs.microsoft.com/en-us/azure/monitoring-and-diagnostics/monitoring-overview-activity-logs#export-the-activity-log-with-a-log-profile). A Log Profile configures how Activity Logs are exported.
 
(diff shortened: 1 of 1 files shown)


website/docs/r/monitor_log_profile.html.markdown
```diff
@@ -10,6 +10,8 @@ description: |-
 
 Manages a [Log Profile](https://docs.microsoft.com/en-us/azure/monitoring-and-diagnostics/monitoring-overview-activity-logs#export-the-activity-log-with-a-log-profile). A Log Profile configures how Activity Logs are exported.
 
+-> **NOTE:** It's only possible to configure one Log Profile per Subscription. If you are trying to create more than one Log Profile, an error with `StatusCode=409` will occur.
```
tombuildsstuff: it'd be good to catch this status code in the resource and throw a more helpful error message too (we should be able to do this [here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_monitor_log_profile.go#L119)) - what do you think?

JunyiYi: The status code itself stands for `Conflict`. The limitation could be one of the reason, but might not likely be the only reason.

PR #2285, opened 2018-11-09
Add `compute` attribute for `route_table_id` and `network_security_group` in `azurerm_subnet`

metacpp:
This PR added `computed` for `route_table_id` and `network_security_group` in `azurerm_subnet`, so that rerun of `terraform plan` will not report unexpected change.

The `<->` is broken in the `Note` part of both pages:
https://www.terraform.io/docs/providers/azurerm/r/subnet_route_table_association.html
https://www.terraform.io/docs/providers/azurerm/r/subnet_network_security_group_association.html

### DIFF
--- a/azurerm/resource_arm_subnet.go
+++ b/azurerm/resource_arm_subnet.go
@@ -44,12 +44,14 @@ func resourceArmSubnet() *schema.Resource {
 			"network_security_group_id": {
 				Type:       schema.TypeString,
 				Optional:   true,
+				Computed:   true,
 				Deprecated: "Use the `azurerm_subnet_network_security_group_association` resource instead.",
 			},
 
 			"route_table_id": {
 				Type:       schema.TypeString,
 				Optional:   true,
+				Computed:   true,
 				Deprecated: "Use the `azurerm_subnet_route_table_association` resource instead.",
 			},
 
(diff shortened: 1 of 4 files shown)


tombuildsstuff:
hi @metacpp 

At this time both of these fields are intentionally not set to Computed, since this breaks the ability to remove these values using the existing resources; as such the documentation at this time requires that users specify the association both as an independent resource and within the `azurerm_subnet` resource. In the next major version of the AzureRM Provider (2.0) - we'll remove these fields from the `azurerm_subnet` resource in favour of the independent resources (which is why these are deprecated).

As such I don't believe we can merge this PR - so I'm going to close this PR for the moment

Thanks!

metacpp:
> At this time both of these fields are intentionally not set to Computed, since this breaks the ability to remove these values using the existing resources

I don't quite understand the scenario here: user sees the values will be cleared through running of `terraform plan` and prefer to apply it?

tombuildsstuff:
> I don't quite understand the scenario here: user sees the values will be cleared through running of terraform plan and prefer to apply it?

@metacpp yes, the configuration defined in code should match the state of the resource in Azure - else Terraform should update the state of the resource in Azure so that it matches the state in code.

At this point in time (until 2.0) the field in the `azurerm_subnet` resource and either the `azurerm_subnet_network_security_group_association` / `azurerm_subnet_route_table_association` resource depending on the field used - since otherwise this'll be a breaking change. The existing (deprecated) fields on the `azurerm_subnet` resource will be removed in 2.0 - at which time this migration is complete). Whilst this isn't ideal this prevents a breaking change where it's not possible to remove a NSG association when not using the `azurerm_subnet_network_security_group_association` resource / a route table when not using the `azurerm_subnet_route_table_association` resource :)

PR #1820, opened 2018-08-24
Adding support for AzureAD auth to AKS via RBAC

timcurless:
I _think_ this is a good starting point. 

1) ~~I would like to refactor to be much more DRY~~ [UPDATE]: Refactoring complete
2) This could use some further testing
3) ~~I have not yet updated docs~~ [UPDATE]: Docs have been updated.

PR #2286, opened 2018-11-09
Add test and sample for AKS with route table.

metacpp:
W/o specifying the route table and associate it with subnet, AKS will try to create a new route table instead and not associate it w/ subnet.

Before AKS team apply the fix, we'd like to update the test to cover customized route table scenario and  encourage user to specify it instead of using default one, which is not working for now.


azurerm/resource_arm_kubernetes_cluster_test.go
```diff
@@ -220,6 +220,31 @@ func TestAccAzureRMKubernetesCluster_internalNetwork(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMKubernetesCluster_internalNetworkRouteTable(t *testing.T) {
```
tombuildsstuff: we should be able to combine this into the `TestAccAzureRMKubernetesCluster_advancedNetworking[X]Complete` test

tombuildsstuff:
Closing as these changes have been pulled into #1820

PR #2306, opened 2018-11-14
cleaned up storage blob validation functions

katbyte:


azurerm/helpers/validate/int.go
```diff
@@ -0,0 +1,72 @@
...
+		if math.Mod(float64(v), float64(divisor)) != 0 {
+			es = append(es, fmt.Errorf("expected %s to be divisible by %d", k, divisor))
+			return
+		}
+
+		return s, es
+	}
+}
+
+// intBetweenDivisibleBy returns a SchemaValidateFunc which tests if the provided value
```
jeffreyCline: Super minor, it would be nice if we could update the name in the comment to match the new function name (e.g. `IntDivisibleBy`)

katbyte: done, merge at your leisure

azurerm/helpers/validate/int.go
```diff
@@ -0,0 +1,72 @@
...
+			es = append(es, fmt.Errorf("expected %s to be divisible by %d", k, divisor))
+			return
+		}
+
+		return s, es
+	}
+}
+
+// IntDivisibleBy returns a SchemaValidateFunc which tests if the provided value
+// is of type int and is between min and max (inclusive) and is divisible by a given number
```
jeffreyCline: @katbyte, do you think we should update the text to the below since the function doesn't actually validate if the value is between two values?

```suggestion
// is of type int and is divisible by a given number
```

azurerm/helpers/validate/int.go
```diff
@@ -0,0 +1,72 @@
...
+			return
+		}
+
+		for _, str := range valid {
+			if v == str {
+				return
+			}
+		}
+
+		es = append(es, fmt.Errorf("expected %q to be one of %v, got %v", k, valid, v))
```
JunyiYi: *minor* `%q` here seems to be inconsistent with other format string.

azurerm/resource_arm_storage_blob.go
```diff
@@ -13,6 +13,10 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
+
```
JunyiYi: *minor* Redundant line breaks?

azurerm/resource_arm_storage_blob.go
```diff
@@ -38,113 +42,81 @@ func resourceArmStorageBlob() *schema.Resource {
...
 				Computed: true,
 			},
+
 			"parallelism": {
 				Type:         schema.TypeInt,
 				Optional:     true,
 				Default:      8,
 				ForceNew:     true,
-				ValidateFunc: validateArmStorageBlobParallelism,
+				ValidateFunc: validation.IntAtLeast(0),
```
JunyiYi: Not sure for the resource, but from the old validation function, this should be:

```suggestion
				ValidateFunc: validation.IntAtLeast(1),
```

azurerm/resource_arm_storage_blob.go
```diff
@@ -38,113 +42,81 @@ func resourceArmStorageBlob() *schema.Resource {
...
 			},
+
 			"attempts": {
 				Type:         schema.TypeInt,
 				Optional:     true,
 				Default:      1,
 				ForceNew:     true,
-				ValidateFunc: validateArmStorageBlobAttempts,
+				ValidateFunc: validation.IntAtLeast(0),
```
JunyiYi: The same comment as before:

```suggestion
				ValidateFunc: validation.IntAtLeast(1),
```

jeffreyCline:
image

azurerm/helpers/validate/int.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"fmt"
+	"math"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+// IntBetweenAndDivisibleBy returns a SchemaValidateFunc which tests if the provided value
+// is of type int and is between min and max (inclusive) and is divisible by a given number
+func IntBetweenAndDivisibleBy(min, max, divisor int) schema.SchemaValidateFunc { // nolint: unparam
```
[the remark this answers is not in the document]
tombuildsstuff: indeed - we have plans for this in Terraform Core once 0.12's out (we're thinking `validation.All(Foo(), Bar())`)

Issue #2322, opened 2018-11-15
[Proposal] 2.0 - Azure Active Directory


tombuildsstuff:
As a part of v2.0 of the AzureRM Provider we're working through some of the longer-term changes we'd like to make.

Over the past year or so we've shipped support for some Azure Active Directory resources within the AzureRM Provider - including support for Applications, Service Principals (and there's PR's/feature requests for more AzureAD resources).

Upon reflecting on this a little - we realised that Azure Active Directory may make more sense as a Provider in it's own right. There's several reasons for this:

## 1. Acceptance Tests

We run Acceptance Tests for all of the Data Sources and Resources in the AzureRM Provider every night - which means that we spin up thousands of resources on demand. This has allowed the AzureRM Provider (along with all the other Terraform Providers) to move pretty quickly - we ship new releases often and have confidence in the resources we're shipping.

All of the AzureAD resources we currently support have acceptance tests, which run nightly too. However there's several Resources/Data Sources we want to add (such as Users) which are considerably harder to write automated acceptance tests for; this is because they (for example) involve sending an email and then completing a Captcha on signup etc. In addition - (almost) every Data Source has an associated Resource - however being unable to write Acceptance Tests for a given resource means we're also unable to support a Data Source either.

Splitting AzureAD out into it's own provider would allow us to manually test the components which can't be automatically tested (prior to a release) whilst still remaining confident in the other resources (since there'd still be automated tests). Since we'd be moving slower in this provider (specifically, we'd be upgrading the Azure SDK for Go less) - shipping support for a resource that's manually tested wouldn't be /as/ big of a deal as it would be in the AzureRM Provider (whilst obviously not being ideal).

## 2. Re-evaluating the design of some of the AzureAD Application/Service Principal resources

After shipping the `azurerm_azuread_application`, `azurerm_azuread_service_principal` and `azurerm_azuread_service_principal_password` resources and seeing some real world feedback - there's sufficient bugs/issues in these resources that I believe these resources require some thought/re-work. At this point in time I'm wondering if it'd make sense to combine the Service Principal resources together - but it requires some prototyping/consultation/additional research to confirm we've covered all the edge-cases.

Splitting AzureAD out into it's own provider would allow us to reevaluate the design of these resources.

## 3. Authentication

[...]

PR #2330, opened 2018-11-15
Expose `provision_vm_agent` in `os_profile_linux_config` for VM and VMSS

### LINKED ISSUE #2264: LinuxConfiguration Support in Compute Resource Provider
In the CRP API version 2018-06-01 we added support for :
  "linuxConfiguration": {
    "provisionVMAgent": false
  }
 
This is not available in azurerm_virtual_machine in os_profile_linux_config.

Can you advise when this can be made available?

Thanks,

metacpp:
1. Resolve #2264, to expose the `provision_vm_agent` in `os_profile_linux_config` for VM and VMSS.
2. Make schema attribute change for `disable_password_authentication` in VMSS to be consistent with definition in doc and VM.

### DIFF
--- a/azurerm/resource_arm_virtual_machine.go
+++ b/azurerm/resource_arm_virtual_machine.go
@@ -483,6 +483,11 @@ func resourceArmVirtualMachine() *schema.Resource {
 							Type:     schema.TypeBool,
 							Required: true,
 						},
+						"provision_vm_agent": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
 						"ssh_keys": {
 							Type:     schema.TypeList,
 							Optional: true,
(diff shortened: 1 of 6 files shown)


tombuildsstuff:
hi @metacpp

Thanks for this PR :)

As mentioned in a couple of other issues recently - we're planning on splitting the `azurerm_virtual_machine` and `azurerm_virtual_machine_scale_set` resources into two new resources (one for Linux and one for Windows) as a part of 2.0. Similar to #2322 we're scoping these changes at the moment and plan to publish a proposal for how we think these will look in the not-too-distant future.

Whilst we could potentially add this functionality to the existing VM and VMSS resources - since this requires changing the hash function and a state migration to include this new field (and thus would require some extensive testing to ensure the upgrade path) - as we don't currently support this functionality I believe it'd be better to instead support it via these new resources instead. As such I'm going to close this PR for the moment - but we'll look to expose this field as a part of the new `azurerm_linux_virtual_machine` / `azurerm_linux_virtual_machine_scale_set` resources as a part of the 2.0 release.

Thanks!

PR #2333, opened 2018-11-16
Fix doc & test issues for `disable_password_authentication` in VMSS

metacpp:
This PR fixes the wrong value setting for `disable_password_authentication ` in VMSS' testing and documentation.


tombuildsstuff:
hi @metacpp

Thanks for this PR :)

As mentioned in a couple of other issues recently - we're planning on splitting the `azurerm_virtual_machine` and `azurerm_virtual_machine_scale_set` resources into two new resources (one for Linux and one for Windows) as a part of 2.0. Similar to #2322 we're scoping these changes at the moment and plan to publish a proposal for how we think these will look in the not-too-distant future.

As this change would require changes to users Terraform Configurations (and one of the considerations which requires investigation for the new VM Scale Set resources is if we should remove the `disable_password_authentication` field in favour of inferring it) - I think it'd be better to make this change in the 2.0 release as a part of these new resources, as such I'm going to close this PR for the moment.

Thanks!

metacpp:
@metacpp I've cleaned up the commit history on this branch.

For the value `disable_password_authentication`, the reason that I set it with `false`, is because of `admin_password` in the `os_profile`.  If it's `true`, `admin_password` will never be validated and used. That's why I also did the change on L to correct the password format.

If `ssh_keys` is used here as authentication way, shall we remove the `admin_password` here?

Thoughts?

metacpp:
@tombuildsstuff I've updated the PR for your suggestions.

PR #2343, opened 2018-11-16
Unify Resource Name Validation

jeffreyCline:
Introduces method to validate Azure Resource names which enforce the Azure standards for naming resources. Currently there isn't a standard way to do the validation which causes the validation to not be consistent within the provider.

 1. First two characters must be a number or a letter
 2. Last characters must be a number or a letter
 3. No consecutive hyphens, underscores, parentheses, or periods
 4. Min and Max length for values vary depending on the resource
 5. Value can not start or end with any special character

### DIFF
--- a/azurerm/helpers/azure/generic.go
+++ b/azurerm/helpers/azure/generic.go
@@ -5,10 +5,10 @@ import (
 	"regexp"
 )
 
-const hyphenUnderscoreParenthesesPeriod string = "-_\\(\\)\\."
-const hyphenUnderscorePeriod string = "-_\\."
+const hyphenUnderscoreParenthesesPeriod string = `-_\(\)\.`
+const hyphenUnderscorePeriod string = `-_\.`
 const hyphenUnderscore string = "-_"
-const hyphenPeriod string = "-\\."
+const hyphenPeriod string = `-\.`
 const hyphen string = "-"
 const none string = ""
 const alphanumericLower string = "a-z0-9"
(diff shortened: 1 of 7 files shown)


JunyiYi:
Thanks @jeffreyCline for the contributions. This is very helpful. However you do have some conflicts waiting for your resolution.

tombuildsstuff:
hey @jeffreyCline 

Thanks for this PR - apologies this has been sat for a while, we've been trying to work out the best path forward here.

We've been chatting internally with the other provider maintainers around some longer term functionality which we'd like to add - one of these things has been a validation method `All` which takes a series of validation functions; which would allow us to mix & match validation functions. Whilst ultimately this would lead to the same result as in this PR I believe this'd be a lot more approachable / maintainable in the long-run as additional validation requirements appear; to give an example:

```
ValidateFunc: validation.All(
  validation.StringLength(1, 100),
  validation.StartsWithAlphanumeric(),
  validation.ShouldBeLowerCase(),
  validation.DisallowHyphens(),
  // ...
)
```

The intention is that this method will live in Terraform Core where some of these common validation functions could also live - whilst we don't have a timeframe for this we could add these methods to the AzureRM Provider for the moment and then migrate this over to Terraform Core at a later date.

A couple of things of note around validation in general however:

- In general we should take the validation requirements from the Azure Portal since historically these have been more correct in practice - whereas the Azure CLI/Documentation/SDK are all auto-generated from the Swagger where the validation requirements historically are non-existent (or wrong).
- Go as a language favours duplicating code verses refactoring ahead of time - whilst there are some common patterns here I believe the `validation.All` method outlined above should significantly help with this. As you've done in this method it's still valuable to split the validation out into their own functions rather than defining them in-line within the Schema so that we can (more) easily add test cases covering them.

Given we're looking to approach this problem from a different direction I'm going to close this PR for the moment - but I'd like to thank you for this contribution :)

Thanks!

Issue #2345, opened 2018-11-17
Enable RBAC without AAD


raphaelquati:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

AKS with RBAC + AAD is not a requirement; you can enable RBAC without AAD.

### New or Affected Resource(s)

* resource_arm_kubernetes_cluster

### Potential Terraform Configuration

```hcl
resource "azurerm_kubernetes_cluster" "k8s" {
    name = "test"
    location = "East US"
    resource_group_name = "test"
    dns_prefix = "test"
    agent_pool_profile {
        name            = "agentpool"
        count           = "${var.agent_count}"
        vm_size         = "Standard_F4s_v2"
        os_type         = "Linux"
        os_disk_size_gb = 30
    }

    linux_profile {
        admin_username = "ubuntu"
        ssh_key {
            key_data = "${file("${var.ssh_public_key}")}"
        }
    }

    service_principal {
        client_id     = "${var.client_id}"
        client_secret = "${var.client_secret}"
    }

    network_profile {
        network_plugin = "azure"
    }

    role_based_access_control {
        enabled = true
    }
}    
```

### References

* https://docs.microsoft.com/en-us/azure/templates/Microsoft.ContainerService/2018-03-31/managedClusters#ManagedClusterProperties

PR #2347, opened 2018-11-17
Enable RBAC without AAD

raphaelquati:
Implementation for #2345

### DIFF
--- a/azurerm/resource_arm_kubernetes_cluster_test.go
+++ b/azurerm/resource_arm_kubernetes_cluster_test.go
@@ -597,7 +597,7 @@ resource "azurerm_kubernetes_cluster" "test" {
   }
 
   role_based_access_control {
-		enaled=true
+		enabled=true
     azure_active_directory {
       server_app_id     = "%s"
       server_app_secret = "%s"
--- a/azurerm/resource_arm_kubernetes_cluster_test.go
+++ b/azurerm/resource_arm_kubernetes_cluster_test.go
@@ -122,6 +122,7 @@ func TestAccAzureRMKubernetesCluster_roleBasedAccessControl(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testCheckAzureRMKubernetesClusterExists(resourceName),
 					resource.TestCheckResourceAttr(resourceName, "role_based_access_control.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "role_based_access_control.0.enabled", "true"),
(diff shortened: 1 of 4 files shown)


tombuildsstuff:
hi @raphaelquati 

Thanks for this PR :)

During the investigation into RBAC recently it appeared that whilst you can enable/disable RBAC for AKS at this time it's tied to AzureAD. For this reason we've designed the block such that if/when additional methods are available for RBAC in the future that we can support them as nested items in the `role_based_access_control` block, by making the AzureAD block optional.

As such I believe we should be able to infer the value of the `enable_rbac` field and as such we should be able to not expose it - do you have a reference which shows that it's possible to Enable RBAC without Azure AD?

Thanks!

joakimhellum-in:
To be clear: I completely agree on RBAC being related to Azure AD for user authentication, and we are currently re-deploying all our clusters to enable this. Also because it's in the security [best practices](https://github.com/Azure/k8s-best-practices/blob/master/Security_securing_a_cluster.md#master-endpoint-security) .

However it is possible to only enable RBAC without Azure AD integration at the moment.

1. Using the Azure Portal we can in fact only enable RBAC and not the Azure AD integration:
image

2. Using Azure CLI RBAC is enabled by default (`--enable-rbac`):
image

But you would need to enable Azure AD by setting the parameters (`--aad-client-app-id`, `--aad-server-app-id`, `--aad-server-app-secret` and `--aad-tenant-id`):
image

Reference: https://docs.microsoft.com/en-us/cli/azure/aks?view=azure-cli-latest#az-aks-create

So I think multiple users might have an issue with this update unless they upgrade their clusters to enable Azure AD integration.

raphaelquati:
@tombuildsstuff,

I try to follow the same "block idea" of others blocks (like oms_agent and http_application_routing), with the "enabled" flag, instead of inferring the presence of block.

azurerm/resource_arm_kubernetes_cluster_test.go
```diff
@@ -596,6 +597,7 @@ resource "azurerm_kubernetes_cluster" "test" {
   }
 
   role_based_access_control {
+		enaled=true
```
timfpark: it looks like you have a typo here: enaled -> enabled

ams0:
Please prioritize this together with #2421 , the current implementation forces AAD integration whenever RBAC is needed (which is most of the times), and that's not ideal nor the AKS defaults. Let me know if I can help testing it.

tombuildsstuff:
As discussed above - closing this in favour of #2495

PR #2354, opened 2018-11-19
New Data Sources: `azurerm_lb` / `azurerm_lb_backend_address_pool`

### LINKED ISSUE #872: Feature request: Data source for load balancer backend address pools
I have a use case where I'm making a load balancer and its backend pool in one module, and then I would like to create a data source to import it in a different module so that I can reference it in the definition of a NIC.

This use case is particularly hindered by the lack of a load balancer backend address pool data source because azurerm_network_interface takes a list of load_balancer_backend_address_pools_ids, which means without the data source, I'm having to hard code my subscription ID into my NIC module.

Thanks for all your work on this Terraform provider; it's been a pleasure to use!

tombuildsstuff:
Fixes #872

### DIFF
--- a/azurerm/data_source_loadbalancer.go
+++ b/azurerm/data_source_loadbalancer.go
@@ -1,0 +1,210 @@
+package azurerm
+
+import (
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-04-01/network"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmLoadBalancer() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmLoadBalancerRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"location": locationForDataSourceSchema(),
+
+			"sku": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"frontend_ip_configuration": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"subnet_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"private_ip_address": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"public_ip_address_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"private_ip_address_allocation": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"load_balancer_rules": {
+							Type:     schema.TypeSet,
+							Computed: true,
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+
+						"inbound_nat_rules": {
+							Type:     schema.TypeSet,
+							Computed: true,
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+
+						"zones": zonesSchemaComputed(),
+					},
+				},
+			},
+
+			"private_ip_address": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"private_ip_addresses": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"tags": tagsForDataSourceSchema(),
+		},
+	}
+}
+
+func dataSourceArmLoadBalancerRead(d *schema.ResourceData, meta interface{}) error {
+	name := d.Get("name").(string)
+	resourceGroup := d.Get("resource_group_name").(string)
+
+	client := meta.(*ArmClient).loadBalancerClient
+	ctx := meta.(*ArmClient).StopContext
+
+	resp, err := client.Get(ctx, resourceGroup, name, "")
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return fmt.Errorf("Load Balancer %q was not found in Resource Group %q", name, resourceGroup)
+		}
+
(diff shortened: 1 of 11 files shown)


website/azurerm.erb
```diff
@@ -120,6 +120,14 @@
                     <a href="/docs/providers/azurerm/d/kubernetes_cluster.html">azurerm_kubernetes_cluster</a>
                 </li>
 
+                <li<%= sidebar_current("docs-azurerm-datasource-load-balancer-x") %>>
+                    <a href="/docs/providers/azurerm/d/loadbalancer.html">azurerm_lb</a>
+                </li>
+
+                <li<%= sidebar_current("docs-azurerm-datasource-load-balancer-backend-address-pool") %>>
+                    <a href="/docs/providers/azurerm/d/loadbalancer_backend_address_pool.html">azurerm_lb_backend_address_pool</a>
```
tombuildsstuff: the name of these matches the name of the resource pages - which is why they're different; although we may choose to redirect them in future

azurerm/data_source_loadbalancer.go
```diff
@@ -0,0 +1,210 @@
...
+							Computed: true,
+						},
+
+						"private_ip_address_allocation": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"load_balancer_rules": {
+							Type:     schema.TypeSet,
```
paddycarver: How are you anticipating people will reference this? The interpolation would need the value before it could know the key to reference it by, and I think count-based loops don't actually work.

azurerm/data_source_loadbalancer.go
```diff
@@ -0,0 +1,210 @@
...
+func dataSourceArmLoadBalancer() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmLoadBalancerRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
katbyte: Validation of at least NoZeroValues` might be nice here?

azurerm/data_source_loadbalancer.go
```diff
@@ -0,0 +1,210 @@
...
+
+			"resource_group_name": resourceGroupNameSchema(),
```
katbyte: Should this be
```suggestion
			"resource_group_name": resourceGroupNameForDataSourceSchema(),
```

website/docs/d/loadbalancer.html.markdown
```diff
@@ -0,0 +1,56 @@
...
+
+* `location` - The Azure location where the Load Balancer exists.
+
+* `sku` - The SKU of the Load Balancer.
+
+* `tags` - A mapping of tags assigned to the resource.
+
+---
+
+A `frontend_ip_configuration` block exports the following:
```
katbyte: This block is missing some of the properties (8 above, 6 here)

website/docs/d/loadbalancer.html.markdown
```diff
@@ -0,0 +1,56 @@
...
+
+## Argument Reference
+
+* `name` - (Required) Specifies the name of the Load Balancer.
+
+* `resource_group_name` - (Required) The name of the Resource Group in which the Load Balancer exists.
+
+## Attributes Reference
+
+The following attributes are exported:
```
katbyte: It looks like `private_ip_address` and `private_ip_addresses` properties are missing?

azurerm/data_source_loadbalancer_backend_address_pool.go
```diff
@@ -0,0 +1,91 @@
...
+			},
+
+			"loadbalancer_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: azure.ValidateResourceID,
+			},
+
+			"backend_ip_configurations": {
+				Type:     schema.TypeSet,
```
katbyte: Should this be a list?

azurerm/data_source_loadbalancer_backend_address_pool.go
```diff
@@ -0,0 +1,91 @@
...
+				Computed: true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.NoZeroValues,
+				},
+				Set: schema.HashString,
+			},
+
+			"load_balancing_rules": {
+				Type:     schema.TypeSet,
```
katbyte: Should this be a list?

azurerm/data_source_loadbalancer.go
tombuildsstuff: no because the ordering will change outside of this, it's unlikely this field will be referenced so we should be able to just remove it for the moment

azurerm/data_source_loadbalancer_backend_address_pool.go
tombuildsstuff: (as above) removing this as it's more likely folks reference this for the ID

azurerm/data_source_loadbalancer_backend_address_pool.go
tombuildsstuff: (as above)

tombuildsstuff:
Tests pass:

screenshot 2019-01-07 at 12 36 47

azurerm/data_source_loadbalancer.go
```diff
@@ -0,0 +1,179 @@
...
+		if zs := config.Zones; zs != nil {
+			zones = *zs
+		}
+		ipConfig["zones"] = zones
+
+		if props := config.FrontendIPConfigurationPropertiesFormat; props != nil {
+			ipConfig["private_ip_address_allocation"] = props.PrivateIPAllocationMethod
+
+			if subnet := props.Subnet; subnet != nil {
+				ipConfig["subnet_id"] = *subnet.ID
```
katbyte: Is this a potential panic?

azurerm/data_source_loadbalancer.go
```diff
@@ -0,0 +1,179 @@
...
+			}
+
+			if pip := props.PrivateIPAddress; pip != nil {
+				ipConfig["private_ip_address"] = *pip
+			}
+
+			if pip := props.PublicIPAddress; pip != nil {
+				ipConfig["public_ip_address_id"] = *pip.ID
```
katbyte: Is this a potential panic?

azurerm/data_source_loadbalancer.go
tombuildsstuff: 👍

azurerm/data_source_loadbalancer.go
tombuildsstuff: 👍 fixed

PR #2378, opened 2018-11-22
azurerm_mssql_elasticpool: moved `elastic_pool_properties` properties to the top level

katbyte:
Also fixes some potential crash points.


azurerm/resource_arm_mssql_elasticpool.go
```diff
@@ -119,39 +120,70 @@ func resourceArmMsSqlElasticPool() *schema.Resource {
...
 
+			"state": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"creation_date": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
tombuildsstuff: I think we’d be better to remove the state and creation date fields entirely here?

azurerm/resource_arm_mssql_elasticpool.go
```diff
@@ -119,39 +120,70 @@ func resourceArmMsSqlElasticPool() *schema.Resource {
...
+
+			"zone_redundant": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+
+			"license_type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
tombuildsstuff: I can’t think of when this field would be likely to be used from within Terraform - should we remove it?

azurerm/resource_arm_mssql_elasticpool.go
```diff
@@ -119,39 +120,70 @@ func resourceArmMsSqlElasticPool() *schema.Resource {
...
 						"license_type": {
-							Type:     schema.TypeString,
-							Computed: true,
+							Type:       schema.TypeString,
+							Computed:   true,
+							Deprecated: "This property has been moved to the top level",
 						},
 					},
 				},
 			},
```
tombuildsstuff: Can we add docs for these new fields?

azurerm/resource_arm_mssql_elasticpool.go
katbyte: Good call

katbyte: I can see them potentially being useful, but i do agree those would be edge cases.

katbyte:
tests pass:
screen shot 2018-11-22 at 15 21 57

PR #2386, opened 2018-11-26
Updating vendor dependency/API version for "policy"

fayusohenson:
Needed in order to support "deployIfNotExists" effect on custom policy definitions.

Before change:
```bash
* azurerm_policy_definition.policy2: 1 error(s) occurred:

* azurerm_policy_definition.policy2: policy.DefinitionsClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidPolicyRuleEffect" Message="The policy definition '*****' rule is invalid. The policy 'effect' property value cannot be 'deployIfNotExists'."
```

After change:
```bash
Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```


katbyte:
Tests pass:
 
screen shot 2018-11-26 at 10 22 54

PR #2395, opened 2018-11-27
Event hub namespace with KafkaEnabled property

### LINKED ISSUE #1417: Feature Request : Azure Event Hub kafka Enabled
_This issue was originally opened by @Ycallaer as hashicorp/terraform#18293. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi,
According to the manual , the block to create event hub looks as follows:
```
resource "azurerm_eventhub_namespace" "test" {
  name                = "acceptanceTestEventHubNamespace"
  sku                 = "Standard"
  capacity            = 1

  tags {
    environment = "Production"
  }
}
```
However, the latest event hub in azure allows you to create kafka enabled clusters (meaning kafka libraries can be used to send events).
This checkbox is currently not reflected in the terraform API.

Would it be possible to introduce it ?

Thanks

fernandoBRS:
Implemented feature request #1417 with `kafkaEnabled` property

(fixes #1417)


fernandoBRS:
Hi @katbyte, yes I can add it in the documentation. Is it better to do a different PR for the documentation?

katbyte:
Same PR is fine @fernandoBRS 🙂

fernandoBRS:
@katbyte the preview version of the Event Hub that supports Kafka has a model structure a bit different compared to the current supported version on Terraform. For example, we can compare the [models.go from the 2017-04-01 version](https://github.com/Azure/azure-sdk-for-go/blob/v18.0.0/services/eventhub/mgmt/2017-04-01/eventhub/models.go) with the [models.go from the preview version](https://github.com/Azure/azure-sdk-for-go/blob/v18.0.0/services/preview/eventhub/mgmt/2018-01-01-preview/eventhub/models.go).

I would suggest two approaches:

- Create a *azurerm_eventhub_preview_namespace*
- Refactor the *azurerm_eventhub_namespace* to use the new API

Do you have any guidance in these approaches?

Thanks!

katbyte:
@fernandoBRS,

Looking at the two files the only difference seems to be the additional two fields:
screen shot 2018-11-28 at 13 32 00

I'm not sure what refactoring would be required aside from adding property?

katbyte:
the `2017-04-01/eventhub` package gets this property in version 22 of the sdk. It might be best to wait until we upgrade.

azurerm/data_source_eventhub_namespace_test.go
```diff
@@ -83,7 +84,9 @@ resource "azurerm_eventhub_namespace" "test" {
   resource_group_name      = "${azurerm_resource_group.test.name}"
   sku                      = "Standard"
   capacity                 = "2"
-  auto_inflate_enabled     = true
+	auto_inflate_enabled     = true
```
jgoldschmidt-apivista: Looks like line 86 and 87 are duplicates.

azurerm/resource_arm_eventhub_namespace_test.go
```diff
@@ -361,6 +386,25 @@ resource "azurerm_eventhub_namespace" "test" {
...
+  name     = "acctestRG-%d"
+  location = "%s"
+}
+
+resource "azurerm_eventhub_namespace" "test" {
+  name                = "acctesteventhubnamespace-%d"
+  location            = "${azurerm_resource_group.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  sku                 = "Basic"
+  sku                 = "Basic"
```
jgoldschmidt-apivista: Lines 401 and 401 are duplicates.

azurerm/data_source_eventhub_namespace_test.go
katbyte: A change like this usually means the line break when from `\n` <--> `\n\r`

katbyte:
Yep, this should definitely be added to a test

azurerm/data_source_eventhub_namespace_test.go
jgoldschmidt-apivista: Sorry, I don't understand. Are you saying lines 87 and 88 are not duplicates, but appear so due to github formatting? It looks to me like they are two green + lines with the same content.

katbyte: Oh my apologies, you are definitely correct. Line 88 didn't appear in the conversation view.

fernandoBRS:
Good news, thanks for the support @katbyte ! 🚀

PR #2410, opened 2018-11-29
New Resource: SignalR service `azurerm_signalr_service`

JunyiYi:
In this pull request, I introduced Azure SignalR (used to create real-time web application) to Terraform. AFAIK, there are no issues or feature requests yet.

I put it under "Messaging" section because it is somehow similar to notification hub.


azurerm/resource_arm_signalr.go
```diff
@@ -0,0 +1,212 @@
...
+	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		if !response.WasNotFound(future.Response()) {
+			return fmt.Errorf("Error deleting SignalR %q (resource group %q): %+v", name, resourceGroup, err)
+		}
+	}
+
+	return nil
+}
+
+func resourceArmSignalRCustomizeDiff(diff *schema.ResourceDiff, v interface{}) error {
```
katbyte: Very minor, not a blocker, but i typically put these right under the schema

azurerm/resource_arm_signalr.go
```diff
@@ -0,0 +1,212 @@
...
+	sku := diff.Get("sku_name").(string)
+	if sku == "Free_F1" {
+		if capacity := diff.Get("capacity").(int); capacity != 1 {
```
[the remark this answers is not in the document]
JunyiYi: Thanks @tombuildsstuff for the explanation. I'm curious about is it the best practice to do a more complicated check in `expand` functions? What would `CustomizeDiff` be used to do?

JunyiYi:
Thanks @tombuildsstuff and @katbyte for reviewing the code. I have changed it accordingly, please take a look.

azurerm/resource_arm_signalr_service.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	return nil
+}
+
+func expandSignalRServiceSku(input []interface{}) *signalr.ResourceSku {
+	v := input[0].(map[string]interface{})
+	name := v["name"].(string)
+	if name == "Free" {
+		name = "Free_F1"
```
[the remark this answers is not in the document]
JunyiYi: It is what @tombuildsstuff suggested before:

```
sku {
  name = "Free"
  capacity = 1
}
```

And for your second question, the answer is "Yes". SignalR service does not follow the convention of `sku`, they only accept `Free_F1` and `Standard_S1` no matter what `capacity` is.

azurerm/resource_arm_signalr.go
```diff
@@ -0,0 +1,212 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.NoZeroValues,
+			},
+
+			"location": locationSchema(),
+
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"sku_name": {
```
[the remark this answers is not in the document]
JunyiYi: This API only accepts `Free_F1` no matter what `capacity` is.

JunyiYi:
@tombuildsstuff @katbyte , after talking with the service team, they do need to use fixed sku names (i.e. `Free_F1` and `Standard_S1`) for all capacities, because they want to adapt capacity changes easily. Quote from service team:

> Initially `capacity` is allowed to be 1-10, now it’s currency values (1,2,5,...,100). We may add more supported numbers in future). To keep the same SKU name, we can support more (or less) units by slightly updating our UI/Validation logic while keeping the UX/interface and consistent and stable.

So I will change the `sku name` to only allow `Free_F1` and `Standard_S1`.

website/docs/r/signalr_service.html.markdown
```diff
@@ -0,0 +1,74 @@
...
+
+* `sku` - A `sku` block as documented below.
+
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+---
+
+A `sku` block supports the following:
+
+* `name` - (Required) Specifies which tier to use. Valid values are `Free` and `Standard`.
```
katbyte: We should update these too
```suggestion
* `name` - (Required) Specifies which tier to use. Valid values are `Free_F!` and `Standard_S1`.
```

website/docs/r/signalr_service.html.markdown
```diff
@@ -0,0 +1,74 @@
...
+  name     = "terraform-signalr"
+  location = "West US"
+}
+
+resource "azurerm_signalr_service" "example" {
+  name                = "tfex-signalr"
+  location            = "${azurerm_resource_group.example.location}"
+  resource_group_name = "${azurerm_resource_group.example.name}"
+  sku {
+    name     = "Free"
```
katbyte: and here
```suggestion
    name     = "Free_F1"
```

katbyte:
Tests pass:
screen shot 2018-12-12 at 10 56 17

azurerm/resource_arm_signalr_service.go
```diff
@@ -0,0 +1,212 @@
...
+	v := input[0].(map[string]interface{})
+	return &signalr.ResourceSku{
+		Name:     utils.String(v["name"].(string)),
+		Capacity: utils.Int32(int32(v["capacity"].(int))),
+	}
+}
+
+func flattenSignalRServiceSku(input *signalr.ResourceSku) []interface{} {
+	result := make(map[string]interface{})
+	if input != nil {
```
tombuildsstuff: if input's nil we're going to set an empty sku block in the state (so there'll be an object there but it'll be empty) - can we update this to:

```
if input == nil {
  return []interface{}{}
}

...
```

this way it'll show that the Sku block isn't being returned from the API from the fact that Terraform will show the block being removed, rather than the fields being blank

PR #2433, opened 2018-12-03
Upgrade `network` SDK to `2018-08-01`

metacpp:
This PR is used to upgrade to `network` SDK to `2018-08-01`, which will be used to support `network_profile` for ACI in a later PR.


azurerm/resource_arm_firewall_network_rule_collection.go
```diff
@@ -93,7 +93,6 @@ func resourceArmFirewallNetworkRuleCollection() *schema.Resource {
 								Type: schema.TypeString,
 								ValidateFunc: validation.StringInSlice([]string{
 									string(network.Any),
-									string(network.ICMP),
```
katbyte: Does this constant no longer exist or has it been renamed? Removing this value is a breaking change

azurerm/helpers/azure/firewall.go
```diff
@@ -30,7 +30,7 @@ func FirewallFixIPConfiguration(input *[]network.AzureFirewallIPConfiguration) (
 				Subnet: &network.SubResource{
 					ID: config.Subnet.ID,
 				},
-				InternalPublicIPAddress: &network.SubResource{
```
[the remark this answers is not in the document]
katbyte: It shouldn't unless the property name `internal_public_ip_address_id` is no longer indicative of what the property is. Was the field renamed or removed?

azurerm/resource_arm_firewall_network_rule_collection.go
JunyiYi: Same question, why removing it? The constant is still there: https://github.com/Azure/azure-sdk-for-go/blob/master/services/network/mgmt/2018-08-01/network/models.go#L423.

katbyte:
I agree with  @hbuckle. If `public_ip_address_id` is now the correct name of the field then it should definitely be renamed to match, `internal_public_ip_address_id` is a bit of a weird claim 🤔

azurerm/resource_arm_firewall_network_rule_collection.go
metacpp: Good catch. `ICMP` is not included in the v21.3 but included since v22.0. I will still keep the raw literal string here until we upgrade to v22.0.

metacpp: It's removed in v21.3 and added back in v22.0.

metacpp:
> I agree with @hbuckle. If `public_ip_address_id` is now the correct name of the field then it should definitely be renamed to match, `internal_public_ip_address_id` is a bit of a weird claim 🤔

I just updated the PR, please help with taking a look at it.

azurerm/helpers/azure/firewall.go
```diff
@@ -3,10 +3,10 @@ package azure
...
 
-	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-04-01/network"
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-08-01/network"
 )
 
-// The API requires InternalPublicIPAddress to be set when for a CreateOrUpdate
+// The API requires PublicIPAddress to be set when for a CreateOrUpdate
 // operation, but Get operations return the property as PublicIPAddress
 // so we need to go through and copy the value to the correct property.
 func FirewallFixIPConfiguration(input *[]network.AzureFirewallIPConfiguration) (*[]network.AzureFirewallIPConfiguration, error) {
```
tombuildsstuff: we should be able to remove this method since it's no longer necessary since the SDK's been fixed

metacpp: Agree.

metacpp:
@katbyte @tombuildsstuff just resolved the comments and updated the PRs.

azurerm/resource_arm_firewall.go
```diff
@@ -247,7 +257,16 @@ func expandArmFirewallIPConfigurations(d *schema.ResourceData) (*[]network.Azure
 		data := configRaw.(map[string]interface{})
 		name := data["name"].(string)
 		subnetId := data["subnet_id"].(string)
-		intPubID := data["internal_public_ip_address_id"].(string)
+
+		exist := false
```
hbuckle: This doesn't work as the properties are empty strings when unset, so exist is always true

```suggestion
		pubID := data["internal_public_ip_address_id"].(string)
		if pubID == "" {
			pubID = data["public_ip_address_id"].(string)
		}

		if pubID == "" {
			return nil, nil, nil, fmt.Errorf("one of `ip_configuration.0.internal_public_ip_address_id` or `ip_configuration.0.public_ip_address_id` must be set")
		}
```

PR #2437, opened 2018-12-04
example: ensuring that winrm is correctly configured

tombuildsstuff:
The default within Azure's been changed such that generated certificates are no longer defaulted to Server/Client Authorization - which is why this fails.

### DIFF
--- a/examples/virtual-machines/provisioners/windows/2-certificates.tf
+++ b/examples/virtual-machines/provisioners/windows/2-certificates.tf
@@ -62,6 +62,9 @@ resource "azurerm_key_vault_certificate" "main" {
     }
 
     x509_certificate_properties {
+      # Server Authentication = 1.3.6.1.5.5.7.3.1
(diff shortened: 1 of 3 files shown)


examples/virtual-machines/provisioners/windows/2-certificates.tf
```diff
@@ -62,6 +62,9 @@ resource "azurerm_key_vault_certificate" "main" {
     }
 
     x509_certificate_properties {
+      # Server Authentication = 1.3.6.1.5.5.7.3.1
+      # Client Authentication = 1.3.6.1.5.5.7.3.2
+      extended_key_usage = [ "1.3.6.1.5.5.7.3.1" ]
```
tombuildsstuff: whilst these comments are in a weird place, they don't really belong in the field specific docs below (but they are inherently useful, given they're interesting values)

PR #2439, opened 2018-12-04
server_app_secret and client_secret added

neumanndaniel:
`server_app_secret` and `client_secret` are missing in the documentation for AKS. If you run terraform plan, terraform complains about the missing options.

### DIFF
--- a/website/docs/d/kubernetes_cluster.html.markdown
+++ b/website/docs/d/kubernetes_cluster.html.markdown
@@ -99,6 +99,8 @@ A `azure_active_directory` block exports the following:
 
 * `server_app_id` - The Server ID of an Azure Active Directory Application.
 
+* `server_app_secret` - The Server Secret of an Azure Active Directory Application.
+
 * `tenant_id` - The Tenant ID used for Azure Active Directory Application.
 ---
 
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hi @neumanndaniel 

Thanks for this PR :)

The AzureRM Provider supports both a Resource (which provisions/manages a Kubernetes Cluster) and a Data Source (which references an existing cluster in a read-only manner) for Kubernetes Clusters. Within the documentation we prefix the titles for Data Sources with `Data Source: [name]` to try and make this clearer - however that's not necessarily sufficient as a differentiator between them unfortunately (and is something we're thinking about how to improve for the future).

Taking a quick look into this PR this updates the documentation for [the Data Source](https://www.terraform.io/docs/providers/azurerm/d/kubernetes_cluster.html) where these fields are intentionally not available (since they're not returned from the API) - however you're referring to using the [Resource](https://www.terraform.io/docs/providers/azurerm/r/kubernetes_cluster.html) where these [fields are available](https://www.terraform.io/docs/providers/azurerm/r/kubernetes_cluster.html#client_app_id).

As such whilst I'd like to thank you for this contribution - since I believe this PR was intended to target the Resource (which contains these fields) I'm going to close this PR for the moment.

Thanks!

PR #2442, opened 2018-12-04
Deprecate azurerm_app_service_plan properties block for top level fields

katbyte:


azurerm/resource_arm_app_service_plan.go
```diff
@@ -191,9 +238,13 @@ func resourceArmAppServicePlanRead(d *schema.ResourceData, meta interface{}) err
 			return fmt.Errorf("Error setting `properties`: %+v", err)
 		}
 
-		if props.MaximumNumberOfWorkers != nil {
-			d.Set("maximum_number_of_workers", int(*props.MaximumNumberOfWorkers))
+		if profile := props.HostingEnvironmentProfile; profile != nil {
+			d.Set("app_service_environment_id", profile.ID)
 		}
+
+		d.Set("maximum_number_of_workers", props.MaximumNumberOfWorkers)
```
tombuildsstuff: since this is a `*int32` we need to cast this to an int here

PR #2463, opened 2018-12-06
Add virtual machines data sources

inkel:
This pull request adds two new Virtual Machine data sources:

* `azurerm_virtual_machine` to retrieve information about a single Virtual Machine.
* `azurerm_virtual_machines` to retrieve information about many Virtual Machines.

### `azurerm_virtual_machine`
```hcl
data "azurerm_virtual_machine" "test" {
  name                = "production"
  resource_group_name = "networking"
}

output "virtual_machine_id" {
  value = "${data.azurerm_virtual_machine.test.id}"
}
```

### `azurerm_virtual_machines`
```hcl
data "azurerm_virtual_machines" "test" {
  name_prefix         = "prod-"
  resource_group_name = "networking"
}

output "virtual_machine_ids" {
  value = "${join(", ", data.azurerm_virtual_machine.test.ids)}"
}

output "virtual_machine_names" {
  value = "${join(", ", data.azurerm_virtual_machine.test.names)}"
}
```

## TODO
* [ ] add tests for `azurerm_virtual_machines`
* [ ] add additional parameters for `azurerm_virtual_machine` (e.g. `location`)
* [ ] add additional parameters for `azurerm_virtual_machines` (e.g. `location`)


tombuildsstuff:
hey @inkel 

Thanks for this PR :)

Out of interest what's the use-case you're looking to solve with this Data Source? We've previously discussed adding a Data Source for a Virtual Machine but ultimately found that most of the information you'd want to obtain from it can be found via another Data Source (e.g. the IP Addresses of the VM can be obtained using the `azurerm_network_interface` Data Source instead)

One thing of particular note here is that the `azurerm_virtual_machine` resource will be being replaced in 2.0 of the AzureRM Provider in favour of two split-out resources (one for a Linux VM / one for a Windows VM) which are going to have different schemas than the VM resource - as such I think it could make sense to to hold off and add support for this for the newer resources instead?

Thanks!

tombuildsstuff:
hey @inkel 

Taking a little time to think this through some more - I could definitely see this being useful, but only for the ID of the Virtual Machine (and not the other fields).

Based on the scenario you've described above - I believe it should be possible to achieve this either using one data source either for a single machine:

```
data "azurerm_virtual_machine" "example" {
  name                = "vm-name-1"
  resource_group_name = "example-resources"
}

output "virtual_machine_id" {
  value = "${data.azurerm_virtual_machine.example.id}"
}

module "monitoring" {
  source              = "./modules/monitoring"
  virtual_machine_ids = [ "${data.azurerm_virtual_machine.example.id}" ]
}
```

or for multiple virtual machines:

```
data "azurerm_virtual_machine" "example" {
  count               = 5
  name                = "vm-name-${count.index}"
  resource_group_name = "example-resources"
}

output "virtual_machine_id" {
  value = "${data.azurerm_virtual_machine.example.id}"
}

module "monitoring" {
  source              = "./modules/monitoring"
  virtual_machine_ids = [ "${data.azurerm_virtual_machine.example.*.id}" ]
}
```

This'd allow the Virtual Machine ID to be used with the `azurerm_virtual_machine_data_disk_attachment` resource - presumably we'd also want to update the `azurerm_virtual_machine_extension` resource to take the Virtual Machine ID in addition to/instead of the Virtual Machine Name (this has previously been requested in another issue - but I can't immediately find it at the moment)

As such I believe by using Count with the Data Source that the `azurerm_virtual_machines` Data Source wouldn't be necessary - what do you think?

Thanks!

inkel:
> As such I believe by using Count with the Data Source that the `azurerm_virtual_machines` Data Source wouldn't be necessary - what do you think?

I like this idea! Let me try it locally and see how does thing look, but I'm positive it will work. Also, it will remove the need of adding an additional data resource (and I'll get to avoid writing more tests 😛)

I'm more than ok with only returning the VM ID.

> This'd allow the Virtual Machine ID to be used with the `azurerm_virtual_machine_data_disk_attachment` resource - presumably we'd also want to update the `azurerm_virtual_machine_extension` resource to take the Virtual Machine ID in addition to/instead of the Virtual Machine Name (this has previously been requested in another issue - but I can't immediately find it at the moment)

I'm not sure I'm following you on this comment, care to explain it again to me? Thanks!

inkel:
Yup, the `data` plus `count` attribute will work more than ok. What would be the next steps, then?

* make `azurerm_virtual_machine` only return `id`
* remove `azurerm_virtual_machines`
* profit? 😬

tombuildsstuff:
@inkel 

>> This'd allow the Virtual Machine ID to be used with the azurerm_virtual_machine_data_disk_attachment resource - presumably we'd also want to update the azurerm_virtual_machine_extension resource to take the Virtual Machine ID in addition to/instead of the Virtual Machine Name (this has previously been requested in another issue - but I can't immediately find it at the moment)
> I'm not sure I'm following you on this comment, care to explain it again to me? Thanks!

There's been a feature request recently to enable the `azurerm_virtual_machine_extension` resource to allow specifying a Virtual Machine ID rather than the Virtual Machine Name as an argument - this was more thinking out loud that this would also be enabled from this Data Source too :)

> Yup, the data plus count attribute will work more than ok. What would be the next steps, then?

Yep basically as you've mentioned above - ensure the `azurerm_virtual_machine` data source does a Get to ensure the VM exists and then just set the ID (and a test) - but that should be all, let me take a quick look through 👍 

Thanks!

inkel:
I've should've known better, but the GitHub suggestion feature, awesome as it is, only applied to one line, so when I applied your suggestion to rename the function it broke the build 😞

Anyway, that's fixed now, alongside with all your other suggestions, so this should be ready for another review. Thank you @tombuildsstuff for your time and feedback!

tombuildsstuff:
Tests pass:

screenshot 2018-12-14 at 10 50 46

inkel:
Yoohoo! Thanks! It feels so good to contribute, you are an amazing maintainer! 👏 



PS: any ideas on when to expect a new release with this data source? 😬

Issue #2462, opened 2018-12-06
Crash azurerm_route_table name with only spaces in it


pedrohdz:
The `azurerm_route_table` data source is crashing `terraform` when a name consisting only of spaces is provided.

Yes, if it hurts don't do it.  ;-) . Thing is that this can make a simple bug difficult to figure out, as it was in my case. (Had to do with a string that was being processed to calculate the name. In this instance there was an error in processing resulting in a space in the name.)

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

```
Terraform v0.11.10
azurerm" (1.19.0)
```

### Affected Resource(s)

* Data source `azurerm_route_table`

### Terraform Configuration Files

```hcl
provider "azurerm" {
}

resource "azurerm_resource_group" "test" {
  name = "testResourceGroup"
  location = "West US"
}

data "azurerm_route_table" "source" {
  resource_group_name = "${azurerm_resource_group.test.name}"
  # Note the space ( ) in the name.
  name = " "
}
```

### Debug Output

[Output Gist](https://gist.github.com/pedrohdz/0e5287b2fbdd8634a7af9475a8cf1d3c)

### Panic Output

[crash.log](https://gist.github.com/pedrohdz/2f1b45944625a969844d6cf2068213e0)

### Expected Behavior

Anything but a panic?  A validation error would be the way to go.

### Actual Behavior

```
panic: runtime error: invalid memory address or nil pointer dereference
```

### Steps to Reproduce

1. `terraform init`
2. `terraform apply`

### Important Factoids

**Nope!**

### References

It seems like the issue is cropping up to the use of `validation.NoZeroValues` in [data_source_route_table.go](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/data_source_route_table.go#L20)?  Might be better to use one of the regex validators?

jeffreyCline:
@pedrohdz can you verify that the target of your data source already exists?

jeffreyCline:
[...]

PR #2477, opened 2018-12-08
Change all references from NoZeroValues to NoEmptyStrings

jeffreyCline:
Ref #2462


jeffreyCline:
It ended up being VS Code... while my personal git account had the correct settings, the global settings were not set, which are the settings VS Code uses to do the commit to the repository. I had to set the global settings by doing a `git config --global core.autocrlf true` and then re-submit the code, which appears to have fixed the line endings getting switched issue.

PR #2490, opened 2018-12-11
assign policy to a managementgroup

lawrenae:
This pull request adds the ability to optionally define a policy on a **management group** level. The existing functionality -- define on a subscription or resource group is still intact.

I've done so by adding `management_group_id` to `azurerm_policy_definition`. IE:

```hcl
data "azurerm_management_group" "test" {
  group_id = "00000000-0000-0000-0000-000000000000"
}

resource "azurerm_policy_definition" "policy" {
  name         = "accTestPolicy"
  policy_type  = "Custom"
  mode         = "Indexed"
  display_name = "acceptance test policy definition"
  management_group_id = "${data.azurerm_management_group.test.id}"
...
}
```

btw, `azurerm_policy_assignment` will need no change to support this.

Feedback most welcome -- this is my first PR for the azurerm terraform provider

### DIFF
--- a/azurerm/resource_arm_policy_definition.go
+++ b/azurerm/resource_arm_policy_definition.go
@@ -61,6 +61,7 @@ func resourceArmPolicyDefinition() *schema.Resource {
 			"management_group_id": {
 				Type:     schema.TypeString,
 				Optional: true,
+				ForceNew: true,
 			},
 
 			"display_name": {
@@ -170,7 +171,7 @@ func resourceArmPolicyDefinitionCreateUpdate(d *schema.ResourceData, meta interf
 		return fmt.Errorf("Error waiting for Policy Definition %q to become available: %s", name, err)
 	}
 
-	resp, err := getPolicy(ctx, client, name, managementGroupID)
+	resp, err := getPolicyDefinition(ctx, client, name, managementGroupID)
 	if err != nil {
 		return err
 	}
@@ -189,9 +190,9 @@ func resourceArmPolicyDefinitionRead(d *schema.ResourceData, meta interface{}) e
 		return err
 	}
 
-	managementGroupID := d.Get("management_group_id").(string)
+	managementGroupID := parseManagementGroupIdFromPolicyId(d.Id())
 
-	resp, err := getPolicy(ctx, client, name, managementGroupID)
+	resp, err := getPolicyDefinition(ctx, client, name, managementGroupID)
 
 	if err != nil {
 		if utils.ResponseWasNotFound(resp.Response) {
@@ -255,7 +256,7 @@ func resourceArmPolicyDefinitionDelete(d *schema.ResourceData, meta interface{})
 		return err
 	}
 
-	managementGroupID := d.Get("management_group_id").(string)
+	managementGroupID := parseManagementGroupIdFromPolicyId(d.Id())
 
 	var resp autorest.Response
 	if managementGroupID == "" {
@@ -285,9 +286,20 @@ func parsePolicyDefinitionNameFromId(id string) (string, error) {
 	return components[len(components)-1], nil
 }
 
+func parseManagementGroupIdFromPolicyId(id string) string {
+	r, _ := regexp.Compile("managementgroups/(.+)/providers/.*$")
+
+	if r.MatchString(id) {
+		parms := r.FindAllStringSubmatch(id, -1)[0]
+		return parms[1]
+	}
+
+	return ""
+}
+
 func policyDefinitionRefreshFunc(ctx context.Context, client policy.DefinitionsClient, name string, managementGroupID string) resource.StateRefreshFunc {
 	return func() (interface{}, string, error) {
-		res, err := getPolicy(ctx, client, name, managementGroupID)
+		res, err := getPolicyDefinition(ctx, client, name, managementGroupID)
 
 		if err != nil {
 			return nil, strconv.Itoa(res.StatusCode), fmt.Errorf("Error issuing read request in policyAssignmentRefreshFunc for Policy Assignment %q: %s", name, err)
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_policy_definition_test.go
```diff
@@ -85,14 +144,29 @@ func testCheckAzureRMPolicyDefinitionDestroy(s *terraform.State) error {
 	return nil
 }
 
-func testAzureRMPolicyDefinition_basic(ri int) string {
-	return fmt.Sprintf(`
+func testAzureRMPolicyDefinition_basic(ri int, managementGroup bool) string {
```
tombuildsstuff: rather than adding the extra logic inside of this, could we revert this function and add a new one which returns the policy definition inside of a management group?

website/docs/r/policy_definition.html.markdown
```diff
@@ -67,6 +69,8 @@ The following arguments are supported:
 
 * `description` - (Optional) The description of the policy definition.
 
+* `management_group_id` - (Optional) the management group id (IE: `${data.azurerm_management_group.test.group_id`) where where to define this policy
```
tombuildsstuff: We tend to phrase this as:

```suggestion
* `management_group_id` - (Optional) The ID of the Management Group where this policy should be defined. Changing this forces a new resource to be created.
```

azurerm/resource_arm_policy_definition.go
```diff
@@ -56,6 +58,11 @@ func resourceArmPolicyDefinition() *schema.Resource {
 				}, true),
 			},
 
+			"management_group_id": {
+				Type:     schema.TypeString,
+				Optional: true,
```
tombuildsstuff: since it's not possible to migrate between policy definitions between management group's, I think this also wants:

```suggestion
				Optional: true,
				ForceNew: true,
```

azurerm/resource_arm_policy_definition_test.go
```diff
@@ -35,6 +36,64 @@ func TestAccAzureRMPolicyDefinition_basic(t *testing.T) {
...
+		if !ok {
+			return fmt.Errorf("not found: %s", policyName)
+		}
+
+		policyName := rs.Primary.Attributes["name"]
+
+		rs, ok = s.RootModule().Resources[managementGroupName]
+		if !ok {
+			return fmt.Errorf("not found: %s", managementGroupName)
+		}
```
tombuildsstuff: since the `management_group_id` key is available on the resource, we should be able to use this instead of querying the (separate) management group resource - as such I don't think this is needed?

website/docs/r/policy_definition.html.markdown
lawrenae: I'll update as suggested, but want to leave in the example. I was confused myself at first because `azurerm_management_group` has both an `.id` and a `.group_id`. 

Alternatively, I could modify it to something like "The Group ID (and *not* the ID) of the Management Group.."

Thoughts?

azurerm/resource_arm_policy_definition.go
```diff
@@ -235,7 +255,14 @@ func resourceArmPolicyDefinitionDelete(d *schema.ResourceData, meta interface{})
 		return err
 	}
 
-	resp, err := client.Delete(ctx, name)
+	managementGroupID := d.Get("management_group_id").(string)
```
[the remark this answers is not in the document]
lawrenae: makes sense -- updating this caused a bit better refactoring elsewhere to.

azurerm/resource_arm_policy_definition.go
```diff
@@ -255,20 +282,41 @@ func parsePolicyDefinitionNameFromId(id string) (string, error) {
...
+
+func resourceArmPolicyDefinitionImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	r, _ := regexp.Compile("managementgroups/(.+)/providers/.*/policyDefinitions/(.+)$")
+	if r.MatchString(d.Id()) {
+		parms := r.FindAllStringSubmatch(d.Id(), -1)[0]
+		d.Set("management_group_id", parms[1])
+		d.Set("name", parms[2])
+
+		return []*schema.ResourceData{d}, nil
+	} else {
```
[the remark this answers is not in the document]
lawrenae: Yes, agreed on removing the `else`.

azurerm/resource_arm_policy_definition.go
```diff
@@ -255,20 +282,41 @@ func parsePolicyDefinitionNameFromId(id string) (string, error) {
...
+		//import a subscription based policy as before
+		return schema.ImportStatePassthrough(d, meta)
+	}
+}
+
+func getPolicy(ctx context.Context, client policy.DefinitionsClient, name string, managementGroupID string) (res policy.Definition, err error) {
```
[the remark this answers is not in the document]
lawrenae: updated

azurerm/resource_arm_policy_definition_test.go
```diff
@@ -35,6 +36,64 @@ func TestAccAzureRMPolicyDefinition_basic(t *testing.T) {
...
+
+		managementGroupID := rs.Primary.Attributes["group_id"]
```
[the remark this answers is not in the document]
lawrenae: yes, this noticeably simplified the test. thanks

azurerm/resource_arm_policy_definition_test.go
lawrenae: yes, thats not as DRY, but much clearer. done.

azurerm/resource_arm_policy_definition.go
lawrenae: updated.

azurerm/resource_arm_policy_definition_test.go
lawrenae: yes, this noticeably simplified the test. thanks

website/docs/r/policy_definition.html.markdown
katbyte: If you'd like to add that information i would add it in a notes box under that line, or to the example at the top of this page

lawrenae: ok. cool.

lawrenae:
@tombuildsstuff I've incorporated the feedback -- let me know what you think!

katbyte:
tests pass:
screen shot 2018-12-27 at 18 47 15

PR #2511, opened 2018-12-13
Requiring Import (PR 1 of ...)

tombuildsstuff:
Whilst the Requiring Import change can be feature-toggled; unfortunately the Timeouts change can't - as such we're going to roll out Requiring Import and Timeouts separately.

This is feature-toggled off by default, but can be enabled by an Environment Variable for testing (which we'll enable in 2.0)

Supersedes #1746

### DIFF
--- a/azurerm/feature_flags.go
+++ b/azurerm/feature_flags.go
@@ -1,0 +1,4 @@
+package azurerm
+
+// This file contains feature flags for functionality which will prove more challenging to implement en-mass
+var requireResourcesToBeImported = false
--- a/azurerm/feature_flags.go
+++ b/azurerm/feature_flags.go
@@ -1,4 +1,9 @@
 package azurerm
 
+import (
+	"os"
+	"strings"
+)
+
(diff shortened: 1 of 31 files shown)


tombuildsstuff:
Given this is feature toggled off (and this is a low-risk change across multiple resources) we'll handle the test output for this via the nightly acceptance tests

pixelicous:
@tombuildsstuff @katbyte so this is going to be implemented only on terraform 2.0 with provider 1.21? if you use this feature toggle it means you will need to set timeouts for all resources from there on out?

thanks in advance

tombuildsstuff:
@pixelicous there's two bits which have been split out from the other PR here: requiring resources to be imported when they already exist (which is what's partially covered in this PR an is feature-toggled off for the moment); the other is the custom timeouts, which isn't covered in this PR and I'm unsure if we can feature-toggle; but in either case both of those features will land in version 2.0 of the AzureRM Provider (the version of Terraform Core is unrelated) :)

PR #2513, opened 2018-12-14
Add `possible_outbound_ip_addresses` to app service

tomasaschan:
This is useful for e.g. setting up firewall rules for Azure SQL - since an App Service can be rescheduled, it's possible that the `outbound_ip_addresses` field output at deploy-time does not catch all IP addresses you need to open up to be future-safe.

Note: I couldn't get the code to build and run tests on my Windows machine (neither from PowerShell, inspecting the makefile to see what was required, nor from WSL), so I've basically just looked at #700 to see what I need to do to support this. (I think this could have been made easier with a `.gitattributes` and an `.editorconfig` at the project root, that - at the very least - specify line endings to be what we expect.) Please feel free to request changes if I missed something 😄

### DIFF
--- a/azurerm/data_source_app_service.go
+++ b/azurerm/data_source_app_service.go
@@ -101,6 +101,11 @@ func dataSourceArmAppService() *schema.Resource {
 				Computed: true,
 			},
 
+			"possible_outbound_ip_addresses": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
 			"source_control": {
 				Type:     schema.TypeList,
 				Computed: true,
--- a/azurerm/data_source_app_service.go
+++ b/azurerm/data_source_app_service.go
@@ -183,6 +183,7 @@ func dataSourceArmAppServiceRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("https_only", props.HTTPSOnly)
 		d.Set("default_site_hostname", props.DefaultHostName)
 		d.Set("outbound_ip_addresses", props.OutboundIPAddresses)
+		d.Set("possible_outbound_ip_addresses", props.PossibleOutboundIPAddresses)
 	}
 
 	if err := d.Set("app_settings", flattenAppServiceAppSettings(appSettingsResp.Properties)); err != nil {
--- a/website/docs/r/app_service.html.markdown
+++ b/website/docs/r/app_service.html.markdown
@@ -205,6 +205,8 @@ The following attributes are exported:
 
 * `outbound_ip_addresses` - A comma separated list of outbound IP addresses - such as `52.23.25.3,52.143.43.12`
 
+* `possible_outbound_ip_addresses` - A comma separated list of outbound IP addresses - such as `52.23.25.3,52.143.43.12,52.143.43.17` - not all of which are necessarily in use. Superset of `outbound_ip_addresses`.
+
 * `source_control` - A `source_control` block as defined below, which contains the Source Control information when `scm_type` is set to `LocalGit`.
 
 * `site_credential` - A `site_credential` block as defined below, which contains the site-level credentials used to publish to this App Service.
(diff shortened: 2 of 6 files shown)


katbyte:
@tomasaschan,

I think this may be better to export as a `TypeList` rather than a comma separated field. However in that case we should probably also update `oubound_ip_addresses` to match?

tomasaschan:
Yeah, I actually thought so to, but decided to follow #700 as closely as possible. The rationale in the discussion there was that it's easy enough to get that list using `split` if one needs it.

I think it's reasonable that they match, but I don't really care which way as long as they both go the same way :) But since changing `outbound_ip_addresses` would be a breaking change, while this isn't, I also figured this can be added with less friction, and a new PR that changes the return type of both attributes can be made in the 2.0 scope.

katbyte:
Rational decision to me @tomasaschan, CSV for now it is!

azurerm/data_source_app_service.go
```diff
@@ -183,6 +183,7 @@ func dataSourceArmAppServiceRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("https_only", props.HTTPSOnly)
 		d.Set("default_site_hostname", props.DefaultHostName)
 		d.Set("outbound_ip_addresses", props.OutboundIPAddresses)
+		d.Set("possible_outbound_ip_addresses", props.PossibleOutboundIPAddresses)
```
metacpp: It looks like that you missed the `possible_outbound_ip_addresses` in the schema ?

metacpp: It looks like that you missed the `possible_outbound_ip_addresses` in the schema ?

website/docs/r/app_service.html.markdown
```diff
@@ -205,6 +205,8 @@ The following attributes are exported:
 
 * `outbound_ip_addresses` - A comma separated list of outbound IP addresses - such as `52.23.25.3,52.143.43.12`
 
+* `possible_outbound_ip_addresses` - A comma separated list of outbound IP addresses - such as `52.23.25.3,52.143.43.12,52.143.43.17` - not all of which are necessarily in use. Superset of `outbound_ip_addresses`.
```
metacpp: It looks like that you missed adding this to the documentation for the data source.

katbyte:
For the test you could just add it to the basic one, like:

```go
func TestAccDataSourceAzureRMAppService_basic(t *testing.T) {
	dataSourceName := "data.azurerm_app_service.test"
	rInt := acctest.RandInt()
	location := testLocation()

		PreCheck:  func() { testAccPreCheck(t) },
		Providers: testAccProviders,
		Steps: []resource.TestStep{
			{
				Config: testAccDataSourceAppService_basic(rInt, location),
					resource.TestCheckResourceAttrSet(dataSourceName, "app_service_plan_id"),
					resource.TestCheckResourceAttrSet(dataSourceName, "possible_outbound_ip_addresses"),
					resource.TestCheckResourceAttrSet(dataSourceName, "outbound_ip_addresses"),
					resource.TestCheckResourceAttr(dataSourceName, "tags.%", "0"),
				),
			},
		},
	})
}
```

tomasaschan:
@katbyte @metacpp Thanks for the review! I honestly hadn't thought about the data source getting the same properties, but of course that makes sense. I've added the new property to the schema for the data source as well.

I've also added a few assertions to the basic tests for both resource and data source, as well as documentation of these properties to the data source (it seems the reason I missed it in the first place is that the `outbound_ip_addresses` property was not documented either).

tombuildsstuff:
Data Source tests pass:

screenshot 2018-12-20 at 13 32 11

tombuildsstuff:
Resource tests pass:

screenshot 2018-12-20 at 14 12 18

tomasaschan:
Woop woop! Thanks for taking this - will significantly simplify my deployments, once released 🎉

Issue #2516, opened 2018-12-14
[azurerm_redis_cache] imported state no fill "maxmemory_policy"


raphaelquati:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

```
Terraform v0.11.10
+ provider.azurerm v1.20.0
```

### Affected Resource(s)
azurerm_redis_cache

### Terraform Configuration Files
```hcl
resource "azurerm_redis_cache" "redis" {
    name = "teste"
    resource_group_name = "${azurerm_resource_group.teste.name}"
    location = "${azurerm_resource_group.teste.location}"
    enable_non_ssl_port = false
    capacity = 0
    family = "C"
    sku_name = "Basic"
    redis_configuration  { 
        maxmemory_policy = "volatile-lru"
        maxmemory_delta = 2
        maxmemory_reserved = 2
    }
}
```

### Expected Behavior
Field maxmemory_policy filled.
### Actual Behavior
maxmemory_policy = ""

### Steps to Reproduce
terraform import azurerm_redis_cache.redis /subscriptions/xxxx/resourceGroups/xxxxx/providers/Microsoft.Cache/Redis/teste
terraform plan

### Notes

After import, maxmemory_policy  in state is empty

![image](https://user-images.githubusercontent.com/815086/50025398-913e7180-ffcc-11e8-953e-955f9fbd0542.png)

Checking the info in Azure Portal:

![image](https://user-images.githubusercontent.com/815086/50025462-c34fd380-ffcc-11e8-85af-7cb08e58d69c.png)

But calling Azure REST API's, the maxmemory_policy is missing!

![image](https://user-images.githubusercontent.com/815086/50025544-0c078c80-ffcd-11e8-88d4-39d0598f0b43.png)

Azurerm_redis_cache forces "maxmemory_policy" to have a value. So It's impossible to exec **terraform apply** after a **terraform import**: Azure rejects the "update":

![image](https://user-images.githubusercontent.com/815086/50025782-d3b47e00-ffcd-11e8-9339-476a96e5f13f.png)

katbyte:
[...]

PR #2517, opened 2018-12-14
Including default value for maxmemory-policy

raphaelquati:
As noted in docs (https://docs.microsoft.com/en-us/azure/azure-cache-for-redis/cache-configure), volatile-lru is the default value for maxmemory-policy.

Fix for #2516


katbyte:
@raphaelquati,

As this is not a bug in the provider i am going to close this. I have added the details of why in #2517.

PR #2521, opened 2018-12-17
Added new data source "azurerm_ip_ranges"

gerardaus:
This pull request includes a new datasource called "azurerm_ip_ranges". This allows a user to pull in the Compute IP address ranges (including SQL ranges) used by the Microsoft Azure Datacenters.

### DIFF
--- a/azurerm/data_source_ip_ranges.go
+++ b/azurerm/data_source_ip_ranges.go
@@ -1,0 +1,123 @@
+package azurerm
+
+import (
+	"encoding/xml"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"sort"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"golang.org/x/net/html"
+)
+
+type Result struct {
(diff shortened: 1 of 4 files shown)


tombuildsstuff:
hey @gerardaus 

Thanks for this PR - apologies for the delayed review here, we're still trying to catch up from over the Christmas break.

Out of interest which use-case are you looking to solve with this data? Within Azure it's possible to configure the firewall to only accept permissions from Azure IP Addresses by using the IP Address `0.0.0.0` (which refers to the Azure Data Centres) - so I'm wondering if that would be suitable instead?

Thanks!

gerardaus:
hi @tombuildsstuff 

Did not know that was possible.  Thanks.

I guess the other use case is from outside of Azure, from other clouds or on-prem not directly networked with Azure.

katbyte:
@gerardaus,

I'm happy to hear `0.0.0.0` will solve this problem for you. In many cross-cloud instances the azure resources will export more specific IP's directly related to that resource that can be used and I think we should encourage their use rather then a blanket "all of azure's IPs" list.

As such i am going to close this, please let us know if there is a specific use missed that only this will solve.

PR #2527, opened 2018-12-17
updated "id" description

asubmani:
in Azure every resource has a unique "resource Id". The `id` attribute in this datasource incorecctly mentions Id as being the Log Analytics Workspace Id.  Here `Id` outputs the unique "Azure Resource Id" of the Log Analytics Workspace.


katbyte:
Hi @asubmani,

Thank you for opening this PR. While I do think this clarifies what the ID represents, the majority of the existing documentation uses the old phrasing. What do you think @tombuildsstuff ?

PR #2532, opened 2018-12-18
Firewall app rules

hbuckle:
Adds support for Azure firewall application rules. I'll follow up with the docs shortly
I wasn't sure how to get the `ConflictsWith` to work with lists of more than one element - any pointers?


katbyte:
@hbuckle,

It will work just fine. Because your saying that property `a` conflicts with property `b.0` or `b.0.c` it does not matter how many items are in list b. There will always be first, 0, element 🙂

hbuckle:
@katbyte the requirement here is that properties of items in the list need to conflict with each other, but there could be multiple items in the list.
So this example config should be valid

```hcl
resource "azurerm_firewall_application_rule_collection" "test" {
  name                = "acctestarc"
  azure_firewall_name = "${azurerm_firewall.test.name}"
  priority            = 100
  action              = "Allow"

  rule {
    name = "rule1"

    source_addresses = [
      "192.168.2.1",
    ]

    target_fqdns = [
      "*.microsoft.com",
    ]

    protocol {
      port = 8443
      type = "Https"
    }

    protocol {
      port = 8080
      type = "Http"
    }
  }

  rule {
    name = "rule2"

    source_addresses = [
      "10.0.0.0/16",
    ]

    fqdn_tags = [
      "test",
    ]
  }
}
```

but using `rule.0.` produces the error `Error: azurerm_firewall_application_rule_collection.test: "rule.1.fqdn_tags": conflicts with rule.0.target_fqdns` - so instead of `.0.` I need something that refers to the current item in the list

katbyte:
Ah sorry @hbuckle, i misunderstood.

unfortunately i think they best way is to just check that in the create/update function and return an error if there is there is a conflict. `CustomizeDiff` could also be used to catch these errors during a plan, however it won't work for interpolated values so we are avoiding it at the moment.

hbuckle:
OK I've added the validation into `expandArmFirewallApplicationRules`
I'm not sure if that's the best place for it as it only triggers when the rules are created, not during the plan phase

azurerm/resource_arm_firewall_application_rule_collection_test.go
```diff
@@ -0,0 +1,746 @@
...
+		}
+
+		read.AzureFirewallPropertiesFormat.ApplicationRuleCollections = &rules
+		future, err := client.CreateOrUpdate(ctx, resourceGroup, firewallName, read)
+		if err != nil {
+			return fmt.Errorf("Error removing Application Rule Collection from Firewall: %+v", err)
+		}
+
+		err = future.WaitForCompletionRef(ctx, client.Client)
+		if err != nil {
```
katbyte: _minor_ these lines could be combined:
```go
		if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
```

azurerm/resource_arm_firewall_application_rule_collection.go
```diff
@@ -0,0 +1,448 @@
...
+						"name": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validation.NoZeroValues,
+						},
+						"description": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"source_addresses": {
```
katbyte: Could we move this up one so all the required values are at the top?

azurerm/resource_arm_firewall_application_rule_collection.go
```diff
@@ -0,0 +1,448 @@
...
+							Optional: true,
+							MinItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"port": {
+										Type:         schema.TypeInt,
+										Optional:     true,
+										ValidateFunc: validate.PortNumber,
+									},
+									"type": {
```
katbyte: Same here, required properties should be at the top.

azurerm/resource_arm_firewall_application_rule_collection.go
```diff
@@ -0,0 +1,448 @@
...
+	azureRMLockByName(firewallName, azureFirewallResourceName)
+	defer azureRMUnlockByName(firewallName, azureFirewallResourceName)
+
+	firewall, err := client.Get(ctx, resourceGroup, firewallName)
+	if err != nil {
+		return fmt.Errorf("Error retrieving Firewall %q (Resource Group %q): %+v", firewallName, resourceGroup, err)
+	}
+
+	if firewall.AzureFirewallPropertiesFormat == nil {
+		return fmt.Errorf("Error expanding Firewall %q (Resource Group %q): `properties` was nil", firewallName, resourceGroup)
```
katbyte: Would this read better as:
```suggestion
		return fmt.Errorf("Unable to parse Firewall %q (Resource Group %q): `properties` was nil", firewallName, resourceGroup)
```

azurerm/resource_arm_firewall_application_rule_collection.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+	props := *firewall.AzureFirewallPropertiesFormat
+
+	if props.ApplicationRuleCollections == nil {
+		return fmt.Errorf("Error expanding Firewall %q (Resource Group %q): `properties.ApplicationRuleCollections` was nil", firewallName, resourceGroup)
```
katbyte: Same here:
```suggestion
		return fmt.Errorf("Unable to parse Firewall %q (Resource Group %q): `properties.ApplicationRuleCollections` was nil", firewallName, resourceGroup)
```

azurerm/resource_arm_firewall_application_rule_collection.go
```diff
@@ -0,0 +1,448 @@
...
+
+	firewall.AzureFirewallPropertiesFormat.ApplicationRuleCollections = &ruleCollections
+
+	future, err := client.CreateOrUpdate(ctx, resourceGroup, firewallName, firewall)
+	if err != nil {
+		return fmt.Errorf("Error creating/updating Application Rule Collection %q in Firewall %q (Resource Group %q): %+v", name, firewallName, resourceGroup, err)
+	}
+
+	err = future.WaitForCompletionRef(ctx, client.Client)
+	if err != nil {
```
katbyte: _minor_ but these could be merged:
```go
	if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
```

azurerm/resource_arm_firewall_application_rule_collection.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+	props.ApplicationRuleCollections = &applicationRules
+
+	future, err := client.CreateOrUpdate(ctx, resourceGroup, firewallName, firewall)
+	if err != nil {
+		return fmt.Errorf("Error deleting Application Rule Collection %q from Firewall %q (Resource Group %q): %+v", name, firewallName, resourceGroup, err)
+	}
+
+	err = future.WaitForCompletionRef(ctx, client.Client)
+	if err != nil {
```
katbyte: _minor_ these could be on the same line:

```go
	if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
```

azurerm/resource_arm_firewall_application_rule_collection.go
```diff
@@ -0,0 +1,448 @@
...
+
+		ruleTargetFqdns := make([]string, 0)
+		for _, v := range rule["target_fqdns"].(*schema.Set).List() {
+			ruleTargetFqdns = append(ruleTargetFqdns, v.(string))
+		}
+
+		output := network.AzureFirewallApplicationRule{
+			Name:            utils.String(ruleName),
+			Description:     utils.String(ruleDescription),
+			SourceAddresses: &ruleSourceAddresses,
```
katbyte: This could be replaced with:

```go
SourceAddresses: utils.ExpandStringArray(rule["source_addresses"].(*schema.Set).List())
```

azurerm/resource_arm_firewall_application_rule_collection.go
```diff
@@ -0,0 +1,448 @@
...
+	outputs := make([]network.AzureFirewallApplicationRule, 0)
+
+	for _, input := range inputs {
+		rule := input.(map[string]interface{})
+
+		ruleName := rule["name"].(string)
+		ruleDescription := rule["description"].(string)
+
+		ruleSourceAddresses := make([]string, 0)
+		for _, v := range rule["source_addresses"].(*schema.Set).List() {
```
katbyte: We have a function that handles this common functionality: `ExpandStringArray`. All these loops could be replaced with it i believe.

azurerm/resource_arm_firewall_application_rule_collection.go
```diff
@@ -0,0 +1,448 @@
...
+	for _, rule := range *rules {
+		output := make(map[string]interface{})
+		if ruleName := rule.Name; ruleName != nil {
+			output["name"] = *ruleName
+		}
+		if ruleDescription := rule.Description; ruleDescription != nil {
+			output["description"] = *ruleDescription
+		}
+		if ruleSourceAddresses := rule.SourceAddresses; ruleSourceAddresses != nil {
+			output["source_addresses"] = sliceToSet(*ruleSourceAddresses)
```
katbyte: Could we move this to `helpers/set/set.go` as it's now used in multiple files?

azurerm/resource_arm_firewall_application_rule_collection.go
katbyte: You could also write a complimentary `set.ToStringSlice(i interface{}{})` function that explicitly knows about sets.

azurerm/resource_arm_firewall_application_rule_collection_test.go
```diff
@@ -0,0 +1,746 @@
...
+					testCheckAzureRMFirewallApplicationRuleCollectionExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "name", "acctestarc"),
+					resource.TestCheckResourceAttr(resourceName, "priority", "100"),
+					resource.TestCheckResourceAttr(resourceName, "action", "Allow"),
+					resource.TestCheckResourceAttr(resourceName, "rule.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.#", "2"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.0.port", "8080"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.0.type", "Http"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.port", "8443"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.port", "Https"),
```
katbyte: Should this be
```suggestion
					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.type", "Https"),
```

azurerm/resource_arm_firewall_application_rule_collection_test.go
```diff
@@ -0,0 +1,746 @@
...
+					testCheckAzureRMFirewallApplicationRuleCollectionExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "name", "acctestarc"),
+					resource.TestCheckResourceAttr(resourceName, "priority", "100"),
+					resource.TestCheckResourceAttr(resourceName, "action", "Allow"),
+					resource.TestCheckResourceAttr(resourceName, "rule.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.#", "2"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.0.port", "8000"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.0.type", "Http"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.port", "8001"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.port", "Https"),
```
katbyte:
```suggestion
					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.type", "Https"),
```

azurerm/resource_arm_firewall_application_rule_collection_test.go
```diff
@@ -0,0 +1,746 @@
...
+					testCheckAzureRMFirewallApplicationRuleCollectionExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "name", "acctestarc"),
+					resource.TestCheckResourceAttr(resourceName, "priority", "100"),
+					resource.TestCheckResourceAttr(resourceName, "action", "Allow"),
+					resource.TestCheckResourceAttr(resourceName, "rule.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.#", "2"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.0.port", "9000"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.0.type", "Https"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.port", "9001"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.port", "Http"),
```
katbyte:
```suggestion
					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.type", "Http"),
```

azurerm/resource_arm_firewall_application_rule_collection_test.go
```diff
@@ -0,0 +1,746 @@
...
+					testCheckAzureRMFirewallApplicationRuleCollectionExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "name", "acctestarc"),
+					resource.TestCheckResourceAttr(resourceName, "priority", "100"),
+					resource.TestCheckResourceAttr(resourceName, "action", "Allow"),
+					resource.TestCheckResourceAttr(resourceName, "rule.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.#", "2"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.0.port", "8000"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.0.type", "Http"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.port", "8001"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.port", "Https"),
```
katbyte:
```suggestion
					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.type", "Https"),
```

azurerm/resource_arm_firewall_application_rule_collection_test.go
```diff
@@ -0,0 +1,746 @@
...
+			{
+				Config: testAccAzureRMFirewallApplicationRuleCollection_multipleProtocols(ri, location),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMFirewallApplicationRuleCollectionExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "name", "acctestarc"),
+					resource.TestCheckResourceAttr(resourceName, "priority", "100"),
+					resource.TestCheckResourceAttr(resourceName, "action", "Allow"),
+					resource.TestCheckResourceAttr(resourceName, "rule.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.#", "2"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.0.port", "8080"),
```
katbyte: I think this should be
```suggestion
					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.0.port", "8000"),
```

azurerm/resource_arm_firewall_application_rule_collection_test.go
```diff
@@ -0,0 +1,746 @@
...
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.0.type", "Http"),
+					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.port", "8443"),
```
katbyte: and:
```suggestion
					resource.TestCheckResourceAttr(resourceName, "rule.0.protocol.1.port", "8001"),
```

hbuckle:
I think I've fixed up all the test attributes, and also updated the utility functions

azurerm/resource_arm_firewall_application_rule_collection.go
katbyte: We are not in an expand function here so I feel like a different term should be used. But it's rather minor and not a big deal 🙂 I'll leave it up to you.

hbuckle: I went with `Error retrieving` then it matches the Read function

katbyte:
tests pass:
screen shot 2019-01-03 at 16 17 58

PR #2565, opened 2018-12-21
azurerm_role_assignment: fix import when use role name instead of id

katbyte:
attempt to fix:

``` error. Status=409 Code="RoleAssignmentExists" Message="The role assignment already exists."
        --- PASS: TestAccAzureRMRoleAssignment/basic/custom (42.86s)
        --- PASS: TestAccAzureRMRoleAssignment/basic/builtin (32.28s)
        --- FAIL: TestAccAzureRMRoleAssignment/basic/roleName (23.27s)
        	testing.go:538: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        		
        		(map[string]string) {
        		}
        		
        		
        		(map[string]string) (len=1) {
        		 (string) (len=20) "role_definition_name": (string) (len=20) "Log Analytics Reader"
        		}
```

### DIFF
--- a/azurerm/resource_arm_role_assignment.go
+++ b/azurerm/resource_arm_role_assignment.go
@@ -2,6 +2,7 @@ package azurerm
 
 import (
 	"fmt"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/suppress"
 	"log"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
Tests pass:

screenshot 2018-12-23 at 14 56 28

PR #2577, opened 2018-12-27
Format remaining acc test TF blocks

katbyte:
formatted blocks with unquoted printf placeholders


azurerm/resource_arm_automation_schedule_test.go
```diff
@@ -445,8 +445,8 @@ resource "azurerm_automation_schedule" "test" {
   automation_account_name = "${azurerm_automation_account.test.name}"
   frequency               = "Month"
   interval                = "1"
-  month_days              = [%d]
-}	
+  month_days              = ["%d"]
```
tombuildsstuff: since this is a digit I don't think we need to quote this?

```suggestion
  month_days              = [%d]
```

katbyte: Tool doesn't handle arrays properly without quotes, so this was a quick hack to get it to work. reverting

PR #2583, opened 2018-12-30
Added Speech Services Kind

lqdev:
When trying to provision a resource of `Speech` kind, there are errors. To correct for this, I added kind `SpeechServices` which should be compatible with the kind being used by Azure currently for that resource.

### DIFF
--- a/azurerm/resource_arm_cognitive_account.go
+++ b/azurerm/resource_arm_cognitive_account.go
@@ -59,6 +59,7 @@ func resourceArmCognitiveAccount() *schema.Resource {
 					"Recommendations",
 					"SpeakerRecognition",
 					"Speech",
+					"SpeechServices",
 					"SpeechTranslation",
 					"TextAnalytics",
 					"TextTranslation",
--- a/azurerm/resource_arm_cognitive_account_test.go
+++ b/azurerm/resource_arm_cognitive_account_test.go
@@ -38,6 +38,33 @@ func TestAccAzureRMCognitiveAccount_basic(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMCognitiveAccount_speechServices(t *testing.T) {
+	resourceName := "azurerm_cognitive_account.test"
+	ri := acctest.RandInt()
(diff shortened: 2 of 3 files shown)


katbyte:
Hi @lqdev,

Thank you for this addition. Looking at the resource & SDK its clear it needs to be updated to use the SDK enum values as I don't see `Speech`, but then I also don't see `SpeechServices`. Is this a new addition to the service? because the possible values the SDK recognizes are:

```go
	// BingSpeech ...
	BingSpeech Kind = "Bing.Speech"
	// CustomSpeech ...
	CustomSpeech Kind = "CustomSpeech"
	// SpeakerRecognition ...
	SpeakerRecognition Kind = "SpeakerRecognition"
	// SpeechTranslation ...
	SpeechTranslation Kind = "SpeechTranslation"
```

The documentation will require updating with any new values.

katbyte (dismissed):
Hi @lqdev,

Thank you for this addition. Looking at the resource & SDK its clear it needs to be updated to use the SDK enum values as I don't see `Speech`, but then I also don't see `SpeechServices`. Is this a new addition to the service? because the possible values the SDK recognizes are:

```go
	// BingSpeech ...
	BingSpeech Kind = "Bing.Speech"
	// CustomSpeech ...
	CustomSpeech Kind = "CustomSpeech"
	// SpeakerRecognition ...
	SpeakerRecognition Kind = "SpeakerRecognition"
	// SpeechTranslation ...
	SpeechTranslation Kind = "SpeechTranslation"
```

The documentation will require updating with any new values.

neilbailie:
Hi,

I've hit this problem before and raised an issue see -> https://github.com/terraform-providers/terraform-provider-azurerm/issues/2294 which covers the origin of the missing kind.

Neil

katbyte:
@lqdev,

We have updated the SDK so this should be unblocked now 🙂

lqdev:
Thanks! @katbyte

tombuildsstuff (approved):
LGTM - thanks for this @lqdev! I hope you don't mind but I've pushed a commit adding a test and docs for this new value

PR #2592, opened 2019-01-03
Requiring Imports (PR 4 of ...)

tombuildsstuff:
Continuing down to the end of the `F` resources

### DIFF
--- a/azurerm/resource_arm_application_insights_api_key.go
+++ b/azurerm/resource_arm_application_insights_api_key.go
@@ -85,7 +85,7 @@ func resourceArmApplicationInsightsAPIKeyCreate(d *schema.ResourceData, meta int
 	resGroup := id.ResourceGroup
 	appInsightsName := id.Path["components"]
 
-	if requireResourcesToBeImported && d.IsNewResource() {
+	if requireResourcesToBeImported {
 		var existing insights.ApplicationInsightsComponentAPIKey
 		existing, err = client.Get(ctx, resGroup, appInsightsName, name)
 		if err != nil {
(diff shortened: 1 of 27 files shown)


azurerm/resource_arm_application_insights_api_key.go
```diff
@@ -85,7 +85,7 @@ func resourceArmApplicationInsightsAPIKeyCreate(d *schema.ResourceData, meta int
 	resGroup := id.ResourceGroup
 	appInsightsName := id.Path["components"]
 
-	if requireResourcesToBeImported && d.IsNewResource() {
```
katbyte: All the other resources still have ` d.IsNewResource() `, thou as it is a create only function i can see why it is not needed.

azurerm/resource_arm_function_app.go
```diff
@@ -215,6 +216,7 @@ func resourceArmFunctionAppCreate(d *schema.ResourceData, meta interface{}) erro
 	log.Printf("[INFO] preparing arguments for AzureRM Function App creation.")
 
 	name := d.Get("name").(string)
+	resGroup := d.Get("resource_group_name").(string)
```
katbyte: _minor_ should we rename this to `resourceGroup`?

azurerm/resource_arm_application_insights_api_key.go
tombuildsstuff: it's only really required if it's a shared function between Create & Update - I didn't feel it added any value having it here tbh (since there's no way to make this code generic enough to go in a function without sacrificing readability imo?)

PR #2608, opened 2019-01-07
DNS: Parsing `dnsZones` rather than `dnszones` since the ID format changed

### LINKED ISSUE #2493: Documentation update, import dns resources
Hi Team,
I have recently needed to import some dns resources, but the documentation for this cases is not correct.
Changing "/dns**Z**one/"  with "/dns**z**one/" solve this issue.

_I have not reviewed other dns resources, but it seems that the same update is needed._

Thanks!!

Terraform version v0.11.9
Terraform azurerm v.1.19.0
SO Windows 10

### Affected Resource(s)
**Documentation located at**

- https://www.terraform.io/docs/providers/azurerm/r/dns_zone.html
- https://www.terraform.io/docs/providers/azurerm/r/dns_a_record.html
- possibly more...

tombuildsstuff:
Switching to use dnsZones rather than dnszones to match the ID format changed to now be:

> /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/dnsZones/{zoneName}/{recordType}/{relativeRecordSetName}

Fixes #2493

### DIFF
--- a/azurerm/resource_arm_dns_a_record.go
+++ b/azurerm/resource_arm_dns_a_record.go
@@ -121,7 +121,7 @@ func resourceArmDnsARecordRead(d *schema.ResourceData, meta interface{}) error {
 
 	resGroup := id.ResourceGroup
 	name := id.Path["A"]
-	zoneName := id.Path["dnszones"]
+	zoneName := id.Path["dnsZones"]
 
 	resp, err := dnsClient.Get(ctx, resGroup, zoneName, name, dns.A)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


azurerm/resource_arm_dns_a_record.go
```diff
@@ -121,7 +121,7 @@ func resourceArmDnsARecordRead(d *schema.ResourceData, meta interface{}) error {
 
 	resGroup := id.ResourceGroup
 	name := id.Path["A"]
-	zoneName := id.Path["dnszones"]
+	zoneName := id.Path["dnsZones"]
```
metacpp: Shall we define a constant string to hold this value ?

metacpp: Can we define a constant string to hold this value instead of updating every occurrence in the near future ?

tombuildsstuff: whilst that'd potentially save a few lines here and there, I'd rather we added support in the SDK for parsing Resource ID's to be honest: 
https://github.com/Azure/azure-sdk-for-go/issues/3080#issuecomment-432080173 (which would mean we could remove most of this logic altogether)

tombuildsstuff:
Upon further investigation this appears to be a bug in the Swagger where the API response - closing this in favour of https://github.com/Azure/azure-rest-api-specs/issues/5014

PR #2611, opened 2019-01-07
Fixes for handling the `dns_prefix` property of an AKS cluster

tomasaschan:
I got help from Microsoft Support today because one of my clusters were badly broken. Eventually, we figured out that it was because of a bad `dns_prefix`.

When I tried to use Terraform to change the prefix it didn't complain, until the underlying ARM call failed because the `dns_prefix` can only be set when the cluster is created, and then never changed.

This PR fixes both these issues: it forces recreation of the cluster if the `dns_prefix` changes, and it adds the missing validation for DNS prefixes. See the respective commit messages for more details on each change.

### DIFF
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -1127,7 +1127,7 @@ func validateKubernetesClusterAgentPoolName() schema.SchemaValidateFunc {
 
 func validateKubernetesClusterDnsPrefix() schema.SchemaValidateFunc {
 	return validation.StringMatch(
-		regexp.MustCompile("^[a-zA-Z][a-zA-Z0-9\\-]{0,43}[a-zA-Z0-9]$"),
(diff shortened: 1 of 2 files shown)


tomasaschan:
@tombuildsstuff Thanks!

In the meantime I got my local go env working again and noticed a build error, so I force-pushed a new commit that includes a previously missing `,` as well as your suggestion for the validation rules in the documentation.

tomasaschan:
Finally figured out how to use the provider locally as well, so now I have tested it on my setup and it seems to do what I intended it to :) I haven't touched the tests, and couldn't find anything in them about the `dns_prefix` property, so I doubt anything exciting will happen there.

Thanks for the speedy review - I'm already looking forward to the release of 1.21!

tombuildsstuff:
@tomasaschan thanks for pushing those changes - would you mind reverting that from the changelog? in the larger providers (AWS/Azure/GCP) we only change that directly on master to try and reduce conflicts

tomasaschan:
@katbyte Not at all, thanks for the help! :)

PR #2627, opened 2019-01-09
Azure Event Hub documentation: Kafka enabled property added on namespace

fernandoBRS:
Azure Event Hub documentation update based on #2395 and #1417


fernandoBRS:
cc @katbyte

PR #2633, opened 2019-01-09
Chunked file transfers for azurerm_data_lake_store_file

dyindude:
Files larger than 4MB fail to upload to ADLS with `azurerm_data_lake_store_file` due to lack of implementation of the API's `buffersize` parameter in the Azure SDK for go. There is an open issue for this on that project here: https://github.com/Azure/azure-sdk-for-go/issues/3231

The referenced issue contains a workaround which transfers the file using 4MB chunks (which appears to be a static value in the API: https://docs.microsoft.com/en-us/rest/api/datalakestore/webhdfs-filesystem-apis )

I have implemented the workaround in this commit, and tested the change in a terraform project I was encountering this error with. I tried my best to make it match the style of the code already present, so let me know if something else needs to be changed.

I am relatively new to golang, so I welcome any constructive feedback! Thanks


katbyte:
Ideally generate a large local file rather then keep something in the code itself

dyindude:
I took a stab at it, using the existing `TestAccAzureRMDataLakeStoreFile_basic` test as a template. Let me know if something needs to be changed.

katbyte:
Thanks @dyindude!

I hope you don't mind but I changed it to use a tmp file and to remove it after the test so I could merge it now 🙂 now LGTM

dyindude:
I don't mind at all, I have learned a lot from this process :) thanks

PR #2638, opened 2019-01-10
Sort data sources and resources map keys

inkel:
This PR sorts alphabetically the keys in the provider's `DataSourcesMap` and `ResourcesMap`.

Additionally it also formats `ResourcesMap` using `go fmt`, so perhaps it's best to review the changes without including whitespace only changes.

This one was first [identified in a different PR](https://github.com/terraform-providers/terraform-provider-azurerm/pull/2625#discussion_r246306617).


katbyte:
I hope you don't mind but I pushed some changes to fix the build 🙂

PR #2650, opened 2019-01-11
allowing the partner_id to be set to an empty string

tombuildsstuff:
Fixing a bug where the Partner ID being set to an empty string would cause a validation error


tombuildsstuff:
Travis build passes, but isn't reporting back:

screenshot 2019-01-11 at 17 34 16

Issue #2409, opened 2018-11-29
WAF upload file size limit


jay7x:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I realised today that Azure Application Gateway with WAF enabled is cutting our file uploads to 100Mb.
This seems to be documented (check link in references). But I found no way to change this in Azure terraform provider. It'd be great to have this.

### New or Affected Resource(s)

* azurerm_application_gateway
* waf_configuration

### Potential Terraform Configuration

```hcl
waf_configuration {
  ...
  file_upload_limit = 500
}
```

### References

* https://docs.microsoft.com/en-us/azure/application-gateway/application-gateway-waf-configuration

ghostinthewires:
Hi, Looked at updating this but when I run the tests it deletes the resources before it completes. Does anyone else get this?

ghostinthewires:
Should this be closed now?

PR #2666, opened 2019-01-14
azurerm_application_gateway: Adding upload limit

ghostinthewires:
This PR introduces the ability to set the file upload limit in the Azure Application Gateway WAF (requested in #2409)


website/docs/r/application_gateway.html.markdown
```diff
@@ -13,102 +13,102 @@ Manages an Application Gateway.
...
+locals {
+  backend_address_pool_name      = "${azurerm_virtual_network.test.name}-beap"
+  frontend_port_name             = "${azurerm_virtual_network.test.name}-feport"
+  frontend_ip_configuration_name = "${azurerm_virtual_network.test.name}-feip"
+  http_setting_name              = "${azurerm_virtual_network.test.name}-be-htst"
+  listener_name                  = "${azurerm_virtual_network.test.name}-httplstn"
+  request_routing_rule_name      = "${azurerm_virtual_network.test.name}-rqrt"
+}
+
+resource "azurerm_application_gateway" "network" {
```
metacpp: Looks like we missed the example usage for `waf_configuration` from the very beginning, can you add it together with `file_upload_limit_mb` ?

tombuildsstuff: we tend to keep the examples as minimal as possible - since there's a default value of 100MB here (which IMO should cover most use-cases) I'd suggest it'd be better to leave this out of the example?

ghostinthewires:
Hi,

I've implemented the changes as suggested. Hopefully should be good to go now!

Just let me know if there is anything else

Issue #2671, opened 2019-01-14
Service bus subscription name not able to start with number


RiverPhillips:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
Terraform: 0.11.11
Azurerm Provider 1.21.0

### Affected Resource(s)

* azurerm_servicebus_subscription
* azurerm_servicebus_subscription_rule

### Expected Behavior
Should be able to create a subscription/subscription rule starting with a number

### Actual Behavior

Output says invalid, subscription/subscription rule name must start with letter and number even though it starts with a number

### Steps to Reproduce

Create a subscription or subscription rule starting with a number

1. `terraform apply`

PR #2672, opened 2019-01-14
Allow service bus subscriptions and subscription rules to start with number

River991:
This PR allows subscription names and Rules to start with numbers resolving #2671

### DIFF
--- a/azurerm/helpers/azure/servicebus.go
+++ b/azurerm/helpers/azure/servicebus.go
@@ -28,7 +28,7 @@ func ValidateServiceBusQueueName() schema.SchemaValidateFunc {
 
 func ValidateServiceBusSubscriptionName() schema.SchemaValidateFunc {
 	return validation.StringMatch(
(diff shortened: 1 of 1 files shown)


azurerm/helpers/azure/servicebus.go
```diff
@@ -28,7 +28,7 @@ func ValidateServiceBusQueueName() schema.SchemaValidateFunc {
 
 func ValidateServiceBusSubscriptionName() schema.SchemaValidateFunc {
 	return validation.StringMatch(
-		regexp.MustCompile("^[a-zA-Z][-._a-zA-Z0-9]{0,48}([a-zA-Z0-9])?$"),
+		regexp.MustCompile("^[a-zA-0-9][-._a-zA-Z0-9]{0,48}([a-zA-Z0-9])?$"),
```
katbyte: I think this should be
```suggestion
		regexp.MustCompile("^[a-zA-Z0-9][-._a-zA-Z0-9]{0,48}([a-zA-Z0-9])?$"),
```

River991:
I just noticed the error, updated the PR

RiverPhillips:
That's great, thanks for the fast review

PR #2694, opened 2019-01-17
Normalize all documentation line endings to LF instead of CRLF.

JunyiYi:
I used a script to find all documentation files which have the wrong line endings (either completely Windows CRLF or a mixed one like `monitor_action_group.html.markdown`).

So I changed all line-endings back to the Linux one (LF).


katbyte:
Thanks for this @JunyiYi! LGTM 👍

PR #2705, opened 2019-01-17
govendor -> go modules

tombuildsstuff:
Attempt # 2 at switching over to Go Modules


GNUmakefile
```diff
@@ -4,6 +4,7 @@ PKG_NAME=azurerm
 
 #make sure we catch schema errors during testing
 TF_SCHEMA_PANIC_ON_ERROR=1
+GO111MODULE=on
```
appilon: With it configured as is, `go` is going to look for dependencies in the module cache. We want it to build from `vendor/` so add
```
GOFLAGS=-mod=vendor
```

tombuildsstuff: Pushed a commit to add this

Issue #2710, opened 2019-01-18
Support for Azure Media Services


seushermsft:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

To deploy Azure Media Services via Terraform, you currently have to use an ARM template, which oesn't allow Terraform to manage the actual resources that are deployed. Instead, there should be a resource of type `azurerm_media_services` that manages Azure Media Services

### New or Affected Resource(s)

* azurerm_media_services

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
resource "azurerm_resource_group" "testrg" {
  name     = "amstestrg"
  location = "westus"
}

resource "azurerm_storage_account" "testsa" {
  name                     = "amstestsa"
  resource_group_name      = "${azurerm_resource_group.testrg.name}"
  location                 = "${azurerm_resource_group.testrg.location}"
  account_tier             = "Standard"
  account_replication_type = "GRS"

  tags {
    environment = "staging"
  }
}

resource "azurerm_media_services" "ams" {

        name = "amstest"
        location = "${azurerm_resource_group.testrg.location}"
        resource_group_name = "${azurerm_resource_group.testrg.name}"
		
        storage_account {
				  id = "${azurerm_storage_account.testsa.id}"
				  is_primary = true
		    }
}
```

### References

* #0000

Lucretius:
@tombuildsstuff looks like we can close this one

tombuildsstuff:
@Lucretius thanks for the heads up, closing :)

PR #2711, opened 2019-01-18
New Resource: `azurerm_media_service_account`

seushermsft:
This change implements the creation and storage account configuration for Media Services, but doesn't expose other features such as the StreamingEndpoints.

Issue: #2710


azurerm/resource_arm_media_services.go
```diff
@@ -0,0 +1,194 @@
...
+
+			"storage_account": {
+				Type:     schema.TypeSet,
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
```
katbyte: Could we validate this with `azure.ValidateResourceID`?

azurerm/resource_arm_media_services.go
```diff
@@ -0,0 +1,194 @@
...
+
+						"is_primary": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
```
katbyte: Could we be explicit with our default here?

azurerm/resource_arm_media_services_test.go
```diff
@@ -0,0 +1,298 @@
...
+		if err != nil {
+			return err
+		}
+
+		resourceGroup := resourceId.ResourceGroup
+		name := resourceId.Path["mediaservices"]
+
+		log.Printf("Deleting Media Services Account '%s' in Resource Group '%s'", name, resourceGroup)
+		_, err = client.Delete(ctx, resourceGroup, name)
+		if err != nil {
```
katbyte: _Very minor_n but these two lines could be joined:
```go
		if _, err = client.Delete(ctx, resourceGroup, name); err != nil {
```

azurerm/resource_arm_media_services_test.go
```diff
@@ -0,0 +1,298 @@
...
+		
+			tags {
+			environment = "staging"
+			}
+	  }
+	  
+	  resource "azurerm_media_services" "ams" {
+	  
+			  name = "%[1]s"
+			  location = "${azurerm_resource_group.testrg.location}"
```
katbyte: Could we align these assignments?

azurerm/resource_arm_media_services_test.go
```diff
@@ -0,0 +1,298 @@
...
+	
+		tags {
+		environment = "staging"
+		}
+  }
+	  
+	  resource "azurerm_media_services" "ams" {
+	  
+			  name = "%[1]s"
+			  location = "${azurerm_resource_group.testrg.location}"
```
katbyte: And align these here too

examples/media-services/multiple_storage_accounts/main.tf
```diff
@@ -0,0 +1,53 @@
...
+  tags {
+    environment = "staging"
+  }
+}
+*/
+
+resource "azurerm_media_services" "ams" {
+
+        name = "amstest"
+        location = "${azurerm_resource_group.testrg.location}"
```
katbyte: Could we align these? you can run `terraform fmt .` to automatically format all files in the directly

website/docs/r/media_services.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+  resource_group_name      = "${azurerm_resource_group.testrg.name}"
+  location                 = "${azurerm_resource_group.testrg.location}"
+  account_tier             = "Standard"
+  account_replication_type = "GRS"
+}
+
+resource "azurerm_media_services" "ams" {
+
+        name = "amstest"
+        location = "${azurerm_resource_group.testrg.location}"
```
katbyte: Could we align these assignments?

website/docs/r/media_services.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) Specifies the name of the MariaDB Server. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the resource group in which to create the MariaDB Server. Changing this forces a new resource to be created.
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `storage_account` - (Required) A `storage_account` block as defined below. Multiple storage_account arguments can be provided, but only one can be marked as `is_primary`=true.
```
katbyte: Could we `quote` true, this also might be better moved into a note block. WDYT?

website/docs/r/media_services.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+---
+
+A `storage_account` block supports the following:
+
+* `id` - (Required) Specifies the Id for the storage account that will be associated with the Media Services instance.
+
+* `is_primary` - (Required) Specifies whether the storage account should be the primary account or not. A Media Services account supports associating multiple storage accounts, but only one can be marked as primary.
```
katbyte: Could we specify the default is `false` here?

azurerm/resource_arm_media_services_test.go
```diff
@@ -0,0 +1,298 @@
...
+)
+
+func init() {
+	resource.AddTestSweepers("azurerm_media_services", &resource.Sweeper{
+		Name: "azurerm_media_services",
+		F:    testSweepMediaServicesAccount,
+	})
+}
+
+func testSweepMediaServicesAccount(region string) error {
```
katbyte: We can also remove the sweeper functions, azure rm does not used them anymore.

azurerm/resource_arm_media_services.go
seushermsft: Added the call to azure.ValidateResourceID

azurerm/resource_arm_media_services_test.go
seushermsft: Joined the two lines

azurerm/resource_arm_media_services.go
seushermsft: Added a default value of false

azurerm/resource_arm_media_services_test.go
seushermsft: Fixed the alignment

azurerm/resource_arm_media_services_test.go
seushermsft: Fixed the alignment

examples/media-services/multiple_storage_accounts/main.tf
seushermsft: I ran this for each of the examples. I had no idea this existed, but its a nice feature.

website/docs/r/media_services.html.markdown
seushermsft: Fixed the alignment

website/docs/r/media_services.html.markdown
seushermsft: I `quoted` "true" and moved this into `## Note`

azurerm/resource_arm_media_services_test.go
seushermsft: Removed.

azurerm/resource_arm_media_services_test.go
```diff
@@ -0,0 +1,298 @@
...
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/services/mediaservices/mgmt/2018-07-01/media"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func init() {
```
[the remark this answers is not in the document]
seushermsft: Removed.

azurerm/resource_arm_media_services_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMMediaServicesAccountDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: generateTemplate_multiplestorage(resourceName, testLocation()),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					checkAccAzureRMMediaServicesAccount("azurerm_media_services.ams", testLocation(), "2"),
```
tombuildsstuff: can we remove this function and instead check the values persisted in the state? there's an example of this here https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_dev_test_windows_virtual_machine_test.go#L27

azurerm/config.go
```diff
@@ -228,6 +229,9 @@ type ArmClient struct {
 	managementGroupsClient             managementgroups.Client
 	managementGroupsSubscriptionClient managementgroups.SubscriptionsClient
 
+	// Media
+	mediaServicesClient media.MediaservicesClient
```
tombuildsstuff: since this is a new Resource Provider - could we add `microsoft.media` (case-sensitive) to the list of Resource Providers which Terraform will register: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/required_resource_providers.go#L44

examples/media-services/non_terraform_managed_storage/main.tf
```diff
@@ -0,0 +1,25 @@
+# Can you also get the id by adding a azurerm_storage_account that references the
+# same properties as the existing storage account.
+variable "existing_storage_account_id" {
+  default = "/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/testrg/providers/Microsoft.Storage/storageAccounts/teststorage"
```
tombuildsstuff: rather than hard-coding this ID (which we'd recommend against doing) - could we instead use [the Storage Account Data Source](https://www.terraform.io/docs/providers/azurerm/d/storage_account.html) instead?

azurerm/resource_arm_media_services_test.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+	}
+
+	return nil
+}
+
+func checkAccAzureRMMediaServicesAccount(resourceName string, location string, storageCount string) resource.TestCheckFunc {
+	// It would be ideal to also validate which storage account was Primary, but that isn't straight forward.
+	// The key for the storage account's Primary flag is non-deterministic (ex: storage_account.1137153885.is_primary) and
+	// isn't based on the storage account name.
```
tombuildsstuff: heh indeed, that said we should be able to remove this since the `import` stage verifies the config matches the local state

website/docs/r/media_services.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The `resource id` of the Media Services account.
+
+## Import
+
+MariaDB Server's can be imported using the `resource id`, e.g.
```
tombuildsstuff:
```suggestion
Media Service Account's can be imported using the `resource id`, e.g.
```

website/docs/r/media_services.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) Specifies the name of the MariaDB Server. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the resource group in which to create the MariaDB Server. Changing this forces a new resource to be created.
```
tombuildsstuff:
```suggestion
* `resource_group_name` - (Required) The name of the resource group in which to create the Media Service Account. Changing this forces a new resource to be created.
```

website/docs/r/media_services.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+		is_primary  = true
+	}
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) Specifies the name of the MariaDB Server. Changing this forces a new resource to be created.
```
tombuildsstuff:
```suggestion
* `name` - (Required) Specifies the name of the Media Service Account. Changing this forces a new resource to be created.
```

website/docs/r/media_services.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+## Import
+
+MariaDB Server's can be imported using the `resource id`, e.g.
+
+```shell
+terraform import azurerm_media_services.ams1 /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/mygroup1/providers/Microsoft.Media/mediaservices/ams1
+```
+## Note
+
+Multiple `storage_account` arguments can be provided to `azurerm_media_services`, but only one can be marked as `is_primary`=`true`.
```
tombuildsstuff: can we remove this in favour of a yellow box below the `is_primary` field by doing something like this:

```
~> **NOTE:** Whilst multiple `storage_account` blocks can be specified - one of them must be set to the primary
```

seushermsft: I will let you and katbyte fight this one out. I originally moved it from is_primary (though not using ~>) to the notes area due to a previous suggestion. I'm happy with it in either place.

katbyte: That is my bad, this is what i intended 😅 I should have said note box not block, sorry about that.

azurerm/resource_arm_media_services.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+
+		return fmt.Errorf("Error reading Media Services Account: %+v", err)
+	}
+
+	d.Set("name", resp.Name)
+	if location := resp.Location; location != nil {
+		d.Set("location", azureRMNormalizeLocation(*location))
+	}
+	flattenAndSetTags(d, resp.Tags)
```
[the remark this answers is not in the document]
seushermsft: I've added it in a similar way that resource_arm_metric_alertrule.go does.

azurerm/resource_arm_media_services_test.go
seushermsft: checkAccAzureRMMediaServicesAccount uses functions, such as resource.TestCheckResourceAttr internally to check the state. It only exists to remove the duplication since both acceptance tests have the same validation logic with slightly different values (ex: # of storage accounts).

azurerm/resource_arm_media_services_test.go
seushermsft: @tombuildsstuff - I disappeared for a week to work on some other stuff. Can you let me know if I'm misunderstanding the two remaining active comments?

tingulfsen:
@tombuildsstuff and @katbyte is there any chance you could take a look at this again regarding the latest comment from @seushermsft above? I would love for this PR to be merged, since I need the functionality provided.

azurerm/resource_arm_media_services_account.go
```diff
@@ -0,0 +1,223 @@
...
+	ctx := meta.(*ArmClient).StopContext
+
+	accountName := d.Get("name").(string)
+	location := azureRMNormalizeLocation(d.Get("location").(string))
+	tags := d.Get("tags").(map[string]interface{})
+	resourceGroup := d.Get("resource_group_name").(string)
+
+	storageAccounts := expandAzureRmStorageAccounts(d)
+	err := validateStorageConfiguration(storageAccounts)
+	if err != nil {
```
tombuildsstuff: we could probably combine these

vendor/github.com/Azure/azure-sdk-for-go/services/mediaservices/mgmt/2018-07-01/media/accountfilters.go
```diff
@@ -0,0 +1,487 @@
+package media
```
tombuildsstuff: can we vendor these files in? update: this requires rebasing on top of master - so I'm going to push a commit for this

azurerm/resource_arm_media_services_account.go
```diff
@@ -0,0 +1,217 @@
...
+					},
+				},
+			},
+
+			"tags": tagsSchema(),
+
+			"media_service_account_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
tombuildsstuff: in retrospect looks like this field isn't being deserialized correctly; will push a commit to remove this

tombuildsstuff: upstream issue: https://github.com/Azure/azure-sdk-for-go/issues/4124

tombuildsstuff:
Turns out there's a couple of bugs in the Azure API and SDK which mean we're going to have to ship support for this without Tags and the Media Service ID field:

* `media_service_id` isn't deserialized correctly: https://github.com/Azure/azure-sdk-for-go/issues/4124
* `tags` field isn't always returned: https://github.com/Azure/azure-rest-api-specs/issues/5249

For the moment I've removed support for these fields - but when these bugs are resolved we should be able to ship support for these.

tombuildsstuff:
Tests pass:

screenshot 2019-02-22 at 18 21 55

PR #2716, opened 2019-01-18
Export GO MODULES in GNUMakefile

metacpp:
1. Export the `GO111MODULE` and `GOFLAGS` in `GNUmakefile` before starting any sub-shell instance.
2. Add `modcheck` target to benefit building/testing process.
3. Disable `GO111MODULE` for `gometalinter` to avoid timeout issue.


metacpp:
@tombuildsstuff The `go mod` functionality is currently disabled in Travis CI. I am not sure if it's by intention or not. But the checksum issue is really noisy each time when you'd like to have a vendor copy for new package.

go.sum
```diff
@@ -68,7 +68,7 @@ github.com/hashicorp/go-azure-helpers v0.0.0-20181211121309-38db96513363 h1:9d0j
 github.com/hashicorp/go-azure-helpers v0.0.0-20181211121309-38db96513363/go.mod h1:Y5ejHZY3jQby82dOASJzyQ2xZw37zs+D5x6AaOC6O5E=
 github.com/hashicorp/go-cleanhttp v0.0.0-20170211013415-3573b8b52aa7 h1:67fHcS+inUoiIqWCKIqeDuq2AlPHNHPiTqp97LdQ+bc=
 github.com/hashicorp/go-cleanhttp v0.0.0-20170211013415-3573b8b52aa7/go.mod h1:JpRdi6/HCYpAwUzNwuwqhbovhLtngrth3wmdIIUrZ80=
-github.com/hashicorp/go-getter v0.0.0-20180226183729-64040d90d4ab h1:ap3umB+Fouf/i1/N1q0AZLjOxSg1s8RIQ159wc7wNPE=
+github.com/hashicorp/go-getter v0.0.0-20180226183729-64040d90d4ab h1:q9prbneWtMiyRZWcIZM5gFEIp7ROi+/iXd1fbLTY9T4=
```
tombuildsstuff: I've run `go mod vendor` and can't reproduce this, where are you seeing this issue?

GNUmakefile
```diff
@@ -4,8 +4,10 @@ PKG_NAME=azurerm
 
 #make sure we catch schema errors during testing
 TF_SCHEMA_PANIC_ON_ERROR=1
-GO111MODULE=on
-GOFLAGS=-mod=vendor
+
+# Turn on Go modules and use local vendor copy by default
+export GO111MODULE=on
+export GOFLAGS=-mod=vendor
```
tombuildsstuff: Go Modules should be enabled within Travis since we get a `go get` error if we don't run the tooling command with `GO111MODULE=off` - as such I don't believe this change is necessary? is this a Windows specific bug?

go.sum
metacpp: I always see this error in my MacBook Pro, even the module cache folder, $GOPATH/pkg/mod, is cleaned.

metacpp: This seems to be a known issue in Go before 1.11.4:
https://github.com/golang/go/issues/29281

katbyte: maybe that is why i am not seeing the issue, running go 1.11.4 here

GNUmakefile
tombuildsstuff: @metacpp your PR's running `go mod vendor` outside of the makefile [where this functionality is turned on](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/GNUmakefile#L6-L8) - as such this functionality won't be enabled outside of the Makefile.

go.sum
tombuildsstuff: @metacpp please upgrade to Go 1.11.4 to resolve this issue

tombuildsstuff:
hi @metacpp 

Thanks for this PR - it appears this can be resolved by upgrading to Go 1.11.4 (since this is a breaking change within Golang) - since we should be on the latest `1.11.x` release, would you mind upgrading? As such I'm going to close this PR for the moment since this isn't required.

Thanks!

GNUmakefile
metacpp: Yes, the above Travis result link is to `prove` `GO111MODULE` is disabled by default.

I did another change to test running of `go mod` in GNUMakefile. And it is also not working:
https://travis-ci.org/terraform-providers/terraform-provider-azurerm/builds/483537544?utm_source=github_status&utm_medium=notification (edited) 

Please take a look at this link: 
https://stackoverflow.com/questions/27282660/setting-environment-variables-in-a-makefile

Each line of recipe will be a sub-shell instance, to make things simpler here, we'd better just `export` them in GNUMakefile.

metacpp:
I've removed the change on the `go.sum` file, considering that it's hard reproduce in other environment. But I still keep the change on GNUMakefile, coz it's easily to reproduce on local/travis environment.

metacpp:
Looks like that running `go mod verify` in Travis has below issue:
```
o: golang.org/x/sync@v0.0.0-20181221193216-37e7f081c4d4: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /home/travis/gopath/pkg/mod/cache/vcs/55179c5d8c4db2eaed9fae4682d4c84a1fd3612df666b372bef3bbb997c9601f: exit status 128:
	fatal: remote error: 
	
	
	Invalid authentication credentials.
	
	Please generate a new identifier:
	  https://go.googlesource.com/new-password
```

tombuildsstuff:
Closing as discussed offline

PR #2717, opened 2019-01-19
Fix bugs of `sidebar_current` for documents.

JunyiYi:
This PR fixes the following issues in all data source and resource documentations:
* No item is highlighted
* Multiple unrelated items are highlighted
* Hierarchical items are not highlighted

### DIFF
--- a/website/docs/r/log_analytics_workspace.html.markdown
+++ b/website/docs/r/log_analytics_workspace.html.markdown
@@ -1,7 +1,7 @@
 ---
 layout: "azurerm"
 page_title: "Azure Resource Manager: azurerm_log_analytics_workspace"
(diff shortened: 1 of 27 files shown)


website/docs/r/log_analytics_workspace.html.markdown
```diff
@@ -1,7 +1,7 @@
 ---
 layout: "azurerm"
 page_title: "Azure Resource Manager: azurerm_log_analytics_workspace"
-sidebar_current: "docs-azurerm-resource-oms-log-analytics-workspace"
+sidebar_current: "docs-azurerm-oms-log-analytics-workspace-x"
```
metacpp: Why do you add `-x` here ? And it seems that every change in this PR has it.

PR #2720, opened 2019-01-19
Fix `sku` docs for `azurerm_databricks_workspace`

petems:
`sku` is lowercase only:

```
`Error: module.foo.azurerm_databricks_workspace.main: expected sku to be one of [standard premium], got Standard`
```


katbyte:
Thanks for the documentation update @petems, LGTM 💯

PR #2725, opened 2019-01-21
Redis vnet lock

hbuckle:
We've noticed some failures when destroying vnet attached redis caches which I think are due to other resources using the same vnet, so I've added vnet and subnet locking


tombuildsstuff:
Ignoring a failing test (due to the Redis API being unavailable) - the tests pass:

screenshot 2019-01-23 at 18 44 16

Issue #2719, opened 2019-01-19
Add support for the Azure IotHub fallback route


IvanovOleg:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When you create an IoT Hub with a custom route via terraform, the [IoTHub fallback route](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-d2c#fallback-route) is disabled by default which is not suitable. If you enable this parameter in portal and run terraform again, it disables it.

### New or Affected Resource(s)

* azurerm_iothub

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

* #0000

danielscholl:
The default action of creating a hub via cli or portal is to have the IoT fallback route enabled.  The default action of the terraform creating a hub is to have the IoT fallback route disabled with no option to enable it.

## Terraform Configuration

```
provider "azurerm" {
  version = "=1.21.0"
}

resource "azurerm_resource_group" "test" {
  name     = "IoTEdgeResources"
  location = "eastus2"
}

resource "azurerm_iothub" "test" {
  name                = "tf-hub"
  resource_group_name = "${azurerm_resource_group.test.name}"
  location            = "${azurerm_resource_group.test.location}"

  sku {
    name     = "S1"
    tier     = "Standard"
    capacity = "1"
  }
}
```

## Azure CLI Commands

```bash
# Create a Hub
az iot hub create --resource-group IoTEdgeResources --name test-cli-hub --sku S1 --location eastus2

# View fallbackRoute status
az iot hub list -g IoTEdgeResources --query \
  "[].{name:name, fallback:properties.routing.fallbackRoute.isEnabled}" \
  -otable
```

## Creation Result

```
Name             Fallback
-----------      ----------
test-cli-hub     True
test-tf-hub      False
```

BlueBasher:
Terraform is indeed disabling the fallback route by default.
[...]

PR #2764, opened 2019-01-25
Add option in IoTHub to Enable Fallback Route

BlueBasher:
Added the option to specify if the IoTHub fallback route should be enabled or not.
It defaults to enabled since the Azure Portal, CLI or ARM all default to true. This is a change in the behaviour of Terraform since Terraform used to default to disabled.
See also issue #2719 

PS. apologies for the multiple commits, first time Go developer and couldn't get the build to run locally.


azurerm/resource_arm_iothub.go
```diff
@@ -248,6 +248,12 @@ func resourceArmIotHub() *schema.Resource {
 				},
 			},
 
+			"enable_fallback_route": {
```
katbyte: Given that the API takes more values should mirror it and possible include them? 
```
"fallback_route": {
	Type:     schema.TypeList,
	MaxItems: 1,
	Optional: true,
	Elem: &schema.Resource{
			"enabled": {
				Type:     schema.TypeBool,
				Required: true,
			},
...
```
and the rest

azurerm/resource_arm_iothub.go
```diff
@@ -308,6 +315,12 @@ func resourceArmIotHubCreateUpdate(d *schema.ResourceData, meta interface{}) err
 			Routing: &devices.RoutingProperties{
 				Endpoints: endpoints,
 				Routes:    routes,
+				FallbackRoute: &devices.FallbackRouteProperties{
+					Name:          utils.String("$fallback"),
```
katbyte: Shouldn't we allow the user to set these and configure the default values to match the portal?

azurerm/resource_arm_iothub.go
```diff
@@ -248,6 +248,12 @@ func resourceArmIotHub() *schema.Resource {
 				},
 			},
 
+			"enable_fallback_route": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
```
katbyte: For now we should keep the default terraform uses as to not break anything, and then when 2.0 comes out we can change the default.
```suggestion
				Default:  false, // todo change to true in 2.0
```

azurerm/resource_arm_iothub.go
BlueBasher: That's indeed better. I modified the code for this.

azurerm/resource_arm_iothub.go
BlueBasher: I understand, modified the default to false.

azurerm/resource_arm_iothub.go
BlueBasher: Modified it so the defaults match the portal and a user can set their own values.
For the fallback route there can only be 1 endpoint so added that as a string with the default value, instead of a list.

BlueBasher:
Hi @katbyte,
FYI: I've finished the changes you requested in the review.

azurerm/resource_arm_iothub.go
```diff
@@ -740,6 +810,28 @@ func flattenIoTHubRoute(input *devices.RoutingProperties) []interface{} {
 	return results
 }
 
+func flattenIoTHubFallbackRoute(input *devices.RoutingProperties) []interface{} {
+	output := make(map[string]interface{})
+
+	if input.FallbackRoute != nil {
```
katbyte: _minor_ This would be a bit cleaner if we:
```go
if route := input.FallbackRoute; route != nil {
```

azurerm/resource_arm_iothub.go
```diff
@@ -740,6 +810,28 @@ func flattenIoTHubRoute(input *devices.RoutingProperties) []interface{} {
...
+		if condition := input.FallbackRoute.Condition; condition != nil {
+			output["condition"] = *condition
+		}
+		if endpointNames := input.FallbackRoute.EndpointNames; endpointNames != nil {
+			output["endpoint_name"] = (*endpointNames)[0]
+		}
+		if isEnabled := input.FallbackRoute.IsEnabled; isEnabled != nil {
+			output["enabled"] = *isEnabled
+		}
+		output["source"] = input.FallbackRoute.Source
```
katbyte: I believe this is a pointer? so we should be checking for nil:
```go
		if v := route.Source; v != nil {
			output["enabled"] = *v
		}
```

website/docs/r/iothub.html.markdown
```diff
@@ -133,6 +139,20 @@ A `route` block supports the following:
 
 * `enabled` - (Required) Used to specify whether a route is enabled.
 
+---
+
+A `fallback_route` block supports the following:
+
+* `name` - (Optional) The name of the route. The name can only include alphanumeric characters, periods, underscores, hyphens, has a maximum length of 64 characters, and must be unique.
```
katbyte: We typically don't list the validation requirements here and let the validation function do inform.

website/docs/r/iothub.html.markdown
```diff
@@ -34,9 +34,9 @@ resource "azurerm_storage_container" "test" {
 }
 
 resource "azurerm_iothub" "test" {
-  name                = "test"
-  resource_group_name = "${azurerm_resource_group.test.name}"
-  location            = "${azurerm_resource_group.test.location}"
+  name                  = "test"
```
katbyte: These changes seem spurious?

azurerm/resource_arm_iothub_test.go
```diff
@@ -298,3 +322,35 @@ resource "azurerm_iothub" "test" {
...
+	return fmt.Sprintf(`
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-%d"
+  location = "%s"
+}
+
+resource "azurerm_iothub" "test" {
+  name                  = "acctestIoTHub-%d"
+  resource_group_name   = "${azurerm_resource_group.test.name}"
+	location              = "${azurerm_resource_group.test.location}"
```
katbyte: Could we fix the formatting here?

azurerm/resource_arm_iothub_test.go
```diff
@@ -113,6 +113,30 @@ func TestAccAzureRMIotHub_customRoutes(t *testing.T) {
...
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMIotHubDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMIotHub_fallbackRoute(rInt, testLocation()),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMIotHubExists(resourceName),
+				),
```
katbyte: Could we check some of the values returned in state?? ie:
```go
resource.TestCheckResourceAttr("azurerm_iothub.test", "fallback_route.0.name", "fallback"),
...
```

and such for the rest of them?

azurerm/resource_arm_iothub.go
```diff
@@ -248,6 +248,51 @@ func resourceArmIotHub() *schema.Resource {
...
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							Default:      "$fallback",
+							ValidateFunc: validation.StringLenBetween(0, 64),
```
katbyte: Could we use a do a bit stronger of validation here:
```
				ValidateFunc: validation.StringMatch(
					regexp.MustCompile("^[-_.a-zA-Z0-9]{1,64$"),
					"Fallback Route Name name can only include alphanumeric characters, periods, underscores, hyphens, has a maximum length of 64 characters, and must be unique.",
				),
```

azurerm/resource_arm_iothub_test.go
BlueBasher: Tab vs spaces :( Fixed it.

azurerm/resource_arm_iothub_test.go
BlueBasher: Added additional checks for all the fallback fields.

azurerm/resource_arm_iothub.go
BlueBasher: Updated the validation for both FallbackRoute and Route.

BlueBasher:
Hi @katbyte,
I've updated the code based on your comments.

BlueBasher:
Hi @katbyte,
the expandIoTHubFallbackRoute method didn't check if a fallback route was actually present. That resulted in the "index out of range".
I've changed the code so now it does check and if no fallback route present, then it uses nil instead.

maxbog:
@katbyte, @BlueBasher Hi, I am very interested in this PR, will this be included in 1.23.0? Do you need any help with moving it forward?

katbyte:
Hi @maxbog, @BlueBasher,

I am hoping to have it included in 1.23, right now the only thing holding it up is the tests don't pass:

multiple:
```
------- Stdout: -------
=== RUN   TestAccAzureRMIotHubConsumerGroup_operationsMonitoringEvents
=== PAUSE TestAccAzureRMIotHubConsumerGroup_operationsMonitoringEvents
=== CONT  TestAccAzureRMIotHubConsumerGroup_operationsMonitoringEvents
--- FAIL: TestAccAzureRMIotHubConsumerGroup_operationsMonitoringEvents (334.78s)
    testing.go:538: Step 0 error: After applying this step, the plan was not empty:
        
        DIFF:
        
        UPDATE: azurerm_iothub.test
          fallback_route.#: "1" => "0"
```

katbyte:
As well as:

```
--- FAIL: TestAccAzureRMIotHub_fallbackRoute (62.55s)
        	* azurerm_iothub.test: 1 error occurred:
        	* azurerm_iothub.test: Error creating/updating IotHub "acctestIoTHub-190218234111779608" (Resource Group "acctestRG-190218234111779608"): devices.IotHubResourceClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="Failed" Message="The async operation failed." InnerError={"unmarshalError":"json: cannot unmarshal number into Go struct field serviceError2.code of type string"} AdditionalInfo=[{"code":400108,"httpStatusCode":"BadRequest","message":"EndpointName:fallback is not from a valid type. If you contact a support representative please include this correlation identifier: ede0b60b-e576-456b-8896-c12b7b75a936, timestamp: 2019-02-19 07:41:21Z, errorcode: IH400108."}]
```

Looks like the API may have changed 🤔

katbyte:
@BlueBasher,

I hope you don't mind but i pushed an update to fix the non-empty plan failure. Opted to ignore the default fallback route, another option would be to mark those properties as computed but i'm not sure it makes sense. WDYT?

BlueBasher:
@katbyte 
Definitely don't mind, most important thing is we get the functionality merged :)
Your change looks fine to me, think ignoring the default is the best option.

katbyte:
@BlueBasher,

I spent some more time trying to figure out why it was 400'ing: updated the API and removed name (it doesn't seem to actually do anything if you set it?) and did a create then update that finally got rid of the 400.

However now it doesn't seem to be actually setting enabled to true so i am going to push what i have for now and pick this up later in the week.

azurerm/resource_arm_iothub.go
```diff
@@ -573,6 +633,28 @@ func expandIoTHubEndpoints(d *schema.ResourceData, subscriptionId string) (*devi
...
+	fallbackRouteList := d.Get("fallback_route").([]interface{})
+	if len(fallbackRouteList) == 0 {
+		return nil
+	}
+
+	fallbackRouteMap := fallbackRouteList[0].(map[string]interface{})
+
+	source := fallbackRouteMap["source"].(string)
+	condition := fallbackRouteMap["condition"].(string)
+	endpointNames := make([]string, 0)
```
maxbog: endpointNames is not used anymore

azurerm/resource_arm_iothub.go
```diff
@@ -313,6 +362,11 @@ func resourceArmIotHubCreateUpdate(d *schema.ResourceData, meta interface{}) err
 		Tags: expandTags(tags),
 	}
 
+	if !d.IsNewResource() {
```
maxbog: I don't really see the point of this diff.
When I removed it and set the `Properties.Routing.FallbackRoute` property to `fallbackRoute` in the line above, fallback routes started working

maxbog:
@katbyte, @BlueBasher I made some minor modifications to make the code work, please look at my comments

azurerm/resource_arm_iothub_test.go
```diff
@@ -298,3 +325,34 @@ resource "azurerm_iothub" "test" {
...
+
+  sku {
+    name     = "S1"
+    tier     = "Standard"
+    capacity = "1"
+  }
+
+  fallback_route {
+    source         = "DeviceMessages"
+    endpoint_names = ["fallback"]
```
maxbog: For the test to pass, this is needed
```suggestion
    endpoint_names = ["events"]
```

azurerm/resource_arm_iothub.go
katbyte: indeed! thanks

azurerm/resource_arm_iothub_test.go
katbyte: Well that was it! thanks

katbyte:
Thank you very much for pointing us in the right direction @maxbog! tests pass and this LGTM now. Will merge it shortly @BlueBasher

BlueBasher:
@katbyte @maxbog Thanks for all the help to get this merged!

PR #2774, opened 2019-01-26
Reverted ParallelTest for tests that shouldn't be Parallel

katbyte:


metacpp:
@katbyte Thanks for the PR, LGTM.

Issue #2757, opened 2019-01-24
New Data Source for built in policies


lawrenae:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Azure comes with a variety of built-in policies both at the subscription and management group levels (at least according to the APIs: https://docs.microsoft.com/en-us/rest/api/resources/policydefinitions/listbuiltin and 
https://docs.microsoft.com/en-us/rest/api/resources/policydefinitions/listbymanagementgroup)

I'd like to see something similar to https://www.terraform.io/docs/providers/azurerm/d/builtin_role_definition.html but for policies

### New or Affected Resource(s)

Something along the lines of a datasource called `azurerm_builtin_policy_definition`

### Potential Terraform Configuration

```hcl
# default to subscription
data "azurerm_builtin_policy_definition" "allowed_resource_types" {
  name = "Allowed resource types"
}

output "policy_id" {
   value = "${data.azurerm_builtin_policy_definition.allowed_resource_types.id}"
}

# for management groups
data "azurerm_builtin_policy_definition" "allowed_resource_types" {
  name = "Allowed resource types"
  management_group_id = "${data.azurerm_management_group.tenant_root_group.group_id}"
}

output "policy_id" {
   value = "${data.azurerm_builtin_policy_definition.allowed_resource_types.id}"
}
```

PR #2788, opened 2019-01-28
add built-in policies for both subscriptions and management-groups

lawrenae:
this is meant as a fix for #2757 

Feedback most welcome, of course


katbyte:
@lawrenae, should we expose more of the properties as the builtin role definition datasource does?

lawrenae:
@katbyte sure. I don't know how useful those extra properties would be to folks, but it should be easy to add them. Will see what I can do!

lawrenae:
@katbyte I added 
* `display_name` - (Required) Specifies the name of the built-in Policy Definition.
* `id` - the ID of the built-in Policy Definition.
* `description` - the Description of the built-in Policy.
* `type` - the Type of Policy.
* `name` - The Name of the Policy Definition

I see a few things in the portal that I couldnt find in the API -- `effect` and `category` that might also be useful.

azurerm/data_source_builtin_policy_definition.go
```diff
@@ -0,0 +1,83 @@
...
+				Optional: true,
+			},
+			"type": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
```
katbyte: Other properties i think we should export after looking at the resource are:
- `policy_type`
- `policy_rule`
- `metadata`
- `parameters`

azurerm/data_source_builtin_policy_definition.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+		}
+	}
+
+	if policyDefinition.ID == nil {
+		return fmt.Errorf("Error loading Policy Definition List: could not find policy '%s'", name)
+	}
+
+	d.SetId(*policyDefinition.ID)
+	d.Set("name", *policyDefinition.Name)
```
katbyte: We don't have to dereference these here, `d.Set` will do it for us.

```suggestion
	d.Set("name", policyDefinition.Name)
```

website/docs/d/builtin_policy_definition.markdown
```diff
@@ -0,0 +1,35 @@
...
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_builtin_policy_definition"
+sidebar_current: "docs-azurerm-datasource-builtin-policy-definition"
+description: |-
+  Get information about an existing built-in Policy Definition.
+---
+
+# Data Source: azurerm_builtin_policy_definition
+
+Use this data source to access information about a built-in Policy Definition. To access information about a custom Policy Definition, [please see the `azurerm_policy_definition` data source](../r/policy_definition.html) instead.
```
katbyte: The policy data source doesn't exist.. yet

azurerm/data_source_builtin_policy_definition.go
```diff
@@ -0,0 +1,83 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"type": {
+				Type:     schema.TypeString,
+				Optional: true,
```
katbyte: This should be computed as we are only returning the value here

lawrenae:
@katbyte I like the idea of combining them. I will get to working on that, hopefully very soon

azurerm/data_source_policy_definition.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+func dataSourceArmPolicyDefinition() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmPolicyDefinitionRead,
+		Schema: map[string]*schema.Schema{
+			"display_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
```
katbyte: Could we validate this with `validate.NoZerovalues`?

azurerm/data_source_policy_definition.go
```diff
@@ -0,0 +1,80 @@
...
+			"management_group_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
```
katbyte: The could be validate with `azure.VAlidateresourceIDOrEmpty`

azurerm/data_source_policy_definition.go
```diff
@@ -0,0 +1,80 @@
...
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
```
katbyte: As this is a data source this should be computed not optional
```suggestion
				Computed: true,
```

azurerm/data_source_policy_definition.go
```diff
@@ -0,0 +1,80 @@
...
+			},
+			"type": {
+				Type:     schema.TypeString,
+				Optional: true,
```
katbyte: As this is a data source this should be computed not optional
```suggestion
				Computed: true,
```

azurerm/data_source_policy_definition.go
```diff
@@ -0,0 +1,80 @@
...
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
```
katbyte: As this is a data source this should be computed not optional
```suggestion
				Computed: true,
```

azurerm/data_source_policy_definition.go
```diff
@@ -0,0 +1,80 @@
...
+
+	if err != nil {
+		return fmt.Errorf("Error loading Policy Definition List: %+v", err)
+	}
+
+	var policyDefinition policy.Definition
+
+	for policyDefinitions.NotDone() {
+		def := policyDefinitions.Value()
+		if *def.DisplayName == name {
```
katbyte: You may want to check if `def.DisplayName` is ` nil` before dereferencing it

azurerm/data_source_policy_definition_test.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func testAccDataSourceBuiltInPolicyDefinitionAtManagementGroup(name string) string {
+	return fmt.Sprintf(`
+
+data "azurerm_client_config" "current" {}
+
+data "azurerm_policy_definition" "test" {
+  display_name = "%s"
+  management_group_id = "${data.azurerm_client_config.current.tenant_id}"
```
katbyte: Could we align these assignments?

azurerm/data_source_policy_definition.go
lawrenae: I went with `validate.NoEmptyStrings` instead -- cool?

lawrenae:
@katbyte I appreciate the additional feedback, and think I've got this ready for review (again).

lawrenae:
@katbyte -- good catches on the docs. updated!

katbyte:
Thanks @lawrenae, LGTM now 🙂

PR #2789, opened 2019-01-28
New Resource: `azurerm_network_interface_application_security_group_association`

tombuildsstuff:
This PR introduces a new resource which associations Network Interfaces to Application Security Groups. At this time this association needs to be configured both on the `azurerm_network_interface` resource and using the `azurerm_network_interface_application_security_group_association ` resource.

Come 2.0 the existing `application_security_group_ids` field in the `azurerm_network_interface ` resource will be removed (as will the other deprecated fields) and the `azurerm_network_interface_application_security_group_association ` resource will be the only way to configure these. Whilst this isn't ideal from a UX perspective in the short-term it mirrors the other fields, which ensures the UX is the same for all the fields.


tombuildsstuff:
Ignoring a couple of intermittent test failures the tests pass:

screenshot 2019-01-29 at 08 48 16

PR #2806, opened 2019-01-30
Limits conforming to Azure docs: https://docs.microsoft.com/en-us/azu…

georgespatton:
This should fix https://github.com/terraform-providers/terraform-provider-azurerm/issues/2136 as well as https://github.com/terraform-providers/terraform-provider-azurerm/issues/2624 to follow the Azure documentation here:  https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/overview


georgespatton:
@katbyte I added a test case, let me know what you think.

tombuildsstuff:
hey @georgespatton 

Thanks for pushing those changes :)

As a part of the v1.22 release we're renaming the `azurerm_autoscale_setting` resource to `azurerm_monitor_autoscale_setting` - as such this change will need to be made in both places. So that we can get this merged and released as a part of the v1.22 release - I'm going to push a commit which also makes this change to that resource (and updates the documentation), I hope you don't mind!

Thanks!

georgespatton:
Sounds great, thank you!

tombuildsstuff:
@georgespatton apologies, I don't entirely understand how but I appear to have broken this PR upon pushing the updates to your fork. I hope you don't mind but I'm going to open another PR which includes these commits - you'll still get the credits/attribution for this (since these include your commits) - and this should go out as a part of v1.22 - apologies either way!

PR #2816, opened 2019-01-31
Documentation: adding the initial 2.0 guide

tombuildsstuff:
This PR adds the initial 2.0 upgrade guide which will go live as a part of v1.22

Rather than making a specific section in the sidebar for this I've combined the "Authentication" guides and this into a single "Guides" block

### DIFF
--- a/website/docs/guides/2.0-upgrade-guide.html.markdown
+++ b/website/docs/guides/2.0-upgrade-guide.html.markdown
@@ -3,7 +3,7 @@ layout: "azurerm"
 page_title: "Azure Resource Manager: 2.0 Upgrade Guide"
 sidebar_current: "docs-azurerm-guide-2.0-upgrade"
 description: | -
-    This page documents how to migrate between two resources in the Azure Provider which have been renamed.
+    Azure Resource Manager: 2.0 Upgrade Guide
 
 ---
 
@@ -16,6 +16,36 @@ Since then we've added support for 191 Resources, 58 Data Sources and have launc
 Version 2.0 of the AzureRM Provider is a major release and as such includes some larger-scale changes which are outlined in this document.
 
 -> **NOTE:** This guide is a Work In Progress and as such fields may be added to this guide until version 2.0 of the AzureRM Provider is released.
+
+### Pinning your Provider Version
+
+We recommend pinning the version of each Provider you use in Terraform - you can do this using the `version` attribute in the `provider` block, either to a specific version of the AzureRM Provider, like so:
+
+```hcl
+provider "azurerm" {
+  version = "=1.22.0"
+}
+```
+
+.. or to any 1.x release:
+
+```hcl
+provider "azurerm" {
+  version = "~> 1.x"
+}
+```
+
+More information on [how to pin the version of a Terraform Provider being used can be found on the Terraform Website](https://www.terraform.io/docs/configuration/providers.html#provider-versions).
+
+Once version 2.0 of the AzureRM Provider is released - you can then upgrade to it by updating the version specified in the Provider block, like so:
+
+```hcl
+provider "azurerm" {
+  version = "=2.0.0"
+}
+```
+
+## What's coming in Version 2.0 of the AzureRM Provider?
 
 At a high level, we're intending for version 2.0 to include the following changes:
 
@@ -42,6 +72,8 @@ In order to match the behaviour of other Terraform Providers version 2.0 of the 
 A resource with the ID /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/group1 already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for `azurerm_resource_group` for more information.
 ```
 
+Information on how to import a given Resource can be found in the documentation for that Resource - for example here's how to [here's how to import a Resource Group in Terraform](/docs/providers/azurerm/r/resource_group.html#import).
+
 You can opt into this behaviour in version 1.22 of the AzureRM Provider by setting the Environment Variable `ARM_PROVIDER_STRICT` to `true`.
 
 ## Custom Timeouts for Resources
@@ -64,7 +96,7 @@ resource "azurerm_resource_group" "test" {
 }
 ```
 
-We intend to support the `timeout` block in version 2.0 of the Azure Provider - which will allow timeouts to be specified on resources (as shown above). [This feature request is being tracked here](https://github.com/terraform-providers/terraform-provider-azurerm/issues/171) and will form part of the 2.0 release of the AzureRM Provider.
+We intend to support the `timeouts` block in version 2.0 of the Azure Provider - which will allow timeouts to be specified on resources (as shown above). [This feature request is being tracked here](https://github.com/terraform-providers/terraform-provider-azurerm/issues/171) and will form part of the 2.0 release of the AzureRM Provider.
 
 ## New Resources for Virtual Machines and Virtual Machine Scale Sets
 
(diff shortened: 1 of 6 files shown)


website/docs/guides/2.0-upgrade-guide.html.markdown
```diff
@@ -0,0 +1,348 @@
+---
+layout: "azurerm"
+page_title: "Azure Active Directory: 2.0 Upgrade Guide"
```
bflad: Copypasta 🍝

website/docs/guides/2.0-upgrade-guide.html.markdown
```diff
@@ -0,0 +1,348 @@
...
+sidebar_current: "docs-azurerm-guide-2.0-upgrade"
+description: | -
+    This page documents how to migrate between two resources in the Azure Provider which have been renamed.
```
bflad: Pretty sure this document talks about more than just this. Let's leave the description generic.

```suggestion
    Azure Resource Manager: 2.0 Upgrade Guide
```

website/docs/guides/2.0-upgrade-guide.html.markdown
```diff
@@ -0,0 +1,348 @@
...
+
+# v2.0 of the AzureRM Provider
+
+Terraform initially shipped support for the AzureRM Provider back in December 2015.
+
+Since then we've added support for 191 Resources, 58 Data Sources and have launched a couple of related Providers in the form of [the Azure Active Directory Provider](https://www.terraform.io/docs/providers/azuread/index.html) and [the Azure Stack Provider](https://www.terraform.io/docs/providers/azurestack/index.html).
+
+Version 2.0 of the AzureRM Provider is a major release and as such includes some larger-scale changes which are outlined in this document.
+
+-> **NOTE:** This guide is a Work In Progress and as such fields may be added to this guide until version 2.0 of the AzureRM Provider is released.
```
bflad: "fields" could be confusing, maybe "additional information"?

```suggestion
-> **NOTE:** This guide is a Work In Progress and additional information may be added to this guide until version 2.0 of the AzureRM Provider is released.
```

website/docs/guides/2.0-upgrade-guide.html.markdown
```diff
@@ -0,0 +1,348 @@
...
+
+At a high level, we're intending for version 2.0 to include the following changes:
+
+* Existing Resources will be required to be imported
```
bflad: We might want to clarify this a bit as I read this as "I have to do something to every one of my existing Terraform AzureRM resources?":

```suggestion
* When adding pre-existing AzureRM infrastructure as new resources in Terraform, importing those Terraform resources will be required
```

website/docs/guides/2.0-upgrade-guide.html.markdown
```diff
@@ -0,0 +1,348 @@
...
+* Custom Timeouts will be available on Resources - this will allow you to specify a custom timeout for provisioning the resource in your Terraform Configuration [using the `timeouts` block](https://www.terraform.io/docs/configuration/resources.html#timeouts).
+* New resources for Virtual Machines and Virtual Machine Scale Sets
+* Removing Fields, Data Sources and Resources which have been deprecated
+
+Each of these topics is covered in more detail below, however please note that this guide is a Work In Progress until version 2.0 of the AzureRM Provider is released and thus things may be added/changed as necessary.
+
```
bflad: It may be worth adding a Markdown table of contents as this guide is long.

website/docs/guides/2.0-upgrade-guide.html.markdown
```diff
@@ -0,0 +1,348 @@
...
+Splitting the Virtual Machine Scale Set resource in two allows us to both provide finer-grain validation for this resource and update the schema.
+
+The existing `azurerm_virtual_machine_scale_set` Resource will continue to be available in it's current form, however it will eventually be deprecated and we recommend using the new resources going forward.
+
+---
+
+We've spent the past few months laying the groundwork for these changes - and whilst we appreciate that your Terraform Configurations may require code changes to upgrade to 2.0 - we take Semantic Versioning seriously and so try our best to limit these changes to major versions.
+
+### Pinning your Provider Version
```
bflad: You may want to consider putting this information at the top so people can immediately see:
* How to consistently upgrade
* How to consistently prevent the upgrade

website/docs/guides/2.0-upgrade-guide.html.markdown
```diff
@@ -0,0 +1,348 @@
...
+Whilst this may allow resources to work in some cases, it leads to hard-to-diagnose bugs in others (which could have been caught during `terraform plan`).
+
+In order to match the behaviour of other Terraform Providers version 2.0 of the AzureRM Provider will require that existing resources are imported into the state prior to use. This means that Terraform will be checking for the presence of an existing resource prior to creating it - and will return an error similar to below:
+
+```
+A resource with the ID /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/group1 already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for `azurerm_resource_group` for more information.
+```
+
+You can opt into this behaviour in version 1.22 of the AzureRM Provider by setting the Environment Variable `ARM_PROVIDER_STRICT` to `true`.
+
```
bflad: You may want to include information about where to find documentation for how to import each resource, e.g. at the bottom of every resource documentation page.

website/docs/guides/2.0-upgrade-guide.html.markdown
```diff
@@ -0,0 +1,348 @@
...
+  location = "West Europe"
+
+  timeouts {
+    create = "10m"
+    delete = "30m"
+  }
+}
+```
+
+We intend to support the `timeout` block in version 2.0 of the Azure Provider - which will allow timeouts to be specified on resources (as shown above). [This feature request is being tracked here](https://github.com/terraform-providers/terraform-provider-azurerm/issues/171) and will form part of the 2.0 release of the AzureRM Provider.
```
bflad: Did you mean `timeout` or `timeouts` here? `timeouts` is the Terraform core implementation.

website/docs/guides/2.0-upgrade-guide.html.markdown
```diff
@@ -0,0 +1,348 @@
...
+* New resources for Virtual Machines and Virtual Machine Scale Sets
+* Removing Fields, Data Sources and Resources which have been deprecated
+
+Each of these topics is covered in more detail below, however please note that this guide is a Work In Progress until version 2.0 of the AzureRM Provider is released and thus things may be added/changed as necessary.
+
+## Requiring Resources to be Imported
+
+Terraform allows for existing resources which have been created outside of Terraform to be Imported into Terraform's State. Once a resource is imported into the state, it's possible for Terraform to track changes and manage this resource. The Azure Provider allows Importing existing resources into the state (using `terraform import`) for (almost) every resource.
+
+Version 2.0 of the Azure Provider aims to solve an issue where it's possible to unintentionally import resources into the state by running `terraform apply`. To explain this further, the majority of Azure's API's are Upserts - which means that a resource will be updated if it exists, otherwise it'll be created.
```
katbyte: I think it would be good to mention (and stress) that along with an unintentional import changes could be made to an existing resource by accident.

website/docs/guides/2.0-upgrade-guide.html.markdown
```diff
@@ -0,0 +1,348 @@
...
+
+Where the unique identifier for (most) Azure resources is the `name` (rather than for example an `aws_instance` where AWS will generate a different unique identifier) - it's possible that users may have unintentionally imported resources into Terraform when running `terraform apply` on an existing resource.
```
katbyte: Should we change this to `azurerm_policy?
```suggestion
Where the unique identifier for (most) Azure resources is the `name` (rather than for example an `azurerm_policy_definition` where the provider will generate a different unique identifier) - it's possible that users may have unintentionally imported resources into Terraform when running `terraform apply` on an existing resource.
```

website/docs/guides/2.0-upgrade-guide.html.markdown
tombuildsstuff: What about:

> Changes when importing existing resources - when bringing Resources provisioned outside of Terraform under Terraform's control, these resources will now be required to be imported.

website/docs/guides/2.0-upgrade-guide.html.markdown
tombuildsstuff: I intentionally picked AWS here since almost(?) all resources have a unique identifier (whereas they're the exception in Azure, rather than the rule) - so I think this is worth leaving?

website/docs/guides/2.0-upgrade-guide.html.markdown
katbyte: Sounds better to me

PR #2820, opened 2019-02-01
Switch Key Vault sub resources to use Key Vault ID instead of URL

### LINKED ISSUE #2396: NXDOMAIN on mykeyvaultname.vault.azure.net should be (able to be) treated as azurerm_key_vault_{secret,certificate} deleted on refresh
### Community Note

### Description

It is generally the expected/desired behavior of Terraform that when a resource is deleted out of band and a subsequent plan or apply is run (after refresh) that terraform proposes to recreate any such resources present in the module but removed from state

In the case of an Azure keyvault's deletion, a terraform module which specifies that keyvault and (some of) that keyvault's contents ought to propose the recreation of the keyvault and those contents.

It does not.

What you get instead is a crashed refresh, with (potentially many) messages like
>azurerm_key_vault_certificate.ssl_cert: azurerm_key_vault_certificate.ssl_cert: keyvault.BaseClient#GetCertificate: Failure sending request: StatusCode=0 -- Original Error: Get https://mykeyvaultname.vault.azure.net/certificates/example-certificate/?api-version=2016-10-01: dial tcp: lookup mykeyvaultname.vault.azure.net on [2001:4898::1050:5050]:53: no such host

This is a result of the fact of implementation that a keyvault represents an independent hostname endpoint of the keyvault API, thus when a keyvault is removed, the entire API surface for that kevault is likewise removed, so the refresh fails on trying to get the {secret, certificate} state.

The code for this refresh https://github.com/terraform-providers/terraform-provider-azurerm/blob/f7248221c808b1ce605ff9d2857cfbc68da2e4db/azurerm/resource_arm_key_vault_secret.go#L160-L168

handles 404 on the secret itself as a warn and remove from state, but the NXDOMAIN on the keyvault itself simply crashes the refresh. But it is always the case that a deleted key vault's contents are also deleted, so I contend this sort of failure ought to be considered a deletion as well. 

I believe that the resource graph will reasonably ensure a recreate is not attempted until after a recreate of the vault itself has been done

I think this would be a reasonable default (always) behavior, but if the current behavior were considered for any reason desirable, it would be nice if there were at least some way to opt-in to this behavior

- azurerm_key_vault_certificate
- azurerm_key_vault_secret

katbyte:
This will allow us to easily check if the key vault exists before doing a update/read/delete on it causing a NXDOMAIN error (as the domain used for the API call vanishes when the key vault is deleted)

fixes #2396

### DIFF
--- a/azurerm/helpers/azure/key_vault.go
+++ b/azurerm/helpers/azure/key_vault.go
@@ -3,6 +3,7 @@ package azure
 import (
 	"context"
 	"fmt"
+	"log"
 
 	"github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2018-02-14/keyvault"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
@@ -38,6 +39,53 @@ func GetKeyVaultBaseUrlFromID(ctx context.Context, client keyvault.VaultsClient,
 	}
 
 	return *resp.Properties.VaultURI, nil
+}
+
+func GetKeyVaultIDFromBaseUrl(ctx context.Context, client keyvault.VaultsClient, keyVaultUrl string) (string, error) {
+
+	list, err := client.ListComplete(ctx, utils.Int32(1))
+	if err != nil {
+		return "", fmt.Errorf("Error GetKeyVaultId unable to list Key Vaults %v", err)
+	}
+
+	for list.NotDone() {
+		v := list.Value()
+		if v.ID == nil {
+			log.Printf("[DEBUG]GetKeyVaultId: v.ID was nil, continuing")
+			continue
+		}
+
+		vid, err := ParseAzureResourceID(*v.ID)
+		if err != nil {
+			log.Printf("[DEBUG] GetKeyVaultId: unable to parse v.ID (%s): %v", *v.ID, err)
+			continue
+		}
+		resourceGroup := vid.ResourceGroup
+		name := vid.Path["vaults"]
+
+		//resp does not appear to contain the vault properties, so lets fech them
+		get, err := client.Get(ctx, resourceGroup, name)
+		if err != nil {
+			log.Printf("[DEBUG] GetKeyVaultId: Error making Read request on KeyVault %q (Resource Group %q): %+v", name, resourceGroup, err)
+			continue
+		}
+
+		if get.ID == nil || get.Properties == nil || get.Properties.VaultURI == nil {
+			log.Printf("[DEBUG] GetKeyVaultId: KeyVault %q (Resource Group %q) has nil ID, properties or vault URI", name, resourceGroup)
+			continue
+		}
+
+		if keyVaultUrl == *get.Properties.VaultURI {
+			return *get.ID, nil
+		}
+
+		e := list.NextWithContext(ctx)
+		if e != nil {
+			return "", fmt.Errorf("Error GetKeyVaultId: Error getting next value on KeyVault %q (Resource Group %q): %+v", name, resourceGroup, err)
+		}
+	}
+
+	return "", fmt.Errorf("Error GetKeyVaultId unable to find Key Vault with url %q", keyVaultUrl)
 }
 
 func KeyVaultExists(ctx context.Context, client keyvault.VaultsClient, keyVaultId string) (bool, error) {
--- a/azurerm/helpers/azure/key_vault.go
+++ b/azurerm/helpers/azure/key_vault.go
@@ -23,7 +23,7 @@ func GetKeyVaultBaseUrlFromID(ctx context.Context, client keyvault.VaultsClient,
 
 	vaultName, ok := id.Path["vaults"]
 	if !ok {
-		return "", fmt.Errorf("resource id does not contain `valuts`: %q", keyVaultId)
+		return "", fmt.Errorf("resource id does not contain `vaults`: %q", keyVaultId)
 	}
 
 	resp, err := client.Get(ctx, resourceGroup, vaultName)
@@ -43,7 +43,7 @@ func GetKeyVaultBaseUrlFromID(ctx context.Context, client keyvault.VaultsClient,
 
 func GetKeyVaultIDFromBaseUrl(ctx context.Context, client keyvault.VaultsClient, keyVaultUrl string) (string, error) {
 
-	list, err := client.ListComplete(ctx, utils.Int32(1))
+	list, err := client.ListComplete(ctx, utils.Int32(1000))
 	if err != nil {
 		return "", fmt.Errorf("Error GetKeyVaultId unable to list Key Vaults %v", err)
 	}
@@ -102,7 +102,7 @@ func KeyVaultExists(ctx context.Context, client keyvault.VaultsClient, keyVaultI
 
 	vaultName, ok := id.Path["vaults"]
 	if !ok {
-		return false, fmt.Errorf("resource id does not contain `valuts`: %q", keyVaultId)
+		return false, fmt.Errorf("resource id does not contain `vaults`: %q", keyVaultId)
 	}
 
 	resp, err := client.Get(ctx, resourceGroup, vaultName)
--- a/azurerm/helpers/azure/key_vault.go
+++ b/azurerm/helpers/azure/key_vault.go
@@ -1,0 +1,73 @@
+package azure
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2018-02-14/keyvault"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func GetKeyVaultBaseUrlFromID(ctx context.Context, client keyvault.VaultsClient, keyVaultId string) (string, error) {
+
+	if keyVaultId == "" {
+		return "", fmt.Errorf("keyVaultId is empty")
+	}
+
+	id, err := ParseAzureResourceID(keyVaultId)
+	if err != nil {
+		return "", err
+	}
+	resourceGroup := id.ResourceGroup
+
+	vaultName, ok := id.Path["vaults"]
+	if !ok {
+		return "", fmt.Errorf("resource id does not contain `valuts`: %q", keyVaultId)
+	}
+
+	resp, err := client.Get(ctx, resourceGroup, vaultName)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return "", fmt.Errorf("Error unable to find KeyVault %q (Resource Group %q): %+v", vaultName, resourceGroup, err)
+		}
+		return "", fmt.Errorf("Error making Read request on KeyVault %q (Resource Group %q): %+v", vaultName, resourceGroup, err)
+	}
+
+	if resp.Properties == nil || resp.Properties.VaultURI == nil {
+		return "", fmt.Errorf("vault (%s) response properties or VaultURI is nil", keyVaultId)
+	}
+
+	return *resp.Properties.VaultURI, nil
+}
+
+func KeyVaultExists(ctx context.Context, client keyvault.VaultsClient, keyVaultId string) (bool, error) {
+
+	if keyVaultId == "" {
+		return false, fmt.Errorf("keyVaultId is empty")
+	}
+
+	id, err := ParseAzureResourceID(keyVaultId)
+	if err != nil {
+		return false, err
+	}
+	resourceGroup := id.ResourceGroup
+
+	vaultName, ok := id.Path["vaults"]
+	if !ok {
+		return false, fmt.Errorf("resource id does not contain `valuts`: %q", keyVaultId)
+	}
+
+	resp, err := client.Get(ctx, resourceGroup, vaultName)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return false, nil
+		}
+		return false, fmt.Errorf("Error making Read request on KeyVault %q (Resource Group %q): %+v", vaultName, resourceGroup, err)
+	}
+
+	if resp.Properties == nil || resp.Properties.VaultURI == nil {
+		return false, fmt.Errorf("vault (%s) response properties or VaultURI is nil", keyVaultId)
+	}
+
+	return true, nil
+}
--- a/azurerm/resource_arm_key_vault_certificate.go
+++ b/azurerm/resource_arm_key_vault_certificate.go
@@ -96,6 +96,7 @@ func resourceArmKeyVaultCertificate() *schema.Resource {
 			"vault_id": {
 				Type:         schema.TypeString,
 				Optional:     true, //todo required in 2.0
+				Computed:     true, //todo required in 2.0
 				ForceNew:     true,
 				ValidateFunc: azure.ValidateResourceID,
 			},
@@ -383,6 +384,12 @@ func resourceArmKeyVaultCertificateCreate(d *schema.ResourceData, meta interface
 		}
 
 		keyVaultBaseUrl = pKeyVaultBaseUrl
+	} else {
+		id, err := azure.GetKeyVaultIDFromBaseUrl(ctx, vaultClient, keyVaultBaseUrl)
+		if err != nil {
+			return fmt.Errorf("Error unable to find key vault ID from URL %q for certificate %q: %+v", keyVaultBaseUrl, name, err)
+		}
+		d.Set("vault_id", id)
 	}
 
 	if requireResourcesToBeImported {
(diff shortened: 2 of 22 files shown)


azurerm/helpers/azure/key_vault.go
```diff
@@ -0,0 +1,121 @@
...
+	if resp.Properties == nil || resp.Properties.VaultURI == nil {
+		return "", fmt.Errorf("vault (%s) response properties or VaultURI is nil", keyVaultId)
+	}
+
+	return *resp.Properties.VaultURI, nil
+}
+
+func GetKeyVaultIDFromBaseUrl(ctx context.Context, client keyvault.VaultsClient, keyVaultUrl string) (string, error) {
+
+	list, err := client.ListComplete(ctx, utils.Int32(1))
```
tombuildsstuff: this'll only return a single keyvault - from the SDK docs:
> // top - maximum number of results to return.

in other resources we set this to 1000 - I'd suggest it's worth doing the same here

```suggestion
	list, err := client.ListComplete(ctx, utils.Int32(1000))
```

azurerm/resource_arm_key_vault_certificate.go
```diff
@@ -19,14 +19,70 @@ import (
...
+			continue
+		}
+
+		if get.ID == nil || get.Properties == nil || get.Properties.VaultURI == nil {
+			log.Printf("[DEBUG] Key Vault Child import: KeyVault %q (Resource Group %q) has nil ID, properties or vault URI", name, resourceGroup)
+			continue
+		}
+
+		if id.KeyVaultBaseUrl == *get.Properties.VaultURI {
+			d.Set("vault_id", *get.ID)
```
tombuildsstuff: we shouldn't need to de-reference this here - d.Set should take care of that?

azurerm/resource_arm_key_vault_certificate.go
```diff
@@ -19,14 +19,70 @@ import (
...
+func resourceArmKeyVaultChildResourceImporter(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	client := meta.(*ArmClient).keyVaultClient
+	ctx := meta.(*ArmClient).StopContext
+
+	id, err := azure.ParseKeyVaultChildID(d.Id())
+	if err != nil {
+		return []*schema.ResourceData{d}, fmt.Errorf("Error Unable to parse ID (%s) for Key Vault Child import: %v", d.Id(), err)
+	}
+
+	list, err := client.ListComplete(ctx, utils.Int32(1))
```
tombuildsstuff: (as above) I believe this should be 1000?

azurerm/resource_arm_key_vault_certificate.go
```diff
@@ -37,10 +93,21 @@ func resourceArmKeyVaultCertificate() *schema.Resource {
 				ValidateFunc: azure.ValidateKeyVaultChildName,
 			},
 
+			"vault_id": {
```
tombuildsstuff: to better match the other resources I feel like this could make more sense as `key_vault_id`?

azurerm/resource_arm_key_vault_certificate.go
```diff
@@ -19,14 +19,70 @@ import (
...
+	if err != nil {
+		return []*schema.ResourceData{d}, fmt.Errorf("Error Unable to parse ID (%s) for Key Vault Child import: %v", d.Id(), err)
+	}
+
+	list, err := client.ListComplete(ctx, utils.Int32(1))
+	if err != nil {
+		return []*schema.ResourceData{d}, fmt.Errorf("Error Unable to list Key Vaults for Key Vault Child import: %v", err)
+	}
+
+	for list.NotDone() {
```
tombuildsstuff: could we not reuse the helper function here?

azurerm/resource_arm_key_vault_certificate.go
```diff
@@ -384,21 +472,31 @@ func resourceArmKeyVaultCertificateRead(d *schema.ResourceData, meta interface{}
...
+	cert, err := client.GetCertificate(ctx, id.KeyVaultBaseUrl, id.Name, "")
 	if err != nil {
 		if utils.ResponseWasNotFound(cert.Response) {
 			log.Printf("[DEBUG] Certificate %q was not found in Key Vault at URI %q - removing from state", id.Name, id.KeyVaultBaseUrl)
 			d.SetId("")
 			return nil
 		}
 
-		return err
+		return fmt.Errorf("Error reading Key Vault Certificate Policy: %+v", err)
```
tombuildsstuff:
```suggestion
		return fmt.Errorf("Error reading Key Vault Certificate: %+v", err)
```

azurerm/resource_arm_key_vault_key.go
```diff
@@ -31,10 +31,21 @@ func resourceArmKeyVaultKey() *schema.Resource {
 				ValidateFunc: azure.ValidateKeyVaultChildName,
 			},
 
+			"vault_id": {
```
tombuildsstuff: (as above) would this make more sense as `key_vault_id`?

azurerm/resource_arm_key_vault_secret.go
```diff
@@ -30,10 +30,21 @@ func resourceArmKeyVaultSecret() *schema.Resource {
 				ValidateFunc: azure.ValidateKeyVaultChildName,
 			},
 
+			"vault_id": {
```
tombuildsstuff: (as above) would this make more sense as `key_vault_id`?

azurerm/helpers/azure/key_vault.go
```diff
@@ -0,0 +1,121 @@
+package azure
```
tombuildsstuff: can we also apply these changes to the Key Vault Access Policy resource and the matching Data Sources too?

website/docs/r/key_vault_secret.html.markdown
```diff
@@ -82,7 +82,7 @@ The following arguments are supported:
 
 * `value` - (Required) Specifies the value of the Key Vault Secret.
 
-* `vault_uri` - (Required) Specifies the URI used to access the Key Vault instance, available on the `azurerm_key_vault` resource.
+* `vault_id` - (Required) Specifies the ID of the Key Vault instance the certificate resides in, available from the `azurerm_key_vault.id` property.
```
tombuildsstuff: would it make sense to make this consistent with the other ID fields?

```suggestion
* `vault_id` - (Required) The ID of the Key Vault where the Secret should be created.
```

website/docs/r/key_vault_key.html.markdown
```diff
@@ -81,7 +81,7 @@ The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the Key Vault Key. Changing this forces a new resource to be created.
 
-* `vault_uri` - (Required) Specifies the URI used to access the Key Vault instance, available on the `azurerm_key_vault` resource.
+* `vault_id` - (Required) Specifies the ID of the Key Vault instance the certificate resides in, available from the `azurerm_key_vault.id` property.
```
tombuildsstuff: would it make sense to make this consistent with the other ID fields?

```suggestion
* `vault_id` - (Required) The ID of the Key Vault where the Key should be created.
```

website/docs/r/key_vault_certificate.html.markdown
```diff
@@ -225,7 +225,7 @@ The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the Key Vault Certificate. Changing this forces a new resource to be created.
 
-* `vault_uri` - (Required) Specifies the URI used to access the Key Vault instance, available on the `azurerm_key_vault` resource.
+* `vault_id` - (Required) Specifies the ID of the Key Vault instance the certificate resides in, available from the `azurerm_key_vault.id` property.
```
tombuildsstuff: would it make sense to make this consistent with the other ID fields?

```suggestion
* `vault_id` - (Required) The ID of the Key Vault where the Certificate should be created.
```

azurerm/resource_arm_key_vault_certificate.go
katbyte:
```suggestion
	list, err := client.ListComplete(ctx, utils.Int32(1000))
```

azurerm/resource_arm_key_vault_certificate.go
katbyte:
```suggestion
			d.Set("vault_id", get.ID)
```

azurerm/resource_arm_key_vault_certificate.go
katbyte: Should we change `key_vault.vault_uri `key_vault.key_vault_uri` as well then?

azurerm/resource_arm_key_vault_certificate.go
katbyte: Reuse which helper function?

azurerm/helpers/azure/key_vault.go
katbyte: Key Vault Access Policy doesn't use a uri, but the vault name. Would we just be changing it to be consistent and deprecating the `name` property?

azurerm/resource_arm_key_vault_certificate.go
tombuildsstuff: I think this should be able to call `GetKeyVaultBaseUrlFromID`?

azurerm/resource_arm_key_vault_certificate.go
tombuildsstuff: could do - but it's called `vault_uri` in the API (whereas the other is a Terraform specific field) - so I feel like it'd be best not too - WDYT?

azurerm/helpers/azure/key_vault.go
tombuildsstuff: yep - this bug would still be possible with the Access Policy resource otherwise?

azurerm/resource_arm_key_vault_certificate.go
katbyte: It seems inconsistent to have one be `key_vault_id` and one `vault_uri` when the are essential the same

azurerm/helpers/azure/key_vault.go
katbyte: I don't believe so, the api it is using is the same one as we use to now check to see if the keyvault exists.

katbyte:
ds tests:
```
[21:22:13] kt@snowbook:~/hashi/..3../terraform-providers/terraform-provider-azurerm▸keyvault/refactor_children$ testazure TestAccDataSourceAzureRMKeyVault
==> Fixing source code with gofmt...
# This logic should match the search logic in scripts/gofmtcheck.sh
gofmt -s -w `find . -name '*.go' | grep -v vendor`
TF_ACC=1 go test ./azurerm -v -test.run=TestAccDataSourceAzureRMKeyVault -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccDataSourceAzureRMKeyVaultAccessPolicy_key
=== PAUSE TestAccDataSourceAzureRMKeyVaultAccessPolicy_key
=== RUN   TestAccDataSourceAzureRMKeyVaultAccessPolicy_secret
=== PAUSE TestAccDataSourceAzureRMKeyVaultAccessPolicy_secret
=== RUN   TestAccDataSourceAzureRMKeyVaultAccessPolicy_certificate
=== PAUSE TestAccDataSourceAzureRMKeyVaultAccessPolicy_certificate
=== RUN   TestAccDataSourceAzureRMKeyVaultAccessPolicy_keySecret
=== PAUSE TestAccDataSourceAzureRMKeyVaultAccessPolicy_keySecret
=== RUN   TestAccDataSourceAzureRMKeyVaultAccessPolicy_keyCertificate
=== PAUSE TestAccDataSourceAzureRMKeyVaultAccessPolicy_keyCertificate
=== RUN   TestAccDataSourceAzureRMKeyVaultAccessPolicy_secretCertificate
=== PAUSE TestAccDataSourceAzureRMKeyVaultAccessPolicy_secretCertificate
=== RUN   TestAccDataSourceAzureRMKeyVaultAccessPolicy_keySecretCertificate
=== PAUSE TestAccDataSourceAzureRMKeyVaultAccessPolicy_keySecretCertificate
=== RUN   TestAccDataSourceAzureRMKeyVaultKey_complete
=== PAUSE TestAccDataSourceAzureRMKeyVaultKey_complete
=== RUN   TestAccDataSourceAzureRMKeyVaultSecret_basic
=== PAUSE TestAccDataSourceAzureRMKeyVaultSecret_basic
=== RUN   TestAccDataSourceAzureRMKeyVaultSecret_complete
=== PAUSE TestAccDataSourceAzureRMKeyVaultSecret_complete
=== RUN   TestAccDataSourceAzureRMKeyVault_basic
=== PAUSE TestAccDataSourceAzureRMKeyVault_basic
=== RUN   TestAccDataSourceAzureRMKeyVault_complete
=== PAUSE TestAccDataSourceAzureRMKeyVault_complete
=== RUN   TestAccDataSourceAzureRMKeyVault_networkAcls
=== PAUSE TestAccDataSourceAzureRMKeyVault_networkAcls
=== CONT  TestAccDataSourceAzureRMKeyVaultAccessPolicy_key
=== CONT  TestAccDataSourceAzureRMKeyVaultKey_complete
=== CONT  TestAccDataSourceAzureRMKeyVault_complete
=== CONT  TestAccDataSourceAzureRMKeyVaultSecret_complete
=== CONT  TestAccDataSourceAzureRMKeyVault_basic
=== CONT  TestAccDataSourceAzureRMKeyVaultSecret_basic
=== CONT  TestAccDataSourceAzureRMKeyVaultAccessPolicy_keyCertificate
=== CONT  TestAccDataSourceAzureRMKeyVaultAccessPolicy_keySecretCertificate
=== CONT  TestAccDataSourceAzureRMKeyVaultAccessPolicy_secretCertificate
--- PASS: TestAccDataSourceAzureRMKeyVaultAccessPolicy_keySecretCertificate (22.98s)
--- PASS: TestAccDataSourceAzureRMKeyVaultAccessPolicy_key (23.09s)
=== CONT  TestAccDataSourceAzureRMKeyVault_networkAcls
--- PASS: TestAccDataSourceAzureRMKeyVaultAccessPolicy_keyCertificate (23.21s)
=== CONT  TestAccDataSourceAzureRMKeyVaultAccessPolicy_certificate
--- PASS: TestAccDataSourceAzureRMKeyVaultAccessPolicy_certificate (20.36s)
=== CONT  TestAccDataSourceAzureRMKeyVaultAccessPolicy_keySecret
--- PASS: TestAccDataSourceAzureRMKeyVaultAccessPolicy_secretCertificate (20.69s)
=== CONT  TestAccDataSourceAzureRMKeyVaultAccessPolicy_secret
--- PASS: TestAccDataSourceAzureRMKeyVaultAccessPolicy_secret (20.68s)
--- PASS: TestAccDataSourceAzureRMKeyVaultAccessPolicy_keySecret (20.80s)
--- PASS: TestAccDataSourceAzureRMKeyVaultSecret_basic (217.07s)
--- PASS: TestAccDataSourceAzureRMKeyVault_complete (218.80s)
--- PASS: TestAccDataSourceAzureRMKeyVault_basic (220.04s)
--- PASS: TestAccDataSourceAzureRMKeyVaultSecret_complete (231.18s)
--- PASS: TestAccDataSourceAzureRMKeyVaultKey_complete (231.94s)
--- PASS: TestAccDataSourceAzureRMKeyVault_networkAcls (273.55s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	296.690s
[21:30:09] kt@snowbook:~/hashi/..3../terraform-providers/terraform-provider-azurerm▸keyvault/refactor_children$
```

azurerm/helpers/azure/key_vault.go
tombuildsstuff: this is less the API and more the fact that when the Key Vault's recreated the Terraform Core bug where resources aren't picked up as requiring recreation until the second `plan` is still possible

katbyte: ah yes, either way i changed it to ID

azurerm/resource_arm_key_vault_certificate.go
mbfrahry: Just following up on this thread. Do we need to do anything with this?

katbyte: totally forgot, thanks for the reminder!

hbuckle:
It seems that `vault_uri` no longer works at all now (I know it is deprecated, but it should still work) as the logic to handle `vault_uri` or `key_vault_id` is only present in `resourceArmKeyVaultKeyCreate` - the other functions are only using `key_vault_id`

error:
```
azurerm_key_vault_key.key: Error checking if key vault "" for Key "key" in Vault at url "https://somevault.vault.azure.net/" exists: keyVaultId is empty
```

katbyte:
Thanks for the heads up @hbuckle. This is definitely a regression and being tracked in #2865, with a fix in #2874

ivanignatiev:
Verification helper functions are searching for KeyVault in the current context when KeyVault can be placed completely in other subscription which allowed to access if subscriptions are under the same tenant. 

Deprecation of `vault_uri ` parameter is blocking deployments in the cases of multi-subscription environments and raises "Resource Group not found" error.

Issue #2567, opened 2018-12-21
Output the access keys and endpoint url for azurerm_cognitive_account


perbergland:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

To create a complete environment where the api key and endpoint are used by some downstream application the resource needs to output these values, or at minimum the api keys.

### New or Affected Resource(s)

* azurerm_cognitive_account

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

* #0000

perbergland:
Background: We use the ComputerVision service

PR #2825, opened 2019-02-01
Added keys to output for cog service account

JimPaine:
When creating a cognitive service account the keys were not returned so you were unable to configure downstream services. 

Related Issue #2567 

Key things to note:
- I couldn't see tests around the other outputs so haven't modified them, i.e. "endpoint" happy to add if it is felt to be required.
- I also noticed that some resource change the terminology around key names for eg. key1 -> primary key but I have left to match the RESTful api and GO SDK.

### DIFF
--- a/azurerm/resource_arm_cognitive_account.go
+++ b/azurerm/resource_arm_cognitive_account.go
@@ -232,7 +232,7 @@ func resourceArmCognitiveAccountRead(d *schema.ResourceData, meta interface{}) e
 
 	if err != nil {
 		if utils.ResponseWasNotFound(resp.Response) {
-			log.Printf("[DEBUG] Cognitive Services Account %q was not found in Resource Group %q - removing from state!", name, resourceGroup)
+			log.Printf("[DEBUG] Not able to obtain keys for Cognitive Services Account %q in Resource Group %q - removing from state!", name, resourceGroup)
 			d.SetId("")
 			return nil
 		}
--- a/azurerm/resource_arm_cognitive_account.go
+++ b/azurerm/resource_arm_cognitive_account.go
@@ -100,6 +100,16 @@ func resourceArmCognitiveAccount() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"key1": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"key2": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -192,6 +202,7 @@ func resourceArmCognitiveAccountRead(d *schema.ResourceData, meta interface{}) e
 	name := id.Path["accounts"]
 
 	resp, err := client.GetProperties(ctx, resourceGroup, name)
+
 	if err != nil {
 		if utils.ResponseWasNotFound(resp.Response) {
 			log.Printf("[DEBUG] Cognitive Services Account %q was not found in Resource Group %q - removing from state!", name, resourceGroup)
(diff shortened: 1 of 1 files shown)


JimPaine:
Hi @katbyte It is also in the portal as key1 and key2 so while it may seem a little odd it is probably best to keep it aligned? I will work on the requested changes so thanks for the feedback.

azurerm/resource_arm_cognitive_account.go
```diff
@@ -209,14 +220,33 @@ func resourceArmCognitiveAccountRead(d *schema.ResourceData, meta interface{}) e
...
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[DEBUG] Not able to obtain keys for Cognitive Services Account %q in Resource Group %q - removing from state!", name, resourceGroup)
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	if key1 := keys.Key1; key1 != nil {
+		d.Set("key1", key1)
```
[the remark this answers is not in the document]
JimPaine: Wasn't aware of this, so thank you. Shall make the change

azurerm/resource_arm_cognitive_account.go
```diff
@@ -209,14 +220,33 @@ func resourceArmCognitiveAccountRead(d *schema.ResourceData, meta interface{}) e
...
 
 	if props := resp.AccountProperties; props != nil {
 		d.Set("endpoint", props.Endpoint)
 	}
 
+	keys, err := client.ListKeys(ctx, resourceGroup, name)
+
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[DEBUG] Not able to obtain keys for Cognitive Services Account %q in Resource Group %q - removing from state!", name, resourceGroup)
```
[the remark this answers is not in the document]
JimPaine: Yeah, I wasn't sure about this. I was thinking that the resource can't be used without the keys so fail hard. But actually because it is a 2 step process to read the resource and then the keys it probably makes more sense to return the error

JimPaine:
Tests and documents updated

azurerm/resource_arm_cognitive_account.go
```diff
@@ -100,6 +100,18 @@ func resourceArmCognitiveAccount() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"key1": {
```
tombuildsstuff: whilst I appreciate this is named `key1` in the API - could we name this `primary_access_key` for consistency with the other resources?

azurerm/resource_arm_cognitive_account.go
```diff
@@ -100,6 +100,18 @@ func resourceArmCognitiveAccount() *schema.Resource {
...
+				Type:      schema.TypeString,
+				Computed:  true,
+				Sensitive: true,
+			},
+
+			"key2": {
```
tombuildsstuff: whilst I appreciate this is named `key2` in the API - could we name this `secondary_access_key` for consistency with the other resources?

azurerm/resource_arm_cognitive_account.go
JimPaine: Updated properties, docs and test

website/docs/r/cognitive_account.html.markdown
```diff
@@ -67,6 +67,10 @@ The following attributes are exported:
 
 * `endpoint` - The endpoint used to connect to the Cognitive Service Account.
 
+* `primary_access_key` - A key that can be used when connecting to the Cognitive Service Account.
```
tombuildsstuff: could we make this:

```suggestion
* `primary_access_key` - A primary access key which can be used to connect to the Cognitive Service Account.
```

website/docs/r/cognitive_account.html.markdown
```diff
@@ -67,6 +67,10 @@ The following attributes are exported:
...
+
+* `secondary_access_key` - The second key that can be used as an alternative to key1 when connecting to the Cognitive Service Account.
```
tombuildsstuff: could we make this:

```suggestion
* `secondary_access_key` - The secondary access key which can be used to connect to the Cognitive Service Account.
```

tombuildsstuff:
Tests pass:

screenshot 2019-02-04 at 15 22 46

PR #2908, opened 2019-02-18
`azurerm_application_gateway` - support for redirect rules

RustyF:
**WIP**.  I still need to write the tests and update the docs but I'm creating this PR now so you guys can take an early look. 

I'm a go newbie and now using GoLand after grief with VSCode.  My formatting is screwed also :)


azurerm/resource_arm_application_gateway.go
```diff
@@ -444,6 +448,65 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+
+			"redirect_configuration": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
```
katbyte: could we validate this? with at least `validate.NoEmptyStrings()`?

azurerm/resource_arm_application_gateway.go
```diff
@@ -444,6 +448,65 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+								"Temporary",
+								"Found",
+								"SeeOther",
+							}, true),
+						},
+
+						"target_listener_name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
```
katbyte: could we validate this? with at least `validate.NoEmptyStrings()`?

RustyF:
Thanks @katbyte for that initial review and categorisation.  Of course, I will make the suggested changes.  

One question on semantic validation, there are more mutually exclusive settings now. Do you typically leave these to be checked by Azure or would you like me to try and add some validation in the code? E.g. you can't have a  target listener and a target URL.

katbyte:
@RustyF,

Usually we try to do as mch validation at the schema level so errors are caught during plan, or at the very least before trying the API as frequently the errors returned by azure are mysterious and unhelpful. But it's not always viable or required.

If you can add a check to `flattenApplicationGatewayRedirectConfigurations()` so an understandable error is floated up to the user that would suffice 🙂

azurerm/resource_arm_application_gateway.go
```diff
@@ -444,6 +448,55 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+
+						"redirect_type": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"Permanent",
```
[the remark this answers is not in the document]
RustyF: Done

azurerm/resource_arm_application_gateway.go
```diff
@@ -408,18 +407,28 @@ func resourceArmApplicationGateway() *schema.Resource {
 						"backend_address_pool_name": {
 							Type:     schema.TypeString,
 							Optional: true,
+							ConflictsWith: []string{"redirect_configuration_name"},
```
katbyte: Unfortunately as these are nested conflicts with won't work. The only way to actually validate this is upon expand check for the conflict and bubble up an error if they are both set.

RustyF: ok willdo

RustyF: Replaced with validation as suggested

RustyF:
Quick update, I've made all the suggested changes, added some validation. I've also added a couple of tests which should give it reasonable coverage.  I haven't actually run the tests yet 😮 as I'm having unrelated `make test` problems but will try again tomorrow.  Other than that, subject to obvious issues or things I've overlooked/not understood, it should be code-complete/stable.

azurerm/resource_arm_application_gateway.go
```diff
@@ -444,6 +449,68 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+
+						"redirect_type": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(network.Permanent),
+								string(network.Temporary),
+								string(network.Found),
+								string(network.SeeOther),
+							}, true),
```
tombuildsstuff: we're making all new fields case-insensitive - as such could we make this:

```suggestion
							}, false),
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -1865,6 +1971,14 @@ func flattenApplicationGatewayRequestRoutingRules(input *[]network.ApplicationGa
 	for _, config := range *input {
 		if props := config.ApplicationGatewayRequestRoutingRulePropertiesFormat; props != nil {
 
+			if props.BackendAddressPool != nil && props.RedirectConfiguration != nil {
```
tombuildsstuff: since this is a pointer to a string (`*string`) - I think we also want to check these aren't `*props.BackendAddressPool != ""` too, in an inner-check

azurerm/resource_arm_application_gateway.go
```diff
@@ -1865,6 +1971,14 @@ func flattenApplicationGatewayRequestRoutingRules(input *[]network.ApplicationGa
...
+				return nil, fmt.Errorf("[ERROR] Conflict between `backend_address_pool_name` and `redirect_configuration_name` (back-end pool not applicable when redirection specified)")
+			}
+
+			if props.BackendHTTPSettings != nil && props.RedirectConfiguration != nil {
```
tombuildsstuff: (same here)

azurerm/resource_arm_application_gateway.go
```diff
@@ -1925,6 +2039,121 @@ func flattenApplicationGatewayRequestRoutingRules(input *[]network.ApplicationGa
...
+		if props := config.ApplicationGatewayRedirectConfigurationPropertiesFormat; props != nil {
+
+			if config.TargetListener == nil && config.TargetURL == nil {
+				return nil, fmt.Errorf("[ERROR] Specify either `target_listener_name` or `target_url`")
+			}
+			if props.TargetListener != nil && config.TargetURL != nil {
+				return nil, fmt.Errorf("[ERROR] Conflict between `target_listener_name` and `target_url` (redirection is either to URL or target listener)")
+			}
+
+			if config.TargetURL != nil && config.IncludePath != nil {
```
tombuildsstuff: (same here) can we check if this is `!= nil && != ""`

azurerm/resource_arm_application_gateway.go
```diff
@@ -1925,6 +2039,121 @@ func flattenApplicationGatewayRequestRoutingRules(input *[]network.ApplicationGa
...
+func flattenApplicationGatewayRedirectConfigurations(input *[]network.ApplicationGatewayRedirectConfiguration) ([]interface{}, error) {
+	results := make([]interface{}, 0)
+	if input == nil {
+		return results, nil
+	}
+
+	for _, config := range *input {
+		if props := config.ApplicationGatewayRedirectConfigurationPropertiesFormat; props != nil {
+
+			if config.TargetListener == nil && config.TargetURL == nil {
```
tombuildsstuff: (same here) can we check if this is `!= nil && != ""`

azurerm/resource_arm_application_gateway.go
```diff
@@ -1925,6 +2039,121 @@ func flattenApplicationGatewayRequestRoutingRules(input *[]network.ApplicationGa
...
+				return nil, fmt.Errorf("[ERROR] Specify either `target_listener_name` or `target_url`")
+			}
+			if props.TargetListener != nil && config.TargetURL != nil {
```
tombuildsstuff: (same here) can we check if this is `!= nil && != ""`

azurerm/resource_arm_application_gateway.go
```diff
@@ -2097,6 +2335,9 @@ func expandApplicationGatewayURLPathMaps(d *schema.ResourceData, gatewayID strin
 				DefaultBackendHTTPSettings: &network.SubResource{
 					ID: utils.String(defaultBackendHTTPSettingsID),
 				},
+				DefaultRedirectConfiguration: &network.SubResource{
+					ID: utils.String(defaultRedirectConfigurationID),
+				},
```
tombuildsstuff: given these 3 ID's are now optional, I have a feeling that we'll need to update this to conditionally set them, for example:

```
if defaultRedirectConfigurationID != "" {
  output.ApplicationGatewayURLPathMapPropertiesFormat.DefaultRedirectConfiguration = &network.SubResource{
    ID: utils.String(defaultRedirectConfigurationID),
  },
}
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -2158,7 +2409,17 @@ func flattenApplicationGatewayURLPathMaps(input *[]network.ApplicationGatewayURL
 						ruleOutput["name"] = *rule.Name
 					}
 
+					// TODO: Note below: ruleProps versus props - bug?
 					if ruleProps := rule.ApplicationGatewayPathRulePropertiesFormat; props != nil {
```
tombuildsstuff: agreed, this looks like a bug; would you mind fixing this? :)

azurerm/resource_arm_application_gateway.go
```diff
@@ -2158,7 +2409,17 @@ func flattenApplicationGatewayURLPathMaps(input *[]network.ApplicationGatewayURL
...
+
+						if ruleProps.BackendAddressPool != nil && ruleProps.RedirectConfiguration != nil {
```
tombuildsstuff: (same here) could we check != nil && != ""?

azurerm/resource_arm_application_gateway.go
```diff
@@ -2158,7 +2409,17 @@ func flattenApplicationGatewayURLPathMaps(input *[]network.ApplicationGatewayURL
...
+							return nil, fmt.Errorf("[ERROR] Conflict between `backend_address_pool_name` and `redirect_configuration_name` (back-end pool not applicable when redirection specified)")
+						}
+
+						if ruleProps.BackendHTTPSettings != nil && ruleProps.RedirectConfiguration != nil {
```
tombuildsstuff: (same here) could we check != nil && != ""?

website/docs/r/application_gateway.html.markdown
```diff
@@ -369,6 +372,22 @@ A `custom_error_configuration` block supports the following:
...
 
+---
+
+A `redirect_configuration` block supports the following:
+
+* `name` - (Required) Unique name of the redirect configuration block
+
+* `redirect_type` - (Required) The type of redirect. Possible values are `Permanent`, `Temporary`, `Found` and `SeeOther`
+
+* `target_listener_name` - (Optional) The name of the listener to redirect to. This parameter is mutually exclusive with `target_url`.
```
tombuildsstuff: for consistency with other resources, can we make this:

```suggestion
* `target_listener_name` - (Optional) The name of the listener to redirect to. This field must be set when `target_url` is set.
```

website/docs/r/application_gateway.html.markdown
```diff
@@ -369,6 +372,22 @@ A `custom_error_configuration` block supports the following:
...
+
+* `target_url` - (Optional) The Url to redirect the request to. This parameter is mutually exclusive with `target_listener_name`.
```
tombuildsstuff: for consistency with other resources, can we make this:

```suggestion
* `target_url` - (Optional) The Url to redirect the request to. This field must be set when `target_listener_name` is set.
```

azurerm/resource_arm_application_gateway.go
RustyF: done

azurerm/resource_arm_application_gateway.go
RustyF: Hi @tombuildsstuff,  I total get the string example - I need to get my C++ head back on ;). However, in the example, `props.BackendAddressPool` here is a `SubResource *`.  So should I also be doing `props.BackendAddressPool != nil' and `props.BackendAddressPool.ID != nil`  and `*props.BackendAddressPool.ID != nil`  ? Or am I losing the plot?

azurerm/resource_arm_application_gateway.go
RustyF: Now done this with utility func

azurerm/resource_arm_application_gateway.go
RustyF: Done

azurerm/resource_arm_application_gateway.go
RustyF: again, these are SubResource * - Feels like we need a predicate func like `ValidSubReference()` which returns true if pointing to a valid resource ID.

azurerm/resource_arm_application_gateway.go
RustyF: As above

website/docs/r/application_gateway.html.markdown
RustyF: Yes, sure, I assume you meant "field must NOT be set when `target_url` is set" ;)  It's a mutual exclusion.

website/docs/r/application_gateway.html.markdown
RustyF: Done

website/docs/r/application_gateway.html.markdown
RustyF: Done. I made the changes from my branch

azurerm/resource_arm_application_gateway.go
tombuildsstuff: ah I'd missed this was a `*SubResource` in which case, yes we'll want to make this:

```
hasBackendAddressPool := props.BackendAddressPool != nil && props.BackendAddressPool.ID != nil && *props.BackendAddressPool.ID != ""
hasRedirectConfiguration := props.RedirectConfiguration != nil && props.RedirectConfiguration.ID != nil && * props.RedirectConfiguration.ID != ""
if hasBackendAddressPool && hasRedirectConfiguration {
  # ...
}
```

azurerm/resource_arm_application_gateway.go
tombuildsstuff: we probably could split this out into a private function - but be aware in Golang private functions (starting with a lower-case character) are shared throughout the package - so the naming would need to be more `applicationGatewayValidateSubResource(...)`

azurerm/resource_arm_application_gateway.go
RustyF: Done, I think

azurerm/resource_arm_application_gateway.go
RustyF: Done

azurerm/resource_arm_application_gateway.go
RustyF: Done

azurerm/resource_arm_application_gateway.go
RustyF: Done

azurerm/resource_arm_application_gateway.go
RustyF: done

RustyF:
Quick update, I've made the outstanding recommended changes.  I tried the acceptance tests. The basic rule with redirect timed out but otherwise appeared to be created OK so that gives me some faith.  However, the second, path-based-rule-with-redirect test fails with error 400. I haven't yet worked out the cause but it must be related to path-based redirect rules.

RustyF:
Update, the first acceptance test works, i.e. basic rule with re-direction to HTTP listener. It just takes a while to run and obviously timed out yesterday.  I'm still diagnosing the path map redirection test.

RustyF:
Ran out of time today but think I've resolved the problems with path URL map-based redirects.  Will run the acceptance tests tomorrow and confirm.

RustyF:
Acceptance tests are passing so we are pretty much there now, I think 😉.  I'm going to review the docs again as I haven't looked at them since the final tweaks.

RustyF:
Please take a final look at the code.  As a go newbie, I won't be doing idiomatic things yet.

RustyF:
Doc updates done.

RustyF:
Let me know if you need me to do anything?

azurerm/helpers/validate/strings.go
```diff
@@ -18,3 +18,7 @@ func NoEmptyStrings(i interface{}, k string) ([]string, []error) {
 
 	return nil, nil
 }
+
+func NilOrEmpty(s *string) bool {
```
katbyte: I'm not sure the value of this function? i feel it may be better to simply have the test inline..

azurerm/resource_arm_application_gateway.go
```diff
@@ -12,6 +10,7 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+	"log"
```
katbyte: could we goimports this file? `log` should be above

azurerm/helpers/validate/strings.go
RustyF: I agree it's providing borderline value - dropped.

azurerm/resource_arm_application_gateway.go
RustyF: Done

azurerm/resource_arm_application_gateway.go
RustyF: Sorry - yes, already done

azurerm/resource_arm_application_gateway.go
RustyF: Done - thanks for the tip

azurerm/helpers/validate/strings.go
RustyF: Done

azurerm/resource_arm_application_gateway.go
RustyF: Done

RustyF:
I ran the acceptance tests again successfully.

RustyF:
@katbyte , @tombuildsstuff thanks for your support and patience on this 👍

OffColour:
@RustyF Massive thanks for your work on this. Started using it yesterday and works perfectly!

RustyF:
Thanks for the feedback - good to know

PR #2920, opened 2019-02-20
Allow 0 capacity for servicebus

hbuckle:
Allow the capacity to be set to 0 for service bus namespace. This is so we can create a module with the option for the premium sku - currently we have to maintain 2 modules.

Example module
```hcl
variable "sku" {
  default = "Basic"
}

variable "capacity" {
  default = 0
}

resource "azurerm_servicebus_namespace" "test" {
  name                = "name"
  sku                 = "${var.sku}"
  capacity            = "${var.capacity}"
}
```

Usage
```hcl
module "servicebusbasic" {
  sku = "Basic"
}

module "servicebuspremium" {
  sku      = "Premium"
  capacity = 4
}
```


hbuckle:
@katbyte technically you can only set capacity for the premium sku and it has to be 1,2,4
The API will accept 0 for basic or standard but it has no effect - so this is kind of a workaround but it enables the scenario outlined above to have a single module that supports multiple skus

katbyte:
@hbuckle,

Tried setting a basic SKU to `1` and see what your saying now, API just sets it to a 0. Could we put the validation back in and instead have it enforce 0 for basic and 1,2,4 for the premium SKUs? As well as adding a note to the documentation stating this too.

tombuildsstuff:
@katbyte 

> Could we put the validation back in and instead have it enforce 0 for basic and 1,2,4 for the premium SKUs? As well as adding a note to the documentation stating this too.

Taking a look through I believe this is now covered, since the validation method only allows the 4 possible values (0, 1, 2 and 4) and the Create method validates it's either 0 or not - as such I think this is fine? 👍

hbuckle:
Correct, the validation has moved to create and should only allow 0 for non-premium skus

azurerm/resource_arm_servicebus_namespace.go
```diff
@@ -145,6 +131,12 @@ func resourceArmServiceBusNamespaceCreateUpdate(d *schema.ResourceData, meta int
 	}
 
 	if capacity, ok := d.GetOk("capacity"); ok {
+		if !strings.EqualFold(sku, string(servicebus.Premium)) && capacity.(int) > 0 {
+			return fmt.Errorf("Service Bus SKU %q only supports `capacity` of 0", sku)
+		}
+		if strings.EqualFold(sku, string(servicebus.Premium)) && capacity.(int) == 0 {
+			return fmt.Errorf("Service Bus SKU %q only supports `capacity` of 1, 2 or 4", sku)
+		}
 		parameters.Sku.Capacity = utils.Int32(int32(capacity.(int)))
```
tombuildsstuff: @hbuckle so I'd missed this, but `d.GetOk` doesn't work for non-string types in Terraform 0.11 - which meant this test currently fails:

```
------- Stdout: -------
=== RUN   TestAccAzureRMServiceBusNamespace_premiumCapacity
=== PAUSE TestAccAzureRMServiceBusNamespace_premiumCapacity
=== CONT  TestAccAzureRMServiceBusNamespace_premiumCapacity
--- FAIL: TestAccAzureRMServiceBusNamespace_premiumCapacity (467.65s)
    testing.go:531: Step 0, expected error:
        
        After applying this step, the plan was not empty:
        
        DIFF:
        
        UPDATE: azurerm_servicebus_namespace.test
          capacity: "1" => "0"
        
        STATE:
        
        azurerm_resource_group.test:
          ID = /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/acctestRG-190222103918073308
          provider = provider.azurerm
          location = westeurope
          name = acctestRG-190222103918073308
          tags.% = 0
        azurerm_servicebus_namespace.test:
          ID = /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/acctestRG-190222103918073308/providers/Microsoft.ServiceBus/namespaces/acctestservicebusnamespace-190222103918073308
          provider = provider.azurerm
          capacity = 1
          default_primary_connection_string = Endpoint=sb://acctestservicebusnamespace-190222103918073308.servicebus.windows.net/;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=XXXXXX=
          default_primary_key = XXXXXX=
          default_secondary_connection_string = Endpoint=sb://acctestservicebusnamespace-190222103918073308.servicebus.windows.net/;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=KQQ3rbM+XXXXXX=
          default_secondary_key = KQQ3rbM+XXXXXX=
          location = westeurope
          name = acctestservicebusnamespace-190222103918073308
          resource_group_name = acctestRG-190222103918073308
          sku = premium
          tags.% = 0
        
          Dependencies:
            azurerm_resource_group.test
        
        To match:
        
        Service Bus SKU "Premium" only supports `capacity` of 1, 2 or 4
```

could we update this to always send the `capacity` field?

hbuckle: I can't seem to get this to work, even if always passing a value. The problem is when you don't set anything for capacity it is still interpreted as being 0 by Terraform so you still get a diff. Is there a way to distinguish between a user explicitly setting 0 and leaving the value unset?

katbyte: @hbuckle, unfortunately i don't think there is any way for us to determine if 0 was specifically set or the value was left empty. Maybe the best solution is to mark the property as required so we can property do validation, WDYT?

hbuckle: @katbyte - yes I was thinking that might be the way. I can default it to 0 and add validation for the premium sku. It would be a breaking change if people had used premium and left it unset, but hopefully that won't be too many people

tombuildsstuff:
Tests pass:

screenshot 2019-03-07 at 11 18 34

PR #2937, opened 2019-02-22
provisionAfterExtension field for chaining multiple extensions…

imvalient:
I am currently working with scale sets extensions and found out there is no way to define dependencies between scale-set extensions at ARM template level, through terraform. 

Reading through the GO Azure SDK I found this feature exists, so I made an implementation that translates a terraform array of extensions dependencies to the AzureRM json based `provisionAfterExtensions` attribute.

Usage
```terraform
resource "azurerm_virtual_machine_scale_set" "test_scale_set" {
    name = "scale-set-test-extension"
    ...
    
    extension {
        name = "first-extension"
        publisher = "Microsoft.Compute"
        ...
    }

    extension {
        name = "second-extension"
        publisher = "Microsoft.Compute"
        ...
        provision_after_extensions = ["first-extension"]
    }
}
```

This way the second extension will start only when the first one finishes, instead of running both at the same time.

As far as i know virtual machine extensions also do not implement this feature, you can define a terraform dependece using `dependsOn`, this could work at creation time but, once created, if the machine scales up, the dependency is lost as its not defined in the azure rm template, if this approach is alright we could add it to that resource as well.

### DIFF
--- a/azurerm/resource_arm_virtual_machine_scale_set.go
+++ b/azurerm/resource_arm_virtual_machine_scale_set.go
@@ -21,10 +21,12 @@ import (
 
 func resourceArmVirtualMachineScaleSet() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceArmVirtualMachineScaleSetCreateUpdate,
-		Read:   resourceArmVirtualMachineScaleSetRead,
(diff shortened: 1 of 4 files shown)


imvalient:
The requested changed have been submitted.

Issue #2935, opened 2019-02-22
Service Fabric - Node Type - Placement Properties and Capacities


steve-hawkins:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

For multi node type clusters you will more than likely want to have some sort of control over placement of services, for more information [refer to this article](https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-cluster-resource-manager-cluster-description#node-properties-and-placement-constraints)

### New or Affected Resource(s)

* azurerm_service_fabric_cluster

### Potential Terraform Configuration

```hcl
resource "azurerm_resource_group" "test" {
  name     = "test-cluster"
  location = "uksouth"
}

resource "azurerm_service_fabric_cluster" "test" {
  name                = "test-cluster"
  resource_group_name = "${azurerm_resource_group.test.name}"
  location            = "${azurerm_resource_group.test.location}"
  reliability_level   = "Bronze"
  upgrade_mode        = "Automatic"
  vm_image            = "Windows"
  management_endpoint = "http://example:80"

  node_type {
    name                 = "first"
    placement_properties {
      "HasSSD" = "true"
    }
    capacities {
      "ClientConnections" = "20000"
      "MemoryGB" = "8"
    }
    instance_count       = 3
    is_primary           = true
    client_endpoint_port = 2020
    http_endpoint_port   = 80
  }

  tags {
    "Hello" = "World"
  }
}
```

### References

#2936

tombuildsstuff:
Fixed via #2936 - thanks again @steve-hawkins :)

PR #2950, opened 2019-02-26
Add placement_properties to service_fabric_cluster resource

maxbog:
Also, add test and documentation


steve-hawkins:
Hi @maxbog placement properties and capacity tags were added recently via issue #2935

maxbog:
@steve-hawkins Wow, when I started working on that, there was no support for them :) Thanks for the find, I'll cancel this PR.

PR #2951, opened 2019-02-26
Fixes to IoTHub diff algorithm

maxbog:
Specific fixes:
- properties `batch_frequency_in_seconds`, `max_chunk_size_in_bytes`, `encoding`, `container_name`, `file_name_format` are only applicable to endpoints of type `AzureIotHub.StorageContainer`
- connection_string diff algorithm takes into account the masked Access Keys when reading the connection string from Azure
- connection_string diff algorithm takes into account the default port for EventHub endpoints

### DIFF
--- a/azurerm/resource_arm_iothub.go
+++ b/azurerm/resource_arm_iothub.go
@@ -23,7 +23,9 @@ import (
 
 func suppressIfTypeIsNot(t string) schema.SchemaDiffSuppressFunc {
 	return func(k, old, new string, d *schema.ResourceData) bool {
-		return d.Get("type").(string) != t
+		path := strings.Split(k, ".")
+		path[len(path)-1] = "type"
+		return d.Get(strings.Join(path, ".")).(string) != t
 	}
 }
 
--- a/azurerm/resource_arm_iothub.go
+++ b/azurerm/resource_arm_iothub.go
@@ -21,6 +21,23 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
+func suppressIfTypeIsNot(t string) schema.SchemaDiffSuppressFunc {
+	return func(k, old, new string, d *schema.ResourceData) bool {
+		return d.Get("type").(string) != t
+	}
+}
+
+func supressWhenAll(fs ...schema.SchemaDiffSuppressFunc) schema.SchemaDiffSuppressFunc {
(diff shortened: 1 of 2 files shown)


maxbog:
@tombuildsstuff Is there anyone that can review the changes? One of the things this fixes is secret leakage to the console. Right now, connection strings with secret access keys are shown in plain text while printing the execution plan.

maxbog:
@katbyte Thanks for the review, I extracted the obvious parts to a different PR: #3007 
I will remove those changes from this PR.

As for the `suppressIfTypeIsNot` checks:
The problem is that the `endpoint` element supports different types of endpoints (Storage container, Service Bus, Event Hub). However, the properties that have suppressed diffs are only applicable to `AzureIotHub.StorageContainer` type. This, combined with them being optional and having a default value, caused the diff to show changes in their value when creating an endpoint of a different type. 

For example when I defined the following simple endpoint:
```terraform
endpoint {
    type = "AzureIotHub.EventHub"

    connection_string = "${azurerm_eventhub_authorization_rule.device_messages_iothubroutes.primary_connection_string}"
    name = "device-messages"
}
```

The second apply, which should not show any changes, actually shows this:
```
endpoint.0.batch_frequency_in_seconds: "0" => "300"
endpoint.0.max_chunk_size_in_bytes:    "0" => "314572800"
endpoint.0.type:                       "" => "AzureIotHub.EventHub"
```
Of course, this does not actually result in any changes being made, so the diff output is incorrect.

The approach adopted here with the `SuppressDiffFunc` is used in AWS provider, that's where I got the inspiration:
https://github.com/terraform-providers/terraform-provider-aws/blob/042ed092aaeafb5e4beef583adb0d6bb8695c9c4/aws/resource_aws_lb.go#L176

However, please do not merge this PR for now, as I think there may be an issue with `DiffSuppressFunc` of the `connection_string` property.

maxbog:
@katbyte I believe the PR can be merged now, I changed the diff supression logic a bit for `connection_string` property

maxbog:
@katbyte I fixed the code, the culprit was: https://github.com/terraform-providers/terraform-provider-azurerm/commit/c7503196414d40b62da32be7eb71721a7f45ca54

I also extended the acceptance test to account for `EventHub` endpoint type.

maxbog:
@katbyte: Any comments?

katbyte:
Apologies for the delay @maxbog, my attention was elsewhere. Tests now pass and so this LGTM 🚀

PR #2981, opened 2019-03-01
further fixes to  test tf configs

katbyte:


azurerm/resource_arm_availability_set.go
```diff
@@ -147,12 +146,8 @@ func resourceArmAvailabilitySetRead(d *schema.ResourceData, meta interface{}) er
...
 
 	if props := resp.AvailabilitySetProperties; props != nil {
-		if v := props.PlatformUpdateDomainCount; v != nil {
-			d.Set("platform_update_domain_count", strconv.Itoa(int(*v)))
-		}
-		if v := props.PlatformFaultDomainCount; v != nil {
-			d.Set("platform_fault_domain_count", strconv.Itoa(int(*v)))
-		}
+		d.Set("platform_update_domain_count", props.PlatformUpdateDomainCount)
+		d.Set("platform_fault_domain_count", props.PlatformFaultDomainCount)
```
tombuildsstuff: is this a regression for 0.12 that works in 0.11? if so do we want a 0.12-code specific feature toggle?

katbyte: This was an accidental unnecessary fix i pushed into the last PR like this

PR #2967, opened 2019-02-28
New Resource: `azurerm_eventgrid_event_subscription`

mbfrahry:
This PR adds eventgrid event subscription to the azure provider. Addresses #1794. 

```
=== RUN   TestAccAzureRMEventGridEventSubscription_basic
=== PAUSE TestAccAzureRMEventGridEventSubscription_basic
=== RUN   TestAccAzureRMEventGridEventSubscription_eventhub
=== PAUSE TestAccAzureRMEventGridEventSubscription_eventhub
=== RUN   TestAccAzureRMEventGridEventSubscription_update
=== PAUSE TestAccAzureRMEventGridEventSubscription_update
=== RUN   TestAccAzureRMEventGridEventSubscription_filter
=== PAUSE TestAccAzureRMEventGridEventSubscription_filter
=== CONT  TestAccAzureRMEventGridEventSubscription_basic
=== CONT  TestAccAzureRMEventGridEventSubscription_eventhub
=== CONT  TestAccAzureRMEventGridEventSubscription_update
=== CONT  TestAccAzureRMEventGridEventSubscription_filter
--- PASS: TestAccAzureRMEventGridEventSubscription_filter (178.35s)
--- PASS: TestAccAzureRMEventGridEventSubscription_basic (178.35s)
--- PASS: TestAccAzureRMEventGridEventSubscription_eventhub (238.15s)
--- PASS: TestAccAzureRMEventGridEventSubscription_update (259.45s)
```

PR #2986, opened 2019-03-01
New resource: `azurerm_api_management_property`

mbfrahry:
This PR adds `azurerm_api_management_property` to the provider.

```
=== RUN   TestAccAzureRMAPIManagementProperty_basic
=== PAUSE TestAccAzureRMAPIManagementProperty_basic
=== RUN   TestAccAzureRMAPIManagementProperty_update
=== PAUSE TestAccAzureRMAPIManagementProperty_update
=== CONT  TestAccAzureRMAPIManagementProperty_basic
=== CONT  TestAccAzureRMAPIManagementProperty_update
--- PASS: TestAccAzureRMAPIManagementProperty_basic (1603.38s)
--- PASS: TestAccAzureRMAPIManagementProperty_update (1992.85s)
```


azurerm/resource_arm_api_management_property.go
```diff
@@ -0,0 +1,166 @@
...
+	if _, err := client.CreateOrUpdate(ctx, resourceGroup, serviceName, name, parameters, ""); err != nil {
+		return fmt.Errorf("Error creating or updating Property %q (Resource Group %q / API Management Service %q): %+v", name, resourceGroup, serviceName, err)
+	}
+
+	resp, err := client.Get(ctx, resourceGroup, serviceName, name)
+	if err != nil {
+		return fmt.Errorf("Error retrieving Property %q (Resource Group %q / API Management Service %q): %+v", name, resourceGroup, serviceName, err)
+	}
+	if resp.ID == nil {
+		return fmt.Errorf("Cannot read ID for Group %q (Resource Group %q / API Management Service %q)", name, resourceGroup, serviceName)
```
katbyte: I think this should be:
```suggestion
		return fmt.Errorf("Cannot read ID for Property %q (Resource Group %q / API Management Service %q)", name, resourceGroup, serviceName)
```

website/azurerm.erb
```diff
@@ -71,6 +71,10 @@
                     <a href="/docs/providers/azurerm/d/api_management_product.html">azurerm_api_management_product</a>
                 </li>
 
+                <li<%= sidebar_current("docs-azurerm-datasource-api-management-property-x") %>>
```
katbyte: I only see a resource .go and no datasource?

website/azurerm.erb
```diff
@@ -857,6 +861,10 @@
                   <a href="/docs/providers/azurerm/r/eventgrid_domain.html">azurerm_eventgrid_domain</a>
                 </li>
 
+                <li<%= sidebar_current("docs-azurerm-resource-messaging-eventgrid-event-subscription") %>>
```
katbyte: I think this is an error? 🤔

website/azurerm.erb
tombuildsstuff: from what I can see, I don't think this Resource warrants a Data Source at this time - so I think this is fine 👍

website/azurerm.erb
tombuildsstuff: this appears to be otherwise missing from the sidebar (and shipped in #2967) - so this seems ok to me?

website/docs/r/api_management_property.html.markdown
```diff
@@ -0,0 +1,75 @@
...
+
+* `resource_group_name` - (Required) The name of the Resource Group in which the API Management Property should exist. Changing this forces a new resource to be created.
+
+* `api_management_name` - (Required) The name of the [API Management Service](api_management.html) in which the API Management Property should exist. Changing this forces a new resource to be created.
+
+* `display_name` - (Required) The display name of this API Management Property.
+
+* `value` - (Required) The value of this API Management Property.
+
+* `secret` - (Optional) Specifies whether the API Management Property is secret. Valid values are `true` or `false`. The default value is `false`.
```
tombuildsstuff: I feel like it could be worth a yellow info box here explaining this isn't making it sensitive within Terraform e.g.

```
~> **NOTE:** setting the field `secret` to `true` doesn't make this field sensitive in Terraform, instead it ...
```

website/azurerm.erb
mbfrahry: Yea, this didn't make it in #2967 so I'm just adding it here

website/docs/r/api_management_property.html.markdown
mbfrahry: Done!

website/azurerm.erb
mbfrahry: You're right here @katbyte, moved this link to the correct postition

PR #2987, opened 2019-03-02
AKS : Advanced Networking / Calico Network Policy

thatInfrastructureGuy:
This PR aims to add Calico Network policy to AKS as described in [MS Docs](https://docs.microsoft.com/en-us/azure/aks/use-network-policies).

Added `network_policy` to `kubernetes_cluster` resource.

PR Inspiration: https://github.com/terraform-providers/terraform-provider-azurerm/pull/1479

### DIFF
--- a/azurerm/data_source_kubernetes_cluster.go
+++ b/azurerm/data_source_kubernetes_cluster.go
@@ -618,7 +618,7 @@ func flattenKubernetesClusterDataSourceNetworkProfile(profile *containerservice.
 	values["network_plugin"] = profile.NetworkPlugin
 
 	if profile.NetworkPolicy != "" {
-		values["network_policy"] = profile.NetworkPolicy
+		values["network_policy"] = string(profile.NetworkPolicy)
 	}
 
 	if profile.ServiceCidr != nil {
--- a/azurerm/data_source_kubernetes_cluster.go
+++ b/azurerm/data_source_kubernetes_cluster.go
@@ -249,6 +249,11 @@ func dataSourceArmKubernetesCluster() *schema.Resource {
 							Computed: true,
 						},
 
+						"network_policy": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
 						"service_cidr": {
 							Type:     schema.TypeString,
 							Computed: true,
@@ -611,6 +616,10 @@ func flattenKubernetesClusterDataSourceNetworkProfile(profile *containerservice.
 	values := make(map[string]interface{})
 
 	values["network_plugin"] = profile.NetworkPlugin
+
+	if profile.NetworkPolicy != "" {
+		values["network_policy"] = profile.NetworkPolicy
+	}
 
 	if profile.ServiceCidr != nil {
 		values["service_cidr"] = *profile.ServiceCidr
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -43,12 +43,6 @@ func resourceArmKubernetesCluster() *schema.Resource {
 					return nil
 				}
 
-				networkPolicy := profile["network_policy"].(string)
-
-				if networkPolicy != "calico" {
-					return nil
-				}
-
 				dockerBridgeCidr := profile["docker_bridge_cidr"].(string)
 				dnsServiceIP := profile["dns_service_ip"].(string)
 				serviceCidr := profile["service_cidr"].(string)
@@ -1069,7 +1063,7 @@ func flattenKubernetesClusterNetworkProfile(profile *containerservice.NetworkPro
 	values["network_plugin"] = profile.NetworkPlugin
 
 	if profile.NetworkPolicy != "" {
-		values["network_policy"] = profile.NetworkPolicy
+		values["network_policy"] = string(profile.NetworkPolicy)
 	}
 
 	if profile.ServiceCidr != nil {
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -43,6 +43,12 @@ func resourceArmKubernetesCluster() *schema.Resource {
 					return nil
 				}
 
+				networkPolicy := profile["network_policy"].(string)
+
+				if networkPolicy != "calico" {
+					return nil
+				}
+
 				dockerBridgeCidr := profile["docker_bridge_cidr"].(string)
 				dnsServiceIP := profile["dns_service_ip"].(string)
 				serviceCidr := profile["service_cidr"].(string)
@@ -309,6 +315,16 @@ func resourceArmKubernetesCluster() *schema.Resource {
 							ValidateFunc: validation.StringInSlice([]string{
 								string(containerservice.Azure),
 								string(containerservice.Kubenet),
+							}, false),
+						},
+
+						"network_policy": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+							ForceNew: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(containerservice.Calico),
 							}, false),
 						},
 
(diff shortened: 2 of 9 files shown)


azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -43,6 +43,12 @@ func resourceArmKubernetesCluster() *schema.Resource {
 					return nil
 				}
 
+				networkPolicy := profile["network_policy"].(string)
+
+				if networkPolicy != "calico" {
+					return nil
```
katbyte: Did you mean to return an error here?

I don't think this is required as validation should prevent any other value?

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -1049,6 +1068,10 @@ func flattenKubernetesClusterNetworkProfile(profile *containerservice.NetworkPro
 
 	values["network_plugin"] = profile.NetworkPlugin
 
+	if profile.NetworkPolicy != "" {
+		values["network_policy"] = profile.NetworkPolicy
```
katbyte: I think we should be casting this to a string?
```suggestion
		values["network_policy"] = string(profile.NetworkPolicy)
```

azurerm/data_source_kubernetes_cluster.go
```diff
@@ -612,6 +617,10 @@ func flattenKubernetesClusterDataSourceNetworkProfile(profile *containerservice.
 
 	values["network_plugin"] = profile.NetworkPlugin
 
+	if profile.NetworkPolicy != "" {
+		values["network_policy"] = profile.NetworkPolicy
```
katbyte: I think we should cast this to a string?
```suggestion
		values["network_policy"] = string(profile.NetworkPolicy)
```

azurerm/resource_arm_kubernetes_cluster_test.go
```diff
@@ -1122,3 +1170,143 @@ resource "azurerm_kubernetes_cluster" "test" {
...
+  }
+
+  service_principal {
+    client_id     = "%s"
+    client_secret = "%s"
+  }
+
+  network_profile {
+    network_plugin = "%s"
+		network_policy = "%s"
```
katbyte: Could we fix the alignment here?

azurerm/resource_arm_kubernetes_cluster_test.go
```diff
@@ -1122,3 +1170,143 @@ resource "azurerm_kubernetes_cluster" "test" {
...
+  }
+
+  service_principal {
+    client_id     = "%s"
+    client_secret = "%s"
+  }
+
+  network_profile {
+		network_plugin     = "%s"
+		network_policy     = "%s"
```
katbyte: could we fix the alignment here?

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -146,6 +146,8 @@ A `network_profile` block supports the following:
 
 -> **NOTE:** When `network_plugin` is set to `azure` - the `vnet_subnet_id` field in the `agent_pool_profile` block must be set.
 
+* `network_policy` - (Optional) Sets up network policy to be used with Azure CNI. [Network policy allows us to control the traffic flow between pods](https://docs.microsoft.com/en-us/azure/aks/use-network-policies). This field can only be set when `network_plugin` is set to `azure`. Currently only accepted value is `calico` which installs Calico daemonset. Changing this forces a new resource to be created.
```
katbyte:
```suggestion
* `network_policy` - (Optional) Sets up network policy to be used with Azure CNI. [Network policy allows us to control the traffic flow between pods](https://docs.microsoft.com/en-us/azure/aks/use-network-policies). This field can only be set when `network_plugin` is set to `azure`. Currently the only accepted value is `calico`, which installs Calico daemonset. Changing this forces a new resource to be created.
```

azurerm/resource_arm_kubernetes_cluster.go
ghostinthewires: Was thinking something like this might work:

"network_policy": {
							Type:     schema.TypeString,
							Optional: true,
							ForceNew: true,
								string(containerservice.Calico),
							}, false),
						},

katbyte: Could we not suppress the case difference? We are slowly moving towards making these sorts of fields case sensitive due to the odd case where it matters on the API side.

thatInfrastructureGuy:
@katbyte Thanks for reverting back. Yes. To enable network policy in your subscription please follow:

```
# To create an AKS with network policy, first enable a feature flag on your subscription. To register the EnableNetworkPolicy feature flag, use the az feature register command as shown in the following example:
az feature register --name EnableNetworkPolicy --namespace Microsoft.ContainerService

# It takes a few minutes for the status to show Registered. You can check on the registration status using the az feature list command:
az feature list -o table --query "[?contains(name, 'Microsoft.ContainerService/EnableNetworkPolicy')].{Name:name,State:properties.state}"

# When ready, refresh the registration of the Microsoft.ContainerService resource provider using the az provider register command:
az provider register --namespace Microsoft.ContainerService
```

Source : https://docs.microsoft.com/en-us/azure/aks/use-network-policies#before-you-begin

thatInfrastructureGuy:
> Also getting many of these:
```
* azurerm_virtual_network.test: 1 error occurred:
        	* azurerm_virtual_network.test: Error waiting for completion of Virtual Network "acctestvirtnet190305190900434025" (Resource Group "acctestRG-190305190900434025"): Code="InternalServerError" Message="An error occurred." Details=[]
```
I believe the Virtual Network errors might be related to above comment as VNET creation is not getting completed because of unknown parameter networkProfile.networkPolicy.

thatInfrastructureGuy:
@katbyte Is there anything I can assist with? Can you please guide me for the documentation of labels. Eg: What does `waiting-response` stand for?

Thanks for helping me out!

jfcoz:
@katbyte , is there something more to change ?

RichardFowles89:
Hi guys, do you have any idea when this will be available for general release? We want to implement it at work.

Thanks!

dannydombrowski:
Hey @katbyte it has been 15 days since your last activity on this issue. Could you please let us know if there is anything else left before merging this issue?

katbyte:
Sorry for the delay here @dannydombrowski, @RichardFowles89, @jfcoz & @thatInfrastructureGuy our attention was elsewhere.

I have verified that the tests now pass for us and this will get into `v1.24` 🙂

RichardFowles89:
That's great - thank you! I will keep my eye out.

RichardFowles89:
Great! Thank you for letting me know

soggychipsnz:
Has anyone had success with provisioning a network-policy enabled AKS cluster with the new provider? 

When attempting to provision an AKS cluster with this enabled, my deployment eventually fails with:
azurerm_kubernetes_cluster.aks_cluster: Error waiting for completion of Managed Kubernetes Cluster "REMOVED" (Resource Group REMOVED"): Code="NodesNotReady" Message=""

The same state deploys fine without the network_policy enabled. 
RBAC and the Azure network plugin are set, also using 1.24.0 of the azurerm provider.
The network_profile network policy is being set to the value of "calico"
If I diff the exported ARM template of the failed deployment with one that has been successfully provisioned via ARM template, nothing stands out as incorrect.

Any assistance on this one or confirmation that it is working for others would be appreciated.

thatInfrastructureGuy:
@soggychipsnz I am running into same issue. I redeployed (not restarted) the VMs and did a terraform apply again to fix it.

soggychipsnz:
Thanks for confirming, I was wondering if I was missing something obvious.
How does one redeploy individual nodes from a managed container service?  (assuming the nodes are the VMs you are referring to)

I have opened a case with MS to check out the failed instance as well just incase they have anything to add.

thatInfrastructureGuy:
Just navigate to `MC_${RESOURCE_GROUP}_${CLUSTER_NAME}` resource group. On each VM; click the redeploy button available in the portal.
Screen Shot 2019-04-08 at 10 58 07 AM

soggychipsnz:
Cheers for the feedback. I'll give that a go after I get closure from my case I have open with MS as they might be able to provide more understanding as to why this issue is happening.

thatInfrastructureGuy:
@soggychipsnz https://github.com/MicrosoftDocs/azure-docs/issues/28567

Seems like azure backend issue which is currently being fixed. No ETA.

feiskyer:
Yep, this is bug for v1.12.x. versions below that are still working

thatInfrastructureGuy:
@feiskyer Thanks for the update. Is there a recommended path forward for users? Eg: 
1. Using AKS 1.12 with [azure-npm](https://github.com/Azure/azure-container-networking/tree/master/npm)  // OR
2. AKS <= 1.11 with Calico.

feiskyer:
@thatInfrastructureGuy yep, both of those are working. it's up to users to decide which one is preferred.

Issue #1576, opened 2018-07-15
Bug Fixes/Enhancements to Application Gateway


tombuildsstuff:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

* `azurerm_application_gateway`

---

The `azurerm_application_gateway` resource is currently missing a selection of fields and also has some bugs which need resolving. Unfortunately there's [a bug in the Application Gateway API where the Application Gateway isn't actually deleted](https://github.com/Azure/azure-rest-api-specs/issues/2187) which prevents us from proceeding this work, since our tests for these resources are failing around 80% of the time due to the Application Gateway not being deleted (meaning this fills up our quota's).

Rather than trying to track the status of this bug across multiple issues - I'm opening this meta-issue to keep track of these bugs and enhancements in one place. Once the bug in the API is resolved - it should be possible to add these enhancements/investigate fixing these bugs; however these are blocked at the moment.

## Blocking API issues

- ~[The Application Gateway API returns that an Application Gateway has been deleted when it hasn't](https://github.com/Azure/azure-rest-api-specs/issues/2187) (which was previously tracked in this repository in https://github.com/terraform-providers/terraform-provider-azurerm/issues/608)~

## Enhancements

 - ~[Support for Disabled Rule Groups](https://github.com/terraform-providers/terraform-provider-azurerm/issues/451)~ (fixed in #3394)
 - ~[Support for Redirect Rules](https://github.com/terraform-providers/terraform-provider-azurerm/issues/552)~
 - ~[Support for Affinity cookie name](https://github.com/terraform-providers/terraform-provider-azurerm/issues/1519)~ (fixed in #3434)
 - ~[Support for Connection draining](https://github.com/terraform-providers/terraform-provider-azurerm/issues/1519)~ (fixed in #2778)
 - ~[Support for Hostname](https://github.com/terraform-providers/terraform-provider-azurerm/issues/1875)~ (fixed in #2990)
[...]

PR #2990, opened 2019-03-03
azurerm_application_gateway: Support for Hostname

mcharriere:
Added support for hostname in backend http settings. #1576

### DIFF
--- a/azurerm/resource_arm_application_gateway.go
+++ b/azurerm/resource_arm_application_gateway.go
@@ -920,6 +920,16 @@ func resourceArmApplicationGatewayCreateUpdate(d *schema.ResourceData, meta inte
 		},
 	}
 
+	for _, backendHttpSettings := range *backendHTTPSettingsCollection {
+		backendHttpSettingsProperties := *backendHttpSettings.ApplicationGatewayBackendHTTPSettingsPropertiesFormat
+		hostName := *backendHttpSettingsProperties.HostName
+		pick := *backendHttpSettingsProperties.PickHostNameFromBackendAddress
+
+		if hostName != "" && pick {
+			return fmt.Errorf("Only one of `host_name` or `pick_host_name_from_backend_address` can be set")
+		}
+	}
+
 	for _, probe := range *probes {
 		probeProperties := *probe.ApplicationGatewayProbePropertiesFormat
 		host := *probeProperties.Host
@@ -1263,10 +1273,6 @@ func expandApplicationGatewayBackendHTTPSettings(d *schema.ResourceData, gateway
 		hostName := v["host_name"].(string)
 		pickHostNameFromBackendAddress := v["pick_host_name_from_backend_address"].(bool)
 		requestTimeout := int32(v["request_timeout"].(int))
-
-		if hostName != "" && pickHostNameFromBackendAddress {
-			return fmt.Errorf("Only one of `host_name` or `pick_host_name_from_backend_http_settings` can be set")
-		}
 
 		setting := network.ApplicationGatewayBackendHTTPSettings{
 			Name: &name,
--- a/azurerm/resource_arm_application_gateway.go
+++ b/azurerm/resource_arm_application_gateway.go
@@ -146,6 +146,11 @@ func resourceArmApplicationGateway() *schema.Resource {
 							}, true),
 						},
 
+						"host_name": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+
 						"pick_host_name_from_backend_address": {
 							Type:     schema.TypeBool,
 							Optional: true,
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_application_gateway.go
```diff
@@ -146,6 +146,11 @@ func resourceArmApplicationGateway() *schema.Resource {
 							}, true),
 						},
 
+						"host_name": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
```
katbyte: Could we get some validation here? at the very least `validate.NoZeroValues`

azurerm/resource_arm_application_gateway.go
```diff
@@ -1263,6 +1279,7 @@ func expandApplicationGatewayBackendHTTPSettings(d *schema.ResourceData, gateway
 			ApplicationGatewayBackendHTTPSettingsPropertiesFormat: &network.ApplicationGatewayBackendHTTPSettingsPropertiesFormat{
 				CookieBasedAffinity:            network.ApplicationGatewayCookieBasedAffinity(cookieBasedAffinity),
 				Path:                           utils.String(path),
+				HostName:                       utils.String(hostName),
```
katbyte: Do we always want to set this? how does the API behave when pick hostname is true?

azurerm/resource_arm_application_gateway.go
```diff
@@ -1255,6 +1270,7 @@ func expandApplicationGatewayBackendHTTPSettings(d *schema.ResourceData, gateway
 		port := int32(v["port"].(int))
 		protocol := v["protocol"].(string)
 		cookieBasedAffinity := v["cookie_based_affinity"].(string)
+		hostName := v["host_name"].(string)
```
katbyte: this is an optional value so we should only be setting it when its not empty?

azurerm/resource_arm_application_gateway.go
```diff
@@ -915,6 +920,16 @@ func resourceArmApplicationGatewayCreateUpdate(d *schema.ResourceData, meta inte
 		},
 	}
 
+	for _, backendHttpSettings := range *backendHTTPSettingsCollection {
+		backendHttpSettingsProperties := *backendHttpSettings.ApplicationGatewayBackendHTTPSettingsPropertiesFormat
+		hostName := *backendHttpSettingsProperties.HostName
```
katbyte: Shouldn't we be nil checking this?

katbyte: oh i see its always set.

azurerm/resource_arm_application_gateway_test.go
```diff
@@ -260,6 +260,32 @@ func TestAccAzureRMApplicationGateway_probesPickHostNameFromBackendHTTPSettings(
 	})
 }
 
+func TestAccAzureRMApplicationGateway_backendHttpSettingsHostName(t *testing.T) {
```
katbyte: Could we get another test with 3 steps:
 - set pick true
 - set hostname
 - unset hostname

katbyte:
As suspected i think always setting the property causes issues:

```
------- Stdout: -------
=== RUN   TestAccAzureRMApplicationGateway_probesPickHostNameFromBackendHTTPSettings
=== PAUSE TestAccAzureRMApplicationGateway_probesPickHostNameFromBackendHTTPSettings
=== CONT  TestAccAzureRMApplicationGateway_probesPickHostNameFromBackendHTTPSettings
--- FAIL: TestAccAzureRMApplicationGateway_probesPickHostNameFromBackendHTTPSettings (107.04s)
        	* azurerm_application_gateway.test: 1 error occurred:
        	* azurerm_application_gateway.test: Error Creating/Updating Application Gateway "acctestag-190305192949044967" (Resource Group "acctestRG-190305192949044967"): network.ApplicationGatewaysClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="ApplicationGatewayBackendHttpSettingsHostNameFieldConflict" Message="HostName field may only be specified if pickHostNameFromBackendAddress is set to false in context '/subscriptions/c0a607b2-6372-4ef3-abdb-dbe52a7b56ba/resourceGroups/acctestRG-190305192949044967/providers/Microsoft.Network/applicationGateways/acctestag-190305192949044967/backendHttpSettingsCollection/acctest-vnet-190305192949044967-be-htst'." Details=[]
```

mcharriere:
Oh  @katbyte thanks for pointing out that. I'll fix it soon.

azurerm/resource_arm_application_gateway.go
metacpp: What if we'd like to `unset` this field through empty string ?

website/docs/r/application_gateway.html.markdown
```diff
@@ -204,6 +204,8 @@ A `backend_http_settings` block supports the following:
 
 * `request_timeout` - (Required) The request timeout in seconds, which must be between 1 and 86400 seconds.
 
+* `host_name` - (Optional) Host header to be sent to the backend servers. Cannot be set if `pick_host_name_from_backend_address` is set to `true`.
```
metacpp:
```suggestion
* `host_name` - (Optional) Host header to be sent to the backend servers. Should be skipped if `pick_host_name_from_backend_address` is set to `true`.
```

azurerm/resource_arm_application_gateway.go
mcharriere: No, I don't think so. My mistake. I'm sending a fix right now.

azurerm/resource_arm_application_gateway.go
mcharriere: I'm sending a fix for this too

azurerm/resource_arm_application_gateway.go
mcharriere: Since the value is not always set, should I stil add this validation? Which case would this cover?

website/docs/r/application_gateway.html.markdown
katbyte: I don't think we should make this change. `Cannot be set` is more inline with the existing documentation so is fine as is as well as is definitive in it should not be set.

azurerm/resource_arm_application_gateway.go
mcharriere: @metacpp I've tested setting and unsetting both parameters and it works well. This PR also includes an error checking for pick_host_name_from_backend_address == true and host_name set.

azurerm/resource_arm_application_gateway_test.go
mcharriere: I have a doubt regarding the test. If I set pick true and then I try to set hostname it will throw an error. Until here it's fine, i can check for that error. But I dont understand the following step. I think that unsetting hostname won't do anything because of the previous error.

Should I add a test just to check for that error (both pick and host_name set)?

azurerm/resource_arm_application_gateway.go
tombuildsstuff: +1 I think since this is Optional NoZeroValues doesn't make sense here

azurerm/resource_arm_application_gateway_test.go
mcharriere: @katbyte I've added the test TestAccAzureRMApplicationGateway_backendHttpSettingsHostNameAndPick checking for the error and it's passing.
Please let me know if you think there is something else to check for.

PR #2994, opened 2019-03-04
New Resource: `azurerm_api_management_logger`

JunyiYi:
A simple resource based on `azurerm_api_management` service. It manages the loggers inside that service.

**Several points to mention**
1. I use `Update()` instead of `CreateOrUpdate()` because `IsBuffered` could only be updated by `Update()`.
2. `LoggerType` is not updatable, thus I mark both `eventhub` and `application_insights` as `ForceNew`.
3. Properties under `eventhub` and `application_insights` are updatable through `Credentials`, however.

One of work items for #1177.


azurerm/resource_arm_api_management_logger_test.go
```diff
@@ -0,0 +1,431 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMApiManagementLogger_basicEventHub(ri, location),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMApiManagementLoggerExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "buffered", "true"),
+					resource.TestCheckResourceAttr(resourceName, "eventhub.#", "1"),
+					resource.TestCheckResourceAttrSet(resourceName, "eventhub.0.name"),
+					resource.TestCheckResourceAttrSet(resourceName, "eventhub.0.connection_string"),
+					resource.TestCheckNoResourceAttr(resourceName, "application_insights.#"),
```
tombuildsstuff: as this should be set to an empty array (above) we should be able to check there's 0 elements here:

```suggestion
					resource.TestCheckResourceAttr(resourceName, "application_insights.#", "0"),
```

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+
+* `name` - (Required) The name of this Logger, which must be unique within the API Management Service. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the Resource Group in which the API Management Service should be exist. Changing this forces a new resource to be created.
+
+* `api_management_name` - (Required) The name of the API Management Service. Changing this forces a new resource to be created.
+
+* `eventhub` - (Optional) An `eventhub` block is documented below.
+
+* `application_insights` - (Optional) An `application_insights` block is documented below.
```
tombuildsstuff:
```suggestion
* `application_insights` - (Optional) An `application_insights` block as documented below.
```

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) The name of this Logger, which must be unique within the API Management Service. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the Resource Group in which the API Management Service should be exist. Changing this forces a new resource to be created.
+
+* `api_management_name` - (Required) The name of the API Management Service. Changing this forces a new resource to be created.
+
+* `eventhub` - (Optional) An `eventhub` block is documented below.
```
tombuildsstuff:
```suggestion
* `eventhub` - (Optional) An `eventhub` block as documented below.
```

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+
+* `application_insights` - (Optional) An `application_insights` block is documented below.
+
+* `buffered` - (Optional) Indicates whether records are buffered in the Logger before publishing. Defaults to `true`.
+
+* `description` - (Optional) A description of this Logger.
```
tombuildsstuff: can we sort these Optional values alphabetically?

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+
+---
+
+An `eventhub` block supports the following:
+
+* `name` - (Required) The name of an `azurerm_eventhub`.
```
tombuildsstuff: This isn't necessarily managed by Terraform - so this needs to be:

```suggestion
* `name` - (Required) The name of an EventHub.
```

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+
+* `connection_string` - (Required) The connection string of an `azurerm_eventhub_namespace`.
```
tombuildsstuff: This isn't necessarily managed by Terraform - so this needs to be:

```suggestion
* `connection_string` - (Required) The connection string of an EventHub Namespace.
```

presumably this could also be the connection string scoped to this EventHub rather than to the EventHub Namespace?

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+
+---
+
+An `application_insights` block supports the following:
+
+* `instrumentation_key` - (Required) The instrumentation key of an `azurerm_application_insights`.
```
tombuildsstuff: This isn't necessarily managed by Terraform - so this needs to be:

```suggestion
* `instrumentation_key` - (Required) The instrumentation key used to push data to Application Insights.
```

website/docs/r/api_management_logger.html.markdown
JunyiYi: This should be the connection string to the namespace, you can specify the eventhub name in `eventhub.name` attribute.

azurerm/resource_arm_api_management_logger.go
```diff
@@ -0,0 +1,273 @@
...
+	resp, err := client.Get(ctx, resourceGroup, serviceName, name)
+	if err != nil {
+		return fmt.Errorf("Error retrieving Logger %q (Resource Group %q / API Management Service %q): %+v", name, resourceGroup, serviceName, err)
+	}
+	if resp.ID == nil {
+		return fmt.Errorf("Cannot read Logger %q (Resource Group %q / API Management Service %q) ID", name, resourceGroup, serviceName)
+	}
+	d.SetId(*resp.ID)
+
+	return resourceArmApiManagementLoggerRead(d, meta)
```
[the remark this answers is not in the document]
JunyiYi: But I do not think we need to update `buffered` in `Create` call.

azurerm/resource_arm_api_management_logger.go
```diff
@@ -0,0 +1,256 @@
...
+
+	d.Set("name", resp.Name)
+	d.Set("resource_group_name", resourceGroup)
+	d.Set("api_management_name", serviceName)
+
+	if properties := resp.LoggerContractProperties; properties != nil {
+		d.Set("buffered", properties.IsBuffered)
+		d.Set("description", properties.Description)
+		if err := d.Set("eventhub", flattenArmApiManagementLoggerEventHub(d, properties)); err != nil {
+			return fmt.Errorf("Error setting `eventhub` for Logger %q (Resource Group %q / API Management Name %q): %s", name, resourceGroup, serviceName, err)
```
tombuildsstuff: we can update this to:

```suggestion
			return fmt.Errorf("Error setting `eventhub`: %s", err)
```

azurerm/resource_arm_api_management_logger.go
```diff
@@ -0,0 +1,256 @@
...
+
+func expandArmApiManagementLoggerApplicationInsights(input []interface{}) map[string]*string {
+	credentials := make(map[string]*string)
+	ai := input[0].(map[string]interface{})
+	credentials["instrumentationKey"] = utils.String(ai["instrumentation_key"].(string))
+	return credentials
+}
+
+func flattenArmApiManagementLoggerEventHub(d *schema.ResourceData, properties *apimanagement.LoggerContractProperties) []interface{} {
+	if properties.LoggerType != apimanagement.AzureEventHub {
```
tombuildsstuff: we should probably set this in all circumstances; since Terraform's state should match the resource's state in Azure

azurerm/resource_arm_api_management_logger_test.go
```diff
@@ -0,0 +1,409 @@
...
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{"application_insights.0.instrumentation_key"},
+			},
+		},
+	})
+}
+
+func TestAccAzureRMApiManagementLogger_basicEventHubAppInsightsUpdate(t *testing.T) {
```
tombuildsstuff: we can make this:

```suggestion
func TestAccAzureRMApiManagementLogger_update(t *testing.T) {
```

azurerm/resource_arm_api_management_logger_test.go
```diff
@@ -0,0 +1,409 @@
...
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMApiManagementLoggerDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMApiManagementLogger_complete(ri, location, "Logger from Terraform test", "false"),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMApiManagementLoggerExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "description", "Logger from Terraform test"),
```
tombuildsstuff: since it's very similar, I think we could probably add a final step to the other `_update` test to add a description and then remove this test case?

azurerm/resource_arm_api_management_logger_test.go
```diff
@@ -0,0 +1,409 @@
...
+					resource.TestCheckResourceAttr(resourceName, "buffered", "false"),
+					resource.TestCheckResourceAttr(resourceName, "eventhub.#", "0"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccAzureRMApiManagementLogger_basicCompleteUpdate(t *testing.T) {
```
tombuildsstuff: (same here) I think we could roll this into the other `update` test?

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
+---
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_api_management_logger"
+sidebar_current: "docs-azurerm-resource-api-management-logger"
+description: |-
+  Manages an API Management Logger.
```
tombuildsstuff: for consistency with the other APIM resources could we make this:

```suggestion
  Manages a Logger within an API Management Service.
```

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+---
+
+# azurerm_api_management_logger
+
+Manages an API Management Logger.
```
tombuildsstuff: for consistency with the other APIM resources could we make this:

```suggestion
Manages a Logger within an API Management Service.
```

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+    name     = "Developer"
+    capacity = 1
+  }
+}
+
+resource "azurerm_api_management_logger" "example" {
+  name                = "example-logger"
+  api_management_name = "${azurerm_api_management.example.name}"
+  resource_group_name = "${azurerm_resource_group.example.name}"
+  description         = "This logger is created by Terraform."
```
tombuildsstuff: since `description` is optional could we remove it from this example? we've [got a TODO](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/examples/todo.txt#L9) to add some additional examples for API Management once the other resources have been added

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+```
+
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of this Logger, which must be unique within the API Management Service. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the Resource Group in which the API Management Service should be exist. Changing this forces a new resource to be created.
```
tombuildsstuff: since this is referring to the RG of an existing APIM Service - we can phrase this as:

```suggestion
* `resource_group_name` - (Required) The name of the Resource Group in which the API Management Service exists. Changing this forces a new resource to be created.
```

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+
+* `api_management_name` - (Required) The name of the API Management Service. Changing this forces a new resource to be created.
+
+* `application_insights` - (Optional) An `application_insights` block as documented below.
+
+* `buffered` - (Optional) Indicates whether records are buffered in the Logger before publishing. Defaults to `true`.
```
tombuildsstuff: we tend to phrase Boolean's with an explanation of what they do in an active sense, e.g.:

```suggestion
* `buffered` - (Optional) Specifies whether records should be buffered in the Logger prior to publishing. Defaults to `true`.
```

website/docs/r/api_management_logger.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+
+An `eventhub` block supports the following:
+
+* `name` - (Required) The name of an EventHub.
+
+* `connection_string` - (Required) The connection string of an EventHub Namespace.
+
+---
+
+An `application_insights` block supports the following:
```
tombuildsstuff: _minor_ could we move this block above the `eventhub` block so that it's also sorted alphabetically?

tombuildsstuff:
Tests pass:

Screenshot 2019-03-21 at 10 48 09

PR #2791, opened 2019-01-29
Add Connection Monitor resource and Network Watcher data source

lewinski:
This PR adds a new resource to manage [Connection Monitors](https://docs.microsoft.com/en-us/azure/network-watcher/network-watcher-monitoring-overview#monitoring), which allow network connectivity monitoring between VMs and an endpoint (another VM, IP, or FQDN) via the existing Network Watcher resources.

Because Network Watchers are limited to one per region per subscription, this PR also adds a Network Watcher data source. This may be useful when multiple Terraform configurations using Network Watchers, Connection Monitors, or Packet Captures are targeted at the same subscription.

PR #2997, opened 2019-03-04
Correct the changelog entry for #2791

lewinski:
I noticed the changelog entries for #2791 looked wrong so this corrects them.


katbyte:
Thats for this @lewinski ! LGTM

PR #3009, opened 2019-03-06
New resource:  `azurerm_iothub_shared_access_policy`

### LINKED ISSUE #2201: Feature Request: Set azurerm_iothub shared access policies
### Community Note

### Description

Hey!
It would be great to be able to set shared access policies for IoT Hubs.

* azurerm_iothub

This could potentially be configured like so:

```hcl
resource "azurerm_iothub_shared_access_policy" "sap" {
  name            = "sap"
  iothub_id       = "${azurerm_iothub.iothub.id}"
  registry_read   = true
  registry_write  = true
  service_connect = true
  device_connect  = true
}

```

maxbog:
also, add test and docs

(resolves #2201)


azurerm/resource_arm_iothub.go
```diff
@@ -114,12 +114,12 @@ func resourceArmIotHub() *schema.Resource {
 
 			"shared_access_policy": {
 				Type:     schema.TypeList,
-				Computed: true,
+				Optional: true,
```
tombuildsstuff: this is going to need to become:

```suggestion
				Type: schema.TypeSet,
				Optional: true,
				Computed: true,
```

This needs to remain `computed` since it's possible that users may only want to retrieve these and not set the values; otherwise this'll be a breaking change.

In addition - since these can now be defined in-line this'll need to be a `TypeSet`, which is a breaking change - but is required since these could be returned in any order at this point (and may differ from the ordering defined in a users config). Unfortunately this would also be a breaking change; as such I believe defining these in-line may not be the best way to expose these fields.

Instead I believe it would make sense to expose a Virtual Resource (ala the azurerm_virtual_Machine_data_disk_attachment` resource which looks up the VM and adds the Disk, then returns) as an independent resource - rather than defining these in-line. This would also allow these to be retrieved as a Data Source.

Unfortunately this would require the use of a Lock to ensure concurrent changes aren't made; however it should be possible to do so - as it appears there's no independent API available to manage Shared Access Policies within an IoTHub.

What do you think?

tombuildsstuff:
hey @maxbog 

Thanks for this PR :)

Taking a quick look through this I believe the Shared Access Policies may be better managed as a separate resource, rather than defined in-line (I've left a comment inline with more information) - WDYT?

Thanks!

maxbog:
hey @tombuildsstuff,
This was actually my original idea, but I didn't feel comfortable enough with golang and terraform internals to implement this from scratch. However, now that I have something similar I can look at, I think I'll manage to create a separate resource.

However, there is a matter of the default access policies that are created together with the IoTHub. The new resource could "take over" the default policies if a policy with the same name is defined in the .tf file.

For example, when we would have a .tf file as below, I would expect the `azurerm_iothub_shared_access_policy.iothubowner` to redefine the default iothubowner policy.
```hcl
resource "azurerm_iothub" "test" {
  resource_group_name = "${azurerm_resource_group.iothub_sap.name}"
  location = "${azurerm_resource_group.iothub_sap.location}"

  name = "test_iothub"

  sku {
    capacity = 1
    name = "S1"
    tier = "Standard"
  }
}

resource "azurerm_iothub_shared_access_policy" "iothubowner" {
  name = "iothubowner"
  registry_read = true
  registry_write = true
}
```

Also, I think that the user should have control over whether the default policies should be created, so I was thinking about a new `provision_default_shared_access_policies` property, with a default value of `true`, to not break backward compatibility. 

Taking the above example further, if the `azurerm_iothub.test.provision_default_shared_access_policies` property was set to `true`, I would expect to provision an IoTHub with a single Access Policy. 

However, if it were to `false`, I would like to have all the default policies provisioned and the iothubowner taken over by the `azurerm_iothub_shared_access_policy.iothubowner` resource.

Another thing is the existing `shared_access_policy` attribute. I would expect that after updating any of the `azurerm_iothub_shared_access_policy` resources, the `azurerm_iothub` resource should be "refreshed" with the latest changes, but I have no idea whether this is actually possible. Or, we could write a note in the documentation that the two cannot be reliably used together.

Could you point me in the right direction here? 

All in all, I think that the existing attribute should be deprecated in favour of a new data source and actually removed in v2.0. What do you think?

maxbog:
@tombuildsstuff: I created a new `azurerm_iothub_shared_access_policy` resource, reverted the changes in this PR and renamed it. Could you please review the code?

maxbog:
@tombuildsstuff I'm stumped why the build is failing :(
In the terraform repo, tagged 0.11.9, there clearly are .go files, but Travis complains that the directory does not contain any.

Here is the file that I am using: 
https://github.com/hashicorp/terraform/blob/v0.11.9/helper/customdiff/compose.go

Do you have any pointers to why the build is failing?

tombuildsstuff:
@maxbog sorry missed these pings, will take a look now

maxbog:
@tombuildsstuff Sorry for pestering, but did you have any time to review the PR? Is there anything I could do to make the review easier for you?

azurerm/resource_arm_iothub_shared_access_policy.go
```diff
@@ -0,0 +1,356 @@
...
+
+		return fmt.Errorf("Error loading IotHub Shared Access Policy %q (Resource Group %q): %+v", iothubName, resourceGroup, err)
+	}
+
+	iothub, err := client.Get(ctx, resourceGroup, iothubName)
+	if err != nil {
+		return fmt.Errorf("Error loading IotHub %q (Resource Group %q): %+v", iothubName, resourceGroup, err)
+	}
+
+	_ = d.Set("name", keyName)
```
katbyte: I don't think we need the _ =?  🤔 
```suggestion
d.Set("name", keyName)
```

azurerm/resource_arm_iothub_shared_access_policy_test.go
```diff
@@ -0,0 +1,234 @@
+package azurerm
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+	"regexp"
```
katbyte: Could we goimports this file?

azurerm/resource_arm_iothub_shared_access_policy_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+  tags {
+    "purpose" = "testing"
+  }
+}
+
+resource "azurerm_iothub_shared_access_policy" "test" {
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  iothub_name         = "${azurerm_iothub.test.name}"
+  name            = "acctest"
```
katbyte: Could we fix the alignment here?

azurerm/resource_arm_iothub_shared_access_policy.go
```diff
@@ -0,0 +1,356 @@
...
+			rights.deviceConnect = true
+		case "serviceconnect":
+			rights.serviceConnect = true
+		}
+	}
+
+	return rights
+}
+
+func resourceArmIotHubSharedAccessPolicyRead(d *schema.ResourceData, meta interface{}) error {
```
katbyte: Could we place this and delete after the CreateUpdate function?

azurerm/resource_arm_iothub_shared_access_policy.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+func validateRegistryWriteImpliesRegistryRead(d *schema.ResourceDiff, meta interface{}) error {
+	if d.Get("registry_write").(bool) && !d.Get("registry_read").(bool) {
+		return fmt.Errorf("registry_read key must be set to true when registry_write is set to true")
+	}
+	return nil
+
+}
+func validateAtLeastOneIsTrue(parameterNames ...string) schema.CustomizeDiffFunc {
```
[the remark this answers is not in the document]
maxbog: I'll go with the single CustomizeDiff function

azurerm/resource_arm_iothub_shared_access_policy.go
maxbog: fixed

azurerm/resource_arm_iothub_shared_access_policy_test.go
maxbog: done

azurerm/resource_arm_iothub_shared_access_policy_test.go
maxbog: done

azurerm/resource_arm_iothub_shared_access_policy.go
maxbog: done

go.mod
```diff
@@ -1,28 +1,23 @@
 module github.com/terraform-providers/terraform-provider-azurerm
 
 require (
-	cloud.google.com/go v0.34.0 // indirect
```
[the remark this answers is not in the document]
maxbog: I reverted the changes after making one CustomizeDiff function instead of using customizeDiff.all

azurerm/resource_arm_iothub_shared_access_policy.go
```diff
@@ -0,0 +1,356 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"iothub_name": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: suppress.CaseDifference,
+				ValidateFunc:     validate.IoTHubName,
+			},
+			"resource_group_name": resourceGroupNameSchema(),
```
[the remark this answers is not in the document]
maxbog: done

azurerm/resource_arm_iothub_shared_access_policy.go
```diff
@@ -0,0 +1,356 @@
...
+			"name": {
```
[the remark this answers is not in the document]
maxbog: done

azurerm/resource_arm_iothub_shared_access_policy.go
```diff
@@ -0,0 +1,356 @@
...
+				Sensitive: true,
+				Computed:  true,
+			},
+			"secondary_connection_string": {
+				Type:      schema.TypeString,
+				Sensitive: true,
+				Computed:  true,
+			},
+		},
+		CustomizeDiff: customdiff.All(
```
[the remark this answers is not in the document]
maxbog: Great idea! done

website/docs/r/iothub_shared_access_policy.html.markdown
```diff
@@ -0,0 +1,86 @@
...
+---
+
+# azurerm_iothub_shared_access_policy
+
+Manages an IotHub Shared Access Policy
+
+## Example Usage
+
+```hcl
+resource "azurerm_resource_group" "test" {
```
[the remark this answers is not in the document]
maxbog: sure, done

maxbog:
@katbyte Thanks for the review! I addressed all the issues

azurerm/resource_arm_iothub_shared_access_policy.go
```diff
@@ -0,0 +1,354 @@
...
+					"The shared access policy key name must not be empty, and must not exceed 64 characters in length.  The shared access policy key name can only contain alphanumeric characters, exclamation marks, periods, underscores and hyphens."),
+			},
+
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"iothub_name": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: suppress.CaseDifference,
```
katbyte: What is the reasoning behind suppressing the case here?

azurerm/resource_arm_iothub_shared_access_policy.go
```diff
@@ -0,0 +1,354 @@
...
+	if err != nil {
+		return fmt.Errorf("Error loading IotHub %q (Resource Group %q): %+v", iothubName, resourceGroup, err)
+	}
+
+	d.Set("name", keyName)
+	d.Set("iothub_name", iothubName)
+	d.Set("resource_group_name", resourceGroup)
+
+	d.Set("primary_key", accessPolicy.PrimaryKey)
+	d.Set("primary_connection_string", getSharedAccessPolicyConnectionString(*iothub.Properties.HostName, keyName, *accessPolicy.PrimaryKey))
```
katbyte: So we don't check for an error with primitives, but we do for complex objects (confusing right)

So could we check this here?
```suggestion
	if err := d.Set("primary_connection_string", getSharedAccessPolicyConnectionString(*iothub.Properties.HostName, keyName, *accessPolicy.PrimaryKey)), err != nil {
		return fmt.ErrorF("error setting `primary_connection_string `: %v")
	}
```

sorry i didn't point this out last review

azurerm/resource_arm_iothub_shared_access_policy.go
```diff
@@ -0,0 +1,354 @@
...
+	d.Set("secondary_key", accessPolicy.SecondaryKey)
+	d.Set("secondary_connection_string", getSharedAccessPolicyConnectionString(*iothub.Properties.HostName, keyName, *accessPolicy.SecondaryKey))
```
katbyte: And here

website/docs/r/iothub_shared_access_policy.html.markdown
```diff
@@ -0,0 +1,86 @@
...
+    capacity = "1"
+  }
+}
+
+resource "azurerm_iothub_shared_access_policy" "example" {
+  name                = "test"
+  resource_group_name = "${azurerm_resource_group.example.name}"
+  iothub_name         = "${azurerm_iothub.example.name}"
+  
+  registry_read = true
```
katbyte:
```suggestion
  registry_read  = true
```

azurerm/resource_arm_iothub_shared_access_policy.go
maxbog: IOTHub name is case-insensitive, that's why the case is suppressed here. I just looked at the azurerm_iothub resource and saw that the case is not suppresed there. This may be dangerous, as simply changing the case of the name, which should not have any effect, as it is still the same resource, will recreate the IoTHub and all the data stored in it will be lost :( 
However, I can remove the suppress func if you have some kind of convention to always have the IDs case-sensitive.

website/docs/r/iothub_shared_access_policy.html.markdown
maxbog: done

azurerm/resource_arm_iothub_shared_access_policy.go
maxbog: done

azurerm/resource_arm_iothub_shared_access_policy.go
maxbog: done

azurerm/resource_arm_iothub_shared_access_policy.go
katbyte: @maxbog, while you are correct that this api seems to be case insensitive, others are not and we are trying to be more consistent across the provider. Given the existing resources are and a plan will notify the user of deletion if they change the case we would like the supress removed. 

Additionally, if casing is "iothub" and a user wishes it to be "IOThub" the supress will prevent this.

Thanks!

maxbog: OK then, I removed the suppress.

maxbog:
BTW this solves #2201

azurerm/resource_arm_iothub_shared_access_policy.go
katbyte: Thank you 🙏

PR #3037, opened 2019-03-12
Add Monitoring Metrics Publisher role assignment to Kubernetes cluster for Azure Monitor

jeffreyCline:
Enabling the ability to see custom Metrics in Azure Monitor in addition to the Metrics and logs in Azure Log Analytics Workspace. This functionality has been enabled in PowerShell and the Azure CLI tools, the service team requested Terraform also support this functionality. Per the service team, since the subscription provisioning the Kubernetes cluster will be required to have the `Microsoft.Authorization/roleAssignments/write` permission to grant the `Monitoring Metrics Publisher` role assignment, they are requesting that this not be a breaking failure. If the role assignment fails allow the deployment of the Kubernetes cluster should continue and silently fail the role assignment.

### DIFF
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -724,7 +724,7 @@ func resourceArmKubernetesClusterDelete(d *schema.ResourceData, meta interface{}
 	return nil
 }
 
-func azureMonitorRoleAssignment(d *schema.ResourceData, meta interface{}) error {
+func azureMonitorRoleAssignment(d *schema.ResourceData, meta interface{}) {
 	servicePrincipalProfile := expandAzureRmKubernetesClusterServicePrincipal(d)
 
 	// Get the root roleDefinitionID full path (i.e. providers/Microsoft.Authorization/roleDefinitions/3913510d-42f4-4e42-8a64-420c390055eb)
@@ -732,7 +732,7 @@ func azureMonitorRoleAssignment(d *schema.ResourceData, meta interface{}) error 
 
 	if rootRoleDefinitionID == "" {
 		log.Printf("[INFO] Unable to locate 'Monitoring Metrics Publisher' Root Role Definition Id\n")
-		return nil
+		return
 	}
 
 	// Set the scope to the AKS cluster (i.e. subscriptions/<subscription id>/resourcegroups/<resource group name>/providers/Microsoft.ContainerService/managedClusters/<resource name>)
@@ -751,7 +751,7 @@ func azureMonitorRoleAssignment(d *schema.ResourceData, meta interface{}) error 
 
 	if clientServicePrincipalObjectId == "" {
 		log.Printf("[DEBUG] Unable to lookup Service Principal Object ID\n")
-		return nil
+		return
 	}
 
 	properties := authorization.RoleAssignmentCreateParameters{
@@ -765,8 +765,6 @@ func azureMonitorRoleAssignment(d *schema.ResourceData, meta interface{}) error 
 	if err := resource.Retry(300*time.Second, retryAzureMonitorRoleAssignment(scope, name, properties, meta)); err != nil {
 		log.Printf("[DEBUG] Error adding Role Assignment to AKS cluster: %+v\n", err)
 	}
-
-	return nil
 }
 
 func getServicePrincipalObjectID(applicationId string, meta interface{}) string {
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -596,10 +596,8 @@ func resourceArmKubernetesClusterCreateUpdate(d *schema.ResourceData, meta inter
 	d.SetId(*read.ID)
 
 	if addonProfiles != nil && *addonProfiles["omsagent"].Enabled {
-		log.Printf("[INFO] adding role assignment for visualization of custom metrics in Azure Monitor.")
-		if err := azureMonitorRoleAssignment(d, meta); err != nil {
-			log.Printf("[INFO] Error adding Azure Monitor Role Assignment: %+v", err)
-		}
+		log.Printf("[INFO] adding role assignment for visualization of custom metrics in Azure Monitor.\n")
+		azureMonitorRoleAssignment(d, meta)
 	}
 
 	return resourceArmKubernetesClusterRead(d, meta)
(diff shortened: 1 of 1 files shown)


azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -715,6 +724,117 @@ func resourceArmKubernetesClusterDelete(d *schema.ResourceData, meta interface{}
...
+	properties := authorization.RoleAssignmentCreateParameters{
+		RoleAssignmentProperties: &authorization.RoleAssignmentProperties{
+			// Monitoring Metrics Publisher Role
+			RoleDefinitionID: utils.String(roleDefinitionID),
+			PrincipalID:      utils.String(clientServicePrincipalObjectId),
+		},
+	}
+
+	if err := resource.Retry(300*time.Second, retryAzureMonitorRoleAssignment(scope, name, properties, meta)); err != nil {
+		log.Printf("[DEBUG] Error adding Role Assignment to AKS cluster: %+v\n", err)
```
katbyte: Shouldn't we error out and fail the deployment?

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -715,6 +724,117 @@ func resourceArmKubernetesClusterDelete(d *schema.ResourceData, meta interface{}
...
+	ctx := meta.(*ArmClient).StopContext
+
+	var servicePrincipal *graphrbac.ServicePrincipal
+
+	filter := fmt.Sprintf("appId eq '%s'", applicationId)
+	log.Printf("[DEBUG] Looking up Service Principal objectID Using filter %q\n", filter)
+
+	apps, err := client.ListComplete(ctx, filter)
+	if err != nil {
+		log.Printf("[DEBUG] Error listing Service Principals: %+v\n", err)
```
katbyte: Shouldn't we error out and fail the deployment?

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -715,6 +724,117 @@ func resourceArmKubernetesClusterDelete(d *schema.ResourceData, meta interface{}
...
+	// Get the root roleDefinitionID full path (i.e. providers/Microsoft.Authorization/roleDefinitions/3913510d-42f4-4e42-8a64-420c390055eb)
+	rootRoleDefinitionID := getRootRoleDefinitionID(meta)
+
+	if rootRoleDefinitionID == "" {
+		log.Printf("[INFO] Unable to locate 'Monitoring Metrics Publisher' Root Role Definition Id\n")
+		return
+	}
+
+	// Set the scope to the AKS cluster (i.e. subscriptions/<subscription id>/resourcegroups/<resource group name>/providers/Microsoft.ContainerService/managedClusters/<resource name>)
+	scope := strings.TrimPrefix(d.Id(), "/")
```
ganga1980: Jeffrey, scope should be fully qualified resource id that includes "/". what is the reason to trim the "/" prefix?

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -715,6 +724,117 @@ func resourceArmKubernetesClusterDelete(d *schema.ResourceData, meta interface{}
...
+	log.Printf("[INFO] Role Assignment Scope is: %s\n", scope)
+
+	// UUID Name of the Role Assignment in this scope
```
ganga1980: probable we can add comment saying this UUID for "Monitoring Metrics Pubisher" role or move to constants file if there is anything like that in TF.

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -715,6 +724,117 @@ func resourceArmKubernetesClusterDelete(d *schema.ResourceData, meta interface{}
...
+	name := "3913510d-42f4-4e42-8a64-420c390055eb"
+
+	// Build the roleDefinitionID for this subscription (i.e. subscriptions/<subscription id>/providers/Microsoft.Authorization/roleDefinitions/<Role Assignment Name>)
```
ganga1980: roleDefinitionID resource Id should have the prefix "/" as well.  without the "/" prefix, its going to fail unless TF provider adds the prefix back before sending the ARM REST API to the azure endpoint.

azurerm/resource_arm_kubernetes_cluster.go
ganga1980: > Shouldn't we error out and fail the deployment?

Role assignment requires 'Microsoft.Authorization/roleAssignments/write' permission and this permission only granted to owner and administrator built-in roles by default.  From monitoring side, we want this role assignment optional hence we shouldn't fail the deployment. Without this role assignment, they get other monitoring data.

azurerm/resource_arm_kubernetes_cluster.go
jeffreyCline: because the REST API adds the leading `/`, if I don't trim the prefix I end up with a scope that starts like `//subscriptions/`

ganga1980: > because the REST API adds the leading `/`, if I don't trim the prefix I end up with a scope that starts like `//subscriptions/`

If this is the case, then we are good on this.

azurerm/resource_arm_kubernetes_cluster.go
jeffreyCline: Yes, the `/` prefix is added at the time of the API call itself...

azurerm/resource_arm_kubernetes_cluster.go
katbyte: However silently failing does not provide any feedback to the user expecting the role to be set. it becomes non obvious as to why metrics are not being published.

I think a new property `assign_monitor_role` separating out the option so that if it does fail the user gets an error messaging suggestion actions they can take an if its not something they want it cant just be unset it and continue with their deployment.

jeffreyCline:
I was able to confirm that the Role Assignment to the AKS Cluster is deleted when the AKS Cluster itself is destroyed...

```
PS C:\WINDOWS\system32> Get-AzureRmRoleAssignment -Scope "/subscriptions/...-67a141095a76"

RoleAssignmentId   : /subscriptions/...-67a141095a76/resourcegroups/aks-rg/providers/Microsoft.ContainerService/managedClusters/aks1/providers/Microsoft.Authorization/roleAssignments/3913510d-42f4-4e42-8a64-420c390055eb
Scope              : /subscriptions/...-67a141095a76/resourcegroups/aks-rg/providers/Microsoft.ContainerService/managedClusters/aks1
DisplayName        : terraform-testing-new
SignInName         :
RoleDefinitionName : Monitoring Metrics Publisher
RoleDefinitionId   : 3913510d-42f4-4e42-8a64-420c390055eb
ObjectId           : ...-3e31a3c3ec37
ObjectType         : ServicePrincipal
CanDelegate        : False

PS C:\WINDOWS\system32> Get-AzureRmRoleAssignment -Scope "/subscriptions/...-67a141095a76"

<Role was deleted when the resource was deleted.>
```

PR #3073, opened 2019-03-19
New Resource: `azurerm_api_management_api_version_set`

metacpp:
This PR is to add `azurerm_api_management_api_version_set` resource to better support `azurerm_api_management_api` resource.


website/docs/r/api_management_api_version_set.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_api_management_version_set"
+sidebar_current: "docs-azurerm-resource-api-management-api-version-set"
+description: |-
+  Manages an API Management API Version Set.
+---
+
+# azurerm_api_management_version_set
+
+Manages an API Management API Version Set.
```
tombuildsstuff:
```suggestion
Manages an API Version Set within a API Management Service.
```

website/docs/r/api_management_api_version_set.html.markdown
```diff
@@ -0,0 +1,78 @@
+---
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_api_management_version_set"
+sidebar_current: "docs-azurerm-resource-api-management-api-version-set"
+description: |-
+  Manages an API Management API Version Set.
```
tombuildsstuff: to match the other resources this can become:

```suggestion
  Manages an API Version Set within a API Management Service.
```

azurerm/resource_arm_api_management_api_version_set.go
```diff
@@ -0,0 +1,218 @@
...
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+
+			"display_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+
+			"versioning_schema": {
```
JunyiYi: It seems that we can infer this value from HCL. For example, if user provides `header_name`, it would be `Header`. In this way, we do not need to do `CustomizeDiff` check, and the user will not duplicate the effort to match those attributes.

tombuildsstuff: that'd be true if `header` and `query` were the only two options - but `segment` being the third option (without a separate field) makes this hard (not at this time, since we can set this if both header & query are set) in the future if a fourth value is added, as such I think it's worth keeping this for the moment?

azurerm/resource_arm_api_management_api_version_set.go
```diff
@@ -0,0 +1,218 @@
...
+	return &schema.Resource{
+		Create: resourceArmApiManagementApiVersionSetCreateUpdate,
+		Read:   resourceArmApiManagementApiVersionSetRead,
+		Update: resourceArmApiManagementApiVersionSetCreateUpdate,
+		Delete: resourceArmApiManagementApiVersionSetDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		CustomizeDiff: func(diff *schema.ResourceDiff, v interface{}) error {
```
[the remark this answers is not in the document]
metacpp: While I can move it to CreateOrUpdate method, but I still don't quite understand the principle here. Do you know why the `CustomDiff` in AKS a is different than your suggestion here ?

metacpp:
The PR has been updated.

azurerm/resource_arm_api_management_api_version_set.go
tombuildsstuff: this is `scheme` and not `schema` in the SDK - will push a commit to fix

tombuildsstuff:
Tests pass with the updated code:

Screenshot 2019-04-03 at 15 33 28

azurerm/resource_arm_api_management_api_version_set.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+
+	var vHeaderName, vQueryName *string
+	if v, ok := d.GetOk("version_header_name"); ok {
+		vHeaderName = utils.String(v.(string))
+	}
+	if v, ok := d.GetOk("version_query_name"); ok {
+		vQueryName = utils.String(v.(string))
+	}
```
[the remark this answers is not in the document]
metacpp: You changed the behavior from `nil` pointer to pointer to empty string.

PR #3080, opened 2019-03-19
provision_after_extensions property for azurerm_virtual_machine_scale_set extension

maxbog:
This property allows sequencing provisioning of VMSS extensions.
Also, some documentation.


katbyte:
Hi @maxbog,

Thank you for the PR, however as there is an existing PR that predates this one i am going to close it in favour of #2937 (and give it a review). Neither are in a meragable state as a state migration function is required due to the changing of a hash.

PR #3090, opened 2019-03-20
Add d.IsNewResource() to require imports where missing

katbyte:


azurerm/resource_arm_virtual_machine_data_disk_attachment.go
```diff
@@ -142,7 +142,7 @@ func resourceArmVirtualMachineDataDiskAttachmentCreateUpdate(d *schema.ResourceD
 	}
 
 	if d.IsNewResource() {
-		if requireResourcesToBeImported {
+		if requireResourcesToBeImported && d.IsNewResource() {
```
Lucretius: It looks like `d.IsNewResource()` is called in the line above.  Otherwise, LGTM.

katbyte: thanks! good catch

PR #3092, opened 2019-03-20
[AKS] make service principal optional

Lucretius:
This partially addresses https://github.com/terraform-providers/terraform-provider-azurerm/issues/2931

I will try to see what needs to be done for the SSH key piece in another PR.

The service principal is just made optional.  It looks like the handling of a nil service principal is the same as the already optional `linux_profile`, so it appeared all I needed to do was simply make it optional instead of required.

### DIFF
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -170,7 +170,7 @@ func resourceArmKubernetesCluster() *schema.Resource {
 			// TODO: 2.0 - we should be able to make this a List to be able to detect changes in the Client Secret
 			"service_principal": {
 				Type:     schema.TypeSet,
-				Required: true,
+				Optional: true,
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
--- a/azurerm/data_source_kubernetes_cluster_test.go
+++ b/azurerm/data_source_kubernetes_cluster_test.go
@@ -295,7 +295,7 @@ func TestAccDataSourceAzureRMKubernetesCluster_addOnProfileRouting(t *testing.T)
 }
 
 func testAccDataSourceAzureRMKubernetesCluster_basic(rInt int, clientId string, clientSecret string, location string) string {
-	r := testAccAzureRMKubernetesCluster_basic(rInt, clientId, clientSecret, location)
+	r := testAccAzureRMKubernetesCluster_basic(rInt, location)
 	return fmt.Sprintf(`
 %s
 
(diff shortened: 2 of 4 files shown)


azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -170,7 +170,7 @@ func resourceArmKubernetesCluster() *schema.Resource {
 			// TODO: 2.0 - we should be able to make this a List to be able to detect changes in the Client Secret
 			"service_principal": {
 				Type:     schema.TypeSet,
-				Required: true,
+				Optional: true,
 				MaxItems: 1,
```
katbyte: Should we mark this as computed so we can get the auto generated SP ID? (if possible)

Lucretius:
Let me check - I think there is another test which sets a service principal so I did not bother adding a new one.  I was wondering about setting that computed property - I can certainly add that in - does it matter if someone defines the service principal block?  It will only be computed in the case it’s not defined.

katbyte:
Oh apologies, if there was another test setting it no need for another

katbyte:
We already have a few instances where there are optional computed sets/lists like this, so at the moment it is fine and really the only way to do achieve it.

katbyte:
So running the tests here i am getting:

```

------- Stdout: -------
=== RUN   TestAccAzureRMKubernetesCluster_addAgent
=== PAUSE TestAccAzureRMKubernetesCluster_addAgent
=== CONT  TestAccAzureRMKubernetesCluster_addAgent
--- FAIL: TestAccAzureRMKubernetesCluster_addAgent (66.28s)
        	* azurerm_kubernetes_cluster.test: 1 error occurred:
        	* azurerm_kubernetes_cluster.test: Error creating/updating Managed Kubernetes Cluster "acctestaks190321035335400399" (Resource Group "acctestRG-190321035335400399"): containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="Required parameter servicePrincipalProfile is missing (null)." Target="servicePrincipalProfile"  
```

is there anything we need to do on our end to enable this behaviour?

Lucretius:
Interesting - should be nothing to do on your end to configure it to work.  I wonder why the automated build did not pick that up.  I will attempt to make an aks cluster with a local binary when I get back home and see what is going on.

Lucretius:
So, after some digging it appears I may have jumped the gun on this.  There appears to be an inconsistency between the REST API, the validation in the GO SDK, and the Powershell CLI.

The REST API [here](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/containerservice/resource-manager/Microsoft.ContainerService/stable/2018-03-31/managedClusters.json#L567) seems to indicate that the service profile is required.  The Go SDK [here](https://github.com/Azure/azure-sdk-for-go/blob/master/services/containerservice/mgmt/2017-08-31/containerservice/managedclusters.go#L70) seems to indicate that there is no rule to enforce (look at Linux Profile above with the same configuration and which is not required).  The powershell CLI exposes [functionality to automatically create a service principal with a kubernetes cluster](https://docs.microsoft.com/en-us/azure/aks/kubernetes-service-principal#automatically-create-and-use-a-service-principal).  I am going to close this PR and just open some issue upstream.

PR #3093, opened 2019-03-21
[Application Gateway] Add WAF properties for request body checking

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/3088 by including `request_body_check` and `max_request_body_size_kb` properties.  WAF test within the application gateway test suite has been updated to verify the new configuration values.  Documentation has been updated to align with [the WAF documentation](https://docs.microsoft.com/en-us/azure/application-gateway/application-gateway-waf-configuration#waf-request-size-limits).

### DIFF
--- a/azurerm/resource_arm_application_gateway.go
+++ b/azurerm/resource_arm_application_gateway.go
@@ -811,6 +811,17 @@ func resourceArmApplicationGateway() *schema.Resource {
 							ValidateFunc: validation.IntBetween(1, 500),
 							Default:      100,
 						},
+						"request_body_check": {
+							Type:     schema.TypeBool,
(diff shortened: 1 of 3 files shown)


Lucretius:
> 
> 
> This LGTM however I an curious as to where you got the default for `max_request_body_size_kb`? does the portal default to 128?

The WAF documentation I linked in the PR is where I got the default value from:

> The maximum request body size field is specified in KBs and controls overall request size limit excluding any file uploads. This field can range from 1-KB minimum to 128-KB maximum value. The default value for request body size is 128 KB.

katbyte:
Awesome! thanks

PR #3100, opened 2019-03-21
[Image] zone resilient property addition

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/1581

Updated documentation.  Note about where zone resiliency can be used comes from the [Azure documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/capture-image-resource#create-an-image-of-a-vm-using-powershell)

Added a property to an existing test function which specifies the storage type so that it could be reused for the `zone_redundant` test.  Added a possible skip to the test, as not all zones support ZRS and did not want false positives interfering with the test.

### DIFF
--- a/azurerm/data_source_image.go
+++ b/azurerm/data_source_image.go
@@ -196,6 +196,8 @@ func dataSourceArmImageRead(d *schema.ResourceData, meta interface{}) error {
 				return fmt.Errorf("[DEBUG] Error setting AzureRM Image Data Disks error: %+v", err)
 			}
 		}
+
+		d.Set("zone_resilient", profile.ZoneResilient)
 	}
 
 	flattenAndSetTags(d, img.Tags)
--- a/azurerm/data_source_image.go
+++ b/azurerm/data_source_image.go
@@ -40,6 +40,11 @@ func dataSourceArmImage() *schema.Resource {
 
 			"location": locationForDataSourceSchema(),
 
+			"zone_resilient": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+
 			"os_disk": {
 				Type:     schema.TypeList,
 				Computed: true,
--- a/azurerm/resource_arm_image.go
+++ b/azurerm/resource_arm_image.go
@@ -298,12 +298,12 @@ func resourceArmImageRead(d *schema.ResourceData, meta interface{}) error {
 				return fmt.Errorf("[DEBUG] Error setting AzureRM Image OS Disk error: %+v", err)
 			}
 		}
-
 		if disks := resp.StorageProfile.DataDisks; disks != nil {
 			if err := d.Set("data_disk", flattenAzureRmImageDataDisks(disks)); err != nil {
 				return fmt.Errorf("[DEBUG] Error setting AzureRM Image Data Disks error: %+v", err)
 			}
 		}
+		d.Set("zone_resilient", resp.StorageProfile.ZoneResilient)
 	}
 
 	flattenAndSetTags(d, resp.Tags)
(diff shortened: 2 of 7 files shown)


Lucretius:
This is not yet ready to merge.  I am having issues running the integration tests seeing the following:
```
azurerm_virtual_machine.testsource: Code="ImageNotFound" Message="The platform image 'Canonical:UbuntuServer:16.04-LTS:latest' is not available. Verify that all fields in the storage profile are correct."
```

Seems to be something local to my branch, I am working on it.

azurerm/resource_arm_image_test.go
```diff
@@ -1388,3 +1433,14 @@ resource "azurerm_virtual_machine_scale_set" "testdestination" {
 }
 `, rInt, location, rInt, rInt, rInt, hostName, rInt, rInt, userName, password, rInt, userName, password, rInt)
 }
+
+// Region list pulled from https://docs.microsoft.com/en-us/azure/storage/common/storage-redundancy-zrs
+func supportsZRS(location string) bool {
+	zrsRegions := []string{"westus2", "francecentral", "southeastasia", "westeurope", "northeurope", "japaneast", "uksouth", "eastus", "eastus2", "centralus"}
```
tombuildsstuff: this only covers Azure Public (there's different Azure Regions which are supported e.g. Government, China, Germany) - as such I think we'd be best to remove this for the moment? (we're running the tests in some of these regions)

azurerm/data_source_image.go
```diff
@@ -40,6 +40,11 @@ func dataSourceArmImage() *schema.Resource {
 
 			"location": locationForDataSourceSchema(),
 
+			"zone_resilient": {
```
tombuildsstuff: could we set this field in the state here: https://github.com/terraform-providers/terraform-provider-azurerm/pull/3100/files#diff-21521ec4fe0d38353d87860312cd25a4R187

azurerm/resource_arm_image.go
```diff
@@ -35,6 +35,13 @@ func resourceArmImage() *schema.Resource {
 
 			"resource_group_name": resourceGroupNameSchema(),
 
+			"zone_resilient": {
```
tombuildsstuff: could we also Set this in the state here using the object returned from the API? https://github.com/terraform-providers/terraform-provider-azurerm/pull/3100/files#diff-95b4abb1c5040a8fd4f3c3674821b7b6R292

website/docs/r/image.html.markdown
```diff
@@ -58,11 +58,14 @@ The following arguments are supported:
     the image. Changing this forces a new resource to be created.
 * `location` - (Required) Specified the supported Azure location where the resource exists.
     Changing this forces a new resource to be created.
+* `zone_resilient` - (Optional) Is zone resiliency enabled?  Defaults to `false`.  Changing this forces a new resource to be created.
```
tombuildsstuff: we're trying to keep these sorted Required -> Optional and then Alphabetically within the fields; could we move this to below Tags (and above the `Note`?)

azurerm/resource_arm_image_test.go
Lucretius: Sure, my concern had been that I did not know in which locations these tests were going to be run, and did not want to have an acceptance test failure simply due to the fact that ZRS storage is not supported in some locations.

azurerm/data_source_image.go
Lucretius: Good catch thanks!

azurerm/resource_arm_image_test.go
```diff
@@ -601,6 +641,7 @@ resource "azurerm_image" "test" {
   name                = "accteste"
   location            = "${azurerm_resource_group.test.location}"
   resource_group_name = "${azurerm_resource_group.test.name}"
+  zone_redundant      = %t
```
katbyte: Should this be `zone_resilient`?
```suggestion
  zone_resilient      = %t
```

Lucretius:
Ah nice catch @katbyte - I think I got mixed up looking at another resource which had the `zone_redundant` property

PR #3103, opened 2019-03-22
New Resource: `azurerm_api_management_subscription`

tombuildsstuff:
This PR adds support for API Management Subscriptions, which are a part of #1177

Of particular note here is that the User ID and Product ID need to be fully qualified Resource ID's (which is what the API returns); however the API allows a short-hand too at creation time (`/products/0000`) - but there's validation to confirm the ID is a Resource ID since this is what's actually returned from the API.

### DIFF
--- a/azurerm/resource_arm_api_management_subscription_test.go
+++ b/azurerm/resource_arm_api_management_subscription_test.go
@@ -163,8 +163,6 @@ func testCheckAzureRMAPIManagementSubscriptionDestroy(s *terraform.State) error 
 				return err
 			}
 		}
-
-		return nil
 	}
 	return nil
 }
--- a/azurerm/resource_arm_api_management_subscription_test.go
+++ b/azurerm/resource_arm_api_management_subscription_test.go
@@ -1,0 +1,298 @@
+package azurerm
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func TestAccAzureRMAPIManagementSubscription_basic(t *testing.T) {
+	resourceName := "azurerm_api_management_subscription.test"
+	ri := tf.AccRandTimeInt()
+	location := testLocation()
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMAPIManagementSubscriptionDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMAPIManagementSubscription_basic(ri, location),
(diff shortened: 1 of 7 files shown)


tombuildsstuff:
Tests pass:

Screenshot 2019-03-22 at 11 18 54

azurerm/resource_arm_api_management_subscription_test.go
```diff
@@ -0,0 +1,298 @@
...
+		ctx := testAccProvider.Meta().(*ArmClient).StopContext
+		resp, err := client.Get(ctx, resourceGroup, serviceName, subscriptionId)
+
+		if err != nil {
+			if !utils.ResponseWasNotFound(resp.Response) {
+				return err
+			}
+		}
+
+		return nil
```
bflad: Should this be removed so it loops through all `azurerm_api_management_subscription` resources in the state to verify they have been deleted? I really wish we had a provider linter to catch these 😄

website/docs/r/api_management_subscription.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+
+# azurerm_api_management_subscription
+
+Manages a Subscription within a API Management Service.
+
+
+## Example Usage
+
+```hcl
+XXXX
```
bflad: Needs to be filled in 😄

website/docs/r/api_management_subscription.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - The ID of the API Management Subscription.
+
+## Import
+
+API Management Properties can be imported using the `resource id`, e.g.
```
bflad: Properties or Subscriptions?

```suggestion
API Management Subscriptions can be imported using the `resource id`, e.g.
```

azurerm/resource_arm_api_management_subscription_test.go
tombuildsstuff: it shouldn't matter since the test cases only allow for one to be successfully provisioned, so I don't think that's necessary? (e.g. we could remove this line but that doesn't change the test expectation so it's fine?)

website/docs/r/api_management_subscription.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+
+* `resource_group_name` - (Required) The name of the Resource Group in which the API Management Service exists. Changing this forces a new resource to be created.
+
+* `product_id` - (Required) The ID of the Product which should be assigned to this Subscription. Changing this forces a new resource to be created.
+
+* `user_id` - (Required) The ID of the User which should be assigned to this Subscription. Changing this forces a new resource to be created.
+
+---
+
+* `state` - (Optional) The state of this Subscription. Possible values are `Active`, `Cancelled`, `Expired`, `Rejected`, `XXX`, `Submitted` and `Suspended`. Defaults to `Submitted`.
```
tombuildsstuff: we can also remove the `XXX` here:

```suggestion
* `state` - (Optional) The state of this Subscription. Possible values are `Active`, `Cancelled`, `Expired`, `Rejected`, `Submitted` and `Suspended`. Defaults to `Submitted`.
```

tombuildsstuff:
Tests pass:

Screenshot 2019-03-24 at 06 10 34

PR #3104, opened 2019-03-22
add zone_redundant as editable flag for mssql_elasticpool

bs-matil:
Hi,
we need zone_redundant elastic pools and therefore implemented the flag on the mssql_elasticpool resource.

Feel free to add any comments.

### DIFF
--- a/azurerm/resource_arm_mssql_elasticpool.go
+++ b/azurerm/resource_arm_mssql_elasticpool.go
@@ -177,7 +177,6 @@ func resourceArmMsSqlElasticPool() *schema.Resource {
 			"zone_redundant": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
 			},
 
 			"tags": tagsSchema(),
@@ -203,7 +202,6 @@ func resourceArmMsSqlElasticPoolCreateUpdate(d *schema.ResourceData, meta interf
 	elasticPoolName := d.Get("name").(string)
 	serverName := d.Get("server_name").(string)
 	resGroup := d.Get("resource_group_name").(string)
-	zoneredundant := d.Get("zone_redundant").(bool)
 
 	if requireResourcesToBeImported && d.IsNewResource() {
 		existing, err := client.Get(ctx, resGroup, serverName, elasticPoolName)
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_mssql_elasticpool_test.go
```diff
@@ -474,7 +506,8 @@ resource "azurerm_mssql_elasticpool" "test" {
   resource_group_name = "${azurerm_resource_group.test.name}"
   location            = "${azurerm_resource_group.test.location}"
   server_name         = "${azurerm_sql_server.test.name}"
-  max_size_gb         = %.7[6]f
+	max_size_gb         = %.7[6]f
```
katbyte: COuld we fix the alignment here? we use spaces not tabs in TF configs.

azurerm/resource_arm_mssql_elasticpool.go
```diff
@@ -176,7 +176,8 @@ func resourceArmMsSqlElasticPool() *schema.Resource {
 
 			"zone_redundant": {
 				Type:     schema.TypeBool,
-				Computed: true,
+				Optional: true,
+				Default:  false,
```
katbyte: Could we remove the default here and let the API decide by conditionally sending the value?

azurerm/resource_arm_mssql_elasticpool.go
```diff
@@ -227,6 +229,7 @@ func resourceArmMsSqlElasticPoolCreateUpdate(d *schema.ResourceData, meta interf
 		Tags:     expandTags(tags),
 		ElasticPoolProperties: &sql.ElasticPoolProperties{
 			PerDatabaseSettings: expandAzureRmMsSqlElasticPoolPerDatabaseSettings(d),
+			ZoneRedundant:       &zoneredundant,
```
katbyte: I think we should be conditionally setting this:

```go
if v, ok := d.GetOkExists("zone_redundant").(bool); ok {
	elasticPool. ElasticPoolProperties.ZoneRedundant = &v
}
```

This is because if someone has set the value via the portal or other means and not updated there terraform when we release this, we'll override their value.

bs-matil: thanks for this hint. didn't noticed it.

bs-matil:
@katbyte I now embedded all requested changes.

website/docs/r/mssql_elasticpool.html.markdown
```diff
@@ -92,14 +92,14 @@ The following arguments are supported:
 
 ---
 
+* `zone_redundant` - (Optional) Whether or not this elastic pool is zone redundant. `tier` needs to be `Premium` for `DTU` based  or `BusinessCritical` for `vCore` based `sku`. Defaults to `false`.
```
bs-matil: @katbyte  should we leave the Defaults to false here? As its the API behavior anyways its still may a fair hint?

katbyte: I would leave it as yes that is what the api does by default.

PR #3110, opened 2019-03-24
[Storage Account] dfs, web endpoints, file secondary endpoint

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/3082

I also noticed that the new API version provided a way to get the secondary File endpoints, so that has been added as well.

### DIFF
--- a/azurerm/resource_arm_storage_account.go
+++ b/azurerm/resource_arm_storage_account.go
@@ -1082,95 +1082,24 @@ func getBlobConnectionString(blobEndpoint *string, acctName *string, acctKey *st
 }
 
 func flattenAndSetAzureRmStorageAccountPrimaryEndpoints(d *schema.ResourceData, primary *storage.Endpoints) error {
-	var blobEndpoint, blobHost string
-	if primary != nil {
-		if v := primary.Blob; v != nil {
-			blobEndpoint = *v
-
-			u, err := url.Parse(*v)
-			if err != nil {
-				return fmt.Errorf("invalid blob endpoint for parsing: %q", *v)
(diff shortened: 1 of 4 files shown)


azurerm/resource_arm_storage_account.go
```diff
@@ -1146,5 +1225,49 @@ func flattenAndSetAzureRmStorageAccountSecondaryEndpoints(d *schema.ResourceData
 	d.Set("secondary_table_endpoint", tableEndpoint)
 	d.Set("secondary_table_host", tableHost)
 
+	var webEndpoint, webHost string
+	if secondary != nil {
+		if v := secondary.Web; v != nil {
```
katbyte: I wonder if it would be worthwhile to add a helper function to this file 🤔 
```go
if secondary != nil {
  webEndpoint, webHost err = parseEndPoint(secondary.Web)
  if err != nil { return err }
}
```

Lucretius: Yeah, I can refactor to add one in.

Lucretius:
I've added a helper function to remove most of the duplication.

Lucretius:
@katbyte turns out that I shouldn't pass both `secondary` and `secondary.Blob` (or any subproperty) into the function when the property itself could potentially be nil!  I added nil checks to both the primary and secondary and ran the previously failing acceptance test to verify.

tombuildsstuff:
Ignoring a known (existing) test failure, the tests pass:

Screenshot 2019-03-29 at 11 08 18

PR #3111, opened 2019-03-24
[Redis] Add min tls redis

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/2839 by adding MinimumTLSVersion as a configurable property to the `azurerm_redis_cache` resource.

### DIFF
--- a/azurerm/resource_arm_redis_cache.go
+++ b/azurerm/resource_arm_redis_cache.go
@@ -266,7 +266,7 @@ func resourceArmRedisCacheCreate(d *schema.ResourceData, meta interface{}) error
 				Family:   family,
 				Name:     sku,
 			},
-			MinimumTLSVersion:  expandMinimumTlsVersion(d.Get("minimum_tls_version").(string)),
+			MinimumTLSVersion:  redis.TLSVersion(d.Get("minimum_tls_version").(string)),
 			RedisConfiguration: expandRedisConfiguration(d),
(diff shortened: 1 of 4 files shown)


azurerm/resource_arm_redis_cache.go
```diff
@@ -473,6 +486,7 @@ func resourceArmRedisCacheRead(d *schema.ResourceData, meta interface{}) error {
 	if props := resp.Properties; props != nil {
 		d.Set("ssl_port", props.SslPort)
 		d.Set("hostname", props.HostName)
+		d.Set("minimum_tls_version", props.MinimumTLSVersion)
```
katbyte: Should we be converting this to a string?
```suggestion
		d.Set("minimum_tls_version", string(props.MinimumTLSVersion))
```

azurerm/resource_arm_redis_cache_test.go
```diff
@@ -595,7 +595,8 @@ resource "azurerm_redis_cache" "test" {
   capacity            = 1
   family              = "C"
   sku_name            = "Basic"
-  enable_non_ssl_port = false
+	enable_non_ssl_port = false
```
katbyte: Could we fix the alignment here?

Lucretius:
Should be all set now, I added the attribute check into the basic test

PR #3118, opened 2019-03-25
Add Probe support for Azure Container Instances

neilpeterson:
Issue - https://github.com/terraform-providers/terraform-provider-azurerm/issues/2916

I've taken my best stab on style. If I've missed any standards etc. happy to update.


azurerm/resource_arm_container_group_test.go
```diff
@@ -729,7 +813,7 @@ resource "azurerm_container_group" "test" {
         "node-name" = "acctestContainerGroup"
       }
     }
-  }
+	}
```
katbyte: Could we revert this?
```suggestion
  }
```

azurerm/resource_arm_container_group_test.go
```diff
@@ -605,7 +645,29 @@ resource "azurerm_container_group" "test" {
     secure_environment_variables = {
       "secureFoo"  = "secureBar"
       "secureFoo1" = "secureBar1"
-    }
+		}
```
katbyte: terraform configs should use spaces not tabs, as such could we revert this?
```suggestion
    }
```

azurerm/resource_arm_container_group_test.go
```diff
@@ -605,7 +645,29 @@ resource "azurerm_container_group" "test" {
...
+
+		readiness_probe {
+			exec = ["cat","/tmp/healthy"]
```
katbyte: Could we align these assignments?

azurerm/resource_arm_container_group_test.go
```diff
@@ -605,7 +645,29 @@ resource "azurerm_container_group" "test" {
     secure_environment_variables = {
       "secureFoo"  = "secureBar"
       "secureFoo1" = "secureBar1"
-    }
+		}
+
+		readiness_probe {
```
katbyte: An change this to be spaces not tabs here.

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1235,56 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+		httpget := make(map[string]interface{})
+		httpget["path"] = *input.HTTPGet.Path
+		httpget["port"] = *input.HTTPGet.Port
+		httpget["scheme"] = input.HTTPGet.Scheme
+
+		output["httpget"] = []interface{}{httpget}
+
+	}
+
+	if FailureThreshold := input.FailureThreshold; FailureThreshold != nil {
```
katbyte: Might i suggest using v instead of FailureThreshold here to better match the existing provider style?

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1235,56 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+		output["failure_threashold"] = input.FailureThreshold
```
katbyte: This should be be `*FailureThreshold`, and with above suggestion:
```suggestion
		output["failure_threashold"] = *v
```

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1235,56 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+	}
+
+	if InitialDelaySeconds := input.InitialDelaySeconds; InitialDelaySeconds != nil {
+		output["inital_delay_seconds"] = input.InitialDelaySeconds
```
katbyte: Same here

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1235,56 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+	}
+
+	if PeriodSeconds := input.PeriodSeconds; PeriodSeconds != nil {
+		output["period_seconds"] = input.PeriodSeconds
```
katbyte: and here

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1235,56 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+	}
+
+	if SuccessThreshold := input.SuccessThreshold; SuccessThreshold != nil {
+		output["sucess_threashold"] = input.SuccessThreshold
```
katbyte: and here

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1235,56 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+	}
+
+	if TimeoutSeconds := input.TimeoutSeconds; TimeoutSeconds != nil {
+		output["timeout_seconds"] = input.TimeoutSeconds
```
katbyte: and here

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1235,56 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+		var exec []string
+		for _, v := range *input.Exec.Command {
+			exec = append(exec, v)
+		}
+		output["exec"] = exec
+	}
+
+	if HTTPGet := input.HTTPGet; HTTPGet != nil {
+		httpget := make(map[string]interface{})
+		httpget["path"] = *input.HTTPGet.Path
```
katbyte: These values should be nil checked before being dereferenced

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1235,56 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+	outputs := make([]interface{}, 0)
+	if input == nil {
+		return outputs
+	}
+
+	output := make(map[string]interface{})
+
+	if Exec := input.Exec; Exec != nil {
+		var exec []string
+		for _, v := range *input.Exec.Command {
```
katbyte: We should nil check this

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1235,56 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+			exec = append(exec, v)
+		}
+		output["exec"] = exec
+	}
+
+	if HTTPGet := input.HTTPGet; HTTPGet != nil {
+		httpget := make(map[string]interface{})
+		httpget["path"] = *input.HTTPGet.Path
+		httpget["port"] = *input.HTTPGet.Port
+		httpget["scheme"] = input.HTTPGet.Scheme
```
katbyte: This should be converted to a string before assingment

azurerm/resource_arm_container_group.go
```diff
@@ -762,6 +920,73 @@ func expandContainerVolumes(input interface{}) (*[]containerinstance.VolumeMount
...
+
+		if timeoutSeconds := probeConfig["timeout_seconds"].(int); timeoutSeconds > 0 {
+			probe.TimeoutSeconds = utils.Int32(int32(timeoutSeconds))
+		}
+
+		commands := probeConfig["exec"].([]interface{})
+		if len(commands) > 0 {
+			execRaw := make([]string, 0, len(commands))
+
+			for _, command := range commands {
```
katbyte: I think you could use `utils.ExpandStringArray` here

azurerm/resource_arm_container_group.go
```diff
@@ -762,6 +920,73 @@ func expandContainerVolumes(input interface{}) (*[]containerinstance.VolumeMount
...
+	probeRaw := input.([]interface{})
+
+	if len(probeRaw) == 0 {
+		return nil
+	}
+
+	for _, p := range probeRaw {
+		probeConfig := p.(map[string]interface{})
+
+		if initialDelaySeconds := probeConfig["inital_delay_seconds"].(int); initialDelaySeconds > 0 {
```
katbyte: _minor and not a blocker_ but you could use v here to simplify this a little:
```go
		if v := probeConfig["inital_delay_seconds"].(int); v > 0 {
			probe.InitialDelaySeconds = utils.Int32(int32(v))
		}
```

But totally optional change

azurerm/resource_arm_container_group.go
katbyte: Could potentially use `utils.FlattenStringArray`

azurerm/resource_arm_container_group.go
```diff
@@ -661,6 +809,16 @@ func expandContainerGroupContainers(d *schema.ResourceData) (*[]containerinstanc
 			}
 		}
 
+		if v, ok := data["liveness_probe"]; ok {
+			livenessProbe := expandContainerProbe(v)
+			container.ContainerProperties.LivenessProbe = livenessProbe
```
katbyte: _minor_ personally i would combine these two lines

```
container.ContainerProperties.LivenessProbe = expandContainerProbe(v)
```

azurerm/resource_arm_container_group.go
```diff
@@ -307,6 +307,154 @@ func resourceArmContainerGroup() *schema.Resource {
...
+									"timeout_seconds": {
+										Type:     schema.TypeInt,
+										Optional: true,
+										ForceNew: true,
+									},
+								},
+							},
+						},
+
+						"readiness_probe": {
```
katbyte: Given that these two properties share the exact same schema, would it make sense to create a helper function that returns the schema?

I am also not going to make the same comments above here.

azurerm/resource_arm_container_group.go
```diff
@@ -307,6 +307,154 @@ func resourceArmContainerGroup() *schema.Resource {
...
+										Type:     schema.TypeInt,
+										Optional: true,
+										ForceNew: true,
+									},
+
+									"timeout_seconds": {
+										Type:     schema.TypeInt,
+										Optional: true,
+										ForceNew: true,
+									},
```
katbyte: Could we validate this and set the portals default? (and document the default)

azurerm/resource_arm_container_group.go
```diff
@@ -307,6 +307,154 @@ func resourceArmContainerGroup() *schema.Resource {
...
+
+									"sucess_threashold": {
+										Type:     schema.TypeInt,
+										Optional: true,
+										ForceNew: true,
+									},
```
katbyte: Could we validate this and set the portals default? (and document the default)

azurerm/resource_arm_container_group.go
```diff
@@ -307,6 +307,154 @@ func resourceArmContainerGroup() *schema.Resource {
...
+
+									"failure_threashold": {
+										Type:     schema.TypeInt,
+										Optional: true,
+										ForceNew: true,
+									},
```
katbyte: Could we validate this and set the portals default? (and document the default)

azurerm/resource_arm_container_group.go
```diff
@@ -307,6 +307,154 @@ func resourceArmContainerGroup() *schema.Resource {
...
+
+									"period_seconds": {
+										Type:     schema.TypeInt,
+										Optional: true,
+										ForceNew: true,
+									},
```
katbyte: Could we validate this and set the portals default? (and document the default)

azurerm/resource_arm_container_group.go
```diff
@@ -307,6 +307,154 @@ func resourceArmContainerGroup() *schema.Resource {
...
+												},
+											},
+										},
+									},
+
+									"inital_delay_seconds": {
+										Type:     schema.TypeInt,
+										Optional: true,
+										ForceNew: true,
+									},
```
katbyte: Could we validate this and set the portals default? (and document the default)

azurerm/resource_arm_container_group.go
```diff
@@ -307,6 +307,154 @@ func resourceArmContainerGroup() *schema.Resource {
...
+												"path": {
+													Type:     schema.TypeString,
+													Optional: true,
+													ForceNew: true,
+												},
+												"port": {
+													Type:     schema.TypeInt,
+													Optional: true,
+													ForceNew: true,
+												},
```
katbyte: Could we validate this with `validate.PortNumber`

azurerm/resource_arm_container_group.go
```diff
@@ -307,6 +307,154 @@ func resourceArmContainerGroup() *schema.Resource {
...
+							Optional: true,
+							ForceNew: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"exec": {
+										Type:     schema.TypeList,
+										Optional: true,
+										ForceNew: true,
+										Elem:     &schema.Schema{Type: schema.TypeString},
```
katbyte: Should we be validating this to prevent empty elements?

neilpeterson:
Thanks a bunch, @katbyte I will address each and also the linting issues picked up by Travis CI.

azurerm/resource_arm_container_group.go
neilpeterson: @katbyte yep, I was wondering how to properly handle this. I would be happy to do this. Are you aware of any other places in the provider that have a similar function that I can use as guidance?

neilpeterson:
@katbyte I've addressed all feedback with the following notes:

- Because the two probe types share a scheme, use a helper function to populate properties ([link](https://github.com/terraform-providers/terraform-provider-azurerm/pull/3118/files#r268812783)). I am happy to do this assuming we can preserve the end user experience. Do you have a sample of where a similar helper function was used that I can draw from?
- Validate and add portal defailts. From what I can tell the Azure APIs are not enforcing default values nor a range of expected values ([link](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/containerinstance/resource-manager/Microsoft.ContainerInstance/stable/2018-10-01/containerInstance.json#L1108)). This fetaure is also not avaliable .via the Azure portal. I have added default taken from [this Kuberntes document](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes). I have contacted some MSFT folk to fund out what the expectation is here.

katbyte:
@neilpeterson,

Here is an example of a scheme function:
```go
func SchemaDevTestVirtualMachineInboundNatRule() *schema.Schema {
	return &schema.Schema{
		Type:     schema.TypeSet,
		Optional: true,
		// since these aren't returned from the API
		ForceNew: true,
		Elem: &schema.Resource{
				"protocol": {
					Type:     schema.TypeString,
					Required: true,
						string(dtl.TCP),
						string(dtl.UDP),
					}, false),
				},

				"backend_port": {
					Type:         schema.TypeInt,
					Required:     true,
					ForceNew:     true,
					ValidateFunc: validate.PortNumber,
				},

				"frontend_port": {
					Type:     schema.TypeInt,
					Computed: true,
				},
			},
		},
	}
}
```

As for the validation if there are no easy win there don't bother yourself too much trying to find some, its not a blocker to merge without 🙂

neilpeterson:
@katbyte thanks a bunch. I've added the helper function and removed the defaults to be more in line with the API. When we sort the Azure end I will come back and add default values back in.

With that, I believe this is ready for another review. Thanks for your help.

azurerm/helpers/azure/container_group.go
```diff
@@ -0,0 +1,91 @@
...
+								ValidateFunc: validate.PortNumber,
+							},
+							"scheme": {
+								Type:     schema.TypeString,
+								Optional: true,
+								ForceNew: true,
+								ValidateFunc: validation.StringInSlice([]string{
+									string(containerinstance.HTTP),
+									string(containerinstance.HTTPS),
+								}, true),
```
katbyte: Could we change this ti false? we are moving towards case sensitivity for properties like this

azurerm/resource_arm_container_group.go
```diff
@@ -661,6 +665,15 @@ func expandContainerGroupContainers(d *schema.ResourceData) (*[]containerinstanc
 			}
 		}
 
+		if v, ok := data["liveness_probe"]; ok {
+			container.ContainerProperties.LivenessProbe = expandContainerProbe(v)
+		}
+
+		if v, ok := data["readiness_probe"]; ok {
+			readinessProbe := expandContainerProbe(v)
+			container.ContainerProperties.ReadinessProbe = readinessProbe
```
katbyte: Could we make this one line for consistency with the above assignment?

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1084,61 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+		return outputs
+	}
+
+	output := make(map[string]interface{})
+
+	if v := input.Exec; v != nil {
+		output["exec"] = *v.Command
+	}
+
+	if v := input.HTTPGet; v != nil {
```
katbyte: _minor_ i would change this to
```suggestion
	if get := input.HTTPGet; v != nil {
```

so we can use it below

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1084,61 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+		httpget := make(map[string]interface{})
+
+		if v := input.HTTPGet.Path; v != nil {
```
katbyte:
```suggestion
		if v := get.Path; v != nil {
```

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1084,61 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+			httpget["path"] = *v
+		}
+
+		if v := input.HTTPGet.Port; v != nil {
```
katbyte:
```suggestion
		if v := get.Port; v != nil {
```

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1084,61 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+			httpget["port"] = *v
+		}
+
+		if input.HTTPGet.Scheme != "" {
```
katbyte:
```suggestion
		if get.Scheme != "" {
```

azurerm/resource_arm_container_group.go
```diff
@@ -1002,6 +1084,61 @@ func flattenContainerVolumes(volumeMounts *[]containerinstance.VolumeMount, cont
...
+			httpget["scheme"] = input.HTTPGet.Scheme
```
katbyte:
```suggestion
			httpget["scheme"] = get.Scheme
```

azurerm/helpers/azure/container_group.go
neilpeterson: @katbyte I am struggling a bit with this one. The Azure API always returns `Http` or `Https`. When setting `ignoreCase` to `false` tests on this value begin to fail. 

My assumption is that these tests should work.

```
    liveness_probe {
      httpget {
        path   = "/"
        port   = 443
        scheme = "Http"
      }
      inital_delay_seconds = 1
      period_seconds       = 1
      failure_threashold   = 1
      sucess_threashold    = 1
      timeout_seconds      = 1
    }
```

and then:

```
resource.TestCheckResourceAttr(resourceName,"container.0.liveness_probe.0.httpget.0.scheme", "Http"),
```

However, it seems I am hitting [this](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/vendor/github.com/hashicorp/terraform/helper/validation/validation.go#L90), which returns:

```
testing.go:538: Step 0 error: config is invalid: azurerm_container_group.test: expected container.0.liveness_probe.0.httpget.0.scheme to be one of [http https], got Http
```

I am still investigating but if you have any advice, I would appreciate it :).

Thanks

azurerm/helpers/azure/container_group.go
```diff
@@ -0,0 +1,91 @@
...
+							},
+						},
+					},
+				},
+
+				"inital_delay_seconds": {
```
tombuildsstuff:
```suggestion
				"initial_delay_seconds": {
```

azurerm/helpers/azure/container_group.go
neilpeterson: Thanks, Tom. I figured something was going on here and had considered this approach. Appreciate the assist and validation.

neilpeterson:
@tombuildsstuff, thanks a bunch. Changes all make sense, really appreciate the assistance in wrapping this up.

tombuildsstuff:
Tests pas:

Screenshot 2019-04-08 at 05 00 35

PR #3121, opened 2019-03-26
New Resource: `azurerm_api_management_api_operation`

tombuildsstuff:
This PR introduces a new resource for API Management Operations, as a part of #1177 

After spending /a long time/ with the nested sets, I ended up porting most elements (header, response etc) to Lists, since the API appears to be consistent in it's ordering which is helpful to avoid 3-level deep nested sets. There's also a TODO here to add an additional acceptance test when API Management Schema's are supported, but it's fine for now.

Tests pass:

Screenshot 2019-03-26 at 11 23 08

### DIFF
--- a/azurerm/helpers/azure/api_management.go
+++ b/azurerm/helpers/azure/api_management.go
@@ -1,8 +1,13 @@
 package azure
 
 import (
+	"fmt"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/services/apimanagement/mgmt/2018-01-01/apimanagement"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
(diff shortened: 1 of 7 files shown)


azurerm/helpers/azure/api_management.go
```diff
@@ -70,3 +75,225 @@ func SchemaApiManagementUserDataSourceName() *schema.Schema {
...
+
+		output := apimanagement.RepresentationContract{
+			ContentType: utils.String(contentType),
+			Sample:      utils.String(sample),
+		}
+
+		contentTypeIsFormData := strings.EqualFold(contentType, "multipart/form-data") || strings.EqualFold(contentType, "application/x-www-form-urlencoded")
+
+		// Representation formParameters can only be specified for form data content types (multipart/form-data, application/x-www-form-urlencoded)
+		if contentTypeIsFormData {
```
alexsomesan: Neat approach to block-wide validations!

PR #3123, opened 2019-03-26
New Resource: `azurerm_api_management_authorization_server`

tombuildsstuff:
This PR introduces support for managing Authorization Servers within an API Management Service

### DIFF
--- a/azurerm/resource_arm_api_management_authorization_server.go
+++ b/azurerm/resource_arm_api_management_authorization_server.go
@@ -1,0 +1,484 @@
+package azurerm
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/Azure/azure-sdk-for-go/services/apimanagement/mgmt/2018-01-01/apimanagement"
+	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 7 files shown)


azurerm/resource_arm_api_management_authorization_server.go
```diff
@@ -0,0 +1,492 @@
...
+			"authorization_endpoint": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+
+			"client_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
```
katbyte: Could we validate this as a resource id:

tombuildsstuff:
Tests pass:

Screenshot 2019-03-29 at 09 58 05

azurerm/resource_arm_api_management_authorization_server.go
tombuildsstuff: this is a user specifyable value (which can be anything) so unfortunately this is the best we can do here

PR #3143, opened 2019-03-28
New Resource: `azurerm_api_management_openid_connect_provider`

tombuildsstuff:
This PR adds support for API Management OpenID Connect Providers, a part of #1177

### DIFF
--- a/azurerm/resource_arm_api_management_openid_connect_provider.go
+++ b/azurerm/resource_arm_api_management_openid_connect_provider.go
@@ -1,0 +1,170 @@
+package azurerm
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/Azure/azure-sdk-for-go/services/apimanagement/mgmt/2018-01-01/apimanagement"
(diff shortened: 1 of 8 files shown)


tombuildsstuff:
Tests pass:

Screenshot 2019-03-28 at 23 07 22

azurerm/resource_arm_api_management_openid_connect_provider.go
```diff
@@ -0,0 +1,170 @@
...
+
+			"resource_group_name": resourceGroupNameSchema(),
+
+			"api_management_name": azure.SchemaApiManagementName(),
+
+			"client_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				Sensitive:    true,
+				ValidateFunc: validate.NoEmptyStrings,
```
katbyte: Should we validate this is an id rather then just not empty?

tombuildsstuff: this is a user-specifyable Client ID, so could be anything unfortunately

PR #3144, opened 2019-03-29
`azurerm_application_gateway` - support for `zones`

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/2293

Updated documentation and added an acceptance test.  Note that as I did not want to add more flags to the test template I had to add a public IP resource (standard_v2 AGW requires a standard SKU public IP - different from the IP used for the rest of the tests).  The only downside is we make an extra, unused dynamic IP address since I also rely on the template to get the rest of the resources.  I could refactor to add more flags to the template function and avoid the creation of this resource if desired.

### DIFF
--- a/azurerm/resource_arm_application_gateway.go
+++ b/azurerm/resource_arm_application_gateway.go
@@ -33,6 +33,8 @@ func resourceArmApplicationGateway() *schema.Resource {
 
 			"location": locationSchema(),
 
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
Tests pass:

Screenshot 2019-03-29 at 10 28 44

PR #3145, opened 2019-03-29
New Resource: azurerm_api_management_policy

JunyiYi:
In this PR, I introduced the global API Management Policy.

Since the backend service will always re-format the XML content (for example, change line endings to `\r\n`, change quotation marks, using `\t` to indent) of the policy; I implement a XML ignore diff function.

I also thought of wrapping all the policies into an abstracted structure in Terraform. However that is not doable due to [advanced recursive control flows](https://docs.microsoft.com/en-us/azure/api-management/api-management-advanced-policies#choose).

`xml_link` requires a public-accessible HTTP endpoint of an valid configuration XML file. That's why I have not included any acceptance tests. Maybe we can add some files in `examples` folder and add the test cases later?


tombuildsstuff:
hey @JunyiYi 

Thanks for this PR :)

Taking a look into this it appears it's only possible to specify one Policy for the API Management Service (since there's no unique name per top-level policy) - as such rather than making this it's own resource I believe we should roll this up into the parent API Management Resource. I hope you don't mind, but I'm going to pull some of the changes you've made here out into a new PR which merges this into the parent `azurerm_api_management` resource so that we can get this merged.

Thanks!

PR #3151, opened 2019-04-01
r/api_management: support for `sign_in`, `sign_up` and `policy` blocks

### LINKED ISSUE #1177: Feature Request: API Management support

tombuildsstuff:
This PR adds support for the `sign_in` and `sign_up` settings blocks - and also adds support for the `policy` block, which was previously added as a separate resource by @JunyiYi in #3145 (I've rolled this up into this PR since it's only possible to have one per API, thus it makes more sense within this resource).

This is the last PR from our list covered in #1177 - as such this fixed #1177

### DIFF
--- a/azurerm/helpers/suppress/xml.go
+++ b/azurerm/helpers/suppress/xml.go
@@ -1,0 +1,47 @@
+package suppress
+
+import (
+	"encoding/xml"
+	"io"
+	"reflect"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func XmlDiff(_, old, new string, _ *schema.ResourceData) bool {
+	oldTokens, err := expandXmlTokensFromString(old)
+	if err != nil {
+		return false
+	}
+
+	newTokens, err := expandXmlTokensFromString(new)
+	if err != nil {
+		return false
+	}
+
+	return reflect.DeepEqual(oldTokens, newTokens)
+}
+
+// This function will extract all XML tokens from a string, but ignoring all white-space tokens
+func expandXmlTokensFromString(input string) ([]xml.Token, error) {
+	decoder := xml.NewDecoder(strings.NewReader(input))
+	tokens := make([]xml.Token, 0)
+	for {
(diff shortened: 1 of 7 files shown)


azurerm/helpers/suppress/xml.go
```diff
@@ -0,0 +1,47 @@
...
+	"reflect"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func XmlDiff(_, old, new string, _ *schema.ResourceData) bool {
+	oldTokens, err := expandXmlTokensFromString(old)
+	if err != nil {
+		return false
```
mbfrahry: Just confirming that we don't want to handle about the errors that `expandXmlTokensFromString` returns

tombuildsstuff: since this is a diff suppress func it should be fine, it'll just mean a diff is shown vs no diff (which is fine) 👍

tombuildsstuff:
Tests pass:

Screenshot 2019-04-03 at 10 27 00

cpdohert:
Forgive me if I'm not understanding something about API management policies here, but there are four different scopes for APIM policies: global, product, API and API operation.  We make heavy use of product-scoped policies and I was hoping to make use of terraform for this now that the api_management resource has been merged - but it doesn't seem like there's any way to do that with the policy block as implemented.  Also, if it's true that there can only be one policy per API, doesn't it make more sense for the policy {} block to be implemented in api_management_api than api_management?  Implementing it in api_management would imply global policy scope.

tombuildsstuff:
@cpdohert this PR added support for Global Policies - we can look to add support for the other APIM policy resources but there’s no issue/feature request tracking them at the moment as far as I recall - would you mind opening a new issue for that?

Issue #1500, opened 2018-07-05
Eliminate repetition when specifying SKU sizing


Supermathie:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When specifying the sku for postgres and redis (and probably some other resources) you need to repeat yourself when specifying SKUs, example:

```
resource "azurerm_postgresql_server" "test" {
  name                = "postgresql-server-1"

  sku {
    name = "B_Gen4_2"
    capacity = 2
    tier = "Basic"
    family = "Gen4"
  }
…
}

resource "azurerm_redis_cache" "test" {
  name                = "tf-redis-basic"
  family              = "C"
  sku_name            = "Basic"
…
}

resource "azurerm_application_gateway" "network" {
  name                = "my-application-gateway-12345"
  sku {
    name           = "Standard_Small"
    tier           = "Standard"
    capacity       = 2
  }
…
}
```

There is unnecessary repetition here - e.g. we can eliminate `name` from the pg resource and `family` from redis since they can be 100% inferred from the other values

### New or Affected Resource(s)

* azurerm_postgresql_server
* azurerm_redis_cache
* azurerm_application_gateway

### Potential Terraform Configuration

```hcl
resource "azurerm_postgresql_server" "test" {
  name                = "postgresql-server-1"

  sku {
    capacity = 2
    tier = "Basic"
    family = "Gen4"
  }
…
}

resource "azurerm_postgresql_server" "test" {
  name                = "postgresql-server-1"

  sku {
    name = "B_Gen4_2"
  }
…
}

resource "azurerm_redis_cache" "test" {
  name                = "tf-redis-basic"
  sku_name            = "Basic"
…
}

resource "azurerm_application_gateway" "network" {
  name                = "my-application-gateway-12345"
  sku {
    name           = "Standard_Small"
    capacity       = 2
  }
…
}

```

### References

ø

Supermathie:
Regarding the choice of:

```
  sku {
    capacity = 2
    tier = "Basic"
    family = "Gen4"
  }
```

vs.

```
  sku {
    name = "B_Gen4_2"
  }
```

[...]

PR #3154, opened 2019-04-01
azurerm_api_management_api - deprecate `sku` in favour of the `sku_name` property

jeffreyCline:
Continuation of addressing issue #1500, Flatten the `SKU` block into a single attribute and unifying the handling of `SKU` across all resources within the Azure Terraform provider.

Example:

```
resource "azurerm_api_management_api" "test" {
  name      = "api-management-1"
  sku_name  = "Developer_1"
}
```


r0bnet:
Thanks for this. Without this feature it's a nightmare to create something like t-shirt sizes if you need to define 3 values for each size (for instance for Postgres). Hope this will find it's way into 2.0.0

azurerm/data_source_api_management.go
```diff
@@ -39,21 +39,9 @@ func dataSourceApiManagementService() *schema.Resource {
 				Computed: true,
 			},
 
-			"sku": {
-				Type:     schema.TypeList,
+			"sku_name": {
```
katbyte: This is a breaking change. We should retain sku until 2.0.

azurerm/resource_arm_api_management.go
```diff
@@ -77,15 +80,28 @@ func resourceArmApiManagementService() *schema.Resource {
 								string(apimanagement.SkuTypePremium),
 							}, false),
 						},
+
 						"capacity": {
 							Type:         schema.TypeInt,
 							Required:     true,
-							ValidateFunc: validation.IntAtLeast(0),
+							ValidateFunc: validation.IntAtLeast(1),
```
katbyte: How come this value changed?

azurerm/resource_arm_api_management.go
```diff
@@ -77,15 +80,28 @@ func resourceArmApiManagementService() *schema.Resource {
...
 						},
 					},
 				},
 			},
 
+			"sku_name": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"sku"},
+				ValidateFunc: azure.SkuNameStringInSlice([]string{
```
katbyte: This function seems rather specific to this service?

azurerm/resource_arm_api_management.go
```diff
@@ -366,6 +382,14 @@ func resourceArmApiManagementServiceCreateUpdate(d *schema.ResourceData, meta in
 	client := meta.(*ArmClient).apiManagementServiceClient
 	ctx := meta.(*ArmClient).StopContext
 
+	log.Printf("[INFO] validating SKU for API Management Service.")
```
katbyte: this seems like a mistake?

azurerm/resource_arm_api_management.go
```diff
@@ -562,8 +585,21 @@ func resourceArmApiManagementServiceRead(d *schema.ResourceData, meta interface{
 		}
 	}
 
-	if err := d.Set("sku", flattenApiManagementServiceSku(resp.Sku)); err != nil {
-		return fmt.Errorf("Error setting `sku`: %+v", err)
+	if skuName == "" {
```
katbyte: As mentioned in other PR, we can just set both and simplify this logic

azurerm/resource_arm_api_management.go
```diff
@@ -850,21 +886,49 @@ func flattenAzureRmApiManagementMachineIdentity(identity *apimanagement.ServiceI
 }
 
 func expandAzureRmApiManagementSku(d *schema.ResourceData) *apimanagement.ServiceSkuProperties {
```
katbyte: Can we split this into

`expandAzureRmApiManagementSku()` and `expandAzureRmApiManagementSkuName()` the the logic to pick in the calling function?

azurerm/resource_arm_api_management_api_test.go
```diff
@@ -33,6 +33,43 @@ func TestAccAzureRMApiManagementApi_basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, "soap_pass_through", "false"),
+					resource.TestCheckResourceAttr(resourceName, "is_current", "true"),
+					resource.TestCheckResourceAttr(resourceName, "is_online", "false"),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
```
katbyte: We shouldn't be doing this.

azurerm/resource_arm_api_management_api_test.go
```diff
@@ -59,6 +96,10 @@ func TestAccAzureRMApiManagementApi_wordRevision(t *testing.T) {
 				ResourceName:      resourceName,
 				ImportState:       true,
 				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
```
katbyte: We shouldn't be doing this.

azurerm/resource_arm_api_management_api_test.go
```diff
@@ -113,6 +154,10 @@ func TestAccAzureRMApiManagementApi_soapPassthrough(t *testing.T) {
...
```
katbyte: We shouldn't be doing this.

azurerm/resource_arm_api_management_api_test.go
```diff
@@ -139,6 +184,8 @@ func TestAccAzureRMApiManagementApi_importSwagger(t *testing.T) {
 				ImportState:       true,
 				ImportStateVerify: true,
 				ImportStateVerifyIgnore: []string{
+					"sku",
```
katbyte: We shouldn't be doing this.

azurerm/resource_arm_api_management_api_test.go
```diff
@@ -168,6 +215,8 @@ func TestAccAzureRMApiManagementApi_importWsdl(t *testing.T) {
...
```
katbyte: And so on

azurerm/resource_arm_api_management_test.go
```diff
@@ -30,6 +31,58 @@ func TestAccAzureRMApiManagement_basic(t *testing.T) {
 				ResourceName:      resourceName,
 				ImportState:       true,
 				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
```
katbyte: We shouldn't be doing this.

website/docs/r/api_management.html.markdown
```diff
@@ -46,7 +43,9 @@ The following arguments are supported:
 
 * `publisher_email` - (Required) The email of publisher/company.
 
-* `sku` - (Required) A `sku` block as documented below.
+* `sku`  - (Required) A `sku` block as documented below
```
katbyte: Can we mark this as deprecated?

azurerm/resource_arm_api_management.go
jeffreyCline: Per the SDK implementation:
```
// Capacity - Capacity of the SKU (number of deployed units of the SKU). The default value is 1.
```

azurerm/resource_arm_api_management.go
jeffreyCline: There are lots of `SKU`'s that have the format of `Name_Capacity` so I will be able to reuse this function in other resources flatten `SKU` PR's. I have updated the name so it is more descriptive of what it is actually doing.

azurerm/resource_arm_api_management.go
jeffreyCline: Removed the line of code...

azurerm/resource_arm_api_management.go
jeffreyCline: I now set both and removed the logic...

azurerm/resource_arm_api_management.go
jeffreyCline: Done.

azurerm/data_source_api_management.go
jeffreyCline: kk

azurerm/resource_arm_api_management_api_test.go
jeffreyCline: Fixed.

azurerm/resource_arm_api_management_api_test.go
jeffreyCline: Fixed.

azurerm/resource_arm_api_management_test.go
jeffreyCline: Fixed.

azurerm/resource_arm_api_management_api_test.go
jeffreyCline: Fixed.

azurerm/resource_arm_api_management_api_test.go
jeffreyCline: Fixed.

azurerm/resource_arm_api_management_api_test.go
jeffreyCline: Fixed.

website/docs/r/api_management.html.markdown
jeffreyCline: Fixed.

azurerm/data_source_api_management.go
katbyte: It appears your still removing the old property

azurerm/resource_arm_api_management.go
katbyte: that says default not minimum? thou it does make sense

jeffreyCline:
Added `SKU` back into the DataSource...

website/docs/r/api_management.html.markdown
```diff
@@ -46,7 +43,9 @@ The following arguments are supported:
 
 * `publisher_email` - (Required) The email of publisher/company.
 
-* `sku` - (Required) A `sku` block as documented below.
+* `sku`  - (Deprecated) A `sku` block as documented below
+
+* `sku_name` - (Required) `sku_name` is a string consisting of two parts separated by an underscore(_). The fist part is the `name`, valid values include: `Developer`, `Basic`, `Standard` and `Premium`. The second part is the `capacity` (e.g. the number of deployed units of the `sku`), which must be a positive `integer` (e.g. `Developer_1`).
```
katbyte: You'll need to escape the _ as its all in italics now? or is it just the guthub viewer here

website/docs/r/api_management.html.markdown
```diff
@@ -185,12 +184,14 @@ A `security` block supports the following:
...
 
-A `sku` block supports the following:
+A `sku` block supports the following: (Deprecated)
 
-* `name` - (Required) Specifies the Pricing Tier for the API Management Service. Possible values include: `Developer`, `Basic`, `Standard` and `Premium`.
+* `name` - (Required) Specifies the Pricing Tier for the API Management Service. Possible values include: Developer, Basic, Standard and Premium.
 
 * `capacity` - (Required) Specifies the Pricing Capacity for the API Management Service.
 
+-> **Note:** This property has been deprecated in favour of the `sku_name` property and will be removed in version 2.0 of the provider.
```
katbyte: Could we add this to the upgrading to 2.0 guide found in the guides directory?

website/docs/r/api_management.html.markdown
jeffreyCline: Fixed.

website/docs/r/api_management.html.markdown
jeffreyCline: Sure... added both the data source and the resource.

azurerm/resource_arm_api_management.go
jeffreyCline: Reverted code and made it like it was, to avoid potential impact to existing users...

PR #3173, opened 2019-04-04
iothub IP filters

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/1699 by adding the ability to configure IP Filter rules for the iothub resource.  Updated the documentation accordingly and added a test.

### DIFF
--- a/azurerm/resource_arm_iothub.go
+++ b/azurerm/resource_arm_iothub.go
@@ -505,7 +505,7 @@ func resourceArmIotHubRead(d *schema.ResourceData, meta interface{}) error {
 			return fmt.Errorf("Error setting `fallbackRoute` in IoTHub %q: %+v", name, err)
 		}
 
-		ipFilterRules := flattenIPFilterRules((properties.IPFilterRules))
+		ipFilterRules := flattenIPFilterRules(properties.IPFilterRules)
 		if err := d.Set("ip_filter_rule", ipFilterRules); err != nil {
 			return fmt.Errorf("Error setting `ip_filter_rule` in IoTHub %q: %+v", name, err)
 		}
--- a/azurerm/resource_arm_iothub.go
+++ b/azurerm/resource_arm_iothub.go
@@ -970,9 +970,16 @@ func flattenIPFilterRules(in *[]devices.IPFilterRule) []interface{} {
 
 	for _, r := range *in {
 		rawRule := make(map[string]interface{})
-		rawRule["name"] = *r.FilterName
+
+		if r.FilterName != nil {
+			rawRule["name"] = *r.FilterName
+		}
+
 		rawRule["action"] = string(r.Action)
-		rawRule["ip_mask"] = *r.IPMask
+
+		if r.IPMask != nil {
+			rawRule["ip_mask"] = *r.IPMask
+		}
 		rules = append(rules, rawRule)
 	}
 	return rules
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_iothub.go
```diff
@@ -323,6 +323,32 @@ func resourceArmIotHub() *schema.Resource {
 				},
 			},
 
+			"ip_filter_rule": {
+				Type:     schema.TypeList,
```
tombuildsstuff: as there could be multiple of these (and they could be returned in any order) - I think this could be better as a Set?

```suggestion
				Type:     schema.TypeSet,
```

azurerm/resource_arm_iothub.go
```diff
@@ -476,6 +504,12 @@ func resourceArmIotHubRead(d *schema.ResourceData, meta interface{}) error {
 		if err := d.Set("fallback_route", fallbackRoute); err != nil {
 			return fmt.Errorf("Error setting `fallbackRoute` in IoTHub %q: %+v", name, err)
 		}
+
+		ipFilterRules := flattenIPFilterRules((properties.IPFilterRules))
```
tombuildsstuff: we can remove the duplicate `()` here 😄

```suggestion
		ipFilterRules := flattenIPFilterRules(properties.IPFilterRules)
```

azurerm/resource_arm_iothub.go
```diff
@@ -907,3 +941,39 @@ func validateIoTHubFileNameFormat(v interface{}, k string) (warnings []string, e
 
 	return warnings, errors
 }
+func expandIPFilterRules(d *schema.ResourceData) *[]devices.IPFilterRule {
+	ipFilterRuleList := d.Get("ip_filter_rule").([]interface{})
```
tombuildsstuff: if we make this a Set then we'll want this to be:

```suggestion
	ipFilterRuleList := d.Get("ip_filter_rule").(*schema.Set).List()
```

azurerm/resource_arm_iothub.go
```diff
@@ -907,3 +941,39 @@ func validateIoTHubFileNameFormat(v interface{}, k string) (warnings []string, e
...
+	rules := make([]interface{}, 0)
+	if in == nil {
+		return rules
+	}
+
+	for _, r := range *in {
+		rawRule := make(map[string]interface{})
+		rawRule["name"] = *r.FilterName
+		rawRule["action"] = string(r.Action)
+		rawRule["ip_mask"] = *r.IPMask
```
tombuildsstuff: if the API response is malformed there's the possibility that these could be nil (and as such crash here) - could we nil check around r.FilterName and r.IPMask here?

Lucretius:
I've added the desired changes.  Thanks for the feedback!

azurerm/resource_arm_iothub.go
```diff
@@ -323,6 +323,32 @@ func resourceArmIotHub() *schema.Resource {
...
 
+			"ip_filter_rule": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
```
katbyte: Should we validate this with `NoEmptyStrings`?

tombuildsstuff: 👍 agreed - since this otherwise LGTM I'm going to push a commit to add this so we can merge it, I hope you don't mind @Lucretius

Lucretius:
@tombuildsstuff not a problem, thanks for doing that!

tombuildsstuff:
Tests pass:

Screenshot 2019-04-08 at 04 50 42

PR #3174, opened 2019-04-04
AKS enable autoscaling

Lucretius:
This bumps the containerservice SDK - fixing a few pieces that the new API changed.

The goal here is to get the kubernetes VMSS working, which would resolve https://github.com/terraform-providers/terraform-provider-azurerm/issues/2900

I have added a test to both the AKS resource and the AKS data source but I am making a few assumptions I need to validate. 
1) If autoscaling is enabled is the type always `VirtualMachineScaleSets` (conversely, if it is not enabled is it always `AvailabilitySet`)?  [This link](https://docs.microsoft.com/en-us/azure/aks/cluster-autoscaler#install-aks-preview-cli-extension) seems to affirm this assumption.
2) ~Is the allowed `min_count` 0 or 1?  I've set it to 0 in the test, so if it is not allowed that test should fail.~ Min count must be 1 or greater.  Setting to 0 caused the API to error with an invalid `min_count` value.


Lucretius:
I suppose we can start reviewing this, with the existing test coverage coupled with the new tests we should be fine to just run this against the acceptance test suite.

Lucretius:
I am testing this PR out now, using the local build to create some kubernetes clusters.  Hold off on any review until I work the kinks out of this.

Lucretius:
So, I haven't run the full test suite as I am still using just a personal subscription but I have run the new tests (both with VMSS autoscale off and VMSS autoscale on) and they've passed.  I learned that I cannot have `min_count` and `max_count` set for just a VMSS without autoscaling and have updated the docs and resource rules to reflect this.

This PR includes the work to both allow VMSS as well as allow autoscaling (I had originally thought these things one and the same otherwise I would have split it up).  Feel free to dig in and poke holes in the code now :)

rbankole:
when can we expect this to be merged into master?

Krueladin:
@katbyte Is anyone looking at adding this to v11 anytime soon? Thanks!

jlpedrosa:
Hi @katbyte  @Lucretius 

One note. Please keep in mind that this will introduce some issues. Azure API behaves differently if the type of agent pools are AvailabilitySets or VMScaleSets. If the cluster is created with VMScaleSets, the API stops allowing the same request being resent, claiming that any modification to the agent pool must be done through the agent pool API.

`"message": "Node pool update via managed cluster not allowed. Use per nodepool operations."`

So for old clusters it will work as before, but clusters using this feature will fail to update or if the  first request fails to create the cluster (timeout, network error... what ever) terraform will never be able to create the cluster.

PR #3183, opened 2019-04-05
validation for identity_ids

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/3059

`identity_ids` is still optional,  but now if it is specified it must
1) Not be an empty array
2) Each entry must not be a UUID.

~Not sure if there is another option for this field aside from UUIDs - but if not, we can create a stronger validation checking that all entries are UUIDs.~

### DIFF
--- a/azurerm/resource_arm_virtual_machine.go
+++ b/azurerm/resource_arm_virtual_machine.go
@@ -102,8 +102,10 @@ func resourceArmVirtualMachine() *schema.Resource {
 						"identity_ids": {
 							Type:     schema.TypeList,
 							Optional: true,
+							MinItems: 1,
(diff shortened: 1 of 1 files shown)


adamday2:
Is this validating that it **is** a UUID? I was under the impression that this field was of the format:

"/subscriptions/<subscription_id>resourcegroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>"

Lucretius:
Hmm the tests seemed to be checking for a UUID, if this is incorrect I can swap this to a less strict “not empty” check.  Definitely don’t want to break this!

PR #3195, opened 2019-04-05
alter identity_id check to be non-empty

Lucretius:
Fixes a bug I introduced in https://github.com/terraform-providers/terraform-provider-azurerm/pull/3183.  Big thanks for the catch by @adamday2 - I have fixed the validation to just check for non-zero values.  I did not see an existing check for an exact resource ID regex - something to consider at a later date I suppose.


katbyte:
Thanks for the speedy fix @Lucretius, LGTM 👍

PR #3159, opened 2019-04-03
New Resource: `azurerm_data_factory`

mbfrahry:
Adding a new resource `azurerm_data_factory`

This is taking the majority of @hbuckle's work from #1720 and modifying it/updating tests.

```
--- PASS: TestAccAzureRMDataFactory_disappears (67.48s)
--- PASS: TestAccAzureRMDataFactory_tags (72.79s)
--- PASS: TestAccAzureRMDataFactory_basic (72.79s)
--- PASS: TestAccAzureRMDataFactory_identity (72.82s)
--- PASS: TestAccAzureRMDataFactory_github (86.10s)
--- PASS: TestAccAzureRMDataFactory_tagsUpdated (86.10s)
```

Fixes #1720

PR #3205, opened 2019-04-08
New Resource: `azurerm_data_factory_linked_service_sql_server`

mbfrahry:
This PR adds the first linked service for data factory into the azurerm provider. We can move the common pieces of linked services into a umbrella resource as we learn more about what is shared between linked services. 

Dependent on #3159 

```
--- PASS: TestAzureRmDataFactoryLinkedServiceConnectionStringDiff (0.00s)
--- PASS: TestAccAzureRMDataFactoryLinkedServiceSQLServer_basic (94.96s)
```


katbyte:
I'll review this once #3159 is merged and the diff is a tad cleaner

mbfrahry:
@katbyte, this should look a bit cleaner now

azurerm/resource_arm_data_factory_linked_service_sql_server.go
```diff
@@ -0,0 +1,357 @@
...
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+
+			"parameters": {
+				Type:     schema.TypeMap,
+				Optional: true,
+			},
+
+			"annotations": {
+				Type:     schema.TypeList,
```
tombuildsstuff: would this make more sense as a Set than a List?

azurerm/resource_arm_data_factory_linked_service_sql_server.go
```diff
@@ -0,0 +1,357 @@
...
+
+	parameters := flattenDataFactoryLinkedServiceParameters(sqlServer.Parameters)
+	if err := d.Set("parameters", parameters); err != nil {
+		return fmt.Errorf("Error setting `parameters`: %+v", err)
+	}
+
+	if connectVia := sqlServer.ConnectVia; connectVia != nil {
+		if connectVia.ReferenceName != nil {
+			d.Set("integration_runtime_name", *connectVia.ReferenceName)
+		}
```
tombuildsstuff: since this is a string pointer; we can allow d.Set to de-reference this for us:

```
d.Set("integration_runtime_name", connectVia.ReferenceName)
```

azurerm/resource_arm_data_factory_linked_service_sql_server.go
```diff
@@ -0,0 +1,357 @@
...
+	if properties := sqlServer.SQLServerLinkedServiceTypeProperties; properties != nil {
+		if properties.ConnectionString != nil {
+			val, ok := properties.ConnectionString.(string)
+			if !ok {
+				log.Printf("[DEBUG] Skipping connection string %q since it's not a string", val)
+			} else {
+				d.Set("connection_string", properties.ConnectionString.(string))
+			}
+		}
+	}
```
tombuildsstuff: I think it'd be better to make this:

```
connectionString := ""
if properties := sqlServer.SQLServerLinkedServiceTypeProperties; properties != nil {
    if properties.ConnectionString != nil {
    val, ok := properties.ConnectionString.(string)
    if ok {
      connectionString = val
    } else {
      log.Printf("[DEBUG] Skipping connection string %q since it's not a string", val)
    }
  }
}

d.Set("connection_string", connectionString)
```

azurerm/resource_arm_data_factory_linked_service_sql_server.go
```diff
@@ -0,0 +1,357 @@
...
+func azureRmDataFactoryLinkedServiceConnectionStringDiff(k, old string, new string, d *schema.ResourceData) bool {
+	oldSplit := strings.Split(strings.ToLower(old), ";")
+	newSplit := strings.Split(strings.ToLower(new), ";")
+
+	sort.Strings(oldSplit)
+	sort.Strings(newSplit)
+
+	// We need to remove the password from the new string since it isn't returned from the api
+	for i, v := range newSplit {
+		if strings.HasPrefix(v, "password") {
```
tombuildsstuff: IIRC the keys for a connection string can be case insensitive; as such we may want to make this:

```suggestion
		if strings.HasPrefix(strings.ToLower(v), "password") {
```

azurerm/resource_arm_data_factory_linked_service_sql_server.go
```diff
@@ -0,0 +1,357 @@
...
+			newSplit = append(newSplit[:i], newSplit[i+1:]...)
+		}
+	}
+
+	if len(oldSplit) != len(newSplit) {
+		return false
+	}
+
+	// We'll error out if we find any differences between the old and the new connectiong strings
```
tombuildsstuff:
```suggestion
	// We'll error out if we find any differences between the old and the new connection strings
```

azurerm/resource_arm_data_factory_linked_service_sql_server.go
```diff
@@ -0,0 +1,357 @@
...
+		return &datafactory.IntegrationRuntimeReference{
+			ReferenceName: &integrationRuntimeName,
+			Type:          &typeString,
+		}
+	}
+	return nil
+}
+
+func expandDataFactoryLinkedServiceParameters(d *schema.ResourceData) map[string]*datafactory.ParameterSpecification {
+	input := d.Get("parameters").(map[string]interface{})
```
tombuildsstuff: in general we're making this cast prior to these expand methods; so that the schema isn't handled within them; which would make this:

```suggestion
func expandDataFactoryLinkedServiceParameters(input map[string]interface{}) map[string]*datafactory.ParameterSpecification {
```

azurerm/resource_arm_data_factory_linked_service_sql_server.go
```diff
@@ -0,0 +1,357 @@
...
+			Type:         datafactory.ParameterTypeString,
+			DefaultValue: v.(string),
+		}
+	}
+
+	return output
+}
+
+func expandDataFactoryLinkedServiceAdditionalProperties(d *schema.ResourceData) map[string]interface{} {
+	input := d.Get("additional_properties").(map[string]interface{})
```
tombuildsstuff: (as above) we could parse this prior to this method and make this:

```suggestion
func expandDataFactoryLinkedServiceAdditionalProperties(input map[string]interface{}) map[string]interface{} {
```

azurerm/resource_arm_data_factory_linked_service_sql_server.go
```diff
@@ -0,0 +1,357 @@
...
+
+			output[k] = val
+		}
+	}
+
+	return output
+}
+
+func expandDataFactoryLinkedServiceAnnotations(d *schema.ResourceData) *[]interface{} {
+	input := d.Get("annotations").([]interface{})
```
tombuildsstuff: (as above)

azurerm/resource_arm_data_factory_linked_service_sql_server.go
```diff
@@ -0,0 +1,357 @@
...
+	for _, annotation := range input {
+		annotations = append(annotations, annotation.(string))
+	}
+
+	return &annotations
+}
+
+func flattenDataFactoryLinkedServiceAnnotations(input *[]interface{}) []string {
+	annotations := make([]string, 0)
+	if input != nil {
```
tombuildsstuff: rather than nesting this - we'd generally inverse this:

```suggestion
	if input == nil {
	  return annotations
	}
```

website/docs/r/data_factory_linked_service_sql_server.html.markdown
```diff
@@ -0,0 +1,75 @@
+---
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_data_factory_linked_service_sql_server"
+sidebar_current: "docs-azurerm-resource-data-factory-linked-service-sql-server"
+description: |-
+  Manage an Azure Data Factory Linked Service SQL Server.
```
tombuildsstuff: this could make more sense as:

```suggestion
  Manage a Linked Service (connection) between a SQL Server and Azure Data Factory.
```

website/docs/r/data_factory_linked_service_sql_server.html.markdown
```diff
@@ -0,0 +1,75 @@
...
+---
+
+# azurerm_data_factory_linked_service_sql_server
+
+Manage an Azure Data Factory Linked Service SQL Server.
```
tombuildsstuff: (as above)

website/docs/r/data_factory_linked_service_sql_server.html.markdown
```diff
@@ -0,0 +1,75 @@
...
+  name                = "example"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  data_factory_name   = "${azurerm_data_factory.test.name}"
+  connection_string   = "Integrated Security=False;Data Source=test;Initial Catalog=test;User ID=test;Password=test"
+  parameters {
+	  "foo" = "example1"
+  }
+  additional_properties {
+	  "bar" = "example2"
+  }
```
tombuildsstuff: since both of these are optional can we remove these from the example?

website/docs/r/data_factory_linked_service_sql_server.html.markdown
```diff
@@ -0,0 +1,75 @@
...
+sidebar_current: "docs-azurerm-resource-data-factory-linked-service-sql-server"
+description: |-
+  Manage an Azure Data Factory Linked Service SQL Server.
+---
+
+# azurerm_data_factory_linked_service_sql_server
+
+Manage an Azure Data Factory Linked Service SQL Server.
+
+## Example Usage
```
tombuildsstuff: since there's a connection string here we should also include the State file disclaimer:

```
~> **Note:** All arguments including the client secret will be stored in the raw state as plain-text. [Read more about sensitive data in state](/docs/state/sensitive-data.html).
```

website/docs/r/data_factory_linked_service_sql_server.html.markdown
```diff
@@ -0,0 +1,75 @@
...
+
+* `parameters` - (Optional) A map of parameters to associate with the Data Factory Linked Service SQL Server.
+
+* `additional_properties` - (Optional) A map of additional properties to associate with the Data Factory Linked Service SQL Server.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The Data Factory Linked Service SQL Server ID.
```
tombuildsstuff: I think this'd read cleaner as:

```suggestion
* `id` - The ID of the Data Factory Linked Service.
```

azurerm/resource_arm_data_factory_linked_service_sql_server.go
mbfrahry: I didn't have any issues with ordering and annotations. What're you thinking the other benefit of swapping this would be?

azurerm/resource_arm_data_factory_linked_service_sql_server.go
mbfrahry: This was an issue that I took care of a few lines up with:
```
oldSplit := strings.Split(strings.ToLower(old), ";")
newSplit := strings.Split(strings.ToLower(new), ";")
```

azurerm/resource_arm_data_factory_linked_service_sql_server.go
tombuildsstuff: then that's fine - some of the Azure API's are picky about ordering and some aren't, it depends on the API unfortunately 🙃

PR #3206, opened 2019-04-08
Blob metadata

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/1400

- Added support for `metadata` property for storage blobs
- Updated markdown by adding property
- Added metadata to basic test (if desired, I can split this into a *separate* test for metadata, but seeing as metadata is like a tag it looks like we normally just include those in basic tests)

### DIFF
--- a/azurerm/resource_arm_storage_blob.go
+++ b/azurerm/resource_arm_storage_blob.go
@@ -580,7 +580,7 @@ func resourceArmStorageBlobUpdate(d *schema.ResourceData, meta interface{}) erro
 	}
 
 	if d.HasChange("metadata") {
-		blob.Metadata = expandBlobMetadata(d)
+		blob.Metadata = expandStorageAccountBlobMetadata(d)
 
 		opts := &storage.SetBlobMetadataOptions{}
 		if err := blob.SetMetadata(opts); err != nil {
--- a/azurerm/resource_arm_storage_blob.go
+++ b/azurerm/resource_arm_storage_blob.go
@@ -114,6 +114,14 @@ func resourceArmStorageBlob() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validation.IntAtLeast(1),
 			},
+
+			"metadata": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_storage_blob.go
```diff
@@ -201,6 +209,13 @@ func resourceArmStorageBlobCreate(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
+	blob.Metadata = expandBlobMetadata(d)
```
katbyte: As these functions are package wide could we rename this to
```suggestion
	blob.Metadata = expandStorageAccountBlobMetadata(d)
```

azurerm/resource_arm_storage_blob.go
```diff
@@ -564,6 +579,15 @@ func resourceArmStorageBlobUpdate(d *schema.ResourceData, meta interface{}) erro
 		return fmt.Errorf("Error setting properties of blob %s (container %s, storage account %s): %+v", id.blobName, id.containerName, id.storageAccountName, err)
 	}
 
+	if d.HasChange("metadata") {
+		blob.Metadata = expandBlobMetadata(d)
```
katbyte: As these functions are package wide could we rename this to
```suggestion
		blob.Metadata = expandStorageAccountBlobMetadata(d)
```

azurerm/resource_arm_storage_blob.go
```diff
@@ -632,6 +662,7 @@ func resourceArmStorageBlobRead(d *schema.ResourceData, meta interface{}) error
 		log.Printf("[INFO] URL for %q is empty", id.blobName)
 	}
 	d.Set("url", u)
+	d.Set("metadata", flattenBlobMetadata(blob.Metadata))
```
katbyte: and here:
```suggestion
	d.Set("metadata", flattenStorageAccountBlobMetadata(blob.Metadata))
```

azurerm/resource_arm_storage_blob.go
```diff
@@ -114,6 +114,14 @@ func resourceArmStorageBlob() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validation.IntAtLeast(1),
 			},
+
+			"metadata": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
```
katbyte: Should we validate this with `NoEmptyStrings`?

Lucretius: I think that is a good call

Lucretius:
Thanks @katbyte for the feedback, I've renamed the functions as desired.

bpoland:
Thanks everyone!

PR #3207, opened 2019-04-08
Add new Trafficmanager Routing Methods

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/2966

I've gone and bumped the trafficmanager SDK to the version mentioned in the PR.  `make test` ran just fine, we will see how the acceptance tests perform.

The PR mentioned two new routing methods (which are `Subnet` and `MultiValue`); I've added the option to set them both.


website/docs/r/traffic_manager_profile.html.markdown
```diff
@@ -69,6 +69,8 @@ The following arguments are supported:
     - `Performance` - Traffic is routed via the User's closest Endpoint
     - `Weighted` - Traffic is spread across Endpoints proportional to their `weight` value.
     - `Priority` - Traffic is routed to the Endpoint with the lowest `priority` value.
+    - `MultiValue`- All healthy Endpoints are returned.  MultiValue routing method works only if all the endpoints of type ‘External’ and are specified as IPv4 or IPv6 addresses.
```
katbyte: These should probably be alphabetically sorted

PR #3219, opened 2019-04-09
remove "Computed" from schema

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/3146

From the discussion I had in the issue, it looks like this feature can be updated after the resource is spun up.  I believe that since it is `Computed` in the schema, updates to the existing resource are ignored.  This helper function is used by both the `app_service` and `app_service_slot` resources.  The default will be `false`.  If the app service requires a 32 bit resource the effect will be that this needs to be specified.


tombuildsstuff:
Tests pass - thanks for this @Lucretius 👍 

Screenshot 2019-04-16 at 13 56 02

PR #3236, opened 2019-04-12
New Resource: `azurerm_data_factory_dataset_sql_server_table`

mbfrahry:
This PR adds `azurerm_data_factory_dataset_sql_server_table` to the Data Factory collection and starts to consolidate some of the methods used between the resources into a single source under `datafactory.go`

```
--- PASS: TestAccAzureRMDataFactoryDatasetSQLServerTable_basic (94.35s)
```


azurerm/data_factory.go
```diff
@@ -0,0 +1,74 @@
...
+		output[k] = &datafactory.ParameterSpecification{
+			Type:         datafactory.ParameterTypeString,
+			DefaultValue: v.(string),
+		}
+	}
+
+	return output
+}
+
+func expandDataFactoryAdditionalProperties(input map[string]interface{}) map[string]interface{} {
```
katbyte: Do we really need this? 🤔  should you be able to use the source map without copying it to a new one?

azurerm/data_factory.go
```diff
@@ -0,0 +1,74 @@
...
+			}
+
+			output[k] = val
+		}
+	}
+
+	return output
+}
+
+func expandDataFactoryAnnotations(input []interface{}) *[]interface{} {
```
katbyte: and here just take a pointer of the source slice?

azurerm/data_factory.go
```diff
@@ -0,0 +1,74 @@
...
+	annotations := make([]interface{}, 0)
+
+	for _, annotation := range input {
+		annotations = append(annotations, annotation.(string))
+	}
+
+	return &annotations
+}
+
+func flattenDataFactoryAnnotations(input *[]interface{}) []string {
```
katbyte: I think you could use `utils.FlattenStringArray` instead of this?

azurerm/resource_arm_data_factory_dataset_sql_server_table_test.go
```diff
@@ -0,0 +1,219 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+)
+
+func TestAccAzureRMDataFactoryDatasetSQLServerTable_basic(t *testing.T) {
```
katbyte: I think this should be:
```suggestion
func TestAccAzureRMDataFactoryDatasetSQLServerTable_update(t *testing.T) {
```

katbyte: Could we create a basic test that is a single step with the bare minimum properties required?

azurerm/resource_arm_data_factory_dataset_sql_server_table_test.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+resource "azurerm_data_factory_dataset_sql_server_table" "test" {
+  name                = "acctestds%d"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  data_factory_name   = "${azurerm_data_factory.test.name}"
+  linked_service_name = "${azurerm_data_factory_linked_service_sql_server.test.name}"
+ 
+  description = "test description"
+	annotations = ["test1", "test2", "test3"]
```
katbyte: alignment
```suggestion
  annotations = ["test1", "test2", "test3"]
```

azurerm/resource_arm_data_factory_dataset_sql_server_table_test.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+resource "azurerm_data_factory_dataset_sql_server_table" "test" {
+  name                = "acctestds%d"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  data_factory_name   = "${azurerm_data_factory.test.name}"
+  linked_service_name = "${azurerm_data_factory_linked_service_sql_server.test.name}"
+ 
+  description = "test description 2"
+	annotations = ["test1", "test2"]
```
katbyte:
```suggestion
  annotations = ["test1", "test2"]
```

azurerm/resource_arm_data_factory_dataset_sql_server_table_test.go
```diff
@@ -0,0 +1,219 @@
...
+  table_name  = "testTable"
+  folder      = "testFolder"
+
+  parameters {
+		"foo" = "test1"
```
katbyte:
```suggestion
    "foo" = "test1"
```

azurerm/resource_arm_data_factory_dataset_sql_server_table_test.go
```diff
@@ -0,0 +1,219 @@
...
+		"bar" = "test2"
```
katbyte:
```suggestion
    "bar" = "test2"
```

katbyte (approved):
Thanks for the PR @mbfrahry, aside from a few comments i've left inline this LGTM 👍

azurerm/data_factory.go
mbfrahry: There needs to be a facepalm emoji in github

azurerm/data_factory.go
mbfrahry: Weirdly enough, I think it'd be a call to utils.ExpandStringArray

mbfrahry: On second thought. I think it's needed because it's an *[]interface{} and we'd need to do some finagling to get it to a []string

katbyte: Ahh yea sorry, its so close to the signature, and as much as i'd want to ask if terraform would accept the interface array i know we should check they are all strings.

mbfrahry: No worries!

PR #3242, opened 2019-04-14
batch tests: fix resource group names

katbyte:


azurerm/data_source_batch_account_test.go
```diff
@@ -60,7 +60,7 @@ func TestAccDataSourceAzureRMBatchAccount_complete(t *testing.T) {
 func testAccDataSourceAzureRMBatchAccount_basic(rInt int, rString string, location string) string {
 	return fmt.Sprintf(`
 resource "azurerm_resource_group" "test" {
-  name     = "testaccbatch%d"
+  name     = "testaccRG-%d-batch"
```
tombuildsstuff: if we're changing these, could we make these `acctest` so the Dalek will pick them up?

katbyte: 🤦‍♀️ That was my intent but i apparently only got halfway there hah

PR #3246, opened 2019-04-15
Update subscription.html.markdown

KbLev:
Adding example for return attribute (Id).

### DIFF
--- a/website/docs/d/subscription.html.markdown
+++ b/website/docs/d/subscription.html.markdown
@@ -26,7 +26,7 @@ output "current_subscription_display_name" {
 
 ## Attributes Reference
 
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hey @KbLev 

Thanks for this PR :)

Within the Azure Provider we (almost always) use the `id` attribute to refer to the Resource ID of the resource within Azure - and such this is implied by convention. In the future we plan to add a glossary containing some of these common terms so it's clearer (and link back) - however we plan to do this across all resources so that it's consistent. As such whilst I'd like to thank you for this contribution, rather than diverging from the other resources here I'm going to close this PR for the moment.

Thanks!

PR #3253, opened 2019-04-16
Enable correlation request ID header in ARM API.

metacpp:
This PR is created to pass unique client generated correlation request id to HTTP extension header,  `x-ms-correlation-request-id`, which can be used to correlate different ARM HTTP requests from same client session.


paultyng:
We should hold merging this until 1.25 ships.

PR #3255, opened 2019-04-16
Storage Account: Enable DefaultAction in Network rules.

thatInfrastructureGuy:
**Introduction:** DefaultAction is the toggle which enables/disables network rules in azure storage account. 
Setting it to `Deny` enables storage account firewall settings and `Allow` disables the firewalls.

**Problem:** When using azure storage account as module with `network_rules` block, there is no good way to disable network security apart from removing the `network_rules` block entirely from the root module. 

**PR objective:** This PR aims to solve this problem by implementing DefaultAction variable. 

Storage Account - *Module* with `network_rules` block: 

```
 network_rules {
    default_action             = "${var.default_action}"
    virtual_network_subnet_ids = ["${var.virtual_network_subnet_ids}"]
    bypass                     = ["${var.bypass}"]
    ip_rules                   = ["${var.ip_rules}"]
  }
```

Storage Account - *Deployment 1* with no firewall:
```
 network_rules {
    default_action             = "Allow"
    virtual_network_subnet_ids = ""
    bypass                     = ""
    ip_rules                   = ""
  }
```

Storage Account - *Deployment 2* with firewall:
```
 network_rules {
    virtual_network_subnet_ids = "${azurerm_subnet.test.id"
    bypass                     = ["Logging", "Metrics"]
    ip_rules                   = ["127.0.0.1", "127.0.0.2"]
  }
```


azurerm/resource_arm_storage_account.go
```diff
@@ -185,6 +185,15 @@ func resourceArmStorageAccount() *schema.Resource {
 							Elem:     &schema.Schema{Type: schema.TypeString},
 							Set:      schema.HashString,
 						},
+						"default_action": {
+							Type:     schema.TypeString,
+							Optional: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(storage.DefaultActionAllow),
+								string(storage.DefaultActionDeny),
+							}, true),
```
tombuildsstuff: we're making all new fields in the provider case-sensitive - as such can we make this:

```suggestion
							}, false),
```

azurerm/resource_arm_storage_account.go
```diff
@@ -185,6 +185,15 @@ func resourceArmStorageAccount() *schema.Resource {
...
+							Default: string(storage.DefaultActionAllow),
```
tombuildsstuff: from a security perspective this feels like a default that could trip people up - it could be better to make this Required and let users intentionally select it?

azurerm/resource_arm_storage_account.go
```diff
@@ -909,6 +918,11 @@ func expandStorageAccountNetworkRules(d *schema.ResourceData) *storage.NetworkRu
 	networkRule := networkRules[0].(map[string]interface{})
 	networkRuleSet := &storage.NetworkRuleSet{}
 
+	if networkRule["default_action"].(string) == "Allow" {
```
tombuildsstuff: we should probably make this:

```suggestion
	if networkRule["default_action"].(string) == string(storage.DefaultActionAllow) {
```

thatInfrastructureGuy:
@tombuildsstuff Thanks for quick review. :) I have made the requested changes. I do agree `default_action` should be required field and having to let users decide on the value of it.

thatInfrastructureGuy:
```
        FAIL: TestAccAzureRMStorageAccount_networkRulesDeleted (150.14s)
    testing.go:538: Step 1 error: After applying this step, the plan was not empty:

        DIFF:

        UPDATE: azurerm_storage_account.testsa
          network_rules.#:                                       "1" => "0"
          network_rules.0.default_action:                        "Allow" => ""
          network_rules.0.ip_rules.#:                            "1" => "0"
          network_rules.0.ip_rules.3619153832:                   "127.0.0.1" => ""
          network_rules.0.virtual_network_subnet_ids.#:          "1" => "0"
          network_rules.0.virtual_network_subnet_ids.1226267014: "/subscriptions/xxx/r
esourceGroups/acctestAzureRMSA-190416064647690939/providers/Microsoft.Network/virtualNetworks/acctestvirtnet19041606464
7690939/subnets/acctestsubnet190416064647690939" => ""
                                                                                                      
        STATE:

        azurerm_resource_group.testrg:
          network_rules.# = 1
          network_rules.0.bypass.# = 1
          network_rules.0.bypass.625562768 = AzureServices
          network_rules.0.default_action = Allow
          network_rules.0.ip_rules.# = 1
          network_rules.0.ip_rules.3619153832 = 127.0.0.1
          network_rules.0.virtual_network_subnet_ids.# = 1
          network_rules.0.virtual_network_subnet_ids.1226267014 = /subscriptions/xxx/r
esourceGroups/acctestAzureRMSA-190416064647690939/providers/Microsoft.Network/virtualNetworks/acctestvirtnet19041606464
7690939/subnets/acctestsubnet190416064647690939
```

I believe this happens because the rules itself are not deleted. Only the flag `"defaultAction": "Allow"` is set on the backend.

tombuildsstuff:
@thatInfrastructureGuy 

> I believe this happens because the rules itself are not deleted. Only the flag "defaultAction": "Allow" is set on the backend.

Indeed - [because the NetworkRuleSet created when `Allow` is set](https://github.com/terraform-providers/terraform-provider-azurerm/pull/3255/files#diff-f4d2502d01b7fb77deb7bb221025fb51R922) doesn't contain any IP Rules/Virtual Network Rules, these aren't being overridden. If an empty list is specified for those then this should pass 👍

thatInfrastructureGuy:
@tombuildsstuff Thanks for the pointer. However, I am not able to  override those IP Rules/Virtual Network Rules even after specifying empty list. I tried the below snippet: 

```
return &storage.NetworkRuleSet{
			DefaultAction:       storage.DefaultActionAllow,
			Bypass:              storage.Bypass(""),
			IPRules:             &[]storage.IPRule{},
			VirtualNetworkRules: &[]storage.VirtualNetworkRule{},
		}
```

I dont know what I am missing. Can you please guide me here?

azurerm/resource_arm_storage_account.go
```diff
@@ -909,6 +917,11 @@ func expandStorageAccountNetworkRules(d *schema.ResourceData) *storage.NetworkRu
 	networkRule := networkRules[0].(map[string]interface{})
 	networkRuleSet := &storage.NetworkRuleSet{}
 
+	if networkRule["default_action"].(string) == string(storage.DefaultActionAllow) {
+		// If Default Access is enabled then no network rules are set.
```
katbyte: Is this required? will the API be displeased if you still provide them?

ashishkulk: API is not displeased if we provide network rules.  However, I am not able to override those IP Rules/Virtual Network Rules even after specifying empty list. 

```
if networkRule["default_action"].(string) == string(storage.DefaultActionAllow) {
		// If Default Access is enabled then no network rules are set.
		return &storage.NetworkRuleSet{
			DefaultAction:       storage.DefaultActionAllow,
			Bypass:              storage.Bypass(""),
			IPRules:             &[]storage.IPRule{},
			VirtualNetworkRules: &[]storage.VirtualNetworkRule{},
		}
	}
```
Also, API creates random key value pair which is hard to determine. Eg:

```
          network_rules.0.bypass.625562768 = AzureServices
          network_rules.0.default_action = Allow
          network_rules.0.ip_rules.# = 1
          network_rules.0.ip_rules.3619153832 = 127.0.0.1
```
Hence, I cant unset `625562768` or `3619153832` .

azurerm/resource_arm_storage_account.go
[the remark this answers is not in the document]
ashishkulk: @katbyte Forgive me because I dont have much knowledge regarding enums in golang. I think you are asking if we can use enums so that we dont have to do string comparison as enum comparison (int) is much faster. Correct me if I am wrong, from what I gather we dont have a way to specify enums.

https://github.com/Azure/azure-sdk-for-go/blob/master/services/storage/mgmt/2018-02-01/storage/models.go#L118

ashishkulk:
Added response in the comments.

azurerm/resource_arm_storage_account.go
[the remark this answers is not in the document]
katbyte: what i mean is convert the constant to a string and use that instead of hard coding the value as you have, enum was probably not the best terminology to use!

azurerm/resource_arm_storage_account.go
katbyte: Ahh so if you specify a default action azure auto populates that property with some values, could you potentially mark it as computed if that is the case?

azurerm/resource_arm_storage_account.go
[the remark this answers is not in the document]
thatInfrastructureGuy: Understood. I believe you were looking at old commit. I did get same feedback from Tom and [it was updated to constant](
https://github.com/terraform-providers/terraform-provider-azurerm/pull/3255/commits/15d95645b0b675618fbc5dddcfe1a34c7caa33a2).

azurerm/resource_arm_storage_account.go
thatInfrastructureGuy: Done. Marked variables as computed. Also added test to reflect the changes.

katbyte: You are still returning an object with just the default action here.

What i was getting at is not returning an object here, setting DefaultAction and all the other properties regardless of what default action is.

katbyte:
@thatInfrastructureGuy,

See my comment left inline. I am still wondering why you are conditionally setting the properties on create/update. Does the API throw an error if you set them?

katbyte:
👋 @thatInfrastructureGuy,

I hope you don't mind but i've pushed a commit demonstrating what I was referring to. I also changed the property to optional as a new required property in an existing block could potentially break peoples configuration.

thatInfrastructureGuy:
@katbyte Sorry for late response. I was on vacation for a bit and never really got back to this PR. Thanks so much for review and the commit. I am totally good with it.

PR #3311, opened 2019-04-26
IC for adding container_configuration to batch pool resource.

tshauck:
Hi -- this PR adds the `container_configuration` block to the batch pool resource.

This setting is necessary for the Pool to support containerized workloads. There are more settings not included in this PR, e.g. setting a default registry. I'd like to add that at some point, but this seemed like a good intermediate step since this is my first PR and if the docker is enabled then the Tasks can still set the registry or other items.

### DIFF
--- a/azurerm/data_source_batch_pool.go
+++ b/azurerm/data_source_batch_pool.go
@@ -112,14 +112,11 @@ func dataSourceArmBatchPool() *schema.Resource {
 			},
 			"container_configuration": {
 				Type:     schema.TypeList,
-				Optional: true,
-				MaxItems: 1,
+                Computed: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"type": {
 							Type:         schema.TypeString,
-							Optional:     true,
-							ValidateFunc: validate.NoEmptyStrings,
 						},
 					},
 				},
--- a/azurerm/data_source_batch_pool.go
+++ b/azurerm/data_source_batch_pool.go
@@ -110,6 +110,20 @@ func dataSourceArmBatchPool() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"container_configuration": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validate.NoEmptyStrings,
+						},
+					},
+				},
+			},
 			"certificate": {
 				Type:     schema.TypeList,
 				Optional: true,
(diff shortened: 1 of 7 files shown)


azurerm/helpers/azure/batch_pool.go
```diff
@@ -214,6 +230,22 @@ func ExpandBatchPoolImageReference(list []interface{}) (*batch.ImageReference, e
 	return imageRef, nil
 }
 
+// ExpandBatchPoolContainerConfiguration expands the Batch pool container configuration
+func ExpandBatchPoolContainerConfiguration(list []interface{}) (*batch.ContainerConfiguration, error) {
+	if len(list) == 0 {
+		return nil, fmt.Errorf("Error: container configuration should be defined")
```
katbyte: We should just return nil here? i don't think this is an actual error

website/docs/d/batch_pool.html.markdown
```diff
@@ -47,6 +47,8 @@ The following attributes are exported:
```
katbyte: I think you missed adding the resource documentation?

azurerm/resource_arm_batch_pool.go
```diff
@@ -567,6 +588,15 @@ func resourceArmBatchPoolRead(d *schema.ResourceData, meta interface{}) error {
 			d.Set("node_agent_sku_id", props.DeploymentConfiguration.VirtualMachineConfiguration.NodeAgentSkuID)
 		}
 
+		if props.DeploymentConfiguration != nil &&
```
katbyte: Could we change this to:
```go
		if dcfg := props.DeploymentConfiguration; dcfg != nil
			if vmcfg := dcfg.VirtualMachineConfiguration; vmcfg != nil {
				d.Set("container_configuration", azure.FlattenBatchPoolContainerConfiguration(vmcfg. ContainerConfiguration)) //this function should check if its nil and return nothing if it is
```

azurerm/data_source_batch_pool.go
```diff
@@ -274,6 +288,17 @@ func dataSourceArmBatchPoolRead(d *schema.ResourceData, meta interface{}) error
 			}
 		}
 
+		if props.DeploymentConfiguration != nil &&
```
katbyte: Same comment as below in the resource. we should nest a bunch of assignments and if's

azurerm/data_source_batch_pool.go
```diff
@@ -110,6 +110,20 @@ func dataSourceArmBatchPool() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"container_configuration": {
```
katbyte: This is a data source so this should all be computed, not optional and needs no validatino or max items.

azurerm/data_source_batch_pool.go
tshauck: Cool, thanks for the comment, looks much batter. I made the update and slightly refactored the later code since it also uses `dcfg.VirtualMachineConfiguration`.

azurerm/helpers/azure/batch_pool.go
```diff
@@ -191,6 +191,22 @@ func FlattenBatchPoolCertificateReferences(armCertificates *[]batch.CertificateR
...
+
+	if armContainerConfiguration == nil {
+		return nil
+	}
+
+	if armContainerConfiguration.Type != nil {
+		result["type"] = *armContainerConfiguration.Type
+	}
+
+	return result
```
katbyte: I think this needs to be:
```suggestion
	return []interface{}{result}
```
to fix:

```
------- Stdout: -------
=== RUN   TestAccDataSourceAzureRMBatchPool_complete
=== PAUSE TestAccDataSourceAzureRMBatchPool_complete
=== CONT  TestAccDataSourceAzureRMBatchPool_complete

------- Stderr: -------
panic: container_configuration: '': source data must be an array or slice, got map
```

website/docs/r/batch_pool.html.markdown
```diff
@@ -128,6 +132,8 @@ The following arguments are supported:
 
 * `certificate` - (Optional) One or more `certificate` blocks that describe the certificates to be installed on each compute node in the pool.
 
+* `container_configuration` - (Optional) The container configuration used in the pool's VMs.
```
katbyte: Looks like your missing the type attribute description here

tshauck:
@katbyte Thanks for the comments! I think I've made the changes in the last two commits: https://github.com/terraform-providers/terraform-provider-azurerm/compare/75879f6..f925fdc. I rebased off master as I was getting some errors I think were resolved later commits.

PR #3318, opened 2019-04-26
bugfix: azurerm_stream_analytics_output_eventhub does not set the output format

maxbog:


azurerm/resource_arm_stream_analytics_output_eventhub_test.go
```diff
@@ -97,6 +97,35 @@ func TestAccAzureRMStreamAnalyticsOutputEventHub_json(t *testing.T) {
...
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMStreamAnalyticsOutputEventHubDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMStreamAnalyticsOutputEventHub_jsonArrayFormat(ri, location),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMStreamAnalyticsOutputEventHubExists(resourceName),
+				),
```
katbyte: We should probably add some checks here:
```go
resource.TestCheckResourceAttr(resourceName, "serialization.0.format", "Array"),
```

and for the rest of the serialization block

maxbog: done here and for all other types of serialization

PR #3323, opened 2019-04-28
test cfg: acctestrg -> acctestRG

katbyte:
For consistency with existing test tf


katbyte:
Split off the non resource group ones into #3326 as some tests now fail

PR #3333, opened 2019-04-29
Fix: Ignore case difference in Resource Group Name for Data Factory resources

mbfrahry:
This PR addresses is a band aid to a bug in the api that returns lower case resource group names. 

That issue can be followed at https://github.com/Azure/azure-rest-api-specs/issues/5788


katbyte:
merged before review click worked, LGTM 👍

PR #3343, opened 2019-04-30
User Assigned Identity: add corresponding data source

logachev:
- Created data source for User Assigned Identity
- Added relevant tests
- Updated documentation


website/docs/d/user_assigned_identity.html.markdown
```diff
@@ -0,0 +1,42 @@
...
+
+output "uai_principal_id" {
+  value = "${data.azurerm_user_assigned_identity.test.principal_id}"
+}
+```
+
+## Argument Reference
+
+* `name` - (Required) Specifies the name of the user assigned identity.
+* `resource_group_name` - (Required) Specifies the name of the resource group.
```
katbyte: Could we use wording consistant with the rest of the provider:
```suggestion
* `resource_group_name` - (Required) The Name of the Resource Group in which the User Assigned Identity exists.
```

website/docs/d/user_assigned_identity.html.markdown
```diff
@@ -0,0 +1,42 @@
...
+
+## Attributes Reference
```
katbyte: I think you missed adding location to the documentation?

website/docs/d/user_assigned_identity.html.markdown
logachev: It depends which resource you use as a reference :) (I think I was looking at virtual machine & virtual network)

azurerm/data_source_user_assigned_identity_test.go
```diff
@@ -0,0 +1,90 @@
...
+					resource.TestCheckResourceAttr(dataSourceName, "tags.%", "1"),
+					testEqualResourceAttr(dataSourceName, resourceName, "principal_id"),
+					testEqualResourceAttr(dataSourceName, resourceName, "client_id"),
+				),
+			},
+		},
+	})
+}
+
+func testEqualResourceAttr(dataSourceName string, resourceName string, attrName string) resource.TestCheckFunc {
```
logachev: I noticed that in other data source tests there is no any checks that data source resource contains same information as a resource.
Do you think this is valuable to move this function to some shared place and reuse in other data source tests?

Issue #1536, opened 2018-07-11
Feature Request: Full sslPolicy support for azurerm_application_gateway


cdhunt:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It looks like a single property (`disabledSslProtocols`) of the sslPolicy configuration object was added to the resource. We would like access to the full object.

### New or Affected Resource(s)

* azurerm_application_gateway

### Potential Terraform Configuration

```hcl
  ssl_policy {
    disabled_ssl_protocols = list
    policy_type = string
    policy_name = string
    cipher_suites = list
    min_protocol_version = list
  }
```

### References

* https://docs.microsoft.com/en-us/azure/templates/microsoft.network/applicationgateways#applicationgatewaysslpolicy-object

stack72:
@katbyte / @tombuildsstuff is this a duplicate of this? https://github.com/terraform-providers/terraform-provider-azurerm/issues/619

tombuildsstuff:
Duplicate of #619 - thanks for pointing that out @stack72 :)

cdhunt:
There's no mention of sslPolicy in #619. Will that be included in that work?

tombuildsstuff:
Reopening this to split this out from #1576

cdhunt:
Are there still blockers for this or is it now technically possible?

tombuildsstuff:
@cdhunt I believe it should be possible for someone to implement this once #2054 is merged - it requires a little thought around deprecating the existing `disabled_ssl_protocols` block (as we'll want to move it into the new `ssl_policy` block) probably as a part of 2.0

cdhunt:
Yeah, dealing with the existing `disabled_ssl_protocols` block without a breaking change was my concern. Thanks.

bs-matil:
I will soon file a PR regards this topic. I feel our Implementation which was done a while ago before this discussion may not fully fit. But I will adopt it in any way needed to get it upstream. See ya later :).

jstewart612:
[...]

PR #3360, opened 2019-05-02
application gateway: add advanced ssl_policies

bs-matil:
This PR should fix the Issue #1536
Feature Request: Full sslPolicy support for azurerm_application_gateway 

It adds support for advanced ssl_polcies. and deprecates the old options

### DIFF
--- a/azurerm/resource_arm_application_gateway.go
+++ b/azurerm/resource_arm_application_gateway.go
@@ -14,6 +14,15 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
+// See https://github.com/Azure/azure-sdk-for-go/blob/master/services/network/mgmt/2018-04-01/network/models.go
+func possibleArmApplicationGatewaySslCipherSuiteValues() []string {
+	cipherSuites := make([]string, 0)
+	for _, cipherSuite := range network.PossibleApplicationGatewaySslCipherSuiteValues() {
+		cipherSuites = append(cipherSuites, string(cipherSuite))
+	}
+	return cipherSuites
+}
+
 func resourceArmApplicationGateway() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceArmApplicationGatewayCreateUpdate,
@@ -635,8 +644,9 @@ func resourceArmApplicationGateway() *schema.Resource {
 
 			// TODO: @tombuildsstuff deprecate this in favour of a full `ssl_protocol` block in the future
 			"disabled_ssl_protocols": {
-				Type:     schema.TypeList,
-				Optional: true,
+				Type:       schema.TypeList,
+				Optional:   true,
+				Deprecated: "has been replaced by `ssl_policy`.`disabled_protocols`",
 				Elem: &schema.Schema{
 					Type:             schema.TypeString,
 					DiffSuppressFunc: suppress.CaseDifference,
@@ -645,6 +655,66 @@ func resourceArmApplicationGateway() *schema.Resource {
 						string(network.TLSv11),
 						string(network.TLSv12),
 					}, true),
+				},
+			},
+
+			"ssl_policy": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"disabled_protocols": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Schema{
+								Type:             schema.TypeString,
+								DiffSuppressFunc: suppress.CaseDifference,
+								ValidateFunc: validation.StringInSlice([]string{
+									string(network.TLSv10),
+									string(network.TLSv11),
+									string(network.TLSv12),
+								}, true),
+							},
+						},
+
+						"policy_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+								ValidateFunc: validation.StringInSlice([]string{
+									string(network.Custom),
+									string(network.Predefined),
+								}, true),
+							},
+						},
+
+						"policy_name": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+
+						"cipher_suites": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Schema{
+								Type:             schema.TypeString,
+								DiffSuppressFunc: suppress.CaseDifference,
+								ValidateFunc:     validation.StringInSlice(possibleArmApplicationGatewaySslCipherSuiteValues(), true),
+							},
+						},
+
+						"min_protocol_version": {
+							Type:             schema.TypeString,
+							Optional:         true,
+							DiffSuppressFunc: suppress.CaseDifference,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(network.TLSv10),
+								string(network.TLSv11),
+								string(network.TLSv12),
+							}, true),
+						},
+					},
 				},
 			},
 
@@ -1228,6 +1298,10 @@ func resourceArmApplicationGatewayRead(d *schema.ResourceData, meta interface{})
 			return fmt.Errorf("Error setting `disabled_ssl_protocols`: %+v", setErr)
 		}
 
+		if setErr := d.Set("ssl_policy", flattenApplicationGatewaySslPolicy(props.SslPolicy)); setErr != nil {
+			return fmt.Errorf("Error setting `ssl_policy`: %+v", setErr)
+		}
+
 		d.Set("enable_http2", props.EnableHTTP2)
 
 		httpListeners, err := flattenApplicationGatewayHTTPListeners(props.HTTPListeners)
@@ -1661,16 +1735,86 @@ func flattenApplicationGatewayConnectionDraining(input *network.ApplicationGatew
 }
 
 func expandApplicationGatewaySslPolicy(d *schema.ResourceData) *network.ApplicationGatewaySslPolicy {
-	vs := d.Get("disabled_ssl_protocols").([]interface{})
-	results := make([]network.ApplicationGatewaySslProtocol, 0)
+	policy := network.ApplicationGatewaySslPolicy{}
+	vs := d.Get("ssl_policy").([]interface{})
+	if len(vs) == 0 {
+		return &policy
+	}
+	v := vs[0].(map[string]interface{})
 
-	for _, v := range vs {
-		results = append(results, network.ApplicationGatewaySslProtocol(v.(string)))
+	disabledSSLPolicies := make([]network.ApplicationGatewaySslProtocol, 0)
+	for _, policy := range v["disabled_protocols"].([]interface{}) {
+		disabledSSLPolicies = append(disabledSSLPolicies, network.ApplicationGatewaySslProtocol(policy.(string)))
+	}
+	if len(disabledSSLPolicies) == 0 {
+		for _, policy := range d.Get("disabled_ssl_protocols").([]interface{}) {
+			disabledSSLPolicies = append(disabledSSLPolicies, network.ApplicationGatewaySslProtocol(policy.(string)))
+		}
 	}
 
-	return &network.ApplicationGatewaySslPolicy{
-		DisabledSslProtocols: &results,
+	if len(disabledSSLPolicies) > 0 {
+		policy = network.ApplicationGatewaySslPolicy{
+			DisabledSslProtocols: &disabledSSLPolicies,
+		}
+	} else {
+		policyType := network.ApplicationGatewaySslPolicyType(v["policy_type"].(string))
+
+		if policyType == network.Predefined {
+			policyName := network.ApplicationGatewaySslPolicyName(v["policy_name"].(string))
+
+			policy = network.ApplicationGatewaySslPolicy{
+				PolicyType: policyType,
+				PolicyName: policyName,
+			}
+		} else if policyType == network.Custom {
+			minProtocolVersion := network.ApplicationGatewaySslProtocol(v["min_protocol_version"].(string))
+
+			cipherSuites := make([]network.ApplicationGatewaySslCipherSuite, 0)
+			for _, cipherSuite := range v["cipher_suites"].([]interface{}) {
+				cipherSuites = append(cipherSuites, network.ApplicationGatewaySslCipherSuite(cipherSuite.(string)))
+			}
+
+			policy = network.ApplicationGatewaySslPolicy{
+				PolicyType:         policyType,
+				MinProtocolVersion: minProtocolVersion,
+				CipherSuites:       &cipherSuites,
+			}
+		}
 	}
+
+	return &policy
+}
+
+func flattenApplicationGatewaySslPolicy(input *network.ApplicationGatewaySslPolicy) []interface{} {
+	results := make([]interface{}, 0)
+
+	if input == nil {
+		return results
+	}
+
+	output := map[string]interface{}{}
+	output["policy_name"] = input.PolicyName
+	output["policy_type"] = input.PolicyType
+	output["min_protocol_version"] = input.MinProtocolVersion
+
+	if input.CipherSuites != nil {
+		cipherSuites := make([]interface{}, 0)
+		for _, v := range *input.CipherSuites {
+			cipherSuites = append(cipherSuites, string(v))
+		}
+		output["cipher_suites"] = cipherSuites
+	}
+
+	if input.DisabledSslProtocols != nil {
+		disabledSslProtocols := make([]interface{}, 0)
+		for _, v := range *input.DisabledSslProtocols {
+			disabledSslProtocols = append(disabledSslProtocols, string(v))
+		}
+		output["disabled_protocols"] = disabledSslProtocols
+	}
+
+	results = append(results, output)
+	return results
 }
 
 func flattenApplicationGatewayDisabledSSLProtocols(input *network.ApplicationGatewaySslPolicy) []interface{} {
--- a/azurerm/resource_arm_application_gateway.go
+++ b/azurerm/resource_arm_application_gateway.go
@@ -642,7 +642,7 @@ func resourceArmApplicationGateway() *schema.Resource {
 				},
 			},
 
-			// TODO: @tombuildsstuff deprecate this in favour of a full `ssl_protocol` block in the future
+			// TODO: remove in 2.0
 			"disabled_ssl_protocols": {
 				Type:       schema.TypeList,
 				Optional:   true,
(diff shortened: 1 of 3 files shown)


website/docs/r/application_gateway.html.markdown
```diff
@@ -379,6 +381,28 @@ A `url_path_map` block supports the following:
...
+---
+
+A `ssl_policy` block supports the following:
+
+* `policy_type` - (Optional) The Type of the Policy. Required if `policy_name` is set. Possible values are `Predefined` and `Custom`. Not compatible with `disabled_protocols`.
+
+For `policy_type`=`Predefined`:
+
+* `policy_name` - (Optional) The Name of the Policy e.g AppGwSslPolicy20170401S. Required if `policy_type` is set to `Predefined`. Possible values can chang over time and 
+are published here https://docs.microsoft.com/de-de/azure/application-gateway/application-gateway-ssl-policy-overview. Not compatible with `disabled_protocols`.
```
jstewart612: The English version if that helps https://docs.microsoft.com/en-us/azure/application-gateway/application-gateway-ssl-policy-overview

bs-matil: @jstewart612 thanks. How could that happen :P.

bs-matil:
unfortunately a tests of mine is currently broken after rebase. I'm on it. please anywas consider to check the code for its general functionallity

bs-matil:
I found the issue. Its caused due to the fact that disable_protocols now is deprecated but still exists as option. I resolved that by creating a new ssl_policy block in parallel. 
The problem here is that this does not work as both states need to be out of sync by definition. I handled this now as it has been done for fqdn and backend_ip addresses. There both blocks are marked as computed so inconsistencies are basically ignored. Actually I don't like this either. Is there a better known solution to such issues?

website/docs/r/application_gateway.html.markdown
jstewart612: It is a genuine mystery to one and all!  ;)

In all seriousness, I don't think this provider's documentation is localized to any other languages, and thus, I think most are, perhaps unfairly, expecting off-world links to go to sites in the same language.

It's the most minor of minor points and really anyone who knows enough how to use this ought know how to go find the English version.... but yeah.

bs-matil: @jstewart612 anyways fixed it :) 👍

azurerm/resource_arm_application_gateway.go
```diff
@@ -630,6 +641,67 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"disabled_protocols": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Computed: true,
+							Elem: &schema.Schema{
+								Type:             schema.TypeString,
+								DiffSuppressFunc: suppress.CaseDifference,
```
tombuildsstuff: since this is a new field can we make this case sensitive by removing this?

```suggestion
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -630,6 +641,67 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+								ValidateFunc: validation.StringInSlice([]string{
+									string(network.TLSv10),
+									string(network.TLSv11),
+									string(network.TLSv12),
+								}, true),
```
tombuildsstuff: since this is a new field can we make this case sensitive by making this false?

```suggestion
								}, false),
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -630,6 +641,67 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+
+						"policy_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+								ValidateFunc: validation.StringInSlice([]string{
+									string(network.Custom),
+									string(network.Predefined),
+								}, true),
```
tombuildsstuff: since this is a new field can we make this case sensitive by making this false?

```suggestion
								}, false),
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -630,6 +641,67 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+
+						"cipher_suites": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Schema{
+								Type:             schema.TypeString,
+								DiffSuppressFunc: suppress.CaseDifference,
```
tombuildsstuff: since this is a new field can we make this case sensitive by removing this?

```suggestion
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -630,6 +641,67 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+								ValidateFunc:     validation.StringInSlice(possibleArmApplicationGatewaySslCipherSuiteValues(), true),
```
tombuildsstuff: since this is a new field can we make this case sensitive by making this false?

```suggestion
								ValidateFunc:     validation.StringInSlice(possibleArmApplicationGatewaySslCipherSuiteValues(), false),
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -630,6 +641,67 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+							},
+						},
+
+						"min_protocol_version": {
+							Type:             schema.TypeString,
+							Optional:         true,
+							DiffSuppressFunc: suppress.CaseDifference,
```
tombuildsstuff: since this is a new field can we make this case sensitive by removing this?

```suggestion
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -630,6 +641,67 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+							ValidateFunc: validation.StringInSlice([]string{
+								string(network.TLSv10),
+								string(network.TLSv11),
+								string(network.TLSv12),
+							}, true),
```
tombuildsstuff: since this is a new field can we make this case sensitive by making this false?

```suggestion
							}, false),
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -1560,16 +1636,86 @@ func flattenApplicationGatewayConnectionDraining(input *network.ApplicationGatew
 }
 
 func expandApplicationGatewaySslPolicy(d *schema.ResourceData) *network.ApplicationGatewaySslPolicy {
-	vs := d.Get("disabled_ssl_protocols").([]interface{})
-	results := make([]network.ApplicationGatewaySslProtocol, 0)
+	policy := network.ApplicationGatewaySslPolicy{}
+	vs := d.Get("ssl_policy").([]interface{})
+	if len(vs) == 0 {
+		return &policy
```
tombuildsstuff: if `ssl_policy` isn't specified but `disabled_ssl_protocols` is (e.g. upgrading with no code changes) - at the moment that'll cause none to be disabled (since it's read below https://github.com/terraform-providers/terraform-provider-azurerm/pull/3360/files#diff-b945a735bb28605a8bcb81914018d807R1650) - can we parse out `disabled_ssl_protocols` here?

azurerm/resource_arm_application_gateway.go
```diff
@@ -1560,16 +1636,86 @@ func flattenApplicationGatewayConnectionDraining(input *network.ApplicationGatew
...
+	if len(disabledSSLPolicies) == 0 {
+		for _, policy := range d.Get("disabled_ssl_protocols").([]interface{}) {
+			disabledSSLPolicies = append(disabledSSLPolicies, network.ApplicationGatewaySslProtocol(policy.(string)))
+		}
+	}
+
+	if len(disabledSSLPolicies) > 0 {
+		policy = network.ApplicationGatewaySslPolicy{
+			DisabledSslProtocols: &disabledSSLPolicies,
+		}
```
tombuildsstuff: due to the way Azure's API works (in that it fields with nil values are ignored, rather than reset), this needs to be set in either case (even if it's an empty array)

azurerm/resource_arm_application_gateway.go
```diff
@@ -1560,16 +1636,86 @@ func flattenApplicationGatewayConnectionDraining(input *network.ApplicationGatew
...
+	output["policy_name"] = input.PolicyName
+	output["policy_type"] = input.PolicyType
+	output["min_protocol_version"] = input.MinProtocolVersion
+
+	if input.CipherSuites != nil {
+		cipherSuites := make([]interface{}, 0)
+		for _, v := range *input.CipherSuites {
+			cipherSuites = append(cipherSuites, string(v))
+		}
+		output["cipher_suites"] = cipherSuites
```
tombuildsstuff: can we move this one line down (outside of the if statement) to ensure that it's always set, which highlights that there's nothing returned here to the user

azurerm/resource_arm_application_gateway.go
```diff
@@ -1560,16 +1636,86 @@ func flattenApplicationGatewayConnectionDraining(input *network.ApplicationGatew
...
+	}
+
+	if input.DisabledSslProtocols != nil {
+		disabledSslProtocols := make([]interface{}, 0)
+		for _, v := range *input.DisabledSslProtocols {
+			disabledSslProtocols = append(disabledSslProtocols, string(v))
+		}
+		output["disabled_protocols"] = disabledSslProtocols
```
tombuildsstuff: (as above) can we move this one line down (outside of the if statement) to ensure that it's always set, which highlights that there's nothing returned here to the user

azurerm/resource_arm_application_gateway_test.go
```diff
@@ -1939,9 +2005,259 @@ resource "azurerm_application_gateway" "test" {
     capacity = 1
   }
 
-  disabled_ssl_protocols = [
-    "TLSv1_0",
-  ]
```
tombuildsstuff: can we add a test covering the existing behaviour (this ^) to ensure it still works for users?

website/docs/r/application_gateway.html.markdown
```diff
@@ -147,7 +147,9 @@ The following arguments are supported:
 
 * `authentication_certificate` - (Optional) One or more `authentication_certificate` blocks as defined below.
 
-* `disabled_ssl_protocols` - (Optional) A list of SSL Protocols which should be disabled on this Application Gateway. Possible values are `TLSv1_0`, `TLSv1_1` and `TLSv1_2`.
+* `disabled_ssl_protocols` - Deprecated: replaced by `ssl_policy`. (Optional) A list of SSL Protocols which should be disabled on this Application Gateway. Possible values are `TLSv1_0`, `TLSv1_1` and `TLSv1_2`. Not compatible with `ssl_policy`
```
tombuildsstuff: can we make this:

```suggestion
* `disabled_ssl_protocols` - (Optional / **Deprecated**) A list of SSL Protocols which should be disabled on this Application Gateway. Possible values are `TLSv1_0`, `TLSv1_1` and `TLSv1_2`.

~> **NOTE:** `disabled_ssl_protocols ` has been deprecated in favour of `disabled_protocols` in the `ssl_policy` block.
```

website/docs/r/application_gateway.html.markdown
```diff
@@ -147,7 +147,9 @@ The following arguments are supported:
...
+
+* `ssl_policy` (Optional) ssl policys block as defined below.
```
tombuildsstuff: can we make this:

```suggestion
* `ssl_policy` (Optional) A `ssl_policy` block as defined below.
```

website/docs/r/application_gateway.html.markdown
```diff
@@ -379,6 +381,28 @@ A `url_path_map` block supports the following:
...
+
+* `min_protocol_version` - (Optional) The minimal TLS version Required if `policy_type` is set to `Custom`. Possible values can change over time and are published here https://docs.microsoft.com/en-us/azure/application-gateway/application-gateway-ssl-policy-overview. Not compatible with `disabled_protocols`
+
+* `cipher_suites` - (Optional) A List of accepted cipher suites Required if `policy_type` is set to `Custom`. Example ["TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256", "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384", "TLS_RSA_WITH_AES_128_GCM_SHA256"]. Possible values can change over time and are published here https://docs.microsoft.com/en-us/azure/application-gateway/application-gateway-ssl-policy-overview. Not compatible with `disabled_protocols`
+
+For `disabled_protocols`:
+
+* `disabled_protocols` - (Optional) A list of SSL Protocols which should be disabled on this Application Gateway. Possible values are `TLSv1_0`, `TLSv1_1` and `TLSv1_2`. Not compatible with `policy_name` / `policy_type` 
+
+
```
tombuildsstuff: since we've got an explicit allow list unfortunately we're going to need to list these out - as such could we make this block:

```
A `ssl_policy` block supports the following:

* `disabled_protocols` - (Optional) A list of SSL Protocols which should be disabled on this Application Gateway. Possible values are `TLSv1_0`, `TLSv1_1` and `TLSv1_2`.

~> **NOTE:** `disabled_protocols` cannot be set when `policy_name` or `policy_type` are set.

* `policy_type` - (Optional) The Type of the Policy. Possible values are `Predefined` and `Custom`.

~> **NOTE:** `policy_type` is Required when `policy_name` is set - cannot be set if `disabled_protocols` is set.

When using a `policy_type` of `Predefined` the following fields are supported:

* `policy_name` - (Optional) The Name of the Policy e.g AppGwSslPolicy20170401S. Required if `policy_type` is set to `Predefined`. Possible values can change over time and 
are published here https://docs.microsoft.com/en-us/azure/application-gateway/application-gateway-ssl-policy-overview. Not compatible with `disabled_protocols`.

When using a `policy_type` of `Custom` the following fields are supported:

* `cipher_suites` - (Required) A List of accepted cipher suites. Possible values are: `TLS_DHE_DSS_WITH_AES_128_CBC_SHA`, `TLS_DHE_DSS_WITH_AES_128_CBC_SHA256`, `TLS_DHE_DSS_WITH_AES_256_CBC_SHA`, `TLS_DHE_DSS_WITH_AES_256_CBC_SHA256`, `TLS_DHE_RSA_WITH_AES_128_CBC_SHA`, `TLS_DHE_RSA_WITH_AES_128_GCM_SHA256`, `TLS_DHE_RSA_WITH_AES_256_CBC_SHA`, `TLS_DHE_RSA_WITH_AES_256_GCM_SHA384`, `TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA`, `TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256`, `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256`, `TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA`, `TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384`, `TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384`, `TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA`, `TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256`, `TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA`, `TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384`, `TLS_RSA_WITH_3DES_EDE_CBC_SHA`, `TLS_RSA_WITH_AES_128_CBC_SHA`, `TLS_RSA_WITH_AES_128_CBC_SHA256`, `TLS_RSA_WITH_AES_128_GCM_SHA256`, `TLS_RSA_WITH_AES_256_CBC_SHA`, `TLS_RSA_WITH_AES_256_CBC_SHA256` and `TLS_RSA_WITH_AES_256_GCM_SHA384`.

* `min_protocol_version` - (Required) The minimal TLS version. Possible values are `TLSv1_0`, `TLSv1_1` and `TLSv1_2`.
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -617,8 +626,10 @@ func resourceArmApplicationGateway() *schema.Resource {
 
 			// TODO: @tombuildsstuff deprecate this in favour of a full `ssl_protocol` block in the future
```
tombuildsstuff: since we've done this can we make this:

```suggestion
			// TODO: remove in 2.0
```

website/docs/r/application_gateway.html.markdown
tombuildsstuff: the Azure docs auto redirect to the users locale if you leave off the locale prefix, so we can make this: `https://docs.microsoft.com/azure/application-gateway/application-gateway-ssl-policy-overview` :)

bs-matil:
@tombuildsstuff I now added another test wich also unfortunately showed an issue that we need also to mark the ssl_policy block with computed otherwise using disabled_ssl_proctocols would cause a changed resource in the ssl_policy block even if the inner disabled_proctocols block is marked as computed. 

Are you okay with handling it like that?

bs-matil:
Please do NOT merge right now there is something wrong in the handling with Computed inputs. 
Can we have a chat @tombuildsstuff ? I found the same issue in the fqdn implementation. Is there a slack or something?

If not here the textual version:
The deprecation currently is done by the "hack" to set deprecated and successor block to Computed:true. This enables the user to set only one of them while the inconsistent state of the other is ignored. That so far was expected AFAIK. 
But unfortunately there is anther side effect to this. Example `fqdns` vs `fqdn_list` in `backend_address_pool`:
What happens there is the everything works just fine for the new block `fqdns` because the block is handled with priority. What I mean by that is that every input for `fqdn_list` is dropped when `fqdn` is set. 
So good so far. Now one would expect that this does not matter because if `fqdns` is not set the `fqdn_list` will work anyways. This not the case because of the Computed:true flag the input of `fqdns` is always set to the current state if a state has been written once. The result is that if one writes `fqdn_list` once he will never be able to change any written value again with the same block because its always over riden by the existing state and the input is fully ignored. 


Back to our topic. 

How should I proceed here? I can do the very same handling which means ignoring anything but the initial input of `disable_ssl_protcols`? This will allow to create a resource once with the desired setting but users will never be able to change this particular state. 

Alternatively we could "drop" disable_proctols till 2.0 and active the new path just then.

Thanks for any help

bs-matil:
I now implemented "plan a" which is as the `fqdns` handling.

katbyte:
Is this ready for another review @bs-matil ?

bs-matil:
@katabyte yes it is. 
@tombuildsstuff,@katabyte: regards my discussion point. Do I understand the way of deprecation and that it's intended to be like that correctly?

bs-matil:
@katbyte  rebased against latest merges. Can be reviewed

bs-matil:
@katbyte test has been fixed.. Sorry for the inconvenience.

PR #3381, opened 2019-05-06
fix error in app service plan and make pre-check

r0bnet:
There was an obsolete `properties` block which is not valid anymore. Fixed it in documentation.
The other change checks if `kind` is set to `Linux` and `reserved` is either `false` or not set. This is a proposal. If this change is not wanted feel free to skip it.

PR was created because of this issue: https://github.com/terraform-providers/terraform-provider-azurerm/issues/3379

### DIFF
--- a/azurerm/resource_arm_app_service_plan.go
+++ b/azurerm/resource_arm_app_service_plan.go
@@ -192,7 +192,7 @@ func resourceArmAppServicePlanCreateUpdate(d *schema.ResourceData, meta interfac
 	}
 
 	reserved, reservedExists := d.GetOkExists("reserved")
-	if kind == "Linux" {
+	if strings.EqualsFold(kind, "Linux") {
 		if !reserved.(bool) || !reservedExists {
 			return fmt.Errorf("Reserved has to be set to true when using kind Linux")
 		}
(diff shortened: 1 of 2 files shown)


azurerm/resource_arm_app_service_plan.go
```diff
@@ -191,8 +191,15 @@ func resourceArmAppServicePlanCreateUpdate(d *schema.ResourceData, meta interfac
 		appServicePlan.AppServicePlanProperties.PerSiteScaling = utils.Bool(v.(bool))
 	}
 
-	if v, exists := d.GetOkExists("reserved"); exists {
-		appServicePlan.AppServicePlanProperties.Reserved = utils.Bool(v.(bool))
+	reserved, reservedExists := d.GetOkExists("reserved")
+	if kind == "Linux" {
```
tombuildsstuff: could we make this:

```suggestion
	if strings.EqualsFold(kind, "Linux") {
```

r0bnet: Sure, feel free. Wasn't aware of that method ;)

r0bnet:
There seems to be another problem, can you check that @tombuildsstuff ?

PR #3384, opened 2019-05-06
:pencil2: fixing the address_spaces schema property for vNET

bgauduch:
PR regarding issue https://github.com/terraform-providers/terraform-provider-azurerm/issues/3373

Provider compiled & tested with modifications (both vNET resource and datasource).
The doc `d/virtual_network.html.markdown` seem's to be already up-to-date.

Any feedbacks and reviews greatly appreciated 😄

### DIFF
--- a/azurerm/resource_arm_virtual_network.go
+++ b/azurerm/resource_arm_virtual_network.go
@@ -41,7 +41,7 @@ func resourceArmVirtualNetwork() *schema.Resource {
 
 			"location": locationSchema(),
 
-			"address_space": {
+			"address_spaces": {
 				Type:     schema.TypeList,
 				Required: true,
 				MinItems: 1,
@@ -216,7 +216,7 @@ func resourceArmVirtualNetworkRead(d *schema.ResourceData, meta interface{}) err
 
 	if props := resp.VirtualNetworkPropertiesFormat; props != nil {
 		if space := props.AddressSpace; space != nil {
-			d.Set("address_space", space.AddressPrefixes)
+			d.Set("address_spaces", space.AddressPrefixes)
 		}
 
 		if err := d.Set("ddos_protection_plan", flattenVirtualNetworkDDoSProtectionPlan(props)); err != nil {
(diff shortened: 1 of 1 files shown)


bgauduch:
Hi @katbyte,

Thanks for your review and feedback !

I was wondering if you have any process for this kind of change afterward opening this PR, well now I know 👌
By the way, some kind of contributing guidelines with this repository workflow would definitely be a plus for newcomers like me 😉

I agree with you: we should make the doc match the actual behavior first as a quick fix, but in the end the attribut should be updated.
I will revert my changes on `resource_arm_virtual_network.go` and update the documentation as well as this PR title for now.

But some things do confuse me in your feedback and in the azure documentation / usage of vNET in general:
-  I'm not sure which property name should be used. `address_prefixes` is referenced both in the [azure doc on vNET](https://docs.microsoft.com/fr-fr/azure/virtual-network/quick-create-powershell#create-the-virtual-network) and [in the az CLI](https://docs.microsoft.com/en-us/cli/azure/network/vnet?view=azure-cli-latest#az-network-vnet-create), but `address_spaces` is used [in the API doc](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/virtualnetworks/createorupdate#request-body). I believe Terraform Providers should stick to the API, WDYT ?
- It looks like the API allow only one address space per vNET according to the API doc, but one can in fact bind multiple spaces to one vNET using the az CLI... really confusing.
- Beside, `address_prefixes` seems to be related mostly to [azure subnets](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/subnets/createorupdate#request-body), which do supports multiple prefixes according to the API (which make sense).

I will open another improvement issue regarding this attribut naming.

Anyway, it looks like the Azure documentation on vNET (API, CLI, PWSH, web) is not quite up-to-date / homogeneous, I might take some time opening issues / PR on it.

Regards,
Baptiste

bgauduch:
Well, the documentation is already up-to-date with the current attribute naming (`address_space` in everything related to vNET, and `address_spaces` in everything related to vNET datasource). 👍 

So this PR can be closed, I will make a summary in the issue.

Regards,
Baptiste

PR #3389, opened 2019-05-07
Add support to disable authentication for Azure Redis caches

code-haven:
By default, Azure Redis cache instances have authentication enabled. To disable it, we'll need to set `authnotrequired=yes` as part of `redis_configuration`. This PR adds the changes for that.


azurerm/resource_arm_redis_cache.go
```diff
@@ -186,6 +186,12 @@ func resourceArmRedisCache() *schema.Resource {
 							Optional:  true,
 							Sensitive: true,
 						},
+						"authnotrequired": {
```
tombuildsstuff: I think it'd be clearer to invert this and make this a bool (with a default value of 'true'); and then handle the parsing to/from `yes/no` internally:

```suggestion
						"enable_authentication": {
```

WDYT?

code-haven: Thanks Tom, `enable_authentication` is definitely a better name as opposed to `authnotrequired`. I have made the necessary changes to reflect this.

code-haven:
@tombuildsstuff I have made suggested changes. Apologies for the force pushes -- wanted to keep the changes under a single commit.

azurerm/resource_arm_redis_cache_test.go
```diff
@@ -640,6 +640,28 @@ func TestAccAzureRMRedisCache_SubscribeAllEvents(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMRedisCacheWithoutAuth(t *testing.T) {
```
tombuildsstuff: in order to match the other test cases can we make this:

```suggestion
func TestAccAzureRMRedisCache_withoutAuth(t *testing.T) {
```

tombuildsstuff:
@code-haven no worries about the force pushes, I hope you don't mind I had to push a commit to master; for whatever reason we can push to rebases to branches of a fork, but if we do that to the master branch it breaks the PR, so this was simplest 🤷‍♂

code-haven:
@tombuildsstuff That's fine. Thank you!

However, there is a problem(or bug ?) I discovered yesterday while testing this and was confirmed by the solutions architect from Microsoft I'm working with at work:

If we have a subnet that has a Redis cache launched with`enable_authentication = true`, we can't add a second one with `enable_authentication = false` under that subnet. 

```
* azurerm_redis_cache.second-redis-delete: Error issuing create request for Redis Cache second-redis-delete (resource group anon-azure-perf-load-rg): redis.Client#Create: Failure sending request: StatusCode=400 -- Original Error: Code="BadRequest" Message="This subnet could not be used for the request because it contains resource navigation links, such as '<link to first redis>'. Please try with a different subnet.
```

However, we can launch the second one with `enable_authentication = true` and then change it post creation to `false`. Do I need to document this?

On the other hand, everything works fine on an empty subnet.

code-haven:
@katbyte @tombuildsstuff 
It looks like Azure rejects `authnotrequired` parameter if `subnet_id` is not specified.

As a solution, I made `enable_authentication` optional and only available for caches launched within a subnet so that we'll end up sending the parameter only if the instance is launched within a subnet.

What do you guys think?

azurerm/resource_arm_redis_cache.go
```diff
@@ -660,7 +673,16 @@ func expandRedisConfiguration(d *schema.ResourceData) map[string]*string {
 		output["aof-storage-connection-string-1"] = utils.String(v.(string))
 	}
 
-	return output
+	// Redis authentication can only be disabled if it is launched inside a VNET.
+	if v, ok := d.GetOkExists("redis_configuration.0.enable_authentication"); ok {
+		if _, isPrivate := d.GetOk("subnet_id"); !isPrivate {
```
tombuildsstuff: checking the tests - we also need to verify that authentication's disabled here:

```suggestion
		if _, isPrivate := d.GetOk("subnet_id"); !isPrivate && !v {
```

code-haven: It was something similar but this caused the acceptance test cases to fail as @katbyte had commented - this is because Azure does not accept `authnotrequired` to be passed at all if the cache is public. That's the reason there's a blanket check whether if its public or not

tombuildsstuff: got it - in which case I think we want this set a level higher? e.g.

```
	if v, ok := d.GetOkExists("redis_configuration.0.enable_authentication"); ok {
          authenticationEnabled := v.(bool)
          if !authenticationEnabled {
             // assertion and assignment
          }
```

WDYT?

(p.s. sorry for messing up the commits, unfortunately I can't push to your master branch without breaking the PR)

code-haven: Ok, so I think there's some communication gap here. So there are two ways to look at the condition:
i) Don't allow `enable_authentication` to be set for public caches.
ii) Allow `enable_authentication` to be set for public caches but only accept `true` as the value.

The changes in the PR assumed we are going with i). Do we want option ii)? If so I can make the necessary changes for this.

code-haven: Made the changes for this
```
		authEnabled := v.(bool)
		_, isPrivate := d.GetOk("subnet_id")

		// Redis authentication can only be disabled if it is launched inside a VNET.
		if !isPrivate {
			if !authEnabled {
				return nil, fmt.Errorf("Cannot set `enable_authentication` to `false` when `subnet_id` is not set")
			}
		} else {
			value := isAuthNotRequiredAsString(authEnabled)
			output["authnotrequired"] = utils.String(value)
		}
	}
```

Is this fine ??

tombuildsstuff:
Tests pass:

Screenshot 2019-05-14 at 19 50 14

azurerm/resource_arm_redis_cache.go
tombuildsstuff: @code-haven looks good - thanks for pushing that 👍

PR #3391, opened 2019-05-07
resource_arm_application_gateway: add exclusion configuration for waf…

bs-matil:
This PR adds an `exclusion` block to the `waf_configuration` this enables the user to set exclusions from WAF scanning.

### DIFF
--- a/azurerm/resource_arm_application_gateway.go
+++ b/azurerm/resource_arm_application_gateway.go
@@ -919,6 +919,29 @@ func resourceArmApplicationGateway() *schema.Resource {
 							ValidateFunc: validation.IntBetween(1, 128),
 							Default:      128,
 						},
+
+						"exclusion": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"match_variable": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+
+									"selector_match_operator": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+
+									"selector": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+								},
+							},
+						},
 					},
 				},
 			},
@@ -2612,6 +2635,23 @@ func expandApplicationGatewayWafConfig(d *schema.ResourceData) *network.Applicat
 	requestBodyCheck := v["request_body_check"].(bool)
 	maxRequestBodySizeInKb := v["max_request_body_size_kb"].(int)
 
+	exclusions := make([]network.ApplicationGatewayFirewallExclusion, 0)
+	for _, exclusion := range v["exclusion"].([]interface{}) {
+		exclusionMap := exclusion.(map[string]interface{})
+
+		matchVariable := exclusionMap["match_variable"].(string)
+		selectorMatchOperator := exclusionMap["selector_match_operator"].(string)
+		selector := exclusionMap["selector"].(string)
+
+		exclusionList := network.ApplicationGatewayFirewallExclusion{
+			MatchVariable:         utils.String(matchVariable),
+			SelectorMatchOperator: utils.String(selectorMatchOperator),
+			Selector:              utils.String(selector),
+		}
+
+		exclusions = append(exclusions, exclusionList)
+	}
+
 	return &network.ApplicationGatewayWebApplicationFirewallConfiguration{
 		Enabled:                utils.Bool(enabled),
 		FirewallMode:           network.ApplicationGatewayFirewallMode(mode),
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_application_gateway.go
```diff
@@ -2620,6 +2660,7 @@ func expandApplicationGatewayWafConfig(d *schema.ResourceData) *network.Applicat
 		FileUploadLimitInMb:    utils.Int32(int32(fileUploadLimitInMb)),
 		RequestBodyCheck:       utils.Bool(requestBodyCheck),
 		MaxRequestBodySizeInKb: utils.Int32(int32(maxRequestBodySizeInKb)),
+		Exclusions:             &exclusions,
```
mbfrahry: We want to have the above code in a separate expand function, similar to what I suggested in the other review

azurerm/resource_arm_application_gateway.go
```diff
@@ -2657,6 +2698,29 @@ func flattenApplicationGatewayWafConfig(input *network.ApplicationGatewayWebAppl
...
+			}
+
+			if exclusionList.Selector != nil {
+				exclusionListOutput["selector"] = *exclusionList.Selector
+			}
+
+			exclusionLists = append(exclusionLists, exclusionListOutput)
+		}
+
+		output["exclusion"] = exclusionLists
```
mbfrahry: We want to have the above code in a flatten method, similar to what I suggested in the other review

azurerm/resource_arm_application_gateway_test.go
```diff
@@ -536,6 +536,50 @@ func TestAccAzureRMApplicationGateway_connectionDraining(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.exclusion.2.selector_match_operator", "StartsWith"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.exclusion.2.selector", "ORIGIN"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.exclusion.3.match_variable", "RequestHeaderNames"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.exclusion.3.selector_match_operator", "Contains"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.exclusion.3.selector", "ORIGIN"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.exclusion.4.match_variable", "RequestHeaderNames"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.exclusion.4.selector_match_operator", ""),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.exclusion.4.selector", ""),
+				),
+			},
```
mbfrahry: Let's add another test step here to confirm that we can update the added fields properly

website/docs/r/application_gateway.html.markdown
```diff
@@ -397,6 +397,18 @@ A `waf_configuration` block supports the following:
 
 * `max_request_body_size_kb` - (Optional) The Maximum Request Body Size in KB.  Accepted values are in the range `1`KB to `128`KB.  Defaults to `128`KB.
 
+* `exclusion` - (Optional) one or many `exclusion` blocks as defined below.
```
mbfrahry:
```suggestion
* `exclusion` - (Optional) One or more `exclusion` blocks as defined below.
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -919,6 +919,29 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"match_variable": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+
+									"selector_match_operator": {
+										Type:     schema.TypeString,
+										Optional: true,
```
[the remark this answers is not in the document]
bs-matil: actually the problem is that it not "really" defaults to equals any. The reason is that the API is diffrent than the azure ui. In the UI you would set "equals any" in the API an nil feaule is treated as equals_any. Actually I don't know how to face this best to the user.

bs-matil:
@mbfrahry I submitted fixes for "all" issues. Please review again.

azurerm/resource_arm_application_gateway.go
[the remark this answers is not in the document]
bs-matil: I decided to change the documentation and remove the reference to the ui at all

mbfrahry:
Thanks @bs-matil! The conflicts just need to be resolved from between this and your other PR

bs-matil:
@mbfrahry  closed in favor of #3407

PR #3394, opened 2019-05-07
resource_arm_application_gateway: add disabled_rule_group to waf_configuration 

bs-matil:
This PR add a `disabled_rule_group` block to the `waf_configuration` block. This enables a user to disable waf rules

### DIFF
--- a/azurerm/resource_arm_application_gateway.go
+++ b/azurerm/resource_arm_application_gateway.go
@@ -949,14 +949,15 @@ func resourceArmApplicationGateway() *schema.Resource {
 											"REQUEST-941-APPLICATION-ATTACK-XSS",
 											"REQUEST-942-APPLICATION-ATTACK-SQLI",
 											"REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION",
-										}, true),
+										}, false),
 									},
 
 									"rules": {
 										Type:     schema.TypeList,
 										Optional: true,
 										Elem: &schema.Schema{
-											Type: schema.TypeInt,
+											Type:         schema.TypeInt,
+											ValidateFunc: validation.IntAtLeast(1),
 										},
 									},
 								},
@@ -2655,29 +2656,6 @@ func expandApplicationGatewayWafConfig(d *schema.ResourceData) *network.Applicat
 	requestBodyCheck := v["request_body_check"].(bool)
 	maxRequestBodySizeInKb := v["max_request_body_size_kb"].(int)
 
-	disabledRuleGroups := make([]network.ApplicationGatewayFirewallDisabledRuleGroup, 0)
-	for _, disabledRuleGroup := range v["disabled_rule_group"].([]interface{}) {
-		disabledRuleGroupMap := disabledRuleGroup.(map[string]interface{})
-
-		ruleGroupName := disabledRuleGroupMap["rule_group_name"].(string)
-
-		rules := make([]int32, 0)
-		for _, rule := range disabledRuleGroupMap["rules"].([]interface{}) {
-			rules = append(rules, int32(rule.(int)))
-		}
-
-		if len(rules) == 0 {
-			rules = nil
-		}
-
-		ruleGroup := network.ApplicationGatewayFirewallDisabledRuleGroup{
-			RuleGroupName: utils.String(ruleGroupName),
-			Rules:         &rules,
-		}
-
-		disabledRuleGroups = append(disabledRuleGroups, ruleGroup)
-	}
-
 	return &network.ApplicationGatewayWebApplicationFirewallConfiguration{
 		Enabled:                utils.Bool(enabled),
 		FirewallMode:           network.ApplicationGatewayFirewallMode(mode),
@@ -2686,7 +2664,7 @@ func expandApplicationGatewayWafConfig(d *schema.ResourceData) *network.Applicat
 		FileUploadLimitInMb:    utils.Int32(int32(fileUploadLimitInMb)),
 		RequestBodyCheck:       utils.Bool(requestBodyCheck),
 		MaxRequestBodySizeInKb: utils.Int32(int32(maxRequestBodySizeInKb)),
-		DisabledRuleGroups:     &disabledRuleGroups,
+		DisabledRuleGroups:     expandApplicationGatewayFirewallDisabledRuleGroup(v["disabled_rule_group"].([]interface{})),
 	}
 }
 
@@ -2712,35 +2690,8 @@ func flattenApplicationGatewayWafConfig(input *network.ApplicationGatewayWebAppl
 		output["rule_set_version"] = *input.RuleSetVersion
 	}
 
-	ruleGroups := make([]interface{}, 0)
-	if disabledRuleGroups := input.DisabledRuleGroups; disabledRuleGroups != nil {
-		for _, ruleGroup := range *disabledRuleGroups {
-			ruleGroupOutput := map[string]interface{}{}
-
-			if ruleGroup.RuleGroupName != nil {
-				ruleGroupOutput["rule_group_name"] = *ruleGroup.RuleGroupName
-			}
-
-			ruleOutputs := make([]interface{}, 0)
-			if rules := ruleGroup.Rules; rules != nil {
-				for _, rule := range *rules {
-					ruleOutputs = append(ruleOutputs, rule)
-				}
-			}
-			ruleGroupOutput["rules"] = ruleOutputs
-
-			ruleGroups = append(ruleGroups, ruleGroupOutput)
-		}
-
-		output["disabled_rule_group"] = ruleGroups
-	}
-
-	if input.RequestBodyCheck != nil {
-		output["request_body_check"] = *input.RequestBodyCheck
-	}
-
-	if input.MaxRequestBodySizeInKb != nil {
-		output["max_request_body_size_kb"] = *input.MaxRequestBodySizeInKb
+	if input.DisabledRuleGroups != nil {
+		output["disabled_rule_group"] = flattenApplicationGateWayDisabledRuleGroups(input.DisabledRuleGroups)
 	}
 
 	if input.FileUploadLimitInMb != nil {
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_application_gateway.go
```diff
@@ -919,6 +919,49 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+											"REQUEST-933-APPLICATION-ATTACK-PHP",
+											"REQUEST-941-APPLICATION-ATTACK-XSS",
+											"REQUEST-942-APPLICATION-ATTACK-SQLI",
+											"REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION",
+										}, true),
+									},
+
+									"rules": {
+										Type:     schema.TypeList,
+										Required: true,
```
timja: This should be optional
```suggestion
										Optional: true,
```
and will disable the whole group if none is specified

bs-matil: done

website/docs/r/application_gateway.html.markdown
```diff
@@ -399,6 +401,14 @@ A `waf_configuration` block supports the following:
 
 ---
 
+A `disabled_rule_group` block supports the following:
+
+* `rule_group_name` - (Required) The rule group where specific rules should be disabled. Accepted values are:  `crs_20_protocol_violations`, `crs_21_protocol_anomalies`, `crs_23_request_limits`, `crs_30_http_policy`, `crs_35_bad_robots`, `crs_40_generic_attacks`, `crs_41_sql_injection_attacks`, `crs_41_xss_attacks`, `crs_42_tight_security`, `crs_45_trojans`, `REQUEST-911-METHOD-ENFORCEMENT`, `REQUEST-913-SCANNER-DETECTION`, `REQUEST-920-PROTOCOL-ENFORCEMENT`, `REQUEST-921-PROTOCOL-ATTACK`, `REQUEST-930-APPLICATION-ATTACK-LFI`, `REQUEST-931-APPLICATION-ATTACK-RFI`, `REQUEST-932-APPLICATION-ATTACK-RCE`, `REQUEST-933-APPLICATION-ATTACK-PHP`, `REQUEST-941-APPLICATION-ATTACK-XSS`, `REQUEST-942-APPLICATION-ATTACK-SQLI`, `REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION`
+
+* `rules` - (Required) A list of rule numbers which should be disabled in that group. Example: [921110, 921151, 921180] if group is set to `REQUEST-921-PROTOCOL-ATTACK`
```
timja: Need to update here too, something like:
```suggestion
* `rules` - (Optional) A list of rule numbers which should be disabled in that group. Example: [921110, 921151, 921180] if group is set to `REQUEST-921-PROTOCOL-ATTACK`. Disables all rules in the group if `rules` is not specified.
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -2645,6 +2712,37 @@ func flattenApplicationGatewayWafConfig(input *network.ApplicationGatewayWebAppl
...
+				for _, rule := range *rules {
+					ruleOutputs = append(ruleOutputs, rule)
+				}
+			}
+			ruleGroupOutput["rules"] = ruleOutputs
+
+			ruleGroups = append(ruleGroups, ruleGroupOutput)
+		}
+
+		output["disabled_rule_group"] = ruleGroups
```
mbfrahry: We should also move the above code into a flatten method that returns an empty `[]interface{}` if input.DisableRuleGroups is nil
```suggestion
		output["disabled_rule_group"] = flattenApplicationGateWayDisabledRuleGroups(input.DisabledRuleGroups)
```

azurerm/resource_arm_application_gateway_test.go
```diff
@@ -536,6 +536,43 @@ func TestAccAzureRMApplicationGateway_connectionDraining(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.disabled_rule_group.0.rule_group_name", "REQUEST-921-PROTOCOL-ATTACK"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.disabled_rule_group.0.rules.0", "921110"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.disabled_rule_group.0.rules.1", "921151"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.disabled_rule_group.0.rules.2", "921180"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.disabled_rule_group.1.rule_group_name", "REQUEST-930-APPLICATION-ATTACK-LFI"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.disabled_rule_group.1.rules.0", "930120"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.disabled_rule_group.1.rules.1", "930130"),
+					resource.TestCheckResourceAttr(resourceName, "waf_configuration.0.disabled_rule_group.2.rule_group_name", "REQUEST-942-APPLICATION-ATTACK-SQLI"),
+				),
+			},
```
mbfrahry: We should add an extra test step here that just confirms that we can update the disabled rules

azurerm/resource_arm_application_gateway.go
```diff
@@ -2645,6 +2712,37 @@ func flattenApplicationGatewayWafConfig(input *network.ApplicationGatewayWebAppl
...
+	}
+
+	if input.RequestBodyCheck != nil {
```
mbfrahry: It looks like this was already set. https://github.com/terraform-providers/terraform-provider-azurerm/blob/3cc0aa473a4da0afa544f9cd7211a047fcbee651/azurerm/resource_arm_application_gateway.go#L2750 Was there a reason this was re-added here or can we remove this?

azurerm/resource_arm_application_gateway.go
```diff
@@ -2645,6 +2712,37 @@ func flattenApplicationGatewayWafConfig(input *network.ApplicationGatewayWebAppl
...
+		output["request_body_check"] = *input.RequestBodyCheck
+	}
+
+	if input.MaxRequestBodySizeInKb != nil {
```
mbfrahry: It looks like this was already set. https://github.com/terraform-providers/terraform-provider-azurerm/blob/3cc0aa473a4da0afa544f9cd7211a047fcbee651/azurerm/resource_arm_application_gateway.go#L2754
Was there a reason this was re-added here or can we remove this?

website/docs/r/application_gateway.html.markdown
```diff
@@ -391,6 +391,8 @@ A `waf_configuration` block supports the following:
 
 * `rule_set_version` - (Required) The Version of the Rule Set used for this Web Application Firewall.
 
+* `disabled_rule_group` - (Optional) one or many `disabled_rule_group` blocks as defined below.
```
mbfrahry: We should keep this message consistent with the other attributes of a similar nature
```suggestion
* `disabled_rule_group` - (Optional) One or more `disabled_rule_group` blocks as defined below.
```

website/docs/r/application_gateway.html.markdown
```diff
@@ -399,6 +401,14 @@ A `waf_configuration` block supports the following:
 
 ---
 
+A `disabled_rule_group` block supports the following:
+
+* `rule_group_name` - (Required) The rule group where specific rules should be disabled. Accepted values are:  `crs_20_protocol_violations`, `crs_21_protocol_anomalies`, `crs_23_request_limits`, `crs_30_http_policy`, `crs_35_bad_robots`, `crs_40_generic_attacks`, `crs_41_sql_injection_attacks`, `crs_41_xss_attacks`, `crs_42_tight_security`, `crs_45_trojans`, `REQUEST-911-METHOD-ENFORCEMENT`, `REQUEST-913-SCANNER-DETECTION`, `REQUEST-920-PROTOCOL-ENFORCEMENT`, `REQUEST-921-PROTOCOL-ATTACK`, `REQUEST-930-APPLICATION-ATTACK-LFI`, `REQUEST-931-APPLICATION-ATTACK-RFI`, `REQUEST-932-APPLICATION-ATTACK-RCE`, `REQUEST-933-APPLICATION-ATTACK-PHP`, `REQUEST-941-APPLICATION-ATTACK-XSS`, `REQUEST-942-APPLICATION-ATTACK-SQLI`, `REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION`
+
+* `rules` - (Optional) A list of rule numbers which should be disabled in that group. Example: [921110, 921151, 921180] if group is set to `REQUEST-921-PROTOCOL-ATTACK`. Disables all rules in the group if `rules` is not specified.
```
mbfrahry: An example usage isn't recommended in descriptions for attributes.
```suggestion
* `rules` - (Optional) A list of rules which should be disabled in that group.  Disables all rules in the specified group if `rules` is not specified.
```

azurerm/resource_arm_application_gateway.go
```diff
@@ -919,6 +919,49 @@ func resourceArmApplicationGateway() *schema.Resource {
...
+											"REQUEST-942-APPLICATION-ATTACK-SQLI",
+											"REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION",
+										}, true),
+									},
+
+									"rules": {
+										Type:     schema.TypeList,
+										Optional: true,
+										Elem: &schema.Schema{
+											Type: schema.TypeInt,
```
mbfrahry: We should add additional validation here just to confirm that the number they pass in is greater than 0

mbfrahry (changes requested):
Hey @bs-matil. Thanks for submitting this PR. Things are generally good but we just want to make sure these features are consistent with how everything else is done in this resource with additional expand and flatten methods, a little bit more validation, and an additional test check. Other than that, things are great!

azurerm/resource_arm_application_gateway.go
bs-matil: auto merge... will fix it

bs-matil:
@mbfrahry I submitted fixes for "all" issues. Please review again.

mbfrahry (approved):

t0klian:
It is not possible to disable OWASP3,0 rule 200004 from **General** rule group with implemented solution

PR #3407, opened 2019-05-09
resource_arm_application_gateway: add exclusion configuration for waf

bs-matil:
the successor of #3391 which I, unfortunately, can't change any more. 
@mbfrahry please merge this instead.


mbfrahry:
This looks great! Thanks @bs-matil

PR #3437, opened 2019-05-14
[WIP] Bug Fix: `azurerm_application_gateway` - updatable gateway subnet

mbfrahry:
Tests passing: 

image

### DIFF
--- a/azurerm/resource_arm_application_gateway.go
+++ b/azurerm/resource_arm_application_gateway.go
@@ -1889,7 +1889,7 @@ func expandApplicationGatewayIPConfigurations(d *schema.ResourceData) (*[]networ
 		oldVS := oldRaw.([]interface{})
 		newVS := newRaw.([]interface{})
 
-		// If we're creating the applicaiton gateway return the current gateway ip configuration.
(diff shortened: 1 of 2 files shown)


azurerm/resource_arm_application_gateway.go
```diff
@@ -1175,6 +1176,17 @@ func resourceArmApplicationGatewayCreateUpdate(d *schema.ResourceData, meta inte
 		gateway.ApplicationGatewayPropertiesFormat.WebApplicationFirewallConfiguration = expandApplicationGatewayWafConfig(d)
 	}
 
+	if stopApplicationGateway {
+		future, err := client.Stop(ctx, resGroup, name)
+		if err != nil {
```
katbyte: What will happen here if it's already stopped? 🤔

katbyte (approved):
Aside from one comment LGTM 👍

azurerm/resource_arm_application_gateway.go
mbfrahry: Just checked by adding another stop section and it passed just fine. I think we're good!

mbfrahry:
Addressing one of the points of #1576

PR #3473, opened 2019-05-20
`azurerm_traffic_manager_profile`  - support for the `interval_in_seconds`, `timeout_in_seconds`, and `tolerated_number_of_failures` propeties

maikelvdh:


azurerm/resource_arm_traffic_manager_profile.go
```diff
@@ -316,6 +341,18 @@ func resourceAzureRMTrafficManagerMonitorConfigHash(v interface{}) int {
 		if v, ok := m["path"]; ok && v != "" {
 			buf.WriteString(fmt.Sprintf("%s-", m["path"].(string)))
 		}
+
+		if v, ok := m["interval_in_seconds"]; ok && v != "" {
```
katbyte: As you are changing the hash function you will need to add a state migration function & test.

website/docs/r/traffic_manager_profile.html.markdown
```diff
@@ -97,6 +100,12 @@ The `monitor_config` block supports:
 
 * `path` - (Optional) The path used by the monitoring checks. Required when `protocol` is set to `HTTP` or `HTTPS` - cannot be set when `protocol` is set to `TCP`.
 
+* `interval_in_seconds` - (Optional) The interval used to check the endpoint health from a Traffic Manager probing agent. You can specify two values here: 30 seconds (normal probing) and 10 seconds (fast probing). If no values are provided, the profile sets to a default value of 30 seconds.
```
katbyte: Could we quote the values and for consistency work it like:
```suggestion
* `interval_in_seconds` - (Optional) The interval used to check the endpoint health from a Traffic Manager probing agent in seconds. Possible values are `30` (normal probing) and `10` (fast probing). Defaults to `30` seconds.
```

website/docs/r/traffic_manager_profile.html.markdown
```diff
@@ -97,6 +100,12 @@ The `monitor_config` block supports:
...
+
+* `timeout_in_seconds` - (Optional) The amount of time the Traffic Manager probing agent should wait before considering that check a failure when a health check probe is sent to the endpoint. If the Probing Interval is set to 30 seconds, then you can set the Timeout value between 5 and 10 seconds. If no value is specified, it uses a default value of 10 seconds. If the Probing Interval is set to 10 seconds, then you can set the Timeout value between 5 and 9 seconds. If no Timeout value is specified, it uses a default value of 9 seconds.
+
+* `tolerated_number_of_failures` - (Optional) The number of failures a Traffic Manager probing agent tolerates before marking that endpoint as unhealthy. Its value can range between 0 and 9. A value of 0 means a single monitoring failure can cause that endpoint to be marked as unhealthy. If no value is specified, it uses the default value of 3.
```
katbyte:
```suggestion
* `tolerated_number_of_failures` - (Optional) The number of failures a Traffic Manager probing agent tolerates before marking that endpoint as unhealthy. Its value can range between `0` and 9. A value of `0` means a single monitoring failure can cause that endpoint to be marked as unhealthy. Defaults to `3`.
```

website/docs/r/traffic_manager_profile.html.markdown
```diff
@@ -97,6 +100,12 @@ The `monitor_config` block supports:
 
 * `path` - (Optional) The path used by the monitoring checks. Required when `protocol` is set to `HTTP` or `HTTPS` - cannot be set when `protocol` is set to `TCP`.
 
+* `interval_in_seconds` - (Optional) The interval used to check the endpoint health from a Traffic Manager probing agent. You can specify two values here: 30 seconds (normal probing) and 10 seconds (fast probing). If no values are provided, the profile sets to a default value of 30 seconds.
+
+* `timeout_in_seconds` - (Optional) The amount of time the Traffic Manager probing agent should wait before considering that check a failure when a health check probe is sent to the endpoint. If the Probing Interval is set to 30 seconds, then you can set the Timeout value between 5 and 10 seconds. If no value is specified, it uses a default value of 10 seconds. If the Probing Interval is set to 10 seconds, then you can set the Timeout value between 5 and 9 seconds. If no Timeout value is specified, it uses a default value of 9 seconds.
```
katbyte:
```suggestion
* `timeout_in_seconds` - (Optional) The amount of time the Traffic Manager probing agent should wait before considering that check a failure when a health check probe is sent to the endpoint. If the Probing Interval is set to `30` seconds, then you can set the Timeout value between `5` and `10` seconds. If no value is specified, it uses a default value of `10` seconds. If the Probing Interval is set to `10` seconds, then you can set the Timeout value between `5` and `9` seconds. If no Timeout value is specified, it uses a default value of `9` seconds.
```

katbyte: Also adding some validation to enforce this would be nice.

katbyte:
Additionally i am seeing test failures:
```
------- Stdout: -------
=== RUN   TestAccAzureRMTrafficManagerProfile_geographic
=== PAUSE TestAccAzureRMTrafficManagerProfile_geographic
=== CONT  TestAccAzureRMTrafficManagerProfile_geographic
--- FAIL: TestAccAzureRMTrafficManagerProfile_geographic (67.31s)
    testing.go:568: Step 0 error: errors during apply:
        
        Error: trafficmanager.ProfilesClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="BadRequest" Message="The Monitor 'intervalInSeconds' property may only have two values: 10 or 30. Monitor timeout must be between 5 and 0 seconds."
        
          on /opt/teamcity-agent/temp/buildTmp/tf-test502625891/main.tf line 7:
          (source code not available)
        
        
FAIL
```

website/docs/r/traffic_manager_profile.html.markdown
maikelvdh: Is there any example available to validate a property of scheme based on the value of another property? I couldn't find such example, and not sure if it would be feasible here, as for the optional parameters I am relying on the default defined by Azure rather then replicate them here.

tombuildsstuff: unfortunately that's not possible at this time (whilst it's possible via `customizeDiff`, that won't work if the fields interpolated) - as such the best place to do that is within the Create/Update function at this point in time

mbfrahry:
Hey @maikelvdh, I fixed up some of the concerns but was unable to add them to this PR. I've opened up a separate PR at #4002 and will close this one in it's place. Thanks for starting this conversation and I'm looking forward to getting this into the provider.

PR #3475, opened 2019-05-20
Update api_management.go

simongh:
The api management api resource has a required property for path. This property can be left blank, according to the Azure docs (& how we're using it). The validation expression however requires it to be at least 3 characters long.

I've adjusted the regex to allow an empty path and to also enforce the maximum length as specified in Azure API spec (https://github.com/Azure/azure-rest-api-specs/blob/8578473fb7799590c01a7ce243a53121ababdb76/specification/apimanagement/resource-manager/Microsoft.ApiManagement/stable/2019-01-01/definitions.json#L129).


simongh:
I'm not sure what failed there. I think the linter doesn't like my change. I don't really use Go, so I'm not sure what I need to fix.

simongh:
@katbyte that great, thanks

PR #3499, opened 2019-05-22
Role Assignment - Treat name as case insensitive

kiall:
When the wrong case is used for a role name, e.g. acrpull instead of the
correct AcrPull, azure will accept the request and return the correct
casing. This results in Terraform believing it needs to update the
resource incorrectly.

Example of incorrect behaviour:

    -/+ module.aks_simple.module.acr.azurerm_role_assignment.acr_spn_pull_role_assn (new resource required)
          role_definition_name:      "AcrPull" => "acrpull" (forces new resource)

### DIFF
--- a/azurerm/resource_arm_role_assignment.go
+++ b/azurerm/resource_arm_role_assignment.go
@@ -50,12 +50,13 @@ func resourceArmRoleAssignment() *schema.Resource {
 			},
 
 			"role_definition_name": {
-				Type:          schema.TypeString,
-				Optional:      true,
-				Computed:      true,
-				ForceNew:      true,
-				ConflictsWith: []string{"role_definition_id"},
-				ValidateFunc:  validateRoleDefinitionName,
+				Type:             schema.TypeString,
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hi @kiall 

Thanks for opening this PR :)

Case Sensitivity's an interesting problem in Azure (and Terraform) - whilst certain parts of Azure (and the API Specifications) define certain fields (e.g. Resource ID's) as case insensitive - other API's which use that information sometimes require that they're cased correctly.

For this reason whilst historically we've opted to treat some fields as case insensitive - on the whole we're moving towards making the provider case sensitive. As an example, this means that any new fields we add are case-sensitive for both the field value and the validation (which works around these bugs, and also matches the behaviour of other Terraform providers).

That said - in this instance since this field is only used for searching for the role name to obtain it's ID, rather than to be used as a value passed into another service, I think this makes sense (and is a nice UX improvement) - so thanks for this PR 👍🏻

Thanks!

Issue #2152, opened 2018-10-24
azurerm_dev_test_virtual_network only denies use of public IPs on second apply


PatrickMcAfee-Eagle:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

terraform -v
Terraform v0.11.9
* provider.azurerm v1.17.0

### Affected Resource(s)

* azurerm_dev_test_virtual_network

### Terraform Configuration Files

```hcl
resource "azurerm_dev_test_lab" "lab" {
  name                = "${lower(var.name)}"
  location            = "${var.location}"
  resource_group_name = "${var.resource_group_name}"
}

resource "azurerm_dev_test_virtual_network" "base_network" {
  name                = "${azurerm_dev_test_lab.lab.name}-base-network"
  lab_name            = "${azurerm_dev_test_lab.lab.name}"
  resource_group_name = "${azurerm_dev_test_lab.lab.resource_group_name}"

  subnet {
    use_public_ip_address           = "Deny"
    use_in_virtual_machine_creation = "Allow"
  }
}

```
### Expected Behavior

Terraform should set "use_public_ip_address" to DENY on the first apply

### Actual Behavior

Terraform sets "use_public_ip_address" to DENY on the *second* run of apply

### Steps to Reproduce

1. `terraform apply`
1. `terraform plan` <-- will show that DENY needs to be set
1. `terraform apply`
1. `terraform plan` <-- will show that no changes are needed

mbfrahry:
Thanks for opening this issue @PatrickMcAfee-Eagle. This has been merged and will make it into the next release 1.29.0

PR #3501, opened 2019-05-22
Bug Fix: `azurerm_dev_test_virtual_network`

mbfrahry:
For some reason, `subnets` only gets modified during update and not create so I've split out the CreateUpdate function into their own create/update functions. 

Addressing #2152
```
--- PASS: TestAccAzureRMDevTestVirtualNetwork_subnet (520.77s)
--- PASS: TestAccAzureRMDevTestVirtualNetwork_basic (520.77s)
```


katbyte (approved):
LGTM 👍

Issue #3445, opened 2019-05-15
Enable setting maximumElasticWorkerCount for App Service Plan / Premium Functions


mikhailshilkov:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Azure Functions Premium Plan (sku=EPx) comes with a possibility to limit the maximum scale-out by setting the property `maximumElasticWorkerCount` on the Plan itself. By default, the value equals 1 which means no scale-out, so we need to be able to set the value in the terraform templates.

Note, this is a different property than `maximumNumberOfWorkers`.

### Affected Resource(s)

* azurerm_app_service_plan

### Potential Terraform Configuration

```hcl
data "azurerm_app_service_plan" "test" {
  name                = "search-app-service-plan"
  resource_group_name = "search-service"
  maximum_elastic_worker_count = "20"
}
```

### References

* https://azure.microsoft.com/en-us/blog/announcing-the-azure-functions-premium-plan-for-enterprise-serverless-workloads/
* https://docs.microsoft.com/en-us/azure/templates/microsoft.web/2018-02-01/serverfarms
* https://docs.microsoft.com/ru-ru/dotnet/api/microsoft.azure.management.websites.models.appserviceplan.maximumelasticworkercount?view=azure-dotnet

PR #3547, opened 2019-05-29
provider/azurerm: Add maximum_elastic_worker_count to app_service_plan

stack72:
@tombuildsstuff so I didn't add this to the `deprecated` properties block and went with the new way of `maximum_number_of_workers`

Fixes: #3445

```
▶ acctests azurerm TestAccDataSourceAzureRMAppServicePlan_premiumSKU
=== RUN   TestAccDataSourceAzureRMAppServicePlan_premiumSKU
=== PAUSE TestAccDataSourceAzureRMAppServicePlan_premiumSKU
=== CONT  TestAccDataSourceAzureRMAppServicePlan_premiumSKU
--- PASS: TestAccDataSourceAzureRMAppServicePlan_premiumSKU (87.91s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	87.978s


▶ acctests azurerm TestAccAzureRMAppServicePlan_premiumConsumptionPlan
=== RUN   TestAccAzureRMAppServicePlan_premiumConsumptionPlan
=== PAUSE TestAccAzureRMAppServicePlan_premiumConsumptionPlan
=== CONT  TestAccAzureRMAppServicePlan_premiumConsumptionPlan
--- PASS: TestAccAzureRMAppServicePlan_premiumConsumptionPlan (83.97s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	84.024s
```

### DIFF
--- a/azurerm/data_source_app_service_plan.go
+++ b/azurerm/data_source_app_service_plan.go
@@ -74,6 +74,11 @@ func dataSourceAppServicePlan() *schema.Resource {
 				Computed: true,
 			},
 
+			"maximum_elastic_worker_count": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
 			"is_xenon": {
 				Type:     schema.TypeBool,
 				Computed: true,
@@ -118,6 +123,10 @@ func dataSourceAppServicePlanRead(d *schema.ResourceData, meta interface{}) erro
 
 		if props.MaximumNumberOfWorkers != nil {
 			d.Set("maximum_number_of_workers", int(*props.MaximumNumberOfWorkers))
+		}
+
+		if props.MaximumElasticWorkerCount != nil {
+			d.Set("maximum_elastic_worker_count", int(*props.MaximumElasticWorkerCount))
 		}
 	}
 
--- a/azurerm/resource_arm_app_service_plan.go
+++ b/azurerm/resource_arm_app_service_plan.go
@@ -118,6 +118,7 @@ func resourceArmAppServicePlan() *schema.Resource {
 				},
 			},
 
+			/// AppServicePlanProperties
 			"app_service_environment_id": {
 				Type:          schema.TypeString,
 				Optional:      true,
@@ -138,6 +139,13 @@ func resourceArmAppServicePlan() *schema.Resource {
 				Optional:      true,
 				Computed:      true,
 				ConflictsWith: []string{"properties.0.reserved"},
+			},
+
+			"maximum_elastic_worker_count": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validation.IntAtLeast(0),
 			},
 
 			"maximum_number_of_workers": {
@@ -216,6 +224,10 @@ func resourceArmAppServicePlanCreateUpdate(d *schema.ResourceData, meta interfac
 		}
 	}
 
+	if v, exists := d.GetOkExists("maximum_elastic_worker_count"); exists {
+		appServicePlan.AppServicePlanProperties.MaximumElasticWorkerCount = utils.Int32(int32(v.(int)))
+	}
+
 	if reservedExists {
 		appServicePlan.AppServicePlanProperties.Reserved = utils.Bool(reserved.(bool))
 	}
@@ -286,6 +298,10 @@ func resourceArmAppServicePlanRead(d *schema.ResourceData, meta interface{}) err
 
 		if props.MaximumNumberOfWorkers != nil {
 			d.Set("maximum_number_of_workers", int(*props.MaximumNumberOfWorkers))
+		}
+
+		if props.MaximumElasticWorkerCount != nil {
+			d.Set("maximum_elastic_worker_count", int(*props.MaximumElasticWorkerCount))
 		}
 
 		d.Set("per_site_scaling", props.PerSiteScaling)
--- a/website/docs/d/app_service_plan.html.markdown
+++ b/website/docs/d/app_service_plan.html.markdown
@@ -42,10 +42,11 @@ output "app_service_plan_id" {
 
 * `tags` - A mapping of tags assigned to the resource.
 
-* `maximum_number_of_workers` - The maximum number of workers supported with the App Service Plan's sku.
+* `maximum_elastic_worker_count` - The maximum number of total workers allowed for this ElasticScaleEnabled App Service Plan.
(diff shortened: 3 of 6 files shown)


azurerm/resource_arm_app_service_plan.go
```diff
@@ -114,6 +114,9 @@ func resourceArmAppServicePlan() *schema.Resource {
 				},
 			},
 
+			"tags": tagsSchema(),
```
katbyte: Could we keep this at the end of the resource schema?

azurerm/resource_arm_app_service_plan.go
```diff
@@ -136,12 +139,16 @@ func resourceArmAppServicePlan() *schema.Resource {
 				ConflictsWith: []string{"properties.0.reserved"},
 			},
 
-			"maximum_number_of_workers": {
+			"maximum_elastic_worker_count": {
 				Type:     schema.TypeInt,
+				Optional: true,
 				Computed: true,
```
katbyte: What is the reasoning behind this being computed?

azurerm/resource_arm_app_service_plan.go
```diff
@@ -136,12 +139,16 @@ func resourceArmAppServicePlan() *schema.Resource {
 				ConflictsWith: []string{"properties.0.reserved"},
 			},
 
-			"maximum_number_of_workers": {
+			"maximum_elastic_worker_count": {
```
katbyte: Could we add some validation to this property?

website/docs/d/app_service_plan.html.markdown
```diff
@@ -44,6 +44,8 @@ output "app_service_plan_id" {
 
 * `maximum_number_of_workers` - The maximum number of workers supported with the App Service Plan's sku.
 
+* `maximum_elastic_worker_count` - The maximum number of total workers allowed for this ElasticScaleEnabled App Service Plan.
```
katbyte: Could we put this before `maximum_number_of_workers`?

website/docs/r/app_service_plan.html.markdown
```diff
@@ -107,6 +107,7 @@ The following arguments are supported:
 
 * `capacity` - (Optional) Specifies the number of workers associated with this App Service Plan.
 
+* `maximum_elastic_worker_count` - The maximum number of total workers allowed for this ElasticScaleEnabled App Service Plan.
```
katbyte: Could we move this out of the sku block documentation to where it belongs?

azurerm/data_source_app_service_plan.go
```diff
@@ -73,6 +73,11 @@ func dataSourceAppServicePlan() *schema.Resource {
 				Computed: true,
 			},
 
+			"maximum_elastic_worker_count": {
```
katbyte: Could we add this to the `testAccDataSourceAppServicePlan_complete` test?

azurerm/resource_arm_app_service_plan.go
```diff
@@ -136,12 +139,16 @@ func resourceArmAppServicePlan() *schema.Resource {
...
 				Type:     schema.TypeInt,
+				Optional: true,
 				Computed: true,
 			},
 
-			"tags": tagsSchema(),
+			"maximum_number_of_workers": {
```
katbyte: Could we add this to a test? perhaps the `TestAccAzureRMAppServicePlan_completeWindows` one

azurerm/resource_arm_app_service_plan.go
stack72: I'm not sure what type of validation you want to add to an Int here?

azurerm/resource_arm_app_service_plan.go
stack72: Because by default this is 1 - so if the user doesn't set this value then it will be 1 coming back from the API and therefore will cause a diff

website/docs/d/app_service_plan.html.markdown
stack72: I am happy to do this - I am not sure why though but 🤷‍♂

website/docs/r/app_service_plan.html.markdown
stack72: Good catch!!

stack72:
Hi @katbyte, do you have any thoughts on what validation you want added in the question above? I'd love to get this PR finished up :)

azurerm/data_source_app_service_plan.go
stack72: I added a new test here because it's a new SKU and Kind

azurerm/resource_arm_app_service_plan.go
stack72: Added to `TestAccAzureRMAppServicePlan_premiumConsumptionPlan` as it was the correct Kind and SKU

stack72:
@katbyte / @tombuildsstuff so I looked into what Validation was available. The API does have no mention of it:

https://docs.microsoft.com/en-us/azure/templates/microsoft.web/2018-02-01/serverfarms#AppServicePlanProperties

No does the SDK. 

I believe this is ready for review again

azurerm/resource_arm_app_service_plan.go
tombuildsstuff: (chatting offline, we're adding `IntAtLeast: 0` here)

website/docs/d/app_service_plan.html.markdown
```diff
@@ -42,10 +42,11 @@ output "app_service_plan_id" {
 
 * `tags` - A mapping of tags assigned to the resource.
 
-* `maximum_number_of_workers` - The maximum number of workers supported with the App Service Plan's sku.
+* `maximum_elastic_worker_count` - The maximum number of total workers allowed for this ElasticScaleEnabled App Service Plan.
 
 * `is_xenon` - A flag that indicates if it's a xenon plan (support for Windows Container)
 
+* `maximum_number_of_workers` - The maximum number of workers supported with the App Service Plan's sku.
```
tombuildsstuff: to not break the formatting, can we make this:

```suggestion
* `maximum_number_of_workers` - The maximum number of workers supported with the App Service Plan's sku.

```

azurerm/resource_arm_app_service_plan.go
katbyte: ta 👍

PR #3585, opened 2019-06-04
Update azurerm_resource_group documentation

dbourcet:
The  `name` attribute can be exported, but the documentation don't mention it yet.

### DIFF
--- a/website/docs/r/resource_group.html.markdown
+++ b/website/docs/r/resource_group.html.markdown
@@ -37,11 +37,9 @@ The following arguments are supported:
 
 ## Attributes Reference
 
-The following attributes are exported:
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hi @dbourcet 

Thanks for this PR :)

In Terraform, all arguments which can be specified on a resource are also exported - as such, rather than duplicating these attributes in the docs, I think we'd be better to keep this in check by updating the text:

```
The following attributes are exported
```

to:

```
In addition to the arguments above, the following attributes are exported:
```

which should make this clearer - WDYT?

Thanks!

dbourcet:
Oh, I didn't know that, thanks for the clarification. Yes, the modification you suggest is good, as it will prevent some people like me to be misleaded.

PR #3599, opened 2019-06-05
Improve docs on create_mode for sql_database

WheresAlice:
'See below' wasn't all that helpful.  I've had a go at improving this, but haven't actually used this resource myself.

### DIFF
--- a/website/docs/r/sql_database.html.markdown
+++ b/website/docs/r/sql_database.html.markdown
@@ -50,7 +50,7 @@ The following arguments are supported:
 
 * `server_name` - (Required) The name of the SQL Server on which to create the database.
 
-* `create_mode` - (Optional) Specifies how to create the database. Defaults to `Default`, ie. create a new database.  This can also be set to `PointInTimeRestore` to restore from a snapshot.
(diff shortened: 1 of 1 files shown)


website/docs/r/sql_database.html.markdown
```diff
@@ -50,7 +50,7 @@ The following arguments are supported:
 
 * `server_name` - (Required) The name of the SQL Server on which to create the database.
 
-* `create_mode` - (Optional) Specifies the type of database to create. Defaults to `Default`. See below for the accepted values/
+* `create_mode` - (Optional) Specifies how to create the database. Defaults to `Default`, ie. create a new database.  This can also be set to `PointInTimeRestore` to restore from a snapshot.
```
katbyte:
```suggestion
* `create_mode` - (Optional) Specifies how to create the database. Must be either `Default` to create a new database or `PointInTimeRestore` to restore from a snapshot. Defaults to `Default`.
```

WheresAlice:
Yes, that looks good.  I've applied that change.

Issue #1712, opened 2018-08-02
IoT Device Provisioning Resource


jgardner04:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Add a resource for Azure IoT Device Provisioning Service. This would likely be a new resource; something like azurerm_iotdps. This command would be completed through the Azure-CLI through a command like that below.

```hcl
az iot dps create --name MyDps --resource-group MyResourceGroup --location westus
```

### New or Affected Resource(s)

This would be a new resource. Something like: 

* azurerm_iotdps
OR
* azurerm_iot_dps

### Potential Terraform Configuration

```hcl
resource "azurerm_iot_dps" "test" {
  name                             = "test"
  resource_group_name  = "test"
  location                         = "West US"
  sku                                = "S1"

  tags {
    "purpose" = "testing"
  }
}
```
An extended version of this would include the options for access-policy, certificate, and linked-hub. That might look something like:

```hcl
resource "azurerm_iot_dps" "test" {
  name                             = "test"
  resource_group_name  = "test"
  location                         = "West US"
  sku                                = "S1"
  access_policy {
    name = "test_policy"
    rights = "DeviceConnect"
    shared_access_policy {
      primary_key = "123"
      secondary_key = "abc"
    }
  }
  certificate {
    contents = "${base64encode(file("certificate-to-import.pfx"))}"
    password = "Pass@word1"
  }
  linked-hub = "${azurerm_iothub.test.name}"

  tags {
    "purpose" = "testing"
  }
}
```

ezhaar:
what about the possibility to link an iot hub to the newly created dps? that would be great.

ukphillips:
Agreed we need the IoT hub linkage to make this useful - Glad to see this started but without the hubs it does not add a lot of value.

mbfrahry:
Hey @ezhaar and @ukphillips, #3922 should finish up this issue by adding support for linked hubs

ezhaar:
@mbfrahry I found another issue with the DPS resource. I filed a bug #22228

PR #3618, opened 2019-06-08
New Resource: `azurerm_iot_dps`

mbfrahry:
Start of #1712. The other pieces of that issue will be separate resources 
```
--- PASS: TestAccAzureRMIotDPS_update (277.78s)
--- PASS: TestAccAzureRMIotDPS_basic (373.70s)
```


website/docs/r/iot_dps.html.markdown
```diff
@@ -0,0 +1,70 @@
...
+}
+
+resource "azurerm_iot_dps" "test" {
+  name                = "test"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  location            = "${azurerm_resource_group.test.location}"
+
+  sku {
+    name     = "S1"
+    tier     = "Standard"
```
katbyte:
```suggestion
   tier     = "Standard"
```

mbfrahry: I don't think we need this one. This change changes the spaces from 4 to 3

PR #3645, opened 2019-06-13
New Resource: `azurerm_iot_dps_certificate`

mbfrahry:
Another piece of the puzzle for #1712

```
--- PASS: TestAccAzureRMIotDPSCertificate_basic (259.34s)
--- PASS: TestAccAzureRMIotDPSCertificate_update (374.39s)
```

### DIFF
--- a/azurerm/resource_arm_iot_dps_cerificate.go
+++ b/azurerm/resource_arm_iot_dps_cerificate.go
@@ -67,7 +67,7 @@ func resourceArmIotDPSCertificateCreateOrUpdate(d *schema.ResourceData, meta int
 		}
 
 		if existing.ID != nil && *existing.ID != "" {
-			return tf.ImportAsExistsError("azurerm_iot_dps", *existing.ID)
+			return tf.ImportAsExistsError("azurerm_iot_dps_certificate", *existing.ID)
 		}
 	}
 
--- a/azurerm/resource_arm_iot_dps_cerificate.go
+++ b/azurerm/resource_arm_iot_dps_cerificate.go
@@ -1,0 +1,149 @@
+package azurerm
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+
+	"github.com/Azure/azure-sdk-for-go/services/provisioningservices/mgmt/2018-01-22/iothub"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func resourceArmIotDPSCertificate() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceArmIotDPSCertificateCreateOrUpdate,
+		Read:   resourceArmIotDPSCertificateRead,
+		Update: resourceArmIotDPSCertificateCreateOrUpdate,
+		Delete: resourceArmIotDPSCertificateDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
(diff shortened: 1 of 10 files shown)


azurerm/resource_arm_iot_dps_cerificate.go
```diff
@@ -0,0 +1,149 @@
...
+	resp, err := client.Get(ctx, name, resourceGroup, iotDPSName, "")
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return nil
+		}
+		return fmt.Errorf("Error retrieving IoT Device Provisioning Service Certificate %q (Device Provisioning Service %q / Resource Group %q): %+v", name, iotDPSName, resourceGroup, err)
+	}
+
+	// TODO address this delete call if https://github.com/Azure/azure-rest-api-specs/pull/6311 get's merged
+	if _, err := client.Delete(ctx, resourceGroup, *resp.Etag, iotDPSName, name, "", nil, nil, iothub.ServerAuthentication, nil, nil, nil, ""); err != nil {
```
katbyte: majestic

azurerm/resource_arm_iot_dps_cerificate.go
```diff
@@ -0,0 +1,149 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.IoTHubName,
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
```
tombuildsstuff: is there a reason this is case insensitive?

```suggestion
			"resource_group_name": azure.SchemaResourceGroupName(),
```

azurerm/resource_arm_iot_dps_cerificate.go
```diff
@@ -0,0 +1,149 @@
...
+			"iot_dps_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.IoTHubName,
+			},
+
+			"certificate_content": {
+				Type:         schema.TypeString,
+				Required:     true,
```
tombuildsstuff: this is a Sensitive value, as such can we make this:

```suggestion
				Required:     true,
				Sensitive: true,
```

azurerm/resource_arm_iot_dps_cerificate.go
```diff
@@ -0,0 +1,149 @@
...
+				ValidateFunc: validation.StringLenBetween(1, 10000),
```
tombuildsstuff: is there a reason for picking this number? I feel like it'd be better to use `NoEmptyStrings` here since certificates can be any length?

azurerm/resource_arm_iot_dps_cerificate.go
```diff
@@ -0,0 +1,149 @@
...
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error retrieving IoT Device Provisioning Service Certificate %q (Device Provisioning Service %q / Resource Group %q): %+v", name, iotDPSName, resourceGroup, err)
+	}
+
+	d.Set("name", resp.Name)
+	d.Set("resource_group_name", resourceGroup)
+	d.Set("iot_dps_name", iotDPSName)
+
```
tombuildsstuff: can we add a comment here to clarify that `certificate_content` isn't returned from the API?

azurerm/resource_arm_iot_dps_cerificate.go
tombuildsstuff: since there's a potential panic here if `resp.Etag` is nil - could we nil-check this?

azurerm/resource_arm_iot_dps_cerificate.go
mbfrahry: Nerp! Swapped

azurerm/resource_arm_iot_dps_cerificate.go
mbfrahry: I wasn't sure how best to validate this and grabbed it from another certificate attribute. I'll change it to `NoEmptyStrings`

PR #3654, opened 2019-06-13
squashed sf create / update to allow AAD update

steve-hawkins:
it looks like the REST specs are a little misleading as the [Create](https://docs.microsoft.com/en-us/rest/api/servicefabric/sfrp-api-clusters_create) can also update

I need to be able to add AAD to some existing clusters without forcing a new cluster resource and this seemed to be the only way (also the way this is achieved via the Portal)


mbfrahry:
Thanks opening this PR @steve-hawkins! This should make it into the next release

Issue #3641, opened 2019-06-12
Changing tags of azurerm_search_service forces new resource


maxbog:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
Terraform v0.12.1
+ provider.azurerm v1.30.1

### Affected Resource(s)

* `azurerm_search_service`

### Terraform Configuration Files

```hcl
resource "azurerm_search_service" "common" {
  location = var.location
  resource_group_name = azurerm_resource_group.services.name

  name = "example-search"
  sku = "standard"

  replica_count = 1
  partition_count = 1

  tags = {
    tag1 = "value1"
  }
}
```

changed to: 

```hcl
resource "azurerm_search_service" "common" {
  location = var.location
  resource_group_name = azurerm_resource_group.services.name

  name = "example-search"
  sku = "standard"

  replica_count = 1
  partition_count = 1

  tags = {
    tag1 = "value2"
  }
}
```

### Expected Behavior

`tag1` value should have changed without affecting the resource

### Actual Behavior

terraform wants to recreate the search service to change the tag

### Steps to Reproduce

1. Change a value of any `azurerm_search_service` tag
2. `terraform apply`

### References

* https://github.com/terraform-providers/terraform-provider-azurerm/blob/41f071c23b449df7f4dfe9aa57ac81c3386e9e04/azurerm/resource_arm_search_service.go#L73

Why does the referenced line have `tagsForceNewSchema` applied instead of `tagsSchema`?

maxbog:
Ok, I tried to fix it by changing the schema, but unfortunately, when I try to update a tag value, I get an error:
```
azurerm_search_service.test: Modifying... [id=/subscriptions/***/resourceGroups/acctest-search/providers/Microsoft.Search/searchServices/ttestsearch]

Error: doesn't support update

  on test.tf line 10, in resource "azurerm_search_service" "test":
  10: resource "azurerm_search_service" "test" {
```

However, I can easily update the resource using the REST API. Maybe this is some issue with the Azure SDK or something else in the provider code that could cause that error?

[...]

PR #3660, opened 2019-06-14
arm_search_service - changing tags should not force new resource

maxbog:
Changing tags should not force creating new search service.

However, the new test fails with the error:
```
=== RUN   TestAccAzureRMSearchService_tagUpdate
=== PAUSE TestAccAzureRMSearchService_tagUpdate
=== CONT  TestAccAzureRMSearchService_tagUpdate
--- FAIL: TestAccAzureRMSearchService_tagUpdate (99.17s)
    testing.go:568: Step 2 error: errors during apply:

        Error: doesn't support update

          on C:\Users\11024186\AppData\Local\Temp\tf-test975401293\main.tf line 7:
          (source code not available)


FAIL
FAIL    github.com/terraform-providers/terraform-provider-azurerm/azurerm       102.504s
```

Not sure what is the problem, as the raw REST calls work without any issue. Please see #3641 for more information

### DIFF
--- a/azurerm/resource_arm_search_service.go
+++ b/azurerm/resource_arm_search_service.go
@@ -17,6 +17,7 @@ func resourceArmSearchService() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceArmSearchServiceCreateUpdate,
 		Read:   resourceArmSearchServiceRead,
+		Update: resourceArmSearchServiceCreateUpdate,
 		Delete: resourceArmSearchServiceDelete,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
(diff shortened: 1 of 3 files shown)


mbfrahry:
Hey @maxbog. This change looks good and I can see where you got hung up. Terraform likes to validate provider code and remove any unnecessary calls where it can. Before this change, Terraform saw that all attributes in the schema were either `ForceNew` or `Computed`. This means that no attributes can be updated and so Terraform let the provider maintainer know that Update is not supported in this resource. Unfortunately, Terraform doesn't let you know that a resource now supports Update when an attribute removes that `ForceNew` tag. For you to get up and running, go ahead and add `Update: resourceArmSearchServiceCreateUpdate` [here](https://github.com/terraform-providers/terraform-provider-azurerm/pull/3660/files#diff-2d240a4e84dcf9bcfd65e264d2e6afefR20) and give the test another go.

maxbog:
@mbfrahry Thanks for the help and a great explanation! I added the `Update` function and the tests pass :D

mbfrahry:
Awesome! These changes look good to me and should make it into the next release! Thanks so much for this @maxbog

PR #3695, opened 2019-06-18
Note added for forced casing at runtime.

jblose:
While working with the Azure MySQL PaaS it was identified that the `name` of the resource was forced to lower case in Azure at apply time; If uppercase letters were used in the Terraform Plan it was immediately out-of-date once applied and flagged the freshly created resource Destroy/Create at next `planning`.

### DIFF
--- a/azurerm/helpers/azure/mysql.go
+++ b/azurerm/helpers/azure/mysql.go
@@ -1,0 +1,15 @@
+package azure
+
+import (
+	"fmt"
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
+)
+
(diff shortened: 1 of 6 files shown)


jblose:
@katbyte Thank you for following up on the PR. That works for me, correcting the provider over the documentation is the right way to go long term. 

I may try and report it on the Azure project as well since "normally" when attempting to name something incorrectly Azure returns an `Error Code = 400` with some flavor of Exception so the implementer understands something is incorrect versus quietly being helpful.

jblose:
> @jblose,
> 
> Hope you don't mind but i've pushed those changes to get this merged

No worries at all and thank you, I was hoping to get around to it this past weekend and ended up sidetracked on another project. Keep fighting the good fight!

PR #3697, opened 2019-06-18
Switch client registration to use a shared type & convert A-D

katbyte:


azurerm/config.go
```diff
@@ -366,24 +360,33 @@ func getArmClient(c *authentication.Config, skipProviderRegistration bool, partn
...
-	client.cognitive = cognitive.BuildClient(endpoint, c.SubscriptionID, partnerId, auth, skipProviderRegistration)
-	client.containers = containers.BuildClient(endpoint, c.SubscriptionID, partnerId, auth, skipProviderRegistration)
+	o := &ar.ClientOptions{
+		Authorizer:                 auth,
+		ProviderName:               "AzureRM",
+		PartnerId:                  partnerId,
+		PollingDuration:            60 * time.Minute,
+		SkipProviderReg:            skipProviderRegistration,
+		EnableCorrelationRequestID: true,
+	}
```
tombuildsstuff: we can move the (ARM)Endpoint and SubscriptionID in here too, since they're reused on most things; then just pass in `options` below

azurerm/config.go
```diff
@@ -63,6 +57,14 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/streamanalytics/mgmt/2016-03-01/streamanalytics"
 	trafficmanagerSvc "github.com/Azure/azure-sdk-for-go/services/trafficmanager/mgmt/2018-04-01/trafficmanager"
 	"github.com/Azure/azure-sdk-for-go/services/web/mgmt/2018-02-01/web"
+	mainStorage "github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+	az "github.com/Azure/go-autorest/autorest/azure"
```
tombuildsstuff: can we remove this alias here?

```suggestion
	"github.com/Azure/go-autorest/autorest/azure"
```

azurerm/internal/ar/client.go
```diff
@@ -10,26 +10,36 @@ import (
...
 )
 
-func ConfigureClient(client *autorest.Client, auth autorest.Authorizer, partnerId string, skipProviderReg bool) {
-	setUserAgent(client, partnerId)
-	client.Authorizer = auth
-	client.Sender = sender.BuildSender("AzureRM")
-	client.PollingDuration = 60 * time.Minute
-	client.SkipResourceProviderRegistration = skipProviderReg
-	client.RequestInspector = azure.WithCorrelationRequestID(azure.CorrelationRequestID())
+type ClientOptions struct {
```
tombuildsstuff: can we rename the package this is in to `common` since this contains common client bits, rather than `ar` which isn't descriptive to new users trying to navigate the codebase?

azurerm/internal/ar/client.go
```diff
@@ -10,26 +10,36 @@ import (
...
 }
 
-func setUserAgent(client *autorest.Client, partnerID string) {
-	// TODO: This is the SDK version not the CLI version, once we are on 0.12, should revisit
+func setUserAgent(client *autorest.Client, providerName, partnerID string) {
 	tfUserAgent := httpclient.UserAgentString()
 
-	pv := version.ProviderVersion
-	providerUserAgent := fmt.Sprintf("%s terraform-provider-azurerm/%s", tfUserAgent, pv)
+	providerUserAgent := fmt.Sprintf("%s terraform-provider-azurerm/%s", tfUserAgent, version.ProviderVersion)
```
tombuildsstuff: I don't think this is worth moving this logic out to the common library (which I assume is the ultimate intention here), since this is something you need to change to be able to debug things when working with authentication; so can we remove the `providerName` and hard-code that here?

azurerm/internal/ar/client.go
```diff
@@ -10,26 +10,36 @@ import (
...
+type ClientOptions struct {
+	Authorizer                 autorest.Authorizer
+	ProviderName               string
+	PartnerId                  string
+	PollingDuration            time.Duration
+	SkipProviderReg            bool
+	EnableCorrelationRequestID bool
+}
+
+func ConfigureClient(c *autorest.Client, o *ClientOptions) {
```
tombuildsstuff: since different clients use different authentication options here; this won't work as-is - instead we'll need to pass in (at least) the Authorizer to use for the particular client here

tombuildsstuff: we could probably also make this a method on the 'Options' object here, e.g.

```suggestion
func (options ClientOptions) ConfigureClient(c *autorest.Client, auth autorest.Authorizer) {
```

azurerm/internal/ar/client.go
```diff
@@ -10,26 +10,36 @@ import (
...
 
-func ConfigureClient(client *autorest.Client, auth autorest.Authorizer, partnerId string, skipProviderReg bool) {
-	setUserAgent(client, partnerId)
-	client.Authorizer = auth
-	client.Sender = sender.BuildSender("AzureRM")
-	client.PollingDuration = 60 * time.Minute
-	client.SkipResourceProviderRegistration = skipProviderReg
-	client.RequestInspector = azure.WithCorrelationRequestID(azure.CorrelationRequestID())
+type ClientOptions struct {
+	Authorizer                 autorest.Authorizer
```
tombuildsstuff: since there's currently 3 authorizers (and about to be a 4th) we need to be more descriptive with this:

```suggestion
	ResourceManagerAuthorizer                 autorest.Authorizer
```

PR #3704, opened 2019-06-19
Bugfix: Error creating VM with private IP in DTL.

kmsisco:
Creating a VM in a DTL that doesn't allow public IPs results in the following error messages even when "disallow_public_ip_address" is true.

> Error: Error waiting for creation/update of DevTest Linux Virtual Machine "terraform-test" (Lab "xxx" / Resource Group "..."): Code="RequestDisallowedByPolicy" Message="Resource 'xxx' was disallowed by policy. Policy identifiers: '[{\"policyAssignment\":{\"name\":\"No Public IP Policy\",\"id\":\"/subscriptions/xxx/providers/Microsoft.Authorization/policyAssignments/xxx\"},\"policyDefinition\":{\"name\":\"NoPublicIPPolicyDefinition\",\"id\":\"/subscriptions/xxx/providers/Microsoft.Authorization/policyDefinitions/NoPublicIPPolicyDefinition\"}}]'."

This change only sets the NetworkInterface property if inbound_nat_rules was defined to allow Azure to use its defaults.


kmsisco:
@katbyte  Moving 181 into the if has the same effect as the current code.   I am setting disallow_public_ip_address to true to get a private IP but I don't want it to initialize the nic. When the structure is passed Azure tries to allocate a public IP for some reason.

katbyte:
Ah, could you write a test with that specific configuration the fails before the fix, and passes after?

Thanks @kmsisco

kmsisco:
These changes depend on #3717

kmsisco:
@katbyte 

I merged the changes that fixed the virtual network issue and now the tests pass.

katbyte:
@kmsisco,

The test is now failing with:
```
 DIFF:
        
        CREATE: azurerm_policy_assignment.test
          description:          "" => "No Public IP Policy Test"
          display_name:         "" => "No Public IP Policy"
          id:                   "" => "<computed>"
          identity:             "" => "<computed>"
          name:                 "" => "test-policy-assignment"
          policy_definition_id: "" => "<computed>"
          scope:                "" => "/subscriptions/1a6092a6-137e-4025-9a7c-ef77f76f2c02"
        CREATE: azurerm_policy_definition.test
          display_name: "" => "No Public IP Policy"
          id:           "" => "<computed>"
          metadata:     "" => "<computed>"
          mode:         "" => "Indexed"
          name:         "" => "NoPublicIPPolicyDefinition"
          policy_rule:  "" => "  {\n    \"if\": {\n      \"anyOf\": [{\n        \"source\": \"action\",\n        \"like\": \"Microsoft.Network/publicIPAddresses/*\"\n      }]\n    },\n    \"then\": {\n      \"effect\": \"deny\"\n    }\n  }\n"
          policy_type:  "" => "Custom"
        
```

Is this something unique to our environment maby 🤔

PR #3718, opened 2019-06-23
New resource: azurerm_private_dns_zone

timja:
Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/3714


azurerm/resource_arm_private_dns_zone_test.go
```diff
@@ -0,0 +1,219 @@
...
+			},
+			{
+				Config:      testAccAzureRMPrivateDnsZone_requiresImport(ri, location),
+				ExpectError: testRequiresImportError("azurerm_private_dns_zone"),
+			},
+		},
+	})
+}
+
+func TestAccAzureRMPrivateDnsZone_withTags(t *testing.T) {
```
timja: This test is currently failing for me locally.

The tags get correctly applied but don't seem to be in the state, not sure what is going on here.

Any help appreciated

timja: It's happening because the response from the API only has the old values, its returning stale data.

timja: I've added a sleep, unless anyone has any better ideas...

bentterp:
This is good! 
I just can't figure out how to use it for VNET auto-registrations. Do we have another resource type coming for that, or am I just not getting it? Tried creating a link in the portal and then generating a plan, no changes required.

timja:
It needs a new resource type, no issue created for it yet AFAIK.
I just created the first resource required for private dns.

bentterp:
#3758 :-D

I'm quite good at creating issues, no good at writing the code for the PR's....

PR #3721, opened 2019-06-24
Resource arm analysis services server

r0bnet:
Created this resource as it was requested in an open issue:
https://github.com/terraform-providers/terraform-provider-azurerm/issues/1734
I omitted the `gateway_resource_id` and `backup_blob_container_uri` for now as i had problems in testing it (for gateway resource id i'm not even sure for it is needed).

There is also currently no acc test for the `admin_users` because i'm having difficulties of creating them in the test. Manual tests were successful though.


tlaur:
Brilliant, well done on working on this, will take a closer look and take this out for a test spin later.

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+	return validation.StringInSlice(connectionModes, true)
+}
+
+func expandAnalysisServicesServerSku(sku string) *analysisservices.ResourceSku {
+	return &analysisservices.ResourceSku{
+		Name: &sku,
+	}
+}
+
+func flattenAnalysisServicesServerSku(sku *analysisservices.ResourceSku) string {
```
katbyte: I don't think this function is necessary.

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+func validateQuerypoolConnectionMode() schema.SchemaValidateFunc {
+	connectionModes := make([]string, len(analysisservices.PossibleConnectionModeValues()))
+	for i, v := range analysisservices.PossibleConnectionModeValues() {
+		connectionModes[i] = string(v)
+	}
+
+	return validation.StringInSlice(connectionModes, true)
+}
+
+func expandAnalysisServicesServerSku(sku string) *analysisservices.ResourceSku {
```
katbyte: I don't think this function is necessary.

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+		output["range_start"] = *fwRule.RangeStart
+		output["range_end"] = *fwRule.RangeEnd
+
+		fwRules = append(fwRules, output)
+	}
+
+	return firewallSettings.EnablePowerBIService, fwRules
+}
+
+func flattenAnalysisServicesServerQuerypoolConnectionMode(serverProperties *analysisservices.ServerProperties) *string {
```
katbyte: I don't think this function is necessary.

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+			RangeStart:       utils.String(fwRule["range_start"].(string)),
+			RangeEnd:         utils.String(fwRule["range_end"].(string)),
+		}
+	}
+	firewallSettings.FirewallRules = &fwRules
+
+	return &firewallSettings
+}
+
+func flattenAnalysisServicesServerAdminUsers(serverAdministrators *analysisservices.ServerAdministrators) *[]string {
```
katbyte: I don't think this function is necessary.

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+	for _, admin := range adminUsers.List() {
+		if adm, ok := admin.(string); ok {
+			members = append(members, adm)
+		}
+	}
+
+	return &analysisservices.ServerAdministrators{Members: &members}
+}
+
+func expandAnalysisServicesServerQuerypoolConnectionMode(d *schema.ResourceData) *analysisservices.ConnectionMode {
```
katbyte: I don't think this function is necessary.

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+	if fwSettings := serverProperties.IPV4FirewallSettings; len(*fwSettings.FirewallRules) > 0 && fwSettings.EnablePowerBIService == nil {
+		return fmt.Errorf("`enable_power_bi_service` must be set if there is at least one firewall rule")
+	}
+
+	tags := d.Get("tags").(map[string]interface{})
+
+	analysisServicesServer := analysisservices.Server{
+		Name:             &name,
+		Location:         &location,
+		Sku:              expandAnalysisServicesServerSku(sku),
```
katbyte: Given this is just .a string i don't think we need an expand function here

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+}
+
+func resourceArmAnalysisServicesServerCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).analysisServicesServerClient
+	ctx := meta.(*ArmClient).StopContext
+
+	log.Printf("[INFO] preparing arguments for Azure ARM Analysis Services Server creation.")
+
+	name := d.Get("name").(string)
+	resGroup := d.Get("resource_group_name").(string)
```
katbyte: _minor_ Ideally we should call this `resourceGroup` for consistency

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+
+	d.Set("name", name)
+	d.Set("resource_group_name", resGroup)
+
+	if location := server.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+
+	d.Set("sku", flattenAnalysisServicesServerSku(server.Sku))
```
katbyte: This should simply be:
```suggestion
	d.Set("sku", server.Sku)
```

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+	serverProperties := expandAnalysisServicesServerMutableProperties(d)
+
+	if fwSettings := serverProperties.IPV4FirewallSettings; len(*fwSettings.FirewallRules) > 0 && fwSettings.EnablePowerBIService == nil {
+		return fmt.Errorf("`enable_power_bi_service` must be set if there is at least one firewall rule")
+	}
+
+	tags := d.Get("tags").(map[string]interface{})
+
+	analysisServicesServer := analysisservices.ServerUpdateParameters{
+		Sku:                     expandAnalysisServicesServerSku(sku),
```
katbyte: As above

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+		Name: &sku,
+	}
+}
+
+func flattenAnalysisServicesServerSku(sku *analysisservices.ResourceSku) string {
+	return *sku.Name
+}
+
+func expandAnalysisServicesServerProperties(d *schema.ResourceData) *analysisservices.ServerProperties {
+	adminUsers := expandAnalysisServicesServerAdminUsers(d)
```
katbyte: I don't think there is a need for this function

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+
+	serverProperties := analysisservices.ServerProperties{AsAdministrators: adminUsers}
+
+	//if backupBlobContainerUri, ok := d.GetOk("backup_blob_container_uri"); ok {
+	//	serverProperties.BackupBlobContainerURI = backupBlobContainerUri.(*string)
+	//}
+
+	serverProperties.IPV4FirewallSettings = expandAnalysisServicesServerFirewallSettings(d)
```
katbyte: I don't think there is a need for this function

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+
+	if connectionMode := expandAnalysisServicesServerQuerypoolConnectionMode(d); connectionMode != nil {
```
katbyte: I don't think there is a need for this function

go.sum
```diff
@@ -16,6 +16,7 @@ github.com/Azure/azure-sdk-for-go v21.3.0+incompatible/go.mod h1:9XXNKU+eRnpl9mo
 github.com/Azure/azure-sdk-for-go v29.0.0+incompatible/go.mod h1:9XXNKU+eRnpl9moKnB4QOLf1HestfXbmab5FXxiDBjc=
 github.com/Azure/azure-sdk-for-go v30.0.0+incompatible h1:6o1Yzl7wTBYg+xw0pY4qnalaPmEQolubEEdepo1/kmI=
 github.com/Azure/azure-sdk-for-go v30.0.0+incompatible/go.mod h1:9XXNKU+eRnpl9moKnB4QOLf1HestfXbmab5FXxiDBjc=
+github.com/Azure/azure-sdk-for-go v30.0.0+incompatible/go.mod h1:9XXNKU+eRnpl9moKnB4QOLf1HestfXbmab5FXxiDBjc=
```
katbyte: This looks like a spurious duplicate line?

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+	if location := server.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+
+	d.Set("sku", flattenAnalysisServicesServerSku(server.Sku))
+
+	if serverProps := server.ServerProperties; serverProps != nil {
+		d.Set("admin_users", *flattenAnalysisServicesServerAdminUsers(server.AsAdministrators))
+
+		//if serverProps.BackupBlobContainerURI != nil {
```
katbyte: If this isn't required we should remove the comment

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,449 @@
...
+	d.Set("resource_group_name", resGroup)
+
+	if location := server.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+
+	d.Set("sku", flattenAnalysisServicesServerSku(server.Sku))
+
+	if serverProps := server.ServerProperties; serverProps != nil {
+		d.Set("admin_users", *flattenAnalysisServicesServerAdminUsers(server.AsAdministrators))
```
katbyte: Idealy we should be using serverprops instead of server

katbyte: Also as this is a top level property we shouldn't need a flatten function

azurerm/resource_arm_analysis_services_server_test.go
```diff
@@ -0,0 +1,379 @@
...
+			},
+		},
+	})
+}
+
+func TestAccAzureRMAnalysisServicesServer_firewallSettings(t *testing.T) {
+	resourceName := "azurerm_analysis_services_server.test"
+	ri := tf.AccRandTimeInt()
+
+	config1 := testAccAzureRMAnalysisServicesServer_firewallSettings(ri, testLocation(), true, make([]map[string]string, 0))
```
katbyte: Instead of passing in parameters like this could we just have straight HCL configs instead?

azurerm/resource_arm_analysis_services_server_test.go
```diff
@@ -0,0 +1,379 @@
...
+					resource.TestCheckResourceAttr(resourceName, "ipv4_firewall_rule.1.name", "test2"),
+					resource.TestCheckResourceAttr(resourceName, "ipv4_firewall_rule.1.range_start", "226.202.187.57"),
+					resource.TestCheckResourceAttr(resourceName, "ipv4_firewall_rule.1.range_end", "226.208.192.47"),
+				),
+			},
+		},
+	})
+}
+
+// Currently unsure how to test it as the email addresses need to exist in the AD
```
katbyte: Would it be possible to create the user's via the azuread provider & use them here?

azurerm/resource_arm_analysis_services_server_test.go
```diff
@@ -0,0 +1,379 @@
...
+
+func testAccAzureRMAnalysisServicesServer_basic(rInt int, location string) string {
+	return fmt.Sprintf(`
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-%d"
+  location = "%s"
+}
+
+resource "azurerm_analysis_services_server" "test" {
+  name                = "acctestass%d"
```
katbyte: Could we ensure capitals work for the name here?
```suggestion
  name                = "acctestASS%d"
```

azurerm/resource_arm_analysis_services_server_test.go
```diff
@@ -0,0 +1,379 @@
...
+  location            = "${azurerm_resource_group.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  sku 				  = "B1"
+
+  tags = {
+	label = "test1"
+	env   = "prod"
```
katbyte: Could we ensure capitals work here?
```suggestion
	ENV   = "Prod"
```

r0bnet:
Thanks @katbyte. I currently don't have the possibility to check your annotations but will try to come back to it next week.

azurerm/resource_arm_analysis_services_server.go
r0bnet: removed (in case you mean we can do it inline)

azurerm/resource_arm_analysis_services_server.go
r0bnet: removed (in case you mean we can do it inline)

azurerm/resource_arm_analysis_services_server.go
r0bnet: removed (in case you mean we can do it inline)

azurerm/resource_arm_analysis_services_server.go
r0bnet: removed (in case you mean we can do it inline)

azurerm/resource_arm_analysis_services_server.go
r0bnet: removed (in case you mean we can do it inline)

azurerm/resource_arm_analysis_services_server.go
r0bnet: removed (in case you mean we can do it inline)

azurerm/resource_arm_analysis_services_server.go
r0bnet: renamed the variable in all places

azurerm/resource_arm_analysis_services_server.go
r0bnet: done

azurerm/resource_arm_analysis_services_server.go
r0bnet: done

azurerm/resource_arm_analysis_services_server.go
r0bnet: i'm not sure if i understand correctly. IMHO this function does more than i want to have simply inline. Can you describe in detail?

azurerm/resource_arm_analysis_services_server.go
r0bnet: i'm not sure if i understand correctly. IMHO this function does more than i want to have simply inline. Can you describe in detail?

go.sum
r0bnet: True. Merging error. Removed

azurerm/resource_arm_analysis_services_server.go
r0bnet: removed

azurerm/resource_arm_analysis_services_server.go
r0bnet: > Idealy we should be using serverprops instead of server

don't understand completely. Can you go into more detail?

azurerm/resource_arm_analysis_services_server_test.go
r0bnet: Talked about this with @tombuildsstuff and he proposed to work with a env var which points to a email address in the azure ad. That okay for you? Then everyone can test in their own azure tenant.

azurerm/resource_arm_analysis_services_server_test.go
r0bnet: renamed accordingly

azurerm/resource_arm_analysis_services_server_test.go
r0bnet: changed to uppercase

azurerm/resource_arm_analysis_services_server.go
r0bnet: i'm not sure if i understand correctly. IMHO this function does more than i want to have simply inline. Can you describe in detail?

stefan-rapp:
Do you know when the creation of an Azure Analysis Service resource will be part of the next Version? Need this feature for a customer project? Thx

azurerm/resource_arm_analysis_services_server_test.go
r0bnet: had to revert it as only lowercase is allowed

azurerm/resource_arm_analysis_services_server_test.go
r0bnet: You now need to specify two email addresses as env variables to run this test. Unfortunately it's not working as expected -> test fails but i don't know why.

r0bnet:
So now i fixed the failing test and changed most of the stuff that i was told to.
@stefan-rapp we'll have to wait what the maintainers will say. Sorry for being so late. Hope this will be included in 1.33.0 at least.

stefan-rapp:
Hi @r0bnet, Nice! Thank you very much. Appreciate it! 👍

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,406 @@
...
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			//"backup_blob_container_uri": {
+			//	Type:         schema.TypeString,
+			//	Optional:     true,
+			//	Sensitive:    true,
+			//	ValidateFunc: validate.URLIsHTTPOrHTTPS,
+			//},
```
tombuildsstuff: could we remove this for the moment?

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,406 @@
...
+		return fmt.Errorf("Error creating Analysis Services Server %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("Error waiting for completion of Analysis Services Server %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	server, err := client.GetDetails(ctx, resourceGroup, name)
+	if err != nil {
+		return err
```
tombuildsstuff: to make this more descriptive for users - could we wrap this error?

```suggestion
		return fmt.Errorf("Error retrieving Analytics Services Server %q (Resource Group %q): %+v", name, resourceGroup, err)
```

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	d.Set("name", name)
+	d.Set("resource_group_name", resourceGroup)
+
+	if location := server.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+
+	d.Set("sku", *server.Sku.Name)
```
tombuildsstuff: there's a potential crash here if this is nil/a bad API response; could we make this:

```
if server.Sku != nil {
  d.Set("sku", server.Sku.Name)
}
```

`d.Set` will handle the property being a pointer and de-reference it to an empty string if required

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,406 @@
...
+	if serverProps := server.ServerProperties; serverProps != nil {
+		if serverProps.AsAdministrators == nil || serverProps.AsAdministrators.Members == nil {
+			d.Set("admin_users", []string{})
+		} else {
+			d.Set("admin_users", *serverProps.AsAdministrators.Members)
+		}
+
+		enablePowerBi, fwRules := flattenAnalysisServicesServerFirewallSettings(serverProps)
+		d.Set("enable_power_bi_service", enablePowerBi)
+		d.Set("ipv4_firewall_rule", fwRules)
```
tombuildsstuff: since this is a complex object could we check for errors when setting here?

```suggestion
		if err := d.Set("ipv4_firewall_rule", fwRules); err != nil {
			return fmt.Errorf("Error setting `ipv4_firewall_rule`: %s", err)
		}
```

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,406 @@
...
+
+	firewallSettings := serverProperties.IPV4FirewallSettings
+
+	enablePowerBi = utils.Bool(false)
+	if firewallSettings.EnablePowerBIService != nil {
+		enablePowerBi = firewallSettings.EnablePowerBIService
+	}
+
+	fwRules = make([]interface{}, 0)
+	for _, fwRule := range *firewallSettings.FirewallRules {
```
tombuildsstuff: if `firewallSettings.FirewallRules` is nil (for example, a bad server response) there's a potential crash here; as such could we make this:

```suggestion
	if firewallSettings.FirewallRules != nil {
		for _, fwRule := range *firewallSettings.FirewallRules {
		...
	}
```

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,406 @@
...
+		output := make(map[string]interface{})
+		output["name"] = *fwRule.FirewallRuleName
```
tombuildsstuff: there's a potential panic here if this is nil - as such could we nil-check around this?

```suggestion
		if fwRule.FirewallRuleName != nil {
			output["name"] = *fwRule.FirewallRuleName
		}
```

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,406 @@
...
+		output["range_start"] = *fwRule.RangeStart
```
tombuildsstuff: (same here)

```suggestion
		if fwRule.RangeStart != nil {
			output["range_start"] = *fwRule.RangeStart
		}
```

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,406 @@
...
+		output["range_end"] = *fwRule.RangeEnd
```
tombuildsstuff: (same here)

```suggestion
		if fwRule.RangeEnd != nil {
			output["range_end"] = *fwRule.RangeEnd
		}
```

azurerm/resource_arm_analysis_services_server.go
tombuildsstuff: it doesn't make much difference either way - so I think this is fine as-is

azurerm/resource_arm_analysis_services_server.go
tombuildsstuff: it doesn't make much difference either way - so I think this is fine as-is

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,406 @@
...
+		return fmt.Errorf("Error creating Analysis Services Server %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("Error waiting for completion of Analysis Services Server %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	server, err := client.GetDetails(ctx, resourceGroup, name)
+	if err != nil {
+		return err
```
tombuildsstuff: could we wrap this error to provide a better explanation to users if this fails?

```suggestion
		return fmt.Errorf("Error retrieving Analysis Services Server %q (Resource Group %q): %s", name, resourceGroup, err)
```

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	if server.ID == nil {
+		return fmt.Errorf("Cannot read ID of Analysis Services Server %q (Resource Group %q)", name, resourceGroup)
+	}
+
+	d.SetId(*server.ID)
```
tombuildsstuff: since the ID shouldn't have changed during the Update (and this is a separate method) could we remove this (and the nil-check above?)

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,406 @@
...
+}
+
+func expandAnalysisServicesServerFirewallSettings(d *schema.ResourceData) *analysisservices.IPv4FirewallSettings {
+	firewallSettings := analysisservices.IPv4FirewallSettings{}
+
+	if enablePowerBi, exists := d.GetOkExists("enable_power_bi_service"); exists {
+		firewallSettings.EnablePowerBIService = utils.Bool(enablePowerBi.(bool))
+	} else {
+		firewallSettings.EnablePowerBIService = utils.Bool(false)
+	}
```
tombuildsstuff: since the default value for this will be `false` we can just default this above:

```suggestion
	firewallSettings := analysisservices.IPv4FirewallSettings{
		EnablePowerBIService: utils.Bool(d.Get("enable_power_bi_service").(bool))
	}
```

azurerm/resource_arm_analysis_services_server_test.go
```diff
@@ -0,0 +1,409 @@
...
+				),
+			},
+		},
+	})
+}
+
+// ARM_ACC_EMAIL1 and ARM_ACC_EMAIL2 must be set and existing emails in the tenant's AD to work properly
+func TestAccAzureRMAnalysisServicesServer_adminUsers(t *testing.T) {
+	const ArmAccEmail1 = "ARM_ACC_EMAIL1"
+	const ArmAccEmail2 = "ARM_ACC_EMAIL2"
```
tombuildsstuff: could we make these keys more descriptive? perhaps `ARM_ACCTEST_ADMIN_EMAIL1` and `ARM_ACCTEST_ADMIN_EMAIL2`?

r0bnet:
@tombuildsstuff resolved all the issues :)

tombuildsstuff:
@r0bnet restarted travis :)

azurerm/resource_arm_analysis_services_server.go
```diff
@@ -0,0 +1,403 @@
...
+			RangeEnd:         utils.String(fwRule["range_end"].(string)),
+		}
+	}
+	firewallSettings.FirewallRules = &fwRules
+
+	return &firewallSettings
+}
+
+func flattenAnalysisServicesServerFirewallSettings(serverProperties *analysisservices.ServerProperties) (enablePowerBi *bool, fwRules []interface{}) {
+	if serverProperties.IPV4FirewallSettings == nil {
```
tombuildsstuff: there's a potential crash here if `serverProperties` is nil - so could we make this:

```suggestion
	if serverProperties == nil || serverProperties.IPV4FirewallSettings == nil {
```

website/docs/r/analysis_services_server.html.markdown
```diff
@@ -0,0 +1,84 @@
+---
```
tombuildsstuff: could we add this new resource [to the sidebar](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/azurerm.erb#L389-L398)?

tombuildsstuff:
@r0bnet I hope you don't mind but the tests identified an issue so I pushed a commit to fix that; but the tests now pass / this LGTM 👍 

Screenshot 2019-07-22 at 10 38 49

PR #3725, opened 2019-06-25
r/kubernetes_cluster: fix Azure CLI command for enabling VMSS support

invidian:
'az feature show' only shows the status of the feature.
'az feature register' should be proposed for actually enabling it.

Signed-off-by: Mateusz Gozdek <mgozdekof@gmail.com>


invidian:
Actually, it seems that also 2nd step is required, according to [docs](https://docs.microsoft.com/en-us/azure/aks/cluster-autoscaler#register-scale-set-feature-provider), which is `az provider register --namespace Microsoft.ContainerService`. Perhaps it would be better to just point to the docs then.

invidian:
Thanks for quick merge!

PR #3737, opened 2019-06-27
Increase timeout for azurerm_app_service_plan to accommodate creation of Isolated SKUs that take a long time to provision

nmiodice:
`azurerm_app_service_plan` resources that are created with an `Isolated` sku can take quite some time to create because they resources being created are not shared and take a long time to create within Azure itself. 

This change increases the timeout to 3 hours when creating `azurerm_app_service_plan` resources. This covers nearly all of the scenarios, as most will create in <= 1.5 hours.

Note: I'm aware that this will be fixed by https://github.com/terraform-providers/terraform-provider-azurerm/issues/171 but I thought that since this is an expected scenario that we should fix it up front


nmiodice:
Thanks @katbyte -- one question, this change looks slated for `1.31.0` which is unreleased according to https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/CHANGELOG.md. When will this actually release? What controls this release and how is the cadence set?

Issue #3522, opened 2019-05-24
Create a storage_account_network_rules resource


MattMencel:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Create a `azurerm_storage_account_network_rules` resource so they can be managed separately from the `azurerm_storage_account` resource.

### New or Affected Resource(s)

* azurerm_storage_account_network_rules
* azurerm_storage_account

### Potential Terraform Configuration

```hcl
resource "azurerm_storage_account_network_rules" "network_rules" {
  name = "network_rules
  storage_account = "${azurerm_storage_account.storage_account.id}"
  bypass = "None"
  ip_rules = "${var.myip}"
  virtual_network_subnet_ids = ["${var.my_vnet_list}"]
}
```

lawrenae:
we were looking into this (or similar) recently at work -- is what this is proposing effectively different than https://www.terraform.io/docs/providers/azurerm/r/storage_account.html#network_rules ?

MattMencel:
I'm proposing a separate resource if possible.  Similar to the way it works with `azurerm_sql_firewall_rule`.  I've got a customer that would like to manage the network_rules outside of the storage_account resource.

evenh:
This functionality is very much needed, thanks for reporting and opening up a PR @MattMencel.

derekrprice:
See #4186 for why this is needed.  Currently, a storage account can't be created that allows access from the IPs from a Function App without creating a circular reference that Terraform can't resolve.

geertn:
Not sure if this affects your work but please take note of:
https://github.com/terraform-providers/terraform-provider-azurerm/issues/4574

Just a FYI

tombuildsstuff:
Fixed via #5082

PR #3741, opened 2019-06-27
Initial commit for storage_account_network_rules_association

MattMencel:
This is an initial PR for #3522. Just the test file for `azurerm_storage_account_network_rule_association`.  Looking for feedback and direction for this before I move forward.  If I'm totally going in the wrong direction let me know.

Since network rules can be applied to Keyvaults and Storage Accounts.... I'm wondering if the network rule resource should be `azurerm_network_rules`?  That could be too similar and be confused with `azurerm_network_security_rule` used with NSGs.

``` terraform
resource "azurerm_network_rules" testnr" {
  name                     = "acctestnr%d"
  resource_group_name      = "${azurerm_resource_group.testrg.name}"
  location                 = "${azurerm_resource_group.testrg.location}"
	
  network_rules {
    ip_rules                   = ["127.0.0.1"]
    virtual_network_subnet_ids = ["${azurerm_subnet.test.id}"]
  }
}

resource "azurerm_storage_account_network_rules_association" "testsanra" {
  storage_account_id        = "${azurerm_storage_account_testsa.id}"
  network_rules_id          = "${azurerm_network_rules.testnr.id}"
}
```

Maybe `azurerm_network_acls` then instead?

``` terraform
resource "azurerm_network_acls" testna" {
  name                     = "acctestnr%d"
  resource_group_name      = "${azurerm_resource_group.testrg.name}"
  location                 = "${azurerm_resource_group.testrg.location}"
	
  network_rules {
    ip_rules                   = ["127.0.0.1"]
    virtual_network_subnet_ids = ["${azurerm_subnet.test.id}"]
  }
}

resource "azurerm_storage_account_network_acls_association" "testsanaa" {
  storage_account_id = "${azurerm_storage_account_testsa.id}"
  network_acls_id = "${azurerm_network_rules.testna.id}"
}
```


magodo:
I also looked into this #3522 recently. I came up same idea as your solution but encountered some issue.

I found all  **association** resources in azure binds two **concrete** resources. However, there doesn't exist `azurerm_network_rule` (or anything alike) resource in azure. So I wonder service team should first provide such a resource, then you can proceed to implement both `azurerm_network_rule` and `azurerm_storage_account_network_rules_association` resources in terrafrom.

IMHO, the association resources apply to the case that there already exist two resources in azure, as time goes by, some use case shows high correlation between those two resources, so some one create an associate resource to associate those two concrete resources by id, so as to tell terraform there dependencies.

On the other hand, I have some further info to share with you. I find some existing resources(i will refer to them as **attaching resources**), e.g. [`azurerm_servicebus_queue_authorization_rule`](https://www.terraform.io/docs/providers/azurerm/r/servicebus_queue_authorization_rule.html),  play as part of some concrete resource (e.g. `azurerm_servicebus_queue`). The implementation to those **attaching resources** are based on the concrete resources' backend client in azure Go SDK.

One precondition in this case is that these backend services have native support for those **attaching resources** in API spec (i.e. they have dedicated API path). However, `azurerm_storage_account` has no such support for network rules... 😥

So I wonder how are you going to preceed?

mbfrahry:
Hey @MattMencel. Thanks for taking the time to write out this test file and think about how to add this resource into the provider. Apologies for not taking a look at it sooner but we just added #5082 which adds `azurerm_storage_account_network_rules` to the provider. As of now we're just going to mirror the functionality of the current network rules inside `azurerm_storage_account` into its own resource. That should unlock the circular dependency chain that was occurring. I'm going to close this down but let me know if that PR isn't going to do what we were looking to solve.

PR #3746, opened 2019-06-27
Data source for azurerm_dev_test_virtual_network

kmsisco:
Added a new data source for getting the configuration of a virtual network in a Dev Test Lab.

### DIFF
--- a/azurerm/data_source_dev_test_virtual_network.go
+++ b/azurerm/data_source_dev_test_virtual_network.go
@@ -1,0 +1,182 @@
+package azurerm
+
+import (
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/services/devtestlabs/mgmt/2016-05-15/dtl"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmDevTestVirtualNetwork() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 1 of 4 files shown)


azurerm/data_source_dev_test_virtual_network.go
```diff
@@ -0,0 +1,182 @@
...
+
+			"unique_identifier": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"allowed_subnets": {
+				Type:     schema.TypeList,
+				Computed: true,
+				MaxItems: 1,
```
tombuildsstuff: this won't have any effect since this field is Computed - as such could we remove this?

```suggestion
```

website/docs/d/dev_test_virtual_network.html.markdown
```diff
@@ -0,0 +1,61 @@
...
+
+* `allowed_subnets` - The list of subnets enabled for the virtual network as defined below.
+* `subnet_overrides` - The list of permission overrides for the subnets as defined below.
+* `unique_identifier` - The unique immutable identifier of the virtual network.
+
+---
+
+An `allowed_subnets` block supports the following:
+
+*	`allow_public_ip` - Indicates if this subnet allows public IP addresses. Possible values are `Allow`, `Default` and `Deny`.
```
tombuildsstuff:
```suggestion
* `allow_public_ip` - Indicates if this subnet allows public IP addresses. Possible values are `Allow`, `Default` and `Deny`.
```

kmsisco:
Thanks!

kmsisco:
@katbyte That failure is the same error that AFAICT is preventing subnets from working in dev test labs.  I have a fix for it in PR #3717 but no one has picked it up. For example, virtual machines cannot be created in the DTL and their acceptance tests also fail (InvalidSubnetUsed).

If you merge the fix from #3717 then the tests will pass.

PR #3757, opened 2019-06-29
Update CosmosDB to accept virtual_network_subnet_ids list

### LINKED ISSUE #3625: New resource for specifying Cosmos DB virtual network rules
### Community Note

### Description

In the azurerm_cosmosdb_account resource, configuring allowed subnets is via multiple virtual_network_rule blocks, 1 per subnet. This isn't very flexible if I want to wrap the resource into a consumable module for my customers, and unlike similar implementations such as:
azurerm_sql_virtual_network_rule
azurerm_mysql_virtual_network_rule
azurerm_postgresql_virtual_network_rule

It would be better if the resource could consume either a list of subnet ids or configured via a new resource. This was briefly discussed in the original PR however the rationale at the time pointed to a new resource cosmosdb_account_virtual_network_rule which was never implemented. 

* azurerm_cosmosdb_virtual_network_rule

```hcl
resource "azurerm_cosmosdb_virtual_network_rule" "example" {
  server_name         = "${azurerm_cosmosdb_account.example.name}"
  subnet_id           = "${azurerm_subnet.subnet.id}"
}
```

### References

[https://www.terraform.io/docs/providers/azurerm/r/sql_virtual_network_rule.html](url)
[https://github.com/terraform-providers/terraform-provider-azurerm/pull/1961](url)

notchairmk:
Adds ability to pass in list of virtual_network_subnet_ids. Fixes #3625 

Started to go down the path of emulating network_rules block in resources like storage account, but that didn't match the underlying CosmosDB model from the Azure SDK. Definitely open to suggestions.

### DIFF
--- a/azurerm/data_source_cosmos_db_account.go
+++ b/azurerm/data_source_cosmos_db_account.go
@@ -111,17 +111,12 @@ func dataSourceArmCosmosDbAccount() *schema.Resource {
 				Computed: true,
 			},
 
-			"virtual_network_rule": {
-				Type:     schema.TypeList,
-				Computed: true,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"id": {
-							Type:     schema.TypeString,
-							Computed: true,
-						},
-					},
+			"virtual_network_subnet_ids": {
+				Type: schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
 				},
+				Computed: true,
 			},
 
 			"enable_multiple_write_locations": {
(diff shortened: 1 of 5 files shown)


notchairmk:
Thanks for taking the time to respond.

The main motivation that we've ran into (and potentially the same for the user who opened the issue) is that there is not currently a way to dynamically set the virtual_network_rule block for with versions of Terraform pre-v0.12. Ideally for our use-case, we'd like to pass in a list of subnet IDs and be able to have them dynamically map to the underlying  virtual network rules.

Totally open for other directions as well, if there's a more promising way forward.

Do you see a way of being able to accommodate this with the API as it currently exists? Or would you prefer to hold off for now?

Issue #3763, opened 2019-07-01
Virtual machine scale set remove the publicIPAddressConfiguration when performing simple plan change


alexsapran:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

```bash
$ terraform -v
Terraform v0.11.14
+ provider.azurerm v1.30.1
+ provider.random v2.1.2
+ provider.template v2.1.2
```

### Affected Resource(s)

* `azurerm_virtual_machine_scale_set`

### Terraform Configuration Files

```hcl
resource "azurerm_virtual_machine_scale_set" "main" {
  name                 = "${local.resource_name}"
  location             = "${var.location}"
  resource_group_name  = "${var.resource_group}"
  overprovision        = "${var.overprovision}"
  upgrade_policy_mode  = "Manual"
  automatic_os_upgrade = "${var.automatic_os_upgrade}"

  single_placement_group = "${var.single_placement_group}"
  priority               = "${var.priority}"

  tags {
    environment       = "${var.environment}"
  }

  zones = ["${var.zones}"]

  sku {
    name     = "${var.sku_name}"
    tier     = "${var.sku_tier}"
    capacity = "${var.capacity_count}"
  }

  os_profile {
    computer_name_prefix = "${local.computer_prefix}"
    admin_username       = "${var.admin_username}"
    custom_data          = "${var.custom_data}"
  }

  os_profile_linux_config {
    disable_password_authentication = true

    ssh_keys = {
      path     = "/home/ubuntu/.ssh/authorized_keys"
      key_data = "${file(var.ssh_key)}"
    }
  }

  network_profile {
    name                      = "${local.network_profile_name}-network"
    primary                   = true
    network_security_group_id = "${var.nsg_id}"
    accelerated_networking    = "${var.enable_accelerated_networking}"

    ip_configuration {
      name                                   = "${local.network_profile_name}"
      primary                                = true
      subnet_id                              = "${var.subnet_id}"
      application_security_group_ids         = ["${var.asg_ids}"]
[...]

PR #3767, opened 2019-07-02
azurerm_virtual_machine_scale_set - prevent `public_ip_address_configuration` from being lost during update

alexsapran:
Why 
----
This commit is fixing #3763 when we perform a simple update on an existing scale-set the definition of the scale-set was changed on the `publicIPAddressConfiguration` field. 

This means that any new instances that would be started from the scale-set would not have public IPs when it would be expected to have public IP.

What
----
* Create acceptance tests to mitigate the issue

Start with creating some test coverage around the issue. 

* Fix the parsing for the public ip configuration

Fix the actual issue that caused the public ip configuration of a scale-set to be removed.

Fixes: #3763

### DIFF
--- a/azurerm/resource_arm_virtual_machine_scale_set.go
+++ b/azurerm/resource_arm_virtual_machine_scale_set.go
@@ -1326,8 +1326,7 @@ func flattenAzureRmVirtualMachineScaleSetNetworkProfile(profile *compute.Virtual
 						config["primary"] = *properties.Primary
 					}
 
-					if properties.PublicIPAddressConfiguration != nil {
-						publicIpInfo := properties.PublicIPAddressConfiguration
+					if publicIpInfo := properties.PublicIPAddressConfiguration ; publicIpInfo!= nil {
 						publicIpConfigs := make([]map[string]interface{}, 0, 1)
 						publicIpConfig := make(map[string]interface{})
 						publicIpConfig["name"] = *publicIpInfo.Name
--- a/azurerm/resource_arm_virtual_machine_scale_set.go
+++ b/azurerm/resource_arm_virtual_machine_scale_set.go
@@ -1326,14 +1326,14 @@ func flattenAzureRmVirtualMachineScaleSetNetworkProfile(profile *compute.Virtual
 						config["primary"] = *properties.Primary
 					}
 
-					if publicIpInfo := properties.PublicIPAddressConfiguration ; publicIpInfo!= nil {
+					if publicIpInfo := properties.PublicIPAddressConfiguration; publicIpInfo != nil {
 						publicIpConfigs := make([]map[string]interface{}, 0, 1)
 						publicIpConfig := make(map[string]interface{})
 						if publicIpName := publicIpInfo.Name; publicIpName != nil {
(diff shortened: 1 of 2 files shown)


azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -1331,8 +1331,9 @@ func flattenAzureRmVirtualMachineScaleSetNetworkProfile(profile *compute.Virtual
 						publicIpConfigs := make([]map[string]interface{}, 0, 1)
 						publicIpConfig := make(map[string]interface{})
 						publicIpConfig["name"] = *publicIpInfo.Name
-						publicIpConfig["domain_name_label"] = *publicIpInfo.VirtualMachineScaleSetPublicIPAddressConfigurationProperties.DNSSettings
+						publicIpConfig["domain_name_label"] = *publicIpInfo.VirtualMachineScaleSetPublicIPAddressConfigurationProperties.DNSSettings.DomainNameLabel
```
kwilczynski: Nice catch!

kwilczynski:
Hi @jeffreyCline, the fix is very simple and appears to solve the problem. Do you think you could have a look for us?

azurerm/resource_arm_virtual_machine_scale_set.go
katbyte: Indeed! We are going to definitely need to add some.. many nil checks to this as the original code was lacking many. My suggestion is something like this:
```go
if ipCfg := properties.PublicIPAddressConfiguration; ipCfg != nil {
						block := make(map[string]interface{})
						if v := ipCfg.Name; v != nil {
							block["name"] = *v
						}

						if ipProperties := ipCfg.VirtualMachineScaleSetPublicIPAddressConfigurationProperties; ipProperties != nil {
							if dns := ipProperties.DNSSettings; dns != nil {
								if v := dns.DomainNameLabel; v != nil {
									block["domain_name_label"] = *v
								}
							}
							if v := ipProperties.IdleTimeoutInMinutes; v != nil {
								block["idle_timeout"] = *v
							}
						}
						config["public_ip_address_configuration"] = []map[string]interface{}{block}
					}
```

kwilczynski:
Hi @katbyte and @jeffreyCline, thank you so much for the review!

I believe @alexsapran will be able to include the `nil` checks in the morning. Which brings me to a question: do you have more release-worthy features lined up to warrant a new version release, so that we could upgrade to an official upstream release ourselves? If no, then perhaps enough bug fixes lined up to make a point-release a reality?

katbyte:
@kwilczynski,

given that we just released last friday and there are only a handful unreleased so most likely release date will be next week.

alexsapran:
Hi @jeffreyCline and @katbyte and thank you for taking the time to review my PR. 

I have pushed 2 new commits ea555d0 and 1498394. The first is just to follow the pattern of the nil checks and the second is to follow your recommendation around the additional Nil checks that you requested.

Please let me know what do you think and if you are still +1

kwilczynski:
Hi @katbyte, thank you so much for your help and for cutting the release. Much appreciated!

Issue #3755, opened 2019-06-29
Provision to pass taint value for the agents in AKS cluster


smuthusa:
Currently, we do not have an option to pass the taint value to the agents created in AKS. Please refer for more details https://github.com/terraform-providers/terraform-provider-azurerm/issues/3753.

PR #3786, opened 2019-07-04
Upgrade Azure Go SDK to v31.0.0

dmoser04:
To enable implementation of #3755 I'd like to upgrade the Azure Go SDK dependency to 31.0.0


dmoser04:
Hi @katbyte 

Thank you for your comment. I reverted the update of the containerservice API

Issue #3025, opened 2019-03-08
Data Source for Availability Zones


bpoland:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When deploying VMs, it would be nice to dynamically get the list of availability zones in the region I'm deploying to.  Microsoft's documentation isn't great but it seems that you can only use certain VM SKUs with availability zones, and I found there's an az command that shows the zones for a particular sku:

```az vm list-skus --location eastus2 --zone```

So, maybe a similar VM SKU related data source would work here.  It seems a bit confusing, but I guess that's how Microsoft has organized things...

### New or Affected Resource(s)

* data.azurerm_virtual_machine_sku

### Potential Terraform Configuration

```hcl
data "azurerm_virtual_machine_sku" "test" {
  name     = "Standard_D2s_v3"
  location = "eastus2"
}
```

Then, in the VM you could set the zone:
```hcl
zones = "${(count.index % data.azurerm_virtual_machine_sku.test.zones) + 1}"
```

(as a side note, I feel like the azurerm_virtual_machine documentation on availability zones is a bit confusing -- why isn't the field just "zone" with a single zone number to put the VM in?)

MattMencel:
Hi @tombuildsstuff is there a technical reason this was pushed to the 2.0.0 milestone?  Waiting on API support or something?

I've got a client request for this feature and would be willing to invest some time unless it can't be done at the moment.

tombuildsstuff:
hey @MattMencel 

> is there a technical reason this was pushed to the 2.0.0 milestone? Waiting on API support or something?

From our side we've got plans to write replacements for the VM and VMSS resources in 2.0; coupled with the comment above (which makes a good point: `(as a side note, I feel like the azurerm_virtual_machine documentation on availability zones is a bit confusing -- why isn't the field just "zone" with a single zone number to put the VM in?)`) - which is a breaking change hence we initially bucketed this for 2.0. However I don't believe there's any technical blockers for adding this new Data Source beforehand (we bucketed this since it's related to the other work).

> I've got a client request for this feature and would be willing to invest some time unless it can't be done at the moment.

[...]

PR #3798, opened 2019-07-05
 New Data Source: `azurerm_compute_resource_sku`

MattMencel:
Adding a data resource for Compute Resource SKUs so we can get access to the availability zone data for a particular SKU.

#3025 

I added support for most of the available attributes. I'm not 100% sure of the response data for `restrictions` so there are still a few things marked `//TODO` in the code.

**EXAMPLE USAGE:**
``` terraform
provider "azurerm" {}

data "azurerm_compute_resource_skus" "test" {
  name     = "Standard_DS2_v2"
  location = "centralus"
}

output "name" {
  value = "${data.azurerm_compute_resource_skus.test.name}"
}

output "resource_type" {
  value = "${data.azurerm_compute_resource_skus.test.resource_type}"
}

output "tier" {
  value = "${data.azurerm_compute_resource_skus.test.tier}"
}

output "size" {
  value = "${data.azurerm_compute_resource_skus.test.size}"
}

output "family" {
  value = "${data.azurerm_compute_resource_skus.test.family}"
}

output "location_info" {
  value = "${data.azurerm_compute_resource_skus.test.location_info}"
}

output "zones" {
  value = "${data.azurerm_compute_resource_skus.test.location_info.0.zones}"
}

output "capabilities" {
  value = "${data.azurerm_compute_resource_skus.test.capabilities}"
}

output "costs" {
  value = "${data.azurerm_compute_resource_skus.test.costs}"
}

output "restrictions" {
  value = "${data.azurerm_compute_resource_skus.test.restrictions}"
}
```

**OUTPUT:**
```
data.azurerm_compute_resource_skus.test: Refreshing state...

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

capabilities = [
    {
        name = MaxResourceVolumeMB,
        value = 14336
    },
    {
        name = OSVhdSizeMB,
        value = 1047552
    },
    {
        name = vCPUs,
        value = 2
    },
    {
        name = HyperVGenerations,
        value = V1,V2
    },
    {
        name = MemoryGB,
        value = 7
    },
    {
        name = MaxDataDiskCount,
        value = 8
    },
    {
        name = LowPriorityCapable,
        value = True
    },
    {
        name = PremiumIO,
        value = True
    },
    {
        name = vCPUsAvailable,
        value = 2
    },
    {
        name = ACUs,
        value = 210
    },
    {
        name = vCPUsPerCore,
        value = 1
    },
    {
        name = CombinedTempDiskAndCachedIOPS,
        value = 8000
    },
    {
        name = CombinedTempDiskAndCachedReadBytesPerSecond,
        value = 67108864
    },
    {
        name = CombinedTempDiskAndCachedWriteBytesPerSecond,
        value = 67108864
    },
    {
        name = CachedDiskBytes,
        value = 92341796864
    },
    {
        name = UncachedDiskIOPS,
        value = 6400
    },
    {
        name = UncachedDiskBytesPerSecond,
        value = 100663296
    },
    {
        name = EphemeralOSDiskSupported,
        value = True
    }
]
family = standardDSv2Family
location_info = [
    {
        location = centralus,
        zones = [1 2 3]
    }
]
name = Standard_DS2_v2
resource_type = virtualMachines
restrictions = []
size = DS2_v2
tier = Standard
zones = [
    1,
    2,
    3
]
```


MattMencel:
Yeah, that makes sense. I was just following along with the pattern in the API (resourceSkusClient).

I also see I'm failing tests and need to put some additional fixes in.

MattMencel:
Hi @katbyte,

The resource is renamed to `azurerm_compute_resource_sku`. I also fixed the failing tests.

At some point, if someone really wanted to retrieve the information for multiple SKUs, it would be possible with some additions to the code. Or I could undo my new changes, and have the data resource accept a list of SKUs which it would then return the data for.

I kind of like it only returning one SKU at a time though.

katbyte:
Thanks for the revisions @MattMencel,

I think it's best for this resource to only return for a single resource. If in the future we wanted to return all available SKUs or multiple i think that would be best as a separate resource `azurerm_compute_resource_skus`

azurerm/data_source_compute_resource_sku.go
```diff
@@ -0,0 +1,442 @@
...
+			if err := d.Set("restrictions", restrictions); err != nil {
+				return fmt.Errorf("Error setting `restrictions`: %+v", err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func Contains(a []string, x string) bool {
```
katbyte: Could we move this into the utils package? and rename it: `StringSliceContains()`

azurerm/data_source_compute_resource_sku.go
```diff
@@ -0,0 +1,442 @@
...
+		output["reason_code"] = prop.ReasonCode
+
+		result = append(result, output)
+	}
+	return result
+}
+
+// TODO
+
+// func flattenRestrictionInfo(input *[]compute.ResourceSkuRestrictionInfo) []interface{} {
```
katbyte: Is there a reason this is still a a todo?

azurerm/data_source_compute_resource_sku_test.go
```diff
@@ -0,0 +1,36 @@
...
+	name := "Standard_DS2_v2"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceAzureRMComputeResourceSku_basic(name, location),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(dataSourceName, "name"),
```
katbyte: We should check that the name matches what we expect
```suggestion
					resource.TestCheckResourceAttr(dataSourceName, "name", name),
```

katbyte: An check that some of the other properties are set

azurerm/provider.go
```diff
@@ -516,8 +516,361 @@ func Provider() terraform.ResourceProvider {
 			},
 		},
 
-		DataSourcesMap: dataSources,
-		ResourcesMap:   resources,
+		DataSourcesMap: map[string]*schema.Resource{
```
katbyte: we'll need to revert this change

website/docs/d/compute_resource_sku.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+
+* `family` - The Family of this particular SKU.
+
+* `kind` - The Kind of resources that are supported in this SKU.
+
+* `capacity` - Specifies the number of virtual machines in the scale set.
+
+* `location_info` - A list of locations and availability zones in those locations where the SKU is available.
+
+* `api_versions` - (TODO) The api versions that support this SKU.
```
katbyte: Looks like this is no longer todo?

website/docs/d/compute_resource_sku.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+
+* `costs` - (TODO) Metadata for retrieving price info.
```
katbyte: Looks like this is no longer todo?

azurerm/data_source_compute_resource_sku.go
```diff
@@ -0,0 +1,442 @@
...
+
+	filteredResourceSkus := make([]compute.ResourceSku, 0)
+	for resourceSkus.NotDone() {
+		shouldInclude := false
+		sku := resourceSkus.Value()
+
+		if v, ok := d.GetOkExists("location"); ok {
+			if location := v.(string); location != "" {
+				if Contains(*sku.Locations, location) && *sku.Name == name {
+					shouldInclude = true
```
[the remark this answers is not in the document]
MattMencel: Just clarifying because I'm not 100% positive this is what you meant... but like this?

``` golang

	resourceSkus, err := client.ListComplete(ctx)

	if err != nil {
		return fmt.Errorf("Error loading Compute Resource SKUs: %+v", err)
	}
        for resourceSkus.NotDone() {
		// shouldInclude := false
		// sku := resourceSkus.Value()

		// if v, ok := d.GetOkExists("location"); ok {
		// 	if location := v.(string); location != "" {
		// 		if Contains(*sku.Locations, location) && *sku.Name == name {
		// 			shouldInclude = true
		// 		}
		// 	}
		// } else {
		// 	if *sku.Name == name {
		// 		shouldInclude = true
		// 	}
		// }

		// if shouldInclude {
		// 	filteredResourceSkus = append(filteredResourceSkus, sku)
		// }

		sku := resourceSkus.Value()
		filteredResourceSkus = append(filteredResourceSkus, sku)
		err = resourceSkus.NextWithContext(ctx)
		if err != nil {
			return fmt.Errorf("Error loading Compute Resource SKUs: %+v", err)
		}
	}
```

Assuming this is correct, I'll remove the commented lines.

MattMencel:
Updated my branch to the latest commits and made some corrections. The lint test failed due to an out of memory error.

katbyte:
Hi @MattMencel,

Thank you for this PR and the changes, but upon internal discussion we have decided that the use case of getting the zones for a region/service is probably better served by a more purpose specific `azurerm_zones` datasource that allows a user to specify a Resource (Microsoft.Compute/virtualMachines)/ and a region (westus) returning just the zones. It seems that is the only output from this datasource that would be useful in terraform and is required for other resource types as well. Given this, whilst I’d like to thank you for this PR, since we’re looking to take a different direction here I’m going to close this PR - but apologies this sat idle for so long.

PR #3330, opened 2019-04-29
App Service: Add schedule backup functionality

namku:
Add schedule backup funcionality and documentation.

PR #3804, opened 2019-07-07
r/app_service: support for backups

tombuildsstuff:
This PR supersedes #3330 by @namku - which adds support for Backups to App Services

PR #3330 was sent from the master branch - as such unfortunately it's not possible to rebase it without breaking that PR (due to Github's branch protection) - therefore this includes all of the original commits by @namku plus the changes requested in #3330 (so that we can get this merged).


azurerm/helpers/azure/app_service_schedule_backup.go
```diff
@@ -0,0 +1,207 @@
...
+								Type:     schema.TypeBool,
+								Optional: true,
+								Default:  false,
+							},
+
+							"retention_period_in_days": {
+								Type:         schema.TypeInt,
+								Optional:     true,
+								Default:      30,
+								ValidateFunc: validateRetentionPeriod,
```
katbyte: Would it make more sense to use the build in validation here:
```suggestion
								ValidateFunc: validation.IntBetween(0, 1000),
```

azurerm/helpers/azure/app_service_schedule_backup.go
```diff
@@ -0,0 +1,207 @@
...
+				"schedule": {
+					Type:     schema.TypeList,
+					Required: true,
+					MaxItems: 1,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"frequency_interval": {
+								Type:         schema.TypeInt,
+								Required:     true,
+								ValidateFunc: validateFrequencyInterval,
```
katbyte: Would it make more sense to use the built in validation function here:
```suggestion
								ValidateFunc: validation.IntBetween(0, 9999999),
```

azurerm/helpers/azure/app_service_schedule_backup.go
```diff
@@ -0,0 +1,207 @@
...
+	if input.Enabled != nil {
+		output["enabled"] = *input.Enabled
+	}
+	if input.StorageAccountURL != nil {
+		output["storage_account_url"] = *input.StorageAccountURL
+	}
+
+	schedules := make([]interface{}, 0)
+	if input.BackupSchedule != nil {
+		v := *input.BackupSchedule
```
katbyte: might be a little friendly if we used schedule instead of v here:
```suggestion
		schedule := *input.BackupSchedule
```

tombuildsstuff: schedule's used immediately below - so this is intentionally different since the output objects use that (`schedule` & `schedules`)

tombuildsstuff:
Tests pass:

Screenshot 2019-07-17 at 19 21 09

PR #3807, opened 2019-07-08
New resource - azurerm_storage_account_blob_settings

### LINKED ISSUE #1070: Storage Soft Delete Support
### Community Note

### Description

Hello,

Could we get support for Storage Soft Delete?

Seperate PR requested to add support on the Azure Go SDK - https://github.com/Azure/azure-sdk-for-go/issues/1504

https://azure.microsoft.com/en-us/blog/soft-delete-for-azure-storage-blobs-now-in-public-preview/
https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-soft-delete

Thank you.

benjamin37:
I was not sure if it's the right choice to create a new resource for these storage account settings instead of putting it to the storage account resources. I decided to create a separate resource since it's a separate api call and there are many settings that can be done. I'm happy for any feedback regarding this point.

I only took care of the soft delete settings of the storage account. Other settings like Cors rules can also be set with this client call. so this resource can easily be extended.

Rest API Reference: https://docs.microsoft.com/en-us/rest/api/storagerp/blobservices

(fixes #1070)


katbyte:
Related to #2046 ?

benjamin37:
Hi @katbyte,

I guess this pull request is not related to #2046. This pull request is about soft delete for storage accounts and #2046 is about storage account encryption settings which needs soft delete for *key vaults* (not storage accounts). The Azure REST APIs are different ones.

benjamin37:
Hi @tombuildsstuff & @katbyte,

any news on this pr?

azurerm/resource_arm_storage_account_blob_settings.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	d.Set("resource_group_name", resourceGroupName)
+	d.Set("storage_account_name", storageAccountName)
+
+	if props := resp.BlobServicePropertiesProperties; props != nil {
+		if policy := props.DeleteRetentionPolicy; policy != nil {
+			if policy.Enabled != nil {
+				d.Set("enable_soft_delete", policy.Enabled)
+			}
```
tombuildsstuff: `d.Set` will handle this being a nilable-pointer; as such we can remove the if statement to make this:

```suggestion
			d.Set("enable_soft_delete", policy.Enabled)
```

azurerm/resource_arm_storage_account_blob_settings.go
```diff
@@ -0,0 +1,177 @@
...
+			if policy.Days != nil {
+				d.Set("soft_delete_retention_days", policy.Days)
+			}
```
tombuildsstuff: (as above) `d.Set` will handle this being a nilable pointer (for simple types) as such we can remove the if statement and make this:

```suggestion
			d.Set("soft_delete_retention_days", policy.Days)
```

azurerm/resource_arm_storage_account_blob_settings.go
```diff
@@ -0,0 +1,177 @@
...
+				Optional: true,
+				Default:  false,
+			},
+
+			"soft_delete_retention_days": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      7,
+				ValidateFunc: validation.IntBetween(1, 365),
+			},
```
tombuildsstuff: since the `azurerm_storage_account` resource already supports configuring `queue_properties`, since you can only have one of these would this make more sense within that resource e.g.

```
resource "azurerm_storage_account" "test" {
  ...

  blob_properties {
    soft_deletion_policy {
      days = 7
    }
  }
}
```

we could then assume the `soft_deletion_policy` block being omited/empty wants this disabled - which means that users wouldn't need to set `enabled` or not - WDYT?

from Terraform's perspective that'd create an Expand / Flatten function like so:

```
func expandStorageAccountBlobProperties(input []interface{}) storage.BlobServiceProperties {
	properties := storage.BlobServiceProperties{
		BlobServicePropertiesProperties: &storage.BlobServicePropertiesProperties{
			DeleteRetentionPolicy: &storage.DeleteRetentionPolicy{
				Enabled: utils.Bool(false),
			},
		},
	}
	if len(input) == 0 {
		return properties
	}

	raw := input[0].(map[string]interface{})
	deletionPolicyRaw := raw["delete_retention_policy"].([]interface{})
	if len(deletionPolicyRaw) > 0 {
		policyRaw := deletionPolicyRaw[0].(map[string]interface{})
		days := policyRaw["days"].(int)
		properties.BlobServicePropertiesProperties.DeleteRetentionPolicy.Enabled = utils.Bool(true)
		properties.BlobServicePropertiesProperties.DeleteRetentionPolicy.Days = utils.Int32(int32(days))
	}

	return properties
}

func flattenStorageAccountBlobProperties(input *storage.BlobServiceProperties) []interface{} {
	enabled := false
	days := 0

	if input != nil && input.BlobServicePropertiesProperties != nil && input.BlobServicePropertiesProperties.DeleteRetentionPolicy != nil {
		if policy := input.BlobServicePropertiesProperties.DeleteRetentionPolicy; policy != nil {
			if policy.Enabled != nil {
				enabled = *policy.Enabled
			}

			if policy.Days != nil {
				days = int(*policy.Days)
			}
		}
	}

	output := make([]interface{}, 0)
	if enabled {
		output = append(output, map[string]interface{}{
			"days": days,
		})
	}

	return []interface{}{
		map[string]interface{}{
			"delete_retention_policy": output,
		},
	}
}
```

WDYT?

azurerm/resource_arm_storage_account_blob_settings.go
```diff
@@ -0,0 +1,177 @@
...
+			if !utils.ResponseWasNotFound(existing.Response) {
+				return fmt.Errorf("error checking for presence of existing Storage Account Blob Settings (Storage Account %q / Resource Group %q): %s",
+					storageAccountName, resourceGroupName, err)
+			}
+		}
+
+		if existing.ID != nil && *existing.ID != "" {
+			return tf.ImportAsExistsError(storageAccountBlobSettingsResourceName, *existing.ID)
+		}
+	}
```
tombuildsstuff: since the Storage Account will always have properties, this won't work/we'd need to remove this

azurerm/resource_arm_storage_account_blob_settings.go
benjamin37: Sounds fine for me, I'll change that!

azurerm/resource_arm_storage_account_blob_settings.go
benjamin37: done

azurerm/resource_arm_storage_account_blob_settings.go
benjamin37: done

azurerm/resource_arm_storage_account_blob_settings.go
benjamin37: done

benjamin37:
Hi @tombuildsstuff,

I've moved the soft delete resource to the storage account as you suggested. It feels to better fit there. Can you have an other look at the PR? 

Thanks and kind regards
Benjamin

azurerm/resource_arm_storage_account.go
```diff
@@ -1484,6 +1568,27 @@ func flattenStorageAccountVirtualNetworks(input *[]storage.VirtualNetworkRule) [
...
+			deleteRetentionPolicy := make([]interface{}, 0)
+			attr := make(map[string]interface{})
+
+			attr["days"] = int(*deletePolicy.Days)
+
+			blobProperties["delete_retention_policy"] = append(deleteRetentionPolicy, attr)
+		}
+	}
+
+	return []interface{}{blobProperties}
```
tombuildsstuff: we need to ensure that `delete_retention_policy` is always set to a value, as such could we make this:

```
```suggestion
	return []interface{}{
		"delete_retention_policy": deleteRetentionPolicies,
	}
```
```

File: azurerm/resource_arm_storage_account.go [hunk 6]
```diff
@@ -1484,6 +1568,27 @@ func flattenStorageAccountVirtualNetworks(input *[]storage.VirtualNetworkRule) [
 	return virtualNetworks
 }
 
+func flattenBlobProperties(input storage.BlobServiceProperties) []interface{} {
+	if input.BlobServicePropertiesProperties == nil {
+		return []interface{}{}
+	}
+
+	blobProperties := make(map[string]interface{})
```
tombuildsstuff: due to changes in Terraform 0.12 we need to ensure that all nested fields have a value; as such can we make the `delete_retention_policies` field a top-level field so that we can set this regardless (and conditionally add elements to it)

```suggestion
	deleteRetentionPolicies := make([]interface{}, 0)
```

azurerm/resource_arm_storage_account.go
```diff
@@ -1484,6 +1568,27 @@ func flattenStorageAccountVirtualNetworks(input *[]storage.VirtualNetworkRule) [
...
+
+	if deletePolicy := input.BlobServicePropertiesProperties.DeleteRetentionPolicy; deletePolicy != nil {
+		if enabled := deletePolicy.Enabled; enabled != nil && *enabled {
+			deleteRetentionPolicy := make([]interface{}, 0)
+			attr := make(map[string]interface{})
+
+			attr["days"] = int(*deletePolicy.Days)
+
+			blobProperties["delete_retention_policy"] = append(deleteRetentionPolicy, attr)
```
tombuildsstuff: given there's always a value for all fields (even if a `nil` response is returned from the API - we can always set this field into the nested object via:

```suggestion
			deleteRetentionPolicies = append(deleteRetentionPolicies, map[string]interface{}{
				"days": days,
			})
```

azurerm/resource_arm_storage_account.go
```diff
@@ -1484,6 +1568,27 @@ func flattenStorageAccountVirtualNetworks(input *[]storage.VirtualNetworkRule) [
...
+	if input.BlobServicePropertiesProperties == nil {
+		return []interface{}{}
+	}
+
+	blobProperties := make(map[string]interface{})
+
+	if deletePolicy := input.BlobServicePropertiesProperties.DeleteRetentionPolicy; deletePolicy != nil {
+		if enabled := deletePolicy.Enabled; enabled != nil && *enabled {
+			deleteRetentionPolicy := make([]interface{}, 0)
+			attr := make(map[string]interface{})
```
tombuildsstuff: (which means we can remove these)

website/docs/r/storage_account.html.markdown
```diff
@@ -122,6 +124,12 @@ The following arguments are supported:
 
 ---
 
+A `blob_properties` block supports the following:
+
+* `delete_retention_policy` - (Optional) block as defined below.
```
tombuildsstuff:
```suggestion
* `delete_retention_policy` - (Optional) A `delete_retention_policy ` block as defined below.
```

website/docs/r/storage_account.html.markdown
```diff
@@ -142,6 +150,12 @@ A `custom_domain` block supports the following:
 * `name` - (Optional) The Custom Domain Name to use for the Storage Account, which will be validated by Azure.
 * `use_subdomain` - (Optional) Should the Custom Domain Name be validated by using indirect CNAME validation?
 
+---
+
+A `delete_retention_policy` block supports the following:
+
+* `days` - (Optional) Specifies the number of days that the blob should be retained. The minimum specified value can be 1 and the maximum value can be 365. Defaults to 7.
```
tombuildsstuff: can we quote these values:

```suggestion
* `days` - (Optional) Specifies the number of days that the blob should be retained, between `1` and `365` days. Defaults to `7`.
```

azurerm/resource_arm_storage_account.go
```diff
@@ -1162,6 +1209,19 @@ func resourceArmStorageAccountRead(d *schema.ResourceData, meta interface{}) err
 		return fmt.Errorf("Unable to locate Storage Account %q!", name)
 	}
 
+	blobClient := storageClient.BlobServicesClient
+
+	blobProps, err := blobClient.GetServiceProperties(ctx, resGroup, name)
+	if err != nil {
+		if blobProps.Response.Response != nil && !utils.ResponseWasNotFound(blobProps.Response) {
```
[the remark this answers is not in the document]
benjamin37: done

azurerm/resource_arm_storage_account.go
benjamin37: done

azurerm/resource_arm_storage_account.go
benjamin37: done

azurerm/resource_arm_storage_account.go
benjamin37: done

website/docs/r/storage_account.html.markdown
benjamin37: done

website/docs/r/storage_account.html.markdown
benjamin37: done

azurerm/resource_arm_storage_account.go
benjamin37: done

azurerm/resource_arm_storage_account.go
```diff
@@ -1484,6 +1568,27 @@ func flattenStorageAccountVirtualNetworks(input *[]storage.VirtualNetworkRule) [
...
+	}
+
+	blobProperties := make(map[string]interface{})
+
+	if deletePolicy := input.BlobServicePropertiesProperties.DeleteRetentionPolicy; deletePolicy != nil {
+		if enabled := deletePolicy.Enabled; enabled != nil && *enabled {
+			deleteRetentionPolicy := make([]interface{}, 0)
+			attr := make(map[string]interface{})
+
+			attr["days"] = int(*deletePolicy.Days)
```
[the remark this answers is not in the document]
benjamin37: Done, but I've got one question to this one. 

In the end `days == 0` when the Azure API returns that the `DeleteRetentionPolicy` is `enabled` but the number of days are missing. This should never happen I guess, but `0` is not a valid value for `days`. This will lead to an error I guess. Shouldn't we just throw an error with appropriate message in this case?

benjamin37:
Hi @tombuildsstuff,

thanks for the re-review. I've added your proposals and got one questions about the `days=0` (mentioned in the review comments).

Kind regards
Benjamin

azurerm/resource_arm_storage_account.go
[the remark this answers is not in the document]
tombuildsstuff: The error wouldn't be something the user can do much about unfortunately - in this instance we're setting this to `0` since there's a nested object/list here, and Terraform 0.12 requires that all keys are set in nested elements - so this is mostly to keep Terraform Core happy more than anything

tombuildsstuff:
👋 running the tests it appears this feature's only available for Blob Storage accounts - could we conditionally lookup this field, depending on the account type being used:

```
------- Stdout: -------
=== RUN   TestAccAzureRMStorageAccount_fileStorageWithUpdate
=== PAUSE TestAccAzureRMStorageAccount_fileStorageWithUpdate
=== CONT  TestAccAzureRMStorageAccount_fileStorageWithUpdate
--- FAIL: TestAccAzureRMStorageAccount_fileStorageWithUpdate (47.46s)
    testing.go:569: Step 0 error: errors during apply:
        
        Error: Error reading blob properties for AzureRM Storage Account "unlikely23exst2acctbhsf": storage.BlobServicesClient#GetServiceProperties: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="FeatureNotSupportedForAccount" Message="Blob is not supported for the account."

...

FAIL
```

benjamin37:
Ups, didn't know that. Blob settings won't be accepted for File Storage Accounts. The Test should pass now.

benjamin37:
Hi @katbyte,

thanks for merging! It's not a new resource anymore, we decided to place this in the storage account itself. Maybe you want to rephrase the `CHANGELOG.md`.

Kind regards
Benjamin

Issue #3812, opened 2019-07-09
Virtual Machine scaleset ignores everything under the network_profile


alexsapran:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

```
$ terraform --version
Terraform v0.11.14
+ provider.azurerm v1.30.1
+ provider.template v2.1.2
```

### Affected Resource(s)

* `azurerm_virtual_machine_scale_set`

### Terraform Configuration Files

```hcl
resource "azurerm_resource_group" "test" {
  name     = "alex-test"
  location = "eastus2"
}

resource "azurerm_virtual_network" "test" {
  name                = "acctvn"
  address_space       = ["10.0.0.0/16"]
  location            = "${azurerm_resource_group.test.location}"
  resource_group_name = "${azurerm_resource_group.test.name}"
}

resource "azurerm_subnet" "test" {
  name                 = "acctsub"
  resource_group_name  = "${azurerm_resource_group.test.name}"
  virtual_network_name = "${azurerm_virtual_network.test.name}"
  address_prefix       = "10.0.2.0/24"
}

resource "azurerm_public_ip" "test" {
  name                = "test"
  location            = "${azurerm_resource_group.test.location}"
  resource_group_name = "${azurerm_resource_group.test.name}"
  allocation_method   = "Static"
  domain_name_label   = "${azurerm_resource_group.test.name}"

  tags = {
    environment = "staging"
  }
}

resource "azurerm_lb" "test" {
  name                = "test"
  location            = "${azurerm_resource_group.test.location}"
  resource_group_name = "${azurerm_resource_group.test.name}"

  frontend_ip_configuration {
    name                 = "PublicIPAddress"
    public_ip_address_id = "${azurerm_public_ip.test.id}"
  }
}

resource "azurerm_lb_backend_address_pool" "bpepool" {
  resource_group_name = "${azurerm_resource_group.test.name}"
  loadbalancer_id     = "${azurerm_lb.test.id}"
  name                = "BackEndAddressPool"
}

resource "azurerm_lb_nat_pool" "lbnatpool" {
[...]

PR #3821, opened 2019-07-10
azurerm_virtual_machine_scale_set  - changes made to  `network profile` should now be properly reflected during update

alexsapran:
The acceptance tests added here are describing the issue as reported by
this ticket #3812

Initially tried to remove the custom `Set` method which calculates the hashcode similar to what is suggested by https://github.com/hashicorp/terraform/issues/10520 and https://github.com/hashicorp/terraform/issues/15420 but then we got into a position that the test framework was generating a perpetual diff.

So the current solution was to include the other fields in the hashcode calculation. 

Many thanks to @kwilczynski 

Fixes: #3812
Co-authored-by: Krzysztof Wilczynski <kw@linux.com>


azurerm/resource_arm_virtual_machine_scale_set_test.go
```diff
@@ -174,6 +174,67 @@ func TestUnitAzureRMVirtualMachineScaleSet_basicPublicIP_simpleUpdate(t *testing
 	})
 }
 
+func TestUnitAzureRMVirtualMachineScaleSet_updateNetworkProfile(t *testing.T) {
```
jeffreyCline: Hi @alexsapran can we update the test function name to start with `TestAccAzureRMVirtualMachineScaleSet` instead of `TestUnitAzureRMVirtualMachineScaleSet`

azurerm/resource_arm_virtual_machine_scale_set_test.go
```diff
@@ -174,6 +174,67 @@ func TestUnitAzureRMVirtualMachineScaleSet_basicPublicIP_simpleUpdate(t *testing
...
+				Config: updatedConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVirtualMachineScaleSetIPForwarding(resourceName, true),
+				),
+			},
+		},
+	})
+}
+
+func TestUnitAzureRMVirtualMachineScaleSet_updateNetworkProfile_ipconfiguration_dns_name_label(t *testing.T) {
```
jeffreyCline: same thing here... can we update the test function name to start with `TestAccAzureRMVirtualMachineScaleSet` instead of `TestUnitAzureRMVirtualMachineScaleSet`

alexsapran:
@jeffreyCline I have updated the test method names.

Issue #3722, opened 2019-06-24
Support for setting access policies on storage containers


chpspiir:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Managing retention policies (immutability and legal holds) on storage containers is not yet supported by azurerm_storage_container. It seems that only ACL policies are supported.

It would be great to manage these policies at the infrastructure-level rather than at code level.

I looked at the API docs for the blob service, and I am having trouble figuring out what operation to use:
https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-rest-api. 

Is this because it isn't exposed in the REST API? Or is it a naming confusion that I am not aware of?

If someone could point me in the right direction, I would gladly take a stab at implementing it. 

### New or Affected Resource(s)

* azurerm_storage_container

### Potential Terraform Configuration

```hcl
resource "azurerm_storage_container" "test" {
  name                  = "vhds"
  resource_group_name   = "${azurerm_resource_group.test.name}"
  storage_account_name  = "${azurerm_storage_account.test.name}"
  container_access_type = "private"

  retention_policy {
    type = "legal-hold/immutable-retention" # It might be up for debate whether legal hold and immutable retention should get their own blocks. Mostly for UX.
    retention_time = "7d"
  }
}
```

### References

https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-rest-api. 

* #0000

vbrinza:
(y)

tombuildsstuff:
hey @chpspiir 

Thanks for opening this issue.

This functionality lives within the Data Plane API's for Azure Storage, rather than within the Resource Manager API's - which as such needs a new version of the Azure Storage SDK [than the version we're currently using](http://github.com/Azure/azure-sdk-for-go/tree/master/storage). Unfortunately we're unable to use [the replacement SDK](https://github.com/Azure/azure-storage-blob-go) - since we've been blocked on this for a while (and this is now blocking a large number of issues) we've ended up writing our own SDK, which I'm just finishing up work on at the moment.

[...]

PR #3830, opened 2019-07-11
r/storage_share: support for configuring ACL's and MetaData

tombuildsstuff:
This PR is dependent on #3722 

This PR adds support for configuring both ACL's and MetaData for the `azurerm_storage_share` resource. It also updates the documentation for both the Storage Share and the Storage Share Directory resource.

### DIFF
--- a/azurerm/resource_arm_storage_share_test.go
+++ b/azurerm/resource_arm_storage_share_test.go
@@ -91,6 +91,43 @@ func TestAccAzureRMStorageShare_disappears(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMStorageShare_metaData(t *testing.T) {
+	ri := tf.AccRandTimeInt()
+	rs := strings.ToLower(acctest.RandString(11))
+	location := testLocation()
+	resourceName := "azurerm_storage_share.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMStorageShareDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMStorageShare_metaData(ri, rs, location),
+				Check: resource.ComposeTestCheckFunc(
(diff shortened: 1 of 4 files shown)


azurerm/resource_arm_storage_share_test.go
```diff
@@ -223,32 +297,107 @@ func testCheckAzureRMStorageShareDestroy(s *terraform.State) error {
...
 func testAccAzureRMStorageShare_basic(rInt int, rString string, location string) string {
+	template := testAccAzureRMStorageShare_template(rInt, rString, location)
 	return fmt.Sprintf(`
-resource "azurerm_resource_group" "test" {
-  name     = "acctestRG-%[1]d"
-  location = "%[2]s"
+%s
+
+resource "azurerm_storage_share" "test" {
+  name                 = "testshare%s"
```
katbyte: Should this be
```suggestion
  name                 = "acctestShare%s"
```

azurerm/resource_arm_storage_share_test.go
```diff
@@ -223,32 +297,107 @@ func testCheckAzureRMStorageShareDestroy(s *terraform.State) error {
...
 
-  tags = {
-    environment = "staging"
+resource "azurerm_storage_share" "test" {
+  name                 = "testshare%s"
+  resource_group_name  = "${azurerm_resource_group.test.name}"
+  storage_account_name = "${azurerm_storage_account.test.name}"
+
+  metadata {
+    hello = "world"
```
katbyte: Should we be checking case here?
```suggestion
    HellO = "WorlD!"
```

tombuildsstuff: nope - as in the other PR the casing on the keys is important (needs to be lower-case) - since this affects both I'll send a PR to fix that separately

azurerm/resource_arm_storage_share_test.go
tombuildsstuff: nope - this is a Data Plane API rather than a Resource Manager API; as such this won't be found directly by the Dalek. Instead the Dalek will find the Storage Account which uses the prefix `acctest` so this will get detected/deleted 👍

PR #3832, opened 2019-07-11
r/storage_queue: switching to use the new SDK

tombuildsstuff:
This PR switches the `azurerm_storage_queue` resource over to using the new SDK. It also adds support for configuring MetaData (support for the larger `service_properties` field will come in a subsequent PR)


azurerm/resource_arm_storage_queue.go
```diff
@@ -31,12 +33,19 @@ func resourceArmStorageQueue() *schema.Resource {
...
-			"resource_group_name": azure.SchemaResourceGroupName(),
+
 			"storage_account_name": {
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.NoEmptyStrings,
```
katbyte: There is a `validateArmStorageAccountName` function we could use here?

azurerm/resource_arm_storage_queue.go
```diff
@@ -71,157 +80,156 @@ func validateArmStorageQueueName(v interface{}, k string) (warnings []string, er
...
+			return tf.ImportAsExistsError("azurerm_storage_queue", resourceID)
 		}
 	}
 
-	log.Printf("[INFO] Creating queue %q in storage account %q", name, storageAccountName)
-	options := &storage.QueueServiceOptions{}
-	err = queueReference.Create(options)
-	if err != nil {
-		return fmt.Errorf("Error creating storage queue on Azure: %s", err)
+	if _, err := client.Create(ctx, accountName, accountName, metaData); err != nil {
```
katbyte: accountName twice? 🤔

azurerm/resource_arm_storage_queue_test.go
```diff
@@ -217,8 +245,62 @@ func testAccAzureRMStorageQueue_requiresImport(rInt int, rString string, locatio
...
 `, template)
 }
+
+func testAccAzureRMStorageQueue_metaData(rInt int, rString string, location string) string {
+	template := testAccAzureRMStorageQueue_template(rInt, rString, location)
+	return fmt.Sprintf(`
+%s
+
+resource "azurerm_storage_queue" "test" {
+  name                 = "mysamplequeue-%d"
```
katbyte: Should this be 
```suggestion
  name                 = "acctestQueue-%d"
```

azurerm/resource_arm_storage_queue_test.go
```diff
@@ -217,8 +245,62 @@ func testAccAzureRMStorageQueue_requiresImport(rInt int, rString string, locatio
...
+  storage_account_name = "${azurerm_storage_account.test.name}"
+
+  metadata = {
+    hello = "world"
```
katbyte: Should we be checking case here:
```suggestion
    HellO = "WorlD!"
```

azurerm/resource_arm_storage_queue_test.go
tombuildsstuff: it doesn't need to be (and infact you can't use upper-case characters here) - since this won't be picked up by the Dalek since this is a Data Plane API rather than a Resource Manager API; instead this'll be picked up by the Storage Account being prefixed with `acctest` - so I think we're good here 👍

azurerm/resource_arm_storage_queue_test.go
tombuildsstuff: the API's pedantic about keys here:

```
metadata.Rick:        "" => "M0rty"
metadata.hello:       "world" => "world"
metadata.rick:        "M0rty" => ""
```

since this is a common method, I'll add some validation in the next PR to ensure these keys are lower-cased

PR #3849, opened 2019-07-15
New resource: azurerm_private_dns_zone_a_record

patst:
Allows to create a DNS A record for a private DNS zone.

I tested the changes manually using terraform v0.12.4 and everything worked fine.

The tests (e.g. ` make testacc TESTARGS='-run=TestAccAzureRMPrivateDnsARecord_basic'`) fails at the moment because of the resource group name.
It seems everything than lower case characters  in the resource group name trigger a resource recreation. I could use some help where to look to fix that.

```
--- FAIL: TestAccAzureRMPrivateDnsARecord_basic (103.13s)
    testing.go:568: Step 0 error: After applying this step, the plan was not empty:

        DIFF:

        DESTROY/CREATE: azurerm_private_dns_a_record.test
          id:                  "/subscriptions/3ec0e19a-e080-49fd-8781-b7d5db8ae47b/resourceGroups/acctestrg-190715081152640912/providers/Microsoft.Network/privateDnsZones/acctestzone190715081152640912.com/A/myarecord190715081152640912" => "<computed>"
          name:                "myarecord190715081152640912" => "myarecord190715081152640912"
          records.#:           "2" => "2"
          records.0:           "1.2.3.4" => "1.2.3.4"
          records.1:           "1.2.4.5" => "1.2.4.5"
          resource_group_name: "acctestrg-190715081152640912" => "acctestRG-190715081152640912" (forces new resource)
          tags:                "" => "<computed>"
          ttl:                 "300" => "300"
          zone_name:           "acctestzone190715081152640912.com" => "acctestzone190715081152640912.com"

```

### DIFF
--- a/azurerm/resource_arm_private_dns_a_record.go
+++ b/azurerm/resource_arm_private_dns_a_record.go
@@ -28,7 +28,8 @@ func resourceArmPrivateDnsARecord() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"resource_group_name": azure.SchemaResourceGroupName(),
+			// TODO: make this case sensitive once the API's fixed https://github.com/Azure/azure-rest-api-specs/issues/6641
+			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
 
 			"zone_name": {
 				Type:     schema.TypeString,
@@ -162,7 +163,7 @@ func resourceArmPrivateDnsARecordDelete(d *schema.ResourceData, meta interface{}
 }
 
 func flattenAzureRmPrivateDnsARecords(records *[]privatedns.ARecord) []string {
-	results := make([]string, 0, len(*records))
+	results := make([]string, 0)
 
 	if records != nil {
 		for _, record := range *records {
--- a/azurerm/resource_arm_private_dns_a_record.go
+++ b/azurerm/resource_arm_private_dns_a_record.go
@@ -164,11 +164,16 @@ func resourceArmPrivateDnsARecordDelete(d *schema.ResourceData, meta interface{}
 
 func flattenAzureRmPrivateDnsARecords(records *[]privatedns.ARecord) []string {
 	results := make([]string, 0)
+	if records == nil {
+		return results
+	}
 
-	if records != nil {
-		for _, record := range *records {
-			results = append(results, *record.Ipv4Address)
+	for _, record := range *records {
+		if record.Ipv4Address == nil {
+			continue
 		}
+
+		results = append(results, *record.Ipv4Address)
 	}
 
 	return results
--- a/azurerm/resource_arm_private_dns_a_record.go
+++ b/azurerm/resource_arm_private_dns_a_record.go
@@ -1,0 +1,188 @@
+package azurerm
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/services/privatedns/mgmt/2018-09-01/privatedns"
+	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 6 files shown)


azurerm/resource_arm_private_dns_a_record.go
```diff
@@ -0,0 +1,188 @@
...
+	resp, err := dnsClient.Delete(ctx, resGroup, zoneName, privatedns.A, name, "")
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("Error deleting Private DNS A Record %s: %+v", name, err)
+	}
+
+	return nil
+}
+
+func flattenAzureRmPrivateDnsARecords(records *[]privatedns.ARecord) []string {
+	results := make([]string, 0, len(*records))
```
tombuildsstuff: whilst we're nil-checking it below, there's a potential crash here if `records` is nil - instead we can just create a 0 element array and expand it as needed (whilst it won't be as efficient as pre-allocating this up-front, I'd guess we're talking about 10 elements top - so we're not overly concerned with that performance hit 😄)

website/docs/r/private_dns_a_record.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The Private DNS A Record ID.
+
+## Import
+
+A records can be imported using the `resource id`, e.g.
```
tombuildsstuff: can we make this:

```suggestion
Private DNS A Records can be imported using the `resource id`, e.g.
```

azurerm/resource_arm_private_dns_a_record.go
```diff
@@ -0,0 +1,188 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
```
tombuildsstuff: > It seems everything than lower case characters in the resource group name trigger a resource recreation. I could use some help where to look to fix that.

Unfortunately this would be a bug in the Azure API - whilst officially the Azure API Specification says that the Resource Group names are case insensitive; some API's require that they're case sensitive; as such Terraform treats them as case-sensitive by default. In this instance I'd recommend [opening a bug on the Rest API Specs Repository](https://github.com/Azure/azure-rest-api-specs) where this can be tracked; and adding a TODO with a link to that (see below for an example) - and we can work around this in the interim using:

```suggestion
			// TODO: make this case sensitive once the API's fixed [link]
			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
```

patst:
hi @tombuildsstuff 

Thanks for the review! 
I made the changes and the tests are working now, thanks for the help! 👍

azurerm/resource_arm_private_dns_a_record.go
```diff
@@ -0,0 +1,189 @@
...
+
+	return nil
+}
+
+func flattenAzureRmPrivateDnsARecords(records *[]privatedns.ARecord) []string {
+	results := make([]string, 0)
+
+	if records != nil {
+		for _, record := range *records {
+			results = append(results, *record.Ipv4Address)
```
tombuildsstuff: there's a crash here if this is nil; could we nil-check this?

```suggestion
			if record.Ipv4Address == nil {
				continue
			}
			results = append(results, *record.Ipv4Address)
```

tombuildsstuff:
Tests pass:

Screenshot 2019-07-17 at 18 17 29

PR #3865, opened 2019-07-17
r/storage_blob: adding additional acceptance tests

### LINKED ISSUE #676: azurerm_storage_blob private source_uri is not supported
Trying to copy blob from a storage account to another (different DC locations) within the same current subscription via source_uri parameter and this ends up to fail with a 404. Doing the same operation with the SAS url works fine.
```
* azurerm_storage_blob.test: Error creating storage blob on Azure: storage: service returned error: StatusCode=404, ErrorCode=CannotVerifyCopySource, ErrorMessage=The specified resource does not exist.
```
### Terraform Version
Terraform v0.11.1
Azurerm provider 1.0.0

### Affected Resource(s)
- azurerm_storage_blob

### Expected Behavior
 azurerm_storage_blob able to copy "private" files without any SAS key in source_uri

### Actual Behavior
 azurerm_storage_blob only access to public uris

The current workaround would be to use blobxfer synccopy

tombuildsstuff:
This PR rewrites the acceptance tests for the `azurerm_storage_blob` resource to ensure that all of the currently supported use-cases (and a couple of new ones coming soon, feature toggled off) are covered. With these tests we should be able to swap out [the old Storage SDK](https://github.com/Azure/azure-sdk-for-go/tree/master/storage) for [the replacement Storage SDK](https://github.com/tombuildsstuff/giovanni)

In addition this also removes the `ConflictsWith` from the `content_type` <-> `source_uri` field, since otherwise this leads to a continual diff.

Fixes #676

### DIFF
--- a/azurerm/internal/services/storage/client.go
+++ b/azurerm/internal/services/storage/client.go
@@ -26,6 +26,8 @@ type Client struct {
 // NOTE: this temporarily diverges from the other clients until we move this client in here
 // once we have this, can take an Options like everything else
 func BuildClient(accountsClient storage.AccountsClient) *Client {
+	// TODO: switch Storage Containers to using the storage.BlobContainersClient
+	// (which should fix #2977) when the storage clients have been moved in here
 	return &Client{
 		accountsClient: accountsClient,
 	}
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
Tests pass:

Screenshot 2019-07-17 at 16 05 13

PR #3785, opened 2019-07-04
kubernetes_cluster: support specifying node_resource_group

invidian:
Closes #3354

Signed-off-by: Mateusz Gozdek <mateusz@kinvolk.io>

PR #3869, opened 2019-07-18
Added support for custom resource group name in AKS

kamil-mrzyglod:
Hi,

I wanted to use Terraform for my project, but I found one particular thing missing - lack of possibility to enter a custom name for a node resource group in Azure Kubernetes Service. To make a long story short - by default, AKS creates an additional resource group holding all the infrastructure resources and keeping them separate from the managed cluster. That RG has a dummy name what makes following naming conventions difficult. This is why Microsoft added a `nodeResourceGroup` parameter in the 2019-04-01 API version.

From my investigation, Terraform supports that parameter but only as an output. This PR adds support for `nodeResourceGroup` by introducing a `custom_node_resource_group` parameter. I did not use `node_resource_group` name as it was already added as a computed result of running an AKS deployment.

Kamil

### DIFF
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -75,6 +75,12 @@ func resourceArmKubernetesCluster() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupName(),
 
+			"custom_node_resource_group": {
+				Type:         schema.TypeString,
+				Optional:     true,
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
hey @kamil-mrzyglod 

Thanks for this PR :)

Taking a look through, whilst this PR looks good it appears to be a duplicate of #3785 (which is now in a position where that can be merged) - as such whilst I'd like to thank you for this contribution, I hope you don't mind but I'm going to close this PR in favour of #3785.

Thanks!

PR #3871, opened 2019-07-18
New Resource: `azurerm_cosmosdb_sql_container`

DenheenJ:
Create resource that creates a cosmosdb sql container. Only creates the properties that cannot be changed after creation.

```
resource "azurerm_cosmosdb_sql_container" "test_container" {
  name                = "test-container"
  resource_group_name = "${azurerm_cosmosdb_account.test_cosmos_account.resource_group_name}"
  account_name        = "${azurerm_cosmosdb_account.test_cosmos_account.name}"
  database_name       = "${azurerm_cosmosdb_sql_database.test_db.name}"
  partition_key_paths = "/definition/id"
  unique_key_policy {
    unique_keys {
      paths = ["/definition/something", "/definition/somethingelse"]
    {
  }
}
```

creates unique key policy json as below

    "uniqueKeyPolicy": {
      "uniqueKeys": [
        {
          "paths": [
            "/definition/something",
            "/definition/somethingelse"
          ]
        }
      ]
    }

creates partition key json as below

    "partitionKey": {
      "kind": "Hash",
      "paths": [
        "/definition/id"
      ]
    },


azurerm/resource_arm_cosmosdb_sql_container_test.go
```diff
@@ -0,0 +1,117 @@
...
+		if resp.StatusCode == http.StatusNotFound {
+			return fmt.Errorf("Bad: Cosmos Container '%s' (account: '%s') does not exist", name, account)
+		}
+
+		return nil
+	}
+}
+
+func testAccAzureRMCosmosDbSqlContainer_basic(rInt int, location string) string {
+	res := fmt.Sprintf(`
```
katbyte: Could we change this to
```suggestion
	return fmt.Sprintf(`
```

azurerm/resource_arm_cosmosdb_sql_container.go
```diff
@@ -0,0 +1,215 @@
...
+	var paths []string
+
+	for _, i := range input.(*schema.Set).List() {
+		b := i.(map[string]interface{})
+		c := &[]string{b["path"].(string)}
+		d := *c
+		e := strings.Join(d, "")
+		paths = append(paths, e)
+	}
+	outputs = append(outputs, documentdb.UniqueKey{
```
katbyte: This could be
```suggestion
	return append(outputs, documentdb.UniqueKey{
```

azurerm/resource_arm_cosmosdb_sql_container.go
```diff
@@ -0,0 +1,215 @@
...
+
+	return nil
+}
+
+func expandSQLContainerUniqueKeyPolicy(input interface{}) *[]documentdb.UniqueKey {
+	outputs := make([]documentdb.UniqueKey, 0)
+	var paths []string
+
+	for _, i := range input.(*schema.Set).List() {
+		b := i.(map[string]interface{})
```
katbyte: Could you explain this code? couldn't it instead be:
```
paths = append(paths, b["path"].(string))
```

DenheenJ: i'll try and see what happens, thanks. Before i was just having issues with the value the Azure SDK is returning.

azurerm/resource_arm_cosmosdb_sql_container.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	d.Set("resource_group_name", id.ResourceGroup)
+	d.Set("account_name", id.Account)
+	d.Set("name", name)
+	d.Set("database_name", id.Database)
+	if props := resp.SQLContainerProperties; props != nil {
+		d.Set("name", props.ID)
+	}
+
```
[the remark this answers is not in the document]
katbyte: Yep, you should be reading all values back

azurerm/resource_arm_cosmosdb_sql_container.go
katbyte: I'm not sure why you are making it an string array pointer, dereferencing it, and then joining it?

`&[]string{b["path"].(string)}` makes a string array of b["path"].(string)
`*` then deferences it
`strings.Join()` joins the array,consisting of just `b["path"].(string)`
leaving a string consisting of `b["path"].(string)` if i'm not mistaken?

azurerm/resource_arm_cosmosdb_sql_container.go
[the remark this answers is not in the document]
DenheenJ: hi @katbyte, thanks for that. now reading all values back that are created.

azurerm/resource_arm_cosmosdb_sql_container.go
DenheenJ: I've completely redone this now so may have to do a fresh review.

DenheenJ:
@sawyerh https://www.terraform.io/docs/providers/azurerm/r/cosmosdb_sql_container.html
seems to not be rendering on the index though.

PR #3875, opened 2019-07-18
storage: using the environment as available

tombuildsstuff:
Threading the current environment through to the ClientOptions package so that we can specify it when creating the Clients for the Storage SDK

This allows the Storage SDK to work on multiple clouds


tombuildsstuff:
Tests pass:

Screenshot 2019-07-19 at 10 08 55

PR #3659, opened 2019-06-14
Support for auth_settings in app service slot

MattiasAng:

PR #3897, opened 2019-07-22
r/app_service_slot: support for auth settings

tombuildsstuff:
This PR includes the changes made by @MattiasAng in #3659. Since that PR otherwise looks good, but is unfortunately from the `master` branch, it's not easily in a position where we can rebase that; as such this PR includes the changes made by @MattiasAng in #3659.

After fixing up the merge conflicts I also wrapped the error messages to give a little more explanation when something went wrong

Tests pass:

Screenshot 2019-07-22 at 13 04 51

### DIFF
--- a/azurerm/resource_arm_app_service_slot.go
+++ b/azurerm/resource_arm_app_service_slot.go
@@ -14,9 +14,9 @@ import (
 
 func resourceArmAppServiceSlot() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceArmAppServiceSlotCreateUpdate,
+		Create: resourceArmAppServiceSlotCreate,
 		Read:   resourceArmAppServiceSlotRead,
-		Update: resourceArmAppServiceSlotCreateUpdate,
+		Update: resourceArmAppServiceSlotCreate,
 		Delete: resourceArmAppServiceSlotDelete,
 		Importer: &schema.ResourceImporter{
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_app_service_slot.go
```diff
@@ -186,46 +188,107 @@ func resourceArmAppServiceSlotCreateUpdate(d *schema.ResourceData, meta interfac
...
+	resourceGroup := id.ResourceGroup
+	appServiceName := id.Path["sites"]
+	slot := id.Path["slots"]
+
+	location := azure.NormalizeLocation(d.Get("location").(string))
+	appServicePlanId := d.Get("app_service_plan_id").(string)
+	siteConfig := azure.ExpandAppServiceSiteConfig(d.Get("site_config"))
+	enabled := d.Get("enabled").(bool)
+	httpsOnly := d.Get("https_only").(bool)
+	tags := d.Get("tags").(map[string]interface{})
```
alexsomesan: Should the outcome of this assertion be tested for potential errors?

tombuildsstuff: nah - we do this throughout the provider since the schema's shared - such that whilst this /could/ be an issue, the entire provider would fail if something was wrong; such that this is fine/would be caught by the tests 👍

PR #3901, opened 2019-07-22
New Resource: `azurerm_sql_failover_group`

### LINKED ISSUE #2989: New Resource: Azure SQL Failover Groups
### Community Note

### Description

Add Support Azure SQL Failover Groups. 
https://docs.microsoft.com/en-us/azure/templates/microsoft.sql/2015-05-01-preview/servers/failovergroups

simongh:
I've created a resource for adding SQL Failover Groups. I've added tests & documentation. Hopefully there's not much that I've missed. Still a novice when it comes to Go

(fixes #2989)


leewilson86:
I am very interested in this as a native resource. I currently have this functionality in place using ARM templates, but a proper resource would be much nicer.

It looks like you’ve covered all the key points. 

The only questions I have are:

- does this resource work properly on deletion, when you have databases synchronised into the secondary region? As failover groups will create databases in the secondary region for you, when using ARM templates, TF becomes unaware of the secondary databases existence, meaning an error is thrown by the ARM API on delete as the FOGs contain databases. Have you handled and tested for this eventuality? 

- does this work ok with databases in elastic pools? On creation, update and delete?

- have you considered that in the primary region a database may be in an elastic pool, but when may be moved out as a Single SQL Database? Likewise, databases may be moved from one elastic pool into another, and a Single SQL Database may be moved into an elastic pool. Ideally the secondary region should always mirror the primary region so when failover happens infrastructure is identical. Does your work account for this?

simongh:
Failover groups do create a secondary if they don't exist, but there is no way with the API to prevent this or as far as I can tell make TF create a resource for the secondary. If you have this requirement, you can crate the primary & secondary, then add the failover group. You're then tracking everything. Doing this also gives you control over the sizing etc of the secondary.

Elastic pools shouldn't affect the databases. Certainly the docs make no mention of this, but then the docs are fairly light on detail.

With regard to moving or sizing the databases, failover groups have no control over this. if you need this control, it's best to create & manage the resources as specific items.

This seems to be the intent of failover groups as defined in Azure. I'm not sure that making it do more is the right thing, but then I'm not always right ;-)

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,356 @@
...
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
katbyte: Could we possible validate this? at least with `NoZeroValues`

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,356 @@
...
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: azure.ValidateResourceID,
+						},
+
+						"location": azure.SchemaLocationForDataSource(),
```
katbyte: Is this supposed to be a datasource schema? ie computed

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,356 @@
...
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"mode": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(sql.ReadOnlyEndpointFailoverPolicyDisabled),
+								string(sql.ReadOnlyEndpointFailoverPolicyEnabled),
+							}, true),
```
katbyte: Could we make this case sensitive?
```suggestion
							}, false),
```

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,356 @@
...
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"mode": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(sql.Automatic),
+								string(sql.Manual),
+							}, true),
```
katbyte: Could we make this case sensitive?
```suggestion
							}, false),
```

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,356 @@
...
+		if existing.ID != nil && *existing.ID != "" {
+			return tf.ImportAsExistsError("azurerm_sql_failover_group", *existing.ID)
+		}
+	}
+
+	tags := d.Get("tags").(map[string]interface{})
+
+	properties := sql.FailoverGroup{
+		FailoverGroupProperties: &sql.FailoverGroupProperties{
+			//Databases =
```
katbyte: THis should probably be removed? 🤔

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,356 @@
...
+	v := vs[0].(map[string]interface{})
+
+	mode := sql.ReadWriteEndpointFailoverPolicy(v["mode"].(string))
+	graceMins := int32(v["grace_minutes"].(int))
+
+	policy := &sql.FailoverGroupReadWriteEndpoint{
+		FailoverPolicy: mode,
+	}
+
+	if mode != "manual" {
```
katbyte: Is it a requirement to not set grace min when mode is manual? if so would it make sense to validate that?

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,356 @@
...
+func expandReadOnlyPolicy(d *schema.ResourceData) *sql.FailoverGroupReadOnlyEndpoint {
+	vs := d.Get("readonly_endpoint_failover_policy").([]interface{})
+	if len(vs) == 0 {
+		return nil
+	}
+
+	v := vs[0].(map[string]interface{})
+	mode := sql.ReadOnlyEndpointFailoverPolicy(v["mode"].(string))
+
+	policy := &sql.FailoverGroupReadOnlyEndpoint{
```
katbyte: This could be
```suggestion
	return &sql.FailoverGroupReadOnlyEndpoint{
```

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,356 @@
...
+	return &partners
+}
+
+func flattenPartnerServers(input *[]sql.PartnerInfo) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0)
+
+	if input != nil {
+		for _, server := range *input {
+			info := make(map[string]interface{})
+			info["id"] = *server.ID
```
katbyte: We should be nil checking this
```suggestion
	if v :=server.ID; v != nil {
			info["id"] = *v
	}
```

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,356 @@
...
+			info["location"] = *server.Location
```
katbyte: This should also be nil checked

website/docs/r/sql_failover_group.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+  Manages a SQL Failover Group.
+---
+
+# azurerm_sql_failover_group
+
+Create a failover group of databases on a collection of Azure SQL servers
+
+## Example Usage
+
+```hcl
```
katbyte: Could we format all TF in this example block so the assignments are aligned?

website/docs/r/sql_failover_group.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+resource "azurerm_resource_group" "test" {
```
katbyte: Could we use example instead of test for resource names?
```suggestion
resource "azurerm_resource_group" "example" {
```

website/docs/r/sql_failover_group.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+}
+
+resource azurerm_sql_database db1 {
+  name = "db1"
+  resource_group_name = "${azurerm_sql_server.primary.resource_group_name}"
+  location = "${azurerm_sql_server.primary.location}"
+  server_name = "${azurerm_sql_server.primary.name}"
+}
+
+resource "azurerm_sql_failover_group" "test" {
```
katbyte:
```suggestion
resource "azurerm_sql_failover_group" "example" {
```

website/docs/r/sql_failover_group.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+`partner_servers` supports the following:
+
+* `id` - (Required) the SQL server ID
+
+`read_write_endpoint_failover_policy` supports the following:
+
+* `mode` - (Required) the failover mode. Possible values are `Manual`, `Automatic`
+* `grace_minutes` - Applies only if `mode` is `Automatic`. The grace period in minutes before failover with data loss is attempted
```
katbyte:
```suggestion

* `grace_minutes` - Applies only if `mode` is `Automatic`. The grace period in minutes before failover with data loss is attempted
```

website/docs/r/sql_failover_group.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+    grace_minutes = 60
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the failover group
```
katbyte: We should document changing this forces a new resource

website/docs/r/sql_failover_group.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+
+* `resource_group_name` - (Required) The name of the resource group containing the SQL server
+
+* `server_name` - (Required) The name of the primary SQL server
```
katbyte: We should document changing this forces a new resource

azurerm/resource_arm_sql_failover_group.go
simongh: Yes. The failover group is attached to the sql server, which defines its location. I thought it would be useful to have the failover group return its location as an attribute

azurerm/resource_arm_sql_failover_group.go
simongh: It's not required when the mode is manual

simongh:
I think I've addressed your comments

azurerm/resource_arm_sql_failover_group.go
katbyte: This still needs to be nil checked

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("Error deleting SQL Failover Group %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	return err
+}
+
+func expandReadWritePolicy(d *schema.ResourceData) *sql.FailoverGroupReadWriteEndpoint {
```
katbyte: Could we rename this to:
```suggestion
func expandSqlFailoverGroupReadWritePolicy(d *schema.ResourceData) *sql.FailoverGroupReadWriteEndpoint {
```

as its package wide?

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	if mode != "manual" {
+		policy.FailoverWithDataLossGracePeriodMinutes = utils.Int32(graceMins)
+	}
+
+	return policy
+}
+
+func flattenReadWritePolicy(input *sql.FailoverGroupReadWriteEndpoint) []interface{} {
```
katbyte: and this to
```suggestion
func flattenSqlFailoverGroupReadWritePolicy(input *sql.FailoverGroupReadWriteEndpoint) []interface{} {
```

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,357 @@
...
+
+	policy["mode"] = string(input.FailoverPolicy)
+
+	if input.FailoverWithDataLossGracePeriodMinutes != nil {
+		policy["grace_minutes"] = *input.FailoverWithDataLossGracePeriodMinutes
+	}
+	return []interface{}{policy}
+}
+
+func expandReadOnlyPolicy(d *schema.ResourceData) *sql.FailoverGroupReadOnlyEndpoint {
```
katbyte:
```suggestion
func expandSqlFailoverGroupReadOnlyPolicy(d *schema.ResourceData) *sql.FailoverGroupReadOnlyEndpoint {
```

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,357 @@
...
+
+	v := vs[0].(map[string]interface{})
+	mode := sql.ReadOnlyEndpointFailoverPolicy(v["mode"].(string))
+
+	return &sql.FailoverGroupReadOnlyEndpoint{
+		FailoverPolicy: mode,
+	}
+}
+
+func flattenReadOnlyPolicy(input *sql.FailoverGroupReadOnlyEndpoint) []interface{} {
```
katbyte:
```suggestion
func flattenSqlFailoverGroupReadOnlyPolicy(input *sql.FailoverGroupReadOnlyEndpoint) []interface{} {
```

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,357 @@
...
+		return []interface{}{}
+	}
+
+	policy := make(map[string]interface{})
+	policy["mode"] = string(input.FailoverPolicy)
+
+	return []interface{}{policy}
+}
+
+func expandPartnerServers(d *schema.ResourceData) *[]sql.PartnerInfo {
```
katbyte:
```suggestion
func expandSqlFailoverGroupPartnerServers(d *schema.ResourceData) *[]sql.PartnerInfo {
```

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,357 @@
...
+		id := info["id"].(string)
+		partners = append(partners, sql.PartnerInfo{
+			ID: &id,
+		})
+	}
+
+	return &partners
+}
+
+func flattenPartnerServers(input *[]sql.PartnerInfo) []map[string]interface{} {
```
katbyte:
```suggestion
func flattenSqlFailoverGroupPartnerServers(input *[]sql.PartnerInfo) []map[string]interface{} {
```

azurerm/resource_arm_sql_failover_group.go
katbyte: Unless there's a reason not to (manual and gravce min causes the API to error) I think we should just send all properties here.

simongh:
It causes an error if you set the grace minutes & the mode to manual

```
Failure sending request: StatusCode=400 -- Original Error: Code="FailoverGroupCreateOrUpdateRequestInvalidReadWriteEndpointFieldsForManualPolicy" Message="Grace period value should not be provided when failover policy Manual is selected for the read-write endpoint."
```

simongh:
I've renamed the funcs as suggested. I'm not sure I follow where the nil check you mentioned needs to go. Partner Server is required & you always need at least 1.

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,357 @@
...
+	if props := resp.FailoverGroupProperties; props != nil {
+		if err := d.Set("read_write_endpoint_failover_policy", flattenSqlFailoverGroupReadWritePolicy(props.ReadWriteEndpoint)); err != nil {
+			return fmt.Errorf("Error setting `read_write_endpoint_failover_policy`: %+v", err)
+		}
+
+		if err := d.Set("readonly_endpoint_failover_policy", flattenSqlFailoverGroupReadOnlyPolicy(props.ReadOnlyEndpoint)); err != nil {
+			return fmt.Errorf("Error setting `read_only_endpoint_failover_policy`: %+v", err)
+		}
+
+		d.Set("databases", set.FromStringSlice(*props.Databases))
```
katbyte: We will need to nil check this:
```suggestion
		if props.Databases != nil {
			d.Set("databases", set.FromStringSlice(*props.Databases))
		}
```

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,357 @@
...
+		d.Set("role", string(props.ReplicationRole))
+
+		if err := d.Set("partner_servers", flattenSqlFailoverGroupPartnerServers(props.PartnerServers)); err != nil {
+			return fmt.Errorf("Error setting `partnet_servers`: %+v", err)
```
katbyte: typo:
```suggestion
			return fmt.Errorf("Error setting `partner_servers`: %+v", err)
```

azurerm/resource_arm_sql_failover_group.go
katbyte: ie 
```suggestion
		if v := server.Location; v != nil {
			info["location"] = *v
		}
```

azurerm/resource_arm_sql_failover_group.go
```diff
@@ -0,0 +1,357 @@
...
+
+func flattenSqlFailoverGroupPartnerServers(input *[]sql.PartnerInfo) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0)
+
+	if input != nil {
+		for _, server := range *input {
+			info := make(map[string]interface{})
+
+			if v := server.ID; v != nil {
+				info["id"] = *server.ID
```
katbyte: we can use v here
```suggestion
				info["id"] = *v
```

simongh:
Gotacha, thanks!

I've addressed those comments & rebased to the most recent master, fixing the merge conflict.

simongh:
Don't mind at all.

Failover groups require the partner servers to be in different regions, ideally region replication partners, but any 2 region will do. The tests use testLocation() to get the location in which to create the resources. I'm not sure how to get 2 different locations for the test, other than hard-coding in 2 different regions.

katbyte:
@simongh,

we can use `testAltLocation()` to get another location 🙂

katbyte:
@simongh,

I have added the 2nd location and more of the tests pass now! however i am still getting the following failures:
```
--- FAIL: TestAccAzureRMSqlFailoverGroup_disappears (256.35s)
    testing.go:568: Step 0 error: errors during second follow-up refresh:

        Error: insufficient items for attribute "partner_servers"; must have at least 1


    testing.go:629: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.

        Error: errors during refresh: insufficient items for attribute "partner_servers"; must have at least 1

        State: <nil>
```
and
```
--- FAIL: TestAccAzureRMSqlFailoverGroup_withTags (419.26s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error issuing create/update request for SQL Failover Group "acctestsfg190810054852875950" (Resource Group "acctestRG-190810054852875950", Server "acctestmssql190810054852875950-primary"): sql.FailoverGroupsClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="FailoverGroupUnableToPerformGroupOperationOnDatabases" Message="The operation cannot be performed due to multiple errors." Details=[{"code":"45138","message":"The destination database name 'acctestdb190810054852875950' already exists on the server 'acctestmssql190810054852875950-secondary'."}]

          on /var/folders/bc/2t9ylvbn4lj1_nzj7bxj2kzc0000gn/T/tf-test149816177/main.tf line 36:
          (source code not available)


FAIL
```

simongh:
Thanks for doing that.

I suspect the errors may be down to what happens when you add a database to a failover group. If the database doesn't have a replica on the partner server, one get's created. That may be what's happening here & then the tests fail because there's now a database on the secondary that TF isn't tracking.

We may be able to fix the tests by defining the secondary db in the test HCL, so that both the primary & secondary are tracked by TF before being added to the failover group.

Whilst that may fix the test, is it the right thing to do? I could update the documentation to explain that this could create untracked resources?

leewilson86:
@simongh I’ve encountered this error before when creating FOGs with ARM templates (using the azurerm_deployment_template resource).

If you create databases in both regions with matching names, according to Azure documentation, they are supposed to be linked up, but I always get an error similar to what you mentioned when the FOG is applied to the primary. I assume this is down to the never ending list of poor Azure documentation. If this worked as Azure mentions then I guess like you say making it clear in the documentation that you should create databases on the primary and in the partner locations is the best way to go about it.

I think your idea of defining the databases for both primary and partner locations in the tests is a potentially a bad one, as it is just hiding the issue to get tests to pass. End users are still going to see this issue. We need TF (or something) to effectively delete the databases in the partner locations as well.

Another thing to consider, is what happens when we delete an azurerm_sql_database_server resource which is tracking a partner server. If you do as you say, and only create the database in the primary region, when you create the FOG, it’ll automatically create the database in the partner locations. But this is going to be a problem for the DELETE action of azurerm_sql_database_server resource on a partner database server, as the partner database server will contain databases which TF is unaware of/not tracking, so it will fail to delete the resource and TF will crash due to the Azure API not allowing you to delete a server before all databases are deleted.

I think there’s a lot more to look into for this. Issues spiral beyond just managing the FOG, there are definite issues with deleting partner serves with databases in place. I’ve spent a lot of time trying to get the same functionality working with ARM templates in the absent of a TF resource and I keep hitting he same issue. I had the plan of getting this working in ARM templates and then looking to replicate the functionality in the azurerm TF provider so we have a proper resource to manage FOGs, but as of yet, I can’t think of an elegant way of doing this. And I think you are about to run into the same issues as me.

So please make sure the azurerm_sql_database_server resource is fully tested when it is tracking a partner database server.

simongh:
@leewilson86  I don't think changing the tests is a hiding anything. Having tested it myself, a secondary db gets created, which isn't tracked by TF, but it's existence doesn't prevent the server from being deleted.

@katbyte could the tests be picking up on the fact that an untracked db was created on the secondary server? I've spotted & fixed another issue with the read-write policy. I've also removed the tags support, since although the API supports them, they don't appear anywhere in the UI.

katbyte:
@simongh,

First, even if the tags don't show up in the portal i think we should leave them in? I think that test is failing because it is the only test that attempts to update the resource? and for some reason it is trying to create a new DB. If you create another test that calls update i imagine you will see the same error?

for `insufficient items for attribute "partner_servers"; must have at least 1` i have a feeling the API isn't returning the correct number of `partner_servers` ? at some point.

@leewilson86, could you suggest a terraform config/test that we could write to ensure we have implemented it correctly?

simongh:
It could be. I've tried doing updates, both by changing the resource & by changing the HCL, all without failure.

If you delete a server it also appears to remove the failover group.
If you delete the failover group, it does leave behind the replicated database it made, but then running `apply` just re-adds the failover group ok & destroy destroys as expected.

I'm at a loss as to what's causing the errors. I'll add the tags support back in though. I'm running the acceptance tests now & I'll see what I can figure out.

simongh:
I found a couple of things - typos in the tests. The update now passes. The delete is still failing, even though I thought I found the cause. I'm fairly certain it's something that the test is doing, since it works when I manually run the same HCL.

Since I'm only half certain of what the test is actually doing, would you mind casting an eye over the test? I've likely missed something simple.

katbyte:
👋 @simongh,

So i consulted core internally, and it sounds like that error means upon read, we got zeropartnert servers instead of the expected 1. Looking at the requests i see
```
{
	"location": "West Europe",
	"tags": {},
	"properties": {
		"readWriteEndpoint": {
			"failoverPolicy": "Automatic",
			"failoverWithDataLossGracePeriodMinutes": 60
		},
		"readOnlyEndpoint": {
			"failoverPolicy": "Disabled"
		},
		"replicationRole": "Primary",
		"replicationState": "CATCH_UP",
		"partnerServers": [{
			"id": "/subscriptions/x/resourceGroups/acctestRG-190813120733021836/providers/Microsoft.Sql/servers/acctestmssql190813120733021836-secondary",
			"location": "West US",
			"replicationRole": "Secondary"
		}],
		"databases": ["/subscriptions/x/resourceGroups/acctestRG-190813120733021836/providers/Microsoft.Sql/servers/acctestmssql190813120733021836-primary/databases/acctestdb190813120733021836"]
	},
	"id": "/subscriptions/x/resourceGroups/acctestRG-190813120733021836/providers/Microsoft.Sql/servers/acctestmssql190813120733021836-primary/failoverGroups/acctestsfg190813120733021836",
	"name": "acctestsfg190813120733021836",
	"type": "Microsoft.Sql/servers/failoverGroups"
}
```

a delete, and then a
```{
	"location": "West Europe",
	"tags": {},
	"properties": {
		"readWriteEndpoint": {
			"failoverPolicy": "Automatic",
			"failoverWithDataLossGracePeriodMinutes": 60
		},
		"readOnlyEndpoint": {
			"failoverPolicy": "Disabled"
		},
		"replicationRole": "Primary",
		"replicationState": "CATCH_UP",
		"partnerServers": [],
		"databases": []
	},
	"id": "/subscriptions/x/resourceGroups/acctestRG-190813120733021836/providers/Microsoft.Sql/servers/acctestmssql190813120733021836-primary/failoverGroups/acctestsfg190813120733021836",
	"name": "acctestsfg190813120733021836",
	"type": "Microsoft.Sql/servers/failoverGroups"
}
```

So it seems like the API starts a delete and gets a:
```
{
	"operation": "DropFailoverGroup",
	"startTime": "2019-08-13T19:11:58.047Z"
}
```

and when terraform does a get on it next, it still returns the resource as its not done deleting yet. You may have to add a poll there so the delete call doesn't return until it succeeds?

simongh:
> You may have to add a poll there so the delete call doesn't return until it succeeds?

Do you some existing resource that does this that I can look at it & get the test working?

katbyte:
@simongh,

the cosmos DB resources do this:
```go
//the SDK now will return a `WasNotFound` response even when still deleting
	stateConf := &resource.StateChangeConf{
		Pending:    []string{"Deleting"},
		Target:     []string{"NotFound"},
		Timeout:    60 * time.Minute,
		MinTimeout: 30 * time.Second,
		Refresh: func() (interface{}, string, error) {

			resp, err2 := client.Get(ctx, resourceGroup, name)
			if err2 != nil {

					return resp, "NotFound", nil
				}
				return nil, "", fmt.Errorf("Error reading CosmosDB Account %q after delete (Resource Group %q): %+v", name, resourceGroup, err2)
			}

			return resp, "Deleting", nil
		},
	}
	if _, err = stateConf.WaitForState(); err != nil {
		return fmt.Errorf("Waiting forCosmosDB Account %q to delete (Resource Group %q): %+v", name, resourceGroup, err)
	}
```

simongh:
I think I've found the problem. The resource does a `future.WaitForCompletionRef` and the acceptance test didn't. I modified the test to use this & it's working as expected.

I've also updated the documentation to outline what happens with secondary databases

katbyte:
I hope you don't mind but i pushed some changes to fix the merge errors.

simongh:
That's great. Thanks for your help.

PR #3913, opened 2019-07-23
azurerm_role_definition - enture role_definition_id is correctly set if left empty after after create

notchairmk:
We encountered the issue where `role_definition_id` remained nil after create, returning 409s on `RoleDefinitionsClient.CreateOrUpdate` with code `RoleDefinitionWithSameNameExists`. 

Updating role definition resource was failing because `resourceArmRoleDefinitionCreateUpdate` checks if `role_definition_id` is nil and, if so, creates a new uuid  and calls `client.CreateOrUpdate` which attempts to create a new resource instead of updating.

Thanks!


notchairmk:
Updated to parse `role_definition_id` and added a test. Can't figure out why CI lint is failing intermittently though.

Issue #3929, opened 2019-07-25
1.32 azurerm_api_management_backend: 'certificate' should not be empty


maximbaz:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.12.5
AzureRM Provider: 1.32.0

### Affected Resource(s)

* `azurerm_api_management_backend`

### Terraform Configuration Files

```hcl
resource "azurerm_api_management_backend" "backend" {
  name                = "XXX"
  title               = "XXX"
  resource_group_name = "XXX"
  api_management_name = "XXX"
  protocol            = "http"
  url                 = "XXX"
  resource_id         = "XXX"

  credentials {
      header = {
        x-functions-key = "XXX"
      }
    }
}
```

### `terraform apply` output

```
Error: Error creating/updating backend "backend" (API Management Service "XXX" / Resource Group "XXX"): apimanagement.BackendClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="ValidationError" Message="One or more fields contain incorrect values:" Details=[{"code":"ValidationError","message":"'certificate' should not be empty.","target":"credentials.certificate"}]
```

### Notes

I'm creating `azurerm_api_management_backend` with an intention to allow API management to access an Azure Function. 

The docs describe `credentials.certificate` to be an optional field, and I've checked the code that it is indeed marked as optional.

If I understand this correctly, the Azure itself is refusing the request with the error message that `credentials.certificate` is required. I'm not sure how to report an upstream issue (if it is indeed an upstream issue), please help.

/cc @hbuckle, @katbyte who worked on the PR #3676 introducing this functionality.
/cc @AndiA92

mbfrahry:
Hey @maximbaz, thanks for opening this issue and apologies that you ran into it in the first place. We found and fixed the issue and it'll make it into the next release.

maximbaz:
Thanks, I tried locally and confirm that the fix is working 👍

PR #3931, opened 2019-07-25
Sending nil certificate when no certificate is specified

mbfrahry:
This PR addresses #3929 by sending a nil certificate object instead of an empty certificate object to the api.


mbfrahry:
image

PR #3946, opened 2019-07-26
clarification on which API to use

bytejunkie:
I suspect since this documentation was written the new API has been released, but the required permissions are in the Legacy. I dont feel like this is clear enough, so have added some text to make it more obvious.

### DIFF
--- a/website/docs/r/azuread_application.html.markdown
+++ b/website/docs/r/azuread_application.html.markdown
@@ -13,7 +13,7 @@ Manages an Application within Azure Active Directory.
 
 ~> **NOTE:** The Azure Active Directory resources have been split out into [a new AzureAD Provider](http://terraform.io/docs/providers/azuread/index.html) - as such the AzureAD resources within the AzureRM Provider are deprecated and will be removed in the next major version (2.0). Information on how to migrate from the existing resources to the new AzureAD Provider [can be found here](../guides/migrating-to-azuread.html).
 
--> **NOTE:** If you're authenticating using a Service Principal then it must have permissions to both `Read and write all applications` and `Sign in and read user profile` within the `Windows Azure Active Directory` API.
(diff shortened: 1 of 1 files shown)


website/docs/r/azuread_application.html.markdown
```diff
@@ -13,7 +13,7 @@ Manages an Application within Azure Active Directory.
 
 ~> **NOTE:** The Azure Active Directory resources have been split out into [a new AzureAD Provider](http://terraform.io/docs/providers/azuread/index.html) - as such the AzureAD resources within the AzureRM Provider are deprecated and will be removed in the next major version (2.0). Information on how to migrate from the existing resources to the new AzureAD Provider [can be found here](../guides/migrating-to-azuread.html).
 
--> **NOTE:** If you're authenticating using a Service Principal then it must have permissions to both `Read and write all applications` and `Sign in and read user profile` within the `Windows Azure Active Directory` API.
+-> **NOTE:** If you're authenticating using a Service Principal then it must have permissions to both `Read and write all applications` and `Sign in and read user profile` within the `Windows Azure Active Directory` API. Note that this is the Legacy Azure Active Directory API, not the new Microsoft Graph API.
```
katbyte:
```suggestion
-> **NOTE:** If you're authenticating using a Service Principal then it must have permissions to both `Read and write all applications` and `Sign in and read user profile` within the legacy `Windows Azure Active Directory` API.
```

bytejunkie: I dont think your edit adds any more clarity. if you note in mine i used the specific names as per the msft page, which is very important in saving users time. I often feel like TF docco is too brief, too terse, this is one of those times.

image

image

katbyte: 🤔 How about:

```suggestion
-> **NOTE:** If you're authenticating using a Service Principal then it must have permissions to both `Read and write all applications` and `Sign in and read user profile` within the legacy `Azure Active Directory Graph` API, *not* the `Microsoft Graph` API.
```

bytejunkie:
I think that's clearer, dont you?

Matt

katbyte:
@bytejunkie,

are you referring to my last suggestion? the email reply went to the top level.

katbyte:
I've done with the latest suggestion, thanks for the input!

PR #3949, opened 2019-07-28
azurerm_function_app - support for cors

hbuckle:
Another copy and paste job from the app service resource to support cors for functions

### DIFF
--- a/azurerm/resource_arm_function_app.go
+++ b/azurerm/resource_arm_function_app.go
@@ -199,6 +199,26 @@ func resourceArmFunctionApp() *schema.Resource {
 							Optional: true,
 							Computed: true,
 						},
+						"cors": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Computed: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"allowed_origins": {
+										Type:     schema.TypeSet,
+										Required: true,
+										Elem:     &schema.Schema{Type: schema.TypeString},
+									},
+									"support_credentials": {
+										Type:     schema.TypeBool,
+										Optional: true,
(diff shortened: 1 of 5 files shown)


azurerm/resource_arm_function_app.go
```diff
@@ -199,6 +199,26 @@ func resourceArmFunctionApp() *schema.Resource {
 							Optional: true,
 							Computed: true,
 						},
+						"cors": {
```
katbyte: If this is the same schema used in `app_service` could we maybe create a shares schema function here?

azurerm/resource_arm_function_app.go
```diff
@@ -681,6 +701,12 @@ func expandFunctionAppSiteConfig(d *schema.ResourceData) web.SiteConfig {
 		siteConfig.LinuxFxVersion = utils.String(v.(string))
 	}
 
+	if v, ok := config["cors"]; ok {
+		corsSettings := v.(interface{})
+		expand := azure.ExpandAppServiceCorsSettings(corsSettings)
```
katbyte: As this is now used by two resources might be more aptly named:
```suggestion
		expand := azure.ExpandWebCorsSettings(corsSettings)
```

azurerm/resource_arm_function_app.go
```diff
@@ -709,6 +735,8 @@ func flattenFunctionAppSiteConfig(input *web.SiteConfig) []interface{} {
 		result["linux_fx_version"] = *input.LinuxFxVersion
 	}
 
+	result["cors"] = azure.FlattenAppServiceCorsSettings(input.Cors)
```
katbyte: As above

hbuckle:
@katbyte - I've moved the relevant bits into a new web.go helper

PR #3987, opened 2019-08-01
azurerm_monitor_diagnostic_setting - added LogAnalyticsDestinationType

KimShepherdVibrato:
Added the argument log_analytics_destination_type to the resource azurerm_monitor_diagnostic_setting.

This allows logs sent to a Log Analytics Workspace to go to dedicated resource tables, instead of to the AzureDiagnostics table.  This is only relevant to Azure Data Factory right now, but is planned to be extended to all resources over time.

https://azure.microsoft.com/en-au/updates/adf-logs-in-dedicated-tables/
https://docs.microsoft.com/en-us/azure/azure-monitor/platform/diagnostic-logs-stream-log-store#azure-diagnostics-vs-resource-specific

### DIFF
--- a/azurerm/resource_arm_monitor_diagnostic_setting.go
+++ b/azurerm/resource_arm_monitor_diagnostic_setting.go
@@ -73,6 +73,19 @@ func resourceArmMonitorDiagnosticSetting() *schema.Resource {
 				ValidateFunc: azure.ValidateResourceID,
 			},
 
+			"log_analytics_destination_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
+					v := val.(string)
+					if v != "Dedicated" {
+						errs = append(errs, fmt.Errorf("%q must be 'Dedicated'", key, v))
+					}
+					return
+				},
+			},
+
 			"log": {
 				Type:     schema.TypeSet,
 				Optional: true,
@@ -232,6 +245,11 @@ func resourceArmMonitorDiagnosticSettingCreateUpdate(d *schema.ResourceData, met
 	if storageAccountId != "" {
 		properties.DiagnosticSettings.StorageAccountID = utils.String(storageAccountId)
 		valid = true
+	}
+
+	logAnalyticsDestinationType := d.Get("log_analytics_destination_type").(string)
+	if logAnalyticsDestinationType != "" {
+		properties.DiagnosticSettings.LogAnalyticsDestinationType = utils.String(logAnalyticsDestinationType)
 	}
 
 	if !valid {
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_monitor_diagnostic_setting.go
```diff
@@ -73,6 +73,19 @@ func resourceArmMonitorDiagnosticSetting() *schema.Resource {
 				ValidateFunc: azure.ValidateResourceID,
 			},
 
+			"log_analytics_destination_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
```
katbyte: I believe this could be:
```suggestion
				ValidateFunc: validation.StringInSlice([]string{"Dedicated"}),
```

azurerm/resource_arm_monitor_diagnostic_setting.go
```diff
@@ -234,6 +247,11 @@ func resourceArmMonitorDiagnosticSettingCreateUpdate(d *schema.ResourceData, met
 		valid = true
 	}
 
+	logAnalyticsDestinationType := d.Get("log_analytics_destination_type").(string)
+	if logAnalyticsDestinationType != "" {
```
katbyte: This could be simplified to
```suggestion
	if v := d.Get("log_analytics_destination_type").(string); v != "" {
		properties.DiagnosticSettings.LogAnalyticsDestinationType = &v
	}
```

katbyte: Could we also check that if a type is specified that a `log_analytics_workspace_id` is too?

KimShepherdVibrato: I've made the simplification and added a check that the workspaceId is set.

azurerm/resource_arm_monitor_diagnostic_setting.go
KimShepherdVibrato: I've made the change to the ValidateFunc.  The property seems to be case sensitive so I've set ignoreCase to false

KimShepherdVibrato:
I've added the log_analytics_destination_type to the logAnalyticsWorkspace.  Do you think this is adequate, or should I create a dedicated test?

KimShepherdVibrato:
Hi @katbyte ,

I must admit the test changes were a bit of Cargo Cult Coding, I didn't really understand and I think I must have just run the unit test suite.

I thought I had tested using the log_analytics_destination_type with a resource it wasn't enabled for, but I must have been mistaken as trying again this morning I get an empty string when refreshing on a key vault.

I've reverted the change to the existing test and added a new test: TestAccAzureRMMonitorDiagnosticSetting_logAnalyticsWorkspaceDedicated

This test uses an Azure Datafactory to correctly test the Dedicated diagnostics.  I've confirmed the test passes for me.

tombuildsstuff:
Tests pass:

Screenshot 2019-08-09 at 10 50 16

PR #3994, opened 2019-08-01
Updated api_management_api_policy for XML file explanation

d4rkd0s:
Added a note for those seeking to utilize `-PolicyFilePath` from https://docs.microsoft.com/en-us/powershell/module/azurerm.apimanagement/set-azurermapimanagementpolicy?view=azurermps-6.13.0#parameters

Tested in Azure, this works and would prevent work on implementing PolicyFilePath as another parameter.

### DIFF
--- a/website/docs/r/api_management_api_policy.html.markdown
+++ b/website/docs/r/api_management_api_policy.html.markdown
@@ -46,7 +46,7 @@ The following arguments are supported:
 
 * `resource_group_name` - (Required) The name of the Resource Group in which the API Management Service exists. Changing this forces a new resource to be created.
 
-* `xml_content` - (Optional) The XML Content for this Policy.
+* `xml_content` - (Optional) The XML Content for this Policy as a string. Note instead of parameter for Microsoft's PolicyFilePath. xml-content can take a file by using Terraform's [file function](https://www.terraform.io/docs/configuration/functions/file.html)
 
 * `xml_link` - (Optional) A link to a Policy XML Document, which must be publicly available.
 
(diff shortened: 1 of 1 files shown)


d4rkd0s:
Originally I was going open up a Feature Request for adding support for `PolicyFilePath` but after researching I found a simple solution that wouldn't require a new feature. I've documented this simple solution for others looking for the same functionality, having an answer available in the docs.

website/docs/r/api_management_api_policy.html.markdown
```diff
@@ -46,7 +46,7 @@ The following arguments are supported:
 
 * `resource_group_name` - (Required) The name of the Resource Group in which the API Management Service exists. Changing this forces a new resource to be created.
 
-* `xml_content` - (Optional) The XML Content for this Policy.
+* `xml_content` - (Optional) The XML Content for this Policy as a string. Note instead of parameter for Microsoft's PolicyFilePath. xml-content can take a file by using Terraform's [file function](https://www.terraform.io/docs/configuration/functions/file.html)
```
katbyte: I think this could sound better as:
```suggestion
* `xml_content` - (Optional) The XML Content for this Policy as a string. An XML file can be used here with Terraform's [file function](https://www.terraform.io/docs/configuration/functions/file.html) that is similar to Microsoft's `PolicyFilePath` option.
```
WDYT?

d4rkd0s:
@katbyte I like it, I've accepted the "suggestion" on the wording change. Thanks for that, I was a bit indifferent on the wording I had come up with.

PR #4013, opened 2019-08-06
azurerm_container_group - allow empty log_type

miikka:
This is not documented anywhere, as far as I can tell, but if you leave out the log type when creating a container group, you get the behavior of both `ContainerInsights` and `ContainerInstanceLogs` at once: there will be both the performance metrics (insights) and the container log output. This is the default when you create a container via Azure CLI. This patch enables the same via Terraform.

### DIFF
--- a/azurerm/resource_arm_container_group.go
+++ b/azurerm/resource_arm_container_group.go
@@ -382,7 +382,7 @@ func resourceArmContainerGroup() *schema.Resource {
 
 									"log_type": {
 										Type:     schema.TypeString,
-										Required: true,
+										Optional: true,
 										ForceNew: true,
 										ValidateFunc: validation.StringInSlice([]string{
 											string(containerinstance.ContainerInsights),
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_container_group.go
```diff
@@ -1241,23 +1241,26 @@ func expandContainerGroupDiagnostics(input []interface{}) *containerinstance.Con
...
+		logAnalytics.LogType = containerinstance.LogAnalyticsLogType(logType)
+
+		metadataMap := analyticsV["metadata"].(map[string]interface{})
+		metadata := make(map[string]*string)
+		for k, v := range metadataMap {
+			strValue := v.(string)
+			metadata[k] = &strValue
+		}
+
+		logAnalytics.Metadata = metadata
```
tombuildsstuff: double-checking this: is it possible to set the Metadata without a log type?

miikka: No – if you try to do it, Azure responds with status 400 and a message saying that you can't set metadata with log type `'<null>'`. I'll look into creating an acceptance test.

miikka:
I added an acceptance test. This is my first time contributing to this project, so not 100% sure if I got it right, but at least it passes with my changes and fails without them.

miikka:
Looks like the Travis build failed because there was some kind of intermittent failure with Github. There's a bunch of entries like this in [the logs](https://travis-ci.org/terraform-providers/terraform-provider-azurerm/jobs/568360681) for `make test`:

```
go: github.com/smartystreets/goconvey@v0.0.0-20180222194500-ef6db91d284a: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /home/travis/gopath/pkg/mod/cache/vcs/1693d9c59c0c1838de53cecb8601eec3084f084265b72c82e78de941f1c8f58c: exit status 128:
	fatal: unable to access 'https://github.com/smartystreets/goconvey/': Failed to connect to github.com port 443: Connection timed out
```

tombuildsstuff:
Tests pass:

Screenshot 2019-08-08 at 16 30 04

PR #4022, opened 2019-08-06
refactor: moving Locks into the `internal` directory

tombuildsstuff:
Continues moving things into the internal directory, which allows us to refactor the Data Sources and Resources into Packages longer-term


tombuildsstuff:
AccTests for LB's look good 👍

PR #4039, opened 2019-08-08
provider: misc golang inspection cleanup

katbyte:


azurerm/resource_arm_storage_share.go
```diff
@@ -313,8 +313,7 @@ func resourceArmStorageShareDelete(d *schema.ResourceData, meta interface{}) err
 		return fmt.Errorf("Error building File Share Client for Storage Account %q (Resource Group %q): %s", id.AccountName, *resourceGroup, err)
 	}
 
-	deleteSnapshots := true
-	if _, err := client.Delete(ctx, id.AccountName, id.ShareName, deleteSnapshots); err != nil {
+	if _, err := client.Delete(ctx, id.AccountName, id.ShareName, true); err != nil {
```
tombuildsstuff: can we revert this, so that this is clearer what `true` is here?

katbyte: can do, just to used to seeing param names in goland 🙃

PR #4048, opened 2019-08-09
New Resource: `azurerm_mariadb_virtual_network_rule`

dintskirveli:


azurerm/resource_arm_mariadb_virtual_network_rule.go
```diff
@@ -0,0 +1,242 @@
...
+			}
+		}
+
+		if existing.ID != nil && *existing.ID != "" {
+			return tf.ImportAsExistsError("azurerm_mariadb_virtual_network_rule", *existing.ID)
+		}
+	}
+
+	// due to a bug in the API we have to ensure the Subnet's configured correctly or the API call will timeout
+	// BUG: https://github.com/Azure/azure-rest-api-specs/issues/3719
```
katbyte: according to the link this was closed in nov 2018?

dintskirveli: This is a direct copy of the MySQL version of this code, with relevant API calls replaced. Because that comment references PostgresSQL, I imagine the MySQL version was implemented the same way -- a copy. I imagine the tests will require the same (minimal) effort, as will the documentation. We are already using this to provision MariaDB servers

azurerm/resource_arm_mariadb_virtual_network_rule.go
```diff
@@ -0,0 +1,242 @@
...
+	subnetsClient := meta.(*ArmClient).subnetClient
```
caius: Looks like this needs to be `subnetsClient := meta.(*ArmClient).network.SubnetsClient` on master now, else `make build` fails with

```
# github.com/terraform-providers/terraform-provider-azurerm/azurerm
azurerm/resource_arm_mariadb_virtual_network_rule.go:80:36: meta.(*ArmClient).subnetClient undefined (type *ArmClient has no field or method subnetClient)
```

dintskirveli: No problem, will update.

azurerm/resource_arm_mariadb_virtual_network_rule.go
katbyte: Most likey, however it does indicate this is no longer needed. So could we remove the workaround in this PR as it seems like its no longer required?

dintskirveli: @katbyte, I thought about this some more, and I'm not comfortable taking this out for a couple of reasons:

- If the code is ever refactored to avoid duplication (and it should be), it will make things more difficult for the person making the change because it introduces an additional difference that needs to be inspected.
- The behavior here is reasonable. It's a sanity check for making sure the subnet exists, and it'll 404 if it doesn't with a clear and easy to understand error.

azurerm/resource_arm_mariadb_virtual_network_rule.go
dintskirveli: Done 👍

azurerm/resource_arm_mariadb_virtual_network_rule.go
katbyte: That's a fair response, we'll investigate this & the other resources separately from this PR.

azurerm/resource_arm_mariadb_virtual_network_rule_test.go
```diff
@@ -261,7 +261,7 @@ resource "azurerm_mariadb_server" "test" {
   resource_group_name          = "${azurerm_resource_group.test.name}"
   administrator_login          = "acctestun"
   administrator_login_password = "H@Sh1CoR3!"
-  version                      = "5.6"
+  version                      = "10.2"
```
dintskirveli: Ah, thanks @katbyte! Still working out app credentials to get acceptance tests to run on our end.

PR #4050, opened 2019-08-09
provider: yet one more set of client registration changes

katbyte:


azurerm/internal/common/client.go
```diff
@@ -15,14 +15,17 @@ import (
...
-	GraphEndpoint               string
-	KeyVaultAuthorizer          autorest.Authorizer
-	ResourceManagerAuthorizer   autorest.Authorizer
-	ResourceManagerEndpoint     string
-	StorageAuthorizer           autorest.Authorizer
-	SubscriptionId              string
-	PartnerId                   string
+	SubscriptionId string
+	TenantID       string
+	PartnerId      string
```
tombuildsstuff: not a blocker for this PR, but we should make these all `Id` or `ID`

azurerm/internal/services/compute/client.go
```diff
@@ -0,0 +1,67 @@
...
+	GalleryImagesClient        compute.GalleryImagesClient
+	GalleryImageVersionsClient compute.GalleryImageVersionsClient
+	ImagesClient               compute.ImagesClient
+	SnapshotsClient            compute.SnapshotsClient
+	UsageClient                compute.UsageClient
+	VMExtensionImageClient     compute.VirtualMachineExtensionImagesClient
+	VMExtensionClient          compute.VirtualMachineExtensionsClient
+	VMScaleSetClient           compute.VirtualMachineScaleSetsClient
+	VMClient                   compute.VirtualMachinesClient
+	VMImageClient              compute.VirtualMachineImagesClient
```
tombuildsstuff: in a subsequent PR (since this affects everything) in retrospect all of the clients here should be pointers; so that we're forced to initialize these (which ensures the authorizer is added, which in turn helps provider developers debug issues)

Issue #4062, opened 2019-08-12
data_source_storage_account_sas doesn't validate dates. Want to PR fix


lawrencegripper:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently the `data_source_storage_account_sas` allows you to provide invalid dates. 

Expected: resource to error if it cannot create a valid SAS token. 
Actual: resource creates invalid sas token and causes failure downstream which are harder to debug. 

I'd like to submit a PR to add a `ValidateFunc` for the data fields. Sound good?

https://github.com/terraform-providers/terraform-provider-azurerm/blob/6f4e1e0ccd80b52ee7fae203d92515b4c5088017/azurerm/data_source_storage_account_sas.go#L101

### New or Affected Resource(s)

- azurerm_data_source_storage_account_sas

### Potential Terraform Configuration

```hcl
data "azurerm_storage_account_sas" "sastoken" {
  connection_string = "${azurerm_storage_account.storage.primary_connection_string}"
  https_only        = true

  resource_types {
    service   = false
    container = false
    object    = true
  }

  services {
    blob  = true
    queue = false
    table = false
    file  = false
  }

  start  = "2019-01-01"
  expiry = "2020-06-31" # This will fail as not US format so doesn't confirm to ISO standard

  permissions {
    read    = true
    write   = false
    delete  = false
    list    = false
    add     = false
    create  = false
    update  = false
    process = false
  }
}
```

r0bnet:
Good idea. Keep in mind that the `start` and `expiry` fields accept any date in ISO-8601 format, so you need to validate that and not simply if only the date part is correct. See: https://docs.microsoft.com/de-de/rest/api/storageservices/create-account-sas#specifying-the-signature-validity-interval
So feel free and go ahead with a PR. I think you don't necessarily need to create an issue for this kind of minor optimisations.

PR #4064, opened 2019-08-12
Add date validator to data_source_storage_account_sas

lawrencegripper:
This PR adds an `ISO8601` validator for use with the `data_source_storage_account_sas` data source. 

There isn't currently a validator for the field so any string can be used and invalid `sas` token are generated (for example a `"2000-06-31"` but there is no 31st of 6th). 

I use [github.com/btubbs/datetime](github.com/btubbs/datetime) to parse the date. 

I've added some tests and run the existing integration tests, all seems good :)

### Why not use existing `RFC3339` validator?

The `ISO8601` standard is more relaxed than `RFC3339` so it would fail on a number of valid inputs. For example `ISO8601` allows `2000-01-01` without a time where as `RGC3339` doesn't. 

### Issue reference

#4062


azurerm/helpers/validate/time_test.go
```diff
@@ -47,6 +47,60 @@ func TestRFC3339Time(t *testing.T) {
...
+			Errors: 0,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.Time, func(t *testing.T) {
+			_, errors := ISO8601DateTime(tc.Time, "test")
+
+			if len(errors) != tc.Errors {
+				t.Fatalf("Expected ISO8601DateTime to have %d not %d errors for %q got date", tc.Errors, len(errors), tc.Time)
```
tombuildsstuff:
```suggestion
				t.Fatalf("Expected ISO8601DateTime to have %d but got %d errors for %q", tc.Errors, len(errors), tc.Time)
```

azurerm/helpers/validate/time.go
```diff
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/Azure/go-autorest/autorest/date"
+	iso8601 "github.com/btubbs/datetime"
```
tombuildsstuff: internal heads up: licence is MIT ✅

lawrencegripper:
@tombuildsstuff no problem, sorry for the slow fix my side. Log message updated. Ready to go! :+1:

PR #4072, opened 2019-08-13
Add Batch Pool Container Registry configuration

ennui93:
Add a `container_registries` block to the batch pool container configuration resource.

This setting is necessary for Azure Batch service Pools to support custom container images (i.e. from private container repositories).


azurerm/data_source_batch_pool.go
```diff
@@ -119,6 +119,16 @@ func dataSourceArmBatchPool() *schema.Resource {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
+						"container_registries": {
+							Type:     schema.TypeSet,
+							Computed: true,
+							Elem: &schema.Schema{
+								Type: schema.TypeMap,
```
katbyte: Instead of a map could we actually write out the schema?

ennui93:
Hi @katbyte, thank you for your review. I have revamped the PR with the schema as requested. Please review and provide any pertinent feedback at your convenience.

website/docs/r/batch_pool.html.markdown
```diff
@@ -247,6 +256,18 @@ A `resource_file` block supports the following:
 
 ~> **Please Note:** Exactly one of `auto_storage_container_name`, `storage_container_url` and `auto_user` must be specified.
 
+---
+
+A `container_registries` block supports the following:
+
+* `registry_server` - (Optional) The container registry URL. The default is "docker.io".
```
katbyte: Could we add that this is a force new?
```suggestion
* `registry_server` - (Optional) The container registry URL. The default is "docker.io". Changing this forces a new resource to be created.
```

website/docs/r/batch_pool.html.markdown
```diff
@@ -247,6 +256,18 @@ A `resource_file` block supports the following:
...
+
+* `user_name` - (Optional) The user name to log into the registry server.
```
katbyte: And here

website/docs/r/batch_pool.html.markdown
```diff
@@ -247,6 +256,18 @@ A `resource_file` block supports the following:
...
+
+* `password` - (Optional) The password to log into the registry server.
```
katbyte: And here

website/docs/r/batch_pool.html.markdown
```diff
@@ -77,6 +77,13 @@ EOF
 
   container_configuration {
     type = "DockerCompatible"
+    container_registries = [
+      {
+        registry_server = "docker.io"
```
katbyte: Could we align these assignments?

azurerm/resource_arm_batch_pool.go
```diff
@@ -124,6 +126,31 @@ func resourceArmBatchPool() *schema.Resource {
...
+									"registry_server": {
+										Type:     schema.TypeString,
+										Required: true,
+										ForceNew: true,
+									},
+									"user_name": {
+										Type:     schema.TypeString,
+										Required: true,
+										ForceNew: true,
+									},
```
katbyte: Could we validate this with at the very least `NoEmptyStrings`

azurerm/resource_arm_batch_pool.go
```diff
@@ -124,6 +126,31 @@ func resourceArmBatchPool() *schema.Resource {
...
+							Optional:   true,
+							ForceNew:   true,
+							ConfigMode: schema.SchemaConfigModeAttr,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"registry_server": {
+										Type:     schema.TypeString,
+										Required: true,
+										ForceNew: true,
+									},
```
katbyte: Could we validate this with at the very least `NoEmptyStrings`

azurerm/resource_arm_batch_pool.go
```diff
@@ -124,6 +126,31 @@ func resourceArmBatchPool() *schema.Resource {
...
+									"user_name": {
+										Type:     schema.TypeString,
+										Required: true,
+										ForceNew: true,
+									},
+									"password": {
+										Type:      schema.TypeString,
+										Required:  true,
+										Sensitive: true,
+									},
```
katbyte: Could we validate this with at the very least `NoEmptyStrings`

azurerm/resource_arm_batch_pool_test.go
```diff
@@ -311,7 +311,11 @@ func TestAccAzureRMBatchPool_container(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testCheckAzureRMBatchPoolExists(resourceName),
 					resource.TestCheckResourceAttr(resourceName, "container_configuration.0.type", "DockerCompatible"),
+					resource.TestCheckResourceAttr(resourceName, "container_configuration.0.container_registries.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "container_configuration.0.container_registries.0.registry_server", "myContainerRegistry.azurecr.io"),
+					resource.TestCheckResourceAttr(resourceName, "container_configuration.0.container_registries.0.user_name", "myUserName"),
 				),
+				ExpectNonEmptyPlan: true,
```
katbyte: We shouldn't need this.. i presume it is because we are reading in the password each time.

website/docs/r/batch_pool.html.markdown
```diff
@@ -247,6 +256,18 @@ A `resource_file` block supports the following:
...
+
+A `container_registries` block supports the following:
+
+* `registry_server` - (Optional) The container registry URL. The default is "docker.io".
+
+* `user_name` - (Optional) The user name to log into the registry server.
+
+* `password` - (Optional) The password to log into the registry server.
+
+~> **NOTE:** Changes to the password field will not trigger a recreation of the pool, which is necessary to change container registry settings. In order to change the password for a container registry, either taint the Batch Pool and trigger an update, or manually destroy and re-create the pool.
```
katbyte: If we don't read in the password each time updates to password should be reflected?

azurerm/helpers/azure/batch_pool.go
```diff
@@ -203,10 +204,44 @@ func FlattenBatchPoolContainerConfiguration(armContainerConfiguration *batch.Con
...
+	if armContainerRegistry == nil {
+		return result
+	}
+	if registryServer := armContainerRegistry.RegistryServer; registryServer != nil {
+		result["registry_server"] = *registryServer
+	}
+	if userName := armContainerRegistry.UserName; userName != nil {
+		result["user_name"] = *userName
+	}
+	// We won't usually get the password back, but we include it here anyways
```
katbyte: If this is usually not returned we should not be setting it. What you can do is re use the password from state that you could get via `d.Get()`

ennui93:
Hi @katbyte, thanks again for your second review. I believe I have addressed all of your comments.

You are correct that I was unaware how to pass the password supplied by the user into the state. The solution looks much better now.

Please review at your convenience.

PR #4076, opened 2019-08-13
provider: add descriptions to provider schema

katbyte:


azurerm/provider.go
```diff
@@ -421,83 +421,96 @@ func Provider() terraform.ResourceProvider {
...
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("ARM_SKIP_CREDENTIALS_VALIDATION", false),
+				Description: "This will cause the AzureRM Provider to skip verifying the credentials being used are valid.",
 			},
 
 			"skip_provider_registration": {
 				Type:        schema.TypeBool,
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("ARM_SKIP_PROVIDER_REGISTRATION", false),
+				Description: "This will cause the the AzureRM Provider to skip registering any required Resource Providers that arn't already registered",
```
tombuildsstuff:
```suggestion
				Description: "Should the AzureRM Provider skip registering all of the Resource Providers that it supports, if they're not already registered?",
```

website/docs/index.html.markdown
```diff
@@ -88,7 +88,7 @@ The following arguments are supported:
 
 * `client_id` - (Optional) The Client ID which should be used. This can also be sourced from the `ARM_CLIENT_ID` Environment Variable.
 
-* `environment` - (Optional) The Cloud Environment which should be used. Possible values are `public`, `usgovernment`, `german` and `china`. Defaults to `public`. This can also be sourced from the `ARM_ENVIRONMENT` environment variable.
+* `environment` - (Optional) The Cloud Environment which should be used. Possible values are `public`, `usgovernment`, `german`, and `china`. Defaults to `public`. This can also be sourced from the `ARM_ENVIRONMENT` environment variable.
```
tombuildsstuff: I don't think we need this comma here?

```suggestion
* `environment` - (Optional) The Cloud Environment which should be used. Possible values are `public`, `usgovernment`, `german` and `china`. Defaults to `public`. This can also be sourced from the `ARM_ENVIRONMENT` environment variable.
```

katbyte: Its an oxford comma: https://www.grammarly.com/blog/what-is-the-oxford-comma-and-why-do-people-care-so-much-about-it/

Its not entirely necessary here because the backticks make it clear its not `german and china` but I've gotten into the habit of using them everywhere and I think it makes sense to leave in?

tombuildsstuff: this was mostly for consistency with the other docs pages, which use this format - but I'm fine either way 👍

Issue #4085, opened 2019-08-14
azurerm_notification_hub_authorization_rule attempt to create multiple rules in one go fails


andrey-dubnik:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.12.6
+ provider.azurerm v1.32.1

### Affected Resource(s)

azurerm_notification_hub_authorization_rule

### Terraform Configuration Files

```hcl
resource "azurerm_notification_hub_authorization_rule" "hosting" {
  for_each = {
      authrule_1 = {
                  listen = true
                  },
      authrule_2 = {
                  send = true
                  },
      authrule_3 = {
                  manage = true
                  }
  }
  notification_hub_name = var.notificationhub_name
  namespace_name      = var.notificationhub_namespace_name
  name                = each.key

  resource_group_name = var.resource_group_name

  listen    = (lookup(each.value, "manage", null) != null ? true : lookup(each.value, "listen", null))
  send      = (lookup(each.value, "manage", null) != null ? true : lookup(each.value, "send", null))
  manage    = lookup(each.value, "manage", null)
}
```

### Debug Output

```sh
# module.notificationhubnamespace.azurerm_notification_hub_authorization_rule.hosting["authrule_1"]: Creating... |  
-- | --
  | # module.notificationhubnamespace.azurerm_notification_hub_authorization_rule.hosting["authrule_2"]: Creating... |  
  | # module.notificationhubnamespace.azurerm_notification_hub_authorization_rule.hosting["authrule_3"]: Creation complete after 6s [id=/subscriptions/***/resourceGroups/***/providers/Microsoft.NotificationHubs/namespaces/nhnazewtmlit001tftest/notificationHubs/nhazewtmlit001tftest/AuthorizationRules/authrule_3] |  
  | # |  
[...]

PR #4087, opened 2019-08-14
azurerm_notification_hub_authorization_rule fix to be able to create multiple rules in one go

adubnik:
Fixes the issue #4085

### DIFF
--- a/azurerm/resource_arm_notification_hub_authorization_rule.go
+++ b/azurerm/resource_arm_notification_hub_authorization_rule.go
@@ -5,6 +5,7 @@ import (
 	"log"
 
 	"github.com/Azure/azure-sdk-for-go/services/notificationhubs/mgmt/2017-04-01/notificationhubs"
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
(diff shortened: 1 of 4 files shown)


azurerm/resource_arm_notification_hub_authorization_rule.go
```diff
@@ -106,21 +107,24 @@ func resourceArmNotificationHubAuthorizationRuleCreateUpdate(d *schema.ResourceD
 		},
 	}
 
-	if _, err := client.CreateOrUpdateAuthorizationRule(ctx, resourceGroup, namespaceName, notificationHubName, name, parameters); err != nil {
-		return fmt.Errorf("Error creating Authorization Rule %q (Notification Hub %q / Namespace %q / Resource Group %q): %+v", name, notificationHubName, namespaceName, resourceGroup, err)
-	}
+	return resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
```
tombuildsstuff: since we don't support custom timeouts at this point (it's coming to every resource in 2.0, so the provider behaves consistently) - as such could we hard code this for the moment? e.g.

```suggestion
	return resource.Retry(5 * time.Minute, func() *resource.RetryError {
```

adubnik: Updated with Locks instead, looks cleaner and no retries are needed.

I was thinking about delete but opted not to add more code for the non existing problem as delete does not seem to be affected by this issue as cleanup is always consistent... Anyway updated Delete with the locks as well so should be good.

BTW comment on the azureRMLockByName says deprecated, is there something else to use in future?

azurerm/resource_arm_notification_hub_authorization_rule_test.go
```diff
@@ -134,6 +134,61 @@ func TestAccAzureRMNotificationHubAuthorizationRule_send(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMNotificationHubAuthorizationRule_multi(t *testing.T) {
+	resourceOneName := "azurerm_notification_hub_authorization_rule.test1"
+	resourceTwoName := "azurerm_notification_hub_authorization_rule.test2"
+	resourceThreeName := "azurerm_notification_hub_authorization_rule.test2"
```
tombuildsstuff: typo here:

```suggestion
	resourceThreeName := "azurerm_notification_hub_authorization_rule.test3"
```

tombuildsstuff:
@adubnik thanks for pushing those changes

> BTW comment on the azureRMLockByName says deprecated, is there something else to use in future?

we're gradually moving those functions (everything in `deprecated.go`) into `azurerm/internal/...` - but we don't want to break all the open PR's; as such those exist as a migration point. For the moment it should be fine to use that, when we remove those functions in the near future we'll switch over to the new method, so that should be fine 👍 

Thanks!

tombuildsstuff:
Tests pass:

Screenshot 2019-08-14 at 18 51 50

PR #4089, opened 2019-08-14
New resource: `azurerm_virtual_wan` #3505

katbyte:
new PR for #3505 so i can resolve merge conflicts


azurerm/resource_arm_virtual_wan_test.go
```diff
@@ -0,0 +1,216 @@
...
+func testAccAzureRMVirtualWan_requiresImport(rInt int, location string) string {
+	return fmt.Sprintf(`
+%s
+
+resource "azurerm_virtual_wan" "import" {
+  name                = "${azurerm_virtual_wan.test.name}"
+  resource_group_name = "${azurerm_virtual_wan.test.resource_group_name}"
+  location            = "${azurerm_virtual_wan.test.location}"
+}
+`, testAccAzureRMVirtualWan_basic(rInt, location))
```
tombuildsstuff: for consistency with the other resources (so we can regex match these in the future) is it worth matching the format of the others:

```suggestion
template := testAccAzureRMVirtualWan_basic(rInt, location)
return fmt.Sprintf(`
...
`, template)
```

Issue #3765, opened 2019-07-01
Support for 'UltraSSDEnabled' flag for a VM


normseq:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

In order to be able to map UltraSSD disks to a VM , the VM needs to be created with the ultra-ssd-enabled flag in CLI

az vm create --subscription $subscription -n $vmname -g $rgname --image Win2016Datacenter **--ultra-ssd-enabled** --zone $zone --authentication-type password --admin-password $password --admin-username $user --attach-data-disks $diskname --size Standard_D4s_v3 --location $location

Without this property set, mapping an UltraSSD disk to a VM results in the following error
_"Error: Error updating Virtual Machine "ultravm01" (Resource Group "ultraRG") with Disk "ultravm01-disk1": compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="StorageAccountType UltraSSD_LRS can be used only when additionalCapabilities.ultraSSDEnabled is set." Target="managedDisk.storageAccountType"_

### New or Affected Resource(s)

- azurerm_virtual_machine

### Potential Terraform Configuration

```hcl
resource "azurerm_virtual_machine" "ultravm" {
    name                  = "ultravm01"
    resource_group_name   = "${azurerm_resource_group.ultragroup.name}"
    network_interface_ids = ["${azurerm_network_interface.ultravmnic.id}"]
    additional_capabilities {
        ultra_ssd_enabled = true
    }

```

### References
**CLI**
https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disks-enable-ultra-ssd
**Property details**
https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.management.compute.models.additionalcapabilities.ultrassdenabled?view=azure-dotnet

* #0000

venkey1000:
Hi, 
I am working on this.

PR #4102, opened 2019-08-15
`azurerm_managed_disk` - add support for the UltraSSD `disk_iops_read_write` & `disk_mbps_read_write` properties

jeffreyCline:
[fixes: #3765 ]


azurerm/data_source_managed_disk.go
```diff
@@ -82,6 +92,12 @@ func dataSourceArmManagedDiskRead(d *schema.ResourceData, meta interface{}) erro
 		if diskSize := props.DiskSizeGB; diskSize != nil {
 			d.Set("disk_size_gb", *diskSize)
 		}
+		if diskIOPS := props.DiskIOPSReadWrite; diskIOPS != nil {
+			d.Set("disk_iops_read_write", *diskIOPS)
+		}
```
tombuildsstuff: this can just be:

```suggestion
d.Set("disk_iops_read_write", props.DiskIOPSReadWrite)
```

azurerm/data_source_managed_disk.go
```diff
@@ -82,6 +92,12 @@ func dataSourceArmManagedDiskRead(d *schema.ResourceData, meta interface{}) erro
...
+		if diskMBps := props.DiskMBpsReadWrite; diskMBps != nil {
+			d.Set("disk_mbps_read_write", *diskMBps)
+		}
```
tombuildsstuff: this can just be:

```
d.Set("disk_mbps_read_write", props.DiskMBpsReadWrite)
```

azurerm/resource_arm_managed_disk.go
```diff
@@ -267,6 +297,12 @@ func resourceArmManagedDiskRead(d *schema.ResourceData, meta interface{}) error
 		if osType := props.OsType; osType != "" {
 			d.Set("os_type", string(osType))
 		}
+		if diskIOPS := props.DiskIOPSReadWrite; diskIOPS != nil {
+			d.Set("disk_iops_read_write", *diskIOPS)
+		}
+		if diskMBps := props.DiskMBpsReadWrite; diskMBps != nil {
+			d.Set("disk_mbps_read_write", *diskMBps)
+		}
```
tombuildsstuff: (same here)

azurerm/data_source_managed_disk.go
jeffreyCline: Fixed.

azurerm/data_source_managed_disk.go
jeffreyCline: Fixed.

azurerm/resource_arm_managed_disk.go
jeffreyCline: Fixed.

jeffreyCline:
> Thanks for the PR @jeffreyCline,
> In addition to my inline comments could we get a test with the ultraSSD type and the new properties?

I didn't add the test as this is a private preview and you need to set up your subscription to use this in one of the three regions it is currently in. I believe that it is supposed to GA on Oct. 1st, so I would be happy to add one then.

azurerm/data_source_managed_disk.go
```diff
@@ -81,7 +91,13 @@ func dataSourceArmManagedDiskRead(d *schema.ResourceData, meta interface{}) erro
 
 	if props := resp.DiskProperties; props != nil {
 		if diskSize := props.DiskSizeGB; diskSize != nil {
-			d.Set("disk_size_gb", *diskSize)
+			d.Set("disk_size_gb", props.DiskSizeGB)
+		}
+		if diskIOPS := props.DiskIOPSReadWrite; diskIOPS != nil {
```
katbyte: We don't need the check here as d.Set() will do it for us:
```suggestion
					d.Set("disk_iops_read_write", props.DiskIOPSReadWrite)
```

azurerm/data_source_managed_disk.go
```diff
@@ -81,7 +91,13 @@ func dataSourceArmManagedDiskRead(d *schema.ResourceData, meta interface{}) erro
...
+			d.Set("disk_iops_read_write", props.DiskIOPSReadWrite)
+		}
+		if diskMBps := props.DiskMBpsReadWrite; diskMBps != nil {
```
katbyte: Same here

azurerm/resource_arm_managed_disk.go
```diff
@@ -268,6 +298,12 @@ func resourceArmManagedDiskRead(d *schema.ResourceData, meta interface{}) error
 		if osType := props.OsType; osType != "" {
 			d.Set("os_type", string(osType))
 		}
+		if diskIOPS := props.DiskIOPSReadWrite; diskIOPS != nil {
```
katbyte: We don't need the check here as `d.Set` will do it for us:
```suggestion
		d.Set("disk_iops_read_write", props.DiskIOPSReadWrite)
```

azurerm/resource_arm_managed_disk.go
```diff
@@ -268,6 +298,12 @@ func resourceArmManagedDiskRead(d *schema.ResourceData, meta interface{}) error
...
+			d.Set("disk_iops_read_write", props.DiskIOPSReadWrite)
+		}
+		if diskMBps := props.DiskMBpsReadWrite; diskMBps != nil {
```
katbyte: Same here

website/docs/d/managed_disk.html.markdown
```diff
@@ -110,5 +110,7 @@ resource "azurerm_virtual_machine" "test" {
 * `source_resource_id` - ID of an existing managed disk that the current resource was created from.
 * `os_type` - The operating system for managed disk. Valid values are `Linux` or `Windows`
 * `disk_size_gb` - The size of the managed disk in gigabytes.
+* `disk_iops_read_write` - The number of IOPS allowed for this disk. One operation can transfer between 4k and 256k bytes.  Only applicable to disks of type `UltraSSD`.
```
katbyte:
```suggestion
* `disk_iops_read_write` - The number of IOPS allowed for this disk. One operation can transfer between 4k and 256k bytes.
```

website/docs/d/managed_disk.html.markdown
```diff
@@ -110,5 +110,7 @@ resource "azurerm_virtual_machine" "test" {
...
+* `disk_mbps_read_write` - The bandwidth allowed for this disk. Only applicable to disks of type `UltraSSD`.
```
katbyte:
```suggestion
* `disk_mbps_read_write` - The bandwidth allowed for this disk. 
```

website/docs/r/managed_disk.html.markdown
```diff
@@ -104,13 +104,19 @@ The following arguments are supported:
...
 
+* `disk_iops_read_write` - (Optional) The number of IOPS allowed for this disk; only settable for UltraSSD disks. One operation can transfer between 4k and 256k bytes.
+
+* `disk_mbps_read_write` - (Optional) The bandwidth allowed for this disk; only settable for UltraSSD disks. MBps means millions of bytes per second.
+
+-> **Note**: A `storage_account_type` of type `UltraSSD_LRS` and the arguments `disk_iops_read_write` and `disk_mbps_read_write` are currently in private preview and are not available to subscriptions that have not requested onboarding to `Azure Ultra Disk Storage` private preview. `Azure Ultra Disk Storage` is only available in `East US 2`, `North Europe`, and `Southeast Asia` regions. For more information see the `Azure Ultra Disk Storage` [product documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/disks-enable-ultra-ssd), [product blog](https://azure.microsoft.com/en-us/blog/announcing-the-general-availability-of-azure-ultra-disk-storage/) and [FAQ](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/faq-for-disks#ultra-disks).
```
katbyte: Given this applies to the "type" could we move it up to under the type??

website/docs/r/managed_disk.html.markdown
```diff
@@ -104,13 +104,19 @@ The following arguments are supported:
...
+
 * `encryption_settings` - (Optional) an `encryption_settings` block as defined below.
 
 * `tags` - (Optional) A mapping of tags to assign to the resource.
 
 * `zones` - (Optional) A collection containing the availability zone to allocate the Managed Disk in.
 
--> **Please Note**: Availability Zones are [only supported in several regions at this time](https://docs.microsoft.com/en-us/azure/availability-zones/az-overview).
+-> **Note**: Availability Zones are [only supported in several regions at this time](https://docs.microsoft.com/en-us/azure/availability-zones/az-overview).
```
katbyte:
```suggestion
-> **Note**: Availability Zones are [only supported in select regions at this time](https://docs.microsoft.com/en-us/azure/availability-zones/az-overview).
```

website/docs/d/managed_disk.html.markdown
mybayern1974: @katbyte , may I suggest to keep the "_Only applicable to disks of type UltraSSD_" since it is a matter of fact. Setting this value against other disk types won't take effect so we would like to explicitly have users be aware of this.

website/docs/d/managed_disk.html.markdown
mybayern1974: @katbyte , may I suggest to keep the "_Only applicable to disks of type UltraSSD_" since it is a matter of fact. Setting this value against other disk types won't take effect so we would like to explicitly have users be aware of this.

katbyte: This is a datasource so users cannot set these values.

website/docs/r/managed_disk.html.markdown
jeffreyCline: kk... moved. :)

azurerm/data_source_managed_disk.go
jeffreyCline: Fixed.

azurerm/data_source_managed_disk.go
jeffreyCline: Fixed.

azurerm/resource_arm_managed_disk.go
jeffreyCline: Fixed.

azurerm/resource_arm_managed_disk.go
jeffreyCline: Fixed.

jeffreyCline:
@kt
> Even thou it is in preview i'd like to see the tests for these properties. I believe our subscription is now whitelisted so we should be able to run them.

I have added the tests to both data source and resource.

PR #4103, opened 2019-08-16
Replace usages of deprecated validation, suppress and lock functions

katbyte:


katbyte:
I do wonder if `azure.ParseAzureResourceID` should be `azure.ParseResourceID` 🤔

tombuildsstuff:
> I do wonder if azure.ParseAzureResourceID should be azure.ParseResourceID 🤔

Possibly for readability - tbh I think we'd be better to try and move off these altogether by sending a PR adding ID Parsing functions to the Go SDK Generator, which can then generate a struct [as we do in Giovanni](https://github.com/tombuildsstuff/giovanni/blob/master/storage/2018-11-09/file/files/resource_id.go#L11-L63) (although this is for Resource ID's not Data Plane API's, so the logic would be considerably simpler) - which would enable us to remove these magic strings altogether

PR #4110, opened 2019-08-17
[WIP]New Resource: Web Application Firewall Policies

jeffreyCline:
Adding WAF support for Azure


jeffreyCline:
This is not the correct API for FrontDoor. The correct WAF is embedded within the FrontDoor API.

PR #4128, opened 2019-08-21
implement azurerm_storage_filesystem

tcz001:


tombuildsstuff:
Looks like this is blocked on this bug: https://github.com/Azure/azure-sdk-for-go/pull/5572 which will need to be fixed in the Swagger?

PR #4137, opened 2019-08-22
azurerm_subnet - Fixed delegation (fixes #3730)

### LINKED ISSUE #3730: subnet delegation actions handling is incorrect.
### Community Note

```bash
$ terraform version
Terraform v0.12.3
+ provider.azurerm v1.30.1
```

### Affected Resource(s)

* `azurerm_subnet`

```hcl
resource "azurerm_subnet" "test" {
  name                 = "test"
  resource_group_name  = "test-rg"
  virtual_network_name = "test-vnet"
  address_prefix       = "10.1.0.0/24"

  delegation {
    name = md5("Microsoft.Sql/managedInstances")
    service_delegation {
      name = "Microsoft.Sql/managedInstances"
      actions = [
        "Microsoft.Network/virtualNetworks/subnets/action" 
        # "Microsoft.Network/virtualNetworks/subnets/join/action"
        # "Microsoft.Network/virtualNetworks/subnets/prepareNetworkPolicies/action"
      ]
    }
  }
}
```

### Expected Behavior

Provider should accept all possible delegation action or if Azure API disallows delegation action modification (which, looking at the CLI it does not), then provider should not try to set or modify them.

### Actual Behavior

I'm fairly new to Azure API, but behaviour of the azurerm_subnet resource seems incorrect:

Provider allows me to create subnet delegation with empty actions list or with a single value in it (the one that's not commented out). Once subnet is created though, it looks like Azure API disregards actions list and create delegation with 3 options (full list above):
```
  # azurerm_subnet.test will be updated in-place
  ~ resource "azurerm_subnet" "test" {
        address_prefix       = "10.1.0.0/24"
        id                   = "..."
        ip_configurations    = []
        name                 = "test"
        resource_group_name  = "test-rg"
        service_endpoints    = []
        virtual_network_name = "test-vnet"

      ~ delegation {
            name = "5807eb7c8e7ae01071afecff5e0a7466"

          ~ service_delegation {
              ~ actions = [
                  - "Microsoft.Network/virtualNetworks/subnets/join/action",
                  - "Microsoft.Network/virtualNetworks/subnets/prepareNetworkPolicies/action",
                  + "Microsoft.Network/virtualNetworks/subnets/action",
                ]
                name    = "Microsoft.Sql/managedInstances"
            }
        }
    }
```

As a result terraform wants to change subnet delegation on every apply. And it is not possible to set all the actions because provider accepts only one: 

```
Error: expected delegation.0.service_delegation.0.actions.1 to be one of [Microsoft.Network/virtualNetworks/subnets/action], got Microsoft.Network/virtualNetworks/subnets/join/action
```

### Steps to Reproduce

1. create the plan from snippet above
2. `terraform apply`
3. `terrafrom plan`

bentterp:
This should fix #3730 

I've removed and deployed delegation, and removed again. When deployed, the delegation shows up in the console without complaints.

After applying a change, the next plan is empty.

### DIFF
--- a/azurerm/resource_arm_subnet.go
+++ b/azurerm/resource_arm_subnet.go
@@ -111,6 +111,7 @@ func resourceArmSubnet() *schema.Resource {
 										Elem: &schema.Schema{
 											Type: schema.TypeString,
 											ValidateFunc: validation.StringInSlice([]string{
+												"Microsoft.Network/virtualNetworks/subnets/action",
 												"Microsoft.Network/virtualNetworks/subnets/join/action",
 												"Microsoft.Network/virtualNetworks/subnets/prepareNetworkPolicies/action",
 											}, false),
--- a/azurerm/resource_arm_subnet.go
+++ b/azurerm/resource_arm_subnet.go
@@ -111,7 +111,8 @@ func resourceArmSubnet() *schema.Resource {
 										Elem: &schema.Schema{
 											Type: schema.TypeString,
 											ValidateFunc: validation.StringInSlice([]string{
-												"Microsoft.Network/virtualNetworks/subnets/action",
+												"Microsoft.Network/virtualNetworks/subnets/join/action",
+												"Microsoft.Network/virtualNetworks/subnets/prepareNetworkPolicies/action",
 											}, false),
 										},
 									},
--- a/website/docs/r/subnet.html.markdown
+++ b/website/docs/r/subnet.html.markdown
@@ -74,7 +74,9 @@ The following arguments are supported:
 ---
 
 A `delegation` block supports the following:
+
 * `name` (Required) A name for this delegation.
+
 * `service_delegation` (Required) A `service_delegation` block as defined below.
 
 ---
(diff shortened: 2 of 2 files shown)


azurerm/resource_arm_subnet.go
```diff
@@ -111,7 +111,8 @@ func resourceArmSubnet() *schema.Resource {
 										Elem: &schema.Schema{
 											Type: schema.TypeString,
 											ValidateFunc: validation.StringInSlice([]string{
-												"Microsoft.Network/virtualNetworks/subnets/action",
```
tombuildsstuff: is this value no longer valid?

website/docs/r/subnet.html.markdown
```diff
@@ -85,7 +85,7 @@ A `service_delegation` block supports the following:
 
 * `name` - (Required) The name of service to delegate to. Possible values include: `Microsoft.BareMetal/AzureVMware`, 
 `Microsoft.BareMetal/CrayServers`, `Microsoft.Batch/batchAccounts`, `Microsoft.ContainerInstance/containerGroups`, `Microsoft.Databricks/workspaces`, `Microsoft.HardwareSecurityModules/dedicatedHSMs`, `Microsoft.Logic/integrationServiceEnvironments`, `Microsoft.Netapp/volumes`, `Microsoft.ServiceFabricMesh/networks`, `Microsoft.Sql/managedInstances`, `Microsoft.Sql/servers`, `Microsoft.Web/hostingEnvironments` or `Microsoft.Web/serverFarms`.
-* `actions` - (Optional) A list of Actions which should be delegated. Possible values include: `Microsoft.Network/virtualNetworks/subnets/action`.
+* `actions` - (Required) A list of Actions which should be delegated. Required values are: `Microsoft.Network/virtualNetworks/subnets/join/action ,  Microsoft.Network/virtualNetworks/subnets/prepareNetworkPolicies/action`.
```
tombuildsstuff: AFAIK it's possible for users to pick and choose these values, rather than requiring all of them to be set?

```suggestion
* `actions` - (Required) A list of Actions which should be delegated. Possible values are: `Microsoft.Network/virtualNetworks/subnets/join/action ,  Microsoft.Network/virtualNetworks/subnets/prepareNetworkPolicies/action`.
```

bentterp: The only thing I have to compare with is that if the delegation is done manually in the webportal then you get this result. I don't know if there are other possible actions, but these two are required (at least this week....) for the delegation to work.

azurerm/resource_arm_subnet.go
tombuildsstuff: this still appears to be a valid value: https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/Microsoft.Network/stable/2018-08-01/examples/VirtualNetworkGetWithServiceAssociationLink.json#L56 - as such can we add this back in?

website/docs/r/subnet.html.markdown
tombuildsstuff: since it's possible for users to pick from one of these values ([including the original one](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/Microsoft.Network/stable/2018-08-01/examples/VirtualNetworkGetWithServiceAssociationLink.json#L56)) could we add this old value back in/update the language here?

tombuildsstuff:
Tests pass:

Screenshot 2019-08-29 at 14 58 15

bentterp:
Yes, you can add all 3 actions like this
```
      name = "subnet service delegation"
      service_delegation {
        name     = delegation.value
        actions  = [
          "Microsoft.Network/virtualNetworks/subnets/join/action",
          "Microsoft.Network/virtualNetworks/subnets/prepareNetworkPolicies/action",
          "Microsoft.Network/virtualNetworks/subnets/action"
        ]
      }
```
and the plan will be sane and the apply will be successfull.

But every subsequent apply will want to do this
```
      ~ delegation {
            name = "subnet service delegation"

          ~ service_delegation {
              ~ actions = [
                    "Microsoft.Network/virtualNetworks/subnets/join/action",
                    "Microsoft.Network/virtualNetworks/subnets/prepareNetworkPolicies/action",
                  + "Microsoft.Network/virtualNetworks/subnets/action",
                ]
                name    = "Microsoft.Sql/managedInstances"
            }
        }
```
And that apply will also be successfull, and then the next time you get the same diff between TF state and what's deployed and so it goes on and on.

This is not a problem for an ARM template because it doesn't care about state - I care about state, that's why I use Terraform 😍 

However, if you leave out "Microsoft.Network/virtualNetworks/subnets/action" then state will match subsequent interrogations, and delegation will still work.

I'm perfectly okay with what you chose to do, but it will create confusion if we don't at least make it clear in the docs what is happening and how to get rid of it.

//Bent

PR #4140, opened 2019-08-22
r/api_management_api_policy: sending `policy` as Raw XML

### LINKED ISSUE #3918: Creating a API Management policy for an API gives an ValidationError
### Community Note

```
$ terraform -v
Terraform v0.12.5
+ provider.azurerm v1.31.0
```

### Affected Resource(s)

* `azurerm_api_management_api_policy`

```hcl

# Create default policy for example API
resource "azurerm_api_management_api_policy" "apim-api-example-policy" {

  resource_group_name = "${var.customer}-rg"
  api_management_name = azurerm_api_management.apim.name
  api_name            = azurerm_api_management_api.apim-api-example.name

# Set the policy here
  xml_content = file("${path.module}/default-api-policy.xml")

}
```

The XML file referenced is available in this gist: https://gist.github.com/tverhoeven/abf1054ff868ee9f602b4308fe606449

### Debug Output

https://gist.github.com/tverhoeven/abf1054ff868ee9f602b4308fe606449

### Expected Behavior

The policy is created without throwing an error.

### Actual Behavior

Terraform report a error, status 400, ValidationError coming from the AzureRM:

`Error: Error creating or updating API Policy (Resource Group "poc-test-rg" / API Management Service "poc-test-apim" / API "poc-test-api-example"): apimanagement.APIPolicyClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="ValidationError" Message="One or more fields contain incorrect values:" Details=[{"code":"ValidationError","message":"'\u003c', hexadecimal value 0x3C, is an invalid attribute character. Line 10, position 63.","target":"representation"}]`

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

I've tried this with both inline XML in the terraform, or with external files containing the XML. I've also tried with the external file formatted as ASCII or UTF-8 (checked with the file command). I always get the same error.

To me it looks like there is an UTF-8 encoding to many happening. See this snippet from the debug output: "\u003cinbound\u003e\n".

I also get this both with Terraform running on macOS and Ubuntu Linux. So it is not OS specific.

### References

* #0000

riordanp:
Should fix #3918

I have updated the existing test case for policy links and string policies. I will try and figure out how to run the full test suite tomorrow. The issue might affect the other policy resources as well so I'll check that at the same time.

### DIFF
--- a/azurerm/resource_arm_api_management_api_policy.go
+++ b/azurerm/resource_arm_api_management_api_policy.go
@@ -92,7 +92,7 @@ func resourceArmApiManagementAPIPolicyCreateUpdate(d *schema.ResourceData, meta 
 
 	if xmlLink != "" {
 		parameters.PolicyContractProperties = &apimanagement.PolicyContractProperties{
-			ContentFormat: apimanagement.Rawxml,
+			ContentFormat: apimanagement.RawxmlLink,
 			PolicyContent: utils.String(xmlLink),
 		}
 	}
--- a/azurerm/resource_arm_api_management_api_policy.go
+++ b/azurerm/resource_arm_api_management_api_policy.go
@@ -83,7 +83,14 @@ func resourceArmApiManagementAPIPolicyCreateUpdate(d *schema.ResourceData, meta 
 	xmlContent := d.Get("xml_content").(string)
 	xmlLink := d.Get("xml_link").(string)
 
-	if xmlContent != "" {
+	if xmlLink != "" {
+		parameters.PolicyContractProperties = &apimanagement.PolicyContractProperties{
+			ContentFormat: apimanagement.RawxmlLink,
+			PolicyContent: utils.String(xmlLink),
+		}
+	} else if xmlContent != "" {
+		// this is intentionally an else-if since `xml_content` is computed
+
 		// clear out any existing value for xml_link
 		if !d.IsNewResource() {
 			d.Set("xml_link", "")
@@ -92,13 +99,6 @@ func resourceArmApiManagementAPIPolicyCreateUpdate(d *schema.ResourceData, meta 
 		parameters.PolicyContractProperties = &apimanagement.PolicyContractProperties{
 			ContentFormat: apimanagement.Rawxml,
 			PolicyContent: utils.String(xmlContent),
-		}
-	}
-
-	if xmlLink != "" {
-		parameters.PolicyContractProperties = &apimanagement.PolicyContractProperties{
-			ContentFormat: apimanagement.RawxmlLink,
-			PolicyContent: utils.String(xmlLink),
 		}
 	}
 
--- a/azurerm/resource_arm_api_management_api_policy.go
+++ b/azurerm/resource_arm_api_management_api_policy.go
@@ -2,6 +2,7 @@ package azurerm
 
 import (
 	"fmt"
+	"html"
 	"log"
 	"time"
 
@@ -151,9 +152,14 @@ func resourceArmApiManagementAPIPolicyRead(d *schema.ResourceData, meta interfac
 	d.Set("api_name", apiName)
 
 	if properties := resp.PolicyContractProperties; properties != nil {
+		policyContent := ""
+		if pc := properties.PolicyContent; pc != nil {
+			policyContent = html.UnescapeString(*pc)
+		}
+
 		// when you submit an `xml_link` to the API, the API downloads this link and stores it as `xml_content`
 		// as such there is no way to set `xml_link` and we'll let Terraform handle it
-		d.Set("xml_content", properties.PolicyContent)
+		d.Set("xml_content", policyContent)
 	}
 
 	return nil
--- a/azurerm/resource_arm_api_management_api_policy.go
+++ b/azurerm/resource_arm_api_management_api_policy.go
@@ -8,9 +8,9 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/apimanagement/mgmt/2018-01-01/apimanagement"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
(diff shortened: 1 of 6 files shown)


azurerm/resource_arm_api_management_api_policy.go
```diff
@@ -74,14 +74,14 @@ func resourceArmApiManagementAPIPolicyCreateUpdate(d *schema.ResourceData, meta
...
-			ContentFormat: apimanagement.XML,
+			ContentFormat: apimanagement.Rawxml,
 			PolicyContent: utils.String(xmlContent),
 		}
 	}
 
 	if xmlLink != "" {
 		parameters.PolicyContractProperties = &apimanagement.PolicyContractProperties{
-			ContentFormat: apimanagement.XMLLink,
+			ContentFormat: apimanagement.Rawxml,
```
tombuildsstuff: since this downloads the XML from the specified link, I have a feeling this needs to be:

```suggestion
			ContentFormat: apimanagement.RawxmlLink,
```

which should still allow this to be treated as RawXML, but ensure it's downloaded rather than interpreted as XML directly

riordanp:
I have run the acceptance tests and updated test cases for the other policy resources, which also seem to fail.

The fix has sort of worked, rather  than failing during `TestAccAzureRMApiManagementAPIPolicy_basic` with a 400 error, it now fails during `TestAccAzureRMApiManagementAPIPolicy_update` with the message 

```
--- FAIL: TestAccAzureRMApiManagementAPIPolicy_update (2298.50s)
    testing.go:568: Step 1 error: After applying this step, the plan was not empty:
```

with xml_content in the diff changing like:

```
xml_content:         "<policies>\r\n\t<inbound>\r\n\t\t<set-variable name=\"abc\" value=\"@(context.Request.Headers.GetValueOrDefault(&quot;X-Header-Name&quot;, &quot;&quot;))\" />\r\n\t\t<find-and-replace from=\"xyz\" to=\"abc\" />\r\n\t</inbound>\r\n</policies>" => "<policies>\n  <inbound>\n    <set-variable name=\"abc\" value=\"@(context.Request.Headers.GetValueOrDefault(\"X-Header-Name\", \"\"))\" />\n    <find-and-replace from=\"xyz\" to=\"abc\" />\n  </inbound>\n</policies>\n"
```

It looks like API-M is adding Windows style line endings (I'm running from git bash on Windows) and encoding spaces as tabs? Any suggestions @tombuildsstuff?

tombuildsstuff:
hey @riordanp 

Sorry for the delayed response - I'd forgotten to post the test results here 🤦‍♂

> It looks like API-M is adding Windows style line endings (I'm running from git bash on Windows) and encoding spaces as tabs, but the DiffSuppressFunc on the property should prevent this type of thing?

That was my guess - digging into this it appears this is because Golang's XML parser doesn't consider the line:

```
<set-variable name="abc" value="@(context.Request.Headers.GetValueOrDefault("X-Header-Name", ""))" />
```

to be valid XML, since parsing it [via the DiffSuppressFunc](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/helpers/suppress/xml.go#L40-L42) returns the error: `Error parsing XML: XML syntax error on line 3: attribute name without = in element` - where any error returned means the Diff won't be suppressed.

That said - it appears that the quotes within the quotes returned from Azure here are encoded, as such we should be able to update the tests to account for this. I'm taking a look at this at the moment and I'm hoping that with ignoring tabs/newlines should allow this to work as expected; will update in a bit :)

jeanpaulsmit:
For me this issue is really blocking, as it can't be easily done using ARM templates just for API policies.
I see this is part of v1.36, when is this scheduled for release?

tombuildsstuff:
hey @riordanp

I spent some time yesterday digging into this - there's a couple of problems in tandem:

Since the XML Content can contain expressions, the Azure API returns this XML Encoded, which varies from the state locally. Whilst it's possible to XML Encode the XML Content from the Terraform Configuration - unfortunately this applies the encoding to the entire file (e.g. `<policies>..` becomes `&gt;policies&lt;`) which makes this invalid XML when it hits Azure. I've also spent some time trying the inverse of this - however whilst Golang itself has an EscapeString method, there's no UnescapeString method to normalise this.

In order to work around this I've had to do a combination of things which definitely isn't ideal, but appears to solve the issue. We now HTML Unencode the Policy from the API (since, this is close enough to XML Unencoding for our purposes) - however this also required a custom diff function to first try parsing this as XML (if there's no expressions) and then fall back to string comparison, which whilst isn't ideal seems to work for these purposes.

I've run the tests for this PR (and added another for the policy content on it's own) - as such I hope you don't mind but so that we can get this merged I've pushed those changes (and rebased this on top of master).

Thanks!

tombuildsstuff:
Tests pass:

Screenshot 2019-11-01 at 07 50 44

jeanpaulsmit:
Awesome, although I managed to work around this by encoding the specific expression in the policy, I'm very happy to see this solved. Thanks.

tombuildsstuff:
@jeanpaulsmit once this is merged you may need to alter the local config so that it's decoded - but I believe that scenario should be covered by the tests 👍

jeanpaulsmit:
@tombuildsstuff is this also fixed for links to externally stored policies? I'm still getting "Expecting white space" when I use a policy containing this in the inbound section:
<set-variable name="token" value="@(context.Request.Headers.GetValueOrDefault("field"))" />

while using this script:
resource "azurerm_api_management_api_policy" "apiPolicy" {
  api_name            = "${azurerm_api_management_api.api.name}"
  api_management_name = "${data.azurerm_api_management.apim.name}"
  resource_group_name = "${data.azurerm_resource_group.rg.name}"

  xml_link = "https://some-url/policy.xml"
}

tombuildsstuff:
@jeanpaulsmit this hasn't shipped yet - but it's also fixed for external/URI downloaded policies when it ships :)

jeanpaulsmit:
@tombuildsstuff aha, so that will be available in v1.37! Cool and thanks again.

PR #4168, opened 2019-08-27
`arm_role_assignment`: add `pricipal_type` and `skip_servicepricipal_aad_check` properties

jeffreyCline:
Including the `PrincipalType` parameter will eliminate a race condition where sometimes if you try to create an identity and then assign it immediately, the assignment fails because the identity hasn’t replicated in AAD yet.

### DIFF
--- a/azurerm/resource_arm_role_assignment.go
+++ b/azurerm/resource_arm_role_assignment.go
@@ -14,7 +14,6 @@ import (
 	"github.com/hashicorp/go-uuid"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
-	"github.com/hashicorp/terraform/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
@@ -66,23 +65,11 @@ func resourceArmRoleAssignment() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"principal_type": {
-				Type:     schema.TypeString,
+			"skip_service_principal_aad_check": {
+				Type:     schema.TypeBool,
 				Optional: true,
-				Computed: true,
 				ForceNew: true,
-				ValidateFunc: validation.StringInSlice([]string{
-					string(authorization.Application),
-					string(authorization.DirectoryObjectOrGroup),
-					string(authorization.DirectoryRoleTemplate),
-					string(authorization.Everyone),
-					string(authorization.ForeignGroup),
-					string(authorization.Group),
-					string(authorization.MSI),
-					string(authorization.ServicePrincipal),
-					string(authorization.Unknown),
-					string(authorization.User),
-				}, false),
+				Default: false,
 			},
 		},
 	}
@@ -145,10 +132,10 @@ func resourceArmRoleAssignmentCreate(d *schema.ResourceData, meta interface{}) e
 		},
 	}
 
-	principalType := d.Get("principal_type").(string)
+	skipPrincipalCheck := d.Get("skip_service_principal_aad_check").(bool)
 
-	if principalType != "" {
-		properties.RoleAssignmentProperties.PrincipalType = authorization.PrincipalType(principalType)
+	if skipPrincipalCheck {
+		properties.RoleAssignmentProperties.PrincipalType = authorization.ServicePrincipal
 	}
 
 	if err := resource.Retry(300*time.Second, retryRoleAssignmentsClient(scope, name, properties, meta)); err != nil {
@@ -189,7 +176,6 @@ func resourceArmRoleAssignmentRead(d *schema.ResourceData, meta interface{}) err
 		d.Set("scope", props.Scope)
 		d.Set("role_definition_id", props.RoleDefinitionID)
 		d.Set("principal_id", props.PrincipalID)
-		d.Set("principal_type", props.PrincipalType)
 
 		//allows for import when role name is used (also if the role name changes a plan will show a diff)
 		if roleId := props.RoleDefinitionID; roleId != nil {
--- a/azurerm/resource_arm_role_assignment.go
+++ b/azurerm/resource_arm_role_assignment.go
@@ -69,7 +69,7 @@ func resourceArmRoleAssignment() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 				ForceNew: true,
-				Default: false,
+				Default:  false,
 			},
 		},
 	}
--- a/azurerm/resource_arm_role_assignment.go
+++ b/azurerm/resource_arm_role_assignment.go
@@ -176,7 +176,7 @@ func resourceArmRoleAssignmentRead(d *schema.ResourceData, meta interface{}) err
 		d.Set("scope", props.Scope)
 		d.Set("role_definition_id", props.RoleDefinitionID)
 		d.Set("principal_id", props.PrincipalID)
-		d.Set("skip_service_principal_aad_check", props.PrincipalType == authorization.ServicePrincipal)
+		d.Set("skip_service_principal_aad_check", d.Get("skip_service_principal_aad_check").(bool))
 
 		//allows for import when role name is used (also if the role name changes a plan will show a diff)
 		if roleId := props.RoleDefinitionID; roleId != nil {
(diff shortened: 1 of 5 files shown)


azurerm/resource_arm_role_assignment_test.go
```diff
@@ -448,7 +492,7 @@ resource "azuread_group" "test" {
   name = "acctestspa-%d"
 }
 
-resource "azurerm_role_assignment" "test" {
+resource ServicePrincipal {
```
tombuildsstuff: this should be

```suggestion
resource "azurerm_role_assignment" "test" {
```

azurerm/resource_arm_role_assignment.go
```diff
@@ -64,6 +65,24 @@ func resourceArmRoleAssignment() *schema.Resource {
...
+				ValidateFunc: validation.StringInSlice([]string{
+					string(authorization.Application),
+					string(authorization.DirectoryObjectOrGroup),
+					string(authorization.DirectoryRoleTemplate),
+					string(authorization.Everyone),
+					string(authorization.ForeignGroup),
+					string(authorization.Group),
+					string(authorization.MSI),
+					string(authorization.ServicePrincipal),
+					string(authorization.Unknown),
```
tombuildsstuff: presumably `Unknown` shouldn't be in this list?

website/docs/r/role_assignment.html.markdown
```diff
@@ -133,6 +134,8 @@ The following arguments are supported:
 
 ~> **NOTE:** The Principal ID is also known as the Object ID (ie not the "Application ID" for applications).
 
+* `PrincipalType` - (Optional) The principal type of the assigned `principal_id`. Possible values include: `User`, `Group`, `ServicePrincipal`, `Unknown`, `DirectoryRoleTemplate`, `ForeignGroup`, `Application`, `MSI`, `DirectoryObjectOrGroup`, `Everyone`.
```
tombuildsstuff: can we sort these alphabetically (and remove 'Unknown' since this doesn't make sense?)

jeffreyCline:
> Could you also explain how exposing this property fixing AD replication eventually consistency? Those two things seem unrelated.

per conversation with the service team: "...there is one more parameter that I recommend you set on the role assignment create, including this parameter will eliminate a race condition where sometimes if you try to create an identity and then assign it immediately, the assignment fails because the identity hasn’t replicated in AAD yet.

In a REST call to ARM you would include:

“principalType”:”ServicePrincipal”

… in the body of the PUT request to create the role assignment.  I’m not sure what that translates to Terraform-wise."

azurerm/resource_arm_role_assignment.go
tombuildsstuff: just because it's in the list supported by the API doesn't mean it should be exposed to users unfortunately, Unknown is generally present to indicate a value that's unsupported on this API version (for example if it's been created on a newer/older version which is no longer supported)

tombuildsstuff:
@jeffreyCline 

> ..there is one more parameter that I recommend you set on the role assignment create, including this parameter will eliminate a race condition where sometimes if you try to create an identity and then assign it immediately, the assignment fails because the identity hasn’t replicated in AAD yet.

Sounds like we want to default this to `ServicePrincipal` then?

jeffreyCline:
@tombuildsstuff
> Sounds like we want to default this to ServicePrincipal then?

This attribute maps to the type of the Principal. If you are dealing with newly provisioned managed identities then setting to `ServicePrincipal` should be fine. In other cases, setting this to `ServicePrincipal` can lead to errors. I would suggest setting it only where needed. This is only needed for cases where principal is newly created and used within say 2-5 mins. of creation.

azurerm/resource_arm_role_assignment.go
jeffreyCline: @tombuildsstuff I have asked the service team about the expected behavior and the purpose of the `Unknown` value.

azurerm/resource_arm_role_assignment.go
```diff
@@ -164,6 +191,12 @@ func resourceArmRoleAssignmentRead(d *schema.ResourceData, meta interface{}) err
 		d.Set("role_definition_id", props.RoleDefinitionID)
 		d.Set("principal_id", props.PrincipalID)
 
+		principalType := d.Get("principal_type").(string)
+
+		if principalType != "" {
```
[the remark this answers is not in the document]
jeffreyCline: Okie-dokie... fixed. :)

azurerm/resource_arm_role_assignment.go
```diff
@@ -163,6 +176,7 @@ func resourceArmRoleAssignmentRead(d *schema.ResourceData, meta interface{}) err
 		d.Set("scope", props.Scope)
 		d.Set("role_definition_id", props.RoleDefinitionID)
 		d.Set("principal_id", props.PrincipalID)
+		d.Set("skip_service_principal_aad_check", d.Get("skip_service_principal_aad_check").(bool))
```
tombuildsstuff: since we're just reading this value back in we can remove this field (and add it to the ignore list during the import tests)

azurerm/resource_arm_role_assignment.go
```diff
@@ -64,6 +64,13 @@ func resourceArmRoleAssignment() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+
+			"skip_service_principal_aad_check": {
```
tombuildsstuff: thinking about this, I'm fine with this approach, but is there any way that we could look this information up so that users don't need to specify it?

jeffreyCline:
> hey @jeffreyCline
> Thanks for pushing those changes - this now LGTM but I had one question around whether we could infer this field rather than making users specify it, WDYT?
> Thanks!

If we wanted to do that then I believe we would have to do a get on the service principal against AAD and see if it comes back with something or not, if the service principal returns not found we could then add this value to the call automatically and avoid having this value at all. Just depends on how you want to do it.

azurerm/resource_arm_role_assignment.go
jeffreyCline: Sure thing.

azurerm/resource_arm_role_assignment.go
katbyte: given this is all a hack around broken AAD replication, and that there might be other reasons to set the type property (or at the least know it) i'm more inclined to just expose the SDK/API type field and users can set it as desired.

jeffreyCline: Per the service team:
* The only valid values that are publicly exposed by our APIs are: **User**, **Group**, and **ServicePrincipal**.

* This property is more relevant on the read path than write. **For writes, the only relevant scenario is for ServicePrincipal**. I see no value add in specifying `PrincipalType` as User or Group. The service anyways checks against AAD and fails the call if the principal is not found or the principal type in AAD doesn’t match what the request specified. Only for `ServicePrincipal`, if AAD returns principal not found, then role assignment create still proceeds in the new api-version.

* **Don’t set this parameter unless explicitly dealing with newly created service principal.**

* As stated above this attribute makes more sense in the read scenario, the API will return User, Group, ServicePrincipal based on the type of the principal in AAD.

website/docs/r/role_assignment.html.markdown
```diff
@@ -133,12 +133,16 @@ The following arguments are supported:
...
 
+* `skip_service_principal_aad_check` - (Optional) If the `principal_id` is a newly provisioned `Service Principal` set this value to `true` to skip the `Azure Active Directory` check which may fail due to replication lag. This argument is only valid if the `principal_id` is a `Service Principal` identity. If it is not a `Service Principal` identity it will cause the role assignment to fail. Defaults to `false`.
+
 ## Attributes Reference
 
 The following attributes are exported:
 
 * `id` - The Role Assignment ID.
 
+* `principal_type` - The type of the `principal_id`, e.g. User, Group, Service Principal, Application, etc.
```
katbyte: I think we can leave this as:
```suggestion
* `principal_type` - The type of the `principal_id`.
```

azurerm/resource_arm_role_assignment_test.go
```diff
@@ -24,8 +24,9 @@ func TestAccAzureRMRoleAssignment(t *testing.T) {
 			"requiresImport": testAccAzureRMRoleAssignment_requiresImport,
 		},
 		"assignment": {
-			"sp":    testAccAzureRMActiveDirectoryServicePrincipal_servicePrincipal,
-			"group": testAccAzureRMActiveDirectoryServicePrincipal_group,
+			"sp":     testAccAzureRMActiveDirectoryServicePrincipal_servicePrincipal,
+			"spType": testAccAzureRMActiveDirectoryServicePrincipal_servicePrincipalWithType,
```
katbyte: this might be better as:
```suggestion
			"spSkip": testAccAzureRMActiveDirectoryServicePrincipal_servicePrincipalWithAadSkip,
```

PR #4170, opened 2019-08-28
Update resource_arm_mariadb_server.go to support version 10.3

babinkos:
Add MariaDB version 10.3 to supported version values. Added element "10.3" to slice of strings.

### DIFF
--- a/azurerm/resource_arm_mariadb_server.go
+++ b/azurerm/resource_arm_mariadb_server.go
@@ -123,6 +123,7 @@ func resourceArmMariaDbServer() *schema.Resource {
 				ForceNew: true,
 				ValidateFunc: validation.StringInSlice([]string{
 					"10.2",
+					"10.3",
 				}, false),
 			},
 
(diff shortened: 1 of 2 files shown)


azurerm/resource_arm_mariadb_server.go
```diff
@@ -123,6 +123,7 @@ func resourceArmMariaDbServer() *schema.Resource {
 				ForceNew: true,
 				ValidateFunc: validation.StringInSlice([]string{
 					"10.2",
+					"10.3",
```
tombuildsstuff: can we add this [to the documentation too](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/mariadb_server.html.markdown#L66)? we'd generally phrase that as:

```
* `version` - (Required) Specifies the version of MariaDB to use. Possible values are `10.2` and `10.3`. Changing this forces a new resource to be created.
```

babinkos:
Hi ,
Yes, looks good !

babinkos:
Hi,
Documentation updated with new MariaDB version values

PR #4179, opened 2019-08-28
Switching `azurerm_storage_blob` over to the new Storage SDK

tombuildsstuff:
Since this is quite a complex resource (in that there's different kinds of blobs) we're going to do this migration to the new SDK in phases. This first PR switches the SDK over to using the new Giovanni Storage SDK but doesn't add support for the new features - nor does it address the open TODO's/add validation/switch the test suite over - which will come in a subsequent PR.

This PR fixes #2245

PR #4205, opened 2019-09-01
storage: caching the Resource Group Name / Account Key

tombuildsstuff:
This PR introduces caching for both the Resource Group Name and the Account Key for a Storage Account. Due to the way this works the test check functions had to be updated to confirm we get an error (404) rather than the resource group name being not found

### DIFF
--- a/azurerm/internal/services/storage/helpers.go
+++ b/azurerm/internal/services/storage/helpers.go
@@ -4,11 +4,23 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 )
 
+var (
+	accountKeysCache        = map[string]string{}
+	resourceGroupNamesCache = map[string]string{}
+	writeLock               = sync.RWMutex{}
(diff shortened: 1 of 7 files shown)


tombuildsstuff:
Ignoring a couple of known test failures, which will be fixed via #4179 and another PR (tomorrow) the tests pass:

Screenshot 2019-09-01 at 11 38 20

azurerm/internal/services/storage/helpers.go
```diff
@@ -4,11 +4,23 @@ import (
...
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 )
 
+var (
+	accountKeysCache        = map[string]string{}
+	resourceGroupNamesCache = map[string]string{}
```
tombuildsstuff: 🤔 about this some more; perhaps we want to force-clear this cache when the account's deleted

tombuildsstuff:
The Travis failure is because a dependency is offline - but the tests pass:

Screenshot 2019-09-02 at 10 48 39

PR #4210, opened 2019-09-02
New Data Source: azurerm_sql_database

leewilson86:
A new data source to be able to query Azure SQL Databases.

### DIFF
--- a/azurerm/data_source_sql_database.go
+++ b/azurerm/data_source_sql_database.go
@@ -59,7 +59,6 @@ func dataSourceSqlDatabase() *schema.Resource {
 
 			"tags": tags.Schema(),
 
-
 			"failover_group_id": {
 				Type:     schema.TypeString,
 				Computed: true,
--- a/azurerm/data_source_sql_database.go
+++ b/azurerm/data_source_sql_database.go
@@ -1,0 +1,187 @@
+package azurerm
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+	"github.com/Azure/azure-sdk-for-go/services/preview/sql/mgmt/2015-05-01-preview/sql"
+)
+
+func dataSourceSqlDatabase() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmSqlDatabaseRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
+
+			"location": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"server_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"create_mode": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"source_database_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"restore_point_in_time": {
+				Type:         schema.TypeString,
+				Computed:     true,
+			},
+
+			"edition": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"collation": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"max_size_bytes": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"requested_service_objective_id": {
+				Type:     schema.TypeString,
+				Computed: true,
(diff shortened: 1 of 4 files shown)


azurerm/data_source_sql_database.go
```diff
@@ -0,0 +1,116 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"location": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"name": {
```
katbyte: This should be at the top & first.

azurerm/data_source_sql_database.go
```diff
@@ -0,0 +1,116 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"failover_group_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"location": {
```
katbyte: Followed by this

azurerm/data_source_sql_database.go
```diff
@@ -0,0 +1,116 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"read_scale": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
```
katbyte: And the rest should follow the ordering found in the resource file.

azurerm/data_source_sql_database.go
```diff
@@ -0,0 +1,116 @@
...
+		return fmt.Errorf("Error retrieving SQL Database %q (server %q / resource group %q): %s", name, serverName, resourceGroup, err)
+	}
+
+	if id := resp.ID; id != nil {
+		d.SetId(*id)
+	}
+
+	d.Set("collation", resp.Collation)
+
+	if dsLocation := resp.ElasticPoolName; dsLocation != nil {
```
katbyte: We don't need this check as `d.Set` will check for nilness

azurerm/data_source_sql_database.go
```diff
@@ -0,0 +1,116 @@
...
+		d.Set("default_secondary_location", dsLocation)
+	}
+
+	d.Set("edition", string(resp.Edition))
+
+	if ep := resp.ElasticPoolName; ep != nil {
+		d.Set("elastic_pool_name", ep)
```
katbyte: Same here:
```suggestion
		d.Set("elastic_pool_name", props.ElasticPoolName)
```

azurerm/data_source_sql_database.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	if fogID := resp.FailoverGroupID; fogID != nil {
```
katbyte: and here

azurerm/data_source_sql_database.go
```diff
@@ -0,0 +1,116 @@
...
+		d.Set("failover_group_id", fogID)
+	}
+
+	d.Set("location", azure.NormalizeLocation(*resp.Location))
+
+	readScale := false
+	if resp.ReadScale == sql.ReadScaleEnabled {
+		readScale = true
+	}
+	d.Set("read_scale", readScale)
```
katbyte: This can be replaced with:
```suggestion
	d.Set("read_scale", props.ReadScale == sql.ReadScaleEnabled)
```

azurerm/data_source_sql_database.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		return fmt.Errorf("Error retrieving SQL Database %q (server %q / resource group %q): %s", name, serverName, resourceGroup, err)
+	}
+
+	if id := resp.ID; id != nil {
+		d.SetId(*id)
+	}
+
+	d.Set("collation", resp.Collation)
```
katbyte: Most of these properties actually reside in the anonymous property `resp.DatabaseProperties` that could be nil. Thus we need to nil check it:
```
if props := respo.DatabaseProperties; props != nil {
    d.Set("collation", props.Collation)
```

leewilson86:
@katbyte thanks for the feedback. I have addressed your comments.

MattMencel:
If we wanted to add a way to list all the databases in a single azure_sql_server instance, would you want that in a separate data resource? `azurerm_sql_databases` ?

leewilson86:
@MattMencel that’s a good question and actually something I have been thinking about myself. I’m just not sure if it’s a good idea to have data source return a list of database. What format would we return the database? Would it be an array of database resource IDs or database names? A map? A list of maps? 

If this is a reasonable feature to add I think it probably best belongs with the existing `azurerm_sql_server` data source like this: `data.azurerm_sql_server.server[0].databases`.

However, would it be better to have databases defined as vars (possibly in .tfvars) and passed into the relevant resources for creation then those vars can be used to query for database details with this data source I’ve just created? This is pretty much how the Terraform config I’m working with operates.


Maybe @katbyte can advise?

The data source I’ve added her is purely for the intention of querying a specific database on a given server.

katbyte:
@MattMencel, @leewilson86 is right in that the best place for that is the `azurerm_sql_server` datasource as `azurerm_sql_server.databases.

PR #4212, opened 2019-09-02
tests: add tf linting R checks & fix errors

katbyte:


GNUmakefile
```diff
@@ -61,17 +68,17 @@ testacc: fmtcheck
 debugacc: fmtcheck
 	TF_ACC=1 dlv test $(TEST) --headless --listen=:2345 --api-version=2 -- -test.v $(TESTARGS)
 
+website-lint:
+	@echo "==> Checking website against linters..."
+	@misspell -error -source=text -i hdinsight website/
+
```
tombuildsstuff: alphabetically this one wants to be below the other one?

azurerm/resource_arm_function_app.go
```diff
@@ -205,7 +205,6 @@ func resourceArmFunctionApp() *schema.Resource {
 						"virtual_network_name": {
 							Type:     schema.TypeString,
 							Optional: true,
-							Computed: false,
```
tombuildsstuff: is this intentional?

katbyte: Yea, `Computed: false,` doesn't actually mean anything right?

GNUmakefile
katbyte: yestrue but aesthetically and functionally I think it's better above given the last two massive blocks run together and i spent a while skimming the file looking for it

PR #4213, opened 2019-09-02
provider: deprecate and move final functions in provider.go

katbyte:


azurerm/resource_arm_application_gateway.go
```diff
@@ -25,6 +25,15 @@ func possibleArmApplicationGatewaySslCipherSuiteValues() []string {
 	return cipherSuites
 }
 
+func base64EncodedStateFunc(v interface{}) string {
```
tombuildsstuff: is this intentionally here, if not should this go into `deprecated.go`?

azurerm/utils/base64.go
```diff
@@ -0,0 +1,20 @@
...
+func Base64EncodeIfNot(data string) string {
+	// Check whether the data is already Base64 encoded; don't double-encode
+	if Base64IsEncoded(data) {
+		return data
+	}
+	// data has not been encoded encode and return
+	return base64.StdEncoding.EncodeToString([]byte(data))
+}
+
+func Base64IsEncoded(data string) bool {
```
tombuildsstuff: since this isn't used outside of this package we can make this private:

```suggestion
func base64IsEncoded(data string) bool {
```

azurerm/resource_arm_virtual_machine.go
```diff
@@ -26,6 +28,22 @@ import (
...
+func userDataStateFunc(v interface{}) string {
+	switch s := v.(type) {
+	case string:
+		s = utils.Base64EncodeIfNot(s)
+		hash := sha1.Sum([]byte(s))
+		return hex.EncodeToString(hash[:])
+	default:
+		return ""
+	}
+}
```
tombuildsstuff: should these be moved into `deprecated.go` for now?

azurerm/helpers/state/state.go
```diff
@@ -0,0 +1,9 @@
+package state
+
+import "strings"
+
+// ignoreCaseStateFunc is a StateFunc from helper/schema that converts the
```
tombuildsstuff:
```suggestion
// IgnoreCase is a StateFunc from helper/schema that converts the
```

azurerm/resource_arm_virtual_machine.go
```diff
@@ -26,6 +28,22 @@ import (
 
 var virtualMachineResourceName = "azurerm_virtual_machine"
 
+// todo move into internal/services/compute/something.go
```
tombuildsstuff: as discussed offline, perhaps `./azurerm/internal/(tf/)state`?

azurerm/resource_arm_application_gateway.go
katbyte: at the moment it is only used by this resource, so moved it here

Issue #3696, opened 2019-06-18
azurerm_app_service unable to configure source control.


JayDoubleu:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently source_control block only applies to localGit. There seems to be no way to configure source control for any other scm methods. 
Using this module with scm_type set to other than None or LocalGit results in error:

[...]

PR #4214, opened 2019-09-02
New Resource: `azurerm_app_service_source_control_token`

joakimhellum-in:
Reference to #3696.

This is first of multiple pull requests to enable support for [App Service deployment](https://docs.microsoft.com/en-us/azure/app-service/deploy-continuous-deployment) in Terraform AzureRM provider. 

This adds new resource `azurerm_app_service_source_control_token` for updating source control tokens. Another way of doing this is to run the [`az webapp deployment source update-token`](https://docs.microsoft.com/en-us/cli/azure/webapp/deployment/source?view=azure-cli-latest#az-webapp-deployment-source-update-token) command.

Have not implemented `Delete` for this resource because it only supports `Update`. Users can however set `token` to empty string. 

~~Not really sure about this resource, but~~ it would be nice to manage this in Terraform. Any advice is much appreciated.

### DIFF
--- a/azurerm/resource_arm_app_service_source_control_token.go
+++ b/azurerm/resource_arm_app_service_source_control_token.go
@@ -7,8 +7,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/web/mgmt/2018-02-01/web"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
-	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
-	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/locks"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
@@ -17,10 +15,10 @@ var appServiceSourceControlTokenResourceName = "azurerm_app_service_source_contr
 
 func resourceArmAppServiceSourceControlToken() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceArmAppServiceSourceControlTokenCreateOrUpdate,
+		Create: resourceArmAppServiceSourceControlTokenCreateUpdate,
 		Read:   resourceArmAppServiceSourceControlTokenRead,
-		Update: resourceArmAppServiceSourceControlTokenCreateOrUpdate,
-		Delete: schema.Noop,
+		Update: resourceArmAppServiceSourceControlTokenCreateUpdate,
+		Delete: resourceArmAppServiceSourceControlTokenDelete,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
@@ -38,14 +36,15 @@ func resourceArmAppServiceSourceControlToken() *schema.Resource {
 			},
 
 			"token": {
-				Type:     schema.TypeString,
-				Required: true,
+				Type:      schema.TypeString,
+				Required:  true,
+				Sensitive: true,
 			},
 		},
 	}
 }
 
-func resourceArmAppServiceSourceControlTokenCreateOrUpdate(d *schema.ResourceData, meta interface{}) error {
+func resourceArmAppServiceSourceControlTokenCreateUpdate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*ArmClient).web.BaseClient
 	ctx := meta.(*ArmClient).StopContext
 
@@ -56,19 +55,6 @@ func resourceArmAppServiceSourceControlTokenCreateOrUpdate(d *schema.ResourceDat
 
 	locks.ByName(scmType, appServiceSourceControlTokenResourceName)
 	defer locks.UnlockByName(scmType, appServiceSourceControlTokenResourceName)
-
-	if features.ShouldResourcesBeImported() && d.IsNewResource() {
-		existing, err := client.GetSourceControl(ctx, scmType)
-		if err != nil {
-			if !utils.ResponseWasNotFound(existing.Response) {
-				return fmt.Errorf("Error checking for presence of existing App Service Source Control Token (Type %q): %s", scmType, err)
-			}
-		}
-
-		if existing.ID != nil && *existing.ID != "" {
-			return tf.ImportAsExistsError(appServiceSourceControlTokenResourceName, *existing.ID)
-		}
-	}
 
 	properties := web.SourceControl{
 		SourceControlProperties: &web.SourceControlProperties{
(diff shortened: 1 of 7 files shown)


katbyte:
@joakimhellum-in,

Am i reading this right in that you can only have one token per type subscription? 🤔

joakimhellum-in:
@katbyte,

Yes 😔 the source control token is per subscription but the source control configuration is per app.

This is alternative to configuring it in in the portal or CLI. When users configure this in the portal, then it uses the `Azure App Service` GitHub OAuth App, which acts as the authenticated user. Meaning all other users with the proper permissions will act as the user who first authenticated in this context, and can read the token. 

This is how it looks in the portal:
image

And this is the CLI command for the same: [`az webapp deployment source update-token`](https://docs.microsoft.com/en-us/cli/azure/webapp/deployment/source?view=azure-cli-latest#az-webapp-deployment-source-update-token)

We find it more helpful to automate this with Terraform, compared to using portal or CLI, because we have alot of subscriptions in the organization.

joakimhellum-in:
@katbyte for what it's worth we are making more pull requests over next few days that may help better understand how this fits together with App Service deployment.

azurerm/resource_arm_app_service_source_control_token.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/locks"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+var appServiceSourceControlTokenResourceName = "azurerm_app_service_source_control_token"
+
+func resourceArmAppServiceSourceControlToken() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceArmAppServiceSourceControlTokenCreateOrUpdate,
```
katbyte: Could we use `CreateUpdate` for consistency with other resources?
```suggestion
		Create: resourceArmAppServiceSourceControlTokenCreateOrUpdate,
```

azurerm/resource_arm_app_service_source_control_token.go
```diff
@@ -0,0 +1,119 @@
...
+		Read:   resourceArmAppServiceSourceControlTokenRead,
+		Update: resourceArmAppServiceSourceControlTokenCreateOrUpdate,
+		Delete: schema.Noop,
```
katbyte: Should we implement delete as setting the token to "" ? It stands to reason that if the resource is deleted the data should be cleared out. WDYT?

website/docs/r/app_service_source_control_token.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+page_title: "Azure Resource Manager: azurerm_app_service_source_control_token"
+sidebar_current: "docs-azurerm-resource-app-service-source-control-token"
+description: |-
+  Manages an App Service source control token.
+
+---
+
+# azurerm_app_service_source_control_token
+
+Manages an App Service source control token.
```
katbyte: It might be good to have a note here  how this is global to the subscription/type?

azurerm/resource_arm_app_service_source_control_token.go
joakimhellum-in: This sounds very nice. Many thanks. Will add this tomorrow.

azurerm/resource_arm_app_service_source_control_token.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		if existing.ID != nil && *existing.ID != "" {
+			return tf.ImportAsExistsError(appServiceSourceControlTokenResourceName, *existing.ID)
+		}
+	}
+
+	properties := web.SourceControl{
+		SourceControlProperties: &web.SourceControlProperties{
+			Token: utils.String(token),
```
tombuildsstuff: should we also be exposing the other fields here:

```
// SourceControlProperties sourceControl resource specific properties
type SourceControlProperties struct {
	// Token - OAuth access token.
	Token *string `json:"token,omitempty"`
	// TokenSecret - OAuth access token secret.
	TokenSecret *string `json:"tokenSecret,omitempty"`
	// RefreshToken - OAuth refresh token.
	RefreshToken *string `json:"refreshToken,omitempty"`
	// ExpirationTime - OAuth token expiration.
	ExpirationTime *date.Time `json:"expirationTime,omitempty"`
}
```

azurerm/resource_arm_app_service_source_control_token.go
```diff
@@ -0,0 +1,119 @@
...
+					"BitBucket",
+					"Dropbox",
+					"GitHub",
+					"OneDrive",
+				}, false),
+			},
+
+			"token": {
+				Type:     schema.TypeString,
+				Required: true,
```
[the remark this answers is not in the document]
joakimhellum-in: 🤦‍♂️

azurerm/resource_arm_app_service_source_control_token.go
joakimhellum-in: Have checked the [Azure](https://docs.microsoft.com/en-us/azure/app-service/deploy-continuous-deployment#authorize-azure-app-service) and [Azure Stack](https://docs.microsoft.com/en-us/azure-stack/operator/azure-stack-app-service-configure-deployment-sources) documentation for configuring deployment sources, and made some more tests with the different source control providers. Based on this tests we should add the `token_secret` field. I cannot find any docs regarding `refresh_token` and `expiration_time`. I assume they should be set together, but don't know how to properly test it, since I don't know in what context they should be used. Let me know if we still should add `refresh_token` and `expiration_time`. Thanks!

website/docs/r/app_service_source_control_token.html.markdown
joakimhellum-in: Have made an attempt to fix this, but it might need to be adjusted to proper English, sorry.

azurerm/resource_arm_app_service_source_control_token.go
joakimhellum-in: Have made an attempt to fix this. Not sure what the most explicit and readable way to do this is.

joakimhellum-in:
Thanks @katbyte @tombuildsstuff for your patience here, it feels alot better now with `Delete` implemented.

> Since this is likely to be a common ask, I'm wondering if it's worth adding [an example to the examples folder](https://github.com/terraform-providers/terraform-provider-azurerm/tree/master/examples/app-service) as a part of this PR - WDYT?

Yes, we are making another pull request during today or tomorrow to support setting `repo_url`, `branch`, and `is_manual_integration` properties in the `source_control` block of an App Service. I guess this should be complete example with both a public and private repo specified in the `source_control` block (with and without a token).

azurerm/resource_arm_app_service_source_control_token.go
```diff
@@ -0,0 +1,146 @@
...
+
+func resourceArmAppServiceSourceControlTokenDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).web.BaseClient
+	ctx := meta.(*ArmClient).StopContext
+
+	scmType := d.Id()
+
+	// Delete cleans up existing tokens by setting their values to ""
+	token := ""
+	tokenSecret := ""
```
tombuildsstuff: out of interest, does this disable this functionality, or just make this password-less? if so, do we want to generate some random (e.g. a GUID) value here instead?

website/docs/r/app_service_source_control_token.html.markdown
```diff
@@ -0,0 +1,41 @@
...
+
+~> **NOTE:** A source control token is global and required only once per Azure account.
```
tombuildsstuff: would this be clearer as:

```suggestion
~> **NOTE:** Source Control Token's are configured at the subscription level, not on each App Service - as such this can only be configured Subscription-wide
```

azurerm/resource_arm_app_service_source_control_token.go
joakimhellum-in: Yes, this functionality is "disabled" when token is set to `""`, so it works pretty well for implementing `Delete` :) 

Example: If you first delete the source control token (set token tok ""), then try to deploy an app service with source control and continuous integration (which configures webhooks into online repos like GitHub), you will receive this error from the API:
```
Error: Error updating Source Control for App Service "test-56ea" (Resource Group "test-56ea"): web.AppsClient#CreateOrUpdateSourceControl: Failure sending request: StatusCode=404 -- Original Error: Code="NotFound" Message="Cannot find SourceControlToken with name GitHub." Details=[{"Message":"Cannot find SourceControlToken with name GitHub."},{"Code":"NotFound"},{"ErrorEntity":{"Code":"NotFound","ExtendedCode":"51004","Message":"Cannot find SourceControlToken with name GitHub.","MessageTemplate":"Cannot find {0} with name {1}.","Parameters":["SourceControlToken","GitHub"]}}]
```
If the token is set to anything else than `""` it will try to authenticate with it.

tombuildsstuff:
Tests pasS:

Screenshot 2019-09-06 at 11 34 44

PR #4226, opened 2019-09-03
azurerm_api_management: invert ssl tls flags

### LINKED ISSUE #3125: azurerm_api_management.security boolean settings are the wrong way round
### Community Note

Terraform: 0.11.13
AzureRM: 1.23

### Affected Resource(s)

* `azurerm_api_management`

```hcl
resource "azurerm_api_management" "apim" {
  name                = "apim-${var.environment}"
  location            = "${azurerm_resource_group.apim.location}"
  resource_group_name = "${azurerm_resource_group.apim.name}"
  publisher_name      = "Sean"
  publisher_email     = "${var.apim_email}"
  sku {
    name     = "${var.apim_sku}"
    capacity = "${var.apim_capacity}"
  }
  security {
    # These parameters are the wrong way round - true means enable
    disable_backend_ssl30 = false
    disable_backend_tls10 = false
    disable_backend_tls11 = false
    disable_frontend_ssl30 = false
    disable_frontend_tls10 = false
    disable_frontend_tls11 = false
    disable_triple_des_chipers = false
  }
}
```

### Expected Behavior

Setting a disable_<protocol> setting to true should disable the protocol.
Setting a disable_<protocol> setting to false should enable the protocol.

### Actual Behavior

Setting a disable_<protocol> setting to true enables the protocol.
Setting a disable_<protocol> setting to false disables the protocol.

### Steps to Reproduce

1. `terraform apply`

alastairtree:
This is a naive, untested fix, but I think it should correct the wrongly inverted properties?

disable_backend*** and disable_frontend**** terraform properties for APIM are all the wrong way round - disable in terraform should not enable them in AzureRM

Fixes #3125


alastairtree:
New properties makes sense so it lines up with the ARM API, but I kinda think somewhere has to make a breaking change or fix it properly - it is currently a broken security issue. In my case I found it because a security audit noticed that the old SSL we thought we had disabled was actually wide open. Not convinced deprecating the old properties is really good enough - they insecurely do the exact opposit of what they say they should! Although I understand that some people might have deliberatley passed in the inverted wrong value to achieve the desired outcome, i suspect more people are just getting the wrong behaviour because the code is wrong and needs fixing. Either way the experience is not good, so someone just needs to make a decision.

katbyte:
@alastairtree, if we flip it we'll break anyone using it property. By creating new properties and deprecating the old it means users will get a message stating exactly why they should switch, and then when we release 2.0 in the upcoming months we can remove the incorrect ones entirely.

katbyte:
Hi @alastairtree, as i haven't received a response nor can push to you branch i have opened #4226 with my suggested fix and am going to close this in favour of it. Thanks again for the PR!

PR #4235, opened 2019-09-04
storage: switching to use SharedKey for authentication

tombuildsstuff:
This PR switches the Blobs and Containers Clients over to using SharedKey rather than SharedKeyLite for authentication, which fixes uploading data using a premium storage account. This PR also adds some additional test cases in this area


tombuildsstuff:
Tests pass:

Screenshot 2019-09-04 at 10 14 51

stack72:
Oh this is awesome! This will kill that other bug I talked about today :)

PR #4245, opened 2019-09-04
New Resource: `azurerm_bot_channels_registration`

mbfrahry:


mbfrahry:
image

azurerm/resource_arm_bot_channels_registration.go
```diff
@@ -0,0 +1,270 @@
...
+					string(botservice.F0),
+					string(botservice.S1),
+				}, false),
+			},
+
+			"microsoft_app_id": {
+				Type:         schema.TypeString,
+				ForceNew:     true,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
```
katbyte: I think we could use the resource ID validation here?

azurerm/resource_arm_bot_channels_registration.go
```diff
@@ -0,0 +1,270 @@
...
+				return fmt.Errorf("Error generating import ID for Bot Channels Registration %q (Resource Group %q): %+v", name, resourceGroup, err)
+			}
+
+			return tf.ImportAsExistsError("azurerm_bot", id)
+		}
+	}
+
+	t := d.Get("tags").(map[string]interface{})
+	displayName := d.Get("display_name").(string)
+	if displayName == "" {
```
katbyte: Is this mimicking the portal behaviour?

azurerm/resource_arm_bot_channels_registration.go
```diff
@@ -0,0 +1,270 @@
...
+		Location: utils.String(d.Get("location").(string)),
+		Sku: &botservice.Sku{
+			Name: botservice.SkuName(d.Get("sku").(string)),
+		},
+		Kind: botservice.KindBot,
+		Tags: tags.Expand(t),
+	}
+
+	_, err := client.Create(ctx, resourceGroup, name, bot)
+	if err != nil {
```
katbyte: We could merge these two lines:
```
	if _, err := client.Create(ctx, resourceGroup, name, bot); err != nil {```

File: azurerm/resource_arm_bot_channels_registration.go [hunk 4]
```diff
@@ -0,0 +1,270 @@
...
+		Location: utils.String(d.Get("location").(string)),
+		Sku: &botservice.Sku{
+			Name: botservice.SkuName(d.Get("sku").(string)),
+		},
+		Kind: botservice.KindBot,
+		Tags: tags.Expand(t),
+	}
+
+	_, err := client.Update(ctx, resourceGroup, name, bot)
+	if err != nil {
```
katbyte: These two lines could be merged?

website/docs/r/bot_channels_registration.markdown
```diff
@@ -0,0 +1,71 @@
...
+
+* `resource_group_name` - (Required) The name of the resource group in which to create the Bot Channels Registration. Changing this forces a new resource to be created.
+
+* `location` - (Required) The supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `sku` - (Required) The SKU of the Bot Channels Registration. Valid values include `F0` or `S1`. Changing this forces a new resource to be created.
+
+* `microsoft_app_id` - (Required) The Microsoft Application ID for the Bot Channels Registration. Changing this forces a new resource to be created.
+
+* `display_name` - (Optional) The name of the Bot Channels Registration will be displayed as.
```
katbyte: We should note that this defaults the name if not specified?

azurerm/resource_arm_bot_channels_registration.go
mbfrahry: Sure is!

website/docs/r/bot_channels_registration.markdown
```diff
@@ -0,0 +1,71 @@
...
+The following attributes are exported:
+
+* `id` - The Bot Channels Registration ID.
+
+## Import
+
+Bot Channels Registration can be imported using the `resource id`, e.g.
+
+```shell
+terraform import azurerm_bot_channels_registration.example /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/example/providers/Microsoft.BotService/botServices/example
```
tombuildsstuff: since this is a new Resource Provider we also [need to add this to the list of Resource Providers which are registered automatically (which is case sensitive `Microsoft.BotService`](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/required_resource_providers.go#L21)

PR #4246, opened 2019-09-05
[WIP] `azurerm_app_service`: support for setting fields within the `source_control` block

joakimhellum-in:
Reference to #4214

Not sure how to best manage setting fields within `source_control` block together with `scm_type` field within `site_config` block for best user experience. I only ever used this field for `LocalGit` and setting it when using other source control providers might cause conflict (ref. #3696). It seems this field and allowed values was added without thinking about it's functionality. Will test this further over the next few days, then commit the changes. A suggestion is to ~~move `scm_type` from `site_config` block to `source_control` block, rename it to `type` and then modify it's behaviour~~ not make breaking changes, which makes this a bit of hack. Any feedback is appreciated.

FYI the changes currently proposed in this pull request will not work because there is a conflict when setting `scm_type` within `site_config` block before updating `source_control`. 

TODO:
- [ ] Add examples to the examples folder

### DIFF
--- a/azurerm/data_source_app_service.go
+++ b/azurerm/data_source_app_service.go
@@ -211,7 +211,7 @@ func dataSourceArmAppServiceRead(d *schema.ResourceData, meta interface{}) error
 		return err
 	}
 
-	scm := flattenAppServiceSourceControl(scmResp.SiteSourceControlProperties)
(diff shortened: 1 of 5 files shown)


joakimhellum-in:
Have not found any good way to implement this without causing breaking changes currently, so closing this in favor of #4277

PR #4253, opened 2019-09-05
refactor: adding a Base Client

tombuildsstuff:
This PR continues the refactoring into Service Packages by introduces a Base Client and moves the StopContext into it (since this is used everywhere)

Since this BaseClient is nested within the ArmClient, all of the fields should be available both at the top level and the inner level, e.g.

```
type Client struct {
  StopContext context.Context
}

type ArmClient struct {
  common.Client
}

func foo() {
  client := ArmClient{}

  client.StopContext // available
  client.Client.StopContext // available
}
```

.. which means that we should be able to move fields into the Base class gradually, leave the references pointing to the top-level `meta.(*ArmClient).StopContext` and then find & replace the `meta.(*ArmClient)` to `meta.(*common.Client)` later on


tombuildsstuff:
Since the tests for this will affect all resources in the same manner, it should be sufficient to run one set of tests to confirm this change is fine - so the App Service tests pass:

Screenshot 2019-09-12 at 07 27 22

Issue #4262, opened 2019-09-05
git.apache.org/thrift.git unavailable and fails make build


damoodamoo:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
TF 0.12.7
go 1.12

### Affected Resource(s)

* `make build` (when building the provider)

### Expected Behavior

`make build` should succeed on a fresh cloned copy of the repo. 

### Actual Behavior

Hangs on line getting git.apache.org/thrift. Eventually times out and fails. Other repos having the same issue - https://github.com/google/go-cloud/issues/1635. 

```
go: git.apache.org/thrift.git@v0.0.0-20180902110319-2566ecd5d999: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /home/damoo/repos/go/pkg/mod/cache/vcs/83dba939f95a790e497d565fc4418400145a1a514f955fa052f662d56e920c3e: exit status 128:
        fatal: unable to access 'https://git.apache.org/thrift.git/': Failed to connect to git.apache.org port 443: Connection timed out
go: error loading module requirements
GNUmakefile:20: recipe for target 'build' failed
make: *** [build] Error 1
```

### Steps to Reproduce

1. Use go 1.12 (having cleaned out packages)
2. Clone the repo
3. Run `make build`

damoodamoo:
Solution to this seems to be to use `go 1.13` - as mentioned in another related issue in the `go-cloud` repo. Added issue here for visibility.

katbyte:
Hi @damoodamoo,

We are not about to update the requirements to 1.13 yet. However you should be able to build the provider by using the vendor directory. Set the environment variable: `GOFLAGS=-mod=vendor`.

damoodamoo:
Ok thanks - it’s good for me on 1.13 but hopefully this might save someone else new to the repo a few mins. Cheers!

PR #4266, opened 2019-09-05
Updated README to stipulate go-113

damoodamoo:
Addressing #4262 - tweak to README to stipulate go version 1.13 is required to run `make build`


katbyte:
Hi @damoodamoo, go 1.13 is not actually required. The apache git org is down at the moment and i believe it will be restored. for now you can switch to vendoring mod. For more information please see my comment on your issue.

PR #4277, opened 2019-09-08
New Resource: `azurerm_app_service_source_control`

joakimhellum-in:
Reference to https://github.com/terraform-providers/terraform-provider-azurerm/issues/3696#issuecomment-529224458 and #4246

This is attempt to add support for managing source control for an App Service without causing breaking changes. More can be done later to improve this :)

It would be nice to have your blessing for this new resource ~~before adding all the examples and their READMEs, because it's quite extensive (configuring OAuth App/Personal Access Token in Butbucket and GitHub, configuring source control for manual integration, and source control tokens for continuous integration)~~, have added very basic examples.

TODO:
- [x] Add examples to examples folder


joakimhellum-in:
Because of change in priorities I'm no longer able to follow up on this pull request, and since the forked repo no longer exists, I will close this pull request. Sorry about that.

PR #4288, opened 2019-09-09
`azurerm_devspace_controller` Bump API version

nexxai:
Fixing after rebasing/reforking


nexxai:
Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/3805

nexxai:
@katbyte Sorry, I didn't even clue into the fact that I was still on the preview branch.  Upgraded to 2019-04-01 (full public).

katbyte:
Looks like there is a problem with the latest API version too:
```
testing.go:568: Step 0 error: After applying this step, the plan was not empty:
        
        DIFF:
        
        DESTROY/CREATE: azurerm_devspace_controller.test
          data_plane_fqdn:                          "azds-d4a0111e-cb75-45e2-9b14-424a2874f154.weu.azds.io" => "<computed>"
          host_suffix:                              "rwxnnlkqxm.weu.azds.io" => "suffix" (forces new resource)
```
😞

tombuildsstuff:
👋 

Taking a quick look here - unfortunately I can't immediately find the API documentation for DevSpace, but I'm wondering if the `host_suffix` field has become Computed in this latest API version, rather than being user-specifiable?

katbyte:
Opened https://github.com/Azure/azure-sdk-for-go/issues/5990 on the SDK to get more context into the issue

tombuildsstuff:
hey @nexxai 

Taking a closer look into this it appears the `host_suffix` field is now configured by Azure rather than being user-configurable - as such we need to update the API version being used we need to make this field `Computed` (meaning we'll read the value from whatever Azure tells us) rather than user-configurable.

Since fixing this is required to update the Azure SDK (which is a dependency for some other things we're working on) and this PR has conflicts (and for some reason doesn't show the origin repo/branch) - whilst I'd like to thank you for this contribution so that we can get this merged I'm going to close this PR in favour of a new one which updates the API version and makes this field user configurable - I hope you don't mind!

Thanks!

PR #4220, opened 2019-09-03
r/postgresql_server: support storage auto-grow

evenh:
Add support for Azure PostgreSQL storage auto-grow. It went into GA on
June 26th 2019.

It is enabled by default, to conform with the behaviour described in
the Azure CLI docs .

This resolves #3885

PR #4302, opened 2019-09-12
Mariadb storage autogrow

r0bnet:
Requested here: https://github.com/terraform-providers/terraform-provider-azurerm/issues/4295

This is the MariaDB part only. MySQL to follow

### DIFF
--- a/azurerm/resource_arm_mariadb_server.go
+++ b/azurerm/resource_arm_mariadb_server.go
@@ -153,6 +153,15 @@ func resourceArmMariaDbServer() *schema.Resource {
 								string(mariadb.Disabled),
 							}, false),
 						},
+
+						"storage_autogrow": {
+							Type:     schema.TypeString,
+							Optional: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(mariadb.StorageAutogrowEnabled),
+								string(mariadb.StorageAutogrowDisabled),
+							}, false),
+						},
 					},
 				},
 			},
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_mariadb_server.go
```diff
@@ -153,6 +153,15 @@ func resourceArmMariaDbServer() *schema.Resource {
 								string(mariadb.Disabled),
 							}, false),
 						},
+
+						"storage_autogrow": {
+							Type:     schema.TypeString,
+							Optional: true,
+							ValidateFunc: validation.StringInSlice([]string{
```
tombuildsstuff: in the PostgreSQL version #4220 this is defaulted to 'Enabled' - do we want to do the same here?

azurerm/resource_arm_mariadb_server.go
```diff
@@ -153,6 +153,15 @@ func resourceArmMariaDbServer() *schema.Resource {
 								string(mariadb.Disabled),
 							}, false),
 						},
+
+						"storage_autogrow": {
```
tombuildsstuff: could we make this `auto_grow` to match the PostgreSQL resource?

azurerm/resource_arm_mariadb_server_test.go
```diff
@@ -344,6 +372,7 @@ resource "azurerm_mariadb_server" "test" {
     storage_mb            = 51200
     backup_retention_days = 7
     geo_redundant_backup  = "Disabled"
+	storage_autogrow      = "Disabled"
```
tombuildsstuff: for the Basic test could we omit this (since this is only the optional properties)

PR #4331, opened 2019-09-16
compute: upgrading the API Version to 2019-07-01

tombuildsstuff:
This PR upgrades the version of the Compute API to 2019-07-01, which is required for building support for the new Virtual Machine & Virtual Machine Scale Set resources for 2.0


tombuildsstuff:
Ran the test suite and the Acceptance Tests look ok 👍

PR #4335, opened 2019-09-16
keyvault_certificate: Store raw data as hex

jen20:
This commit changes the `certificate_data` field of an `azurerm_keyvault_certificate` to a hexadecimal representation, in common with the `thumbprint`. This is a better approach than calling `string(x)` directly on a byte array, since certificate data usually contains invalid UTF-8, and therefore cannot be processed by a wide array of tools downstream.

Documentation is updated to reflect the new format, but since this is a computed value, no state migration function is supplied.


tombuildsstuff:
Tests pass:

Screenshot 2019-09-18 at 17 23 41

PR #4339, opened 2019-09-16
azurerm_kubernetes_cluster: pass service_pricipal.client_secret during read

katbyte:


invidian:
Looks like because of that now Terraform wants to re-create AKS clusters. Snippet from plan:
```
16:34:51        service_principal.0.client_secret:                      <sensitive> => <sensitive> (forces new resource)
16:34:51        tags.%:                                                 "1" => "1"
```

Issue #3748, opened 2019-06-28
azurerm_public_ip_prefix does not contain a "Data Source" version


vsabella:
The resource azurerm_public_ip_prefix does not contain an equivalent "data source" object, so it is difficult to create a global public ip prefix shared by many different terraform modules.

PR #4340, opened 2019-09-17
Adding data source 'azurerm_public_ip_prefix'

midacts:
This PR is in response to #3748.

I still do not fully understand how everything works (although I would like to learn more), but I think this is all that is required to get this data source added and working.

### Acceptance Tests
The test runs successfully.
It creates a `public_ip_prefix` and also creates a data source of that resource that the test verifies against.
image.

### Website Documentation
I am pretty sure I got everything correct (but that is one of the main things I am not 100% sure about).

in the `Attributes Reference` section, I did not add `zones` as it is specific region only (that is what the resource object says).

### DIFF
--- a/azurerm/data_source_public_ip_prefix.go
+++ b/azurerm/data_source_public_ip_prefix.go
@@ -38,9 +38,9 @@ func dataSourceArmPublicIpPrefix() *schema.Resource {
 				Computed: true,
 			},
 
-			"zones": azure.SchemaSingleZone(),
+			"zones": azure.SchemaZonesComputed(),
 
-			"tags": tags.Schema(),
+			"tags": tags.SchemaDataSource(),
 		},
 	}
 }
@@ -50,16 +50,26 @@ func dataSourceArmPublicIpPrefixRead(d *schema.ResourceData, meta interface{}) e
 	ctx := meta.(*ArmClient).StopContext
 
 	name := d.Get("name").(string)
-	resGroup := d.Get("resource_group_name").(string)
-	resp, err := client.Get(ctx, resGroup, name, "")
+	resourceGroup := d.Get("resource_group_name").(string)
+	resp, err := client.Get(ctx, resourceGroup, name, "")
 	if err != nil {
 		if utils.ResponseWasNotFound(resp.Response) {
-			return fmt.Errorf("Error: Public IP prefix %q was not found", name)
+			return fmt.Errorf("Error: Public IP prefix %q was not found in Resource Group %q", name, resourceGroup)
 		}
-		return err
+		return fmt.Errorf("Error retrieving Public IP Prefix %q (Resource Group %q): %+v", name, resourceGroup, err)
 	}
 
 	d.SetId(*resp.ID)
-
-	return resourceArmPublicIpPrefixRead(d, meta)
+	d.Set("zones", resp.Zones)
+	if location := resp.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	if sku := resp.Sku; sku != nil {
+		d.Set("sku", string(sku.Name))
+	}
+	if props := resp.PublicIPPrefixPropertiesFormat; props != nil {
+		d.Set("prefix_length", props.PrefixLength)
+		d.Set("ip_prefix", props.IPPrefix)
+	}
+	return tags.FlattenAndSet(d, resp.Tags)
 }
--- a/azurerm/data_source_public_ip_prefix.go
+++ b/azurerm/data_source_public_ip_prefix.go
@@ -1,0 +1,65 @@
+package azurerm
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmPublicIpPrefix() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmPublicIpPrefixRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"location": azure.SchemaLocationForDataSource(),
+
+			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
+
+			"sku": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"prefix_length": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
+			"ip_prefix": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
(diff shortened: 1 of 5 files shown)


azurerm/data_source_public_ip_prefix.go
```diff
@@ -0,0 +1,65 @@
...
+			},
+
+			"ip_prefix": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"zones": azure.SchemaSingleZone(),
+
+			"tags": tags.Schema(),
```
tombuildsstuff: since this is a Data Source these are ReadOnly - as such could we make this:

```suggestion
			"tags": tags.SchemaDataSource(),
```

azurerm/data_source_public_ip_prefix.go
```diff
@@ -0,0 +1,65 @@
...
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
+			"ip_prefix": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"zones": azure.SchemaSingleZone(),
```
tombuildsstuff: since this is a Data Source these are ReadOnly - as such could we make this:

```suggestion
			"zones": azure.SchemaZonesComputed(),
```

azurerm/data_source_public_ip_prefix.go
```diff
@@ -0,0 +1,65 @@
...
+func dataSourceArmPublicIpPrefixRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).network.PublicIPPrefixesClient
+	ctx := meta.(*ArmClient).StopContext
+
+	name := d.Get("name").(string)
+	resGroup := d.Get("resource_group_name").(string)
+	resp, err := client.Get(ctx, resGroup, name, "")
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return fmt.Errorf("Error: Public IP prefix %q was not found", name)
```
tombuildsstuff: could we wrap this to contain more information:

```suggestion
			return fmt.Errorf("Error: Public IP prefix %q was not found in Resource Group %q", name, resourceGroup)
```

azurerm/data_source_public_ip_prefix.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		return err
```
tombuildsstuff: could we wrap this error to make this clearer what's failed:

```suggestion
		return fmt.Errorf("Error retrieving Public IP Prefix %q (Resource Group %q): %+v", name, resourceGroup, err)
```

azurerm/data_source_public_ip_prefix.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	d.SetId(*resp.ID)
+
+	return resourceArmPublicIpPrefixRead(d, meta)
```
tombuildsstuff: in-case the Schema's diverge over time - rather than calling out to the Resource's Read Function could we instead set these inline here?

```suggestion
	d.Set("zones", resp.Zones)
	if location := resp.Location; location != nil {
		d.Set("location", azure.NormalizeLocation(*location))
	}

	if sku := resp.Sku; sku != nil {
		d.Set("sku", string(sku.Name))
	}

	if props := resp.PublicIPPrefixPropertiesFormat; props != nil {
		d.Set("prefix_length", props.PrefixLength)
		d.Set("ip_prefix", props.IPPrefix)
	}

	return tags.FlattenAndSet(d, resp.Tags)
```

midacts:
Thanks for your review and your suggestions.

I will take a look at those and how they work and commit the changes.

azurerm/data_source_public_ip_prefix.go
midacts: Thanks for finding this.

I was curious what the difference is?
It [looks like](https://gist.github.com/midacts/17125d9b0dd669d973218fddea1ebd60/revisions) `SchemaDataSource` removes these:

- `Optional: true`
- `ValidateFunc: Validate` 
  - I see that this calls `Validate` to validate the tags

and adds:

```text
Elem: &schema.Schema{
	Type: schema.TypeString,
},
```

I am not sure what the `Elem` part does.

azurerm/data_source_public_ip_prefix.go
midacts: It looks like the difference is `SchemaSingleZone` has:
- `ForceNew: true`
- `MaxItems: 1`

And `SchemaZonesComputed` has:
`Computed: true`

I'm not sure how to know when to use which one.
(Maybe since there is a `ForceNew`, it shouldn't be used for data sources).

azurerm/data_source_public_ip_prefix.go
midacts: Good idea.
Thanks for checking this.

azurerm/data_source_public_ip_prefix.go
midacts: Good idea.
Thanks for checking this.

azurerm/data_source_public_ip_prefix.go
midacts: Inline makes sense.

I was curious why we wouldn't add the [`name` and `resource group` lines](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_public_ip_prefix.go#L138-L139) here as well?

website/docs/d/public_ip_prefix.html.markdown
```diff
@@ -0,0 +1,39 @@
+---
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_public_ip_prefix"
+sidebar_current: "docs-azurerm-datasource-public-ip-prefix-x"
+description: |-
+  Gets information about an existing Public IP Prefix.
+
+---
+
+# Data Source: azurerm_public_ip_prefix
```
tombuildsstuff: could we add an entry for this new data source [to the sidebar](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/azurerm.erb#L57-L59)?

midacts:
Good catch. (Thanks for finding these).
Do you know if there is documentation stating what is required for new data sources and resources?
Things like - add to the website markdown, add to the sidebar, make your tests.

Or even tests to verify everything in the repo is compliant with these rules (so it wouldn't have to manually be checked).

tombuildsstuff:
@midacts we've got plans to add some specific documented in the near-future (it's being tracked in #2688) - however we've intentionally held off doing this until some code reorg. is done as a part of 2.0 (which we've started working on at the moment) - however once the refactoring is done (and we're able to link to the latest plans/practices) this is something we're definitely interested in adding :)

There is a linter for the other bits - however the Azure Provider diverges from other providers (as an example) as we need to nil-check every field in the response (since Azure may return the representation of a resource at the last time it was modified; rather than how the SDK/API Definition states that it should) - so we need to integrate some Azure specific rules into [the Terraform Provider linter](https://github.com/bflad/tfproviderlint) the broader Terraform team is working on.

tombuildsstuff:
Test passes:

Screenshot 2019-09-22 at 21 49 45

midacts:
Glad to be able to contribute and learn more about golang, Terraform and the azurerm provider for Terraform.

Thanks for your help.

PR #4345, opened 2019-09-17
Update Azure SDK to v33.2.0

r0bnet:
Would like to update to newer SDK version for at least two reasons:

- SQL API Version 2018-06-01-preview
- Kusto API Verison 2019-05-15

Had to change only one part in the kusto_cluster resource. Basic acc test didn't have a problem though.


r0bnet:
There is already a PR for that, didn't see it :/

PR #4367, opened 2019-09-18
New Resource: `azurerm_bot_channel_slack`

mbfrahry:
```
        --- PASS: TestAccAzureRMBotChannelsRegistration/connection/basic (89.74s)
        --- PASS: TestAccAzureRMBotChannelsRegistration/connection/update (98.56s)
```

### DIFF
--- a/azurerm/provider.go
+++ b/azurerm/provider.go
@@ -195,6 +195,7 @@ func Provider() terraform.ResourceProvider {
 		"azurerm_batch_account":                                      resourceArmBatchAccount(),
 		"azurerm_batch_application":                                  resourceArmBatchApplication(),
 		"azurerm_batch_certificate":                                  resourceArmBatchCertificate(),
+		"azurerm_bot_channel_slack":                                  resourceArmBotChannelSlack(),
 		"azurerm_bot_channels_registration":                          resourceArmBotChannelsRegistration(),
 		"azurerm_bot_connection":                                     resourceArmBotConnection(),
 		"azurerm_batch_pool":                                         resourceArmBatchPool(),
--- a/azurerm/resource_arm_bot_channel_slack.go
+++ b/azurerm/resource_arm_bot_channel_slack.go
@@ -1,0 +1,214 @@
+package azurerm
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/Azure/azure-sdk-for-go/services/preview/botservice/mgmt/2018-07-12/botservice"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/response"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func resourceArmBotChannelSlack() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceArmBotChannelSlackCreate,
+		Read:   resourceArmBotChannelSlackRead,
+		Delete: resourceArmBotChannelSlackDelete,
+		Update: resourceArmBotChannelSlackUpdate,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"location": azure.SchemaLocation(),
+
+			"bot_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+
+			"client_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+			"client_secret": {
+				Type:         schema.TypeString,
+				Required:     true,
+				Sensitive:    true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+			"verification_token": {
+				Type:         schema.TypeString,
+				Required:     true,
+				Sensitive:    true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+			"landing_page_url": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+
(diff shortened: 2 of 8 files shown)


azurerm/provider.go
```diff
@@ -195,6 +195,7 @@ func Provider() terraform.ResourceProvider {
 		"azurerm_batch_account":                                      resourceArmBatchAccount(),
 		"azurerm_batch_application":                                  resourceArmBatchApplication(),
 		"azurerm_batch_certificate":                                  resourceArmBatchCertificate(),
+		"azurerm_bot_channel_slack":                                  resourceArmBotChannelSlack(),
 		"azurerm_bot_channels_registration":                          resourceArmBotChannelsRegistration(),
```
tombuildsstuff: should these both be `channel_` or `channels_`?

azurerm/resource_arm_bot_channel_slack.go
```diff
@@ -0,0 +1,219 @@
...
+				ClientID:                utils.String(d.Get("client_id").(string)),
+				ClientSecret:            utils.String(d.Get("client_secret").(string)),
+				VerificationToken:       utils.String(d.Get("verification_token").(string)),
+				LandingPageURL:          utils.String(d.Get("landing_page_url").(string)),
+				IsEnabled:               utils.Bool(true),
+				RegisterBeforeOAuthFlow: utils.Bool(true),
+			},
+			ChannelName: botservice.ChannelNameSlackChannel1,
+		},
+		Location: utils.String(d.Get("location").(string)),
```
tombuildsstuff: we need to normalize this via:

```suggestion
		Location: utils.String(azure.NormalizeLocation(d.Get("location").(string))),
```

azurerm/resource_arm_bot_channel_slack.go
```diff
@@ -0,0 +1,219 @@
...
+		Kind:     botservice.KindBot,
+	}
+
+	if _, err := client.Create(ctx, resourceGroup, botName, botservice.ChannelNameSlackChannel, channel); err != nil {
+		return fmt.Errorf("Error issuing create request for Bot Channel Slack (Resource Group %q / Bot %q): %+v", resourceGroup, botName, err)
+	}
+
+	resp, err := client.Get(ctx, resourceGroup, botName, string(botservice.ChannelNameSlackChannel))
```
tombuildsstuff: do we need to cast this to a string for the Create too?

azurerm/resource_arm_bot_channel_slack.go
```diff
@@ -0,0 +1,219 @@
...
+	id, err := azure.ParseAzureResourceID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	botName := id.Path["botServices"]
+	resp, err := client.Get(ctx, id.ResourceGroup, botName, string(botservice.ChannelNameSlackChannel))
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[INFO] Error reading Bot Channel Slack (Resource Group %q / Bot %q): %+v", id.ResourceGroup, botName, err)
```
tombuildsstuff:
```suggestion
			log.Printf("[INFO] Bot Channel Slack (Resource Group %q / Bot %q) was not found - removing from state!", id.ResourceGroup, botName)
```

azurerm/resource_arm_bot_channel_slack.go
```diff
@@ -0,0 +1,219 @@
...
+			d.SetId("")
+			return nil
+		}
+
+		return fmt.Errorf("Error reading Bot Channel Slack (Resource Group %q / Bot %q): %+v", id.ResourceGroup, botName, err)
```
tombuildsstuff: can we match the ordering with other resources:

```suggestion
		return fmt.Errorf("Error reading Channel Slack for Bot %q (Resource Group %q): %+v", botName, id.ResourceGroup, err)
```

azurerm/resource_arm_bot_channel_slack.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	d.Set("resource_group_name", id.ResourceGroup)
+	d.Set("location", resp.Location)
+	d.Set("bot_name", botName)
+
+	if props := resp.Properties; props != nil {
+		if channel, ok := props.AsSlackChannel(); ok {
+			if channel != nil {
```
tombuildsstuff: I think we can remove this line, since otherwise `ok` would return false?

azurerm/resource_arm_bot_channel_slack.go
```diff
@@ -0,0 +1,219 @@
...
+		},
+		Location: utils.String(d.Get("location").(string)),
+		Kind:     botservice.KindBot,
+	}
+
+	if _, err := client.Update(ctx, resourceGroup, botName, botservice.ChannelNameSlackChannel, channel); err != nil {
+		return fmt.Errorf("Error issuing create request for Bot Channel Slack (Resource Group %q / Bot %q): %+v", resourceGroup, botName, err)
+	}
+
+	resp, err := client.Get(ctx, resourceGroup, botName, string(botservice.ChannelNameSlackChannel))
```
tombuildsstuff: (same here) do we need to cast this to a string in all the calls? it's odd it's inconsistent

website/docs/r/bot_channel_slack.markdown
```diff
@@ -0,0 +1,74 @@
+---
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_bot_channel_slack"
+sidebar_current: "docs-azurerm-resource-bot-channel-slack"
+description: |-
+  Manages a Bot Channel for Slack.
```
tombuildsstuff: does this want to be:

```suggestion
  Manages a Slack integration for a Bot Channel
```

website/docs/r/bot_channel_slack.markdown
```diff
@@ -0,0 +1,74 @@
...
+---
+
+# azurerm_bot_connection
+
+Manages a Bot Channel for Slack.
```
tombuildsstuff: (same here)

website/docs/r/bot_channel_slack.markdown
```diff
@@ -0,0 +1,74 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The Bot Channel ID.
+
+## Import
+
+Bot Channel can be imported using the `resource id`, e.g.
```
tombuildsstuff: does this want to be:

```suggestion
The Slack Channel for a Bot can be imported using the `resource id`, e.g.
```

azurerm/provider.go
mbfrahry: I don't think so. `channels_registration` houses every `channel`

azurerm/resource_arm_bot_channel_slack.go
mbfrahry: No, the SDK is weird in that create/update takes the `ChannelName` constant while the get/delete take strings instead. It's weird

azurerm/resource_arm_bot_channel_slack.go
mbfrahry: Only because the SDK is inconsistent

PR #4371, opened 2019-09-19
r/kusto_cluster: switching to use the constants from the sdk

tombuildsstuff:
Since these are now available in the SDK we can use the constants rather than hard-coding these


tombuildsstuff:
Closing in favour of #4376 which updates the SDK too

PR #4385, opened 2019-09-19
New Resource: azurerm_kusto_eventhub_data_connection

jrauschenbusch:
As discussed [here](https://github.com/terraform-providers/terraform-provider-azurerm/issues/3856) with @r0bnet an initial version for EventHub data connection resource.


website/docs/r/kusto_eventhub_data_connection.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+
+* `eventhub_id` - (Required) Specifies the resource id of the EventHub this data connection will use for ingestion. Changing this forces a new resource to be created.
+
+* `consumer_group` - (Required) Specifies the EventHub consumer group this data connection will use for ingestion. Changing this forces a new resource to be created.
+
+* `table_name` - (Optional) Specifies the target table name used for the message ingestion. Table must exist before resource is created.
+
+* `mapping_rule_name` - (Optional) Specifies the mapping rule used for the message ingestion. Mapping rule must exist before resource is created.
+
+* `data_format` - (Optional) Specifies the data format of the EventHub messages. Allowed values: AVRO, CSV, JSON, MULTIJSON, PSV, RAW, SCSV, SINGLEJSON, SOHSV, TSV, TXT
```
tombuildsstuff: can we quote these values:

```suggestion
* `data_format` - (Optional) Specifies the data format of the EventHub messages. Allowed values: `AVRO`, `CSV`, `JSON`, `MULTIJSON`, `PSV`, `RAW`, `SCSV`, `SINGLEJSON`, `SOHSV`, `TSV` and `TXT`
```

azurerm/resource_arm_kusto_eventhub_data_connection.go
```diff
@@ -0,0 +1,300 @@
...
+			"mapping_rule_name": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validateAzureRMKustoEntityName,
+			},
+
+			"data_format": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validateAzureRMKustoDataConnectionDataFormat,
```
tombuildsstuff: rather than using the `All` method could we use:

```
  string(kusto.Thing),
}, false),
```

whilst the "All" methods are convenient it's possible values will be missed in the documentation if they're added to the SDK - so until we can generate those could we hard-code this for the moment?

website/docs/r/kusto_eventhub_data_connection.html.markdown
jrauschenbusch: np. will do that 👍

azurerm/resource_arm_kusto_eventhub_data_connection.go
jrauschenbusch: done

PR #4389, opened 2019-09-19
New Resource: `azurerm_bot_channel_email`

mbfrahry:


azurerm/resource_arm_bot_channel_email.go
```diff
@@ -0,0 +1,198 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"location": azure.SchemaLocation(),
+
+			"bot_name": {
+				Type:         schema.TypeString,
+				Required:     true,
```
tombuildsstuff: looks like this needs to be ForceNew?

azurerm/resource_arm_bot_channel_email.go
```diff
@@ -0,0 +1,198 @@
...
+
+	return nil
+}
+
+func resourceArmBotChannelEmailUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).bot.ChannelClient
+	ctx := meta.(*ArmClient).StopContext
+
+	botName := d.Get("bot_name").(string)
+	resourceGroup := d.Get("resource_group_name").(string)
```
tombuildsstuff: given this is the Update we should have the ID here - as such can we parse these from the ID?

azurerm/resource_arm_bot_channel_email_test.go
```diff
@@ -0,0 +1,170 @@
...
+		if resp.StatusCode != http.StatusNotFound {
+			return fmt.Errorf("Bot Channel Email still exists:\n%#v", resp.Properties)
+		}
+	}
+
+	return nil
+}
+
+func testAccAzureRMBotChannelEmail_basicConfig(rInt int, location string) string {
+	template := testAccAzureRMBotChannelsRegistration_basicConfig(rInt, location)
```
tombuildsstuff: this test fails because:

>         Error: Error issuing create request for Bot Channels Registration "acctestdf190926083626598033" (Resource Group "acctestRG-190926083626598033"): botservice.BotsClient#Create: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidBotData" Message="Bot is not valid. Errors: The Microsoft App ID is already registered to another bot application..  See https://aka.ms/bot-requirements for detailed requirements."

maybe we need to generate a new UUID for each bot application?

PR #4404, opened 2019-09-21
Adding forwarding_configuration to front door

montross50:
The docs did not include any information about this block. https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_front_door.go#L159 

This should serve at least as a rough draft to make the docs more complete for this resource.

### DIFF
--- a/website/docs/r/front_door.html.markdown
+++ b/website/docs/r/front_door.html.markdown
@@ -174,6 +174,22 @@ The `routing_rule` block supports the following:
 
 * `enabled` - (Optional) `Enable` or `Disable` use of this Backend Routing Rule. Permitted values are `true` or `false`. Defaults to `true`.
 
+* `forwarding_configuration` - (Optional) A `forwarding_configuration` block as defined below.
+
+---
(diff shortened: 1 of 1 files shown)


jrestall:
@montross50 @katbyte It would be useful to provide some guidance/docs on disabling caching as I'm currently researching how to do that with the front door provider.

It looks like based on the below, not providing a forwarding_configuration will automatically disable caching. Might be useful to confirm and document that.

https://github.com/Azure/azure-sdk-for-go/blob/master/services/preview/frontdoor/mgmt/2019-04-01/frontdoor/models.go#L927

It could be considered an issue that you can't disable caching and also use the forwarding_configuration if this is right though, would be better if cachingConfiguration wasn't set [here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/1f32f7a1e897c9332ea9d6480ffcb299e906e9eb/azurerm/resource_arm_front_door.go#L1063) if those two cache settings weren't set.

PR #4409, opened 2019-09-23
azurerm_eventhub_namespace: support for the `network_rulesets` property

katbyte:


azurerm/resource_arm_eventhub_namespace.go
```diff
@@ -283,3 +366,96 @@ func eventHubNamespaceStateStatusCodeRefreshFunc(ctx context.Context, client *ev
...
+				rules = append(rules, eventhub.NWRuleSetIPRules{
+					IPMask: utils.String(rblock["ip_mask"].(string)),
+					Action: eventhub.NetworkRuleIPAction(rblock["action"].(string)),
+				})
+			}
+
+			ruleset.IPRules = &rules
+		}
+	}
+
```
mbfrahry: Do we want to do extra validation to ensure that one of `ip_rule` or `virtual_network_rule` are specified?

katbyte: It works just fine with neither 🙂

PR #4432, opened 2019-09-25
Add zone redundancy for service bus namespace

hbuckle:
Note that the preview API doesn't expose all of the same clients and operations as stable so I created a separate client for the preview API.


tombuildsstuff:
Tests pass:

Screenshot 2019-09-26 at 10 46 44

PR #4433, opened 2019-09-25
Updated policy section to clarify usage

jeanpaulsmit:
In the previous version there was a reference to the Policy block, but nowhere was to be found how to add this to the API Management resource itself. The update shows an example so users don't need to go through the code to find out how to use it.

### DIFF
--- a/website/docs/r/api_management.html.markdown
+++ b/website/docs/r/api_management.html.markdown
@@ -26,6 +26,17 @@ resource "azurerm_api_management" "test" {
   publisher_email     = "company@terraform.io"
 
   sku_name = "Developer_1"
+   
+  policy {
+    xml_content = <<XML
+    <policies>
+      <inbound />
+      <backend />
+      <outbound />
+      <on-error />
+    </policies>
+XML
+  }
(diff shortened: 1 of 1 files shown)


jeanpaulsmit:
Hi @katbyte, thanks for getting back on this. Either is fine by me, I only want to save others from searching how to put a certain block in the TF template. I can imagine the example to be part of the topic where it's described, but I also think it would be helpful to have some kind of generic example on how the optional sections are meant to be defined.

mbfrahry:
Hey @jeanpaulsmit, let's go ahead and make this another full example similar to how we do it in [virtual_network_peering](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/docs/r/virtual_network_peering.html.markdown#example-usage-global-virtual-network-peering) and remove it from the attribute description. We just try to promote consistency in our docs wherever we can and seeing an example at the top of the page should be enough. Another next step would be to also add an example in the [examples](https://github.com/terraform-providers/terraform-provider-azurerm/tree/master/examples) folder.

PR #4439, opened 2019-09-26
Removed ForceNew for resource_arm_function_app for the app_service_plan_id

MarkDordoy:
A function app should not have to be re-created when you re-associate it to another service plan. This PR will bring the `resource_arm_function_app` in line with its closely related `resource_arm_app_service` resource

### DIFF
--- a/azurerm/resource_arm_function_app.go
+++ b/azurerm/resource_arm_function_app.go
@@ -49,7 +49,6 @@ func resourceArmFunctionApp() *schema.Resource {
 			"app_service_plan_id": {
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: true,
 			},
 
 			"enabled": {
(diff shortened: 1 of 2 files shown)


azurerm/resource_arm_function_app.go
```diff
@@ -49,7 +49,6 @@ func resourceArmFunctionApp() *schema.Resource {
 			"app_service_plan_id": {
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: true,
```
tombuildsstuff: since this no longer forces a new resource, could we also [update the documentation](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/function_app.html.markdown#L96)?

MarkDordoy: @tombuildsstuff my bad. PR now includes the update to documentation. 

Thanks for spotting that.

PR #4454, opened 2019-09-29
New Resource/Data Source: `azurerm_App_Service_Certificate_Order` 

hund030:
This PR is code implement for issue https://github.com/terraform-providers/terraform-provider-azurerm/issues/2038


azurerm/resource_arm_app_service_certificate_order.go
```diff
@@ -0,0 +1,488 @@
...
+			},
+
+			"validity_in_years": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      1,
+				ValidateFunc: validation.IntBetween(1, 3),
+			},
+
+			"domain_verification_token": {
```
mbfrahry: Terraform doesn't need to know about every attribute an API returns but instead just attributes that directly affect the resource or could be used in another resource. With that in mind, I don't think any of these `Computed` attributes need to be specified in the schema. I'm open to further discussion though if you believe otherwise.

azurerm/resource_arm_app_service_certificate_order.go
```diff
@@ -0,0 +1,488 @@
...
+	}
+
+	location := azure.NormalizeLocation(d.Get("location").(string))
+	t := d.Get("tags").(map[string]interface{})
+
+	// if v, ok := d.GetOk("certificates"); ok { }
+
+	properties := web.AppServiceCertificateOrderProperties{}
+
+	if v, ok := d.GetOk("product_type"); ok {
```
mbfrahry: We should put every attribute in the properties struct even if Terraform isn't holding it. This allows us to pass empty values to the api to remove certain attributes that aren't present.

azurerm/resource_arm_app_service_certificate_order.go
```diff
@@ -0,0 +1,488 @@
...
+			Type:     schema.TypeInt,
+			Computed: true,
+		},
+
+		"serial_number": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
+
+		"thumbprint": {
```
mbfrahry: Many of the attributes here can be removed but I think `thumbprint` is useful. Also, how many Certificates can this resource hold? If it's just one, we can just have a thumbprint attribute and remove this TypeList entirely

website/docs/d/app_service_certificate_order.html.markdown
```diff
@@ -0,0 +1,150 @@
...
+
+* `serial_number` - Current serial number of the certificate.
+
+* `last_certificate_issuance_time` - Certificate last issuance time.
+
+* `next_auto_renewal_time_stamp` - Time stamp when the certificate would be auto renewed next
+
+* `expiration_time` - Certificate expiration time.
+
+* `is_private_key_external` - true if private key is external; otherwise, false.
```
mbfrahry: We should clean up the language here
```suggestion
* `is_private_key_external` - Whether the private key is external or not. 
```

website/docs/d/app_service_certificate_order.html.markdown
```diff
@@ -0,0 +1,150 @@
...
+
+* `location` - The Azure location where the App Service exists.
+
+* `auto_renew` - true if the certificate should be automatically renewed when it expires; otherwise, false.
+
+* `certificates` - State of the Key Vault secret. A `certificates` block as defined below.
+
+* `csr` - Last CSR that was created for this order.
+
+* `distinguished_name` - Certificate distinguished name.
```
mbfrahry: As tedious as it it we should expand on the descriptions here
```suggestion
* `distinguished_name` - The Distinguished Name for the App Service Certificate Order
```

mbfrahry:
These tests also look to be quite expensive so we should a conditional at the top of each test that skips them if an environmental variable isn't set. The below should work for that
```
if os.Getenv("ARM_RUN_TEST_APP_SERVICE_CERTIFICATE") == "" {
    t.Skip("Skipping as ARM_RUN_TEST_APP_SERVICE_CERTIFICATE" is not specified")
}
```

azurerm/resource_arm_app_service_certificate_order.go
hund030: Remove the computed attributes seem useless.

azurerm/resource_arm_app_service_certificate_order.go
hund030: Remove the computed attributes seem useless.

azurerm/resource_arm_app_service_certificate_order.go
hund030: Get attributes directly and put them into the struct.

PR #4462, opened 2019-09-30
Move linux note

dfredell:
Move the note on Linus app service plan to be below the kind field.


dfredell:
@katbyte This line change is near to your last change, FYI.

PR #4249, opened 2019-09-05
Support app service HTTP blob logging

phekmat:
This PR is an extension of #3520 (and works toward #1082), adding blob support for the HTTP logs

Issue #1082, opened 2018-04-05
Feature request: Support log settings for app services and functions


phekmat:
Like #756 and #1026, it would be nice to support the log configuration for app_service/slot and function_app. 

See https://github.com/Azure/azure-sdk-for-go/blob/v15.0.0/services/web/mgmt/2016-09-01/web/models.go#L16146 and https://docs.microsoft.com/en-us/rest/api/appservice/webapps/updatediagnosticlogsconfig

dijitali:
If this is looked at, it would be great to support all App Service configuration options rather than just log settings.  

These are viewable in the Resource Explorer here:
https://resources.azure.com/subscriptions/<SUBSCRIPTION-ID>/resourceGroups/<RESOURCE-GROUP-NAME>/providers/Microsoft.Web/sites/<APP-NAME>/config

And documented here:
https://docs.microsoft.com/en-us/rest/api/appservice/webapps

vidarw:
@tombuildsstuff @katbyte  Any thoughs on how to name the properties for the best consistency?

![image](https://user-images.githubusercontent.com/1172172/48348025-5f11bb00-e680-11e8-8c32-f19c86f082c2.png)

tombuildsstuff:
@vidarw presumably these'd make the most sense as nested blocks/objects (such that it matches the API) within the App Service / App Service Slot resource, but it's hard to say if they'd be better as an independent (split out) resource until it's prototyped. The other question I think is relevant here is how we'd reset these back to their default values (since there's no "delete/reset" API endpoint/method); whilst it's possible to determine what the default values are (and to submit those) - there's no guarantee those won't change over time.

Hope that helps :)

vidarw:
When you mention default values, you refer to the scenario where first apply values and then later removes the config section from your terraform files? Is the expected behavior to return these settings to their original value?

Shouldn't it be possible to test if MS changes the default by creating a new blank resource without the log section and compare settings with the maintained list?

Something like this?
```
logs {
  application_file_system_log_level =  "On"/"Off" (or should we use true|false)
  application_table_storage_sas_token = "https://..." (document the ability to use the SAS data source)
  application_table_storage_log_level = "Off"
  http_file_system_retention_size = "15"
  http_file_system_retention_days = "30"
}
```
Other suggested names for the section _diagnostic_logs_ (after the UI) _log_settings_/_log_config_ (similar names to the other terraform sections).

tombuildsstuff:
[...]

PR #4473, opened 2019-10-01
Support log settings for app service slot

phekmat:
This PR is an extension of #4249 & #3520 (and works toward #1082), adding diagnostic and HTTP log settings support for the slot


tombuildsstuff:
Tests pass:

Screenshot 2019-10-02 at 07 58 12

Issue #2807, opened 2019-01-30
Announcement: Upcoming Changes in Version 2.0 of the Azure Provider


tombuildsstuff:
Terraform initially shipped support for the AzureRM Provider back in December 2015. Since then we've added support for 191 Resources, 58 Data Sources and have launched a couple of related Providers in the form of [the Azure Active Directory Provider](https://www.terraform.io/docs/providers/azuread/index.html) and [the Azure Stack Provider](https://www.terraform.io/docs/providers/azurestack/index.html).

Version 2.0 of the AzureRM Provider will be a **Major Release** - in that it will include some larger-scale changes not seen in a regular release. A summary of these changes is outlined below - however a full breakdown will be available on the Terraform Website after the release of v1.22.

## Summary

* Existing Resources will be required to be imported
* Custom Timeouts will be available on Resources - this will allow you to specify a custom timeout for provisioning the resource in your Terraform Configuration [using the `timeouts` block](https://www.terraform.io/docs/configuration/resources.html#timeouts).
* New resources for Virtual Machines and Virtual Machine Scale Sets
* Removing Fields, Data Sources and Resources which have been deprecated

A brief summary of each item can be found below - more details will be available in the Azure Provider 2.0 upgrade guide on the Terraform Website once v1.22 has been released.

---

### Existing Resources will be required to be Imported

Terraform allows for existing resources which have been created outside of Terraform to be Imported into Terraform's State. Once a resource is imported into the state, it's possible for Terraform to track changes and manage this resource. The Azure Provider allows Importing existing resources into the state (using `terraform import`) for (almost) every resource.

Version 2.0 of the Azure Provider aims to solve an issue where it's possible to unintentionally import resources into the state by running `terraform apply`. To explain this further, the majority of Azure's API's are Upserts - which means that a resource will be updated if it exists, otherwise it'll be created.

Where the unique identifier for (most) Azure resources is the `name` (rather than for example an `aws_instance` where AWS will generate a different unique identifier) - it's possible that users may have unintentionally imported resources into Terraform when running `terraform apply` on an existing resource.

Whilst this may allow resources to work in some cases, it leads to hard-to-diagnose bugs in others (which could have been caught during `terraform plan`).

In order to match the behaviour of other Terraform Providers version 2.0 of the AzureRM Provider will require that existing resources are imported into the state prior to use. This means that Terraform will be checking for the presence of an existing resource prior to creating it - and will return an error similar to below:

```
[...]

Issue #1747, opened 2018-08-09
Managed SQL database instance


srusru:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

---

It would be nice if we have a terraform feature for Azure SQL Managed database instance.

rbankole:
i second this! :)

Chris-Gray94:
I think this needs to be done now that it is in GA.
https://azure.microsoft.com/en-us/blog/azure-sql-database-managed-instance-general-purpose-tier-general-availability/

tombuildsstuff:
@jeffreyCline have you started looking into this yet? If not I don't think anybody's working on this from HashiCorp/Microsoft's side, so you should be good @mb290 👍 (thanks in advance 😄)

mb290:
@tombuildsstuff maybe I've bitten off much more than I can chew! The RestAPI documentation is available here and relatively straight forward, however i'm really struggling to find a concise list of the options for SKUs: [https://docs.microsoft.com/en-us/rest/api/sql/managedinstances/createorupdate](https://docs.microsoft.com/en-us/rest/api/sql/managedinstances/createorupdate)

There are also lots of dependencies on the VNet that the subnet is in, so i'm trying to figure out the best way to either a) document or b) template this. [https://docs.microsoft.com/en-us/azure/sql-database/sql-database-managed-instance-vnet-configuration#requirements](https://docs.microsoft.com/en-us/azure/sql-database/sql-database-managed-instance-vnet-configuration#requirements)

The final challenge is for the first SQL MI in a Subnet MS take up to 6 hours to provision so TF is going to require a very large timeout.

theharleyquin:
Bump - just checking in to see if there is any movement on this.

tombuildsstuff:
@theharleyquin not at this time - as @mb290 has mentioned the first Managed Instance in a Subnet can take around 6 hours to provision (up to a maximum of 30h) - which makes developing support for this fairly challenging, unfortunately.

thcp:
It seems that no news from Hashicorp on this one.

ganeshmedhekar:
Hi,
[...]

Issue #171, opened 2017-07-13
Resource Timeouts for creation / deletion


mooperd:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Is there a way of configuring timeouts?

```
azurerm_subnet.subnet: Creating...
  address_prefix:            "" => "10.0.2.0/24"
  ip_configurations.#:       "" => "<computed>"
  name:                      "" => "bach"
  network_security_group_id: "" => "<computed>"
  resource_group_name:       "" => "bach"
  route_table_id:            "" => "<computed>"
  virtual_network_name:      "" => "bach"
azurerm_subnet.subnet: Still creating... (10s elapsed)
azurerm_subnet.subnet: Still creating... (20s elapsed)
azurerm_subnet.subnet: Still creating... (30s elapsed)
azurerm_subnet.subnet: Still creating... (40s elapsed)
azurerm_subnet.subnet: Still creating... (50s elapsed)
azurerm_subnet.subnet: Still creating... (1m0s elapsed)
azurerm_subnet.subnet: Still creating... (1m10s elapsed)
azurerm_subnet.subnet: Still creating... (1m20s elapsed)
azurerm_subnet.subnet: Still creating... (1m30s elapsed)
azurerm_subnet.subnet: Still creating... (1m40s elapsed)
azurerm_subnet.subnet: Still creating... (1m50s elapsed)
azurerm_subnet.subnet: Still creating... (2m0s elapsed)
azurerm_subnet.subnet: Still creating... (2m10s elapsed)
azurerm_subnet.subnet: Still creating... (2m20s elapsed)
azurerm_subnet.subnet: Still creating... (2m30s elapsed)
azurerm_subnet.subnet: Still creating... (2m40s elapsed)
azurerm_subnet.subnet: Still creating... (2m50s elapsed)
azurerm_subnet.subnet: Still creating... (3m0s elapsed)
azurerm_subnet.subnet: Still creating... (3m10s elapsed)
azurerm_subnet.subnet: Still creating... (3m20s elapsed)
azurerm_subnet.subnet: Still creating... (3m30s elapsed)
[...]

PR #4475, opened 2019-10-01
Groundwork for Custom Timeouts

tombuildsstuff:
This PR contains the necessary groundwork to support Custom Timeouts on both Resources and Data Sources

Whilst this functionality is controlled by a feature toggle (which is intentionally disabled by default) - in it's current form this functionality has no effect on existing resources, therefore it's intentionally not documented or recommended for use at this time.

When the context changes have been threaded through all of the resources, in a future 1.x release we'll document this behaviour so that users can opt into this - however at this time this functionality is unsupported.

Given this is feature toggled we should be able to add timeouts gradually to resources, internally if the feature toggle is disabled any timeouts defined on resources will be removed, leading to the same behaviour as today - for example:

Given the following configuration:

```
$ cat main.tf
  name = "tom-dev6666"
  location = "West Europe"

  timeouts {
    create = "2s"
    delete = "4s"
  }
}
```

Without the feature enabled:

```
$ terraform apply

Error: Unsupported block type

  on main.tf line 5, in resource "azurerm_resource_group" "test":
   5:   timeouts {

Blocks of type "timeouts" are not expected here.
```

With the feature enabled:

```
$ terraform apply

  + create

  # azurerm_resource_group.test will be created
  + resource "azurerm_resource_group" "test" {
      + id       = (known after apply)
      + location = "westeurope"
      + name     = "tom-dev6666"
      + tags     = (known after apply)

      + timeouts {
          + create = "2s"
          + delete = "4s"
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

azurerm_resource_group.test: Creating...

Error: Error creating resource group: resources.GroupsClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: context deadline exceeded
```

(this obviously works with a more sane timeout configured, 2s/4s are an example)

The (acceptance) test suite itself intentionally _doesn't_ configure any timeouts in the Check Func's since we should continue polling until we have a conclusion

Longer term this unblocks #171 (and related issues like #1747) which forms a part of #2807

### DIFF
--- a/azurerm/config.go
+++ b/azurerm/config.go
@@ -197,6 +197,8 @@ func getArmClient(authConfig *authentication.Config, skipProviderRegistration bo
 	// Key Vault Endpoints
 	keyVaultAuth := authConfig.BearerAuthorizerCallback(sender, oauthConfig)
 
+	timeout := 180 * time.Minute
+
 	o := &common.ClientOptions{
 		SubscriptionId:              authConfig.SubscriptionID,
 		TenantID:                    authConfig.TenantID,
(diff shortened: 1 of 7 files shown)


tombuildsstuff:
Looks like some issues between Travis <-> Github - it's green in Travis:

Screenshot 2019-10-02 at 07 55 03

PR #4476, opened 2019-10-01
Fixes for load balancer and virtual machine scale sets related documentation

arestarh:
…ed documentation.

### DIFF
--- a/website/docs/r/loadbalancer_nat_pool.html.markdown
+++ b/website/docs/r/loadbalancer_nat_pool.html.markdown
@@ -8,7 +8,9 @@ description: |-
 
 # azurerm_lb_nat_pool
 
-Manages a Load Balancer NAT pool. This resource is intended to use with virtual machine scale sets.
+Manages a Load Balancer NAT pool. 
+
+-> **NOTE:** This resource cannot be used with with virtual machines, instead use the `azurerm_lb_nat_rule` resource.
 
 ~> **NOTE** When using this resource, the Load Balancer needs to have a FrontEnd IP Configuration Attached
 
--- a/website/docs/r/loadbalancer_nat_pool.html.markdown
+++ b/website/docs/r/loadbalancer_nat_pool.html.markdown
@@ -8,7 +8,7 @@ description: |-
 
 # azurerm_lb_nat_pool
 
-Manages a Load Balancer NAT pool.
+Manages a Load Balancer NAT pool. This resource is intended to use with virtual machine scale sets.
 
 ~> **NOTE** When using this resource, the Load Balancer needs to have a FrontEnd IP Configuration Attached
 
@@ -67,7 +67,7 @@ The following arguments are supported:
 
 The following attributes are exported:
 
-* `id` - The ID of the Load Balancer to which the resource is attached.
+* `id` - The ID of the Load Balancer NAT pool.
 
 ## Import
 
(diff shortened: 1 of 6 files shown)


website/docs/r/loadbalancer_nat_pool.html.markdown
```diff
@@ -8,7 +8,7 @@ description: |-
 
 # azurerm_lb_nat_pool
 
-Manages a Load Balancer NAT pool.
+Manages a Load Balancer NAT pool. This resource is intended to use with virtual machine scale sets.
```
katbyte: This can be used with VMs as well so can we revert this?
```suggestion
Manages a Load Balancer NAT pool.
```

website/docs/r/loadbalancer_nat_rule.html.markdown
```diff
@@ -8,7 +8,7 @@ description: |-
 
 # azurerm_lb_nat_rule
 
-Manages a Load Balancer NAT Rule.
+Manages a Load Balancer NAT Rule. This resource is intended to use with network interfaces of virtual machines.
```
katbyte: As above could we revert this
```suggestion
Manages a Load Balancer NAT Rule.
```

website/docs/r/loadbalancer_nat_pool.html.markdown
arestarh: Thanks for reviewing.
I might be wrong, but I've made those note(s) because in _ARM Template reference_ (**API version 2019-04-01**) and _REST API reference_ (**API version 2019-07-01**) regarding creating\updating Load Balancer resource in Azure there are some clarifications about using\referencing NAT Pools\NAT Rules:
https://docs.microsoft.com/en-us/azure/templates/microsoft.network/2019-04-01/loadbalancers#LoadBalancerPropertiesFormat
https://docs.microsoft.com/en-us/rest/api/load-balancer/loadbalancers/createorupdate#request-body

website/docs/r/loadbalancer_nat_rule.html.markdown
arestarh: Thanks for reviewing.
I might be wrong, but I've made those note(s) because in _ARM Template reference_ (**API version 2019-04-01**) and _REST API reference_ (**API version 2019-07-01**) regarding creating\updating Load Balancer resource in Azure there are some clarifications about using\referencing NAT Pools\NAT Rules:
https://docs.microsoft.com/en-us/azure/templates/microsoft.network/2019-04-01/loadbalancers#LoadBalancerPropertiesFormat
https://docs.microsoft.com/en-us/rest/api/load-balancer/loadbalancers/createorupdate#request-body

website/docs/r/loadbalancer_nat_pool.html.markdown
katbyte: Talking internally it sounds like while this cannot be used with a VM, it can be used with other azure resource. So instead could we perhaps add a note instead?
```suggestion
Manages a Load Balancer NAT pool. 

-> **NOTE:** This resource cannot be used with with virtual machines, instead use the `azurerm_lb_nat_rule` resource.
```

website/docs/r/loadbalancer_nat_rule.html.markdown
katbyte: as above,
```suggestion
Manages a Load Balancer NAT Rule. 

-> **NOTE:** This resource cannot be used with with virtual machine scale sets, instead use the `azurerm_lb_nat_pool` resource.
```

arestarh:
@katbyte Fixed. Thanks for useful suggestions.

PR #4498, opened 2019-10-03
azurerm_kubernetes_cluster  - add support for the `azure_policy` property in the `addon_profile` block

grahamhayes:
Signed-off-by: Graham Hayes <gr@ham.ie>

### DIFF
--- a/azurerm/data_source_kubernetes_cluster.go
+++ b/azurerm/data_source_kubernetes_cluster.go
@@ -77,6 +77,19 @@ func dataSourceArmKubernetesCluster() *schema.Resource {
 								},
 							},
 						},
+
+						"azure_policy": {
+							Type:     schema.TypeList,
+							Computed: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:     schema.TypeBool,
+										Computed: true,
+									},
+								},
+							},
+						},
 					},
 				},
 			},
@@ -621,6 +634,20 @@ func flattenKubernetesClusterDataSourceAddonProfiles(profile map[string]*contain
 		kubeDashboards = append(kubeDashboards, output)
 	}
 	values["kube_dashboard"] = kubeDashboards
+
+	azurePolicies := make([]interface{}, 0)
+	if azurePolicy := profile["azurepolicy"]; azurePolicy != nil {
+		enabled := false
+		if enabledVal := azurePolicy.Enabled; enabledVal != nil {
+			enabled = *enabledVal
+		}
+
+		output := map[string]interface{}{
+			"enabled": enabled,
+		}
+		azurePolicies = append(azurePolicies, output)
+	}
+	values["azure_policy"] = azurePolicies
 
 	return []interface{}{values}
 }
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -315,6 +315,20 @@ func resourceArmKubernetesCluster() *schema.Resource {
 								},
 							},
 						},
+
+						"azure_policy": {
+							Type:     schema.TypeList,
+							MaxItems: 1,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:     schema.TypeBool,
+										Required: true,
+									},
+								},
+							},
+						},
 					},
 				},
 			},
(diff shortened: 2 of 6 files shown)


azurerm/data_source_kubernetes_cluster.go
```diff
@@ -77,6 +77,20 @@ func dataSourceArmKubernetesCluster() *schema.Resource {
 								},
 							},
 						},
+
+						"azure_policy": {
+							Type:     schema.TypeList,
+							MaxItems: 1,
+							Optional: true,
```
mbfrahry: Attributes in DataSources need to be `Computed: true`.  `MaxItems: 1` Can also be removed as it's only used when Creating/Updating a resource.

azurerm/data_source_kubernetes_cluster.go
```diff
@@ -77,6 +77,20 @@ func dataSourceArmKubernetesCluster() *schema.Resource {
...
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:     schema.TypeBool,
+										Required: true,
```
mbfrahry: Also, this should be `Computed: true` instead
```suggestion
										Computed: true,
```

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -1043,6 +1057,17 @@ func expandKubernetesClusterAddonProfiles(d *schema.ResourceData) map[string]*co
 		}
 	}
 
+	azurePolicy := profile["azure_policy"].([]interface{})
+	if len(azurePolicy) > 0 {
+		value := azurePolicy[0].(map[string]interface{})
```
mbfrahry: `azurePolicy[0]` could be nil so we should do a nil check before casting it to `map[string]interface{}`

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -1043,6 +1057,17 @@ func expandKubernetesClusterAddonProfiles(d *schema.ResourceData) map[string]*co
...
+		enabled := value["enabled"].(bool)
+
+		addonProfiles["azurepolicy"] = &containerservice.ManagedClusterAddonProfile{
+			Enabled: utils.Bool(enabled),
+			Config:  nil,
```
mbfrahry: Is Config something we'll want to implement in the future? If so, can you add something like
```suggestion
                         // TODO add support for AddonProfile Config
			Config:  nil,
```

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -144,6 +144,7 @@ A `addon_profile` block supports the following:
 * `http_application_routing` - (Optional) A `http_application_routing` block.
 * `oms_agent` - (Optional) A `oms_agent` block. For more details, please visit [How to onboard Azure Monitor for containers](https://docs.microsoft.com/en-us/azure/monitoring/monitoring-container-insights-onboard).
 * `kube_dashboard` - (Optional) A `kube_dashboard` block.
+* `azure_policy` - (Optional) A `azure_policy` block.  For more details please visit [Understand Azure Policy for Azure Kubernetes Service](https://docs.microsoft.com/en-ie/azure/governance/policy/concepts/rego-for-aks)
```
mbfrahry: We should add a note as well that this attribute is in preview with information for how to opt-in
```suggestion
* `azure_policy` - (Optional) A `azure_policy` block.  For more details please visit [Understand Azure Policy for Azure Kubernetes Service](https://docs.microsoft.com/en-ie/azure/governance/policy/concepts/rego-for-aks)

-> **Note**: Azure Policy for Azure Kubernetes Service is currently in preview and not available to subscriptions that have not [opted-in](https://docs.microsoft.com/en-us/azure/governance/policy/concepts/rego-for-aks?toc=/azure/aks/toc.json) to join `Azure Policy` preview. 
```

azurerm/resource_arm_kubernetes_cluster.go
grahamhayes: sure - I kept it in line with the other policy addons, but I will add a nil check now.

azurerm/resource_arm_kubernetes_cluster.go
grahamhayes: I don't think we will - I think this is always just going to be a true / false flag.

azurerm/data_source_kubernetes_cluster.go
grahamhayes: Done

azurerm/data_source_kubernetes_cluster.go
grahamhayes: Done

PR #4510, opened 2019-10-04
subnet resource: remove "Deprecated" attribute on "network_security_group_id" and "route_table_id"

khaliddermoumi:
subnet resource: removed the "Deprecated" attribute on "network_security_group_id" and "route_table_id", as they are still required to be used ([see issue 3077](https://github.com/terraform-providers/terraform-provider-azurerm/issues/3077)), causing "deprecated" warnings. They can't be deprecated and have to be used at the same time.

### DIFF
--- a/azurerm/resource_arm_subnet.go
+++ b/azurerm/resource_arm_subnet.go
@@ -48,15 +48,13 @@ func resourceArmSubnet() *schema.Resource {
 			},
 
 			"network_security_group_id": {
-				Type:       schema.TypeString,
-				Optional:   true,
-				Deprecated: "Use the `azurerm_subnet_network_security_group_association` resource instead.",
+				Type:     schema.TypeString,
+				Optional: true,
 			},
 
 			"route_table_id": {
-				Type:       schema.TypeString,
-				Optional:   true,
-				Deprecated: "Use the `azurerm_subnet_route_table_association` resource instead.",
+				Type:     schema.TypeString,
+				Optional: true,
 			},
 
 			"ip_configurations": {


tombuildsstuff:
hi @khaliddermoumi 

Thanks for this PR.

The fields `network_security_group_id` and `route_table_id` are deprecated as they're [scheduled to be removed in 2.0](https://www.terraform.io/docs/providers/azurerm/guides/2.0-upgrade-guide.html#resource-azurerm_subnet). Whilst it's unfortunate that these fields need to be specified in both the `azurerm_subnet` resource and the `_association` resources - we're unable to move exclusively to the `_association` resources until 2.0 to avoid introducing a breaking change. As such whilst I'd like to thank you for this contribution I'm going to close this PR for the moment.

Thanks!

khaliddermoumi:
Hi Tom,
no worries, I don't mind if you don't use this PR. But instead, I believe you should fix the problem. My point was: either remove "Deprecated", or make the "real" fix. I don't think that waiting for many months for 2.0 is ok (2.0 is "9% complete" as of yesterday). People want to have clean code, and I believe you shouldn't have to wait for months to get rid of masses of "Deprecated" warnings. So if you don't want to make a real fix, then remove the "Deprecated" attribute. As this is very simple and risk-free to do, I believe this should be done in the 1.x version soon now.
Thanks for your work, BTW!

tombuildsstuff:
hey @khaliddermoumi

The "9% complete" on the milestone is unfortunately misleading (since some PR's are closed until we're able to move forward with them) - I'm actually half way through writing a more detailed status update for 2.0 at the moment which should clarify that.

Unfortunately it's not possible to make the "real fix" here without bumping the major version number (since we try to follow SemVer, avoiding breakages within Azure) - so whilst I appreciate this is frustrating there's a warning output for this field, this is intentional to raise awareness that this field has been superseded until this is ultimately fixed in 2.0

Thanks!

dossy:
It's confusing (to me) that an interface is deprecated before the alternative, preferred solution, is available or in place.

This shouldn't be deprecated yet.  Wait until 2.0, then deprecate `azurerm_subnet.network_security_group_id` only after the preferred method is available, properly warning people that if they don't update, it will eventually stop working.

Deprecating it now, without the preferred alternative available, becomes a chicken-little warning, and after seeing it long enough, will simply get ignored, and when you do decide to obsolete it and remove it, plenty of people's configurations will break and they will be rightfully surprised, because you've _conditioned them_ to ignore the deprecation warning, because you gave them no choice but to ignore it.

tombuildsstuff:
@dossy it's unfortunate, but this is the only way we can achieve that without introducing a breaking for a use-case. It's worth noting the `network_security_group_id` field is deprecated since this field'll be completely removed in 2.0 - as such whilst we're aware this behaviour isn't ideal it's the best way to alert people to this at this time without introducing a breaking change to the behaviour of this field.

khaliddermoumi:
Hi everybody!
Tom, I can't really follow this argument, as with a "next major version", everybody expects breaking changes anyway. In fact, major version changes are defined as bringing breaking changes. So I can't see why you couldn't leave the field as it is here (without "deprecated"), and remove it in 2.0. I would not be surprised that there are breaking changes when I switch to a new breaking change.
Best Regards!

PR #4522, opened 2019-10-04
refactor: making more clients public

mbfrahry:


azurerm/config.go
```diff
@@ -112,28 +112,29 @@ type ArmClient struct {
...
+	Logic            *logic.Client
+	ManagementGroups *managementgroup.Client
+	Maps             *maps.Client
+	Mariadb          *mariadb.Client
+	Media            *media.Client
+	Monitor          *monitor.Client
+	Msi              *msi.Client
+	Mssql            *mssql.Client
+	Mysql            *mysql.Client
+	postgres         *postgres.Client
```
tombuildsstuff: it might be the Diff - but something doesn't look right here?

PR #4527, opened 2019-10-05
provider: ensure all locks are done in the same order

katbyte:
a follow up on #4501


nitzanm:
@katbyte Thanks for doing this! Looking at the code, and comparing it with my notes in #4501, I think there's still some inconsistency.  Here's what I'm seeing in the code after both #4501 and this are merged:

NSG association:
1. NSG
1. Subnet
1. Vnet

Network interface:
1. NSG
1. Subnet
1. Vnet

resource_arm_firewall.go:
1. Vnet
1. Subnet

resource_arm_network_profile.go:
1. Vnet
1. Subnet

resource_arm_redis_cache.go:
1. Vnet
1. Subnet

I think NSG association and network interface are still in the opposite order than the three files you modified here.  Can you take a quick look to see if I'm getting this right?

PR #4531, opened 2019-10-06
add warning for deprecated properties

scott1138:
when using the association resources for route tables or network security groups subsequent configuration executions will not see the id and while it is option and being deprecated it will still remove your association and require an additional run to fix.

### DIFF
--- a/website/docs/r/subnet.html.markdown
+++ b/website/docs/r/subnet.html.markdown
@@ -61,11 +61,11 @@ The following arguments are supported:
 
 * `network_security_group_id` - (Optional / **Deprecated**) The ID of the Network Security Group to associate with the subnet.
 
--> **NOTE:** At this time Subnet `<->` Network Security Group associations need to be configured both using this field (which is now Deprecated) and/or using the `azurerm_subnet_network_security_group_association` resource. This field is deprecated and will be removed in favour of that resource in the next major version (2.0) of the AzureRM Provider.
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hi @scott1138 

Thanks for this PR :)

In 1.x versions of the Azure Provider these Subnet <-> NSG/Route Table associations need to be specified both in the `azurerm_subnet` and the `_association` resources, rather than solely in the `_association` resources, whilst using `ignore_changes` may fix the plan we don't recommend omiting these fields from the `azurerm_subnet` resource since this association could potentially be removed if the `azurerm_subnet` resource is updated (which is why we recommend using both here, until the older fields are removed in 2.0).

As such whilst I'd like to thank you for this contribution I'm going to close this PR for the moment.

Thanks!

scott1138:
@tombuildsstuff Makes sense.  Should we then make the make the language more clear? 

NOTE: At this time Subnet <-> Network Security Group associations need to be configured both using this field (which is now Deprecated) *and/or* using the azurerm_subnet_network_security_group_association resource

I took the and/or to mean do it in both places or just use the association.

tombuildsstuff:
@scott1138 sorry I'd missed that - I guess the `/or` bit should be removed to make that clearer?

scott1138:
@tombuildsstuff I think that would eliminate the ambiguity.

scott1138:
@tombuildsstuff #4545 to save you the effort :)
Also, thanks for all the amazing work on the provider!

Issue #4378, opened 2019-09-19
multi-tenant deployment using Azure shared image gallery via terraform


nicethomaslearngit:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I have created shared image gallery in tenant1 and I am able to create VMs using this image across all subscriptions in this tenant. Used CLI, powershell and Terraform and it worked fine

When I try to create a VM in Tenant-2 with Shared Image Galley available in Tenant-1, it is failing with below error.

Error: compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=403 -- Original Error: Code="LinkedAuthorizationFailed" Message="The client has permission to perform action 'Microsoft.Compute/galleries/images/versions/read' on scope '/subscriptions/subscription-ID-of-tenant-2/resourceGroups/mygroup/providers/Microsoft.Compute/virtualMachines/sigvm-01', however the current tenant '<Tenant-2 ID>' is not authorized to access linked subscription '<subscription-ID-of-tenant-1>'." 

I am using the version 1.34.0

### New or Affected Resource(s)

* azurerm_1.34.0

### Potential Terraform Configuration

```hcl
####  connecting to tenant -2 ########
provider "azurerm" {
  subscription_id = "${var.subscription_id}"
  client_id       = "${var.client_id}"
  client_secret   = "${var.client_secret}"
  tenant_id       = "${var.tenant_id}"
}

#####  connecting to tenant -1 ######## Provider for fetching the SIG resource
provider "azurerm" {
  alias           = "sig-resources-id"
  subscription_id = "${var.sig_subscription_id}"
  client_id       = "${var.sig_client_id}"
  client_secret   = "${var.sig_client_secret}"
  tenant_id       = "${var.sig_tenant_id}"

```

### References

* #4290

nicethomaslearngit:
Hi @tombuildsstuff 

Could you kindly assist me what changes are needed here for using shared Image galleries as above.

nicethomaslearngit:
Hi @tombuildsstuff 

Have you got a chance to look into this.. thank you

[...]

PR #4548, opened 2019-10-07
Update key vault data source docs for deprecated vault_uri

### LINKED ISSUE #3029: Update documentation for azurerm_key_vault_* according to #2396
The changes in #2396 are not reflected in the documentation. See #2396 discussion.

phekmat:
- Replace the `vault_uri` attribute with `key_vault_id` for `key_vault_key`
- Add a note that the key vault must be in the same subscription as the
configured provider. See #2396 for discussion
- Update example that uses the deprecated `vault_uri`

Fixes #3029

### DIFF
--- a/website/docs/d/key_vault_key.html.markdown
+++ b/website/docs/d/key_vault_key.html.markdown
@@ -33,7 +33,9 @@ The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the Key Vault Key.
 
-* `key_vault_id` - (Required) Specifies the ID of the Key Vault instance where the Secret resides, available on the `azurerm_key_vault` Data Source / Resource. **NOTE:** The vault must be in the same subscription as the provider. If the vault is in another subscription, you must create an aliased provider for that subscription.
+* `key_vault_id` - (Required) Specifies the ID of the Key Vault instance where the Secret resides, available on the `azurerm_key_vault` Data Source / Resource. 
+
+**NOTE:** The vault must be in the same subscription as the provider. If the vault is in another subscription, you must create an aliased provider for that subscription.
 
 ## Attributes Reference
 
--- a/website/docs/d/key_vault_key.html.markdown
+++ b/website/docs/d/key_vault_key.html.markdown
@@ -18,8 +18,8 @@ Use this data source to access information about an existing Key Vault Key.
 
 ```hcl
 data "azurerm_key_vault_key" "test" {
-  name      = "secret-sauce"
-  vault_uri = "https://rickslab.vault.azure.net/"
+  name         = "secret-sauce"
+  key_vault_id = "${data.azurerm_key_vault.existing.id}"
 }
 
 output "key_type" {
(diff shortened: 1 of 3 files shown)


website/docs/d/key_vault_key.html.markdown
```diff
@@ -33,7 +33,7 @@ The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the Key Vault Key.
 
-* `vault_uri` - (Required) Specifies the ID of the Key Vault Key Vault instance where the Key resides, available on the `azurerm_key_vault` Data Source / Resource.
+* `key_vault_id` - (Required) Specifies the ID of the Key Vault instance where the Secret resides, available on the `azurerm_key_vault` Data Source / Resource. **NOTE:** The vault must be in the same subscription as the provider. If the vault is in another subscription, you must create an aliased provider for that subscription.
```
katbyte: I'm not sure this is entirely true as we do support multi-tenant auth allowing cross subscription access now. As such could we remove the note?
```suggestion
* `key_vault_id` - (Required) Specifies the ID of the Key Vault instance where the Secret resides, available on the `azurerm_key_vault` Data Source / Resource. 
```

website/docs/d/key_vault_secret.html.markdown
```diff
@@ -33,8 +33,7 @@ The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the Key Vault Secret.
 
-* `key_vault_id` - (Required) Specifies the ID of the Key Vault instance where the Secret resides, available on the `azurerm_key_vault` Data Source / Resource.
-
+* `key_vault_id` - (Required) Specifies the ID of the Key Vault instance where the Secret resides, available on the `azurerm_key_vault` Data Source / Resource. **NOTE:** The vault must be in the same subscription as the provider. If the vault is in another subscription, you must create an aliased provider for that subscription.
```
katbyte: I'm not sure this is entirely true as we do support multi-tenant auth allowing cross subscription access now. As such could we remove the note?
```suggestion
* `key_vault_id` - (Required) Specifies the ID of the Key Vault instance where the Secret resides, available on the `azurerm_key_vault` Data Source / Resource.
```

phekmat:
> 
> 
> Thank you for correcting these docs @phekmat! LGTM except that i'm not entirely sure about your note wrt key vault. We recently added multi tenant auth allowing cross subscription access so i think we should remove the note? WDYT?

Perhaps I'm not configuring the provider correctly, but I run into the same issue as in https://github.com/terraform-providers/terraform-provider-azurerm/issues/2396#issuecomment-467079855 when I attempt to use a key vault in another subscription, hence the note.

katbyte:
You will have to create a SP with access to both subscriptions in two tenants? i'm not entirely sure as i've not played around with it too much. You can see more details here: https://github.com/terraform-providers/terraform-provider-azurerm/issues/4378

phekmat:
> 
> 
> You will have to create a SP with access to both subscriptions in two tenants? i'm not entirely sure as i've not played around with it too much. You can see more details here: #4378

That seems like a different configuration than I was imagining. We have a single tenant, but multiple subscriptions with a service principal for each subscription. Even with contributor access to the other subscription, the data source can't find the key vault because it's only looking in the subscription the provider is configured with. The solution in the thread also seems to use an aliased provider for the other subscription. 

All this is further complicated by the fact that key vault has its own set of permissions for the data plane.

PR #4592, opened 2019-10-10
`azurerm_api_management_api` - the `version` and `version_set_id` properties can now be set

### LINKED ISSUE #4297: Unable to associate azurerm_api_management_api to azurerm_api_management_api_version_set
### Community Note

### Description
There is no way (as far as I can tell) to specify which version set an api belongs to. I dug into the azurerm_api_management_api resource source code and saw read only `version` and `version_set_id` properties. I would expect these two properties to be writable to allow for the linkage of an api to a version set. 

Terraform v0.12.6
azurerm v1.32.0

### Affected Resource(s)

* `azurerm_api_management_api`
* `azurerm_api_management_api_version_set`

kraihn:
Should fix #4297 

This is my first Terraform provider pull request and step into programming in Go. What am I missing? Is there anything I need to be aware of in terms of a list of items to check or validate? The footprint of the arguments I'm interested in appears to be small.

### DIFF
--- a/azurerm/resource_arm_api_management_api.go
+++ b/azurerm/resource_arm_api_management_api.go
@@ -260,34 +260,21 @@ func resourceArmApiManagementApiCreateUpdate(d *schema.ResourceData, meta interf
 		apiType = apimanagement.APIType(apimanagement.SoapToRest)
 	}
 
-	var params apimanagement.APICreateOrUpdateParameter
+	params := apimanagement.APICreateOrUpdateParameter{
+		APICreateOrUpdateProperties: &apimanagement.APICreateOrUpdateProperties{
+			APIType:                       apiType,
+			Description:                   utils.String(description),
+			DisplayName:                   utils.String(displayName),
+			Path:                          utils.String(path),
+			Protocols:                     protocols,
+			ServiceURL:                    utils.String(serviceUrl),
+			SubscriptionKeyParameterNames: subscriptionKeyParameterNames,
+			APIVersion:                    utils.String(version),
+		},
+	}
+
 	if versionSetId != "" {
-		params = apimanagement.APICreateOrUpdateParameter{
-			APICreateOrUpdateProperties: &apimanagement.APICreateOrUpdateProperties{
-				APIType:                       apiType,
-				Description:                   utils.String(description),
-				DisplayName:                   utils.String(displayName),
-				Path:                          utils.String(path),
-				Protocols:                     protocols,
-				ServiceURL:                    utils.String(serviceUrl),
-				SubscriptionKeyParameterNames: subscriptionKeyParameterNames,
-				APIVersion:                    utils.String(version),
-				APIVersionSetID:               utils.String(versionSetId),
-			},
-		}
-	} else {
-		params = apimanagement.APICreateOrUpdateParameter{
-			APICreateOrUpdateProperties: &apimanagement.APICreateOrUpdateProperties{
-				APIType:                       apiType,
-				Description:                   utils.String(description),
-				DisplayName:                   utils.String(displayName),
-				Path:                          utils.String(path),
-				Protocols:                     protocols,
-				ServiceURL:                    utils.String(serviceUrl),
-				SubscriptionKeyParameterNames: subscriptionKeyParameterNames,
-				APIVersion:                    utils.String(version),
-			},
-		}
+		params.APICreateOrUpdateProperties.APIVersionSetID = utils.String(versionSetId)
 	}
 
 	if _, err := client.CreateOrUpdate(ctx, resourceGroup, serviceName, apiId, params, ""); err != nil {
--- a/azurerm/resource_arm_api_management_api.go
+++ b/azurerm/resource_arm_api_management_api.go
@@ -170,11 +170,13 @@ func resourceArmApiManagementApi() *schema.Resource {
 
 			"version": {
 				Type:     schema.TypeString,
+				Computed: true,
 				Optional: true,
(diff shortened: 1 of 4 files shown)


azurerm/resource_arm_api_management_api.go
```diff
@@ -170,12 +170,12 @@ func resourceArmApiManagementApi() *schema.Resource {
 
 			"version": {
 				Type:     schema.TypeString,
-				Computed: true,
+				Optional: true,
```
tombuildsstuff: I believe this'll need to be Optional & Computed, which means that it's configurable (Optional) but can have a default value (Computed) since Azure auto-assigns a version if one is unspecified?

azurerm/resource_arm_api_management_api.go
```diff
@@ -257,16 +260,34 @@ func resourceArmApiManagementApiCreateUpdate(d *schema.ResourceData, meta interf
...
-			APIType:                       apiType,
-			Description:                   utils.String(description),
-			DisplayName:                   utils.String(displayName),
-			Path:                          utils.String(path),
-			Protocols:                     protocols,
-			ServiceURL:                    utils.String(serviceUrl),
-			SubscriptionKeyParameterNames: subscriptionKeyParameterNames,
-		},
+	var params apimanagement.APICreateOrUpdateParameter
+	if versionSetId != "" {
```
tombuildsstuff: rather than doing this, since the only difference between the two blocks appears to be the `versionSetId` field, we could have a single block and do:

```
if versionSetId != "" {
  params.APICreateOrUpdateProperties.APIVersionSetID = utils.String(versionSetId)
}
```

azurerm/resource_arm_api_management_api.go
```diff
@@ -243,6 +243,9 @@ func resourceArmApiManagementApiCreateUpdate(d *schema.ResourceData, meta interf
 	displayName := d.Get("display_name").(string)
 	serviceUrl := d.Get("service_url").(string)
 
+	version := d.Get("version").(string)
+	versionSetId := d.Get("version_set_id").(string)
```
tombuildsstuff: we also need to set this field back into the state in the Read function below (`resourceArmApiManagementApiRead`) and [add this to the documentation](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/api_management_api.html.markdown#L67) - could we also add an Acceptance Test for this ([example](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/azurerm/resource_arm_api_management_api_test.go#L14-L40))? this'd involve adding a new test setting this field to confirm this works - which can be [run using the instructions in the README](https://github.com/terraform-providers/terraform-provider-azurerm#developing-the-provider)

azurerm/resource_arm_api_management_api.go
kraihn: When I left `computed: true` during testing, removing the `version` and `version_set_id` from a resource resulted in no change on an apply. But I would expect an API to be able to be removed from a version set. Any ideas why I would see this?

azurerm/resource_arm_api_management_api.go
kraihn: Wasn't quite sure how to do this in Go. Thanks for sharing! Changed!

azurerm/resource_arm_api_management_api.go
kraihn: My local state also doesn't appear to persist the `version` and `version_set_id`  between applies when `computed: true`.

azurerm/resource_arm_api_management_api.go
kraihn: I put in a first pass at setting the Read state, but I'm not sure how to validate. Do you have a way to debug with breakpoints?

azurerm/resource_arm_api_management_api.go
kraihn: Azure doesn't require or set a version on the first API in a version set. A version is only required on subsequent versions, and that value also is not generated.

azurerm/resource_arm_api_management_api.go
kraihn: @tombuildsstuff, I believe I have these items covered now.

sam-cogan:
Would be great to get this added, currently it is impossible to create versioned APIs

azurerm/resource_arm_api_management_api.go
```diff
@@ -170,14 +170,25 @@ func resourceArmApiManagementApi() *schema.Resource {
...
 			"version_set_id": {
 				Type:     schema.TypeString,
-				Computed: true,
+				Optional: true,
 			},
 		},
+
+		CustomizeDiff: func(diff *schema.ResourceDiff, v interface{}) error {
+			_, hasVersion := diff.GetOk("version")
+			_, hasVersionSetId := diff.GetOk("version_set_id")
```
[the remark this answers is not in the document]
kraihn: Removed.

website/docs/d/api_management_api.html.markdown
```diff
@@ -35,6 +35,12 @@ output "api_management_api_id" {
 
 * `revision` - (Required) The Revision of the API Management API.
 
+* `version` - (Optional) The Version number of this API, if this API is versioned.
+
+* `version_set_id` - (Optional) The ID of the Version Set which this API is associated with.
+
+-> **NOTE:** When `version` is set, `version_set_id` must also be specified
```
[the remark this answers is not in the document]
kraihn: Moved.

PR #4593, opened 2019-10-10
provider: adding timeout support to resources a -> d

mbfrahry:


azurerm/resource_arm_container_service.go
```diff
@@ -35,6 +35,13 @@ If you're using ACS with Kubernetes, we'd recommend migrating to AKS / the 'azur
 More information can be found here: https://azure.microsoft.com/en-us/updates/azure-container-service-will-retire-on-january-31-2020/
 `,
 
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(80 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(80 * time.Minute),
+			Delete: schema.DefaultTimeout(80 * time.Minute),
+		},
```
tombuildsstuff: this is being removed in 2.0 - so we can remove this one 👍

azurerm/resource_arm_cosmosdb_account.go
```diff
@@ -34,6 +34,13 @@ func resourceArmCosmosDbAccount() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(120 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(120 * time.Minute),
+			Delete: schema.DefaultTimeout(120 * time.Minute),
```
tombuildsstuff: the deployment time for CosmosDB depends on how many replica's are specified - as such I think it'd be prudent to set these to `180m`

Issue #4595, opened 2019-10-11
resource_arm_firewall, resource_arm_firewall_application_rule_collection – Mssql and Threat Intelligence


BrendanThompson:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The default behaviour of the `azurerm_firewall` in respect to [Threat Intelligence](https://docs.microsoft.com/en-us/azure/firewall/threat-intel) is to enable it as **Alert Only**, this however is not configurable via the Terraform resource itself, ideally this is something that should be surfaced in the provider as it's available via the API. ( https://docs.microsoft.com/en-us/rest/api/firewall/azurefirewalls/createorupdate#azurefirewallthreatintelmode )

Secondly the `azurerm_firewall_application_rule_collection` has support via the API to have a protocol of type `Mssql` this should be available through Terraform as well. Keeping in mind this is currently not available in the backing API client library I have raised an issue there too. ( https://github.com/Azure/azure-sdk-for-go/issues/6003 )

### New or Affected Resource(s)

* azurerm_firewall
* azurerm_firewall_application_rule_collection

### Potential Terraform Configuration

```hcl
resource "azurerm_firewall" "this" {
  name                = local.firewall_name
  location            = data.azurerm_resource_group.this.location
  resource_group_name = data.azurerm_resource_group.this.name
  threat_intelligence_mode = "Alert"

  ip_configuration {
    name                 = format("%s-ipconfig", local.firewall_name)
    subnet_id            = module.firewall_subnet.subnet_ids[0]
    public_ip_address_id = azurerm_public_ip.this.id
  }
}

resource "azurerm_firewall_application_rule_collection" "deny_all" {
  count = var.enable_default_rules ? 1 : 0

  name                = "deny-all"
  azure_firewall_name = azurerm_firewall.this.name
  resource_group_name = data.azurerm_resource_group.this.name
  priority            = 65000
  action              = "Deny"

  rule {
    name             = "deny_catch_all"
    source_addresses = ["*"]
    target_fqdns     = ["*"]

    protocol {
      port = "1433"
      type = "Mssql"
    }
  }

}
```

[...]

PR #4596, opened 2019-10-11
feat(application_collection): updated api ver, added mssql protocol

BrendanThompson:
- Updated the API version to `2019-07-01` to allow for use of the Mssql const
- Added Mssql type into Protocol validation

RELATES: #4595


azurerm/resource_arm_firewall_application_rule_collection.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"log"
 
-	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-06-01/network"
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-07-01/network"
```
tombuildsstuff: since we're bumping the API version here could we do this for all of the networking resources?

azurerm/resource_arm_firewall_application_rule_collection.go
```diff
@@ -104,6 +104,7 @@ func resourceArmFirewallApplicationRuleCollection() *schema.Resource {
 										ValidateFunc: validation.StringInSlice([]string{
 											string(network.AzureFirewallApplicationRuleProtocolTypeHTTP),
 											string(network.AzureFirewallApplicationRuleProtocolTypeHTTPS),
+											string(network.AzureFirewallApplicationRuleProtocolTypeMssql),
```
tombuildsstuff: could we add this new value to [the documentation too](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/firewall_application_rule_collection.html.markdown#L121)?

PR #4615, opened 2019-10-14
azurerm_network_security_rule & azurerm_network_security_rule - add support for `Icmp` to the `protocol` property

ignatz42:
Attempt to fix https://github.com/terraform-providers/terraform-provider-azurerm/issues/3937


katbyte:
Hmm, oddly the test is failing with:
```
config is invalid: expected protocol to be one of [* Tcp Udp], got Icmp
```

katbyte:
@ignatz42 

You'll need to add the new value to the security rule resource as well and then the test will pass. I tried to push the required changes to you branch but i don't have permission.

ignatz42:
@katbyte Thank you for the review and finding the test failure.  I've updated the commit with a small docs change and the fix for the failure.

PR #4467, opened 2019-09-30
Adds throughput support for Cosmos DB Mongo Collection

kmfk:
This would resolve #3586 and allow setting the `throughput` property on the collection.

There are several places where support should be added for `throughput` on the Cosmos DB resources. If this ok, I can add in the others.

PR #4616, opened 2019-10-14
[DO NOT MERGE] Adds Throughput to CosmosDB Resources

### LINKED ISSUE #3623: Set throughput at CosmosDB SQL database level
### Community Note

### Description

CosmosDB suports setting throughput at a database level and shared across collections, this is not currently possible when provisioning using terraform. 

This affects the cost and performance so is needed for this resource

*  azurerm_cosmosdb_sql_database 

```hcl
resource "azurerm_cosmosdb_sql_database" "example" {
  name                          = "tfex-cosmos-mongo-db"
  resource_group_name = "${data.azurerm_cosmosdb_account.example.resource_group_name}"
  account_name             = "${data.azurerm_cosmosdb_account.example.name}"
  throughput                 = 400
}
```

kmfk:
As a follow up to #4467, this would add the ability to specify `throughput` on the remaining, supported CosmosDB resources. This would resolve #3623.

### DIFF
--- a/azurerm/resource_arm_cosmosdb_cassandra_keyspace.go
+++ b/azurerm/resource_arm_cosmosdb_cassandra_keyspace.go
@@ -41,6 +41,13 @@ func resourceArmCosmosDbCassandraKeyspace() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validate.CosmosAccountName,
 			},
+
+			"throughput": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      400,
+				ValidateFunc: validate.CosmosThroughput,
+			},
 		},
 	}
 }
@@ -53,6 +60,7 @@ func resourceArmCosmosDbCassandraKeyspaceCreate(d *schema.ResourceData, meta int
 	name := d.Get("name").(string)
 	resourceGroup := d.Get("resource_group_name").(string)
 	account := d.Get("account_name").(string)
+	throughput := d.Get("throughput").(int)
 
 	if features.ShouldResourcesBeImported() && d.IsNewResource() {
 		existing, err := client.GetCassandraKeyspace(ctx, resourceGroup, account, name)
@@ -86,6 +94,23 @@ func resourceArmCosmosDbCassandraKeyspaceCreate(d *schema.ResourceData, meta int
 
 	if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
 		return fmt.Errorf("Error waiting on create/update future for Cosmos Cassandra Keyspace %s (Account %s): %+v", name, account, err)
+	}
+
+	throughputParameters := documentdb.ThroughputUpdateParameters{
+		ThroughputUpdateProperties: &documentdb.ThroughputUpdateProperties{
+			Resource: &documentdb.ThroughputResource{
+				Throughput: utils.Int32(int32(throughput)),
+			},
+		},
+	}
+
+	throughputFuture, err := client.UpdateCassandraKeyspaceThroughput(ctx, resourceGroup, account, name, throughputParameters)
+	if err != nil {
+		return fmt.Errorf("Error setting Throughput for Cosmos Cassandra Keyspace %s (Account %s): %+v", name, account, err)
+	}
+
+	if err = throughputFuture.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("Error waiting on ThroughputUpdate future for Cosmos Cassandra Keyspace %s (Account %s): %+v", name, account, err)
 	}
 
 	resp, err := client.GetCassandraKeyspace(ctx, resourceGroup, account, name)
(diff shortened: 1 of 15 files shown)


kmfk:
@katbyte As mentioned in my previous PR, followed up to add `throughput` to the other ComosDB resources that supported it.

cc @ragoragino

kmfk:
@katbyte I had to fix the tests in Travis and doing so, I ran the Acceptance tests as well.  I actually get an 404 error from each of the Throughput methods exposed in the Go SDK. At first, I thought it was from the initial create not blocking before I was attempting to update throughput.

However, I opened up my REST client and tried the underlying rest endpoints directly and receive a 404 from the route on an existing resource - I don't believe these endpoints work as expected.

For example, I created a CosmosDB Account with a SQL Database and used the API to verify it existed, which it did.  I then hit the endpoint to fetch it's throughput and I get a 404 - its very reproducible.

This would actually block this PR and we should probably revert the change from #4467 until the API is fixed.

kmfk:
For example, using the endpoint to [GET the Database](https://docs.microsoft.com/en-us/rest/api/cosmos-db-resource-provider/databaseaccounts/getsqldatabase):
Screenshot 2019-10-14 16 18 11

And then trying to [GET Throughput](https://docs.microsoft.com/en-us/rest/api/cosmos-db-resource-provider/databaseaccounts/getsqldatabasethroughput):
Screenshot 2019-10-14 16 20 53

This would be the same endpoint hit by the Go SDK, [here](https://github.com/Azure/azure-sdk-for-go/blob/master/services/cosmos-db/mgmt/2015-04-08/documentdb/databaseaccounts.go#L6618).

katbyte:
@kmfk,

The tests for #4467 don't fail with a 404, in fact i can see the throughput is actually being read & updated correctly. (it fails due to indexs not matching)

Maybe this is a transient issue or due to the region you are using? (we run our tests in west europe)

ragoragino:
@katbyte, @kmfk 

We have been also getting 404 with an update of the MongoDB database after its creation.

It seems that the database needs to be created with the throughput property in order to be applicable to all collections. If it is only updated on throughput property (without the previous creation including that property), then the REST request fails with 404.

Here is the gist that reproduces the issue, where the first MongoDB database is created with throughput property (and succeeds), while the second one is only updated after its creation (and only creation succeeds, the update returns 404). If I modify the second case and create the database with the throughput property, the update also succeeds (with a value different than the value used when creating the database):

https://gist.github.com/ragoragino/7f6488a34f0b2d7fda235afcb40fd74b

The header "x-ms-offer-throughput" needs to be sent in the request as documented in https://docs.microsoft.com/en-us/rest/api/cosmos-db/create-a-collection for the creation of the database with the throughput property to work correctly.

kmfk:
@ragoragino I had done some more testing last night and mostly agree with your summary.  It does seem like Mongo Collections and SQL Containers can actually update throughput even if they were first created _without_ throughput explicitly set at creation - regardless of if the Database was created with throughput explicitly set.

While the PUT endpoint is essentially an upsert (create/update) - my testing found that if you create a database without explicitly setting throughput initially, subsequent PUT calls trying to add throughput wouldn't error - but would still not update the throughput. The throughput endpoints will still 404.

So, this aligns with what @katbyte said - the MongoDB Collection succeeded when I changed only that resource. The acceptance test now fails due to the changes above on the MongoDB Database resource - since the collection has a dependency on it and the Database creation fails.

While I could try to catch the 404 errors, it seems like if throughput was not set initially, setting it on an existing Database / Keyspace resource would require a `forceNew`. I can also change this to always set the header to add throughput - but that wouldn't help existing resources created with an earlier version of the provider.

Finally, since (to my knowledge) there is no way to do a conditional `forceNew`, I think it would mean we'd have to set `forceNew` on throughput for the Database and Keyspace resources to maintain backwards compatibility - even though they should technically be able to be modified in place if they were created with throughput explicitly set. 

Sorry for the wall of text - thoughts or anything I'm missing?

pmccloghrylaing:
I'm keen to see this feature get in. Would failing the update and providing an appropriate error be a solution to this, instead of `forceNew`? Something that lets the user know that they'll need to recreate the database in order to set throughput. If it only happens when encountering the 404 error it also means that if/when the Azure API allows these updates they'll start working and the error won't be displayed.

katbyte:
@kmfk, @pmcatominey,

I've been thinking about it over the weekend and I agree. I think what we could do is:
- open an issue on the azure-for-go SDK detailing that the update fails in this situation
- if we get a error wrap it in our own error "Error updating throughput, was this collection/db/etc created with a custom throughput? you may need to recreate it. for more details http://url_to_sdk_issue"
- update the docs

WDYT @kmfk?

katbyte:
@kmfk,

I've opened #5116 to address issues with the existing throughput properties. That technique can now be used for the remaining cosmos resources.

aristosvo:
What if I want to create a collection which I want to use shared throughput? Is that an option or should I open an issue?

katbyte:
Hi @kmfk,

I hope you don't mind but i have opened #5203 with the required changes, as such i'm going to close this PR. Thanks again for all the work you put in!

kmfk:
@katbyte Nope, sorry I didn't get back to this - thanks for pushing it forward!

Issue #4461, opened 2019-09-30
Azure FrontDoor : Caching issue - Unexpected resource modifying 


Lidestyle:
### Terraform (and AzureRM Provider) Version

```sh
terraform -v
Terraform v0.12.9
+ provider.azurerm v1.34.0
```

### Affected Resource(s)

* `azurerm_frontdoor`

### -  --  main.tf
```sh
// Manage Azure FrontDoor
resource "azurerm_frontdoor" "traffic_gateway" {
  name                                         = local.frontdoor_name
  friendly_name                                = local.frontdoor_name
  location                                     = var.location
  resource_group_name                          = var.resource_group
  enforce_backend_pools_certificate_name_check = var.backend_certificate_check # false

  routing_rule {
      name                    = var.routing_rule_name
      accepted_protocols      = ["Https"]
      patterns_to_match       = ["/*"]
      frontend_endpoints      = [var.frontend_endpoint_name]
      forwarding_configuration {
        forwarding_protocol   = var.forwarding_protocol
        backend_pool_name     = var.backend_pool_name
      }
  }

  backend_pool_load_balancing {
    name                            = var.balancing_name
    sample_size                     = local.backand_balancer.sample_size
    successful_samples_required     = local.backand_balancer.successful_samples_required
    additional_latency_milliseconds = local.backend_balancer.additional_latency_milliseconds
  }

  backend_pool_health_probe {
    name                  = var.liveness_probe_name
    path                  = var.healthcheck_api_path
    interval_in_seconds   = 255
    protocol              = "Https"
  }

  backend_pool {
      name            = var.backend_pool_name
      backend {
          host_header = "${var.backend_name}.azure-api.net"
          address     = "${var.backend_name}.azure-api.net"
          http_port   = 80
          https_port  = 443
      }

      load_balancing_name = var.balancing_name
      health_probe_name   = var.liveness_probe_name
  }

  frontend_endpoint {
    name                                      = var.frontend_endpoint_name
    host_name                                 = var.backend_host_name
    web_application_firewall_policy_link_id   = var.waf_enabled != false ? azurerm_frontdoor_firewall_policy.waf_assigned_policy[0].id : null
    custom_https_provisioning_enabled         = false
  }

  tags = {
    environment = var.environment_prefix
  }
}

// Manage FrontDoor firewall policy
resource "azurerm_frontdoor_firewall_policy" "waf_assigned_policy" {
```

### Expected Behavior
```sh 
[...]

PR #4618, opened 2019-10-15
Bug: `azurerm_frontdoor` fix for caching issue

WodansSon:
Real fix for #4461


WodansSon:
image

azurerm/resource_arm_front_door.go
```diff
@@ -172,24 +172,25 @@ func resourceArmFrontDoor() *schema.Resource {
 										Required:     true,
 										ValidateFunc: afd.ValidateBackendPoolRoutingRuleName,
 									},
+									// Remove default value for #4461
 									"cache_use_dynamic_compression": {
 										Type:     schema.TypeBool,
 										Optional: true,
-										Default:  false,
```
katbyte: I think this will end up causing people a diff?

azurerm/resource_arm_front_door.go
```diff
@@ -172,24 +172,25 @@ func resourceArmFrontDoor() *schema.Resource {
...
 									},
+									// Remove default value for #4461
 									"cache_query_parameter_strip_directive": {
 										Type:     schema.TypeString,
 										Optional: true,
 										ValidateFunc: validation.StringInSlice([]string{
 											string(frontdoor.StripAll),
 											string(frontdoor.StripNone),
 										}, false),
-										Default: string(frontdoor.StripNone),
```
katbyte: This is a breaking change? we can't do this until 2.0

azurerm/resource_arm_front_door.go
```diff
@@ -198,7 +199,7 @@ func resourceArmFrontDoor() *schema.Resource {
 											string(frontdoor.HTTPSOnly),
 											string(frontdoor.MatchRequest),
 										}, false),
-										Default: string(frontdoor.MatchRequest),
+										Default: string(frontdoor.HTTPSOnly),
```
katbyte: This is a breaking change? we can't do this until 2.0

azurerm/resource_arm_front_door.go
WodansSon: I don't think this will be an issue since it's data type is bool, it is always `false` by default... WDYT?

azurerm/resource_arm_front_door.go
WodansSon: No, this is not a breaking change because the value is never evaluated unless the `cache_use_dynamic_compression` is set to TRUE. So this is basically a NoOp… WDYT?

azurerm/resource_arm_front_door.go
WodansSon: Reverted default value change to avoide the breaking change, will open issue to update this in the 2.0 timeframe.

azurerm/internal/services/frontdoor/helper.go
```diff
@@ -201,3 +203,90 @@ func FlattenFrontendEndpointLinkSlice(input *[]frontdoor.FrontendEndpointLink) [
...
+// into a ResourceID. We make assumptions about the structure of URLs,
+// which is obviously not good, but the best thing available given the
+// SDK. I had to normalize the key casing of Path because the Front Door API
+// via Portal does not have consistent casing within the resource, for example:
+//
+// In the backendPools block the casing of the HealthProbeSettings is (notice the lowercase 'h'):
+// portal-front-door/ -> healthProbeSettings/healthProbeSettings-1571100669337
+//
+// but in the HealthProbeSettings block the casing of the HealthProbeSettings is (notice the uppercase 'H')::
+// portal-front-door/ -> HealthProbeSettings/healthProbeSettings-1571100669337
```
tombuildsstuff: it'd be worth raising this as a bug in the API - since the ARM Specification states that the same Resource ID should be returned in the same casing for the same API call, wherever it's present.

It's probably also worth calling out specifically that whilst the ARM Specification "technically" says Resource ID's are case insensitive; in practice they're not since both downstream API's and downstream client's don't consume them in that fashion - which I think is a larger discussion we need to have with the ARM Team 🙃

azurerm/resource_arm_front_door.go
WilheimRy: Looks like setting it to StripNone or StripAll won't change the 'caching' field but only the 'Query string caching behaviour'. Is there a way to change the 'caching' field?

PR #4620, opened 2019-10-15
azurerm_cosmosdb_mongo_collection, azurerm_cosmosdb_mongo_database  - support configuring throughput

e96wic:


e96wic:
Changes functionality of https://github.com/terraform-providers/terraform-provider-azurerm/commit/0caed50c5bdba7c78709bc7af8cba1780b6c29dc

katbyte:
hi @e96wic,

 I don't think silently recreating the DB is the best user experience here. We are discussing this in #4616 and i think we should instead error with a message saying the resource needs to be recreated by hand linking to an issue on the azure-for-go SDK detailing the limitation of the API (it does sound like a bug on their end)

WDYT?

e96wic:
Hi @katbyte, where do you think I'm recreating the DB - can you point me to the code? I totally agree with you, silently recreating resources is a bad idea!

katbyte:
whoops 😅 i totally am misremembering this PR from last week. Your entirely right, i will give it a closer look later today/tomorrow. 

It looks like you arebasicallyy doing what was suggested already by adding in some guards and removing the default throughput (which i can now see would cause some problems?)

katbyte:
@e96wic

azurerm/resource_arm_cosmosdb_mongo_collection.go
```diff
@@ -118,20 +120,27 @@ func resourceArmCosmosDbMongoCollectionCreateUpdate(d *schema.ResourceData, meta
...
+				if err != nil {
+					return fmt.Errorf("Error generating import ID for Cosmos Mongo Collection %s (Account %s, Database %s)", name, account, database)
+				}
 
-			return tf.ImportAsExistsError("azurerm_cosmosdb_mongo_collection", id)
+				return tf.ImportAsExistsError("azurerm_cosmosdb_mongo_collection", id)
+			}
+		} else if dbHasThroughputConfigured {
+			createUpdateOptions["throughput"] = utils.String(strconv.Itoa(throughput))
 		}
```
katbyte: I think we'll need to change this to

```suggestion
		if hasThroughput {
			createUpdateOptions["throughput"] = utils.String(strconv.Itoa(throughput))
		}
```

azurerm/resource_arm_cosmosdb_mongo_collection.go
```diff
@@ -118,20 +120,27 @@ func resourceArmCosmosDbMongoCollectionCreateUpdate(d *schema.ResourceData, meta
 	account := d.Get("account_name").(string)
 	database := d.Get("database_name").(string)
 	throughput := d.Get("throughput").(int)
+	dbHasThroughputConfigured := throughput > 0
```
katbyte: If we remove the default we can do:
```suggestion
	throughput , hasThroughput := d.GetOk("throughput)
```

azurerm/resource_arm_cosmosdb_mongo_collection.go
```diff
@@ -165,21 +174,27 @@ func resourceArmCosmosDbMongoCollectionCreateUpdate(d *schema.ResourceData, meta
...
-	}
+		}
 
-	throughputFuture, err := client.UpdateMongoDBCollectionThroughput(ctx, resourceGroup, account, database, name, throughputParameters)
-	if err != nil {
-		return fmt.Errorf("Error setting Throughput for Cosmos MongoDB Collection %s (Account %s, Database %s): %+v", name, account, database, err)
-	}
+		throughputFuture, err := client.UpdateMongoDBCollectionThroughput(ctx, resourceGroup, account, database, name, throughputParameters)
+		if err != nil {
+			_ = d.Set("throughput", nil)
```
katbyte: Why are you setting this to nil?

azurerm/resource_arm_cosmosdb_mongo_collection.go
```diff
@@ -243,11 +258,14 @@ func resourceArmCosmosDbMongoCollectionRead(d *schema.ResourceData, meta interfa
 
 	throughputResp, err := client.GetMongoDBCollectionThroughput(ctx, id.ResourceGroup, id.Account, id.Database, id.Collection)
 	if err != nil {
-		return fmt.Errorf("Error reading Throughput on Cosmos Mongo Collection %s (Account %s, Database %s): %+v", id.Collection, id.Account, id.Database, err)
-	}
-
-	if throughput := throughputResp.Throughput; throughput != nil {
-		d.Set("throughput", int(*throughput))
+		if !utils.ResponseWasNotFound(throughputResp.Response) {
+			_ = d.Set("throughput", nil)
```
katbyte: I don't think we need to set this?

azurerm/resource_arm_cosmosdb_mongo_collection.go
```diff
@@ -243,11 +258,14 @@ func resourceArmCosmosDbMongoCollectionRead(d *schema.ResourceData, meta interfa
...
+			return fmt.Errorf("Error reading Throughput on Cosmos Mongo Collection %s (Account %s, Database %s): %+v", id.Collection, id.Account, id.Database, err)
+		}
+	} else {
+		if throughput := throughputResp.Throughput; throughput != nil {
+			_ = d.Set("throughput", int(*throughput))
```
katbyte: d.Set() will deal with nil values so:
```suggestion
			d.Set("throughput", throughputResp.Throughput)
```

azurerm/resource_arm_cosmosdb_mongo_database.go
```diff
@@ -49,32 +52,47 @@ func resourceArmCosmosDbMongoDatabase() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validate.CosmosAccountName,
 			},
+
+			"throughput": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      nil,
```
katbyte: leaving this out is the same as nil
```suggestion
```

azurerm/resource_arm_cosmosdb_mongo_collection.go
```diff
@@ -78,7 +80,7 @@ func resourceArmCosmosDbMongoCollection() *schema.Resource {
 			"throughput": {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				Default:      400,
+				Default:      nil,
```
katbyte: I think this will technically be a breaking change? While i agree we should make the change i'm not sure we can until 2.0 where we break/change schema

azurerm/resource_arm_cosmosdb_mongo_database.go
```diff
@@ -49,32 +52,47 @@ func resourceArmCosmosDbMongoDatabase() *schema.Resource {
...
+				}
+
+				return tf.ImportAsExistsError("azurerm_cosmosdb_mongo_database", id)
 			}
-
-			return tf.ImportAsExistsError("azurerm_cosmosdb_mongo_database", id)
+		} else if dbHasThroughputConfigured {
+			createUpdateOptions["throughput"] = utils.String(strconv.Itoa(throughput))
 		}
 	}
```
katbyte: Same comments as above

azurerm/resource_arm_cosmosdb_mongo_database.go
```diff
@@ -131,10 +173,22 @@ func resourceArmCosmosDbMongoDatabaseRead(d *schema.ResourceData, meta interface
 		return fmt.Errorf("Error reading Cosmos Mongo Database %s (Account %s): %+v", id.Database, id.Account, err)
 	}
 
-	d.Set("resource_group_name", id.ResourceGroup)
-	d.Set("account_name", id.Account)
+	_ = d.Set("resource_group_name", id.ResourceGroup)
+	_ = d.Set("account_name", id.Account)
```
katbyte: We don't need the assingment
```suggestion
	d.Set("account_name", id.Account)
```

azurerm/resource_arm_cosmosdb_mongo_database.go
```diff
@@ -131,10 +173,22 @@ func resourceArmCosmosDbMongoDatabaseRead(d *schema.ResourceData, meta interface
...
 	if props := resp.MongoDBDatabaseProperties; props != nil {
-		d.Set("name", props.ID)
+		_ = d.Set("name", props.ID)
```
katbyte:
```suggestion
		d.Set("name", props.ID)
```

azurerm/resource_arm_cosmosdb_mongo_database.go
```diff
@@ -131,10 +173,22 @@ func resourceArmCosmosDbMongoDatabaseRead(d *schema.ResourceData, meta interface
...
+	throughputResp, err := client.GetMongoDBDatabaseThroughput(ctx, id.ResourceGroup, id.Account, id.Database)
+	if err != nil {
+		if !utils.ResponseWasNotFound(throughputResp.Response) {
+			_ = d.Set("throughput", nil)
+			return fmt.Errorf("Error reading Throughput on Cosmos Mongo Database %s (Account %s): %+v", id.Database, id.Account, err)
+		}
+	} else {
+		if throughput := throughputResp.Throughput; throughput != nil {
+			_ = d.Set("throughput", int(*throughput))
+		}
```
katbyte: same comments as above

azurerm/resource_arm_cosmosdb_mongo_collection.go
e96wic: That is a good question. If I'm correct it should not break anyones code / resources.
Case 1 - user had no throughput configured for existing resource -> terraform won't re-apply the throughput meaning it won't override whatever is configured in azure.
Case 2 - user creates a new resource -> will be created with the azure default (which should be 1000)
Case 3 - user had throughput configured -> will use that value

In the past my team had quite some trouble with newly added parameters with default values. In some cases it destroyed resources for us. In this case setting the throughput to a default of 400 might scale down collections for users which had them previously configured in terraform, but scaled via azure directly. If users roll their terraform config out automatically this is dangerous.

e96wic: What do you think, @katbyte ?

kmfk:
@e96wic Thanks for this follow up

@katbyte Once this PR is resolved, I'll use it for the basis of fixing #4616

azurerm/resource_arm_cosmosdb_mongo_collection.go
kmfk: Documentation and testing showed 400 being the azure default - but I do agree that defaulting to `nil` here probably makes more sense

katbyte: 👍I think your correct, at worst the user will just have a plan with no affect.

azurerm/resource_arm_cosmosdb_mongo_collection.go
e96wic: looks correct, but don't we update the resource when executing "terraform import" with the current logic?

azurerm/resource_arm_cosmosdb_mongo_collection.go
e96wic: I now moved it to the case where the throughput request failed with a 404 indicating that the resource does not have throughput configured. I came across a case while testing, where I created a resource without throuhput, then tried to set it (which is not allowed) resulting in an error but the throughput was still set to the not applied value in the terraform state.

azurerm/resource_arm_cosmosdb_mongo_collection.go
e96wic: manual tests did not show what I was assuming so I guess it's fine

e96wic:
@katbyte Could you review have another look at this PR? :)

e96wic:
Hi @katbyte, the failing acceptance test actually indicated the desired behavior. The import test step failed because the config did not have throughput configured. Azure sets a default throughput though, which will be imported to the state by my changes.

azurerm/resource_arm_cosmosdb_mongo_collection_test.go
```diff
@@ -24,7 +24,9 @@ func TestAccAzureRMCosmosDbMongoCollection_basic(t *testing.T) {
 				Config: testAccAzureRMCosmosDbMongoCollection_basic(ri, testLocation()),
 				Check: resource.ComposeAggregateTestCheckFunc(
 					testCheckAzureRMCosmosDbMongoCollectionExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "throughput", "400"),
 				),
+				ExpectNonEmptyPlan: true,
```
katbyte: We shouldn't need this, could we remove this and make the test pass?

website/docs/r/cosmosdb_mongo_database.html.markdown
```diff
@@ -36,6 +37,8 @@ The following arguments are supported:
 
 * `account_name` - (Required) The name of the Cosmos DB Mongo Database to create the table within. Changing this forces a new resource to be created.
 
+* `throughput` - (Optional) The throughput of the MongoDB collection (RU/s). Must be set in increments of `100`. The minimum value is `400`. This must be set upon database creation otherwise it cannot be updated without a manual terraform destroy-apply.
```
katbyte: Could we make this
```suggestion
* `throughput` - (Optional) The throughput of the MongoDB collection (RU/s). Must be set in increments of `100`, and defaults to `400`.

-> **NOTE:** The minimum value is `400`. This must be set upon database creation otherwise it cannot be updated without a manual terraform destroy-apply.
```

website/docs/r/cosmosdb_mongo_collection.html.markdown
```diff
@@ -56,7 +56,7 @@ The following arguments are supported:
 * `database_name` - (Required) The name of the Cosmos DB Mongo Database in which the Cosmos DB Mongo Collection is created. Changing this forces a new resource to be created.
 * `default_ttl_seconds` - (Required) The default Time To Live in seconds. If the value is `-1` items are not automatically expired.
 * `shard_key` - (Required) The name of the key to partition on for sharding. There must not be any other unique index keys.
-* `throughput` - (Optional) The throughput of the MongoDB collection (RU/s). Must be set in increments of `100`. The default and minimum value is `400`.
+* `throughput` - (Optional) The throughput of the MongoDB collection (RU/s). Must be set in increments of `100`. The minimum value is `400`. This must be set upon database creation otherwise it cannot be updated without a manual terraform destroy-apply.
```
katbyte: see my other comment

azurerm/resource_arm_cosmosdb_mongo_database.go
```diff
@@ -49,11 +52,17 @@ func resourceArmCosmosDbMongoDatabase() *schema.Resource {
...
+				Type:         schema.TypeInt,
+				Optional:     true,
+				ValidateFunc: validate.CosmosThroughput,
+			},
 		},
 	}
 }
 
-func resourceArmCosmosDbMongoDatabaseCreate(d *schema.ResourceData, meta interface{}) error {
+func resourceArmCosmosDbMongoDatabaseCreateUpdate(d *schema.ResourceData, meta interface{}) error {
```
katbyte: Could we use separate create and update functions? i think it'll simplify things a tad

katbyte:
hi @e96wic,

I am looking into finishing this and it doesn't appear i have permission to push to your branch. Are you still working on this or should i go ahead and open a new PR?

katbyte:
Hey @e96wic, i've opened #5116 that address the comments i've made so we can get this merged. As such i am going to close this in favour of it, thanks for all your work on this change!

PR #4625, opened 2019-10-15
Data source/loadbalancer backend address pool/backend_ip_configurations to be computed

marioanton:
- Adding compute to attribute backend_ip_configurations
- Minor lint issues

### DIFF
--- a/azurerm/data_source_loadbalancer_backend_address_pool.go
+++ b/azurerm/data_source_loadbalancer_backend_address_pool.go
@@ -48,7 +48,7 @@ func dataSourceArmLoadBalancerBackendAddressPool() *schema.Resource {
 }
 
 func dataSourceArmLoadBalancerBackendAddressPoolRead(d *schema.ResourceData, meta interface{}) error {
-	loadBalancerId := d.Get("loadbalancer_id").(string)
+	loadBalancerID := d.Get("loadbalancer_id").(string)
 	name := d.Get("name").(string)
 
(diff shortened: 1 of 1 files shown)


leewilson86:
@tombuildsstuff 

This PR is a fix for a bug my colleague found with the following PR which I raised a few days ago (https://github.com/terraform-providers/terraform-provider-azurerm/pull/4605).

It fixes this error:

```
Resource 'data.azurerm_lb_backend_address_pool.beap' does not have attribute 'backend_ip_configurations' for variable 'data.azurerm_lb_backend_address_pool.beap.backend_ip_configurations'
```

azurerm/data_source_loadbalancer_backend_address_pool.go
```diff
@@ -33,6 +33,7 @@ func dataSourceArmLoadBalancerBackendAddressPool() *schema.Resource {
 			"backend_ip_configurations": {
 				Type:     schema.TypeList,
 				Optional: true,
```
mbfrahry: Any attribute that isn't used to get the resource should remove the `Required/Optional: true` tag in favor of `Computed: true`

mariobede: Thanks, will sort that out.

mariobede:
Changes commited and tested on my side.

marioanton:
Changes commited.

PR #4639, opened 2019-10-17
azurerm_firewall: allow multiple `ip_configuration` blocks

### LINKED ISSUE #4045: azurerm_firewall: support multiple ip_configuration blocks
### Community Note

### Description

Currently it is only possible to have one ip_configuration block per azurerm_firewall resource where Azure Firewall actually supports multiple IP configurations. Having multiple IP addresses associated with a Firewall is essential...

Is this possible to enhance the module and add a feature to enable the multiple ip_configuration blocks?

* azurerm_firewall

```hcl
```

### References

* #0000

houkms:
Allow `enable_ip_configuration` to accept multiple ip_configurations.

(fixes #4045)

### DIFF
--- a/azurerm/data_source_firewall.go
+++ b/azurerm/data_source_firewall.go
@@ -94,8 +94,10 @@ func dataSourceArmFirewallRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if props := read.AzureFirewallPropertiesFormat; props != nil {
-		if err := d.Set("ip_configuration", flattenArmFirewallIPConfigurations(props.IPConfigurations)); err != nil {
-			return fmt.Errorf("Error setting `ip_configuration`: %+v", err)
+		if props.IPConfigurations != nil {
+			if err := d.Set("ip_configuration", flattenArmFirewallIPConfigurations(props.IPConfigurations)); err != nil {
+				return fmt.Errorf("Error setting `ip_configuration`: %+v", err)
+			}
 		}
 	}
 
--- a/azurerm/data_source_firewall.go
+++ b/azurerm/data_source_firewall.go
@@ -94,10 +94,8 @@ func dataSourceArmFirewallRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if props := read.AzureFirewallPropertiesFormat; props != nil {
-		if props.IPConfigurations != nil {
-			if err := d.Set("ip_configuration", flattenArmFirewallIPConfigurations(props.IPConfigurations)); err != nil {
-				return fmt.Errorf("Error setting `ip_configuration`: %+v", err)
-			}
+		if err := d.Set("ip_configuration", flattenArmFirewallIPConfigurations(props.IPConfigurations)); err != nil {
+			return fmt.Errorf("Error setting `ip_configuration`: %+v", err)
 		}
 	}
 
--- a/azurerm/data_source_firewall.go
+++ b/azurerm/data_source_firewall.go
@@ -32,7 +32,7 @@ func dataSourceArmFirewall() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
 
-			"ip_configuration": {
+			"ip_configurations": {
 				Type:     schema.TypeList,
 				Computed: true,
 				MaxItems: 1,
@@ -96,8 +96,8 @@ func dataSourceArmFirewallRead(d *schema.ResourceData, meta interface{}) error {
 
 	if props := read.AzureFirewallPropertiesFormat; props != nil {
 		ipConfigs := flattenArmFirewallIPConfigurations(props.IPConfigurations)
-		if err := d.Set("ip_configuration", ipConfigs); err != nil {
-			return fmt.Errorf("Error setting `ip_configuration`: %+v", err)
+		if err := d.Set("ip_configurations", ipConfigs); err != nil {
+			return fmt.Errorf("Error setting `ip_configurations`: %+v", err)
 		}
 	}
 
--- a/azurerm/data_source_firewall.go
+++ b/azurerm/data_source_firewall.go
@@ -32,7 +32,7 @@ func dataSourceArmFirewall() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
 
-			"ip_configurations": {
+			"ip_configuration": {
 				Type:     schema.TypeList,
 				Computed: true,
 				Elem: &schema.Resource{
@@ -94,9 +94,8 @@ func dataSourceArmFirewallRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if props := read.AzureFirewallPropertiesFormat; props != nil {
-		ipConfigs := flattenArmFirewallIPConfigurations(props.IPConfigurations)
-		if err := d.Set("ip_configurations", ipConfigs); err != nil {
-			return fmt.Errorf("Error setting `ip_configurations`: %+v", err)
+		if err := d.Set("ip_configuration", flattenArmFirewallIPConfigurations(props.IPConfigurations)); err != nil {
+			return fmt.Errorf("Error setting `ip_configuration`: %+v", err)
 		}
 	}
 
--- a/azurerm/data_source_firewall.go
+++ b/azurerm/data_source_firewall.go
@@ -35,7 +35,6 @@ func dataSourceArmFirewall() *schema.Resource {
 			"ip_configurations": {
 				Type:     schema.TypeList,
 				Computed: true,
-				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"name": {
--- a/azurerm/data_source_firewall_test.go
+++ b/azurerm/data_source_firewall_test.go
@@ -21,8 +21,8 @@ func TestAccDataSourceAzureRMFirewall_basic(t *testing.T) {
 			{
 				Config: testAccDataSourceFirewall_basic(rInt, location),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(dataSourceName, "ip_configuration.0.name", "configuration"),
-					resource.TestCheckResourceAttrSet(dataSourceName, "ip_configuration.0.private_ip_address"),
+					resource.TestCheckResourceAttr(dataSourceName, "ip_configurations.0.name", "configuration"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "ip_configurations.0.private_ip_address"),
 				),
 			},
 		},
@@ -63,7 +63,7 @@ resource "azurerm_firewall" "test" {
   location            = "${azurerm_resource_group.test.location}"
   resource_group_name = "${azurerm_resource_group.test.name}"
 
-  ip_configuration {
+  ip_configurations {
     name                 = "configuration"
     subnet_id            = "${azurerm_subnet.test.id}"
     public_ip_address_id = "${azurerm_public_ip.test.id}"
--- a/azurerm/data_source_firewall_test.go
+++ b/azurerm/data_source_firewall_test.go
@@ -21,8 +21,8 @@ func TestAccDataSourceAzureRMFirewall_basic(t *testing.T) {
 			{
 				Config: testAccDataSourceFirewall_basic(rInt, location),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(dataSourceName, "ip_configurations.0.name", "configuration"),
-					resource.TestCheckResourceAttrSet(dataSourceName, "ip_configurations.0.private_ip_address"),
+					resource.TestCheckResourceAttr(dataSourceName, "ip_configuration.0.name", "configuration"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "ip_configuration.0.private_ip_address"),
 				),
 			},
 		},
@@ -63,7 +63,7 @@ resource "azurerm_firewall" "test" {
   location            = "${azurerm_resource_group.test.location}"
   resource_group_name = "${azurerm_resource_group.test.name}"
 
-  ip_configurations {
+  ip_configuration {
     name                 = "configuration"
     subnet_id            = "${azurerm_subnet.test.id}"
     public_ip_address_id = "${azurerm_public_ip.test.id}"
--- a/azurerm/resource_arm_firewall.go
+++ b/azurerm/resource_arm_firewall.go
@@ -419,23 +419,29 @@ func validateFirewallConfigurationSettings(d *schema.ResourceData) error {
 	subnetNumber := 0
 	hasInternal := false
 	hasPublic := false
+
 	for _, configRaw := range configs {
 		data := configRaw.(map[string]interface{})
 		if subnet, exist := data["subnet_id"].(string); exist && subnet != "" {
 			subnetNumber++
 		}
+
 		if internal, exist := data["internal_public_ip_address_id"].(string); exist && internal != "" {
 			hasInternal = true
 		}
+
 		if public, exist := data["public_ip_address_id"].(string); exist && public != "" {
 			hasPublic = true
 		}
 	}
+
 	if subnetNumber != 1 {
 		return fmt.Errorf(`The "ip_configuration" is invalid, %d "subnet_id" have been set, one "subnet_id" should be set among all "ip_configuration" blocks`, subnetNumber)
 	}
+
 	if hasInternal && hasPublic {
 		return fmt.Errorf(`The "ip_configuration" is invalid, both "public_ip_address_id" and "internal_public_ip_address_id" have been set, all defined "ip_configuration" blocks must use the same attribute name.`)
 	}
+
 	return nil
 }
--- a/azurerm/resource_arm_firewall.go
+++ b/azurerm/resource_arm_firewall.go
@@ -97,6 +97,10 @@ func resourceArmFirewallCreateUpdate(d *schema.ResourceData, meta interface{}) e
 	ctx, cancel := timeouts.ForCreateUpdate(meta.(*ArmClient).StopContext, d)
 	defer cancel()
 
+	if err := validateFirewallConfigurationSettings(d); err != nil {
+		return fmt.Errorf("Error creating Firewall %q (Resource Group %q): %+v", d.Get("name").(string), d.Get("resource_group_name").(string), err)
+	}
+
 	log.Printf("[INFO] preparing arguments for AzureRM Azure Firewall creation")
 
 	name := d.Get("name").(string)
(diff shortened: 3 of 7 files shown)


azurerm/data_source_firewall.go
```diff
@@ -32,10 +32,9 @@ func dataSourceArmFirewall() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
 
-			"ip_configuration": {
+			"ip_configurations": {
```
WodansSon: Do you think we should change the name? Wouldn't this break everyone that is already using this resource? I think just removing the `MaxItems` should be enough here, WDYT?

azurerm/data_source_firewall.go
```diff
@@ -96,8 +95,8 @@ func dataSourceArmFirewallRead(d *schema.ResourceData, meta interface{}) error {
 
 	if props := read.AzureFirewallPropertiesFormat; props != nil {
 		ipConfigs := flattenArmFirewallIPConfigurations(props.IPConfigurations)
```
WodansSon:
```suggestion
	if err := d.Set("ip_configuration", flattenArmFirewallIPConfigurations(props.IPConfigurations)); err != nil { 
		return fmt.Errorf("Error setting `ip_configurations`: %+v", err) 
	}
```

azurerm/data_source_firewall_test.go
```diff
@@ -21,8 +21,8 @@ func TestAccDataSourceAzureRMFirewall_basic(t *testing.T) {
 			{
 				Config: testAccDataSourceFirewall_basic(rInt, location),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(dataSourceName, "ip_configuration.0.name", "configuration"),
-					resource.TestCheckResourceAttrSet(dataSourceName, "ip_configuration.0.private_ip_address"),
+					resource.TestCheckResourceAttr(dataSourceName, "ip_configurations.0.name", "configuration"),
```
WodansSon: Please revert all of these references back to `ip_configuration`

azurerm/resource_arm_firewall.go
```diff
@@ -62,24 +62,22 @@ func resourceArmFirewall() *schema.Resource {
...
-							Computed:      true,
-							ValidateFunc:  azure.ValidateResourceID,
-							Deprecated:    "This field has been deprecated. Use `public_ip_address_id` instead.",
-							ConflictsWith: []string{"ip_configuration.0.public_ip_address_id"},
+							Type:         schema.TypeString,
+							Optional:     true,
+							Computed:     true,
+							ValidateFunc: azure.ValidateResourceID,
+							Deprecated:   "This field has been deprecated. Use `public_ip_address_id` instead.",
 						},
```
WodansSon: Why has the `ConflictsWith` been removed here?

azurerm/resource_arm_firewall.go
```diff
@@ -62,24 +62,22 @@ func resourceArmFirewall() *schema.Resource {
...
-							Type:          schema.TypeString,
-							Optional:      true,
-							Computed:      true,
-							ValidateFunc:  azure.ValidateResourceID,
-							ConflictsWith: []string{"ip_configuration.0.internal_public_ip_address_id"},
+							Type:         schema.TypeString,
+							Optional:     true,
+							Computed:     true,
+							ValidateFunc: azure.ValidateResourceID,
 						},
```
WodansSon: Same here, Why has the `ConflictsWith` been removed here?

website/docs/d/firewall.html.markdown
```diff
@@ -36,11 +36,11 @@ The following attributes are exported:
 
 * `id` - The Resource ID of the Azure Firewall.
 
-* `ip_configuration` - A `ip_configuration` block as defined below.
+* `ip_configurations` - A `ip_configurations` block as defined below.
```
WodansSon: We will need to revert all references back to be `ip_configuration`, WDYT?

azurerm/data_source_firewall.go
houkms: Agree and reverted.

azurerm/data_source_firewall.go
houkms: yes, but it's their original coding style. Anyway, it has been modified by your suggestion both in resource and datasource.

azurerm/data_source_firewall_test.go
houkms: reverted.

website/docs/d/firewall.html.markdown
houkms: Agree and reverted.

azurerm/resource_arm_firewall.go
houkms: yeah, it's a hard decision. 

I don't think we should use `"ip_configuration.0.public_ip_address_id"` here anymore since _ip_configuration_ is a list now.

But I couldn't find a method to deal with the situation, because it needs every element of _ip_configuration_ to check conflicts between the two sub-properties. 

can you offer some suggestions?

azurerm/resource_arm_firewall.go
houkms: same.

houkms:
@WodansSon 

Thanks for your suggestions. All the _ip_configurations_ are reverted to _ip_configuration_ now. 
I addressed the reason about `ConflictsWith` inline, waiting for your advice.

azurerm/resource_arm_firewall.go
```diff
@@ -211,8 +209,7 @@ func resourceArmFirewallRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if props := read.AzureFirewallPropertiesFormat; props != nil {
-		ipConfigs := flattenArmFirewallIPConfigurations(props.IPConfigurations)
-		if err := d.Set("ip_configuration", ipConfigs); err != nil {
+		if err := d.Set("ip_configuration", flattenArmFirewallIPConfigurations(props.IPConfigurations)); err != nil {
 			return fmt.Errorf("Error setting `ip_configuration`: %+v", err)
 		}
```
WodansSon: We should add the same nil check here as in the data source above, WDYT?

azurerm/resource_arm_firewall.go
```diff
@@ -62,24 +62,22 @@ func resourceArmFirewall() *schema.Resource {
 						},
 						"subnet_id": {
 							Type:         schema.TypeString,
-							Required:     true,
+							Optional:     true,
```
WodansSon: Is there a reason for changing this to `Optional`? I think this should still be `Required`, WDYT?

azurerm/resource_arm_firewall.go
WodansSon: I would keep the `ConflictsWith` since that will keep the end user mixing and matching the two properties within a single `ip_configuration`. Then I would creating an `ip_configuration` validation function that:
1) Create two bool vars `hasInternal` and `hasPublic`.
2) Loops over all of the `ip_configuration`'s in the schema and look for `internal_public_ip_address_id` and `public_ip_address_id` properties.
a) If it finds an `internal_public_ip_address_id` set the `hasInternal` to `true` and continue the loop.
b) If it finds a `public_ip_address_id` set the `hasPublic` to `true` and continue the loop.
3) Once the code has iterated over all of the `ip_configurations` items in the schema if `hasInternal` is `true` and `hasPublic` is also `true` return an error saying that the `ip_configuration` is invalid and that they must use either one or the other in all defined `ip_configurations` properties.

I would then reference this new function in the `resourceArmFirewallCreateUpdate` function like:

```
func resourceArmFirewallCreateUpdate(d *schema.ResourceData, meta interface{}) error {
	client := meta.(*ArmClient).Network.AzureFirewallsClient
	ctx, cancel := timeouts.ForCreateUpdate(meta.(*ArmClient).StopContext, d)
	defer cancel()

	if err := network.ValidateFirewallConfigurationSettings(d); err != nil {
		return fmt.Errorf(`Error creating Firewall %q (Resource Group %q): the "ip_configuration" is invalid, both "public_ip_address_id" and "internal_public_ip_address_id" have been set, all defined "ip_configuration" blocks must use the same attribute name.`, d.Get("name").(string), d.Get("resource_group_name").(string))
	}
```

azurerm/resource_arm_firewall.go
WodansSon: See above...

azurerm/data_source_firewall.go
```diff
@@ -95,8 +94,7 @@ func dataSourceArmFirewallRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if props := read.AzureFirewallPropertiesFormat; props != nil {
-		ipConfigs := flattenArmFirewallIPConfigurations(props.IPConfigurations)
-		if err := d.Set("ip_configuration", ipConfigs); err != nil {
+		if err := d.Set("ip_configuration", flattenArmFirewallIPConfigurations(props.IPConfigurations)); err != nil {
 			return fmt.Errorf("Error setting `ip_configuration`: %+v", err)
 		}
```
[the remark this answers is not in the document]
houkms: Added

azurerm/resource_arm_firewall.go
houkms: Added.

azurerm/resource_arm_firewall.go
houkms: Thanks for the suggestions. Modified according to your comments.

azurerm/resource_arm_firewall.go
houkms: Yes, the reason is RestApi requires only one `subnet_id` in `ip_configuration`. 0 or more than 1 `subnet_id` will cause errors.

Now, my solution is to change the `subnet_id` to `Optional` and check it in our new introduced `validateFirewallConfigurationSettings` function to make sure there is only one `subnet_id ` in `ip_configuration`.

I have considered other ways like move `subnet_id` outside from `ip_configuration`, but it may make users confused and also bring breaking changes.

What do you think?

houkms:
@WodansSon Thanks for your suggestions. I added the validate function for `ip_configuration` and addressed the reason for the change of `subnet_id`.

azurerm/resource_arm_firewall.go
tombuildsstuff: the documentation needs to be updated to match this being Optional too: https://github.com/terraform-providers/terraform-provider-azurerm/pull/4639/files#diff-0f2b98bda27ee42e45f9395c8f830b0dR77

azurerm/resource_arm_firewall.go
```diff
@@ -407,3 +413,35 @@ func validateAzureFirewallSubnetName(v interface{}, k string) (warnings []string
...
+	configs := d.Get("ip_configuration").([]interface{})
+	subnetNumber := 0
+	hasInternal := false
+	hasPublic := false
+
+	for _, configRaw := range configs {
+		data := configRaw.(map[string]interface{})
+		if subnet, exist := data["subnet_id"].(string); exist && subnet != "" {
+			subnetNumber++
+		}
```
tombuildsstuff: I'm not sure what the purpose of this is.. I think this can be removed?

azurerm/resource_arm_firewall.go
```diff
@@ -297,7 +300,7 @@ func expandArmFirewallIPConfigurations(d *schema.ResourceData) (*[]network.Azure
 	subnetNamesToLock := make([]string, 0)
 	virtualNetworkNamesToLock := make([]string, 0)
 
-	for _, configRaw := range configs {
+	for index, configRaw := range configs {
```
tombuildsstuff: we should update this expand method to return an object; such as:

```
type expandedFirewallIPConfigurations struct {
  ipConfigurations []network.AzureFirewallIPConfiguration
  virtualNetworkNames []string
  subnetNames []string
}
```

which would allow us to make this function:

```
func expandArmFirewallIPConfigurations(input []interface{}) (*expandedFirewallIPConfigurations, error) {
```

where the `validate` logic above can be merged into this

azurerm/resource_arm_firewall.go
```diff
@@ -407,3 +413,35 @@ func validateAzureFirewallSubnetName(v interface{}, k string) (warnings []string
...
+		}
+	}
+
+	if subnetNumber != 1 {
+		return fmt.Errorf(`The "ip_configuration" is invalid, %d "subnet_id" have been set, one "subnet_id" should be set among all "ip_configuration" blocks`, subnetNumber)
+	}
+
+	if hasInternal && hasPublic {
+		return fmt.Errorf(`The "ip_configuration" is invalid, both "public_ip_address_id" and "internal_public_ip_address_id" have been set, all defined "ip_configuration" blocks must use the same attribute name.`)
+	}
```
tombuildsstuff: shouldn't this be inside the for loop?

azurerm/resource_arm_firewall.go
```diff
@@ -407,3 +413,35 @@ func validateAzureFirewallSubnetName(v interface{}, k string) (warnings []string
 
 	return warnings, errors
 }
+
+func validateFirewallConfigurationSettings(d *schema.ResourceData) error {
+	configs := d.Get("ip_configuration").([]interface{})
+	subnetNumber := 0
+	hasInternal := false
+	hasPublic := false
```
tombuildsstuff: I think both of these should be defined in the for loop?

azurerm/data_source_firewall.go
```diff
@@ -95,9 +94,10 @@ func dataSourceArmFirewallRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if props := read.AzureFirewallPropertiesFormat; props != nil {
-		ipConfigs := flattenArmFirewallIPConfigurations(props.IPConfigurations)
-		if err := d.Set("ip_configuration", ipConfigs); err != nil {
-			return fmt.Errorf("Error setting `ip_configuration`: %+v", err)
+		if props.IPConfigurations != nil {
```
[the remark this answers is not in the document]
houkms: removed.

azurerm/resource_arm_firewall.go
```diff
@@ -99,6 +97,10 @@ func resourceArmFirewallCreateUpdate(d *schema.ResourceData, meta interface{}) e
 	ctx, cancel := timeouts.ForCreateUpdate(meta.(*ArmClient).StopContext, d)
 	defer cancel()
 
+	if err := validateFirewallConfigurationSettings(d); err != nil {
+		return fmt.Errorf("Error creating Firewall %q (Resource Group %q): %+v", d.Get("name").(string), d.Get("resource_group_name").(string), err)
```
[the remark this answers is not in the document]
houkms: moved.

azurerm/resource_arm_firewall.go
```diff
@@ -211,9 +213,10 @@ func resourceArmFirewallRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if props := read.AzureFirewallPropertiesFormat; props != nil {
-		ipConfigs := flattenArmFirewallIPConfigurations(props.IPConfigurations)
-		if err := d.Set("ip_configuration", ipConfigs); err != nil {
-			return fmt.Errorf("Error setting `ip_configuration`: %+v", err)
+		if props.IPConfigurations != nil {
```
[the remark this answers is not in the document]
houkms: removed.

azurerm/resource_arm_firewall.go
```diff
@@ -308,36 +311,39 @@ func expandArmFirewallIPConfigurations(d *schema.ResourceData) (*[]network.Azure
...
-		subnetName := subnetID.Path["subnets"]
-		virtualNetworkName := subnetID.Path["virtualNetworks"]
-
-		if !sliceContainsValue(subnetNamesToLock, subnetName) {
-			subnetNamesToLock = append(subnetNamesToLock, subnetName)
-		}
-
-		if !sliceContainsValue(virtualNetworkNamesToLock, virtualNetworkName) {
-			virtualNetworkNamesToLock = append(virtualNetworkNamesToLock, virtualNetworkName)
+			return nil, nil, nil, fmt.Errorf("one of `ip_configuration.%s.internal_public_ip_address_id` or `ip_configuration.%s.public_ip_address_id` must be set", strconv.Itoa(index), strconv.Itoa(index))
```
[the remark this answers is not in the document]
houkms: modified.

azurerm/resource_arm_firewall.go
```diff
@@ -407,3 +413,35 @@ func validateAzureFirewallSubnetName(v interface{}, k string) (warnings []string
...
+		if internal, exist := data["internal_public_ip_address_id"].(string); exist && internal != "" {
+			hasInternal = true
+		}
+
+		if public, exist := data["public_ip_address_id"].(string); exist && public != "" {
+			hasPublic = true
+		}
+	}
+
+	if subnetNumber != 1 {
```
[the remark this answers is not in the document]
houkms: I am afraid we can not.

For `internal_public_ip_address_id` and `public_ip_address_id`
What we need here is to make sure only one of them can appear in all of the `ip_configuration` elements, not only in one element.

For `subnetNumber`
The constraint is that only one `subnet_id` is required and allowed among all the elements of  `ip_configuration` ( Azure RestApi will throw an error if multiple `subnet_id` exist or there is no `subnet_id`)

azurerm/resource_arm_firewall.go
houkms: For subnetNumber:

The constraint is that only one subnet_id is required and allowed among all the elements of ip_configuration ( Azure RestApi will throw an error if multiple subnet_id exist or there is no subnet_id)

website/docs/r/firewall.html.markdown
```diff
@@ -78,6 +78,8 @@ A `ip_configuration` block supports the following:
 
 -> **NOTE** The Subnet used for the Firewall must have the name `AzureFirewallSubnet` and the subnet mask must be at least `/26`.
 
+-> **NOTE** The `subnet_id` is shared among multiple `ip_configuration` blocks, make sure there is one and only one `subnet_id` in all `ip_configuration` blocks.
```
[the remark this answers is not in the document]
houkms: modified.

azurerm/resource_arm_firewall.go
houkms: We are already in `expandArmFirewallIPConfigurations`.

azurerm/resource_arm_firewall.go
```diff
@@ -407,3 +413,35 @@ func validateAzureFirewallSubnetName(v interface{}, k string) (warnings []string
...
 	return warnings, errors
 }
+
+func validateFirewallConfigurationSettings(d *schema.ResourceData) error {
+	configs := d.Get("ip_configuration").([]interface{})
+	subnetNumber := 0
+	hasInternal := false
+	hasPublic := false
+
+	for _, configRaw := range configs {
```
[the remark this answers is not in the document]
houkms: reset to be `_`, as we don't use it any more.

houkms:
@WodansSon I addressed the comments, would you please help review the codes?

houkms:
Hi @katbyte 

It's a property renaming. I think it's a way to deal with breaking change, which was already introduced before this PR.

timja:
I missed this PR and implemented it myself as well, I implemented pretty much the same thing except there's more validation here, and I implemented a test.

I've submitted the test as a PR to your fork @houkms, if you're interested in taking it to give more confidence in this PR:
https://github.com/houkms/terraform-provider-azurerm/pull/1

Either way would love to get this in,
ping @katbyte

houkms:
@timja Thanks for your test codes. It looks good to me. Lets wait for the reviews.

houkms:
The tests fail because my validation function doesn't allow  `internal_public_ip_address_id` and `public_ip_address_id` both exist before creating or updating.

While in the Read function, both of them will be set (see [here)](https://github.com/terraform-providers/terraform-provider-azurerm/pull/4639/files#diff-94986bd1e9d127462440622c4196c315L374-L376) if we use any one of them, which will make they both appear in tfstate thus result in validation error in the next time we call create/update.

CONCLUSION

- I removed the validation of the two properties (follow an existing [resource](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_scheduler_job_collection.go#L116-L128)), and the tests work well now.

PR #4645, opened 2019-10-17
[azurerm_key_vault_certificate] - switches the `emails`, `dns_names `, `upns` of the `subject_alternative_names` property to use `TypeSet`

hbuckle:
I noticed when importing an existing certificate these properties did not come back in the same order as the configuration, so Terraform wanted to recreate the certificate.


hbuckle:
@katbyte yes of course, now updated

Issue #3843, opened 2019-07-12
ServiceHealth value error


funnybirdman:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.11.14

### Affected Resource(s)

* `azurerm_monitor_activity_log_alert`

### Terraform Configuration Files

```resource "azurerm_resource_group" "main" {
  name     = "example-resources"
  location = "West US"
}

resource "azurerm_monitor_action_group" "main" {
  name                = "example-actiongroup"
  resource_group_name = "${azurerm_resource_group.main.name}"
  short_name          = "p0action"

  webhook_receiver {
    name        = "callmyapi"
    service_uri = "http://example.com/alert"
  }
}

resource "azurerm_storage_account" "to_monitor" {
  name                     = "examplesa"
  resource_group_name      = "${azurerm_resource_group.main.name}"
  location                 = "${azurerm_resource_group.main.location}"
  account_tier             = "Standard"
  account_replication_type = "GRS"
}

resource "azurerm_monitor_activity_log_alert" "main" {
  name                = "example-activitylogalert"
  resource_group_name = "${azurerm_resource_group.main.name}"
  scopes              = ["${azurerm_resource_group.main.id}"]
  description         = "This alert will monitor a specific storage account updates."

   criteria {
    category       = "Service Health" #<--- this is the problem
  }

  action {
    action_group_id = "${azurerm_monitor_action_group.main.id}"

    webhook_properties = {
      from = "terraform"
    }
  }
}
```

### Debug Output
NA it succeeds to create. There is no error reporting on Azure side it seems.

### Panic Output
NA

### Expected Behavior
In the created alert the condition should equal `ServiceHealth`
```
"condition": {
      "allOf": [
        {
          "containsAny": null,
          "equals": "ServiceHealth",
          "field": "category",
          "odata.type": null
        }
      ],
      "odata.type": null
    },
```

### Actual Behavior

An alert was created with `Service Health`.
```
 "condition": {
     "allOf": [
       {
         "containsAny": null,
         "equals": "Service Health",
[...]

PR #4646, opened 2019-10-17
Fixing issue #3843

tylerd:
Space in "Service Health" category causes 500 error when executing REST call

### DIFF
--- a/azurerm/resource_arm_monitor_activity_log_alert.go
+++ b/azurerm/resource_arm_monitor_activity_log_alert.go
@@ -75,7 +75,7 @@ func resourceArmMonitorActivityLogAlert() *schema.Resource {
 								"Policy",
 								"Recommendation",
 								"Security",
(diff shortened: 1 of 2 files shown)


azurerm/resource_arm_monitor_activity_log_alert.go
```diff
@@ -75,7 +75,7 @@ func resourceArmMonitorActivityLogAlert() *schema.Resource {
 								"Policy",
 								"Recommendation",
 								"Security",
-								"Service Health",
+								"ServiceHealth",
```
tombuildsstuff: could we [also update this in the documentation](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/monitor_activity_log_alert.html.markdown#L86)?

tylerd: Done

PR #4663, opened 2019-10-19
Added Terraform Azure Partner ID

echuvyrov:
Added Terraform Azure Partner GUID

### DIFF
--- a/azurerm/internal/common/client_options.go
+++ b/azurerm/internal/common/client_options.go
@@ -65,13 +65,13 @@ func setUserAgent(client *autorest.Client, tfVersion, partnerID string) {
 
 	// only one pid can be interpreted currently
 	//  hence, send partner Id if present
-	if partnerID != "" {
-		client.UserAgent = fmt.Sprintf("%s pid-%s", client.UserAgent, partnerID)
-	} else {
-		// otherwise, if pid is not present in config files, send Terraform GUID
-		terraformGUID := "222c6c49-1b0a-5959-a213-6608f9eb8820"
-		client.UserAgent = fmt.Sprintf("%s pid-%s", client.UserAgent, terraformGUID)
+	//  if partner Id isn't present, send Terraform GUID
+	if partnerID == "" {
+
+		// Microsoft’s Terraform Partner ID is a specific GUID pasted below
+		partnerID = "222c6c49-1b0a-5959-a213-6608f9eb8820"
 	}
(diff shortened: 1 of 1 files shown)


azurerm/internal/common/client_options.go
```diff
@@ -63,8 +63,14 @@ func setUserAgent(client *autorest.Client, tfVersion, partnerID string) {
...
 	}
 
+	// only one pid can be interpreted currently
+	//  hence, send partner Id if present
 	if partnerID != "" {
 		client.UserAgent = fmt.Sprintf("%s pid-%s", client.UserAgent, partnerID)
+	} else {
+		// otherwise, if pid is not present in config files, send Terraform GUID
+		terraformGUID := "222c6c49-1b0a-5959-a213-6608f9eb8820"
+		client.UserAgent = fmt.Sprintf("%s pid-%s", client.UserAgent, terraformGUID)
```
StefanSchoof: Why not use only a single Sprintf after the if else (with a local variable), to  eliminate the code duplication?

katbyte:
Hi @echuvyrov,

Could you please explain the rationale for this PR & change? What does that GUID represent?

azurerm/internal/common/client_options.go
katbyte:
```suggestion
if partnerID = "" {
	partnerID := "222c6c49-1b0a-5959-a213-6608f9eb8820"
}
client.UserAgent = fmt.Sprintf("%s pid-%s", client.UserAgent, partnerID)
```

schuettecarsten:
Sorry, this change is absolutely unacceptable.
If no provider_id is specified, then no provider_id should be sent to Azure.

tesharp:
To ask same as @katbyte 
Could you please explain the rationale for this PR & change?
Why has this been added as default?

OliverCole:
@paultyng I understand you want to lock #4747, but I think there is the unresolved question of why @katbyte’s reasonable query was either ignored or answered out of band?

echuvyrov:
Apologies for not responding to this right away.

As [Mitchell mentioned in his response](https://news.ycombinator.com/item?id=21388951), @katbyte and I work together on the Terraform provider for Azure (although she does all the heavy lifting). We have weekly syncs where we discuss issues that our joint customers bring to us and pending PRs. We discussed the Partner ID integration during one of the calls and did not loop back around to respond on the PR what had been discussed. That was a major oversight that we will work to avoid in the future.

OliverCole:
@echuvyrov Thanks for clarifying.

PR #4709, opened 2019-10-24
storage: caching all of the account information

tombuildsstuff:
This PR introduces more extensive caching for Storage Accounts - which allows us to work around a low rate limit due to an unperforment API; which should improve the performance when provisioning large quantities of storage related resources


tombuildsstuff:
Tests pass:

Screenshot 2019-10-28 at 17 55 52

PR #4737, opened 2019-10-28
Remove all expired public preview notices

ashemedai:
MariaDB GA: https://azure.microsoft.com/en-gb/blog/announcing-the-general-availability-of-azure-database-for-mariadb/Public
IP prefix: https://azure.microsoft.com/en-us/updates/public-ip-prefix-general-availability/
Shared Image Gallery: https://azure.microsoft.com/en-us/updates/shared-image-gallery-generally-available/

### DIFF
--- a/website/docs/d/shared_image.html.markdown
+++ b/website/docs/d/shared_image.html.markdown
@@ -11,8 +11,6 @@ description: |-
 
 Use this data source to access information about an existing Shared Image within a Shared Image Gallery.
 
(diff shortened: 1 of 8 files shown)


ashemedai:
@tombuildsstuff Sorry, forgot the data sources. So here's some additional clean ups.

Potentially we might be able to get rid off the notice in `kubernetes_cluster` as well, but not 100% sure. The Azure Firewall at least is also out of public preview.

PR #4742, opened 2019-10-29
upgrade sql version from 2015-05-01-preview to 2017-03-01-preview

yupwei68:


mbfrahry:
Thanks for opening this @yupwei68. It should make it into the next release

Issue #4563, opened 2019-10-08
storage_account - ATP not available in Azure Germany


sigv:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

```plain
Terraform v0.11.14
+ provider.azurerm v1.35.0
```

### Affected Resource(s)

* `azurerm_storage_account`

### Terraform Configuration Files

```hcl
resource "azurerm_resource_group" "primary" {
  name     = "ABC"
  location = "Germany Central"
}

resource "azurerm_storage_account" "storagestandard" {
  name                              = "destoragestandard"
  resource_group_name               = "${azurerm_resource_group.primary.name}"
  location                          = "${azurerm_resource_group.primary.location}"
  account_kind                      = "StorageV2"
  account_tier                      = "Standard"
  account_replication_type          = "LRS"
  enable_https_traffic_only         = true
  enable_advanced_threat_protection = false
}
```

### Error Output

```plain
Error: Error refreshing state: 1 error occurred:
        * azurerm_storage_account.storagestandard: 1 error occurred:
        * azurerm_storage_account.storagestandard: azurerm_storage_account.storagestandard: Error reading the advanced threat protection settings of AzureRM Storage Account "destoragestandard": security.AdvancedThreatProtectionClient#Get: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="InvalidResourceNamespace" Message="The resource namespace 'Microsoft.Security' is invalid."
```

### Expected Behavior

`enable_advanced_threat_protection` is ignored as not available in Azure Germany.

### Actual Behavior

Any action involving the storage account fails with a `HTTP/2.0 404 Not Found` on the AzureRM Response for the APT resource.

### Steps to Reproduce

`terraform apply` / `terraform refresh -target=azurerm_storage_account.storagestandard`

### Important Factoids

Specific to Azure Germany.

Our environments in regular `AzureCloud` do not have trouble with this.

### References

* #3920 APT not available in Azure Government Cloud.

sigv:
PR submitted based on how #3947 was done.

PR #4746, opened 2019-10-29
storage_account: Fix Advanced Threat Protection matching for Azure Germany

sigv:
Although the error message returned from Azure Germany Cloud is precisely "The resource namespace 'Microsoft.Security' is invalid." the error message we are seeing in `r/storage_account` has other contents added on it, so we cannot do an equals check directly. The ATP fix was being pushed as v1.36.0, but this change provides the actual resolution.

Ref #4563, #4564.


sigv:
@katbyte Well, I did a bad assumption that `msg` was the same as debug log's `Message` -- as a result, v1.36.0 did not ship the fix for Azure Germany as in #4564. Properly tested this one locally and can run a `terraform refresh`.

Issue #87, opened 2017-06-13
SQL Server: support for Transparent Data Encryption


hashibot:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

_This issue was originally opened by @anniehedgpeth as hashicorp/terraform#14595. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
0.9.5

### Affected Resource(s)
- azurerm_sql_database

The `azurerm_sql_database` is missing the settings for [Transparent Data Encryption](https://docs.microsoft.com/en-us/sql/relational-databases/security/encryption/transparent-data-encryption-with-azure-sql-database).

achandmsft:
As we haven't heard any requests for this enhancement yet, I will archive this issue. We can reconsider if there is any customer interest around the same.

keithhopkins-gpc:
How to get this re-opened?  We have a requirement for TDE in SQL Database (PaaS).

RustyF:
I too have this requirement.   As an aside, newbie question, is there a workaround that allows one to script TDE inside terraform or would I have to manually add TDE afterwards?

simonlegeek:
I agree this would be very useful otherwise we would have to rely on a script running as a provisioner.

jjgrayston:
+1. Amazed it hasn't been implemented yet.

andreas-schilling:
TDE is enabled by default these days. So this feature would only make sense to enable it for existing setups where it is missing so far.

gusnuf:
This feature is also necessary to integrate TDE with KeyVault, so a custom/user-managed key can be used.

rojekabc:
I think the author of this issue doesn't think about settings for _azurerm_sql_database_, but rather for _azurerm_sql_server_. Look on those az client commands available on the reference:
* az sql server tde-key | Manage a server's encryption protector.
* az sql server tde-key set | Sets the server's encryption protector.
* az sql server tde-key show | Gets a server encryption protector.
[...]

PR #4754, opened 2019-10-29
azurerm_sql_server: Enable SystemAssigned identity

bgs113:
This provides the capability to enable "SystemAssigned" identity on SQL Server resources, as in the following:

```terraform
  name                         = "mysqlserver"
  resource_group_name          = azurerm_resource_group.test.name
  location                     = "azurerm_resource_group.test.location"
  version                      = "12.0"
  administrator_login          = "mradministrator"
  administrator_login_password = "thisIsDog11"

  identity {
    type = "SystemAssigned"
  }

  tags = {
    environment = "production"
  }
}
```

This is a prerequisite to enabling SQL TDE with BYOK using Azure Key Vault (requested in #87)

### DIFF
--- a/azurerm/resource_arm_sql_server.go
+++ b/azurerm/resource_arm_sql_server.go
@@ -10,6 +10,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/response"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/suppress"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
@@ -74,6 +75,33 @@ func resourceArmSqlServer() *schema.Resource {
 				Computed: true,
 			},
 
+			"identity": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: suppress.CaseDifference,
+							ValidateFunc: validation.StringInSlice([]string{
+								"SystemAssigned",
+							}, true),
+						},
+						"principal_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"tenant_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+
 			"tags": tags.Schema(),
 		},
 	}
@@ -113,6 +141,11 @@ func resourceArmSqlServerCreateUpdate(d *schema.ResourceData, meta interface{}) 
 			Version:            utils.String(version),
 			AdministratorLogin: utils.String(adminUsername),
 		},
+	}
+
+	if _, ok := d.GetOk("identity"); ok {
+		sqlServerIdentity := expandAzureRmSqlServerIdentity(d)
+		parameters.Identity = sqlServerIdentity
 	}
 
 	if d.HasChange("administrator_login_password") {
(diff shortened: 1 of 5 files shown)


azurerm/resource_arm_sql_server.go
```diff
@@ -74,6 +75,33 @@ func resourceArmSqlServer() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: suppress.CaseDifference,
```
tombuildsstuff: we're making all new fields case-sensitive, as such can we make this:

```suggestion
```

azurerm/resource_arm_sql_server.go
```diff
@@ -74,6 +75,33 @@ func resourceArmSqlServer() *schema.Resource {
 				Computed: true,
 			},
 
+			"identity": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
tombuildsstuff: since this should only be set if the users configured it - can we remove this:

```suggestion
```

azurerm/resource_arm_sql_server.go
```diff
@@ -74,6 +75,33 @@ func resourceArmSqlServer() *schema.Resource {
...
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: suppress.CaseDifference,
+							ValidateFunc: validation.StringInSlice([]string{
+								"SystemAssigned",
+							}, true),
```
tombuildsstuff: (as above) we're making all new fields case-sensitive - as such can we make this:

```suggestion
							}, false),
```
```suggestion
							}, true),
```

azurerm/resource_arm_sql_server.go
```diff
@@ -173,6 +206,10 @@ func resourceArmSqlServerRead(d *schema.ResourceData, meta interface{}) error {
 		d.Set("location", azure.NormalizeLocation(*location))
 	}
 
+	if identity := resp.Identity; identity != nil {
+		d.Set("identity", flattenAzureRmSqlServerIdentity(*identity))
+	}
```
tombuildsstuff: we need to ensure the `identity` block is always set - as such can we make this:

```suggestion
	if err := d.Set("identity", flattenAzureRmSqlServerIdentity(resp.Identity); err != nil {
		return fmt.Errorf("Error setting `identity`: %+v", err)
	}
```

azurerm/resource_arm_sql_server.go
```diff
@@ -202,3 +239,25 @@ func resourceArmSqlServerDelete(d *schema.ResourceData, meta interface{}) error
 
 	return future.WaitForCompletionRef(ctx, client.Client)
 }
+
+func expandAzureRmSqlServerIdentity(d *schema.ResourceData) *sql.ResourceIdentity {
+	identities := d.Get("identity").([]interface{})
+	identity := identities[0].(map[string]interface{})
```
tombuildsstuff: this assumes there's one element here which may not necessarily be the case - as such can we return an empty object here to ensure this is disabled:

```suggestion
	if len(identities) == 0 {
		return &sql.ResourceIdentity{}
	}
	identity := identities[0].(map[string]interface{})
```

azurerm/resource_arm_sql_server.go
```diff
@@ -202,3 +239,25 @@ func resourceArmSqlServerDelete(d *schema.ResourceData, meta interface{}) error
...
+	identityType := sql.IdentityType(identity["type"].(string))
+	return &sql.ResourceIdentity{
+		Type: identityType,
+	}
+}
+
+func flattenAzureRmSqlServerIdentity(identity sql.ResourceIdentity) []interface{} {
```
tombuildsstuff: since we need to ensure this is always set (even if there's a bad API response) can we make this:

```suggestion
func flattenAzureRmSqlServerIdentity(identity *sql.ResourceIdentity) []interface{} {
	if identity == nil {
		return []interface{}{}
	}
```

azurerm/resource_arm_sql_server.go
```diff
@@ -74,6 +75,33 @@ func resourceArmSqlServer() *schema.Resource {
 				Computed: true,
 			},
 
+			"identity": {
```
tombuildsstuff: could we [add this new block to the documentation](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/sql_server.html.markdown#L42-L54)? there's [an example of how that looks for another resource here](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/app_service.html.markdown#L94))

azurerm/resource_arm_sql_server.go
bgs113: Set this to `false`.

azurerm/resource_arm_sql_server.go
bgs113: Done.

azurerm/resource_arm_sql_server.go
bgs113: Done.

azurerm/resource_arm_sql_server.go
bgs113: Added documentation, including updated data source.

tombuildsstuff:
Tests pass:

Screenshot 2019-10-30 at 20 02 12

PR #4755, opened 2019-10-30
Sql Server Blob Auditing Polices Resource to be added

### LINKED ISSUE #2217: azurerm_sql_server: Audit Configuration
### Community Note

### Description

Add SQL Server audit configuration

- azurerm_sql_server
- azurerm_sql_database

image

yupwei68:
Related issues: https://github.com/terraform-providers/terraform-provider-azurerm/issues/2217

Fixes #2217


yupwei68:
@katbyte Thanks for your comment. I have deleted the extended server blob auditing resource. Please continue reviewing.

katbyte:
Hi @yupwei68,

I spoke with the SQL team to day and what i've found out is the `extended` version of the client & API allow more detail to be collected. So could we have both resources but they be `azurerm_mssql_server_blob_auditing_policies` & `azurerm_mssql_server_blob_extended_auditing_policies` ?

yupwei68:
Hi @katbyte ,
Comparing server blob auditing policies, extended server blob auditing polices has one more attribute "predicate expression"which allows user to filter the audit records. Both could update the SQL server auditing policies. But this attribute is not available on azure portal. That's why I deleted the extended one. I think we could keep align with portal to avoid the users' confusion.

yupwei68:
Hi @katbyte ,I have reset the extended server blob auditing policies resource back. But for the resource name, Could we keep align with other mssql resources such as "`azurerm_sql_server`","`azurerm_sql_database`"? As for "`azurerm_mssql_server_blob_extended_auditing_policies`", there is also an extended auditing polices for sql server database instead of sql server, which might cause confusion. Could we keep the two resource name "`azurerm_sql_server_blob_auditing_policies`" &"`azurerm_sql_extended_server_blob_auditing_policies`"?

yupwei68:
Hi @katbyte, I have reset back the extended blob auditing polices and change the names to `azurerm_mssql_server_blob_auditing_policies` & `azurerm_mssql_server_blob_extended_auditing_policies`. Please continue reviewing.

azurerm/data_source_mssql_server_blob_auditing_policies.go
```diff
@@ -0,0 +1,100 @@
...
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmMsSqlServerBlobAuditingPolicies() *schema.Resource {
```
tombuildsstuff: taking a look through we can remove this Data Source since this doesn't make sense for this Resource

azurerm/data_source_mssql_server_blob_extended_auditing_policies.go
```diff
@@ -0,0 +1,107 @@
...
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmMsSqlServerBlobExtendedAuditingPolicies() *schema.Resource {
```
tombuildsstuff: can we remove this data source?

azurerm/resource_arm_mssql_server_blob_auditing_policies.go
```diff
@@ -0,0 +1,230 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: azure.ValidateMsSqlServerName,
+			},
+
+			"state": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{"Enabled", "Disabled"}, false),
+			},
```
tombuildsstuff: since this can be conditionally Enabled/Disabled - this wants to be nested as a block within the Server resource e.g.

```
  # ...
  auditing_policy {}
}
```

where we can use the presence of the `auditing_policy` block to assume Enabled/Disabled - and update this as necessary, rather than having this as a separate resource

azurerm/resource_arm_mssql_server_blob_extended_auditing_policies.go
```diff
@@ -0,0 +1,240 @@
...
+				Type:      schema.TypeString,
+				Required:  true,
+				Sensitive: true,
+			},
+			"retention_days": {
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
+			"audit_actions_and_groups": {
+				Type:     schema.TypeString,
```
tombuildsstuff: since this is a List - to make this a better UX this wants to be a Set of Strings which we convert to/from a CSV

azurerm/resource_arm_mssql_server_blob_extended_auditing_policies.go
```diff
@@ -0,0 +1,240 @@
...
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
+					v := val.(string)
+					_, err := uuid.FromString(v)
+					if err != nil {
+						errs = append(errs, fmt.Errorf("%q is not in correct format:%+v", key, err))
+					}
+					return
+				},
+			},
+			"is_storage_secondary_key_in_use": {
```
tombuildsstuff: this looks like a Computed, Read-Only field - so I think we can remove this?

azurerm/resource_arm_mssql_server_blob_extended_auditing_policies.go
```diff
@@ -0,0 +1,240 @@
...
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"is_azure_monitor_target_enabled": {
```
tombuildsstuff: this looks like a Computed, Read-Only field - so I think we can remove this?

azurerm/data_source_mssql_server_blob_auditing_policies.go
yupwei68: OK

azurerm/data_source_mssql_server_blob_extended_auditing_policies.go
yupwei68: Yes

azurerm/resource_arm_mssql_server_blob_auditing_policies.go
yupwei68: OK. I have made the corresponding changes in new PR https://github.com/terraform-providers/terraform-provider-azurerm/pull/5036

azurerm/resource_arm_mssql_server_blob_extended_auditing_policies.go
yupwei68: OK. I have made the corresponding changes in new PR #5036

azurerm/resource_arm_mssql_server_blob_extended_auditing_policies.go
yupwei68: This feature is available on portal. So I have reserved this feature. 
image

azurerm/resource_arm_mssql_server_blob_extended_auditing_policies.go
yupwei68: OK. I have removed it in new PR https://github.com/terraform-providers/terraform-provider-azurerm/pull/5036

yupwei68:
Hi @tombuildsstuff I have made the corresponding two changes in new PR DB: #5049 & server: #5036. I choose to implement the `blob_extended_auditing_policy` for it's more inclusive.

PR #4767, opened 2019-10-30
[Website] Fix broken link

cgriggs01:
Adding correct path for documentation link... fixing broken link error


cgriggs01:
How will this change make it into the `stable-website` branch?

Is that something I am responsible for?

Issue #4575, opened 2019-10-09
azurerm_app_service: mark app_settings as sensitive


PSanetra:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform: 0.12.9
AzureRM: 1.35.0

### Affected Resource(s)

* `azurerm_app_service`

### Expected Behavior

The `app_settings` elements of the `azurerm_app_service` should be marked as sensitive. Because there may be sensitive values in the `app_settings`.

### Actual Behavior

The `app_settings` element values are printed to the terraform log during the terraform planning phase if a `azurerm_app_service` resource will be destroyed.

StefanSchoof:
I disagree. 

Sensitive app settings belong into a key vault. And making this senetive, makes people think that the information is protected and make it harder to see changes of not senetive data.

tombuildsstuff:
hi @PSanetra 

Thanks for opening this issue.

Whilst it's certainly possible to store sensitive information in the App Settings block - Microsoft recommends against doing so and (as @StefanSchoof has mentioned) suggests instead storing/retrieving sensitive information from Azure Key Vault instead ([you can find an example of how to do this here](https://medium.com/@patoncrispy/securing-settings-in-dot-net-core-apps-using-azure-key-vault-d4ec82de00b8)).

Unfortunately implementing this change would impact users who are following the recommended advise here (since any changes, for example changing a feature toggle, would be masked in the plan). As such whilst I'd like to thank you for opening this issue, since this behaviour is intentional (due to Microsoft's recommendation) I'm going to close this issue for the moment.

Thanks!

PSanetra:
@tombuildsstuff The article you posted stores the `ClientSecret` for accessing the key vault in the `appSettings.json`. I don't think that is an option if you want to deploy your app as a container, is it?

@StefanSchoof is it possible to access the key vault without a client secret which would be stored inside the app settings and therefore logged in plain text by terraform?

StefanSchoof:
[...]

PR #4771, opened 2019-10-31
azurerm_app_service: mark app_settings as sensitive

aqche:
Sets `Sensitive: true` to `app_settings` for `azurerm_app_service` to address #4575.

### DIFF
--- a/azurerm/data_source_app_service.go
+++ b/azurerm/data_source_app_service.go
@@ -63,6 +63,7 @@ func dataSourceArmAppService() *schema.Resource {
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},
+				Sensitive: true,
 			},
 
 			"connection_string": {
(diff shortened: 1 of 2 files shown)


StefanSchoof:
I added a comment to the issue, where I disagree that doing so is a good idea.

tombuildsstuff:
hi @PSanetra 

Thanks for this PR :)

Whilst it's certainly possible to store sensitive information in the App Settings block - Microsoft recommends against doing so and (as @StefanSchoof has mentioned in #4575) suggests instead storing/retrieving sensitive information from Azure Key Vault instead ([you can find an example of how to do this here](https://medium.com/@patoncrispy/securing-settings-in-dot-net-core-apps-using-azure-key-vault-d4ec82de00b8)).

Unfortunately implementing this change would impact users who are following the recommended advise here (since any changes, for example changing a feature toggle, would be masked in the plan). As such whilst I'd like to thank you for opening this Pull Request, since this behaviour is intentional (due to Microsoft's recommendation) I'm going to close this PR for the moment.

Thanks!

aqche:
@StefanSchoof @tombuildsstuff 

thanks for the comments. makes sense to me! 👍

PR #4777, opened 2019-10-31
azurerm_automation_credential  -  deprecate `account_name` in favour of `automation_account_name`

### LINKED ISSUE #4776: Add input validation to automation resources and change account_name parameter
### Community Note

### Description

Not all automation resources have validation for the automation account, runbook and schedule names. These already exist in the helpers and this PR just adds them to the fields. 

Furthermore, the `azurerm_automation_credential` resource still uses the `account_name` parameter. This PR

* azurerm_automation_account
* azurerm_automation_credential
* azurerm_automation_dsc_configuration
* azurerm_automation_module
* azurerm_automation_runbook
* azurerm_automation_schedule

For the `azurerm_automation_credential`, one could add `automation_account_name` as an alternative to `account_name`. The code can be copied from what is in `azurerm_automation_schedule`.

```hcl
resource "azurerm_automation_credential" "test" {
  name                    = "devicePassword"
  resource_group_name     = "${azurerm_resource_group.test.name}"
  automation_account_name = "${azurerm_automation_account.test.name}"
  username                = "test_user"
  password                = "test_pwd"
}
```

### References

* #0000

draggeta:
This MR does the following:
- Add validation for automation account names, runbooks and schedules
- Adds the `automation_account_name` parameter to `azurerm_automation_credential`
- Updates test and docs for `azurerm_automation_credential`

Closes #4776

### DIFF
--- a/azurerm/resource_arm_automation_credential.go
+++ b/azurerm/resource_arm_automation_credential.go
@@ -41,10 +41,23 @@ func resourceArmAutomationCredential() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupName(),
 
+			//this is AutomationAccountName in the SDK
 			"account_name": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				Deprecated:    "account_name has been renamed to automation_account_name for clarity and to match the azure API",
+				ConflictsWith: []string{"automation_account_name"},
+				ValidateFunc: azure.ValidateAutomationAccountName(),
+			},
+
+			"automation_account_name": {
 				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Optional: true, //todo change to required once account_name has been removed
+				Computed: true,
+				//ForceNew:      true, //todo this needs to come back once account_name has been removed
+				ConflictsWith: []string{"account_name"},
+				ValidateFunc: azure.ValidateAutomationAccountName(),
 			},
 
 			"username": {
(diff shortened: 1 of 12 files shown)


draggeta:
If this is accepted, the depreciation notice should probably be added to the 2.0 depreciation list?

katbyte:
@draggeta, yep! it sure should be

azurerm/resource_arm_automation_credential.go
```diff
@@ -41,10 +41,23 @@ func resourceArmAutomationCredential() *schema.Resource {
...
+				ValidateFunc:  azure.ValidateAutomationAccountName(),
+			},
+
+			"automation_account_name": {
 				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Optional: true, //todo change to required once account_name has been removed
+				Computed: true,
+				//ForceNew:      true, //todo this needs to come back once account_name has been removed
```
katbyte: As we are removing this, it might be worth catching the error on update and checking if its for renaming this property and adding a `the name property needs a manual destroy/re-creation to update for until 2.0` or something to that effect?

azurerm/resource_arm_automation_credential.go
```diff
@@ -75,13 +88,20 @@ func resourceArmAutomationCredentialCreateUpdate(d *schema.ResourceData, meta in
 
 	name := d.Get("name").(string)
 	resGroup := d.Get("resource_group_name").(string)
-	accName := d.Get("account_name").(string)
+	//CustomizeDiff should ensure one of these two is set
```
katbyte: as customize diff is broken (and we are working on a better way to achieve this with a new validator) i think we can remove this
```suggestion
```

azurerm/resource_arm_automation_credential.go
draggeta: I was about to implement this and then realised that this should just work when the `ForceNew` property is set to `true` on both. I've tested it and it works as one expects. Even when changing between `account_name` and `automation_account_name`

katbyte: The use case i would be concerned with there is migrating to the new property. If i am on the old one and then switch to the new property would that cause a recreation?

draggeta: From what I tested, it doesn't. This is probably due to the fact that both are computed. This means that if you don't specify any value, it uses the current value. Switching between account_name and automation_account_name and back doesn't change anything (unless you change the associated value).

katbyte: perfect! 👌

katbyte:
whoops somehow merged before approved, LGTM now @draggeta! thanks!

Issue #4739, opened 2019-10-28
private_dns_zone does not support SRV records


alainvanhoof:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

private_dns_zone does not support SRV records

### New or Affected Resource(s)

Terraform v0.12.12
provider.azurerm v1.35.0

### Potential Terraform Configuration

azurerm_private_dns_zone

### Expected Behavior

Can create SRV records in private DNS zone.

### Actual Behavior

I can create a DNS zone with a private zone type, which is deprecated and gives deprecated warnings, however if I create a private DNS zone, then SRV record are not supported.

phires:
I'm currently looking into implementing all the missing types to the private dns zones, will create a pull request sometime next week.

katbyte:
fixed by #4783###

PR #4783, opened 2019-10-31
new resource 'azurerm_private_dns_srv_record'

phires:
New Resource 'azurerm_private_dns_srv_record' for SRV DNS records in Private DNS Zones.
Fixes issue #4739


azurerm/resource_arm_private_dns_srv_record.go
```diff
@@ -0,0 +1,262 @@
...
+func flattenAzureRmPrivateDnsSrvRecords(records *[]privatedns.SrvRecord) []map[string]interface{} {
+	results := make([]map[string]interface{}, 0)
+
+	if records != nil {
+		for _, record := range *records {
+			results = append(results, map[string]interface{}{
+				"priority": *record.Priority,
+				"weight":   *record.Weight,
+				"port":     *record.Port,
+				"target":   *record.Target,
```
katbyte: we will need to nil check these de-references

azurerm/resource_arm_private_dns_srv_record.go
```diff
@@ -0,0 +1,262 @@
...
+
+	if m, ok := v.(map[string]interface{}); ok {
+		buf.WriteString(fmt.Sprintf("%d-", m["priority"].(int)))
+		buf.WriteString(fmt.Sprintf("%d-", m["weight"].(int)))
+		buf.WriteString(fmt.Sprintf("%d-", m["port"].(int)))
+		buf.WriteString(fmt.Sprintf("%s-", m["target"].(string)))
+	}
+
+	return hashcode.String(buf.String())
+}
```
katbyte: as mentioned above we probably don't need this

azurerm/resource_arm_private_dns_srv_record_test.go
```diff
@@ -0,0 +1,365 @@
...
+  location = "%s"
+}
+
+resource "azurerm_private_dns_zone" "test" {
+  name                = "testzone%d.com"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+}
+
+resource "azurerm_private_dns_srv_record" "test" {
+  name                = "test%d"
```
katbyte: could we use
```suggestion
  name                = "testaccSRV%d"
```

for the dns srv records

azurerm/resource_arm_private_dns_srv_record_test.go
```diff
@@ -0,0 +1,365 @@
...
+		return fmt.Errorf("Private DNS SRV record still exists:\n%#v", resp.RecordSetProperties)
+	}
+
+	return nil
+}
+
+func testAccAzureRMPrivateDnsSrvRecord_basic(rInt int, location string) string {
+	return fmt.Sprintf(`
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-%d"
```
katbyte: and can we use this for the resource group names
```suggestion
  name     = "acctestRG-prvdns-%d"
```

website/docs/r/private_dns_srv_record.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+---
+
+# azurerm_private_dns_srv_record
+
+Enables you to manage DNS SRV Records within Azure Private DNS.
+
+## Example Usage
+
+```hcl
+resource "azurerm_resource_group" "test" {
```
katbyte: could we use example for resource names in docs?
```suggestion
resource "azurerm_resource_group" "example" {
```

website/docs/r/private_dns_srv_record.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+
+The `record` block supports:
+
+* `priority` - (Required) Priority of the SRV record.
+
+* `weight` - (Required) Weight of the SRV record.
+
+* `port` - (Required) Port the service is listening on.
+
+* `target` - (Required) FQDN of the service.
```
katbyte:
```suggestion
* `target` - (Required) The FQDN of the service.
```

website/docs/r/private_dns_srv_record.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+The `record` block supports:
+
+* `priority` - (Required) Priority of the SRV record.
+
+* `weight` - (Required) Weight of the SRV record.
+
+* `port` - (Required) Port the service is listening on.
```
katbyte:
```suggestion
* `port` - (Required) The Port the service is listening on.
```

website/docs/r/private_dns_srv_record.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+
+* `ttl ` - (Required) The Time To Live (TTL) of the DNS record in seconds.
+
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+The `record` block supports:
+
+* `priority` - (Required) Priority of the SRV record.
+
+* `weight` - (Required) Weight of the SRV record.
```
katbyte:
```suggestion
* `weight` - (Required) The Weight of the SRV record.
```

website/docs/r/private_dns_srv_record.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+
+* `zone_name` - (Required) Specifies the Private DNS Zone where the resource exists. Changing this forces a new resource to be created.
+
+* `ttl ` - (Required) The Time To Live (TTL) of the DNS record in seconds.
+
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+The `record` block supports:
+
+* `priority` - (Required) Priority of the SRV record.
```
katbyte:
```suggestion
* `priority` - (Required) The priority of the SRV record.
```

azurerm/resource_arm_private_dns_srv_record.go
phires: Do we really? The incoming record should always be complete as the content is validated in the Schema

website/docs/r/private_dns_srv_record.html.markdown
```diff
@@ -0,0 +1,89 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the DNS SRV Record. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) Specifies the resource group where the resource exists. Changing this forces a new resource to be created.
+
+* `zone_name` - (Required) Specifies the Private DNS Zone where the resource exists. Changing this forces a new resource to be created.
```
tombuildsstuff:
```suggestion
* `zone_name` - (Required) Specifies the Private DNS Zone where the resource exists. Changing this forces a new resource to be created.

* `record` - (Required) One or more `record` blocks as defined below.
```

website/docs/r/private_dns_srv_record.html.markdown
```diff
@@ -0,0 +1,89 @@
...
+
+* `ttl ` - (Required) The Time To Live (TTL) of the DNS record in seconds.
+
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+The `record` block supports:
```
tombuildsstuff: could we add a divider here / update the text to match other resources:

```suggestion
---

A `record` block supports the following:
```

azurerm/resource_arm_private_dns_srv_record.go
tombuildsstuff: the Azure API returns the representation of a resource at the last time it was modified, rather then the latest representation for the API version; as such we've seen (multiple) instances where the API response differs from the schema defined in the SDK, which means these values will be nil - which is why we're cautious parsing the response back from the API - so unfortunately yes (thanks for pushing this)

azurerm/resource_arm_private_dns_srv_record.go
```diff
@@ -0,0 +1,252 @@
...
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.NoEmptyStrings,
```
tombuildsstuff: running the tests it appears due to the API bug outlined below that the name is also lower-cased in the API response. Since DNS addresses themselves need to be lower-cased - I hope you don't mind but I'm going to push a commit to add some validation to this resource (and the other Private DNS resources) to ensure that the name field is lower-cased

tombuildsstuff:
With the validation/test changes the tests now pass:

Screenshot 2019-11-17 at 15 25 14

PR #4788, opened 2019-11-01
Add usage example for app_service_certificate

eweilnau:
I added an example of creating an app_service_certificate using key_vault_secret_id. Having it would have saved me a few hours of trial and error.


eweilnau:
Sounds good to me. I will work on moving it over to the example folder and format.

tombuildsstuff:
hey @eweilnau 

Thanks for this PR - I hope you don't mind but so that we can get this merged I've pushed a commit which moves this to the examples folder (and adds a link from the resource documentation so folks can find it)

Thanks!

eweilnau:
@tombuildsstuff thanks for taking this the "last mile". I was struggling to test the example for lack of a certificate that I could share.

Issue #4542, opened 2019-10-07
Add support for ForwardDeadLetteredMessagesTo (ServiceBus > Subscription)


matt40k:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Please update azurerm_servicebus_subscription to support ForwardDeadLetteredMessagesTo. 
https://docs.microsoft.com/en-us/powershell/module/azurerm.servicebus/new-azurermservicebussubscription?view=azurermps-6.13.0

### New or Affected Resource(s)

* azurerm_XXXXX

### Potential Terraform Configuration

```hcl
resource "azurerm_servicebus_subscription" "example" {
  name                = "tfex_sevicebus_subscription"
  resource_group_name = "${azurerm_resource_group.example.name}"
  namespace_name      = "${azurerm_servicebus_namespace.example.name}"
  topic_name          = "${azurerm_servicebus_topic.example.name}"
  max_delivery_count  = 1
  forward_dead_lettered_messages_to = "example_dead"
}
```

### References

https://docs.microsoft.com/en-us/powershell/module/azurerm.servicebus/new-azurermservicebussubscription?view=azurermps-6.13.0

* #0000

PR #4789, opened 2019-11-01
azurerm_servicebus_subscription  - support for the `forward_dead_lettered_messages_to` property

aqche:
Adds the `forward_dead_lettered_messages_to` argument to `azurerm_servicebus_subscription` to address #4542


aqche:
@katbyte thanks for the review! :)

PR #4792, opened 2019-11-01
Add subcategories to docs for display on the Terraform Registry

justincampbell:
We will soon be displaying documentation for this provider both on [terraform.io](https://www.terraform.io/docs/providers/index.html) and on [the Terraform Registry](https://registry.terraform.io/providers).

Documentation displayed on the Terraform Registry will not use the ERB layout containing the existing navigation, and will instead build the navigation dynamically based on the directory and YAML frontmatter of a file. For providers that group similar resources and data sources by service or use case (subcategories), we'll need to add this information to the Markdown source file.

This PR modifies Resource and Data Source documentation source files by adding a subcategory to the metadata if those files were grouped previously.

For more information about how documentation is rendered on the Terraform Registry, please see this reference: [Terraform Registry - Provider Documentation](https://www.terraform.io/docs/registry/providers/docs.html).


website/azurerm.erb
```diff
@@ -788,9 +820,17 @@
                   <a href="/docs/providers/azurerm/r/virtual_machine_extension.html">azurerm_virtual_machine_extension</a>
                 </li>
 
+                <li>
+                  <a href="/docs/providers/azurerm/r/linux_virtual_machine_scale_set.html">azurerm_linux_virtual_machine_scale_set</a>
+                </li>
```
tombuildsstuff: this hasn't launched yet (it's feature toggled off) - so can we remove this for the moment?

website/azurerm.erb
```diff
@@ -788,9 +820,17 @@
...
+
                 <li>
                   <a href="/docs/providers/azurerm/r/virtual_machine_scale_set.html">azurerm_virtual_machine_scale_set</a>
                 </li>
+
+                <li>
+                  <a href="/docs/providers/azurerm/r/windows_virtual_machine_scale_set.html">azurerm_windows_virtual_machine_scale_set</a>
+                </li>
```
tombuildsstuff: this hasn't launched yet (it's feature toggled off) - so can we remove this for the moment?

tombuildsstuff: this hasn't launched yet (it's feature toggled off) - so can we remove this for the moment?

website/azurerm.erb
tombuildsstuff: this hasn't launched yet (it's feature toggled off) - so can we remove this for the moment?

website/docs/r/private_dns_zone_virtual_network_link.html.markdown
```diff
@@ -1,4 +1,5 @@
 ---
+subcategory: "Private DNS"
```
tombuildsstuff: 🤔 about this would it make more sense to have these as nested sub-categories? e.g. `DNS -> Public/Private`?

justincampbell: Interesting! Not sure what you mean by nested, there's only 1 level of subcategories "DNS" or "Private DNS" and then a "Resources" folder underneath that. I'll DM you a link with a current example.

justincampbell:
@tombuildsstuff Thanks! I removed those 2 from the HTML nav and set their subcategories to "Beta".

website/docs/r/private_dns_zone_virtual_network_link.html.markdown
katbyte: I think ideally we would want to be able to nest further:
`DNS/Public/resources`
`DNS/Private/resources`
`Recovery Services/Backup/resources`
`Recovery Services/Replication/resources`
ect

justincampbell: That won't be possible in the near term, but it's something we can add to the backlog to prioritize!

azurerm/provider.go
```diff
@@ -683,7 +683,7 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
 			SupportsAuxiliaryTenants:       len(auxTenants) > 0,
 
 			// Doc Links
-			ClientSecretDocsLink: "https://www.terraform.io/docs/providers/azurerm/auth/service_principal_client_secret.html",
+			ClientSecretDocsLink: "https://www.terraform.io/docs/providers/azurerm/guides/service_principal_client_secret.html",
```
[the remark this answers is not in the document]
justincampbell: Prepared here, I will wait until both PRs are approved to merge this and immediately deploy the redirects afterward: https://github.com/hashicorp/terraform-website/pull/991

PR #4793, opened 2019-11-01
New Resource: `azurerm_data_factory_trigger_schedule` 

mbfrahry:
```
--- PASS: TestAccAzureRMDataFactoryTriggerSchedule_basic (84.33s)
--- PASS: TestAccAzureRMDataFactoryTriggerSchedule_complete (103.15s)
```

### DIFF
--- a/azurerm/resource_arm_data_factory_trigger_schedule.go
+++ b/azurerm/resource_arm_data_factory_trigger_schedule.go
@@ -54,6 +54,9 @@ func resourceArmDataFactoryTriggerSchedule() *schema.Resource {
 				ValidateFunc: validate.DataFactoryName(),
 			},
 
+			// This time can only be  represented in UTC.
+			// An issue has been filed in the SDK for the timezone attribute that doesn't seem to work
+			// https://github.com/Azure/azure-sdk-for-go/issues/6244
 			"start_time": {
 				Type:             schema.TypeString,
 				Optional:         true,
@@ -62,6 +65,9 @@ func resourceArmDataFactoryTriggerSchedule() *schema.Resource {
 				ValidateFunc:     validate.RFC3339Time, //times in the past just start immediately
 			},
 
+			// This time can only be  represented in UTC.
+			// An issue has been filed in the SDK for the timezone attribute that doesn't seem to work
+			// https://github.com/Azure/azure-sdk-for-go/issues/6244
 			"end_time": {
 				Type:             schema.TypeString,
 				Optional:         true,
@@ -87,6 +93,20 @@ func resourceArmDataFactoryTriggerSchedule() *schema.Resource {
 				Optional:     true,
 				Default:      1,
 				ValidateFunc: validation.IntAtLeast(1),
+			},
+
+			"pipeline_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.DataFactoryPipelineAndTriggerName(),
+			},
+
+			"pipeline_parameters": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
 			},
 
 			"annotations": {
@@ -144,8 +164,18 @@ func resourceArmDataFactoryTriggerScheduleCreateUpdate(d *schema.ResourceData, m
 		props.Recurrence.EndTime = &date.Time{Time: t}
 	}
 
+	reference := &datafactory.PipelineReference{
+		ReferenceName: utils.String(d.Get("pipeline_name").(string)),
+	}
+
 	scheduleProps := &datafactory.ScheduleTrigger{
 		ScheduleTriggerTypeProperties: props,
+		Pipelines: &[]datafactory.TriggerPipelineReference{
+			{
+				PipelineReference: reference,
+				Parameters:        d.Get("pipeline_parameters").(map[string]interface{}),
+			},
+		},
 	}
 
 	if v, ok := d.GetOk("annotations"); ok {
@@ -216,6 +246,16 @@ func resourceArmDataFactoryTriggerScheduleRead(d *schema.ResourceData, meta inte
 			}
 			d.Set("frequency", recurrence.Frequency)
 			d.Set("interval", recurrence.Interval)
+		}
+
+		if pipelines := scheduleTriggerProps.Pipelines; pipelines != nil {
+			if len(*pipelines) > 0 {
+				pipeline := *pipelines
+				if reference := pipeline[0].PipelineReference; reference != nil {
+					d.Set("pipeline_name", reference.ReferenceName)
+				}
+				d.Set("pipeline_parameters", pipeline[0].Parameters)
+			}
 		}
 
 		annotations := flattenDataFactoryAnnotations(scheduleTriggerProps.Annotations)
--- a/azurerm/resource_arm_data_factory_trigger_schedule.go
+++ b/azurerm/resource_arm_data_factory_trigger_schedule.go
@@ -1,0 +1,257 @@
+package azurerm
+
+import (
+	"fmt"
+	"github.com/Azure/go-autorest/autorest/date"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/suppress"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
+	"log"
+	"regexp"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/datafactory/mgmt/2018-06-01/datafactory"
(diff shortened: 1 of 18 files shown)


azurerm/resource_arm_data_factory_trigger_schedule.go
```diff
@@ -0,0 +1,259 @@
...
+	"fmt"
+	"github.com/Azure/go-autorest/autorest/date"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/suppress"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
+	"log"
+	"regexp"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/datafactory/mgmt/2018-06-01/datafactory"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
```
katbyte: could we run goimports on this file?

azurerm/resource_arm_data_factory_trigger_schedule.go
```diff
@@ -0,0 +1,259 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateAzureRMDataFactoryPipelineName,
+			},
+
+			// There's a bug in the Azure API where this is returned in lower-case
+			// BUG: https://github.com/Azure/azure-rest-api-specs/issues/5788
```
katbyte: I suggest reopening this on the sdk as they seem to actually act on issues there... sometimes 😪

azurerm/resource_arm_data_factory_trigger_schedule.go
```diff
@@ -0,0 +1,259 @@
...
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateAzureRMDataFactoryPipelineName,
```
katbyte: validateAzureRMDataFactoryPipelineName? is that the same rules as for a trigger name?

azurerm/resource_arm_data_factory_trigger_schedule.go
```diff
@@ -0,0 +1,259 @@
...
+			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
+
+			"data_factory_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringMatch(
+					regexp.MustCompile(`^[A-Za-z0-9]+(?:-[A-Za-z0-9]+)*$`),
+					`Invalid data_factory_name, see https://docs.microsoft.com/en-us/azure/data-factory/naming-rules`,
+				),
```
katbyte: We should probably put this into a shared validation function as its used in multiple places

azurerm/resource_arm_data_factory_trigger_schedule.go
```diff
@@ -0,0 +1,259 @@
...
+				Default:      1,
+				ValidateFunc: validation.IntAtLeast(1),
+			},
+
+			"timezone": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "UTC",
+				ValidateFunc: validate.NoEmptyStrings,
+			},
```
katbyte: We have a few different timezone validation functions, perhaps one would work here?

azurerm/resource_arm_data_factory_trigger_schedule.go
```diff
@@ -0,0 +1,259 @@
...
+
+			"annotations": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
```
katbyte: Could we validate against `NoEmptyStrings`?

azurerm/resource_arm_data_factory_trigger_schedule_test.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+resource "azurerm_data_factory" "test" {
+  name                = "acctestdf%d"
+  location            = "${azurerm_resource_group.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+}
+
+resource "azurerm_data_factory_trigger_schedule" "test" {
+  name                = "acctestdf%d"
```
katbyte: Could we put some capitals into the name?
```suggestion
  name                = "acctestDFTS%d"
```

azurerm/resource_arm_data_factory_trigger_schedule_test.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+	}
+
+	return nil
+}
+
+func testAccAzureRMDataFactoryTriggerSchedule_basic(rInt int, location string) string {
+	return fmt.Sprintf(`
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-%d"
```
katbyte: I've been doing this lately for tests
```suggestion
  name     = "acctestRG-datafactory-%d"
```

website/docs/r/data_factory_trigger_schedule.html.markdown
```diff
@@ -0,0 +1,71 @@
...
+  interval    = 5
+  frequency   = "Day"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) Specifies the name of the Data Factory Schedule Trigger. Changing this forces a new resource to be created. Must be globally unique. See the [Microsoft documentation](https://docs.microsoft.com/en-us/azure/data-factory/naming-rules) for all restrictions.
```
katbyte: If the API has a `CheckName1 function we should probably make use of it to ensure the name is unique

mbfrahry: Unfortunately, I didn't find one

azurerm/resource_arm_data_factory_trigger_schedule.go
mbfrahry: Sure is. Moved it out into it's own validation

azurerm/resource_arm_data_factory_trigger_schedule.go
mbfrahry: I'm hesitant to add any validation here and might actually just remove this value. It doesn't seem to do anything and the UI just defaults the time to UTC anyway

mbfrahry:
Review addressed but I need to add one more piece before it's ready

azurerm/resource_arm_data_factory_trigger_schedule.go
katbyte: If you pick PST the API still takes the value as UTC?

Then i would definitely remove it, add a note that the time is always UC, and open a bug on the SDK with a link in the code explaining why UTC has been ignored.

mbfrahry: Confirmed that timezone doesn't work no matter the value. It's been removed and an issue has been opened https://github.com/Azure/azure-sdk-for-go/issues/6244

mbfrahry: The Data Factory API attributes tend to accept any value without validation. You can submit anything to timezone and it'll be accepted whether it's valid or not but `Pacific Standard Time` didn't change the format of the time at all so I've submitted an issue on the sdk as a start. https://github.com/Azure/azure-sdk-for-go/issues/6244

mbfrahry:
This should be ready for a final review @katbyte

azurerm/resource_arm_data_factory_trigger_schedule.go
katbyte: _fun_

neil-yechenwei:
@mbfrahry , I noticed you made a PR (https://github.com/terraform-providers/terraform-provider-azurerm/pull/3244) to add support for azurerm_data_factory_pipeline. I found you added a property "activity" in your first commit (https://github.com/terraform-providers/terraform-provider-azurerm/pull/3244/commits/06bd3226b5d97c47b4d33a542231d624c1e31b5e). But you deleted this property in later commit (https://github.com/terraform-providers/terraform-provider-azurerm/pull/3244/commits/6822cfa94215b4e0132114e3da0bb192596ad7af). After learned from the azure doc for data factory (https://docs.microsoft.com/en-us/azure/data-factory/v1/data-factory-copy-activity-tutorial-using-powershell), I think you should retain this property "activity" since I think it's an important thing for data factory pipeline to combine the InputDataset and the OutputDataset. May I know why you removed the property "activity"? If this property is necessary, can it be added back?

Issue #4769, opened 2019-10-31
Dependency graph for VMSS prevents Loadbalancer update


mbrancato:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

```
Terraform v0.12.3
+ provider.azurerm v1.36.1
```

### Affected Resource(s)

* `azurerm_virtual_machine_scale_set`
* `azurerm_lb_probe`
* `azurerm_lb`

The VMSS is dependent on the LB probe and LB resources. When certain changes are made to these causing recreation, the `apply` will fail. To correct this, I see two possible solutions:

1. Modify the VMSS to remove the LB probe dependency, then re-add once the LB probe is recreated.
2. Delete the VMSS first, then recreate. Obviously this could have a larger impact for a small change.

### Terraform Configuration Files

```hcl
variable "resource_group" {}

data "azurerm_resource_group" "test" {
  name = "${var.resource_group}"
}

locals {
  prefix = "test-vmss-bug"
}

resource "azurerm_virtual_network" "test" {
  name                = "${local.prefix}-vnet"
  address_space       = ["10.0.0.0/16"]
  location            = "${data.azurerm_resource_group.test.location}"
  resource_group_name = "${data.azurerm_resource_group.test.name}"
}

resource "azurerm_subnet" "test" {
  name                 = "${local.prefix}-subnet"
  resource_group_name  = "${data.azurerm_resource_group.test.name}"
  virtual_network_name = "${azurerm_virtual_network.test.name}"
  address_prefix       = "10.0.2.0/24"
}

resource "azurerm_lb" "test" {
  name                = "${local.prefix}-lb"
  location            = "${data.azurerm_resource_group.test.location}"
  resource_group_name = "${data.azurerm_resource_group.test.name}"
  #sku                 = "Standard"

  frontend_ip_configuration {
    name                          = "${local.prefix}-lbip"
    subnet_id                     = "${azurerm_subnet.test.id}"
    private_ip_address            = "10.0.2.250"
    private_ip_address_allocation = "Static"
  }
}

[...]

PR #4794, opened 2019-11-02
Add test for changing loadbalancer health probe on VMSS

mbrancato:
This adds a test for updating the VMSS loadbalancer when the VMSS has the `health_probe_id` property assigned to an `azurerm_lb_probe` resource.

For more info: #4769


katbyte:
Hi @mbrancato,

Thank you for opening this PR. As the old VMSS resource is going to be deprecated i will open a new PR with this test for you.

PR #4795, opened 2019-11-02
Update virtual_network_gateway.html.markdown

BDworak:
SKU type required is actually "Basic" and not "Standard". My mistake on last PR.


BDworak:
@tombuildsstuff  #4768 Got the SKUs mixed up. It is actually a basic PIP that is needed, not Standard. Sorry for the mistake.

Issue #4807, opened 2019-11-05
Hyphens not allowed in Azure Bastion ip configuration name


arestarh:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

1.36.1

### Affected Resource(s)

azurerm_bastion_host

### Terraform Configuration Files

```hcl
resource "azurerm_bastion_host" "bastion" {
  name                = "myname-bastion-myenvironment"
  location            = "${azurerm_resource_group.rg.location}"
  resource_group_name = "${azurerm_resource_group.rg.name}"

  ip_configuration {
    name                 = "myname-bastion-myenvironment-configuration"
    subnet_id            = "${azurerm_subnet.bastion.id}"
    public_ip_address_id = "${azurerm_public_ip.bhpip.id}"
  }
}
```

### Expected Behavior

Resource ip configuration name should be myname-bastion-myenvironment-configuration.

### Actual Behavior

This error is generated: **Error**: _lowercase letters, highercase letters numbers only are allowed in "ip_configuration.0.name": "myname-bastion-myenvironment-configuration"_

### Steps to Reproduce

Create a bastion that contains hyphens as part of the ip configuration name

arestarh:
Pretty similar issue https://github.com/terraform-providers/terraform-provider-azurerm/issues/4759 that related to bastion host name was resolved by PR https://github.com/terraform-providers/terraform-provider-azurerm/pull/4766

Seems like we have bastion ip configuration name constraints as well according to https://github.com/terraform-providers/terraform-provider-azurerm/blob/94cf18b287d634ee16aa5e37d3a73d3e8a356fa4/azurerm/resource_arm_bastion_host.go#L211

Unfortunately I didn't find info about bastion ip configuration name constraints neither in
https://docs.microsoft.com/en-us/azure/templates/microsoft.network/2019-06-01/bastionhosts nor in https://docs.microsoft.com/en-us/rest/api/virtualnetwork/bastionhosts/createorupdate

tombuildsstuff:
hey @arestarh 

Thanks for opening this issue.

[...]

PR #4814, opened 2019-11-06
Update bastion host ip config name validation to allow for hyphens

aqche:
Updates the bastion host ip config name validation to match the azure requirements to allow for hyphens in the name. Also updated tests to include hyphen in ip configuration name to cover this issue. 

Fixes: #4807


tombuildsstuff:
Tests pass:

Screenshot 2019-11-06 at 09 48 23

PR #3672, opened 2019-06-14
New resources: `iothub_endpoint_storage_container`, `iothub_endpoint_eventhub`, `iothub_endpoint_servicebus_queue`, `iothub_endpoint_servicebus_topic`

maxbog:
Extracts Endpoint resources from IoTHub.

See #3303

PR #4823, opened 2019-11-07
New resources: `iothub_endpoint_storage_container`, `iothub_endpoint_eventhub`, `iothub_endpoint_servicebus_queue`, `iothub_endpoint_servicebus_topic`

mbfrahry:
This is a continuation of #3672. That extracts `endpoints` out of the iothub resource.


azurerm/resource_arm_iothub_endpoint_storage_container.go
```diff
@@ -0,0 +1,303 @@
...
+					maskedNew := accountKeyRegex.ReplaceAllString(new, "AccountKey=****")
+					return (new == d.Get(k).(string)) && (maskedNew == old)
+				},
+				Sensitive: true,
+			},
+
+			"encoding": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				DiffSuppressFunc: suppress.CaseDifference,
```
katbyte: Is this required? if so does that mean there is an API or SDK bug where its returning a different constant then in the SDK and should we open an issue?

website/docs/r/iothub_endpoint_storage_container.html.markdown
```diff
@@ -0,0 +1,102 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the endpoint. The name must be unique across endpoint types. The following names are reserved:  `events`, `operationsMonitoringEvents`, `fileNotifications` and `$default`.
+
+* `resource_group_name` - (Required) The name of the resource group under which the IotHub Storage Container Endpoint resource has to be created. Changing this forces a new resource to be created.
+
+* `iothub_name` - (Required) The name of the IoTHub to which this Storage Container Endpoint belongs. Changing this forces a new resource to be created.
+
```
katbyte:
```suggestion
```

azurerm/resource_arm_iothub_endpoint_storage_container.go
mbfrahry: Good call. I removed it and ran through the options and everything looked fine.

Issue #4760, opened 2019-10-30
Feature request: azurerm_api_management_diagnostic


molinch:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently it does not seem to be possible to specify the diagnostic settings for API management
This part was mentioned in https://github.com/terraform-providers/terraform-provider-azurerm/issues/1177 but it seems like it got forgotten

![image](https://user-images.githubusercontent.com/5130190/67845181-f1f86f00-faf6-11e9-840c-ed0ed0514463.png)

### New or Affected Resource(s)

- azurerm_api_management_diagnostic
- azurerm_api_management_api_diagnostic

### Potential Terraform Configuration

Basic settings would be as simple as:
```hcl
resource "azurerm_api_management_diagnostic" "api_management_diagnostic" {
  logger = "${azurerm_api_management_logger.api_management_logger.name}"
  sampling = 20
  always_log_errors = true
}
```
But there are optional additional settings as well:
![image](https://user-images.githubusercontent.com/5130190/67845428-603d3180-faf7-11e9-8024-5d0970b25134.png)

### References

See the yaml that references diagnostics:
https://github.com/Azure/azure-api-management-devops-resource-kit/blob/master/example/diagnostics/application-insights.template.json

aqche:
Hi, I was making an attempt at adding the `azurerm_api_management_diagnostic` resource and managed it get it working but since it is the provider is currently using `2018-01-01` version of the `apimanagement` service the resource only supports enabling and disabling Application Insights. Would it be valuable for me to open a PR with the `2018-01-01` `azurerm_api_management_diagnostic` resource?

jeanpaulsmit:
Great feature request, looking for this myself. However, I don't think this should be enabled on APIM level, rather on API level?
Although you can enable it on all API's using the Azure portal, in my view it's a property of an API.
As the logging/diagnostics are going to App Insights, the resource should be called something like api-management-api-appinsights or api-management-api-logger-appinsights to be prepared for additional loggers.

evertonmc:
> Potential Terraform Configuration
[...]

PR #4836, opened 2019-11-08
New Resource: `azurerm_api_management_diagnostic `

aqche:
Partially addresses: #4760

Adds the `azurerm_api_management_diagnostic` resource to the extent that the `2018-01-01` version of the `apimanagement` service allows. As noted in my comment on the feature request, I realize this doesn't have near the amount of options as the `2019-01-01` version would, but perhaps it can provide incremental value/serve as a step to addressing the issue. That being said, this is my first attempt at contributing a new resource to the provider so please let me know if you have any feedback! :)

```
--- PASS: TestAccAzureRMApiManagementDiagnostic_basic (2122.14s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm       2122.193s
```

```
--- PASS: TestAccAzureRMApiManagementDiagnostic_requiresImport (2911.42s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm       2911.465s
```

### DIFF
--- a/azurerm/resource_arm_api_management_diagnostic.go
+++ b/azurerm/resource_arm_api_management_diagnostic.go
@@ -127,7 +127,7 @@ func resourceArmApiManagementDiagnosticRead(d *schema.ResourceData, meta interfa
 	d.Set("api_management_name", serviceName)
 
 	if props := resp.DiagnosticContractProperties; props != nil {
-		d.Set("enabled", *props.Enabled)
+		d.Set("enabled", props.Enabled)
 	}
 
 	return nil
--- a/azurerm/resource_arm_api_management_diagnostic.go
+++ b/azurerm/resource_arm_api_management_diagnostic.go
@@ -36,6 +36,7 @@ func resourceArmApiManagementDiagnostic() *schema.Resource {
 			"identifier": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 				ValidateFunc: validation.StringInSlice([]string{
 					"applicationinsights",
 				}, false),
--- a/azurerm/resource_arm_api_management_diagnostic.go
+++ b/azurerm/resource_arm_api_management_diagnostic.go
@@ -1,0 +1,156 @@
+package azurerm
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/apimanagement/mgmt/2018-01-01/apimanagement"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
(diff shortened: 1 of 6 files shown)


azurerm/resource_arm_api_management_diagnostic.go
```diff
@@ -0,0 +1,156 @@
...
+			Create: schema.DefaultTimeout(30 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"identifier": {
+				Type:     schema.TypeString,
+				Required: true,
```
tombuildsstuff: presumably this is also:

```suggestion
				Required: true,
				ForceNew: true,
```

website/docs/r/api_management_diagnostic.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+  api_management_name = "${azurerm_api_management.test.name}"
+  enabled             = true
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `identifier` - (Required) The diagnostic identifier for the API Management Service. Allowed values are `applicationinsights`.
```
tombuildsstuff: Could we make this:

```suggestion
* `identifier` - (Required) The diagnostic identifier for the API Management Service. At this time the only supported value is `applicationinsights`. Changing this forces a new resource to be created.
```

website/docs/r/api_management_diagnostic.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - The ID of the API Management Diagnostic.
+
+## Import
+
+API Management Diagnostics can be imported using their `resource id`, e.g.
```
tombuildsstuff: to match this with other resources could we make this:

```suggestion
API Management Diagnostics can be imported using the `resource id`, e.g.
```

tombuildsstuff:
@aqche we're preparing to start some fairly large-scale refactoring to the codebase - as such I'm going to push a commit to fix the issues above and rebase this, so that we can get this merged before that happens; I hope you don't mind!

tombuildsstuff:
Tests pass:

Screenshot 2019-12-18 at 22 35 57

aqche:
@tombuildsstuff thanks for the review and taking the time to make the changes as well! 🙏

Issue #3514, opened 2019-05-23
App Configuration resources


Sahasrara:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Please create app_configuration resources so that app configuration stores can be Terraformed.

### New or Affected Resource(s)

* azurerm_app_configuration_store

### Potential Terraform Configuration

```
resource "azurerm_app_configuration_store" "app_config" {
  name                = 
  subscription   = 
  location            =
  resource_group_name =
}
```

brennerm:
Started implementing the new resource at #4859. Feel free to add your comments/wishes there.

mbfrahry:
Thanks for opening this @Sahasrara and for doing the legwork @brennerm.

This has been released in [version 1.39.0 of the provider](https://github.com/terraform-providers/terraform-provider-azurerm/blob/v1.39.0/CHANGELOG.md). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading. As an example:
```hcl
provider "azurerm" {
    version = "~> 1.39.0"
}
# ... other configuration ...
```

PR #4859, opened 2019-11-12
New Resource: `azurerm_app_configuration`

brennerm:
This PR adds the resource `azurerm_app_configuration` that allows creating, deleting and updating Azure App Configuration instances.

Related issue: #3514


brennerm:
Tests and documentation are missing as I would like to agree on the implementation first.

brennerm:
Has been validated with this Terraform config:
```
  name = "test"
  location = "West Europe"
}

resource "azurerm_app_configuration" "test" {
  name = "test"
  location = azurerm_resource_group.test.location
}

output "test" {
  value = azurerm_app_configuration.test
}
```

azurerm/resource_arm_app_configuration.go
```diff
@@ -0,0 +1,320 @@
...
+			},
+
+			"tags": tags.Schema(),
+		},
+	}
+}
+
+func resourceArmAppConfigurationCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).AppConfiguration.AppConfigurationsClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*ArmClient).StopContext, d)
```
mbfrahry: We have a timeout specifically for Create. This is minor but should be swapped for consistency. 
```suggestion
	ctx, cancel := timeouts.ForCreate(meta.(*ArmClient).StopContext, d)
```

azurerm/resource_arm_app_configuration.go
```diff
@@ -0,0 +1,320 @@
...
+		errors = append(errors, fmt.Errorf("%q may only contain alphanumeric characters and dashes and must be between 5-50 chars", k))
+	}
+
+	return warnings, errors
+}
+
+func makeAccessKeyMap(value appconf.APIKey) map[string]string {
+	m := make(map[string]string)
+
+	m["id"] = *value.ID
```
mbfrahry: We need to check that these values aren't nil before setting them, otherwise we could panic

azurerm/resource_arm_app_configuration.go
```diff
@@ -0,0 +1,320 @@
...
+	resultPage, err := client.ListKeys(ctx, resourceGroup, name, "")
+	if err != nil {
+		return fmt.Errorf("Failed to receive access keys for App Configuration %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	values := resultPage.Values()
+	for _, value := range values {
+		accessKey := makeAccessKeyMap(value)
+
+		if strings.HasPrefix(*value.Name, "Primary") {
```
mbfrahry: We should nil check anytime there is a pointer `*` in front of a value (unless we've already check for nil there). If that value is nil, we panic when we try and get the address of it through `*` which causes us to potentially lose statefiles.

azurerm/resource_arm_app_configuration.go
```diff
@@ -0,0 +1,320 @@
...
+				}, false),
+			},
+
+			"endpoint": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"primary_read_key": {
+				Type:      schema.TypeMap,
```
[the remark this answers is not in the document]
brennerm: Hey @mbfrahry, thanks for the review. Would you like to keep the differentiation between primary and secondary keys? If so which way would you prefer:

1. use two different list, one for primary, one for secondary keys
2. add an attribute that tells if this key is primary or secondary

Feel free to suggest another solution.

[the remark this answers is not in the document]
brennerm: After thinking about this I'm not sure if using a list makes it easier for users to access the keys.

Right now the interface for accessing the primary read key secret looks like this: `azurerem_app_configuration.appconf.primary_read_key.secret`

How should the interface look like after switching to a list?

azurerm/resource_arm_app_configuration.go
```diff
@@ -0,0 +1,320 @@
...
+
+	fut, err := client.Delete(ctx, resourceGroup, name)
+	if err != nil {
+		if response.WasNotFound(fut.Response()) {
+			return nil
+		}
+		return fmt.Errorf("Error deleting App Configuration %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	resp, err := fut.Result(*client)
```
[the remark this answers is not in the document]
brennerm: Not sure, good catch. :+1:

azurerm/resource_arm_app_configuration.go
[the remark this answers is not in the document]
mbfrahry: I believe all the logic you have can stay around dictating what key is being referred to. I'm just suggesting we change what object we're packaging it into since a Map is used for when we don't know the keys. Switching to a List would like:

```
"primary_write_key": {
	Type:      schema.TypeList,
        MaxItems: 1,
	Computed:  true,
	Elem: &schema.Schema{
                          "secret": {
				Type:     schema.TypeString,
				Computed: true,
				Sensitive: true,
			},
			"connection_string": {
				Type:     schema.TypeString,
				Computed: true,
				Sensitive: true,
                        },
                },
	},
},
```
I'd argue that `id` isn't necessary unless we're passing it along to another resource.

azurerm/resource_arm_app_configuration.go
brennerm: `nil` check is now done beforehand.

azurerm/resource_arm_app_configuration.go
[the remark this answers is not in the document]
brennerm: From my understand the id is required to connect to the App configuration as well as the secret. At least the connection string contains both of them.

brennerm:
@mbfrahry I have another branch adding a new data source to get a value from an App Configuration. Are you OK with me adding these changes to this PR as both use the same App Configuration client code? Otherwise I'd create a separate PR.

mbfrahry:
Hey @brennerm, adding it here should be fine

azurerm/resource_arm_app_configuration.go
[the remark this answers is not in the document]
mbfrahry: Makes sense to me! Let's keep `id` in

brennerm:
Done

azurerm/data_source_app_configuration_value.go
```diff
@@ -0,0 +1,98 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"content_type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"last_modified": {
```
mbfrahry: I normally like to steer away from time based values being held in Terraform since they tend to be more clutter rather than a benefit. Is there a specific use case you envision this being used in? If not, would you be ok with removing it?

brennerm: Agree :+1:

brennerm:
@mbfrahry Apparently Azure allows upgrading from the `free` to the `standard` SKU but downgrading is not supported. The API responds with the following response:
`StatusCode=400 -- Original Error: Code="SkuDowngradeNotAllowed" Message="ErrorSkuDowngradeNotAllowed"`

How do you normally handle this? Make changing the SKU create a new resource by default or do we allow an upgrade in-place and tell the user that downgrading is not supported?

mbfrahry:
Ohhh, that's a tricky one. We have a way to customize the diff of a resource depending on certain conditions but I don't think we should use that here. 

The API leaves a clear enough message on what is going on when trying to downgrade a sku that a user could act on and deal with themselves. Force newing on our end might add more confusion since we wouldn't be able to leave as helpful of a message as the api about why we need to force new when downgrading. 

I think leaving what to do in the hand of the user is fine here.

brennerm:
Alright, will add a hint to the documentation.

brennerm:
I'm having trouble writing the acceptance tests for the data source as the Azure API currently does not support creating values in an App Configuration. I'd propose to open an issue to make sure the tests are implemented as soon as the API and the SDK has been extended.

mbfrahry:
Shoot, unfortunately, we rarely merge things unless we can test them automatically/nightly since things can go horribly wrong. As such, I think we'll have to pull that datasource out and put it into another PR and keep it in holding until we the api supports creating values in an App Configuration.

brennerm:
@mbfrahry Acceptance tests for the app configuration resource have been added and the data source has been removed from the repo. Can I see the results of the acceptance tests somewhere? I'm not having access to an active Azure subscription so I can't execute them.

katbyte:
@brennerm,

the tests are failing with:
```
------- Stdout: -------
=== RUN   TestAccAzureAppConfiguration_free
=== PAUSE TestAccAzureAppConfiguration_free
=== CONT  TestAccAzureAppConfiguration_free
--- FAIL: TestAccAzureAppConfiguration_free (135.92s)
    testing.go:569: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=3) "sku": (string) (len=4) "free"
        }
        
FAIL

------- Stderr: -------
2019/11/28 19:40:38 [DEBUG] Registering Data Sources for "Compute"..
2019/11/28 19:40:38 [DEBUG] Registering Resources for "Compute"..
```

looks like you need to set the SKU on read

brennerm:
Should be fixed. :+1:

brennerm:
The schema version could be set to 0 as this is the initial version of this resource, right?

azurerm/resource_arm_app_configuration.go
```diff
@@ -0,0 +1,385 @@
...
+			},
+
+			"endpoint": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"primary_read_key": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
```
mbfrahry: This can be removed as `MaxItems` is a validation check for what a user submits rather than a check that the api only returns 1 of this. This goes for `MaxItems` on the blocks down below as well

azurerm/resource_arm_app_configuration.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	resultPage, err := client.ListKeys(ctx, resourceGroup, name, "")
+	if err != nil {
+		return fmt.Errorf("Failed to receive access keys for App Configuration %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	values := resultPage.Values()
+	for _, value := range values {
+		if &value == nil {
```
mbfrahry: Unfortunately, this isn't the right check for nil. `value` is actually fine in this case as it's just an APIKey but the individual attributes in an APIKey are pointers `*`. Those need to be nil checked before pointing to them

azurerm/resource_arm_app_configuration.go
brennerm: Just out of curiosity, I got this from [resource_arm_kubernetes_cluster.go](https://github.com/terraform-providers/terraform-provider-azurerm/blob/fc2de5e6850eb5a7ed68a030c67e11fe83fb40f0/azurerm/resource_arm_kubernetes_cluster.go#L525). Is there some reason why it's necessary there?

mbfrahry:
Hey @brennerm. That nil check still isn't quite right. We want to still set whatever variables we have access too. I went ahead and wrote what it should look like up and merged with master since there were some tricky conflicts but for some reason I can't commit my changes to your branch. I opened a PR on your end with master changes and my three commits on top. Apologies for all the changes coming in!  https://github.com/brennerm/terraform-provider-azurerm/pull/1

brennerm:
@mbfrahry Understand, thought we are going with the all or nothing approach and ignore incomplete access keys which probably shouldn't happen at all.

azurerm/resource_arm_app_configuration.go
mbfrahry: Not that I'm aware! It might just be a mistake. That validation is only checked against what a user submits not what the api returns.

mbfrahry:
Thanks for your patience here @brennerm. This should go out the door in the next release

website/docs/r/app_configuration.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+subcategory: "App Configuration"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_app_configuration"
+sidebar_current: "docs-azurerm-resource-app-configuration"
+description: |-
+  Manages an Azure App Configuration.
+
+---
+
+# azurerm_container_registry
```
devblackops: Copy/paste error here? This should be `#azurerm_app_configuration`

brennerm: Correct, fixed with #5186.

PR #4453, opened 2019-09-27
`azurerm_image`: add option to set the hyper_v_generation for the image

abhinavdahiya:
The new clients require this information see [properties.hyperVGeneration][1] and it supports 2 values `V1` and `V2` [2]

the default is `V1` in azure cli `--hyper-v-generation` [3], so it should be safe to keep that as default.

[1]: https://docs.microsoft.com/en-us/rest/api/compute/images/createorupdate#image
[2]: https://docs.microsoft.com/en-us/rest/api/compute/images/createorupdate#hypervgenerationtypes
[3]: https://docs.microsoft.com/en-us/cli/azure/image?view=azure-cli-latest#optional-parameters

Fixes: https://github.com/terraform-providers/terraform-provider-azurerm/issues/4361

PR #4865, opened 2019-11-13
Support for StorageAccountType on azurerm_shared_image_version

### LINKED ISSUE #4833: azurerm_shared_image_version - Please add missing 'storage_account_type' to support ZRS
### Community Note

### Description

Azure Shared Image Versions supports Zone-Redundant storage (Standard_ZRS). 

This is a must-have if you are running a VM Scale Set that uses ZRS for higher availability. Otherwise, if the zone with the image version fails, the VMSS will fail even though that is available in multiple zones.

Please add this attribute to the AzureRM provider.

azurerm_shared_image_version

Please add to the target_region block an optional:
```
storage_account_type = "ZRS"         # <--- Should translate to "storageAccountType":
                                                             # "Standard_ZRS" in AzureRM
```

```
resource "azurerm_shared_image_version" "my_image" {
  name   = "myimage"
  ...
  target_region {
    name                                     = "EU West"
    regional_replica_count         = "1"
    storage_account_type          = "ZRS"      # <---- Not supported by the AzureRM provider yet
  }
}
```

cwebbtw:
Work in progress fix for https://github.com/terraform-providers/terraform-provider-azurerm/issues/4833

Have made changes to support storage account type in:

- Data source
- Resource
- Documentation
- Acceptance test for ZRS/LRS storage account type

I've made some changes to `azurerm_shared_image_version` to support a `storage_account_type` field, however I'm running into problems running the acceptance tests, which seem to be failing against origin/master. Running tests with:

`make testacc TESTARGS='-run=TestAccAzureRMSharedImageVersion'`

Other people seem to be having similar issues and have active PRs open, see https://github.com/terraform-providers/terraform-provider-azurerm/pull/4453.

```
--- FAIL: TestAccDataSourceAzureRMSharedImageVersion_basic (912.05s)
    testing.go:569: Step 1 error: errors during apply:
        
        Error: compute.ImagesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="Required parameter 'hyperVGeneration' is missing (null)." Target="hyperVGeneration"
        
          on /var/folders/v8/cmph72_n7lvbpnbp2g5ssc0m0000gn/T/tf-test596992779/main.tf line 102:
          (source code not available)
        
        
        
        Error: Code="TargetDiskBlobAlreadyExists" Message="Blob https://accsa191113120502781918.blob.core.windows.net/vhds/myosdisk1.vhd already exists. Please provide a different blob URI as target for disk 'myosdisk1'."
        
          on /var/folders/v8/cmph72_n7lvbpnbp2g5ssc0m0000gn/T/tf-test596992779/main.tf line 64:
          (source code not available)
```

As such, I'm not confident the changes made will support a fix for https://github.com/terraform-providers/terraform-provider-azurerm/issues/4833. It would be good if someone can *review* and make suggestions based on the current inability to run tests.

Fixes #4833


azurerm/data_source_shared_image_version_test.go
```diff
@@ -37,6 +37,16 @@ func TestAccDataSourceAzureRMSharedImageVersion_basic(t *testing.T) {
...
+				),
+			},
+			{
+				Config: testAccDataSourceSharedImageVersion_zrs(rInt, location, username, password, hostname),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(dataSourceName, "tags.%", "0"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "managed_image_id"),
+					resource.TestCheckResourceAttr(dataSourceName, "target_region.#", "1"),
+					resource.TestCheckResourceAttr(dataSourceName, "target_region.0.storage_account_type", "Standard_ZRS"),
 				),
```
tombuildsstuff: since the Storage Account Type within the Resource is ForceNew this probably wants to be a distinct test in it's own right - could we split this out into it's own test?

azurerm/resource_arm_shared_image_version.go
```diff
@@ -282,6 +290,8 @@ func flattenSharedImageVersionTargetRegions(input *[]compute.TargetRegion) []int
 				output["regional_replica_count"] = int(*v.RegionalReplicaCount)
 			}
 
+			output["storage_account_type"] = v.StorageAccountType
```
tombuildsstuff: could we cast this to a string here:
```suggestion
			output["storage_account_type"] = string(v.StorageAccountType)
```

azurerm/resource_arm_shared_image_version_test.go
```diff
@@ -217,6 +217,27 @@ resource "azurerm_shared_image_version" "test" {
 }
 `, template)
 }
+func testAccAzureRMSharedImageVersion_imageVersionZrs(rInt int, location, username, password, hostname string) string {
```
tombuildsstuff: could we add a test for this in the Resource too, provisioning this in a standalone test (as in the other cases above)

tombuildsstuff:
@cwebbtw 

> I've made some changes to azurerm_shared_image_version to support a storage_account_type field, however I'm running into problems running the acceptance tests, which seem to be failing against origin/master. Running tests with:

There's a PR tracking a fix for this in #4453 but unfortunately there's a bug in it we've not been able to track down just yet - at this point this PR otherwise looks good though, so I don't think that's a blocker for this PR (although we do need to dive deeper into #4453 so we can ship that fix)

azurerm/data_source_shared_image_version_test.go
cwebbtw: I would agree, there needs to be an existing case where the resource exists and a change forces the resource to be re-created. I believe I was right in the understanding that changing the storage account requires a new resource to be created?

cwebbtw:
> hey @cwebbtw
> 
> Thanks for this PR :)
> 
> Taking a look through this is looking good - if we can fix up the minor comments (and the tests pass) then this otherwise LGTM 👍
> 
> Thanks!

I will make the changes and will watch the tracker for when the tests pass again on master, will merge the changes to origin to my branch and verify that the tests for adding `storage_account_type` pass. I'm keen on waiting for the tests to pass before we risk a PR merge, as I'm sure is your intention by the test passing comment 👍

MarcelT-NL:
Just wanted to reach out and give you a big "THANK YOU" for the quick manner in which you guys picked my enhancement request up! I will finally be able to run my terraform script again soon :)

MarcelT-NL:
Hi @cwebbtw, 

I am a little confused about the process here - is this fix included in 1.37 and if so, where can I find the ETA of that release? Any comments on this would be appreciated!

Cheers, Marcel

tombuildsstuff:
@cwebbtw 

> I will make the changes and will watch the tracker for when the tests pass again on master, will merge the changes to origin to my branch and verify that the tests for adding storage_account_type pass. I'm keen on waiting for the tests to pass before we risk a PR merge, as I'm sure is your intention by the test passing comment 👍

Since these issues are unrelated I don't think we're necessarily blocked on #4453 (albeit that needs to be fixed prior to shipping) - so providing the tests pass here we should be otherwise good - I'll take a look through now 👍

azurerm/resource_arm_shared_image_version_test.go
```diff
@@ -61,6 +61,58 @@ func TestAccAzureRMSharedImageVersion_basic(t *testing.T) {
...
+				),
+			},
+			{
+				Config: testAccAzureRMSharedImageVersion_imageVersionStorageAccountType(ri, testLocation(), userName, password, hostName, "Standard_ZRS"),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMSharedImageVersionExists(resourceName),
+					resource.TestCheckResourceAttrSet(resourceName, "managed_image_id"),
+					resource.TestCheckResourceAttr(resourceName, "target_region.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "target_region.0.storage_account_type", "Standard_ZRS"),
+				),
```
tombuildsstuff: running the test this fails because the Blob still exists - could we split remove the `Standard_LRS` section above so that this only provisions a ZFS account (and update the test name?)

azurerm/data_source_shared_image_version_test.go
tombuildsstuff: since that's a feature of the Resource rather than the Data Source - can we update the Data Source to only read this value from the Resource? If we wish to confirm this it'd be better to add a separate test for a data source from  a ZFS image, but I don't think that's necessary since it'll be covered by the resource here?

azurerm/resource_arm_shared_image_version_test.go
cwebbtw: I can do this, however I ran the acceptance tests for the `azure_shared_image_version` against origin/master and get a failure due to the errors:

```
        Error: Code="TargetDiskBlobAlreadyExists" Message="Blob https://accsa191113120502781918.blob.core.windows.net/vhds/myosdisk1.vhd already exists. Please provide a different blob URI as target for disk 'myosdisk1'."
        
          on /var/folders/v8/cmph72_n7lvbpnbp2g5ssc0m0000gn/T/tf-test596992779/main.tf line 64:
          (source code not available)
```

cwebbtw: The issue has been that I didn't want to make changes that would affect other resources/data tests pass as part of this PR given I wasn't confident that I was running the acceptance tests in the correct way due to them failing on remote.

Happy to make the changes here though, but I would just double check to confirm that they are actually failing on master, or perhaps I have a local configuration issue?

azurerm/data_source_shared_image_version_test.go
cwebbtw: Apologies @tombuildsstuff, I'm not sure I follow what you mean here. Given we can test that the data source is returning a correct value for the `storage_account_type` attribute (LRS) in the second test step, I've actually removed the third test step and the corresponding resources as it feels like it's not testing anything we don't already test.

tombuildsstuff:
hey @cwebbtw 

Thanks for this PR - apologies for the delayed re-review here!

Taking a look through here it appears there's a few minor things needed to get this merged; namely treating the `storage_account_type` field as a Computed field (since it's optional, but has a default value) and removing the assertions from the tests (since the block `target_region` is a Set rather than a List - thus testing this is easiest via an Import test).

Since this also requires a rebase (to pick up some changes required to make the tests pass) - I hope you don't mind but since this is from your `master` branch we'd rather not force-push to this (in addition there's a known bug in Github around force-pushing to a master branch with branch protection where force-pushing ends up breaking the PR).

As such whilst I'd like to thank you for this contribution, I'm going to close this in favour of #5212 - which includes your commits from this PR (so you'll still get credit for the contributions) plus the necessary changes to allow us to get this merged - I hope you don't mind!

Thanks!

PR #4876, opened 2019-11-14
`azurerm_eventgrid_domain` - Export `primary_access_key` and `secondary_access_key`

hbuckle:


hbuckle:
@mbfrahry - yes indeed, docs updated

Issue #3863, opened 2019-07-17
azurerm_api_management_api lacks ability to set "Subscription required" flag to false


HKochniss:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

1.30

### Affected Resource(s)

* `azurerm_api_management_api`

```hcl
resource "azurerm_api_management_api" "example" {
  depends_on          = ["azurerm_api_management.mgmttest"]
  name                = "example-api"
  resource_group_name = "${azurerm_resource_group.rg.name}"
  api_management_name = "${azurerm_api_management.mgmttest.name}"
  revision            = "1"
  display_name        = "Example API"
  path                = "example"
  protocols           = ["https"]
  service_url         = "https://some_url"
```

### Expected Behavior

By default the `Subscription required` flag should be set to **false** if no `subscription_key_parameter_names` are provided *or* there should be an argument `subscription_required` exposed for this flag. Right now **it seems impossible** to create an API Management Api without a required subscription (thus allowing simple passthrough/proxy scenarios).

### Actual Behavior

Creating an API management always sets the `Subscription required` flag in the portal to *true*

### Steps to Reproduce

Create any `azurerm_api_management` with any `azurerm_api_management_api`.

1. `terraform apply`

### Important Factoids

none

### References
https://www.terraform.io/docs/providers/azurerm/r/api_management_api.html#subscription_key_parameter_names
shows no represenation for the flag
![apim_subscription_flag](https://user-images.githubusercontent.com/1243055/61374233-258ef080-a89c-11e9-9e08-753530437837.png)

* #0000

cdesewell:
Also mentioned here: https://github.com/terraform-providers/terraform-provider-azurerm/issues/3649

HKochniss:
ok, thanks. Still think this should be a separate issue

AndiA92:
hi @tombuildsstuff , is there any plan on implementing this? As @HKochniss pointed out it seems impossible to create an API endpoint without a subscription key.

brunhil:
[...]

Issue #2613, opened 2019-01-07
API Management Consumption Tier


scottrangerio:
Hey all,

API Management has a consumption tier which is currently in preview. It would be great to see the azurerm_api_management resource updated to support the additional tier.

I've had a quick look and the property does exist in the Azure sdk https://github.com/Azure/azure-sdk-for-go/blob/master/services/preview/apimanagement/mgmt/2018-06-01-preview/apimanagement/models.go#L510 so it appears this should just be a case of updating the vendored package, updating the schema to include the additional tier and then testing it.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected Resource(s)

* azurerm_api_management

tombuildsstuff:
Dependent on https://github.com/Azure/azure-rest-api-specs/issues/4409 (since this exposes the new API/SDK version required)

tombuildsstuff:
👋 

The upstream issue's been fixed: https://github.com/Azure/azure-rest-api-specs/issues/4409 - once the SDK's been updated #3335 we should be able to go to the new version of the API Management API - at which point it should be possible to add support for this

Thanks!

mikhailshilkov:
[Azure API Management Consumption Tier is generally available](https://azure.microsoft.com/en-us/updates/azure-api-management-consumption-tier-is-now-generally-available/) since May 29

tminhein:
Is Terraform stopped supporting Azure.  a lot of new features aren't available in terraform.  I.E: 
1. Consumption Tier in Azure apim
2. throughput option in cosmos db database layer.

Those are already reported in Github months and months ago.

tombuildsstuff:
@tminhein this functionality requires upgrading to the new version of the API Management API (2019-01-01) - which unfortunately [doesn't match the API response](https://github.com/Azure/azure-rest-api-specs/pull/6065) - as such we're unable to add support for this until the API/SDK is fixed.

stena-dev:
[...]

PR #4885, opened 2019-11-14
azurerm_api_management_api - support for `subscription_required`

brunhil:
Add `subscription_required` to data and resource for **azurerm_api_management_api**

(fixes ##3863 )

### DIFF
--- a/website/docs/d/api_management_api.html.markdown
+++ b/website/docs/d/api_management_api.html.markdown
@@ -57,6 +57,8 @@ output "api_management_api_id" {
 
 * `subscription_key_parameter_names` - A `subscription_key_parameter_names` block as documented below.
 
+* `subscription_required` - Should this API require a subscription key?
+
 * `version` - The Version number of this API, if this API is versioned.
 
 * `version_set_id` - The ID of the Version Set which this API is associated with.
--- a/website/docs/d/api_management_api.html.markdown
+++ b/website/docs/d/api_management_api.html.markdown
@@ -65,12 +65,6 @@ output "api_management_api_id" {
 
 ---
 
-An `authentication_settings` block exports the following:
-
-* `subscription_key_required` - Whether the API requires a subscription key or not.
-
----
-
 A `subscription_key_parameter_names` block exports the following:
 
 * `header` - The name of the HTTP Header which should be used for the Subscription Key.
--- a/website/docs/d/api_management_api.html.markdown
+++ b/website/docs/d/api_management_api.html.markdown
@@ -65,6 +65,12 @@ output "api_management_api_id" {
 
 ---
 
+An `authentication_settings` block exports the following:
+
+* `subscription_key_required` - Whether the API requires a subscription key or not.
+
+---
+
 A `subscription_key_parameter_names` block exports the following:
 
 * `header` - The name of the HTTP Header which should be used for the Subscription Key.
--- a/azurerm/internal/services/apimanagement/api_management_api_data_source.go
+++ b/azurerm/internal/services/apimanagement/api_management_api_data_source.go
@@ -99,6 +99,11 @@ func dataSourceApiManagementApi() *schema.Resource {
 				},
 			},
 
+			"subscription_required": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+
 			"version": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -147,6 +152,7 @@ func dataSourceApiManagementApiRead(d *schema.ResourceData, meta interface{}) er
 		d.Set("revision", props.APIRevision)
 		d.Set("service_url", props.ServiceURL)
 		d.Set("soap_pass_through", string(props.APIType) == string(apimanagement.SoapPassThrough))
+		d.Set("subscription_required", props.SubscriptionRequired)
 		d.Set("version", props.APIVersion)
 		d.Set("version_set_id", props.APIVersionSetID)
 
--- a/azurerm/internal/services/apimanagement/api_management_api_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_api_resource.go
@@ -163,6 +163,12 @@ func resourceArmApiManagementApi() *schema.Resource {
 				},
 			},
 
+			"subscription_required": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+			},
+
 			"soap_pass_through": {
 				Type:     schema.TypeBool,
 				Optional: true,
@@ -292,6 +298,7 @@ func resourceArmApiManagementApiCreateUpdate(d *schema.ResourceData, meta interf
 	description := d.Get("description").(string)
 	displayName := d.Get("display_name").(string)
 	serviceUrl := d.Get("service_url").(string)
+	subscriptionRequired := d.Get("subscription_required").(bool)
 
 	protocolsRaw := d.Get("protocols").(*schema.Set).List()
 	protocols := expandApiManagementApiProtocols(protocolsRaw)
@@ -310,6 +317,7 @@ func resourceArmApiManagementApiCreateUpdate(d *schema.ResourceData, meta interf
 			ServiceURL:                    utils.String(serviceUrl),
 			SubscriptionKeyParameterNames: subscriptionKeyParameterNames,
 			APIVersion:                    utils.String(version),
+			SubscriptionRequired:          &subscriptionRequired,
 		},
 	}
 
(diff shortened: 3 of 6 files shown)


azurerm/resource_arm_api_management_api.go
```diff
@@ -454,3 +477,30 @@ func flattenApiManagementApiSubscriptionKeyParamNames(paramNames *apimanagement.
...
 	return []interface{}{result}
 }
+
+func expandApiManagementAuthenticationSettings(input []interface{}) *apimanagement.AuthenticationSettingsContract {
+	if len(input) == 0 {
+		return nil
+	}
+
+	v := input[0].(map[string]interface{})
+	subscriptionKeyRequired := v["subscriptionKeyRequired"].(bool)
```
mbfrahry:
```suggestion
	subscriptionKeyRequired := v["subscription_key_required"].(bool)
```

website/docs/d/api_management_api.html.markdown
```diff
@@ -58,12 +58,20 @@ output "api_management_api_id" {
...
+
 * `version` - The Version number of this API, if this API is versioned.
 
 * `version_set_id` - The ID of the Version Set which this API is associated with.
 
 ---
 
+An `authentication_settings` block exports the following:
+
+* `subscription_key_required` - Should this API require a subscription key?
```
mbfrahry:
```suggestion
* `subscription_key_required` - Whether the API requires a subscription key or not.
```

brunhil:
Good catch with the spelling @mbfrahry , it was one of those days yesterday -_-. I'll get the tests added and other changes fixed and push them up.

azurerm/resource_arm_api_management_api_test.go
```diff
@@ -28,6 +28,8 @@ func TestAccAzureRMApiManagementApi_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "soap_pass_through", "false"),
 					resource.TestCheckResourceAttr(resourceName, "is_current", "true"),
 					resource.TestCheckResourceAttr(resourceName, "is_online", "false"),
+					resource.TestCheckResourceAttr(resourceName, "authentication_settings.#", "1"),
```
brunhil: This is set by default on creation of APIs for APIM, so I thought it was easiest to verify on the "basic" test. If you want an explicit test let me know.

azurerm/resource_arm_api_management_api.go
```diff
@@ -454,3 +477,30 @@ func flattenApiManagementApiSubscriptionKeyParamNames(paramNames *apimanagement.
 
 	return []interface{}{result}
 }
+
+func expandApiManagementAuthenticationSettings(input []interface{}) *apimanagement.AuthenticationSettingsContract {
+	if len(input) == 0 {
+		return nil
+	}
+
+	v := input[0].(map[string]interface{})
```
mbfrahry: This caused a test to panic so we should confirm that `input[0] != nil` before casting this to a `map[string]interface{}`

```

------- Stdout: -------
=== RUN   TestAccAzureRMApiManagementApi_importUpdate

------- Stderr: -------
2019/11/15 17:21:04 [DEBUG] Registering Data Sources for "Compute"..
2019/11/15 17:21:04 [DEBUG] Registering Resources for "Compute"..
panic: interface conversion: interface {} is nil, not map[string]interface {}

goroutine 1247 [running]:
github.com/terraform-providers/terraform-provider-azurerm/azurerm.expandApiManagementAuthenticationSettings(...)
	/opt/teamcity-agent/work/458e5e4800bd94f6/src/github.com/terraform-providers/terraform-provider-azurerm/azurerm/resource_arm_api_management_api.go:486
github.com/terraform-providers/terraform-provider-azurerm/azurerm.resourceArmApiManagementApiCreateUpdate(0xc0002c9490, 0x3b50f20, 0xc0014fe000, 0x0, 0x0)
```

azurerm/resource_arm_api_management_api_test.go
```diff
@@ -28,6 +28,8 @@ func TestAccAzureRMApiManagementApi_basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, "authentication_settings.0.subscription_key_required", "true"),
```
mbfrahry: This test now errors. It doesn't look like this is being set properly
```
------- Stdout: -------
=== RUN   TestAccAzureRMApiManagementApi_basic
--- FAIL: TestAccAzureRMApiManagementApi_basic (2612.17s)
    testing.go:569: Step 0 error: Check failed: Check 6/6 error: azurerm_api_management_api.test: Attribute 'authentication_settings.0.subscription_key_required' not found
FAIL
```

azurerm/resource_arm_api_management_api_test.go
```diff
@@ -120,6 +122,33 @@ func TestAccAzureRMApiManagementApi_version(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMApiManagementApi_authenticationSettings(t *testing.T) {
```
mbfrahry: This test is also failing

```
=== RUN   TestAccAzureRMApiManagementApi_authenticationSettings
--- FAIL: TestAccAzureRMApiManagementApi_authenticationSettings (2825.43s)
    testing.go:569: Step 0 error: Check failed: Check 3/3 error: azurerm_api_management_api.test: Attribute 'authentication_settings.0.subscription_key_required' not found
FAIL
```

azurerm/resource_arm_api_management_api.go
```diff
@@ -159,6 +159,21 @@ func resourceArmApiManagementApi() *schema.Resource {
 				},
 			},
 
+			"authentication_settings": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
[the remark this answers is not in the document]
brunhil: So this actually is a misunderstanding on my part...I'll explain more in the message on this PR.

brunhil:
After digging further into this, it turns out we have a few approaches we can take with this.

The 'AuthenticationSettings' field is a 2019-01-01 API Version property. So these changes are actually for 2019-01-01. I was implementing this when looking at the REST API [here](https://docs.microsoft.com/en-us/rest/api/apimanagement/2019-01-01/apis/createorupdate). I had modified the try-it option to point at 2018-01-1 and assumed the documentation on the left updated, it did not :).

After testing/verifying in 2018-01-01, the response body is as follows:

```
{
  "id": "...",
  "type": "Microsoft.ApiManagement/service/apis",
  "name": "test-api-name",
  "properties": {
    "displayName": "test-api-name",
    "apiRevision": "1",
    "description": "",
    "subscriptionRequired": false,
```

Then attempting to set the value in the body for the REST API led to,

```
{
  "error": {
    "code": "ValidationError",
    "message": "SubscriptionRequired property is not supported for versions before 20180601.",
    "details": null
  }
}
```

As a result, I'm marking this PR as WIP. We would need to update to 2019-01-01 API version to support this field, and this is blocked by the upstream issues that #2613 is blocked by.

tombuildsstuff:
Assigning this to the "blocked" milestone since this is blocked on upgrading to 2019-01-01, which is blocked on https://github.com/Azure/azure-rest-api-specs/issues/6372

katbyte:
Looks like this should be close from the blocking issue
```
just yesterday we got signed all required PRs and now preparing the nuget package. Should be ready in nearest days.
``

usamaahmadkhan:
@brunhil can we have an estimated date when this will get released? Blocked as we want to consume this resource from terraform in our module

markti:
what about linking to an authorization_server (oauth / openidc) ??? is this included in this PR?

katbyte:
@brunhil - apim api as been updated on master so this should be unblocked now

brunhil:
Thanks for the note @katbyte , I'll start getting this up to date!

markti:
this should resolve issue #3863 no?

brunhil:
@markti - This PR was simply to expose the ability to turn off Subscription Settings, so that won't be added

markti:
> @markti - This PR was simply to expose the ability to turn off Subscription Settings, so that won't be added

Ok, no problem. This is a great submission as its very needed functionality. 

I'd like to implement the Oauth / OIDC setting...very eager to see this merged so I can work on that...

brunhil:
@markti - It turns out the AuthenticationSettings block isn't required any more for this PR as the **SubscriptionRequired** block is now at the top level.

brunhil:
@katbyte , when this was initially implemented Subscription Required was turned on by default. They may have changed their behavior in the new API version.

danderton:
We're also blocked with this current issue. Do you have an ETA?

brunhil:
Sorry all, forgot to hit push! The test should be fixed if it defaults to false. Let me know if there's any other requested changes. Thanks for reviewing!

3mard:
Thanks you all for this, can't wait for this to get merged

Issue #4319, opened 2019-09-13
Cognitive Services: support for `Luis.Authoring`


tommyJimmy87:
Would be cool to have the possibility to add the new Luis Authoring Cognitive Service and also the Web App Bot as possible resources.

tombuildsstuff:
hi @tommyJimmy87 

Thanks for opening this issue :)

Taking a look into this I believe that LUIS support is available via [the `kind` field within the `azurerm_cognitive_account ` resource](https://www.terraform.io/docs/providers/azurerm/r/cognitive_account.html) - would you be able to take a look and see if that works for you? In addition support for Bots has been merged and [will form part of the v1.34.0 release of the Azure Provider](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/CHANGELOG.md). Since this functionality appears to be supported I'm going to close this issue for the moment - however please let us know if that doesn't work for you and we'll take another look.

Thanks!

tommyJimmy87:
@tombuildsstuff  
Hi thanks for the update about the Bot Service. 

Regarding the LUIS Authoring is a different LUIS Cognitive Service where you can choose the Authoring API which will allow to share applications across different accounts in your organization. From the LUIS cognitive service kind in terraform I cannot see the option to choose the Authoring API.

https://docs.microsoft.com/en-us/azure/cognitive-services/luis/luis-concept-keys

<img width="944" alt="Screen Shot 2019-09-13 at 16 56 38" src="https://user-images.githubusercontent.com/21197220/64872502-9c036100-d647-11e9-88bb-54287a0cc43b.png">

tombuildsstuff:
@tommyJimmy87 so that we could understand that better, would you be able to show the generated ARM Template for that (you should be able to see it by clicking "Automated/Template Deployment" at the bottom of the create screen)? Thanks!

tommyJimmy87:
@tombuildsstuff Sure.

```{
    "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
        "name": {
            "type": "string"
        },
        "authoringLocation": {
            "type": "string"
        },
        "authoringSku": {
            "type": "string"
        }
    },
    "resources": [
        {
            "apiVersion": "2016-02-01-preview",
            "name": "[parameters('name')]",
            "location": "[parameters('authoringLocation')]",
[...]

PR #4888, opened 2019-11-15
Add LUIS.Authoring to cognitive services

enricsoler:
It solves issue #4319 adding LUIS.Authoring cognitive resource type to list and to the docs.


enricsoler:
I think this check failure is not related with this PR 🤔

tombuildsstuff:
@enricsoler 

> I think this check failure is not related with this PR 🤔

Indeed - as you mentioned this was resolved by #4887 but it's still showing as a failure because this isn't rebased on top of master which includes this - but this'll be fine, thanks again for this PR 👍🏻

PR #4889, opened 2019-11-15
New Resource & Data Source: `azurerm_netapp_pool`

neil-yechenwei:
This PR is code implement for issue https://github.com/terraform-providers/terraform-provider-azurerm/issues/3936


azurerm/resource_arm_netapp_pool.go
```diff
@@ -0,0 +1,220 @@
...
+					string(netapp.Standard),
+					string(netapp.Ultra),
+				}, false),
+				DiffSuppressFunc: suppress.CaseDifference,
+			},
+
+			"size": {
+				Type:         schema.TypeInt,
+				Required:     true,
+				ValidateFunc: validation.IntBetween(4398046511104, 549755813888000),
```
mbfrahry: I'd suggest we make this a little easier on users and have this be `size_in_4_tb` and then 
```suggestion
				ValidateFunc: validation.IntBetween(1, 30),
```
at that point we can multiply any number passed in by `4398046511104` when we read it and then divide by `4398046511104` when we set it.

azurerm/resource_arm_netapp_pool.go
```diff
@@ -0,0 +1,220 @@
...
+
+	id, err := azure.ParseAzureResourceID(d.Id())
+	if err != nil {
+		return err
+	}
+	resourceGroup := id.ResourceGroup
+	accountName := id.Path["netAppAccounts"]
+	name := id.Path["capacityPools"]
+
+	_, err = client.Delete(ctx, resourceGroup, accountName, name)
```
mbfrahry: I saw that this returns a future, is there a reason we're using `waitForNetAppPoolToBeDeleted` instead? I've noticed that futures aren't always accurate so I just wanted to confirm that that was the case here

azurerm/resource_arm_netapp_pool.go
```diff
@@ -0,0 +1,220 @@
...
+func netappPoolDeleteStateRefreshFunc(ctx context.Context, client *netapp.PoolsClient, resourceGroupName string, accountName string, name string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		res, err := client.Get(ctx, resourceGroupName, accountName, name)
+		if err != nil {
+			if !utils.ResponseWasNotFound(res.Response) {
+				return nil, "", fmt.Errorf("Error retrieving NetApp Pool %q (Resource Group %q): %s", name, resourceGroupName, err)
+			}
+		}
+
+		if _, err := client.Delete(ctx, resourceGroupName, accountName, name); err != nil {
```
mbfrahry: If the future isn't quite accurate, do you mind commenting about that and also comment about why we're deleting again here

azurerm/resource_arm_netapp_pool_test.go
```diff
@@ -0,0 +1,235 @@
...
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckAzureRMNetAppPoolDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMNetAppPool_basic(ri, location),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMNetAppPoolExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "service_level", "Premium"),
+				),
+			},
```
mbfrahry: Can we add another test import step here?

azurerm/resource_arm_netapp_pool_test.go
```diff
@@ -0,0 +1,235 @@
...
+			{
+				Config: testAccAzureRMNetAppPool_complete(ri, location),
```
mbfrahry: Can we change the size just to confirm we can update that as well?

website/docs/d/netapp_pool.html.markdown
```diff
@@ -0,0 +1,51 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `location` - The Azure Region where the NetApp Pool exists.
+
+* `service_level` - The service level of the file system.
+
+* `size` - Provisioned size of the pool (in bytes). Allowed values are in 4TiB chunks (value must be multiply of 4398046511104).
```
mbfrahry:
```suggestion
* `size` - Provisioned size of the pool (in bytes). Values are in 4TiB chunks.
```

website/docs/r/netapp_pool.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+
+* `name` - (Required) The name of the NetApp Pool. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the resource group where the NetApp Pool should be created. Changing this forces a new resource to be created.
+
+* `account_name` - (Required) The name of the NetApp account.
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `service_level` - (Required) The service level of the file system.
```
mbfrahry: We should note what the allowed values are:

```suggestion
* `service_level` - (Required) The service level of the file system. Valid values include `Premium`, `Standard`, or `Ultra`.
```

mbfrahry (changes requested):
Hey @neil-yechenwei, thanks for opening this PR. It's mostly there and the main thing to discuss would be `size` and making it a little more user friendly

azurerm/resource_arm_netapp_pool.go
katbyte: Also opening an issue on the SDK with a link here so we can track when its fixed

azurerm/data_source_netapp_pool.go
```diff
@@ -0,0 +1,88 @@
...
+		Read: dataSourceArmNetAppPoolRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: validation.StringMatch(
+					regexp.MustCompile(`^[\da-zA-Z_-]{4,64}$`),
+					`The name must be between 4 and 64 characters in length and contain only letters, numbers, underscore or hyphens.`,
+				),
```
katbyte: could we make this a shared function in the netapp package and share it with the resource?

azurerm/data_source_netapp_pool.go
```diff
@@ -0,0 +1,88 @@
...
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
+
+			"location": azure.SchemaLocationForDataSource(),
+
+			"account_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: validation.StringMatch(
```
katbyte: could we make this a shared function in the netapp package and share it with the resource and netapp account resource?

azurerm/data_source_netapp_pool.go
```diff
@@ -0,0 +1,88 @@
...
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	if poolProperties := resp.PoolProperties; poolProperties != nil {
+		if err := d.Set("service_level", poolProperties.ServiceLevel); err != nil {
+			return fmt.Errorf("Error setting `service_level`: %+v", err)
+		}
+
+		if err := d.Set("size", poolProperties.Size); err != nil {
+			return fmt.Errorf("Error setting `size`: %+v", err)
+		}
```
katbyte: Is there a reason you are checking the error for these sets? we typically don't check the error for primitive types

azurerm/resource_arm_netapp_pool.go
```diff
@@ -0,0 +1,220 @@
...
+
+			"service_level": {
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(netapp.Premium),
+					string(netapp.Standard),
+					string(netapp.Ultra),
+				}, false),
+				DiffSuppressFunc: suppress.CaseDifference,
```
katbyte: Why are we suppressing the case here? is the API returning the constant in the wrong casing? if so please open an issue on the SDK and put a link in a comment here explaining why

azurerm/resource_arm_netapp_pool.go
```diff
@@ -0,0 +1,220 @@
...
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	if poolProperties := resp.PoolProperties; poolProperties != nil {
+		if err := d.Set("service_level", poolProperties.ServiceLevel); err != nil {
+			return fmt.Errorf("Error setting `service_level`: %+v", err)
+		}
+
+		if err := d.Set("size", poolProperties.Size); err != nil {
+			return fmt.Errorf("Error setting `size`: %+v", err)
+		}
```
katbyte: same for datasource, why the err checks?

azurerm/resource_arm_netapp_pool_test.go
katbyte: the values in basic should be different in complete where applicable

website/docs/d/netapp_pool.html.markdown
```diff
@@ -0,0 +1,51 @@
...
+```
+
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the NetApp Pool.
+
+* `account_name` - (Required) The name of the NetApp account.
```
katbyte: Could we make this more datasource sounding:
```suggestion
* `account_name` - (Required) The name of the NetApp account where the NetApp pool exists.
```

website/docs/r/netapp_pool.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the NetApp Pool. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the resource group where the NetApp Pool should be created. Changing this forces a new resource to be created.
+
+* `account_name` - (Required) The name of the NetApp account.
```
katbyte:
```suggestion
* `account_name` - (Required) The name of the NetApp account in which the NetApp Pool should be created.
```

website/docs/r/netapp_pool.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `service_level` - (Required) The service level of the file system.
+
+* `size` - (Required) Provisioned size of the pool (in bytes). Allowed values are in 4TiB chunks (value must be multiply of 4398046511104).
```
katbyte: we should quote constants
```suggestion
* `size` - (Required) Provisioned size of the pool (in bytes). Allowed values are in 4TiB chunks (value must be multiply of `4398046511104`).
```

katbyte (changes requested):
Hi @neil-yechenwei,

Looking pretty good here, i've left some comments in addition to matthews tht should be addressed.

azurerm/resource_arm_netapp_pool.go
neil-yechenwei: fixed

azurerm/resource_arm_netapp_pool.go
neil-yechenwei: Yes. The future isn't accurate here. If I don't use "waitForNetAppPoolToBeDeleted" here, it would throw below error.
Error Message:
azurerm_resource_group.example: Refreshing state... [id=/subscriptions/xxxxx/resourceGroups/neiltest21-resources]
azurerm_netapp_account.example: Refreshing state... [id=/subscriptions/xxx/resourceGroups/neiltest21-resources/providers/Microsoft.NetApp/netAppAccounts/neiltest21-netappaccount]
azurerm_netapp_account.example: Destroying... [id=/subscriptions/xxxx/resourceGroups/neiltest21-resources/providers/Microsoft.NetApp/netAppAccounts/neiltest21-netappaccount]
Error: Error deleting NetApp Account "neiltest21-netappaccount" (Resource Group "neiltest21-resources"): netapp.AccountsClient#Delete: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="CannotDeleteResource" Message="Can not delete resource before nested resources are deleted."

azurerm/resource_arm_netapp_pool.go
neil-yechenwei: Sure. Updated.

azurerm/resource_arm_netapp_pool_test.go
neil-yechenwei: Fixed

azurerm/resource_arm_netapp_pool_test.go
neil-yechenwei: Fixed

website/docs/d/netapp_pool.html.markdown
neil-yechenwei: Fixed

website/docs/r/netapp_pool.html.markdown
neil-yechenwei: Fixed

azurerm/data_source_netapp_pool.go
neil-yechenwei: I think I cannot do this since netapp pool name is not allowed to start with underscore or hyphens but netapp account name can be allowed to start with them.

website/docs/r/netapp_pool.html.markdown
neil-yechenwei: It has been removed according by mbfrahry's comment.

website/docs/r/netapp_pool.html.markdown
neil-yechenwei: Fixed

website/docs/d/netapp_pool.html.markdown
neil-yechenwei: Fixed

azurerm/data_source_netapp_pool.go
neil-yechenwei: Moved this validation to netapp package as shared function.

azurerm/data_source_netapp_pool.go
neil-yechenwei: Fixed

azurerm/resource_arm_netapp_pool.go
neil-yechenwei: Fixed

azurerm/resource_arm_netapp_pool.go
neil-yechenwei: Fixed

neil-yechenwei:
@katbyte , @mbfrahry , Thanks for your comments. I've updated code per your comments.

azurerm/data_source_netapp_pool.go
```diff
@@ -0,0 +1,76 @@
...
+
+	d.Set("name", name)
+	d.Set("account_name", accountName)
+	d.Set("resource_group_name", resourceGroup)
+	if location := resp.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	if poolProperties := resp.PoolProperties; poolProperties != nil {
+		d.Set("service_level", poolProperties.ServiceLevel)
+		d.Set("size_in_4_tb", *poolProperties.Size/4398046511104)
```
mbfrahry: This could be nil so we should check that it's not before addressing it
```suggestion
if poolProperties.Size != nil {
		d.Set("size_in_4_tb", *poolProperties.Size/4398046511104)
}
```

azurerm/resource_arm_netapp_pool.go
```diff
@@ -0,0 +1,212 @@
...
+
+	d.Set("name", name)
+	d.Set("resource_group_name", resourceGroup)
+	d.Set("account_name", accountName)
+	if location := resp.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	if poolProperties := resp.PoolProperties; poolProperties != nil {
+		d.Set("service_level", poolProperties.ServiceLevel)
+		d.Set("size_in_4_tb", *poolProperties.Size/4398046511104)
```
mbfrahry: And we should check for nil here too

azurerm/resource_arm_netapp_pool.go
mbfrahry: Perfect! That makes sense

azurerm/resource_arm_netapp_pool.go
mbfrahry: Looks great!

website/docs/r/netapp_pool.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+
+* `resource_group_name` - (Required) The name of the resource group where the NetApp Pool should be created. Changing this forces a new resource to be created.
+
+* `account_name` - (Required) The name of the NetApp account in which the NetApp Pool should be created.
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `service_level` - (Required) The service level of the file system. Valid values include `Premium`, `Standard`, or `Ultra`.
+
+* `size_in_4_tb` - (Required) Provisioned size of the pool (in bytes). Allowed values are in 4TiB chunks (value must be multiply of `4398046511104`).
```
mbfrahry: I don't think we need the last bit.
```suggestion
* `size_in_4_tb` - (Required) Provisioned size of the pool (in bytes). Values are in 4TiB chunks.
```

mbfrahry (changes requested):
This is close! We just have a couple nil checks to add (so we don't panic) and then we'll do one last pass on tests and we're good.

azurerm/data_source_netapp_pool.go
neil-yechenwei: Fixed

azurerm/resource_arm_netapp_pool.go
neil-yechenwei: Fixed

website/docs/r/netapp_pool.html.markdown
neil-yechenwei: Fixed

neil-yechenwei:
@mbfrahry , Thanks for your comments. I've updated code.

mbfrahry (approved):
These changes look great! Thanks for opening this PR. We'll get it into the next release!

PR #4897, opened 2019-11-15
`azurerm_app_service` - Add support for `JAVA` container 

gpduck:


mbfrahry:
Thanks for opening this @gpduck, this change looks great and should make it into the next release

PR #4909, opened 2019-11-18
Add mention of skip_provider_registration in resource provider registration error

### LINKED ISSUE #4799: Improve Error message for failed Provider Register
### Community Note

### Description

If terraform is currently in a subscription without all resources providers regiserd, it tries to register these. This can fail if the user has not enough permissions. The error message is:

```
Error: Error ensuring Resource Providers are registered: Cannot register provider Microsoft.Kusto with Azure Resource Manager: resources.ProvidersClient#Register: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403
Code="AuthorizationFailed" Message="The client 'xxxx@yyy' with object id '<id>' does not have authorization to perform action 'Microsoft.Kusto/register/action' over scope '/subscriptions/<id>' or
the scope is invalid. If access was recently granted, please refresh your credentials.".

  on main.tf line 1, in provider "azurerm":
   1: provider "azurerm" {
```

This message is not very helpful. If you run in this kind of problem you have two options:
 - set `skip_provider_registration` to true
 - let someone with permissions enable all needed providers.

So the error should mention the `skip_provider_registration` option and also list all needed but not activated resource providers and only the first causing an error. So I can ask a person with higher rights to enable them all.

If someone new to terraform and only got access to a limited subscription, this may one of the first errors he encounters, so the message should be very helpful.

n/a

n/a

### References

* The check is done in https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/required_resource_providers.go

StefanSchoof:
Resolves #4799

### DIFF
--- a/azurerm/provider.go
+++ b/azurerm/provider.go
@@ -745,7 +745,7 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
 
 				err := ensureResourceProvidersAreRegistered(ctx, *client.Resource.ProvidersClient, availableResourceProviders, requiredResourceProviders)
 				if err != nil {
-					return nil, fmt.Errorf("Error ensuring Resource Providers are registered. If not other possible consider using the skip_provider_registration option https://www.terraform.io/docs/providers/azurerm/index.html#skip_provider_registration: %s", err)
+					return nil, fmt.Errorf("Error ensuring Resource Providers are registered. If the access method does not have permissions to register resource providers you may want to use the skip_provider_registration flag. Please see https://www.terraform.io/docs/providers/azurerm/index.html#skip_provider_registration for more details: %s", err)
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


azurerm/provider.go
```diff
@@ -745,7 +745,7 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
 
 				err := ensureResourceProvidersAreRegistered(ctx, *client.Resource.ProvidersClient, availableResourceProviders, requiredResourceProviders)
 				if err != nil {
-					return nil, fmt.Errorf("Error ensuring Resource Providers are registered: %s", err)
+					return nil, fmt.Errorf("Error ensuring Resource Providers are registered. If not other possible consider using the skip_provider_registration option https://www.terraform.io/docs/providers/azurerm/index.html#skip_provider_registration: %s", err)
```
katbyte: I think this might be worded better as:
```suggestion
					return nil, fmt.Errorf("Error ensuring Resource Providers are registered. If the access methold does not have permissions to register resource providers you may want to use the skip_provider_registration flag. Please see https://www.terraform.io/docs/providers/azurerm/index.html#skip_provider_registration for more details: %s", err)
```

WDYT?

StefanSchoof: Thanks, that sounds good. I changed it

PR #4917, opened 2019-11-19
App Service: support for php 7.3

ghostinthewires:
This adds support for PHP 7.3 to Azure App Service (including slots)

It also removes support for PHP 5.5 as this is no longer supported in Azure

Documentation also updated

### DIFF
--- a/azurerm/helpers/azure/app_service.go
+++ b/azurerm/helpers/azure/app_service.go
@@ -371,6 +371,7 @@ func SchemaAppServiceSiteConfig() *schema.Schema {
 					Type:     schema.TypeString,
 					Optional: true,
 					ValidateFunc: validation.StringInSlice([]string{
+						"5.5",
 						"5.6",
 						"7.0",
 						"7.1",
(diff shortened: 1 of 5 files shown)


ghostinthewires:
@katbyte @tombuildsstuff - Hi both, 

Are you able to provide an ETA for this being merged (all being good) 
Also which version of the AzureRM provider it may be included in (and subsequent release date of that)

Sorry for the upfront questions, just we need this functionality for a client pretty urgently. 👍

katbyte:
Hi @ghostinthewires,

I'm not sure we can remove support for `5.5` as it may affect other clouds that are not azure public. As such i am going to remove that change so we can get support for 7.3 into 1.37 🙂

ghostinthewires:
@katbyte That makes sense, thanks for that.

Do you have an idea of approx release date for 1.37?

katbyte:
Hopefully this week @ghostinthewires

ghostinthewires:
Awesome! Thanks for the quick merge @katbyte , much appreciated 👍

PR #4933, opened 2019-11-20
Add New Resource & Data Source: azurerm_netapp_volume

### LINKED ISSUE #3936: Implement Azure Netapp Files Support
### Description
Hi, 

With the general release of Azure Netapp Files, we'd like to see support in terraform for this.

Features and description: 
https://azure.microsoft.com/en-gb/blog/azure-netapp-files-is-now-generally-available/
https://cloud.netapp.com/azure-netapp-files

ARM template Documentation
https://docs.microsoft.com/en-us/azure/templates/microsoft.netapp/allversions

Something along the lines of the below could be useful for creating an account & a pool of storage:
```hcl
resource "azurerm_netapp_files" "account" {
  name                 = "${var.name}"
  location             = "${var.location}"
  subscription         = "${var.subscription}"
  resource_group_name  = "${var.resource_group_name}"
}

resource "azurerm_netapp_files" "capacity_pool" {
  name                 = "${var.pool_name}"
  size                 = ""
  properties           = ""
...
}
etc
```
& similar schema for creating volumes of different types (CIFS/NFS etc)

### Community Note

neil-yechenwei:
This PR is code implement for issue https://github.com/terraform-providers/terraform-provider-azurerm/issues/3936

Fixes #3936


azurerm/data_source_netapp_volume.go
```diff
@@ -0,0 +1,141 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"usage_threshold": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
+			"export_policy_rule": {
```
tombuildsstuff: I think we can remove this field - since I don't think it's likely to be used by users?

azurerm/data_source_netapp_volume.go
```diff
@@ -0,0 +1,141 @@
...
+		d.Set("subnet_id", props.SubnetID)
+
+		if props.UsageThreshold != nil {
+			d.Set("usage_threshold", *props.UsageThreshold/1073741824)
+		}
+		if props.ExportPolicy != nil {
+			if err := d.Set("export_policy_rule", flattenArmExportPolicyRule(props.ExportPolicy.Rules)); err != nil {
+				return fmt.Errorf("Error setting `export_policy_rule`: %+v", err)
+			}
+		}
```
tombuildsstuff: (as above) we can remove this

azurerm/internal/services/netapp/validate.go
```diff
@@ -24,3 +24,23 @@ func ValidateNetAppPoolName(v interface{}, k string) (warnings []string, errors
...
+
+func ValidateNetAppVolumeCreationToken(v interface{}, k string) (warnings []string, errors []error) {
+	value := v.(string)
+
+	if !regexp.MustCompile(`^[a-zA-Z][-\da-zA-Z]{0,79}$`).MatchString(value) {
+		errors = append(errors, fmt.Errorf("%q must be between 1 and 80 characters in length and start with letters and contains only letters, numbers or hyphens.", k))
+	}
+
+	return warnings, errors
+}
```
tombuildsstuff: can we add unit tests for this?

azurerm/internal/services/netapp/validate.go
```diff
@@ -24,3 +24,23 @@ func ValidateNetAppPoolName(v interface{}, k string) (warnings []string, errors
...
+
+func ValidateNetAppVolumeName(v interface{}, k string) (warnings []string, errors []error) {
+	value := v.(string)
+
+	if !regexp.MustCompile(`^[a-zA-Z][-_\da-zA-Z]{0,63}$`).MatchString(value) {
+		errors = append(errors, fmt.Errorf("%q must be between 1 and 64 characters in length and start with letters and contains only letters, numbers, underscore or hyphens.", k))
+	}
+
+	return warnings, errors
+}
```
tombuildsstuff: can we add unit tests for this?

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+			},
+
+			"service_level": {
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(netapp.Premium),
+					string(netapp.Standard),
+					string(netapp.Ultra),
+				}, true),
```
tombuildsstuff: this should be case insensitive:

```suggestion
				}, false),
```

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+			"pool_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: aznetapp.ValidateNetAppPoolName,
+			},
+
+			"creation_token": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: aznetapp.ValidateNetAppVolumeCreationToken,
```
tombuildsstuff: given this is a creation token presumably this should be ForceNew?

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+				ValidateFunc: aznetapp.ValidateNetAppVolumeName,
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"location": azure.SchemaLocation(),
+
+			"account_name": {
+				Type:         schema.TypeString,
+				Required:     true,
```
tombuildsstuff: this'll also be ForceNew

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+				ValidateFunc: aznetapp.ValidateNetAppAccountName,
+			},
+
+			"pool_name": {
+				Type:         schema.TypeString,
+				Required:     true,
```
tombuildsstuff: this'll also be ForceNew

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	if props := resp.VolumeProperties; props != nil {
+		d.Set("creation_token", props.CreationToken)
+		d.Set("service_level", props.ServiceLevel)
+		d.Set("subnet_id", props.SubnetID)
+
+		if props.UsageThreshold != nil {
+			d.Set("usage_threshold", *props.UsageThreshold/1073741824)
+		}
+		if props.ExportPolicy != nil {
```
tombuildsstuff: we can move this check inside the flatten function (~which can then return an empty array~ update: it does) to ensure this is always set

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+
+func waitForNetAppVolumeToBeDeleted(ctx context.Context, client *netapp.VolumesClient, resourceGroup, accountName, poolName, name string) error {
+	log.Printf("[DEBUG] Waiting for NetApp Volume Provisioning Service %q (Resource Group %q) to be deleted", name, resourceGroup)
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{"200", "202"},
+		Target:  []string{"404"},
+		Refresh: netappVolumeDeleteStateRefreshFunc(ctx, client, resourceGroup, accountName, poolName, name),
+		Timeout: 20 * time.Minute,
+	}
+	if _, err := stateConf.WaitForState(); err != nil {
```
tombuildsstuff: can we not use the Future to handle this?

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+
+		if _, err := client.Delete(ctx, resourceGroupName, accountName, poolName, name); err != nil {
+			log.Printf("Error reissuing NetApp Volume %q delete request (Resource Group %q): %+v", name, resourceGroupName, err)
+		}
+
+		return res, strconv.Itoa(res.StatusCode), nil
+	}
+}
+
+func expandArmExportPolicyRule(input []interface{}) *netapp.VolumePropertiesExportPolicy {
```
tombuildsstuff: this function name is scoped to the Package not the File - so can we rename this:

```suggestion
func expandArmNetAppVolumeExportPolicyRule(input []interface{}) *netapp.VolumePropertiesExportPolicy {
```

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+				UnixReadWrite:  utils.Bool(unixReadWrite),
+			}
+
+			results = append(results, result)
+		}
+	}
+
+	result := netapp.VolumePropertiesExportPolicy{
+		Rules: &results,
+	}
```
tombuildsstuff: we may as well return this directly here, rather than assigning it to a variable

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+		}
+		if v := item.Nfsv4; v != nil {
+			c["nfsv4"] = *v
+		}
+		if v := item.UnixReadOnly; v != nil {
+			c["unix_read_only"] = *v
+		}
+		if v := item.UnixReadWrite; v != nil {
+			c["unix_read_write"] = *v
+		}
```
tombuildsstuff: all of these values need to be set in all cases; as such can we update this to be:
]
```suggestion
		unixReadWrite := false
		if v := item.UnixReadWrite; v != nil {
			unixReadWrite = *v
		}
		
		...
		
		results = append(results, map[string]interface{}{
			"unix_read_write": unixReadWrite,
		})
```

azurerm/resource_arm_netapp_volume_test.go
```diff
@@ -0,0 +1,322 @@
+package azurerm
```
tombuildsstuff: can we add more tests here? covering changing the subnet, usage threshold, export policy rule etc

azurerm/resource_arm_netapp_volume_test.go
```diff
@@ -0,0 +1,322 @@
...
+}
+
+resource "azurerm_netapp_pool" "test" {
+  name                = "acctest-NetAppPool-%d"
+  location            = "${azurerm_resource_group.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  account_name        = "${azurerm_netapp_account.test.name}"
+  service_level       = "Premium"
+  size_in_4_tb        = "1"
+}
```
tombuildsstuff: rather than defining these the same for all of the tests we're using [template functions](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_windows_virtual_machine_scale_set_test.go#L67) to remove the commonality - can we update this?

examples/netapp/volume/main.tf
```diff
@@ -0,0 +1,63 @@
...
+  usage_threshold     = "100"
+
+  export_policy_rule {
+    rule_index      = "1"
+    allowed_clients = "1.2.3.0/24"
+    cifs            = "false"
+    nfsv3           = "true"
+    nfsv4           = "false"
+    unix_read_only  = "false"
+    unix_read_write = "true"
```
tombuildsstuff: these are booleans - can we remove the quotes?

examples/netapp/volume/main.tf
```diff
@@ -0,0 +1,63 @@
...
+  resource_group_name = "${azurerm_resource_group.example.name}"
+  account_name        = "${azurerm_netapp_account.example.name}"
+  pool_name           = "${azurerm_netapp_pool.example.name}"
+  creation_token      = "my-unique-file-path"
+  service_level       = "Premium"
+  subnet_id           = "${azurerm_subnet.example.id}"
+  usage_threshold     = "100"
+
+  export_policy_rule {
+    rule_index      = "1"
```
tombuildsstuff: this is an integer; can we remove the quotes here?

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+* `creation_token` - A unique file path for the volume. Used when creating mount targets.
+
+* `service_level` - The service level of the file system.
+
+* `subnet_id` - The Azure Resource URI for a delegated subnet. Must have the delegation Microsoft.NetApp/volumes.
+
+* `usage_threshold` - Maximum storage quota allowed for a file system in bytes. This is a soft quota used for alerting only.
+
+* `export_policy_rule` - One `export_policy_rule` block defined below.
```
tombuildsstuff:
```suggestion
* `export_policy_rule` - One or more `export_policy_rule` blocks as defined below.
```

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+* `pool_name` - (Required) The name of the NetApp pool where the NetApp volume exists.
+
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `location` - The Azure Region where the NetApp Volume exists.
+
+* `creation_token` - A unique file path for the volume. Used when creating mount targets.
```
tombuildsstuff:
```suggestion
* `creation_token` - A unique file path for the volume
```

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+* `name` - (Required) The name of the NetApp Volume.
+
+* `resource_group_name` - (Required) The Name of the Resource Group where the NetApp Volume exists.
+
+* `account_name` - (Required) The name of the NetApp account where the NetApp pool exists.
+
+* `pool_name` - (Required) The name of the NetApp pool where the NetApp volume exists.
+
+
```
tombuildsstuff:
```suggestion
```

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+  account_name        = "acctestnetappaccount"
+  pool_name           = "acctestnetapppool"
+  name                = "acctestnetappvolume"
+}
+
+output "netapp_volume_id" {
+  value = "${data.azurerm_netapp_volume.example.id}"
+}
+```
+
```
tombuildsstuff:
```suggestion
```

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+## NetApp Volume Usage
+
+```hcl
+data "azurerm_netapp_volume" "example" {
+  resource_group_name = "acctestRG"
+  account_name        = "acctestnetappaccount"
+  pool_name           = "acctestnetapppool"
+  name                = "acctestnetappvolume"
```
tombuildsstuff: can we use more readable naming here e.g.

```suggestion
  name                = "example-volume"
```

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+The following attributes are exported:
+
+* `location` - The Azure Region where the NetApp Volume exists.
+
+* `creation_token` - A unique file path for the volume. Used when creating mount targets.
+
+* `service_level` - The service level of the file system.
+
+* `subnet_id` - The Azure Resource URI for a delegated subnet. Must have the delegation Microsoft.NetApp/volumes.
```
tombuildsstuff: This is a Data Source so this'll be readonly

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+* `usage_threshold` - Maximum storage quota allowed for a file system in bytes. This is a soft quota used for alerting only.
```
tombuildsstuff: same here

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+* `export_policy_rule` - One `export_policy_rule` block defined below.
+
+---
+
+The `export_policy_rule` block contains the following:
```
tombuildsstuff:
```suggestion
A `export_policy_rule` block exports the following:
```

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+* `nfsv3` - Allows NFSv3 protocol.
+
+* `nfsv4` - Allows NFSv4 protocol.
+
+* `unix_read_only` - Read only access.
+
+* `unix_read_write` - Read and write access.
+
+---
```
tombuildsstuff:
```suggestion
```

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+* `export_policy_rule` - One `export_policy_rule` block defined below.
+
+---
+
+The `export_policy_rule` block contains the following:
+
+* `rule_index` - Order index.
+
+* `allowed_clients` - Client ingress specification as comma separated string with IPv4 CIDRs, IPv4 host addresses and host names.
```
tombuildsstuff: in which case this field wants to become a List/Set of strings - since exposing a comma-separated string isn't a great user experience

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+* `cifs` - Allows CIFS protocol.
+
+* `nfsv3` - Allows NFSv3 protocol.
+
+* `nfsv4` - Allows NFSv4 protocol.
+
+* `unix_read_only` - Read only access.
+
+* `unix_read_write` - Read and write access.
```
tombuildsstuff: what specifically does this mean? I'm assuming this:

```suggestion
* `unix_read_write` - Are the Read and Write permissions enabled on Unix Systems?
```

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+sidebar_current: "docs-azurerm-resource-netapp-volume"
+description: |-
+  Manages a NetApp Volume.
+---
+
+# azurerm_netapp_volume
+
+Manages a NetApp Volume.
+
+
```
tombuildsstuff:
```suggestion
```

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+resource "azurerm_netapp_volume" "example" {
+  name                = "example-netappvolume"
+  location            = "${azurerm_resource_group.example.location}"
+  resource_group_name = "${azurerm_resource_group.example.name}"
+  account_name        = "${azurerm_netapp_account.example.name}"
+  pool_name           = "${azurerm_netapp_pool.example.name}"
+  creation_token      = "my-unique-file-path"
+  service_level       = "Premium"
+  subnet_id           = "${azurerm_subnet.example.id}"
+  usage_threshold     = "100"
```
tombuildsstuff: this is an integer - so can we remove the quotes here?

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+  resource_group_name = "${azurerm_resource_group.example.name}"
+}
+
+resource "azurerm_netapp_pool" "example" {
+  name                = "example-netapppool"
+  location            = "${azurerm_resource_group.example.location}"
+  resource_group_name = "${azurerm_resource_group.example.name}"
+  account_name        = "${azurerm_netapp_account.example.name}"
+  service_level       = "Premium"
+  size_in_4_tb        = "1"
```
tombuildsstuff: this is an integer - so can we remove the quotes here?

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+  location            = "${azurerm_resource_group.example.location}"
+  resource_group_name = "${azurerm_resource_group.example.name}"
+  address_space       = ["10.0.0.0/16"]
+}
+
+resource "azurerm_subnet" "example" {
+  name                 = "example-subnet"
+  resource_group_name  = "${azurerm_resource_group.example.name}"
+  virtual_network_name = "${azurerm_virtual_network.example.name}"
+  address_prefix       = "10.0.2.0/24"
```
tombuildsstuff: this'll make this more readable:

```suggestion
  address_prefix       = "10.0.2.0/24"

```

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `creation_token` - (Required) A unique file path for the volume. Used when creating mount targets.
+
+* `service_level` - (Required) The service level of the file system. Valid values include `Premium`, `Standard`, or `Ultra`.
+
+* `subnet_id` - (Required) The Azure Resource URI for a delegated subnet. Must have the delegation Microsoft.NetApp/volumes.
+
+* `usage_threshold` - (Required) Maximum storage quota allowed for a file system in bytes. This is a soft quota used for alerting only.
+
+* `export_policy_rule` - (Optional) One `export_policy_rule` block defined below.
```
tombuildsstuff: There can be up to 5 of these - as such:

```suggestion
* `export_policy_rule` - (Optional) One or more `export_policy_rule` block defined below.
```

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `pool_name` - (Required) The name of the NetApp pool in which the NetApp Volume should be created.
+
+* `creation_token` - (Required) A unique file path for the volume. Used when creating mount targets.
+
+* `service_level` - (Required) The service level of the file system. Valid values include `Premium`, `Standard`, or `Ultra`.
+
+* `subnet_id` - (Required) The Azure Resource URI for a delegated subnet. Must have the delegation Microsoft.NetApp/volumes.
+
+* `usage_threshold` - (Required) Maximum storage quota allowed for a file system in bytes. This is a soft quota used for alerting only.
```
tombuildsstuff: Perhaps this field wants to be renamed (since it's being converted to GB's above)?

```suggestion
* `usage_threshold` - (Required) The maximum Storage Quota in Gigabytes allowed for a file system.

-> **NOTE:** This is a soft quota used for alerting only.
```

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `account_name` - (Required) The name of the NetApp account in which the NetApp Pool should be created.
+
+* `pool_name` - (Required) The name of the NetApp pool in which the NetApp Volume should be created.
+
+* `creation_token` - (Required) A unique file path for the volume. Used when creating mount targets.
+
+* `service_level` - (Required) The service level of the file system. Valid values include `Premium`, `Standard`, or `Ultra`.
```
tombuildsstuff: can we sort these alphabetically?

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `name` - (Required) The name of the NetApp Volume. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the resource group where the NetApp Volume should be created. Changing this forces a new resource to be created.
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `account_name` - (Required) The name of the NetApp account in which the NetApp Pool should be created.
+
+* `pool_name` - (Required) The name of the NetApp pool in which the NetApp Volume should be created.
```
tombuildsstuff: since this is force new we need to include the "Changing this forces a new resource to be created" text here

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the NetApp Volume. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the resource group where the NetApp Volume should be created. Changing this forces a new resource to be created.
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `account_name` - (Required) The name of the NetApp account in which the NetApp Pool should be created.
```
tombuildsstuff: since this is force new we need to include the "Changing this forces a new resource to be created" text here

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `pool_name` - (Required) The name of the NetApp pool in which the NetApp Volume should be created.
+
+* `creation_token` - (Required) A unique file path for the volume. Used when creating mount targets.
+
+* `service_level` - (Required) The service level of the file system. Valid values include `Premium`, `Standard`, or `Ultra`.
+
+* `subnet_id` - (Required) The Azure Resource URI for a delegated subnet. Must have the delegation Microsoft.NetApp/volumes.
```
tombuildsstuff: we intentionally don't mention "Azure Resource URI" since it's unhelpful - can we update this to be:

```suggestion
* `subnet_id` - (Required) The ID of a Subnet in which the ..., which must have the delegation Microsoft.NetApp/volumes.
```

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `usage_threshold` - (Required) Maximum storage quota allowed for a file system in bytes. This is a soft quota used for alerting only.
+
+* `export_policy_rule` - (Optional) One `export_policy_rule` block defined below.
+
+---
+
+The `export_policy_rule` block contains the following:
```
tombuildsstuff:
```suggestion
An `export_policy_rule` block supports the following:
```

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `rule_index` - (Required) Order index.
```
tombuildsstuff: what does this mean? could we be more descriptive here

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `cifs` - (Required) Allows CIFS protocol.
+
+* `nfsv3` - (Required) Allows NFSv3 protocol.
+
+* `nfsv4` - (Required) Allows NFSv4 protocol.
+
+* `unix_read_only` - (Required) Read only access.
+
+* `unix_read_write` - (Required) Read and write access.
```
tombuildsstuff: what does this mean? could we be more descriptive here

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `allowed_clients` - (Required) Client ingress specification as comma separated string with IPv4 CIDRs, IPv4 host addresses and host names.
+
+* `cifs` - (Required) Allows CIFS protocol.
+
+* `nfsv3` - (Required) Allows NFSv3 protocol.
+
+* `nfsv4` - (Required) Allows NFSv4 protocol.
+
+* `unix_read_only` - (Required) Read only access.
```
tombuildsstuff: what does this mean? could we be more descriptive here

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `export_policy_rule` - (Optional) One `export_policy_rule` block defined below.
+
+---
+
+The `export_policy_rule` block contains the following:
+
+* `rule_index` - (Required) Order index.
+
+* `allowed_clients` - (Required) Client ingress specification as comma separated string with IPv4 CIDRs, IPv4 host addresses and host names.
```
tombuildsstuff: (as above) I think this wants to become an List/Set of strings rather than being a comma separated string; which isn't a great UX here

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The ID of the NetApp Volume.
+
+## Import
+
+NetApp Volume can be imported using the `resource id`, e.g.
```
tombuildsstuff:
```suggestion
NetApp Volumes can be imported using the `resource id`, e.g.
```

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+			},
+
+			"usage_threshold": {
+				Type:         schema.TypeInt,
+				Required:     true,
+				ValidateFunc: validation.IntBetween(100, 4096),
+			},
+
+			"export_policy_rule": {
+				Type:     schema.TypeList,
```
tombuildsstuff: since the rule_index is part of this presumably these could be returned in any order and `export_policy_rule` needs to become a Set?

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+				ValidateFunc: validation.StringInSlice([]string{
+					string(netapp.Premium),
+					string(netapp.Standard),
+					string(netapp.Ultra),
+				}, true),
+			},
+
+			"subnet_id": {
+				Type:         schema.TypeString,
+				Required:     true,
```
tombuildsstuff: presumably this is ForceNew? can we add a test to cover this

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,357 @@
...
+				ValidateFunc: azure.ValidateResourceID,
+			},
+
+			"usage_threshold": {
+				Type:         schema.TypeInt,
+				Required:     true,
```
tombuildsstuff: presumably this is ForceNew? can we add a test to cover this

tombuildsstuff:
(sorry for the double-review, Github returned a 500 🙃)

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: Fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: Fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: FIxed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: Yes

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: FIxed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: Fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

azurerm/data_source_netapp_volume.go
neil-yechenwei: I don't think so. Now both powershell and azure portal for netapp volume have opened the entry to let user edit export policy. And we cannot ensure the users won't use this. If they want to use, I submit another PR for this? Why don't we leave it as optional field? And this field should be optional. See the doc https://docs.microsoft.com/en-us/azure/azure-netapp-files/azure-netapp-files-configure-export-policy.

azurerm/data_source_netapp_volume.go
neil-yechenwei: (as above) we cannot remove this

azurerm/internal/services/netapp/validate.go
neil-yechenwei: fixed

azurerm/internal/services/netapp/validate.go
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: Probably not. Because there is an issue here https://github.com/Azure/azure-sdk-for-go/issues/6374

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: I don't think so. We can left it empty.
image

examples/netapp/volume/main.tf
neil-yechenwei: fixed

examples/netapp/volume/main.tf
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: It shouldn't be ForceNew.
image

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: Read only file system on unix.

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: Read write file system on unix.

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: Specify the index number for the rule.
An export policy can consist of up to five rules. Rules are evaluated according to their order in the list of index numbers. Rules with lower index numbers are evaluated first. For example, the rule with index number 1 is evaluated before the rule with index number 2.

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: I found *_rule also is set as TypeList in existing resources of terraform. Why i cannot use TypeList for this field like others? For more details about rule_index and export_policy, please see this doc https://docs.microsoft.com/en-us/azure/azure-netapp-files/azure-netapp-files-configure-export-policy

azurerm/resource_arm_netapp_volume_test.go
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume_test.go
neil-yechenwei: fixed

neil-yechenwei:
@tombuildsstuff , Thanks for your comments. I've updated code per your comments.

azurerm/resource_arm_netapp_pool_test.go
```diff
@@ -200,7 +200,7 @@ resource "azurerm_netapp_pool" "test" {
   location            = "${azurerm_resource_group.test.location}"
   resource_group_name = "${azurerm_resource_group.test.name}"
   service_level       = "Premium"
-  size_in_4_tb        = "1"
+  size_in_4_tb        = 1
```
katbyte: This has been renamed on master

azurerm/resource_arm_netapp_volume.go
katbyte: added to test, resolved

azurerm/resource_arm_netapp_volume.go
katbyte: Most of those should probably be `TypeSets` too, as order doesn't matter if `rule_index` is used, otherwise if we keep it as a typelist the ordering could be implicitly inferred by the index. WDYT @tombuildsstuff ?

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,361 @@
...
+	if err != nil {
+		return err
+	}
+	resourceGroup := id.ResourceGroup
+	accountName := id.Path["netAppAccounts"]
+	poolName := id.Path["capacityPools"]
+	name := id.Path["volumes"]
+
+	_, err = client.Delete(ctx, resourceGroup, accountName, poolName, name)
+	if err != nil {
```
katbyte: These two lines could be merged

azurerm/resource_arm_netapp_volume.go
katbyte: That is for the pool, have you checked if volume exhibits the same behaviour?

azurerm/resource_arm_netapp_volume_test.go
```diff
@@ -0,0 +1,465 @@
...
+  name                = "acctest-NetAppVolume-%d"
+  location            = "${azurerm_resource_group.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  account_name        = "${azurerm_netapp_account.test.name}"
+  pool_name           = "${azurerm_netapp_pool.test.name}"
+  service_level       = "Premium"
+  creation_token      = "my-unique-file-path-%d"
+  subnet_id           = "${azurerm_subnet.test.id}"
+  usage_threshold     = 101
+  export_policy_rule {
```
katbyte: Could we have multiple rules here?

azurerm/resource_arm_netapp_volume_test.go
```diff
@@ -0,0 +1,465 @@
...
+					resource.TestCheckResourceAttr(resourceName, "usage_threshold", "101"),
+					resource.TestCheckResourceAttr(resourceName, "export_policy_rule.#", "1"),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
```
katbyte: Could we have another step going back to basic?

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+
+* `cifs` - Allows CIFS protocol.
+
+* `nfsv3` - Allows NFSv3 protocol.
+
+* `nfsv4` - Allows NFSv4 protocol.
+
+* `unix_read_only` - Read only file system type on unix.
+
+* `unix_read_write` - Read and write file system type on unix.
```
katbyte: Could we change the wording to sound like a data source?
```
An `export_policy_rule` block exports the following:

* `nfsv4` - Is the NFSv4 protocol allowed.
```

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `cifs` - (Required) Allows CIFS protocol.
+
+* `nfsv3` - (Required) Allows NFSv3 protocol.
+
+* `nfsv4` - (Required) Allows NFSv4 protocol.
+
+* `unix_read_only` - (Required) Read only file system type on unix.
+
+* `unix_read_write` - (Required) Read and write file system type on unix.
```
katbyte: could we improve the wording here:
```
* `rule_index` - (Required) The index number of the rule.


* `cifs` - (Required) Is the CIFS protocol allowed?

* `unix_read_only` - (Required) Is the file system on unix read only?
```

and so forth

azurerm/resource_arm_netapp_pool_test.go
neil-yechenwei: thanks for reminder. I've updated.

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: Yes. I already double checked before against this issue. both volume and pool have the same issue. So I also mentioned the volume also exists the same issue in this issue link https://github.com/Azure/azure-sdk-for-go/issues/6374

image

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume_test.go
neil-yechenwei: fxied

azurerm/resource_arm_netapp_volume_test.go
neil-yechenwei: fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

neil-yechenwei:
@katbyte , Thanks for your comments. I've updated code per your comments.

azurerm/data_source_netapp_volume.go
tombuildsstuff: whilst users may configure this on the Resource, the Data Source is Read Only and I can't envisage a scenario where users are likely to use this in the Data Source (since this needs to be a Set too) - for now let's remove it and we can add it back in if users request it

azurerm/data_source_netapp_volume.go
tombuildsstuff: as above we should be fine to remove this since users aren't likely to access this in a Data Source imo

azurerm/resource_arm_netapp_volume.go
tombuildsstuff: this is a Terraform specific thing where all inner fields need to be specified

neil-yechenwei: so you mean I don’t need to change the code here as you suggested, right?

neil-yechenwei: If I use your suggested code, it would add at least one item. Then the behavior is different with the resource in azure portal because the behavior of this resource in azure portal allows not to add any item. So we need to keep align with azure portal, then I don’t need to change the code as you suggested, right?

azurerm/data_source_netapp_volume.go
neil-yechenwei: removed

azurerm/data_source_netapp_volume.go
neil-yechenwei: removed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

azurerm/resource_arm_netapp_pool.go
```diff
@@ -44,6 +44,7 @@ func resourceArmNetAppPool() *schema.Resource {
 			"account_name": {
 				Type:         schema.TypeString,
 				Required:     true,
+				ForceNew:     true,
```
katbyte: This is a breaking change? as well as the documentation should also be updated.

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,372 @@
...
+				MaxItems: 5,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"rule_index": {
+							Type:         schema.TypeInt,
+							Required:     true,
+							ValidateFunc: validation.IntBetween(1, 5),
+						},
+						"allowed_clients": {
+							Type:     schema.TypeList,
```
katbyte: This should probably be a set too>

azurerm/resource_arm_netapp_volume.go
katbyte: Could we link back to that issue from here/create a new one update that to reflect it also affects this call?

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+
+* `pool_name` - (Required) The name of the NetApp pool where the NetApp volume exists.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `location` - The Azure Region where the NetApp Volume exists.
+
+* `creation_token` - A unique file path for the volume.
```
katbyte: Could we use data source language here

website/docs/d/netapp_volume.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+
+* `service_level` - The service level of the file system.
+
+* `subnet_id` - The ID of a Subnet in which the NetApp Volume, which must have the delegation Microsoft.NetApp/volumes.
```
katbyte: grammer:
```suggestion
* `subnet_id` - The ID of a Subnet in which the NetApp Volume resides.
```

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `export_policy_rule` - (Optional) One or more `export_policy_rule` block defined below.
+
+---
+
+An `export_policy_rule` block supports the following:
+
+* `rule_index` - (Required) The index number of the rule.
+
+* `allowed_clients` - (Required) Client ingress specification as list with IPv4 CIDRs, IPv4 host addresses.
```
katbyte: Could we fix the grammar here?

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `account_name` - (Required) The name of the NetApp account in which the NetApp Pool should be created. Changing this forces a new resource to be created.
+
+* `creation_token` - (Required) A unique file path for the volume. Used when creating mount targets. Changing this forces a new resource to be created.
+
+* `pool_name` - (Required) The name of the NetApp pool in which the NetApp Volume should be created. Changing this forces a new resource to be created.
+
+* `service_level` - (Required) The service level of the file system. Valid values include `Premium`, `Standard`, or `Ultra`.
+
+* `subnet_id` - (Required) The ID of a Subnet in which the NetApp Volume, which must have the delegation Microsoft.NetApp/volumes. Changing this forces a new resource to be created.
```
katbyte: Could we fix the grammar here?

website/docs/r/netapp_volume.html.markdown
katbyte: Could we elaborate on what the service level means?

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `usage_threshold` - (Required) The maximum Storage Quota in Gigabytes allowed for a file system.
```
katbyte:
```suggestion
* `usage_threshold` - (Required) The maximum Storage Quota allowed for a file system in Gigabytes.
```

azurerm/resource_arm_netapp_pool.go
neil-yechenwei: NetApp Volume and NetApp Snapshot I haven' merged to master branch is associated with NetApp pool. So I think it has no meaning for only using NetApp Pool if NetApp Volume and NetApp Snapshot isn't ready. And it won't change the final result for this resource. So I think this isn't a breaking change. I've updated the documentation. Thanks.

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: Fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/d/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

neil-yechenwei:
Thanks for the comments @katbyte , I've updated the code per your comments.

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,378 @@
...
+						},
+						"allowed_clients": {
+							Type:     schema.TypeSet,
+							Required: true,
+							Elem: &schema.Schema{
+								Type:         schema.TypeString,
+								ValidateFunc: aznetapp.ValidateNetAppVolumeAllowedClients,
+							},
+						},
+						"cifs": {
```
katbyte: Would it make sense for these properties to be named `cifs_enabled` ?

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,378 @@
...
+						},
+						"nfsv4": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+						"unix_read_only": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+						"unix_read_write": {
```
katbyte: should we make sure only one of these is set? or can they both be set?

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+	if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("Error waiting for creation of NetApp Volume %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	resp, err := client.Get(ctx, resourceGroup, accountName, poolName, name)
+	if err != nil {
+		return fmt.Errorf("Error retrieving NetApp Volume %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+	if resp.ID == nil {
```
katbyte: could we also check if the ID is empty?

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,378 @@
...
+	if location := resp.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	if props := resp.VolumeProperties; props != nil {
+		d.Set("creation_token", props.CreationToken)
+		d.Set("service_level", props.ServiceLevel)
+		d.Set("subnet_id", props.SubnetID)
+
+		if props.UsageThreshold != nil {
+			d.Set("usage_threshold", *props.UsageThreshold/1073741824)
```
katbyte: what units is `usage_threshold` in? could we make the name `usage_threshold_in_units`

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `export_policy_rule` - (Optional) One or more `export_policy_rule` block defined below.
+
+---
+
+An `export_policy_rule` block supports the following:
+
+* `rule_index` - (Required) The index number of the rule.
+
+* `allowed_clients` - (Required) The allowed clients as set with IPv4 CIDRs, IPv4 host addresses.
```
katbyte: Could we fix the grammar here?
```suggestion
* `allowed_clients` - (Required) A list of allowed clients IPv4 addresses.
```

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `creation_token` - (Required) A unique file path for the volume. Used when creating mount targets. Changing this forces a new resource to be created.
+
+* `pool_name` - (Required) The name of the NetApp pool in which the NetApp Volume should be created. Changing this forces a new resource to be created.
+
+* `service_level` - (Required) The target performance of the file system. Valid values include `Premium`, `Standard`, or `Ultra`.
+
+* `subnet_id` - (Required) The ID of the Subnet within the NetApp Volume, which must have the delegation Microsoft.NetApp/volumes. Changing this forces a new resource to be created.
+
+* `usage_threshold` - (Required) The maximum Storage Quota allowed for a file system in Gigabytes.
```
katbyte: the name here seems different then the description?
Should this property perhaps be called `storage_quota_in_gb`?

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `name` - (Required) The name of the NetApp Volume. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the resource group where the NetApp Volume should be created. Changing this forces a new resource to be created.
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `account_name` - (Required) The name of the NetApp account in which the NetApp Pool should be created. Changing this forces a new resource to be created.
+
+* `creation_token` - (Required) A unique file path for the volume. Used when creating mount targets. Changing this forces a new resource to be created.
```
katbyte: Should this perhaps be called something different? creation token seems odd if its more the "volume path"?

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `pool_name` - (Required) The name of the NetApp pool in which the NetApp Volume should be created. Changing this forces a new resource to be created.
+
+* `service_level` - (Required) The target performance of the file system. Valid values include `Premium`, `Standard`, or `Ultra`.
+
+* `subnet_id` - (Required) The ID of the Subnet within the NetApp Volume, which must have the delegation Microsoft.NetApp/volumes. Changing this forces a new resource to be created.
```
katbyte: Is this description correct? should it be

```suggestion
* `subnet_id` - (Required) The ID of the Subnet the NetApp Volume resides in, which must have the `Microsoft.NetApp/volumes` delegation. Changing this forces a new resource to be created.
```

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+
+* `usage_threshold` - (Required) The maximum Storage Quota allowed for a file system in Gigabytes.
+
+* `export_policy_rule` - (Optional) One or more `export_policy_rule` block defined below.
+
+---
+
+An `export_policy_rule` block supports the following:
+
+* `rule_index` - (Required) The index number of the rule.
```
katbyte: Are rules evaluated in order? does order matter here? if no would it make sense to remove the index?

azurerm/data_source_netapp_volume.go
```diff
@@ -0,0 +1,99 @@
...
+
+	resp, err := client.Get(ctx, resourceGroup, accountName, poolName, name)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return fmt.Errorf("Error: NetApp Volume %q (Resource Group %q) was not found", name, resourceGroup)
+		}
+		return fmt.Errorf("Error reading NetApp Volume %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	d.SetId(*resp.ID)
```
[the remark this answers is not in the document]
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: It cannot be removed because rules are evaluated according to their order in the list of index numbers so that we need to allow user to specify the rule index to customize the sequence of the rule.
See more details from https://docs.microsoft.com/en-us/azure/azure-netapp-files/azure-netapp-files-configure-export-policy.

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: I am confused. You want this field name is updated to "usage_threshold_in_units" or "storage_quota_in_gb"? I think "storage_quota_in_gb" is preferred so that I updated this field name to "storage_quota_in_gb".

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: just need to set only one of these. So update these two fields as optional.

neil-yechenwei:
Thank you for the comments @katbyte , I've updated code per your comments.

website/docs/r/netapp_volume.html.markdown
katbyte: 👍

azurerm/data_source_netapp_volume.go
```diff
@@ -0,0 +1,110 @@
...
+	resp, err := client.Get(ctx, resourceGroup, accountName, poolName, name)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return fmt.Errorf("Error: NetApp Volume %q (Resource Group %q) was not found", name, resourceGroup)
+		}
+		return fmt.Errorf("Error reading NetApp Volume %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	if resp.ID == nil && *resp.ID != "" {
+		return fmt.Errorf("Error retrieving NetApp Volume %q (Resource Group %q): ID was nil", name, resourceGroup)
```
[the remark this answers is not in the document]
neil-yechenwei: fixed

azurerm/internal/services/netapp/validation_test.go
```diff
@@ -0,0 +1,182 @@
...
+			expected: false,
+		},
+	}
+
+	for _, v := range testData {
+		t.Logf("[DEBUG] Testing %q..", v.input)
+
+		_, errors := ValidateNetAppVolumeAllowedClients(v.input, "allowed_clients")
+		actual := len(errors) == 0
+		if v.expected != actual {
```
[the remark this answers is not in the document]
neil-yechenwei: I think I cannot merge the code as you suggested since the variable "actual" in line 179 is referenced.

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,389 @@
...
+							Type:         schema.TypeInt,
+							Required:     true,
+							ValidateFunc: validation.IntBetween(1, 5),
+						},
+						"allowed_clients": {
+							Type:     schema.TypeSet,
+							Required: true,
+							Elem: &schema.Schema{
+								Type:         schema.TypeString,
+								ValidateFunc: aznetapp.ValidateNetAppVolumeAllowedClients,
```
katbyte: I think we have an existing IP address validation function?

azurerm/resource_arm_netapp_volume.go
```diff
@@ -0,0 +1,389 @@
...
+						},
+						"cifs_enabled": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+						"nfsv3": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+						"nfsv4": {
```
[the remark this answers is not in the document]
katbyte: this still needs to be addressed

website/docs/r/netapp_volume.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+}
+
+resource "azurerm_subnet" "example" {
+  name                 = "example-subnet"
+  resource_group_name  = "${azurerm_resource_group.example.name}"
+  virtual_network_name = "${azurerm_virtual_network.example.name}"
+  address_prefix       = "10.0.2.0/24"
+
+  delegation {
+    name = "testdelegation"
```
katbyte:
```suggestion
    name = "netapp"
```

azurerm/resource_arm_netapp_volume.go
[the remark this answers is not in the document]
neil-yechenwei: fixed

website/docs/r/netapp_volume.html.markdown
neil-yechenwei: fixed

azurerm/resource_arm_netapp_volume.go
neil-yechenwei: fixed

neil-yechenwei:
@katbyte , I already used the command you provided before for testing. Below is my test result whose test cases are all passed. I am using the test command `make testacc TEST=./azurerm TESTARGS="-parallel 11 -test.run=TestAccAzureRMNetAppVolume"`. If it's different with yours, could you share your test command. So I think maybe it's just your test environment issue. Could you double check whether you deleted all NetApp related resources in Azure portal before testing? After you deleted all NetApp related resource, please rerun the test cases.

v-cheye@NeilDevBox12 MINGW64 /c/terraform/projects/terraform-provider-azurerm (patch_netapp_volume)
$ make testacc TEST=./azurerm TESTARGS="-parallel 11 -test.run=TestAccAzureRMNetAppVolume"
TF_ACC=1 go test ./azurerm -v -parallel 11 -test.run=TestAccAzureRMNetAppVolume -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMNetAppVolume_basic
=== PAUSE TestAccAzureRMNetAppVolume_basic
=== RUN   TestAccAzureRMNetAppVolume_requiresImport
--- SKIP: TestAccAzureRMNetAppVolume_requiresImport (0.00s)
    resource_arm_netapp_volume_test.go:42: Skipping since resources aren't required to be imported
=== RUN   TestAccAzureRMNetAppVolume_complete
=== PAUSE TestAccAzureRMNetAppVolume_complete
=== RUN   TestAccAzureRMNetAppVolume_update
=== PAUSE TestAccAzureRMNetAppVolume_update
=== RUN   TestAccAzureRMNetAppVolume_updateSubnet
=== PAUSE TestAccAzureRMNetAppVolume_updateSubnet
=== RUN   TestAccAzureRMNetAppVolume_updateExportPolicyRule
=== PAUSE TestAccAzureRMNetAppVolume_updateExportPolicyRule
=== CONT  TestAccAzureRMNetAppVolume_basic
=== CONT  TestAccAzureRMNetAppVolume_updateSubnet
=== CONT  TestAccAzureRMNetAppVolume_update
=== CONT  TestAccAzureRMNetAppVolume_complete
=== CONT  TestAccAzureRMNetAppVolume_updateExportPolicyRule
--- PASS: TestAccAzureRMNetAppVolume_complete (571.17s)
--- PASS: TestAccAzureRMNetAppVolume_update (774.01s)
--- PASS: TestAccAzureRMNetAppVolume_basic (968.05s)
--- PASS: TestAccAzureRMNetAppVolume_updateExportPolicyRule (1141.21s)
--- PASS: TestAccAzureRMNetAppVolume_updateSubnet (1546.39s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm       1561.007s

PR #4941, opened 2019-11-21
r/resource_group: fixing a bug in the deletion assertion

tombuildsstuff:
Noticed this whilst passing through during a talk yesterday


tombuildsstuff:
Tests pass:

Screenshot 2019-11-21 at 12 31 52

PR #4951, opened 2019-11-21
`azurerm_netapp_pool`: `size_in_4_tb` -> `size_in_tb`

mbfrahry:
This PR moves `size_in_4_tb` to just `size_in_tb`. It also validates that size can be passed in as any number between `4` and `500` and not just multiples of `4`

### DIFF
--- a/azurerm/resource_arm_netapp_pool_test.go
+++ b/azurerm/resource_arm_netapp_pool_test.go
@@ -79,7 +79,7 @@ func TestAccAzureRMNetAppPool_complete(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testCheckAzureRMNetAppPoolExists(resourceName),
 					resource.TestCheckResourceAttr(resourceName, "service_level", "Standard"),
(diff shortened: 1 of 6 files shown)


azurerm/resource_arm_netapp_pool_test.go
```diff
@@ -236,7 +236,7 @@ resource "azurerm_netapp_pool" "test" {
   location            = "${azurerm_resource_group.test.location}"
   resource_group_name = "${azurerm_resource_group.test.name}"
   service_level       = "Standard"
-  size_in_4_tb        = "2"
+  size                = "8"
```
katbyte: Should we have a test with a non multiple of 4?

mbfrahry: Swapped to `15`
```
--- PASS: TestAccAzureRMNetAppPool_basic (101.94s)
--- PASS: TestAccAzureRMNetAppPool_complete (106.63s)
--- PASS: TestAccAzureRMNetAppPool_update (123.98s)
```

PR #4955, opened 2019-11-22
comman marshal supports bool/int/float slice

magodo:
Currently *common_marshal.go* only supports expand/flatten utilities for
string slice, however, there are 3 more primitive types in terraform:
bool, int, float.

I find it would be helpful if there is direct support in utility so that user could get rid of hassel. For example, in *resource_arm_monitor_autoscale_setting.go*, exists following function:

```go
func expandAzureRmMonitorAutoScaleSettingRecurrence(input []interface{}) *insights.Recurrence {
   if len(input) == 0 {
      return nil
   }

   recurrenceRaw := input[0].(map[string]interface{})

   timeZone := recurrenceRaw["timezone"].(string)
   days := make([]string, 0)
   for _, dayItem := range recurrenceRaw["days"].([]interface{}) {
      days = append(days, dayItem.(string))
   }

   hours := make([]int32, 0)
   for _, hourItem := range recurrenceRaw["hours"].([]interface{}) {
      hours = append(hours, int32(hourItem.(int)))
   }

   minutes := make([]int32, 0)
   for _, minuteItem := range recurrenceRaw["minutes"].([]interface{}) {
      minutes = append(minutes, int32(minuteItem.(int)))
   }

   return &insights.Recurrence{
      // API docs say this has to be `Week`.
      Frequency: insights.RecurrenceFrequencyWeek,
      Schedule: &insights.RecurrentSchedule{
         TimeZone: utils.String(timeZone),
         Days:     &days,
         Hours:    &hours,
         Minutes:  &minutes,
      },
   }
}
```

By leveraging utility functions, it could be simplified as: 

```go
func expandAzureRmMonitorAutoScaleSettingRecurrence(input []interface{}) *insights.Recurrence {
   if len(input) == 0 {
      return nil
   }

   recurrenceRaw := input[0].(map[string]interface{})

   return &insights.Recurrence{
      // API docs say this has to be `Week`.
      Frequency: insights.RecurrenceFrequencyWeek,
      Schedule: &insights.RecurrentSchedule{
         TimeZone: utils.String(recurrenceRaw["timezone"].(string)),
         Days:     utils.ExpandStringSlice(recurrenceRaw["days"].([]interface{})),
         Hours:    utils.ExpandInt32Slice(recurrenceRaw["hours"].([]interface{})),
         Minutes:  utils.ExpandInt32Slice(recurrenceRaw["minutes"].([]interface{})),
      },
   }
}
```

The implementation here follows the shcema type definition as described
in [official page](https://www.terraform.io/docs/extend/schemas/schema-types.html).

The deal here is that:

- for numeric types always expand from or flatten to `int`
- for float types always expand from or flatten to `float64`

Addtionally, add test case for each function in *common_marshal.go*.

### DIFF
--- a/azurerm/utils/common_marshal.go
+++ b/azurerm/utils/common_marshal.go
@@ -21,3 +21,289 @@ func FlattenStringSlice(input *[]string) []interface{} {
 	}
 	return result
 }
+
+func ExpandBoolSlice(input []interface{}) *[]bool {
+	result := make([]bool, 0)
+	for _, item := range input {
+		if item != nil {
+			result = append(result, item.(bool))
+		} else {
+			result = append(result, false)
+		}
+	}
+	return &result
+}
+
+func FlattenBoolSlice(input *[]bool) []interface{} {
+	result := make([]interface{}, 0)
+	if input != nil {
+		for _, item := range *input {
+			result = append(result, item)
+		}
+	}
+	return result
+}
+
+func ExpandUintSlice(input []interface{}) *[]uint {
+	result := make([]uint, 0)
+	for _, item := range input {
+		if item != nil {
+			result = append(result, uint(item.(int)))
+		} else {
+			result = append(result, 0)
+		}
+	}
+	return &result
+}
+
+func FlattenUintSlice(input *[]uint) []interface{} {
+	result := make([]interface{}, 0)
+	if input != nil {
+		for _, item := range *input {
+			result = append(result, int(item))
+		}
+	}
+	return result
+}
+
+func ExpandUint8Slice(input []interface{}) *[]uint8 {
+	result := make([]uint8, 0)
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
hi @magodo 

Thanks for this PR.

Taking a look through here whilst I could potentially see value in adding some of these functions - Terraform intentionally differs from the Azure SDK for Go (and the Azure API) in several areas to provide a better user experience. For example rather than returning a slice of Boolean's we'd instead carve this up into a more useful experience for users - which ultimately means that we wouldn't end up using these methods. As such whilst like to thank you for this contribution - at this time I'm going to close this PR for the moment.

Thanks!

magodo:
Hi @tombuildsstuff, I see, thank you for your detailed explaination.

PR #4959, opened 2019-11-22
provider: being more descriptive with the resource provider error

tombuildsstuff:
This PR extends the work done by @StefanSchoof in #4909 to be more descriptive about why we do this, and the API errors returned if Resource Provider Registration is disabled - and also splits this onto multiple lines to aid copy/pasting for users on Windows.


StefanSchoof:
@tombuildsstuff Great, this should be reduce the problems people running in this problem.

Have you seen that I also tried to include all missing providers in the error message: https://github.com/hashicorp/go-azure-helpers/pull/46? So a user can request all missing providers to register by an admin in one step.

PR #4961, opened 2019-11-22
Virtual Wan: refactoring & temporarily removing from the release

tombuildsstuff:
This PR does several things:

* Removes the `azurerm_virtual_wan` resource from the release, since `virtual_network_connections` need to be managed in a separate resource to allow users to be explicit around ordering 
* Refactors the `azurerm_virtual_wan` Data Source & Resource to ensure that fields are always set, to ensure diff's are always shown
* Updates the documentation to ensure that all fields are descriptive

Whilst it's unfortunate to pull a resource post-merging - prior experience with the Networking API's has shown that Virtual Network Connections need to be managed as independent resources, to allow users to enforce ordering upon deletion to work around issues in the Azure API.

To be able to ship the `azurerm_virtual_wan` resource we'll then split the Virtual Network Connections out to their own resource - at which point we can then ship this resource.

### DIFF
--- a/azurerm/data_source_virtual_hub.go
+++ b/azurerm/data_source_virtual_hub.go
@@ -128,33 +128,37 @@ func dataSourceArmVirtualHubRead(d *schema.ResourceData, meta interface{}) error
 	}
 	if props := resp.VirtualHubProperties; props != nil {
 		d.Set("address_prefix", props.AddressPrefix)
-		if props.VirtualWan != nil {
-			if err := d.Set("virtual_wan_id", props.VirtualWan.ID); err != nil {
-				return fmt.Errorf("Error setting `virtual_wan_id`: %+v", err)
-			}
-		}
-		if props.VpnGateway != nil {
-			if err := d.Set("s2s_vpn_gateway_id", props.VpnGateway.ID); err != nil {
-				return fmt.Errorf("Error setting `s2s_vpn_gateway_id`: %+v", err)
-			}
-		}
-		if props.P2SVpnGateway != nil {
-			if err := d.Set("p2s_vpn_gateway_id", props.P2SVpnGateway.ID); err != nil {
-				return fmt.Errorf("Error setting `p2s_vpn_gateway_id`: %+v", err)
-			}
-		}
+
+		var expressRouteGatewayId *string
 		if props.ExpressRouteGateway != nil {
-			if err := d.Set("express_route_gateway_id", props.ExpressRouteGateway.ID); err != nil {
-				return fmt.Errorf("Error setting `express_route_gateway_id`: %+v", err)
-			}
+			expressRouteGatewayId = props.ExpressRouteGateway.ID
 		}
+		d.Set("express_route_gateway_id", expressRouteGatewayId)
+
+		var p2sVpnGatewayId *string
+		if props.P2SVpnGateway != nil {
+			p2sVpnGatewayId = props.P2SVpnGateway.ID
+		}
+		d.Set("p2s_vpn_gateway_id", p2sVpnGatewayId)
+
+		if err := d.Set("route", flattenArmVirtualHubRoute(props.RouteTable)); err != nil {
+			return fmt.Errorf("Error setting `route`: %+v", err)
+		}
+
+		var vpnGatewayId *string
+		if props.VpnGateway != nil {
+			vpnGatewayId = props.VpnGateway.ID
+		}
+		d.Set("s2s_vpn_gateway_id", vpnGatewayId)
+
+		var virtualWanId *string
+		if props.VirtualWan != nil {
+			virtualWanId = props.VirtualWan.ID
+		}
+		d.Set("virtual_wan_id", virtualWanId)
+
 		if err := d.Set("virtual_network_connection", flattenArmVirtualHubVirtualNetworkConnection(props.VirtualNetworkConnections)); err != nil {
 			return fmt.Errorf("Error setting `virtual_network_connection`: %+v", err)
-		}
-		if props.RouteTable != nil {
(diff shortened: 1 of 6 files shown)


azurerm/data_source_virtual_hub.go
```diff
@@ -128,34 +128,38 @@ func dataSourceArmVirtualHubRead(d *schema.ResourceData, meta interface{}) error
...
-				return fmt.Errorf("Error setting `virtual_wan_id`: %+v", err)
-			}
-		}
-		if props.VpnGateway != nil {
-			if err := d.Set("s2s_vpn_gateway_id", props.VpnGateway.ID); err != nil {
-				return fmt.Errorf("Error setting `s2s_vpn_gateway_id`: %+v", err)
-			}
+
+		var expressRouteGatewayId *string
+		if props.ExpressRouteGateway != nil {
```
neil-yechenwei: Gateway fields should be readonly since gateway fields here is associated with virtual hub after double confirmed with service owner of this api. see my another PR https://github.com/terraform-providers/terraform-provider-azurerm/pull/4964

azurerm/data_source_virtual_hub.go
```diff
@@ -128,34 +128,38 @@ func dataSourceArmVirtualHubRead(d *schema.ResourceData, meta interface{}) error
...
+			expressRouteGatewayId = props.ExpressRouteGateway.ID
 		}
+		d.Set("express_route_gateway_id", expressRouteGatewayId)
+
+		var p2sVpnGatewayId *string
```
neil-yechenwei: same as above. Gateway fields should be readonly since gateway fields here is associated with virtual hub after double confirmed with service owner of this api. see my another PR #4964

azurerm/data_source_virtual_hub.go
```diff
@@ -128,34 +128,38 @@ func dataSourceArmVirtualHubRead(d *schema.ResourceData, meta interface{}) error
...
-				return fmt.Errorf("Error setting `express_route_gateway_id`: %+v", err)
-			}
+		d.Set("p2s_vpn_gateway_id", p2sVpnGatewayId)
+
+		if err := d.Set("route", flattenArmVirtualHubRoute(props.RouteTable)); err != nil {
+			return fmt.Errorf("Error setting `route`: %+v", err)
 		}
+
+		var vpnGatewayId *string
+		if props.VpnGateway != nil {
```
neil-yechenwei: same as above. Gateway fields should be readonly since gateway fields here is associated with virtual hub after double confirmed with service owner of this api. see my another PR #4964

azurerm/data_source_virtual_hub.go
```diff
@@ -128,34 +128,38 @@ func dataSourceArmVirtualHubRead(d *schema.ResourceData, meta interface{}) error
...
+			vpnGatewayId = props.VpnGateway.ID
+		}
+		d.Set("s2s_vpn_gateway_id", vpnGatewayId)
+
+		var virtualWanId *string
```
neil-yechenwei: this field should be added attribute "forceNew" since virtual hub needs to be rebuilt when this field is changed. Thus the behavior here is the same as the resource in azure portal.

azurerm/data_source_virtual_hub.go
```diff
@@ -128,34 +128,38 @@ func dataSourceArmVirtualHubRead(d *schema.ResourceData, meta interface{}) error
...
+		if props.VirtualWan != nil {
+			virtualWanId = props.VirtualWan.ID
+		}
+		d.Set("virtual_wan_id", virtualWanId)
```
neil-yechenwei: this field should be added attribute "forceNew" since virtual hub needs to be rebuilt when this field is changed. Thus the behavior here is the same as the resource in azure portal. see my another PR #4964

azurerm/resource_arm_virtual_hub.go
```diff
@@ -71,6 +71,7 @@ func resourceArmVirtualHub() *schema.Resource {
 				ValidateFunc: azure.ValidateResourceID,
 			},
 
+			// TODO: these need to be managed via a separate resource
 			"virtual_network_connection": {
```
neil-yechenwei: You mean I need to add new data resource file and resource file for this field as a separate resource?

azurerm/resource_arm_virtual_hub.go
```diff
@@ -106,6 +107,7 @@ func resourceArmVirtualHub() *schema.Resource {
 			},
 
 			"route": {
+				// TODO: confirm this wants to be a TypeSet
 				Type:     schema.TypeList,
```
neil-yechenwei: No index field here so that i think this field as list should be ok, right?

tombuildsstuff: no, in testing this needs to be a set

azurerm/resource_arm_virtual_hub.go
tombuildsstuff: I've got a PR coming adding these - but yes, these need to be managed as separate resources

azurerm/data_source_virtual_hub.go
tombuildsstuff: indeed - I've got a PR upcoming which handles this (and adds the dependencies for this resource to be functional)

website/azurerm.erb
```diff
@@ -457,10 +457,6 @@
                 <li<%= sidebar_current("docs-azurerm-datasource-virtual-network-gateway-connection") %>>
                     <a href="/docs/providers/azurerm/d/virtual_network_gateway_connection.html">azurerm_virtual_network_gateway_connection</a>
                 </li>
-
```
katbyte: I'm not sure this will remove it from the new registry docs? 🤔

tombuildsstuff: I've raised this with the team separately / via Slack

PR #4964, opened 2019-11-22
Update schema for azurerm_virtual_hub

neil-yechenwei:
After double confirmed with the service owner of this api, so updated the field virtual_wan_id as forceNew and these three gateway fields as readonly.

### DIFF
--- a/azurerm/resource_arm_virtual_hub.go
+++ b/azurerm/resource_arm_virtual_hub.go
@@ -111,18 +111,18 @@ func resourceArmVirtualHub() *schema.Resource {
 			},
 
 			"s2s_vpn_gateway_id": {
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
hi @neil-yechenwei 

Thanks for this PR.

After spending some time investigating the Virtual WAN resource it appears there's some larger changes required here - as such unfortunately we're having to pull it from the release for the moment, as is being done in #4951.

Having spent some time on this yesterday I've got a branch locally which includes both these changes and the larger changes/other resources required to ship the Virtual WAN resource - as such whilst I'd like to thank you for this contribution I'm going to close this PR for the moment in favour of an upcoming PR which includes these larger changes.

Thanks!

PR #4966, opened 2019-11-23
Storage network rules are now stored in state, regardless of the valu…

samart:
…es of IPRules and VNetRules, which are optional.  Plan now detects changes properly.

Moved check.  

Network rules now stored in statefile

"network_rules": [
  {
    "bypass": [
      "AzureServices",
      "Logging",
      "Metrics"
    ],
    "default_action": "Deny",
    "ip_rules": [],
    "virtual_network_subnet_ids": []
  }
],

and Plan now diffs this section properly.

      ~ network_rules {
          ~ bypass                     = [
                "AzureServices",
                "Logging",
              - "Metrics",
            ]
            ip_rules                   = []
            virtual_network_subnet_ids = []
        }

Validated configuration in Azure portal as valid.
Screen Shot 2019-11-23 at 4 40 23 PM

### DIFF
--- a/azurerm/resource_arm_storage_account.go
+++ b/azurerm/resource_arm_storage_account.go
@@ -1406,7 +1406,6 @@ func expandQueuePropertiesCors(input []interface{}) *queues.Cors {
 }
 
 func flattenStorageAccountNetworkRules(input *storage.NetworkRuleSet) []interface{} {
-
 	networkRules := make(map[string]interface{})
 
 	if len(*input.IPRules) > 0 {
--- a/azurerm/resource_arm_storage_account.go
+++ b/azurerm/resource_arm_storage_account.go
@@ -1406,13 +1406,15 @@ func expandQueuePropertiesCors(input []interface{}) *queues.Cors {
 }
 
 func flattenStorageAccountNetworkRules(input *storage.NetworkRuleSet) []interface{} {
-	if len(*input.IPRules) == 0 && len(*input.VirtualNetworkRules) == 0 {
-		return []interface{}{}
-	}
+
 	networkRules := make(map[string]interface{})
 
-	networkRules["ip_rules"] = schema.NewSet(schema.HashString, flattenStorageAccountIPRules(input.IPRules))
-	networkRules["virtual_network_subnet_ids"] = schema.NewSet(schema.HashString, flattenStorageAccountVirtualNetworks(input.VirtualNetworkRules))
+	if len(*input.IPRules) > 0 {
+		networkRules["ip_rules"] = schema.NewSet(schema.HashString, flattenStorageAccountIPRules(input.IPRules))
+	}
+	if len(*input.VirtualNetworkRules) > 0 {
+		networkRules["virtual_network_subnet_ids"] = schema.NewSet(schema.HashString, flattenStorageAccountVirtualNetworks(input.VirtualNetworkRules))
+	}
 	networkRules["bypass"] = schema.NewSet(schema.HashString, flattenStorageAccountBypass(input.Bypass))
 	networkRules["default_action"] = string(input.DefaultAction)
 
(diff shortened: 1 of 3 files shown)


samart:
This fixes, what seems to be a bug in the storage account resource, where on reading from azure, it does not save the state of network rules in some cases.

azurerm/resource_arm_storage_account.go
```diff
@@ -1406,13 +1406,14 @@ func expandQueuePropertiesCors(input []interface{}) *queues.Cors {
...
-	if len(*input.IPRules) == 0 && len(*input.VirtualNetworkRules) == 0 {
-		return []interface{}{}
-	}
 	networkRules := make(map[string]interface{})
 
-	networkRules["ip_rules"] = schema.NewSet(schema.HashString, flattenStorageAccountIPRules(input.IPRules))
-	networkRules["virtual_network_subnet_ids"] = schema.NewSet(schema.HashString, flattenStorageAccountVirtualNetworks(input.VirtualNetworkRules))
+	if len(*input.IPRules) > 0 {
+		networkRules["ip_rules"] = schema.NewSet(schema.HashString, flattenStorageAccountIPRules(input.IPRules))
+	}
```
tombuildsstuff: can we remove this `if` check here? the flatten function will return an empty array when this is nil - which'll ensure we track diff's here:

```suggestion
	networkRules["ip_rules"] = schema.NewSet(schema.HashString, flattenStorageAccountIPRules(input.IPRules))
```

azurerm/resource_arm_storage_account.go
```diff
@@ -1406,13 +1406,14 @@ func expandQueuePropertiesCors(input []interface{}) *queues.Cors {
...
+	if len(*input.VirtualNetworkRules) > 0 {
+		networkRules["virtual_network_subnet_ids"] = schema.NewSet(schema.HashString, flattenStorageAccountVirtualNetworks(input.VirtualNetworkRules))
+	}
```
tombuildsstuff: can we remove this `if` check here? the flatten function will return an empty array when this is nil - which'll ensure we track diff's here:

```suggestion
	networkRules["virtual_network_subnet_ids"] = schema.NewSet(schema.HashString, flattenStorageAccountVirtualNetworks(input.VirtualNetworkRules))
```

samart: @tombuildsstuff now that i am removing the checks completely, the current test "TestAccAzureRMStorageAccount_networkRules" is actually good enough, i think. 

https://github.com/terraform-providers/terraform-provider-azurerm/blob/0fd80ac109826ceb3c2cd6ca791b5943e45ec0c4/azurerm/resource_arm_storage_account_test.go#L636

 previously, it didn't test the conditional branch and now that i'm removing the branches completely.

what do you think?

azurerm/resource_arm_storage_account.go
```diff
@@ -1406,13 +1406,14 @@ func expandQueuePropertiesCors(input []interface{}) *queues.Cors {
 }
 
 func flattenStorageAccountNetworkRules(input *storage.NetworkRuleSet) []interface{} {
-	if len(*input.IPRules) == 0 && len(*input.VirtualNetworkRules) == 0 {
-		return []interface{}{}
-	}
```
tombuildsstuff: 🤔 reading this again I believe this wants to check if `input` is nil, rather than that these specific properties are nil

tombuildsstuff:
hey @samart 

Thanks for pushing those changes - apologies for the delayed re-review here!

After spending some time looking/thinking through this, I believe it may be possible to achieve this [by making this nil-check instead check the top-level `input` object as mentioned in this comment](https://github.com/terraform-providers/terraform-provider-azurerm/pull/4966/files#r359407454) rather than checking the inner-values.

I've spent a little time confirming that this is the case and have opened #5210 which includes these changes - as such whilst I'd like to thank you for this contribution, so that we can get this merged I hope you don't mind but I'm going to close this in favour of #5210.

Thanks!

samart:
Ok, thanks.   I can test the other change when its merged as well.

PR #4982, opened 2019-11-26
MsSql database blob auditing policies to be added

yupwei68:
MsSql database blob auditing policies to be added


tombuildsstuff:
hi @yupwei68 

Thanks for this PR :)

Taking a look though there's two bits of functionality within this Pull Request which both need to be designed slightly differently.

Since both of these resources can be conditionally Enabled/Disabled - rather than being independent resources, these need to be nested within the resources they're modifying. In this instance that's within the `azurerm_sql_server` and `azurerm_sql_database` resources - where we can use the presence of the sub-block to signify if these should be Enabled/Disabled - for example:

```hcl
  # ...
}
```

where the omission of an `access_policy` block means that this should be Disabled - where as:

```hcl
  # ...

  access_policy { ... }
}
```

the presence of the `access_policy` block means this should be Enabled with the given configuration.

Since both of the pieces of functionality in this Pull Request need to be refactored in the same manner and these changes will be fairly large - rather than trying to handle this in a single Pull Request, I'm going to suggest that we close this Pull Request in favour of two new PR's, which:

1. Split the two pieces of functionality into nested objects within the `azurerm_sql_server` and `azurerm_sql_database` resources (respectively).
2. Remove the Data Sources (and don't add this information to the `azurerm_sql_server` and `azurerm_sql_database` resources) - since the information returned here isn't likely to be directly used by users.

As such whilst I'd like to thank you for this contribution - I'm going to close this Pull Request for the moment.

Thanks!

yupwei68:
Hi @tombuildsstuff I have made the corresponding two changes in new PR DB: https://github.com/terraform-providers/terraform-provider-azurerm/pull/5049 & server: https://github.com/terraform-providers/terraform-provider-azurerm/pull/5036. I choose to implement the `blob_extended_auditing_policy` for it's more inclusive.

PR #4984, opened 2019-11-26
New Resource: `azurerm_bot_channel_ms_teams`

HiltonGiesenow:
I'm very new to Terraform - just dipping my toes in - but I need an MsTeams channel registration for an Azure Bot. I've tried to write it here, to contribute to the project, but I've never used Go at all, and also have no idea how to test this new resource, or how to use it in a project. So, this is a PR but also a call for help...

I've gotten Go and Make installed and gotten the project to compile, on Windows. I tried replacing the terraform-provider-azurerm.exe with the newly-generated one but that didn't work.


azurerm/resource_arm_bot_channel_msteams.go
```diff
@@ -0,0 +1,218 @@
...
+
+			"location": azure.SchemaLocation(),
+
+			"bot_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+
+			"enabled": {
```
mbfrahry: It looks like there is no way for this to be disabled when we're using this resource. We can remove this value and pass in true when we create this structure down below.

azurerm/resource_arm_bot_channel_msteams.go
```diff
@@ -0,0 +1,218 @@
...
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
+			"calling_web_hook": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validate.NoEmptyStrings,
```
mbfrahry: There is some additional validation we need here. The API only accepts https and it adds a `/` at the end of the URL. Unfortunately, the API accepts a url without a `/` at the end but always returns that url with a `/` at the end. This causes a permanent diff within Terraform as it tries to set a url without `/` and the api always returns one with `/`.

azurerm/resource_arm_bot_channel_msteams_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/services/preview/botservice/mgmt/2018-07-12/botservice"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func testAccAzureRMBotChannelMsTeams_basic(t *testing.T) {
+	// if ok := skipMsTeamsChannel(); ok {
```
mbfrahry: This can be removed since an MsTeams channel doesn't need any credentials to be configured

azurerm/resource_arm_bot_channel_msteams_test.go
```diff
@@ -0,0 +1,191 @@
...
+  location            = "${azurerm_bot_channels_registration.test.location}"
+  resource_group_name = "${azurerm_resource_group.test.name}"
+  enabled 			  = false
+  calling_web_hook    = "http://example2.com"
+  enable_calling      = false
+}
+`, template)
+}
+
+// func skipMsTeamsChannel() bool {
```
mbfrahry: This can be removed as well

azurerm/resource_arm_bot_channel_msteams_test.go
```diff
@@ -0,0 +1,191 @@
...
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMBotChannelMsTeamsExists(resourceName),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
```
mbfrahry: `ImportStateVerifyIgnore` is normally only called when a value isn't returned from the API and so we can't check whether it has been imported and set into state correctly. From my brief bit of testing, it does look like these values are returned from the api just fine. Did you see something different? If not, can we remove all the `ImportStateVerifyIgnore` from this file?

azurerm/resource_arm_bot_channel_msteams_test.go
```diff
@@ -0,0 +1,191 @@
...
+					"calling_web_hook",
+					"enable_calling",
+				},
+			},
+		},
+	})
+}
+
+func testAccAzureRMBotChannelMsTeams_update(t *testing.T) {
+	// if ok := skipMsTeamsChannel(); ok {
```
mbfrahry: This can be removed as well

azurerm/provider.go
```diff
@@ -217,6 +217,7 @@ func Provider() terraform.ResourceProvider {
 		"azurerm_batch_certificate":                                  resourceArmBatchCertificate(),
 		"azurerm_bot_channel_email":                                  resourceArmBotChannelEmail(),
 		"azurerm_bot_channel_slack":                                  resourceArmBotChannelSlack(),
+		"azurerm_bot_channel_msteams":                                resourceArmBotChannelMsTeams(),
```
mbfrahry: It looks like the UI has this as MSTeams but would it be more readable to split it out a little here to look like `azurerm_bot_channel_ms_teams`?

tombuildsstuff: it might be worth making this `microsoft_teams` rather than `ms` so this is clearer - WDYT?

azurerm/resource_arm_bot_channel_msteams_test.go
HiltonGiesenow: This is a bit beyond my knowledge here, but happy to make whatever changes are needed to the PR. What exactly needs to be changed?

azurerm/resource_arm_bot_channel_msteams.go
HiltonGiesenow: Sounds good, but not sure how to implement this?

azurerm/provider.go
HiltonGiesenow: If we look at sources like this
1 - https://github.com/Azure/azure-sdk-for-go/blob/7a9d2769e4a581b0b1bc609c71b59af043e05c98/services/preview/botservice/mgmt/2018-07-12/botservice/models.go
and 2 - https://docs.microsoft.com/en-us/cli/azure/bot/msteams?view=azure-cli-latest

It seems like "MsTeams" is the consistent use Microsoft themselves make - might be better if we keep it this way, especially if someone is migrating from something else?

HiltonGiesenow:
Thanks guys for all the help, especially considering I'm out of my depth, so it added extra work your side! Comments above and some of the changes implemented/checked in

HiltonGiesenow:
@tombuildsstuff @mbfrahry hey guys, just wondering if you've had a chance to look more at this? It seems like we're very close to finishing it.

Out of interest, Matthew it sounds like you've maybe worked on deploying bots with Terraform? If so, I'm curious if you've ever managed to get an MsTeams bot to work? It requires a lot more around the appid and apppassword than a normal bot, so I'm trying to figure out the ins and outs (it's turning out there are some interesting constraints on the Service Principal that aren't obvious when you create it easily in the portal UI)

azurerm/resource_arm_bot_channel_msteams_test.go
mbfrahry: This should be able to be removed since I saw these values being returned from the api. My concern was that it's been added so maybe you didn't see these values being returned from the api or did you copy this section in from another resource?

azurerm/resource_arm_bot_channel_msteams.go
mbfrahry: So you should be able to write a ValidateFunction that checks that the string starts with `https` and ends with `/`. 

```
func validateCallingWebHook() schema.SchemaValidateFunc {
	return func(i interface{}, k string) (warnings []string, errors []error) {
		value := i.(string)
		if !strings.HasPrefix(value, "https://" ) || !strings.HasSuffix(value, "/") {
			errors = append(errors, fmt.Errorf("invalid `calling_web_hook`, must start with `https://` and end with `/`"))
		}

		return warnings, errors
	}
}
```

That'll need to be checked and tested as well

azurerm/provider.go
mbfrahry: I agree that it should be kept as MsTeams but will push again that it should be `azurerm_bot_channel_ms_teams`. Microsoft makes the distinction between MsTeams with capitals more often than not and we can mirror that but splitting out the words from each other

mbfrahry:
I haven't taken a look at implementing a MS Team Bot Service. I'd be curious to hear more about it but it might be easier to track/comment on that in a new issue/feature request?

HiltonGiesenow:
Sorry for the delay, been a hectic few days. I've made the changes - hope they're correct... As I said, I've never used Go before this, so I managed to get the "fmt" command working, and got it to build, but I still don't know how to actually test my changes. If there's still  work to be done after this checkin, maybe you can give me some guidance so I can take that work off you if necessary?

mbfrahry:
All good. Testing this specific resource isn't very straightforward. I don't mind taking it from here as you've done the bulk of the work and testing this resource is different than testing every other resource on Azure. Do you mind if I take it off your hands and then walk you through how I'm going to test it and then how you can test it locally when I'm finished?

HiltonGiesenow:
Sounds great!

mbfrahry:
Thanks for the work here @HiltonGiesenow. To test this through our testing framework, you need to setup quite a few environment variables that are required to authenticate with your account.
```
"ARM_CLIENT_ID",
"ARM_CLIENT_SECRET",
"ARM_SUBSCRIPTION_ID",
"ARM_TENANT_ID",
"ARM_TEST_LOCATION",
"ARM_TEST_LOCATION_ALT",
"ARM_TEST_LOCATION_ALT2",
```
From there you can run 
```
make testacc TESTARGS='-run=TestAccAzureRMBotChannel'
```
which will run through all the tests with that prefix. This will spin up and tear down actual resources on your account which could be expensive depending on how often you run them and how long they stay up but this is how we test Terraform code. Does that make sense.

Other than that, thank you again for the work here and this should make it into the next release

HiltonGiesenow:
glad to have been of help, hopefully the first of many. I'll have to learn Go soon I guess ;>.

On testing - the other  stuff I can figure out, but what ClientId/ClientSecret am I using? What I mean is, I assume I need to create an App, but what setup does it need exactly?

mbfrahry:
Have you seen our guide for how to setup a service principal and client secret? https://www.terraform.io/docs/providers/azurerm/guides/service_principal_client_secret.html. That might be a good place to start if you haven't.

HiltonGiesenow:
ah, no, hadn't seen that - I'm just running interactively so far - thanks so much, will give it a go!

Issue #4932, opened 2019-11-20
Support for extra info storageaccount name validation error


sanderobdeijn:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When creating a storage account name validation takes place and if the name isn't correct you get the following error:
Error: name can only consist of lowercase letters and numbers, and must be between 3 and 24 characters long

When using interpolation is it sometimes difficult to know what the used name was. Is it possible that the used name is shown in the error?

### New or Affected Resource(s)

* azurerm_storage_account

### Potential Terraform Configuration

Not Applicable
### References

Not Applicable

PR #5027, opened 2019-11-29
print storage account name in validation error

skdeep12:
Prints the storage account name in validation error as required in #4932 

Fixes: #4932


sanderobdeijn:
thank you @skdeep12

PR #5033, opened 2019-12-01
refactor: moving clients A -> C

tombuildsstuff:
This PR relocates Clients A through C into the common `clients` package which (in time) enables us to refactor Data Sources/Resources. I've done this in a small chunk (A through C) since the `compute` package was already present - so this seemed like a good point to break this up.

Whilst it's probably a little early - I've taken the opportunity to hook up the client registration for each package at the same time

A test from within the affected packages runs - so this is hooked up correctly 👍 

```
$ acctests azurerm TestAccAzureRMContainerGroup_SystemAssignedIdentity
=== RUN   TestAccAzureRMContainerGroup_SystemAssignedIdentity
=== PAUSE TestAccAzureRMContainerGroup_SystemAssignedIdentity
=== CONT  TestAccAzureRMContainerGroup_SystemAssignedIdentity
--- PASS: TestAccAzureRMContainerGroup_SystemAssignedIdentity (199.92s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	199.974s
```


azurerm/internal/acceptance/testing.go
```diff
@@ -0,0 +1,94 @@
...
+			t.Fatalf("`%s` must be set for acceptance tests!", variable)
+		}
+	}
+}
+
+func Location() string {
+	return os.Getenv("ARM_TEST_LOCATION")
+}
+
+func AltLocation() string {
```
katbyte: Part of me wants to rename this to `LocationAlt` while we do this but thats my OCD talking

tombuildsstuff: The other option would be to make Location a struct returning Primary/Default, Secondary and Ternary?

PR #5045, opened 2019-12-03
Update default config to VMSS+SLB during aks create

neil-yechenwei:
This PR is the code implement for the issue https://github.com/terraform-providers/terraform-provider-azurerm/issues/4465

### DIFF
--- a/azurerm/resource_arm_kubernetes_cluster.go
+++ b/azurerm/resource_arm_kubernetes_cluster.go
@@ -126,7 +126,7 @@ func resourceArmKubernetesCluster() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 							ForceNew: true,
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
hi @neil-yechenwei 

Thanks for this PR.

The `agent_pool_profile` block within the `azurerm_kubernetes_cluster` resource is Deprecated and will be removed in 2.0, since it's been replaced with the `default_node_pool` block and the separate `azurerm_kubernetes_node_pool` resource to work around a breaking change to the AKS API.

With regard to updating the default Load Balancer type to "Standard", since changing this default value would be a breaking change we're intentionally holding off updating this until we're closer to 2.0 - as such whilst I'd like to thank you for this contribution I'm going to close this PR for the moment.

Thanks!

PR #5046, opened 2019-12-03
Healthcare resource. CORS block is fixed to avoid unnecessary diffs

kutsovmqs:
The issue is that default CORS values are create in Azure, even if none are specified in the TF script. If none are specified in the TF script then any TF plan or apply will see that there is a CORS Configuration and want to delete it accordingly, since it's not in the script.


kutsovmqs:
Hi @tombuildsstuff . Could you tell me what I need to do to move this PR forward?

tombuildsstuff:
Running the tests for this and comparing them with master - both are currently failing due to a behavioural change in the API; as such I'm going to merge this for the moment then send a second PR to fix that up - thanks again @kutsovmqs

PR #5048, opened 2019-12-03
New resource: dedicated_host_group (compute)

magodo:
Add new resource `dedicated_host_group`, which belongs to Compute Provider.

> A host group is a resource that represents a collection of dedicated hosts. You create a host group in a region and an availability zone, and add hosts to it.

Please refer to this [link](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/dedicated-hosts#groups-hosts-and-vms) for more info.


website/docs/d/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,61 @@
+---
```
katbyte: Could we add this to the index.erb

website/docs/r/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,70 @@
+---
```
katbyte: Could we add this file to the index.erb

magodo:
@katbyte, this PR is filed before the refactor you've made. Now I think I'd better to reflect the refactoring to this resource with your comment applied and raise another PR.

azurerm/data_source_dedicated_host_group.go
```diff
@@ -0,0 +1,101 @@
...
+	d.Set("name", name)
+	d.Set("resource_group_name", resourceGroupName)
+	if location := resp.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	if dedicatedHostGroupProperties := resp.DedicatedHostGroupProperties; dedicatedHostGroupProperties != nil {
+		if err := d.Set("hosts", flattenArmDedicatedHostGroupSubResourceReadOnly(dedicatedHostGroupProperties.Hosts)); err != nil {
+			return fmt.Errorf("Error setting `hosts`: %+v", err)
+		}
+		d.Set("platform_fault_domain_count", int(*dedicatedHostGroupProperties.PlatformFaultDomainCount))
```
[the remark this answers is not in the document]
magodo: Why is that? I assume a `schema.TypeInt` corresponds to `int`, not `int32`.

magodo:
Replaced by #5307

PR #5064, opened 2019-12-04
resource_arm_virtual_machine.go: Add read and update support for license_type.

NilsBusche:
At the moment `license_type` was not read and refreshed. This PR adds read support for this attribute.

Furthermore you could only set `license_type` to one of `Windows_Client` or `Windows_Server` but not back to the default. The Azure API uses `None` as default so it should be used in Terraform as well. This allows you to switch back.

### DIFF
--- a/azurerm/resource_arm_virtual_machine.go
+++ b/azurerm/resource_arm_virtual_machine.go
@@ -159,13 +159,13 @@ func resourceArmVirtualMachine() *schema.Resource {
 			},
 
 			"license_type": {
-				Type:             schema.TypeString,
-				Optional:         true,
-				Computed:         true,
-				DiffSuppressFunc: suppress.CaseDifference,
+				Type:     schema.TypeString,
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
hey @NilsBusche 

We're introducing replacements for the `azurerm_virtual_machine` and `azurerm_virtual_machine_scale_set` resources in the next major version of the Azure Provider (2.0 - more details can be found in #2807).

Since both of these resources are being superseded in the near future - we're instead looking to add this functionality to the replacement resources instead (in this case the upcoming `azurerm_linux_vitual_machine` and `azurerm_windows_virtual_machine` resources).

As such whilst I'd like to thank you for this contribution, I'm going to close this Pull Request for the moment - I've opened #5290 (which I've assigned to the 2.0 milestone - and tagged you in, so you should be subscribed to updates 🤞) to ensure this functionality is supported by the new resources.

Thanks!

PR #5072, opened 2019-12-05
azurerm_subnet - make `network_security_group_id` as computed

### LINKED ISSUE #4983: Subnet NSG deprecation
### Community Note

Provider 1.37

### Affected Resource(s)

azurerm_subnet_network_security_group_association
azurerm_subnet

### Expected Behavior

Terraform apply gives a warning that network_security_group_id has been deprecated and to ALSO use azurerm_subnet_network_security_group_association ready for when the change is implemented.

### Actual Behavior

Terraform apply gives a warning that network_security_group_id has been deprecated and to use azurerm_subnet_network_security_group_association instead.
This results in the the NSG being removed from the subnet, reapplied on a second run, removed on a third and so on.

### Steps to Reproduce

Use azurerm_subnet_network_security_group_association without network_security_group_id in subnet

### Important Factoids

The documentation on this was improved on several months ago, but the Terraform message itself should also reflect this as it could lead to someone making the recommended change and removing the NSGs from their subnets.

njuCZ:
when use `azurerm_subnet_network_security_group_association` without `network_security_group_id` in subnet, 
first terraform apply, everything is OK
but terraform plan will show `network_security_group_id` state is not consistent, and the second time terraform apply, the terraform will set subnet network security group null

Though `network_security_group_id` is a deprecated field and will be removed soon, but currently I think it's better for fix the bug

(fixes #4983)

### DIFF
--- a/azurerm/resource_arm_subnet.go
+++ b/azurerm/resource_arm_subnet.go
@@ -60,6 +60,7 @@ func resourceArmSubnet() *schema.Resource {
 			"network_security_group_id": {
 				Type:       schema.TypeString,
 				Optional:   true,
+				Computed:   true,
 				Deprecated: "Use the `azurerm_subnet_network_security_group_association` resource instead.",
 			},
 
--- a/azurerm/resource_arm_subnet_network_security_group_association_test.go
+++ b/azurerm/resource_arm_subnet_network_security_group_association_test.go
@@ -236,7 +236,6 @@ resource "azurerm_subnet" "test" {
   resource_group_name       = "${azurerm_resource_group.test.name}"
   virtual_network_name      = "${azurerm_virtual_network.test.name}"
   address_prefix            = "10.0.2.0/24"
-  network_security_group_id = "${azurerm_network_security_group.test.id}"
 }
 
 resource "azurerm_network_security_group" "test" {


tombuildsstuff:
hi @njuCZ 

Thanks for this PR :)

At this the `network_security_group_id` field is intentionally not set to Computed, since this breaks the ability to remove this values using the existing resources - instead the documentation at this time requires that users specify the association for this field (and the `route_table_id` field) both as an independent resource and within the `azurerm_subnet` resource at this time. In the upcoming major version of the AzureRM Provider (2.0) - we'll remove these fields from the `azurerm_subnet` resource in favour of the independent resources (which is why these are deprecated).

Whilst we appreciate this isn't the best user experience until 2.0 - unfortunately making this change would be a breaking behavioural change which would catch users out, which is why we've intentionally avoided making these fields `Computed` at this time. As such whilst I'd like to thank you for this contribution I'm going to close this PR for the moment.

Thanks!

njuCZ:
@tombuildsstuff  thank you for pointing the point: making a field `computed` will make terraform can not detect changes,  so deleting can not work.
By the way, should we make `subnet` field in `azurerm_virtual_network` deprecated ? because it's also computed, and we can use resource `azurerm_subnet`. because currently if you use the `subnet` field, you can not remove it by setting it null

tombuildsstuff:
@njuCZ no that's a separate valid use-case since Subnets can be defined inline (to ensure only those exist) or externally - in this instance this is to work around a design issue in the Azure Networking API's

PR #5075, opened 2019-12-05
Bug: List storage accounts returns only first page

benjamin37:
When calling the current list storage account method it just returns the first page of storage accounts if the response is a paginated response. When the storage account that the provider is looking for is not in the first page I get the error message 

> Unable to locate Storage Account $STORAGE_ACCOUNT_NAME!

How to reproduce? In one of our subscriptions this happens when updating storage accounts relative consistently (about 10% of the calls). Sadly I found no setup to reproduce this paginated response consistently. If you find a way to consistently reproduce a paginated response form list storage accounts please let me know.

### DIFF
--- a/azurerm/internal/services/storage/helpers.go
+++ b/azurerm/internal/services/storage/helpers.go
@@ -56,7 +56,7 @@ func (client Client) AddToCache(accountName string, props storage.Account) error
 	accountsLock.Lock()
 	defer accountsLock.Unlock()
 
-	account, err := client.populateAccountDetails(accountName, props)
+	account, err := populateAccountDetails(accountName, props)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
👋 

I've run the tests for this PR but they fail due to a separate issue in master which has since been fixed (#5113), since this PR is branched off from that commit.

I've rebased this PR locally and run the Storage Container tests and this looks good:

```
$ acctests azurerm TestAccAzureRMStorageContainer_basic
=== RUN   TestAccAzureRMStorageContainer_basic
=== PAUSE TestAccAzureRMStorageContainer_basic
=== CONT  TestAccAzureRMStorageContainer_basic
--- PASS: TestAccAzureRMStorageContainer_basic (126.26s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	126.317s
```

.. as such I believe we're good here - thanks again for this PR / apologies for the delay with this one.

Thanks!

PR #5081, opened 2019-12-05
azurerm_api_management_api - correctly setting the soap API type when `soap_pass_through` is true

### LINKED ISSUE #5080: azurerm_api_management_api soap_pass_through does not work when importing WSDL
### Community Note

Terraform v0.12.16  
azurerm v1.37.0 (and older)

### Affected Resource(s)

* `azurerm_api_management_api`

```hcl
resource "azurerm_api_management_api" "submitter" {  
  api_management_name = "..."
  resource_group_name = "..."
  
  display_name = "Some display name"  
  name = "some-name"  
  path = "some-path"  
  protocols = ["https"]  
  revision = "1"  
  service_url = "https://example.org"  
  soap_pass_through = true  
  
 import {  
  content_format = "wsdl"  
  content_value = <<XML 
        [WSDL goes here]
     XML
  }
}
```
### Debug Output

Omitted, but I have a trace log of the bug and a working version - if needed.

### Panic Output

No panic.

### Expected Behavior

Expect a new Azure API Management API to be created with SOAP pass-through.
The property `soap_pass_through` is set to `true` in the state file, but that does not mean that the end-result is correct.
The correct property is not sent to the Azure REST API when the WSDL is sent in (during the import phase), meaning that the generated API operations are the same as having `soap_pass_through=false`.

### Actual Behavior

The API operations are not created as expected with `soap_pass_through=true`. If you try to create a API in API Management and choose "Import WSDL" and "SOAP pass-through", none of the operations will not have any policies. 
Instead, all endpoints are generated with policies that transform from JSON to XML and vice verca - the behavior that is expected if you do not choose SOAP pass-through.

### Steps to Reproduce

Use the configuration above and create a API in an existing API Management instance.

### Important Factoids

I have found the bug and fixed it. I have verified it by building my own version of the provider and testing it. 
PR is coming.

### References

None. I will submit a PR later.

tnicolaysen:
I am not completely sure if the modified block (line 222-229) is 100 % correct.
Not sure if there can be other combinations of those properties.

The change has been tested and verified locally for both soap_pass_through = true and false

Fixes #5080
Details in PR

### DIFF
--- a/azurerm/resource_arm_api_management_api.go
+++ b/azurerm/resource_arm_api_management_api.go
@@ -216,6 +216,18 @@ func resourceArmApiManagementApiCreateUpdate(d *schema.ResourceData, meta interf
 		}
 	}
 
+	var apiType apimanagement.APIType
+	var soapApiType apimanagement.SoapAPIType
+
+	soapPassThrough := d.Get("soap_pass_through").(bool)
+	if soapPassThrough {
+		apiType = apimanagement.Soap
+		soapApiType = apimanagement.SoapPassThrough
+	} else {
+		apiType = apimanagement.HTTP
+		soapApiType = apimanagement.SoapToRest
+	}
+
 	// If import is used, we need to send properties to Azure API in two operations.
 	// First we execute import and then updated the other props.
 	if vs, hasImport := d.GetOk("import"); hasImport {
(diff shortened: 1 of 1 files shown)


tnicolaysen:
API docs: https://docs.microsoft.com/en-us/rest/api/apimanagement/2019-01-01/apis/createorupdate

tnicolaysen:
@torresdal Do you want to have a look?

torresdal:
@tnicolaysen been a long time since I contributed here, so better leave that to the great folks at HashiCorp 😊 (cc @ tombuildsstuff)

PR #5082, opened 2019-12-05
New Resource: `azurerm_storage_account_network_rules`

mbfrahry:
Addresses #3522
```
--- PASS: TestAccAzureRMStorageAccountNetworkRules_basic (159.91s)
--- PASS: TestAccAzureRMStorageAccountNetworkRules_update (168.75s)
```


azurerm/resource_arm_storage_account_network_rules.go
```diff
@@ -0,0 +1,275 @@
...
+				Optional: true,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+					ValidateFunc: validation.StringInSlice([]string{
+						string(storage.AzureServices),
+						string(storage.Logging),
+						string(storage.Metrics),
+						string(storage.None),
+					}, true),
```
katbyte: Is there a reason for the case insensitivy?

azurerm/resource_arm_storage_account_network_rules.go
```diff
@@ -0,0 +1,275 @@
...
+
+			"ip_rules": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"virtual_network_subnet_ids": {
+				Type:     schema.TypeList,
```
katbyte: Should this be a set?

azurerm/resource_arm_storage_account_network_rules.go
```diff
@@ -0,0 +1,275 @@
...
+				Optional: true,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
```
katbyte: And we could validate this is an ID

azurerm/resource_arm_storage_account_network_rules.go
```diff
@@ -0,0 +1,275 @@
...
+						string(storage.Logging),
+						string(storage.Metrics),
+						string(storage.None),
+					}, true),
+				},
+				Set: schema.HashString,
+			},
+
+			"ip_rules": {
+				Type:     schema.TypeList,
```
katbyte: Should this be a a set?

azurerm/resource_arm_storage_account_network_rules.go
```diff
@@ -0,0 +1,275 @@
...
+				Optional: true,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
```
katbyte: and at least validated with no empty strings?

azurerm/resource_arm_storage_account_network_rules.go
```diff
@@ -0,0 +1,275 @@
...
+						string(storage.None),
+					}, true),
+				},
+				Set: schema.HashString,
+			},
+
+			"ip_rules": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
katbyte: as this is an opt computed block we should use the weird config atter mode hack so it can be unset?

azurerm/resource_arm_storage_account_network_rules.go
```diff
@@ -0,0 +1,275 @@
...
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"virtual_network_subnet_ids": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
katbyte: as this is an opt computed block we should use the weird config atter mode hack so it can be unset?

azurerm/resource_arm_storage_account_network_rules_test.go
```diff
@@ -0,0 +1,162 @@
...
+  }
+}
+
+resource "azurerm_storage_account_network_rules" "test" {
+  resource_group_name  = "${azurerm_resource_group.testrg.name}"
+  storage_account_name = "${azurerm_storage_account.testsa.name}"
+
+  default_action             = "Deny"
+  ip_rules                   = ["127.0.0.1"]
+  virtual_network_subnet_ids = ["${azurerm_subnet.test.id}"]
```
katbyte: what happens if we try and create this with nether of these two?

azurerm/resource_arm_storage_account_network_rules_test.go
```diff
@@ -0,0 +1,162 @@
...
+  }
+}
+
+resource "azurerm_storage_account_network_rules" "test" {
+  resource_group_name  = "${azurerm_resource_group.testrg.name}"
+  storage_account_name = "${azurerm_storage_account.testsa.name}"
+
+  default_action             = "Allow"
+  ip_rules                   = ["127.0.0.1"]
+  virtual_network_subnet_ids = ["${azurerm_subnet.test.id}"]
```
katbyte: could we update these to something different & a different amount?

azurerm/resource_arm_storage_account_network_rules_test.go
```diff
@@ -0,0 +1,162 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMStorageAccountExists("azurerm_storage_account.testsa"),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
```
katbyte: Should we add a back to basics step here

website/docs/r/storage_account_network_rules.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+The following arguments are supported:
+
+* `storage_account_name` - (Required) Specifies the name of the storage account. Changing this forces a new resource to be created. This must be unique across the entire Azure service, not just within the resource group.
+
+* `resource_group_name` - (Required) The name of the resource group in which to create the storage account. Changing this forces a new resource to be created.
+
+* `default_action` - (Required) Specifies the default action of allow or deny when no other rules match. Valid options are `Deny` or `Allow`.
+
+* `bypass` - (Optional)  Specifies whether traffic is bypassed for Logging/Metrics/AzureServices. Valid options are
+any combination of `Logging`, `Metrics`, `AzureServices`, or `None`.
```
katbyte: spurious line break?

website/docs/r/storage_account_network_rules.html.markdown
```diff
@@ -0,0 +1,95 @@
...
+* `storage_account_name` - (Required) Specifies the name of the storage account. Changing this forces a new resource to be created. This must be unique across the entire Azure service, not just within the resource group.
+
+* `resource_group_name` - (Required) The name of the resource group in which to create the storage account. Changing this forces a new resource to be created.
+
+* `default_action` - (Required) Specifies the default action of allow or deny when no other rules match. Valid options are `Deny` or `Allow`.
+
+* `bypass` - (Optional)  Specifies whether traffic is bypassed for Logging/Metrics/AzureServices. Valid options are
+any combination of `Logging`, `Metrics`, `AzureServices`, or `None`. 
+
+* `ip_rules` - (Optional) List of public IP or IP ranges in CIDR Format. Only IPV4 addresses are allowed. Private IP address ranges (as defined in [RFC 1918](https://tools.ietf.org/html/rfc1918#section-3)) are not allowed.
```
katbyte: Should e validate the property as such?

azurerm/resource_arm_storage_account_network_rules.go
mbfrahry: Nerp. Came over from some copy pasta

azurerm/resource_arm_storage_account_network_rules.go
mbfrahry: Sure should

azurerm/resource_arm_storage_account_network_rules.go
mbfrahry: Sure should

azurerm/resource_arm_storage_account_network_rules.go
mbfrahry: I don't believe I've seen this

katbyte: Search for `ConfigMode: schema.SchemaConfigModeAttr,`, its basically to force tf to allow array syntax in .12 because otherwise there is no way to send an empty list to clear the property.

website/docs/r/storage_account_network_rules.html.markdown
mbfrahry: Sure should

azurerm/resource_arm_storage_account_network_rules_test.go
mbfrahry: Oh, good call

azurerm/resource_arm_storage_account_network_rules_test.go
mbfrahry: Nothing odd. It creates it with the other parameters specified just fine. Added a test to confirm

azurerm/resource_arm_storage_account_network_rules.go
mbfrahry: Thank you thank you

azurerm/resource_arm_storage_account_network_rules.go
```diff
@@ -0,0 +1,297 @@
...
+	if storageAccount.NetworkRuleSet == nil {
+		return nil
+	}
+
+	// We can't delete a network rule set so we'll just update it back to the default instead
+	opts := storage.AccountUpdateParameters{
+		AccountPropertiesUpdateParameters: &storage.AccountPropertiesUpdateParameters{
+			NetworkRuleSet: &storage.NetworkRuleSet{
+				Bypass:        storage.AzureServices,
+				DefaultAction: storage.DefaultActionDeny,
```
tombuildsstuff: I think this wants to be Allow?

PR #5091, opened 2019-12-06
Allow ports to be set at group level in azurerm_container_group

### LINKED ISSUE #4413: Regarding #1662: Allow Container ~Group~ Ports to be specifically specified rather than aggregating the container ports
This is basically a rehash of issue [#1662](https://github.com/terraform-providers/terraform-provider-azurerm/issues/1662) which was closed as part of [#1930](https://github.com/terraform-providers/terraform-provider-azurerm/issues/1930) but does not seem to fix the issue stated in #1662

A container groups exposed ports are still an aggregation of the defined containers ports, eventhough ARM allows for a subset of ports to be exposed or am I missing something?

See [resource_arm_container_group.go](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_container_group.go)

Thanks

amasover:
Fixes #4413

Note: I suggested an alternate syntax in the linked issue, but decided not to implement it because a schema change like that is difficult for me, being a bit new to Golang/Terraform.

I think this PR breaks backwards compatibility. Previously, Terraform would aggregate ports from `container{}` blocks, but with this PR you must explicitly set ports on both the top-level resource and on individual containers.

I can't see a good way to implement the change without breaking compatibility. The goal is to enable the [use-case described here](https://docs.microsoft.com/en-us/azure/container-instances/container-instances-container-group-ssl). To achieve that goal, we can't assume that we know what ports the user wants exposed outside the container group, because some `container{}` level ports may be intended to be private.


azurerm/resource_arm_container_group.go
```diff
@@ -589,6 +620,7 @@ func resourceArmContainerGroupRead(d *schema.ResourceData, meta interface{}) err
 		if address := props.IPAddress; address != nil {
 			d.Set("ip_address_type", address.Type)
 			d.Set("ip_address", address.IP)
+			d.Set("ports", address.Ports)
```
amasover: Linter was failing here because I needed to flatten the ports and provide this as a `schema.Set`.

amasover:
Maybe we can avoid a breaking change like this:

* If no ports are explicitly exposed at the group level, assume that any exposed container ports are intended to also be exposed at the group level. (this is the behavior as it stands today)
* If no ports are explicitly exposed at the group level, but a new group level variable, `private_ports = true`, is set, assume that no exposed container ports are intended to be exposed at the group level. (Could see a use-case for this, if your app has microservices that need to talk to each other, but you don't need any public ports open).
* If any ports are explicitly exposed at the group level, then expose only those ports. Do not assume all exposed container ports are intended to be public, only expose the ones that are explicitly exposed at the group level. (this is the use-case that I want to enable)

Then maybe in v2 of the provider we could revert to the breaking change suggested in this PR, since v2 will have a few other breaking changes as well. And I guess if v2 is coming sooner rather than later, it might make some sense to just wait a bit, and just roll this PR's change into v2.

I'll leave it up to y'all what direction you want to go with this, especially because I'm a total newbie here, just wanted to throw some ideas out there :smiley_cat:

amasover:
I've updated this PR to be more consistent with the changes from v2.

w0ut0:
This is indeed the exact solution to issue [6409](https://github.com/terraform-providers/terraform-provider-azurerm/issues/6409). This would really come in handy for us.

amasover:
I think I misunderstood something when I presented this option:

> If no ports are explicitly exposed at the group level, but a new group level variable, private_ports = true, is set, assume that no exposed container ports are intended to be exposed at the group level. (Could see a use-case for this, if your app has microservices that need to talk to each other, but you don't need any public ports open).

It looks like Azure API doesn't accept that use-case (we get an error at least from autorest/validation):

> Error: Error creating/updating container group "atlantis-container-group" (Resource Group "aarontest-atlantis"): containerinstance.ContainerGroupsClient#CreateOrUpdate: Invalid input: autorest/validation: validation failed: parameter=containerGroup.ContainerGroupProperties.IPAddress.Ports constraint=Null value=(*[]containerinstance.Port)(nil) details: value can not be null; required parameter

There's one other case I thought of, that it turns out also isn't valid. What if you want to expose a public port, but not expose a container port (no idea why you'd want to do this). It turns out, we don't need to worry about that either, since the Azure API rejects it:

> Original Error: Code="UnreferencedIpAddressPorts" Message="Following ports '443,80' in the 'ipAddress' are not used by any container in container group 'atlantis-container-group'."

Since we can strike out those two cases, I think I can present the following as a way to not break backwards compatibility:

* If no ports are explicitly exposed at the group level, assume that any exposed container ports are intended to also be exposed at the group level. (this is the behavior as it stands today)
* If any ports are explicitly exposed at the group level, then expose only those ports. Do not assume all exposed container ports are intended to be public, only expose the ones that are explicitly exposed at the group level. (this is the use-case that I want to enable)

If I were to make changes to enable that behavior (in doing so, keeping backwards compatibility for existing configs), would you be willing to accept this PR, @katbyte 

Example config (once I make changes for backwards compatibility)...

No ports explicitly exposed at group level (will work the same way it does today):
```hcl
resource "azurerm_container_group" "webapp" {
    name                = "webapp-aci"
    #omitted lines of config...
    container {
        name             = "nginx"
        ports {
            port     = 443
            protocol = "TCP"
        }
    }
    container {
        name             = "webapp"
        ports {
            port     = 80
            protocol = "TCP"
        }
    }
}
```
This will expose both 443 and 80 publicly. Connecting into the public IP of the container group on port 443 will route you to the nginx container, while connecting to port 80 will route you to the webapp container.

```hcl
resource "azurerm_container_group" "webapp" {
    name                = "webapp-aci"
    #omitted lines of config...
    ports {
        port     = 443
        protocol = "TCP"
    }
    container {
        name             = "nginx"
        ports {
            port     = 443
            protocol = "TCP"
        }
    }
    container {
        name             = "webapp"
        ports {
            port     = 80
            protocol = "TCP"
        }
    }
}
```
Adding the new `ports` block at the container group level means that only port 443 is exposed publicly. Connecting to the public IP of the container group on port 443 will route you to nginx, but it is only possible to connect to the webapp on port 80 from inside the container group (so nginx is able to terminate SSL and then route to the unsecured port on the webapp).

There's another use-case that I've thought of, based on the second example Terraform config. I'll explain it here for completeness. What if we want to expose some ports publicly, and map them to certain containers. But we also want a container with no ports exposed. For example, maybe we have a 3rd container that generates reports:

```hcl
    container {
        name             = "report-generator"
    }
```

Its job is to sit inside the container group vnet, and read data from port 80 on the webapp, then send that data outbound to a third-party service. It doesn't need any ports open to accomplish that, as nothing needs to connect to it. The changes currently in this PR support that use-case (and the Azure API allows me to deploy a container with no ports, as long as all public ports are mapped to other containers).

Sorry to overwhelm with detail...the TLDR is, I think it *is* possible to make this change in a backwards compatible way, while still supporting new functionality in a way that makes sense. I can go ahead and modify the PR when I have a bit of time.

Any chance I can get a review once I've made those changes @katbyte?

amasover:
Something else....maybe in the future we want the definitive way to define the first example config from my last post (current provider behavior) like this:

```hcl
resource "azurerm_container_group" "webapp" {
    name                = "webapp-aci"
    #omitted lines of config...
    ports {
        port     = 443
        protocol = "TCP"
    }
    ports {
        port     = 80
        protocol = "TCP"
    }
    container {
        name             = "nginx"
        ports {
            port     = 443
            protocol = "TCP"
        }
    }
    container {
        name             = "webapp"
        ports {
            port     = 80
            protocol = "TCP"
        }
    }
}
```
It's a little more verbose, but it's also maybe a little more clear that those ports are going to be exposed publicly.

I think we still should be able to make non-breaking change that I've outlined in my last post right now. Then at some point in the future, we issue a depreciation warning that ports will become a required field on the container group/resource level.

I don't really have a big preference syntactically. We could also just make the non-breaking change, and never require ports explicitly at the container group/resource level. Personally I just want to be able to deploy my current app that has insecure ports that cannot be exposed publicly. So whatever path is easiest to implement, or easiest for the maintainers is something I'm all for.

jackofallops:
Hi @amasover 

I'm going to close this PR for now, I think a new clean PR may be the best way to approach this issue. The reasoning for this is:
* This has been around long enough (pre-2.0) that there are already merge conflicts due to change
* There are API / SDK changes incoming for this service, which will almost certainly cause further conflicts
* and, most importantly, I think the approach needs to address an incorrect assumption in the original implementation to correctly address the breaking change element.

What I think would be required, as you've correctly covered at least in part above, is:

* There should be a top level set item for the resource, probably `exposed_port` for clarity of its purpose. This should have `Optional`, `ForceNew`, and `Computed` set. 
* This should be set as it is now if not explicitly specified to prevent the breaking change, but can be defined independently if the user requires. 
* The code that deals with this when it is user specified should check that the entries exists in the combined container group ports set. (Sadly this will need to be in the `Create` function, as it's not possible to do at Schema level to catch at plan stage)
* It should be commented with `// change to 'Required' in 3.0 of the provider` so we can find it more easily to make the breaking change at the appropriate time.

If you would prefer to fix the conflicts and continue in this PR I can re-open it later, but I'd recommend updating your master and re-branching from current.

Thanks!

PR #5093, opened 2019-12-06
Update resource azurerm_virtual_machine_scale_set to support Ultra Disk

neil-yechenwei:
This PR is the code implement of the issue https://github.com/terraform-providers/terraform-provider-azurerm/issues/4892


katbyte:
Hi @neil-yechenwei,

Thank you for opening this PR, however we are not making any changes to the existing VM/VMSS as we have re-written then for 2.0 (#2807) which includes this fix. As such and as discussed i am going to close this PR.

PR #5094, opened 2019-12-06
add new resource and data source of azure active directory domain service

njuCZ:
fix terraform-providers/terraform-provider-azuread#94

the reason why I submit in this repo:
1) domainservice is about resourceManager, not graphrbac
2) After creating aadds successfully, we need to set the virtual network dnsserver using domain service controller ip

notes:
1. create domainservice needs `global admin` permission
2. now there can be only one domainservice instance per tenant
3. when creating aadds, two domian service controllers will be deployed. when the first controller succeed and is running, the create api completes, but we still needs to wait for another controller deploying. 
4. because creating aadds needs a virtual network, and after creating success, it needs to set the virtual network dnsserver. So I write this logic in codes, but I wanna is there any better ways to handle matters like this
5. creating aadds in local env costs about 80 mins

there are some knows issues of AADDS, if there is any problem creating, please refer to [know issues](https://docs.microsoft.com/en-us/azure/active-directory-domain-services/troubleshoot-alerts) to solve


katbyte:
Hi @njuCZ,

Thank you for this PR however we are moving all AAD resources out into the azuread provider: https://github.com/terraform-providers/terraform-provider-azuread. Could you please reopen this for that provider? Thanks

As such i am going to close this PR.

PR #5107, opened 2019-12-09
App Service: Adding Import-Time Validation

tombuildsstuff:
This PR adds import-time validation to App Service - by confirming the ID parses correctly during import.

There's also some minor cleanup to remove the use of `resource.SchemaData` from within the helper package, since this shouldn't be present there


tombuildsstuff:
Tests still pass:

Screenshot 2019-12-09 at 21 29 57

tombuildsstuff:
Tests look good 👍

PR #5109, opened 2019-12-09
Storage: switching to use the Authorizers from Azure/go-autorest

tombuildsstuff:
Now that https://github.com/Azure/go-autorest/pull/416 has been merged and released into github.com/Azure/go-autorest we can now switch over to using the authorizers from that repository directly


tombuildsstuff:
Tests still pass:

Screenshot 2019-12-09 at 21 29 57

PR #5116, opened 2019-12-10
azurerm_cosmos_mongo_*: fix throughput & deprecate indexes

katbyte:
continuation of #4620

this also deprecated the indexes property as it seems to be ignored by the API now & `shared_key` has the same affect


azurerm/resource_arm_cosmosdb_mongo_collection.go
```diff
@@ -78,13 +80,14 @@ func resourceArmCosmosDbMongoCollection() *schema.Resource {
...
+				Computed:     true,
 				ValidateFunc: validate.CosmosThroughput,
 			},
 
 			"indexes": {
-				Type:     schema.TypeSet,
-				Optional: true,
+				Type:       schema.TypeSet,
+				Optional:   true,
+				Deprecated: "Indexes are ignored unless they are the shared key so have ben deprecated. ",
```
mbfrahry:
```suggestion
				Deprecated: "Indexes are ignored unless they are the shared key so have been deprecated.",
```

website/docs/r/network_watcher.html.markdown
```diff
@@ -35,7 +35,7 @@ The following arguments are supported:
 
 * `resource_group_name` - (Required) The name of the resource group in which to create the Network Watcher. Changing this forces a new resource to be created.
 
-* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+* `location` - (Required) `Specifies the supported Azure location where the resource exists`. Changing this forces a new resource to be created.
```
mbfrahry: Are these `` intentional?

katbyte: nope! thanks

PR #5120, opened 2019-12-10
Support opencensus tracing

magodo:
Since azure go sdk has opencensus instrumented inside, and azure
provider is based on azure go sdk. We can directly make use of the
tracing provided so as for triage, performance measuring, process
visualization, etc.

This commit implemented follwing exporters:

- zipkin
- jaeger

Other exporters should be able to easily integrate the same way.

The trick here is that I make use of the fact that each CRUD of each
resource will create a context, via a ultimate call to `buildWithTimeout()`.
So I instrument at that function, indicating the current resource name
and the operation name. In order to aggregate all spans in one terraform
command invocation, I started a root span at `main()`, and spawn child
span from that root span at `buildWithTimeout()`.

The usage is like: By default there is no tracing. If user wants to enable tracing, one needs to do following stuff:

- `export TF_AZURE_TRACER=<type>`, where `type` is one of `[jaeger, zipkin]`
- (jaeger only) 
  
  - `export TF_AZURE_TRACE_JAEGER_AGENT_ENDPOINT_URI=<agent_uri>`
    
    E.g. `agent_uri` could be: `localhost:6831`

  - `export TF_AZURE_TRACE_JAEGER_COLLECTION_ENDPOINT_URI=<collection_uri>`

    E.g.  `collection_uri` could be: `http://localhost:14268/api/traces`

- (zipkin only)

  - `export TF_AZURE_TRACE_ZIPKIN_HTTP_REPORTER_URI=<reporter_uri>`

    E.g. `reporter_uri` could be: `http://localhost:19411/api/v2/spans`

  - `export TF_AZURE_TRACE_ZIPKIN_ENDPOINT_URI=<endpoint_uri>`

    E.g. ; `endpoint_uri` could be: `192.168.1.5:5454`


tombuildsstuff:
hi @magodo 

Thanks for this PR.

Whilst the Azure SDK supports tracing unfortunately this isn't something we plan to expose within a Terraform Provider directly; instead we already capture similar performance information via the Test Suite using our CI system today.

It's also worth noting that unfortunately not all resources have a "name" field - as such the tracing involved here wouldn't be holistic in it's current approach - meaning that we'd instead need to pass a unique (set of) identifiers from the resource into this function, which is a substantial amount of work. As such whilst I'd like to thank you for this contribution I'm going to close this Pull Request for the moment.

Thanks!

magodo:
Hi @tombuildsstuff , good to know! Then i'll leave to myself use.
Thank you!

PR #3633, opened 2019-06-11
New resource: azurerm_stream_analytics_reference_input_blob

praneetloke:
I noticed that it was not possible to create inputs for a stream analytics job of type [reference](https://docs.microsoft.com/en-us/azure/stream-analytics/stream-analytics-use-reference-data#azure-blob-storage). The code to represent such a resource already existed, but just wasn't exposed as a resource to be able to use it.

I also added an example, that I tested using my own subscription. Here's a screenshot of how the "reference" input looks.

Screenshot_2019-06-10 Microsoft Azure

PR #5129, opened 2019-12-11
New resource: Add blob reference input as a new resource for stream analytics input resources (rebased)

danielrbradley:
This is a rebased version of #3633 so that I can help investigate the build errors.


katbyte:
Hi @danielrbradley, 

Thanks for opening this PR! looks like the original author has pulled in your changes and as such i am going to close this one.

PR #5139, opened 2019-12-12
subnet: Support address_prefixes for multiple prefixes (ipv6)

### LINKED ISSUE #5140: Subnets with multiple addressPrefixes cannot be used as a source or created
### Community Note

7427e8c593607d30a49fa3d211c38151cf422ce0

### Affected Resource(s)

* `azurerm_subnet`

```hcl
data "azurerm_subnet" "preexisting_worker_subnet" {
  count = var.preexisting_network ? 1 : 0

  resource_group_name  = var.network_resource_group_name
  virtual_network_name = var.virtual_network_name
  name                 = var.worker_subnet
}

locals {
  # empty when subnet has multiple addressPrefixes (2+)
  address_prefix data.azurerm_subnet.preexisting_worker_subnet[0].address_prefix
}
```

### Expected Behavior

* Retrieve address_prefixes when multiple are set on the subnet
* Allow creation of subnets with 2+ address_prefixes (ipv4 and ipv6, for example)

### Actual Behavior

Nil data on retrieval, unable to create with existing API

smarterclayton:
The azure subnet resource supports addressPrefix when one prefix is available and addressPrefixes when two or or more are set (for instance, when configuring an ipv4 and ipv6 dual stack configuration for a network). Update `azurerm_subnet` to allow either `address_prefix` or `address_prefixes` as input, where the latter is a list of string. Check that at least one is set, and mark the older address_prefix as deprecated.

The data source for subnet is modified to always return address_prefixes to simplify retrieval (although this could be changed to always return the first address prefix as address_prefix).

Today, if two prefixes are specified via CLI or Azure UI the data source returns no address_prefix and it is not possible to create a new subnet with two prefixes via Terraform.

Fixes #5140

### DIFF
--- a/azurerm/internal/services/network/data_source_subnet.go
+++ b/azurerm/internal/services/network/data_source_subnet.go
@@ -41,6 +41,12 @@ func dataSourceArmSubnet() *schema.Resource {
 				Computed: true,
 			},
 
+			"address_prefixes": {
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Computed: true,
+			},
+
 			"network_security_group_id": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -102,7 +108,18 @@ func dataSourceArmSubnetRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("virtual_network_name", virtualNetworkName)
 
 	if props := resp.SubnetPropertiesFormat; props != nil {
-		d.Set("address_prefix", props.AddressPrefix)
+		if props.AddressPrefix != nil {
+			d.Set("address_prefix", props.AddressPrefix)
+		}
+		if props.AddressPrefixes == nil {
+			if props.AddressPrefix != nil && len(*props.AddressPrefix) > 0 {
+				d.Set("address_prefixes", []string{*props.AddressPrefix})
+			} else {
+				d.Set("address_prefixes", []string{})
+			}
+		} else {
+			d.Set("address_prefixes", utils.FlattenStringSlice(props.AddressPrefixes))
+		}
 
 		if pe := props.PrivateEndpointNetworkPolicies; pe != nil {
 			d.Set("enforce_private_link_endpoint_network_policies", strings.EqualFold("Disabled", *pe))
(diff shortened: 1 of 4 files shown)


smarterclayton:
Still fixing a few things - it looks like my use of Sets is inappropriate and I'm able to break azure UI at least (if an ipv6 subnet comes before an ipv4 subnet, the azure UI is unable to display information about it).  So it needs to be a List for now.

smarterclayton:
> There won't be a good migration path from here to there unfortunately so it might be best to close this and migrate your changes over to the new format when we're done in a few days. Apologies for the inconvenience we are causing here.

That's fine, we have downstream forks for older versions anyway.

smarterclayton:
> I've been poking around subnet and it seems like you can only specify 2 address prefixes (one ipv4 and one ipv6)

Hrm, I thought you could have multiple ipv6 ranges.  Will test and follow up.

smarterclayton:
Regarding `address_prefixes` I was trying to remain consistent with both the API (where address_prefixes is the preferred field and newer versions of the API return it by default) and the behavior of other similar Azure API singleton -> array transitions I saw in the code base.  Was I mistaken about following the API behavior over the historical behavior?

smarterclayton:
> address_prefixes

You are correct, while the API accepts multiple, the error you get is

```
Multiple address prefixes from same family found in dual stack subnet /subscriptions/xxx/resourceGroups/clayton-ipv6-h6lfz-rg/providers/Microsoft.Network/virtualNetworks/clayton-ipv6-h6lfz-vnet/subnets/clayton-ipv6-h6lfz-worker-subnet. Address prefixes System.Collections.Generic.List`1[System.String] can have only one IPv4 and one IPv6 IP address prefixes.
```

So is the convention to map more closely to the azure experience (for instance, azure cli expects multiple --address-prefixes) or to the implied semantics?  My concern about doing something family specific would be if azure ever changes to allow multiple disjoint subnets within the ipv4 or ipv6 space, in which case the module would need to introduce that new field anyway

katbyte:
> Regarding `address_prefixes` I was trying to remain consistent with both the API (where address_prefixes is the preferred field and newer versions of the API return it by default) and the behavior of other similar Azure API singleton -> array transitions I saw in the code base. Was I mistaken about following the API behavior over the historical behavior?

Basically data sources should always be lists otherwise there is no way to address the values within

fabianofranz:
https://github.com/terraform-providers/terraform-provider-azurerm/pull/6148 is taking over this PR. @smarterclayton this one can be closed, tks!

katbyte:
@smarterclayton,

as @fabianofranz as used this as the base for his PR and you changes are included in it i am going to close this one. Thanks!

PR #5145, opened 2019-12-12
azurerm_virtual_network_gateway_connection  - add "connection_protocol" property

### LINKED ISSUE #5144: Support for connection_protocol in vnet gateway connection
### Community Note

### Description

The provider currently does not support the parameter connection_protocol in the ressource azurerm_virtual_network_gateway_connection. It should be possible to select IKEv1 or IKEv2

* azurerm_virtual_network_gateway_connection

```hcl
resource "azurerm_virtual_network_gateway_connection" "vpnCon" {
    location                           = var.region
    name                               = "vpnc"
    resource_group_name      = azurerm_resource_group.rgVpn.name
    type                                 = "IPSec"
    virtual_network_gateway_id         = azurerm_virtual_network_gateway.vngwVpn.id
    local_network_gateway_id           = azurerm_local_network_gateway.lnetGwVpn.id
    shared_key                         = "...."
    connection_protocol                = "IKEv1"
```

### References

According to documentation this is possible on IPSec connections with SKU VpnGw1 or greater.
https://docs.microsoft.com/en-us/azure/vpn-gateway/vpn-gateway-vpn-faq#ipsecike

* #0000

phires:
This fixes #5144

### DIFF
--- a/azurerm/resource_arm_virtual_network_gateway_connection.go
+++ b/azurerm/resource_arm_virtual_network_gateway_connection.go
@@ -128,11 +128,12 @@ func resourceArmVirtualNetworkGatewayConnection() *schema.Resource {
 			"connection_protocol": {
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 				ForceNew: true,
 				ValidateFunc: validation.StringInSlice([]string{
 					string(network.IKEv1),
 					string(network.IKEv2),
-				}, true),
+				}, false),
 			},
 
 			"ipsec_policy": {
@@ -379,9 +380,7 @@ func resourceArmVirtualNetworkGatewayConnectionRead(d *schema.ResourceData, meta
 		d.Set("shared_key", conn.SharedKey)
 	}
 
-	if string(conn.ConnectionProtocol) != "" {
-		d.Set("connection_protocol", string(conn.ConnectionProtocol))
-	}
+	d.Set("connection_protocol", string(conn.ConnectionProtocol))
 
 	if conn.ExpressRouteGatewayBypass != nil {
 		d.Set("express_route_gateway_bypass", conn.ExpressRouteGatewayBypass)
--- a/azurerm/resource_arm_virtual_network_gateway_connection.go
+++ b/azurerm/resource_arm_virtual_network_gateway_connection.go
@@ -503,7 +503,6 @@ func getArmVirtualNetworkGatewayConnectionProperties(d *schema.ResourceData) (*n
 	if v, ok := d.GetOk("connection_protocol"); ok {
 		connectionProtocol := v.(string)
 		props.ConnectionProtocol = network.VirtualNetworkGatewayConnectionProtocol(connectionProtocol)
-
 	}
 
 	if v, ok := d.GetOk("ipsec_policy"); ok {
--- a/azurerm/resource_arm_virtual_network_gateway_connection.go
+++ b/azurerm/resource_arm_virtual_network_gateway_connection.go
@@ -125,6 +125,16 @@ func resourceArmVirtualNetworkGatewayConnection() *schema.Resource {
 				Computed: true,
 			},
 
+			"connection_protocol": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(network.IKEv1),
+					string(network.IKEv2),
+				}, true),
+			},
+
 			"ipsec_policy": {
 				Type:     schema.TypeList,
 				Optional: true,
@@ -369,6 +379,10 @@ func resourceArmVirtualNetworkGatewayConnectionRead(d *schema.ResourceData, meta
 		d.Set("shared_key", conn.SharedKey)
 	}
 
+	if string(conn.ConnectionProtocol) != "" {
+		d.Set("connection_protocol", string(conn.ConnectionProtocol))
+	}
+
 	if conn.ExpressRouteGatewayBypass != nil {
 		d.Set("express_route_gateway_bypass", conn.ExpressRouteGatewayBypass)
 	}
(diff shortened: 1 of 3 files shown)


azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -125,6 +125,16 @@ func resourceArmVirtualNetworkGatewayConnection() *schema.Resource {
...
 			},
 
+			"connection_protocol": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(network.IKEv1),
+					string(network.IKEv2),
+				}, true),
```
tombuildsstuff: we're making all new fields case-sensitive going forward - so can we make this:

```suggestion
				}, false),
```

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -369,6 +379,10 @@ func resourceArmVirtualNetworkGatewayConnectionRead(d *schema.ResourceData, meta
 		d.Set("shared_key", conn.SharedKey)
 	}
 
+	if string(conn.ConnectionProtocol) != "" {
+		d.Set("connection_protocol", string(conn.ConnectionProtocol))
+	}
```
tombuildsstuff: can we remove this `if` statement here?

website/docs/r/virtual_network_gateway_connection.html.markdown
```diff
@@ -245,6 +245,12 @@ The following arguments are supported:
     Site-to-Site or VNet-to-VNet connection is created whereas ExpressRoute
     connections do not need a shared key.
 
+* `connection_protocol` - (Optional) The IKE protocol version to use. Possible
+    values are `IKEv1` and `IKEv2`. Only valid for IPSec connections on virtual
+    network gateways with SKU `VpnGw1`, `VpnGw2`, `VpnGw3`, `VpnGw1AZ`, `VpnGw2AZ`
+    or `VpnGw3AZ`. Changing the protocol will force a new connection to be created. 
+    Defaults to `IKEv2`.
```
tombuildsstuff: There's no default set above? As such should we set this to:

```suggestion
```

website/docs/r/virtual_network_gateway_connection.html.markdown
```diff
@@ -245,6 +245,12 @@ The following arguments are supported:
     Site-to-Site or VNet-to-VNet connection is created whereas ExpressRoute
     connections do not need a shared key.
 
+* `connection_protocol` - (Optional) The IKE protocol version to use. Possible
+    values are `IKEv1` and `IKEv2`. Only valid for IPSec connections on virtual
+    network gateways with SKU `VpnGw1`, `VpnGw2`, `VpnGw3`, `VpnGw1AZ`, `VpnGw2AZ`
+    or `VpnGw3AZ`. Changing the protocol will force a new connection to be created.
```
tombuildsstuff: could we split this line:

```
Only valid for IPSec connections on virtual network gateways with SKU `VpnGw1`, `VpnGw2`, `VpnGw3`, `VpnGw1AZ`, `VpnGw2AZ` or `VpnGw3AZ`.
```

into a blue info box below the description? We can do this using:

```
-> Only valid for IPSec connections on Virtual Network Gateways using the SKU `VpnGw1`, `VpnGw2`, `VpnGw3`, `VpnGw1AZ`, `VpnGw2AZ` or `VpnGw3AZ`.
```

in addition could we use the text:

```suggestion
Changing this value forces a new resource to be created.
```

to match the rest of the documentation?

azurerm/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -125,6 +125,16 @@ func resourceArmVirtualNetworkGatewayConnection() *schema.Resource {
 				Computed: true,
 			},
 
+			"connection_protocol": {
```
tombuildsstuff: could we add an acceptance test covering this new field?

azurerm/resource_arm_virtual_network_gateway_connection.go
phires: I was thinking about that, but that API documentation does want to have either IKEv1 or IKEv2 as a value. Also wouldn't that be a breaking change? Currently the value is never set. If we suddenly start to set it, it would cause all connections to be recreated, as the type can't be changed without recreation.

website/docs/r/virtual_network_gateway_connection.html.markdown
phires: It is implicit set to IKEv2 by the API if the value is emtpy. Hence the `if` check above

azurerm/resource_arm_virtual_network_gateway_connection.go
phires: Wouldn't this break during conversion to `network.VirtualNetworkGatewayConnectionProtocol` on line 505?

azurerm/resource_arm_virtual_network_gateway_connection.go
phires: will do

website/docs/r/virtual_network_gateway_connection.html.markdown
phires: Also I wouldn't want to introduce it as a default value on the provider, as it would cause all existing connections to be recreated. But as the default if not set is IKEv2 I have no better idea on how to document this...

azurerm/resource_arm_virtual_network_gateway_connection.go
phires: added:
`$ make testacc TESTARGS='-run=TestAccAzureRMVirtualNetworkGatewayConnection_connectionprotocol'`
looong time after...
`--- PASS: TestAccAzureRMVirtualNetworkGatewayConnection_connectionprotocol (2512.56s)
PASS`

azurerm/resource_arm_virtual_network_gateway_connection.go
katbyte: It shouldn't as we are validating the strings against the correct case

azurerm/resource_arm_virtual_network_gateway_connection.go
katbyte: It might be best to mark it as computed in that case?

tombuildsstuff:
Tests pass:

Screenshot 2019-12-18 at 10 53 43

PR #5158, opened 2019-12-13
Add note about SLB req for availability_zones

rudolphjacksonm:
Adds note to `availability_zones` to make it clearer that Standard Load Balancer SKU is required to enable Availability Zones.

### DIFF
--- a/website/docs/r/kubernetes_cluster.html.markdown
+++ b/website/docs/r/kubernetes_cluster.html.markdown
@@ -168,6 +168,8 @@ A `agent_pool_profile` block supports the following:
 
 * `availability_zones` - (Optional) Availability zones for nodes. The property `type` of the `agent_pool_profile` must be set to `VirtualMachineScaleSets` in order to use availability zones.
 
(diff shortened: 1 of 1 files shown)


website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -168,6 +168,8 @@ A `agent_pool_profile` block supports the following:
 
 * `availability_zones` - (Optional) Availability zones for nodes. The property `type` of the `agent_pool_profile` must be set to `VirtualMachineScaleSets` in order to use availability zones.
 
+-> **NOTE:** If you have provided the `load_balancer_sku` argument with the value `Basic`, you will not be able use `availability_zones`. You must use the `Standard` SKU.
```
tombuildsstuff: we can probably flip this around to:

```suggestion
-> **NOTE:** To configure Availability Zones the `load_balancer_sku` must be set to `Standard`
```

WDYT?

tombuildsstuff:
Ignoring a couple of known test failures, the tests pass:

Screenshot 2019-12-16 at 07 19 14

Issue #4506, opened 2019-10-04
Managed Identities for AKS


fxgsell:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

AKS released support for managed identity in preview, it can be used with the cli by adding the flag `--enable-managed-identity`. (preview is public but the subscription must be opted-in the preview)

In the preview period, a service principal is still required but eventually this requirement in AKS will be removed completely.

This will also required new exported attributes (in an identity block):

* principal_id - The Principal ID for the Service Principal associated with the Managed Service Identity.
* tenant_id - The Tenant ID for the Service Principal associated with the Managed Service Identity.

### New or Affected Resource(s)

* azurerm_kubernetes_cluster

### Potential Terraform Configuration
2 options:

Either a simple flag:
```hcl
resource "azurerm_kubernetes_cluster" "test" {
  enable_managed_identity = true
  ...
}
```

Or a new identity block: (Example here is for Managed identity this should also support specifying a SP like other resources using identities)
```hcl
resource "azurerm_kubernetes_cluster" "test" {
  ...
  identity {
    type = "SystemAssigned"
  }
}
```

### References

* https://docs.microsoft.com/en-us/azure/aks/use-managed-identity

sauryadas:
at GA the SA managed identity will be created by default, no explicit flag will be required

fxgsell:
Ok, we will still need to export the new identity information and remove/make optional the existing fields.

mbrancato:
While the service identity may be created by default, there is still a schema change required for this resource as the `service_principal` block is required today. Also, explicit SP assignment is still supported as I understand it, so making this block optional seems good. It might also be good to align the name of this block with other resources supporting MSI like `azurerm_app_service` and `azurerm_virtual_machine`.

shubham124:
Can someone share a link which contains aks creation through terraform using user managed identity.
It will  be really helpful as I am facing so many challenges and also above suggestion was not working.

jkdey:
I get the following when attempting to set a managed_cluster_identity block:

```
validate: error: failed to validate Terraform configuration in .

[...]

PR #5168, opened 2019-12-14
`azurerm_kubernetes_cluster` - add `managed_cluster_identity` support

mbfrahry:
Address #4506


website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -281,6 +283,12 @@ A `linux_profile` block supports the following:
 
 ---
 
+A `mangaged_cluster_identity` block supports the following:
```
tracypholmes:
```suggestion
A `managed_cluster_identity` block supports the following:
```

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -792,6 +824,13 @@ func resourceArmKubernetesClusterUpdate(d *schema.ResourceData, meta interface{}
 		existing.ManagedClusterProperties.WindowsProfile = windowsProfile
 	}
 
+	if d.HasChange("managed_cluster_identity") {
+		updateCluster = true
+		managedClusterIdentityRaw := d.Get("managed_cluster_identity").([]interface{})
+		managedClusterIdentity := expandKubernetesClusterManagedClusterIdentity(managedClusterIdentityRaw)
+		existing.Identity = managedClusterIdentity
```
katbyte: minor but this could be
```
		existing.Identity = expandKubernetesClusterManagedClusterIdentity(d.Get("managed_cluster_identity")) 
```

if the expand sig is changed

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -1403,6 +1446,20 @@ func expandKubernetesClusterRoleBasedAccessControl(input []interface{}, provider
...
+		return nil
+	}
+
+	val := input[0].(map[string]interface{})
+
+	managedClusterIdentity := &containerservice.ManagedClusterIdentity{
+		Type: containerservice.ResourceIdentityType(val["type"].(string)),
+	}
+
+	return managedClusterIdentity
```
katbyte: This could also be 
```
return &containerservice.ManagedClusterIdentity{
		Type: containerservice.ResourceIdentityType(val["type"].(string)),
	}
```

as there is no value is assigning it to a var

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -1403,6 +1446,20 @@ func expandKubernetesClusterRoleBasedAccessControl(input []interface{}, provider
 	return rbacEnabled, aad
 }
 
+func expandKubernetesClusterManagedClusterIdentity(input []interface{}) *containerservice.ManagedClusterIdentity {
+	if len(input) == 0 || input[0] == nil {
+		return nil
+	}
+
+	val := input[0].(map[string]interface{})
```
katbyte: would `values` make more sense?
```suggestion
	values := input[0].(map[string]interface{})
```

azurerm/resource_arm_kubernetes_cluster.go
```diff
@@ -1529,3 +1586,20 @@ func flattenKubernetesClusterKubeConfigAAD(config kubernetes.KubeConfigAAD) []in
...
 	}
 }
+
+func flattenKubernetesClusterManagedClusterIdentity(input *containerservice.ManagedClusterIdentity) []interface{} {
+	if input == nil {
+		return []interface{}{}
+	}
+
+	identity := make(map[string]interface{})
+	if input.PrincipalID != nil {
```
katbyte: I think with .12 we need to ensure these are set now?
```suggestion
	input.PrincipalID = ""
	if input.PrincipalID != nil {
```

PR #5184, opened 2019-12-16
refactor: moving the config builder / introducing the `acceptance` package

tombuildsstuff:
This PR introduces the new `acceptance` package which will replace the existing test infrastructure; the general intention is to reduce boilerplate where possible. In order to demonstrate this I've updated the tests for the Resource Group resource to use this new pattern - which I believe should suffice for now.

In addition this PR moves most of the logic of `config.go` out of the main package - which should enable the final steps required to move Data Sources & Resources.

Finally, this also introduces a feature toggle for Dynamic Test Locations; which allows us to dynamically distribute our test load across Azure Regions, to better distribute this across Azure; however at this time this is feature toggled off by default.

Tests pass:

```
$ acctests azurerm TestAccAzureRMResourceGroup_
=== PAUSE TestAccAzureRMResourceGroup_basic
=== RUN   TestAccAzureRMResourceGroup_requiresImport
--- SKIP: TestAccAzureRMResourceGroup_requiresImport (0.00s)
    resource_arm_resource_group_test.go:35: Skipping since resources aren't required to be imported
=== RUN   TestAccAzureRMResourceGroup_disappears
=== PAUSE TestAccAzureRMResourceGroup_disappears
=== PAUSE TestAccAzureRMResourceGroup_withTags
=== CONT  TestAccAzureRMResourceGroup_basic
--- PASS: TestAccAzureRMResourceGroup_basic (78.66s)
=== CONT  TestAccAzureRMResourceGroup_withTags
--- PASS: TestAccAzureRMResourceGroup_withTags (99.51s)
=== CONT  TestAccAzureRMResourceGroup_disappears
--- PASS: TestAccAzureRMResourceGroup_disappears (67.29s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	245.525s
```

### DIFF
--- a/azurerm/internal/acceptance/data.go
+++ b/azurerm/internal/acceptance/data.go
@@ -11,8 +11,8 @@ import (
 )
 
 type TestData struct {
-	// Location is a set of Azure Regions which should be used for this Test
-	Location Regions
+	// Locations is a set of Azure Regions which should be used for this Test
+	Locations Regions
 
 	// RandomString is a random integer which is unique to this test case
 	RandomInteger int
@@ -58,9 +58,9 @@ func BuildTestData(t *testing.T, resourceType string, resourceLabel string) Test
 	}
 
 	if features.UseDynamicTestLocations() {
-		testData.Location = SupportedLocations()
+		testData.Locations = availableLocations()
 	} else {
-		testData.Location = Regions{
+		testData.Locations = Regions{
 			Primary:   Location(),
 			Secondary: AltLocation(),
 			Ternary:   AltLocation2(),
--- a/azurerm/internal/acceptance/data.go
+++ b/azurerm/internal/acceptance/data.go
@@ -1,0 +1,71 @@
+package acceptance
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
+	"github.com/terraform-providers/terraform-provider-azuread/azuread/helpers/tf"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
+)
+
+type TestData struct {
+	// Location is a set of Azure Regions which should be used for this Test
+	Location Regions
+
+	// RandomString is a random integer which is unique to this test case
+	RandomInteger int
+
+	// RandomString is a random 5 character string is unique to this test case
+	RandomString string
+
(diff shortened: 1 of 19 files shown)


azurerm/internal/acceptance/data.go
```diff
@@ -0,0 +1,71 @@
...
+
+type TestData struct {
+	// Location is a set of Azure Regions which should be used for this Test
+	Location Regions
+
+	// RandomString is a random integer which is unique to this test case
+	RandomInteger int
+
+	// RandomString is a random 5 character string is unique to this test case
+	RandomString string
```
katbyte: Where is this used? i've been replacing it where possible by doing a % on the rint

azurerm/internal/acceptance/data.go
```diff
@@ -0,0 +1,71 @@
...
+		ResourceName:    fmt.Sprintf("%s.%s", resourceType, resourceLabel),
+		Environment:     *env,
+		EnvironmentName: EnvironmentName(),
+
+		resourceType:  resourceType,
+		resourceLabel: resourceLabel,
+	}
+
+	if features.UseDynamicTestLocations() {
+		testData.Location = SupportedLocations()
```
katbyte: how will this handel tests that have to run in a specific location? or a location we don't have in any env var?

Might it make sense to pass in from a test "supported locations" and then if any of the ENV vars match its used? 

It seems `SupportedLocations` might be better phrased as `DesiredLocations` as support indicates it'll work

azurerm/internal/acceptance/data.go
tombuildsstuff: Storage/anything Windows related (Redis, VM's etc) - using a combination of letters/numbers gives us a higher chance of uniqueness when we're only allowed 5 characters

azurerm/internal/acceptance/data.go
tombuildsstuff: for locations we don't have in an environment var (e.g. this is only available in this 1 region) we can continue hard-coding them as we do today.

we're only hard-coding to a single location in the tests AFAIK - so I think we can skip passing in `supported locations` for now, but if we need it that makes sense to me 👍 

in retrospect this can become a private function, so I guess we could split the difference and make this `availableLocations` since they're what the user running the test is saying is available to be used?

tombuildsstuff:
Tests pass:

Screenshot 2019-12-17 at 11 25 31

PR #5191, opened 2019-12-17
refactor: moving the top-level Client into Packages

tombuildsstuff:
This PR does several interdependent things which unfortunately makes it both a) quite large and b) **a breaking change for all open PR's**. Whilst this is unfortunate - we're now at a point where there's not much we can to do work around this - but it's probably easier to review this commit-by-commit.

Firstly this removes `ArmClient` and replaces it with `clients.Client` - which has been an inner variable within the struct for some time; so there shouldn't be many code changes here directly.

Next this then moves the Azure Provider init into `./azurerm/internal/provider` - however until all Data Sources & Resources have been moved from the root this needs to take a list of existing Data Sources & Resources.

Finally, this switches the use of `testAccProvider` and `testAccProviders` for `acceptance.AzureProvider` and `acceptance.SupportedProviders` respectively.

---

Once this has been merged we'll need to do one more large PR moving all Data Sources & Resources into Packages - unfortunately until this is fully complete we may be unable to run the acceptance tests (needs confirming) - so this may need to be done in a single PR based on this one.. 😞.

We'll also need to update the Tests, Makefile & CI to ensure tests are run using the new structure, rather than assuming everything's present in the root - but again that's one for the follow up PR.


tombuildsstuff:
Tests pass:

```
$ acctests azurerm TestAccAzureRMResourceGroup_basic
=== PAUSE TestAccAzureRMResourceGroup_basic
=== CONT  TestAccAzureRMResourceGroup_basic
--- PASS: TestAccAzureRMResourceGroup_basic (82.69s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	82.739s
```

tombuildsstuff:
Rebased on top of master and the resource group tests still pass:

```
$ acctests azurerm TestAccAzureRMResourceGroup_basic
=== PAUSE TestAccAzureRMResourceGroup_basic
=== CONT  TestAccAzureRMResourceGroup_basic
--- PASS: TestAccAzureRMResourceGroup_basic (81.95s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	81.994s
```

PR #5210, opened 2019-12-18
r/storage_account: handling when network_rules is empty

tombuildsstuff:
This PR supersedes the work done by @samart in #4966 - by instead confirming that the top-level `props.NetworkRuleSet` object is nil (and returning a default) rather than if the nested elements inside of it are nil

I've also added a couple of nil-checks to prevent some crashes that I spotted whilst passing through


tombuildsstuff:
Tests pass:

Screenshot 2019-12-18 at 17 33 06

PR #5211, opened 2019-12-18
Fix reference format with output resource

Willis0826:


tombuildsstuff:
hey @Willis0826 

Thanks for this PR :)

As [announced in this blog post we're deprecating support for Terraform 0.11](https://www.hashicorp.com/blog/deprecating-terraform-0-11-support-in-terraform-providers/) in the not-too-distant-future and as such we're gradually replacing usage examples with syntax from Terraform 0.11 to use the syntax from Terraform 0.12 instead.

Taking a look through here it appears that this switches back to using the Terraform 0.11 syntax - as such whilst I'd like to thank you for this contribution since we're deprecating this I'm going to close this PR for the moment, as we've intentionally removed Terraform 0.11 syntax from this example. We'd also recommend upgrading to Terraform 0.12 - since as mentioned above support for Terraform 0.11 is being deprecated in the not-too-distant future.

Thanks!

PR #3244, opened 2019-04-15
New Resource: `azurerm_data_factory_pipeline`

mbfrahry:
```
--- PASS: TestAccAzureRMDataFactoryPipeline_basic (78.39s)
--- PASS: TestAccAzureRMDataFactoryPipeline_update (97.29s)
```

PR #5226, opened 2019-12-20
Move resource files into service packages

katbyte:


neil-yechenwei:
@tombuildsstuff , @mbfrahry , I noticed you made a PR (#3244) to add support for azurerm_data_factory_pipeline. I found you added a property "activity" in your first commit (06bd322). But you deleted this property in later commit (6822cfa). After learned from the azure doc for data factory (https://docs.microsoft.com/en-us/azure/data-factory/v1/data-factory-copy-activity-tutorial-using-powershell), I think you should retain this property "activity" since I think it's an important thing for data factory pipeline to combine the InputDataset and the OutputDataset. May I know why you removed the property "activity"? If this property is necessary, can it be added back?

PR #5236, opened 2019-12-23
F/azure sdk for go updatev37.2

xuzhang3:
* update dependency azure-sdk-for-go from v36.2->v37.2
* refresh vendor files


katbyte:
Hi @xuzhang3,

Thank you for this PR! However we are going to upgrade directly to v38 once 1.40 is release so I have opened #5335 to that effect. As such i am going to close this PR.

PR #5240, opened 2019-12-23
provider - fix travis OOM errors

katbyte:


katbyte:
It seems like it is the unused linter that is causing the OOM error. I have removed it from now and opened an issue on [golang-ci](https://github.com/golangci/golangci-lint/issues/898)

katbyte:
(ready to be reviewed)

PR #5249, opened 2019-12-25
Add new resource & data source: azurerm_disk_encryption_set

ArcturusZhang:
This is part of update of the [SSE-CMK feature](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disk-encryption), adding a new resource and a data source: `azurerm_disk_encryption_set`.

This feature requires Keyvault to enable `enable_soft_delete` and `enable_purge_protection`, which are both not implemented in terraform. And the creation of `azurerm_disk_encryption_set` depends on those two switches, otherwise the service will return errors complaining about this.
Therefore I did some workaround in the test of `azurerm_disk_encryption_set`. I will change the workaround back when these two features are implemented in `azurerm_key_vault`.

Currently the fields that could accept user input in a DiskEncryptionSet is all required, therefore I do not put a complete test, since all the fields are covered in the basic test.

### DIFF
--- a/azurerm/internal/services/compute/data_source_disk_encryption_set.go
+++ b/azurerm/internal/services/compute/data_source_disk_encryption_set.go
@@ -61,5 +61,5 @@ func dataSourceArmDiskEncryptionSetRead(d *schema.ResourceData, meta interface{}
 		d.Set("location", azure.NormalizeLocation(*location))
 	}
 
-	return nil
+	return tags.FlattenAndSet(d, resp.Tags)
 }
--- a/azurerm/internal/services/compute/data_source_disk_encryption_set.go
+++ b/azurerm/internal/services/compute/data_source_disk_encryption_set.go
@@ -1,0 +1,137 @@
+package compute
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmDiskEncryptionSet() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmDiskEncryptionSetRead,
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+
+			"location": azure.SchemaLocationForDataSource(),
+
+			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
+
+			"active_key": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"key_url": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"source_vault_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+
+			"identity": {
+				Type:     schema.TypeList,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"principal_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"tenant_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+
+			"previous_keys": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"key_url": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"source_vault_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+
+			"tags": tags.SchemaDataSource(),
+		},
+	}
+}
+
+func dataSourceArmDiskEncryptionSetRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Compute.DiskEncryptionSetsClient
+	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	name := d.Get("name").(string)
+	resourceGroup := d.Get("resource_group_name").(string)
+
+	resp, err := client.Get(ctx, resourceGroup, name)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return fmt.Errorf("Error: Disk Encryption Set %q (Resource Group %q) was not found", name, resourceGroup)
+		}
+		return fmt.Errorf("Error reading Disk Encryption Set %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	d.SetId(*resp.ID)
+
+	d.Set("name", name)
+	d.Set("resource_group_name", resourceGroup)
+	if location := resp.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	if encryptionSetProperties := resp.EncryptionSetProperties; encryptionSetProperties != nil {
+		if err := d.Set("active_key", flattenArmDiskEncryptionSetKeyVaultAndKeyReference(encryptionSetProperties.ActiveKey)); err != nil {
+			return fmt.Errorf("Error setting `active_key`: %+v", err)
+		}
+		if err := d.Set("previous_keys", flattenArmDiskEncryptionSetKeyVaultAndKeyReferenceArray(encryptionSetProperties.PreviousKeys)); err != nil {
+			return fmt.Errorf("Error setting `previous_keys`: %+v", err)
+		}
+	}
+	if identity := resp.Identity; identity != nil {
+		if err := d.Set("identity", flattenArmDiskEncryptionSetIdentity(identity)); err != nil {
+			return fmt.Errorf("Error setting `identity`: %+v", err)
+		}
+	}
+
+	return nil
+}
--- a/azurerm/internal/services/compute/data_source_disk_encryption_set.go
+++ b/azurerm/internal/services/compute/data_source_disk_encryption_set.go
@@ -32,62 +32,6 @@ func dataSourceArmDiskEncryptionSet() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
 
-			"active_key": {
-				Type:     schema.TypeList,
-				Computed: true,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"key_url": {
-							Type:     schema.TypeString,
-							Computed: true,
-						},
-						"source_vault_id": {
-							Type:     schema.TypeString,
-							Computed: true,
-						},
-					},
-				},
-			},
-
-			"identity": {
-				Type:     schema.TypeList,
-				Computed: true,
-				MaxItems: 1,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"type": {
-							Type:     schema.TypeString,
-							Computed: true,
-						},
-						"principal_id": {
-							Type:     schema.TypeString,
-							Computed: true,
-						},
-						"tenant_id": {
-							Type:     schema.TypeString,
-							Computed: true,
-						},
-					},
-				},
-			},
-
-			"previous_keys": {
-				Type:     schema.TypeList,
-				Computed: true,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"key_url": {
-							Type:     schema.TypeString,
-							Computed: true,
-						},
-						"source_vault_id": {
-							Type:     schema.TypeString,
-							Computed: true,
-						},
-					},
-				},
-			},
-
 			"tags": tags.SchemaDataSource(),
 		},
 	}
@@ -115,19 +59,6 @@ func dataSourceArmDiskEncryptionSetRead(d *schema.ResourceData, meta interface{}
 	d.Set("resource_group_name", resourceGroup)
 	if location := resp.Location; location != nil {
 		d.Set("location", azure.NormalizeLocation(*location))
-	}
-	if encryptionSetProperties := resp.EncryptionSetProperties; encryptionSetProperties != nil {
-		if err := d.Set("active_key", flattenArmDiskEncryptionSetKeyVaultAndKeyReference(encryptionSetProperties.ActiveKey)); err != nil {
-			return fmt.Errorf("Error setting `active_key`: %+v", err)
-		}
-		if err := d.Set("previous_keys", flattenArmDiskEncryptionSetKeyVaultAndKeyReferenceArray(encryptionSetProperties.PreviousKeys)); err != nil {
-			return fmt.Errorf("Error setting `previous_keys`: %+v", err)
-		}
-	}
-	if identity := resp.Identity; identity != nil {
-		if err := d.Set("identity", flattenArmDiskEncryptionSetIdentity(identity)); err != nil {
-			return fmt.Errorf("Error setting `identity`: %+v", err)
-		}
 	}
 
 	return nil
--- a/azurerm/internal/services/compute/data_source_disk_encryption_set.go
+++ b/azurerm/internal/services/compute/data_source_disk_encryption_set.go
@@ -18,10 +18,7 @@ func dataSourceArmDiskEncryptionSet() *schema.Resource {
 		Read: dataSourceArmDiskEncryptionSetRead,
 
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(30 * time.Minute),
-			Read:   schema.DefaultTimeout(5 * time.Minute),
-			Update: schema.DefaultTimeout(30 * time.Minute),
-			Delete: schema.DefaultTimeout(30 * time.Minute),
+			Read: schema.DefaultTimeout(5 * time.Minute),
 		},
 
 		Schema: map[string]*schema.Schema{
--- a/azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
+++ b/azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
@@ -276,14 +276,20 @@ func flattenArmDiskEncryptionSetIdentity(input *compute.EncryptionSetIdentity) [
 		return make([]interface{}, 0)
 	}
 
-	result := make(map[string]interface{})
-
-	result["type"] = string(input.Type)
-	if principalId := input.PrincipalID; principalId != nil {
-		result["principal_id"] = *principalId
+	t := string(input.Type)
+	principalId := ""
+	if input.PrincipalID != nil {
+		principalId = *input.PrincipalID
 	}
-	if tenantId := input.TenantID; tenantId != nil {
-		result["tenant_id"] = *tenantId
+	tenantId := ""
+	if input.TenantID != nil {
+		tenantId = *input.TenantID
 	}
-	return []interface{}{result}
+	return []interface{}{
+		map[string]interface{} {
+			"type": t,
+			"principal_id": principalId,
+			"tenant_id": tenantId,
+		},
+	}
 }
--- a/azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
+++ b/azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
@@ -6,8 +6,8 @@ import (
 	"log"
 	"time"
 
-	"github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2018-02-14/keyvault"
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2018-02-14/keyvault"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
--- a/azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
+++ b/azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
@@ -12,7 +12,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
-	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
@@ -50,27 +49,28 @@ func resourceArmDiskEncryptionSet() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupName(),
 
-			"key_vault_key_uri": {
+			"key_vault_key_id": {
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     true,
-				ValidateFunc: validate.NoEmptyStrings,
+				ValidateFunc: azure.ValidateKeyVaultChildId,
 			},
 
 			"identity": {
-				Type:     schema.TypeList,
-				Optional: true,
-				Computed: true,
+				Type: schema.TypeList,
+				// whilst the API Documentation shows optional - attempting to send nothing returns:
+				// `Required parameter 'ResourceIdentity' is missing (null)`
+				// hence this is required
+				Required: true,
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"type": {
 							Type:     schema.TypeString,
-							Optional: true,
+							Required: true,
 							ValidateFunc: validation.StringInSlice([]string{
 								string(compute.SystemAssigned),
 							}, false),
-							Default: string(compute.SystemAssigned),
 						},
 						"principal_id": {
 							Type:     schema.TypeString,
@@ -91,6 +91,7 @@ func resourceArmDiskEncryptionSet() *schema.Resource {
 
 func resourceArmDiskEncryptionSetCreateUpdate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*clients.Client).Compute.DiskEncryptionSetsClient
+	vaultClient := meta.(*clients.Client).KeyVault.VaultsClient
 	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
 	defer cancel()
 
@@ -109,45 +110,37 @@ func resourceArmDiskEncryptionSetCreateUpdate(d *schema.ResourceData, meta inter
 		}
 	}
 
+	keyVaultKeyId := d.Get("key_vault_key_id").(string)
+	keyVaultDetails, err := diskEncryptionSetRetrieveKeyVault(ctx, vaultClient, keyVaultKeyId)
+	if err != nil {
+		return fmt.Errorf("Error validating Key Vault %q (Resource Group %q) for Disk Encryption Set: %+v", err)
+	}
+	if !keyVaultDetails.softDeleteEnabled {
+		return fmt.Errorf("Error validating Key Vault %q (Resource Group %q) for Disk Encryption Set: Soft Delete must be enabled but it isn't!", err)
+	}
+	if !keyVaultDetails.purgeProtectionEnabled {
+		return fmt.Errorf("Error validating Key Vault %q (Resource Group %q) for Disk Encryption Set: Purge Protection must be enabled but it isn't!", err)
+	}
+
 	location := azure.NormalizeLocation(d.Get("location").(string))
-	//activeKey := d.Get("active_key").([]interface{})
+	identityRaw := d.Get("identity").([]interface{})
 	t := d.Get("tags").(map[string]interface{})
 
-	keyURL := d.Get("key_vault_key_uri").(string)
-	vaultClient := meta.(*clients.Client).KeyVault.VaultsClient
-	keyID, err := azure.ParseKeyVaultChildID(keyURL)
-	if err != nil {
-		return fmt.Errorf("Error creating Disk Encryption Set %q (Resource Group %q): %+v", name, resourceGroup, err)
-	}
-	keyVaultID, err := azure.GetKeyVaultIDFromBaseUrl(ctx, vaultClient, keyID.KeyVaultBaseUrl)
-	if err != nil {
-		return fmt.Errorf("Error retrieving the Resource ID the Key Vault at URL %q: %s", keyID.KeyVaultBaseUrl, err)
-	}
-	if keyVaultID == nil {
-		return fmt.Errorf("Error creating Disk Encryption Set %q (Resource Group %q): Unable to determine the Resource ID for the Key Vault at URL %q", name, resourceGroup, keyID.KeyVaultBaseUrl)
-	}
-
-	diskEncryptionSet := compute.DiskEncryptionSet{
+	params := compute.DiskEncryptionSet{
 		Location: utils.String(location),
 		EncryptionSetProperties: &compute.EncryptionSetProperties{
 			ActiveKey: &compute.KeyVaultAndKeyReference{
-				KeyURL: &keyURL,
+				KeyURL: utils.String(keyVaultKeyId),
 				SourceVault: &compute.SourceVault{
-					ID: keyVaultID,
+					ID: utils.String(keyVaultDetails.keyVaultId),
 				},
 			},
 		},
-		Tags: tags.Expand(t),
+		Identity: expandArmDiskEncryptionSetIdentity(identityRaw),
+		Tags:     tags.Expand(t),
 	}
 
-	diskEncryptionSet.Identity = expandArmDiskEncryptionSetIdentity(d)
-
-	// validate whether the keyvault has soft-delete and purge-protection enabled
-	if err := validateKeyVault(ctx, meta.(*clients.Client).KeyVault.VaultsClient, resourceGroup, *keyVaultID); err != nil {
-		return fmt.Errorf("Error creating Disk Encryption Set %q (Resource Group %q): %+v", name, resourceGroup, err)
-	}
-
-	future, err := client.CreateOrUpdate(ctx, resourceGroup, name, diskEncryptionSet)
+	future, err := client.CreateOrUpdate(ctx, resourceGroup, name, params)
 	if err != nil {
 		return fmt.Errorf("Error creating Disk Encryption Set %q (Resource Group %q): %+v", name, resourceGroup, err)
 	}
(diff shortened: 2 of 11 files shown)


azurerm/internal/services/compute/data_source_disk_encryption_set.go
```diff
@@ -0,0 +1,134 @@
...
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"source_vault_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
```
tombuildsstuff: this doesn't make sense to expose in the data source - can we remove this?

azurerm/internal/services/compute/data_source_disk_encryption_set.go
```diff
@@ -0,0 +1,134 @@
...
+						"principal_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"tenant_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
```
tombuildsstuff: this doesn't make sense to expose in the data source - can we remove this?

azurerm/internal/services/compute/data_source_disk_encryption_set.go
```diff
@@ -0,0 +1,134 @@
...
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"source_vault_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
```
tombuildsstuff: this doesn't make sense to expose in the data source - can we remove this?

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,387 @@
...
+				Type:     schema.TypeList,
+				Required: true,
+				// the forceNew is enabled because currently key rotation is not supported, you cannot change the key vault and key associated with disk encryption set.
+				ForceNew: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"key_url": {
+							Type:         schema.TypeString,
+							Required:     true,
```
tombuildsstuff: (seeing the note above) then this also wants to be ForceNew

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,387 @@
...
+							ValidateFunc: validate.NoEmptyStrings,
+						},
+						"source_vault_id": {
+							Type:         schema.TypeString,
+							Required:     true,
```
tombuildsstuff: (seeing the note above) then this also wants to be ForceNew

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,387 @@
...
+							ValidateFunc: validate.NoEmptyStrings,
+						},
+					},
+				},
+			},
```
tombuildsstuff: we can remove this, since it's not something users are likely to use in Terraform?

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,387 @@
...
+		Tags: tags.Expand(t),
+	}
+
+	if v, ok := d.GetOk("identity"); ok {
+		diskEncryptionSet.Identity = expandArmDiskEncryptionSetIdentity(v.([]interface{}))
+	} else {
+		diskEncryptionSet.Identity = &compute.EncryptionSetIdentity{
+			Type: compute.SystemAssigned,
+		}
+	}
```
tombuildsstuff: can we move this logic inside the expand function?

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,387 @@
...
+	if resp.ID == nil {
+		return fmt.Errorf("Cannot read Disk Encryption Set %q (Resource Group %q) ID", name, resourceGroup)
+	}
+	d.SetId(*resp.ID)
+
+	return resourceArmDiskEncryptionSetRead(d, meta)
+}
+
+func validateKeyVaultAndKey(ctx context.Context, meta interface{}, resourceGroup string, keyVaultAndKey *compute.KeyVaultAndKeyReference) error {
```
tombuildsstuff: rather than passing in meta here can you pass in the specific clients needed here?

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,387 @@
...
+		return err
+	}
+	resourceGroup := id.ResourceGroup
+	name := id.Path["diskEncryptionSets"]
+
+	future, err := client.Delete(ctx, resourceGroup, name)
+	if err != nil {
+		if response.WasNotFound(future.Response()) {
+			return nil
+		}
```
tombuildsstuff: we should be able to remove this, since this'll get caught by the Read

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,387 @@
...
+		return fmt.Errorf("Error deleting Disk Encryption Set %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		if !response.WasNotFound(future.Response()) {
```
tombuildsstuff: do we need this, this looks like a bug in the Swagger/API?

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,387 @@
...
+	if keyUrl := input.KeyURL; keyUrl != nil {
+		result["key_url"] = *keyUrl
+	}
+	if sourceVault := input.SourceVault; sourceVault != nil {
+		if sourceVaultId := sourceVault.ID; sourceVaultId != nil {
+			result["source_vault_id"] = *sourceVaultId
+		}
+	}
+
+	return []interface{}{result}
```
tombuildsstuff: both of these fields need to always have a value set:

```suggestion
	keyUrl := ""
	if input.KeyURL != nil {
		keyUrl = *input.KeyURL
	}
	
	sourceVaultId := ""
	if input.SourceVault != nil && input.SourceVault.ID != nil {
		sourceVaultId = *input.SourceVault.ID
	}

	return []interface{}{
		map[string]interface{}{
			"key_url": keyUrl,
			"source_vault_id": sourceVaultId,
		},
	}
```

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,387 @@
...
+	for _, item := range *input {
+		v := make(map[string]interface{})
+
+		if sourceVault := item.SourceVault; sourceVault != nil {
+			if sourceVaultId := sourceVault.ID; sourceVaultId != nil {
+				v["source_vault_id"] = *sourceVaultId
+			}
+		}
+
+		results = append(results, v)
```
tombuildsstuff: (same here) can we update this as above?

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,387 @@
...
+	result := make(map[string]interface{})
+
+	result["type"] = string(input.Type)
+	if principalId := input.PrincipalID; principalId != nil {
+		result["principal_id"] = *principalId
+	}
+	if tenantId := input.TenantID; tenantId != nil {
+		result["tenant_id"] = *tenantId
+	}
+	return []interface{}{result}
```
tombuildsstuff: (same here) can we ensure these always have a value as above

azurerm/internal/services/compute/tests/data_source_disk_encryption_set_test.go
```diff
@@ -0,0 +1,65 @@
...
+			{
+				Config:  testAccPrepareKeyvaultAndKey(resourceGroup, location, vaultName, keyName),
+				Destroy: false,
+				Check:   resource.ComposeTestCheckFunc(),
+			},
+			// This step is not negligible, without this step, the final step will fail on refresh complaining `Disk Encryption Set does not exist`
+			{
+				PreConfig: func() { enableSoftDeleteAndPurgeProtectionForKeyvault(resourceGroup, vaultName) },
+				Config:    testAccAzureRMDiskEncryptionSet_basic(resourceGroup, location, vaultName, keyName, desName),
+			},
```
tombuildsstuff: presumably we also need a couple of steps at the end to first delete the Disk Encryption Set, then Purge & Disable Soft Delete before deleting the Key Vault?

azurerm/internal/services/compute/tests/data_source_disk_encryption_set_test.go
```diff
@@ -0,0 +1,65 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(dataSourceName, "identity.#", "1"),
+					resource.TestCheckResourceAttr(dataSourceName, "identity.0.type", "SystemAssigned"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "identity.0.principal_id"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "identity.0.tenant_id"),
+					resource.TestCheckResourceAttr(dataSourceName, "active_key.#", "1"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "active_key.0.source_vault_id"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "active_key.0.key_url"),
+				),
+			},
```
tombuildsstuff: presumably we need a couple of steps at the end to first delete the Disk Encryption Set, then Purge & Disable Soft Delete before deleting the Key Vault?

azurerm/internal/services/compute/tests/data_source_disk_encryption_set_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMDiskEncryptionSetDestroy,
+		Steps: []resource.TestStep{
+			// These two steps are used for setting up a valid keyVault, which enables soft-delete and purge protection.
+			{
+				Config:  testAccPrepareKeyvaultAndKey(resourceGroup, location, vaultName, keyName),
+				Destroy: false,
```
tombuildsstuff: this isn't needed, since `destroy: false` is the default

```suggestion
```

azurerm/internal/services/compute/tests/data_source_disk_encryption_set_test.go
```diff
@@ -0,0 +1,65 @@
...
+				Check:   resource.ComposeTestCheckFunc(),
+			},
+			// This step is not negligible, without this step, the final step will fail on refresh complaining `Disk Encryption Set does not exist`
+			{
+				PreConfig: func() { enableSoftDeleteAndPurgeProtectionForKeyvault(resourceGroup, vaultName) },
```
tombuildsstuff: FWIW we could roll this into the "Check" above, since it's relevant at that step

azurerm/internal/services/compute/tests/resource_arm_disk_encryption_set_test.go
```diff
@@ -0,0 +1,243 @@
...
+	desName := fmt.Sprintf("acctestdes-%d", ri)
+	location := acceptance.Location()
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMDiskEncryptionSetDestroy,
+		Steps: []resource.TestStep{
+			// This test step is temporary due to freezing of functions in keyVault.
+			// After applying soft-delete and purge-protection in keyVault, this extra step can be removed.
```
tombuildsstuff:
```suggestion
			// TODO: After applying soft-delete and purge-protection in keyVault, this extra step can be removed.
```

azurerm/internal/services/compute/tests/resource_arm_disk_encryption_set_test.go
```diff
@@ -0,0 +1,243 @@
...
+			{
+				Config:  testAccPrepareKeyvaultAndKey(resourceGroup, location, vaultName, keyName),
+				Destroy: false,
```
tombuildsstuff: (as above) we can remove this

```suggestion
```

azurerm/internal/services/compute/tests/resource_arm_disk_encryption_set_test.go
```diff
@@ -0,0 +1,243 @@
...
+					testCheckAzureRMDiskEncryptionSetExists(resourceName),
+					resource.TestCheckResourceAttrSet(resourceName, "active_key.0.source_vault_id"),
+					resource.TestCheckResourceAttrSet(resourceName, "active_key.0.key_url"),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
```
tombuildsstuff: (as above) do we need a couple of steps here to first delete the DES, then purge/disable soft delete to allow the key vault to be torn down?

azurerm/internal/services/compute/tests/resource_arm_disk_encryption_set_test.go
```diff
@@ -0,0 +1,243 @@
...
+		armClient := acceptance.AzureProvider.Meta().(*clients.Client)
+		client := armClient.KeyVault.VaultsClient
+		ctx := armClient.StopContext
+		vaultPatch := keyvault.VaultPatchParameters{
+			Properties: &keyvault.VaultPatchProperties{
+				EnableSoftDelete:      utils.Bool(true),
+				EnablePurgeProtection: utils.Bool(true),
+			},
+		}
+		log.Printf("[LOG] Updating")
```
tombuildsstuff: could we ensure we log what we're updating here?

```suggestion
		log.Printf("[DEBUG] Enabling Soft Delete & Purge Protection on Key Vault %q (Resource Group %q)..", vaultName, resourceGroup)
```

azurerm/internal/services/compute/validation.go
```diff
@@ -76,3 +76,18 @@ func validateName(maxLength int) func(i interface{}, k string) (warnings []strin
...
+
+func ValidateDiskEncryptionSetName(i interface{}, k string) (warnings []string, errors []error) {
+	v, ok := i.(string)
+	if !ok {
+		errors = append(errors, fmt.Errorf("expected type of %s to be string", k))
+		return
+	}
+
+	// Supported characters for the name are a-z, A-Z, 0-9 and _. The maximum name length is 80 characters.
+	// despite the swagger says the constraint above, but a name that contains hyphens will also pass.
```
tombuildsstuff: can we confirm this validation with the portal, which is (unfortunately) the most reliable source of this information

website/docs/d/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,66 @@
...
+
+* `location` - Resource location
+
+* `active_key` - One `active_key` block defined below.
+
+* `identity` - A `identity` block defined below.
+
+* `previous_keys` - One or more `previous_key` block defined below.
+
+* `tags` - Resource tags
```
tombuildsstuff: can we update this to use the terminology used for `tags` in every other resource?

website/docs/d/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,66 @@
...
+* `resource_group_name` - (Required) The name of the Resource Group where the Disk Encryption Set exists.
+
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - Resource Id
+
+* `location` - Resource location
```
tombuildsstuff: can we update this to use the terminology used for `location` in every other resource?

website/docs/d/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,66 @@
...
+* `name` - (Required) The name of the Disk Encryption Set exists.
+
+* `resource_group_name` - (Required) The name of the Resource Group where the Disk Encryption Set exists.
+
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - Resource Id
```
tombuildsstuff: can we update this to use the terminology used for `id` in every other resource?

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
+---
+subcategory: "Compute"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_disk_encryption_set"
+sidebar_current: "docs-azurerm-resource-disk-encryption-set"
+description: |-
+  Manage Azure DiskEncryptionSet instance.
```
tombuildsstuff: this isn't an instance?

```suggestion
  Manages a Disk Encryption Set.
```

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_disk_encryption_set"
+sidebar_current: "docs-azurerm-resource-disk-encryption-set"
+description: |-
+  Manage Azure DiskEncryptionSet instance.
+---
+
+# azurerm_disk_encryption_set
+
+Manage an Azure DiskEncryptionSet instance.
```
tombuildsstuff: this isn't an instance?

```suggestion
Manages a Disk Encryption Set.
```

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+-> **NOTE:** The DiskEncryptionSet service is currently in public preview and are only available in a limited set of regions. 
+
+## Example Usage
+
+```hcl
+data "azurerm_client_config" "current" {}
+
+resource "azurerm_resource_group" "test" {
+  name     = "des-example-rg"
+  location = "westus2"
```
tombuildsstuff: we use West Europe for all examples:

```suggestion
  location = "West Europe"
```

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+
+-> **NOTE:** The DiskEncryptionSet service is currently in public preview and are only available in a limited set of regions. 
+
+## Example Usage
+
+```hcl
+data "azurerm_client_config" "current" {}
+
+resource "azurerm_resource_group" "test" {
+  name     = "des-example-rg"
```
tombuildsstuff: we use "example-resources" for all new examples

```suggestion
  name     = "example-resources"
```

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+    key_url         = azurerm_key_vault_key.test.id
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the disk encryption set that is being created. The name can't be changed after the disk encryption set is created. Supported characters for the name are a-z, A-Z, 0-9, _ and -. The maximum name length is 80 characters. Changing this forces a new resource to be created.
```
tombuildsstuff: can we make this match the rest of the documentation:

```suggestion
* `name` - (Required) The name of the Disk Encryption Set. Changing this forces a new resource to be created.
```

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+
+* `resource_group_name` - (Required) Specifies the name of the Resource Group in which the Disk Encryption Set should exist. Changing this forces a new resource to be created.
```
tombuildsstuff:
```suggestion
* `resource_group_name` - (Required) Specifies the name of the Resource Group where the Disk Encryption Set should exist. Changing this forces a new resource to be created.
```

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+
+* `location` - (Required) Specifies the Azure Region where the Disk Encryption Set exists. Changing this forces a new resource to be created.
+
+* `active_key` - (Required) One `active_key` block defined below.
```
tombuildsstuff: can we make this match the rest of the documentation:

```suggestion
* `active_key` - (Required) An `active_key` block defined below.
```

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+
+* `key_url` - (Required) Specifies the URL pointing to a key or secret in KeyVault.
+
+-> **NOTE** Access to the KeyVault must be granted for this Disk Encryption Set, if you want to further use this Disk Encryption Set in a Managed Disk or Virtual Machine, or Virtual Machine Scale Set. For instructions, please refer to the doc of [Server side encryption of Azure managed disks](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disk-encryption).
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `previous_keys` - One or more `previous_key` block defined below.
```
tombuildsstuff: (as above) we can remove this, since it's not useful directly in Terraform:

```suggestion
```

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+
+* `identity` - (Optional) A `identity` block defined below.
+
+* `tags` - (Optional) A mapping of tags to assign to the Disk Encryption Set.
+
+---
+
+The `active_key` block supports the following:
+
+* `source_vault_id` - (Required) Specifies the resource ID of the KeyVault containing the key or secret.
```
tombuildsstuff: we don't use the phrasing "Resource ID" here, since it conflicts with Terraform's phrasing - instead we make this `the ID of ..`

```suggestion
* `source_vault_id` - (Required) Specifies the ID of the Key Vault where the associated Key or Secret should be sourced from.
```

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+
+* `key_url` - (Required) Specifies the URL pointing to a key or secret in KeyVault.
```
tombuildsstuff:
```suggestion
* `key_url` - (Required) Specifies the URL to a Key Vault Key (either from a Key Vault Key, or the Key URL for the Key Vault Secret).
```

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+
+The following attributes are exported:
+
+* `previous_keys` - One or more `previous_key` block defined below.
+
+* `id` - The ID of the Disk Encryption Set.
+
+---
+
+A `identity` block supports the following:
```
tombuildsstuff:
```suggestion
A `identity` block exports the following:
```

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+
+~> **NOTE:** When `type` is set to `SystemAssigned`, identity the Principal ID can be retrieved after the Disk Encryption Set has been created. See [documentation](https://docs.microsoft.com/en-us/azure/active-directory/managed-service-identity/overview) for additional information.
+
+---
+
+The `previous_key` block contains the following:
+
+* `source_vault_id` - (Required) Specifies the resource ID of the KeyVault containing the key or secret.
+
+* `key_url` - (Required) Specifies the URL pointing to a key or secret in KeyVault.
```
tombuildsstuff: (as above) we can remove this

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+
+## Import
+
+Disk Encryption Set can be imported using the `resource id`, e.g.
+
+```shell
+terraform import azurerm_disk_encryption_set.test /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/group1/providers/Microsoft.Compute/diskEncryptionSets/des1
```
tombuildsstuff:
```suggestion
terraform import azurerm_disk_encryption_set.test /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/group1/providers/Microsoft.Compute/diskEncryptionSets/encryptionSet1
```

azurerm/internal/services/compute/data_source_disk_encryption_set.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	if encryptionSetProperties := resp.EncryptionSetProperties; encryptionSetProperties != nil {
+		if err := d.Set("active_key", flattenArmDiskEncryptionSetKeyVaultAndKeyReference(encryptionSetProperties.ActiveKey)); err != nil {
+			return fmt.Errorf("Error setting `active_key`: %+v", err)
+		}
+		if err := d.Set("previous_keys", flattenArmDiskEncryptionSetKeyVaultAndKeyReferenceArray(encryptionSetProperties.PreviousKeys)); err != nil {
+			return fmt.Errorf("Error setting `previous_keys`: %+v", err)
+		}
+	}
+	if identity := resp.Identity; identity != nil {
```
tombuildsstuff: we need to ensure that `identity` is set in all circumstances, so we can move this check inside the flatten function

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,387 @@
...
+				ForceNew: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"key_url": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validate.NoEmptyStrings,
+						},
+						"source_vault_id": {
```
tombuildsstuff: we should be able to infer this from the Key URL - meaning instead we could make a single top-level field for `key_vault_key_uri` ?

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+sidebar_current: "docs-azurerm-resource-disk-encryption-set"
+description: |-
+  Manage Azure DiskEncryptionSet instance.
+---
+
+# azurerm_disk_encryption_set
+
+Manage an Azure DiskEncryptionSet instance.
+
+-> **NOTE:** The DiskEncryptionSet service is currently in public preview and are only available in a limited set of regions.
```
tombuildsstuff: can we make the phrasing consistent with other resources:

```suggestion
-> **NOTE:** Disk Encryption Sets are currently in Public Preview and are only available in a limited set of regions. 
```

how do users opt-into this?

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+description: |-
+  Manage Azure DiskEncryptionSet instance.
+---
+
+# azurerm_disk_encryption_set
+
+Manage an Azure DiskEncryptionSet instance.
+
+-> **NOTE:** The DiskEncryptionSet service is currently in public preview and are only available in a limited set of regions. 
+
```
tombuildsstuff: since this requires that the Key Vault is configured in a particular way - can we also document that here:

```

-> **NOTE:** At this time the Key Vault used to store the Active Key for this Disk Encryption Set must have both Soft Delete & Purge Protection enabled - which are not yet supported by Terraform - instead you can configure this using [a provisioner](https://www.terraform.io/docs/provisioners/local-exec.html) or [the `azurerm_template_deployment` resource](https://www.terraform.io/docs/providers/azurerm/r/template_deployment.html).

```

tombuildsstuff:
Sorry for the double review - Github bug 🙃

azurerm/internal/services/compute/data_source_disk_encryption_set.go
ArcturusZhang: Done

azurerm/internal/services/compute/data_source_disk_encryption_set.go
ArcturusZhang: Done

azurerm/internal/services/compute/data_source_disk_encryption_set.go
ArcturusZhang: Done

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
ArcturusZhang: Done

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
ArcturusZhang: Done

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
ArcturusZhang: Sure, now I am passing *client,Client to this method

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
ArcturusZhang: Done

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
ArcturusZhang: Oh... this is a mistake. We do not need to check `WasNotFound` here

azurerm/internal/services/compute/tests/data_source_disk_encryption_set_test.go
ArcturusZhang: Done

azurerm/internal/services/compute/tests/data_source_disk_encryption_set_test.go
ArcturusZhang: No, we cannot. Both soft-delete and purge-protection are irreversible. once turning on, there is no turning back.
Soft-delete means that the key vault could be recovered after deletion, if you want to get rid of it permanently, you can purge it. Soft-deleted key vault will remain in Azure for 90 days.
Purge-protection means that you cannot purge this key vault.
In this case, the key vault can be deleted (soft deleted, recoverable) but it will persist its name for 90 days.

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: Done

website/docs/d/disk_encryption_set.html.markdown
ArcturusZhang: No problem

website/docs/d/disk_encryption_set.html.markdown
ArcturusZhang: Done

website/docs/d/disk_encryption_set.html.markdown
ArcturusZhang: Done. I also removed those fields that do not exist in  this data sources any more.

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: DOne

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: Let me ask the service team and then fill this information in.

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: Added. And do you think we should add the provisioner command in the example?

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: Done

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: Done

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: The field is deleted and now inferred from the key_url

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: Done

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: Done

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: Done

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: Removed

ArcturusZhang:
Hi @tombuildsstuff I have resolved some comments, and left some replies. Please have a look.
Additionally, I cannot pass the acctest on my side now. It works well when I am submitting this PR.
It reports:
```
=== RUN   TestAccAzureRMDiskEncryptionSet_basic
=== PAUSE TestAccAzureRMDiskEncryptionSet_basic
=== CONT  TestAccAzureRMDiskEncryptionSet_basic
--- FAIL: TestAccAzureRMDiskEncryptionSet_basic (0.07s)
    testing.go:569: Step 0 error: Error initializing context: 2 problems:

        - Could not satisfy plugin requirements:
        Plugin reinitialization required. Please run "terraform init".

        Plugins are external binaries that Terraform uses to access and manipulate
        resources. The configuration provided requires plugins which can't be located,
        don't satisfy the version constraints, or are otherwise incompatible.

        Terraform automatically discovers provider requirements from your
        configuration, including providers used in child modules. To see the
        requirements and constraints from each module, run "terraform providers".

        - provider "azurerm" is not available
```
Could you please help me on this? What is possibly causing this error? Or is there some kinds of caches terraform are using during the acctest?

azurerm/internal/services/compute/validation.go
ArcturusZhang: I have got the replies from the service team, and changed the regex.

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: Done

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: Done

website/docs/r/disk_encryption_set.html.markdown
ArcturusZhang: This block has been removed per your suggestion

ArcturusZhang:
Hi @tombuildsstuff I have fixed the acc-tests: 
```
=== RUN   TestAccAzureRMDiskEncryptionSet_basic
=== PAUSE TestAccAzureRMDiskEncryptionSet_basic
=== CONT  TestAccAzureRMDiskEncryptionSet_basic
--- PASS: TestAccAzureRMDiskEncryptionSet_basic (367.77s)
PASS
=== RUN   TestAccDataSourceAzureRMDiskEncryptionSet_basic
=== PAUSE TestAccDataSourceAzureRMDiskEncryptionSet_basic
=== CONT  TestAccDataSourceAzureRMDiskEncryptionSet_basic
--- PASS: TestAccDataSourceAzureRMDiskEncryptionSet_basic (428.47s)
PASS
```

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
ArcturusZhang: I have modified this to make sure all these field would have a value, please have a look~

tombuildsstuff:
@ArcturusZhang hope you don't mind but I've rebased this on top of master to fix the merge conflicts

ArcturusZhang:
> @ArcturusZhang Dapeng Zhang FTE hope you don't mind but I've rebased this on top of master to fix the merge conflicts

Sure, thanks.
But the CI is failing on `website-lint`, how do I fix this? Because i did not really catch any useful information on the log

ArcturusZhang:
Hi @tombuildsstuff I should have fixed the CI failure, please have a look 🚀

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,291 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateDiskEncryptionSetName,
+			},
+
+			"location": azure.SchemaLocation(),
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"key_vault_key_uri": {
```
tombuildsstuff: this wants to be:

```suggestion
			"key_vault_key_id": {
```

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,291 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.NoEmptyStrings,
```
tombuildsstuff: we can validate this is a Key Vault Child (Key/Secret) ID using:

```suggestion
				ValidateFunc: azure.ValidateKeyVaultChildID,
```

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,291 @@
...
+			},
+
+			"identity": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
tombuildsstuff:
```suggestion
```

azurerm/internal/services/compute/resource_arm_disk_encryption_set.go
```diff
@@ -0,0 +1,291 @@
...
+			"key_vault_key_uri": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+
+			"identity": {
+				Type:     schema.TypeList,
+				Optional: true,
```
tombuildsstuff: after spending some time playing with this, this wants to be Required:

```suggestion
				Required: true,
```

since if this is unspecified the API returns: `Required parameter 'ResourceIdentity' is missing (null)`

azurerm/internal/services/compute/tests/resource_arm_disk_encryption_set_test.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func TestAccAzureRMDiskEncryptionSet_basic(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_disk_encryption_set", "test")
+	resourceGroup := fmt.Sprintf("acctestRG-%d", data.RandomInteger)
+	vaultName := fmt.Sprintf("vault%d", data.RandomInteger)
+	keyName := fmt.Sprintf("key-%s", data.RandomString)
+	desName := fmt.Sprintf("acctestdes-%d", data.RandomInteger)
```
tombuildsstuff: (same here)

website/docs/r/disk_encryption_set.html.markdown
```diff
@@ -0,0 +1,121 @@
...
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The ID of the Disk Encryption Set.
+
+---
+
+A `identity` block exports the following:
```
tombuildsstuff: this is the properties supported by this block, rather than exported by it - so we need to move this up & duplicate this for the exported fields (`principal_id` and `tenant_id`)

tombuildsstuff:
Tests pass:

Screenshot 2020-01-10 at 11 53 44

ArcturusZhang:
Thanks @tombuildsstuff !

PR #5251, opened 2019-12-25
Update VMSS for SSE-CMK feature

ArcturusZhang:
This is part of update of the [SSE-CMK feature](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disk-encryption), adding new fields for `azurerm_virtual_machine_scale_set`.


tombuildsstuff:
hi @ArcturusZhang 

Thanks for this PR :)

At this point in time the existing `azurerm_virtual_machine_scale_set` resource is feature-frozen since it's being superseded in 2.0 - as such whilst I'd like to thank you for this contribution, I'm going to close this in favour of #5252 (which you've also submitted, which adds this to the replacement VMSS resources).

Thanks!

Issue #5257, opened 2019-12-26
Support for data migration service


magodo:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Support data migration service and its related resources. It is a fully managed service designed to enable seamless migrations from multiple database sources to Azure data platforms with minimal downtime (online migrations). More info [here](https://docs.microsoft.com/en-us/azure/dms/dms-overview).

The resource in terraform is meant to support resource provision/destroy. Not going to handle any procedure task (e.g. only support create/destroy a migration task but not start/stop a migration task).

### New or Affected Resource(s)

* azurerm_data_migration_service
* azurerm_data_migration_project
* azurerm_data_migration_task

### Potential Terraform Configuration

```hcl
resource "azurerm_data_migration_service" "example" {
	name                = "example-dms"
	location            = azurerm_resource_group.example.location
	resource_group_name = azurerm_resource_group.example.name
	virtual_subnet_id   = azurerm_subnet.example.id
	sku_name            = "Standard_1vCores"
}

resource "azurerm_data_migration_service_project" "example" {
	name                = "example-dms-project"
	location            = azurerm_resource_group.example.location
	resource_group_name = azurerm_resource_group.example.name
        service_name = azurerm_data_migration_service.example.name
	source_platform = "SQL"
        target_platform = "SQLDB"
}

resource "azurerm_data_migration_service_task_sql_server" "test" {
  name                       = "example-dms-task"
  service_name               = azurerm_data_migration_service.example.name
  project_name               = azurerm_data_migration_service_project.example.name
  resource_group_name        = azurerm_resource_group.example.name
  databases {
    source_database_name = "FooDB"
    target_database_name = "SomeAzureSqlDB"
    source_tables        = ["dbo.SomeTable"]
    target_tables        = ["dbo.SomeTable"]
  }
  source_connection_info {
    server_address           = "1.2.3.4"
    server_port              = 1433
    user_name                = "SA"
    password                 = "secret"
    encrypt_connection       = true
    trust_server_certificate = true
  }
  target_connection_info {
    server_address           = "some-sql-server.database.windows.net"
[...]

PR #5258, opened 2019-12-26
New Resource & Data Source: azurerm_data_migration_service

magodo:
Add new resource: data migration service.

This is the first resource under DataMigration resource provider.

There are more resources to come (which are not included in this PR):

- data migration project
- data migration task

I will implement those two resources after this PR is merged since they depends on data migration service.

Check feature request #5257 for more details.


magodo:
This PR depends on #5259

azurerm/internal/services/datamigration/data_source_data_migration_service.go
```diff
@@ -0,0 +1,100 @@
...
+func dataSourceArmDataMigrationService() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmDataMigrationServiceRead,
+
+		Schema: map[string]*schema.Schema{
+			"resource_group_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
```
katbyte: Could we use the shared schema here

azurerm/internal/services/datamigration/data_source_data_migration_service.go
```diff
@@ -0,0 +1,100 @@
...
+
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
```
katbyte: Could we validate this properly and not just with no zero values

azurerm/internal/services/datamigration/data_source_data_migration_service.go
```diff
@@ -0,0 +1,100 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"kind": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"provisioning_state": {
```
katbyte: I don't think we should be exposing this?

azurerm/internal/services/datamigration/data_source_data_migration_service.go
```diff
@@ -0,0 +1,100 @@
...
+	if location := resp.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	d.Set("resource_group_name", resourceGroup)
+	d.Set("name", resp.Name)
+	if serviceProperties := resp.ServiceProperties; serviceProperties != nil {
+		d.Set("provisioning_state", string(serviceProperties.ProvisioningState))
+		d.Set("virtual_subnet_id", serviceProperties.VirtualSubnetID)
+	}
+	if err := d.Set("sku_name", resp.Sku.Name); err != nil {
```
katbyte: We need to nil check SKU & cast Name to a string

azurerm/internal/services/datamigration/data_source_data_migration_service.go
```diff
@@ -0,0 +1,100 @@
...
+		return fmt.Errorf("Error setting `sku_name`: %+v", err)
```
katbyte: we don't need to check this for an error

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,247 @@
...
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"location": azure.SchemaLocation(),
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"name": {
```
katbyte: name should be the first property

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,247 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.NoEmptyStrings,
```
katbyte: could we properly validate this

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,247 @@
...
+			},
+
+			"virtual_subnet_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: azure.ValidateResourceID,
```
katbyte: could we validate this is a Azure ID?

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,247 @@
...
+			},
+
+			"sku_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.NoEmptyStrings,
```
katbyte: We should validate this is one of the correct SKU values?

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,247 @@
...
+				Default:      "Cloud",
+				ValidateFunc: validation.StringInSlice([]string{"Cloud"}, false),
+			},
+
+			"delete_running_tasks": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
+			"tags": tags.Schema(),
```
katbyte: tags should always be last

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,247 @@
...
+
+			"type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"provisioning_state": {
```
katbyte: i don't think we should expose this?

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	d.Set("resource_group_name", resourceGroup)
+	d.Set("name", name)
+	if serviceProperties := resp.ServiceProperties; serviceProperties != nil {
+		d.Set("provisioning_state", string(serviceProperties.ProvisioningState))
+		d.Set("virtual_subnet_id", serviceProperties.VirtualSubnetID)
+	}
+	if err := d.Set("sku_name", resp.Sku.Name); err != nil {
+		return fmt.Errorf("Error setting `sku_name`: %+v", err)
+	}
```
katbyte: we don't need to check for an error here, should case name to a string, and need to nil check SKU

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		return err
+	}
+	resourceGroup := id.ResourceGroup
+	name := id.Path["services"]
+
+	parameters := datamigration.Service{}
+
+	if t, ok := d.GetOk("tags"); ok {
+		parameters.Tags = tags.Expand(t.(map[string]interface{}))
```
katbyte: This should be `Tags: tags.Expand(t),` ?

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,247 @@
...
+
+			"kind": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				Default:      "Cloud",
+				ValidateFunc: validation.StringInSlice([]string{"Cloud"}, false),
+			},
+
+			"delete_running_tasks": {
```
katbyte: THis isn't; force new and is not in the update function? should this be in tf?

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,247 @@
...
+
+	id, err := azure.ParseAzureResourceID(d.Id())
+	if err != nil {
+		return err
+	}
+	resourceGroup := id.ResourceGroup
+	name := id.Path["services"]
+
+	var deleteRunningTasks *bool
+	if toDelete, ok := d.GetOk("delete_running_tasks"); ok {
```
katbyte: This won't work if the resource cfg is removed

azurerm/internal/services/datamigration/tests/resource_arm_data_migration_service_test.go
```diff
@@ -0,0 +1,210 @@
...
+			},
+			{
+				Config:      testAccAzureRMDataMigrationService_requiresImport(data),
+				ExpectError: acceptance.RequiresImportError("azurerm_data_migration_service"),
+			},
+		},
+	})
+}
+
+func TestAccAzureRMDataMigrationService_complete(t *testing.T) {
```
katbyte: the complete task should not have an update step, there should be a seperate update test

azurerm/internal/services/datamigration/tests/resource_arm_data_migration_service_test.go
```diff
@@ -0,0 +1,210 @@
...
+		return nil
+	}
+
+	return nil
+}
+
+func testAccAzureRMDataMigrationService_base(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-%d"
```
katbyte:
```suggestion
  name     = "acctestRG-dm-%d"
```

website/docs/d/data_migration_service.html.markdown
```diff
@@ -0,0 +1,56 @@
...
+
+* `sku_name` - The resource's sku name
+
+* `type` - The resource type chain (e.g. virtualMachines/extensions)
+
+* `kind` - The resource kind.
+
+* `provisioning_state` - The resource's provisioning state
+
+* `tags` - Resource tags.
```
katbyte: Could we explain what these properties do/make it read and sound like the rest of the data source documentation?

website/docs/r/data_migration_service.html.markdown
```diff
@@ -0,0 +1,82 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `type` - (Optional) The resource type chain (e.g. virtualMachines/extensions)
+
+* `provisioning_state` - The resource's provisioning state
+
+* `id` - Resource ID.
```
katbyte: could we make this read like the rest of the resources? and be a be more clear what some of these properties do? ie type

katbyte (changes requested):
Thanks for the PR @magodo, iv'e left some comments inline that need to be addressed before we merge this

katbyte (changes requested):
Thanks for the PR @magodo, iv'e left some comments inline that need to be addressed before we merge this

azurerm/internal/services/datamigration/data_source_data_migration_service.go
magodo: I will nil check `resp.Sku`. While for the casting, I think `d.Set()` already handle the case where you are passing in a pointer to a primitive type (`resp.Sku.Name` here is of type `*string`) and seems not necessary. Or no?

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
magodo: I will nil check resp.Sku. While for the casting, I think d.Set() already handle the case where you are passing in a pointer to a primitive type (resp.Sku.Name here is of type *string) and seems not necessary. Or no?

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
magodo: Because `d.GetOk()` returns an `interface{}` while `tags.Expand()` expect a `map[string]interface{}`. Need a type cast here to make compiler happy.

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
magodo: What do you mean by *Azure ID*. The validation function used here is the same one as you did in resource: **resource_arm_eventhub_namespace**.

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
magodo: After a reconsideration, I think we can just hide this property from user and force it as `true`. The meaning of this property is:

> deleteRunningTasks - delete the resource even if it contains running tasks

This is always guaranteed in terraform if **task** is another resource which depends on **data migration service**. So terraform has already handled it properly. For those **task** created out-of-bound, we can just force delete them.

azurerm/internal/services/datamigration/data_source_data_migration_service.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmDataMigrationService() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmDataMigrationServiceRead,
+
```
katbyte: We'll need to add timeouts to this resource

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,241 @@
...
+	return &schema.Resource{
+		Create: resourceArmDataMigrationServiceCreate,
+		Read:   resourceArmDataMigrationServiceRead,
+		Update: resourceArmDataMigrationServiceUpdate,
+		Delete: resourceArmDataMigrationServiceDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
```
katbyte: We'll need to add timeouts to this resource

azurerm/internal/services/datamigration/data_source_data_migration_service.go
```diff
@@ -0,0 +1,88 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validateName,
+			},
+			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
```
katbyte:
```suggestion

			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
```

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,241 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateName,
+			},
+
+			"location": azure.SchemaLocation(),
+
+			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
+
+			"virtual_subnet_id": {
```
katbyte: Could we rename this to
```suggestion
			"subnet_id": {
```

to match other resources? no where else do we use the term `virtual_subnet_id`

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,241 @@
...
+				Optional:     true,
+				ForceNew:     true,
+				Default:      "Cloud",
+				ValidateFunc: validation.StringInSlice([]string{"Cloud"}, false),
+			},
+
+			"type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
katbyte: Does this make sense?

azurerm/internal/services/datamigration/data_source_data_migration_service.go
```diff
@@ -0,0 +1,88 @@
...
+
+			"sku_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
katbyte: I don't think this makes sense? no other resource exports it

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
katbyte: Given ever where else we don't check if this exists or not i think this can be:
```
parameters := datamigration.Service{
	Tags := tags.Expand(d.Get("tags").(map[string]interface{}))
```

azurerm/internal/services/datamigration/tests/resource_arm_data_migration_service_test.go
```diff
@@ -0,0 +1,231 @@
...
+			{
+				Config: testAccAzureRMDataMigrationService_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMDataMigrationServiceExists(data.ResourceName),
+				),
+			},
+			{
+				Config:      testAccAzureRMDataMigrationService_requiresImport(data),
+				ExpectError: acceptance.RequiresImportError("azurerm_data_migration_service"),
+			},
```
katbyte: Could we use the new `data.RequiresImportErrorStep` function?

katbyte (changes requested):
Thanks for the revisions @magodo, i've left some more comments inline

azurerm/internal/services/datamigration/data_source_data_migration_service.go
magodo: I add a 30-5-30-30 minutes timeout to this resource. In my personal test, the creation takes longest, ~20min. There is no SLA for timeout in service document though.

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
magodo: Right, this is fine to remove..

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,241 @@
...
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateName,
+			},
+
+			"location": azure.SchemaLocation(),
+
+			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
```
tombuildsstuff: this needs to be case-sensitive

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,241 @@
...
+				}, false),
+			},
+
+			"kind": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				Default:      "Cloud",
+				ValidateFunc: validation.StringInSlice([]string{"Cloud"}, false),
+			},
```
tombuildsstuff: in which case we can remove this field and hard-code this for the moment

azurerm/internal/services/datamigration/data_source_data_migration_service.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmDataMigrationService() *schema.Resource {
```
tombuildsstuff: I think we can remove this Data Source, since this isn't something users are likely to use directly from Terraform?

azurerm/internal/services/datamigration/tests/resource_arm_data_migration_service_test.go
```diff
@@ -0,0 +1,228 @@
...
+					testCheckAzureRMDataMigrationServiceExists(data.ResourceName),
+				),
+			},
+			{
+				Config: testAccAzureRMDataMigrationService_complete(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMDataMigrationServiceExists(data.ResourceName),
+					resource.TestCheckResourceAttr(data.ResourceName, "tags.name", "test"),
+				),
+			},
```
tombuildsstuff: can we add an import step here?

azurerm/internal/services/datamigration/tests/resource_arm_data_migration_service_test.go
```diff
@@ -0,0 +1,228 @@
...
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMDataMigrationServiceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMDataMigrationService_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMDataMigrationServiceExists(data.ResourceName),
+				),
+			},
```
tombuildsstuff: can we add an import step here?

website/docs/r/data_migration_service.html.markdown
```diff
@@ -0,0 +1,77 @@
+---
```
tombuildsstuff: can we add this to the sidebar too?

website/docs/r/data_migration_service.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) Specify the name of the data migration service. Changing this forces a new resource to be created.
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) Name of the resource group in which to create the data migration service. Changing this forces a new resource to be created.
+
+* `virtual_subnet_id` - (Required) The ID of the virtual subnet resource to which the data migration service should be joined. Changing this forces a new resource to be created.
```
tombuildsstuff: should we be making this `subnet_id` to match other resources?

website/docs/r/data_migration_service.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+
+* `sku_name` - (Required) The sku name of the data migration service. Currently only one of "Premium_4vCores", "Standard_1vCores", "Standard_2vCores", "Standard_4vCores" is allowed. Changing this forces a new resource to be created.
```
tombuildsstuff: can we quote these to match other resources:

```suggestion
* `sku_name` - (Required) The sku name of the data migration service. Possible values are `Standard_1vCores`, `Standard_2vCores`, `Standard_4vCores` and `Premium_4vCores`. Changing this forces a new resource to be created.
```

website/docs/r/data_migration_service.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+
+* `kind` - (Optional) The resource kind. Only 'Cloud' (the default) is supported. Changing this forces a new resource to be created.
```
tombuildsstuff: (as above) we can remove this

website/docs/r/data_migration_service.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+
+* `tags` - (Optional) A mapping of tags to assigned to the resource.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `type` - (Optional) The resource type chain (e.g. virtualMachines/extensions)
```
tombuildsstuff: this field doesn't seem useful to export - can we remove it?

website/docs/r/data_migration_service.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+
+* `id` - Resource ID of data migration service.
```
tombuildsstuff: we don't use the term "Resource ID" since it's ambiguous between Terraform and Azure - could we instead make this:

```suggestion
* `id` - The ID of Data Migration Service.
```

to match the other resources

website/docs/r/data_migration_service.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+
+## Import
+
+Data Migration Service can be imported using the `resource id`, e.g.
```
tombuildsstuff:
```suggestion
Data Migration Services can be imported using the `resource id`, e.g.
```

website/docs/r/data_migration_service.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+# azurerm_data_migration_service
+
+Manage Azure Data Migration Service instance.
+
+## Example Usage
+
+```hcl
+resource "azurerm_resource_group" "example" {
+  name     = "example-rg"
+  location = "northeurope"
```
tombuildsstuff: we use West Europe for all examples - as such could we make this:

```suggestion
  location = "West Europe"
```

website/docs/r/data_migration_service.html.markdown
```diff
@@ -0,0 +1,77 @@
+---
+subcategory: "Data Migration"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_data_migration_service"
+sidebar_current: "docs-azurerm-resource-data-migration-service"
+description: |-
+  Manage Azure Service instance.
```
tombuildsstuff:
```suggestion
  Manages a Data Migration Service.
```

website/docs/r/data_migration_service.html.markdown
```diff
@@ -0,0 +1,77 @@
...
+---
+
+# azurerm_data_migration_service
+
+Manage Azure Data Migration Service instance.
```
tombuildsstuff:
```suggestion
Manages a Data Migration Service.
```

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,241 @@
...
+	skuName := d.Get("sku_name").(string)
+	subnetID := d.Get("subnet_id").(string)
+	location := d.Get("location").(string)
+
+	parameters := datamigration.Service{
+		Location: utils.String(location),
+		ServiceProperties: &datamigration.ServiceProperties{
+			VirtualSubnetID: utils.String(subnetID),
+		},
+		Sku: &datamigration.ServiceSku{Name: utils.String(skuName)},
```
tombuildsstuff: _minor_ could we new-line this for readability?

```suggestion
		Sku: &datamigration.ServiceSku{
			Name: utils.String(skuName),
		},
```

tombuildsstuff (changes requested):
hey @magodo 

Thanks for this PR :)

Taking a look through this is looking good to me - if we can remove the data source, fix up the pending comments (and the tests pass) then this should otherwise be good 👍 

Thanks!

azurerm/internal/services/datamigration/data_source_data_migration_service.go
magodo: Acutally there are some upcoming resources on top of DMS (see #5257 for more info). It might be some rare cases that user A created a DMS service and define different DMS projects for different migration scenario into different terraform modules, inside where user might just want to pass in an input variable indicating in which DMS server to create the DMS project/tasks.

magodo: So I think we can just leave the data source there and remove the `kind` properties for now?

azurerm/internal/services/datamigration/tests/resource_arm_data_migration_service_test.go
magodo: Just wondering what is the rule we are following here 🤔. Is it like after every apply we should add a imoprt verfication?

website/docs/r/data_migration_service.html.markdown
magodo: Do you mean we should not add it into sidebar for now?

website/docs/r/data_migration_service.html.markdown
magodo: Yes, my mistake 😥

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
magodo: Sure

azurerm/internal/services/datamigration/tests/resource_arm_data_migration_service_test.go
katbyte: basically yes thou with a few exceptions. Every step should have an import to check the values in azure are as expected.

azurerm/internal/services/datamigration/data_source_data_migration_service.go
```diff
@@ -0,0 +1,82 @@
...
+	d.Set("id", resp.ID)
+	if location := resp.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	d.Set("resource_group_name", resourceGroup)
+	d.Set("name", resp.Name)
+	if serviceProperties := resp.ServiceProperties; serviceProperties != nil {
+		d.Set("subnet_id", serviceProperties.VirtualSubnetID)
+	}
+	if resp.Sku != nil && resp.Sku.Name != nil {
```
katbyte: we don't need to check name and d.set will
```suggestion
	if resp.Sku != nil {
```

azurerm/internal/services/datamigration/data_source_data_migration_service.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return fmt.Errorf("Error: Data Migration Service (Service Name %q / Group Name %q) was not found", name, resourceGroup)
+		}
+		return fmt.Errorf("Error reading Data Migration Service (Service Name %q / Group Name %q): %+v", name, resourceGroup, err)
+	}
+
+	d.SetId(*resp.ID)
+
+	d.Set("id", resp.ID)
```
katbyte: d.SetId will set this

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[INFO] Data Migration Service %q does not exist - removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error reading Data Migration Service (Service Name %q / Group Name %q): %+v", name, resourceGroup, err)
+	}
+
+	d.Set("id", resp.ID)
```
katbyte: We should be using `d.SetID()` here?

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,232 @@
...
+	if location := resp.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	d.Set("resource_group_name", resourceGroup)
+	d.Set("name", name)
+	if serviceProperties := resp.ServiceProperties; serviceProperties != nil {
+		d.Set("subnet_id", serviceProperties.VirtualSubnetID)
+	}
+	if resp.Sku != nil && resp.Sku.Name != nil {
```
katbyte: We don't need to check name is nil as d.set will

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,232 @@
...
+	id, err := azure.ParseAzureResourceID(d.Id())
+	if err != nil {
+		return err
+	}
+	resourceGroup := id.ResourceGroup
+	name := id.Path["services"]
+
+	// For dms tasks created via terraform, the deletion order has already handled the dependency between dms task and dms service. In which case, dms service will not start to delete until
+	// dms tasks have been deleted.
+	// For dms tasks created out of terraform, it is user's responsibility to ensure the deletion order. And for dms service, it makes sense to just force delete outstanding tasks.
```
katbyte:
```suggestion
	// For dms tasks created outside of terraform, it is user's responsibility to ensure the deletion order. And for dms service, it makes sense to just force delete outstanding tasks.
```

katbyte: Could we also add a note about this in the docs?

website/docs/d/data_migration_service.html.markdown
```diff
@@ -0,0 +1,51 @@
...
+
+* `id` - The ID of Data Migration Service.
+
+* `location` - Azure location where the resource exists.
+
+* `subnet_id` - The ID of the virtual subnet resource to which the data migration service exists.
+
+* `sku_name` - The sku name of data migration service.
+
+* `kind` - The resource kind.
```
katbyte: the property has been removed so we should remove it here too

website/docs/r/data_migration_service.html.markdown
katbyte: It should be added to the website index erb so it shows up in the nav

website/docs/d/data_migration_service.html.markdown
```diff
@@ -0,0 +1,51 @@
+---
```
katbyte: This needs to be added to the sidebar

website/docs/r/data_migration_service.html.markdown
katbyte: same comment

katbyte (changes requested):
Thanks for the updates @magodo, in addition to my comments it looks like you have a test failure:

```
Test Failed

------- Stdout: -------
=== RUN   TestAccAzureRMDataMigrationService_complete

panic: Invalid address to set: []string{"id"}



goroutine 328 [running]:

github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*ResourceData).Set(0xc000a317a0, 0x4061c0c, 0x2, 0x39b0140, 0xc0013a0ea0, 0x0, 0xdf8475800)

	/opt/teamcity-agent/work/dbea4b84960ec72a/pkg/mod/github.com/hashicorp/terraform-plugin-sdk@v1.1.1/helper/schema/resource_data.go:193 +0x354

github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/datamigration.resourceArmDataMigrationServiceRead(0xc000a317a0, 0x3aa3100, 0xc001122240, 0x0, 0x0)

	/opt/teamcity-agent/work/dbea4b84960ec72a/src/github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/datamigration/resource_arm_data_migration_service.go:158 +0x5ba

github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/datamigration.resourceArmDataMigrationServiceCreate(0xc000a317a0, 0x3aa3100, 0xc001122240, 0x0, 0x0)

	/opt/teamcity-agent/work/dbea4b84960ec72a/src/github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/datamigration/resource_arm_data_migration_service.go:133 +0xa38

github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Resource).Apply(0xc000bfb980, 0xc000796050, 0xc00162c2a0, 0x3aa3100, 0xc001122240, 0xc001630301, 0xc001984150, 0xc0016303d0)

	/opt/teamcity-agent/work/dbea4b84960ec72a/pkg/mod/github.com/hashicorp/terraform-plugin-sdk@v1.1.1/helper/schema/resource.go:305 +0x365

github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Provider).Apply(0xc00148e400, 0xc001039918, 0xc000796050, 0xc00162c2a0, 0xc0016303e0, 0xc0008b9160, 0x3a7eae0)

	/opt/teamcity-agent/work/dbea4b84960ec72a/pkg/mod/github.com/hashicorp/terraform-plugin-sdk@v1.1.1/helper/schema/provider.go:294 +0x99

github.com/hashicorp/terraform-plugin-sdk/internal/helper/plugin.(*GRPCProviderServer).ApplyResourceChange(0xc0008b84b8, 0x47dcf00, 0xc001942cc0, 0xc0015e3680, 0xc0008b84b8, 0xc001942cc0, 0xc00255da80)

	/opt/teamcity-agent/work/dbea4b84960ec72a/pkg/mod/github.com/hashicorp/terraform-plugin-sdk@v1.1.1/internal/helper/plugin/grpc_provider.go:885 +0x882

github.com/hashicorp/terraform-plugin-sdk/internal/tfplugin5._Provider_ApplyResourceChange_Handler(0x3e50c60, 0xc0008b84b8, 0x47dcf00, 0xc001942cc0, 0xc0015e3620, 0x0, 0x47dcf00, 0xc001942cc0, 0xc001629180, 0x370)

	/opt/teamcity-agent/work/dbea4b84960ec72a/pkg/mod/github.com/hashicorp/terraform-plugin-sdk@v1.1.1/internal/tfplugin5/tfplugin5.pb.go:3189 +0x217

google.golang.org/grpc.(*Server).processUnaryRPC(0xc002be5080, 0x482eee0, 0xc001252000, 0xc0015f9100, 0xc0013fc0c0, 0x71d1ae0, 0x0, 0x0, 0x0)

	/opt/teamcity-agent/work/dbea4b84960ec72a/pkg/mod/google.golang.org/grpc@v1.23.0/server.go:995 +0x460

google.golang.org/grpc.(*Server).handleStream(0xc002be5080, 0x482eee0, 0xc001252000, 0xc0015f9100, 0x0)

	/opt/teamcity-agent/work/dbea4b84960ec72a/pkg/mod/google.golang.org/grpc@v1.23.0/server.go:1275 +0xd97

google.golang.org/grpc.(*Server).serveStreams.func1.1(0xc001097ea0, 0xc002be5080, 0x482eee0, 0xc001252000, 0xc0015f9100)

	/opt/teamcity-agent/work/dbea4b84960ec72a/pkg/mod/google.golang.org/grpc@v1.23.0/server.go:710 +0xbb

created by google.golang.org/grpc.(*Server).serveStreams.func1

	/opt/teamcity-agent/work/dbea4b84960ec72a/pkg/mod/google.golang.org/grpc@v1.23.0/server.go:708 +0xa1

FAIL	github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/datamigration/tests	942.276s
```

magodo:
Hi Kat, I rerun the test locally after apply your review comment, it works as expected. Not sure the cause of the error attached above.

katbyte:
@magodo, we run tests with `TF_SCHEMA_PANIC_ON_ERROR="1"` as it'll do more checks such as erroring when you try and set an invalid property

magodo:
@katbyte, I run the acc test via makefile of the project. I think it had already export `TF_SCHEMA_PANIC_ON_ERROR` beforehead. Would you please recheck? (maybe the error has something to do with the review comment you pointed previously, about set `Id` a second time?)

katbyte (changes requested):

katbyte:
Tests are passing for us not 🎉

azurerm/internal/services/datamigration/tests/resource_arm_data_migration_service_test.go
magodo: As far as i know, for normal test step (instead of import test step), it will always verify that `Plan()` is empty right after each `Apply()` and we don't have a perpetual diff issue. It will do this with TWO plans. One without a refresh.
So is there anything more a import step will check?

azurerm/internal/services/datamigration/resource_arm_data_migration_service.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return err
+	}
+	resourceGroup := id.ResourceGroup
+	name := id.Path["services"]
+
+	// For dms tasks created via terraform, the deletion order has already handled the dependency between dms task and dms service. In which case, dms service will not start to delete until
+	// dms tasks have been deleted.
+	// For dms tasks created outside of terraform, it is user's responsibility to ensure the deletion order. And for dms service, it makes sense to just force delete outstanding tasks.
+	toDeleteRunningTasks := true
```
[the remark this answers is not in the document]
magodo: Either way should work with proper documentation. But yes, your suggestion is more safe, so I'll change per your suggestion :)

azurerm/internal/services/datamigration/tests/resource_arm_data_migration_service_test.go
katbyte: Yes, it will ensure that a read results in the same data being read into the schema - ie all values are being read back accurately

katbyte (approved):
Thanks for the changes @magodo, LGTM 👍

katbyte (changes requested):
@magodo, looks like CI errors needs to be fixed up, but once they are thus should be gtg

magodo:
@katbyte 
The remaining CI error is because that the list of allowed resource providers doesn't contain this new resource provider.

azurerm/internal/services/databasemigration/resource_arm_database_migration_service.go
```diff
@@ -0,0 +1,227 @@
...
+			}
+		}
+		if existing.ID != nil && *existing.ID != "" {
+			return tf.ImportAsExistsError("azurerm_data_migration_service", *existing.ID)
+		}
+	}
+
+	skuName := d.Get("sku_name").(string)
+	subnetID := d.Get("subnet_id").(string)
+	location := d.Get("location").(string)
```
katbyte: Could we normalize the locaation: 
```suggestion
azure.NormalizeLocation(d.Get("location").(string))
```

azurerm/internal/services/databasemigration/resource_arm_database_migration_service.go
```diff
@@ -0,0 +1,227 @@
...
+
+	parameters := datamigration.Service{
+		Location: utils.String(location),
```
katbyte: _minor_ we can just do these inline if you'd like
```suggestion
		Location: utils.String(azure.NormalizeLocation(d.Get("location").(string))),
```

azurerm/internal/services/databasemigration/resource_arm_database_migration_service.go
```diff
@@ -0,0 +1,227 @@
...
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[INFO] Data Migration Service %q does not exist - removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error reading Data Migration Service (Service Name %q / Group Name %q): %+v", id.Name, id.ResourceGroup, err)
+	}
+
+	d.SetId(*resp.ID)
```
katbyte: Could we chheck for nil and that ID isn't empty?

azurerm/internal/services/databasemigration/tests/resource_arm_database_migration_service_test.go
```diff
@@ -0,0 +1,233 @@
...
+					resource.TestCheckResourceAttrSet(data.ResourceName, "subnet_id"),
+					resource.TestCheckResourceAttr(data.ResourceName, "sku_name", "Standard_1vCores"),
+					resource.TestCheckResourceAttr(data.ResourceName, "tags.name", "test"),
+				),
+			},
+			data.ImportStep(),
+		},
+	})
+}
+
```
katbyte: Could we also do an update test here where we go from basic -> complete -> basic?

azurerm/internal/services/databasemigration/tests/resource_arm_database_migration_service_test.go
```diff
@@ -0,0 +1,233 @@
...
+		return nil
+	}
+
+	return nil
+}
+
+func testAccAzureRMDatabaseMigrationService_base(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+resource "azurerm_resource_group" "test" {
```
katbyte: Could we add a provider block here?

azurerm/internal/services/databasemigration/tests/resource_arm_database_migration_service_test.go
```diff
@@ -0,0 +1,233 @@
...
+  name     = "acctestRG-dms-%d"
```
katbyte: Given this is DataBase migrations service, could we maybe use 
```suggestion
  name     = "acctestRG-dbms-%d"
```

WDYT?

website/docs/r/database_migration_service.html.markdown
```diff
@@ -0,0 +1,74 @@
...
+* `sku_name` - (Required) The sku name of the database migration service. Possible values are `Premium_4vCores`, `Standard_1vCores`, `Standard_2vCores` and `Standard_4vCores`. Changing this forces a new resource to be created.
+
+* `tags` - (Optional) A mapping of tags to assigned to the resource.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The ID of Database Migration Service.
+
```
katbyte: COuld we include the timeout documentation here?

katbyte (changes requested):
Thanks for the revisions @magodo,

I've left some more (mostly) minor comments inline that one addressed this should be good to merge!

katbyte (approved):
Thanks for those changes @magodo. LGTM now 👍

PR #5262, opened 2019-12-26
provider - fix various test failures from the refactoring

katbyte:


azurerm/internal/services/healthcare/tests/resource_arm_healthcare_service_test.go
```diff
@@ -146,7 +149,7 @@ resource "azurerm_healthcare_service" "test" {
     "${data.azurerm_client_config.current.service_principal_object_id}",
   ]
 }
-`, data.RandomInteger, location, data.RandomInteger)
+`, data.RandomInteger, location, data.RandomInteger/10) //name can only be 24 chars long
```
tombuildsstuff: could we use RandomString here instead:

```suggestion
`, data.RandomInteger, location, data.RandomString)
```

azurerm/internal/services/healthcare/tests/resource_arm_healthcare_service_test.go
```diff
@@ -206,5 +209,5 @@ resource "azurerm_healthcare_service" "test" {
     allow_credentials  = true
   }
 }
-`, data.RandomInteger, location, data.RandomInteger)
+`, data.RandomInteger, location, data.RandomInteger/10) //name can only be 24 chars long
```
tombuildsstuff: (same here)

azurerm/internal/services/healthcare/tests/resource_arm_healthcare_service_test.go
katbyte: I don't see the benefit from using random string here? this works and retains the date information from random int

PR #5278, opened 2019-12-31
r/kubernetes_cluster: renaming `managed_cluster_identity` to `identity`

tombuildsstuff:
Making this consistent with the other resources, which all use `identity` as the name of the MSI block

This hasn't shipped yet - thus this isn't a breaking change


tombuildsstuff:
Tests look good 👍

PR #5293, opened 2020-01-02
Fix example casing, add some defaults

drdamour:


website/docs/r/app_service_certificate_order.html.markdown
```diff
@@ -25,7 +25,7 @@ resource "azurerm_app_service_certificate_order" "example" {
   resource_group_name = "${azurerm_resource_group.example.name}"
   location            = "global"
   distinguished_name  = "CN=example.com"
-  product_type        = "standard"
+  product_type        = "Standard"
```
drdamour: was already cased correctly in docs. got this error
```
Error: Error setting `product_type` for App Service Certificate Order "usos1cer01-inmp-dev" (Resource Group "usos1rgp01-certs"), either `Standard` or `WildCard`
```

when using all lower case.

website/docs/r/app_service_certificate_order.html.markdown
```diff
@@ -25,7 +25,7 @@ resource "azurerm_app_service_certificate_order" "example" {
   resource_group_name = "${azurerm_resource_group.example.name}"
   location            = "global"
```
drdamour: this global location confused me...is this some tf doc practice, or does it have special meaning, or is it just a random val to be different than the resource group?

drdamour: ah i just tried some other region and sure enough it must be global 
```
Error: Error creating/updating App Service Certificate Order "usos1cer01-inmp-dev" (Resource Group "usos1rgp01-certs"): web.AppServiceCertificateOrdersClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="LocationNotAvailableForResourceType" Message="The provided location 'westus2' is not available for resource type 'Microsoft.CertificateRegistration/certificateOrders'. List of available regions for the resource type is 'global'
```

drdamour:
thanks @katbyte how long does it take to roll out?

Issue #5044, opened 2019-12-02
New resources: azurerm_api_management_identity_provider_xxx


molinch:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently it does not seem to be possible to create identity providers for API management
This part was mentioned in https://github.com/terraform-providers/terraform-provider-azurerm/issues/1177 but it seems like it got forgotten

![image](https://user-images.githubusercontent.com/5130190/69978835-05975b00-1525-11ea-8ebf-be708825386f.png)

### New or Affected Resource(s)

- azurerm_api_management_identity_provider_aad
- azurerm_api_management_identity_provider_aad_b2c
- so on so forth, there is also Facebook, Google, Microsoft, Twitter, Username/password
However I believe the most important ones, for enterprise usages, are aad and aad b2c. The others are just nice to have.

### Potential Terraform Configuration

For azurerm_api_management_identity_provider_aad it could be:
```hcl
resource "azurerm_api_management_identity_provider" "api_management_identity_provider" {
  client_id = "123456789"
  client_secret = "*******"
  signin_tenant = "mytenant"
  allowed_tenants = [ "mytenant1", "mytenant2" ]
}
```

fsettele:
The provider Username/password is not a real provider. It's managed by `properties.enabled` of type https://docs.microsoft.com/en-us/rest/api/apimanagement/2019-12-01/signupsettings/createorupdate#portalsignupsettings. Which means, when adding the following snippet to the `azurerm_api_management ` resource definition it enables the Username/password provider for the API Management service.
```
resource "azurerm_api_management" "apim" {
...
  sign_up {
    enabled = true # This enables the Username and Password Identity Provider
    terms_of_service {
      consent_required = false
      enabled = false
    }
...
}
```

favoretti:
Since this issue seems to have been addressed in the latest versions of the provider (or a valid workaround was provided) - I'm going to close it. Please open a new updated bug report if this is still relevant. Thank you.

PR #5306, opened 2020-01-04
New Resource: `azurerm_api_management_identity_provider_twitter`

aqche:
Partially addresses: #5044

Adds the `azurerm_api_management_identity_provider_twitter` resource.

```
--- PASS: TestAccAzureRMApiManagementIdentityProviderTwitter_update (1802.14s)
--- PASS: TestAccAzureRMApiManagementIdentityProviderTwitter_basic (1858.36s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/apimanagement/tests 1858.456s
```


azurerm/internal/services/apimanagement/resource_arm_api_management_identity_provider_twitter.go
```diff
@@ -0,0 +1,159 @@
...
+	resourceGroup := d.Get("resource_group_name").(string)
+	serviceName := d.Get("api_management_name").(string)
+	clientID := d.Get("api_key").(string)
+	clientSecret := d.Get("api_secret_key").(string)
+
+	if features.ShouldResourcesBeImported() && d.IsNewResource() {
+		existing, err := client.Get(ctx, resourceGroup, serviceName, apimanagement.Twitter)
+		if err != nil {
+			if !utils.ResponseWasNotFound(existing.Response) {
+				return fmt.Errorf("Error checking for presence of existing Identity Provider %q (API Management Service %q / Resource Group %q): %s", apimanagement.Twitter, serviceName, resourceGroup, err)
```
katbyte: Could we add the type to all wrapped errors:
```suggestion
				return fmt.Errorf("Error checking for presence of existing Twitter Identity Provider %q (API Management Service %q / Resource Group %q): %s", apimanagement.Twitter, serviceName, resourceGroup, err)
```

aqche: The first `%q` in the format string uses the `apimanagement.Twitter` value so the resulting error message will note that it is the Twitter Identity Provider for that APIM. Is that sufficient without applying your suggested change?

katbyte: Oh i didn't notice that at all, yep that'll do quite nicely! thanks

PR #5307, opened 2020-01-04
New Data Source/Resource: `azurerm_dedicated_host_group`

magodo:
Add new resource `dedicated_host_group`, which belongs to Compute Provider.

> A host group is a resource that represents a collection of dedicated hosts. You create a host group in a region and an availability zone, and add hosts to it.

Please refer to this [link](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/dedicated-hosts#groups-hosts-and-vms) for more info.

### DIFF
--- a/azurerm/internal/services/compute/data_source_dedicated_host_group.go
+++ b/azurerm/internal/services/compute/data_source_dedicated_host_group.go
@@ -3,6 +3,7 @@ package compute
 import (
 	"fmt"
 	"regexp"
+	"time"
 
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
 
@@ -18,6 +19,10 @@ import (
 func dataSourceArmDedicatedHostGroup() *schema.Resource {
 	return &schema.Resource{
 		Read: dataSourceArmDedicatedHostGroupRead,
+
+		Timeouts: &schema.ResourceTimeout{
+			Read: schema.DefaultTimeout(5 * time.Minute),
+		},
 
 		Schema: map[string]*schema.Schema{
 			"name": {
--- a/azurerm/internal/services/compute/data_source_dedicated_host_group.go
+++ b/azurerm/internal/services/compute/data_source_dedicated_host_group.go
@@ -1,0 +1,98 @@
+package compute
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmDedicatedHostGroup() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmDedicatedHostGroupRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringMatch(regexp.MustCompile(`^[^_\W][\w-.]{0,78}[\w]$`), ""),
+			},
+
+			"location": azure.SchemaLocationForDataSource(),
+
+			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
+
+			"host_ids": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"platform_fault_domain_count": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
+			"zones": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"tags": tags.SchemaDataSource(),
+		},
+	}
+}
+
+func dataSourceArmDedicatedHostGroupRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Compute.DedicatedHostGroupsClient
+	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	name := d.Get("name").(string)
+	resourceGroupName := d.Get("resource_group_name").(string)
+
+	resp, err := client.Get(ctx, resourceGroupName, name)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return fmt.Errorf("Error: Dedicated Host Group %q (Resource Group %q) was not found", name, resourceGroupName)
+		}
+		return fmt.Errorf("Error reading Dedicated Host Group %q (Resource Group %q): %+v", name, resourceGroupName, err)
+	}
+
+	d.SetId(*resp.ID)
+
+	d.Set("name", name)
+	d.Set("resource_group_name", resourceGroupName)
+	if location := resp.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	if props := resp.DedicatedHostGroupProperties; props != nil {
+		if props.Hosts != nil {
+			var hostIds []string
+			for _, item := range *props.Hosts {
+				hostIds = append(hostIds, *item.ID)
+			}
+			if err := d.Set("host_ids", hostIds); err != nil {
+				return fmt.Errorf("Error setting `hosts`: %+v", err)
+			}
+		}
+		d.Set("platform_fault_domain_count", int(*props.PlatformFaultDomainCount))
+	}
+
+	d.Set("zones", utils.FlattenStringSlice(resp.Zones))
+
+	return nil
+}
--- a/azurerm/internal/services/compute/data_source_dedicated_host_group.go
+++ b/azurerm/internal/services/compute/data_source_dedicated_host_group.go
@@ -30,14 +30,6 @@ func dataSourceArmDedicatedHostGroup() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
 
-			"host_ids": {
-				Type:     schema.TypeList,
-				Computed: true,
-				Elem: &schema.Schema{
-					Type: schema.TypeString,
-				},
-			},
-
 			"platform_fault_domain_count": {
 				Type:     schema.TypeInt,
 				Computed: true,
@@ -80,19 +72,10 @@ func dataSourceArmDedicatedHostGroupRead(d *schema.ResourceData, meta interface{
 		d.Set("location", azure.NormalizeLocation(*location))
 	}
 	if props := resp.DedicatedHostGroupProperties; props != nil {
-		if props.Hosts != nil {
-			var hostIds []string
-			for _, item := range *props.Hosts {
-				hostIds = append(hostIds, *item.ID)
-			}
-			if err := d.Set("host_ids", hostIds); err != nil {
-				return fmt.Errorf("Error setting `hosts`: %+v", err)
-			}
-		}
 		d.Set("platform_fault_domain_count", int(*props.PlatformFaultDomainCount))
 	}
 
 	d.Set("zones", utils.FlattenStringSlice(resp.Zones))
 
-	return nil
+	return tags.FlattenAndSet(d, resp.Tags)
 }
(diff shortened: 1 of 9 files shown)


azurerm/internal/services/compute/data_source_dedicated_host_group.go
```diff
@@ -0,0 +1,98 @@
...
+		if props.Hosts != nil {
+			var hostIds []string
+			for _, item := range *props.Hosts {
+				hostIds = append(hostIds, *item.ID)
+			}
+			if err := d.Set("host_ids", hostIds); err != nil {
+				return fmt.Errorf("Error setting `hosts`: %+v", err)
+			}
+		}
+		d.Set("platform_fault_domain_count", int(*props.PlatformFaultDomainCount))
```
tombuildsstuff: Azure returns the representation of a resource at the last time it was modified, thus this could be nil - as such we need to make this:

```suggestion
		platformFaultDomainCount := 0
		if props.PlatformFaultDomainCount != nil {
			platformFaultDomainCount = int(*props.PlatformFaultDomainCount)
		}
		d.Set("platform_fault_domain_count", platformFaultDomainCount)
```

azurerm/internal/services/compute/data_source_dedicated_host_group.go
```diff
@@ -0,0 +1,98 @@
...
+
+			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
+
+			"host_ids": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
```
tombuildsstuff: this value's too dynamic to be useful in a data source - can we remove this?

azurerm/internal/services/compute/data_source_dedicated_host_group.go
```diff
@@ -0,0 +1,98 @@
...
+			if err := d.Set("host_ids", hostIds); err != nil {
+				return fmt.Errorf("Error setting `hosts`: %+v", err)
+			}
+		}
+		d.Set("platform_fault_domain_count", int(*props.PlatformFaultDomainCount))
+	}
+
+	d.Set("zones", utils.FlattenStringSlice(resp.Zones))
+
+	return nil
```
tombuildsstuff: there's tags here too so this wants to be:

```suggestion
	return tags.FlattenAndSet(d, resp.Tags)
```

azurerm/internal/services/compute/resource_arm_dedicated_host_group.go
```diff
@@ -0,0 +1,203 @@
...
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.StringMatch(regexp.MustCompile(`^[1-9]\d*$`), ""),
+				},
+				// Currently only one endpoint is allowed.
+				// we'll leave this open to enhancement when they add multiple zones support.
+				MaxItems: 1,
+			},
```
tombuildsstuff: we can use `SingleZoneSchema` here instead

azurerm/internal/services/compute/resource_arm_dedicated_host_group.go
```diff
@@ -0,0 +1,203 @@
...
+		return fmt.Errorf("Error reading Dedicated Host Group %q (Resource Group %q): %+v", name, resourceGroupName, err)
+	}
+
+	d.Set("name", name)
+	d.Set("resource_group_name", resourceGroupName)
+	if location := resp.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	if props := resp.DedicatedHostGroupProperties; props != nil {
+		d.Set("platform_fault_domain_count", int(*props.PlatformFaultDomainCount))
```
tombuildsstuff: (as above) can we nil-check this and ensure we set a default value if this is nil here?

azurerm/internal/services/compute/tests/resource_arm_dedicated_host_group_test.go
```diff
@@ -0,0 +1,180 @@
...
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMDedicatedHostGroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMDedicatedHostGroup_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMDedicatedHostGroupExists(data.ResourceName),
+				),
+			},
+			data.RequiresImportErrorStep(func(data acceptance.TestData) string {
```
tombuildsstuff: since they have the same method signature we can simplify this to:

```suggestion
			data.RequiresImportErrorStep(testAccAzureRMDedicatedHostGroup_requiresImport),
```

azurerm/internal/services/compute/tests/resource_arm_dedicated_host_group_test.go
```diff
@@ -0,0 +1,180 @@
...
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-compute-%d"
+  location = "%s"
+}
+
+resource "azurerm_dedicated_host_group" "test" {
+  resource_group_name 			= "${azurerm_resource_group.test.name}"
+  name 							= "acctestDHG-compute-%s"
+  location            			= "${azurerm_resource_group.test.location}"
+  platform_fault_domain_count 	= 2
```
tombuildsstuff: _minor_ could we make this spacing consistent here?

website/docs/d/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,52 @@
+---
+subcategory: "Compute"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_dedicated_host_group"
+sidebar_current: "docs-azurerm-datasource-dedicated-host-group"
```
tombuildsstuff: this is no longer needed:

```suggestion
```

website/docs/d/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+data "azurerm_dedicated_host_group" "example" {
+  resource_group_name = "example-rg"
+  name                = "example-dedicated-host-group" 
+}
+
+output "id" {
+  value = data.azurerm_dedicated_host_group.example.id
+}
+```
+
```
tombuildsstuff: we can remove this extra whitespace:

```suggestion
```

website/docs/d/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+page_title: "Azure Resource Manager: azurerm_dedicated_host_group"
+sidebar_current: "docs-azurerm-datasource-dedicated-host-group"
+description: |-
+  Gets information about an existing Dedicated Host Group
+---
+
+# Data Source: azurerm_dedicated_host_group
+
+Use this data source to access information about an existing Dedicated Host Group.
+
```
tombuildsstuff: we can remove this extra whitespace:

```suggestion
```

website/docs/d/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The ID of the Dedicated Host Group.
+
+* `location` - The Azure location where the Dedicated Host Group exists.
+
+* `host_ids` - A list of ID of all dedicated hosts in the Dedicated Host Group.
```
tombuildsstuff: as above - we can remove this:

```suggestion
```

website/docs/d/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+
+* `platform_fault_domain_count` - Number of fault domains that the Dedicated Host Group can span.
+
+* `zones` - Availability Zone the Dedicated Host Group is scoped. Only single zone is supported. If it is absent, it means the group is using all zones in the region.
```
tombuildsstuff: since this is a Data Source this is read-only:

```suggestion
* `zones` - The Availability Zones in which this Dedicated Host Group is located
```

website/docs/r/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,59 @@
+---
+subcategory: "Compute"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_dedicated_host_group"
+sidebar_current: "docs-azurerm-resource-dedicated-host-group"
+description: |-
+  Manage Azure DedicatedHostGroup instance.
```
tombuildsstuff:
```suggestion
  Manages a Dedicated Host Group
```

website/docs/r/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+---
+
+# azurerm_dedicated_host_group
+
+Manage a Azure Dedicated Host Group instance.
```
tombuildsstuff:
```suggestion
Manages a Dedicated Host Group
```

website/docs/r/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,59 @@
+---
+subcategory: "Compute"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_dedicated_host_group"
+sidebar_current: "docs-azurerm-resource-dedicated-host-group"
```
tombuildsstuff:
```suggestion
```

website/docs/d/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+
+The following attributes are exported:
+
+* `id` - The ID of the Dedicated Host Group.
+
+* `location` - The Azure location where the Dedicated Host Group exists.
+
+* `host_ids` - A list of ID of all dedicated hosts in the Dedicated Host Group.
+
+* `platform_fault_domain_count` - Number of fault domains that the Dedicated Host Group can span.
```
tombuildsstuff:
```suggestion
* `platform_fault_domain_count` - The number of fault domains the Dedicated Host Group spans.
```

website/docs/r/dedicated_host_group.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+
+* `name` - (Required) Specifies the name of the Dedicated Host Group. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) Specifies the name of the resource group the Dedicated Host Group is located in. Changing this forces a new resource to be created.
+
+* `location` - (Required) The Azure location where the Dedicated Host Group exists. Changing this forces a new resource to be created.
+
+* `platform_fault_domain_count` - (Required) Number of fault domains that the host group can span. Changing this forces a new resource to be created.
+
+* `zones` - (Optional) Availability Zone to use for this host group. Only single zone is supported. The zone can be assigned only during creation. If not provided, the group supports all zones in the region. If provided, enforces each host in the group to be in the same zone. Changing this forces a new resource to be created.
```
tombuildsstuff:
```suggestion
* `zones` - (Optional) A list of Availability Zones in which the Dedicated Host Group should be located

-> **NOTE:** At this time only a single zone is supported.
```

azurerm/internal/services/compute/data_source_dedicated_host_group.go
magodo: I didn't quite understand the point here. Would you please kindly provide more clue about in which case this might be null?

azurerm/internal/services/compute/data_source_dedicated_host_group.go
magodo: Agree!

azurerm/internal/services/compute/resource_arm_dedicated_host_group.go
magodo: What if azure supports multiple zones later? Then should we introduce a breaking change?

azurerm/internal/services/compute/data_source_dedicated_host_group.go
katbyte: if you don't do this the property won't exist. we need to make sure we set it to something

azurerm/internal/services/compute/resource_arm_dedicated_host_group.go
katbyte: No, it would still be the same name and type, just a different min/max item.

magodo: Ooh, I mis-understood the comment 😨

azurerm/internal/services/compute/data_source_dedicated_host_group.go
magodo: Do you mean the property will be missing in response? Why didn't we caught this fault in acctest?

magodo: Though I think this certain field is intrinsincly present in response, for consistence, a protection still makes sense here.

tombuildsstuff:
Tests pass:

Screenshot 2020-01-09 at 12 24 25

magodo:
@tombuildsstuff Thank you 🎉

PR #5316, opened 2020-01-06
azurerm_storage_account - fix performance issue for accounts that don't support queues 

magodo:
Not all storage account supports queue. Without this condition, user with unsupported kind will suffer a block (for about 10min) while reading the resource.

### DIFF
--- a/azurerm/internal/services/storage/resource_arm_storage_account.go
+++ b/azurerm/internal/services/storage/resource_arm_storage_account.go
@@ -1240,6 +1240,10 @@ func resourceArmStorageAccountRead(d *schema.ResourceData, meta interface{}) err
 	}
 
 	// queue is only available for certain tier and kind (as specified below)
+	if resp.Sku == nil {
+		return fmt.Errorf("Error retrieving Storage Account %q (Resource Group %q): `sku` was nil", name, resourceGrup)
+	}
+	
 	if resp.Sku.Tier == storage.Standard {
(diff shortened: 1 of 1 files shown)


azurerm/internal/services/storage/resource_arm_storage_account.go
```diff
@@ -1239,20 +1239,25 @@ func resourceArmStorageAccountRead(d *schema.ResourceData, meta interface{}) err
 		}
 	}
 
-	queueClient, err := storageClient.QueuesClient(ctx, *account)
-	if err != nil {
-		return fmt.Errorf("Error building Queues Client: %s", err)
-	}
+	// queue is only available for certain tier and kind (as specified below)
+	if resp.Sku.Tier == storage.Standard {
```
tombuildsstuff: resp.Sku is a nilable object here, so we'd need to nil-check this first:

```suggestion
	if resp.Sku == nil {
		return fmt.Errorf("Error retrieving Storage Account %q (Resource Group %q): `sku` was nil", name, resourceGrup)
	}
	
	if resp.Sku.Tier == storage.Standard {
```

magodo: Good to know 😉

tombuildsstuff:
@magodo thanks for pushing those changes - I hope you don't mind but so that we can get this merged I've pushed a commit to resolve the linting issue :)

PR #5330, opened 2020-01-07
Added app_service_hybrid_connection resource and basic creation test

jackofallops:
plenty left to do (docs, test coverage) - but functional on a basic level

### DIFF
--- a/azurerm/internal/services/web/registration.go
+++ b/azurerm/internal/services/web/registration.go
@@ -34,5 +34,6 @@ func (r Registration) SupportedResources() map[string]*schema.Resource {
 		"azurerm_app_service_virtual_network_swift_connection": resourceArmAppServiceVirtualNetworkSwiftConnection(),
 		"azurerm_app_service":                                  resourceArmAppService(),
 		"azurerm_function_app":                                 resourceArmFunctionApp(),
+		"azurerm_app_service_hybrid_connection":                resourceArmAppServiceHybridConnection(),
 	}
 }
--- a/azurerm/internal/services/web/registration.go
+++ b/azurerm/internal/services/web/registration.go
@@ -28,12 +28,12 @@ func (r Registration) SupportedResources() map[string]*schema.Resource {
 		"azurerm_app_service_certificate":                      resourceArmAppServiceCertificate(),
 		"azurerm_app_service_certificate_order":                resourceArmAppServiceCertificateOrder(),
 		"azurerm_app_service_custom_hostname_binding":          resourceArmAppServiceCustomHostnameBinding(),
+		"azurerm_app_service_hybrid_connection":                resourceArmAppServiceHybridConnection(),
 		"azurerm_app_service_plan":                             resourceArmAppServicePlan(),
 		"azurerm_app_service_slot":                             resourceArmAppServiceSlot(),
 		"azurerm_app_service_source_control_token":             resourceArmAppServiceSourceControlToken(),
 		"azurerm_app_service_virtual_network_swift_connection": resourceArmAppServiceVirtualNetworkSwiftConnection(),
 		"azurerm_app_service":                                  resourceArmAppService(),
 		"azurerm_function_app":                                 resourceArmFunctionApp(),
-		"azurerm_app_service_hybrid_connection":                resourceArmAppServiceHybridConnection(),
 	}
 }
--- a/azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
+++ b/azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
@@ -1,0 +1,186 @@
+package web
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/go-azure-helpers/response"
+
+	"github.com/Azure/azure-sdk-for-go/services/web/mgmt/2018-02-01/web"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func resourceArmAppServiceHybridConnection() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceArmAppServiceHybridConnectionCreateUpdate,
+		Read:   resourceArmAppServiceHybridConnectionRead,
+		Update: resourceArmAppServiceHybridConnectionCreateUpdate,
+		Delete: resourceArmAppServiceHybridConnectionDelete,
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"app_service_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"resource_group_name": azure.SchemaResourceGroupName(),
+			"namespace_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringLenBetween(6, 50),
+			},
+			"relay_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"relay_arm_uri": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: azure.ValidateResourceID,
+			},
+			"hostname": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"port": {
+				Type:         schema.TypeInt,
+				Required:     true,
+				ValidateFunc: validate.PortNumberOrZero,
+			},
+			"service_bus_namespace": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"service_bus_suffix": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"send_key_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"send_key_value": {
+				Type:      schema.TypeString,
+				Required:  true,
+				Sensitive: true,
+			},
+		},
+	}
+}
+
+func resourceArmAppServiceHybridConnectionCreateUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Web.AppServicesClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	name := d.Get("app_service_name").(string)
+	resGroup := d.Get("resource_group_name").(string)
+	namespaceName := d.Get("namespace_name").(string)
+	relayName := d.Get("relay_name").(string)
+	relayNameArmUri := d.Get("relay_arm_uri").(string)
+	hostname := d.Get("hostname").(string)
+	port := int32(d.Get("port").(int))
+	serviceBusSuffix := d.Get("service_bus_suffix").(string)
+	serviceBusNamespace := d.Get("service_bus_namespace").(string)
+	sendKeyName := d.Get("send_key_name").(string)
+	sendKeyValue := d.Get("send_key_value").(string)
+
+	hybridConnectionProperties := web.HybridConnectionProperties{
+		ServiceBusNamespace: &serviceBusNamespace,
+		RelayName:           &relayName,
+		RelayArmURI:         &relayNameArmUri,
+		Hostname:            &hostname,
+		Port:                &port,
+		SendKeyName:         &sendKeyName,
+		SendKeyValue:        &sendKeyValue,
+		ServiceBusSuffix:    &serviceBusSuffix,
+	}
+
+	connectionEnvelope := web.HybridConnection{}
+	connectionEnvelope.HybridConnectionProperties = &hybridConnectionProperties
+
+	if _, err := client.CreateOrUpdateHybridConnection(ctx, resGroup, name, namespaceName, relayName, connectionEnvelope); err != nil {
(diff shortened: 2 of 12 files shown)


azurerm/internal/services/web/registration.go
```diff
@@ -34,5 +34,6 @@ func (r Registration) SupportedResources() map[string]*schema.Resource {
 		"azurerm_app_service_virtual_network_swift_connection": resourceArmAppServiceVirtualNetworkSwiftConnection(),
 		"azurerm_app_service":                                  resourceArmAppService(),
 		"azurerm_function_app":                                 resourceArmFunctionApp(),
+		"azurerm_app_service_hybrid_connection":                resourceArmAppServiceHybridConnection(),
```
katbyte: Could we order this list alphabetically?

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
```diff
@@ -0,0 +1,186 @@
...
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"app_service_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
katbyte: Could we validate this? preferably with regex and a function to share

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
```diff
@@ -0,0 +1,186 @@
...
+			"resource_group_name": azure.SchemaResourceGroupName(),
+			"namespace_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringLenBetween(6, 50),
```
katbyte: Are any and all characters valid?

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
```diff
@@ -0,0 +1,186 @@
...
+		RelayArmURI:         &relayNameArmUri,
+		Hostname:            &hostname,
+		Port:                &port,
+		SendKeyName:         &sendKeyName,
+		SendKeyValue:        &sendKeyValue,
+		ServiceBusSuffix:    &serviceBusSuffix,
+	}
+
+	connectionEnvelope := web.HybridConnection{}
+	connectionEnvelope.HybridConnectionProperties = &hybridConnectionProperties
```
katbyte: This can be simplified to:
```
	 web.HybridConnection{
		HybridConnectionProperties: & web.HybridConnectionProperties{
			ServiceBusNamespace: utils.String(d.Get("service_bus_namespace").(string)),
			RelayName:           utils.String(d.Get("relay_name").(string)),
			...
		}
	}
```

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
```diff
@@ -0,0 +1,186 @@
...
+	namespaceName := id.Path["hybridConnectionNamespaces"]
+	relayName := id.Path["relays"]
+
+	resp, err := client.DeleteHybridConnection(ctx, resGroup, name, namespaceName, relayName)
+	if err != nil {
+		if response.WasNotFound(resp.Response) {
+			return nil
+		}
+		return err
+	}
```
katbyte: Could we simplify this (and wrap the error):
```
	if err != nil {
		if !response.WasNotFound(resp.Response) {
			return fmt.Errorf("Error deleting App Service Hybrid Connection%q (Resource Group %q, Relay %q): %+v", name, resGroup, relayName, err)
		}
	}
```

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
```diff
@@ -0,0 +1,186 @@
...
+func resourceArmAppServiceHybridConnectionDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Web.AppServicesClient
+	ctx, cancel := timeouts.ForDelete(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	id, err := azure.ParseAzureResourceID(d.Id())
+	if err != nil {
+		return err
+	}
+	resGroup := id.ResourceGroup
```
katbyte: for consistency could we make this
```suggestion
	resourceGroup := id.ResourceGroup
```

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
```diff
@@ -0,0 +1,186 @@
...
+func resourceArmAppServiceHybridConnectionRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Web.AppServicesClient
+	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	id, err := azure.ParseAzureResourceID(d.Id())
+	if err != nil {
+		return err
+	}
+	resGroup := id.ResourceGroup
```
katbyte: for consistency could we make this 
```suggestion
	resourceGroup := id.ResourceGroup
```

azurerm/internal/services/web/tests/resource_arm_app_service_hybrid_connection_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+resource "azurerm_app_service" "test" {
+  name                = "acctestAS-%d"
+  location            = azurerm_resource_group.test.location
+  resource_group_name = azurerm_resource_group.test.name
+  app_service_plan_id = azurerm_app_service_plan.test.id
+}
+
+resource "azurerm_relay_namespace" "test" {
+  name                = "acctestrn-%d"
```
katbyte: Could we try and use capitals here (if the resource allows it) - this is to ensure the API handles casing correctly.
```suggestion
  name                = "acctest-RN-%d"
```

azurerm/internal/services/web/tests/resource_arm_app_service_hybrid_connection_test.go
```diff
@@ -0,0 +1,147 @@
...
+  location            = azurerm_resource_group.test.location
+  resource_group_name = azurerm_resource_group.test.name
+
+  sku_name = "Standard"
+}
+
+resource "azurerm_relay_hybrid_connection" "test" {
+	name                 = "acctestrnhc-%d"
```
katbyte: And here
```suggestion
	name                 = "acctest-RHC-%d"
```

azurerm/internal/services/web/tests/resource_arm_app_service_hybrid_connection_test.go
```diff
@@ -0,0 +1,147 @@
...
+	resource_group_name  = azurerm_resource_group.test.name
+	relay_namespace_name = azurerm_relay_namespace.test.name
+	user_metadata        = "metadatatest"
+  }
+
+resource "azurerm_servicebus_namespace" "test" {
+  name                = "acctestservicebusnamespace-%d"
```
katbyte:
```suggestion
  name                = "acctest-SBN-%d"
```

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
```diff
@@ -0,0 +1,186 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringLenBetween(6, 50),
+			},
+			"relay_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"relay_arm_uri": {
```
katbyte: Given that this actually takes the ID would it make sense to rename it?
```suggestion
			"relay_id": {
```

and mention in the docs it maps to the arm uri property

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
```diff
@@ -0,0 +1,186 @@
...
+				ForceNew: true,
+			},
+			"resource_group_name": azure.SchemaResourceGroupName(),
+			"namespace_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringLenBetween(6, 50),
+			},
+			"relay_name": {
```
katbyte: Given we are getting the relay ID below, we could extract the name from it instead?

azurerm/internal/services/web/registration.go
jackofallops: Yep, not a problem. Will be in next push.

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
jackofallops: I've found no documentation to say otherwise, only documented validation I found is length.  I'll try some obscure chars and see what the API response comes back with.

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
jackofallops: Will do.  Except `relayName` as that's used again later for `CreateOrUpdateHybridConnection`

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
jackofallops: I was torn on this one, the API refers to it unhelpfully as relayArmURI, so anyone following the documentation would be (probably?) expecting this. My thought was to leave this and put appropriate comment in the doc page for it.

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
katbyte: hopefully its one of the nice APIs that actually return an error with the regex required

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
katbyte: We often slightly rename or correct the API to what a user would expect, `relay_id` is really clear to someone in autocomplete or a reading the tf, and makes  `relay_id = relat.test.id` make sense, while uri is only helpful to those coming from the API, in which case having that in the docs would assist them (and i think the latter case would be the les common)

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
```diff
@@ -0,0 +1,205 @@
...
+				ValidateFunc: azure.ValidateResourceID,
+			},
+			"namespace_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"relay_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
katbyte: Could we move all computed properties to the bottom? generally we go required -> optional -> computed

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
```diff
@@ -0,0 +1,205 @@
...
+
+	namespaceName := relayId.Path["namespaces"]
+	if namespaceName == "" {
+		return fmt.Errorf("Failed to parse relay namespace from ID: %s", relayArmURI)
+	}
+
+	relayName := relayId.Path["hybridConnections"]
+	if relayName == "" {
+		return fmt.Errorf("Failed to parse relay name from ID: %s", relayArmURI)
+	}
```
katbyte: So since you added the checks you might as add a parse ID function for that resource! We are planning on slowly creating an ID and parse function for each resource ID so we get these checks & import id validation. Here is an example in the code right now:

```
type AppServiceResourceID struct {
	ResourceGroup string
	Name          string
}

func ParseAppServiceID(input string) (*AppServiceResourceID, error) {
	id, err := azure.ParseAzureResourceID(input)
	if err != nil {
		return nil, fmt.Errorf("[ERROR] Unable to parse App Service ID %q: %+v", input, err)
	}

	appService := AppServiceResourceID{
		ResourceGroup: id.ResourceGroup,
	}

	if appService.Name, err = id.PopSegment("sites"); err != nil {
		return nil, err
	}

	if err := id.ValidateNoEmptySegments(input); err != nil {
		return nil, err
	}

	return &appService, nil
}
```

this will also simplify things elsewhere

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
```diff
@@ -0,0 +1,205 @@
...
+		RelayArmURI:         &relayArmURI,
+		Hostname:            utils.String(d.Get("hostname").(string)),
+		Port:                &port,
+		SendKeyName:         utils.String(d.Get("send_key_name").(string)),
+		SendKeyValue:        utils.String(d.Get("send_key_value").(string)),
+		ServiceBusSuffix:    utils.String(d.Get("service_bus_suffix").(string)),
+	}
+
+	connectionEnvelope := web.HybridConnection{}
+	connectionEnvelope.HybridConnectionProperties = &hybridConnectionProperties
```
katbyte: This could still become
```
	connectionEnvelope := web.HybridConnection{
		HybridConnectionProperties = &hybridConnectionProperties
```

or more consistent with the rest of the provider:

```
	connectionEnvelope := web.HybridConnection{
		HybridConnectionProperties = &web.HybridConnectionProperties{
		ServiceBusNamespace: utils.String(d.Get("service_bus_namespace").(string)),
...
```

azurerm/internal/services/web/tests/resource_arm_app_service_hybrid_connection_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func TestAccAzureRMAppServiceHybridConnection_basic(t *testing.T) {
```
tombuildsstuff: we need to add support for & a requires import test here - see other resources for an example (this forms part of #2807)

jackofallops:
Bug report opened on the API: https://github.com/Azure/azure-rest-api-specs/issues/8108
This blocks completion as property being set by the API is incorrect based on the request, preventing the acceptance test passing.

azurerm/internal/services/web/resource_arm_app_service_hybrid_connection.go
jackofallops: also getting the relay_namespace now too.

tombuildsstuff:
Since this [is currently blocked on a bug in the Azure API](https://github.com/Azure/azure-rest-api-specs/issues/8108) - I'm going to close this PR for the moment - however we'll take another look once the upstream issue's been fixed

PR #5336, opened 2020-01-08
New data source: 'azurerm_iothub_dps'

brunhil:
Add the 'azurerm_iothub_dps' data source. 

I chose not to include the **SKU** property as I didn't think it was too critical for a data source. I imagine that the more important values would be the ones included. I can easily add it if you think it's required.


tombuildsstuff:
Test passes:

Screenshot 2020-01-08 at 08 04 07

PR #5337, opened 2020-01-08
provider: format acctest tf blocks again

katbyte:


jackofallops:
lgtm also (late to the party, still sorting out notifications)

PR #5339, opened 2020-01-08
r/kubernetes_cluster: switching back to parallel tests

tombuildsstuff:
Trialled switching back to regular (non-parallel'd) test cases to work around API performance issues, but this ultimately wasn't fruitful

### DIFF
--- a/azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
+++ b/azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
@@ -237,6 +237,28 @@ func resourceArmKubernetesCluster() *schema.Resource {
 				},
 			},
 
+			"service_principal": {
+				Type:     schema.TypeList,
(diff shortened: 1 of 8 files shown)


azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
```diff
@@ -237,6 +237,28 @@ func resourceArmKubernetesCluster() *schema.Resource {
 				},
 			},
 
+			"service_principal": {
```
katbyte: I think you'll want to revert this change?

tombuildsstuff: nah this is just a minor regrouping from where the `service_principal` block went from required -> optional -> required in the last PR and I didn't want travis to block the release - so this should be fine 👍

PR #5341, opened 2020-01-08
Add new test function `RandomIntOfLegth()` 

katbyte:


azurerm/internal/acceptance/data.go
```diff
@@ -82,3 +83,30 @@ func BuildTestData(t *testing.T, resourceType string, resourceLabel string) Test
 
 	return testData
 }
+
+func (td *TestData) RandomIntOfLength(len int) int {
```
tombuildsstuff: some resources (e.g. Storage/Windows VM's) require this to support 4-8 chars; so I think we could make this `MMYYR..`?

azurerm/internal/acceptance/data_test.go
```diff
@@ -0,0 +1,59 @@
+package acceptance
+
+import (
+	"testing"
+)
+
+// since this depends on GetAuthConfig which lives in this package
+// unfortunately this has to live in a different package to the other func
```
tombuildsstuff: this doesn't appear to be true?

azurerm/internal/acceptance/data.go
```diff
@@ -82,3 +83,30 @@ func BuildTestData(t *testing.T, resourceType string, resourceLabel string) Test
...
+	// len should not be
+	//  - greater then 18, longest a int can represent
+	//  - less then 8, as that gives us YYMMDDRR
+	if 8 > len || len > 18 {
+		panic(fmt.Sprintf("Invalid Test: RandomIntOfLength: len is not between 6 or 18 inclusive"))
```
tombuildsstuff: 6 or 8?

azurerm/internal/acceptance/data.go
```diff
@@ -82,3 +83,30 @@ func BuildTestData(t *testing.T, resourceType string, resourceLabel string) Test
...
+	r := td.RandomInteger % 100
+
+	// 18 - just return the int
+	if len >= 18 {
+		return td.RandomInteger
+	}
+
+	// 16-17 just strip off the last 1-2 digits
+	if len >= 16 {
+		return td.RandomInteger / int(math.Pow10(18-len))
```
tombuildsstuff: I think this could be simplified to:

```suggestion
		return td.RandomInteger[0:15]
```

azurerm/internal/acceptance/data.go
```diff
@@ -82,3 +83,30 @@ func BuildTestData(t *testing.T, resourceType string, resourceLabel string) Test
...
+	}
+
+	// 8-15 remove the last x digits
+	a := td.RandomInteger / int(math.Pow10(18-len+2))
+
+	// multiply by 100 and add last two digits of randomness back in
+	return (a * 100) + r
```
tombuildsstuff: I have a feeling this function could be simplified to:

```suggestion
	return td.RandomInteger[0:len]
```

azurerm/internal/acceptance/data.go
katbyte: results in a compilation error as its a int not a string

azurerm/internal/acceptance/data.go
katbyte: results in a compilation error as its a int not a string

tombuildsstuff: in which case we should be able to do:

```go
s := strconv.Itoa(123456)
v := s[0:15]
i, _ := strconv.Atoi(v)
return i
```

PR #5356, opened 2020-01-09
analysisservices - add ID parsing function

katbyte:


azurerm/internal/services/analysisservices/parse/analysis_services_server.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+)
+
+type AnalysisServicesServerId struct {
+	ResourceGroup string
+	Name          string
+}
+
+func AnalysisServicesServerID(input string) (*AnalysisServicesServerId, error) {
```
tombuildsstuff: can we add unit tests for this parser, similar to the other parsers?

PR #5363, opened 2020-01-10
Add ability to specify test timeout for Acceptance Tests

brunhil:
Based off @tombuildsstuff 's comment in Slack, I decided to document this and add the ability to set the test timeout.


brunhil:
I'm not the best at make files, took my inital stab at providing the service name.

PR #5382, opened 2020-01-12
azurerm_iothub & azurerm_iothub_dps: deprecate the sku.tier property

katbyte:
partially addresses #1500


WodansSon:
Should this be flattened to a `sku_name`?
### For example:
```
  sku_name = "S_1"
```
### Instead of:
```
sku { 
    name     = "S1" 
    capacity = "1" 
}
```

WDYT?

PR #5385, opened 2020-01-13
Upgrade azure-sdk-for-go to v38.1.0 and frontdoor to 2019-11-01

timja:
https://github.com/Azure/azure-sdk-for-go/releases


timja:
Acceptance tests are failing, looking into it

azurerm/internal/services/frontdoor/tests/resource_arm_front_door_firewall_policy_test.go
```diff
@@ -272,7 +272,7 @@ resource "azurerm_frontdoor_firewall_policy" "test" {
     }
   }
 
-  %s
+  %[3]s
```
timja: This test must have been failing before this PR, this was passing in the region text before, in my case "West Europe" was randomly being placed in the resource body

PR #5386, opened 2020-01-13
F/virtual machine ids

jackofallops:
Addition of VirtualMachineID and VirtualMachineExtensionID Parse & Validation

### DIFF
--- a/azurerm/internal/services/compute/virtual_machine_extension.go
+++ b/azurerm/internal/services/compute/virtual_machine_extension.go
@@ -1,0 +1,49 @@
+package compute
+
+import (
+	"fmt"
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+)
+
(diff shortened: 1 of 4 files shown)


azurerm/internal/services/compute/virtual_machine_extension.go
```diff
@@ -0,0 +1,49 @@
...
+		ResourceGroup: id.ResourceGroup,
+	}
+
+	if virtualMachineExtension.Name, err = id.PopSegment("extensions"); err != nil {
+		return nil, err
+	}
+
+	if virtualMachineExtension.VirtualMachine, err = id.PopSegment("virtualMachines"); err != nil {
+		return nil, err
+	}
```
tombuildsstuff: since the ID's built up of multiple key-vaule pairs, the ordering could be important here in time; should we make this:

a) parse VM Name
b) parse name?

PR #5393, opened 2020-01-14
F/virtual machine extension by ID fixes #2456

### LINKED ISSUE #2456: azurerm_virtual_machine_extension should also accept virtual_machine_id instead of virtual_machine_name
### Community Note

### Description

Currently the azure_virtual_machine_extension resource requires to specify the concerned VM with the parameter "virtual_machine_name". This is not consistent with other resources that require the "virtual_machine_id" parameter. Also, I think that it would be safer to use the id instead of the name.

When developing a module that has both a disk attachment and a vm extension, this is a bit cumbersome because it requires my module interface to require both the id and the name of the VM. One possible solution would also be to create a azure_virtual_machine_extension data source to resolve the name from the id (but still: having only ids would be a cleaner interface in my opinion).

* azure_virtual_machine_extension

jackofallops:
This change updates the VM extension resource to use the `virtual_machine_id` and deprecates the use of `virtual_machine_name`.

### DIFF
--- a/azurerm/internal/services/compute/registration.go
+++ b/azurerm/internal/services/compute/registration.go
@@ -45,7 +45,7 @@ func (r Registration) SupportedResources() map[string]*schema.Resource {
 		"azurerm_shared_image":                         resourceArmSharedImage(),
 		"azurerm_snapshot":                             resourceArmSnapshot(),
 		"azurerm_virtual_machine_data_disk_attachment": resourceArmVirtualMachineDataDiskAttachment(),
-		"azurerm_virtual_machine_extension":            resourceArmVirtualMachineExtensions(),
+		"azurerm_virtual_machine_extension":            resourceArmVirtualMachineExtension(),
 		"azurerm_virtual_machine_scale_set":            resourceArmVirtualMachineScaleSet(),
 		"azurerm_virtual_machine":                      resourceArmVirtualMachine(),
 	}
--- a/azurerm/internal/services/compute/resource_arm_virtual_machine_extension.go
+++ b/azurerm/internal/services/compute/resource_arm_virtual_machine_extension.go
@@ -91,7 +91,16 @@ func resourceArmVirtualMachineExtension() *schema.Resource {
 				DiffSuppressFunc: structure.SuppressJsonDiff,
 			},
 			// TODO: Remove location and resource_group_name in 2.0
-			"location":            azure.SchemaLocationDeprecated(),
+			//"location":            azure.SchemaLocationDeprecated(),
+			"location": {
+				Type:             schema.TypeString,
+				ForceNew:         true,
+				Optional:         true,
+				Computed:         true,
+				StateFunc:        azure.NormalizeLocation,
+				DiffSuppressFunc: azure.SuppressLocationDiff,
+				Deprecated:       "location is no longer used",
+			},
 			"resource_group_name": azure.SchemaResourceGroupNameDeprecated(),
 			"tags":                tags.Schema(),
 		},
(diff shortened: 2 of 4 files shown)


azurerm/internal/services/compute/registration.go
```diff
@@ -45,7 +45,7 @@ func (r Registration) SupportedResources() map[string]*schema.Resource {
 		"azurerm_shared_image":                         resourceArmSharedImage(),
 		"azurerm_snapshot":                             resourceArmSnapshot(),
 		"azurerm_virtual_machine_data_disk_attachment": resourceArmVirtualMachineDataDiskAttachment(),
-		"azurerm_virtual_machine_extension":            resourceArmVirtualMachineExtensions(),
+		"azurerm_virtual_machine_extension":            resourceArmVirtualMachineExtension(),
```
katbyte: Nice catch 👍

azurerm/internal/services/compute/resource_arm_virtual_machine_extension.go
```diff
@@ -85,26 +90,67 @@ func resourceArmVirtualMachineExtensions() *schema.Resource {
...
+
+		virtualMachineName = v.Name
+		resourceGroup = v.ResourceGroup
+
+		virtualMachine, err := vmClient.Get(ctx, resourceGroup, virtualMachineName, "")
+		if err != nil {
+			return fmt.Errorf("Error getting Virtual Machine %q (Resource Group %q): %+v", name, resourceGroup, err)
+		}
+
+		location = *virtualMachine.Location
```
katbyte: I think we'll need to nil check this

azurerm/internal/services/compute/tests/resource_arm_virtual_machine_extension_test.go
```diff
@@ -246,6 +274,111 @@ resource "azurerm_virtual_machine" "test" {
...
+
+  tags = {
+    environment = "Production"
+  }
+}
+`, data.RandomInteger, data.Locations.Primary, data.RandomInteger, data.RandomInteger, data.RandomInteger, data.RandomInteger, data.RandomInteger, data.RandomInteger, data.RandomInteger)
+}
+
+// TODO: Remove in 2.0
+func testAccAzureRMVirtualMachineExtension_deprecated(data acceptance.TestData) string {
```
katbyte: THis seems to be testing the new property so shouldn't be removed in 2.0/deprecated? 🤔

azurerm/internal/services/compute/tests/resource_arm_virtual_machine_extension_test.go
```diff
@@ -40,6 +40,34 @@ func TestAccAzureRMVirtualMachineExtension_basic(t *testing.T) {
...
+				),
+			},
+			data.ImportStep("protected_settings"),
+			{
+				Config: testAccAzureRMVirtualMachineExtension_deprecatedUpdate(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVirtualMachineExtensionExists(data.ResourceName),
+					resource.TestMatchResourceAttr(data.ResourceName, "settings", regexp.MustCompile("whoami")),
+				),
+			},
```
katbyte: And another import step here

azurerm/internal/services/compute/resource_arm_virtual_machine_extension.go
```diff
@@ -40,16 +40,21 @@ func resourceArmVirtualMachineExtensions() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
-
```
katbyte: minor, generally we leave spaces between properties now, but not a blocker.

azurerm/internal/services/compute/tests/resource_arm_virtual_machine_extension_test.go
jackofallops: This config uses the deprecated fields (name / location) which won't be available post-2.0, now commented as such.

PR #5457, opened 2020-01-21
Feature: `azurem_application_insights` add support for `retention_in_days`

neil-yechenwei:
Fix the issue of https://github.com/terraform-providers/terraform-provider-azurerm/issues/5441

### DIFF
--- a/azurerm/internal/services/applicationinsights/resource_arm_application_insights.go
+++ b/azurerm/internal/services/applicationinsights/resource_arm_application_insights.go
@@ -64,6 +64,23 @@ func resourceArmApplicationInsights() *schema.Resource {
 				}, true),
 			},
 
+			"retention_in_days": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  90,
+				ValidateFunc: validation.IntInSlice([]int{
+					30,
+					60,
+					90,
+					120,
+					180,
+					270,
+					365,
+					550,
+					730,
+				}),
+			},
+
 			"sampling_percentage": {
 				Type:         schema.TypeFloat,
 				Optional:     true,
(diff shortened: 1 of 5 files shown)


azurerm/internal/services/applicationinsights/resource_arm_application_insights.go
```diff
@@ -64,6 +64,23 @@ func resourceArmApplicationInsights() *schema.Resource {
 				}, true),
 			},
 
+			"retention_in_days": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  90,
```
mbfrahry: Unfortunately, this is a breaking change since the API doesn't provide `retention_in_days` by default. This means that users migrating to this update will see a change like `retention_in_days   = 0 -> 90`.  We can't have this so this default should be removed.

mbfrahry (changes requested):
Hey @neil-yechenwei, there is a breaking change here that I've documented below.

azurerm/internal/services/applicationinsights/resource_arm_application_insights.go
katbyte: If the API returns a default we can probably just use computed instead

website/docs/d/application_insights.html.markdown
```diff
@@ -35,4 +35,5 @@ output "application_insights_instrumentation_key" {
 * `application_type` - The type of the component.
 * `instrumentation_key` - The instrumentation key of the Application Insights component.
 * `location` - The Azure location where the component exists.
+* `retention_in_days` - The retention period in days. Possible values are `30`, `60`, `90`, `120`, `180`, `270`, `365`, `550` or `730`.
```
katbyte: This is a datasource so can we word this as such

katbyte (changes requested):

azurerm/internal/services/applicationinsights/resource_arm_application_insights.go
neil-yechenwei: API returns "nil". So I removed the default value.

website/docs/d/application_insights.html.markdown
neil-yechenwei: fixed

neil-yechenwei:
@mbfrahry  @katbyte , Thanks for your comments. I've updated code.

mbfrahry (approved):
LGTM! And tests are passing. Thanks for taking the time to get write this out @neil-yechenwei

Issue #5321, opened 2020-01-06
terraform plan doesn't normalize IPv6 addresses in azurerm_dns_aaaa_record resources


bbkane:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

```
λ terraform -v
Terraform v0.12.18
+ provider.azurerm v1.39.0
```

### Affected Resource(s)

* `azurerm_dns_aaaa_record`

### Terraform Configuration Files

```hcl
terraform {
  backend "local" {
    path = "terraform.tfstate"
  }
}

provider "azurerm" {
  # whilst the `version` attribute is optional, we recommend pinning to a given version of the Provider
  version = "~> 1.39"
}


resource "azurerm_resource_group" "B12_azure_aaaa_mismatch_repro" {
  name     = "12_azure_aaaa_mismatch_repro"
  location = "West US 2"
}

resource "azurerm_dns_zone" "BbbkaneScom" {
  name                = "bbkane.com"
  resource_group_name = azurerm_resource_group.B12_azure_aaaa_mismatch_repro.name
}


resource "azurerm_dns_aaaa_record" "azure_aaaa_record" {

  name                = "name-ipv6"
  zone_name           = azurerm_dns_zone.BbbkaneScom.name
  resource_group_name = azurerm_resource_group.B12_azure_aaaa_mismatch_repro.name
  ttl                 = 3600
  records             = ["2607:f8b0:4005:0800:0000:0000:0000:1003"]
}
```

### Debug Output

From `terraform plan` after resource creation (`TF_LOG=TRACE`): https://gist.github.com/bbkane/1e427ff5d614cb5b7c7878e4aa51b9b8 

### Panic Output

No panic

### Expected Behavior

With the above code:

```
terraform init  # initialize azure provider
terraform apply   # create resource group, zone, and AAAA record
terraform plan  # shouldn't do anything - all things already created and code not changed
```

### Actual Behavior

```
terraform init  # initialize azure provider
terraform apply   # create resource group, zone, and AAAA record
terraform plan  # wants to "update" already created AAAA record with uncompressed version
[...]

PR #5459, opened 2020-01-21
Bug: `azurerm_dns_aaaa_record` - Normalize IPv6 addresses

phires:
This is a work in progress try to fix issue #5321 .
Sadly somehow the DiffSuppressFunc does not compare the correct set values with each other - and I have no idea why. 
If adding a logline to the func I get something like this:

```
old 2607:f8b0:4005:800::1003, new 
old 2201:1234:1234::1, new 2201:1234:1234::1
old , new 2607:f8b0:4005:0800:0000:0000:0000:1003
```

Anyone any clues?


azurerm/helpers/azure/network.go
```diff
@@ -0,0 +1,17 @@
+package azure
+
+import (
+	"fmt"
+	"net"
+)
+
```
katbyte: I don't think we need this function?

azurerm/internal/services/dns/resource_arm_dns_aaaa_record.go
```diff
@@ -207,6 +213,13 @@ func resourceArmDnsAaaaRecordDelete(d *schema.ResourceData, meta interface{}) er
 	return nil
 }
 
+func ipv6AddressDiffSuppress(_, old, new string, _ *schema.ResourceData) bool {
```
katbyte: could we move this into the suppress package for `suppress.IPv6Compression`

azurerm/internal/services/dns/resource_arm_dns_aaaa_record.go
```diff
@@ -48,9 +49,14 @@ func resourceArmDnsAAAARecord() *schema.Resource {
 			},
 
 			"records": {
-				Type:          schema.TypeSet,
-				Optional:      true,
-				Elem:          &schema.Schema{Type: schema.TypeString},
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
```
katbyte: And we should add a `validate.IPv6Address`

azurerm/internal/services/dns/resource_arm_dns_aaaa_record.go
```diff
@@ -207,6 +213,13 @@ func resourceArmDnsAaaaRecordDelete(d *schema.ResourceData, meta interface{}) er
 	return nil
 }
 
+func ipv6AddressDiffSuppress(_, old, new string, _ *schema.ResourceData) bool {
+	oldIp := net.ParseIP(old)
+	newIp := net.ParseIP(new)
+
+	return oldIp.Equal(newIp)
```
katbyte: This should look more like this:
```
	oldIp := net.ParseIP(old)
	if oldIp == nil {
		return false
	}

	newIp := net.ParseIP(new)
	if newIp == nil {
		return false
	}

	return oldIp.Equal(newIp)
```

azurerm/helpers/azure/network.go
phires: That's correct, it was a leftover from earlier tries. 
Removed.

katbyte:
👋 @phires, Is this reaady to be looked at outside the failing travis test?

WodansSon:
@phires, would it make sense to just extend the `validate.IPv6Address` function to include this compressed IPv6 validation functionality instead of exposing a `suppress` function? At any rate, just a question... thanks for this! 🚀

katbyte:
Pretty sure the validation function will allow the compressed form in, so it won't really help

phires:
> wave @phires, Is this reaady to be looked at outside the failing travis test?

I've removed the unused import (shame on me), would be great to have a look onto this, why the DiffSuppressFunc is not working correctly.

katbyte:
Hey @phires, i believe this is because we are trying to diff suppress on a set and they behave differently then a list. I think what we'll need to do is setup a custom hash function an a state migration to switch to the new hashs.

WodansSon:
@phires, I believe @katbyte is totally correct in her assessment of what is happening here. I was playing around with this too and added some debug spew to your code just to see what the values were when entering and exiting your function and what I saw is consistent with what @katbyte is saying:

*** = Enter function
--- = Exit function
```
*** Str Old: 2607:f8b0:4005:800::1003 :: Str New:  ***
--- Old: 2607:f8b0:4005:800::1003 :: New: <nil> :: Equal: false ---

*** Str Old:  :: Str New: 2607:f8b0:4005:0800:0000:0000:0000:1003 ***
--- Old: <nil> :: New: 2607:f8b0:4005:800::1003 :: Equal: false ---
```

phires:
> @phires, I believe @katbyte is totally correct in her assessment of what is happening here. I was playing around with this too and added some debug spew to your code just to see what the values were when entering and exiting your function and what I saw is consistent with what @katbyte is saying:
> 
> *** = Enter function
> --- = Exit function
> 
> ```
> *** Str Old: 2607:f8b0:4005:800::1003 :: Str New:  ***
> --- Old: 2607:f8b0:4005:800::1003 :: New: <nil> :: Equal: false ---
> 
> *** Str Old:  :: Str New: 2607:f8b0:4005:0800:0000:0000:0000:1003 ***
> --- Old: <nil> :: New: 2607:f8b0:4005:800::1003 :: Equal: false ---
> ```

Yes, see my original PR text, I had added a similar func already ;)

So, a custom hash function it is. Any pointers where to start?

WodansSon:
@phires sure, I played around with this a bit and here is what I did...

In the `./azurerm/helpers/azure` directory I created a file called `ipv6.go`. I then wrote the below code for that file:
```
package azure

import (
	"net"

	"github.com/hashicorp/terraform-plugin-sdk/helper/hashcode"
)

func NormalizeIPv6Address(ipv6 interface{}) string {
	return net.ParseIP(ipv6.(string)).String()
}

func HashIPv6Address(ipv6 interface{}) int {
	return hashcode.String(NormalizeIPv6Address(ipv6))
}
```

Then in the resource I updated the "records" schema `Set` function to be:
```
Set:           azure.HashIPv6Address,
```
I also removed the `DiffSuppressFunc` and removed the `github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/suppress` package import declaration.

Finally, I modified the expand function:
```
Orig code:
ipv6 := v.(string)

New code:
ipv6 := azure.NormalizeIPv6Address(v)
```

and the flatten function:
```
Orig code:
results = append(results, *record.Ipv6Address)

New code:
results = append(results, azure.NormalizeIPv6Address(*record.Ipv6Address))
```

That's it... I think that should work now all we need now is to move the test cases and update them to use the new normalize functions and I think we are gold. 😃

WodansSon:
@phires doing it this way I don't think we will need the state migration either, but play with it and see how it goes.

WodansSon:
👋 @phires just curious, did you want to make the changes to your code or are you ok with me pushing to your branch? Thanks again for the PR!

phires:
Hi, I've implemented the changes and also build some tests for the new funcs.

WodansSon:
image

Issue #5227, opened 2019-12-20
Validation Helper: CIDR network


phires:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

Potentially all that are using the validators

### Expected Behavior

* A CIDR value of 123.456.789.999/10 should be invalid
* A CIDR value of 0.0.0.0 should be invalid, as it is not a valid CIDR notation without the network part

### Actual Behavior

Above examples are valid

### Steps to Reproduce

1. `terraform apply`

### References

* #0000

phires:
There is a validator in go (net.ParseCIDR).
I would refactor these validation methods, although I'm not too sure of the implications of making an IP without network suffix (e.g. 0.0.0.0) failing the check.

katbyte:
@phires, that would be a great addition. You could make it an additional parameter or have multiple such as `validate.CIDR` and `validate.CIDRAllowNoSuffix` or `validate.CIDRorIP`

favoretti:
Since this issue has been reported a long time ago and relates to the version of provider we no longer support - I'm going to close it. Please open a new updated bug report on current versions of terraform and provider if this is still relevant. Thank you.

PR #5461, opened 2020-01-21
Fix for Issue #5227

phires:
This resolves issue  #5227.

CIDR notations are now checked using the respective go function instead of a regex.
This leads to old notations without the network suffix to be invalid and could possible break some deployments. 
Additionally a validation func is added for these invalid notations (CIDRAllowNoSuffix).

Also, as more and more services now allow IPv4 and IPv6 a func has been added to validate both versions (IPAddress)

### DIFF
--- a/azurerm/helpers/validate/network.go
+++ b/azurerm/helpers/validate/network.go
@@ -94,7 +94,17 @@ func validateIPAddress(i interface{}, k string, allowEmpty bool) (warnings []str
 }
 
 // CIDR validations allow IPv4 and IPv6 CIDR notations as defined in RFC 4632 and RFC 4291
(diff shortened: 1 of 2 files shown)


katbyte:
Hey @phires, thank you for doing this.

We recently moved the non azure specific validators into the plugin SDK: https://github.com/hashicorp/terraform-plugin-sdk/pull/296 & https://github.com/hashicorp/terraform-plugin-sdk/pull/301

Is there anything here not already reflected by those PRs? - we have decided to remove the `AndEmpty` ones as we can no combine them with `Any` or `All`

phires:
Moving the validators up to the SDK is a great idea, as all providers participate of those now.
So I guess this PR can be closed, as everything seems to be there.

katbyte:
Closing as requested 👍

PR #5470, opened 2020-01-22
Add data source for azurerm_api_management_api_version_set

neil-yechenwei:
Fix the issue of https://github.com/terraform-providers/terraform-provider-azurerm/issues/5217

### DIFF
--- a/website/docs/d/api_management_api_version_set.html.markdown
+++ b/website/docs/d/api_management_api_version_set.html.markdown
@@ -1,0 +1,47 @@
+---
+subcategory: "API Management"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_api_management_api_version_set"
+description: |-
+  Gets information about an API Version Set within an existing API Management Service.
+---
+
+# Data Source: azurerm_api_management_api_version_set
+
+Uses this data source to access information about an API Version Set within an API Management Service.
+
+## Example Usage
+
+```hcl
+data "azurerm_api_management_api_version_set" "example" {
+  resource_group_name = "acctestRG"
+  api_management_name = "acctestApiManagement"
+  name                = "example-api-version-set"
+}
+
+output "api_management_api_version_set_id" {
+  value = "${data.azurerm_api_management_api_version_set.example.id}"
+}
+```
+
+## Argument Reference
+
+* `name` - (Required) The name of the API Version Set.
+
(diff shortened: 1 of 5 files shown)


website/docs/d/api_management_api_version_set.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+}
+```
+
+## Argument Reference
+
+* `name` - (Required) The name of the API Version Set.
+
+* `resource_group_name` - (Required) The name of the Resource Group in which the parent API Management Service exists.
+
+* `api_management_name` - (Required) The name of the [API Management Service](api_management.html) in which the API Version Set should exist.
```
katbyte:
```suggestion
* `api_management_name` - (Required) The name of the [API Management Service](api_management.html) in which the API Version exists.
```

website/docs/d/api_management_api_version_set.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+## Attributes Reference
+
+* `id` - The ID of the API Version Set.
+
+* `description` - The description of API Version Set.
+
+* `display_name` - The display name of this API Version Set.
+
+* `versioning_schema` - The value that determines where the API Version identifer will be located in a HTTP request. Possible values are `Header`, `Query` and `Segment`.
```
katbyte: We don't need to state the possible values for data sources
```suggestion
* `versioning_schema` - The value that determines where the API Version identifier will be located in a HTTP request.
```

website/docs/d/api_management_api_version_set.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+# Data Source: azurerm_api_management_api_version_set
+
+Uses this data source to access information about an API Version Set within an API Management Service.
+
+## Example Usage
+
+```hcl
+data "azurerm_api_management_api_version_set" "example" {
+  resource_group_name = "acctestRG"
```
katbyte: Could we use example names here
```suggestion
  resource_group_name = "example-resources"
```

website/docs/d/api_management_api_version_set.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+  api_management_name = "acctestApiManagement"
```
katbyte:
```suggestion
  api_management_name = "example-api-management"
```

PR #5471, opened 2020-01-22
azurerm_traffic_manager_profile - add expected_status_code_ranges

katbyte:
also changes some TypeSets -> TypeLists as the API ignores all but the first element

### DIFF
--- a/azurerm/internal/services/trafficmanager/resource_arm_traffic_manager_profile.go
+++ b/azurerm/internal/services/trafficmanager/resource_arm_traffic_manager_profile.go
@@ -1,19 +1,19 @@
 package trafficmanager
 
 import (
-	"bytes"
 	"fmt"
 	"log"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


azurerm/internal/services/trafficmanager/resource_arm_traffic_manager_profile.go
```diff
@@ -194,7 +212,7 @@ func resourceArmTrafficManagerProfileCreateUpdate(d *schema.ResourceData, meta i
 
 	read, err := client.Get(ctx, resGroup, name)
 	if err != nil {
-		return err
+		return fmt.Errorf("Error reading TrafficManager profile %s (resource group %s): %v", name, resGroup, err)
```
jackofallops: Should we update this to `resourceGroup` as elsewhere in the provider?

katbyte: Not a bad idea! done

PR #5485, opened 2020-01-22
azurerm_netapp_volume  - support NFSv4.1, vol sizing issue

paulomarquesc:
Hi Terraform team,

I'm part of Azure NetApp team at Microsoft and I'm providing this PR that adds support for NFSv4.1 since we have a partner that needs this support to work internally and with their own customers.

This is a summary of changes I made:

* Bumped up the ANF's Go SDK version to the latest (**2019-10-01**)
* Implemented an attribute name change for the ExportPolicyRule that got renamed to **Nfsv41** since api version **2019-07-01** and applied the same renaming concept to Go's own object properties to avoid confusion.
* Implemented the property called **protocol_types** as additional volume property since this is the one that controls what protocol the volume will support (it is defined as a list following our API definition but only one is supported) and included validation.
* Implemented some changes on Tests so we can check if a **NFSv4.1** volume gets created, plus testing for **NFSv3** value on basic volume test
* Stepped down the ServiceLevel from **Premium** to **Standard** on Tests so developer incur in less costs while working on ANF resource.
* Fixed a  bug related to volume size that was limiting a volume to 4TB while the correct maximum size is 100TB.

The acceptance tests are working fine as you can see the results below:

image

Please let me know if this PR looks good or if any changes are needed.

Regards

Paulo


azurerm/internal/services/netapp/data_source_netapp_volume.go
```diff
@@ -61,6 +62,18 @@ func dataSourceArmNetAppVolume() *schema.Resource {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
+
+			"protocol_types": {
```
katbyte: if only one is allowed this should be
```suggestion
			"protocol": {
```

I think type isn't really needed either ?

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -89,6 +89,18 @@ func resourceArmNetAppVolume() *schema.Resource {
 				ValidateFunc: azure.ValidateResourceID,
 			},
 
+			"protocol_types": {
```
katbyte: Same as above comment

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -122,7 +134,7 @@ func resourceArmNetAppVolume() *schema.Resource {
 							Type:     schema.TypeBool,
 							Required: true,
 						},
-						"nfsv4_enabled": {
+						"nfsv41_enabled": {
```
katbyte: What is the point of these enabled functions if we have the protocol property ? if that property is only cifs, can nfsv3 still be enabled?

For that matter can you enabled cifs and nfs3 for the same export rule?

katbyte: Also by the nature of changing the name of this property the hashcode will change and a state migration will be required

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -176,6 +189,7 @@ func resourceArmNetAppVolumeCreateUpdate(d *schema.ResourceData, meta interface{
 			CreationToken:  utils.String(volumePath),
 			ServiceLevel:   netapp.ServiceLevel(serviceLevel),
 			SubnetID:       utils.String(subnetId),
+			ProtocolTypes:  utils.ExpandStringSlice(protocolTypes),
```
katbyte: This seems to imply multiple types can be passed in here?

azurerm/internal/services/netapp/data_source_netapp_volume.go
paulomarquesc: This name aligns with all other client tools, e.g. SDKs, ARM Templates, Portal, Powershell and CLI so this name will be very familiar to the users.

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
paulomarquesc: This name aligns with all other client tools, e.g. SDKs, ARM Templates, Portal, Powershell and CLI so this name will be very familiar to the users.

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
paulomarquesc: Yes. the SDK expects an array.

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
paulomarquesc: For the first question, these properties are specific to the NFS export policy and needs to be separately defined as specified by the APIs. 

For the second question, as I commented above, we will have this support as soon as ANF gets support for multiprotocol, and yes, you can define as of now cifs + nfsv3, cifs + nfsv4 but not nfsv3 + nfsv4 on the export policy, the latter will cause a 500 response. But even if the multiprotocol combinations as of today are not returning an error, it doesn't mean it will work as expected since we didn't add support for it yet.

Regarding the last comment of this conversation, not sure if I follow what this comment mean, I very new to Terraform in general and this is my first contribution :-). Can you provide more details on what does that mean?

azurerm/internal/services/netapp/data_source_netapp_volume.go
paulomarquesc: About the documentation, I'll look at it as well.

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
katbyte: So because there is a hashcode created for each element of this list, adding a new property will change the computed hashcode. So going from version 1.41-1.42 with the same config will produce a diff.

We could also simplify this to a `protocols_enabled` property instead so in to future we don't have to worry about adding a new property and doing another state migration?

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -114,22 +129,36 @@ func resourceArmNetAppVolume() *schema.Resource {
...
+
 						"nfsv3_enabled": {
 							Type:     schema.TypeBool,
 							Required: true,
 						},
+
+						"nfsv41_enabled": {
+							Type:          schema.TypeBool,
+							Optional:      true,
+							ConflictsWith: []string{"export_policy_rule.nfsv4_enabled"},
```
katbyte: As this is in a set we won't be able to do a conflicts with

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -316,15 +365,28 @@ func expandArmNetAppVolumeExportPolicyRule(input []interface{}) *netapp.VolumePr
...
+			// var nfsv41Enabled bool
+
+			// _, nfsv4Ok := v["nfsv4_enabled"].(bool)
+
+			// if nfsv4Ok {
+			// 	nfsv41Enabled = v["nfsv4_enabled"].(bool)
+			// } else {
+			// 	// TODO: Keep only this line in 2.0
+			// 	nfsv41Enabled = v["nfsv41_enabled"].(bool)
+			// }
```
katbyte: This may be tricker then i thought due to it being a set, it might not be possible with a bunch of bools as there is no way to determine which one is set or both are set. Going to a list of protocols would work, otherwise might have to make this change in 2.0

paulomarquesc: Yes, this sounds a good idea, we can create the object with all items set as false and whatever protocol is in the list we make the change in the object to true.

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
paulomarquesc: Since we will work on a list of enabled protocols, should we just remove the conflicts with?

paulomarquesc: @katbyte, thinking about the protocols_enabled property, are you thinking about an array or comma separated string? I don't recall seeing any comma separated string.

paulomarquesc:
Hi @katbyte , I completed the changes, please let me know if they're ok now or if I need to change anything else.

azurerm/internal/services/netapp/data_source_netapp_volume.go
```diff
@@ -61,6 +62,18 @@ func dataSourceArmNetAppVolume() *schema.Resource {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
+
+			"protocols": {
+				Type:     schema.TypeList,
+				Computed: true,
+				MaxItems: 2,
```
katbyte: this is a data source so we don't need the max items

azurerm/internal/services/netapp/data_source_netapp_volume.go
```diff
@@ -61,6 +62,18 @@ func dataSourceArmNetAppVolume() *schema.Resource {
...
+				Elem: &schema.Schema{Type: schema.TypeString,
+					ValidateFunc: validation.StringInSlice([]string{
```
katbyte: We don't need validation on a data source

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -89,6 +90,19 @@ func resourceArmNetAppVolume() *schema.Resource {
 				ValidateFunc: azure.ValidateResourceID,
 			},
 
+			"protocols": {
+				Type:     schema.TypeList,
+				ForceNew: true,
+				Optional: true,
+				MaxItems: 2,
```
katbyte: You can only select two protocols? or do two conflict?

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -89,6 +90,19 @@ func resourceArmNetAppVolume() *schema.Resource {
 				ValidateFunc: azure.ValidateResourceID,
 			},
 
+			"protocols": {
+				Type:     schema.TypeList,
```
katbyte: This should be a type set

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -114,22 +129,25 @@ func resourceArmNetAppVolume() *schema.Resource {
...
-							Required: true,
-						},
-						"nfsv4_enabled": {
-							Type:     schema.TypeBool,
+
+						"protocols_enabled": {
+							Type:     schema.TypeList,
 							Required: true,
+							MaxItems: 1,
+							MinItems: 1,
```
katbyte: You cannot set more then 1 per export?

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -114,22 +129,25 @@ func resourceArmNetAppVolume() *schema.Resource {
 								ValidateFunc: validate.CIDR,
 							},
 						},
-						"cifs_enabled": {
```
katbyte: We will need to keep these around as deprecated. Switch them to optional +computed and then always set them

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -114,22 +129,25 @@ func resourceArmNetAppVolume() *schema.Resource {
...
-						"nfsv3_enabled": {
-							Type:     schema.TypeBool,
-							Required: true,
-						},
-						"nfsv4_enabled": {
-							Type:     schema.TypeBool,
+
+						"protocols_enabled": {
+							Type:     schema.TypeList,
 							Required: true,
```
katbyte: And then this can become optional, and we use the _enabled functions, unless this is set (we can fix this all up in 2.0)

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
paulomarquesc: Only one combination of two values will work, cifs+nfsv3 or cifs+nfsv4.1, three items implies that the user is adding nfsv3 or nfvs4.1 together and this is a combination that fails. I want to write a validator function to check if nfsv3 and nfsv4 are at the same time defined in the list but it looks like the validation is done on item by item, if you have an example where we can validade if two items are in the list that would be great, otherwise we can skip to a next update since this may take more time to be done.

azurerm/internal/services/netapp/data_source_netapp_volume.go
paulomarquesc: Fixed.

azurerm/internal/services/netapp/data_source_netapp_volume.go
paulomarquesc: Fixed,

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
paulomarquesc: Fixed.

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
paulomarquesc: Good catch, yes we can do 2 but same validations applies as I mentioned before.

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
katbyte: You can add a validation function to the list block itself:
```
func ListOfUniqueStrings(i interface{}, k string) (warnings []string, errors []error) {
	v, ok := i.([]interface{})
	if !ok {
		errors = append(errors, fmt.Errorf("expected type of %q to be List", k))
		return warnings, errors
	}

	for _, e := range v {
		if _, eok := e.(string); !eok {
			errors = append(errors, fmt.Errorf("expected %q to only contain string elements, found :%v", k, e))
			return warnings, errors
		}
	}

	for n1, i1 := range v {
		for n2, i2 := range v {
			if i1.(string) == i2.(string) && n1 != n2 {
				errors = append(errors, fmt.Errorf("expected %q to not have duplicates: found 2 or more of %v", k, i1))
				return warnings, errors
			}
		}
	}

	return warnings, errors
}
```

katbyte: i'd also document the restriction

paulomarquesc: I'm specifying single value for now on doc support for multi-protocol is on the road map but not public ETA. With this in mind we can define maxitem=1 on both for now, what do you think?

katbyte: 👍 sounds good

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
katbyte: 👍

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
paulomarquesc: Hi @katbyte , any idea why when I changed this to schema.TypeSet I started getting validation error on my test?

--- FAIL: TestAccAzureRMNetAppVolume_nfsv41 (425.90s)
    testing.go:569: Step 0 error: Check failed: Check 2/2 error: azurerm_netapp_volume.test: Attribute 'protocols.0' not found
FAIL

Here is snippet of the test:

```go
func TestAccAzureRMNetAppVolume_nfsv41(t *testing.T) {
	data := acceptance.BuildTestData(t, "azurerm_netapp_volume", "test")

		PreCheck:     func() { acceptance.PreCheck(t) },
		Providers:    acceptance.SupportedProviders,
		CheckDestroy: testCheckAzureRMNetAppVolumeDestroy,
		Steps: []resource.TestStep{
			{
				Config: testAccAzureRMNetAppVolume_nfsv41(data),
					testCheckAzureRMNetAppVolumeExists(data.ResourceName),
					resource.TestCheckResourceAttr(data.ResourceName, "protocols.0", "NFSv4.1"),
				),
			},
			data.ImportStep(),
		},
	})
}
```

paulomarquesc: I think I solved it, I used the hash that the Set generates and I don't get the error as follows:

```go
resource.TestCheckResourceAttr(data.ResourceName, "protocols.3098200649", "NFSv4.1"),
```

paulomarquesc:
Hi @katbyte , just pushed all fixes. Is there a way to prevent the "destroy" action programmatically for Azure NetApp Files resources? I sent an email thread raising this concern so here is just a summary. A simple change in the TF file on `protocols` attribute from NFSv3 to NFSv4.1 will cause the volume to be destroyed then recreated. ANF does not support soft deletes so once the volume is delete it is gone for good so I want to check with you if there is a way that we can prevent this from happening. I understand that the `terraform plan` will show the user that something will be destroyed but in these cases where the damage can be permanent we should better prevent it.

katbyte:
@paulomarquesc, if you remove the `ForceNew` attribute it won't destroy it. can you update that property without re-creating the resource?

azurerm/internal/services/netapp/data_source_netapp_volume.go
```diff
@@ -61,6 +61,12 @@ func dataSourceArmNetAppVolume() *schema.Resource {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
+
+			"protocols": {
+				Type:     schema.TypeSet,
```
katbyte: Data sources should be `TypeList`

website/docs/r/netapp_volume.html.markdown
```diff
@@ -87,6 +88,8 @@ The following arguments are supported:
 
 * `service_level` - (Required) The target performance of the file system. Valid values include `Premium`, `Standard`, or `Ultra`.
 
+* `protocols` - (Optional) The target volume protocol expressed as a list. Supported single value include `CIFS`, `NFSv3`, or `NFSv4.1`. If argument is not defined it will default to `NFSv3`. **VERY IMPORTANT**, please do not use this attribute to change the protocol used to create the volume, e.g. from NFSv3 to NFSv4.1 this will cause your current NFSv3 volume to destroyed then redeployed as NFSv4.1. Azure NetApp Files does not support soft deletes, once the volume is deleted there is no way to reverse it so if the plan shows that a volume will be destroyed only apply it if you're sure you can get it deleted.
```
katbyte: We usually just put 
```suggestion
* `protocols` - (Optional) The target volume protocol expressed as a list. Supported single value include `CIFS`, `NFSv3`, or `NFSv4.1`. If argument is not defined it will default to `NFSv3`. Changing this forces a new resource to be created and data will be lost.
```

paulomarquesc:
> @paulomarquesc, if you remove the `ForceNew` attribute it won't destroy it. can you update that property without re-creating the resource?

No, as of today, once the volume is created as NFSv3 or NFSv4.1 we can't change it without recreating the volume. Thanks, I'll leave it without the ForceNew and let the platform bubble up an error to be safe.

azurerm/internal/services/netapp/data_source_netapp_volume.go
paulomarquesc: Fixed.

paulomarquesc:
@katbyte, one last item is related to the `export_policy_rule`, I tried some combinations like `Optional: true`, `Computed: true`  this is related to the deprecation of `cifs_enabled`, `nfsv3_enabled` and `nfsv4_enabled` attributes. If the user provides only the _enabled attributes and not `protocols_enabled` it will cause an update, in the other hand if the user provides only `protocols_enabled` and not the other ones it will also cause an update. The only time it does not cause the update is if they provide all of those attributes as follows:

```
  export_policy_rule  {
   rule_index        = 1
   allowed_clients   = ["0.0.0.0/0"]
   cifs_enabled      = false
   nfsv3_enabled     = true
   nfsv4_enabled     = false
   protocols_enabled = ["NFSv3"]
   unix_read_only    = false
   unix_read_write   = true
  }
```

Is there another way to prevent this or should we be ok for the timeframe where we remove the deprecated ones on v2?

paulomarquesc:
Hi @katbyte , any updates in this?

paulomarquesc:
Hi @tombuildsstuff and @katbyte , since this got moved from milestone v1.44.0 to milestone v2.0.0, can I just remove all related code that makes it non-breaking?

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -114,22 +129,47 @@ func resourceArmNetAppVolume() *schema.Resource {
 								ValidateFunc: validate.CIDR,
 							},
 						},
+
+						"protocols_enabled": {
+							Type:     schema.TypeSet,
+							Optional: true,
+							Computed: true,
+							MaxItems: 2,
```
mbfrahry: The docs say only one value is supported at this time. Is that right? If so, we should match that here? From local testing, it seemed like the API accepted all three values being passed in at the same time.

paulomarquesc: Hi @mbfrahry, to your first question, this will be supported in the future and @katbyte asked me to use TypeSet and I tried to reduce the items to 1 but TypeSet does not supports it, only TypeList.
Regarding the second part, as of today it is not supported to have more than one at this point, we have on the road map to support two, CIFS+NFS3 and CIFS+NFS4 but never the three of them or NFS3+NFS4 combination. Have you checked if the export policy got created? It will not throw an error, it just don't get the export policy created.

paulomarquesc:
> Hey @paulomarquesc, these changes are looking off to a good start but I've done some local testing and some of the constraints aren't quite right. I think we can squeeze this into 2.0 so we should be able to remove all the deprecated fields

Awesome, I'll work on removing all the things that makes it non-breaking changes.

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
mbfrahry: Ah, that makes sense. Thanks for clarifying. I just saw the values were put into the template but I didn't deploy the template.

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
paulomarquesc: > I think we'll want to add ForceNew to this as well. As `ProtocolTypesCanNotBeChanged`

@mbfrahry, not sure I understood what is needed here, do you have more details on the "As `ProtocolTypesCanNotBeChanged`" sentence?

paulomarquesc:
Hi @mbfrahry, @katbyte, just saw that v2 was released, can I still remove all extra code for non-breaking changes? And I also still have a last question that needs clarification regarding adding back the ForceNew attribute above, can you please take a look and clarify?

mbfrahry:
Hey @paulomarquesc. Unfortunately, we still have to keep the breaking changes in until the next major release. For your ForceNew question, I saw we just have ForceNew there. Are there any options that you can swap it to other than NFSv3 or NFSv4.1? If so, we could write a `CustomizeDiff` that will ForceNew when those values are being changed from but I don't think it's necessary

paulomarquesc:
> Hey @paulomarquesc. Unfortunately, we still have to keep the breaking changes in until the next major release. For your ForceNew question, I saw we just have ForceNew there. Are there any options that you can swap it to other than NFSv3 or NFSv4.1? If so, we could write a `CustomizeDiff` that will ForceNew when those values are being changed from but I don't think it's necessary

Hi @mbfrahry, there is no migration path from one protocol to another, the volume needs to be recreated but we need to block the customer to inadvertently make this change and end up by deleting production volumes, the only way I see is to remove `ForceNew` or implement an error if this change is coming, I looked at other parts of the AzureRm code and it looks like there is a way to raise the error.

About the breaking changes, ok, i'll keep those,

Another thing, I just completed resolving all the conflicts that came up since master got all the v2.0 code merged in and now I cannot compile my branch anymore, I get this error:

```
# github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate
..\..\helpers\validate\strings.go:34:9: undefined: validation.StringIsNotEmpty
azurerm\internal\provider\services.go:66: running "go": exit status 2
make: *** [generate] Error 1
```
Any ideas how to solve this?

paulomarquesc:
@mbfrahry, I need you guys help here, I tried using the following code change to do a pre-validation but it looks like it does not work the way I was envisioned:

```go
	if d.HasChange("protocols") {
		return fmt.Errorf("`protocols` cannot be changed to a different protocol, a new volume needs to be created. `ProtocolTypesCanNotBeChanged`")
	}
```

When I added back the ForceNew = true, I ended up with my volume deleted and the error message came after it, so no use for this option.

As I mentioned before to @katbyte, unless there is a clear solution you can help me with, that the customer volume will not be deleted before any sort of validation, we will need to leave the ForceNew option out and let Azure bubble up the error.

This review process is already taking a long time and this is impacting one of the largest Azure customer that requires this change. I'm open for a joint session or any other solution that can speed up this PR.

Please let me know how we can make this happen.

katbyte:
@paulomarquesc,

`ForceNew` is already on the name and many other resources where accidental deletion would be less than ideal. I don't see any reason to not mark this attribute as force new as the plan will show it to the user and it is inline with all other terraform resources. For users who are worried about accidental deletion.

```
resource "resource" "name" {
  lifecycle {
    prevent_destroy = true
  }
}
```

for further information see https://www.terraform.io/docs/configuration/resources.html#prevent_destroy

paulomarquesc:
@katbyte, thanks, that is really helpful. I'll even add to the example and make a note in the documentation.

I still have another question pending here: https://github.com/terraform-providers/terraform-provider-azurerm/pull/5485#issuecomment-582918925

Can you please check? No matter what I do, it always removes and places the export policy back if using TypeSet instead of TypeList.

katbyte:
@paulomarquesc, that looks like it may have to do with the export policies being a type set.. it might go away if you use a type list? might not be possibleto do the deprecations with typeset =/

paulomarquesc:
@katbyte, yes, I didn't have this issue when originally I made this as TypeList.

katbyte:
@paulomarquesc, the protocols or the export policy?

katbyte:
Also if you hope onto our community slack we can have a real time discussion there

paulomarquesc:
@katbyte , what is the community name? I couldn't find you guys there.

paulomarquesc:
> @paulomarquesc, the protocols or the export policy?

The issue is with Export policy. Protocols is working fine.

PR #5494, opened 2020-01-23
azurerm_api_management: Deploy APIM in a vnet

pacon-vib:
Long time user, first time contributor. This PR exposes the existing functionality in the Azure SDK to tell an API Management (APIM) service that it needs to live in a vnet, in either internal or external mode. In a vnet in internal mode, we need to export the service's private IP address so that we can create DNS records pointing to it, so this PR does that too.

I'm posting this just to get some early 👀 on it to check I'm on the right track, but I believe I have some more things to do before this is ready to merge. Here's my TODO list, please let me know if there's anything I've missed:
- [ ] Add documentation for new properties
- [ ] Modify test cases to set and read new properties

I'm pretty sure `virtualNetworkTypeStringToType()` is less elegant than it could be, and may not accord with naming conventions, so I'd be grateful for your suggestions. I'm still finding my way around all the helper functions (and around Golang itself), so any pointers would be helpful.

### DIFF
--- a/azurerm/internal/services/apimanagement/resource_arm_api_management.go
+++ b/azurerm/internal/services/apimanagement/resource_arm_api_management.go
@@ -62,6 +62,14 @@ func resourceArmApiManagementService() *schema.Resource {
 				},
 			},
 
+			"private_ip_addresses": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
 			"publisher_name": {
 				Type:         schema.TypeString,
 				Required:     true,
@@ -460,6 +468,21 @@ func resourceArmApiManagementService() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"vnet_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"External",
+					"Internal",
+					"None",
+				}, false),
+			},
+
+			"subnet_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


azurerm/internal/services/apimanagement/resource_arm_api_management.go
```diff
@@ -514,6 +537,12 @@ func resourceArmApiManagementServiceCreateUpdate(d *schema.ResourceData, meta in
 			CustomProperties:       customProperties,
 			Certificates:           certificates,
 			HostnameConfigurations: hostnameConfigurations,
+			VirtualNetworkType:     virtualNetworkTypeStringToType(utils.String(d.Get("vnet_type").(string))),
```
katbyte: We should be able to go
```suggestion
			VirtualNetworkType:     apimanagement.VirtualNetworkType(d.Get("vnet_type").(string))),
```

azurerm/internal/services/apimanagement/resource_arm_api_management.go
```diff
@@ -62,6 +62,14 @@ func resourceArmApiManagementService() *schema.Resource {
 				},
 			},
 
+			"private_ip_addresses": {
```
katbyte: Could we put computed values at the end of the block?

azurerm/internal/services/apimanagement/resource_arm_api_management.go
```diff
@@ -514,6 +537,12 @@ func resourceArmApiManagementServiceCreateUpdate(d *schema.ResourceData, meta in
 			CustomProperties:       customProperties,
 			Certificates:           certificates,
 			HostnameConfigurations: hostnameConfigurations,
+			VirtualNetworkType:     virtualNetworkTypeStringToType(utils.String(d.Get("vnet_type").(string))),
+			VirtualNetworkConfiguration: &apimanagement.VirtualNetworkConfiguration{
+				Vnetid:           utils.String(""),
+				Subnetname:       utils.String(""),
```
katbyte: COuld we also support these two values?

azurerm/internal/services/apimanagement/resource_arm_api_management.go
```diff
@@ -657,6 +686,7 @@ func resourceArmApiManagementServiceRead(d *schema.ResourceData, meta interface{
 		d.Set("management_api_url", props.ManagementAPIURL)
 		d.Set("scm_url", props.ScmURL)
 		d.Set("public_ip_addresses", props.PublicIPAddresses)
+		d.Set("private_ip_addresses", props.PrivateIPAddresses)
```
katbyte: We should we reading in the other new values here too

azurerm/internal/services/apimanagement/resource_arm_api_management.go
```diff
@@ -62,6 +62,14 @@ func resourceArmApiManagementService() *schema.Resource {
 				},
 			},
```
katbyte: Could we put these new values into a `virtual_network` block?

tombuildsstuff:
hey @pacon-vib 

Thanks for this PR :)

Since we've not heard back from you here and another PR's been opened which also adds support for Virtual Networks to API Management (but which contains some of the changes mentioned above) - I hope you don't mind but I'm going to close this in favour of #5769.

Thanks!

PR #5522, opened 2020-01-25
Fix import of stream analytics jobs that have been created via Azure Portal

r0bnet:
Hey,

i found out that importing an existing Stream Analytics job that was created via the Portal didn't work. The reason was that a `transformation` object has a name property. It's used to read & update the transformation. Unfortunately it was called `Transformation` in this provider but Azure calls it `main` when you create the job in the portal.

In this PR i changed the name to be also `main`. In order to keep backward compatibility i'm reading the name of the transformation before updating it.

I tested the following:
- Create a Job via Portal
  - Import it via TF (with PR changes)
  - Update it via TF (with PR changes)
- Create a Job via the current 1.41.0 version of the provider
  - Import it via TF (with PR changes)
  - Update it via TF (with PR changes)

All tests completed without any error.

I also ran all acc tests for this resource successfully.

If you want me to add any further tests pls let me know.

### DIFF
--- a/azurerm/internal/services/streamanalytics/resource_arm_stream_analytics_job.go
+++ b/azurerm/internal/services/streamanalytics/resource_arm_stream_analytics_job.go
@@ -162,7 +162,7 @@ func resourceArmStreamAnalyticsJobCreateUpdate(d *schema.ResourceData, meta inte
 
 	// needs to be defined inline for a Create but via a separate API for Update
 	transformation := streamanalytics.Transformation{
-		Name: utils.String("Transformation"),
+		Name: utils.String("main"),
 		TransformationProperties: &streamanalytics.TransformationProperties{
 			StreamingUnits: utils.Int32(int32(streamingUnits)),
 			Query:          utils.String(transformationQuery),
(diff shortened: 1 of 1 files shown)


azurerm/internal/services/streamanalytics/resource_arm_stream_analytics_job.go
```diff
@@ -162,7 +162,7 @@ func resourceArmStreamAnalyticsJobCreateUpdate(d *schema.ResourceData, meta inte
 
 	// needs to be defined inline for a Create but via a separate API for Update
 	transformation := streamanalytics.Transformation{
-		Name: utils.String("Transformation"),
+		Name: utils.String("main"),
```
katbyte: Would it make sense to instead expose this property with a computed default of "main"? 🤔 that way if the portal changes at some point or the cli is differet code updates won't be required

r0bnet: We could do that but i changed the implementation so that when the resource is read the transformation is read also (inline) so that we always have the correct name.

tombuildsstuff: this is an implementation detail more than anything, so I don't think it's worth exposing this here?

tombuildsstuff:
Tests pass:

Screenshot 2020-02-11 at 10 05 27

PR #5526, opened 2020-01-26
azurerm_netapp_volume - support mount_ip_addresses

### LINKED ISSUE #5416: Support for [Azure NetApp Files] - Create volume no IP address in the attributes
### Community Note

### Description

When an Azure NetApp files volume is created, it does not output the full mount path. Currently, only the volume path (name) is provided in the output. Can we please add the IP address for the volume as well to the attributes section for the data source? 

azurerm_netapp_volume

```hcl
```

### References
https://www.terraform.io/docs/providers/azurerm/d/netapp_volume.html

* #0000

neil-yechenwei:
Fixes #5416


azurerm/internal/services/netapp/resource_arm_netapp_volume.go
```diff
@@ -388,3 +388,23 @@ func flattenArmNetAppVolumeExportPolicyRule(input *netapp.VolumePropertiesExport
...
+func flattenArmNetAppVolumeMountTargets(input interface{}) []interface{} {
+	results := make([]interface{}, 0)
+	if input == nil {
+		return results
+	}
+
+	for _, item := range input.([]interface{}) {
+		if item != nil {
+			v := item.(map[string]interface{})
+			ipAddress := v["ipAddress"]
```
tombuildsstuff: this needs to be:

```suggestion
			ipAddress := v["ip_address"].(string)
```

azurerm/internal/services/netapp/data_source_netapp_volume.go
```diff
@@ -42,6 +42,19 @@ func dataSourceArmNetAppVolume() *schema.Resource {
 				ValidateFunc: ValidateNetAppPoolName,
 			},
 
+			"mount_targets": {
+				Type:     schema.TypeSet,
```
tombuildsstuff: this needs to be a TypeList if it's a Computed field else it's not easily indexable - however - what are the mount targets here, a list of IP Addresses which should be used to mount the volume? If so it's likely this'd make more sense as a list of strings being `mount_ip_addresses`?

can we also update the test to confirm this field is correctly set

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
neil-yechenwei: It cannot be updated to "ip_address" since api returns ""mountTargets":[{"provisioningState":"Succeeded","ipAddress":"10.0.2.4"}]" for property mountTargets.

azurerm/internal/services/netapp/data_source_netapp_volume.go
neil-yechenwei: Fixed

neil-yechenwei:
Hi @tombuildsstuff , Thanks for your comments. I've updated code.

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
tombuildsstuff: ah so this nested object isn't modelled as an object in the Swagger? in which case can we send a PR to the Swagger repo to fix that

website/docs/d/netapp_volume.html.markdown
```diff
@@ -43,6 +43,8 @@ The following attributes are exported:
 
 * `location` - The Azure Region where the NetApp Volume exists.
 
+* `mount_ip_addresses` - A list of IPv4 Addresses which should be used to mount the volume.
```
tombuildsstuff: this has also been added to the Resource - so can we document this field there too?

neil-yechenwei: I didn't add this field "mount_ip_addresses" to the resource file. I just added this to data source file. So I think we don't need to document this field to resource md file.

neil-yechenwei:
> hey @neil-yechenwei
> 
> Thanks for making those changes - taking a look through whilst this looks good it appears we also need to send a PR to fix the Swagger prior to merging this, else breaking changes to the API Response won't be caught via the ARM "breaking changes" process - would you be able to send a PR to fix that in the Swagger and link that in the code here so that we can switch it out once fixed?
> 
> Thanks!

@tombuildsstuff , I am not sure what you mean. Do you mean I need to fix MountTargets type as exact object not interface{} like ExportPolicy in the Swagger, right? If not, may I know what & why I need to fix in the swagger?

neil-yechenwei:
@katbyte , If you point to this issue, I've confirmed with service owner. They will fix this issue in the future but no ETA. So I've added a comment in the code and filed an issue (https://github.com/Azure/azure-rest-api-specs/issues/8604) on rest api spec. Once they fix this issue, I will make an another PR to update code. So could you help to merge this PR first?

website/docs/d/netapp_volume.html.markdown
katbyte: Is there a reason we are only adding it to the datasource and not the resource too? @neil-yechenwei

neil-yechenwei: You can find user's usage preference from (https://github.com/terraform-providers/terraform-provider-azurerm/issues/5416), seems they want use data source to retrieve this property value.

jtracey93:
Hi all, been watching this PR closely. Any ideas when it may go live so it’s available for use? Would really like to use TF to deploy some SMB ANF volumes. 
Thanks

katbyte:
@jtracey93, this is currently blocked on an update to the rest api swagger, once that has been made and an updated SDK released we can get this in.

azurerm/internal/services/netapp/resource_arm_netapp_volume.go
ArcturusZhang: When it comes to `mountTargets`, it reminds me that I changed that some time ago... and it may be a mistake. I will take a look at this and see how to change it correctly (and when changing the swagger we should let the service team know and approve, which may take a long time).

katbyte:
@neil-yechenwei, now that #6419 is meged i think this has ben unblocked!

neil-yechenwei:
@katbyte & @tombuildsstuff , I've applied the fix for property mountTargets after new go sdk is upgraded. Thanks.

katbyte:
@neil-yechenwei - could we add the property to the resource to? thanks!

neil-yechenwei:
@katbyte , I also added it to resource. Thanks.

PR #5531, opened 2020-01-27
`azurerm_kubernetes_cluster` & `azurerm_kubernetes_cluster_node_pool`:  Support for `node_labels`

### LINKED ISSUE #4968: azurerm_kubernetes_cluster does not support labels in an agent_pool_profile
### Community Note

### Description

It is possible to apply k8s labels to a node pool with the Azure CLI, but not with Terraform.  `node_taints` are available, which is great, but we use a combination of taints and label matching to place pods. It would be great to add `node_labels` in the same style as `node_taints`.

dintel:
Add support for `node_labels` in `azurerm_kubernetes_cluster` and `azurerm_kubernetes_cluster_node_pool` resources.
Fixes #4968


azurerm/internal/services/containers/data_source_kubernetes_cluster.go
```diff
@@ -176,6 +176,14 @@ func dataSourceArmKubernetesCluster() *schema.Resource {
 							Computed: true,
 						},
 
+						"node_labels": {
+							Type:     schema.TypeMap,
+							Optional: true,
```
mbfrahry: It doesn't look like this is used to get the resource and as such should be swapped to 
```suggestion
							Computed: true,
```

azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
```diff
@@ -1230,6 +1238,10 @@ func expandKubernetesClusterAgentPoolProfiles(input []interface{}, isNewResource
 			return nil, fmt.Errorf("Can't create an AKS cluster with autoscaling enabled but not setting min_count or max_count")
 		}
 
+		if nodeLabels := utils.ExpandMapStringPtrString(config["node_labels"].(map[string]interface{})); len(nodeLabels) > 0 {
```
mbfrahry: Do we want to empty node label struct to the api so people can remove the labels?

azurerm/internal/services/containers/kubernetes_nodepool.go
```diff
@@ -286,6 +298,14 @@ func FlattenDefaultNodePool(input *[]containerservice.ManagedClusterAgentPoolPro
 		name = *agentPool.Name
 	}
 
+	var nodeLabels map[string]string
+	if agentPool.NodeLabels != nil {
+		nodeLabels = make(map[string]string)
```
mbfrahry: Might be misreading here but can't we use your nifty flatten function here?

azurerm/internal/services/containers/kubernetes_nodepool.go
```diff
@@ -90,6 +90,14 @@ func SchemaDefaultNodePool() *schema.Schema {
 					ValidateFunc: validation.IntBetween(1, 100),
 				},
 
+				"node_labels": {
+					Type:     schema.TypeMap,
+					Optional: true,
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
```
mbfrahry: I think we'll want to add validation here that prevents people from adding empty labels unless an empty label is valid?

azurerm/utils/common_marshal.go
```diff
@@ -12,6 +12,15 @@ func ExpandStringSlice(input []interface{}) *[]string {
 	return &result
 }
 
+func ExpandMapStringPtrString(input map[string]interface{}) map[string]*string {
+	result := make(map[string]*string)
+	for k, v := range input {
+		s := v.(string)
+		result[k] = &s
```
mbfrahry: I think we can consolidate this just a bit
```suggestion
		result[k] = utils.String(v.(string))
```

azurerm/utils/common_marshal.go
```diff
@@ -21,3 +30,11 @@ func FlattenStringSlice(input *[]string) []interface{} {
 	}
 	return result
 }
+
+func FlattenMapStringPtrString(input map[string]*string) map[string]interface{} {
+	result := make(map[string]interface{})
+	for k, v := range input {
+		result[k] = *v
```
mbfrahry: We'll want to confirm that `v` isn't nil before pointing to it

mbfrahry (changes requested):
Hey @dintel, this looks good so far but I've made a couple minor suggestions. It'd also be good to add some tests to confirm that we can create, update, and remove them

azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
dintel: From my tests of Azure API node labels can not be changed.
Whenever you try to alter node labels API accepts the update, but labels are never changed.

azurerm/internal/services/containers/kubernetes_nodepool.go
dintel: `ExpandMapStringPtrString` is almost the same, but it accepts `map[string]interface{}`, so it can not be used here.

azurerm/internal/services/containers/kubernetes_nodepool.go
dintel: Empty label is valid.

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -239,6 +241,8 @@ A `default_node_pool` block supports the following:
 
 * `max_pods` - (Optional) The maximum number of pods that can run on each agent. Changing this forces a new resource to be created.
 
+* `node_labels` - (Optional) A map of Kubernetes labels which should be applied to nodes in the agent pool (e.g `{"name": "value"}`).
```
tombuildsstuff: could we make this:

```suggestion
* `node_labels` - (Optional) A map of Kubernetes labels which should be applied to nodes in the Default Node Pool.
```

(since this is a Map users can imply it's:

```
foo = {
  Hello = "World"
}
```

as such we can remove this example)

tombuildsstuff (changes requested):
hey @dintel 

Thanks for this PR / pushing those changes - I've taken a look through and left some other minor comments inline but this is otherwise looking good 👍 

Thanks!

azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
```diff
@@ -223,6 +223,14 @@ func resourceArmKubernetesCluster() *schema.Resource {
 							ForceNew: true,
 						},
 
+						"node_labels": {
+							Type:     schema.TypeMap,
+							Optional: true,
```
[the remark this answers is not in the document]
dintel: @tombuildsstuff since agent the `agent_pool_profile` block is deprecated, do not you want to drop it here too?

dintel:
@tombuildsstuff @mbfrahry I made changes as you guys requested. Could you please review again?

dintel:
@tombuildsstuff @mbfrahry could you guys review this PR? I made some changes to reflect new version changes.

mbfrahry (changes requested):
Hey @dintel, I'm struggling to get a couple of the tests working.
```
--- FAIL: TestAccAzureRMKubernetesCluster_nodeLabels (80.00s)
    testing.go:640: Step 0 error: errors during apply:
        
        Error: Error creating Managed Kubernetes Cluster "acctestaks200304233511150885" (Resource Group "acctestRG-200304233511150885"): containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="CanNotTaintAllAgentPools" Message="Operation failed due to insufficient nodes for system pod scheduling. Placing custom taints on all node pools is not supported. Retry the operation and leave at least one node pool untainted."
```
Looking at what we are sending to the  api, it seems like the label is being appended to the taint attribute.
```
"nodeLabels": {},
"nodeTaints": ["key=value:PreferNoSchedule"],
```

And then the datasource config seems off 
```
--- FAIL: TestAccDataSourceAzureRMKubernetesCluster_nodeLabels (0.02s)
    testing.go:640: Step 0 error: Missing item separator: On /opt/teamcity-agent/temp/buildTmp/tf-test852087197/main.tf line 19: Expected a comma to mark the beginning of the next item.
```

dintel:
@mbfrahry Please check now.

mbfrahry (approved):
LGTM! Thanks for your taking the time to get this in

PR #5533, opened 2020-01-27
Add Network Rule Set to Servicebus Namespace

guillaumestevens:
Add a network rule set block to define network restrictions (ip & vnet). Based on Network rule set already present in eventhub_namespace resource.
tests are updated, however, when i'm running them on my laptop, he crash (CPU overload).


guillaumestevens:
hi @jackofallops 

ok i'll do the split. 

Guillaume

guillaumestevens:
Hi @jackofallops

i've just pushed the change. 

Guillaume

guillaumestevens:
Hello, 

i've a little problem. When Azure create a servicebus namespace, the network rule sets is created (with the name default, can't be change) with just the properties "default_action = allow". 

I've tried to uptade to the provider new version 2.X but i've a problem with the feature 'ShouldResourcesBeImported'.
I don't know how to deal with that. Users should'nt have to import this new resource at the servicebus creation. How to do that automatically ? 

Thanks for your help

katbyte:
Hey @guillaumestevens,

Sorry we have taken so long to get back to this. It seems there is a more complete & up to date pr at #6379 without conflict. As such i'm going to close this and i thank you again for opening the PR!

PR #5539, opened 2020-01-27
azurerm_front_door - Add minimum_tls_version property

### LINKED ISSUE #5538: azurerm_frontdoor HTTPS provisioning request fails
### Community Note

```
Terraform v0.11.14
+ provider.azuread v0.7.0
+ provider.azurerm v1.42.0
+ provider.null v2.1.2
+ provider.random v2.2.1
+ provider.template v2.1.2
```

### Affected Resource(s)

* `azurerm_frontdoor`

```hcl
resource "azurerm_frontdoor" "frontdoor" {
  name                = "${var.name}"
  friendly_name       = "${var.name}"

  enforce_backend_pools_certificate_name_check = true

  backend_pool {
    name                = "${local.backend_name}"
    health_probe_name   = "web-health-probe"
    load_balancing_name = "web-lb"
    backend {
      address     = "${var.backend_web_fqdn}"
      host_header = "${var.backend_web_fqdn}"
      http_port   = 80
      https_port  = 443
    }
  }

  backend_pool_load_balancing {
    name = "web-lb"
  }

  backend_pool_health_probe {
    name                = "web-health-probe"
    path                = "/_health_check"
    protocol            = "Https"
    interval_in_seconds = "30"
  }

  frontend_endpoint {
    name                                    = "${local.frontend_name}"
    host_name                               = "${var.frontend_web_fqdn}"
    web_application_firewall_policy_link_id = "${azurerm_frontdoor_firewall_policy.waf_policy.id}"
    custom_https_provisioning_enabled       = true
    custom_https_configuration {
      certificate_source = "FrontDoor"
    }
  }

  frontend_endpoint {
    custom_https_provisioning_enabled = false
    host_name                         = "${var.name}.azurefd.net"
    name                              = "defaultHostName"
    session_affinity_enabled          = true
    session_affinity_ttl_seconds      = 0
  }

  routing_rule {
    name               = "web-primary-rule"
    accepted_protocols = ["Https"]
    frontend_endpoints = ["${local.frontend_name}"]
    patterns_to_match  = ["/*"]
    enabled            = true

    forwarding_configuration {
      backend_pool_name   = "${local.backend_name}"
      forwarding_protocol = "HttpsOnly"
      cache_enabled       = false
    }
  }
}
```

### Expected Behavior

Custom Domain HTTPS is enabled successfully.

### Actual Behavior

Error:

```
azurerm_frontdoor.frontdoor: Unable enable Custom Domain HTTPS for Frontend Endpoint "mycustomdomain.example.com" (Resource Group "REDACTED"): Error enabling Custom Domain HTTPS for Frontend Endpoint: frontdoor.FrontendEndpointsClient#EnableHTTPS: Failure sending request: StatusCode=400 -- Original Error: Code="BadRequest" Message="\"minimumTlsVersion\" is a mandatory parameter."
```

### Steps to Reproduce

1. Setup CNAME DNS record to point to new Azure Frontdoor frontend
2. Run `terraform apply` with custom https provisioning enabled on custom domain

### References

The `minimumTlsVersion` parameter is required in the current API version:
https://docs.microsoft.com/en-us/rest/api/frontdoorservice/frontdoor/frontendendpoints/enablehttps

The SDK version was upgraded in #5385, thus upgrading the Frontdoor API version used. I believe this is the source of the bug.

sean-nixon:
Fixes #5538 

The latest version of the Front Door API requires a `minimumTlsVersion` property set in the request to enable custom domain HTTPS. This PR adds support for this property with a default of "1.0" to preserved backwards compatibility. Since September 2019, [Azure is making TLS 1.2 the default minimum.](https://docs.microsoft.com/en-us/azure/frontdoor/front-door-faq#what-tls-versions-are-supported-by-azure-front-door-service) Once version 2.0 is released, I think the default should be updated to 1.2 to align with Azure's defaults. I've included a TODO comment and note in the upgrade guide to mark this.

### DIFF
--- a/azurerm/internal/services/frontdoor/resource_arm_front_door.go
+++ b/azurerm/internal/services/frontdoor/resource_arm_front_door.go
@@ -408,6 +408,15 @@ func resourceArmFrontDoor() *schema.Resource {
 										}, false),
 										Default: string(frontdoor.CertificateSourceFrontDoor),
 									},
+									"minimum_tls_version": {
+										Type:     schema.TypeString,
+										Optional: true,
+										ValidateFunc: validation.StringInSlice([]string{
+											string(frontdoor.OneFullStopTwo),
+											string(frontdoor.OneFullStopZero),
+										}, false),
+										Default: string(frontdoor.OneFullStopTwo),
+									},
 									"provisioning_state": {
 										Type:     schema.TypeString,
 										Computed: true,
@@ -1265,6 +1274,8 @@ func flattenArmFrontDoorFrontendEndpoint(d *schema.ResourceData, input *[]frontd
 						chc["certificate_source"] = string(frontdoor.CertificateSourceFrontDoor)
 					}
 
+					chc["minimum_tls_version"] = string(customHTTPSConfiguration.MinimumTLSVersion)
+
 					if provisioningState := properties.CustomHTTPSProvisioningState; provisioningState != "" {
 						chc["provisioning_state"] = provisioningState
 						if provisioningState == frontdoor.CustomHTTPSProvisioningStateEnabled || provisioningState == frontdoor.CustomHTTPSProvisioningStateEnabling {
@@ -1502,8 +1513,11 @@ func makeCustomHttpsConfiguration(customHttpsConfiguration map[string]interface{
 	// https://github.com/Azure/azure-sdk-for-go/issues/6882
 	defaultProtocolType := "ServerNameIndication"
 
+	minTLSVersion := customHttpsConfiguration["minimum_tls_version"].(string)
+
 	customHTTPSConfigurationUpdate := frontdoor.CustomHTTPSConfiguration{
-		ProtocolType: &defaultProtocolType,
+		ProtocolType:      &defaultProtocolType,
+		MinimumTLSVersion: frontdoor.MinimumTLSVersion(minTLSVersion),
 	}
 
 	if customHttpsConfiguration["certificate_source"].(string) == "AzureKeyVault" {
--- a/azurerm/internal/services/frontdoor/resource_arm_front_door.go
+++ b/azurerm/internal/services/frontdoor/resource_arm_front_door.go
@@ -410,12 +410,7 @@ func resourceArmFrontDoor() *schema.Resource {
 									},
 									"minimum_tls_version": {
 										Type:     schema.TypeString,
-										Optional: true,
-										ValidateFunc: validation.StringInSlice([]string{
-											string(frontdoor.OneFullStopTwo),
-											string(frontdoor.OneFullStopZero),
-										}, false),
-										Default: string(frontdoor.OneFullStopZero), // TODO: Update default to TLS 1.2 in version 2.0
+										Computed: true,
 									},
 									"provisioning_state": {
 										Type:     schema.TypeString,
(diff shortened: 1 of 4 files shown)


WodansSon:
@sean-nixon I noticed that if you set the `TLS` to 1.0 then attempt to change it to `TLS` 1.2, the apply executes but the `TLS` does not get updated to 1.2 which ends up in a constant plan loop of always wanting to change the `TLS`. Do you think we should add a check to see if the certificate has already been deployed and then expose and error stating something like you need to revoke the 1.0 cert before you can issue a 1.2 cert?

sean-nixon:
@WodansSon Thanks for catching that issue with updates. I agree it might make sense to make the value computed given the Azure Portal behavior of completely hiding this. However, I would prefer not to always send 1.0. TLS 1.0 is no longer PCI compliant and a lot of organizations have policies to always disabling TLS 1.0, so it's not really an option for me personally.

What would you think of instead using the minimum TLS version config if it exists? If it doesn't exist already, we can then safely default to TLS 1.2 which would match the behavior in the Azure Portal.

Alternatively, I'm seeing if I can update the code to properly update the min TLS version. I think it may be possible (currently testing this), but the downside is it's essentially a complete redeployment of custom HTTPS certificates so it takes a long time. My test is still running after almost 2 hours, so I'm not sure if it will eventually error out.

WodansSon:
@sean-nixon yep, that totally makes sense... If the value is already set in azure keep sending whatever value that was be it 1.0 (for backward compat) or 1.2, and if it is not yet defined always send 1.2. I agree that It is kind of painful that it can take up to 8 hours to completely revoke the cert once it has been deployed, but on the bright side with the new version of the API it now takes only about 10 minutes to deploy a Front Door managed cert, where in the old API it could take up to 20 hours. 😄

timja:
Have you tested this and TLS 1.2 works with front door? I implemented this locally but when I scanned it with ssllabs it still said it was using 1.0 but possibly I was doing something wrong

timja:
(Even if it doesn’t work much better to fix creation of new endpoints, unfortunately when I was testing in the upgrade I must’ve just been updating them)

sean-nixon:
@timja I created two frontends, one with 1.2 and one with 1.0 and ran Qualys checks on both of them. It does seem like 1.0 is technically supported even if you select 1.2, but only with clients that don't support SNI. There's a marked difference between the two reports:

TLS 1.2 Frontend:
image


TLS 1.0 Frontend:
image

timja:
Hmm okay I didn't notice this when I was testing, otherwise I would have submitted it earlier, bit unexpected though

sean-nixon:
@WodansSon I took a stab at making the attribute computed. I also removed the `customHttpsConfiguration` from the main Create/Update request. It doesn't actually affect the resource unless specifically sending an enableHttps request using the frontend client, and it wouldn't be possible to send a sane default for `minimum_tls_version` without separating out the Create and Update functions which seemed excessive.

WodansSon:
@sean-nixon Agreed. Hope you don't mind, but I added a test to validate the functionality and pushed it to your branch. If the test passes I am going to merge this code and get it in for the v1.43 release of the provider. Thanks for your help, this is great! 🚀

sean-nixon:
@WodansSon I definitely don't mind. I hadn't added one because of the external dependencies. What are the typical practices around acceptance tests that have external dependencies that need to be set up before they can pass? For example, in this case a custom domain and CNAME record for validation would need to be created and the frontend name updated before the tests could pass.

WodansSon:
@sean-nixon those are more difficult, and the need to continually test it is made on a case by case basis. However, here we can work around that by using the FrontDoor managed Custom HTTPS cert provided by Microsoft, so as to not add any external dependencies. A truly custom HTTPS cert would be very difficult to code, so this way we can at least verify that the settings get set. 😃

sean-nixon:
Oh, I see what you did now. I misread your configuration. Does this actually succeed for you locally? I didn't think the default `*.azurefd.net` domain can have custom HTTPS settings applied to it. The default domain has HTTPS enabled by default by Azure. I believe I was getting errors for this a while back when trying to enable HTTPS using a provisioner and the Azure CLI.

WodansSon:
Yep, it works just fine for me... they may have changed something in the new API that allows this.

WodansSon:
image

PR #5542, opened 2020-01-28
compute: Preparation for the Virtual Machine Resources

tombuildsstuff:
This PR introduces a few things needed for the upcoming Linux/Windows Virtual Machine resources (as outlined in #2807).

Firstly this introduces the concept of the Features block which is Optional in 1.x but will become Required in 2.0. This will be used to define the behaviour of select resources (at this time only the new Virtual Machine resource [deleting the os disk on deletion], but in time the new Virtual Machine Scale Set [rolling instances] and Key Vault resources [soft delete]). What this looks like in practice:

```
provider "azurerm" {
    features {      
      virtual_machine {
        delete_os_disk_on_deletion = true
      }
    }
}
```

(since there's no Acceptance Tests covering the Provider block directly, I've added unit tests covering the expand in this instance)

Secondly this introduces a bunch of Parsers/Validators across both Compute and Networking for resources which are required by the Virtual Machine resource - but it /intentionally/ doesn't retrofit these into other resources since that's a larger/out-of-scope task.

Finally this introduces some helper functions which are required by the upcoming Virtual Machine resources - mostly to make that PR easier to review, since it's a large resource.

### DIFF
--- a/azurerm/internal/services/compute/virtual_machine_instance.go
+++ b/azurerm/internal/services/compute/virtual_machine_instance.go
@@ -1,0 +1,36 @@
+package compute
+
+import (
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+)
+
+// shouldBootVirtualMachine determines if the Virtual Machine should be started after
+// the Virtual Machine has been shut down for maintainence. This means that Virtual Machines
+// which are already stopped can be updated but will not be started
+func shouldBootVirtualMachine(instanceView compute.VirtualMachineInstanceView) bool {
+	if instanceView.Statuses != nil {
+		for _, status := range *instanceView.Statuses {
+			if status.Code == nil {
+				continue
+			}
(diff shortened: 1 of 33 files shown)


azurerm/internal/services/compute/virtual_machine_instance.go
```diff
@@ -0,0 +1,37 @@
...
+			state := strings.ToLower(*status.Code)
+			if !strings.HasPrefix(state, "PowerState/") {
+				continue
+			}
+
+			state = strings.TrimPrefix(state, "powerstate/")
+			switch strings.ToLower(state) {
+			case "deallocating":
+			case "deallocated":
+			case "stopped":
```
jackofallops: Might need/want to add `stopping` here?

tombuildsstuff: agreed - rather than pushing to this branch & rebasing #5550 I'll make this change there since it's dependent on this one

PR #5562, opened 2020-01-30
r/(linux|windows)_virtual_machine_scale_set: using the features block to control roll instances

tombuildsstuff:
This PR switches the `azurerm_linux_virtual_machine_scale_set` and `azurerm_windows_virtual_machine_scale_set` resources over to using the `features` block to control rolling instances when required, rather than using a property on the resources (since this won't be present during import).

This allows the feature to be controlled via:

```
provider "azurerm" {
  features {
    virtual_machine_scale_set {
      roll_instances_when_required = true|false
    }
  }
}
```

The `features` block itself will be documented along with the beta documentation in the not-too-distant future.

### DIFF
--- a/azurerm/internal/provider/features.go
+++ b/azurerm/internal/provider/features.go
@@ -26,6 +26,20 @@ func schemaFeatures() *schema.Schema {
 						},
 					},
 				},
+
+				"virtual_machine_scale_set": {
+					Type:     schema.TypeList,
+					Optional: true,
+					MaxItems: 1,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"roll_instances_when_required": {
+								Type:     schema.TypeBool,
+								Required: true,
+							},
+						},
+					},
+				},
 			},
 		},
 	}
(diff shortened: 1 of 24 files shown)


azurerm/internal/provider/features.go
```diff
@@ -26,6 +26,20 @@ func schemaFeatures() *schema.Schema {
...
+
+				"virtual_machine_scale_set": {
+					Type:     schema.TypeList,
+					Optional: true,
+					MaxItems: 1,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"roll_instances_when_required": {
+								Type:     schema.TypeBool,
+								Required: true,
```
katbyte: Should this be optional for when more options go in, or is the intention to require all provider features to be explicitly states in every provider block?

katbyte:
Would it perhaps make sense to use `meta.(*clients.Client).Compute.Feature.DeleteOSDiskOnDeletion` for features 🤔

tombuildsstuff:
> Would it perhaps make sense to use meta.(*clients.Client).Compute.Feature.DeleteOSDiskOnDeletion for features 🤔

Potentially in time - my thinking for the moment is that these /could/ span multiple clients (e.g. mssql vs sql, frontdoor vs networking - all of which need to be refactored in time) - until we've done that I think it's easiest for these to stay in the same package for the moment (and it's fairly trivial to move these into packages at a later date if we want to do that?)

azurerm/internal/provider/features.go
tombuildsstuff: that makes sense once we've added a secondary field here (which is why the parsing logic below handles this) but since there's only a single element it didn't feel worth making this field optional for the moment (since otherwise you may as well omit the `virtual_machine_scale_set` block?)

tombuildsstuff:
Ignoring some expected test failures (since the API's broken, with a disclaimer in the docs) the tests pass:

Screenshot 2020-02-03 at 10 54 03

Issue #5434, opened 2020-01-17
Destroy storage account with network rules is impossible (hangs forever)


KrisPetkov:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
* Terraform v0.11.11
* Provider: v1.41

### Affected Resource(s)

* `azurerm_storage_account`

### Terraform Configuration Files

```hcl
provider "azurerm" {
  subscription_id = "${var.azure_subscription_id}"
  client_id = "${var.azure_client_id}"
  client_secret = "${var.azure_client_secret}"
  tenant_id = "${var.azure_tenant_id}"
}

resource "azurerm_storage_account" "my_storage_azure" {
  name                      = "${var.myazurestorage_name}"
  resource_group_name       = "${var.resource_group_name}"
  location                  = "${var.location}"
  account_kind              = "StorageV2"
  account_tier              = "Standard"
  access_tier               = "Hot"
  account_replication_type  = "ZRS"
  enable_https_traffic_only = true

  network_rules {
    default_action             = "Deny"
    virtual_network_subnet_ids = ["${var.platform_subnets}","${var.bosh_subnets}"]
  }
}

resource "azurerm_postgresql_server" "my_postgresql_server" {
  name                = "${var.mypostgredb_name}"
  location            = "${var.location}"
  resource_group_name = "${var.resource_group_name}"

  sku {
    name     = "GP_Gen5_2"
    capacity = 2
    tier     = "GeneralPurpose"
    family   = "Gen5"
  }

  storage_profile {
    storage_mb            = 5120
    backup_retention_days = 14
    geo_redundant_backup  = "Disabled"
  }

  administrator_login          = "vcap"
  administrator_login_password = "${var.db_adm_password}"
  version                      = "10"
  ssl_enforcement              = "Enabled"
}

resource "azurerm_postgresql_database" "my_postgre_db" {
  name                = "mypostgredb"
  resource_group_name = "${var.resource_group_name}"
  server_name         = "${azurerm_postgresql_server.my_postgresql_server.name}"
  charset             = "UTF8"
[...]

Issue #3156, opened 2019-04-02
Storage Accounts not destroying (1hr+)


runecalico:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

.\terraform.exe -v
Terraform v0.11.13
+ provider.azurerm v1.23.0

### Affected Resource(s)

* `azurerm_storage_account`

### Terraform Configuration Files

```hcl

resource "azurerm_storage_account" "repo" {
  depends_on = ["module.databricks"]
  name       = "fzeu2dtabrklinpdsto00"

  resource_group_name       = "${module.databricks.databricks_rg}"
  location                  = "${var.databricks_region}"
  account_tier              = "Standard"
  account_kind              = "StorageV2"
  account_replication_type  = "LRS"
  access_tier               = "Hot"
  enable_blob_encryption    = "True"
  enable_file_encryption    = "True"
  enable_https_traffic_only = "True"

  network_rules = {
    ip_rules                   = ["${module.azure_lookups.ip_rules}"]
    virtual_network_subnet_ids = ["${local.virtual_network_subnet_ids}"]
    bypass                     = ["AzureServices"]
  }

  tags {
    Product         = "${var.product_key}"
    ApplicationName = "${var.application_name}"

    #CostCenter      = "${var.costcenter}"


    #DeployScenario  = "${var.deploy_scenario}"

    Environment = "${var.environment}"
    CreatedBy   = "${var.created_by}"
    Version     = "${var.version}"
  }
}

resource "azurerm_storage_container" "repo_containers" {
  count = "${length(local.repo_containers)}"

  name                  = "${local.repo_containers[count.index]}"
  resource_group_name   = "${module.databricks.databricks_rg}"
  storage_account_name  = "${azurerm_storage_account.repo.name}"
  container_access_type = "private"
}

#Blob storage account for data scientist workspace
resource "azurerm_storage_account" "sandbox" {
  depends_on                = ["azurerm_storage_account.repo"]
  name                      = "fzeu2dtabrksbnpdsto01"
  resource_group_name       = "${module.databricks.databricks_rg}"
[...]

PR #5563, opened 2020-01-30
Fix for issue 5434

KrisPetkov:
Fix for issue #5434 (and related issues: #5530 , #3156 )

* **Problem:** If you have a storage account with more than one network rule set - the destroy action hangs forever and the real destroy/deletion of the resource never happens. If you have a storage account with zero or one network rule attached then this problem is irrelevant - the destroy action completes successfully.

* **Reason:** In the `resourceArmStorageAccountDelete` method located in _azurerm/internal/services/storage/resource_arm_storage_account.go_ there is a string array which is passed to the mutex lock method `locks.MultipleByName`. The reason for the problem is that the creation of this array does not cover all possible cases. The logic of the method assumes that you have different virtual networks and inserts them into the array used for locking but does not work correctly in the corner case where you have one virtual network with few sub-networks and you are assigning these sub-networks as a network rules to your storage account. Hence the array passed to the lock method contains n-times strings with the same name (same strings). When this array goes to the mutex the first string is used to lock, the second string is used to lock etc. but at the end all the strings are the same and when the first unlock comes - everything is OK but when a second, third or what so ever unlock hits... dead lock there is nothing to unlock because the first unlock already used the string and it was removed and all other unlocks are trying to use something that simply does not exists.

* **Resolution:** Check if there are subnets and attach them to the network name hence all the strings used to populate the array are unique and the lock-unlock problem is resolved.

### DIFF
--- a/azurerm/internal/services/storage/resource_arm_storage_account.go
+++ b/azurerm/internal/services/storage/resource_arm_storage_account.go
@@ -1305,6 +1305,12 @@ func resourceArmStorageAccountDelete(d *schema.ResourceData, meta interface{}) e
 					}
 
 					networkName := id.Path["virtualNetworks"]
+
+					subnetName := id.Path["subnets"]
+					if subnetName != "" {
+						networkName = fmt.Sprintf("%s-%s", networkName, subnetName)
+					}
+
 					for _, virtualNetworkName := range virtualNetworkNames {
 						if networkName == virtualNetworkName {
 							continue


KrisPetkov:
@tombuildsstuff  I'm sorry but I just have noticed that I made the commit with one of my accounts and made the PR with my other account. I don't know if this is a problem. If it is I will make a new commit and new PR.

KrisPetkov:
I will close it and make a new PR

PR #5565, opened 2020-01-30
azurerm_storage_account - fix hanging destroy caused by multiple network rules 

KrisPetkov:
Fix for issue #5434 (and related issues: #5530 , #3156 )

* **Problem:** If you have a storage account with more than one network rule set - the destroy action hangs forever and the real destroy/deletion of the resource never happens. If you have a storage account with zero or one network rule attached then this problem is irrelevant - the destroy action completes successfully.

* **Reason:** In the `resourceArmStorageAccountDelete` method located in _azurerm/internal/services/storage/resource_arm_storage_account.go_ there is a string array which is passed to the mutex lock method `locks.MultipleByName`. The reason for the problem is that the creation of this array does not cover all possible cases. The logic of the method assumes that you have different virtual networks and inserts them into the array used for locking but does not work correctly in the corner case where you have one virtual network with few sub-networks and you are assigning these sub-networks as a network rules to your storage account. Hence the array passed to the lock method contains n-times strings with the same name (same strings). When this array goes to the mutex the first string is used to lock, the second string is used to lock etc. but at the end all the strings are the same and when the first unlock comes - everything is OK but when a second, third or what so ever unlock hits... dead lock there is nothing to unlock because the first unlock already used the string and it was removed and all other unlocks are trying to use something that simply does not exists.

* **Resolution:** Check if there are subnets and attach them to the network name hence all the strings used to populate the array are unique and the lock-unlock problem is resolved.

### DIFF
--- a/azurerm/internal/locks/lock.go
+++ b/azurerm/internal/locks/lock.go
@@ -1,9 +1,8 @@
 package locks
 
 import (
-	"fmt"
-
 	"github.com/hashicorp/terraform-plugin-sdk/helper/mutexkv"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/common"
 )
 
 // armMutexKV is the instance of MutexKV for ARM resources
@@ -20,11 +19,10 @@ func ByName(name string, resourceType string) {
 }
 
 func MultipleByName(names *[]string, resourceType string) {
-	for i, name := range *names {
-		// at the end of every array item add its index. This way we guarantee that this item will be unique (no duplicates are possible)
-		uniqueValue := fmt.Sprintf("%s-arrIdx-%d", name, i)
+	newSlice := common.RemoveDuplicatesFromStringArray(*names)
 
-		ByName(uniqueValue, resourceType)
+	for _, name := range newSlice {
+		ByName(name, resourceType)
 	}
 }
 
@@ -38,10 +36,9 @@ func UnlockByName(name string, resourceType string) {
 }
 
 func UnlockMultipleByName(names *[]string, resourceType string) {
-	for i, name := range *names {
-		// at the end of every array item add its index. We need to add this sufix because it was added during the lock process
-		uniqueValue := fmt.Sprintf("%s-arrIdx-%d", name, i)
+	newSlice := common.RemoveDuplicatesFromStringArray(*names)
 
-		UnlockByName(uniqueValue, resourceType)
+	for _, name := range newSlice {
+		UnlockByName(name, resourceType)
 	}
 }
--- a/azurerm/internal/locks/lock.go
+++ b/azurerm/internal/locks/lock.go
@@ -1,6 +1,10 @@
 package locks
 
-import "github.com/hashicorp/terraform-plugin-sdk/helper/mutexkv"
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/mutexkv"
+)
 
 // armMutexKV is the instance of MutexKV for ARM resources
 var armMutexKV = mutexkv.NewMutexKV()
@@ -16,8 +20,11 @@ func ByName(name string, resourceType string) {
 }
 
 func MultipleByName(names *[]string, resourceType string) {
-	for _, name := range *names {
-		ByName(name, resourceType)
+	for i, name := range *names {
+		// at the end of every array item add its index. This way we guarantee that this item will be unique (no duplicates are possible)
+		uniqueValue := fmt.Sprintf("%s-arrIdx-%d", name, i)
+
+		ByName(uniqueValue, resourceType)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


KrisPetkov:
Hi @katbyte 

Ok, I can make this change. So your idea is to change the `locks.MultipleByName` method so if there are any duplicate strings, they should be skipped or be handled like adding some suffix so they become unique?

KrisPetkov:
Hi @katbyte 

I have provided a new commit with a more generic solution of the problem. Please have a look and share your opinion. I was thinking to make the solution more complex by providing an additional method that will check if the passed array have duplicated elements and then to execute the logic from my last commit. But I don't know if it worth to make it more complex (to introduce a hole new method which will iterate all over the passed array and seeks for duplicates)?

azurerm/internal/locks/lock.go
```diff
@@ -16,8 +20,11 @@ func ByName(name string, resourceType string) {
 }
 
 func MultipleByName(names *[]string, resourceType string) {
-	for _, name := range *names {
-		ByName(name, resourceType)
+	for i, name := range *names {
+		// at the end of every array item add its index. This way we guarantee that this item will be unique (no duplicates are possible)
+		uniqueValue := fmt.Sprintf("%s-arrIdx-%d", name, i)
```
katbyte: This has the same issue as the previous solution, if two resources call this with resource N, but they have different indexes we could deadlock. What we should do is just ensure there are no duplicates:
```
func SliceUniqString(elements []string) []sring {
	seen := make(map[int]bool, len(s))
	j := 0
	for _, v := range s {
		if _, ok := seen[v]; ok {
			continue
		}
		seen[v] = true
		elements[j] = v
		j++
	}
	return elements[:j]
}
```

katbyte: This also abstracts out the actual lock names, which could introduce hard to debug issues

azurerm/internal/locks/lock.go
```diff
@@ -31,7 +38,10 @@ func UnlockByName(name string, resourceType string) {
 }
 
 func UnlockMultipleByName(names *[]string, resourceType string) {
-	for _, name := range *names {
-		UnlockByName(name, resourceType)
+	for i, name := range *names {
+		// at the end of every array item add its index. We need to add this sufix because it was added during the lock process
+		uniqueValue := fmt.Sprintf("%s-arrIdx-%d", name, i)
```
katbyte: And then we don't need to worry about it here, unlocking twice in a row i think should be fine?

azurerm/internal/locks/lock.go
KrisPetkov: Right, I got your idea. I have a prepared similar solution. I was trying to avoid the deletion/remove of duplicated strings and that's why I proposed my previous solution but maybe you are right and this is the right way. I have committed the new changes

azurerm/internal/locks/lock.go
KrisPetkov: Actually we need to handle the change here, also, because if you try to unlock already unlocked mutex (by using the same lock string) you will get the following error (I tested this scenarion few times):
`fatal error: sync: unlock of unlocked mutex`

katbyte: cool thaks for checking that

azurerm/internal/locks/lock.go
KrisPetkov: @katbyte I know that you are busy working on v2.0.0 but just wanted to ask if I need to make some other changes or the last ones are OK? And do you think that this fix will be merged in the next v1.44? I'm asking because it is an essential bug for our team and currently blocking some of our tasks and I just want to have some more information.

katbyte:
Fix for issue #5434 (and related issues: #5530 , #3156 )

* **Problem:** If you have a storage account with more than one network rule set - the destroy action hangs forever and the real destroy/deletion of the resource never happens. If you have a storage account with zero or one network rule attached then this problem is irrelevant - the destroy action completes successfully.

* **Reason:** In the `resourceArmStorageAccountDelete` method located in _azurerm/internal/services/storage/resource_arm_storage_account.go_ there is a string array which is passed to the mutex lock method `locks.MultipleByName`. The reason for the problem is that the creation of this array does not cover all possible cases. The logic of the method assumes that you have different virtual networks and inserts them into the array used for locking but does not work correctly in the corner case where you have one virtual network with few sub-networks and you are assigning these sub-networks as a network rules to your storage account. Hence the array passed to the lock method contains n-times strings with the same name (same strings). When this array goes to the mutex the first string is used to lock, the second string is used to lock etc. but at the end all the strings are the same and when the first unlock comes - everything is OK but when a second, third or what so ever unlock hits... dead lock there is nothing to unlock because the first unlock already used the string and it was removed and all other unlocks are trying to use something that simply does not exists.

* **Resolution:** Check if there are subnets and attach them to the network name hence all the strings used to populate the array are unique and the lock-unlock problem is resolved.

PR #5570, opened 2020-01-30
`network_security_rule`: suppress case diff on `resource_group_name`

notchairmk:
This PR addresses an issue where a network security rule can be created with `resource_group_name` having different casing than the network security group to which it's being attached.

For example
```
resource "azurerm_network_security_group" "example" {
  resource_group_name = "acctestRG-123"
  ...
}

resource "azurerm_network_security_rule" "example" {
  resource_group_name = "acctestrg-123"
  ...
}
```
would successfully apply, but produces non-empty plan after apply.

### DIFF
--- a/azurerm/internal/services/network/resource_arm_network_security_rule.go
+++ b/azurerm/internal/services/network/resource_arm_network_security_rule.go
@@ -41,7 +41,7 @@ func resourceArmNetworkSecurityRule() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"resource_group_name": azure.SchemaResourceGroupName(),
+			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
 
 			"network_security_group_name": {
 				Type:     schema.TypeString,
--- a/azurerm/internal/services/network/tests/resource_arm_network_security_rule_test.go
+++ b/azurerm/internal/services/network/tests/resource_arm_network_security_rule_test.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"fmt"
 	"net/http"
+	"strings"
 	"testing"
 
 	"github.com/hashicorp/go-azure-helpers/response"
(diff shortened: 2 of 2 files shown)


katbyte:
Hi @notchairmk,

This is by design and I'm not sure why this would be required? Unless absolutely necessary due to an api bug we default to case sensitive to ward off strange unforseen behaviours. If you address back to the original resource:
```
resource "azurerm_network_security_group" "example" {
  resource_group_name = "acctestRG-123"
  ...
}

resource "azurerm_network_security_rule" "example" {
  resource_group_name = azurerm_network_security_group.example.resource_group_name
  ...
}
```

you are guaranteed to get the same & correct case.

notchairmk:
It does seem like a potential API bug, specifically in that you can create a network security group with inconsistent resource group name case compared to the resource group it exists within. It's not possible within the Azure CLI or portal, those only allow exact matches.

The issue we're running into is a network security group not being managed by Terraform, instead created by an Azure service. The case of the resource group name within the NSG resource doesn't match the resource group name case.

So maybe a better example would be:
```
data "azurerm_resource_group" "example" {
  resource_group_name = "acctestRG"
}

# equivalent of below NSG exists in `acctestRG` resource group
#
# resource "azurerm_network_security_group" "example" {
#  resource_group_name = "acctestrg"
#  name = "acctestNSG"
#  ...
# }

resource "azurerm_network_security_rule" "example" {
  resource_group_name = data.azurerm_resource_group.example.name
  network_security_group_name = "acctestNSG"
  ...
}
``` 
This example never converges.

It's totally possible that this is the wrong approach to take, if instead it's a bug that should be fixed on the API side.

katbyte:
@notchairmk, 100% this is a bug that should to be fixed on the API side. What service is this if i may ask?

notchairmk:
@katbyte The service in this case is AKS. 

But the bug itself appears to be in the response coming back from hitting the `securityRules` REST endpoint. The ID of the resource coming back has a resource group with case matching the NSG it's attached to, rather than the name of resource group the rule was created with. 

As a comparison, this behavior isn't popping up in e.g. virtual networks/subnets. We can create a virtual network with `resource_group_name = "acctestVNET"` and a subnet with `resource_group_name = "acctestvnet"` and the resource IDs returned contain `acctestVNET` and `acctestvnet` respectively.

notchairmk:
Opened [Azure/azure-rest-api-specs#8309](https://github.com/Azure/azure-rest-api-specs/issues/8309), which has been closed out, saying it would require coordination between ARM and resource providers. Is there anywhere else upstream I could open an issue to get it resolved?

katbyte:
@notchairmk, i typically open an issue on the go sdk

katbyte:
@notchairmk, i reopened your issue on the sdk: https://github.com/Azure/azure-sdk-for-go/issues/7738, so i'm going to close this for now. do feel free to open  an issue if you'd like to keep visibility or discuss more.

PR #5579, opened 2020-01-31
Updated managed_disk resource for disk_size_gb updates

jackofallops:
Works to address https://github.com/terraform-providers/terraform-provider-azurerm/issues/2534

This resource should be used with `azurerm_virtual_machine_data_disk_attachment` for correct behaviour in https://github.com/terraform-providers/terraform-provider-azurerm/issues/1668 also.

### DIFF
--- a/azurerm/internal/services/compute/resource_arm_managed_disk.go
+++ b/azurerm/internal/services/compute/resource_arm_managed_disk.go
@@ -15,6 +15,7 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/compute/resource_arm_managed_disk.go
```diff
@@ -281,28 +282,212 @@ func resourceArmManagedDiskCreateUpdate(d *schema.ResourceData, meta interface{}
...
+	// if we are attached to a VM we bring down the VM as necessary for the operations which are not allowed while it's online
+	if disk.ManagedBy != nil {
+		virtualMachine, err := ParseVirtualMachineID(*disk.ManagedBy)
+		if err != nil {
+			return fmt.Errorf("Error parsing VMID %q for disk attachment: %+v", *disk.ManagedBy, err)
+		}
+		// check instanceView State
+		vmClient := meta.(*clients.Client).Compute.VMClient
+
+		instanceView, err := vmClient.InstanceView(ctx, virtualMachine.ResourceGroup, virtualMachine.Name)
```
tombuildsstuff: we need to lock on the Virtual Machine resource group/name here before doing these manipulations

PR #5590, opened 2020-02-01
azurerm_lb - support for the `private_ip_address_version` property

smarterclayton:
Azure allows load balancers to have an allocated private ip address
version in their frontend configuration to select between IPv4 and
IPv6, like NICs. Make setting the version on load balancers consistent
with NICs.


azurerm/internal/services/network/data_source_loadbalancer.go
```diff
@@ -141,10 +147,14 @@ func dataSourceArmLoadBalancerRead(d *schema.ResourceData, meta interface{}) err
...
 						privateIpAddresses = append(privateIpAddresses, *feipProps.PrivateIPAddress)
 					}
+					if privateIpAddressVersion == "" {
+						privateIpAddressVersion = string(feipProps.PrivateIPAddressVersion)
+					}
 				}
 			}
 
 			d.Set("private_ip_address", privateIpAddress)
+			d.Set("private_ip_address_version", privateIpAddressVersion)
```
mbfrahry: Setting `private_ip_address_version` at this level causes a crash because it lives under `frontend_ip_configuration`. I believe we can just remove this line

azurerm/internal/services/network/resource_arm_loadbalancer.go
```diff
@@ -279,11 +290,13 @@ func resourceArmLoadBalancerRead(d *schema.ResourceData, meta interface{}) error
 
 						privateIpAddresses = append(privateIpAddresses, *feipProps.PrivateIPAddress)
 					}
+					privateIpAddressVersion = string(feipProps.PrivateIPAddressVersion)
 				}
 			}
 
 			d.Set("private_ip_address", privateIpAddress)
 			d.Set("private_ip_addresses", privateIpAddresses)
+			d.Set("private_ip_address_version", privateIpAddressVersion)
```
mbfrahry: Same here

smarterclayton:
Updated, PTAL

Issue #3621, opened 2019-06-09
Feature Request: Add support for HTTP2 to azurerm_api_management


NickMetz:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
It should be possible to enable HTTP/2 Client-side protocol in azurerm_api_management (default it is disabled). 

![apmi](https://user-images.githubusercontent.com/26707368/59165102-a973f100-8b16-11e9-94f5-7bca8ecd6d3d.png)

Please be aware that there is currently a bug #3125 in security block, that may will also impact this feature request. 

**UPDATE:**  Maby we can separate this setting to a new block, it seems not to rely on Security see api call.

### Affected Resource(s)

* azurerm_api_management

### Potential Terraform Configuration

```hcl
resource "azurerm_api_management" "test" {
  name                = "example-apim"
  location            = "${azurerm_resource_group.test.location}"
  resource_group_name = "${azurerm_resource_group.test.name}"
  publisher_name      = "My Company"
  publisher_email     = "company@terraform.io"

  sku {
    name     = "Developer"
    capacity = 1
   }
  
   protocols {
      enable_http2 = false
   }
}

```
This is how it looks like on API (default is false)
```
    ],
    "privateIPAddresses": null,
    "additionalLocations": null,
    "virtualNetworkConfiguration": null,
    "customProperties": {
      "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Backend.Protocols.Ssl30": "false",
      "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Backend.Protocols.Tls10": "false",
      "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Backend.Protocols.Tls11": "false",
      "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Ciphers.TripleDes168": "false",
      "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Protocols.Ssl30": "false",
      "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Protocols.Tls10": "false",
[...]

PR #5593, opened 2020-02-03
`azurerm_api_management` - Add http2 protocol support for API Management

wasfree:
This PR will fix issue #3621 and add support to enable http2 protocol for API Management.

### DIFF
--- a/azurerm/internal/services/apimanagement/resource_arm_api_management.go
+++ b/azurerm/internal/services/apimanagement/resource_arm_api_management.go
@@ -213,7 +213,7 @@ func resourceArmApiManagementService() *schema.Resource {
 						"enable_http2": {
 							Type:     schema.TypeBool,
 							Optional: true,
-							Computed: true,
+							Default:  false,
 						},
 					},
 				},
--- a/azurerm/internal/services/apimanagement/resource_arm_api_management.go
+++ b/azurerm/internal/services/apimanagement/resource_arm_api_management.go
@@ -28,6 +28,7 @@ var apimFrontendProtocolSsl3 = "Microsoft.WindowsAzure.ApiManagement.Gateway.Sec
 var apimFrontendProtocolTls10 = "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Protocols.Tls10"
 var apimFrontendProtocolTls11 = "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Protocols.Tls11"
 var apimTripleDesCiphers = "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Ciphers.TripleDes168"
+var apimHttp2Protocol = "Microsoft.WindowsAzure.ApiManagement.Gateway.Protocols.Server.Http2"
 
 func resourceArmApiManagementService() *schema.Resource {
 	return &schema.Resource{
@@ -196,6 +197,23 @@ func resourceArmApiManagementService() *schema.Resource {
 								string(apimanagement.CertificateAuthority),
 								string(apimanagement.Root),
 							}, false),
+						},
+					},
+				},
+			},
+
+			"protocols": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+
+						"enable_http2": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Computed: true,
 						},
 					},
 				},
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/apimanagement/resource_arm_api_management.go
```diff
@@ -201,6 +202,23 @@ func resourceArmApiManagementService() *schema.Resource {
...
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+
+						"enable_http2": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Computed: true,
```
mbfrahry: How do you feel about removing `Computed: true` and instead just have`Default: false` so it matches the way the api is implemented? Optional Computed can get a little wonky so we should limit it where we can.

website/docs/r/api_management.html.markdown
```diff
@@ -163,6 +165,12 @@ A `proxy` block supports the following:
 
 ---
 
+A `protocols` block supports the following:
+
+* `enable_http2` - (Optional) Should HTTP/2 support be enabled on the client-facing side of the gateway?
```
mbfrahry: If we do add `Default: false` we should make a note of it here as well that it defaults to false.

mbfrahry (changes requested):
Hey @NickMetz. This change looks good with just a minor suggestion to better match the api

azurerm/internal/services/apimanagement/resource_arm_api_management.go
```diff
@@ -201,6 +202,23 @@ func resourceArmApiManagementService() *schema.Resource {
...
 			},
 
+			"protocols": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+
```
tombuildsstuff: I have a feeling the linting will want this line removed?

```suggestion
```

website/docs/r/api_management.html.markdown
```diff
@@ -163,6 +165,12 @@ A `proxy` block supports the following:
 
 ---
 
+A `protocols` block supports the following:
+
+* `enable_http2` - (Optional) Should HTTP/2 support be enabled on the client-facing side of the gateway? Defaults to `false`.
```
tombuildsstuff: just double-checking here since [the key is `Server.Http2`](https://github.com/terraform-providers/terraform-provider-azurerm/pull/5593/files#diff-6c4c0fbff8f080a23c6180cb0ab1e6d0R31)?

azurerm/internal/services/apimanagement/resource_arm_api_management.go
```diff
@@ -201,6 +202,23 @@ func resourceArmApiManagementService() *schema.Resource {
 				},
 			},
 
+			"protocols": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
tombuildsstuff: can we add a TODO to remove this in 2.0?

```suggestion
				Computed: true, //  TODO: remove in 2.0
```

tombuildsstuff (changes requested):
hey @NickMetz 

Thanks for pushing those changes - I've taken a look through and left some comments inline but this is looking good to me, if we can fix up the remaining ones (and the tests pass) then this should otherwise be good 👍 

Thanks!

tombuildsstuff:
Acceptance tests look fine:

Screenshot 2020-02-10 at 12 18 34

website/docs/r/api_management.html.markdown
NickMetz: Hi @tombuildsstuff, 
thanks for your feedback. So your suggestion is to rename `enable_http2` to `server_http2` ?

azurerm/internal/services/apimanagement/resource_arm_api_management.go
NickMetz: Hi @mbfrahry,
thanks for your comment. It's now set to `Default: false`.

website/docs/r/api_management.html.markdown
NickMetz: Added Defaults to `false` in documentation.

website/docs/r/api_management.html.markdown
tombuildsstuff: as discussed offline, we should be able to update the documentation slightly here - but this is otherwise fine as-is

azurerm/internal/services/apimanagement/resource_arm_api_management.go
NickMetz: Removed line

azurerm/internal/services/apimanagement/resource_arm_api_management.go
NickMetz: Added comment as suggested.

website/docs/r/api_management.html.markdown
NickMetz: Updated description to: Should HTTP/2 be supported by the API Management Service? Defaults to `false`.

NickMetz:
Hi @tombuildsstuff , @mbfrahry,
all required changes has been done and acceptance tests passed.

mbfrahry (approved):
LGTM!

PR #5597, opened 2020-02-03
r/(linux|windows)_virtual_machine_scale_set: ultra ssd's now need to …

tombuildsstuff:
…be in an availability zone:

Looks like a behavioural change in the API since the resource was first built:

```
Code="InvalidParameter" Message="Feature 'UltraSSD' can be used only with Virtual Machine Scale Sets in an Availability Zone." Target="additionalCapabilities.ultraSSDEnabled"
```


tombuildsstuff:
Linux VMSS passes:

Screenshot 2020-02-03 at 19 51 01

tombuildsstuff:
Windows VMSS passes:

Screenshot 2020-02-03 at 20 17 59

Issue #5553, opened 2020-01-29
Support for kusto EventGrid


noam-shapira:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Add support for kusto EventGrid 

### New or Affected Resource(s)

* azurerm_kusto

### References
https://docs.microsoft.com/bs-latn-ba/azure/data-explorer/ingest-data-event-grid

* #0000

favoretti:
Since this issue seems to have been addressed in the latest versions of the provider - I'm going to close it. Please open a new updated bug report if this is still relevant. Thank you.

PR #5607, opened 2020-02-04
New Resource: `azurerm_kusto_eventgrid_data_connection`

noam-shapira:
add event grid data connection for azure kusto service 

#5553 add event grid data connection

### DIFF
--- a/azurerm/internal/services/kusto/registration.go
+++ b/azurerm/internal/services/kusto/registration.go
@@ -19,9 +19,10 @@ func (r Registration) SupportedDataSources() map[string]*schema.Resource {
 // SupportedResources returns the supported Resources supported by this Service
 func (r Registration) SupportedResources() map[string]*schema.Resource {
 	return map[string]*schema.Resource{
-		"azurerm_kusto_cluster":                  resourceArmKustoCluster(),
-		"azurerm_kusto_database":                 resourceArmKustoDatabase(),
-		"azurerm_kusto_database_principal":       resourceArmKustoDatabasePrincipal(),
-		"azurerm_kusto_eventhub_data_connection": resourceArmKustoEventHubDataConnection(),
+		"azurerm_kusto_cluster":                   resourceArmKustoCluster(),
+		"azurerm_kusto_database":                  resourceArmKustoDatabase(),
+		"azurerm_kusto_database_principal":        resourceArmKustoDatabasePrincipal(),
+		"azurerm_kusto_eventhub_data_connection":  resourceArmKustoEventHubDataConnection(),
+		"azurerm_kusto_eventgrid_data_connection": resourceArmKustoEventGridDataConnection(),
 	}
 }
--- a/azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
+++ b/azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
@@ -1,0 +1,285 @@
+package kusto
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/kusto/mgmt/2019-05-15/kusto"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func resourceArmKustoEventGridDataConnection() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceArmKustoEventGridDataConnectionCreateUpdate,
+		Read:   resourceArmKustoEventGridDataConnectionRead,
+		Update: resourceArmKustoEventGridDataConnectionCreateUpdate,
+		Delete: resourceArmKustoEventGridDataConnectionDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(60 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(60 * time.Minute),
+			Delete: schema.DefaultTimeout(60 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateAzureRMKustoDataConnectionName,
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"location": azure.SchemaLocation(),
+
+			"cluster_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateAzureRMKustoClusterName,
+			},
+
+			"database_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateAzureRMKustoDatabaseName,
+			},
+
+			"storage_account_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: azure.ValidateResourceID,
+			},
+
+			"eventhub_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: azure.ValidateResourceID,
+			},
+
+			"consumer_group": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: azure.ValidateEventHubConsumerName(),
+			},
+
+			"table_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validateAzureRMKustoEntityName,
+			},
+
+			"mapping_rule_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validateAzureRMKustoEntityName,
+			},
+
+			"data_format": {
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(kusto.MULTIJSON),
+					string(kusto.JSON),
+					string(kusto.CSV),
+					string(kusto.TSV),
+					string(kusto.SCSV),
+					string(kusto.SOHSV),
+					string(kusto.PSV),
+					string(kusto.TXT),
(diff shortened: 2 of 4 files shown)


azurerm/internal/services/kusto/registration.go
```diff
@@ -19,9 +19,10 @@ func (r Registration) SupportedDataSources() map[string]*schema.Resource {
...
 	return map[string]*schema.Resource{
-		"azurerm_kusto_cluster":                  resourceArmKustoCluster(),
-		"azurerm_kusto_database":                 resourceArmKustoDatabase(),
-		"azurerm_kusto_database_principal":       resourceArmKustoDatabasePrincipal(),
-		"azurerm_kusto_eventhub_data_connection": resourceArmKustoEventHubDataConnection(),
+		"azurerm_kusto_cluster":                   resourceArmKustoCluster(),
+		"azurerm_kusto_database":                  resourceArmKustoDatabase(),
+		"azurerm_kusto_database_principal":        resourceArmKustoDatabasePrincipal(),
+		"azurerm_kusto_eventhub_data_connection":  resourceArmKustoEventHubDataConnection(),
+		"azurerm_kusto_eventgrid_data_connection": resourceArmKustoEventGridDataConnection(),
```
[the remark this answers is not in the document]
noam-shapira: Hey @mbfrahry thanks for comments, I will fix the order.

Regards accept test, 
Eventgrid data connection cannot be created without target table. 
Kusto RM doesn't provide a way to create table ( as it not part of the control plan)
So, I can't add test that doesn't depend on external resource.

@mbfrahry what do you recommend?

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
```diff
@@ -0,0 +1,284 @@
...
+		Create: resourceArmKustoEventGridDataConnectionCreateUpdate,
+		Read:   resourceArmKustoEventGridDataConnectionRead,
+		Update: resourceArmKustoEventGridDataConnectionCreateUpdate,
+		Delete: resourceArmKustoEventGridDataConnectionDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Timeouts: &schema.ResourceTimeout{
```
ilayrn: The timeouts do not make sense. Read timeout should be 1 min. All others should be 30 mins

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
```diff
@@ -0,0 +1,284 @@
...
+			Update: schema.DefaultTimeout(60 * time.Minute),
+			Delete: schema.DefaultTimeout(60 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateAzureRMKustoDataConnectionName,
```
ilayrn: IMHO the validation should be done against ARM calling dataConnectionValidation and not locally. Validation can be changed in the server-side and it doesn't make sense to duplicate it in client code and to have to change it. This command is valid to all other validate calls in this file

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+	}
+}
+
+func resourceArmKustoEventGridDataConnectionCreateUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Kusto.DataConnectionsClient
+	ctx, cancel := timeouts.ForDelete(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	log.Printf("[INFO] preparing arguments for Azure Kusto EventGrid Data Connection creation.")
```
ilayrn: Data Explorer

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
```diff
@@ -0,0 +1,284 @@
...
+
+	name := d.Get("name").(string)
+	resourceGroup := d.Get("resource_group_name").(string)
+	clusterName := d.Get("cluster_name").(string)
+	databaseName := d.Get("database_name").(string)
+
+	if features.ShouldResourcesBeImported() && d.IsNewResource() {
```
ilayrn: Why is this block required? ARM template is idempotent so there's no reason to check if the resource exists.

mbfrahry: This block is required because Azure tends to upsert any changes regardless of whether it's a new resource or not. This provides challenges for Terraform users as inadvertent changes may get applied on a resource created by something other than Terraform. This check forces a user to first import a resource that has been created outside of Terraform before making any changes on it. It's an easy simple check to prevent users from accidentally making changes to  their infrastructure.

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
mbfrahry: I'm not sure where Data Explorer comes from but this resource is called EventGridDataConnection and it'd be good to keep that matched in the logfile to prevent any confusion should anyone need to peruse the logfile for any reason.

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
mbfrahry: Doing a check here is actually quite helpful as it prevents Terraform from stopping mid execution because a validation check had failed. This check prevents anything from being created until the validation succeeds. Also not all APIs gracefully handle validation or incorrect data so we err on the side of caution. The APIs might not offer any information about why a value is wrong so this check ensures there is an error a user can take action on. Lastly, this check happens before the api is every hit which ensures these issues are discovered faster.

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
mbfrahry: These timeouts chosen as a default across all of our resources. They may seem larger than they should be but it's just an extra precaution in case anything wonky happens on Azure's side that causes these calls to take longer than usual.

azurerm/internal/services/kusto/registration.go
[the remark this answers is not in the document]
mbfrahry: Hey @noam-shapira, unfortunately I don't think we'll be able to take this in without being able to run nightly acceptance tests on it. Is the issue that Terraform doesn't currently manage the resource that's missing or there isn't a way for Terraform to manage it?

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
ilayrn: I meant: "[INFO] preparing arguments for Azure Data Explorer EventGrid Data Connection creation."

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
ilayrn: Yes, I understand the trade-offs. The pro is that the validation is done client-side and it fails fast in case the input is not valid. It also reduces the number of calls against the API. The con is that (some of) the validation logic duplicated and the same code exists in the server-side - if it changes there, someone needs to remember to update the client-side validation. In case the API handles the validation gracefully and offers information about why the value is wrong, do you find it reasonable to validate by calling the validate API? or is it preferable to deal with the changes when they occur?

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
ilayrn: Ok, they are pretty high though. Read operations receive timeout from Azure ARM API after 20 seconds so a client timeout of 1 min seems reasonable.

azurerm/internal/services/kusto/registration.go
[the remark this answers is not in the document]
noam-shapira: @mbfrahry , sorry for the delay.
Tables and Mapping are not part of the control plan, so its not managed by arm resources.
I plan to add direct REST call to the resource, in order to create the Table and Mapping.

What do you think?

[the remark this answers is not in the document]
patrickbsf: If a way was provided in Terraform to execute Kusto management calls (.create-merge table, . create table xxx ingestion xx, ...) that would be FANTASTIC.  It's a significant hole for me right now (in addition to these data connections).  The data connections need created, but they won't work at all unless the tables and mappings are already there as well.
I had resigned myself to having terraform shell out to something I wrote myself to execute kusto management calls but I really don't want to have to resort to that.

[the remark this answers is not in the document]
patrickbsf: To add to that and why Terraform here would be wonderful - because Kusto requires tables/mappings to be created up-front having it be part of Terraform is pretty important.  Kusto however doesn't have the ability for idempotent changes - there's no create if not exist kind of operation.  .create-merge table is ok for table creations, but .create table xxx ingestion xx mapping will fail hard if the mapping already exists.  deleting a mapping that doesn't exist is also a hard error.
For terraform to offer idempotent table/mapping creation operations would be wonderful.

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
katbyte: @ilayrn, at the moment we do all validation locally and do not reach out to the API.

azurerm/internal/services/kusto/resource_arm_kusto_eventgrid_data_connection.go
jrauschenbusch: `Azure Data Explorer` is just a marketing name. `Kusto` is the MS-internal name of the product, which will not change in the future. Therefore, the azure-sdk-for-go also uses the identifier `kusto` instead of `adx`or `dataexplorer`.

azurerm/internal/services/kusto/registration.go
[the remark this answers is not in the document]
jrauschenbusch: Makes totally sense to me. But i would prefer to use the azure-sdk-for-go in favor of own REST calls. There is already a ticket opened by @mbfrahry to request the Kusto data support within the SDK: https://github.com/Azure/azure-sdk-for-go/issues/7916. As soon as this is solved i could start with an implementation. Until then we could use the ticket https://github.com/terraform-providers/terraform-provider-azurerm/issues/6392 to define a design of required resources.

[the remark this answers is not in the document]
jrauschenbusch: What about using https://github.com/Azure/azure-kusto-go for table and mapping creation? But i don't know if it's stable enough to use it.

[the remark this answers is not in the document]
patrickbsf: Actually, as much as I love Go, it unfortunately means either separately building a Go binary as part of a completely independent pipeline and getting it to where terraform is running, or having the Go compiler so that a go run could be executed.  The same configuration Terraform was using to authenticate with Azure would also need to be used by that same Go code.  It would then have to be a shelled-out 'local' execution tied to a resource.  

I will likely use the azure-kusto-go library for other things, but for the initial 'bake' of an environment using it (or the python lib, etc.) would likely be met with a bit of resistance.

katbyte:
Hey @noam-shapira - thank you for this PR. However as its blocked by support for kusto in the SDK (Azure/azure-sdk-for-go#7916) i am going to close this. Please do feel to reopen this once its supported. Thanks again

PR #5610, opened 2020-02-04
ssh key simple validation and error messaging

### LINKED ISSUE #5605: Support for SSH key format errors
### Community Note

### Description

When using a EdDSA key(or any format other than RSA), Terraform only displays an error warning about "Invalid Parameter". 
My suggestion is to change this message to something informing about the supported key formats.

```
Supported SSH key formats

Azure currently supports SSH protocol 2 (SSH-2) RSA public-private key pairs with a minimum length of 2048 bits. Other key formats such as ED25519 and ECDSA are not supported.
```

https://docs.microsoft.com/en-us/azure/virtual-machines/linux/ssh-from-windows

jackofallops:
Fixes #5605 

Added simple validation function to check supplied string carries valid key signature type and is of minimum length or greater.

### DIFF
--- a/azurerm/internal/services/compute/ssh_keys.go
+++ b/azurerm/internal/services/compute/ssh_keys.go
@@ -138,9 +138,14 @@ func ValidateSSHKey(i interface{}, k string) (warnings []string, errors []error)
 
 	switch validSig {
 	case "ssh-rsa":
-		strLen := len(strings.TrimSpace(v))
-		if strLen < 379 {
-			return nil, []error{fmt.Errorf("expected %q to have a key size of at least 2048 bits", k)}
+		keyParts := strings.Fields(v)
+		if len(keyParts) > 1 {
+			strLen := len(keyParts[1])
+			if strLen < 371 {
+				return nil, []error{fmt.Errorf("expected %q to have a key size of at least 2048 bits", k)}
+			}
+		} else {
+			return nil, []error{fmt.Errorf("expected %q to be valid of at least 2 parts, signature and key", k)}
 		}
 	case "":
 		return nil, []error{fmt.Errorf("Bad: Azure currently supports SSH protocol 2 (SSH-2) RSA public-private key pairs with a minimum length of 2048 bits")}
--- a/azurerm/internal/services/compute/ssh_keys.go
+++ b/azurerm/internal/services/compute/ssh_keys.go
@@ -3,6 +3,7 @@ package compute
 import (
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/compute/ssh_keys.go
```diff
@@ -111,3 +112,44 @@ func parseUsernameFromAuthorizedKeysPath(input string) *string {
...
+			validSig = t
+		}
+	}
+
+	switch validSig {
+	case "ssh-rsa":
+		keyParts := strings.Fields(v)
+		if len(keyParts) > 1 {
+			strLen := len(keyParts[1])
+			if strLen < 371 {
```
tombuildsstuff: I don't know enough about SSH keys to see the specifics but I feel like we'd be better to parse the SSH key using the crypto library rather than checking the string length?

website/docs/r/virtual_machine.html.markdown
```diff
@@ -266,7 +266,7 @@ A `plan` block supports the following:
 
 A `ssh_keys` block supports the following:
 
-* `key_data` - (Required) The Public SSH Key which should be written to the `path` defined above.
+* `key_data` - (Required) The Public SSH Key which should be written to the `path` defined above. Azure only supports RSA SSH2 key signatures of at least 2048 bits in length
```
tombuildsstuff:
```suggestion
* `key_data` - (Required) The Public SSH Key which should be written to the `path` defined above.

~> **Note:** Azure only supports RSA SSH2 key signatures of at least 2048 bits in length
```

azurerm/internal/services/compute/ssh_keys.go
```diff
@@ -24,7 +25,7 @@ func SSHKeysSchema(isVirtualMachine bool) *schema.Schema {
 					Type:         schema.TypeString,
 					Required:     true,
 					ForceNew:     isVirtualMachine,
-					ValidateFunc: validation.StringIsNotEmpty,
+					ValidateFunc: ValidateSSHKey,
```
tombuildsstuff: this is used in the new VM resources not the old one (AFAIK) - could we update the documentation for those in addition/instead?

azurerm/internal/services/compute/ssh_keys.go
katbyte: agreed

azurerm/internal/services/compute/ssh_keys.go
jackofallops: Docs for the new resources already make mention of the restriction for key signature and length I believe, but I'll double check.

azurerm/internal/services/compute/ssh_keys.go
jackofallops: It's not core go SDK (yet) means using importing `golang.org/x/crypto/ssh`, will rework for that as discussed offline.
NB: I don't believe I've ever seen a case where a 2048bit key in SSH2 notation is anything other than 371 characters.

PR #5614, opened 2020-02-04
r/storage: authenticating to the Blob/Queues API's using AzureAD Auth

### LINKED ISSUE #3651: Provider authentication to blob using Azure AD
### Community Note

### Description
Azure Storage now supports [authentication using Azure AD](https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad), in addition to authentication with a SAS token or access keys. The provider also supports authentication with Azure AD service principal, but look like it's using the credentials to get access keys, and then use them to access the storage. I noticed that when I tried to use SP credentials and encountered the following error:
```
Error retrieving keys for Storage Account "<>": storage.AccountsClient#ListKeys: Failure responding to request: StatusCode=403
```
It happened because I granted the SP contributor role to the blob, not to the storage (least privilege!). It was solved once I granted it a contributor role to the entire storage account.

Please consider support authentication with AD credentials.

tombuildsstuff:
This adds support for authenticating to the Blob & Queue Storage API's using AzureAD authentication via a Feature Toggle on the Provider block.

Unfortunately this isn't supported for the Files & Tables API's but is documented to that affect

Fixes #3651

### DIFF
--- a/azurerm/internal/provider/provider.go
+++ b/azurerm/internal/provider/provider.go
@@ -175,6 +175,13 @@ func AzureProvider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("ARM_SKIP_PROVIDER_REGISTRATION", false),
 				Description: "Should the AzureRM Provider skip registering all of the Resource Providers that it supports, if they're not already registered?",
 			},
+
+			"storage_use_azuread": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("ARM_STORAGE_USE_AZUREAD", false),
+				Description: "Should the AzureRM Provider use AzureAD to access the Storage Data Plane API's?",
+			},
 		},
 
 		DataSourcesMap: dataSources,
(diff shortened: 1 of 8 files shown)


azurerm/internal/provider/provider.go
```diff
@@ -175,6 +175,13 @@ func AzureProvider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("ARM_SKIP_PROVIDER_REGISTRATION", false),
 				Description: "Should the AzureRM Provider skip registering all of the Resource Providers that it supports, if they're not already registered?",
 			},
+
+			"storage_use_azuread": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("ARM_STORAGE_USE_AZUREAD", false),
+				Description: "Should the AzureRM Provider use AzureAD to access the Storage Data Plane API's?",
+			},
```
tombuildsstuff: Note: this is intentionally a top-level feature and not in the `features` block since it's authentication related

azurerm/internal/services/storage/client/client.go
```diff
@@ -24,15 +24,17 @@ type Client struct {
 	FileSystemsClient        *filesystems.Client
 	ManagementPoliciesClient storage.ManagementPoliciesClient
 	BlobServicesClient       storage.BlobServicesClient
-	environment              az.Environment
+
+	environment az.Environment
+	storageAuth *autorest.Authorizer
```
katbyte: Given this is not for all storage as some don't support it atm, and just storage ad, auth might it make more sense to be
```suggestion
	storageAdAuth *autorest.Authorizer
```

tombuildsstuff:
Tests pass:

Screenshot 2020-02-05 at 17 33 12

azurerm/internal/services/storage/client/client.go
tombuildsstuff: 👍

PR #5625, opened 2020-02-06
devspace - add ID parsing function

tracypholmes:


azurerm/internal/services/devspace/parse/devspace_controller_test.go
```diff
@@ -0,0 +1,73 @@
...
+			Name:  "DevSpace Controller ID",
+			Input: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resGroup1/providers/Microsoft.DevSpaces/controllers/Controller1",
+			Expected: &DevSpaceControllerId{
+				Name:          "Controller1",
+				ResourceGroup: "resGroup1",
+			},
+		},
+		{
+			Name:     "Wrong Casing",
+			Input:    "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resGroup1/providers/Microsoft.DevSpaces/Controllers/",
```
jackofallops: Should have a value on there also or will fail on missing value also, possibly hiding a condition?

tracypholmes: Oooh..good catch. TY - I also think you've answered something I was wondering about!

PR #5626, opened 2020-02-06
healthcare - add ID parsing function

tracypholmes:


azurerm/internal/services/healthcare/parse/healthcare_service_test.go
```diff
@@ -0,0 +1,73 @@
...
+			Name:  "Healthcare Service ID",
+			Input: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resGroup1/providers/Microsoft.HealthcareApis/services/Service1",
+			Expected: &HealthcareServiceId{
+				Name:          "Service1",
+				ResourceGroup: "resGroup1",
+			},
+		},
+		{
+			Name:     "Wrong Casing",
+			Input:    "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resGroup1/providers/Microsoft.HealthcareApis/Services/",
```
jackofallops: Should have a value on there also or will fail on missing value also, possibly hiding a condition?

tracypholmes: Indeed. Thank you for catching that!

PR #5636, opened 2020-02-07
`r/azurerm_disk_encryption_set`: add ID parsing function

ArcturusZhang:


azurerm/internal/services/compute/resource_arm_managed_disk.go
```diff
@@ -144,7 +145,7 @@ func resourceArmManagedDisk() *schema.Resource {
 				// TODO: make this case-sensitive once this bug in the Azure API has been fixed:
 				//       https://github.com/Azure/azure-rest-api-specs/issues/8132
 				DiffSuppressFunc: suppress.CaseDifference,
-				ValidateFunc:     azure.ValidateResourceID,
+				ValidateFunc:     validate.DiskEncryptionSetID,
```
tombuildsstuff: unfortunately we can't use this here until the API returns this in the correct case

ArcturusZhang: Done

ArcturusZhang:
Hi @tombuildsstuff I have revert the validation in `azurerm_managed_disk`.

azurerm/internal/services/compute/resource_arm_managed_disk.go
tombuildsstuff: @ArcturusZhang I think you may have forgotten to push that commit? :)

ArcturusZhang: 😂how silly of me

ArcturusZhang:
Hi @tombuildsstuff well, this time I have pushed the changes. 😂

Issue #1903, opened 2018-09-10
Add support for static website to azurerm_storage_account


hughtopping:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Static website hosting for Azure Storage is now in public preview. [See here for Azure announcement.](https://azure.microsoft.com/en-us/blog/azure-storage-static-web-hosting-public-preview/)

Currently this can be added with the Azure Portal UI, but is missing from the `azurerm_storage_account` Terraform resource. 

### New or Affected Resource(s)

* azurerm_storage_account

### Potential Terraform Configuration

```hcl
resource "azurerm_storage_account" "static_website" {
  account_replication_type  = "RAGRS"
  account_tier              = "Standard"
  account_kind              = "StorageV2"
  location                  = "westeurope"
  name                      = "static_website_storage_account"
  resource_group_name       = "static_website_rg"
  enable_https_traffic_only = true

  static_website {
    enabled             = true
    index_document_name = "index.html"
    error_document_path = "error.html"
  }
}
```

### References

* https://azure.microsoft.com/en-us/blog/azure-storage-static-web-hosting-public-preview/

0x7f:
For those who can not wait until this feature lands in stable terraform and still want to manage their storage with static website hosting enabled via terraform, I have a possible workaround.

First, I tried creating [azurerm_template_deployment](https://www.terraform.io/docs/providers/azurerm/r/template_deployment.html) and enable the static website feature via an ARM template. Unfortunately this is not possible, because static website feature is also not available via arm templates yet, see [here](https://docs.microsoft.com/en-us/azure/templates/Microsoft.Storage/2018-07-01/storageAccounts#StorageAccountPropertiesCreateParameters). It is only possible via Blob Service REST API since API Version [2018-03-28](https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-service-properties#request) which you can use via azure cli.

So the workaround is to run enable static website via azure cli using terraform's [`local-exec`](https://www.terraform.io/docs/provisioners/local-exec.html) provisioner. Based on your example it could look like:

```tf
[...]

PR #5649, opened 2020-02-08
`azurerm_storage_account` - add support for static website

aqche:
Fixes: #1903 

Adds support for static web hosting to `azurerm_storage_account`.

```
=== RUN   TestAccAzureRMStorageAccount_blobAccountProperties
=== PAUSE TestAccAzureRMStorageAccount_blobAccountProperties
=== CONT  TestAccAzureRMStorageAccount_blobAccountProperties
--- PASS: TestAccAzureRMStorageAccount_blobAccountProperties (149.02s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/storage/tests       149.069s
```


azurerm/internal/services/storage/resource_arm_storage_account.go
```diff
@@ -452,6 +453,40 @@ func resourceArmStorageAccount() *schema.Resource {
...
+									"enabled": {
+										Type:     schema.TypeBool,
+										Required: true,
+									},
+									"index_document": {
+										Type:         schema.TypeString,
+										Optional:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"error_document_404_path": {
```
StefanSchoof: Why here a path postfix? In the azure cli it is 404-document and index-document. See https://docs.microsoft.com/de-de/azure/storage/blobs/storage-blob-static-website-how-to?tabs=azure-cli

website/docs/r/storage_account.html.markdown
```diff
@@ -229,6 +233,22 @@ A `queue_properties` block supports the following:
...
+
+* `static_website` - (Optional) A `static_website` block as defined below.
+
+---
+
+A `static_website` block supports the following:
+
+* `enabled` - (Required) Indicates whether static website support is enabled for the given account.
+
+* `index_document` - (Optional) The webpage that Azure Storage serves for requests to the root of a website or any subfolder. For example, index.html. The value is case-sensitive.
```
StefanSchoof: Should the documentation mention the default value for index and 404?

website/docs/r/storage_account.html.markdown
```diff
@@ -229,6 +233,22 @@ A `queue_properties` block supports the following:
...
 
+---
+
+A `blob_account_properties` block supports the following:
+
+* `static_website` - (Optional) A `static_website` block as defined below.
+
+---
+
+A `static_website` block supports the following:
```
StefanSchoof: Should the documentation mention that a $web container is created and the files should be placed inside?

azurerm/internal/services/storage/resource_arm_storage_account.go
aqche: I named these arguments based on the names from `giovanni`. 

https://github.com/tombuildsstuff/giovanni/blob/89b9b60252d146d1412bb2cbefa1183c30f4c9c0/storage/2018-11-09/blob/accounts/models.go#L20-L27

It's also worth noting in Azure PowerShell, the input is called `ErrorDocument404Path` as well.

https://docs.microsoft.com/en-us/powershell/module/az.storage/enable-azstoragestaticwebsite?view=azps-3.4.0

website/docs/r/storage_account.html.markdown
aqche: I believe they both default to no value. From my testing, if no index document is specified, the primary web endpoint just returns a 404. If no 404 document is specified, Azure will return a default generic 404 page (added note for this).

website/docs/r/storage_account.html.markdown
aqche: Good call, added!

aqche:
@tombuildsstuff thanks for the review, updated the code with your suggestions!

```
--- PASS: TestAccAzureRMStorageAccount_staticWebsiteProperties (148.88s)
```

aqche:
Fixed up the tflint errors, looks like website-lint is failing from something unrelated to this PR.

azurerm/internal/services/storage/resource_arm_storage_account.go
```diff
@@ -452,6 +453,27 @@ func resourceArmStorageAccount() *schema.Resource {
 				},
 			},
 
+			"static_website": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
tombuildsstuff: since we'll only flatten a `static_website` block when there's one enabled we can remove Computed here (since there'll be 0 blocks when omited)

```suggestion
```

tombuildsstuff:
Ignoring some known test failures (which'll be fixed via #5710) - the tests look good 👍

PR #5657, opened 2020-02-10
tests: change configs to .12 (M-N)

katbyte:


tombuildsstuff:
As discussed offline I'm going to close this for the moment since otherwise this is going to conflict with every open PR - once we've got the PR count down for 2.0 we should be good to re-run this

PR #5670, opened 2020-02-11
Corrected Prefix Length

johnwildes:
The maximum IP prefix length that can be reserved (according to docs.microsoft) is `/28` or 16 usable addresses.

https://docs.microsoft.com/en-us/azure/virtual-network/public-ip-address-prefix#constraints

### DIFF
--- a/website/docs/r/public_ip_prefix.html.markdown
+++ b/website/docs/r/public_ip_prefix.html.markdown
@@ -45,7 +45,7 @@ The following arguments are supported:
 
 -> **Note** Public IP Prefix can only be created with Standard SKUs at this time.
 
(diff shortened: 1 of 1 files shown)


johnwildes:
@mbfrahry I updated the module, PR is here https://github.com/terraform-providers/terraform-provider-azurerm/pull/5693

mbfrahry:
Thanks for that @johnwildes. Can we keep the changes to a single pr though? It helps us have a single point of reference when looking at how/why a change was made. Also rebasing with master should allow the travis tests to pass.

johnwildes:
@mbfrahry I think I understand what you're saying.  I opened this PR for the documentation, and instead of updating the actual code in this PR as well, I opened a new PR.  I'll remember this for the future and just submit one.  I don't see a way to fix this from here.

tombuildsstuff:
The travis failures are unrelated (and have since been fixed in master) - this LGTM 👍

PR #5679, opened 2020-02-11
Add form recognizer as a Cognitive account kind

apollo2030:
FormRecognizer is a new service which is in preview on azure: https://azure.microsoft.com/en-in/services/cognitive-services/form-recognizer/

### DIFF
--- a/azurerm/internal/services/cognitive/resource_arm_cognitive_account.go
+++ b/azurerm/internal/services/cognitive/resource_arm_cognitive_account.go
@@ -74,6 +74,7 @@ func resourceArmCognitiveAccount() *schema.Resource {
 					"CustomVision.Training",
 					"Emotion",
 					"Face",
+					"FormRecognizer",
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/cognitive/resource_arm_cognitive_account.go
```diff
@@ -74,6 +74,7 @@ func resourceArmCognitiveAccount() *schema.Resource {
 					"CustomVision.Training",
 					"Emotion",
 					"Face",
+					"FormRecognizer",
```
tombuildsstuff: could we also [add this to the documentation](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/cognitive_account.html.markdown#L45)?

apollo2030: Done

PR #5681, opened 2020-02-11
Add FormRecognizer as a Cognitive Account Kind

stack72:
This service is currently in preview - https://azure.microsoft.com/en-in/services/cognitive-services/form-recognizer/

The API takes a string and doesn't perform any validation so we should
allow Terraform to do the same

@tombuildsstuff not sure if Terraform supports preview features or not but thought I'd take a chance


tombuildsstuff:
Duplicate of #5679 - closing in favour of that one :)

PR #5708, opened 2020-02-12
docs: removing the Beta signifier from docs

tombuildsstuff:
These resources will no longer be in Beta - thus we can remove this label from them


tombuildsstuff:
The travis error will be resolved once #5707 is merged

Issue #5541, opened 2020-01-28
AKS node pool k8s version not being updated


jstevans:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
* `terraform` version: 0.12.8
* `azurerm provider` version: 1.41

### Affected Resource(s)

* `azurerm_kubernetes_cluster`

### Terraform Configuration Files

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key: https://keybase.io/hashicorp

resource "azurerm_kubernetes_cluster" "k8s" {
  name                = var.cluster_info.name
  location            = azurerm_resource_group.k8s.location
  dns_prefix          = var.cluster_info.dns_prefix
  resource_group_name = azurerm_resource_group.k8s.name
  kubernetes_version  = var.kubernetes_version

  role_based_access_control {
    enabled = ! local.aks_aad_skip_rbac
    dynamic "azure_active_directory" {
      for_each = "${! local.aks_aad_skip_rbac ? list(local.aks_rbac_setting) : []}"
      content {
        client_app_id     = local.aks_rbac_setting.client_app_id
        server_app_id     = local.aks_rbac_setting.server_app_id
        server_app_secret = local.aks_rbac_setting.server_app_secret
        tenant_id         = local.aks_rbac_setting.tenant_id
      }
    }
  }

  default_node_pool {
    name               = var.agent_pool.name
    node_count         = var.agent_pool.count
    vm_size            = "Standard_DS2_v2"
    type               = "VirtualMachineScaleSets"
    os_disk_size_gb    = 30
    max_pods           = 30
    availability_zones = local.sanitized_availability_zones

    enable_auto_scaling= true
    min_count = 3
    max_count = 12
  }

  service_principal {
    client_id     = var.aks_login.client_id
    client_secret = var.aks_login.client_secret
  }

  addon_profile {
    oms_agent {
      enabled                    = true
[...]

PR #5741, opened 2020-02-13
Couple default_node_pool OrchestratorVersion to cluster k8s_version

jstevans:
Per #5541, currently AKS node pool versions can never be updated. This
occurred due to a change in behavior in new ARM behavior now that AKS
clusters can have multiple agent pools.

There's a more involved fix in the discussion on that issue which involves
exposing OrchestratorVersion as a settable attribute on agent_pool_profiles,
but this change focuses on recovering the old behavior.

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_nodepool.go
+++ b/azurerm/internal/services/containers/kubernetes_nodepool.go
@@ -164,10 +164,10 @@ func ExpandDefaultNodePool(d *schema.ResourceData) (*[]containerservice.ManagedC
 		// at this time the default node pool has to be Linux or the AKS cluster fails to provision with:
 		// Pods not in Running status: coredns-7fc597cc45-v5z7x,coredns-autoscaler-7ccc76bfbd-djl7j,metrics-server-cbd95f966-5rl97,tunnelfront-7d9884977b-wpbvn
 		// Windows agents can be configured via the separate node pool resource
-		OsType: containerservice.Linux,
+		OsType:              containerservice.Linux,
+		OrchestratorVersion: utils.String(d.Get("kubernetes_version").(string)),
 
 		//// TODO: support these in time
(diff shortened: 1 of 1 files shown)


derek-burdick:
Just some feedback from a user of AKS and multiple node pools. I never want to upgrade the control plane and node pools at the same time. I also do not want to run terraform apply twice in a row without a change and have node pools upgraded.

In my opinion, waiting until OrchestratorVersion can be set in terraform is more preferable than node pools taking the setting from the cluster.

jackofallops:
Hi @jstevans 
Thanks for the PR.
The case for running disparate compatible versions between master / workers is something I've experienced myself, so I'm inclined to agreed with @derek-burdick that we shouldn't take the version from the control plane exclusively but possibly extend the node pool schema to support an optional field for it?  What are your thoughts?  
For reference, the k8s skew policy is described [here](https://kubernetes.io/docs/setup/release/version-skew-policy/)

jstevans:
Hey folks,

I'm a single-cluster user whose ability to manage his cluster has regressed due to this new feature, so personally I'd prefer to return ASAP to the old behavior that we were all previously living with 😄 I get the concern from users who have onboarded to multiple agent-pool configs, though. 

However, I'm afraid I don't have the bandwidth to make the more-involved change that @tombuildsstuff suggested in #5541. I also don't have the expertise to weigh the tradeoff that @derek-burdick requests there, either. Please free to make another change in lieu of this one.

It'd be awesome to be unbroken in the near future, though, so someone please pick this up who wants the more involved fix. To expound -- my clusters are under a security policy that prevents me from upgrading except via TF, so my node pools are completely stuck right now.

katbyte:
Hi @jstevans,

it looks like @titilambert has opened #6047 which supersedes this by allowing `OrchestratorVersion` to be set 🙂 as such i am going to close this in favour of it.

PR #5742, opened 2020-02-13
Add `azurerm_resources.id` and `azurerm_resources.resources[*].resource_group`

jstevans:
This PR adds to the `azurerm_resources` data source, such that:

* it is possible to provide the Azure resource ID of a resource (e.g. `subscriptions/SUBSCRIPTION/resourceGroup/RESOURCEGROUP/PROVIDER/RESOURCENAME`) in lieu of resourceGroup/provider/name. This is helpful in cases where a Terraform resource implicitly encompasses several Azure resources, and only provides the subscription ID of those implicit resources (e.g. the `public_ip` that is automatically created when creating a k8s cluster).
* each returned resource now also provides its resource group. This is helpful when attempting to perform logical organization of explicitly-managed Azure resources alongside non- or implicitly-managed Azure resources (e.g. when trying to create a new IP address that can be associated with a `kubernetes_cluster`, which requires it to be in the `MC_*` resource group that the original IP address was created in).

### DIFF
--- a/azurerm/internal/services/resource/data_source_resources.go
+++ b/azurerm/internal/services/resource/data_source_resources.go
@@ -185,8 +185,13 @@ func transformResource(res resources.GenericResource) *map[string]interface{} {
 	}
 
 	resID := ""
+	resGroup := ""
 	if res.ID != nil {
 		resID = *res.ID
+		parsedID, err := azure.ParseAzureResourceID(*res.ID)
+		if err == nil {
+			resGroup = parsedID.ResourceGroup
+		}
 	}
 
 	resType := ""
@@ -211,6 +216,7 @@ func transformResource(res resources.GenericResource) *map[string]interface{} {
 
 	return &map[string]interface{}{
 		"name":           resName,
+		"resource_group": resGroup,
 		"id":             resID,
 		"type":           resType,
 		"location":       resLocation,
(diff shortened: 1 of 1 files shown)


jstevans:
@tombuildsstuff Not sure who to request a review from -- could you advise?

katbyte:
@jstevans,

Thank you for this PR, however upon some internal discussion we are not sure if we should make this addition. While resources should comply with the ARM specifications it unfortunately is not guaranteed, particularly when it comes to resources which haven’t been updated. Instead we’d suggest using data sources at a more granular level. Given this, whilst i would like to thank you for this PR i am going to close it as we don't feel like the direction we want to take this data source.

jstevans:
@katbyte as a user of `terraform-provider-azurerm` interested in contributing, I'd love to be part of that discussion! 🙂

Our use case is that we want to manually provision IP addresses for AKS clusters so that we can add them to a traffic manager. Right now, `resource_arm_kubernetes_cluster.go` only emits the resource ID of the resource, so we're doing the gymnastics ourselves of parsing that into resource parts. 

I'm not clear on what's objectionable about this change -- to me, this seems like a natural analog to [`az resource --ids`](https://docs.microsoft.com/en-us/cli/azure/resource?view=azure-cli-latest#az-resource-show), plus outputting `resourceGroup` which is definitely already present in the response object. Could you help me understand?

PR #5745, opened 2020-02-14
Update subnets options and add details for endpoint 

Narsi-1310:
The following changes have been carried out:
- renamed parameter to enforce_private_link_service_network_policies from disable_private_link_service_network_policy_enforcement  for  service and endpoint subnets
- Include private service connection under azurerm_private_endpoint - to provide better view of example.

This is inline with terraform v0.12.20


tombuildsstuff:
@Narsi-1310 hope you don't mind but I've pushed a commit to fix the linting so that we can get this merged 👍

Narsi-1310:
Thanks @tombuildsstuff

PR #5746, opened 2020-02-14
New resource `azurerm_policy_remediation`

ArcturusZhang:
This is a revised version of PR [5277](https://github.com/terraform-providers/terraform-provider-azurerm/pull/5277).

A `azurerm_policy_remediation` is used along with a `azurerm_policy_assignment` to remediate a policy.
A policy can be assigned in 4 different scopes: Subscription, Resource Group, Resource and Management Group. This remediation resource supports all of them using a `scope` field similar to the `azurerm_policy_assignment`.

Since the id of a management group (like `/providers/Microsoft.Management/managementGroups/00000000-0000-0000-0000-000000000000`) is not a valid azure resource id, passing it to `azure.ParseResourceId` will throw errors, I wrote some additional ID parsing functions to deal with this kind of situations.

And additionally, since all of the ids in a remediation will be converted to lower cases, I am using case diff suppression everywhere.


azurerm/internal/services/policyinsights/tests/resource_arm_policy_remediation_test.go
```diff
@@ -0,0 +1,591 @@
...
+
+		return nil
+	}
+
+	return nil
+}
+
+func testAccAzureRMPolicyRemediation_atSubscription(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+data "azurerm_subscription" "current" {}
```
katbyte: We'll need to add a provider block to all these tests

azurerm/internal/services/policyinsights/resource_arm_policy_remediation.go
```diff
@@ -0,0 +1,207 @@
...
+
+			"location_filters": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"policy_definition_reference_id": {
```
katbyte: I don't see any test that make use of this property?

azurerm/internal/services/policyinsights/parse/remediation.go
```diff
@@ -0,0 +1,130 @@
...
+	// {scope}/providers/Microsoft.PolicyInsights/remediations/{name}
+	regex := regexp.MustCompile(`/providers/[Mm]icrosoft\.[Pp]olicy[Ii]nsights/remediations/`)
+	if !regex.MatchString(input) {
+		return nil, fmt.Errorf("Unable to parse Policy Remediation ID %q", input)
+	}
+
+	segments := regex.Split(input, -1)
+
+	if len(segments) != 2 {
+		return nil, fmt.Errorf("Unable to parse Policy Remediation ID %q: Expected 2 segments after splition", input)
```
katbyte:
```suggestion
		return nil, fmt.Errorf("Unable to parse Policy Remediation ID %q: Expected 2 segments after split", input)
```

azurerm/internal/services/policyinsights/parse/remediation.go
```diff
@@ -0,0 +1,130 @@
...
+
+	id := RemediationId{
+		Name:               name,
+		RemediationScopeId: *scopeId,
+	}
+
+	return &id, nil
+}
+
+type RemediationScopeId struct {
```
katbyte: I'm not sure we should share the properties struct like this. not all properties will be populated for each of the remediation types. While they share a similar base we should have 4 derived types for each type of ID.eg PolicyRemediationAtSubscriptionID` and such

azurerm/internal/services/policyinsights/parse/remediation.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return &scopeId, nil
+}
+
+func isManagementGroupId(input string) bool {
+	_, err := ManagementGroupID(input)
+	return err == nil
+}
+
+// TODO -- move this to management group RP directory
```
katbyte: Could we do this as part of this PR & update the resources to use it?

azurerm/internal/services/policyinsights/remediations.go
```diff
@@ -0,0 +1,57 @@
...
+	case parse.AtResourceGroup:
+		return client.CreateOrUpdateAtResourceGroup(ctx, scope.SubscriptionId, scope.ResourceGroup, name, parameters)
+	case parse.AtResource:
+		return client.CreateOrUpdateAtResource(ctx, scope.ScopeId, name, parameters)
+	case parse.AtManagementGroup:
+		return client.CreateOrUpdateAtManagementGroup(ctx, scope.ManagementGroupId, name, parameters)
+	default:
+		return policyinsights.Remediation{}, fmt.Errorf("Error creating Policy Remediation: Invalid scope type %q", scope.Type)
+	}
+}
```
katbyte: Given this will only be called in one resource do we need to have a function for it?

azurerm/internal/clients/client.go
```diff
@@ -46,6 +46,7 @@ import (
 	network "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/client"
 	notificationhub "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/notificationhub/client"
 	policy "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/policy/client"
+	policyinsights "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/policyinsights/client"
```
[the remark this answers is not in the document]
ArcturusZhang: Sure, I will move them to the policy directory

azurerm/internal/services/policyinsights/client/client.go
```diff
@@ -0,0 +1,19 @@
+package client
```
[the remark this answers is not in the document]
ArcturusZhang: Done

azurerm/internal/services/policyinsights/data_source_policy_remediation.go
```diff
@@ -0,0 +1,68 @@
+package policyinsights
```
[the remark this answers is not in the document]
ArcturusZhang: You are right, this resource is like policy assignment. The data source does not make much sense. I have removed it.

azurerm/internal/services/policyinsights/data_source_policy_remediation.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	resp, err := RemediationGetAtScope(ctx, client, name, *scopeId)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return fmt.Errorf("Error: Policy Remediation %q (Scope %q) was not found", name, scope)
+		}
+		return fmt.Errorf("Error reading Policy Remediation %q (Scope %q): %+v", name, scope, err)
+	}
+	if resp.ID == nil || *resp.ID == "" {
```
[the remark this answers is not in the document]
ArcturusZhang: The data source has been removed

azurerm/internal/services/policyinsights/data_source_policy_remediation.go
```diff
@@ -0,0 +1,68 @@
...
+		return fmt.Errorf("Cannot read Policy Remediation %q (Scope %q) ID", name, scope)
+	}
+
```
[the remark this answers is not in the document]
ArcturusZhang: The data source has been removed

azurerm/internal/services/policyinsights/parse/remediation.go
```diff
@@ -0,0 +1,130 @@
...
+	AtResourceGroup
+	AtResource
+)
+
+type RemediationId struct {
+	Name string
+	RemediationScopeId
+}
+
+// TODO: This paring function is currently suppressing every case difference due to github issue: https://github.com/Azure/azure-rest-api-specs/issues/8353.
```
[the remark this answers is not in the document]
ArcturusZhang: Done

azurerm/internal/services/policyinsights/tests/resource_arm_policy_remediation_test.go
ArcturusZhang: Done

azurerm/internal/services/policyinsights/parse/remediation.go
ArcturusZhang: Well, I have already opened a PR for this (and it has been reviewed). Could I just close that PR, and merge those changes into this PR?

azurerm/internal/services/policyinsights/remediations.go
ArcturusZhang: Sure, i will move these functions to the corresponding CRUD functions

azurerm/internal/services/policyinsights/resource_arm_policy_remediation.go
ArcturusZhang: I will add some tests for this property

azurerm/internal/services/policyinsights/remediations.go
ArcturusZhang: Since the GetAtScope function and DeleteAtScope function are called multiple times, I will keep them, but move them to the resource file.

azurerm/internal/services/policyinsights/parse/remediation.go
ArcturusZhang: I have merged [PR 5878](https://github.com/terraform-providers/terraform-provider-azurerm/pull/5878) into this PR.

azurerm/internal/services/policyinsights/parse/remediation.go
ArcturusZhang: I have done some tweaks on the structure of this part. Introduced an interface for `RemediationScopeId` and 4 structs implemented this interface, please have a look again

ArcturusZhang:
Hi @katbyte I have resolved your comment. These are what I have done:
1. Removed directory and package `policyinsights` and move all the files to `policy` package.
1. Removed the data source of `azurerm_policy_remediation`
1. Define an interface `PolicyScopeId` to represent different scope IDs, and declare 4 different structs for 4 different scopes implementing the interface `PolicyScopeId`. These types could be used for other resources in the `azurerm_policy_assignment`, `azurerm_policy_definition` and `azurerm_policy_set_definition`.
1. Refactor the file structure in `parse` and `validate`. Since the scope can be shared within all the resources in `policy` service, I moved the interface`PolicyScopeId` and 4 structs for different scope IDs to the `parse/scope.go` file.

Despite I introduce the parsing functions for `azurerm_policy_definition` and `azurerm_policy_assignment`, I did not change the resource and data source of policy definition and policy assignment for simplicity of this PR. I could create another PR to refine the existing resource in policy package.

azurerm/internal/services/policyinsights/parse/remediation.go
katbyte: awesome, thanks 👍

azurerm/internal/services/policyinsights/resource_arm_policy_remediation.go
ArcturusZhang: Test added

ArcturusZhang:
HI @katbyte some information on the acceptance tests of this resource:
`TestAccAzureRMPolicyRemediation_atSubscription`: PASS
`TestAccAzureRMPolicyRemediation_atResourceGroup`: PASS
`TestAccAzureRMPolicyRemediation_atManagementGroup`: FAIL, the service principal i am using does not have the permission to create management group
`TestAccAzureRMPolicyRemediation_atResource`: PASS
`TestAccAzureRMPolicyRemediation_atSubscriptionWithDefinitionSet`: FAIL, something wrong with the `azurerm_policy_set_definition` resource, it will always return diff in the `metadata` field with something like `createdOn`.

katbyte:
I think this will conflict with the prs: #5981 #6107 ? 🤔

ArcturusZhang:
> I think this will conflict with the prs: #5981 #6107 ? 🤔

Well, in general this PR should not have conflicts with them, because I did not change anything else in the policy service (including the policy_definition resource) other than the addition of policy remediation. Therefore I suppose my PR should not have conflicts with them.

But I think the changes for management group IDs in this PR would benefit those two PR. They are trying to handle the management group in their own resources, which I think is not a very good idea to introduce too much copy and paste work.

ArcturusZhang:
I just noticed that one of them has been approved, then maybe a refinement of how to deal the management group in the policy service should be introduced after that and this PR get merged

azurerm/internal/services/policy/resource_arm_policy_remediation.go
```diff
@@ -0,0 +1,248 @@
...
+func flattenArmRemediationLocationFilters(input *policyinsights.RemediationFilters) []interface{} {
+	if input == nil {
+		return make([]interface{}, 0)
+	}
+
+	return utils.FlattenStringSlice(input.Locations)
+}
+
+// RemediationGetAtScope is a wrapper of the 4 Get functions on RemediationsClient, combining them into one to simplify code.
+func RemediationGetAtScope(ctx context.Context, client *policyinsights.RemediationsClient, name string, scopeId parse.PolicyScopeId) (policyinsights.Remediation, error) {
```
katbyte: Should this be private to this package?
```suggestion
func remediationGetAtScope(ctx context.Context, client *policyinsights.RemediationsClient, name string, scopeId parse.PolicyScopeId) (policyinsights.Remediation, error) {
```

azurerm/internal/services/policy/tests/resource_arm_policy_remediation_test.go
```diff
@@ -0,0 +1,733 @@
...
+			return fmt.Errorf("Policy Insights Remediation not found: %s", resourceName)
+		}
+
+		id, err := parse.PolicyRemediationID(rs.Primary.ID)
+		if err != nil {
+			return err
+		}
+
+		client := acceptance.AzureProvider.Meta().(*clients.Client).Policy.RemediationsClient
+		ctx := acceptance.AzureProvider.Meta().(*clients.Client).StopContext
```
katbyte: could we move these to the top of the anon function?

website/docs/r/policy_remediation.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the Policy Remediation. Changing this forces a new resource to be created.
+
+* `scope` - (Required) The Scope at which the Policy Remediation should be applied, which must be a Resource ID (such as Subscription e.g. `/subscriptions/00000000-0000-0000-0000-000000000000` or a Resource Group e.g.`/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myResourceGroup` or a specified Resource e.g. `/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myResourceGroup/providers/Microsoft.Compute/virtualMachines/myVM`) or a Management Group (e.g. `/providers/Microsoft.Management/managementGroups/00000000-0000-0000-0000-000000000000`). Changing this forces a new resource to be created.
```
katbyte: this might read better as a list?

website/docs/r/policy_remediation.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+* `policy_definition_reference_id` - (Optional) The policy definition reference ID of the individual definition that should be remediated. Required when the policy assignment being remediated assigns a policy set definition.
+
+* `location_filters` - (Optional) A list of the resource locations that will be remediated.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The ID of the Policy Remediation.
+
```
katbyte: could we add the timeout docs here?

robinsk:
Nice to see getting closer to merge. Really looking forward to clean up our subscriptions programmatically.

@ArcturusZhang Thank you for working on this, much appreciated!

azurerm/internal/services/managementgroup/data_source_management_group.go
```diff
@@ -21,8 +22,9 @@ func dataSourceArmManagementGroup() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"group_id": {
-				Type:     schema.TypeString,
-				Required: true,
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.ManagementGroupName,
```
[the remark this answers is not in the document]
ArcturusZhang: Here group_id means this one in its ID:
```
/providers/Microsoft.ManagementGroup/managementGroups/{group_id}
```
the thing that in other resources' IDs at the same position would be called as a `name`, for instance
```
/subscriptions/{subsID}/resourceGroups/group1/providers/Microsoft.Compute/virtualMachines/{name}
```
And another reason is that if we call this `group_id` in the validation function, the validation function would looks like `ManagementGroupGroupID` or `ManagementGroupID`. This would be confusing with the validation function of the real management group ID. Therefore I took this name.

azurerm/internal/services/managementgroup/resource_arm_management_group.go
```diff
@@ -38,10 +40,11 @@ func resourceArmManagementGroup() *schema.Resource {
...
 			"group_id": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
-				ForceNew: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.ManagementGroupName,
```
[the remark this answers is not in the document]
ArcturusZhang: Same reason as the above comment

azurerm/internal/services/policy/parse/definition.go
```diff
@@ -0,0 +1,43 @@
...
+
+type PolicyDefinitionId struct {
+	Name string
+	PolicyScopeId
+}
+
+// TODO: This paring function is currently suppressing every case difference due to github issue: https://github.com/Azure/azure-rest-api-specs/issues/8353
+func PolicyDefinitionID(input string) (*PolicyDefinitionId, error) {
+	// in general, the id of a definition should be:
+	// {scope}/providers/Microsoft.PolicyInsights/remediations/{name}
```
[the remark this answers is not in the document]
ArcturusZhang: Yeah... This regex only matches the middle part, and then I did a split on the ID to separate the `{scope}` part and the `name`. Because the `{scope}` can either be a subscription ID, resource group ID, resource ID, or management group ID, those 4 scopes do not share the same pattern.

azurerm/internal/services/policy/resource_arm_policy_remediation.go
ArcturusZhang: Emmm I am afraid that we cannot do this... because this is also used in the acc tests, the checkDestroy or checkExist functions are using this to query whether the remediation is successfully provisioned.

azurerm/internal/services/policy/tests/resource_arm_policy_remediation_test.go
ArcturusZhang: Sure!

website/docs/r/policy_remediation.html.markdown
ArcturusZhang: Sure

website/docs/r/policy_remediation.html.markdown
ArcturusZhang: Oops, I forgot this. I will add them

ArcturusZhang:
Hi @katbyte I have resolved the comments.

And I just spotted a typo in the doc of `azurerm_policy_set_definition` when I am copying the `timeout` section, and I changed those in this PR, hope you would not mind.

azurerm/internal/services/policy/resource_arm_policy_remediation.go
katbyte: 👍

azurerm/internal/services/policy/parse/definition.go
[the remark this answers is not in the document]
katbyte: kk 👍

azurerm/internal/services/policy/resource_arm_policy_remediation.go
```diff
@@ -0,0 +1,248 @@
...
+	return &result
+}
+
+func flattenArmRemediationLocationFilters(input *policyinsights.RemediationFilters) []interface{} {
+	if input == nil {
+		return make([]interface{}, 0)
+	}
+
+	return utils.FlattenStringSlice(input.Locations)
+}
```
katbyte: i think for both these functions we could just inline the expand/flatten slice functions?

azurerm/internal/services/policy/validate/definition.go
```diff
@@ -0,0 +1,41 @@
...
+
+	if len(segments) != 2 {
+		errors = append(errors, fmt.Errorf("expected %q to have 2 segments after splition", k))
+		return
+	}
+
+	scope := segments[0]
+	// scope should be a resource group ID, or Management Group ID
+	_, err := parse.PolicyScopeID(scope)
+	if err != nil {
```
katbyte: these two lines could be merged

azurerm/internal/services/policy/resource_arm_policy_remediation.go
ArcturusZhang: Emm, in my memory, the inline call was the first version of this, and then I changed them into this... Let me check if there is problems when changing these back

azurerm/internal/services/policy/validate/definition.go
ArcturusZhang: Done

azurerm/internal/services/policy/resource_arm_policy_remediation.go
ArcturusZhang: Done

ArcturusZhang:
Hi @katbyte I resolved the final two minor comments

ArcturusZhang:
I also included some improvement of policy_definition in this PR. Acc test result for the policy definition:
image
But still the test case related with mgmt group fails with authorization error:
```
Status=403 Code="AuthorizationFailed" Message="The client '***' with object id '***' does not have authorization to perform action 'Microsoft.Management/managementGroups/read' over scope '/providers/Microsoft.Management/managementGroups/***' or the scope is invalid. If access was recently granted, please refresh your credentials."
```

kailunshi:
Hi @ArcturusZhang 

First at all, thank you for the great work. I know this is merged and you might not be following it but I discovered that ResourceDiscoveryMode option is not included. https://github.com/Azure/azure-sdk-for-go/blob/master/services/policyinsights/mgmt/2019-10-01/policyinsights/models.go#L2056

Could you please shed some light? let me know if you want me to create a new ticket.

ArcturusZhang:
> Hi @ArcturusZhang
> 
> First at all, thank you for the great work. I know this is merged and you might not be following it but I discovered that ResourceDiscoveryMode option is not included. https://github.com/Azure/azure-sdk-for-go/blob/master/services/policyinsights/mgmt/2019-10-01/policyinsights/models.go#L2056
> 
> Could you please shed some light? let me know if you want me to create a new ticket.

Well, this field initially does not exist in the version of go SDK terraform is using when I first work on this resource (Dec. 2019). I will take some time to add this into this resource, thanks for remind me.

kailunshi:
> > Hi @ArcturusZhang
> > First at all, thank you for the great work. I know this is merged and you might not be following it but I discovered that ResourceDiscoveryMode option is not included. https://github.com/Azure/azure-sdk-for-go/blob/master/services/policyinsights/mgmt/2019-10-01/policyinsights/models.go#L2056
> > Could you please shed some light? let me know if you want me to create a new ticket.
> 
> Well, this field initially does not exist in the version of go SDK terraform is using when I first work on this resource (Dec. 2019). I will take some time to add this into this resource, thanks for remind me.

Makes total sense. azure is fast evolving. :)

Thank you!

PR #5761, opened 2020-02-14
azurerm_function_app - Add WEBSITE_CONTENT & WEBSITE_CONTENTAZUREFILECONNECTIONSTRING  for premium plans

francescopersico:
The actual code only works for consumption plan but WEBSITE_CONTENT settings are required also on premium plans.


francescopersico:
@katbyte 

The WEBSITE_CONTENTSHARE and WEBSITE_CONTENTAZUREFILECONNECTIONSTRING app_settings are required for consumption and **premium plans** as stated here https://docs.microsoft.com/en-us/azure/azure-functions/functions-app-settings#website_contentazurefileconnectionstring.

The code in master branch automatically add these settings but only for consumption plan (the reason is that the premium plan is quite new). What i have done is change the "if" to also include the settings on premium plans.

PR #5765, opened 2020-02-14
New Resource: `azurerm_network_interface_security_group_association`

tombuildsstuff:
This commit introduces a new resource for linking a Network Interface and a Network Security Group which is required to work around a dependency issue in the Azure API where the NSG must be detached in order to be able to delete dependent resources. By introducing a separate resource to do this we can force this in Terraform which also handily works around the issue in the Azure API.

Due to a design limitation which exists in the Azure SDK for Go <-> Azure Resource Manager API - this PR introduces an override from the Go SDK which allows us to remove the Network Security Group ID - which is required for this resource to work.

Ultimately this resource & a patch to the Application Security Group association resource may need to be backported to 1.x - will confirm on Monday.

### DIFF
--- a/azurerm/internal/services/network/network_interface_network_security_group_association_resource.go
+++ b/azurerm/internal/services/network/network_interface_network_security_group_association_resource.go
@@ -1,0 +1,218 @@
+package network
+
+import (
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-09-01/network"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/azuresdkhacks"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/locks"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func resourceArmNetworkInterfaceSecurityGroupAssociation() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceArmNetworkInterfaceSecurityGroupAssociationCreate,
+		Read:   resourceArmNetworkInterfaceSecurityGroupAssociationRead,
+		Delete: resourceArmNetworkInterfaceSecurityGroupAssociationDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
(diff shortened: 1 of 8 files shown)


tombuildsstuff:
Tests pass:

Screenshot 2020-02-17 at 09 20 49

azurerm/internal/services/network/tests/network_interface_network_security_group_association_resource_test.go
```diff
@@ -0,0 +1,281 @@
...
+%s
+
+resource "azurerm_network_interface_security_group_association" "import" {
+  network_interface_id      = azurerm_network_interface_security_group_association.test.network_interface_id
+  network_security_group_id = azurerm_network_interface_security_group_association.test.network_security_group_id
+}
+`, template)
+}
+
+func testAccAzureRMNetworkInterfaceSecurityGroupAssociation_updateNIC(data acceptance.TestData) string {
```
jackofallops: Looks like the NIC isn't in the template, so this will create and not update?

website/docs/r/network_interface_security_group_association.html.markdown
```diff
@@ -0,0 +1,91 @@
...
+* `update` - (Defaults to 30 minutes) Used when updating the association between the Network Interface and the Network Security Group.
+* `read` - (Defaults to 5 minutes) Used when retrieving the association between the Network Interface and the Network Security Group.
+* `delete` - (Defaults to 30 minutes) Used when deleting the association between the Network Interface and the Network Security Group.
+
+## Import
+
+Associations between Network Interfaces and Network Security Group can be imported using the `resource id`, e.g.
+
+```shell
+terraform import azurerm_network_interface_security_group_association.association1 /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/mygroup1/providers/microsoft.network/networkInterfaces/nic1/ipConfigurations/example|/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/group1/providers/Microsoft.Network/networkSecurityGroups/group1
```
jackofallops: Will this not get caught by the shell interpreter as a pipe?

azurerm/internal/services/network/network_interface_network_security_group_association_resource.go
```diff
@@ -0,0 +1,218 @@
...
+		return fmt.Errorf("Error retrieving Network Interface %q (Resource Group %q): %+v", networkInterfaceName, resourceGroup, err)
+	}
+
+	props := read.InterfacePropertiesFormat
+	if props == nil {
+		return fmt.Errorf("Error: `properties` was nil for Network Interface %q (Resource Group %q)", networkInterfaceName, resourceGroup)
+	}
+
+	// first double-check it doesn't exist
+	resourceId := fmt.Sprintf("%s|%s", networkInterfaceId, networkSecurityGroupId)
```
jackofallops: The pipe separator may cause issue with `terraform import` ? (see comment on docs)

azurerm/internal/services/network/tests/network_interface_network_security_group_association_resource_test.go
tombuildsstuff: that's intentional since this is defined in the `Basic` to be able to override it here

website/docs/r/network_interface_security_group_association.html.markdown
tombuildsstuff: it will - but we expect on Windows folks will quote these ID's - there's probably a separate task to go through and update those in the docs 🤔

tombuildsstuff:
Superseded by #5784

PR #5773, opened 2020-02-16
On metric alert rule creation, the 'autoMitigate' property should be set to true by default

sivanguetta:
On metric alert rule creation, the 'autoMitigate' property should be set to true by default.
When 'autoMitigate' is set to false monitor condition not changing from 'Fired' to 'Resolved' automatically as the customers are expected.

### DIFF
--- a/azurerm/internal/services/monitor/resource_arm_monitor_metric_alert.go
+++ b/azurerm/internal/services/monitor/resource_arm_monitor_metric_alert.go
@@ -166,7 +166,7 @@ func resourceArmMonitorMetricAlert() *schema.Resource {
 			"auto_mitigate": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
+				Default:  true,
 			},
 
 			"description": {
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/monitor/resource_arm_monitor_metric_alert.go
```diff
@@ -166,7 +166,7 @@ func resourceArmMonitorMetricAlert() *schema.Resource {
 			"auto_mitigate": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
+				Default:  true,
```
tombuildsstuff: @sivanguetta thanks for this PR - just checking - is this behaviour mirrored in the Azure CLI/Portal? Thanks!

sivanguetta: > @sivanguetta thanks for this PR - just checking - is this behaviour mirrored in the Azure CLI/Portal? Thanks!

In Azure portal - 'auto_mitigate' is always true, the user cannot change this property.

In Azure CLI - in case that 'auto_mitigate' property was not provided, the metric alert rule is created without 'auto_mitigate' property (it is an optional property). 
When metric alert rule is created without 'auto_mitigate', it is considered as 'true'. 

As there was no indication what is the behavior when auto_mitigate property is not provided, I have raised a PR to update the documentation- https://github.com/Azure/azure-rest-api-specs/pull/8412

tombuildsstuff: @sivanguetta makes sense to me 👍

sivanguetta: > @sivanguetta makes sense to me 👍

@tombuildsstuff 
Thank you for the review:-)
I see that all checks have passed but I can't merge the pull request because I'm not authorized.
How can I merge the PR? what is missing?

sivanguetta:
@tombuildsstuff
Thank you for the review:-)
I see that all checks have passed but I can't merge the pull request because I'm not authorized.
How can I merge the PR? what is missing?

PR #5778, opened 2020-02-16
azurerm_cognitive_account - support for the `qna_runtime_endpoint` property

### LINKED ISSUE #4338: azurerm_cognitive_account QnAMaker missing api-properties
When running a terraform apply to create a QnAMaker service. I receive the following error.

```
Error: Error creating Cognitive Services Account "qnamaker-stable" (Resource Group "qnamaker-stable"): cognitiveservices
.AccountsClient#Create: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returne
d an error. Status=400 Code="ApiPropertiesInvalid" Message="The given 'apiProperties' '{}' is invalid. Validation errors
: Required properties are missing from object: qnaRuntimeEndpoint. Path ''." 
```

When adding:
```
api-properties = {
    "qnaRuntimeEndpoint" = module.qnamaker.default_site_hostname
  }
```

I get the following:
```
Error: Unsupported argument

  on qnamaker.tf line 56, in module "qnamaker":
  56:   api-properties = {

An argument named "api-properties" is not expected here.
 ```

For reference creating via CLI looks like this:
```
az cognitiveservices account create -g devexp-qnamaker -n devexp-qnamaker-account --kind QnAMaker --sku S0 --location westus --api-properties qnaRuntimeEndpoint=https://devexp-qnamaker-webapp.azurewebsites.net
```

### PR DESCRIPTION
**Pull request by RJPearson94**
This commit creates a new optional `qna_runtime_endpoint` property at the top level of the cognitive account resource 

Previously if the kind of a cognitive service account was set to `QnAMaker` an error would be thrown from the Microsoft API (see issue for more details). 
This issue was blocked due to missing functionality in the Azure SDK however support to allow the Account Properties and Account API Properties to be set was added in V25. This change allows the `qna_runtime_endpoint` to be set and will allow `QnAMaker` cognitive service accounts to be created via Terraform. 

I have added 2 Acceptance Tests for setting the `qna_runtime_endpoint` with `QnAMaker` set as the kind, 1 as a new resource and the 2nd with the url updated. I tried to add an acceptance test for setting the `qna_runtime_endpoint` on a non `QnAMaker` kind cognitive service account i.e. `Face` however the Microsoft API ignores this property when it is supplied so it is not returned in the response. As a result the tests failed with the following error `ImportStateVerify attributes not equivalent`. I decided remove this test.

The website docs have been updated to match the current functionality.

Fixes #4338

I am still quite new to Go and this is my first commit to the Terraform Providers so any feedback will be greatly appreciated

### CODE REVIEW

File: azurerm/internal/services/cognitive/resource_arm_cognitive_account.go [hunk 1]
```diff
@@ -98,6 +98,32 @@ func resourceArmCognitiveAccount() *schema.Resource {
 				}, false),
 			},
 
+			"properties": {
```
katbyte: Could we remove this block and pull its properties out into the top level?

azurerm/internal/services/cognitive/tests/resource_arm_cognitive_account_test.go
```diff
@@ -144,6 +144,71 @@ func TestAccAzureRMCognitiveAccount_update(t *testing.T) {
...
+					resource.TestCheckResourceAttr(data.ResourceName, "kind", "QnAMaker"),
+					resource.TestCheckResourceAttr(data.ResourceName, "properties.#", "1"),
+					resource.TestCheckResourceAttr(data.ResourceName, "properties.0.api_properties.#", "1"),
+					resource.TestCheckResourceAttr(data.ResourceName, "properties.0.api_properties.0.qna_runtime_endpoint", "https://localhost:8080/"),
+					resource.TestCheckResourceAttrSet(data.ResourceName, "endpoint"),
+					resource.TestCheckResourceAttrSet(data.ResourceName, "primary_access_key"),
+					resource.TestCheckResourceAttrSet(data.ResourceName, "secondary_access_key"),
+				),
+			},
+			{
```
katbyte: could we put a import step here?

RJPearson94:
Closing PR due to issues of additional commit being included. i will reopen a new one

azurerm/internal/services/cognitive/resource_arm_cognitive_account.go
RJPearson94: the api_properties block has now been moved to the root

azurerm/internal/services/cognitive/tests/resource_arm_cognitive_account_test.go
RJPearson94: done

RJPearson94:
Thanks for the feedback @katbyte. I have updated the PR with the changes that you recommended

RJPearson94:
Hi @katbyte, thanks for looking at the PR is it possible to get these updates reviewed please

RJPearson94:
Hey @katbyte, @mbfrahry & @tombuildsstuff, is it possible to get this change reviewed as it would be great to get this out as part of the 2.3.0 release.

azurerm/internal/services/cognitive/resource_arm_cognitive_account.go
```diff
@@ -146,11 +163,16 @@ func resourceArmCognitiveAccountCreate(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf("error expanding sku_name for Cognitive Account %s (Resource Group %q): %v", name, resourceGroup, err)
 	}
 
+	accountProperties := expandCognitiveAccountProperties(d)
+	if err := validateCognitiveAccountProperties(accountProperties, kind); err != nil {
```
katbyte: We should just do the validation in the expand function to be consistent with the rest of the provider

azurerm/internal/services/cognitive/resource_arm_cognitive_account.go
```diff
@@ -98,6 +98,22 @@ func resourceArmCognitiveAccount() *schema.Resource {
 				}, false),
 			},
 
+			"api_properties": {
```
katbyte: Just a random thought, but what do you think about removing this block and just having the sub properties at the top level? it would simplify the code, remove the expand/flatten functions, and provider a cleaner UX. I don't know if grouping them under `api_properties` really does much?

WDYT?

RJPearson94: yeah very good point, I left it in the api_properties block to keep parity with the SDK however I agree it will be cleaner at the top level. I will update the code

azurerm/internal/services/cognitive/resource_arm_cognitive_account.go
RJPearson94: done

azurerm/internal/services/cognitive/resource_arm_cognitive_account.go
RJPearson94: done

RJPearson94:
Hi @katbyte, thanks for the feedback. 

I have made the changes that you recommended.

The Cognitive Service Account for QnAMaker isn't available in all regions (I can't find any documentation stating where these Cognitive Service Accounts can be provisioned). I have been testing with West US (and this is what the Azure Portal defaults to) so I have hardcoded this location in the tests

RJPearson94:
Thanks for making the changes and merging this @katbyte, greatly appreciated 👍

PR #5779, opened 2020-02-16
dns - add parse ID and func

tracypholmes:


azurerm/internal/services/dns/resource_arm_dns_ns_record.go
```diff
@@ -33,6 +31,9 @@ func resourceArmDnsNsRecord() *schema.Resource {
 			Delete: schema.DefaultTimeout(30 * time.Minute),
 		},
 
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
```
jackofallops: Should we use the NsRecord parser here as with the other resources?

tracypholmes: Good catch - TY!

PR #5786, opened 2020-02-17
 todos and add new api_server_access_policy block

jackofallops:
opening for early review
still todo: data_source, docs, nil deref bug fix


tombuildsstuff:
Thinking about this, I think we're better holding off on this change for the moment, since this resource may have further changes we may need to introduce at the same time

PR #5791, opened 2020-02-17
ACR docs update

johnwildes:
I wanted to make a slight correction to the docs based on experience with deploying the resource.

1.) de-emphasizing the `Classic` SKU tier option, as it is not supported on new resources as of the March 2019 date.  It is still listed in the October 2019 API for the resource provider.  Probably can't remove it from the terraform provider to support those folks who have deployed it as `Classic`

2.) Added a note to emphasize that the `georeplication_locations` parameter is only supported on `Premium` SKU tier deployments.  This is of course on the Microsoft Docs, but if you just come here to learn how to deploy the resource, and think you can use it as an `optional` parameter, it'll fail if the SKU is not `Premium`.


johnwildes:
@tombuildsstuff I apologize for the noob ness here.  I forked the `terraform-azure-provider` repo.  Create a branch for my updates to the ACR doc, made my updates, and committed them to that branch.  When I submitted this PR, the only thing I needed to be changed is the `acrDocsUpdate` branch.  Not sure how these other 3 commits are being referenced.  It looks like it's going to pass the checks, but the commit history is going to be a bit whacked, unless someone can help educate me on why those other commits showed up.

tombuildsstuff:
@johnwildes all good about the commits, looks like you PR'd the changes from master to your fork :)

PR #5523, opened 2020-01-25
New Resource: `azurerm_express_route_gateway`

brent-au:
This PR adds the new ExpressRoute Gateway resource which can be deployed into Virtual Hub networks.

PR #5806, opened 2020-02-18
r/express_route_gateway: fixes from #5523

tombuildsstuff:
Travis is being slow so I've intentionally split these out into a new PR rather than pushing to the existing one since these are minor changes


tombuildsstuff:
The acceptance tests are failing due to some transient issues in Azure - but they otherwise look fine 👍

PR #5812, opened 2020-02-19
auth: using the correct authorizer for the service principals client

tombuildsstuff:
fixes a bug introduced in refactoring in #5748


tombuildsstuff:
Tests pass:

Screenshot 2020-02-19 at 08 50 07

PR #5824, opened 2020-02-20
Support more properties for azurerm_kubernetes_cluster

### LINKED ISSUE #5646: Support for new managed cluster load balancer profile settings in kubernetes
### Community Note

### Description

Hi.

Recently in AKS new load balancer profile settings appeared:
https://docs.microsoft.com/en-us/rest/api/aks/managedclusters/createorupdate#managedclusterloadbalancerprofile

These are:
allocatedOutboundPorts - Desired number of allocated SNAT ports per VM. Allowed values must be in the range of 0 to 64000 (inclusive). The default value is 0 which results in Azure dynamically allocating ports.
idleTimeoutInMinutes - Desired outbound flow idle timeout in minutes. Allowed values must be in the range of 4 to 120 (inclusive). The default value is 30 minutes.

We are currently setting this options using az cli and would be happy to set them using azurerm. 

* azurerm_kubernetes_cluster

```
resource "azurerm_kubernetes_cluster" "default" {
...
 network_profile {
  ...
      load_balancer_profile {
      managed_outbound_ip_count = 5
      ...
      allocated_outbound_ports = 10000
      idle_timeout_in_minutes = 4
    }
  }
}
```

### References

* https://docs.microsoft.com/en-us/rest/api/aks/managedclusters/createorupdate#managedclusterloadbalancerprofile

neil-yechenwei:
fixes #5646


neil-yechenwei:
@katbyte , thanks for you comments. I added test case to support new properties.

azurerm/internal/services/containers/tests/resource_arm_kubernetes_cluster_network_test.go
```diff
@@ -1217,6 +1245,52 @@ resource "azurerm_kubernetes_cluster" "test" {
 `, data.RandomInteger, data.Locations.Primary, data.RandomInteger, data.RandomInteger, data.RandomInteger, data.RandomInteger, data.RandomInteger, currentKubernetesVersion, data.RandomInteger, clientId, clientSecret)
 }
 
+func testAccAzureRMKubernetesCluster_standardLoadBalancerProfileWithPortAndTimeoutConfig(data acceptance.TestData, clientId string, clientSecret string) string {
+	return fmt.Sprintf(`
+resource "azurerm_resource_group" "test" {
```
katbyte: We'll need to add a provider block here

azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
```diff
@@ -1487,29 +1499,31 @@ func expandLoadBalancerProfile(d []interface{}, loadBalancerType string) (*conta
 
 	config := d[0].(map[string]interface{})
 
-	var managedOutboundIps *containerservice.ManagedClusterLoadBalancerProfileManagedOutboundIPs
-	var outboundIpPrefixes *containerservice.ManagedClusterLoadBalancerProfileOutboundIPPrefixes
-	var outboundIps *containerservice.ManagedClusterLoadBalancerProfileOutboundIPs
+	loadBalancerProfile := &containerservice.ManagedClusterLoadBalancerProfile{}
```
katbyte: for the sake of brevity could we use
```suggestion
	profile := &containerservice.ManagedClusterLoadBalancerProfile{}
```

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -323,6 +323,10 @@ A `load_balancer_profile` block supports the following:
 
 ~> **NOTE:** These options are mutually exclusive. Note that when specifying `outbound_ip_address_ids` ([azurerm_public_ip](/docs/providers/azurerm/r/public_ip.html)) the SKU must be `Standard`.
 
+* `allocated_outbound_port` - (Optional) Count of desired allocated SNAT port per VM for the cluster load balancer. Must be in the range of [0, 64000].
```
katbyte: Could we be consistent with other docs and
```suggestion
* `allocated_outbound_port` - (Optional) Count of desired allocated SNAT port per VM for the cluster load balancer. Must be between `0` and `64000` inclusive.
```

azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
```diff
@@ -424,6 +424,18 @@ func resourceArmKubernetesCluster() *schema.Resource {
 							Computed: true,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
+									"allocated_outbound_port": {
+										Type:         schema.TypeInt,
+										Optional:     true,
+										Default:      -1,
```
katbyte: What is the reason for -1 here? can't we just do d.GetOk() and only pass along the value if it is specified?

neil-yechenwei: Nope. Because d.GetOk() will return "0" for TypeInt when you didn't specify "allocated_outbound_port" in tf config. Then service side would set this property as "0" but actually I don't want to set it as "0". I just want to set it as nil.

azurerm/internal/services/containers/tests/resource_arm_kubernetes_cluster_network_test.go
neil-yechenwei: Updated

azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
neil-yechenwei: Updated

website/docs/r/kubernetes_cluster.html.markdown
neil-yechenwei: Updated

neil-yechenwei:
@katbyte , Thanks for your comments. I've updated code.

azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
katbyte: `GetOk` returns a `int, bool`, if the property isn't set i believe the boolean will be false? @neil-yechenwei

neil-yechenwei: Yes. The boolean will be false if the property isn't set in tf config. But actually sometimes we don't want to set it as "false". We just want to set it as nil when the property isn't set in tf config.

neil-yechenwei:
@katbyte , I've answered your concern. Please have a look.

azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
katbyte: @neil-yechenwei, does this not work:
```
	if port, ok := config["allocated_outbound_port"].(int); ok {
		profile.AllocatedOutboundPorts = utils.Int32(int32(port))
	}
```

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -273,6 +273,10 @@ A `load_balancer_profile` block supports the following:
 
 ~> **NOTE:** These options are mutually exclusive. Note that when specifying `outbound_ip_address_ids` ([azurerm_public_ip](/docs/providers/azurerm/r/public_ip.html)) the SKU must be `Standard`.
 
+* `allocated_outbound_port` - (Optional) Count of desired allocated SNAT port per VM for the cluster load balancer. Must be between `0` and `64000` inclusive.
```
katbyte: I'm not sure what this property is, an allocated port or the number of ports? count here, (and the API being ports) lead me to believe it's a count? if so we should change the name? maybe
```suggestion
* `outbound_ports_allocated` - (Optional) Number of desired SNAT port for each VM in the clusters load balancer. Must be between `0` and `64000` inclusive.
```

or maybe `outbound_port_count`?

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -273,6 +273,10 @@ A `load_balancer_profile` block supports the following:
...
+
+* `idle_timeout_in_minutes` - (Optional) Desired outbound flow idle timeout in minutes for the cluster load balancer. Must be in the range of [4, 120].
```
katbyte: Could we change this to
```suggestion
* `idle_timeout_in_minutes` - (Optional) Desired outbound flow idle timeout in minutes for the cluster load balancer. Must be between `4` and `120` inclusive.
```

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -273,6 +273,10 @@ A `load_balancer_profile` block supports the following:
...
+
 * `managed_outbound_ip_count` - (Optional) Count of desired managed outbound IPs for the cluster load balancer. Must be in the range of [1, 100].
```
katbyte: could we also fix the range here to?

azurerm/internal/services/containers/tests/resource_arm_kubernetes_cluster_network_test.go
```diff
@@ -1261,6 +1289,56 @@ resource "azurerm_kubernetes_cluster" "test" {
...
 }
 
+func testAccAzureRMKubernetesCluster_standardLoadBalancerProfileWithPortAndTimeoutConfig(data acceptance.TestData, clientId string, clientSecret string) string {
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
+
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-%d"
```
katbyte: could we make this
```suggestion
  name     = "acctestRG-aks-%d"
```

neil-yechenwei: Updated

website/docs/r/kubernetes_cluster.html.markdown
neil-yechenwei: Updated

website/docs/r/kubernetes_cluster.html.markdown
neil-yechenwei: Updated

website/docs/r/kubernetes_cluster.html.markdown
neil-yechenwei: Updated with outbound_ports_allocated

neil-yechenwei:
@katbyte , yes. below code you suggested doesn't work. Because the code "config["allocated_outbound_port"].(int)" would return "0" when tfconfig doesn't specify the property "allocated_outbound_port" and there is no "Default: -1". Then the property "profile.AllocatedOutboundPorts" would be set as "0" and pass to service side and “0” is meaningful for this property in service side. But actually I just want to set this property as "nil" when tf config doesn't specify this property. So if I use below code, I cannot distinguish user specified "0" or nil. 

if port, ok := config["allocated_outbound_port"].(int); ok {
	profile.AllocatedOutboundPorts = utils.Int32(int32(port))
}

neil-yechenwei:
@katbyte , I've updated code. Thanks for your comments.

katbyte:
@neil-yechenwei, have you tried removing the default? If there is no default, and the property is not computed, `ok` should be false in that if statement? and thus nothing will be passed along to the API?

azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go
neil-yechenwei: yes. it doesn't work. Because the code "config["allocated_outbound_port"].(int)" would return "0" when tfconfig doesn't specify the property "allocated_outbound_port" and there is no "Default: -1". Then the property "profile.AllocatedOutboundPorts" would be set as "0" and pass to service side and “0” is meaningful for this property in service side. But actually I just want to set this property as "nil" when tf config doesn't specify this property. So if I use below code, I cannot distinguish user specified "0" or nil.

neil-yechenwei:
@katbyte , I've tried before but it still isn't expected.

tfconfig:
```
provider "azurerm" {
  features {}
}

  name     = "acctestRG-aks-%d"
  location = "%s"
}

resource "azurerm_kubernetes_cluster" "test" {
  name                = "acctestaks%d"
  dns_prefix          = "acctestaks%d"
  kubernetes_version  = "%s"

  linux_profile {
    admin_username = "acctestuser%d"

    ssh_key {
      key_data = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCqaZoyiz1qbdOQ8xEf6uEu1cCwYowo5FHtsBhqLoDnnp7KUTEBN+L2NxRIfQ781rxV6Iq5jSav6b2Q8z5KiseOlvKA/RF2wqU0UPYqQviQhLmW6THTpmrv/YkUCuzxDpsH7DUDhZcwySLKVVe0Qm3+5N2Ta6UYH3lsDf9R9wTP2K/+vAnflKebuypNlmocIvakFWoZda18FOmsOoIVXQ8HWFNCuw9ZCunMSN62QGamCe3dL5cXlkgHYv7ekJE15IA9aOJcM7e90oeTqo+7HTcWfdu0qQqPWY5ujyMw/llas8tsXY85LFqRnr3gJ02bAscjc477+X+j/gkpFoN1QEmt terraform@demo.tld"
    }
  }

  default_node_pool {
    name       = "default"
    node_count = 1
    vm_size    = "Standard_DS2_v2"
  }

  service_principal {
    client_id     = "%s"
    client_secret = "%s"
  }

  network_profile {
    network_plugin    = "kubenet"
    load_balancer_sku = "Standard"
    load_balancer_profile {
      managed_outbound_ip_count = 2
      idle_timeout_in_minutes   = 10
    }
  }
}

tf schema definition:
```
"load_balancer_profile": {
							Type:     schema.TypeList,
							MaxItems: 1,
							ForceNew: true,
							Optional: true,
							Computed: true,
							Elem: &schema.Resource{
									"outbound_ports_allocated": {
										Type:         schema.TypeInt,
										Optional:     true,
										ValidateFunc: validation.IntBetween(0, 64000),
									},
```

the result of debugging:
![image](https://user-images.githubusercontent.com/19754191/76601005-05a2f000-6543-11ea-8e97-21b687001046.png)

**Comment by katbyte** [comment]
Well dang, look at that. Guess nested blocks don't behave in the way I expected them to be like ones on the top level 😞 sorry about that. If that is the case instead of -1 could we use nil for the default?

**Comment by neil-yechenwei** [comment]
@katbyte , seems we cannot use nil for the default. We still encountered the same issue after tested.

tfconfig without "outbound_ports_allocated":
```
provider "azurerm" {
  features {}
}

  name     = "acctestRG-aks-%d"
  location = "%s"
}

resource "azurerm_kubernetes_cluster" "test" {
  name                = "acctestaks%d"
  dns_prefix          = "acctestaks%d"
  kubernetes_version  = "%s"

  linux_profile {
    admin_username = "acctestuser%d"

    ssh_key {
      key_data = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCqaZoyiz1qbdOQ8xEf6uEu1cCwYowo5FHtsBhqLoDnnp7KUTEBN+L2NxRIfQ781rxV6Iq5jSav6b2Q8z5KiseOlvKA/RF2wqU0UPYqQviQhLmW6THTpmrv/YkUCuzxDpsH7DUDhZcwySLKVVe0Qm3+5N2Ta6UYH3lsDf9R9wTP2K/+vAnflKebuypNlmocIvakFWoZda18FOmsOoIVXQ8HWFNCuw9ZCunMSN62QGamCe3dL5cXlkgHYv7ekJE15IA9aOJcM7e90oeTqo+7HTcWfdu0qQqPWY5ujyMw/llas8tsXY85LFqRnr3gJ02bAscjc477+X+j/gkpFoN1QEmt terraform@demo.tld"
    }
  }

  default_node_pool {
    name       = "default"
    node_count = 1
    vm_size    = "Standard_DS2_v2"
  }

  service_principal {
    client_id     = "%s"
    client_secret = "%s"
  }

  network_profile {
    network_plugin    = "kubenet"
    load_balancer_sku = "Standard"
    load_balancer_profile {
      managed_outbound_ip_count = 2
    }
  }
}
```

tf schema definition:
```
"load_balancer_profile": {
							Type:     schema.TypeList,
							MaxItems: 1,
							ForceNew: true,
							Optional: true,
							Computed: true,
							Elem: &schema.Resource{
									"outbound_ports_allocated": {
										Type:         schema.TypeInt,
										Optional:     true,
										Default:      nil,
										ValidateFunc: validation.IntBetween(0, 64000),
									},
```

The result of debugging:
image

File: azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go (hunk 4 above)
Comment (jackofallops) [reply azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go]: `0` is the default in the service for this setting, and this value has special significance in that it instructs the service to dynamically allocate outbound ports as needed.  I think in this case a default value of `0`  and adding `Computed: true,` here would be acceptable, if properly documented, and exhibit the correct behaviour in use.

Comment (neil-yechenwei) [reply azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go]: @jackofallops , I don't think so. If you add `Default: 0` and `Computed: true` for this property, user cannot set it as `nil`. I think we should allow user to not set it. And it would be a breaking change if I update code as you suggested. It would impact other users who don't expect this property. And terraform doesn't allow us to set `Default: 0` and `Computed: true` together on same property. If you do this, it would threw below error message.

property schema:
```
"load_balancer_profile": {
							Type:     schema.TypeList,
							MaxItems: 1,
							ForceNew: true,
							Optional: true,
							Computed: true,
							Elem: &schema.Resource{
									"outbound_ports_allocated": {
										Type:         schema.TypeInt,
										Optional:     true,
										Computed:     true,
										Default:      0,
										ValidateFunc: validation.IntBetween(0, 64000),
									},
```

Error message:
```
            testing.go:640: Step 0 error: config is invalid: Internal validation of the provider failed! This is always a bug
                with the provider itself, and not a user issue. Please report
                this bug:
                
                1 error occurred:
                	* resource azurerm_kubernetes_cluster: outbound_ports_allocated: Default must be nil if computed
                
                
FAIL

Debugger finished with exit code 0

```

Comment (jackofallops) [reply azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go]: Sorry - poor wording there. I believe that a read-back value from the API for `AllocatedOutboundPorts` will contain the default value `0`, not `nil`, so we should be fine to set `Computed` and let the zero value pass through?  The user can not set nil explicitly, and omitting the setting should use the default since it must have a value between 0 and 64,000?

Comment (neil-yechenwei) [reply azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go]: @jackofallops , sorry, I don't think so. Service API would return `nil` when the property isn't set in tfconfig. So I think user is able to set it as `nil` without `breaking change`. If we set this property must be between 0 and 64000, then breaking change would happen. So I have a concern. When breaking change is allowed in terraform? The breaking change in here is allowed?

tfconfig without `outbound_ports_allocated`:
```
provider "azurerm" {
  features {}
}

  name     = "acctestRG-aks-%d"
  location = "westus2"
}

resource "azurerm_kubernetes_cluster" "test" {
  name                = "acctestaks%d"
  dns_prefix          = "acctestaks%d"
  kubernetes_version  = "%s"

  linux_profile {
    admin_username = "acctestuser%d"

    ssh_key {
      key_data = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCqaZoyiz1qbdOQ8xEf6uEu1cCwYowo5FHtsBhqLoDnnp7KUTEBN+L2NxRIfQ781rxV6Iq5jSav6b2Q8z5KiseOlvKA/RF2wqU0UPYqQviQhLmW6THTpmrv/YkUCuzxDpsH7DUDhZcwySLKVVe0Qm3+5N2Ta6UYH3lsDf9R9wTP2K/+vAnflKebuypNlmocIvakFWoZda18FOmsOoIVXQ8HWFNCuw9ZCunMSN62QGamCe3dL5cXlkgHYv7ekJE15IA9aOJcM7e90oeTqo+7HTcWfdu0qQqPWY5ujyMw/llas8tsXY85LFqRnr3gJ02bAscjc477+X+j/gkpFoN1QEmt terraform@demo.tld"
    }
  }

  default_node_pool {
    name       = "default"
    node_count = 1
    vm_size    = "Standard_DS2_v2"
  }

  service_principal {
    client_id     = "%s"
    client_secret = "%s"
  }

  network_profile {
    network_plugin    = "kubenet"
    load_balancer_sku = "Standard"
    load_balancer_profile {
      managed_outbound_ip_count = 2
    }
  }
}
```

terraform schema:
```
"load_balancer_profile": {
							Type:     schema.TypeList,
							MaxItems: 1,
							ForceNew: true,
							Optional: true,
							Computed: true,
							Elem: &schema.Resource{
									"outbound_ports_allocated": {
										Type:         schema.TypeInt,
										Optional:     true,
										Default:      0,
										ValidateFunc: validation.IntBetween(0, 64000),
									},
```

The result of `outbound_ports_allocated` from service API:
image

Comment (jackofallops) [reply azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go]: Hi @neil-yechenwei 
I meant the following for the schema:
```
						"load_balancer_profile": {
							Type:     schema.TypeList,
							MaxItems: 1,
							ForceNew: true,
							Optional: true,
							Computed: true,
							Elem: &schema.Resource{
									"outbound_ports_allocated": {
										Type:         schema.TypeInt,
										Optional:     true,
										Computed:     true,
										ValidateFunc: validation.IntBetween(0, 64000),
									},
```
And can I see the read comes back with a `0` from the API:
image

image


with config:
```
 network_profile {
    network_plugin    = "kubenet"
    load_balancer_sku = "Standard"
    load_balancer_profile {
      managed_outbound_ip_count = 2
      idle_timeout_in_minutes   = 10
    }
```
Does that help explain what I mean?

Comment (neil-yechenwei) [reply azurerm/internal/services/containers/resource_arm_kubernetes_cluster.go]: @jackofallops , sorry, seems I still cannot get your point. I still think your change would cause `breaking change` since some user maybe doesn't expect this change would impact their existing service. So I am asking does terraform allow `breaking change` in this situation. If yes, I can update this PR to `force user to set this property as 0-64000`. So do you mean ` user must set property "outbound_ports_allocated" as 0~64000`? And I've tried your above changes, you actually already set this property `outbound_ports_allocated` as `0` before calling service API so that you would get `0` from service side. Actually service API would return `nil` as default. Please see my below example. If you think I am not right, could you commit your whole changes to this PR for understanding?

tfconfig without property `outbound_ports_allocated`:
```
provider "azurerm" {
  features {}
}

  name     = "acctestRG-aks-%d"
  location = "westus2"
}

resource "azurerm_kubernetes_cluster" "test" {
  name                = "acctestaks%d"
  dns_prefix          = "acctestaks%d"
  kubernetes_version  = "%s"

  linux_profile {
    admin_username = "acctestuser%d"

    ssh_key {
      key_data = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCqaZoyiz1qbdOQ8xEf6uEu1cCwYowo5FHtsBhqLoDnnp7KUTEBN+L2NxRIfQ781rxV6Iq5jSav6b2Q8z5KiseOlvKA/RF2wqU0UPYqQviQhLmW6THTpmrv/YkUCuzxDpsH7DUDhZcwySLKVVe0Qm3+5N2Ta6UYH3lsDf9R9wTP2K/+vAnflKebuypNlmocIvakFWoZda18FOmsOoIVXQ8HWFNCuw9ZCunMSN62QGamCe3dL5cXlkgHYv7ekJE15IA9aOJcM7e90oeTqo+7HTcWfdu0qQqPWY5ujyMw/llas8tsXY85LFqRnr3gJ02bAscjc477+X+j/gkpFoN1QEmt terraform@demo.tld"
    }
  }

  default_node_pool {
    name       = "default"
    node_count = 1
    vm_size    = "Standard_DS2_v2"
  }

  service_principal {
    client_id     = "%s"
    client_secret = "%s"
  }

  network_profile {
    network_plugin    = "kubenet"
    load_balancer_sku = "Standard"
    load_balancer_profile {
      managed_outbound_ip_count = 2
    }
  }
}
```

terraform schema:
```
"load_balancer_profile": {
							Type:     schema.TypeList,
							MaxItems: 1,
							ForceNew: true,
							Optional: true,
							Computed: true,
							Elem: &schema.Resource{
									"outbound_ports_allocated": {
										Type:         schema.TypeInt,
										Optional:     true,
										Computed:     true,
										ValidateFunc: validation.IntBetween(0, 64000),
									},
```

Do not set property `outbound_ports_allocated` before calling service api:
image

Getting value of property `outbound_ports_allocated` from Service API:
image

So service API would return `nil` NOT `0` when you didn't set this property in tfconfig.

neil-yechenwei: @jackofallops , I've double confirmed with this service owner. They said property "AllocatedOutboundPorts" would be set to “0” and property “IdleTimeoutInMinutes” would be set to “30” in service side when we don't set these two properties in request body. So I've updated code. Please have a look. Thanks.

The answer from service owner:
Today, if a customer doesn’t set them explicitly then we don’t fill the two properties from the AKS side but they do get filled on the SLB side with the defaults.

neil-yechenwei:
@katbyte & @jackofallops , I've double confirmed with this service owner. They said property "AllocatedOutboundPorts" would be set to “0” and property “IdleTimeoutInMinutes” would be set to “30” in service side when we don't set these two properties in request body. So I've updated code. Thanks for your suggestions.

The answer from service owner:
_Today, if a customer doesn’t set them explicitly then we don’t fill the two properties from the AKS side but they do get filled on the SLB side with the defaults._

djsly:
nice work @neil-yechenwei , looking forward for this to be merged!

evenh:
Thanks for this work @neil-yechenwei! Is there more work that remains or just reviewing?

etiennetremel:
Thanks for putting this through @neil-yechenwei, any chance to get this merged?

tombuildsstuff:
hey @neil-yechenwei 

Since there's a number of other Pull Requests currently open for AKS - to be able to merge all of these without them all conflicting, I've pulled these commits locally, squashed them, fix the comments and then combined them into a single branch.

As such, whilst I'd like to thank you for opening this Pull Request, I'm going to close this in favour of #7233 which integrates these changes (and the other open AKS PR's - and includes some of the other open Feature Requests).

Thanks!

Issue #2826, opened 2019-02-01
AzureRM TXT record do not support more than 255 characters


jonathanlavoie-optel:
AzureRM TXT record do not support more than 255 characters

**My code:**
`resource "azurerm_dns_txt_record" "google" {
  name                = "google._domainkey"
  zone_name           = "${azurerm_dns_zone.module.name}"
  resource_group_name = "${azurerm_resource_group.module.name}"
  ttl                 = "${var.ttl}"

  record {
    value = "v=DKIM1; k=rsa; p=NqqBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIt7Hs45AQEAu6eALKOV7VkAS75B9tmQQnobrG6f4sKb7x+DlFsBehUx3txjIQscAiMz7IAfohs6agccc0t6DLndt9humMV7CGhOjm84ZufzPPSJT1MZHPh3IhEeW2ktxhFqI88RSzzACGyWSOKhymBbbbNI68nglYtG1OIQ0ZIO1EA/gu63RmroyG2iwxxbX2fG4QGRPvo8tOQyohTccButrh+sUilXtlAhwXQnCzlcakyANB1WPSeAvmWKfkmif/0BfcvbxbSf5oycs/7inV5cTtmhEdfYkogmwFh+RasHI8rlXZy5MkjuOB1TgplQ8upL046KM62Cq5GKqZ1R8lpkrzPPZZoAFwIDAQAB"
  }
}`

**Terraform error:**
`Error: Error applying plan:

1 error(s) occurred:

* module.behaviori_com.azurerm_dns_txt_record.google: 1 error(s) occurred:

[...]

Issue #5547, opened 2020-01-29
 azurerm_dns_txt_record truncates long TXT records on import 


bbkane:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

```
Terraform v0.12.20
+ provider.azurerm v1.42.0
```

### Affected Resource(s)

* `azurerm_dns_txt_record`

### Terraform Configuration Files

```hcl
terraform {
  backend "local" {
    path = "terraform.tfstate"
  }
}


provider "azurerm" {
  version = "~> 1.39"
}

resource "azurerm_resource_group" "B16_repro_long_txt_record" {
  name     = "B16_repro_long_txt_record"
  location = "West US 2"
}

resource "azurerm_dns_zone" "BbbkaneDcom" {
  name                = "bbkane.com"
  resource_group_name = azurerm_resource_group.B16_repro_long_txt_record.name
}

resource "azurerm_dns_txt_record" "long-value-portal_bbkane_com_TXT" {
  name                = "long-value-portal"
  resource_group_name = azurerm_resource_group.B16_repro_long_txt_record.name
  ttl                 = 3600
  zone_name           = azurerm_dns_zone.BbbkaneDcom.name
  record {

    # Need a TXT record with a value > 255 to trigger this (DKIM records, for example)
    # perl -E "print 'a' x 250, 'b' x 250"  | pbcopy
[...]

PR #5837, opened 2020-02-21
azurerm_dns_txt_record  - support records up to 1024 characters in length

njuCZ:
fixa #5547 

According to [dns txt record doc](https://docs.microsoft.com/en-us/azure/dns/dns-zones-records#txt-records), a record length can be up to 1024 characters. To align with azure.cli and powershell, I set each record set length to 254 characters

image


bootc:
I think this also fixes #2826

Issue #3083, opened 2019-03-20
app_service_plan should allow multiple kind


piizei:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
Terraform v0.11.10
provider.azurerm v1.22.1

### Affected Resource(s)

* `azurerm_app_service_plan`

### Issue

App service plan should allow multiple values in kind option:
kind                = "functionapp,linux"

this would reflect the behaviour of the ARM templates. Currently it is impossible to create consumption linux plan for functions.

* #0000

nicr9:
I've been trying to create a linux functionapp with Terraform recently. It's been frustrating 😂 

As an experiment I created the functionapp manually through the webconsole and then imported the `azurerm_app_service_plan` and `azurerm_function_app` so that I could inspect the differences with what terraform was creating.

I believe that the `azurerm_app_service_plan.kind` should remain as `functionapp` and the `azurerm_function_app.kind` field should be allowed to change to `functionapp,linux` somehow.

Check it out:

```
$ terraform show
# azurerm_app_service_plan.linuxapptest:
resource "azurerm_app_service_plan" "linuxapptest" {
    id                        = "..."
    kind                      = "functionapp" <-- Still just a functionapp plan
    location                  = "westus"
    maximum_number_of_workers = 0
    name                      = "..."
    per_site_scaling          = false
    reserved                  = true
    resource_group_name       = "..."

    properties {
        per_site_scaling = false
        reserved         = true
    }

    sku {
        capacity = 0
        size     = "Y1"
        tier     = "Dynamic"
    }
}

# azurerm_function_app.linuxapptest:
resource "azurerm_function_app" "linuxapptest" {
    app_service_plan_id            = "..."
    app_settings                   = {
        "FUNCTIONS_WORKER_RUNTIME"     = "python"
        "WEBSITE_NODE_DEFAULT_VERSION" = "10.14.1"
    }
    client_affinity_enabled        = false
    default_hostname               = "....azurewebsites.net"
    enable_builtin_logging         = false
    enabled                        = true
    https_only                     = false
[...]

PR #5839, opened 2020-02-21
fix for linux function apps

jackofallops:
addresses: #3083

### DIFF
--- a/azurerm/internal/services/web/data_source_function_app.go
+++ b/azurerm/internal/services/web/data_source_function_app.go
@@ -2,6 +2,7 @@ package web
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


Bessonov:
@jackofallops Great! But how it's different from the current approach described here: https://github.com/terraform-providers/terraform-provider-azurerm/issues/5650

jackofallops:
Hi @Bessonov 
Function Apps are something we're planning on taking another look at as the service has evolved and grown over time and, since it is controlled by the same API as Web Apps, needs some careful consideration and investigation, especially in areas that are possibly not well documented (as you mention in your issue, `kind` is something of a mystery!) .  Specifically the change here stopped the provider from removing 'linux'  portion of the `kind` value, which was previously reverting anything received to simply `functionapp` and gave the resource an awareness of OS's other than Windows (which is the default if linux is not present in the string).

PR #5842, opened 2020-02-21
Update AD domain regex for azure netapp files account to allow dash (-)

jknash:
Azure AD domain services allows a dash - in the domain name, but the regex that validates the domain name does not.  I changed the regex from: ^[(\da-zA-Z)\.]{1,255}$ 
to:  ^[(\da-zA-Z-)\.]{1,255}$
It's a subtle change, but we can't use the ANF provider until it is fixed, because our domain name contains a dash -


jknash:
Looking at this, I should have said this is for the azure netapp files account resource in the title.  It's not immediately clear.  Sorry!

PR #5853, opened 2020-02-24
azurerm_cosmosdb_mongo_collection - remove deprecated property

katbyte:


andreish:
as of writting this , documentation at https://www.terraform.io/docs/providers/azurerm/r/cosmosdb_mongo_collection.html still has example with indexes  block . please and note where this applies, and how to migrate !

PR #5875, opened 2020-02-25
New Tool: website-scaffolder

tombuildsstuff:
This tools intended to aid with building new resources, in particular by scaffolding the documentation from the resource definition.

For examples sake I've updated the documentation for both Resource Groups & Search Services using this output - seems to work well 👍🏻

### DIFF
--- a/azurerm/internal/tools/website-scaffold/main.go
+++ b/azurerm/internal/tools/website-scaffold/main.go
@@ -1,0 +1,73 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/provider"
+)
+
+func main() {
+	isResource := flag.Bool("resource", true, "True if this is a Resource, False if this is a Data Source")
+	resourceName := flag.String("name", "", "The name of the Data Source/Resource which should be generated")
+	flag.Parse()
+
+	if err := run(*isResource, *resourceName); err != nil {
+		panic(err)
+	}
+}
+
+func run(isResource bool, resourceName string) error {
+	if isResource {
+		return runForResource(resourceName)
+	}
+
+	return runForDataSource(resourceName)
+}
+
+func runForDataSource(resourceName string) error {
+	var dataSource *schema.Resource
+
+	for _, service := range provider.SupportedServices() {
+		for key, ds := range service.SupportedDataSources() {
+			if key == resourceName {
+				dataSource = ds
+				break
+			}
+		}
+	}
+
+	if dataSource == nil {
+		return fmt.Errorf("Data Source %q was not registered!", resourceName)
+	}
+
+	return buildDocumentationForDataSource(resourceName, dataSource)
+}
+
+func runForResource(resourceName string) error {
+	var resource *schema.Resource
+
+	for _, service := range provider.SupportedServices() {
+		for key, rs := range service.SupportedResources() {
+			if key == resourceName {
+				resource = rs
+				break
+			}
+		}
+	}
+
+	if resource == nil {
+		return fmt.Errorf("Resource %q was not registered!", resourceName)
+	}
+
+	return buildDocumentationForResource(resourceName, resource)
+}
+
+func buildDocumentationForDataSource(resourceName string, resource *schema.Resource) error {
+	return nil
+}
+
+func buildDocumentationForResource(resourceName string, resource *schema.Resource) error {
+	return nil
+}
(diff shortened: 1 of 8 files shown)


azurerm/internal/tools/website-scaffold/main.go
```diff
@@ -0,0 +1,891 @@
...
+			output += gen.processElementForExample(field, indentLevel, v.Elem, isAttribute)
+			continue
+		}
+
+		output += processField(field, v)
+	}
+	return strings.TrimSuffix(output, "\n")
+}
+
+func (gen documentationGenerator) sortFields(input map[string]*schema.Schema) []string {
```
magodo: Alphabetic ordering might not obey the conventions that:

1. Some special fields have special ordering, e.g. `name` comes first and `tags` comes last
2. Except point 1, `ForceNew` + `Required` > `ForceNew` + `Optional` > `Required` > `Optional`

azurerm/internal/tools/website-scaffold/main.go
```diff
@@ -0,0 +1,891 @@
...
+	// assuming we have both optional & required fields - let's add a separarer
+	if len(fields) > 0 && len(optionalFields) > 0 {
+		fields += "---\n\n"
+	}
+	fields += optionalFields
+
+	// first list all of the top-level fields / blocks alphabetically
+
+	// then we need to collect a list of all block names, everywhere
+	blockNames, blocks := gen.uniqueBlockNamesForArgument(gen.resource.Schema)
```
magodo: Given the case that there are two arguments sharing the same block type, then it will be repeated in the document, which seems redundent.

Barely comparing schema definition to predict whether two arguments share a same block type is not so reliable, however,  in this case maybe we can add a **TODO** comment to remind user to choose.

azurerm/internal/tools/website-scaffold/main.go
```diff
@@ -0,0 +1,891 @@
...
+
+	for _, blockName := range blockNames {
```
magodo: Maybe need to sort keys in `blockNames` in some way (either reflect the nested level or simply in alphabetic order).

magodo:
Maybe leveraging the `Description` in schema definition as a source of description of each property, and use `TODO` as a fallback when it is absent.

azurerm/internal/tools/website-scaffold/main.go
```diff
@@ -0,0 +1,891 @@
...
+		friendlyText := timeoutToFriendlyText(*timeouts.Read)
+		timeoutsText += fmt.Sprintf("* `read` - (Defaults to %s) Used when retrieving the %s.\n", friendlyText, gen.brandName)
+	}
+
+	if timeouts.Update != nil {
+		friendlyText := timeoutToFriendlyText(*timeouts.Update)
+		timeoutsText += fmt.Sprintf("* `update` - (Defaults to %s) Used when updating the %s.\n", friendlyText, gen.brandName)
+	}
+
+	if timeouts.Create != nil {
```
jackofallops: Copy paste error here?  Should be delete?

azurerm/internal/tools/website-scaffold/main.go
tombuildsstuff: this is already done within the other method

azurerm/internal/tools/website-scaffold/main.go
tombuildsstuff: we're scaffolding rather than generating - thus this is an acceptable limitation

website/docs/r/search_service.html.markdown
```diff
@@ -8,79 +8,83 @@ description: |-
...
 
 The following arguments are supported:
 
-* `name` - (Required) The name of the Search Service. Changing this forces a new resource to be created.
+* `location` - (Required) The Azure Region where the Search Service should exist. Changing this forces a new Search Service to be created.
+
+* `name` - (Required) The Name which should be used for this Search Service. Changing this forces a new Search Service to be created.
 
-* `resource_group_name` - (Required) The name of the resource group in which to create the Search Service. Changing this forces a new resource to be created.
+* `resource_group_name` - (Required) The name of the Resource Group where the Search Service should exist. Changing this forces a new Search Service to be created.
```
katbyte: Instead of doing alphabetical here should be be consistent with how we order them in the schema/typically right now? name->rg->location ?

website/docs/r/search_service.html.markdown
```diff
@@ -8,79 +8,83 @@ description: |-
...
 * `query_keys` - A `query_keys` block as defined below.
 
+* `secondary_key` - The Secondary Key used for Search Service Administration.
+
 ---
 
 A `query_keys` block exports the following:
 
-* `name` - The name of the query key.
+* `key` - The value of this Query Key.
```
katbyte: Same here, alphabetical vs typical

tombuildsstuff: in this instance these should be alphabetical for computed blocks, so I think this is fine?

website/docs/r/search_service.html.markdown
tombuildsstuff: that's mostly for historical reasons/where there's a larger set of fields - but it's not really giving us any value here, so I think we can ignore that for now

PR #5906, opened 2020-02-26
azurerm_application_gateway - support 125 for V2 sku capacity

neil-yechenwei:
Issue:
Allow sku.capacity up to 125 when using standard_v2 in the azurerm_application_gateway resource since sku.capacity is limited to 32 currently.

Description:
The Capacity of the SKU to use for this Application Gateway. Possible values for V1 SKU must be between 1 and 32, and for V2 SKU must be between 1 and 125.

SKU Reference doc: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits?toc=%2fazure%2fapplication-gateway%2ftoc.json#application-gateway-limits

### DIFF
--- a/azurerm/internal/services/network/resource_arm_application_gateway.go
+++ b/azurerm/internal/services/network/resource_arm_application_gateway.go
@@ -3728,16 +3728,21 @@ func flattenApplicationGatewayCustomErrorConfigurations(input *[]network.Applica
 }
 
 func ApplicationGatewayCustomizeDiff(d *schema.ResourceDiff, _ interface{}) error {
-	tier := d.Get("sku.0.tier").(string)
+	_, hasAutoscaleConfig := d.GetOk("autoscale_configuration.0")
 	capacity, hasCapacity := d.GetOk("sku.0.capacity")
+	tier := d.Get("sku.0.tier").(string)
+
+	if !hasAutoscaleConfig && !hasCapacity {
+		return fmt.Errorf("The Application Gateway must specify either `capacity` or `autoscale_configuration` for the selected SKU tier %q", tier)
+	}
 
 	if hasCapacity {
 		if (strings.EqualFold(tier, string(network.ApplicationGatewayTierStandard)) || strings.EqualFold(tier, string(network.ApplicationGatewayTierWAF))) && (capacity.(int) < 1 || capacity.(int) > 32) {
-			return fmt.Errorf("The capacity of V1 SKU %q must be between 1 and 32.", tier)
+			return fmt.Errorf("The value '%d' exceeds the maximum capacity allowed for a %q V1 SKU, the %q SKU must have a capacity value between 1 and 32", capacity, tier, tier)
 		}
 
 		if (strings.EqualFold(tier, string(network.ApplicationGatewayTierStandardV2)) || strings.EqualFold(tier, string(network.ApplicationGatewayTierWAFV2))) && (capacity.(int) < 1 || capacity.(int) > 125) {
-			return fmt.Errorf("The capacity of V2 SKU %q must be between 1 and 125.", tier)
+			return fmt.Errorf("The value '%d' exceeds the maximum capacity allowed for a %q V2 SKU, the %q SKU must have a capacity value between 1 and 125", capacity, tier, tier)
 		}
 	}
 
--- a/azurerm/internal/services/network/resource_arm_application_gateway.go
+++ b/azurerm/internal/services/network/resource_arm_application_gateway.go
@@ -662,7 +662,7 @@ func resourceArmApplicationGateway() *schema.Resource {
 						"capacity": {
 							Type:         schema.TypeInt,
 							Optional:     true,
-							ValidateFunc: validation.IntBetween(1, 32),
+							ValidateFunc: validation.IntBetween(1, 125),
 						},
 					},
 				},
--- a/azurerm/internal/services/network/resource_arm_application_gateway.go
+++ b/azurerm/internal/services/network/resource_arm_application_gateway.go
@@ -3,6 +3,7 @@ package network
 import (
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-09-01/network"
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/network/resource_arm_application_gateway.go
```diff
@@ -662,7 +662,7 @@ func resourceArmApplicationGateway() *schema.Resource {
 						"capacity": {
 							Type:         schema.TypeInt,
 							Optional:     true,
-							ValidateFunc: validation.IntBetween(1, 32),
+							ValidateFunc: validation.IntBetween(1, 125),
```
katbyte: Could we add a check to the create/update functions to ensure the 32 limit is respected when not using a V2 SKU?

website/docs/r/application_gateway.html.markdown
```diff
@@ -380,7 +380,7 @@ A `sku` block supports the following:
 
 * `tier` - (Required) The Tier of the SKU to use for this Application Gateway. Possible values are `Standard`, `Standard_v2`, `WAF` and `WAF_v2`.
 
-* `capacity` - (Required) The Capacity of the SKU to use for this Application Gateway - which must be between 1 and 10, optional if `autoscale_configuration` is set
+* `capacity` - (Required) The Capacity of the SKU to use for this Application Gateway. Possible values for V1 SKU must be between 1 and 32, and for V2 SKU must be between 1 and 125, optional if `autoscale_configuration` is set.
```
katbyte:
```suggestion
* `capacity` - (Required) The Capacity of the SKU to use for this Application Gateway. When using a V1 SKU this value must be between 1 and 32, and  1 to 125 for a V2 SKU. This property is optional if `autoscale_configuration` is set.
```

neil-yechenwei: fixed

azurerm/internal/services/network/resource_arm_application_gateway.go
neil-yechenwei: Maybe adding CustomizeDiff function is better. I've updated code. Please have a look. Thanks.

neil-yechenwei:
@katbyte , Thanks for your comments. I've updated code. Please have a look.

azurerm/internal/services/network/resource_arm_application_gateway.go
```diff
@@ -3724,3 +3726,20 @@ func flattenApplicationGatewayCustomErrorConfigurations(input *[]network.Applica
 
 	return results
 }
+
+func ApplicationGatewayCustomizeDiff(d *schema.ResourceDiff, _ interface{}) error {
+	tier := d.Get("sku.0.tier").(string)
+	capacity, hasCapacity := d.GetOk("sku.0.capacity")
+
+	if hasCapacity {
+		if (strings.EqualFold(tier, string(network.ApplicationGatewayTierStandard)) || strings.EqualFold(tier, string(network.ApplicationGatewayTierWAF))) && (capacity.(int) < 1 || capacity.(int) > 32) {
```
WodansSon: This is looks good, but do you think we can make the error message a bit more clear... maybe something like... WDYT?
```
if (strings.EqualFold(tier, string(network.ApplicationGatewayTierStandardV2)) || strings.EqualFold(tier, string(network.ApplicationGatewayTierWAFV2))) && (capacity.(int) < 1 || capacity.(int) > 125) {
	return fmt.Errorf("The value '%d' exceeds the maximum capacity allowed for a %q V2 SKU, the %q SKU must have a capacity value between 1 and 125", capacity, tier, tier)
} else if (capacity.(int) < 1 || capacity.(int) > 32) {
	return fmt.Errorf("The value '%d' exceeds the maximum capacity allowed for a %q V1 SKU, the %q SKU must have a capacity value between 1 and 32", capacity, tier, tier)
}
```

azurerm/internal/services/network/resource_arm_application_gateway.go
```diff
@@ -3724,3 +3726,20 @@ func flattenApplicationGatewayCustomErrorConfigurations(input *[]network.Applica
...
+			return fmt.Errorf("The capacity of V1 SKU %q must be between 1 and 32.", tier)
+		}
+
+		if (strings.EqualFold(tier, string(network.ApplicationGatewayTierStandardV2)) || strings.EqualFold(tier, string(network.ApplicationGatewayTierWAFV2))) && (capacity.(int) < 1 || capacity.(int) > 125) {
+			return fmt.Errorf("The capacity of V2 SKU %q must be between 1 and 125.", tier)
+		}
```
WodansSon: This looks really great, but do you think we can make the error message a bit more clear? WDYT?

```suggestion
if (strings.EqualFold(tier, string(network.ApplicationGatewayTierStandardV2)) || strings.EqualFold(tier, string(network.ApplicationGatewayTierWAFV2))) && (capacity.(int) < 1 || capacity.(int) > 125) {
			return fmt.Errorf("The value '%d' exceeds the maximum capacity allowed for a %q V2 SKU, the %q SKU must have a capacity value between 1 and 125", capacity, tier, tier)
		} else if (capacity.(int) < 1 || capacity.(int) > 32) {
			return fmt.Errorf("The value '%d' exceeds the maximum capacity allowed for a %q V1 SKU, the %q SKU must have a capacity value between 1 and 32", capacity, tier, tier)
}
```

azurerm/internal/services/network/resource_arm_application_gateway.go
```diff
@@ -660,9 +661,8 @@ func resourceArmApplicationGateway() *schema.Resource {
 						},
 
 						"capacity": {
-							Type:         schema.TypeInt,
-							Optional:     true,
-							ValidateFunc: validation.IntBetween(1, 32),
+							Type:     schema.TypeInt,
```
WodansSon: The doc's say that this is a required attribute but is optional only if the `autoscale_configuration` attribute is set to `true`. Do you think we can add an additional check to the `CustomizeDiff` to check this condition? I think this may have been overlooked with the initial implementation of this resource.

azurerm/internal/services/network/resource_arm_application_gateway.go
neil-yechenwei: Fixed

azurerm/internal/services/network/resource_arm_application_gateway.go
neil-yechenwei: I think your concern is reasonable. I just add the validation in the CustomizeDiff function. Please have a look. Thanks.

neil-yechenwei:
@WodansSon , Thanks for your comments. I've updated code. Please have a look. Thanks.

WodansSon:
The one test that failed was due to a deprecated attribute as part of 2.0.
image

WodansSon:
image

PR #5923, opened 2020-02-26
azurerm_traffic_manager_profile  - support for custom_header (#2187)

### LINKED ISSUE #2187: Support "Custom Header Settings" field for Traffic Manager Endpoint
### Community Note

### Description

Recently Azure implemented support for adding custom headers to automatically for Traffic Manager endpoint. This functionally is currently missing from the provider

Traffic Manager Endpoint

* azurerm_resource_arm_traffic_manager_endpoint

jstevans:
(fixes #2187)


azurerm/internal/services/trafficmanager/resource_arm_traffic_manager_profile.go
```diff
@@ -106,6 +106,25 @@ func resourceArmTrafficManagerProfile() *schema.Resource {
...
+						"custom_headers": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"name": {
+										Type:             schema.TypeString,
+										Required:         true,
+										ValidateFunc:     validation.StringIsNotEmpty,
+										DiffSuppressFunc: suppress.CaseDifference,
```
katbyte: What is the reason for ignoring case here?

azurerm/internal/services/trafficmanager/resource_arm_traffic_manager_profile.go
```diff
@@ -106,6 +106,25 @@ func resourceArmTrafficManagerProfile() *schema.Resource {
 							},
 						},
 
+						"custom_headers": {
+							Type:     schema.TypeList,
```
katbyte: As this is a map of name->values should we just use a type map here intead?

jstevans: That was my initial thought, since that's how JS does it with the [Headers object constructor][mdn-headers], but

1. a request is technically [allowed to have multiple headers with the same name][stackoverflow-multiple-headers] -- although it's equivalent to a comma-separated list, it can sometimes be nicer (e.g. with long entries) to have separate headers
2. the Azure REST interface uses a list of key-value pairs, and my impression of this codebase was that it tries to cleave closely to Azure where there isn't an opportunity for major simplification.


[mdn-headers]:https://developer.mozilla.org/en-US/docs/Web/API/Headers/Headers
[stackoverflow-multiple-headers]:https://stackoverflow.com/questions/4371328/are-duplicate-http-response-headers-acceptable

azurerm/internal/services/trafficmanager/resource_arm_traffic_manager_profile.go
jstevans: I believe [per standard][rfc-7230] that header names are case-insensitive. Not sure if there's a newer standard, please let me know if there's a reason this should be case sensitive :slightly_smiling_face: 

[rfc-7230]:https://tools.ietf.org/html/rfc7230#section-3.2

azurerm/internal/services/trafficmanager/resource_arm_traffic_manager_profile.go
jstevans: Actually, just realized `azurerm_traffic_manager_endpoint` already implements `custom_header`. Now that I've fixed pluralization, this implementation looks consistent with that one, and both seem consistent with their respective Azure REST equivalents :slightly_smiling_face:

jstevans:
@katbyte due to covid19 I'm unable to get to the computer that has a working `terraform-provider-azurerm` repo, and I can't manually figure out what's wrong with the formatting of this code. The `tflint` job is failing, could I ask you to please run `terrafmt` for me?

jstevans:
@katbyte You rock! 🙌 How do we go about getting this merged?

PR #5931, opened 2020-02-27
Add tags to cluster node pools

WYarde:
I'm excited to present my first attempt to contribute back to Terraform. Thanks in advance for guiding me to get this PR in good shape to get merged.

**Goal of PR**
Add tags to AKS nodepools

**Changes**
- Implemented tags for nodepools resource and datasource
- Extended some of the tests
- Updated documentation

**Attention points**
- ~~Although the change itself is small, this PR ended up touching almost 1K files because of the second commit where I have added the new containerservice sdk to modules.txt (required for nodepool tags), then ran `go mod vendor`. Besides the new files I expected, it updated many files. Is this expected behavior? Should I approach this differently?~~
- ~~I had to resolve compiler errors due to casing of `Web[H]ookProfiles` in `azurerm/internal/services/monitor/resource_arm_monitor_metric_alert.go`. Is it OK to piggy-back this fix?~~
- I tried to run acceptance tests, but even running just the containerservice ones takes 1+ hour. What is the recommended approach to validate and test?


mbfrahry:
Hey @WYarde! Thanks for opening this PR! I noticed that the tags attribute is in all of the various places you're adding to in the version of the sdk we're using. Do some of them not work for you? Not upgrading the sdk would make getting this change in much easier.

WYarde:
@mbfrahry Thank you for looking into this. 

That's a great point! I missed the nuance between API version and SDK version. Indeed, I 'just' need an API version bump which fortunately drastically reduces the size of this PR 💃 .

Let me know your thoughts.

WYarde:
@mbfrahry Just to clarify to avoid confusion - The current containerservice API does not support tags on nodepools. So although sdk upgrade is not needed (see my reply above), I do still need to update the API version

PR #5934, opened 2020-02-28
azurerm_storage_account - allow up to 50 tags

katbyte:


azurerm/internal/services/storage/resource_arm_storage_account.go
```diff
@@ -543,6 +533,8 @@ func resourceArmStorageAccount() *schema.Resource {
 				Computed:  true,
 				Sensitive: true,
 			},
+
+			"tags": tags.Schema(),
```
jackofallops: Storage account tags are still more limited (length) than other resources, we might want to add a specific schema to the tags package that still uses `validateAzureRMStorageAccountTags` and update that to the 50 count?

katbyte: whoops, good catch! thx

PR #6015, opened 2020-03-06
`azurerm_frontdoor`: Upcoming RP change forces all new front door resources to be created in the `Global` location

WodansSon:


WodansSon:
image

azurerm/internal/services/frontdoor/resource_arm_frontdoor.go
```diff
@@ -66,7 +67,12 @@ func resourceArmFrontDoor() *schema.Resource {
 				Required: true,
 			},
 
-			"location": azure.SchemaLocation(),
+			"location": {
+				Type:       schema.TypeString,
+				Optional:   true,
+				Computed:   true,
+				Deprecated: "per resource provider change, 'location' is no longer valid and will always be set to 'Global', however if the Front Door service was created prior to the v2.1.0 release of the provider it may continue to exist in its current location",
```
katbyte: could we change this to
```suggestion
				Deprecated: "Due to the service's API changing 'location' must now always be set to 'Global' for new resources, however if the Front Door service was created prior 2020/03/10 it may continue to exist in a specific current location",
```

azurerm/internal/services/frontdoor/resource_arm_frontdoor.go
```diff
@@ -481,16 +487,37 @@ func resourceArmFrontDoorCreateUpdate(d *schema.ResourceData, meta interface{})
...
+			return fmt.Errorf("locating Front Door %q (Resource Group %q): %+v", name, resourceGroup, err)
+		}
+	} else {
+		preExists = true
+		location = azure.NormalizeLocation(*exists.Location)
+	}
+
+	if hasLocation && preExists {
+		if location != azure.NormalizeLocation(cfgLocation) {
+			return fmt.Errorf("the Front Door %q (Resource Group %q) already exists in %q and cannot be moved to the %q location", name, resourceGroup, location, cfgLocation)
```
katbyte: Should we state that the location is "x" and by placing that back in config it'll work?

WodansSon:
image

PR #6020, opened 2020-03-06
Cannot set max_count and min_count to zero when autoscaling is false

SvenHamers:
when creating a some locic in the node pool it is not possible to set the value to 0 

Example:
    min_count = var.default_autoscaling_enabled ? var.default_min_node_count : 0
    max_count = var.default_autoscaling_enabled ? var.default_min_node_count: 0

The output of the following says it should be possible

else if minCount > 0 || maxCount > 0 {
		return fmt.Errorf("`max_count` and `min_count` must be set to `0` when enable_auto_scaling is set to `false`")


wagnst:
Hi guys.

I think this might have broken something in the `default_node_pool` if this code is also used there (I am not sure). When I update an existing AKS cluster with the new provider 2.1.0 I get:

```
Error: Error updating Default Node Pool "lhg-weur-kfkadevmdw-aks" (Resource Group "lhg-weur-kfkadevmdwaks-rg"): containerservice.AgentPoolsClient#CreateOrUpdate: Invalid input: autorest/validation: validation failed: parameter=parameters.ManagedClusterAgentPoolProfileProperties.Count constraint=Null value=(*int32)(nil) details: value can not be null; required parameter
```

The code bascially is 

```hcl
...
default_node_pool {
    ...
    node_count          = 1
    enable_auto_scaling = false
    ...
}
...
```

I explicitly do NOT set `min_count` and `max_count` at all in that case (autoscaling = false). But when I check the state this is set (again), although if I remove the fields manually from it:

```
    default_node_pool {
        availability_zones    = [
            "1",
        ]
        enable_auto_scaling   = false
        enable_node_public_ip = false
        max_count             = 0
        max_pods              = 250
        min_count             = 0
        name                  = "default"
        node_count            = 1
        node_labels           = {}
        node_taints           = []
        os_disk_size_gb       = 128
        tags                  = {}
        type                  = "VirtualMachineScaleSets"
        vm_size               = "Standard_D2s_v3"
        vnet_subnet_id        = "/subscriptions/xxx/resourceGroups/yyy/providers/Microsoft.Network/virtualNetworks/zzz/subnets/aaa"
    }

```

PR #6067, opened 2020-03-11
Added loadbalancer name-addresses map attribute

tpatzig:
This adds a new attribute to get a name-private_ip_address mapping
of the frontend_ip_configurations, defined on the loadbalancer.
Helpful, if you have manny frontend_ip_configurations, created with a
dynamic loop and need these mapping for further steps or outputs,
otherwise you just rely on the list order.

### DIFF
--- a/azurerm/internal/services/network/resource_arm_lb.go
+++ b/azurerm/internal/services/network/resource_arm_lb.go
@@ -164,6 +164,14 @@ func resourceArmLoadBalancer() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"names_private_ip_addresses": {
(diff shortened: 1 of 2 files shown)


jackofallops:
Hi @tpatzig 
Thanks for this PR.
In review the (potentially) dynamic nature of this information makes it feel like this shouldn't be on the resource.  It's possibly appropriate to have it on the Data Source instead to be able to look it up when it's needed? (on the resource this information will potentially be stale if there has been no configuration change to have updated the state, leading to unpredictable behaviours.). I'm going to close this PR for now, but if you'd like to rework this into the Data Source, we can revisit the review.
Thanks!

Issue #6076, opened 2020-03-11
Support for configuring "Daily Usage Quota" in azurerm_function_app


danielwertheim:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When using the consumption plan for an Azure Function there's a setting to set the "_Daily Usage Quota in GB-s_". Would be great if this could be configured via e.g `daily_usage_quota`

![image](https://user-images.githubusercontent.com/196044/76460830-0de21a80-63df-11ea-971a-8a92e342e819.png)

### New or Affected Resource(s)

* azurerm_function_app

### Potential Terraform Configuration

```hcl
resource "azurerm_function_app" "example" {
  name                      = "test-azure-functions"
  location                  = azurerm_resource_group.example.location
  resource_group_name       = azurerm_resource_group.example.name
  app_service_plan_id       = azurerm_app_service_plan.example.id
  storage_connection_string = azurerm_storage_account.example.primary_connection_string

  daily_usage_quota         = 10000
}
```

### References

* https://docs.microsoft.com/en-us/sandbox/functions-recipes/portal#controlling-how-much-your-function-can-run-in-one-day

danielwertheim:
If there's a workaround, e.g. running a "script" or similar (relatively new to Terraform) feel free to suggest how this can be achieved.

danielwertheim:
This worked fine.

```hcl
resource "null_resource" "test" {
  provisioner "local-exec" {
    command = "az functionapp update -g ${azurerm_resource_group.example.name} -n ${azurerm_function_app.example.name} --set dailyMemoryTimeQuota=5000"
  }
}
```

aqche:
I can look into getting a PR open for this.

PR #6100, opened 2020-03-12
`azurerm_function_app` - support for `daily_memory_time_quota`

aqche:
Fixes: #6076

Adds the `daily_memory_time_quota` option to the `azurerm_function_app` resource.

```
=== RUN   TestAccAzureRMFunctionApp_dailyMemoryTimeQuota
=== PAUSE TestAccAzureRMFunctionApp_dailyMemoryTimeQuota
=== CONT  TestAccAzureRMFunctionApp_dailyMemoryTimeQuota
--- PASS: TestAccAzureRMFunctionApp_dailyMemoryTimeQuota (200.69s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/web/tests   200.725s
```


azurerm/internal/services/web/resource_arm_function_app.go
```diff
@@ -174,6 +174,11 @@ func resourceArmFunctionApp() *schema.Resource {
 				Default:  false,
 			},
 
+			"daily_memory_time_quota": {
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
```
aqche: looks like if the value of `DailyMemoryTimeQuota` is set to 0 or a negative value, then the quota will be turned off. I know `schema.TypeInt` will default to a value of 0 if not provided. Do we want to explicitly set a default value here?

tombuildsstuff: I think that should be fine - since the quota itself only affects functions inside a consumption plan?

tombuildsstuff:
Ignoring a couple of known test errors (and the linting errors fixed in #6118) this otherwise LGTM:

Screenshot 2020-03-16 at 10 22 45

PR #6125, opened 2020-03-16
azurerm_lb: allow ipv6 addresses for the `private_ip_address` property

fabianofranz:
The azure load balancer resource's frontend_ip_configuration argument assumes
the IP address is only going to be IPv4. This removes that restriction to be
in line with the rest of the terraform code that does not have that validation.


fabianofranz:
> Is the issue that the ip address can be ipv4 or ipv6? if that is the case could we change the validation to accept either rather then any string?

Done, changed to either accept IPv4, IPv6, or empty.

fabianofranz:
@katbyte ^

fabianofranz:
@katbyte looks good, thanks!

PR #6105, opened 2020-03-13
azurerm_application_gateway  - support WAF Policies

sirlatrom:
Fixes #4667.

PR #5004, opened 2019-11-27
[Networking] Updating to `2019-09-01` / fixing `azurerm_virtual_hub` / VPN Resources

tombuildsstuff:
We ultimately ended up pulling the `azurerm_virtual_hub` Data Source & Resource from v1.37 since the resource design didn't match how the API worked. As such this PR reintroduces the `azurerm_virtual_hub` resource by introducing the dependencies necessary to use it.

Whilst it's unfortunate this PR is so large - due to the fact that this is already merged it makes sense to bundle these together - however traditionally we'd split these apart, in particular the SDK upgrade.

This PR includes:

* updating the `network` API to `2019-09-01`
* re-introducing the `azurerm_virtual_hub` Data Source / Resources:
    - removing the `express_route_gateway_id` field from the Data Source & Resource
    - removing the `p2s_vpn_gateway_id` field from the Data Source & Resource
    - removing the `s2s_vpn_gateway_id` field from the Data Source & Resource
    - removing the `route` block from the Data Source
    - removing the `virtual_network_connection` block from the Data Source & Resource
* removing the `security_provider_name` field from the `azurerm_virtual_wan` resource since this is no longer available in the API
* New Resource: `azurerm_point_to_site_vpn_gateway`
* New Resource: `azurerm_vpn_gateway`
* New Resource: `azurerm_vpn_server_configuration`

---

## Acceptance Test Results:

Data Source: `azurerm_virtual_hub`:

Screenshot 2019-11-27 at 16 15 49

Data Source: `azurerm_virtual_wan`:

Screenshot 2019-11-27 at 11 47 54

Resource: `azurerm_point_to_site_vpn_gateway`:

Screenshot 2019-11-28 at 17 44 15

Resource: `azurerm_virtual_hub`:

Screenshot 2019-11-27 at 11 47 33

Resource: `azurerm_virtual_hub_connection`:

Screenshot 2019-11-28 at 14 23 24

Resource: `azurerm_vpn_gateway`:

Screenshot 2019-11-28 at 14 23 24

Resource: `azurerm_vpn_server_configuration`:

Screenshot 2019-11-27 at 11 47 06

Resource: `azurerm_virtual_wan`:

Screenshot 2019-11-27 at 12 10 53

PR #6126, opened 2020-03-16
azurerm_web_application_firewall_policy - support managed_rules

### LINKED ISSUE #5354: Unable to create custom Azure Application Gateway WAF rule
### Community Note

Terraform v0.12.19
+ provider.azurerm v1.40.0

### Affected Resource(s)

* ` azurerm_web_application_firewall_policy`

The below code is a subset from the example usage given on the doc page at https://www.terraform.io/docs/providers/azurerm/r/web_application_firewall_policy.html  I receive the same error using the full example from that page.

```hcl
  name     = "example-rg"
  location = "West US 2"
}

resource "azurerm_web_application_firewall_policy" "example" {
  name                = "example-wafpolicy"

  custom_rules {
    name      = "Rule1"
    priority  = 1
    rule_type = "MatchRule"

    match_conditions {
      match_variables {
        variable_name = "RemoteAddr"
      }

      operator           = "IPMatch"
      negation_condition = false
      match_values       = ["192.168.1.0/24", "10.0.0.0/24"]
    }

    action = "Block"
  }
}
```

### Debug Output

https://gist.github.com/darrenk13/ff308c5a34378cac9193ba6dc0b5168f

### Expected Behavior

Custom Application Gateway WAF rule should have been created.

### Actual Behavior

I received an error from Azure and the resource was not created.
```
Error: Error creating Web Application Firewall Policy "example-wafpolicy" (Resource Group "example-rg"): network.WebApplicationFirewallPoliciesClient#CreateOrUpdate: Invalid input: autorest/validation: validation failed: parameter=parameters.WebApplicationFirewallPolicyPropertiesFormat.ManagedRules constraint=Null value=(*network.ManagedRulesDefinition)(nil) details: value can not be null; required parameter

  on main.tf line 6, in resource "azurerm_web_application_firewall_policy" "example":
   6: resource "azurerm_web_application_firewall_policy" "example" {
```

### Steps to Reproduce

1. `terraform apply`

### References

*  https://www.terraform.io/docs/providers/azurerm/r/web_application_firewall_policy.html

### LINKED ISSUE #5727: Support for WAF Policy Managed Rule definitions
### Community Note

### Description

In when configuring the WAF policy. In addition to defining custom rules I want to be able to defined which predefined/managed rules are active on the policy. This is supported by ARM right now and I assume az cli too.

* azurerm_web_application_firewall_policy

```hcl
resource "azurerm_web_application_firewall_policy" "example" {
  name                = "example-wafpolicy"

  managedRuleSet {
    ruleSetType    = "OWASP"
    ruleSetVersion = "3.1"
    
    ruleGroupOverride {
      ruleGroupName = "REQUEST-920-PROTOCOL-ENFORCEMENT"
      disabledRules = [
        "920300",
        "920440"
      ]
    }

    ruleGroupOverride {
      ruleGroupName = "REQUEST-942-APPLICATION-ATTACK-SQLI"
      disabledRules = [
        "942440"
      ]
    }
  }
}
```

sirlatrom:
Adds `managed_rules` to `azurerm_web_application_firewall_policy`. Until this commit, at least since SDK version 2019-09-01 (as imported in the touched file), it was not possible to create a new resource of that type, as reported in #5354.

~Note: Also streamline singular plurality for optional blocks allowing multiple repetitions to comply with how other `azurerm_` resources use singular for block names when there may be multiple specified.~

Fixes #5727.
Fixes #5354.


azurerm/internal/services/network/resource_arm_web_application_firewall_policy.go
```diff
@@ -402,12 +577,102 @@ func flattenArmWebApplicationFirewallPolicyPolicySettings(input *network.PolicyS
 
 	result := make(map[string]interface{})
 
-	result["enabled"] = input.State == network.WebApplicationFirewallEnabledStateDisabled
+	result["enabled"] = input.State == network.WebApplicationFirewallEnabledStateEnabled
```
sirlatrom: This one snuck in, and fixes a bug where `policy_settings.enabled` is set to the opposite of the actual value.

sirlatrom:
Ping @katbyte @tombuildsstuff, this is a prerequisite for #6105 since the test in that one requires successfully creating an `azurerm_web_application_firewall_policy`.

azurerm/internal/services/network/resource_arm_web_application_firewall_policy.go
```diff
@@ -143,6 +143,87 @@ func resourceArmWebApplicationFirewallPolicy() *schema.Resource {
...
+									},
+								},
+							},
+						},
+						"managed_rule_set": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"rule_set_type": {
```
katbyte: I think we can just use type as rule set is implied?
```suggestion
									"type": {
```

azurerm/internal/services/network/resource_arm_web_application_firewall_policy.go
```diff
@@ -143,6 +143,87 @@ func resourceArmWebApplicationFirewallPolicy() *schema.Resource {
...
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.NoZeroValues,
+									},
+									"rule_set_version": {
```
katbyte:
```suggestion
									"version": {
```

azurerm/internal/services/network/resource_arm_web_application_firewall_policy.go
```diff
@@ -332,11 +418,102 @@ func expandArmWebApplicationFirewallPolicyPolicySettings(input []interface{}) *n
...
+func expandArmWebApplicationFirewallPolicyManagedRulesDefinition(input []interface{}) *network.ManagedRulesDefinition {
+	if len(input) == 0 {
+		return nil
+	}
+	v := input[0].(map[string]interface{})
+
+	exclusions := v["exclusion"].([]interface{})
+	managedRuleSets := v["managed_rule_set"].([]interface{})
+
+	result := network.ManagedRulesDefinition{
```
katbyte: We could just return the struct?
```suggestion
	return &network.ManagedRulesDefinition{
```

website/docs/r/web_application_firewall_policy.html.markdown
```diff
@@ -130,6 +163,42 @@ The `policy_setting` block supports the following:
...
+
+---
+
+The `exclusion` block supports the following:
+
+* `match_variable` - (Required) The name of the Match Variable.
+
+* `selector` - (Optional) Describes field of the matchVariable collection.
+
+* `selector_match_operator` - (Required) Describes operator to be matched.
```
katbyte: COuld we list all possible values here?

website/docs/r/web_application_firewall_policy.html.markdown
```diff
@@ -130,6 +163,42 @@ The `policy_setting` block supports the following:
...
+
+* `exclusion` - (Optional) One or more `exclusion` block defined below.
+
+* `managed_rules_set` - (Optional) One or more `managed_rules_set` block defined below.
+
+---
+
+The `exclusion` block supports the following:
+
+* `match_variable` - (Required) The name of the Match Variable.
```
katbyte: COuld we list all possible values here?

azurerm/internal/services/network/resource_arm_web_application_firewall_policy.go
```diff
@@ -48,7 +48,7 @@ func resourceArmWebApplicationFirewallPolicy() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
 
-			"custom_rules": {
+			"custom_rule": {
```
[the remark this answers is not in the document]
sirlatrom: There can be no existing users of the resource as it does not work, as described in #5354. I thought that before making the resource actually work, this would be a good chance to align naming with other resources. The existing docs even mention a `custom_rule` block:

image

The same goes for `match_condition` and `match_variable`.

A single block with a plural name would indicate that the block configures all such elements, e.g. custom rules, whereas multiple blocks with a singular name indicate that they each configure such an element.

website/docs/r/web_application_firewall_policy.html.markdown
sirlatrom: Done in 9fa3f6ea0.

website/docs/r/web_application_firewall_policy.html.markdown
sirlatrom: Done in 9fa3f6ea0.

azurerm/internal/services/network/resource_arm_web_application_firewall_policy.go
sirlatrom: Done in 9fa3f6ea0.

azurerm/internal/services/network/resource_arm_web_application_firewall_policy.go
sirlatrom: Done in 9fa3f6ea0.

azurerm/internal/services/network/resource_arm_web_application_firewall_policy.go
sirlatrom: Done in 9fa3f6ea0.

azurerm/internal/services/network/resource_arm_web_application_firewall_policy.go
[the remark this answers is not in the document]
sirlatrom: I see now that the `ManagedRules` field was introduced in #5004 (2e69260f3a) between dependency `github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-07-01/network` and `github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-09-01/network`, but the `azurerm_web_application_firewall_policy` resource was not updated to reflect this new required field.

This of course means that there will be existing users of this resource that have used the `custom_rules` name for the property before the `ManagedRules` field was introduced, so I suppose I'll revert this change after all, and update the docs to all use the actual property names instead of the singular forms.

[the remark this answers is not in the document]
sirlatrom: Fixed in 7aeb41b5f.

sirlatrom:
> Thanks for making those changes @sirlatrom. I just noticed there is not test for the new properties, could we set them in the complete test? once that's done this should be good to merge.

This should be done in cefd1c8.

sirlatrom:
> 
> 
> @sirlatrom,
> 
> we can't add a new required property, one it clearly works without it 🙂 and it would break the resource for existing users. As can be see from the tests:
> 
> ```
> Test Failed
> 
> ------- Stdout: -------
> === RUN   TestAccAzureRMWebApplicationFirewallPolicy_basic
> === PAUSE TestAccAzureRMWebApplicationFirewallPolicy_basic
> === CONT  TestAccAzureRMWebApplicationFirewallPolicy_basic
> --- FAIL: TestAccAzureRMWebApplicationFirewallPolicy_basic (0.76s)
>     testing.go:640: Step 0 error: config is invalid: "managed_rules": required field is not set
> ```

The property is required on the ARM side since API version 2019-08-01, but was not implemented when this resource was switched to that version. I'll try and add a reasonable default value.

sirlatrom:
@katbyte When providing an empty ManagedRules object I get this API error from Azure, so I think it really must be specified:

> ```
> Error: Error creating Web Application Firewall Policy "foobarPolicy" (Resource Group "rgFooBar"): network.WebApplicationFirewallPoliciesClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="ApplicationGatewayFirewallManagedRuleSetsNoValidCrsRuleSetsAttached" Message="Firewall policy /subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/rgFooBar/providers/Microsoft.Network/ApplicationGatewayWebApplicationFirewallPolicies/foobarPolicy does not have any valid Core Rule Set attached to it." Details=[]
> ```

the essence of which is: **foobarPolicy does not have any valid Core Rule Set attached to it**.

It's all caused by #5004 which updated the SDK version from 2019-07-01 to 2019-09-01 without implementing the required field. See here: https://github.com/terraform-providers/terraform-provider-azurerm/blame/91d54d8ef85c82b2aa38723a12f5709dd9ef1abe/azurerm/resource_arm_web_application_firewall_policy.go#L8.

Anyone with a provider version prior to this change will not be affected. Anyone with the current version and later are currently not able to create firewall policy resources, and will need this fix to be able to do so again. Given how both the SDK and the API say the field is required and cannot just be created without properly configuring it, I would claim it actually _does_ have to be required.

If we have to fill out default values for the ManagedRules field, it has to contain a choice of a managed rule set, which I don't think either of us are able to choose without breaking other things for the end users. Hence, I believe they should choose one themselves when upgrading to the provider version that adds this field.

katbyte:
@sirlatrom, thanks for investigating this. I agree with everything you said and looks like we just have to mark it required and move forward as checking in our CI systems i can see the error and that this has been broken for quite some time. Triggering CI and taking a look now

sirlatrom:
> @sirlatrom, this looks good aside for we need to update the existing tests so they pass, basic with the minimal rules, complete with everything set.

@katbyte Thank you for your patience! I've added the required properties to the tests in 2f15e45, and corrected the `managed_rules.managed_rule_set`'s `Required` to `true` as it is indeed required. I've also adjusted #6105 accordingly. I had also made a typo when referring to the `managed_rule_set` property in the tests.

---

One last thing I saw was that in the `waf_configuration` block, the `rule_set_type` has a default value of `OWASP` and the `rule_set_version` and `disabled_rule_group.rule_group_name` properties have `StringInSlice` validation functions. The default type is `OWASP` and the validation functions include different versions and rule group names across the possible types.

I'm not sure if those defaults and lists of valid values will always be up to date, so while I could add the same ones to the equivalent properties in the `managed_rules.managed_rule_set.rule_group_override.rule_group_name`/`disabled_rules` properties; but I would both have to duplicate those lists of valid values, resulting in double maintenance and the risk of drift, and additionally, those lists may grow out of date resulting in the potential need of the provider needing a new release _just_ to support a new rule set type or version or disabled rule.

WDYT?

sirlatrom:
@katbyte Looks like I have more work to do then. I imagine the tests fail because I forgot to access `managed_rules` as a list, and `policy_settings` I'm not sure about. I'll find time to look at it tomorrow CEST.

sirlatrom:
@katbyte OK, it should be done in 6c30ebdbb now, but I don't have access to run acceptance tests in my current environment.

sirlatrom:
PS: 6731c99 includes the validation functions you suggested in https://github.com/terraform-providers/terraform-provider-azurerm/pull/6126#pullrequestreview-385989469.

sirlatrom:
@katbyte I've tried creating a policy on our subscription, and it comes back with these values when running `az network application-gateway waf-policy show ...`:

```json
...
  "policySettings": {
    "fileUploadLimitInMb": 100,
    "maxRequestBodySizeInKb": 128,
    "mode": "Prevention",
    "requestBodyCheck": true,
    "state": "Enabled"
  },
...
```

So I'm able to set the fields using the patched provider. When I run plan the next time, I'm told there are no changes to be made, so that should indicate the settings are read properly, too. We might be dealing with the defaults, here, which are set to `"mode": "Detection"` and `"state": "Enabled"`.

sirlatrom:
@katbyte One option would be to add the `policy_settings` block to the basic config in the test, even if it is not required.

sirlatrom:
The test that fails looks unrelated.

sirlatrom:
@katbyte Tests pass after rebase on origin/master which includes #6540. Please let me know if acceptance tests still fail after I added `policy_settings` to the basic test.

PR #6127, opened 2020-03-16
azurerm_servicebus_subscription : Fixed default_message_ttl format in docs

rajrao:
Its no longer based on Timespan, but needs to be ISO8601 format.

### DIFF
--- a/website/docs/r/servicebus_subscription.html.markdown
+++ b/website/docs/r/servicebus_subscription.html.markdown
@@ -69,7 +69,7 @@ The following arguments are supported:
     [TimeSpan](#timespan-format) format.
 
 * `default_message_ttl` - (Optional) The TTL of messages sent to this Subscription
-    if no TTL value is set on the message itself. Provided in the [TimeSpan](#timespan-format)
+    if no TTL value is set on the message itself. Provided in the ISO8601 [TimeSpan](#timespan-format)
     format.
 
 * `lock_duration` - (Optional) The lock duration for the subscription, maximum
@@ -92,7 +92,7 @@ The following arguments are supported:
 
 ### TimeSpan Format
 
-Some arguments for this resource are required in the TimeSpan format which is used to represent a length of time. The supported format is documented [here](https://msdn.microsoft.com/en-us/library/se73z7b9(v=vs.110).aspx#Anchor_2)
+Some arguments for this resource are required in the TimeSpan format which is used to represent a length of time. The supported format is documented [here](https://docs.microsoft.com/en-us/cli/azure/servicebus/topic/subscription?view=azure-cli-latest#az-servicebus-topic-subscription-update). Example value: "P3DT0H0M0.0S" (for 3days)
 
 ## Attributes Reference
 
--- a/website/docs/r/servicebus_subscription.html.markdown
+++ b/website/docs/r/servicebus_subscription.html.markdown
@@ -65,15 +65,13 @@ The following arguments are supported:
 * `max_delivery_count` - (Required) The maximum number of deliveries.
 
 * `auto_delete_on_idle` - (Optional) The idle interval after which the
-    Subscription is automatically deleted, minimum of 5 minutes. Provided in the
-    [TimeSpan](#timespan-format) format.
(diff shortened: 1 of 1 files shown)


website/docs/r/servicebus_subscription.html.markdown
```diff
@@ -69,7 +69,7 @@ The following arguments are supported:
     [TimeSpan](#timespan-format) format.
```
magodo: We shall also update the doc for this attribute.

website/docs/r/servicebus_subscription.html.markdown
```diff
@@ -69,7 +69,7 @@ The following arguments are supported:
...
 
 * `default_message_ttl` - (Optional) The TTL of messages sent to this Subscription
-    if no TTL value is set on the message itself. Provided in the [TimeSpan](#timespan-format)
+    if no TTL value is set on the message itself. Provided in the ISO8601 [TimeSpan](#timespan-format)
```
magodo: It is not necessary to provide a link for ISO8601 since it is well-knowned format. So we can replace the wording a bit:
```suggestion
    in ISO8601 format, if no TTL value is set on the message itself.
```

website/docs/r/servicebus_subscription.html.markdown
```diff
@@ -92,7 +92,7 @@ The following arguments are supported:
 
 ### TimeSpan Format
 
-Some arguments for this resource are required in the TimeSpan format which is used to represent a length of time. The supported format is documented [here](https://msdn.microsoft.com/en-us/library/se73z7b9(v=vs.110).aspx#Anchor_2)
```
magodo: Just remove this "TimeSpan Format" as ISO8601 is a well-known format.
(PS: we normally add link to azure rest api instead of azure cli/azure pwsh as a reference, the link for this case is [this](https://docs.microsoft.com/en-us/rest/api/servicebus/subscriptions/createorupdate#request-body))

website/docs/r/servicebus_subscription.html.markdown
```diff
@@ -92,7 +92,7 @@ The following arguments are supported:
...
+Some arguments for this resource are required in the TimeSpan format which is used to represent a length of time. The supported format is documented [here](https://docs.microsoft.com/en-us/cli/azure/servicebus/topic/subscription?view=azure-cli-latest#az-servicebus-topic-subscription-update). Example value: "P3DT0H0M0.0S" (for 3days)
```
WodansSon: When adding a link to the docs site, can we remove the locale from the URL to allow the site to auto-detect the users locale?

```suggestion
Some arguments for this resource are required in the TimeSpan format which is used to represent a length of time. The supported format is documented [here](https://docs.microsoft.com/cli/azure/servicebus/topic/subscription?view=azure-cli-latest#az-servicebus-topic-subscription-update). Example value: "P3DT0H0M0.0S" (for 3days)
```

website/docs/r/servicebus_subscription.html.markdown
rajrao: I think its nice to have the ISO8601 format link, as we can over time add more examples. Also, it allows us to link to the actual documentation. Hope you keep it.

magodo: Of course, then we can follow the way as other resource did (e.g. [iothub](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/docs/r/iothub.html.markdown)), add the link towards wikipedia for ISO8601. The point is that we'd better keep things consistent.

katbyte: I'm inclined to agree, it's easy enough to link to public examples like iothub for thsoe who need the examples while keeping the docs clean,

rajrao: So take out link and examples?

rajrao: Made the documentation consistent with Topic and Queue documentation. Also, please note, that it looks like your provider does not support duplicateDetectionHistoryTimeWindow (https://docs.microsoft.com/en-us/rest/api/servicebus/subscriptions/createorupdate). Let me know and I will add a separate issue for that.

PR #6137, opened 2020-03-17
Doc Fix: `azurerm_application_gateway`

sebader:
In the `backend_http_settings` the field `probe_name` is only optional. The full example in the beginning is already correct and does not include a custom probe.


mbfrahry:
Thanks for this @sebader!

PR #6140, opened 2020-03-17
arm_public_ip: public_ips can be dynamic for ipv6 addresses

fabianofranz:
This limitation has been relaxed for standard load balancers.

Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/6090.

### DIFF
--- a/azurerm/internal/services/network/tests/resource_arm_public_ip_test.go
+++ b/azurerm/internal/services/network/tests/resource_arm_public_ip_test.go
@@ -112,7 +112,7 @@ func TestAccAzureRMPublicIpStatic_basic_withDNSLabel(t *testing.T) {
 	})
 }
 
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/network/tests/resource_arm_public_ip_test.go
```diff
@@ -112,22 +112,6 @@ func TestAccAzureRMPublicIpStatic_basic_withDNSLabel(t *testing.T) {
 	})
 }
 
-func TestAccAzureRMPublicIpStatic_standard_withIPv6_fails(t *testing.T) {
```
katbyte: Could we invert this test so that it is checks that it succeeds?

fabianofranz: @katbyte done, PTAL!

fabianofranz:
Comments addressed.

PR #6141, opened 2020-03-17
Doc Fix: `azurerm_backup_protected_file_share`

pearcec:
This example is broken.


mbfrahry:
Thanks for this @pearcec!

PR #6142, opened 2020-03-17
virtualnetwork: allow address_prefix to be null

fabianofranz:
The azure subnet resource allows for multiple address prefixes using the
address_prefixes argument. The virtual network resource allows inline subnets,
but does not currently support this argument. When the address_prefixes argument
is used, address_prefix will be null when terraform refreshes its state causing
terraform to panic. This code updates the virtual network resource to allow
address_prefixes to be used and for address_prefix to be null.

Tests have been added to test the multiple prefixes. The magic number in the
new tests is the same magic number used in the other VNet tests. I'm not 100%
clear on this number, but it appears to be a known ID. It's origin is
documented (not very well) here:

https://github.com/terraform-providers/terraform-provider-azurerm/pull/1913

This fix addresses several IPv6 related bugs.

### DIFF
--- a/azurerm/internal/services/network/resource_arm_virtual_network.go
+++ b/azurerm/internal/services/network/resource_arm_virtual_network.go
@@ -106,9 +106,15 @@ func resourceArmVirtualNetwork() *schema.Resource {
 						},
 
 						"address_prefix": {
-							Type:         schema.TypeString,
-							Required:     true,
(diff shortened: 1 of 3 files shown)


fabianofranz:
Closed in favor of https://github.com/terraform-providers/terraform-provider-azurerm/pull/6148.

Issue #2977, opened 2019-02-28
Cannot create azurerm_storage_container in azurerm_storage_account that uses network_rules


phil-bevan:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
Terraform v0.11.11
+ provider.azurerm v1.21.0

### Affected Resource(s)

* `azurerm_storage_account`
* `azurerm_storage_container`

### Terraform Configuration Files

```
resource "azurerm_storage_account" "test-storage-acct" {
  name                     = "${var.prefix}storacct"
  resource_group_name      = "${var.resgroup}"
  location                 = "${var.location}"
  account_tier             = "Standard"
  account_replication_type = "LRS"
  network_rules {
    ip_rules                   = ["aaa.bbb.ccc.ddd/ee"]
    virtual_network_subnet_ids = ["${var.subnetid}"]
  }
}
resource "azurerm_storage_container" "provisioning" {
  name                  = "${var.prefix}-provisioning"
  resource_group_name   = "${var.resgroup}"
  storage_account_name  = "${azurerm_storage_account.test-storage-acct.name}"
  container_access_type = "private"
}
```

### Debug Output

* azurerm_storage_container.provisioning: Error creating container "philtesting1-provisioning" in storage account "philtesting1storacct": storage: service returned error: StatusCode=403, ErrorCode=AuthorizationFailure, ErrorMessage=This request is not authorized to perform this operation.
RequestId:a7f9d2e1-701e-00b3-4e74-cf3b34000000
Time:2019-02-28T14:45:53.7885750Z, RequestInitiated=Thu, 28 Feb 2019 14:45:53 GMT, RequestId=a7f9d2e1-701e-00b3-4e74-cf3b34000000, API Version=, QueryParameterName=, QueryParameterValue=

### Expected Behavior

Container can be created in a storage account that uses network rules

### Actual Behavior

After applying a network_rule to a storage account I cannot provision a container into it.  My public IP is included in the address range specified in the network rule.  I can successfully create the container via the Azure portal

### Steps to Reproduce

1. `terraform apply`

[...]

PR #6147, opened 2020-03-17
Use BlobContainersClient for blob container operations

DSakura207:
The pull request replaces custom storage blob container client `giovanni` with the one from Azure Go SDK.
Functions and tests are duplicated from `giovanni` for generating and parsing ID for a blob container, to ensure backward compatibility. Several local functions are added to container resource file to limit changes to local file.

Also `testAccAzureRMStorageContainer_basicAzureADAuth` is changed since the provider blocks are duplicated. `strings.Replace` looks dirty, so I am open to any suggestion to avoid it. 

With those effort, #2977 is possible to fix and containers could have immutability policy and legal holds.

Last but not least, thank you @tombuildsstuff for the custom storage client. It is a great work.

### DIFF
--- a/azurerm/internal/services/storage/client/client.go
+++ b/azurerm/internal/services/storage/client/client.go
@@ -29,6 +29,7 @@ type Client struct {
 	CachesClient             *storagecache.CachesClient
 	StorageTargetsClient     *storagecache.StorageTargetsClient
 	SubscriptionId           string
+	StorageUseAzureAD        bool
 
 	environment   az.Environment
 	storageAdAuth *autorest.Authorizer
@@ -72,6 +73,7 @@ func NewClient(options *common.ClientOptions) *Client {
 
 	if options.StorageUseAzureAD {
 		client.storageAdAuth = &options.StorageAuthorizer
+		client.StorageUseAzureAD = true
 	}
 
 	return &client
--- a/azurerm/internal/services/storage/client/client.go
+++ b/azurerm/internal/services/storage/client/client.go
@@ -25,6 +25,7 @@ type Client struct {
 	FileSystemsClient        *filesystems.Client
 	ManagementPoliciesClient storage.ManagementPoliciesClient
 	BlobServicesClient       storage.BlobServicesClient
+	BlobContainersClient     storage.BlobContainersClient
 	CachesClient             *storagecache.CachesClient
 	StorageTargetsClient     *storagecache.StorageTargetsClient
 	SubscriptionId           string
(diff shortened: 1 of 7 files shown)


tombuildsstuff:
hi @DSakura207 

Thanks for this PR.

Taking a look through whilst on the whole the code for this looks pretty good, unfortunately due to the way the Azure Storage API's (and Storage Permissions work) this approach isn't going to work for most users.

At this point in time Azure Storage supports two different methods for authentication - using the (Shared) Access Key and AzureAD authentication - where this PR now only supports authenticating using AzureAD authentication ([configured here](https://github.com/terraform-providers/terraform-provider-azurerm/pull/6147/files#diff-4ba1726ed3c7657eea626dfc5eaf61c2R49)).

Whilst at first glance this sounds like it should be fine - unfortunately the rollout of the new Azure Storage Permissions has been done in such a way it's caused a large amount of user confusion; such that when we previously defaulted to using AzureAD authentication for Storage and it turns out that most users don't have this permission enabled by default.

After a ton of investigation into this, it appears the way the new Storage Permissions have been rolled out is inconsistent, insofar as new Role Assignments (for say Contributor/Owner) also get assigned the new Storage Permissions - however older/existing Role Assignments at these same permissions do not. Ultimately this meant that when we defaulted to AzureAD authentication last time this caused a considerable amount of confusion amongst users since this differentiation was unclear.

Fundamentally there are 3 main differences between the Storage Data Plane & Storage Resource Manager API's of note here:

* The Resource Manager API's only support authenticating using AzureAD (and not Shared Key authorization)
* The Resource Manager API's don't support managing Blobs (amongst other functionality) - and likely won't anytime soon, since ARM has a 4MB payload limit
* The Identifiers for the Storage Resources are different between the two - for example: `https://{account}.blob.{domain}/{container}` and `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{accountName}/blobServices/default/containers/{containerName}`.

Ultimately I believe we're going to have to continue using the Giovanni SDK (since we need Shared Key authorization) for the foreseeable future. However to be able to fix #2977 requires using the Resource Manager API's - as such I believe we need to take a slightly different approach here.

Perhaps the best way to get this working would be a wrapper around Giovanni/the Resource Manager API's that conditionally calls them, but maintains the Data Plane ID's - WDYT?

Thanks!

DSakura207:
Hello @tombuildsstuff ,

Thank you for the reply. I replaced blob container client because when Terraform runs on Azure Pipelines (Microsoft's CI/CD service) hosted agent, `resourceArmStorageContainerRead` returns 403 error when used 2.1.0 provider. When I switched to `BlobContainersClient` it succeeded.

After a second thought, it maybe caused by the role assignment you said. However the fun is that it succeeded when I run it locally with the same service principal but manually created secret.

A wrapper seems to be the solution. Let me see what I can do.

DSakura207:
Added wrapper and retry, rebased onto newest master, and fixed lint errors.

Now the client does:
- When storage is not set to authenticate with Azure AD
  - Use giovanni for all actions.
  - When an action failed, try to use Azure client to do the same thing.
- When storage is set to authenticate with Azure AD
  - Only uses Azure client.

Hopefully this is okay, since I have to repeat some code for the retry operation.

DSakura207:
@tombuildsstuff can you review code please?

tombuildsstuff:
hey @DSakura207

I've recently opened #9314 which adds a shim layer to the Storage Data Plane resources (Container, File Share, Queues, Tables) - which will allow users to opt to use either the Data Plane or the Resource Manager API, depending on the permissions available.

As a part of that work I've also looked to implement the Resource Manager shim for these resources - unfortunately it appears that the Resource Manager API is missing a significant amount of functionality present in the Data Plane API.

Ultimately this means at this time only the Containers functionality can be switched over to the Resource Manager API - since this is the only one of the 4 which contains the same functionality in both API's. I've raised these issues with Microsoft and they'll be run down internally - when this is available we should be able to implement the resource manager toggle for these resources.

As such, whilst I'd like to thank you for this contribution - and apologies for the delayed review time here - I'm going to close this PR in favour of #9314, which once merged enables us to support this functionality via a Feature Toggle, meaning users can move across gradually rather than introducing a breaking permissions change here.

Thanks!

PR #6168, opened 2020-03-19
Pilot impl for azurermproviderlint

magodo:
Write a new command which will lint the rules currently used for tflint target in *GNUMakefile*, plus adding a simple potential azurerm lint rule as a pilot, which will check the format string of `fmt.Errorf()` or `errors.New()` to eliminate the use of "error" (case-insensitive), which is suggested by Jeffrey.

Note, currently this isn't used in Makefile, so it won't impact the CI. Also there is no README for this since it is just a pilot.


tombuildsstuff:
Closing this in favour of #7313 which includes these changes

PR #6172, opened 2020-03-19
travis: conditionally running the test, lint and tflint steps

tombuildsstuff:
This PR conditionally runs the Travis jobs `make lint`, `make test` and `make tflint` only if there are changes within the ./azurerm directory

Assuming no committed/staged changes to the ./azurerm folder:

```
$ TRAVIS=ci make test
Checking if this should be conditionally run..
No changes committed to ./azurerm - nothing to lint - exiting
```

When then staging a file:

```
$ git add azurerm/foo.txt

$ TRAVIS=ci make test
Checking if this should be conditionally run..
==> Running Unit Tests...
```

Ultimately this should bring the total time for a Travis build down to ~5m rather than ~1h10m when there's no changes to the `./azurerm` folder

### DIFF
--- a/scripts/run-lint.sh
+++ b/scripts/run-lint.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+function checkForConditionalRun {
+  if [ "$TRAVIS" == "ci" ];
+  then
+    echo "Checking if this should be conditionally run.."
+    result=$(git diff --name-only HEAD | grep azurerm/)
+    if [ "$result" = "" ];
+    then
+      echo "No changes committed to ./azurerm - nothing to lint - exiting"
+      exit 0
+    fi
+  fi
+}
+
+function runLinters {
+  echo "==> Checking source code against linters..."
+  golangci-lint run ./...
+}
+
+function main {
+  checkForConditionalRun
+  runLinters
+}
+
+main
--- a/scripts/run-test.sh
+++ b/scripts/run-test.sh
@@ -16,7 +16,7 @@ function checkForConditionalRun {
 function runTests {
   echo "==> Running Unit Tests..."
   go test -i $TEST || exit 1
-	go test -v $TEST "$TESTARGS" -timeout=30s -parallel=4
+  go test -v $TEST "$TESTARGS" -timeout=30s -parallel=4
 }
 
 function main {
--- a/scripts/run-test.sh
+++ b/scripts/run-test.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+function checkForConditionalRun {
+  if [ "$TRAVIS" == "ci" ];
+  then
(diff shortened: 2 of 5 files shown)


scripts/run-lint.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
```
jackofallops: Exit on error?
```suggestion
set -e

```

scripts/run-test.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
```
jackofallops: exit on error?
```suggestion
set -e

```

scripts/run-tflint.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
```
jackofallops: exit on error?
```suggestion
set -e

```

scripts/run-test.sh
```diff
@@ -0,0 +1,27 @@
...
+      echo "No changes committed to ./azurerm - nothing to lint - exiting"
+      exit 0
+    fi
+  fi
+}
+
+function runTests {
+  echo "==> Running Unit Tests..."
+  go test -i $TEST || exit 1
+	go test -v $TEST "$TESTARGS" -timeout=30s -parallel=4
```
jackofallops: globbing / splitting protection
```suggestion
	go test -v "$TEST" "$TESTARGS" -timeout=30s -parallel=4
```

scripts/run-test.sh
```diff
@@ -0,0 +1,27 @@
...
+    then
+      echo "No changes committed to ./azurerm - nothing to lint - exiting"
+      exit 0
+    fi
+  fi
+}
+
+function runTests {
+  echo "==> Running Unit Tests..."
+  go test -i $TEST || exit 1
```
jackofallops: globbing / splitting protection
```suggestion
  go test -i "$TEST" || exit 1
```

scripts/run-test.sh
tombuildsstuff: in this case TEST will always have a value, (since it's defaulted in the Gnumakefile) - whereas TESTARGS won't - so I don't think that's needed?

jackofallops: I'm a fan of `shellcheck`, so this habit is ingrained 😄- I don't think it's essential here though.

scripts/run-lint.sh
tombuildsstuff: turns out we can't, since exiting within the inner function causes an error to be returned from the outer function

scripts/run-test.sh
tombuildsstuff: in testing, apparently not? ¯\_(ツ)_/¯

PR #6175, opened 2020-03-19
azurerm_application_gateway: can combine include_path with target_url

bailsman:
Redirecting say http://foo/path?query to http://bar/
include_path controls whether it will go to
	http://bar/path
or
	http://bar/

This is a perfectly valid option in a redirect_configuration in
combination with target_url. Currently the code gives an error if you
use them in combination and omits it from the request if target_url is
set.

Omitting include_path has strange results in the Azure Portal
(no radio button selected) and "null" defaults to true whereas the
documentation says include_path should default to false.

This change makes the behavior consistent with the documentation.


bailsman:
I've only tested this with a 302 Found redirect using target_url in a redirect_configuration. I don't fully understand the purpose behind this check that include_path should not be used in combination with target_url.

Did the API change since this resource was first designed, or does this change break something in some other usage of redirect_configuration?

Hopefully someone more familiar with this code and the Application Gateway can advise.

mbfrahry:
Added a test! 
```
--- PASS: TestAccAzureRMApplicationGateway_IncludePathWithTargetURL (1752.21s)
```

PR #6179, opened 2020-03-19
Change type of ipv4_firewall_rule for Analysis Services Server

r0bnet:
We faced the problem that the order changes in Azure side when there are more than 2 rules but we don't care about the order. That's why i changed the type to a Set instead of a List. Should fix our problems here.


tombuildsstuff:
Tests look good:

Screenshot 2020-03-19 at 18 09 30

PR #6195, opened 2020-03-20
Add devcontainer and debugging launch task

lawrencegripper:
I wanted to share this as I've recently switched to using VSCode with Devcontainers as my local development environment and think it would be useful to others. 

If this is something that could be useful and merged let me know and I can add a section into docs on how to use. 

What does this PR do?

If you have VSCode, Docker and [VSCode devcontainers](https://code.visualstudio.com/docs/remote/containers) installed this PR allows you to simple clone, "reopen in container" and press F5 to start interactively debugging a test or coding a new resource. 

The golang env and tooling required to run, debug and test the provider are installed and tracked in the `Dockerfile` so users don't have to mess around getting an environment setup.


stuartleeks:
This would be amazing to have!

paultyng:
@lawrencegripper I agree we should have common debug setups in the repo, unfortunately the SDK is introducing a [new test driver](https://github.com/hashicorp/terraform-plugin-sdk/pull/262) that changes the way debug will need to work. With the binary test driver the provider code is not what the debugger will attach to in this setup, the code will be executing an external Terraform binary to call the provider and we haven't fully worked out exactly how to attach delve, etc, but its one of our last remaining tasks for v2. 

In addition, we plan to create common VSCode/Goland debug setups with the new driver and making sure this is still an option. I'll give the team a pointer to this PR to possibly add devcontainers when we do VSCode.

(Personally this is very interesting, I'm on Win10/WSL2 as my primary, so curious if the container model would simplify it).

lawrencegripper:
@paultyng thanks for getting back to me on this one. Shall I leave this one open for the time being of best to close off and revisit when new test driver is setup?

paultyng:
Lets revisit once the new driver is setup, not much actionable here. I don't see a tracking issue for the new testing framework yet, but a PR should be coming this week or next to all the official providers who aren't using it yet.

PR #6207, opened 2020-03-21
Update azurerm_managed_disk.disk_encryption_set_id without recreating the resource

kazimierzbudzyk:
Fixes: https://github.com/terraform-providers/terraform-provider-azurerm/issues/6184

CMK encryption can be enabled without recreating (requires VM to be stopped): https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disk-encryption#enable-on-an-existing-disk


kazimierzbudzyk:
Thank you for the review @katbyte! Applied requested changes, let me know if you have any other comments!

tombuildsstuff:
(Tests look fine)

PR #6218, opened 2020-03-23
azurerm_sql_database: extended_auditing_policy is for azurerm_sql_server

pallxk:
Specifying it on `azurerm_sql_database` results in:

```
Error: Unsupported block type

  on sql_server.tf line 70, in resource "azurerm_sql_database" "test":
  70:   extended_auditing_policy {

Blocks of type "extended_auditing_policy" are not expected here.
```

### DIFF
--- a/website/docs/r/sql_database.html.markdown
+++ b/website/docs/r/sql_database.html.markdown
@@ -25,6 +25,17 @@ resource "azurerm_sql_server" "example" {
   version                      = "12.0"
   administrator_login          = "4dm1n157r470r"
   administrator_login_password = "4-v3ry-53cr37-p455w0rd"
+
+  extended_auditing_policy {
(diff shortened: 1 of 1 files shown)


website/docs/r/sql_database.html.markdown
```diff
@@ -25,6 +25,17 @@ resource "azurerm_sql_server" "example" {
...
+  extended_auditing_policy {
+    storage_endpoint                        = azurerm_storage_account.example.primary_blob_endpoint
+    storage_account_access_key              = azurerm_storage_account.example.primary_access_key
+    storage_account_access_key_is_secondary = true
+    retention_in_days                       = 6
+  }
+
+  tags = {
+    environment = "production"
+  }
```
tombuildsstuff: there's no storage account in this example, so the changes above here can be removed

PR #6228, opened 2020-03-23
netapp_volume: allow up to 102400 MB for the storage_quota_in_gb property

paulomarquesc:
Fixing bug related to 4TiB volume size limit to ANF's current limit of 100TiB.


paulomarquesc:
@katbyte , the issue with CI is not related to my change.

PR #6229, opened 2020-03-23
New data source: azurerm_network_service_tags

r0bnet:
Service Tags / address prefixes are useful when using firewall rules or UDRs in some scenarios.


azurerm/internal/services/network/data_source_service_tags.go
```diff
@@ -0,0 +1,89 @@
...
+	location := d.Get("location").(string)
+	res, err := client.List(ctx, location)
+	if err != nil {
+		return err
+	}
+
+	service := d.Get("service").(string)
+	region := d.Get("region").(string)
+
+	if res.Values == nil {
```
magodo: Let's move this check just after the `List()` call. And if it is nil, can we return error to notify user?

azurerm/internal/services/network/data_source_service_tags.go
```diff
@@ -0,0 +1,89 @@
...
+			if *props.Region == region {
+				addressPrefixes = *props.AddressPrefixes
+			}
+		}
+	}
+
+	err = d.Set("address_prefixes", addressPrefixes)
+	if err != nil {
+		return fmt.Errorf("Error setting `address_prefixes`: %+v", err)
+	}
```
magodo: Can we just do following change? 

```suggestion
	for _, sti := range *res.Values {
		if sti.Name == nil || !isServiceTagOf(*sti.Name, service) {
			continue
		}

		if props := sti.Properties; props != nil {
			if *props.Region == region {
				addressPrefixes := make([]string, 0)
				if props.AddressPrefixes != nil {
					addressPrefixes = *props.AddressPrefixes
				}
				err = d.Set("address_prefixes", addressPrefixes)
				if err != nil {
					return fmt.Errorf("Error setting `address_prefixes`: %+v", err)
				}
				if sti.ID == nil {
					return errors.New("unexpected nil ID for service tag")
				}
				d.SetId(*sti.ID)
				return nil
			}
		}
	}
	return errors.New("specified service tag not found")
}

// isServiceTagOf is used to check whether a service tag name belongs to the service of name `serviceName`.
// Service tag name has format as below:
// - (regional) serviceName.locationName
// - (all) serviceName
func isServiceTagOf(stName, serviceName string) bool {
	stNameComponents := strings.Split(stName, ".")
	if len(stNameComponents) != 1 && len(stNameComponents) != 2 {
		return false
	}
	return stNameComponents[0] == serviceName
}
```

There are several changes here:

1. Instead of using `SystemService` to compare against `service` name, I prefer using the `Name` of the `ServiceTagInformation`. There are two reasons to do so:

   - This allow user to refer to the list in [doc](https://docs.microsoft.com/en-us/azure/virtual-network/service-tags-overview) for a specific tag name, with no need to prefix of `Azure`.
   - Based on above point, some tag name (e.g. Sql) has different case between it in `SystemService`(AzureSQL) and `Name`(Sql). Since `Name` is always the same case as it is in the doc, hence I prefer that.

2. Stop the loop once we got the tag info.
3. Return error if no service tag is found.

azurerm/internal/services/network/registration.go
```diff
@@ -39,6 +39,7 @@ func (r Registration) SupportedDataSources() map[string]*schema.Resource {
 		"azurerm_public_ips":                                dataSourceArmPublicIPs(),
 		"azurerm_public_ip_prefix":                          dataSourceArmPublicIpPrefix(),
 		"azurerm_route_table":                               dataSourceArmRouteTable(),
+		"azurerm_service_tags":                              dataSourceServiceTags(),
```
magodo: Can we rename it as `azurerm_network_service_tags` so as to be easier reason about?

azurerm/internal/services/network/data_source_service_tags.go
```diff
@@ -0,0 +1,89 @@
...
+		},
+	}
+}
+
+func dataSourceServiceTagsRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Network.ServiceTagsClient
+	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	location := d.Get("location").(string)
```
magodo: Can we normalize the location?

```suggestion
	location := azure.NormalizeLocation(d.Get("location").(string))
```

azurerm/internal/services/network/data_source_service_tags.go
```diff
@@ -0,0 +1,89 @@
...
+	res, err := client.List(ctx, location)
+	if err != nil {
+		return err
+	}
+
+	service := d.Get("service").(string)
+	region := d.Get("region").(string)
```
magodo: Can we normalize this?

```suggestion
	region := azure.NormalizeLocation(d.Get("region").(string))
```

azurerm/internal/services/network/registration.go
magodo: Can we rename it as `azurerm_network_service_tags` so as to be easier reasoned about?

azurerm/internal/services/compute/tests/resource_arm_managed_disk_test.go
```diff
@@ -1022,12 +1022,12 @@ resource "azurerm_role_assignment" "disk-encryption-read-keyvault" {
 }
 
 resource "azurerm_managed_disk" "test" {
-  name                   = "acctestd-%d"
-  location               = azurerm_resource_group.test.location
-  resource_group_name    = azurerm_resource_group.test.name
-  storage_account_type   = "Standard_LRS"
-  create_option          = "Empty"
-  disk_size_gb           = 1
+  name                 = "acctestd-%d"
```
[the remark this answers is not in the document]
r0bnet: I think this came in with merge of master. Also i had to re-run `terrafmt` over this file for whatever reason...

r0bnet:
@magodo i changed everything accordingly. Any problems left?

azurerm/internal/services/network/data_source_network_service_tags.go
```diff
@@ -0,0 +1,105 @@
...
+func dataSourceNetworkServiceTags() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNetworkServiceTagsRead,
+
+		Timeouts: &schema.ResourceTimeout{
+			Read: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"location": azure.SchemaLocation(),
```
katbyte: Would this make more sense as `location_for_request` as it seems to only control which location to make the request to?

azurerm/internal/services/network/data_source_network_service_tags.go
```diff
@@ -0,0 +1,105 @@
...
+
+		if props := sti.Properties; props != nil {
+			if *props.Region == locationFilter {
+				addressPrefixes := make([]string, 0)
+				if props.AddressPrefixes != nil {
+					addressPrefixes = *props.AddressPrefixes
+				}
+				err = d.Set("address_prefixes", addressPrefixes)
+				if err != nil {
+					return fmt.Errorf("Error setting `address_prefixes`: %+v", err)
```
katbyte: Could we either use `fmt.Errorf` or `errors.NewF` throughout the entire file for consistency?

azurerm/internal/services/network/data_source_network_service_tags.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+	}
+	return errors.New("specified service tag not found")
+}
+
+// isServiceTagOf is used to check whether a service tag name belongs to the service of name `serviceName`.
+// Service tag name has format as below:
+// - (regional) serviceName.locationName
+// - (all) serviceName
+func isServiceTagOf(stName, serviceName string) bool {
```
katbyte: Is there a reason to have this as a separate function when its only used once?

azurerm/internal/services/network/data_source_network_service_tags.go
```diff
@@ -0,0 +1,105 @@
...
+
+func dataSourceNetworkServiceTagsRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Network.ServiceTagsClient
+	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	location := azure.NormalizeLocation(d.Get("location").(string))
+	res, err := client.List(ctx, location)
+	if err != nil {
+		return err
```
katbyte: Could we wrap this error?

azurerm/internal/services/network/data_source_network_service_tags.go
```diff
@@ -0,0 +1,105 @@
...
+				if sti.ID == nil {
+					return errors.New("unexpected nil ID for service tag")
+				}
+
+				d.SetId(*sti.ID)
+				return nil
+			}
+		}
+	}
+	return errors.New("specified service tag not found")
```
katbyte: If there is a filter should we also include it in the error message as why nothing was found? ie no service tags found globaly for service x, or no service tags found for x in region/location y

azurerm/internal/services/network/data_source_network_service_tags.go
```diff
@@ -0,0 +1,105 @@
...
+	service := d.Get("service").(string)
+	locationFilter := d.Get("location_filter").(string)
+
+	for _, sti := range *res.Values {
+		if sti.Name == nil || !isServiceTagOf(*sti.Name, service) {
+			continue
+		}
+
+		if props := sti.Properties; props != nil {
+			if *props.Region == locationFilter {
```
katbyte: How will this work if the location filter is empty? if i understand correctly some services are "global" so in those cases is Region empty? (could we get a comment to that affect if so) while others are regional and will have a location?

azurerm/internal/services/network/data_source_network_service_tags.go
```diff
@@ -0,0 +1,105 @@
...
+				addressPrefixes := make([]string, 0)
+				if props.AddressPrefixes != nil {
+					addressPrefixes = *props.AddressPrefixes
+				}
+				err = d.Set("address_prefixes", addressPrefixes)
+				if err != nil {
```
katbyte: I think this would be better as:
```suggestion
				if props.AddressPrefixes != nil {
					if err = d.Set("address_prefixes", *props.AddressPrefixes); err != nil {
						return fmt.Errorf("Error setting `address_prefixes`: %+v", err)
```

azurerm/internal/services/network/data_source_network_service_tags.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return err
+	}
+
+	if res.Values == nil {
+		return errors.New("unexpected nil value for service tag information")
+	}
+
+	service := d.Get("service").(string)
+	locationFilter := d.Get("location_filter").(string)
```
katbyte: we'll also need to normalize this like above

azurerm/internal/services/network/data_source_network_service_tags.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+	}
+}
+
+func dataSourceNetworkServiceTagsRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Network.ServiceTagsClient
+	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	location := azure.NormalizeLocation(d.Get("location").(string))
```
katbyte: this can be
```suggestion
	location := azure.NormalizeLocation(d.Get("location"))
```

azurerm/internal/services/network/data_source_network_service_tags.go
katbyte: we also need to nil check Region & normalize it when comparing with locationfilter

website/docs/d/network_service_tags.html.markdown
```diff
@@ -0,0 +1,51 @@
...
+
+The following arguments are supported:
+
+* `location` - (Required) The Azure Region where the Service Tags exists. This value is not used to filter the results, use `region` for that purpose instead. More information can be found here: [Service Tags URL parameters](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/servicetags/list#uri-parameters).
+
+* `service` - (Required) The type of the service for which address prefixes will be fetched. Available service tags can be found here: [Available service tags](https://docs.microsoft.com/en-us/azure/virtual-network/service-tags-overview#available-service-tags).
+
+---
+
+* `location_filter` - (Optional) Changes the scope of the service tags. Can be any value that is also valid for `location`.
```
katbyte: could we expand on this and talk about global vs regional service tags?

azurerm/internal/services/network/data_source_network_service_tags.go
magodo: @katbyte 
The reason to set `address_prefixes` anyway is to ensure there is always something in the `ResourceData` (even an empty list), so that (e.g.) output var could safely refer to `address_prefixes`. Otherwise, if you don't set it, terraform might panic if someone tries to refer to it.

azurerm/internal/services/network/data_source_network_service_tags.go
r0bnet: I would rather specify it in the documentation because imho anyone who'd read `location_for_request` would anyway go to documentation. At least i would.

azurerm/internal/services/network/data_source_network_service_tags.go
r0bnet: will use fmt.Errorf and get rid of errors.New

azurerm/internal/services/network/data_source_network_service_tags.go
r0bnet: yepp

azurerm/internal/services/network/data_source_network_service_tags.go
r0bnet: yepp

azurerm/internal/services/network/data_source_network_service_tags.go
r0bnet: yepp

azurerm/internal/services/network/data_source_network_service_tags.go
r0bnet: done

azurerm/internal/services/network/data_source_network_service_tags.go
r0bnet: I fixed normalization of the Region prop too. But from my understanding you get ALL address prefixes if you leave the location_filter out. Otherwise you get only the region you requested.

azurerm/internal/services/network/data_source_network_service_tags.go
r0bnet: Would like to go here with @magodo proposal

website/docs/d/network_service_tags.html.markdown
r0bnet: adjusted documentation

azurerm/internal/services/network/data_source_network_service_tags.go
r0bnet: would like to leave it as is. If that's a problem i can also change it

azurerm/internal/services/network/data_source_network_service_tags.go
katbyte: 👌 however we still need to nilcheck region beforedereferencingg

r0bnet:
@katbyte all good, i'm fine with you pushing changes. But don't know why the tests are failing. I ran them locally with no error. I just merged master so see if those were temp issues.

PR #6230, opened 2020-03-23
[linux|windows]_virtual_machine - Allow update of OS disk encryption settings

kazimierzbudzyk:
Fixes: https://github.com/terraform-providers/terraform-provider-azurerm/issues/6183

Inspired by how os_disk resizes have to be handled: https://github.com/kazimierzbudzyk/terraform-provider-azurerm/blob/master/azurerm/internal/services/compute/linux_virtual_machine_resource.go#L835


kazimierzbudzyk:
Updated tests to match changes in https://github.com/terraform-providers/terraform-provider-azurerm/pull/6162/commits/3765f3228d81adabf89d960868140e7a6877e003

tombuildsstuff:
@kazimierzbudzyk I've tested this in several Azure Regions but this doesn't appear to be possible at this point in time:

> compute.DisksClient#Update: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="PropertyChangeNotAllowed" Message="Changing property 'encryption.diskEncryptionSetId' is not allowed." Target="encryption.diskEncryptionSetId"

I know this functionality is coming at some point (a few months post GA was the last update I found), but do you happen to know which Azure Region this functionality is supported in?

Thanks!

kazimierzbudzyk:
@tombuildsstuff I don't have a full list, but tested it successfully in `useast` and `centralus`. Was the VM deallocated (although, terraform did it for me correctly as well)?

kazimierzbudzyk:
Tested in `useast` successfully as well, but while doing that I discovered a bug in the implementation. While the API interaction works correctly (verified in the UI), state is not getting updated so the update continues showing up as pending change. I am looking into what could be causing it (field not getting set correctly in `resourceLinuxVirtualMachineRead`?), but if you have any ideas what could be happening here it would be very helpful!

kazimierzbudzyk:
Looks like it's actually an API issue - I managed to reproduce in `az cli`. API returns `diskEncryptionSet` empty for encrypted `os_disk`:
```
      "osDisk": {
        "caching": "ReadWrite",
        "createOption": "FromImage",
        "diffDiskSettings": null,
        "diskSizeGb": null,
        "encryptionSettings": null,
        "image": null,
        "managedDisk": {
          "diskEncryptionSet": null,
          "id": "/subscriptions/${subscriptionId}/resourceGroups/ZAZ-TEST-USEAST/providers/Microsoft.Compute/disks/cmk-os-disk",
          "resourceGroup": "ZAZ-TEST-USEAST",
          "storageAccountType": null
        },
        "name": "cmk-os-disk",
        "osType": "Linux",
        "vhd": null,
        "writeAcceleratorEnabled": false
      }
```

kazimierzbudzyk:
Looks like I was missing changes to read logic, fixed in https://github.com/terraform-providers/terraform-provider-azurerm/pull/6230/commits/09a21c4d9d77d9123402044d40946efabbf53edb.

kazimierzbudzyk:
@tombuildsstuff What do you think about handling the issue by wrapping `PropertyChangeNotAllowed` to return message suggesting selected region doesn't support the feature?

ramankumarlive:
@kazimierzbudzyk @tombuildsstuff  I am the Azure PM for the feature. The feature is already available in all the public and Azure Government regions in GA. It will be available in rest of the Sovereign regions soon. Please see the documentation for the supported regions.


https://docs.microsoft.com/en-us/azure/virtual-machines/windows/disk-encryption#supported-regions

ashrayjain:
Could we get this in 2.5.0 if possible. This is blocking us from picking up CMKs.
Thanks!

Issue #6059, opened 2020-03-10
AzureRm V2: Cannot create keyVault without Subscription Level Permissions


KSLiquid:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

azurerm version 2.0
terraform version v0.12.21

### Affected Resource(s)

* azurerm_key_vault

### Terraform Configuration Files

```hcl
provider "azurerm" {
    version = "2.0"
    features {}
    skip_provider_registration = "true"
}
 
resource "azurerm_key_vault" "AKVDea01" {
    name                            = "mykvname"
    location                        = "westeurope"
    resource_group_name             = "myrgname"
    tenant_id                       = "my-tenant-id"

    sku_name = "standard" 
}
```

### Expected Behavior

I expect the keyVault to be created

### Actual Behavior

I can't create the Keyvault and i get the following error:

Error: Error checking for the presence of an existing Soft-Deleted Key Vault "mykvname" (Location "westeurope"): keyvault.VaultsClient#GetDeleted: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="AuthorizationFailed" Message="The client 'b.xxx@xxx.de' with object id 'xxxxx' does not have authorization to perform action 'Microsoft.KeyVault/locations/deletedVaults/read' over scope '/subscriptions/xxxxx' or the scope is invalid. If access was recently granted, please refresh your credentials."

### Steps to Reproduce

create a new terraform project.
Copy my code from above and try:

`terraform init`
`terraform apply`

### Important Factoids

I do not have access on the subscription level, just on the resource group.
Of course this should not be a problem (and it wasn't).
Im azurerm version 1.28.0 i can create a keyVault without the same permissions i have right know.
the following script works for me:

```hcl
provider "azurerm" {
    version = "1.28.0"
    skip_provider_registration = "true"
}
 
resource "azurerm_key_vault" "AKVDea01" {
    name                            = "mykvname"
    location                        = "westeurope"
    resource_group_name             = "myrgname"
    tenant_id                       = "my-tenant-id"

[...]

PR #6238, opened 2020-03-24
Key vault permission fix

KSLiquid:
This is my humble try to fix the issue #6059.

Please check carefully since this is my first github contribution ever.

The idea is that the check for a soft deleted key vault exists is not nessessary, if the user does not want to recover It anyways.
So I put it in an if stetement checking for that.
This way a user could avoid the issue by setting
`      recover_soft_deleted_key_vaults = false`
in the prodiver block.


KSLiquid:
This solution does not work as I expected. It is nessessary to check for a soft deleted KeyVault with the same name, even if I do NOT want to replace it.

PR #6239, opened 2020-03-24
Issue #6059 - subscription rights for key_vault

Jawvig:
This commit fixes issue #6509, in which some subscription level
rights are required in order to create a key_vault. This occurred
due to the provider always looking for a soft deleted key vault
prior to checking whether the feature
KeyVault.RecoverSoftDeletedKeyVaults was set to true. Thus event
setting the value to false did not stop it searching for the soft
deletions.

The rights it was requiring were to perform action
'Microsoft.KeyVault/locations/deletedVaults/read' over scope
'/subscriptions/xxxxxx'

### DIFF
--- a/azurerm/internal/services/keyvault/resource_arm_key_vault.go
+++ b/azurerm/internal/services/keyvault/resource_arm_key_vault.go
@@ -214,22 +214,24 @@ func resourceArmKeyVaultCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	// before creating check to see if the key vault exists in the soft delete state
-	softDeletedKeyVault, err := client.GetDeleted(ctx, name, location)
-	if err != nil {
-		if !utils.ResponseWasNotFound(softDeletedKeyVault.Response) {
-			return fmt.Errorf("Error checking for the presence of an existing Soft-Deleted Key Vault %q (Location %q): %+v", name, location, err)
-		}
-	}
-
-	// if so, does the user want us to recover it?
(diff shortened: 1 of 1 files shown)


Jawvig:
Hold off on this. One of the acceptance tests is failing:
```
    TestAccAzureRMKeyVault_softDeleteRecoveryDisabled: testing.go:621: Step 3, no error received, but expected a match to:

        An existing soft-deleted Key Vault exists with the Name
```
I'll look into it further.

Jawvig:
It looks to me like the test might be wrong, but I'm getting a bit out of my depth and help would be appreciated. 

I'm looking at the [`softDeleteRecoveryDisabled` test](https://github.com/terraform-providers/terraform-provider-azurerm/blob/edf4033dd8bed7d31f6d5ff3c8ead650fc54aff7/azurerm/internal/services/keyvault/tests/resource_arm_key_vault_test.go#L350)

The [deletion step](https://github.com/terraform-providers/terraform-provider-azurerm/blob/edf4033dd8bed7d31f6d5ff3c8ead650fc54aff7/azurerm/internal/services/keyvault/tests/resource_arm_key_vault_test.go#L1096) does not seem to set the `purge_soft_delete_on_destroy` to `false` in the provider features, and thus it should, I think, permanently delete the key vault rather than soft delete it. 

Then when the resource is attempted to be recreated there's no soft deleted one present and so the expected error message does not appear. Am I missing something?

Jawvig:
I see why the test is failing. I'm going to close this pull request and put some comments into #6239

PR #6240, opened 2020-03-25
provider misc cleanup - rg names & normalize location

katbyte:


tombuildsstuff:
Rather than making this generic method less strictly typed - I've opened #6242 which supersedes this by splitting the `NormalizeLocation` into 3 to give us more granular methods which can be more strongly typed, allowing us to lean on the Go compiler here:

1. `Normalize` - taking a string returning a normalized string
2. `NormalizeNilable` - for flattening nilable location fields to ensure there's always a value
3. `StateFunc` - for flattening this into the state

PR #6241, opened 2020-03-25
add different create_mode for resource `azurerm_postgresql_server`

### LINKED ISSUE #6180: Terraform support for point on time server restore for Azure managed PostgreSQL
Hi,
We are using terraform for our couple cluster deployment for our product in MS Azure. But, in case of managed PostgreSQL, when we do a parallel upgrades we are not able to restore an existing PostgreSQL DB to a new instance with point in time restore option. Azure cli supports the same as below command. Azure managed PostgreSQL server restore (point in time) is not supported in terraform at present.

./az postgres server restore -g DatabaseTestingnPOC -n NewPrimary -s primary --restore-point-in-time "2020-03-17T16:50:30.0000Z"

Can you please help in getting this through terraform.

Thanks best regards
Krishna
Official email - ksingita@cisco.com

njuCZ:
add support of geo_restore, PointInTimeRestore and replica create for resource `azurerm_postgresql_server`.

fix #6180

### DIFF
--- a/azurerm/internal/services/postgres/parse/postgresql_server.go
+++ b/azurerm/internal/services/postgres/parse/postgresql_server.go
@@ -1,0 +1,33 @@
+package parse
+
+import (
+	"fmt"
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+)
+
+type PostgresqlServerId struct {
+	ResourceGroup string
+	Name          string
+}
+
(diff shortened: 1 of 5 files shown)


katbyte:
Hey @njuCZ, i was adding some additional properties to this resource and ended up refactoring it quite a bunch so i've pulled these changes into the resource. As such i am going to close this in favour of #6459.

PR #6250, opened 2020-03-25
Changed management policy to only send set properties

simongh:
There's an issue when any of the optional properties are not set on a management policy. These were being sent to Azure & it was interpreting these as set with a value of zero.

Looking at what the Azure portal does, if these are not intended to be set they should not be sent. I've modified the resource to only add the 3 properties if they are not nil.


tombuildsstuff:
Tests pass:

Screenshot 2020-03-26 at 10 22 52

PR #6260, opened 2020-03-25
azurerm_key_vault - can now be created without subscription level permissions (fixes #6059)

### LINKED ISSUE #6059: AzureRm V2: Cannot create keyVault without Subscription Level Permissions

Jawvig:
This commit fixes #6509 with the following changes:

1) On Key Vault create, only looking at the subscription level for soft deleted vaults if the `RecoverSoftDeletedKeyVaults` flag is set to `true`
2) Remove the (now redundant) error message returned when there's a soft deleted vault of the same name but the user doesn't want to recover it. This is already covered by an error returned by Azure.
2) Fix the acceptance test to expect the error message Azure returns in this instance, rather than the text in the removed error.
3) Fixes the test to add a secret to the vault that will be soft deleted, as Azure will otherwise silently recover the vault anyway (even if the mode is not 'recover') if there is no content in the vault.
4) Fixes the test so that it doesn't purge the vault on destroy.

(fixes #6059 )


Jawvig:
I'd like to link issue #6059 to this PR.

mongrelion:
This is good hunting, @Jawvig
Not a maintainer here but, is there anything I can do to help move this forward? I'm dealing with the same issue and running our own compiled version of the provider is not an option for production.

Jawvig:
@mongrelion I've just promoted it in the Slack channel to see if we can get it into the next release.

mongrelion:
Great!
By the way, I just tested this against our DTA environments and it works like a charm. Not sure if anyone else wants to take it for a spin, since there's already an acceptance test in place.
Also, to clarify, I am in the same boat where I do have access to the resource group but not to the subscription, so that's why the provider fails at the time of retrieving the soft-deleted key vaults.

jacksinclairaudley:
This is going to save our bacon! Thankyou @Jawvig!

A30004028:
Hi @mongrelion , did you manage to get a response on a release for this issue? Thanks

mfolley:
If this makes its way into the next release this is going to make our lives at lot easier. Thanks for the effort @Jawvig

Jawvig:
@a30001539 A couple of us have pushed it on the Slack channel. I know @tombuildsstuff has been knee deep in critical AKS fixes for a while, but looks like that's released now today in 2.5.0. Here's hoping a cheeky comment mention will bump it up! It's quite a simple fix. If it wasn't, I wouldn't have had the skills. :-)

siwon:
Hi, can we have some information about the reason it hadn't been merged in 2.7.0 version ?
Crossed fingers for a merge in 2.8.0 version ! :pray:

PS: Thanks a lot for all your work btw ! :thumbsup:

kamakay:
Hi! Do you have a timeline for the 2.8 release?
Thanks

Jawvig:
Hi @katbyte I confess I'm confused. The new code seems to ignore the `RecoverSoftDeletedKeyVaults` feature - is this now retired? It always attempts to recover, but if it doesn't have permissions it just quietly ignores the fact that there might be one that should have been recovered. So two people running it with different permissions would end up with different results - either a recovered one or a new one.

I feel I'm overlooking something here.

katbyte:
This is to directly fix the linked issue here, unable to create a key vault due to lack of permissions - @jackofallops has been looking into the other issues raised in this PR and will follow up with a PR that fixes the issues around recovering keyvaults & child items within them, but that will not make it into the 2.8. Given the number of upvotes on the linked issue we figured it would be best to fix that issue sooner then later and follow up with the rest in 2.9/2.10.

Jawvig:
I understand it fixes the linked issue. It's just the way it does it is to change the behaviour so that it will silently fail to recover a keyvault if you don't have permissions even when `RecoverSoftDeletedKeyVaults` is set to `true`. Whereas if you have permissions it will recover it. So, two different outcomes just based on permission, and ignoring the feature value.

In the original code in this PR, it would still fail if you didn't have permissions and explicitly set `RecoverSoftDeletedKeyVaults` to `true`, but would successfully create a new key vault if you set `ReceoverSoftDeletedKeyVaults` to `false` (thus solving the linked issue). I guess it's a matter of opinion, and I'm probably a bit biased here, but I reckon this is the better solution. Anyway, I won't argue it further.

And agreed - the effect of the child objects were a separate issue and can be looked at elsewhere. It's just that I noticed when running the tests under different permissions, the test still passed anyway when it shouldn't have done, so I wasn't able to check the change was having any effect until I made those test changes.

katbyte:
If the keyvault is soft deleted i believe it will then fail to create it after silently failing to recover it when it tries to recreate a vault with the same name? While the error may be different i thinkt he outcome is similar

Jawvig:
Yes, you're right! I hadn't considered that. Okay, I'm convinced. 😀

Issue #3182, opened 2019-04-05
Configure Data Sources in a Log Analytics Workspace


josh-barker:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I'd like to be able to configure Data Sources on a Log Analytics Workspace

### New or Affected Resource(s)

* azurerm_log_analytics_workspace

and/or:

* azurerm_log_analytics_workspace_iis_logs
* azurerm_log_analytics_workspace_linux_performance_collection
* azurerm_log_analytics_workspace_linux_performance_object
* azurerm_log_analytics_workspace_linux_syslog
* azurerm_log_analytics_workspace_linux_syslogCollection
* azurerm_log_analytics_workspace_windows_event
* azurerm_log_analytics_workspace_windows_performance_counter

Note: There may be other data sources and properties that can be configured.

### Potential Terraform Configuration

Inline example of configuring "datasource" properties in a Log Analytics workspace

```hcl
resource "azurerm_log_analytics_workspace" "ws" {
  name                = "Demo"
  location            = "US East"
  resource_group_name = "Some-Resource-Group"
  sku                 = "Standard"

  windows_performance_counter { 
    object_name = "LogicalDisk"
    instance_name = "*"
    interval_seconds = 10
    counter_name = "Disk Writes/sec"
  }

  windows_performance_counter { 
    object_name = "LogicalDisk"
    instance_name = "*"
    interval_seconds = 10
    counter_name = "Free Megabytes"
  }

  linux_performance_collection {
    state = "Enabled"
  }

  linux_performance_object {
    performance_counters = ["% Processor Time", "% Privileged Time"]
    object_name = "Processor"
    instance_name = "*"
    interval_seconds = 10
  }
}
```

Configure "datasource" using a separate resource for a Log Analytics workspace.

```hcl
resource "azurerm_log_analytics_workspace_iis_logs" "iis_logs" {
  name                = "iis_logs"
  resource_group_name = "Some-Resource-Group"

  state = "OnPremiseEnabled"
}

resource "azurerm_log_analytics_workspace_linux_performance_collection" "lin_perf_collection" {
[...]

PR #6274, opened 2020-03-27
New Resource: azurerm_log_analytics_datasource_windows_performance_counter

magodo:
This is one of the datasource requested by #3182.

There is a trade-off of interface design, is that for this specific ds, it has several different counters, and user specify one by setting the object_name. So it means user can only set one counter each time. While there are acutally kinds of counters out there.
So maybe we can give user a list to specify them at once. But later I feel it has several problems:

1. we need to iterate the list to call API one by one, unless using some concurrency (which is rare in terraform provider)
2. we need to define different names for those counters, since each counter has its own resource name/id. This is weired and maybe breaking some contract in terraform.

Another point is that currently the `name`, `object_name`, `counter_name`, `instance_name` allow arbitrary string, hence I'm using the `StringIsNotEmpty` validation for them.

Test Result:

```bash
💤  terraform-provider-azurerm [log_analytics_workspace_datasource] ⚡  make testacc TEST=./azurerm/internal/services/loganalytics/tests TESTARGS="-run='TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_'"
TF_ACC=1 go test ./azurerm/internal/services/loganalytics/tests -v -run='TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_' -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_basic
=== PAUSE TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_basic
=== RUN   TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_complete
=== PAUSE TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_complete
=== RUN   TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_update
=== PAUSE TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_update
=== RUN   TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_requiresImport
=== PAUSE TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_requiresImport
=== CONT  TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_basic
=== CONT  TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_requiresImport
=== CONT  TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_update
=== CONT  TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_complete
--- PASS: TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_complete (180.08s)
--- PASS: TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_basic (180.31s)
--- PASS: TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_requiresImport (193.07s)
--- PASS: TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_update (238.72s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/loganalytics/tests  238.765s
```

### DIFF
--- a/azurerm/internal/services/loganalytics/tests/resource_arm_log_analytics_datasource_windows_performance_counter_test.go
+++ b/azurerm/internal/services/loganalytics/tests/resource_arm_log_analytics_datasource_windows_performance_counter_test.go
@@ -1,0 +1,235 @@
+package tests
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/loganalytics/parse"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_basic(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_log_analytics_datasource_windows_performance_counter", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMLogAnalyticsDataSourceWindowsPerformanceCounterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMLogAnalyticsDataSourceWindowsPerformanceCounterExists(data.ResourceName),
+					resource.TestCheckResourceAttr(data.ResourceName, "counter_name", "CPU"),
+					resource.TestCheckResourceAttr(data.ResourceName, "object_name", "CPU"),
+					resource.TestCheckResourceAttr(data.ResourceName, "interval_seconds", "10"),
+					resource.TestCheckResourceAttr(data.ResourceName, "instance_name", "*"),
+				),
+			},
+			data.ImportStep(),
+		},
+	})
+}
+
+func TestAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_complete(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_log_analytics_datasource_windows_performance_counter", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMLogAnalyticsDataSourceWindowsPerformanceCounterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_complete(data),
+				Check: resource.ComposeTestCheckFunc(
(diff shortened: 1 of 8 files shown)


azurerm/internal/services/loganalytics/tests/resource_arm_log_analytics_datasource_windows_performance_counter_test.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func testAccAzureRMLogAnalyticsDataSourceWindowsPerformanceCounter_template(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
+
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-%d"
```
mbfrahry: We're adding the package initials to all our test resource groups which helps better track potential test issues through thee resource group.

```suggestion
  name     = "acctestRG-la-%d"
```

mbfrahry (changes requested):
Hey @magodo, this is definitely a weird resource to craft from what we're given on the api/sdk side but I think you've nailed it. It is weird that we have to do each counter separately but I don't think there is a way to do it in bulk that doesn't feel hacky so let's stick with the way you've done it here.

My main issue is just how we're presenting the attributes to the user which I've pointed out in more detail below.

If this model is also universal across other datasources, it might be worthwhile to tease this out into it's own separate package but that work can be done once more start coming in. 

Thanks for the effort here!

website/docs/r/log_analytics_datasource_windows_performance_counter.html.markdown
```diff
@@ -0,0 +1,82 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) The Name which should be used for this Log Analytics Windows Performance Counter DataSource. Changing this forces a new Log Analytics Windows Performance Counter DataSource to be created.
+
+* `resource_group_name` - (Required) The name of the Resource Group where the Log Analytics Windows Performance Counter DataSource should exist. Changing this forces a new Log Analytics Windows Performance Counter DataSource to be created.
+
+* `workspace_name` - (Required) The name of the Log Analytics Workspace where the Log Analytics Windows Performance Counter DataSource should exist. Changing this forces a new Log Analytics Windows Performance Counter DataSource to be created.
+
+* `counter_name` - (Required) The friendly name of the performance counter.
```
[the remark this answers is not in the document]
magodo: Sure! I didn't come up with that.

azurerm/internal/services/loganalytics/tests/resource_arm_log_analytics_datasource_windows_performance_counter_test.go
magodo: Will do it.

magodo:
@mbfrahry Thank you for review, I have made the change accordingly, please have a check.

Regarding to: 
> If this model is also universal across other datasources, it might be worthwhile to tease this out into it's own separate package but that work can be done once more start coming in.

That is exactly what I was thinking about at first. While after a closely looking into the property definition of each DS(s), I find they has slightly difference among others. I listed some points which will affect code structure below which might hint we may have to implement each DS separately with different schema definitions and undertaking code redundency in some degree:

| Kind  | Sub-kind | Sub-sub-kind | has switch?|
| - | - | - | - |
|Windows Event|eventLogName|eventType(static)|no|
|Windows Performance Counter|objectName|N/A|no|
|IIS|N/A|N/A|yes|
|Linux Performance Counter|objectName|performanceCounters(dynamic)|yes|
|Linux Syslog|syslogName|syslogSeverities(static)|yes|
|Custom Log| **TBD**|||

mbfrahry (approved):
LGTM!

mbfrahry:
Ahhh, that make sense. Thanks for pulling that table together.

PR #6276, opened 2020-03-27
azurerm_management_group - deprecated and rename `group_id` to `name`

ArcturusZhang:
The term `group_id` in management group is sometimes confusing with the UUID of the mgmt group or a full ID of a mgmt group. And a mgmt group can have a human-readable string as its `group_id`. Therefore `group_id` is not quite a good term for this.

I introduced a new attribute `name` for the resource and data source `azurerm_management_group` to provide more friendly usage of management group and reduce the confusion.

### DIFF
--- a/azurerm/internal/services/managementgroup/tests/data_source_management_group_test.go
+++ b/azurerm/internal/services/managementgroup/tests/data_source_management_group_test.go
@@ -37,7 +37,7 @@ resource "azurerm_management_group" "test" {
 }
 
 data "azurerm_management_group" "test" {
-  group_name = azurerm_management_group.test.group_name
+  name = azurerm_management_group.test.name
 }
 `, data.RandomInteger)
 }
--- a/azurerm/internal/services/managementgroup/tests/data_source_management_group_test.go
+++ b/azurerm/internal/services/managementgroup/tests/data_source_management_group_test.go
@@ -37,7 +37,7 @@ resource "azurerm_management_group" "test" {
 }
 
 data "azurerm_management_group" "test" {
-  group_id = azurerm_management_group.test.group_id
+  group_name = azurerm_management_group.test.group_name
 }
 `, data.RandomInteger)
 }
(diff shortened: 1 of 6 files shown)


website/docs/r/management_group.html.markdown
```diff
@@ -39,9 +39,13 @@ resource "azurerm_management_group" "example_child" {
 
 The following arguments are supported:
 
-* `group_id` - (Optional) The UUID for this Management Group, which needs to be unique across your tenant - which will be generated if not provided. Changing this forces a new resource to be created.
+* `name` - (Optional) The name or UUID for this Management Group, which needs to be unique across your tenant. A new UUID will be generated if not provided. Changing this forces a new resource to be created.
 
-* `display_name` - (Optional) A friendly name for this Management Group. If not specified, this'll be the same as the `group_id`.
+* `group_id` - (Optional) The name or UUID for this Management Group, which needs to be unique across your tenant. A new UUID will be generated if not provided. Changing this forces a new resource to be created.
+
+~> **NOTE:** The field `group_id` has been deprecated in favor of `name`.
```
katbyte: We typically can just leave out deprecated properties as unless a user is using them they don't really need to know, and if they are the deprecation message will let them know the new proeprty

azurerm/internal/services/managementgroup/tests/data_source_management_group_test.go
```diff
@@ -37,7 +37,7 @@ resource "azurerm_management_group" "test" {
 }
 
 data "azurerm_management_group" "test" {
-  group_id = azurerm_management_group.test.group_id
+  group_name = azurerm_management_group.test.group_name
```
katbyte: i think this should be
```suggestion
  name = azurerm_management_group.test.name
```

website/docs/r/management_group.html.markdown
ArcturusZhang: You mean to remove this line of deprecation notice? Removed.

azurerm/internal/services/managementgroup/tests/data_source_management_group_test.go
ArcturusZhang: Oops... forgot this. Fixed.

ArcturusZhang:
Hi @katbyte I fixed the data source test, and removed the deprecation notice in the doc. Please let me know if this is not what you meant...

PR #6301, opened 2020-03-30
azurerm_[linux|windows]_virtual_machine - if the `priority` property on read is empty assume it to be `Regular`

kazimierzbudzyk:
Fixes: https://github.com/terraform-providers/terraform-provider-azurerm/issues/6300

### DIFF
--- a/azurerm/internal/services/compute/linux_virtual_machine_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_resource.go
@@ -566,8 +566,13 @@ func resourceLinuxVirtualMachineRead(d *schema.ResourceData, meta interface{}) e
 			return fmt.Errorf("Error setting `secret`: %+v", err)
 		}
 	}
-
-	d.Set("priority", string(props.Priority))
+	// Resources created with azurerm_virtual_machine have priority set to ""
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -566,8 +566,13 @@ func resourceLinuxVirtualMachineRead(d *schema.ResourceData, meta interface{}) e
...
 	}
-
-	d.Set("priority", string(props.Priority))
+	// Resources created with azurerm_virtual_machine have priority set to ""
+	// We need to treat "" as equal to "Regular" to allow migration azurerm_virtual_machine -> azurerm_linux_virtual_machine
+	priority := props.Priority
+	if priority == "" {
+		priority = compute.Regular
+	}
+	d.Set("priority", string(priority))
```
ArcturusZhang, line 575: Could we please do this by assigning the default value first, and then check if response is empty and then assign the real value if it is empty.
```suggestion
	priority := string(compute.Regular)
	if props.Priority != "" {
		priority = string(props.Priority)
	}
	d.Set("priority", priority)
```

kazimierzbudzyk:
Thank you for the review @ArcturusZhang! Made changes according to your comments.

kazimierzbudzyk:
Thanks! @ArcturusZhang what is the process for getting this added to a release?

PR #6309, opened 2020-03-31
New resource: azurerm_private_dns_txt_record

### LINKED ISSUE #5012: Support for azurerm_private_dns_txt_record
### Community Note

### Description

Would be nice to have also resource for adding txt records to private dns.

* azurerm_private_dns_txt_record

```hcl

resource "azurerm_private_dns_txt_record" "example" {
  name                = "foobar"
  zone_name           = "${azurerm_private_dns_zone.example.name}"
  ttl                 = 300
  txt              = "foobar txt string"
}

```

### References

* #0000

manicminer:
Fixes #5012

### DIFF
--- a/azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
+++ b/azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
@@ -163,7 +163,7 @@ func resourceArmPrivateDnsTxtRecordRead(d *schema.ResourceData, meta interface{}
 	d.Set("fqdn", resp.Fqdn)
 
 	if err := d.Set("record", flattenAzureRmPrivateDnsTxtRecords(resp.TxtRecords)); err != nil {
-		return err
+		return fmt.Errorf("setting `record`: %s", err)
 	}
 
 	return tags.FlattenAndSet(d, resp.Metadata)
--- a/azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
+++ b/azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
@@ -106,14 +106,11 @@ func resourceArmPrivateDnsTxtRecordCreateUpdate(d *schema.ResourceData, meta int
 		}
 	}
 
-	ttl := int64(d.Get("ttl").(int))
-	t := d.Get("tags").(map[string]interface{})
-
 	parameters := privatedns.RecordSet{
 		Name: &name,
 		RecordSetProperties: &privatedns.RecordSetProperties{
-			Metadata:   tags.Expand(t),
-			TTL:        &ttl,
+			Metadata:   tags.Expand(d.Get("tags").(map[string]interface{})),
+			TTL:        utils.Int64(int64(d.Get("ttl").(int))),
 			TxtRecords: expandAzureRmPrivateDnsTxtRecords(d),
 		},
 	}
--- a/azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
+++ b/azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
@@ -1,0 +1,240 @@
+package privatedns
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/privatedns/mgmt/2018-09-01/privatedns"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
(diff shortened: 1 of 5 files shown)


azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
```diff
@@ -0,0 +1,237 @@
...
+			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
+
+			"zone_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
+
+			"record": {
```
mbfrahry, line 57: Do you think any other attributes will be place under record? If not, we could probably shorten this to just be a TypeSet with string elements
```
Type:     schema.TypeSet,
					Required: true,
					Elem: &schema.Schema{
						Type: schema.TypeString,
					},
```

azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	d.Set("name", name)
+	d.Set("resource_group_name", resGroup)
+	d.Set("zone_name", zoneName)
+	d.Set("ttl", resp.TTL)
+	d.Set("fqdn", resp.Fqdn)
+
+	if err := d.Set("record", flattenAzureRmPrivateDnsTxtRecords(resp.TxtRecords)); err != nil {
+		return err
```
mbfrahry, line 166: We should expand this out to:
```suggestion
		return fmt.Errorf("setting `record`: %s", err)
```
Just to give us a better picture of what thing we had trouble setting

azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
```diff
@@ -0,0 +1,237 @@
...
+	return results
+}
+
+func expandAzureRmPrivateDnsTxtRecords(d *schema.ResourceData) *[]privatedns.TxtRecord {
+	recordStrings := d.Get("record").(*schema.Set).List()
+	records := make([]privatedns.TxtRecord, len(recordStrings))
+
+	segmentLen := 254
+	for i, v := range recordStrings {
+		record := v.(map[string]interface{})
```
mbfrahry, line 219: I'm always hesitant about making this type of assertion as I've seen cases where this is nil and we panic. Let's add a continue if this is nil and then we can keep the rest of the logic the same.

azurerm/internal/services/privatedns/tests/resource_arm_private_dns_txt_record_test.go
```diff
@@ -0,0 +1,334 @@
...
+		CheckDestroy: testCheckAzureRMPrivateDnsTxtRecordDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMPrivateDnsTxtRecord_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMPrivateDnsTxtRecordExists(data.ResourceName),
+					resource.TestCheckResourceAttr(data.ResourceName, "record.#", "2"),
+				),
+			},
+			{
```
mbfrahry, line 72: Let's add import steps between all these checks and one more step at the end that takes us back to the basic configuration just to confirm it all works

azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
manicminer: Thanks, I wasn't sure about this actually as I wanted to try and keep it similar to the [azurerm_dns_txt_record](https://github.com/terraform-providers/terraform-provider-azurerm/blob/c916aa6db4142add1760a9b39bc668eb8b1dfcde/azurerm/internal/services/dns/resource_arm_dns_txt_record.go#L54-L66) resource for consistency. Do you think it's worth changing both resources (and deprecating the `value` key in the existing resource)?

azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
manicminer: Do you mean the record values may come back as nil? e.g. mitigate like:

```go
	segmentLen := 254
	for i, v := range recordStrings {
		if v == nil {
			continue
		}

		record := v.(map[string]interface{})
```

azurerm/internal/services/privatedns/tests/resource_arm_private_dns_txt_record_test.go
manicminer: Great, thanks! Hoping I've interpreted your suggestion correctly, have pushed some changes.

azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
mbfrahry: Exactly. Looks great!

azurerm/internal/services/privatedns/tests/resource_arm_private_dns_txt_record_test.go
```diff
@@ -0,0 +1,343 @@
...
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMPrivateDnsTxtRecordDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMPrivateDnsTxtRecord_withTags(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMPrivateDnsTxtRecordExists(data.ResourceName),
+					resource.TestCheckResourceAttr(data.ResourceName, "tags.%", "2"),
+				),
+			},
```
mbfrahry, line 105: We want the import step here too

azurerm/internal/services/privatedns/resource_arm_private_dns_txt_record.go
mbfrahry: Ahhh, yea. I get that. I think that's just leftover from trying to match the terraform schema to the api struct but it's not a best practice. I don't think we can match it at this time because we couldn't keep the same name. We could just note it as a todo for 3.0

tombuildsstuff:
Tests pass:

Screenshot 2020-04-08 at 14 28 04

PR #6315, opened 2020-04-01
Add `value` to Attributes

aiell0:
Added `value` as an attribute to the Azure Key Vault.

### DIFF
--- a/website/docs/r/key_vault_secret.html.markdown
+++ b/website/docs/r/key_vault_secret.html.markdown
@@ -98,6 +98,7 @@ The following arguments are supported:
 The following attributes are exported:
 
 * `id` - The Key Vault Secret ID.
+* `value` - The Key Vault Value.
 * `version` - The current version of the Key Vault Secret.
 
 ## Timeouts


tombuildsstuff:
hey @aiell0 

Thanks for opening this PR :)

In Terraform all arguments to a resource are also available as attributes - as such to prevent duplication, the `attributes` section of the documentation only shows the attributes available in addition to the arguments.

We're gradually working to update the text of the attributes section across all resources to match the [Resource Group](https://www.terraform.io/docs/providers/azurerm/r/resource_group.html#attributes-reference) and [Search Service](https://www.terraform.io/docs/providers/azurerm/r/search_service.html#attributes-reference) documentation to make this clearer, by changing:

```
```

to:

```
In addition to the Arguments listed above - the following Attributes are exported:
```

Whilst we could work through making these changes to all the documentation at once - we're taking this as an opportunity to update the documentation at the same time - as such these changes will be rolled out gradually.

As such whilst I'd like to thank you for this contribution, I'm going to close this PR for the moment - that said I believe the changes mentioned above should ultimately make this clearer.

Thanks!

PR #6328, opened 2020-04-01
Update README.md

pearcec:
updating the build target to be correct for isolating a test

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ $ make test
 The majority of tests in the provider are Acceptance Tests - which provisions real resources in Azure. It's possible to run the entire acceptance test suite by running `make testacc` - however it's likely you'll want to run a subset, which you can do using a prefix, by running:
 
 ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,7 +96,7 @@ $ make test
 The majority of tests in the provider are Acceptance Tests - which provisions real resources in Azure. It's possible to run the entire acceptance test suite by running `make testacc` - however it's likely you'll want to run a subset, which you can do using a prefix, by running:
```
tombuildsstuff, line 96: could we update this too?

```suggestion
The majority of tests in the provider are Acceptance Tests - which provisions real resources in Azure. It's possible to run the entire acceptance test suite by running `make acctests` - however it's likely you'll want to run a subset, which you can do using a prefix, by running:
```

pearcec: Is that true?  Looking at the GNUmakefile there is still a target for testacc, which looks like it runs all the tests.  Please verify.

tombuildsstuff: good spot

Issue #6094, opened 2020-03-12
`azurerm_kubernetes_cluster`: `node_count` is always `nil` after initial creation


wagnst:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

* `azurerm_2.1.0`
* `TF 0.12.21`

### Affected Resource(s)

* `azurerm_kubernetes_cluster`
* `default_node_pool`

### Terraform Configuration Files

```hcl
...
default_node_pool {
    ...
    node_count          = 1
    enable_auto_scaling = false
    ...
}
...
```

### Expected Behavior
No error occuring for default_node_pool.

### Actual Behavior

```
Error: Error updating Default Node Pool "lhg-weur-kfkadevmdw-aks" (Resource Group "lhg-weur-kfkadevmdwaks-rg"): containerservice.AgentPoolsClient#CreateOrUpdate: Invalid input: autorest/validation: validation failed: parameter=parameters.ManagedClusterAgentPoolProfileProperties.Count constraint=Null value=(*int32)(nil) details: value can not be null; required parameter
```

I explicitly do NOT set `min_count` and `max_count` at all in that case (autoscaling = false). But when I check the state this is set (again), although if I remove the fields manually from it - maybe thats not relevant and is the default behavior anyways:

```
    default_node_pool {
        availability_zones    = [
            "1",
        ]
        enable_auto_scaling   = false
        enable_node_public_ip = false
        max_count             = 0
        max_pods              = 250
        min_count             = 0
        name                  = "default"
        node_count            = 1
        node_labels           = {}
        node_taints           = []
        os_disk_size_gb       = 128
        tags                  = {}
        type                  = "VirtualMachineScaleSets"
        vm_size               = "Standard_D2s_v3"
        vnet_subnet_id        = "/subscriptions/xxx/resourceGroups/yyy/providers/Microsoft.Network/virtualNetworks/zzz/subnets/aaa"
    }
```

Then if I change the `max_count` and `min_count` to `node_count` (in my example its `1`) I get:

```
[...]

PR #6357, opened 2020-04-03
Fix issue with default_node_pool min/max counts + update documentation

adback03:
The max/min counts in the default node pool should be in the range of 0 to 100, not 1 and 100. This is a similar issue to that of #6020, but was never fixed for the default node pool.

I have also updated both sets of documentation to reflect these changes.

One minor update I snuck in this PR as well is just fixing the documentation for vnet_subnet_id. It is not required.

EDIT: This is potentially related to #6094 as well.

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_nodepool.go
+++ b/azurerm/internal/services/containers/kubernetes_nodepool.go
@@ -68,7 +68,7 @@ func SchemaDefaultNodePool() *schema.Schema {
 				"max_count": {
 					Type:         schema.TypeInt,
 					Optional:     true,
-					ValidateFunc: validation.IntBetween(1, 100),
+					ValidateFunc: validation.IntBetween(0, 100),
 				},
 
 				"max_pods": {
@@ -81,7 +81,7 @@ func SchemaDefaultNodePool() *schema.Schema {
 				"min_count": {
 					Type:         schema.TypeInt,
 					Optional:     true,
-					ValidateFunc: validation.IntBetween(1, 100),
+					ValidateFunc: validation.IntBetween(0, 100),
 				},
 
 				"node_count": {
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
hey @adback03 

Thanks for this PR :)

That the `min_count` and `max_count` fields of the `azurerm_kubernetes_cluster_node_pool` resource can be set to `0` is actually a bug we're intending to fix in the v2.5 release - the root-cause of this being [this error message here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/containers/kubernetes_nodepool.go#L245), where this value should be set to `null` rather than `0` (and this confusion stems from the error message being incorrect).

There's recently been some breaking changes in the AKS API - meaning that unfortunately we're going to have to include some breaking changes in the v2.5 release of the Azure Provider - as such we're going to take this opportunity to fix this bug and require that this field is set to `null` in both the `azurerm_kubernetes_cluster` and `azurerm_kubernetes_cluster_node_pool` resources (and fix the error messages).

These values can be set conditionally using `null` rather than `0` like so:

```
locals {
  enable_auto_scale = false
  fixed_count = 2
  min_count   = 1
  max_count   = 4
}

resource "azurerm_kubernetes_cluster_node_pool" "test" {
  # ...
  enable_auto_scaling = local.enable_auto_scale
  node_count          = local.enable_auto_scale ? null : local.fixed_count
  min_count           = local.enable_auto_scale ? local.min_count : null
  max_count           = local.enable_auto_scale ? local.max_count : null
}
```

As such whilst I'd like to thank you for this contribution, since we're looking to take a different approach here I'm going close this PR for the moment - however the changes to the `vnet_subnet_id` field look good, so to be able to get that in I hope you don't mind but I'm going to pull that change into the other (upcoming) PR so that we can ship that fix.

Thanks!

PR #6379, opened 2020-04-07
New resource - `azurerm_servicebus_namespace_network_rule_set`

ArcturusZhang:
Introduces new resource `azurerm_servicebus_namespace_network_rule_set`

Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/3930


azurerm/internal/services/servicebus/resource_arm_servicebus_namespace_network_rule.go
```diff
@@ -0,0 +1,308 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  string(servicebus.Allow),
+				ValidateFunc: validation.StringInSlice([]string{
+					string(servicebus.Allow),
+					string(servicebus.Deny),
+				}, false),
+			},
+
+			"ip_masks": {
```
aristosvo, line 63: I’d expect this variable to be `ip_rules`, based on other `network_rules`. Any reason to prefer `ip_masks`?

ArcturusZhang: Well, this is the REST API property name. And I am wondering does `ip_masks` usually mean these are the IPs that is allowed or in the opposite?
There is also an API for individually modifying this `CreateOrUpdateIPFilterRules`, but I did not take this name, since this name implies that those IPs are filtered out, but actually based on what is showing on the portal, these IPs should be the IPs that are allowed to pass.

aristosvo: `ip_masks` confused me at first glance, so I would argue for `ip_rules`. But what can I say, I opened a PR for a solved problem 🙄

ArcturusZhang: `ip_rules` works for me! I will change this right away

ArcturusZhang:
Thanks @aristosvo for the review! I have changed the `ip_masks` to `ip_rules`, and also, since I just realized that this resource actually contains a lot of network rules, therefore the name `azurerm_servicebus_namespace_network_rule` may not be so suitable for this, therefore I changed the resource name to `azurerm_servicebus_namespace_network_rule_set` (same as the REST API implies) as well as a bunch of file names. 
Please have a look, thanks!

website/docs/r/servicebus_namespace_network_rule_set.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+
+* `resource_group_name` - (Required) Specifies the name of the Resource Group where the ServiceBus Namespace Network Rule Set should exist. Changing this forces a new resource to be created.
+
+* `namespace_name` - (Required) Specifies the ServiceBus Namespace name to which to attach the ServiceBus Namespace Network Rule Set. Changing this forces a new resource to be created.
+
+~> **NOTE:** The ServiceBus Namespace must be `Premium` in order to attach a ServiceBus Namespace Network Rule Set.
+
+* `default_action` - (Optional) Specifies the default action for the ServiceBus Namespace Network Rule Set. Possible values are `Allow` and `Deny`. Defaults to `Deny`.
+
+* `ip_rules` - (Optional) A list of IP filter masks that are added to allow access to the ServiceBus Namespace.
```
aristosvo, line 78: Based on the `network_acls` of `azurerm_key_vault` and the assumption CIDR Blocks work as well:
```suggestion
* `ip_rules` - (Optional) One or more IP Addresses, or CIDR Blocks which should be able to access the ServiceBus Namespace.
```

website/docs/r/servicebus_namespace_network_rule_set.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+
+* `network_rules` - (Optional) One or more `network_rules` blocks defined below.
```
aristosvo, line 80: Small fix
```suggestion
* `network_rules` - (Optional) One or more `network_rules` blocks as defined below.
```

website/docs/r/servicebus_namespace_network_rule_set.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+
+---
+
+A `network_rules` block supports the following:
+
+* `subnet_id` - (Required) The ID of the subnet you want to allow to access the corresponding ServiceBus Namespace.
```
aristosvo, line 86: Just a suggestion, also based on `azurerm_key_vault`
```suggestion
* `subnet_id` - (Required) The Subnet ID which should be able to access this ServiceBus Namespace.
```

website/docs/r/servicebus_namespace_network_rule_set.html.markdown
ArcturusZhang: Done

website/docs/r/servicebus_namespace_network_rule_set.html.markdown
ArcturusZhang: Done

website/docs/r/servicebus_namespace_network_rule_set.html.markdown
ArcturusZhang: Done

ArcturusZhang:
Hi @aristosvo thanks for the review! I have resolved those comments, please have a look

Issue #6376, opened 2020-04-07
Test Defect about `TestAccDataSourceSubnet_basic`


lrxtom2:
Cause of that no relate resource in tf code.
`func testAccDataSourceSubnet_basic(data acceptance.TestData) string {
	template := testAccDataSourceSubnet_template(data)
	return fmt.Sprintf(
%s

data "azurerm_subnet" "test" {
  name                 = azurerm_subnet.test.name
  virtual_network_name = azurerm_subnet.test.virtual_network_name
  resource_group_name  = azurerm_subnet.test.resource_group_name
}
, template)
}`

**Error Message:**
go test ./tests -v -timeout=0 -run TestAccDataSourceSubnet_basic
=== RUN   TestAccDataSourceSubnet_basic
=== PAUSE TestAccDataSourceSubnet_basic
=== CONT  TestAccDataSourceSubnet_basic
--- FAIL: TestAccDataSourceSubnet_basic (0.18s)
    testing.go:640: Step 0 error: config is invalid: 3 problems:
        
        - Reference to undeclared resource: A managed resource "azurerm_subnet" "test" has not been declared in the root module.
        - Reference to undeclared resource: A managed resource "azurerm_subnet" "test" has not been declared in the root module.
        - Reference to undeclared resource: A managed resource "azurerm_subnet" "test" has not been declared in the root module.
FAIL
FAIL	github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/tests	3.376s
FAIL

PR #6381, opened 2020-04-07
Fix test defect `TestAccDataSourceSubnet_basic`

lrxtom2:
Thanks for reviewing. The test `TestAccDataSourceSubnet_basic` have a test defect because of lock off config in tf code. #6376


lrxtom2:
#6376

PR #6386, opened 2020-04-07
New resource: azurerm_managed_application

neil-yechenwei:
This PR is the implement of the issue terraform-providers/terraform-provider-azurerm#5294


azurerm/internal/services/managedapplications/managed_application_resource.go
```diff
@@ -0,0 +1,309 @@
...
+				ValidateFunc: resource.ValidateResourceGroupID,
+			},
+
+			"application_definition_id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validate.ManagedApplicationDefinitionID,
+			},
+
+			"parameters": {
```
katbyte, line 79: Instead of a json string could we use a map of param<> value and generate the json ourselves?

azurerm/internal/services/managedapplications/managed_application_resource.go
```diff
@@ -0,0 +1,309 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"MarketPlace",
+					"ServiceCatalog",
+				}, false),
+			},
+
+			"managed_resource_group_id": {
```
katbyte, line 66: COuld we make this
```suggestion
			"target_resource_group_name": {
```
and generate the id in code?

azurerm/internal/services/managedapplications/managed_application_resource.go
```diff
@@ -0,0 +1,309 @@
...
+			paramsVal := params.(map[string]interface{})
+			for _, v := range paramsVal {
+				if v != nil {
+					delete(v.(map[string]interface{}), "type")
+				}
+			}
+			json, err := structure.FlattenJsonToString(paramsVal)
+			if err != nil {
+				return fmt.Errorf("failed to serialize JSON from Parameters: %+v", err)
+			}
```
katbyte, line 227: Given this is a map already we def shouldn't be converting it to json for terraform

azurerm/internal/services/managedapplications/managed_application_resource.go
```diff
@@ -0,0 +1,309 @@
...
+			"plan": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
```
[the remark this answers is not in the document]
neil-yechenwei: I didn't find validation rule from doc and api error message. So I have to use validation.StringIsNotEmpty for basic validation.

azurerm/internal/services/managedapplications/managed_application_resource.go
```diff
@@ -0,0 +1,309 @@
...
+
+	if v, ok := d.GetOk("application_definition_id"); ok {
+		parameters.ApplicationDefinitionID = utils.String(v.(string))
+	}
+	if v, ok := d.GetOk("plan"); ok {
+		parameters.Plan = expandManagedApplicationPlan(v.([]interface{}))
+	}
+	if v, ok := d.GetOk("parameters"); ok {
+		expandedParams, err := structure.ExpandJsonFromString(v.(string))
+		if err != nil {
```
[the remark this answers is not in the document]
neil-yechenwei: Seems I cannot do this since "expandedParams" variable is used in line 163

azurerm/internal/services/managedapplications/managed_application_resource.go
neil-yechenwei: updated

azurerm/internal/services/managedapplications/managed_application_resource.go
neil-yechenwei: updated

azurerm/internal/services/managedapplications/managed_application_resource.go
neil-yechenwei: updated

neil-yechenwei:
@katbyte , thanks for your comments. I've updated code.

website/docs/r/managed_application.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+* `name` - (Required) Specifies the name of the Managed Application. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the Resource Group where the Managed Application should exist. Changing this forces a new resource to be created.
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `kind` - (Required) The kind of the managed application. Possible values are `MarketPlace` and `ServiceCatalog`.
+
+* `target_resource_group_name` - (Required) The name of the target resource group which includes resources managed by Managed Application.
```
seiffert, line 74: Did you consider naming this `managed_resource_group_name`? I haven't heard this to be referred to as 'target resource group' yet and "managed resource group" would be consistent with `az` as well:

```
$ az managedapp create --help

Command
    az managedapp create : Create a managed application.

Arguments
    --kind              [Required] : The managed application kind. can be marketplace or
                                     servicecatalog.
    --managed-rg-id -m  [Required] : The resource group managed by the managed application.
    --name -n           [Required] : The managed application name.
    --resource-group -g [Required] : The resource group of the managed application.
...
```

website/docs/r/managed_application.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+* `version` - (Required) Specifies the version of the plan from the marketplace.
+
+* `promotion_code` - (Optional) Specifies the promotion code of the plan. 
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The ID of the Managed Application.
```
seiffert, line 102: I believe it would be convenient to expose the app's outputs as well. Do you think we could add that from the beginning?

website/docs/r/managed_application.html.markdown
neil-yechenwei: @katbyte . I think what seiffert suggests is reasonable.  I found generally it would be named as "managed_resource_group_name" in Azure side after went through the related docs. I think it would be more friendly to azure users. How do you think? If it's fine, I will update it to "managed_resource_group_name"?

website/docs/r/managed_application.html.markdown
neil-yechenwei: According by kt's comments previously, suggests to not add outputs from the beginning. Suggests to add them until user asks. Thus we don't need to maintain the outputs user doesn't need whereas outputs is easily to be added.

website/docs/r/managed_application.html.markdown
katbyte: I don't think the API's name is good or clear as to what it is (or if it should be created first) but i'll leave it up to you @neil-yechenwei - please make sure the docs are clear what it means either way

the CLi tool is auto generated from swagger and often carries over tragic names & descriptions @seiffert - we often will change/rename properties so they is more obvious as to what they do ie target resource group implies the resource will act upon that group. However in this case its not as egregous as i've seen elsewhere so i'll leave it up to the PR author.

website/docs/r/managed_application.html.markdown
katbyte: What are the app outputs you would like to see @seiffert and in what instance would they be sueful?

website/docs/r/managed_application.html.markdown
neil-yechenwei: Thanks both. I updated it to "managed_resource_group_name" and updated the document for the description.

website/docs/r/managed_application.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) Specifies the name of the Managed Application. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the Resource Group where the Managed Application should exist. Changing this forces a new resource to be created.
+
+* `location` - (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created.
+
+* `kind` - (Required) The kind of the managed application. Possible values are `MarketPlace` and `ServiceCatalog`. Changing this forces a new resource to be created.
```
katbyte, line 72:
```suggestion
* `kind` - (Required) The kind of the managed application to deply. Possible values are `MarketPlace` and `ServiceCatalog`. Changing this forces a new resource to be created.
```

website/docs/r/managed_application.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+* `managed_resource_group_name` - (Required) The name of the target resource group which holds all the resources that are required by the managed application. Changing this forces a new resource to be created.
```
katbyte, line 74: I think this should be 
```suggestion
* `managed_resource_group_name` - (Required) The name of the target resource group where all the resources deployed by the managed application will reside. Changing this forces a new resource to be created.
```

website/docs/r/managed_application.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+* `application_definition_id` - (Optional) The fully qualified path of managed application definition ID.
```
katbyte, line 76:
```suggestion
* `application_definition_id` - (Optional) The application definition ID to deply.
```

website/docs/r/managed_application.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+* `parameters` - (Optional) The name and value pairs that define the managed application parameters.
```
katbyte, line 78:
```suggestion
* `parameters` - (Optional) A mapping of name and value pairs to pass to the managed application as parameters.
```

website/docs/r/managed_application.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+* `name` - (Required) Specifies the name of the plan from the marketplace.
+
+* `product` - (Required) Specifies the product of the plan from the marketplace.
+
+* `publisher` - (Required) Specifies the publisher of the plan.
+
+* `version` - (Required) Specifies the version of the plan from the marketplace.
+
+* `promotion_code` - (Optional) Specifies the promotion code of the plan.
```
katbyte, line 96:
```suggestion
* `promotion_code` - (Optional) Specifies the promotion code to use with the plan. 
```

website/docs/r/managed_application.html.markdown
neil-yechenwei: "to deply" - > "to deploy", right? If yes, I updated

website/docs/r/managed_application.html.markdown
neil-yechenwei: updated

website/docs/r/managed_application.html.markdown
neil-yechenwei: updated

website/docs/r/managed_application.html.markdown
neil-yechenwei: updated

website/docs/r/managed_application.html.markdown
neil-yechenwei: updated

neil-yechenwei:
@katbyte , thanks for your comments. For test case "TestAccAzureRMManagedApplication_requiresImport" failure, I've fixed. For test case "TestAccAzureRMManagedApplication_complete", I've supplement the description about how to accept terms for the image while plan is specified. So before you run this test case, you have to run command `Set-AzContext -SubscriptionId "<insert your subscription id>"; Get-AzMarketplaceTerms -Publisher "cisco" -Product "meraki-vmx" -Name "meraki-vmx100" | Set-AzMarketplaceTerms -Accept` first.

My test results:
```
=== RUN   TestAccAzureRMManagedApplication_requiresImport
=== PAUSE TestAccAzureRMManagedApplication_requiresImport
=== CONT  TestAccAzureRMManagedApplication_requiresImport
--- PASS: TestAccAzureRMManagedApplication_requiresImport (572.98s)
PASS

=== RUN   TestAccAzureRMManagedApplication_complete
=== PAUSE TestAccAzureRMManagedApplication_complete
=== CONT  TestAccAzureRMManagedApplication_complete
--- PASS: TestAccAzureRMManagedApplication_complete (1416.08s)
PASS
```

website/docs/r/managed_application.html.markdown
seiffert: Sorry for the late response. The app outputs can be anything the app publisher decides to output,  so this would need to be a generic key/value map. 

Outputs could be URLs to a management interface deployed as part of the app or also just configuration values for clients to use the app. E.g. if you would deploy a managed app that contains a database server, the server name could be an output of the app. By making it an attribute of the resource, the server name could be used to provision the databases in that server.

seiffert: Of course, this  functionality can also be developed in a future PR. I was just looking at this for a potential use case of our team where outputs would be good to have.

azurerm/internal/services/managedapplications/managed_application_resource.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	parameters := managedapplications.Application{
+		Location: utils.String(azure.NormalizeLocation(d.Get("location"))),
+		Kind:     utils.String(d.Get("kind").(string)),
+		Tags:     tags.Expand(d.Get("tags").(map[string]interface{})),
+	}
+
+	if v, ok := d.GetOk("managed_resource_group_name"); ok {
+		targetResourceGroupId := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", meta.(*clients.Client).Account.SubscriptionId, v)
```
katbyte, line 148: Is it possible to target a resource group in a different subscription?

azurerm/internal/services/managedapplications/managed_application_resource.go
```diff
@@ -0,0 +1,313 @@
...
+		parameters.ApplicationProperties = &managedapplications.ApplicationProperties{
+			ManagedResourceGroupID: utils.String(targetResourceGroupId),
+		}
+	}
+	if v, ok := d.GetOk("application_definition_id"); ok {
+		a := v.(string)
+		b := utils.String(a)
+		parameters.ApplicationDefinitionID = b
```
katbyte, line 156: could we just do this
```suggestion
		parameters.ApplicationDefinitionID = utils.String(v.(string))
```

azurerm/internal/services/managedapplications/managed_application_resource.go
```diff
@@ -0,0 +1,313 @@
...
+
+	d.Set("name", id.Name)
+	d.Set("resource_group_name", id.ResourceGroup)
+	d.Set("location", location.NormalizeNilable(resp.Location))
+	d.Set("kind", resp.Kind)
+	if err := d.Set("plan", flattenManagedApplicationPlan(resp.Plan)); err != nil {
+		return fmt.Errorf("setting `plan`: %+v", err)
+	}
+	if props := resp.ApplicationProperties; props != nil {
+		parsedManagedResourceGroupID := strings.Split(*props.ManagedResourceGroupID, "/")
```
katbyte, line 224: could we use the resource group parse id function here maybe?

website/docs/r/managed_application.html.markdown
katbyte: ahh thanks for explaining that makes a lot of sense, @neil-yechenwei could we get an `managed_app_outputs` map as suggested?

website/docs/r/managed_application.html.markdown
```diff
@@ -0,0 +1,121 @@
...
+
+* `product` - (Required) Specifies the product of the plan from the marketplace.
+
+* `publisher` - (Required) Specifies the publisher of the plan.
+
+* `version` - (Required) Specifies the version of the plan from the marketplace.
+
+* `promotion_code` - (Optional) Specifies the promotion code to use with the plan.
+
+~> **NOTE:** When `plan` is specified, legal terms must be accepted for this item on this subscription before creating the Managed Application. For example, to accept the terms using PowerShell, please use `Set-AzContext -SubscriptionId "00000000-0000-0000-0000-000000000000"; Get-AzMarketplaceTerms -Publisher "cisco" -Product "meraki-vmx" -Name "meraki-vmx100" | Set-AzMarketplaceTerms -Accept`.
```
katbyte, line 98: could we make this a bit more generic
```suggestion
~> **NOTE:** When `plan` is specified, legal terms must be accepted for this item on this subscription before creating the Managed Application. For example, to accept the terms using PowerShell, please use `Set-AzContext -SubscriptionId "00000000-0000-0000-0000-000000000000"; Get-AzMarketplaceTerms -Publisher "plan-publisher" -Product "plan-product" -Name "plan-name" | Set-AzMarketplaceTerms -Accept`.
```

website/docs/r/managed_application.html.markdown
neil-yechenwei: @katbyte , may I ask you mean I need to add a Computed map attribute `managed_app_outputs` which includes `name\resource_group_name\kind\managed_resource_group_name\application_definition_id\parameters\plan\tags` in resource file, right? If yes, could we add a data source file to output these properties?

azurerm/internal/services/managedapplications/managed_application_resource.go
neil-yechenwei: No. According by previous test results, they must be in the same subscription otherwise it would threw error message which instructs they must be in the same subscription.

azurerm/internal/services/managedapplications/managed_application_resource.go
neil-yechenwei: updated

azurerm/internal/services/managedapplications/managed_application_resource.go
neil-yechenwei: updated

website/docs/r/managed_application.html.markdown
neil-yechenwei: updated

website/docs/r/managed_application.html.markdown
neil-yechenwei: Added

neil-yechenwei:
@katbyte , Thanks for your comments. I've updated code.

azurerm/internal/services/managedapplications/managed_application_resource.go
katbyte: cool thanks!

Issue #6382, opened 2020-04-07
Missing support for route_table on subnet block for azurerm_virtual_network


JSchenken:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Currently the subnet block for azurerm_virtual_network does not provide/expose a route_table property. This prevents the connection of a subnet to a route table. 

### New or Affected Resource(s)

azurerm_virtual_network

### Potential Terraform Configuration

```hcl
resource "azurerm_route_table" "main" {
  name                = "acctestroutetable%d"
  location            = azurerm_resource_group.test.location
  resource_group_name = azurerm_resource_group.test.name

  route {
    name           = "route1"
    address_prefix = "8.0.0.0/8"
    next_hop_type  = "Internet"
  }
}

resource "azurerm_virtual_network" "test" {
  name                = "acctestvirtnet%d"
  address_space       = ["10.0.0.0/16"]
  location            = azurerm_resource_group.test.location
  resource_group_name = azurerm_resource_group.test.name

  subnet {
    name           = "subnet1"
    address_prefix = "10.0.1.0/24"
    route_table    = azurerm_route_table.main.id
  }

  tags = {
    environment = "staging"
  }
}
```

### References

* #0000

bailsman:
Might this be helpful? https://www.terraform.io/docs/providers/azurerm/r/subnet_route_table_association.html

JSchenken:
ah, yes, that work. Just was thrown by having subnet defined within the azurerm_virtual_network but not with same set of properties/capabilities as when defined on its own. Should the subnet property within azurerm_virtual_network be removed?

njuCZ:
@JSchenken Could you please reopen your PR to fix this issue. Though `azurerm_subnet_route_association` could achieve the same goal, there are some other scenario like #7196 and #7586 which could not use this resource. So your PR will help unblock these users.

PR #6394, opened 2020-04-07
Add route_table support for subnet block in azurerm_virtual_network

JSchenken:
Address #6382


JSchenken:
closing this as the attached issue has an alternate approach to accomplish same end result.

PR #6399, opened 2020-04-08
New resource: `azurerm_advisor_suppression`

yupwei68:
Fix: terraform-providers/terraform-provider-azurerm#6119

Dependent PR: [Azure-sdk-for-go version upgrade to v40.6.0](https://github.com/terraform-providers/terraform-provider-azurerm/pull/6398)
advisor suppression pass


yupwei68:
Hi @katbyte ,I have merged master recommendation code. Sorry that I have to change `azurerm_advisor_recommendations` a little bit. Please continue reviewing.
image

azurerm/internal/services/advisor/data_source_advisor_recommendations.go
```diff
@@ -59,7 +60,7 @@ func dataSourceArmAdvisorRecommendations() *schema.Resource {
 							Computed: true,
 						},
 
-						"recommendation_name": {
+						"recommendation_id": {
```
katbyte, line 63: This is a breaking change, we'll have to support both and deprecate the old property - why are we removing the name of the recomendation? seems valuable?

azurerm/internal/services/advisor/validate/advisor.go
```diff
@@ -0,0 +1,49 @@
...
+		return warnings, errors
+	}
+
+	if v <= 0 || v > 24855 {
+		errors = append(errors, fmt.Errorf("%q is expected to be between 0 and 24855 or -1", k))
+		return warnings, errors
+	}
+
+	return warnings, errors
+}
```
katbyte, line 49: can we just use the validation.IntInRange function for this insteal?

azurerm/internal/services/advisor/tests/resource_arm_advisor_suppression_test.go
```diff
@@ -0,0 +1,221 @@
...
+	return nil
+}
+
+func testAccAzureRMAdvisorSuppression_basic(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
+
+data "azurerm_advisor_recommendations" "test" {}
```
katbyte, line 170: for these tests is there a repeatable recommendation we can use rather they relying on the first? also when the resource is deleted will the suppression stop?

azurerm/internal/services/advisor/resource_arm_advisor_suppression.go
```diff
@@ -0,0 +1,162 @@
...
+			},
+
+			"recommendation_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.AdvisorRecommendationID,
+			},
+
+			"duration_days": {
```
katbyte, line 55: could we make this 
```suggestion
			"duration_of_days": {
```
or
```suggestion
			"duration_in_days": {
```

azurerm/internal/services/advisor/data_source_advisor_recommendations.go
yupwei68: I agree. I'll keep them both.

website/docs/r/advisor_suppression.html.markdown
```diff
@@ -22,7 +22,7 @@ data "azurerm_advisor_recommendations" "example" {}
 resource "azurerm_advisor_suppression" "example" {
   name              = "example-as"
   recommendation_id = data.azurerm_advisor_recommendations.example.recommendations.0.recommendation_id
-  duration_in_days     = "3"
+  duration_in_days  = "3"
```
ArcturusZhang, line 25: I suppose this should be an integer, could we please fix this?
```suggestion
  duration_in_days  = 3
```

yupwei68:
@katbyte It's the same problem of the unresolved comment. It could happen that the resource has been deleted, but the recommendation exists for latency (sometimes more than one day). The suppression will return 404 error in this case. I have not found a reserved recommendation for testing. I have asked Mark if there exists reserved recommendation. Other comments have been resolved. Thanks for your comments.

=== RUN   TestAccAzureRMAdvisorSuppression_basic
--- PASS: TestAccAzureRMAdvisorSuppression_basic (38.17s)
=== RUN   TestAccAzureRMAdvisorSuppression_requiresImport
--- PASS: TestAccAzureRMAdvisorSuppression_requiresImport (48.37s)
=== RUN   TestAccAzureRMAdvisorSuppression_complete
--- PASS: TestAccAzureRMAdvisorSuppression_complete (36.46s)
=== RUN   TestAccAzureRMAdvisorSuppression_update
--- PASS: TestAccAzureRMAdvisorSuppression_update (116.30s)
PASS

katbyte:
@yupwei68 - until the test passes reliably this can't be merged. running it again and again its always the complete test that is failing:
```
 Error: failure in retrieving Advisor Suppressions "acctest-sp-200610210226612028": advisor.SuppressionsClient#Get: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: error response cannot be parsed: "" error: EOF    
```
that seems like a bug in the API/SDK if its returning something that is not able to be parsed

yupwei68:
Blocked by issue: https://github.com/Azure/azure-rest-api-specs/issues/9847

katbyte:
As this is blocked i am going to close it until the API is fixed - at which point we can revive the branch in gh

PR #6426, opened 2020-04-09
cosmosdb_mongo_collection - support `index` & `system_index`

### LINKED ISSUE #4460: azurerm_cosmosdb_mongo_collection does not seem to allow compound indexes
### Community Note

### Description

MongoDB API for CosmosDB supports compound indexes (https://docs.microsoft.com/en-us/azure/cosmos-db/mongodb-indexing). However, there does not seem to be a way to create them through Terraform azurerm_cosmosdb_mongo_collection resource.

neil-yechenwei:
Fix the issue of https://github.com/terraform-providers/terraform-provider-azurerm/issues/4460

Fixes #4460


neil-yechenwei:
@katbyte , after tested previously, I found some system keys can be updated and some system keys cannot be updated. So I think maybe we canot simply ignore system keys. After checked azure cli and powershell for this resource, seems they also didn't ignore system keys. So I think we can align with them. Does it make sense?

katbyte:
@neil-yechenwei - however we do it, maybe separate `index` and `index_system` properties, we should always reading all properties back on read.

neil-yechenwei:
@katbyte , because some system indexes can be updated. If I move them to computed index_system property, it would cause diff on index property. So seems I cannot do this. After double confirmed with service owner, seems only system indexes "_id" and "DocumentDBDefaultIndex" cannot be updated, other system indexes can be updated. So I ignore them.

katbyte:
@neil-yechenwei, we should always be reading in the values the user can set. weather we seperate user settable and system into `indexes` & a computed `system_indexes` ot just ignore the system ones we should be ensure the user's are read back in from the API.

neil-yechenwei:
@katbyte , I just want to double confirm you mean user settable indexes which include user customized index and settable system index should be moved into property "indexes". The system index which is not settable should be moved into "system_indexes", right?

azurerm/internal/services/cosmos/resource_arm_cosmosdb_mongo_collection.go
```diff
@@ -82,6 +82,45 @@ func resourceArmCosmosDbMongoCollection() *schema.Resource {
...
+						"unique": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
+					},
+				},
+			},
+
+			"system_index": {
```
katbyte, line 106: I think this should be `system_indexes` as it'll never be used as seperate blocks?

neil-yechenwei: updated

neil-yechenwei:
@katbyte , thanks for your comments. I've updated code.

azurerm/internal/services/cosmos/resource_arm_cosmosdb_mongo_collection.go
```diff
@@ -264,7 +303,16 @@ func resourceArmCosmosDbMongoCollectionRead(d *schema.ResourceData, meta interfa
 		}
 
 		if props.Indexes != nil {
```
tombuildsstuff, line 305: we need to ensure these fields are always set, so we can remove the if statement here, since the `flatten` function should handle these being nil

azurerm/internal/services/cosmos/resource_arm_cosmosdb_mongo_collection.go
```diff
@@ -321,24 +384,59 @@ func expandCosmosMongoCollectionIndexes(defaultTtl *int) *[]documentdb.MongoInde
...
-	return &outputs
+	return &results
 }
 
-func flattenCosmosMongoCollectionIndexes(indexes *[]documentdb.MongoIndex) *int {
-	var ttl int
-	for _, i := range *indexes {
-		if key := i.Key; key != nil {
-			var ttlInner int32
+func flattenCosmosMongoCollectionIndex(input *[]documentdb.MongoIndex) (*[]map[string]interface{}, *[]map[string]interface{}, *int32) {
```
tombuildsstuff, line 390: this method needs to account for `input` being nil here

azurerm/internal/services/cosmos/resource_arm_cosmosdb_mongo_collection.go
```diff
@@ -321,24 +384,59 @@ func expandCosmosMongoCollectionIndexes(defaultTtl *int) *[]documentdb.MongoInde
...
+	indexes := make([]map[string]interface{}, 0)
+	systemIndexes := make([]map[string]interface{}, 0)
+	var ttl *int32
+
+	for _, v := range *input {
+		index := map[string]interface{}{}
+		systemIndex := map[string]interface{}{}
+
+		if v.Key != nil && v.Key.Keys != nil {
+			key := (*v.Key.Keys)[0]
```
tombuildsstuff, line 400: there's a crash here if there's 0 items in this

website/docs/r/cosmosdb_mongo_collection.html.markdown
```diff
@@ -45,14 +45,33 @@ The following arguments are supported:
...
 
+* `system_indexes` - One or more `system_indexes` blocks as defined below.
+
+---
+
+The `system_indexes` block supports the following:
+
+* `keys` - The list of system keys which are not settable for each Cosmos DB Mongo Collection.
+
+* `unique` - The unique status of the index.
```
tombuildsstuff, line 73: can we add a better description here?

azurerm/internal/services/cosmos/resource_arm_cosmosdb_mongo_collection.go
neil-yechenwei: it always wouldn't be 0 since api always return system key "_id".

website/docs/r/cosmosdb_mongo_collection.html.markdown
neil-yechenwei: updated

azurerm/internal/services/cosmos/resource_arm_cosmosdb_mongo_collection.go
neil-yechenwei: updated

azurerm/internal/services/cosmos/resource_arm_cosmosdb_mongo_collection.go
neil-yechenwei: updated

neil-yechenwei:
@tombuildsstuff , thanks for your comments. I've updated code.

PR #6428, opened 2020-04-09
Update Azure Policy Apply Example

julie-ng:
Azure schema changed and JSON needs to use `listOfAllowedLocations` instead of `allowedLocations`. Otherwise you get this error:

```
Error: policy.AssignmentsClient#Create: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="MissingPolicyParameter" Message="The policy assignment 'allowed_locations' is missing the parameter(s) 'listOfAllowedLocations' as defined in the policy definition 'e56962a6-4747-49cd-b67b-bf8b01975c4c'."
```
That policy definition refers to a built-in Azure Policy - see https://docs.microsoft.com/en-us/azure/governance/policy/samples/built-in-policies#general

### DIFF
--- a/website/docs/r/policy_assignment.html.markdown
+++ b/website/docs/r/policy_assignment.html.markdown
@@ -63,7 +63,7 @@ resource "azurerm_policy_assignment" "example" {
 
   parameters = <<PARAMETERS
 {
-  "allowedLocations": {
+  "listOfAllowedLocations": {
     "value": [ "West Europe" ]
   }
 }


website/docs/r/policy_assignment.html.markdown
```diff
@@ -63,7 +63,7 @@ resource "azurerm_policy_assignment" "example" {
 
   parameters = <<PARAMETERS
 {
-  "allowedLocations": {
+  "listOfAllowedLocations": {
```
tombuildsstuff, line 66: does this also need to be updated above [here](https://github.com/terraform-providers/terraform-provider-azurerm/pull/6428/files#diff-af242adb59cafee47d7689f105331f02L39) and [here](https://github.com/terraform-providers/terraform-provider-azurerm/pull/6428/files#diff-af242adb59cafee47d7689f105331f02R27) too?

katbyte: I think we'll need to change it more places, the policy in the example here above still uses the old name. We might want to change all our custom policies in tests & examples to use the new builtin name?

katbyte:
hey @julie-ng,

Given that this property name depends very much on the supplied policy (and custom ones everywhere in the provider tests & docs don't use this new name) and this would break the example i'm going to close this PR. I think we'll need to change all of the test & example policies to make this change to match the new built in property. Thanks again for the PR and feel free to reopen this with the those changes!

PR #6438, opened 2020-04-10
azurerm_sql_server - support for `connection_policy` property

katbyte:


azurerm/internal/services/sql/resource_arm_sql_server.go
```diff
@@ -71,9 +71,15 @@ func resourceArmSqlServer() *schema.Resource {
 				Sensitive: true,
 			},
 
-			"fully_qualified_domain_name": {
+			"connection_policy": {
```
mbfrahry, line 74: Can we add a test to confirm this works?

PR #6455, opened 2020-04-13
azurerm_network_interface: expose internal_domain_name_suffix

bailsman:
After a network interface has been created in a subnet, Azure assigns it
an internal domain name and exposes this as a property.

The suffix looks something like this:

tcjzv0sul21rwyt5qe31ylgpgo.ax.internal.cloudapp.net

Servers can reach each other internally at, for example:

server0.tcjzv0sul21rwyt5qe31ylgpgo.ax.internal.cloudapp.net

This change exposes the azure provided internal_domain_name_suffix in terraform.

This is useful, for example, when creating a self-signed server certificate to be used internally, if you do want the common name to match the hostname that the other server connects on.

### DIFF
--- a/azurerm/internal/services/network/resource_arm_network_interface.go
+++ b/azurerm/internal/services/network/resource_arm_network_interface.go
@@ -140,6 +140,11 @@ func resourceArmNetworkInterface() *schema.Resource {
 				ValidateFunc: validation.StringIsNotEmpty,
 			},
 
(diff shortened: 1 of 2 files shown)


bailsman:
Oops. I force pushed and may have disappeared your comments, so now I'm not sure what changes you requested. I added a line to the resource documentation explaining the attribute that is now exposed.

PR #6493, opened 2020-04-15
`azurerm_subnet`: Support for multiple prefixes with `address_prefixes` 

fabianofranz:
**Splitting https://github.com/terraform-providers/terraform-provider-azurerm/pull/6148 into separate PR's**

The azure subnet resource supports addressPrefix when one prefix is available and addressPrefixes when two or or more are set (for instance, when configuring an ipv4 and ipv6 dual stack configuration for a network). Update azurerm_subnet to allow either address_prefix or address_prefixes as input, where the latter is a list of string. Check that at least one is set, and mark the older address_prefix as deprecated.

The data source for subnet is modified to always return address_prefixes to simplify retrieval (although this could be changed to always return the first address prefix as address_prefix).

Today, if two prefixes are specified via CLI or Azure UI the data source returns no address_prefix and it is not possible to create a new subnet with two prefixes via Terraform.

Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/5140

### DIFF
--- a/azurerm/internal/services/network/data_source_subnet.go
+++ b/azurerm/internal/services/network/data_source_subnet.go
@@ -40,6 +40,12 @@ func dataSourceArmSubnet() *schema.Resource {
 				Computed: true,
 			},
 
+			"address_prefixes": {
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Computed: true,
+			},
+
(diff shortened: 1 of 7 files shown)


fabianofranz:
@mbfrahry PTAL

fabianofranz:
@mbfrahry oops, my bad, thanks for pointing out. Fixed.

PR #6495, opened 2020-04-16
Add support for HyperVGeneration property of shared image

neil-yechenwei:
Fix the issue of terraform-providers/terraform-provider-azurerm#6316


manicminer:
Hi @neil-yechenwei, thanks for this PR! Coindicentally, I've also worked up a similar change for the same attribute, so I'll go ahead and merge your changes into mine and submit for review, as it looks like you have a workaround for the nested deletion issue.

manicminer:
Resubmitted as #6511

PR #6530, opened 2020-04-17
Enable AKS  AAD integration v2 + SDK version bump

jlpedrosa:
Microsoft introduced a new version of the[ integration between AKS and AAD for RBAC](https://docs.microsoft.com/en-us/azure/aks/managed-aad). 

This eliminates the need of creating specific SP for AAD and the required set of permissions (including admin consent).

This change required version bump of the containers to 2020-03, which introduced an small change in node pool, I've made the change in a way that is backwards compatible. 

I just ran acceptance test on the new test that I've created. 

Help on how to activate the feature in a subscription, as it is preview:

```
az feature register --name AAD-V2 --namespace Microsoft.ContainerService
```
It might take several minutes for the status to show as Registered. You can check the registration status by using the az feature list command:
```
az feature list -o table --query "[?contains(name, 'Microsoft.ContainerService/AAD-V2')].{Name:name,State:properties.state}"
```
When the status shows as registered, refresh the registration of the Microsoft.ContainerService resource provider by using the az provider register command:

```
az provider register --namespace Microsoft.ContainerService
```


vendor/github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2020-03-01/containerservice/models.go
```diff
@@ -2328,6 +2399,7 @@ func (mcp ManagedClusterProperties) MarshalJSON() ([]byte, error) {
 // ManagedClusterPropertiesAutoScalerProfile parameters to be applied to the cluster-autoscaler when
 // enabled
 type ManagedClusterPropertiesAutoScalerProfile struct {
+	BalanceSimilarNodeGroups      *bool   `json:"balance-similar-node-groups,omitempty"`
```
aristosvo, line 2402: AFAIK is this not the latest version, this should be a *string due to the latest 0.42 release, fixed here: https://github.com/Azure/azure-rest-api-specs/pull/8919#issuecomment-621453037
```suggestion
	BalanceSimilarNodeGroups      *string   `json:"balance-similar-node-groups,omitempty"`
```

jlpedrosa: Hi @aristosvo 
I don't think @tombuildsstuff or @katbyte would find ok to update the SDK version in this PR.
If they are, happy to do so.

jlpedrosa:
Rebased

vendor/github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2020-03-01/containerservice/models.go
katbyte: +1, we will never update something in the vendor folder. i will be updating the go sdk version to 42 later today/tomorrow however

aristosvo: @katbyte With regard to bumping the AKS API version, is that also a separate PR?

jlpedrosa:
Rebased

jlpedrosa:
Rebased

PR #6538, opened 2020-04-20
app_service_environment - support for `location` property

neil-yechenwei:
Fix the issue of terraform-providers/terraform-provider-azurerm#6532


neil-yechenwei:
@katbyte , actually doc has already added `location` property.
See the doc: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/docs/d/app_service_environment.html.markdown

PR #6550, opened 2020-04-20
CDN Endpoint: make origin_host_header required 

r0bnet:
When omitting this property the API is always the following error:

```
Error: Error waiting for CDN Endpoint "example" (Profile "example-cdn" / Resource Group "test-rg") to finish creating: Code="BadRequest" Message="{\"ErrorMessage\":\"Errors found in Model: OriginHostHeader must match the regex '(^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9_\\\\-]*[a-zA-Z0-9])\\\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\\\-]*[A-Za-z0-9])$)|((?:[:0-9A-Fa-f]+))'.\"}"
```

Didn't want to do local validation with the above mentioned regex but to simply make this value required instead of optional.

### DIFF
--- a/azurerm/internal/services/cdn/resource_arm_cdn_endpoint.go
+++ b/azurerm/internal/services/cdn/resource_arm_cdn_endpoint.go
@@ -58,8 +58,7 @@ func resourceArmCdnEndpoint() *schema.Resource {
 
 			"origin_host_header": {
 				Type:     schema.TypeString,
-				Optional: true,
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
Ignoring a couple of temporarily failures within Azure - the tests look good 👍

PR #6565, opened 2020-04-21
azurerm_virtual_network_gateway_connection  - shared_key is optional when type is IPSec

neil-yechenwei:
fixes terraform-providers/terraform-provider-azurerm#5402


azurerm/internal/services/network/resource_arm_virtual_network_gateway_connection.go
```diff
@@ -510,23 +510,19 @@ func getArmVirtualNetworkGatewayConnectionProperties(d *schema.ResourceData) (*n
...
+			return nil, fmt.Errorf("`express_route_circuit_id` must be specified when `type` is set to `ExpressRoute`")
 		}
 	}
 
 	if props.ConnectionType == network.IPsec {
 		if props.LocalNetworkGateway2 == nil || props.LocalNetworkGateway2.ID == nil {
-			return nil, fmt.Errorf("`local_network_gateway_id` and `shared_key` must be specified when `type` is set to `IPsec")
-		}
-
-		if props.SharedKey == nil {
```
tombuildsstuff, line 522: can we add an acceptance test to ensure this use-case is supported through the API? Unfortunately ARM Templates can behave differently to the API's themselves - so an acceptance test should confirm this

neil-yechenwei: Fixed

neil-yechenwei:
@tombuildsstuff , Thanks for your suggestion. I've added test case to cover this use-case.

PR #6574, opened 2020-04-22
`azurerm_private_endpoint`: Add support for tags

neil-yechenwei:
Fix the issue of terraform-providers/terraform-provider-azurerm#6544


WodansSon:
image

JamesDLD:
Does this also tags the network interface?

PR #6586, opened 2020-04-22
add traffic_selector_policy block and tests

carloVentrella:
Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/6272

### DIFF
--- a/azurerm/internal/services/network/tests/virtual_network_gateway_connection_resource_test.go
+++ b/azurerm/internal/services/network/tests/virtual_network_gateway_connection_resource_test.go
@@ -113,6 +113,27 @@ func TestAccAzureRMVirtualNetworkGatewayConnection_ipsecpolicy(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMVirtualNetworkGatewayConnection_trafficselectorpolicy(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_virtual_network_gateway_connection", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMVirtualNetworkGatewayConnectionDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMVirtualNetworkGatewayConnection_trafficselectorpolicy(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVirtualNetworkGatewayConnectionExists(data.ResourceName),
+					resource.TestCheckResourceAttr(data.ResourceName, "traffic_selector_policy.0.local_address_ranges.0", "10.66.18.0/24"),
+					resource.TestCheckResourceAttr(data.ResourceName, "traffic_selector_policy.0.local_address_ranges.1", "10.66.17.0/24"),
+					resource.TestCheckResourceAttr(data.ResourceName, "traffic_selector_policy.0.remote_address_ranges.0", "10.1.1.0/24"),
+				),
+			},
+		},
+	})
+}
+
 func TestAccAzureRMVirtualNetworkGatewayConnection_connectionprotocol(t *testing.T) {
 	expectedConnectionProtocol := "IKEv1"
 	data := acceptance.BuildTestData(t, "azurerm_virtual_network_gateway_connection", "test")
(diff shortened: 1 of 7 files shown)


azurerm/internal/services/network/data_source_virtual_network_gateway_connection.go
```diff
@@ -109,6 +109,29 @@ func dataSourceArmVirtualNetworkGatewayConnection() *schema.Resource {
 				Computed: true,
 			},
 
+			"traffic_selector_policy": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"local_address_ranges": {
```
katbyte, line 117: would this make more sense as
```suggestion
						"local_address_cidrs": {
```

to be explicit on what format it is expecting?

azurerm/internal/services/network/data_source_virtual_network_gateway_connection.go
```diff
@@ -109,6 +109,29 @@ func dataSourceArmVirtualNetworkGatewayConnection() *schema.Resource {
...
+							Type:     schema.TypeList,
+							Computed: true,
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+						"remote_address_ranges": {
```
katbyte, line 124: same here

azurerm/internal/services/network/data_source_virtual_network_gateway_connection.go
```diff
@@ -109,6 +109,29 @@ func dataSourceArmVirtualNetworkGatewayConnection() *schema.Resource {
 				Computed: true,
 			},
 
+			"traffic_selector_policy": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
```
katbyte, line 115: can there be more then 1 here? or should this me maxitems of 1

azurerm/internal/services/network/data_source_virtual_network_gateway_connection.go
carloVentrella: I used `local_address_ranges` to respect azure names (https://docs.microsoft.com/en-us/powershell/module/az.network/new-azipsectrafficselectorpolicy?view=azps-3.8.0&viewFallbackFrom=azps-3.6.1).
Also, the format can be better explicited in the documentation.

azurerm/internal/services/network/data_source_virtual_network_gateway_connection.go
carloVentrella: Yep, you're right. I'll add `MaxItems: 1`

carloVentrella:
@katbyte I integrated your suggestions. Thanks

carloVentrella:
@katbyte is it possible to merge this for the next release?
Thank you

azurerm/internal/services/network/tests/virtual_network_gateway_connection_resource_test.go
```diff
@@ -113,6 +113,27 @@ func TestAccAzureRMVirtualNetworkGatewayConnection_ipsecpolicy(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMVirtualNetworkGatewayConnection_trafficselectorpolicy(t *testing.T) {
```
tombuildsstuff, line 116: _minor_ but could we update the test name to match the other resources?

```suggestion
func TestAccAzureRMVirtualNetworkGatewayConnection_trafficSelectorPolicy(t *testing.T) {
```

azurerm/internal/services/network/virtual_network_gateway_connection_data_source.go
```diff
@@ -249,3 +278,18 @@ func flattenArmVirtualNetworkGatewayConnectionDataSourceIpsecPolicies(ipsecPolic
...
 }
+
+func flattenArmVirtualNetworkGatewayConnectionDataSourcePolicyTrafficSelectors(trafficSelectorPolicies *[]network.TrafficSelectorPolicy) []interface{} {
+	schemaTrafficSelectorPolicies := make([]interface{}, 0)
+
+	if trafficSelectorPolicies != nil {
+		for _, trafficSelectorPolicy := range *trafficSelectorPolicies {
+			schemaTrafficSelectorPolicy := make(map[string]interface{})
+			schemaTrafficSelectorPolicy["local_address_cidrs"] = trafficSelectorPolicy.LocalAddressRanges
+			schemaTrafficSelectorPolicy["remote_address_cidrs"] = trafficSelectorPolicy.RemoteAddressRanges
```
tombuildsstuff, line 289: since these are an array of `*[]string`:

```
// TrafficSelectorPolicy an traffic selector policy for a virtual network gateway connection.
type TrafficSelectorPolicy struct {
	// LocalAddressRanges - A collection of local address spaces in CIDR format.
	LocalAddressRanges *[]string `json:"localAddressRanges,omitempty"`
	// RemoteAddressRanges - A collection of remote address spaces in CIDR format.
	RemoteAddressRanges *[]string `json:"remoteAddressRanges,omitempty"`
}
```

we need to convert these to a list of strings, which we can do using `utils.FlattenStringSlice` - would you mind updating this to be:

```suggestion
			schemaTrafficSelectorPolicies = append(schemaTrafficSelectorPolicies, map[string]interface{}{
				"local_address_cidrs": utils.FlattenStringSlice(trafficSelectorPolicy.LocalAddressRanges),
				"remote_address_cidrs": utils.FlattenStringSlice(trafficSelectorPolicy.RemoteAddressRanges),
			})

```

this helps avoid issues when the Azure API returns nil for these properties, which can happen with resources which haven't been updated

azurerm/internal/services/network/virtual_network_gateway_connection_resource.go
```diff
@@ -394,6 +418,13 @@ func resourceArmVirtualNetworkGatewayConnectionRead(d *schema.ResourceData, meta
 		}
 	}
 
+	if conn.TrafficSelectorPolicies != nil {
```
tombuildsstuff, line 421: we need to ensure that `traffic_selector_policy` is set at all times - as such can we remove this? the flatten function should return an empty list if this is nil/empty - which should then show a diff

```suggestion
```

azurerm/internal/services/network/virtual_network_gateway_connection_resource.go
```diff
@@ -587,6 +622,33 @@ func expandArmVirtualNetworkGatewayConnectionIpsecPolicies(schemaIpsecPolicies [
...
+				localAddressRangesArr = append(localAddressRangesArr, l.(string))
+			}
+			trafficSelectorPolicy.LocalAddressRanges = &localAddressRangesArr
+		}
+		if remoteAddressRanges, ok := schemaTrafficSelectorPolicy["remote_address_cidrs"].([]interface{}); ok {
+			remoteAddressRangesArr := make([]string, 0, len(remoteAddressRanges))
+			for _, l := range remoteAddressRanges {
+				remoteAddressRangesArr = append(remoteAddressRangesArr, l.(string))
+			}
+			trafficSelectorPolicy.RemoteAddressRanges = &remoteAddressRangesArr
```
tombuildsstuff, line 643: we should be able to use `utils.ExpandStringSlice` to do the same here fwiw

azurerm/internal/services/network/virtual_network_gateway_connection_resource.go
```diff
@@ -587,6 +622,33 @@ func expandArmVirtualNetworkGatewayConnectionIpsecPolicies(schemaIpsecPolicies [
...
+
+	for _, d := range schemaTrafficSelectorPolicies {
+		schemaTrafficSelectorPolicy := d.(map[string]interface{})
+		trafficSelectorPolicy := &network.TrafficSelectorPolicy{}
+		if localAddressRanges, ok := schemaTrafficSelectorPolicy["local_address_cidrs"].([]interface{}); ok {
+			localAddressRangesArr := make([]string, 0, len(localAddressRanges))
+			for _, l := range localAddressRanges {
+				localAddressRangesArr = append(localAddressRangesArr, l.(string))
+			}
+			trafficSelectorPolicy.LocalAddressRanges = &localAddressRangesArr
```
tombuildsstuff, line 636: we should be able to use `utils.ExpandStringSlice` to do the same here fwiw

azurerm/internal/services/network/virtual_network_gateway_connection_resource.go
```diff
@@ -615,3 +677,18 @@ func flattenArmVirtualNetworkGatewayConnectionIpsecPolicies(ipsecPolicies *[]net
...
 }
+
+func flattenArmVirtualNetworkGatewayConnectionTrafficSelectorPolicies(trafficSelectorPolicies *[]network.TrafficSelectorPolicy) []interface{} {
+	schemaTrafficSelectorPolicies := make([]interface{}, 0)
+
+	if trafficSelectorPolicies != nil {
+		for _, trafficSelectorPolicy := range *trafficSelectorPolicies {
+			schemaTrafficSelectorPolicy := make(map[string]interface{})
+			schemaTrafficSelectorPolicy["local_address_cidrs"] = trafficSelectorPolicy.LocalAddressRanges
+			schemaTrafficSelectorPolicy["remote_address_cidrs"] = trafficSelectorPolicy.RemoteAddressRanges
```
tombuildsstuff, line 688: (same as above) - could we update this to use the `flatten` methods, which handles Azure returning nil for these properties

website/docs/d/virtual_network_gateway_connection.html.markdown
```diff
@@ -72,6 +72,10 @@ output "virtual_network_gateway_connection_id" {
     Only a single policy can be defined for a connection. For details on
     custom policies refer to [the relevant section in the Azure documentation](https://docs.microsoft.com/en-us/azure/vpn-gateway/vpn-gateway-ipsecikepolicy-rm-powershell).
 
+* `traffic_selector_policy` (Optional) A `traffic_selector_policy` which allows to specify traffic selector policy proposal to be used in a virtual network gateway connection.
+    Only one block can be defined for a connection.
+    For details about traffic selectors refer to [the relevant section in the Azure documentation](https://docs.microsoft.com/en-us/azure/vpn-gateway/vpn-gateway-connect-multiple-policybased-rm-ps).
```
tombuildsstuff, line 77: can we remove "optional" from this, since this a read-only value (in the Data Source) - the reference in `ipsec_policy` above is a bug in the documentation unfortunately

could we also add this to [the Resource documentation too](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/docs/r/virtual_network_gateway_connection.html.markdown)?

manicminer:
@carloVentrella thanks for the changes. this LGTM, just waiting on test results.

manicminer:
Tests pass (1 race condition passed 2nd time):

Screenshot 2020-07-13 15 59 04
Screenshot 2020-07-13 15 59 58
Screenshot 2020-07-13 16 22 08

PR #6589, opened 2020-04-23
`azurerm_public_ip_prefix`: Update `prefix_length` validation to accept all valid IPv4 address ranges

WodansSon:
[Feb 11, 2020](https://github.com/terraform-providers/terraform-provider-azurerm/pull/5693) there was a change to the resource to restrict the range to be between 28 and 31 due to the product documentation which states that by default [/28](https://docs.microsoft.com/en-us/azure/virtual-network/public-ip-address-prefix#benefits) is the maximum limit, however also in the documentation it states that you should contact customer support to [change the default limit of /28](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits?toc=/azure/virtual-network/toc.json#publicip-address).

Since there is no way for terraform to know the approved range for a given subscription I have opened up the validation to check that the passed value is between `0` and `31`, which should cover all possible IPv4 scenarios.


WodansSon:
New test failed as expected due to subscription public ip limits
image

Issue #5673, opened 2020-02-11
Terraform wants to change azurerm_monitor_diagnostic_setting log category settings


Tommy-Ten:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.12.20
+ provider.azurerm v1.41.0

### Affected Resource(s)

* `azurerm_monitor_diagnostic_setting`

### Terraform Configuration Files

```
resource "azurerm_monitor_diagnostic_setting" "export_activity_logs" {
  name                           = var.actLogName
  target_resource_id             = data.azurerm_subscription.current.id
  storage_account_id             = data.azurerm_storage_account.logging_activity_logs_archive_storage_account.id
  eventhub_authorization_rule_id = azurerm_eventhub_namespace_authorization_rule.eventhub_auth_rule.id

  log {
    category = "Administrative"
    enabled  = true

    retention_policy {
      enabled = true
      days    = 365
    }
  }

  log {
    category = "Security"
    enabled  = false

    retention_policy {
      enabled = false
    }
  }

  log {
    category = "ServiceHealth"
    enabled  = false

    retention_policy {
      enabled = false
    }
  }

  log {
    category = "Alert"
    enabled  = false

    retention_policy {
      enabled = false
    }
  }

  log {
    category = "Recommendation"
    enabled  = false

    retention_policy {
      enabled = false
    }
  }

  log {
    category = "Policy"
    enabled  = false

    retention_policy {
      enabled = false
    }
  }

  log {
    category = "Autoscale"
    enabled  = false

    retention_policy {
      enabled = false
    }
  }

  log {
    category = "ResourceHealth"
    enabled  = false

    retention_policy {
      enabled = false
    }
  }
}
```
### Expected Behavior

After an initial `terraform apply,` when I run `terraform plan` or another `terraform apply`,  I should see no changes.

### Actual Behavior

After an initial `terraform apply,` when I run `terraform plan` or another `terraform apply`,  I see settings for log categories that I defined in my configuration being changed.

```
Resource actions are indicated with the following symbols:
  ~ update in-place

[...]

PR #6603, opened 2020-04-24
azurerm_monitor_diagnostic_setting  - mark retention_policy & retention_policy  as optional 

nyuen:
modifying the Diagnostic settings schema to set the retention policy to Optional #5673

Some resources such as ActivityLogs do not support Retention Policy causing a change on subsequent Terraform apply even if the diagnostic setting. This is caused by a mismatch between the TFState and the object returned by the REST API


nyuen:
thanks @katbyte, looks like the linter failed, I guess len(policiesRaw) != 0 should be sufficient.
One a side note do you have more inputs one how the tests are run? I've tried to added a failing this and go test monitor resource_arm_monitor_diagnostic_setting_test.go still passed somehow.

katbyte:
heh thats why we have linters 🙂 - and sadly the api can returninconsistentt results depending on many things including region and timing. pushed a commit which should fix that up

PR #6604, opened 2020-04-24
`azurerm_frontdoor` - Add support for `backend_pools_send_receive_timeout_seconds`

### LINKED ISSUE #4558: Azure Frontdoor: support custom timeout for backend pool
### Community Note

### Description

Azure Frontdoor supports setting a custom timeout for the backend. The standard timeout is 30s which is quite short.

See also:
https://feedback.azure.com/forums/217313-networking/suggestions/36482500-allow-configurable-timeout-period-for-front-door
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/frontdoor/resource-manager/Microsoft.Network/stable/2019-05-01/frontdoor.json

In the last link there are 2 backend pool settings:
1. enforceCertificateNameCheck
2. sendRecvTimeoutSeconds

(1) is supported by terraform using the `enforce_backend_pools_certificate_name_check` directive.

Please add a directive to support `sendRecvTimeoutSeconds`. I'm not sure if this directive applies to all backend pools or only a specified one.

* azurerm_frontdoor

```hcl
resource "azurerm_frontdoor" "example" {
  name                                         = "example-FrontDoor"
  location                                     = "${azurerm_resource_group.example.location}"
  resource_group_name                          = "${azurerm_resource_group.example.name}"
  enforce_backend_pools_certificate_name_check = false
  backend_pools_request_timeout = "60"
...
```

### References

https://feedback.azure.com/forums/217313-networking/suggestions/36482500-allow-configurable-timeout-period-for-front-door
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/frontdoor/resource-manager/Microsoft.Network/stable/2019-05-01/frontdoor.json

WodansSon:
(fixes #4558)


azurerm/internal/services/frontdoor/resource_arm_frontdoor.go
```diff
@@ -67,6 +67,13 @@ func resourceArmFrontDoor() *schema.Resource {
 				Required: true,
 			},
 
+			"backend_pools_send_receive_timeout_seconds": {
```
katbyte, line 70: Should this be `backend_pool_send_receive_timeout_in_seconds`?

azurerm/internal/services/frontdoor/resource_arm_frontdoor.go
```diff
@@ -841,6 +855,7 @@ func expandArmFrontDoorBackendPoolsSettings(enforceCertificateNameCheck bool) *f
 
 	result := frontdoor.BackendPoolsSettings{
 		EnforceCertificateNameCheck: enforceCheck,
+		SendRecvTimeoutSeconds:      utils.Int32(backendPoolsSendReceiveTimeoutSeconds),
 	}
```
katbyte, line 859: Are there more settings here? would it make sense to move this into a `backend_pool_configuration` block?

azurerm/internal/services/frontdoor/resource_arm_frontdoor.go
WodansSon: I was originally going to call it that too, but after researching it, it's a global setting for [all backend pools](https://docs.microsoft.com/en-us/rest/api/frontdoorservice/frontdoor/frontdoors/get#backendpoolssettings) thus the plural. 🙂

azurerm/internal/services/frontdoor/resource_arm_frontdoor.go
WodansSon: Again, I was thinking the same thing... but I wanted to avoid a breaking change so I just put it at the top level with the other backend pools settings. Not sure if we should pull it into a block yet, if another setting is exposed in a future API I think it would be appropriate to make the change at that time... Maybe open a 3.0 breaking change issue for this to fix in the future?

katbyte: well we now have 2 so it makes sense to move into a block (especially when we have such long names) if we did it in 3.0 we would still ideally want to depreacate the old properties & support both so people have a migration path.

azurerm/internal/services/frontdoor/resource_arm_frontdoor.go
katbyte: ahh see i would have assumed it was applying to all of them without the s, but there is another property backend_pool? which is different the the other back end pools i can see why heh

WodansSon:
image

PR #6607, opened 2020-04-24
Update servicebus name validation

neil-yechenwei:
fixes terraform-providers/terraform-provider-azurerm#5639

### DIFF
--- a/azurerm/helpers/azure/servicebus.go
+++ b/azurerm/helpers/azure/servicebus.go
@@ -12,13 +12,6 @@ import (
 )
 
 // validation
-func ValidateServiceBusNamespaceName() schema.SchemaValidateFunc {
-	return validation.StringMatch(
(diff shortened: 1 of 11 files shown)


azurerm/helpers/azure/servicebus.go
```diff
@@ -12,13 +12,6 @@ import (
 )
 
 // validation
-func ValidateServiceBusNamespaceName() schema.SchemaValidateFunc {
-	return validation.StringMatch(
-		regexp.MustCompile("^[a-zA-Z][-a-zA-Z0-9]{4,48}[a-zA-Z0-9]$"),
-		"The namespace name can contain only letters, numbers, and hyphens. The namespace must start with a letter, and it must end with a letter or number and be between 6 and 50 characters long.",
-	)
-}
-
```
aristosvo, line 21: Is there a reason why you moved it away from here? Why not just implement this code here?

katbyte: we are moving service specific validation to be in the service package with the resource files, this helper azure package will eventually be removed

PR #6628, opened 2020-04-26
`azurerm_storage_account` - Add support for `change_feed` 

### LINKED ISSUE #6627: Support for Change Feed
### Community Note

### Description
The purpose of the change feed is to provide transaction logs of all the changes that occur to the blobs and the blob metadata in your storage account. The change feed provides ordered, guaranteed, durable, immutable, read-only log of these changes. Client applications can read these logs at any time, either in streaming or in batch mode. The change feed enables you to build efficient and scalable solutions that process change events that occur in your Blob Storage account at a low cost.

* azurerm_storage_account

```hcl
resource "azurerm_storage_account" "test" {
  name                = "acc123"

  location                 = azurerm_resource_group.test.location

  blob_properties {
    cors_rule {
      allowed_origins    = ["http://www.example.com"]
      exposed_headers    = ["x-tempo-*"]
      allowed_headers    = ["x-tempo-*"]
      allowed_methods    = ["GET", "PUT"]
      max_age_in_seconds = "500"
    }

    change_feed {
      enabled = true
    }

    delete_retention_policy {
      days = 300
    }
  }
}
```

### References

* https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-change-feed?tabs=azure-portal

lrxtom2:
Fixes #6627

### DIFF
--- a/azurerm/internal/services/storage/resource_arm_storage_account.go
+++ b/azurerm/internal/services/storage/resource_arm_storage_account.go
@@ -245,6 +245,19 @@ func resourceArmStorageAccount() *schema.Resource {
 								},
 							},
 						},
+						"change_feed": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:     schema.TypeBool,
+										Required: true,
+									},
+								},
+							},
+						},
 					},
 				},
 			},
@@ -1341,7 +1354,20 @@ func expandBlobProperties(input []interface{}) storage.BlobServiceProperties {
 	corsRaw := v["cors_rule"].([]interface{})
 	props.BlobServicePropertiesProperties.Cors = expandBlobPropertiesCors(corsRaw)
 
+	changeFeed := v["change_feed"].([]interface{})
+	props.BlobServicePropertiesProperties.ChangeFeed = expandBlobPropertiesChangeFeed(changeFeed)
+
 	return props
+}
+
+func expandBlobPropertiesChangeFeed(input []interface{}) *storage.ChangeFeed {
+	if len(input) == 0 {
+		return nil
+	}
+
+	changeFeedAttr := input[0].(map[string]interface{})
+
+	return &storage.ChangeFeed{Enabled: utils.Bool(changeFeedAttr["enabled"].(bool))}
 }
 
 func expandBlobPropertiesDeleteRetentionPolicy(input []interface{}) *storage.DeleteRetentionPolicy {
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/storage/resource_arm_storage_account.go
```diff
@@ -245,6 +245,19 @@ func resourceArmStorageAccount() *schema.Resource {
...
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:     schema.TypeBool,
+										Required: true,
+									},
+								},
+							},
+						},
```
katbyte, line 260: Given this is only an true/false value and unless it will change prior to moving out of preview should we make this a top level property `change_feed_enasbled`

lrxtom2: Hi @katbyte , Thanks for reviewing. 
Because the field `change_feed` is a preview field and have three status: `true`, `false`, `nil`. So we decided to use nest block (same as go-sdk).

Reference: 
1. [Document](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-change-feed?tabs=azure-portal)
2. [Reference we are following up](https://github.com/terraform-providers/terraform-provider-azurerm/blob/8dc7ecafca4cc76173fc43a11f24c66de08c869f/azurerm/internal/services/compute/virtual_machine_scale_set.go#L25)

katbyte: @lrxtom2 - i'm not sure what we get by nested it in a block like this? could we please move it the top level as `change_feed_enasbled`. thanks

lrxtom2: It's a great idea but it's a preview feature. If we default any value and user don't register the feature into his subscription will meet an error message below. So it may effect many end users.
`autorest/azure: Service returned an error. Status=<nil> Code="FeatureNotSupportedForAccount" Message="Change Feed is not supported for the account."`
So we refer field `ultra_ssd_enabled` in `virtual machine` to use nest block.

lrxtom2:
@katbyte , Thanks for your review. I replied your comment, please have a check.

azurerm/internal/services/storage/resource_arm_storage_account.go
tombuildsstuff: @lrxtom2 the API should only be returning that error when trying to Enable that functionality - otherwise that's a bug in the API? From Terraform's perspective I'd agree this wants to be added at the top level as a flag

katbyte:
@lrxtom2 - see mine and toms comments - this wants changing to a top level property `change_feed_enabled`. We can use GetOkExists even thou its deprecated(nolint it with a comment) without a default to determine if the user has actually set it or not.

tombuildsstuff:
hi @lrxtom2 

Since we've not heard back from you here I'm going to temporarily close this PR for the moment, rather than leaving this open stale - but if you can make the changes proposed above then we should be able to re-open this and get this merged 👍 

Thanks!

PR #6639, opened 2020-04-27
VM and VMSS validation fixes

manicminer:
Separate validation for VM/VMSS (resource) name, and linux/windows computer names

- Adjust naming rules to meet current API behavior
- Length validation for VMSS computer name prefixes

This maintains the current behavior of assuming the `computer_name` / `computer_name_prefix` from the VM / VMSS name respectively. However since the `name` values now have more relaxed rules, should the specified `name` not be a valid `computer_name` / `computer_name_prefix`, then the latter must also be specified. There are new tests for these cases.

### DIFF
--- a/azurerm/internal/services/compute/resource_arm_windows_virtual_machine_scale_set.go
+++ b/azurerm/internal/services/compute/resource_arm_windows_virtual_machine_scale_set.go
@@ -49,7 +49,7 @@ func resourceArmWindowsVirtualMachineScaleSet() *schema.Resource {
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     true,
-				ValidateFunc: ValidateWindowsName,
+				ValidateFunc: ValidateVmName,
 			},
 
 			"resource_group_name": azure.SchemaResourceGroupName(),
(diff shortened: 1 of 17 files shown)


azurerm/internal/services/compute/resource_arm_windows_virtual_machine_scale_set.go
```diff
@@ -371,6 +370,10 @@ func resourceArmWindowsVirtualMachineScaleSetCreate(d *schema.ResourceData, meta
 	if v, ok := d.GetOk("computer_name_prefix"); ok && len(v.(string)) > 0 {
 		computerNamePrefix = v.(string)
 	} else {
+		_, errs := ValidateWindowsComputerNamePrefix(d.Get("name"), "computer_name_prefix")
+		if len(errs) > 0 {
+			return fmt.Errorf("unable to assume default computer name prefix. %s. Please adjust the %q, or specify an explicit %q", errs[0], "name", "computer_name_prefix")
```
katbyte, line 375: and here

manicminer:
> Thanks @manicminer! Aside from some minor comments LGTM 👍

Fixed, thanks for catching!

PR #6661, opened 2020-04-28
Add health_check_path to site_config

### LINKED ISSUE #6429: Support for HealthCheckPath on SchemaAppServiceSiteConfig
Hi,

### Community Note

### Description

It's now possible to configure [HealthCheckPath on a app_service](https://github.com/projectkudu/kudu/wiki/Health-Check-(Preview)), it looks like it's managed via site_config mapped to this SchemaAppServiceSiteConfig

* azurerm_app_service

Best regards,

sirlatrom:
Fixes #6429.

### DIFF
--- a/azurerm/helpers/azure/app_service.go
+++ b/azurerm/helpers/azure/app_service.go
@@ -450,6 +450,11 @@ func SchemaAppServiceSiteConfig() *schema.Schema {
 					}, false),
 				},
 
+				"health_check_path": {
+					Type:     schema.TypeString,
+					Optional: true,
+				},
+
 				"linux_fx_version": {
 					Type:     schema.TypeString,
 					Optional: true,
@@ -745,6 +750,11 @@ func SchemaAppServiceDataSourceSiteConfig() *schema.Schema {
 				},
 
 				"ftps_state": {
+					Type:     schema.TypeString,
+					Computed: true,
+				},
+
+				"health_check_path": {
 					Type:     schema.TypeString,
 					Computed: true,
 				},
@@ -1486,6 +1496,10 @@ func ExpandAppServiceSiteConfig(input interface{}) (*web.SiteConfig, error) {
 		siteConfig.FtpsState = web.FtpsState(v.(string))
 	}
 
+	if v, ok := config["health_check_path"]; ok {
+		siteConfig.HealthCheckPath = utils.String(v.(string))
+	}
+
 	if v, ok := config["min_tls_version"]; ok {
 		siteConfig.MinTLSVersion = web.SupportedTLSVersions(v.(string))
 	}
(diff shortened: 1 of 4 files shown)


sirlatrom:
@katbyte This should solve #6429. I've added an acceptance test that should pass as well. I've verified the functionality on my org's tenant, but cannot run acceptance tests due to restrictive permissions with regard to creating resource groups.

sirlatrom:
@katbyte I've added docs in 100e193c8. As a side note, I see some attributes are missing from the docs for the `azurerm_app_service_slot` page although they share the same schema for several sub-resources such as `site_settings`, but I don't suppose adding those should be a requirement for this PR to be merged.

azurerm/helpers/azure/app_service.go
```diff
@@ -1605,6 +1619,11 @@ func FlattenAppServiceSiteConfig(input *web.SiteConfig) []interface{} {
 
 	result["scm_type"] = string(input.ScmType)
 	result["ftps_state"] = string(input.FtpsState)
+
+	if input.HealthCheckPath != nil {
+		result["health_check_path"] = *input.HealthCheckPath
+	}
```
tombuildsstuff, line 1625: in general we need to set a value for every key inside a dictionary, which helps us catch when Azure doesn't return a value from the API (otherwise Terraform Core will show a diff between "null" and "empty") e.g.

```suggestion
	healthCheckPath := ""
	if input.HealthCheckPath !=  nil {
		healthCheckPath = *input.HealthCheckPath
	}
	result["health_check_path"] = healthCheckPath
```

however since the rest of this class doesn't, we can ignore this for now

website/docs/r/app_service.html.markdown
```diff
@@ -187,6 +187,8 @@ A `site_config` block supports the following:
 
 * `ftps_state` - (Optional) State of FTP / FTPS service for this App Service. Possible values include: `AllAllowed`, `FtpsOnly` and `Disabled`.
 
+* `health_check_path` - (Optional) The health check path to be pinged by App Service. [For more information - please see the corresponding Kudu Wiki page](https://github.com/projectkudu/kudu/wiki/Health-Check-(Preview)).
```
tombuildsstuff, line 190: since this is in Preview - could we add a note to this effect?

```suggestion
* `health_check_path` - (Optional) The health check path to be pinged by App Service. [For more information - please see the corresponding Kudu Wiki page](https://github.com/projectkudu/kudu/wiki/Health-Check-(Preview)).

~> **Note:** This functionality is in Preview and is subject to changes (including breaking changes) on Azure's end
```

azurerm/helpers/azure/app_service.go
sirlatrom: The API explicitly supports `null` as the value for this field, so I assumed that was what not setting the field would represent.

sirlatrom: FWIW I can change the above to explicitly set `result["health_check_path"] = nil` when `input.HealthCheckPath == nil` if that is supported.

jackofallops:
image
Failures are transient on Azure side

PR #6641, opened 2020-04-27
storage_account_name should reference the name of the storage account

JSkimming:
The recent change to deprecate `storage_connection_string` has resulted in a bug in the documentation where it states `storage_account_name` should use the `id` of the storage account. It really should use the `name`.

Fixes #6633

PR #6662, opened 2020-04-28
use name instead of id

r0bnet:
With `id` it doesn't work since it is validating it (max 24 chars and no special chars...)


CodingAnarchy:
Looks like a duplicate of #6641.

katbyte:
hey @r0bnet, @CodingAnarchy is correct! i'm going to close this in favour of the PR opened first and fixes another config too.

Issue #6091, opened 2020-03-12
Azure Management Group creation with Service Principal returns 403


julioas09:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform version: 0.12.20
Azurerm version: 2.0.0

### Affected Resource(s)

* `azurerm_management_group`

We use a Service Principal to connect to out Azure environment. This SP has Owner role at Root Management Group. If we login to Azure CLI with this SP, we can manage Management Groups without a problem.
When we try to run from terraform, we get a 403 error:

```hcl
Error: Error checking for presence of existing Management Group "d45bbbf8-72bd-486b-ac78-73e07c2e4043": managementgroups.Client#Get: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="AuthorizationFailed" Message="The client 'f0e7a3ad-21cb-4968-aea2-b3ef5ec930d4' with object id 'f0e7a3ad-21cb-4968-aea2-b3ef5ec930d4' does not have authorization to perform action 'Microsoft.Management/managementGroups/read' over scope '/providers/Microsoft.Management/managementGroups/d45bbbf8-72bd-486b-ac78-73e07c2e4043' or the scope is invalid. If access was recently granted, please refresh your credentials." 
```

### Terraform Configuration Files

```hcl
provider "azurerm" {
  version = "=2.0.0"
  features {}
}
resource "azurerm_management_group" "ccoe" {
  display_name = "azure-ccoe"
}
```

### Debug Output

Provided in [Gist](https://gist.github.com/julioas09/1a1e857baf02352ba0d97a4a025e0257)

### Expected Behavior

Management Group is created

### Actual Behavior

Terraform apply fails with error 403 forbidden

### Steps to Reproduce

1. Assign service principal as owner of Root Management Group
2. Set proper local env variables to connect with SP:
export ARM_SUBSCRIPTION_ID="xxxxxx"
export ARM_CLIENT_ID="xxxxxx"
export ARM_CLIENT_SECRET="xxxxx"
[...]

PR #6668, opened 2020-04-28
Fix Management Group CreateUpdate Function

wsf11:
Addresses issue #6091 

Issue: The Azure REST API returns a 403 if either you do not have access to a management group or a management group cannot be found. When creating a new management group, a check is always performed to see if the management group exists already, causing a 403, and an error. 

Fix: Change response not found to response forbidden and log message. In addition, there was a regression in PR #6276 that did not check for the name, rather it was checking for the group_name


wsf11:
@katbyte Let me know if you have any feedback on this PR!

azurerm/utils/response.go
```diff
@@ -36,3 +36,7 @@ func ResponseWasStatusCode(resp autorest.Response, statusCode int) bool { // nol
 
 	return false
 }
+
+func ResponseWasForbidden(resp autorest.Response) bool {
```
mbfrahry, line 40: It looks like this was added recently so we can remove it.

jackofallops:
Quick update - Our CI System still being worked on to be able to run the suite there to be able to merge/release this, but we're getting there... When running locally:

Tests Pass
```
--- PASS: TestAccAzureRMManagementGroup_basic (54.19s)
--- PASS: TestAccAzureRMManagementGroup_requiresImport (58.97s)
--- PASS: TestAccAzureRMManagementGroup_withName (62.22s)
--- PASS: TestAccDataSourceArmManagementGroup_basic (65.46s)
--- PASS: TestAccAzureRMManagementGroup_withSubscriptions (85.23s)
--- PASS: TestAccAzureRMManagementGroup_nested (88.34s)
--- PASS: TestAccAzureRMManagementGroup_multiLevel (121.81s)
--- PASS: TestAccAzureRMManagementGroup_multiLevelUpdated (147.49s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/managementgroup/tests	227.362s
```

jackofallops:
I've looped back around to this now I can test review it properly.
What I've found is that the service appears to respond with a 403 when a `get` is sent for a non-existent group ID, where it _should_ respond with a 404. This behaviour would have been masked pre-2.0 as there was a conditional test on the import / existence check (`features.ShouldResourcesBeImported()`) so unless opted into that feature in 1.x, one would never see it.
As a result, I think this needs a slightly different change for this bug. I'll suggest in a review for that.

There's also the name update bug as mentioned by @ArcturusZhang - But I think that will need some careful thought and probably separating the Create and Update functions, as the change of name appears to change the resource ID of the group also.

jackofallops:
Hi @wsf11 - I hope you don't mind, I've pushed a few changes to address the issue and resolve a couple other items.  Tests are now passing fully:
image
image

403 issue logged [here](https://github.com/Azure/azure-rest-api-specs/issues/9549)

wsf11:
@jackofallops no worries. Glad I could help with this!

PR #6684, opened 2020-04-29
tooling: Provider Defined Builds

tombuildsstuff:
This commit introduces Provider Defined Builds - which allows us to define the Build Configuration used for Acceptance Testing as code in this repository.

As a part of this we dynamically generate Build Configurations based on the Services which exist in this repository - thus adding a new Build Configuration will automatically (once merged) add a new matching Build Configuration.


.teamcity/components/settings.kt
```diff
@@ -0,0 +1,37 @@
...
+        "germany" to LocationConfiguration("germanynortheast", "germanycentral", "", false)
+)
+
+// specifies the list of Azure Environments where tests should be run nightly
+var runNightly = mapOf(
+        "public" to true
+)
+
+// specifies a list of services which should be run with a custom test configuration
+var serviceTestConfigurationOverrides = mapOf(
```
tombuildsstuff, line 18: presumably this list is going to want more tweaking based on trial/error - but that's outside the scope of this PR

PR #6685, opened 2020-04-29
Fix postgresql documentation

danryc:


website/docs/r/postgresql_server.html.markdown
```diff
@@ -51,35 +49,35 @@ The following arguments are supported:
...
 
-* `geo_redundant_backup_enabled` - (Optional) Turn Geo-redundant server backups on/off. This allows you to choose between locally redundant or geo-redundant backup storage in the General Purpose and Memory Optimized tiers. When the backups are stored in geo-redundant backup storage, they are not only stored within the region in which your server is hosted, but are also replicated to a paired data center. This provides better protection and ability to restore your server in a different region in the event of a disaster. This is not support for the Basic tier.
+* `geo_redundant_backup_enabled` - (Optional) Turn Geo-redundant server backups on/off. Possible values are `true` and `false`. This allows you to choose between locally redundant or geo-redundant backup storage in the General Purpose and Memory Optimized tiers. When the backups are stored in geo-redundant backup storage, they are not only stored within the region in which your server is hosted, but are also replicated to a paired data center. This provides better protection and ability to restore your server in a different region in the event of a disaster. This is not support for the Basic tier.
```
jackofallops, line 64: minor typo and intl Eng change.
```suggestion
* `geo_redundant_backup_enabled` - (Optional) Turn Geo-redundant server backups on/off. Possible values are `true` and `false`. This allows you to choose between locally redundant or geo-redundant backup storage in the General Purpose and Memory Optimised tiers. When the backups are stored in geo-redundant backup storage, they are not only stored within the region in which your server is hosted, but are also replicated to a paired data centre. This provides better protection and ability to restore your server in a different region in the event of a disaster. This is not supported for the Basic tier.
```

danryc: Hello Steve
You are right - I saw my mistake when I created new PR and previous commit was there also. Thank you for the tip - next time I'll do in mentioned way.
-- 
Daniel

PR #6705, opened 2020-04-30
`azurerm_app_service` `ip_restriction` - support for `name` & `priority`

### LINKED ISSUE #6388: Support for name and priority in ip_restriction of azurerm_app_service
### Community Note

### Description

The App Service IP restriction feature can optionally take `name` and `priority` parameters. This should be supported in the TF provider as well
https://www.terraform.io/docs/providers/azurerm/r/app_service.html#ip_restriction

image

* azurerm_app_service

Basicially it could look like this:

```hcl
    ip_restriction {
      ip_address  = "1.2.3.4/32" # this exists today
      name       = "1-2-3 restriction"
      priority    = 123
    }
```

aqche:
Fixes: #6388

Adds the `name` and `priority` options to the `azurerm_app_service` resource's `ip_restriction` option.

```
=== RUN   TestAccAzureRMAppService_completeIpRestriction
=== PAUSE TestAccAzureRMAppService_completeIpRestriction
=== CONT  TestAccAzureRMAppService_completeIpRestriction
--- PASS: TestAccAzureRMAppService_completeIpRestriction (143.55s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/web/tests   143.564s
```

```
=== RUN   TestAccDataSourceAzureRMAppService_ipRestriction
=== PAUSE TestAccDataSourceAzureRMAppService_ipRestriction
=== CONT  TestAccDataSourceAzureRMAppService_ipRestriction
--- PASS: TestAccDataSourceAzureRMAppService_ipRestriction (147.63s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/web/tests   147.645s
```

### DIFF
--- a/azurerm/helpers/azure/app_service.go
+++ b/azurerm/helpers/azure/app_service.go
@@ -324,7 +324,7 @@ func SchemaAppServiceSiteConfig() *schema.Schema {
 							"priority": {
 								Type:         schema.TypeInt,
 								Optional:     true,
-								Computed:     true,
+								Default:      65000,
 								ValidateFunc: validation.IntBetween(1, 2147483647),
 							},
 						},
--- a/azurerm/helpers/azure/app_service.go
+++ b/azurerm/helpers/azure/app_service.go
@@ -325,7 +325,7 @@ func SchemaAppServiceSiteConfig() *schema.Schema {
 								Type:         schema.TypeInt,
 								Optional:     true,
 								Computed:     true,
-								ValidateFunc: validation.IntBetween(1, 65000),
+								ValidateFunc: validation.IntBetween(1, 2147483647),
 							},
 						},
 					},
--- a/azurerm/helpers/azure/app_service.go
+++ b/azurerm/helpers/azure/app_service.go
@@ -324,6 +324,7 @@ func SchemaAppServiceSiteConfig() *schema.Schema {
 							"priority": {
 								Type:         schema.TypeInt,
 								Optional:     true,
+								Computed:     true,
 								ValidateFunc: validation.IntAtLeast(1),
 							},
 						},
--- a/azurerm/helpers/azure/app_service.go
+++ b/azurerm/helpers/azure/app_service.go
@@ -319,12 +319,13 @@ func SchemaAppServiceSiteConfig() *schema.Schema {
 							"name": {
 								Type:         schema.TypeString,
 								Optional:     true,
+								Computed:     true,
 								ValidateFunc: validation.StringIsNotEmpty,
 							},
 							"priority": {
 								Type:         schema.TypeInt,
 								Optional:     true,
-								Default:      65000,
+								Computed:     true,
 								ValidateFunc: validation.IntBetween(1, 2147483647),
 							},
 						},
--- a/azurerm/helpers/azure/app_service.go
+++ b/azurerm/helpers/azure/app_service.go
@@ -316,6 +316,16 @@ func SchemaAppServiceSiteConfig() *schema.Schema {
 								Optional:     true,
 								ValidateFunc: validation.StringIsNotEmpty,
 							},
+							"name": {
+								Type:         schema.TypeString,
+								Optional:     true,
+								ValidateFunc: validation.StringIsNotEmpty,
+							},
+							"priority": {
+								Type:         schema.TypeInt,
+								Optional:     true,
+								ValidateFunc: validation.IntAtLeast(1),
+							},
 						},
 					},
 				},
@@ -677,6 +687,14 @@ func SchemaAppServiceDataSourceSiteConfig() *schema.Schema {
 							},
 							"virtual_network_subnet_id": {
 								Type:     schema.TypeString,
+								Computed: true,
+							},
+							"name": {
+								Type:     schema.TypeString,
+								Computed: true,
+							},
+							"priority": {
+								Type:     schema.TypeInt,
 								Computed: true,
 							},
 						},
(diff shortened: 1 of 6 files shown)


azurerm/helpers/azure/app_service.go
```diff
@@ -316,6 +316,17 @@ func SchemaAppServiceSiteConfig() *schema.Schema {
...
+							"name": {
+								Type:         schema.TypeString,
+								Optional:     true,
+								ValidateFunc: validation.StringIsNotEmpty,
+							},
+							"priority": {
+								Type:         schema.TypeInt,
+								Optional:     true,
+								Computed:     true,
+								ValidateFunc: validation.IntAtLeast(1),
```
sebader, line 328: Should you check here as well if priority is maximum 65000?

website/docs/r/app_service.html.markdown
```diff
@@ -309,6 +309,10 @@ A `ip_restriction` block supports the following:
 
 -> **NOTE:** One of either `ip_address` or `virtual_network_subnet_id` must be specified
 
+* `name` - (Optional) The name for this IP Restriction.
+
+* `priority` - (Optional) The priority for this IP Restriction. Restrictions are enforced in priority order.
```
sebader, line 314: Maybe add a note that it defaults to 65000

azurerm/helpers/azure/app_service.go
aqche: good call

website/docs/r/app_service.html.markdown
aqche: sure, thanks for the info

azurerm/helpers/azure/app_service.go
aqche: are you sure about the max value being 65000? after some testing, it looks like the max value is 2147483647 for me.

azurerm/helpers/azure/app_service.go
```diff
@@ -316,6 +316,17 @@ func SchemaAppServiceSiteConfig() *schema.Schema {
...
 							},
+							"name": {
+								Type:         schema.TypeString,
+								Optional:     true,
+								ValidateFunc: validation.StringIsNotEmpty,
+							},
+							"priority": {
+								Type:         schema.TypeInt,
+								Optional:     true,
+								Computed:     true,
```
aqche, line 327: as @sebader noted, priority defaults to 65000 when not provided. Should we make that an explicit default or let it be computed?

azurerm/helpers/azure/app_service.go
sebader: good question actually. I just tried to set a rule via az powershell:
`Add-AzWebAppAccessRestrictionRule: Cannot validate argument on parameter 'Priority'. The 65001 argument is greater than the maximum allowed range of 65000. Supply an argument that is less than or equal to 65000 and then try the command again.`

The portal lets you also set something larger than 65000 and enforces the limit at 2147483647. So I guess you are right, that should be it.

azurerm/helpers/azure/app_service.go
sebader: I'm not sure where the default of 65000 is injected in this case. Might be in the Go SDK? Since in CLI, Powershell and the portal the parameter must be manually specified, there is no default there.

sebader: So might actually be better so explicitly set it?!

sebader:
nice, looks good to me know @aqche Thank you!

azurerm/helpers/azure/app_service.go
```diff
@@ -316,6 +316,17 @@ func SchemaAppServiceSiteConfig() *schema.Schema {
 								Optional:     true,
 								ValidateFunc: validation.StringIsNotEmpty,
 							},
+							"name": {
+								Type:         schema.TypeString,
+								Optional:     true,
+								ValidateFunc: validation.StringIsNotEmpty,
+							},
```
jackofallops, line 323: I think we'll want `Computed: true,` on here.

azurerm/helpers/azure/app_service.go
```diff
@@ -316,6 +316,17 @@ func SchemaAppServiceSiteConfig() *schema.Schema {
...
+							"priority": {
+								Type:         schema.TypeInt,
+								Optional:     true,
+								Default:      65000,
+								ValidateFunc: validation.IntBetween(1, 2147483647),
+							},
```
jackofallops, line 329: I think this should be `Computed: true,` rather than setting a default. If more than one IP Restriction is used this could cause a conflict.  The "default" value is assigned by the platform, rather than it being in the SDK, so can safely be read back from the API if not explicitly set.

azurerm/internal/services/web/tests/resource_arm_app_service_test.go
```diff
@@ -514,6 +514,27 @@ func TestAccAzureRMAppService_oneIpRestriction(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMAppService_completeIpRestriction(t *testing.T) {
```
jackofallops, line 517: Could we have an additional test that updates an app_service by adding, then removing, a 2nd `ip_restriction`? (or extend this one)

jackofallops (changes requested):
Hi @aqche 
Thanks for this PR.  Just a few minor changes needed and I think this should be good to merge.

Ste

aqche:
@jackofallops thanks for the review! made the updates you suggested. let me know what you think!

```
=== RUN   TestAccAzureRMAppService_completeIpRestriction
=== PAUSE TestAccAzureRMAppService_completeIpRestriction
=== CONT  TestAccAzureRMAppService_completeIpRestriction
--- PASS: TestAccAzureRMAppService_completeIpRestriction (266.53s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/web/tests   266.555s```

jackofallops (approved):

jackofallops:
image
Failures are transient at Azure side.

bentterp:
When I compare with the configuration options in the Azure portal, it seems that the action "allow/deny" is missing.

If there were no deny rules, then there would be no need for assigning priorities - similar to NSG rules.

PR #6716, opened 2020-04-30
Add recovery capability for Key Vault Soft Deleted Keys, Secrets and Certificates

jackofallops:
Adds support for recovering Soft Deleted items from KeyVault.  This feature is toggled by the `recover_soft_deleted_key_vaults` setting the Provider's `feature` config.

This requires that Terraform has the `recover` permission on the item type in the KeyVault.


jackofallops:
Tests pass:
image
And the one failed also fixed:
image

Issue #5014, opened 2019-11-28
Policies always update even when no change occurs


eohtake:
Related to #2938

### Terraform (and AzureRM Provider) Version

```
Terraform v0.12.16
+ provider.azurerm v1.34.0
```
### Affected Resource(s)

* `azurerm_policy_definition`

### Terraform Configuration Files

```hcl
resource "azurerm_policy_definition" "policy" {
  name         = "accTestPolicy"
  policy_type  = "Custom"
  mode         = "Indexed"
  display_name = "Acceptance test policy definition"

  metadata = <<METADATA
    {
    "category": "General"
    }
  METADATA

  policy_rule = <<POLICY_RULE
    {
    "if": {
      "not": {
        "field": "location",
        "in": "[parameters('allowedLocations')]"
      }
    },
    "then": {
      "effect": "audit"
    }
  }
POLICY_RULE

  parameters = <<PARAMETERS
    {
    "allowedLocations": {
      "type": "Array",
      "metadata": {
        "description": "The list of allowed locations for resources.",
        "displayName": "Allowed locations",
        "strongType": "location"
      }
    }
  }
PARAMETERS
}
```

### Expected Behavior
Terraform should report no updates in policies that haven't changed during `terraform apply` and `terraform plan`

### Actual Behavior
Terraform always update on refresh, policies that haven't changed.
What Terraform is updating though, are fields in Metadata that are not present in the configuration file, but are present in Azure, so it tries to clear them.

```
              - createdBy = "4-1c33-4966-89ae-23423423423" -> null
              - createdOn = "2019-11-28T10:04:28.3320952Z" -> null
              - updatedBy = "324234-1c33-4234-89ae-234" -> null
              - updatedOn = "2019-11-28T10:47:41.4931002Z" -> null
```

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create
  ~ update in-place

Terraform will perform the following actions:

  # azurerm_policy_definition.policy will be updated in-place
  ~ resource "azurerm_policy_definition" "policy" {
        display_name = "Acceptance test policy definition"
        id           = "/subscriptions/43423423423423-43432423/providers/Microsoft.Authorization/policyDefinitions/accTestPolicy"
      ~ metadata     = jsonencode(
          ~ {
                category  = "General"
              - createdBy = "4-1c33-4966-89ae-23423423423" -> null
[...]

PR #6734, opened 2020-05-01
Custom metadata DiffSuppressFunc - issue #5014

richeney:
Fixes [#5014](https://github.com/terraform-providers/terraform-provider-azurerm/issues/5014)

### DIFF
--- a/azurerm/internal/services/policy/resource_arm_policy_definition.go
+++ b/azurerm/internal/services/policy/resource_arm_policy_definition.go
@@ -2,8 +2,10 @@ package policy
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log"
+	"reflect"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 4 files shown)


richeney:
The Go SDK for azurerm adds four metadata fields - "createdBy", "createdOn", "updatedBy", "updatedOn" - which causes spurious changes in terraform plan. (Users have been using lifecycle ignore blocks on metadata to suppress the "changes".)

I added a custom DiffSuppressFunc that strips those keys (if found) before using reflect.DeepEqual. With the change then valid changes to custom metadata will now be trapped. E.g.:

```hcl
  metadata     = <<METADATA
    {
        "category": "Tags"
    }
METADATA
```

richeney:
I added essentially the same fix to policy definitions as I'd already created for policy set definitions. (As suggested by @keanesf's [comment](https://github.com/terraform-providers/terraform-provider-azurerm/issues/5014#issuecomment-622595629). 

Plus tests for both.

tombuildsstuff:
Tests look good 👍

PR #6742, opened 2020-05-02
Fixes resource_arm_application_gateway with empty string password

gpduck:
Empty string is a valid password

Fixes terraform-providers/terraform-provider-azurerm#6368
Fixes terraform-providers/terraform-provider-azurerm#6295


azurerm/internal/services/network/tests/resource_arm_application_gateway_test.go
```diff
@@ -573,6 +573,25 @@ func TestAccAzureRMApplicationGateway_sslCertificate_keyvault(t *testing.T) {
...
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMApplicationGatewayDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMApplicationGateway_sslCertificateEmptyPassword(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMApplicationGatewayExists(data.ResourceName),
+				),
+			},
+			data.ImportStep(),
```
tombuildsstuff, line 590: this test is failing since the `ssl_certificate.0.data` (and `ssl_certificate.0.password`) fields are read from the config file, rather than the Azure API (as it doesn't return them) - meaning these don't exist at import time (since that doesn't read the config) as such we'll need to update this test to account for that:

```suggestion
			// since these are read from the existing state
			data.ImportStep(
				"ssl_certificate.0.data",
				"ssl_certificate.0.password",
			),
```

I hope you don't mind but since this otherwise LGTM I'm going to push a commit to get this fixed

tombuildsstuff:
Tests look good 👍

Issue #6026, opened 2020-03-07
azurerm_site_recovery_network_mapping fails with "Error checking for presence of existing site recovery network mapping network-primary2secondary"


martenbohlin:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
Terraform v0.12.17
+ provider.azurerm v2.0.0

### Affected Resource(s)

* `azurerm_site_recovery_network_mapping`

### Terraform Configuration Files

This problem is detected by the integration. test TestAccAzureRMSiteRecoveryNetworkMapping_basic

### Expected Behavior

A network mapping should have been created.

### Actual Behavior

The folllowing error:
--- FAIL: TestAccAzureRMSiteRecoveryNetworkMapping_basic (501.84s)
    testing.go:640: Step 0 error: errors during apply:

        Error: Error checking for presence of existing site recovery network mapping mapping-200307180752678515 (vault acctest-vault-200307180752678515): siterecovery.ReplicationNetworkMappingsClient#Get: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="BadRequest" Message="An invalid parameter NetworkMappingName was passed." Details=[{"activityId":"e3c8c514-7a1b-aaac-5e45-ee6cacf41c3a","clientRequestId":"66625d9f-10cb-4fb9-92d7-665e3688fb32","code":"InvalidParameter","message":"An invalid parameter NetworkMappingName was passed.","possibleCauses":"Parameter value is not correct for this action.","recommendedAction":"Correct the parameter value and retry the last action. If the issue persists, contact Support."}]

          on /var/folders/7s/t24np7dn2w5_rg6nhsdjpgr00000gp/T/tf-test089678882/main.tf line 49:
          (source code not available)

### Steps to Reproduce

make testacc TESTARGS='-run=TestAccAzureRMSiteRecoveryNetworkMapping'

### References

* #0000

wenesak:
same here same error, anybody knows of a work around?

tombuildsstuff:
Fixed via #6747

PR #6747, opened 2020-05-03
Workaround for strange response code in recovery services network mappings API. Fixes issue #6026

martenbohlin:
Issue #6026 seems to be caused by the Azure API returning 400 instead of 404 when there are no network-mapping with the name that we ask for. This pull requests adds a workaround for that.

(Is there a way for me to report a bug on the API?)


katbyte:
@martenbohlin - what we typically do is open an issue on the azure-sdk-for gorepo, and then place a link in an comment next to the work aaround so we know why, and can easily check to see if its fixed.

tombuildsstuff:
Test passes:

PR #6752, opened 2020-05-04
New data source - azurerm_virtual_machine_scale_set_network_interfaces

### LINKED ISSUE #7333: azurerm_linux_virtual_machine_scale_set output private ip
### Community Note

### Description

I want to generate an ssh config from terraform inventory using a template. We use a scale set but it is pretty static and we only ever expand for big events. The azurerm_linux_virtual_machine_scale_set is a welcome improvement over the azurerm_virtual_machine_scale_set but it still does not export the private IPs of the VM's it creates.

```hcl
resource "azurerm_linux_virtual_machine_scale_set" "java" {
  instances           = var.java_count
  .....
}
output "java-main" {
    value=azurerm_linux_virtual_machine_scale_set.java
}
```

- azurerm_linux_virtual_machine_scale_set

```hcl
output "java-main" {
    value=azurerm_linux_virtual_machine_scale_set.java.private_ip_adresses # a map of instance id to ip address
}

```

hbuckle:
Data source for scale set NICs - particularly useful to get the IP addresses of a scale set
Resolves #7333


hbuckle:
@tombuildsstuff @katbyte - Anyone available to review this?

hbuckle:
@tombuildsstuff @katbyte - Anyone available to review this?

azurerm/internal/services/network/data_source_virtual_machine_scale_set_network_interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+
+func dataSourceArmVirtualMachineScaleSetNetworkInterfacesRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Network.InterfacesClient
+	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	resGroup := d.Get("resource_group_name").(string)
+	vmssName := d.Get("virtual_machine_scale_set_name").(string)
+
+	resp, err := client.ListVirtualMachineScaleSetNetworkInterfaces(ctx, resGroup, vmssName)
```
ArcturusZhang, line 187: Since this API is a pagable request, could we switch to the `ListVirtualMachineScaleSetNetworkInterfacesComplete` method and use the iterator to iterate the result?

azurerm/internal/services/network/data_source_virtual_machine_scale_set_network_interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+	if err != nil {
+		return fmt.Errorf("Error making Read request on Azure VMSS %q (Resource Group %q): %+v", vmssName, resGroup, err)
+	}
+
+	results := make([]interface{}, 0)
+
+	for _, iface := range resp.Values() {
```
ArcturusZhang, line 194: Actually the result may have multiple pages, therefore you should add an outer loop to iterate all of the pages.
But it would be more clear if you use the `ListVirtualMachineScaleSetNetworkInterfacesComplete` and iterate over the iterator.

azurerm/internal/services/network/data_source_virtual_machine_scale_set_network_interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+		result := make(map[string]interface{})
+
+		if iface.Name != nil {
+			result["name"] = *iface.Name
+		} else {
+			result["name"] = ""
+		}
```
ArcturusZhang, line 201: Usually we would write this like this:
```suggestion
                name := "" // var name string also works
		if iface.Name != nil {
			name = *iface.Name
		}
		
		// other fields
		
		results = append(results, map[string]interface{}{
		      "name": name,
		      // other fields
		})
```

azurerm/internal/services/network/data_source_virtual_machine_scale_set_network_interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+			configs := *iface.IPConfigurations
+
+			result["private_ip_address"] = *configs[0].InterfaceIPConfigurationPropertiesFormat.PrivateIPAddress
+
+			addresses := make([]interface{}, 0)
+			for _, config := range configs {
+				if config.InterfaceIPConfigurationPropertiesFormat != nil {
+					addresses = append(addresses, config.InterfaceIPConfigurationPropertiesFormat.PrivateIPAddress)
+				}
+			}
```
ArcturusZhang, line 237: Could we please extract this into an flatten function? Which may make this more clear

azurerm/internal/services/network/data_source_virtual_machine_scale_set_network_interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+		var appliedDNSServers []string
+		var dnsServers []string
+		if dnsSettings := iface.DNSSettings; dnsSettings != nil {
+			if s := dnsSettings.AppliedDNSServers; s != nil {
+				appliedDNSServers = *s
+			}
+
+			if s := dnsSettings.DNSServers; s != nil {
+				dnsServers = *s
+			}
```
ArcturusZhang, line 254: Actually we have a built-in helper function `utils.FlattenStringSlice()`, could we please use this function instead? By doing this, these nil-checks can be removed.

azurerm/internal/services/network/data_source_virtual_machine_scale_set_network_interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+				result["internal_dns_name_label"] = *dnsSettings.InternalDNSNameLabel
+			} else {
+				result["internal_dns_name_label"] = ""
+			}
+		}
+
+		result["applied_dns_servers"] = appliedDNSServers
+		result["dns_servers"] = dnsServers
+		result["enable_ip_forwarding"] = *iface.EnableIPForwarding
+		result["enable_accelerated_networking"] = *iface.EnableAcceleratedNetworking
```
ArcturusZhang, line 266: These two would also need nil-checks

azurerm/internal/services/network/data_source_virtual_machine_scale_set_network_interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+								},
+							},
+						},
+
+						"dns_servers": {
+							Type:     schema.TypeSet,
+							Computed: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							Set:      schema.HashString,
+						},
```
ArcturusZhang, line 136: Since this is a Computed attribute, we should just use `TypeList` instead of `TypeSet`. And this would also benefit the users when they are trying to use one of the element.

azurerm/internal/services/network/data_source_virtual_machine_scale_set_network_interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"applied_dns_servers": {
+							Type:     schema.TypeSet,
+							Computed: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							Set:      schema.HashString,
+						},
```
ArcturusZhang, line 148: Similar with the above, should use `TypeList`

azurerm/internal/services/network/data_source_virtual_machine_scale_set_network_interfaces.go
```diff
@@ -0,0 +1,275 @@
...
+										Elem:     &schema.Schema{Type: schema.TypeString},
+										Set:      schema.HashString,
+									},
+
+									"application_security_group_ids": {
+										Type:     schema.TypeSet,
+										Computed: true,
+										Elem:     &schema.Schema{Type: schema.TypeString},
+										Set:      schema.HashString,
+									},
```
ArcturusZhang, line 121: Since this is a Computed attribute, we should just use TypeList instead of TypeSet. And this would also benefit the users when they are trying to use one of the element.

hbuckle:
@ArcturusZhang - I think I've addressed everything

azurerm/internal/services/network/virtual_machine_scale_set_network_interfaces_data_source.go
```diff
@@ -0,0 +1,279 @@
...
+
+			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
+
+			"network_interfaces": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+
+						"name": {
```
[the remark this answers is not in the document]
hbuckle: Removed

svyotov:
This looks like a useful/required PR :+1:

hbuckle:
It should be ready to go, @ArcturusZhang can you confirm?

karlbohlmark:
This looks like a really valuable change that is very close to actually happening. Will it be merged or is it advisable to look for workarounds for this instead of waiting for this to be accepted?

hbuckle:
I've resolved the merge conflict so just needs a reviewer

hbuckle:
@ArcturusZhang @tombuildsstuff bumping this one again

JonShelley:
Looks like this fell through the cracks. Any update on this?

katbyte:
Hey @hbuckle - is there a reason for this to be specific to NICs and not a general windows/linux vmss datasource?

hbuckle:
@katbyte I think it was following the model of `azurerm_network_interface` which has its own data source. It is also a separate API to get the nics so made sense as a separate data source

katbyte:
@hbuckle - i'm still not sure what we gain by having this as a separate resource other then part of existing data source giving us `azure_virtual_machine_scale_set.network_interfaces`? 

`azurerm_network_interface` has a direct resource counterpoint, while this just pulls the list of nics for the vmss? thus it makes sense to be part of that datasource rather then sepreate

hbuckle:
@katbyte sure it can go as part of the existing data source. TBH I've forgotten what I even needed this for now.

katbyte:
Sounds good @hbuckle - mind if i close this until you've done so and can open a new PR?

katbyte:
It looks like there is a PR already open to do this! #10585 - as such i'm going to close this in favour of that pr

PR #6774, opened 2020-05-05
[BUG] Analysis Services IPv4 rule name casing issue fix

r0bnet:
I needed to change the hash function of an `ipv4_firewall_rule` to a custom one in order to prevent situations like the following:

```hcl
azurerm_analysis_services_server.analysis_services will be updated in-place
  ~ resource "azurerm_analysis_services_server" "analysis_services" {
        admin_users               = []
        enable_power_bi_service   = true
        id                        = "/subscriptions/dc8fbfc0-2769-421c-921e-f0dae437fa79/resourceGroups/dl4-analysis-services-rg/providers/Microsoft.AnalysisServices/servers/myanalysisservices"
        location                  = "northeurope"
        name                      = "myanalysisservices"
        querypool_connection_mode = "All"
        resource_group_name       = "dl4-analysis-services-rg"
        server_full_name          = "asazure://northeurope.asazure.windows.net/myanalysisservices"
        sku                       = "S1"
        tags                      = {}
      + ipv4_firewall_rule {
          + name        = "DWHMGMT-QDM-VM"
          + range_end   = "13.74.177.213"
          + range_start = "13.74.177.213"
        }
      - ipv4_firewall_rule {
          - name        = "dwhmgmt-qdm-vm" -> null
          - range_end   = "13.74.177.213" -> null
          - range_start = "13.74.177.213" -> null
        }
    }
```

When applying the changes above nothing really changed on server side as it didn't see a difference in the name.

@tombuildsstuff suggested to try to convert it to a list instead but I already switched the rules to Set a few weeks ago because the order sometimes randomly changed. That's why I think a Set is the best option here.

### DIFF
--- a/azurerm/internal/services/analysisservices/analysis_services_server_resource.go
+++ b/azurerm/internal/services/analysisservices/analysis_services_server_resource.go
@@ -3,13 +3,13 @@ package analysisservices
 import (
 	"bytes"
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/helper/hashcode"
 	"log"
 	"regexp"
 	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/analysisservices/mgmt/2017-08-01/analysisservices"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/hashcode"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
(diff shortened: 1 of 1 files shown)


jackofallops (approved):
Thanks @r0bnet 
LGTM 👍  Will fire off the acctests now.

Ste

jackofallops:
Tests pass:
image

PR #6786, opened 2020-05-06
[Enhancement] Allow updating Analysis Services Server when paused

r0bnet:
This PR checks if the Analysis Services Server is paused when it's going to be updated and will start it before. It will also suspend it again afterwards if needed.

Before it was unable to update a paused Analysis Services Server.


azurerm/internal/services/analysisservices/analysis_services_server_resource.go
```diff
@@ -252,6 +252,32 @@ func resourceArmAnalysisServicesServerUpdate(d *schema.ResourceData, meta interf
...
+	if err != nil {
+		if utils.ResponseWasNotFound(serverResp.Response) {
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error retrieving Analysis Services Server %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
+	}
+
+	if serverResp.State != analysisservices.StateSucceeded && serverResp.State != analysisservices.StatePaused {
+		return fmt.Errorf("Error updating Analysis Services Server %q (Resource Group %q): State must be either Succeeded or Paused", id.Name, id.ResourceGroup)
```
tombuildsstuff, line 265: it'd be worth adding the current state here

```suggestion
		return fmt.Errorf("Error updating Analysis Services Server %q (Resource Group %q): State must be either Succeeded or Paused but got %q", id.Name, id.ResourceGroup, serverResp.State)
```

azurerm/internal/services/analysisservices/analysis_services_server_resource.go
```diff
@@ -252,6 +252,32 @@ func resourceArmAnalysisServicesServerUpdate(d *schema.ResourceData, meta interf
 		return err
 	}
 
+	serverResp, err := client.GetDetails(ctx, id.ResourceGroup, id.Name)
+	if err != nil {
+		if utils.ResponseWasNotFound(serverResp.Response) {
+			d.SetId("")
+			return nil
```
tombuildsstuff, line 259: this is the update - so if this is a 404 we should error (since the Read function will be run prior to this - so this'd be changes made OOB during the run)

azurerm/internal/services/analysisservices/analysis_services_server_resource.go
```diff
@@ -271,6 +297,17 @@ func resourceArmAnalysisServicesServerUpdate(d *schema.ResourceData, meta interf
 		return fmt.Errorf("Error waiting for completion of Analysis Services Server %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
 	}
 
+	if isPaused {
+		suspendFuture, err := client.Suspend(ctx, id.ResourceGroup, id.Name)
+		if err != nil {
+			return fmt.Errorf("Error pausing Analysis Services Server %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
```
tombuildsstuff, line 303: maybe worth clarifying:

```suggestion
			return fmt.Errorf("Error re-pausing Analysis Services Server %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
```

azurerm/internal/services/analysisservices/analysis_services_server_resource.go
r0bnet: Done

r0bnet:
@tombuildsstuff added an acceptance test and fixed your comments.

azurerm/internal/services/analysisservices/analysis_services_server_resource.go
```diff
@@ -245,13 +245,35 @@ func resourceArmAnalysisServicesServerUpdate(d *schema.ResourceData, meta interf
...
+	}
+
+	if serverResp.State != analysisservices.StateSucceeded && serverResp.State != analysisservices.StatePaused {
+		return fmt.Errorf("Error updating Analysis Services Server %q (Resource Group %q): State must be either Succeeded or Paused but got %q", id.Name, id.ResourceGroup, serverResp.State)
+	}
+
+	isPaused := serverResp.State == analysisservices.StatePaused
+
+	if isPaused {
+		resumeFuture, err := client.Resume(ctx, id.ResourceGroup, id.Name)
```
katbyte, line 267: Should we add a note to the docs that the resource will briefly unsuspend the server to make updates?

azurerm/internal/services/analysisservices/tests/analysis_services_server_resource_test.go
```diff
@@ -210,6 +211,33 @@ func TestAccAzureRMAnalysisServicesServer_backupBlobContainerUri(t *testing.T) {
...
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMAnalysisServicesServerDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMAnalysisServicesServer_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMAnalysisServicesServerExists(data.ResourceName),
+					testSuspendAzureRMAnalysisServicesServer(data.ResourceName),
```
katbyte, line 226: would it make sense to add a `state` property to the resource so we don't have to do this?

azurerm/internal/services/analysisservices/analysis_services_server_resource.go
r0bnet: Added a note for this in the docs.

azurerm/internal/services/analysisservices/tests/analysis_services_server_resource_test.go
r0bnet: Was thinking about that but didn't find a real use case for that attribute. Didn't want to add it simply to make a test easier.

tombuildsstuff: we need to confirm the state in Azure, rather than in the Statefile - so it's better to check Azure directly here

tombuildsstuff:
Tests pass:

Screenshot 2020-05-07 at 11 21 31

PR #6787, opened 2020-05-06
fix bug when scope is management group for resource `azurerm_role_definition`

### LINKED ISSUE #6695: azurerm_role_definition destroy fails when scoping role at management group level
_This issue was originally opened by @TofferK as hashicorp/terraform#24806. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._
    
<hr>
    

### Terraform Version

```
Terraform v0.12.24
+ provider.azurerm v2.7.0
...
```

```hcl
data "azurerm_subscription" "primary" {
}

data "azurerm_management_group" "top_level" {
  name = "mg-level1-01"
}

resource "azurerm_role_definition" "example" {
  name        = "my-custom-role2"
  scope       = data.azurerm_management_group.top_level.id
  description = "This is a custom role created via Terraform"

  permissions {
    actions     = ["*"]
    not_actions = []
  }

  assignable_scopes = [
    data.azurerm_management_group.top_level.id,
  ]
}
...
```

### Debug Output
 See attached trace output

### Expected Behavior
 The role definition should have been destroyed

### Actual Behavior
 The role definition destroy fails, with an error

### Steps to Reproduce

1. `terraform init`
2. `terraform apply`
3. `terraform destroy`

### References

[azurerm_role_definition_destroy.log](https://github.com/hashicorp/terraform/files/4552645/azurerm_role_definition_destroy.log)

njuCZ:
fix #6695 

when scope is management group, the id format is `/providers/Microsoft.Authorization/roleDefinitions/{id}`
 after func `parseRoleDefinitionId`, the scope is empty string, so delete function will report error.
This will not affect `createUpdate` and `read`

solution: read the real scope from state if scope is empty

### DIFF
--- a/azurerm/internal/services/authorization/resource_arm_role_definition.go
+++ b/azurerm/internal/services/authorization/resource_arm_role_definition.go
@@ -221,6 +221,15 @@ func resourceArmRoleDefinitionDelete(d *schema.ResourceData, meta interface{}) e
 		return err
 	}
 
+	// fix bugs when scope is management group
+	// in this case, the id format is `/providers/Microsoft.Authorization/roleDefinitions/{id}`
+	// after func `parseRoleDefinitionId`, the scope is empty string
+	// read the real scope from state
+	if id.scope == "" {
+		scope := d.Get("scope").(string)
+		id.scope = strings.TrimPrefix(scope, "/")
+	}
+
 	resp, err := client.Delete(ctx, id.scope, id.roleDefinitionId)
 	if err != nil {
 		if !utils.ResponseWasNotFound(resp.Response) {
--- a/azurerm/internal/services/authorization/tests/resource_arm_role_definition_test.go
+++ b/azurerm/internal/services/authorization/tests/resource_arm_role_definition_test.go
@@ -165,6 +165,25 @@ func TestAccAzureRMRoleDefinition_emptyName(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMRoleDefinition_scopeManagementGroup(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_role_definition", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMRoleDefinitionDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMRoleDefinition_scopeManagementGroup(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMRoleDefinitionExists(data.ResourceName),
+				),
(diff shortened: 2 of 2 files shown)


azurerm/internal/services/authorization/resource_arm_role_definition.go
```diff
@@ -221,6 +221,15 @@ func resourceArmRoleDefinitionDelete(d *schema.ResourceData, meta interface{}) e
...
 	}
 
+	// fix bugs when scope is management group
+	// in this case, the id format is `/providers/Microsoft.Authorization/roleDefinitions/{id}`
+	// after func `parseRoleDefinitionId`, the scope is empty string
+	// read the real scope from state
+	if id.scope == "" {
+		scope := d.Get("scope").(string)
+		id.scope = strings.TrimPrefix(scope, "/")
+	}
```
tombuildsstuff, line 231: unfortunately this won't work where the resource has been removed from the Terraform Configuration - instead we'd be better to look up the Role Definition and confirm the level this Role Definition is created at - which should give us that information consistently?

njuCZ:
@tombuildsstuff it seems we could not look up the scope information in the API, So could I wrap it in the id using the format `{id}|{scope}` if the scope is `management group` ?

tombuildsstuff:
@njuCZ based on what I can see, the ID should be in the format: `/{scope}/providers/Microsoft.Authorization/roleDefinitions/{roleDefinitionId}` - so my understanding is that the ID's should be:

* Management Group: `/providers/Microsoft.Management/managementGroups/group1/providers/Microsoft.Authorization/roleDefinitions/{roleDefinitionID}`
* Subscription: `/subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Authorization/roleDefinitions/{roleDefinitionID}`
* Resource Group: `/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/group1/providers/Microsoft.Authorization/roleDefinitions/{roleDefinitionID}`

all of which should have the Scope available as a part of the ID - so it'd be good to confirm the ID's for these, it's likely the ID parsing function needs updating?

njuCZ:
in fact, when the scope is `Management Group`, the id is `/providers/Microsoft.Authorization/roleDefinitions/{roleDefinitionID}`

tombuildsstuff:
@njuCZ so that sounds like a bug in the API?

njuCZ:
@tombuildsstuff ok, I will fire an issue in the azure-rest-api-spec repo first

tombuildsstuff:
hi @njuCZ 

Since this is blocked on [this bug in the Swagger repository](https://github.com/Azure/azure-rest-api-specs/issues/9389) - rather than leaving this issue open until that's fixed, I'm going to close this issue temporarily (it's already in the Blocked milestone). Once the upstream issue has been resolved then we can re-open this and take another look.

Thanks!

PR #6801, opened 2020-05-06
provider: file renaming for readability [services R-S]

tracypholmes:
Moves `resource_arm` and `data_source` prefix from the beginning of filename for easier readability and navigation.


tombuildsstuff:
As discussed offline - we'll need to roll this out more gradually to avoid merge conflicts in the open PR's, so I'm going to close this PR for the moment

Issue #6588, opened 2020-04-23
azurerm_automation_runbook fails with 404 Error - Unless Published version of runbook already exists (v.2)


arcotek-ltd:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.12.24
+ provider.azuread v0.8.0
+ provider.azurerm v2.6.0

### Affected Resource(s)

* `azurerm_automation_runbook`

### Terraform Configuration Files
[https://gist.github.com/arcotek-ltd/389d3d74572ac1b1f74038c429a0c8c6](https://gist.github.com/arcotek-ltd/389d3d74572ac1b1f74038c429a0c8c6)

### Expected Behavior

"Copy" the runbook from the given `publish_content_link` (storage account (URL)) and "install" the runbook in the automation account.

### Actual Behavior
Receive error:
```
Error: Error retrieving Automation Runbook "myScript" (Account "myAA" / Resource Group "myRg"): automation.RunbookClient#Get: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="ResourceNotFound" Message="The Resource 'Microsoft.Automation/automationAccounts/myAA/runbooks/myScript' under resource group 'myRg' was not found."

  on ..\azurerm_automation_runbook\main.tf line 36, in resource "azurerm_automation_runbook" "pool":
  36: resource "azurerm_automation_runbook" "pool" {
```

Why is it looking in the automation account? Of course it doesn't exist. Isn't that the whole point of this module?

According to the documentation:
_publish_content_link - (Required) The published runbook content link._
If this means it needs to be the URI of the runbook in the automation account, why does your example points to a .ps1 file on Github?

_content - (Optional) The desired content of the runbook._
If, in order to get this to create a runbook, why is this attribute optional?

I feel that the documentation could do with a little update to explain exactly what these attributes do in their current form, and the module needs improvement to allow a runbook to be created form a URI.

### Steps to Reproduce

1. `terraform apply`

### References

[...]

Issue #4403, opened 2019-09-20
Error setting the draft Automation Runbook, azurerm_automation_runbook fails with 404 Error - Unless Published version of runbook already exists 


Mounika0207:
**Terraform Configuration Files**
```
resource "azurerm_resource_group" "example" {
  name     = "loggingRG"
  location = "eastus2"
}

resource "azurerm_automation_account" "example" {
  name                = "testaccount"
  location            = "${azurerm_resource_group.example.location}"
  resource_group_name = "${azurerm_resource_group.example.name}"

  sku {
    name = "Basic"
  }
}

data "local_file" "example" {
  filename = "${path.module}/Users/testuser/tftest/Conncetionscript.ps1"
}

resource "azurerm_automation_runbook" "example" {
  name                             = "Conncetionscript"
  location                         = "${azurerm_resource_group.example.location}"
  resource_group_name = "${azurerm_resource_group.example.name}"
  account_name              = "${azurerm_automation_account.example.name}"
  log_verbose                  = "true"
  log_progress                = "true"
  description                    = "testing tf template"
  runbook_type                = "PowerShell"

  depends_on = ["azurerm_automation_account.example"]

  publish_content_link {
    uri = "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/101-automation-runbook-getvms/Runbooks/Get-AzureVMTutorial.ps1"
  }

  content = "${data.local_file.example.filename}"
}
```

**Debug Output**
* azurerm_automation_runbook.runbook_start: Error setting the draft Automation Runbook automation.RunbookDraftClient#ReplaceContent: Failure sending request: StatusCode=404 -- Original Error: Code="ResourceNotFound" Message="The Resource 'Microsoft.Automation/automationAccounts/testaccount/runbooks/Conncetionscript' under resource group 'loggingRG' was not found." "

**Expected Behavior**
Runbook should be created

**Actual Behavior**
Looking for creating runbook which doesn't exist

**Steps to Reproduce**
terraform apply

**Terraform Version**
Terraform v0.12.7
+ provider.azurerm v1.34.0
+ provider.local v1.3.0

ratmav:
i ran into this today with the both the [regular and custom content examples listed in the documentation](https://www.terraform.io/docs/providers/azurerm/r/automation_runbook.html).

MathieuBuisson:
I ran into this problem as well, using the `content` argument in my case.
Terraform version : 0.12.9
[...]

Issue #3086, opened 2019-03-20
azurerm_automation_runbook can't be created, tries to find runbook when not existing.


mariobede:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

╰─ terraform -v
Terraform v0.11.13
+ provider.azurerm v1.23.0

### Affected Resource(s)

* `azurerm_automation_runbook`

### Terraform Configuration Files
```
resource "azurerm_automation_runbook" "runbook" {
  name                = "dasd"
  location            = "northeurop"
  resource_group_name = "en1tstma01aut-rg"
  account_name        = "en1tstma01aut"
  log_verbose         = true
  log_progress        = true
  description         = "powershell"
  runbook_type        = "Powershell"
  content             = "${data.local_file.example.content}"

  publish_content_link {
    uri = "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/101-automation-runbook-getvms/Runbooks/Get-AzureVMTutorial.ps1"
  }
}
```

### Debug Output
```
module.rbk.module.azureAutomationRunbook.azurerm_automation_runbook.runbook: Creating...
  account_name:               "" => "en1tstma01aut"
[...]

PR #6813, opened 2020-05-07
`azurerm_automation_runbook` - add support for create runbook as draft and then publish

njuCZ:
according to https://github.com/Azure/azure-rest-api-specs/blob/master/specification/automation/resource-manager/Microsoft.Automation/stable/2015-10-31/examples/createRunbookAsDraft.json, we could create runbook with no `publish_content_link`

this PR add supports for it. This is also the same experience with the portal

Additionally, I think exposing `publish_content_link` is a little confusing for users, and it tends to reports wield error. I wonder Could we make it deprecated and removed in next version ? 

relevant issues:
#6588 
#4403 
#3086

image

### DIFF
--- a/azurerm/internal/services/automation/automation_runbook_resource.go
+++ b/azurerm/internal/services/automation/automation_runbook_resource.go
@@ -306,7 +306,7 @@ func resourceArmAutomationRunbookDelete(d *schema.ResourceData, meta interface{}
 }
 
 func expandContentLink(inputs []interface{}) *automation.ContentLink {
-	if len(inputs) == 0 {
+	if len(inputs) == 0 || inputs[0] == nil {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/automation/automation_runbook_resource.go
```diff
@@ -298,20 +305,22 @@ func resourceArmAutomationRunbookDelete(d *schema.ResourceData, meta interface{}
 	return nil
 }
 
-func expandContentLink(d *schema.ResourceData) automation.ContentLink {
-	inputs := d.Get("publish_content_link").([]interface{})
+func expandContentLink(inputs []interface{}) *automation.ContentLink {
+	if len(inputs) == 0 {
```
mbfrahry, line 309: We should also check for nil 
```suggestion
	if len(inputs) == 0 || inputs[0] == nil {
```

PR #6819, opened 2020-05-07
[BUG] azurerm_redis_cache backup requires connection string

cemckenzie:
When creating a redis-cache resource, the rdb-storage-connection-string in the redis_configuration is required is rdb_backup_enabled is true.

Currently, an error is thrown during the create:

`redis.Client#Create: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidRequestBody" Message="The value of the parameter 'properties.redisConfiguration.rdb-storage-connection-string' is invalid."`

This change updates the error to:

`Error: Error parsing Redis Configuration: The rdb_storage_connection_string property must be set when rdb_backup_enabled is true
on <PATH>/main.tf line 23`

It also updates the docs.

### DIFF
--- a/azurerm/internal/services/redis/resource_arm_redis_cache.go
+++ b/azurerm/internal/services/redis/resource_arm_redis_cache.go
@@ -674,7 +674,7 @@ func expandRedisConfiguration(d *schema.ResourceData) (map[string]*string, error
 	if v, ok := d.GetOk("redis_configuration.0.rdb_backup_enabled"); ok {
 		backupEnabled := v.(bool)
 		if backupEnabled == true {
-			if v, ok := d.GetOk("redis_configuration.0.rdb_storage_connection_string"); !ok || v.(string) != "" {
+			if connStr, connOk := d.GetOk("redis_configuration.0.rdb_storage_connection_string"); !connOk || connStr.(string) == "" {
 				return nil, fmt.Errorf("The rdb_storage_connection_string property must be set when rdb_backup_enabled is true")
 			}
 		}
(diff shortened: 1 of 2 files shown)


website/docs/r/redis_cache.html.markdown
```diff
@@ -91,7 +91,7 @@ A `redis_configuration` block supports the following:
 
 * `maxfragmentationmemory_reserved` - (Optional) Value in megabytes reserved to accommodate for memory fragmentation. Defaults are shown below.
 
-* `rdb_backup_enabled` - (Optional) Is Backup Enabled? Only supported on Premium SKU's.
+* `rdb_backup_enabled` - (Optional) Is Backup Enabled? Only supported on Premium SKU's. If enabled, `rdb_storage_connection_string` must also be set.
```
katbyte, line 94: We should make this a note so it stands out

azurerm/internal/services/redis/resource_arm_redis_cache.go
```diff
@@ -672,6 +672,12 @@ func expandRedisConfiguration(d *schema.ResourceData) (map[string]*string, error
 
 	// RDB Backup
 	if v, ok := d.GetOk("redis_configuration.0.rdb_backup_enabled"); ok {
+		backupEnabled := v.(bool)
+		if backupEnabled == true {
+			if _, ok := d.GetOk("redis_configuration.0.rdb_storage_connection_string"); !ok {
```
katbyte, line 677: we should also check that its empty
```suggestion
			if v, ok := d.GetOk("redis_configuration.0.rdb_storage_connection_string"); !ok || v.(string) != ""{
```

website/docs/r/redis_cache.html.markdown
cemckenzie: updated to be a -> **NOTE:**

PR #6834, opened 2020-05-09
azurerm_api_management_named_value - fix the non empty plan when `secret` is true

yupwei68:
Fix https://github.com/terraform-providers/terraform-provider-azurerm/issues/6688

image


yupwei68:
Hi @katbyte, I don't know why. I still get tests passed.
image

PR #6858, opened 2020-05-11
`azurerm_eventgrid_topic` - support for `input_schema`, `input_mapping_fields`, and `input_mapping_default_values`

jrauschenbusch:
Adds support for customized input schema and mapping to resource `azurerm_eventgrid_topic`.

Additions to resource `azurerm_eventgrid_topic`:
- input_schema
- input_mapping_fields
- input_mapping_default_values

### DIFF
--- a/azurerm/internal/services/eventgrid/eventgrid_topic_resource.go
+++ b/azurerm/internal/services/eventgrid/eventgrid_topic_resource.go
@@ -315,47 +315,53 @@ func expandAzureRmEventgridTopicInputMapping(d *schema.ResourceData) *eventgrid.
 
 	if imfok {
 		mappings := imf.([]interface{})
-		mapping := mappings[0].(map[string]interface{})
+		if len(mappings) > 0 && mappings[0] != nil {
+			if mapping := mappings[0].(map[string]interface{}); mapping != nil {
 
-		if id := mapping["id"].(string); id != "" {
-			jismp.ID = &eventgrid.JSONField{SourceField: &id}
-		}
+				if id := mapping["id"].(string); id != "" {
+					jismp.ID = &eventgrid.JSONField{SourceField: &id}
+				}
 
-		if eventTime := mapping["event_time"].(string); eventTime != "" {
-			jismp.EventTime = &eventgrid.JSONField{SourceField: &eventTime}
-		}
+				if eventTime := mapping["event_time"].(string); eventTime != "" {
+					jismp.EventTime = &eventgrid.JSONField{SourceField: &eventTime}
+				}
 
-		if topic := mapping["topic"].(string); topic != "" {
-			jismp.Topic = &eventgrid.JSONField{SourceField: &topic}
-		}
+				if topic := mapping["topic"].(string); topic != "" {
+					jismp.Topic = &eventgrid.JSONField{SourceField: &topic}
+				}
 
-		if dataVersion := mapping["data_version"].(string); dataVersion != "" {
-			jismp.DataVersion = &eventgrid.JSONFieldWithDefault{SourceField: &dataVersion}
-		}
+				if dataVersion := mapping["data_version"].(string); dataVersion != "" {
+					jismp.DataVersion = &eventgrid.JSONFieldWithDefault{SourceField: &dataVersion}
+				}
 
-		if subject := mapping["subject"].(string); subject != "" {
-			jismp.Subject = &eventgrid.JSONFieldWithDefault{SourceField: &subject}
-		}
+				if subject := mapping["subject"].(string); subject != "" {
+					jismp.Subject = &eventgrid.JSONFieldWithDefault{SourceField: &subject}
+				}
 
-		if eventType := mapping["event_type"].(string); eventType != "" {
-			jismp.EventType = &eventgrid.JSONFieldWithDefault{SourceField: &eventType}
+				if eventType := mapping["event_type"].(string); eventType != "" {
+					jismp.EventType = &eventgrid.JSONFieldWithDefault{SourceField: &eventType}
+				}
+			}
 		}
 	}
 
 	if imdvok {
 		mappings := imdv.([]interface{})
-		mapping := mappings[0].(map[string]interface{})
+		if len(mappings) > 0 && mappings[0] != nil {
+			if mapping := mappings[0].(map[string]interface{}); mapping != nil {
 
-		if dataVersion := mapping["data_version"].(string); dataVersion != "" {
-			jismp.DataVersion = &eventgrid.JSONFieldWithDefault{DefaultValue: &dataVersion}
-		}
+				if dataVersion := mapping["data_version"].(string); dataVersion != "" {
+					jismp.DataVersion = &eventgrid.JSONFieldWithDefault{DefaultValue: &dataVersion}
+				}
 
-		if subject := mapping["subject"].(string); subject != "" {
-			jismp.Subject = &eventgrid.JSONFieldWithDefault{DefaultValue: &subject}
-		}
+				if subject := mapping["subject"].(string); subject != "" {
+					jismp.Subject = &eventgrid.JSONFieldWithDefault{DefaultValue: &subject}
+				}
 
-		if eventType := mapping["event_type"].(string); eventType != "" {
-			jismp.EventType = &eventgrid.JSONFieldWithDefault{DefaultValue: &eventType}
+				if eventType := mapping["event_type"].(string); eventType != "" {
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/eventgrid/eventgrid_topic_resource.go
```diff
@@ -196,3 +301,134 @@ func resourceArmEventGridTopicDelete(d *schema.ResourceData, meta interface{}) e
...
+		}
+
+		if eventType := mapping["event_type"].(string); eventType != "" {
+			jismp.EventType = &eventgrid.JSONFieldWithDefault{SourceField: &eventType}
+		}
+	}
+
+	if imdvok {
+		mappings := imdv.([]interface{})
+		mapping := mappings[0].(map[string]interface{})
```
mbfrahry, line 347: Similar to the above, this could panic and we should check it before casting

azurerm/internal/services/eventgrid/eventgrid_topic_resource.go
```diff
@@ -51,6 +51,85 @@ func resourceArmEventGridTopic() *schema.Resource {
...
+			},
+
+			"input_mapping_fields": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
```
mbfrahry, line 73: Should we try and force a user to pass in at least one of these attributes? I'm wondering what happens if a user doesn't pass anything into this for whatever reason. We could add 
```
AtLeastOneOf: []string{"input_mapping_files.0.id", "input_mapping_files.0.topic".....}
```
to achieve this

azurerm/internal/services/eventgrid/eventgrid_topic_resource.go
```diff
@@ -51,6 +51,85 @@ func resourceArmEventGridTopic() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"event_type": {
+							Type:     schema.TypeString,
+							ForceNew: true,
+							Optional: true,
```
mbfrahry, line 117: We could also add `AtLeastOneOf` here

mbfrahry (changes requested):
Hey @jrauschenbusch, thanks for opening this! I just have a few things that we could do to harden the attributes and prevent some crashes

azurerm/internal/services/eventgrid/eventgrid_topic_resource.go
jrauschenbusch: I agree. It would be really nice to give some hint to the end user that there's maybe a misuse, but `AtLeastOneOf` is in conflict with the `Optional` flag. But it has to be optional as it the input_mapping is only used as soon as the input_schema is changed to `CustomEventSchema`. Per default `EventGridSchema` is used. In this case an input_mapping is just ignored.

azurerm/internal/services/eventgrid/eventgrid_topic_resource.go
```diff
@@ -196,3 +301,134 @@ func resourceArmEventGridTopicDelete(d *schema.ResourceData, meta interface{}) e
...
+
+	if !imfok && !imdvok {
+		return nil
+	}
+
+	jismp := eventgrid.JSONInputSchemaMappingProperties{}
+
+	if imfok {
+		mappings := imf.([]interface{})
+		mapping := mappings[0].(map[string]interface{})
```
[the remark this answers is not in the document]
jrauschenbusch: fixed

azurerm/internal/services/eventgrid/eventgrid_topic_resource.go
jrauschenbusch: fixed

azurerm/internal/services/eventgrid/eventgrid_topic_resource.go
jrauschenbusch: see answer above

azurerm/internal/services/eventgrid/eventgrid_topic_resource.go
jrauschenbusch: I've made some tests now.

When using `EventGridSchema` or `CloudEventV01Schema` as input_schema, and we add an (empty) input_mapping_fields and/or input_mapping_default_values block, the following error will be reported:
```sh
Error: eventgrid.TopicsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidRequest" Message="InputSchemaMapping must not be provided when InputSchema is EventGridSchema or CloudEventV01Schema."
```

When using `CustomEventSchema` as input_schema and adding empty input_mapping_fields and input_mapping_default_values blocks the following will be reported:
```sh
Error: eventgrid.TopicsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidRequest" Message="Invalid InputSchemaMapping: DataVersion entry cannot be null, and either sourceField or defaultValue must be specified."
```

After adding a sourceField or defaultValue for `data_version` the following is reported:
```sh
Error: eventgrid.TopicsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidRequest" Message="Invalid InputSchemaMapping: Invalid Subject. Either sourceField or defaultValue should be specified."
```

After adding a sourceField or defaultValue for `subject` the following is reported:
```sh
Error: eventgrid.TopicsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidRequest" Message="Invalid InputSchemaMapping: Invalid EventType. Either sourceField or defaultValue should be specified."
```

After adding a sourceField or defaultValue for `event_type` the resource could finally be created.

Hence at Plan time we would need the following checks:
- if `input_schema != "CustomEventSchema"`, then the blocks input_mapping_fields and input_mapping_default_values must not be present
- if `input_schema == "CustomEventSchema"`, then `subject`, `event_type` and `data_version` are required in one of the input_mapping(_default) blocks.

mbfrahry: So I have an idea to make this easier on users and it's to split these attributes out into their own respective lists. `InputSchemaCloudEventSchemaV10` and `InputSchemaEventGridSchema` can just be booleans that conflict with each other like `is_event_grid_schema` and then have `custom_event_schema_input_mapping` with all of these variables in there. That will let us better differentiate the `Required` and `Optional` attributes. While being harder from a maintainer perspective, it should really help users see how all of the pieces here fit together.

jrauschenbusch: I think this is making things even more complex, also for the users. As there are not only `input_mapping_fields` to be considered but also `input_mapping_default_values`.

I think your design proposal would be look like: 
```terraform
resource "azurerm_evengrid_topic" "example" {

   ...

   enable_event_grid_schema = true // Optional, Default,  ConflictsWith ["enable_cloud_event_schema_v10", "custom_event_schema_input_mapping"]

   enable_cloud_event_schema_v10 = true  // Optional, ConflictsWith ["enable_event_grid_schema", "custom_event_schema_input_mapping"]

   custom_event_schema_input_mapping { // Optional,  ConflictsWith ["enable_event_grid_schema", "enable_cloud_event_schema_v10"]
      mapping_fields {
        id = "path.id" // optional
        topic= "path.topic" // optional
        event_time= "path.eventtime" // optional
        subject = "path.subject" // required either here
        event_type = "path.type" //  required either here
        data_version = "path.version"  //  required either here
      }
      default_mappings {
        subject = "foo" // or required here
        event_type = "bar" // or required here
        data_version = "" // or required here
      }
   }
}
```

So far I've  aligned the design with the existing resource `azurerm_eventgrid_domain`. (see https://www.terraform.io/docs/providers/azurerm/r/eventgrid_domain.html#input_schema)

LMKWYT

mbfrahry: Your design is what I was thinking but you're right that the burden is a little too high on both sides. I appreciate that you took the time to write it out though. 

I wasn't aware that `azurerm_eventgrid_domain` also did something similar so I'm a fan of keeping things consistent between that resource and this one.

mbfrahry (approved):
LGTM!

PR #6860, opened 2020-05-11
Update Resource: `azurerm_eventgrid_event_subscription`

### LINKED ISSUE #4308: Event Grid subscription with Service Bus Queue endpoint type
At the moment there is no Endpoint type _Service Bus Queue_ for Event Grid subscriptions. Any plans to implement this in the near future? What is the alternative? Using _azurerm_template_deployment_ instead?

### LINKED ISSUE #4857: azurerm_eventgrid_event_subscription support for Service Bus Queue endpoint
### Community Note

### Description

There are currently 5 endpoint types available when configuring an event grid event subscription (both through Azure portal and az cli):

- Webhook
- Storage Queue
- Event Hub
- Hybrid Connection
- Service Bus Queue

azurerm_eventgrid_event_subscription currently supports the first 4 of these, but not Service Bus Queue.

* azurerm_eventgrid_event_subscription

```
resource "azurerm_eventgrid_event_subscription" "default" {
  name  = "defaultEventSubscription"
  scope = azurerm_resource_group.default.id

  service_bus_queue_endpoint {
    connection_string = azurerm_servicebus_queue_authorization_rule.default.primary_connection_string
  }
}
```

* #0000

jrauschenbusch:
Updates/extends resource `azurerm_eventgrid_event_subscription`

Additions to resource `azurerm_eventgrid_event_subscription`:
- service_bus_queue_endpoint
- service_bus_topic_endpoint
- expiration_time_utc

Fixes #4308
Fixes #4857


azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -144,6 +152,54 @@ func resourceArmEventGridEventSubscription() *schema.Resource {
 				},
 			},
 
+			"service_bus_queue_endpoint": {
```
katbyte, line 155: Given that we can only have 1 endpoint of 1 type, with 1 property, would it make sense to flatten these to: `service_bus_queue_endpoint_id`, `service_bus_topic_endpoint_id`, and `azure_function_endpoint_id` ? WDYT? unless there are known plans to add additional properties to these blocks soon not sure theres any advantage to making these named blocks with a single property

azurerm/internal/services/eventgrid/tests/eventgrid_event_subscription_resource_test.go
```diff
@@ -355,6 +418,116 @@ resource "azurerm_eventgrid_event_subscription" "test" {
 `, data.RandomInteger, data.Locations.Primary, data.RandomInteger, data.RandomInteger, data.RandomInteger)
 }
 
+func testAccAzureRMEventGridEventSubscription_serviceBusQueue(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-%d"
```
katbyte, line 427: would we make this
```suggestion
  name     = "acctestRG-eg-%d"
```

azurerm/internal/services/eventgrid/tests/eventgrid_event_subscription_resource_test.go
```diff
@@ -355,6 +418,116 @@ resource "azurerm_eventgrid_event_subscription" "test" {
...
+  location = "%s"
+}
+resource "azurerm_servicebus_namespace" "example" {
```
katbyte, line 430: and put new lines here to be consistent with the restof the tests
```suggestion
}

resource "azurerm_servicebus_namespace" "example" {
```

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
jrauschenbusch: I totally agree that this design is a bit clumsy. However, as you can see, there are several other types of endpoints all require either a resource ID, a URL, a queue name, or other endpoint-specific details. So I tried to stick to the current approach.

katbyte: I would argue that the other ones should be deprecated and simplified ie webhook_endpoint should become webhook_endpoint_url. I'm ok with `storage_queue_endpoint` being the odd one out as a block as it does have multiple properties to set for the sake of cleaning up everything else.

jrauschenbusch: Here my proposal for the resource schema:
```terraform
{
  name =  "..."
  scope = "..."

  "endpoint" {

    // Possible types:
    //   - event_hub
    //   - storage_queue
    //   - service_bus_queue
    //   - service_bus_topic
    //   - azure_function
    //   - webhook

    "type"         = "event_hub"   // Required

    // **)
    //   - event_hub
    //   - storage_queue
    //   - service_bus_queue
    //   - service_bus_topic
    //   - azure_function
    "id"           = "..."  // Optional (Only required for **)

    "webhook_url"  = "..."  // Optional (Only required for type webhook)

    "queue_name"   = "..."  // Optional (Only required for type storage_queue)
  }

}
```

katbyte: i'm not sure i'm sold on a shared endpoint block. It seems more confusing then a set of top level properties that are explity in their requirements and can easily be validated during plan (including the required properties for the block one and ensuring the ID can be parsed correctly for that type of resoruce) vs a a block where some are required as well as a new property defining the type and a id validation that changes depending on the type. Seems like a lot of added complexity for a more confusing ux?

jrauschenbusch: ah ok. now I got it, this wasn't my favorite either. I just thought you were concerned with the redundancy of the ID attributes. But it was just about nesting in separate blocks for most of the endpoint types except the storage_account_queue, as it has several attributes to be defined. sorry for my delay ... 🤦

jrauschenbusch: Hey @katbyte,

after investigating the latest EventGrid model definition inside the *azure-sdk-for-go* i'm not sure if flattening the endpoint blocks is really a good approach (beside the fact that this would be a breaking change and a lot of issue tickets would be the result).

For example the webhook endpoint supports further configuration options, which are currently not implemented within the azurerm provider resource (batch size):
https://github.com/Azure/azure-sdk-for-go/blob/master/services/preview/eventgrid/mgmt/2020-04-01-preview/eventgrid/models.go#L8168

As there were a lot of updates since the latest *azure-sdk-for-go* release i would think that it's really likely that the other endpoints could also become some more configuration options in the future.

WDYT?

katbyte: If there are additional properties not exposed for config then yea makes sense to leave it as a block! but one which are a single property i'd rather start without the block and then if/when they do add more covert it into a block (also means people will have to be aware of that change)

Theres no need for you to update the existing ones if you don't want, but for schema changes like this what we do is deprecate the old ones, conflict them with the new property then set both on read, and on create get the old one, and if the new one is populated, get and use that instead

jrauschenbusch: ah, ok, that's how it makes sense for me. I'll add the changes to have this "Deprecated / ConflictWith" stuff in it, and hopefully we'll get this PR done soon. In addition to the `storage_queue_endpoint` block, the `webhook_endpoint` block will also remain unchanged, as it makes sense to include support for batch configuration properties in the near future.

azurerm/internal/services/eventgrid/parse/eventgrid_event_subscription.go
```diff
@@ -25,14 +25,7 @@ func EventGridEventSubscriptionID(input string) (*EventGridEventSubscriptionId,
...
 		Scope: segments[0],
-	}
-
-	if eventSubscription.Name, err = id.PopSegment("eventSubscriptions"); err != nil {
-		return nil, err
-	}
-
-	if err := id.ValidateNoEmptySegments(input); err != nil {
-		return nil, err
+		Name:  segments[1],
```
jrauschenbusch, line 28: Hotfix was necessary here for testing purposes.

jrauschenbusch:
@katbyte Can you please review may latest update (Deprecations / ConflictsWith / Constants / Fixes in ExpirationTime expanding / flattening)?

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -69,41 +82,65 @@ func resourceArmEventGridEventSubscription() *schema.Resource {
...
+			},
+
 			"topic_name": {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
 			},
 
-			"storage_queue_endpoint": {
+			AzureFunctionEndpointID: {
```
katbyte, line 97: Could we leave these as
```suggestion
			"azure_function_endpoint_id: {
```

it becomes hard to read the schema when we abstract this out

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -69,41 +82,65 @@ func resourceArmEventGridEventSubscription() *schema.Resource {
...
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: utils.RemoveFromStringArray(PossibleEnpointTypes(), AzureFunctionEndpointID),
```
katbyte, line 100: And ideally here to

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -318,34 +380,60 @@ func resourceArmEventGridEventSubscriptionRead(d *schema.ResourceData, meta inte
...
+			if err := d.Set(EventHubEndpoint, flattenEventGridEventSubscriptionEventHubEndpoint(eventHubEndpoint)); err != nil {
+				return fmt.Errorf("Error setting `%q` for EventGrid Event Subscription %q (Scope %q): %s", EventHubEndpoint, id.Name, id.Scope, err)
 			}
 		}
 		if hybridConnectionEndpoint, ok := props.Destination.AsHybridConnectionEventSubscriptionDestination(); ok {
-			if err := d.Set("hybrid_connection_endpoint", flattenEventGridEventSubscriptionHybridConnectionEndpoint(hybridConnectionEndpoint)); err != nil {
-				return fmt.Errorf("Error setting `hybrid_connection_endpoint` for EventGrid Event Subscription %q (Scope %q): %s", id.Name, id.Scope, err)
+			if err := d.Set(HybridConnectionEndpointID, flattenEventGridEventSubscriptionHybridConnectionEndpointID(hybridConnectionEndpoint)); err != nil {
+				return fmt.Errorf("Error setting `%q` for EventGrid Event Subscription %q (Scope %q): %s", HybridConnectionEndpointID, id.Name, id.Scope, err)
+			}
```
katbyte, line 409: as these are not a block we don't need to have a flatten function, we can just set them directly

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -406,28 +494,61 @@ func resourceArmEventGridEventSubscriptionDelete(d *schema.ResourceData, meta in
...
-	if _, ok := d.GetOk("storage_queue_endpoint"); ok {
-		return expandEventGridEventSubscriptionStorageQueueEndpoint(d)
+	if _, ok := d.GetOk(AzureFunctionEndpointID); ok {
+		return expandEventGridEventSubscriptionAzureFunctionEndpointID(d)
 	}
 
-	if _, ok := d.GetOk("eventhub_endpoint"); ok {
+	if _, ok := d.GetOk(EventHubEndpointID); ok {
+		return expandEventGridEventSubscriptionEventHubEndpointID(d)
+	} else if _, ok := d.GetOk(EventHubEndpoint); ok {
```
katbyte, line 521: and similarly for these we can
```suggestion
	if v, ok := d.GetOk(EventHubEndpointID); ok {
		return &eventgrid.EventHubEventSubscriptionDestination{
			EndpointType: eventgrid.EndpointTypeEventHub,
			EventHubEventSubscriptionDestinationProperties: &eventgrid.EventHubEventSubscriptionDestinationProperties{
				ResourceID: utils.String(v.(string)),
			},)
	} else if _, ok := d.GetOk(EventHubEndpoint); ok {
```

website/docs/r/eventgrid_event_subscription.html.markdown
```diff
@@ -87,22 +99,21 @@ A `storage_queue_endpoint` supports the following:
 * `queue_name` - (Required) Specifies the name of the storage queue where the Event Subscriptio will receive events.
 
 ---
+A `eventhub_endpoint` supports the following:	
 
-A `eventhub_endpoint` supports the following:
+* `eventhub_id` - (Required) Specifies the id of the eventhub where the Event Subscription will receive events.
```
katbyte, line 104: we have some trailing whitespace on these lines
```suggestion
* `eventhub_id` - (Required) Specifies the id of the eventhub where the Event Subscription will receive events.```

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
jrauschenbusch: After completing the refactoring, thoughts came in the same direction. Indeed, (schema) key abstraction is not a good coding practice. I'll change this immediately.

jrauschenbusch:
@katbyte no worries 😄everything should be fixed now. plz check it out.

PR #6861, opened 2020-05-11
Update Resource: Add advanced filter support to `azurerm_eventgrid_event_subscription`

### LINKED ISSUE #3452: azurerm_eventgrid_event_subscription support advanced filtering
### Community Note

### Description

Looking for the advanced filtering option.  Presently it has subject filtering for beginswith and endswith which is good, but my use-case was to use the advance filter to search the subject for a string that's contained in the subject "i.e. virtualMachines, storageAccounts".

* azurerm_eventgrid_event_subscription

subject_filter {
    string_contains = "blah"
}

```hcl
```

### References

* #0000

jrauschenbusch:
Adds advanced_filter support in resource `azurerm_eventgrid_event_subscription`.

Fixes #3452

### DIFF
--- a/azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
+++ b/azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
@@ -228,6 +228,247 @@ func resourceArmEventGridEventSubscription() *schema.Resource {
 				},
 			},
 
+			"advanced_filter": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"bool_equals": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"value": {
+										Type:     schema.TypeBool,
+										Required: true,
+									},
+								},
+							},
+						},
+						"number_greater_than": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"value": {
+										Type:     schema.TypeFloat,
+										Required: true,
+									},
+								},
+							},
+						},
+						"number_greater_than_or_equals": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"value": {
+										Type:     schema.TypeFloat,
+										Required: true,
+									},
+								},
+							}},
+						"number_less_than": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"value": {
+										Type:     schema.TypeFloat,
+										Required: true,
+									},
+								},
+							},
+						},
+						"number_less_than_or_equals": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"value": {
+										Type:     schema.TypeFloat,
+										Required: true,
+									},
+								},
+							},
+						},
+						"number_in": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"values": {
+										Type:     schema.TypeList,
+										Required: true,
+										MaxItems: 5,
+										Elem: &schema.Schema{
+											Type: schema.TypeFloat,
+										},
+									},
+								},
+							},
+						},
+						"number_not_in": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"values": {
+										Type:     schema.TypeList,
+										Required: true,
+										MaxItems: 5,
+										Elem: &schema.Schema{
+											Type: schema.TypeFloat,
+										},
+									},
+								},
+							},
+						},
+						"string_begins_with": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"values": {
+										Type:     schema.TypeList,
+										Required: true,
+										MaxItems: 5,
+										Elem: &schema.Schema{
+											Type: schema.TypeString,
+										},
+									},
+								},
+							},
+						},
+						"string_ends_with": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"values": {
+										Type:     schema.TypeList,
+										Required: true,
+										MaxItems: 5,
+										Elem: &schema.Schema{
+											Type: schema.TypeString,
+										},
+									},
+								},
+							},
+						},
+						"string_contains": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"values": {
+										Type:     schema.TypeList,
+										Required: true,
+										MaxItems: 5,
+										Elem: &schema.Schema{
+											Type: schema.TypeString,
+										},
+									},
+								},
+							},
+						},
+						"string_in": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"values": {
+										Type:     schema.TypeList,
+										Required: true,
+										MaxItems: 5,
+										Elem: &schema.Schema{
+											Type: schema.TypeString,
+										},
+									},
+								},
+							},
+						},
+						"string_not_in": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"key": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"values": {
+										Type:     schema.TypeList,
+										Required: true,
+										MaxItems: 5,
+										Elem: &schema.Schema{
+											Type: schema.TypeString,
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+
 			"storage_blob_dead_letter_destination": {
 				Type:     schema.TypeList,
 				MaxItems: 1,
@@ -306,7 +547,10 @@ func resourceArmEventGridEventSubscriptionCreateUpdate(d *schema.ResourceData, m
 		return fmt.Errorf("One of the following endpoint types must be specificed to create an EventGrid Event Subscription: %q", enpointPropertyNames())
 	}
 
-	filter := expandEventGridEventSubscriptionFilter(d)
+	filter, err := expandEventGridEventSubscriptionFilter(d)
+	if err != nil {
+		return fmt.Errorf("expanding filters for EventGrid Event Subscription %q (Scope %q): %+v", name, scope, err)
+	}
 
 	expirationTime, err := expandEventGridExpirationTime(d)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -176,6 +180,58 @@ func resourceArmEventGridEventSubscription() *schema.Resource {
...
+								string(eventgrid.OperatorTypeStringContains),
+								string(eventgrid.OperatorTypeStringEndsWith),
+								string(eventgrid.OperatorTypeStringIn),
+								string(eventgrid.OperatorTypeStringNotIn),
+							}, false),
+						},
+						"value": {
+							Type:             schema.TypeString,
+							Optional:         true,
+							ConflictsWith:    []string{"advanced_filter.values"},
```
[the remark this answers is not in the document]
jrauschenbusch: yes, i know... 😒there is an ticket that was created an eternity ago in the terrarform-plugin-sdk to address this topic (https://github.com/hashicorp/terraform-plugin-sdk/issues/71).

I already handled it so i remove this in the schema part.

see https://github.com/terraform-providers/terraform-provider-azurerm/pull/6861/commits/5079531a23fd972e9545b2e1e2f296c2873549bd#diff-e960883f0833ba5c3ab01eb6f75e2482R593

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -498,7 +560,129 @@ func expandEventGridEventSubscriptionFilter(d *schema.ResourceData) *eventgrid.E
...
+	case string(eventgrid.OperatorTypeStringEndsWith):
+		return eventgrid.StringEndsWithAdvancedFilter{Key: &key, OperatorType: eventgrid.OperatorTypeStringEndsWith, Values: &values}, nil
+	case string(eventgrid.OperatorTypeStringContains):
+		return eventgrid.StringContainsAdvancedFilter{Key: &key, OperatorType: eventgrid.OperatorTypeStringContains, Values: &values}, nil
+	default:
+		return nil, nil
+	}
+}
+
+func expandScalarAdvancedFilter(key string, operatorType string, value string) (eventgrid.BasicAdvancedFilter, error) {
```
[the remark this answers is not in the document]
jrauschenbusch: Then i would prefer this one:

```terraform
advanced_filter {
  number_less_than {
     key   = "data.filesize"
     value = 42.0
  }
  string_begins_with {
    key    = "data.name"
    values = ["foo", "bar"]
  }
}
```

Multiple nested filter blocks (also equal ones) should then be allowed within the advanced_filter block.

Reasons:
- operator_type defines the type of the filter
- operator_type defines if a single or multiple values have to be passed

WDYT?

azurerm/utils/type.go
```diff
@@ -0,0 +1,67 @@
...
+	numbers := []string{}
+	for _, f := range values {
+		number := FormatFloat(f)
+		numbers = append(numbers, number)
+	}
+	return numbers
+}
+
+func TypeRelaxedEquals(old string, new string) bool {
+	if strings.EqualFold(old, new) {
```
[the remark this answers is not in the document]
jrauschenbusch: It's really been a long time since I've implemented this. But as far as I remember it happened that the advanced_filter should be re-created because the state apparently deviated from the target state. The problem was only solved with this softened validation.

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -176,6 +180,58 @@ func resourceArmEventGridEventSubscription() *schema.Resource {
...
+							Optional:         true,
+							ConflictsWith:    []string{"advanced_filter.values"},
+							DiffSuppressFunc: AdvancedFilterDiffSuppressFunc,
+						},
+						"values": {
+							Type:          schema.TypeList,
+							MinItems:      1,
+							MaxItems:      5,
+							Optional:      true,
+							ConflictsWith: []string{"advanced_filter.value"},
```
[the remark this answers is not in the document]
jrauschenbusch: > hey @jrauschenbusch
> 
> Thanks for this PR :)
> 
> Taking a look through whilst this looks pretty good and works today, unfortunately reusing a single schema field for multiple times will break in a future version of Terraform - and as such we'll need to take a different approach here. Instead I'd suggest introducing a nested block for each type of variable, for example:
> 
> ```
> advanced_filter {
>   integer {
>     key = 3
>     operator_type = "LessThan"
>     values = [1, 2, 3]
>   }
> 
>   # or boolean { .. }
>   # or string { .. }
> }
> ```
> 
> Which allows us to have more granular type information here - and will work going forwards - as such could we update this to use that approach? From a validation side we can then use the "expand" method of the filter to check that only one of `boolean`, `integer` or `string` blocks are specified, which should give us the same thing - WDYT?
> 
> Thanks!

Hey @tombuildsstuff 😄

I think then of course we have to change that. See my proposal within the review comments and LMKWYT.

Thanks!

jrauschenbusch:
@tombuildsstuff Now have changed everything to my latest proposal. Tested it locally. Everything seems to work and schema validation is now a lot easier than before.

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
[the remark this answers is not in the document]
tombuildsstuff: @jrauschenbusch sounds good to me 👍

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -176,6 +176,247 @@ func resourceArmEventGridEventSubscription() *schema.Resource {
...
 			},
 
+			"advanced_filter": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"bool_equals": {
+							Type:     schema.TypeList,
```
tombuildsstuff, line 186: this (along with the other nested blocks) may want changing to a Set depending on how the API behaves with multiple elements, but this is fine for now 👍

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -254,7 +495,10 @@ func resourceArmEventGridEventSubscriptionCreateUpdate(d *schema.ResourceData, m
 		return fmt.Errorf("One of the following endpoint types must be specificed to create an EventGrid Event Subscription: %q", getEnpointTypes())
 	}
 
-	filter := expandEventGridEventSubscriptionFilter(d)
+	filter, err := expandEventGridEventSubscriptionFilter(d)
+	if err != nil {
+		return fmt.Errorf("Error creating/updating EventGrid Event Subscription %q (Scope %q): %s", name, scope, err)
```
tombuildsstuff, line 500: in this instance since this error is expanding the filter, we can raise this via:

```suggestion
		return fmt.Errorf("expanding filters for EventGrid Event Subscription %q (Scope %q): %+v", name, scope, err)
```

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -354,6 +598,9 @@ func resourceArmEventGridEventSubscriptionRead(d *schema.ResourceData, meta inte
 			if err := d.Set("subject_filter", flattenEventGridEventSubscriptionSubjectFilter(filter)); err != nil {
 				return fmt.Errorf("Error setting `subject_filter` for EventGrid Event Subscription %q (Scope %q): %s", id.Name, id.Scope, err)
 			}
+			if err := d.Set("advanced_filter", flattenEventGridEventSubscriptionAdvancedFilter(filter.AdvancedFilters)); err != nil {
```
tombuildsstuff, line 601: there's a panic here if `filter` is nil - can we pass `filter` in here instead?

```suggestion
			if err := d.Set("advanced_filter", flattenEventGridEventSubscriptionAdvancedFilter(filter)); err != nil {
```

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -606,6 +912,112 @@ func flattenEventGridEventSubscriptionSubjectFilter(filter *eventgrid.EventSubsc
...
+	numberGreaterThanOrEquals := make([]interface{}, 0)
+	numberLessThan := make([]interface{}, 0)
+	numberLessThanOrEquals := make([]interface{}, 0)
+	numberIn := make([]interface{}, 0)
+	numberNotIn := make([]interface{}, 0)
+	stringBeginsWith := make([]interface{}, 0)
+	stringEndsWith := make([]interface{}, 0)
+	stringContains := make([]interface{}, 0)
+	stringIn := make([]interface{}, 0)
+	stringNotIn := make([]interface{}, 0)
```
tombuildsstuff, line 932: since the `key`, `value` and `values` properties can be nil/omitted from the Azure API here - it'd be worth adding a couple of inline functions to assist:

```
var flattenValue = func(inputKey, inputValue *string) map[string]interface{} {
  key := ""
  if inputKey != null {
    key = *inputKey
  }
  value := ""
  if inputValue != null {
    value = *inputValue
  }

  return map[string]interface{}{
    "key": key,
    "value": value,
  }
}
var flattenValues = func(inputKey *string, inputValues *[]string) map[string]interface{} {
  key := ""
  if inputKey != null {
    key = *inputKey
  }
  values := make([]string, 0)
  if inputValues != null {
    values = *inputValues
  }

  return map[string]interface{}{
    "key": key,
    "values": values,
  }
}
```

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -606,6 +912,112 @@ func flattenEventGridEventSubscriptionSubjectFilter(filter *eventgrid.EventSubsc
...
+
+	for _, item := range *input {
+		switch f := item.(type) {
+		case eventgrid.BoolEqualsAdvancedFilter:
+			boolEquals = append(boolEquals, map[string]interface{}{
```
tombuildsstuff, line 937: which means this can become:

```suggestion
			boolEquals = append(boolEquals, flattenValues(f.Key, f.Value))
```

(I guess we could also pass in `f`)

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -606,6 +912,112 @@ func flattenEventGridEventSubscriptionSubjectFilter(filter *eventgrid.EventSubsc
...
+		"string_begins_with":            stringBeginsWith,
+		"string_ends_with":              stringEndsWith,
+		"string_contains":               stringContains,
+		"string_in":                     stringIn,
+		"string_not_in":                 stringNotIn,
+	}
+
+	results = append(results, filters)
+
+	return results
```
tombuildsstuff, line 1018: (not an issue, just for info) since we're hard-coding a single element here, we can reduce this to:

```suggestion
	return []interface{}{
		map[string][]interface{}{
 			"bool_equals":                   boolEquals,
 			"number_greater_than":           numberGreaterThan,
 			"number_greater_than_or_equals": numberGreaterThanOrEquals,
 			"number_less_than":              numberLessThan,
 			"number_less_than_or_equals":    numberLessThanOrEquals,
 			"number_in":                     numberIn,
 			"number_not_in":                 numberNotIn,
 			"string_begins_with":            stringBeginsWith,
 			"string_ends_with":              stringEndsWith,
 			"string_contains":               stringContains,
 			"string_in":                     stringIn,
 			"string_not_in":                 stringNotIn,
 		}
	}
```

azurerm/internal/services/eventgrid/parse/eventgrid_event_subscription.go
```diff
@@ -25,14 +25,7 @@ func EventGridEventSubscriptionID(input string) (*EventGridEventSubscriptionId,
...
 		Scope: segments[0],
-	}
-
-	if eventSubscription.Name, err = id.PopSegment("eventSubscriptions"); err != nil {
-		return nil, err
-	}
-
-	if err := id.ValidateNoEmptySegments(input); err != nil {
-		return nil, err
+		Name:  segments[1],
```
tombuildsstuff, line 28: (for passers by) this appears to be a rebase containing a change made the other day - but this is expected 👍

azurerm/internal/services/eventgrid/tests/eventgrid_event_subscription_resource_test.go
```diff
@@ -401,3 +424,52 @@ resource "azurerm_eventgrid_event_subscription" "test" {
...
+    queue_name         = "${azurerm_storage_queue.test.name}"
+  }
+  advanced_filter {
+    number_less_than {
+      key   = "data.contentLength"
+      value = 42.0
+    }
+    number_in {
+      key    = "data.contentLength"
+      values = [1, 1, 2, 3, 5]
```
tombuildsstuff, line 466: is that an intentional duplicate value?

jrauschenbusch: No, other than that it works with the API, there is no deeper meaning in it. Was just part of the Fibonacci sequence.

azurerm/internal/services/eventgrid/parse/eventgrid_event_subscription.go
jrauschenbusch: correct. it was just for testing purposes, as the other fix wasn't merged during the time i worked on this branch.

azurerm/internal/services/eventgrid/tests/eventgrid_event_subscription_resource_test.go
tombuildsstuff: ok, seems reasonable to me :)

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
jrauschenbusch: It seems that duplicate values are allowed (tested by myself). But in fact a set would make more sense.

jrauschenbusch: Unfortunately testing is a lot more complex using sets, as one has to know the computed hash value of the properties to test their values. Do you have an idea how to easily solve this?

jrauschenbusch:
Hey @tombuildsstuff 

I have fixed all PR notes as far as they were easy to implement. If there are other points, please feel free to fix them.

Thanks!

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -733,6 +1039,84 @@ func flattenEventGridEventSubscriptionSubjectFilter(filter *eventgrid.EventSubsc
 	return []interface{}{result}
 }
 
+func flattenEventGridEventSubscriptionAdvancedFilter(input *eventgrid.EventSubscriptionFilter) []interface{} {
+	results := make([]interface{}, 0)
+	if input == nil {
```
tombuildsstuff, line 1044: this also wants to check

```suggestion
	if input == nil || input.AdvancedFilters == nil {
```

jrauschenbusch:
@katbyte Cannot reproduce it using the latest branch version (https://github.com/terraform-providers/terraform-provider-azurerm/pull/6861/commits/63afc0e36826da2bbcd1ebfddb96d6e1e8a261f2). But indeed, using the branch version before applying the patch of @tombuildsstuff (https://github.com/terraform-providers/terraform-provider-azurerm/pull/6861/commits/933a60824128eb102b0af60162bf8834cbe74e0d) this error is thrown.

jrauschenbusch:
@tombuildsstuff any update here? does the ACC test still fail or is it ready to be merged?

katbyte:
Just checked now, same test is failing with:
```
TestAccAzureRMEventGridEventSubscription_advancedFilter: testing.go:640: Step 0 error: config is invalid: Unsupported argument: An argument named "resource_group_name" is not expected here.
```

jrauschenbusch:
@katbyte Oh, i'm so sorry. That was a mistake from my side. I'll invest a bit more time to fix and optimize the acceptance tests. I'll ping you back as soon as i'm done.

jrauschenbusch:
@katbyte I've now fixed/extended the Acc tests and executed it with my Azure account.

Test result:
```
make acctests SERVICE='eventgrid' TESTARGS='-run=TestAccAzureRMEventGridEventSubscription_advancedFilter' TESTTIMEOUT='60m'
TF_ACC=1 go test -v ./azurerm/internal/services/eventgrid/tests/ -run=TestAccAzureRMEventGridEventSubscription_advancedFilter -timeout 60m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMEventGridEventSubscription_advancedFilter
=== PAUSE TestAccAzureRMEventGridEventSubscription_advancedFilter
=== CONT  TestAccAzureRMEventGridEventSubscription_advancedFilter
--- PASS: TestAccAzureRMEventGridEventSubscription_advancedFilter (252.65s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/eventgrid/tests     (cached)
```

Sorry for the circumstances.

PR #6866, opened 2020-05-12
fix perfomance issue of keyvault

### LINKED ISSUE #6196: keyvault: caching of metadata required in larger terraform configurations
### Community Note

Terraform v0.12.23
provider.azurerm v2.2.0

### Affected Resource(s)

* `azurerm_key_vault`
* `azurerm_key_vault_secret`
* `azurerm_virtual_machine_scale_set_extension`

```hcl

resource "azurerm_virtual_machine_scale_set_extension" "vmss_ext_msi" {
  virtual_machine_scale_set_id = azurerm_virtual_machine_scale_set.vmss.id
  name                         = "ManagedIdentityWindowsExtension"
  publisher                    = "Microsoft.ManagedIdentity"
  type                         = "ManagedIdentityExtensionForWindows"
  type_handler_version         = "1.0"
  settings                     = "{\"port\": 50342}"
  auto_upgrade_minor_version   = true
}
resource "azurerm_key_vault_secret" "kv_secret_01" {
  name         = "mySecret"
  value        = var.my_value
  key_vault_id = azurerm_key_vault.kv.id
  content_type = "text/plain"

  tags = {
    name        = "My Parameters"
  }

  depends_on = [
    azurerm_key_vault.kv,
    azurerm_key_vault_access_policy.kv_some_policy,
  ]
}
resource "azurerm_key_vault" "kv" {
  name                            = "my-kv"
  resource_group_name             = azurerm_resource_group.rg.name
  location                        = azurerm_resource_group.rg.location
  tenant_id                       = var.tenant_id
  enabled_for_deployment          = true
  enabled_for_disk_encryption     = true
  enabled_for_template_deployment = true
  purge_protection_enabled        = true
  soft_delete_enabled             = true

  sku_name = "standard"

  network_acls {
    default_action = "Deny"
    bypass         = "AzureServices"

    ip_rules = ["***","***"]

    virtual_network_subnet_ids = [
      azurerm_subnet.sub1.id,
      azurerm_subnet.sub2.id,
      azurerm_subnet.sub3_deployment_machine.id,
    ]
  }

  tags = {
    name        = "my good kv"
  }
}
```

### Expected Behavior

Terraform/AzureRM should always able to create a plan. 

### Actual Behavior

Every "terraform plan" action occurs different errors but the original error is the same which is "Context Deadline Exceeded". Sometimes we can create a plan with the all same resources but today I tried more than 15 times and no luck. Please see below for some errors which I selected uniquely. This kind of error occurs multiple times and for different resources in every plan action. 

```hcl
Error: Error checking if key vault "/subscriptions/***/resourceGroups/***/providers/Microsoft.KeyVault/vaults/***" for Secret "***" in Vault at url "https://***.vault.azure.net/" exists: Error making Read request on KeyVault "***" (Resource Group "***"): keyvault.VaultsClient#Get: Failure sending request: StatusCode=429 -- Original Error: context deadline exceeded

Error: Error retrieving the Resource ID the Key Vault at URL "https://***.vault.azure.net/": Error making Read request on KeyVault "***" (Resource Group "***"): keyvault.VaultsClient#Get: Failure sending request: StatusCode=429 -- Original Error: context deadline exceeded

Error: Error retrieving Extension "ManagedIdentityWindowsExtension" (Virtual Machine Scale Set "***" / Resource Group "***"): compute.VirtualMachineScaleSetExtensionsClient#Get: Failure sending request: StatusCode=429 -- Original Error: context deadline exceeded

Error: Error retrieving the Resource ID the Key Vault at URL "https://***.vault.azure.net/": Error GetKeyVaultId unable to list Key Vaults keyvault.VaultsClient#List: Failure sending request: StatusCode=429 -- Original Error: context deadline exceeded

Error: Error making Read request on AzureRM App Service "***": web.AppsClient#Get: Failure sending request: StatusCode=429 -- Original Error: context deadline exceeded
```

### Steps to Reproduce

1. `terraform plan`

### Important Factoids

* Totally it's refreshing more than 1200 resources while planning. We have almost 80 key vaults and they contain lots of secrets. So most of the errors are related to Key Vaults. 
* These resources are already created and exist on Azure. So I can't use timeouts for creation and deletion. Also read timeouts didn't work. I've added it about 450 resources. I guess we need auto-retry..
* If we use resource targeting for those resources which we see on errors, we can create the plan successfully. So resources and our connections are healthy.
* We are only passing the "-refresh=true" parameter for plan action. 
* I've also tried with a parallelism parameter but no luck.
* Trying to plan from both a VM on Azure and my laptop.

njuCZ:
fix #6196

I think the root cause of this issue is every key/secret/certificate will list all key vaults, which causes the poor performance and Context Deadline Exceeded.

according to https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#objects-identifiers-and-versioning

the key vault name is global unique, and 
```
An object identifier has the following general format:

https://{keyvault-name}.vault.azure.net/{object-type}/{object-name}/{object-version}
```
we could directly get baseURL from keyvault name and get keyvault name from baseURL.

AS for get keyvault id from name, I use resourcesClient instead of listing to improve the performance.

image
image

### DIFF
--- a/azurerm/helpers/azure/key_vault.go
+++ b/azurerm/helpers/azure/key_vault.go
@@ -3,89 +3,41 @@ package azure
 import (
 	"context"
 	"fmt"
+	"regexp"
 
 	"github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2018-02-14/keyvault"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
-func GetKeyVaultBaseUrlFromID(ctx context.Context, client *keyvault.VaultsClient, keyVaultId string) (string, error) {
-	if keyVaultId == "" {
-		return "", fmt.Errorf("keyVaultId is empty")
+func GetKeyVaultNameFromBaseUrl(keyVaultUrl string) (string, error) {
+	groups := regexp.MustCompile(`^https://(.+)\.vault\.azure\.net/?$`).FindStringSubmatch(keyVaultUrl)
+	if len(groups) != 2 {
+		return "", fmt.Errorf("parsing keyVaultUrl: %q, expected group: 2", keyVaultUrl)
 	}
 
-	id, err := ParseAzureResourceID(keyVaultId)
-	if err != nil {
-		return "", err
-	}
-	resourceGroup := id.ResourceGroup
-
-	vaultName, ok := id.Path["vaults"]
-	if !ok {
-		return "", fmt.Errorf("resource id does not contain `vaults`: %q", keyVaultId)
-	}
-
-	resp, err := client.Get(ctx, resourceGroup, vaultName)
-	if err != nil {
-		if utils.ResponseWasNotFound(resp.Response) {
-			return "", fmt.Errorf("Error unable to find KeyVault %q (Resource Group %q): %+v", vaultName, resourceGroup, err)
-		}
-		return "", fmt.Errorf("Error making Read request on KeyVault %q (Resource Group %q): %+v", vaultName, resourceGroup, err)
-	}
-
-	if resp.Properties == nil || resp.Properties.VaultURI == nil {
-		return "", fmt.Errorf("vault (%s) response properties or VaultURI is nil", keyVaultId)
-	}
-
-	return *resp.Properties.VaultURI, nil
+	return groups[1], nil
 }
 
-func GetKeyVaultIDFromBaseUrl(ctx context.Context, client *keyvault.VaultsClient, keyVaultUrl string) (*string, error) {
-	list, err := client.ListComplete(ctx, utils.Int32(1000))
+func GetKeyVaultIDFromBaseUrl(ctx context.Context, client *resources.Client, keyVaultUrl string) (*string, error) {
+	name, err := GetKeyVaultNameFromBaseUrl(keyVaultUrl)
 	if err != nil {
-		return nil, fmt.Errorf("Error GetKeyVaultId unable to list Key Vaults %v", err)
+		return nil, err
 	}
 
-	for list.NotDone() {
-		v := list.Value()
-
-		if v.ID == nil {
-			return nil, fmt.Errorf("v.ID was nil")
-		}
-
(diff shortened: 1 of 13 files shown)


tombuildsstuff:
hi @njuCZ 

Thanks for opening this PR.

Taking a look at #6196 - there's two reasons behind this issue:

1) Terraform is returning a "Context Deadline Exceeded" since the default timeouts are being hit..

2) .. because Azure's returning a HTTP 429 because of too many requests being sent (and so we're being forced to wait and retry) which in turn exceeds the default timeout

Prior to version 2.0 of the Azure Provider all resources shared the same timeout (3 hours) - which whilst convenient in some cases caused issues in others where some Azure Services take a long time to provision (e.g. App Service Environment).

Version 2.0 of the Azure Provider added support for Custom Timeouts on all resources - as such we've gone through all the Data Sources and Resources available in the Azure Provider and added reasonable-ish defaults for these, based on how long it takes to provision/destroy these resources (and we're updating this over time).

Whilst this approach may help a little in the short-term - in practice since this information isn't being cached it likely won't alleviate that too much. Instead we'd be better to take an approach similar to [how Storage Accounts caching works](https://github.com/terraform-providers/terraform-provider-azurerm/blob/49b206606f775e5dbf07f6d6f1e8d7e1ecee8534/azurerm/internal/services/storage/client/helpers.go) by looking this information up and caching it - which would probably give us a small performance boost.

In Terraform we recommend using multiple smaller state files rather than one large state file - primarily to reduce the "blast radius" when something goes wrong - but also to avoid situations like this. Whilst one option would be to work through and add Custom Timeouts to all of the resources (as has been partially tried in #6196) - it's preferable to split this Terraform Configuration up so that it's not doing so much, which would solve this issue.

Whilst this approach may help to alleviate some of the pain in the short-term - even if we cached this we'd end up with the same issue longer-term albeit with a slightly higher threshold. As mentioned above, it should instead be possible to work around this by splitting this configuration into multiple State files - which is our recommendation here - and as such whilst I'd like to thank you for this contribution I'm going to close this PR for the moment.

Thanks!

njuCZ:
@tombuildsstuff thanks for your opinion, that makes sense to me. So what about this issue #6196 ? should we tell the author to split the configuration file and close it ?

PR #6867, opened 2020-05-12
New data source: `azurerm_advisor_recommendations`

yupwei68:
fix: https://github.com/terraform-providers/terraform-provider-azurerm/issues/6119
image


tombuildsstuff:
hi @yupwei68 

Thanks for opening this PR.

Taking a look through here, whilst the code for this looks good - thinking about use-cases for this functionality, it appears this is something that a user would opt to do interactively (using the Azure CLI) or automatically (via the HTTP API's) - rather than something that's useful in Terraform (since there's no means of a user performing actions based on these recommendations).

As such whilst I'd like to thank you for this contribution - since it appears this functionality isn't suited to Terraform I'm going to close this pull request for the moment - but we'd recommend instead using the Azure CLI to run interactive commands like this.

Thanks!

website/docs/d/advisor_recommendations.html.markdown
```diff
@@ -0,0 +1,72 @@
...
+output "recommendations" {
+  value = data.azurerm_advisor_recommendations.example.recommendations
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `categories_filter` - (Optional) Specifies a list of categories in which the Advisor Recommendations will be listed. Possible values are 'HighAvailability', 'Security', 'Performance', 'Cost' and 'OperationalExcellence'.
```
katbyte, line 34: We'll need to quote these correctly
```suggestion
* `categories_filter` - (Optional) Specifies a list of categories in which the Advisor Recommendations will be listed. Possible values are `HighAvailability`, `Security`, `Performance`, `Cost` and `OperationalExcellence`.
```

azurerm/internal/services/advisor/data_source_advisor_recommendations.go
```diff
@@ -0,0 +1,212 @@
...
+func dataSourceArmAdvisorRecommendations() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceArmAdvisorRecommendationsRead,
+
+		Timeouts: &schema.ResourceTimeout{
+			Read: schema.DefaultTimeout(10 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"categories_filter": {
```
katbyte, line 26: Could we change this to
```suggestion
			"filter_by_category": {
```

azurerm/internal/services/advisor/data_source_advisor_recommendations.go
```diff
@@ -0,0 +1,212 @@
...
+					ValidateFunc: validation.StringInSlice([]string{
+						string(advisor.HighAvailability),
+						string(advisor.Security),
+						string(advisor.Performance),
+						string(advisor.Cost),
+						string(advisor.OperationalExcellence)}, true),
+				},
+			},
+
+			"resource_group_names_filter": azure.SchemaResourceGroupNameSetOptional(),
```
katbyte, line 40: and
```suggestion
			"filter_by_resource_groups": azure.SchemaResourceGroupNameSetOptional(),
```

katbyte: I'm also not sure its worth creating a schema function for this given its not duplicated anywhere else/

yupwei68: Here I add a schema function because during the reviewing of this PR, the schema function in `resource_group.go` continues to be added. I need to change all the `ValidateFunc: validateResourceGroupName` to `ValidateFunc: ValidateResourceGroupName`

yupwei68: Do you think I shall remove the schema function?

yupwei68:
@katbyte , Please continue reviewing.

PR #6897, opened 2020-05-13
size property must not be defined if source_uri is used

floushee:
I added additional information to the size property documentation.

### DIFF
--- a/website/docs/r/storage_blob.html.markdown
+++ b/website/docs/r/storage_blob.html.markdown
@@ -54,7 +54,9 @@ The following arguments are supported:
 
 * `type` - (Required) The type of the storage blob to be created. Possible values are `Append`, `Block` or `Page`. Changing this forces a new resource to be created.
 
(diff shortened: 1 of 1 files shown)


website/docs/r/storage_blob.html.markdown
```diff
@@ -54,7 +54,7 @@ The following arguments are supported:
 
 * `type` - (Required) The type of the storage blob to be created. Possible values are `Append`, `Block` or `Page`. Changing this forces a new resource to be created.
 
-* `size` - (Optional) Used only for `page` blobs to specify the size in bytes of the blob to be created. Must be a multiple of 512. Defaults to 0.
+* `size` - (Optional) Used only for `page` blobs to specify the size in bytes of the blob to be created. Must be a multiple of 512. Defaults to 0. Not needed if `source_uri` is defined.
```
katbyte, line 57: I think this might read better if we invert the logic:
```suggestion
* `size` - (Optional) Used only for `page` blobs to specify the size in bytes of the blob to be created. Must be a multiple of 512. Defaults to 0. 

~> **Note:** `size` is required if `source_uri` is not set.
```

PR #5404, opened 2020-01-15
CosmosDB Account: Support zone redundancy for `geo_location`

evenh:

PR #6899, opened 2020-05-13
add zone_redundant flag to azure_cosmosdb_account geo_location

rubbish:
Need to figure out how to write a test for this, I'm not super familiar with the tests in this provider.


katbyte:
@rubbish - there is an existing pr #5404, its also lacking tests and needs some work thou.

katbyte:
Hey @rubbish,

Thanks for the PR but as the older #5404 is more complete and there and been no further commits here i'm going to close this in favour of that one.

PR #6916, opened 2020-05-14
`azurerm_frontdoor` - Expose the `header_frontdoor_id` attribute

### LINKED ISSUE #6741: Retrive frontdoorID from Graph
_This issue was originally opened by @tikicoder as hashicorp/terraform#24832. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._
    
<hr>
    
### Current Terraform Version
```
Terraform Version
Terraform v0.12.24
provider.azurev2.5.0
...
```

### Use-cases
Azure has updated the Front Door Service to have a frontdoorID.  Can TF be updated to pull this information?  This ID is what you can use to lock down services to only accept connections from Front Door.
https://docs.microsoft.com/en-us/azure/frontdoor/front-door-faq

### Attempted Solutions
I manually use postman to get the information.

### Proposal
update TF to use the newer graph API and return the value

WodansSon:
Fixes #6741

### DIFF
--- a/azurerm/internal/services/frontdoor/frontdoor_resource.go
+++ b/azurerm/internal/services/frontdoor/frontdoor_resource.go
@@ -51,7 +51,7 @@ func resourceArmFrontDoor() *schema.Resource {
 				Computed: true,
 			},
 
-			"frontdoor_id": {
+			"header_frontdoor_id": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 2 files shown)


WodansSon:
image

PR #6927, opened 2020-05-14
feature: optionally enhanced location validation

tombuildsstuff:
This commit introduces a new optional feature, for the enhanced validation of Locations.

The Azure MetaData Service - that is: /metadata/endpoints?api-version=2018-01-01 returns information about the locations which are suppported on the Azure Instance that we're connected to. As such, this commit optionally caches this information with the intent of providing more granular validation - to avoid cases where an unsupported location is specified.

This allows Terraform to catch this error during `terraform plan` - rather than failing during `terraform apply` - which is a better user experience.

This functionality is disabled by default at this time - but can conditionally be enabled via the Feature Flag using the Environment Variable `ARM_PROVIDER_ENHANCED_VALIDATION` to `true`.

Example before/with this feature disabled:

```
 $ ARM_PROVIDER_ENHANCED_VALIDATION=false tf plan

  + create

  # azurerm_resource_group.test will be created
  + resource "azurerm_resource_group" "test" {
      + id       = (known after apply)
      + location = "chinanorth"
      + name     = "tom-dev99"

      + timeouts {
          + create = "60m"
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy
```

Example with this feature enabled:

```
$ ARM_PROVIDER_ENHANCED_VALIDATION=true tf plan

Error: "chinanorth" was not found in the list of supported Azure Locations: "westus,westus2,eastus,centralus,centraluseuap,southcentralus,northcentralus,westcentralus,eastus2,eastus2euap,brazilsouth,brazilus,northeurope,westeurope,eastasia,southeastasia,japanwest,japaneast,koreacentral,koreasouth,indiasouth,indiawest,indiacentral,australiaeast,australiasoutheast,canadacentral,canadaeast,uknorth,uksouth2,uksouth,ukwest,francecentral,francesouth,australiacentral,australiacentral2,uaecentral,uaenorth,southafricanorth,southafricawest,switzerlandnorth,switzerlandwest,germanynorth,germanywestcentral,norwayeast,norwaywest"

  on main.tf line 5, in resource "azurerm_resource_group" "test":
   5: resource "azurerm_resource_group" "test" {
```

### DIFF
--- a/azurerm/internal/sdk/locations.go
+++ b/azurerm/internal/sdk/locations.go
@@ -1,0 +1,60 @@
+package sdk
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+type SupportedLocations struct {
+	// Locations is a list of Locations which are supported on this Azure Endpoint.
+	// This could be nil when the user is offline, or the Azure MetaData Service does not have this
+	// information and as such this should be used as best-effort, rather than guaranteed
+	Locations *[]string
+}
+
+type cloudEndpoint struct {
+	Endpoint  string    `json:"endpoint"`
+	Locations *[]string `json:"locations"`
+}
+
+type metaDataResponse struct {
+	CloudEndpoint map[string]cloudEndpoint `json:"cloudEndpoint"`
+}
+
+// AvailableAzureLocations returns a list of the Azure Locations which are available on the specified endpoint
+func AvailableAzureLocations(ctx context.Context, endpoint string) (*SupportedLocations, error) {
+	uri := fmt.Sprintf("https://%s//metadata/endpoints?api-version=2018-01-01", endpoint)
+	client := http.Client{
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+		},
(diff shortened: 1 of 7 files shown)


azurerm/internal/sdk/locations.go
```diff
@@ -0,0 +1,60 @@
...
+	client := http.Client{
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+		},
+	}
+	req, err := http.NewRequestWithContext(ctx, "GET", uri, nil)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := client.Do(req)
```
tombuildsstuff, line 39: note: this intentionally doesn't use Azure/go-autorest since we don't need it here (it's an unauthorized endpoint, and logging this isn't super valuable)

katbyte:
THou i wonder if we should add this to the docs?

tombuildsstuff:
> THou i wonder if we should add this property to the docs?

I'll add a section to the changelog for now, if we turn this on by default then it's probably worth documenting more broadly

PR #6940, opened 2020-05-15
`azurerm_shared_image_version` - `target_region.x.storage_account_type` is now defaulted and multiple `target_region`s can be added/removed

### LINKED ISSUE #5569: Adding target_region block to azurerm_shared_image_version causes replacement
### Community Note

Terraform v0.12.20
+ provider.azurerm v1.42.0

### Affected Resource(s)

* `azurerm_shared_image_version`

Initial configuration:

```hcl
resource "azurerm_shared_image_version" "bamboo-agent-windows-1-0-0" {
  name                = "1.0.0"
  gallery_name        = azurerm_shared_image_gallery.central_image_gallery.name
  image_name          = azurerm_shared_image.bamboo-agent-windows.name
  resource_group_name = azurerm_resource_group.resource-group.name
  location            = "northcentralus"
  managed_image_id    = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/mygroup/providers/Microsoft.Compute/images/build-agent-image"

  target_region {
    name                   = "northcentralus"
    regional_replica_count = "1"
    storage_account_type   = "Standard_LRS"
  }

  tags = {
    environment = azurerm_resource_group.resource-group.tags.environment
  }
}
```
Added new target_region block after applying previous configuration:

```hcl
resource "azurerm_shared_image_version" "bamboo-agent-windows-1-0-0" {
  name                = "1.0.0"
  gallery_name        = azurerm_shared_image_gallery.central_image_gallery.name
  image_name          = azurerm_shared_image.bamboo-agent-windows.name
  resource_group_name = azurerm_resource_group.resource-group.name
  location            = "northcentralus"
  managed_image_id    = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/mygroup/providers/Microsoft.Compute/images/build-agent-image"

  target_region {
    name                   = "northcentralus"
    regional_replica_count = "1"
    storage_account_type   = "Standard_LRS"
  }

  target_region {
    name                   = "southcentralus"
    regional_replica_count = "1"
    storage_account_type   = "Standard_LRS"
  }

  tags = {
    environment = azurerm_resource_group.resource-group.tags.environment
  }
}
```
### Expected Behavior

azurerm_shared_image_version resource is updated in-place with new target_region.

### Actual Behavior

azurerm_shared_image_version change forces replacement

### Steps to Reproduce
1. Create an image version with one target_region block
2. Run apply
3. Modify configuration files to second target_region block
4. Run plan

neil-yechenwei:
fixes terraform-providers/terraform-provider-azurerm#5569

Existing test case already covers this scenario. So no more test case needs to be added.


mbfrahry (changes requested):
Hey @neil-yechenwei, this fixes part of your linked issue but we are still missing a piece. We'll want to remove the `ForceNew: true` and add a `CustomizeDiff` that checks to see if the old/new key being passed in aren't empty and aren't the same. If they aren't the same then we have to `ForceNew` the resource.

neil-yechenwei:
@mbfrahry , thanks for your comments. Seems I still don't understand why I need to remove "ForceNew: true" and add CustomizeDiff since "ValidateFunc" can check if the input is empty and difference between old value and new value. Could you provide some details?

neil-yechenwei:
@mbfrahry , seems "ForceNew: true" is doing the things what you are describing. So May I know why I have to remove "ForceNew: true" and add "CustomizeDiff" to do the same thing?

mbfrahry:
Hey @neil-yechenwei, no problem with the extra explanation. The reason I call out using a separate `CustomizeDiff` is because the current iteration will always force a new resource to be created whenever a new `target_region` is added. This is due to the `ForceNew: true` tag being associated with `storage_account_type`. In practice, we should be able to add and remove  a `target_region` as we please but this isn't the case with the current version. By adding a `CustomizeDiff` we can have a finer granularity on when we `ForceNew` the resource. We can check the old key and the new key to determine what we do with the resource. If either is empty, we know that we're removing or adding a `target_region` and can proceed without recreating the resource. If both are not empty, we now need to recreate the resource since that is not allowed by the api. The best approach may actually be removing the `target_region` when the previous use case occurs and adding a new `target_region` with the new `storage_account_type`.

neil-yechenwei:
@mbfrahry , I tried to use `CustomizeDiff`. Seems "old value" and "new value" are same while ran "tf apply" at second time so that I cannot compare the values. Maybe I missed something. Could you give me some suggestion? Thanks.
```
CustomizeDiff: customdiff.Sequence(
			customdiff.ForceNewIfChange("target_region", func(old, new, meta interface{}) bool {
				oldTargetRegions := old.(*schema.Set).List()
				newTargetRegions := new.(*schema.Set).List()

				// As `target_region` is required property, so `new` cannot be set as empty. So we don't need to check here.
				if len(oldTargetRegions) == 0 {
					return true
				}

				for _, oldTargetRegion := range oldTargetRegions {
					oldTargetRegionBlock := oldTargetRegion.(map[string]interface{})
					for _, newTargetRegion := range newTargetRegions {
						newTargetRegionBlock := newTargetRegion.(map[string]interface{})
						if azure.NormalizeLocation(oldTargetRegionBlock["name"]) == azure.NormalizeLocation(newTargetRegionBlock["name"]) {
							if oldTargetRegionBlock["storage_account_type"] != newTargetRegionBlock["storage_account_type"] {
								return true
							}
						}
					}
				}

				return true
			}),
		),
```

mbfrahry:
Ok, I've just spent quite a bit of time playing with this and you're right that you won't get the correct behavior with `CustomizeDiff`. 

I will note that your implementation will always ForceNew the resource if the number of `target_region`s change. We'd want to `return false` at the end of the function to show that we don't need to ForceNew the resource if things haven't changed. 

CustomizeDiff won't work in a Set where a specific attribute needs to be `ForceNew` because we can't get to the key of the Set where the changed `storage_account_type` lives. The implementation you provided only triggers when we add or remove a `target_region` but we only care about when a `storage_account_type` changes. 

So we have two options: We can just remove the `ForceNew` entirely from `storage_account_type` and rely on the api error that you can't update that field or we can do some complicated logic in the provider to detect when `storage_account_type` is changing and update the resource by removing the affected `target_region` and then re-adding it with the new `storage_account_type`

neil-yechenwei:
@mbfrahry , thanks for your suggestions. I think using CustomizeDiff in a Set should be a common requirement. So I prefer your first option since I assume it should be supported/implemented in the future.

mbfrahry (approved):
LGTM!

PR #7012, opened 2020-05-19
New Resource: `azurerm_time_series_insights_environment` 

### LINKED ISSUE #3844: Support for Microsoft.TimeSeriesInsights
### Community Note

### Description

[Time Series Insights](https://docs.microsoft.com/en-gb/azure/time-series-insights/) is an Azure Resource that cannot be created via Terraform. It could be a supported resource. ~~However, it is not well supported by Microsoft at this time, as it cannot be created via an ARM template~~.

* azurerm_time_series_insights

### References

* TSI Docs: https://docs.microsoft.com/en-gb/azure/time-series-insights/

mbfrahry:
```
--- PASS: TestAccAzureRMTimeSeriesInsightsEnvironment_basic (139.73s)
--- PASS: TestAccAzureRMTimeSeriesInsightsEnvironment_complete (301.78s)
```

Fixes #3844


azurerm/internal/services/timeseriesinsights/parse/time_series_insights_environment.go
```diff
@@ -0,0 +1,33 @@
...
+
+type TimeSeriesInsightsEnvironmentId struct {
+	ResourceGroup string
+	Name          string
+}
+
+func TimeSeriesInsightsEnvironmentID(input string) (*TimeSeriesInsightsEnvironmentId, error) {
+	id, err := azure.ParseAzureResourceID(input)
+	if err != nil {
+		return nil, fmt.Errorf("[ERROR] Unable to parse Time Series Insights Environment ID %q: %+v", input, err)
```
ArcturusZhang, line 17: Could we please modify this error message  a bit to make it to look less like a log rather than an error message?
```suggestion
		return nil, fmt.Errorf("parsing Time Series Insights Environment ID %q: %+v", input, err)
```

azurerm/internal/services/timeseriesinsights/resource_time_series_insights_environment.go
```diff
@@ -0,0 +1,293 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringMatch(
+					regexp.MustCompile(`^[-\w\._\(\)]+$`),
+					"Time Series Insights Environment name must be 1 - 90 characters long, contain only word characters and underscores.",
```
ArcturusZhang, line 50: This regex does not really match the error message below. For instance this regex does not have a constraint of length, and the regex also allows braces. Which one is expected? We should at least modify one of these.

azurerm/internal/services/timeseriesinsights/resource_time_series_insights_environment.go
```diff
@@ -0,0 +1,293 @@
...
+func resourceArmTimeSeriesInsightsEnvironment() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceArmTimeSeriesInsightsEnvironmentCreateUpdate,
+		Read:   resourceArmTimeSeriesInsightsEnvironmentRead,
+		Update: resourceArmTimeSeriesInsightsEnvironmentCreateUpdate,
+		Delete: resourceArmTimeSeriesInsightsEnvironmentDelete,
+		Importer: azSchema.ValidateResourceIDPriorToImport(func(id string) error {
+			_, err := parse.TimeSeriesInsightsEnvironmentID(id)
+			return err
+		}),
```
ArcturusZhang, line 34: Based on the content below, this resource is a polymorphic one. Since different types of this resource share the same ID, could we add a bit more validation here to ensure the user is importing the ID to the right resource?

azurerm/internal/services/timeseriesinsights/resource_time_series_insights_environment.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	resource, ok := resp.Value.AsStandardEnvironmentResource()
+	if !ok {
+		return fmt.Errorf("resource was not a standard Time Series Insights Environment %q (Resource Group %q)", name, resourceGroup)
+	}
+
+	if resource.ID == nil {
+		return fmt.Errorf("cannot read Time Series Insights Environment %q (Resource Group %q) ID", name, resourceGroup)
+	}
```
ArcturusZhang, line 187: Could we add a check to ensure the ID is not nil and not empty?
```suggestion
	if resource.ID == nil || *resource.ID == "" {
		return fmt.Errorf("cannot read Time Series Insights Environment %q (Resource Group %q) ID", name, resourceGroup)
	}
```

azurerm/internal/services/timeseriesinsights/resource_time_series_insights_environment.go
```diff
@@ -0,0 +1,293 @@
...
+
+	id, err := parse.TimeSeriesInsightsEnvironmentID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	response, err := client.Delete(ctx, id.ResourceGroup, id.Name)
+	if err != nil {
+		if !utils.ResponseWasNotFound(response) {
+			return fmt.Errorf("deleting Time Series Insights Environment %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
```
ArcturusZhang, line 253: I suppose if the service returns 404 when you are deleting the resource, it is still something wrong happening. Typically if you are attempting to delete a non-existing resource, the service should return 204

azurerm/internal/services/timeseriesinsights/tests/resource_time_series_insights_environment_test.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+		resp, err := client.Get(ctx, id.ResourceGroup, id.Name, "")
+
+		if err != nil {
+			return nil
+		}
+
+		if resp.StatusCode != http.StatusNotFound {
+			return fmt.Errorf("Time Series Insights Environment still exists: %q", id.Name)
+		}
```
ArcturusZhang, line 118: This reads really weird. If the resource does not exist, the error would be non-nil, and this check would return `nil` immediately. But if something is wrong (for instance 429 returned), the error would be also non-nil. In this case, this case would regard this as "the resource does not exist" and return nil.

On the other hand, the `if resp.StatusCode != http.StatusNotFound` would never succeed. `Get` function would only return nil error when it gets the resource successfully (in other word `200`), in this case the status code would never be 404.
This behaviour is also defined in the swagger [here](https://github.com/Azure/azure-rest-api-specs/blob/b069b0fa4817788d6363007645cae04317864f09/specification/timeseriesinsights/resource-manager/Microsoft.TimeSeriesInsights/preview/2018-08-15-preview/timeseriesinsights.json#L162)

ArcturusZhang: And to be honest, a lot of the existing tests have this kind of check destroy functions, which I suppose it is not so correct. What do you think?

azurerm/internal/services/timeseriesinsights/tests/resource_time_series_insights_environment_test.go
```diff
@@ -0,0 +1,167 @@
...
+				),
+			},
+			data.ImportStep(),
+			{
+				Config: testAccAzureRMTimeSeriesInsightsEnvironment_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMTimeSeriesInsightsEnvironmentExists(data.ResourceName),
+				),
+			},
+			data.ImportStep(),
```
ArcturusZhang, line 63: Does this update test really help? Because in the schema, the `tags` are defined as `ForceNew`, and the config for basic does not have a tag, complete config has a tag. Therefore this update would always be a replacement rather than a in-place update.

website/docs/r/time_series_insights_environment.html.markdown
```diff
@@ -0,0 +1,70 @@
...
+
+* `sku_name` - (Required) Specifies the SKU Name for this Time Series Insights Environment. It is string consisting of two parts separated by an underscore(\_).The fist part is the `name`, valid values include: `S1` and `S2`. The second part is the `capacity` (e.g. the number of deployed units of the `sku`), which must be a positive `integer` (e.g. `S1_1`).
+
+* `data_retention_time` - (Required) Specifies the ISO8601 timespan specifying the minimum number of days the environment's events will be available for query.
+
+* `storage_limited_exceeded_behavior` - (Optional) Specifies the behavior the Time Series Insights service should take when the environment's capacity has been exceeded. Valid values include `PauseIngress` and `PurgeOldData`.
+
+* `partition_key` - (Optional) The name of the event property which will be used to partition data.
+
+* `tags` - (Optional) A mapping of tags to assign to the resource.
```
ArcturusZhang, line 47: Is the `tags` ForceNew as defined in the schema? If it is, we should change this description accordingly.

azurerm/internal/services/timeseriesinsights/parse/time_series_insights_environment.go
mbfrahry: Good catch! Done!

website/docs/r/time_series_insights_environment.html.markdown
mbfrahry: My fault here! Tags now updates without recreating the resource

PR #7040, opened 2020-05-21
Update storage account syntax for Azure Function App

SeanKilleen:
When working with the example azure functions app, I noticed errors around the storage account syntax. After a bit of trial and error, I realized that it appears TF now uses `storage_connection_string` rather than the name / access key combination, causing errors when using the sample.

This PR updates the docs and sample to reflect that change.


SeanKilleen:
@katbyte this may have been an issue of timing. 

Now when I'm running something with 2.11, I see:

> Warning: "storage_connection_string": [DEPRECATED] Deprecated in favor of `storage_account_name` and `storage_account_access_key`

So it looks like this change may not have been correct. 

I'm going to open a new PR to build on it and mention the difference.

PR #7045, opened 2020-05-21
Azure monitor support for azurerm_hdinsight_*_cluster

kosinsky:
Adds support for Azure Monitor for HDInsight clusters:

* azurerm_hdinsight_hadoop_cluster
* azurerm_hdinsight_hbase_cluster
* azurerm_hdinsight_interactive_query_cluster
* azurerm_hdinsight_kafla_cluster
* azurerm_hdinsight_spark_cluster
* azurerm_hdinsight_storm_cluster

```hcl
resource "azurerm_hdinsight_hadoop_cluster" "example" {
  name                = "example-hdicluster"
  location            = "${azurerm_resource_group.example.location}"
  cluster_version     = "3.6"
  tier                = "Standard"

  monitor {
    log_analytics_workspace_id = data.azurerm_log_analytics_workspace.example.workspace_id
    primary_key                = data.azurerm_log_analytics_workspace.example.primary_shared_key
  }

}
```

 Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/6780


ross-p-smith:
Hi @tombuildsstuff - is this likely to get in the next release?

ross-p-smith:
@jackofallops - any chance we can get this one into 2.17.0 with the other HDInisght (external metastore) as it means I can have one testing cycle in my project? We have a lengthy approval process to take new provider versions. Would help me greatly. Ross

kosinsky:
> During the review below it became apparent that the monitor configuration doesn't fit the usual approach to the cluster API interaction in this service. Almost everything about this service requires a cluster rebuild to change, but the monitor config (as I understand it) it entirely configurable out of band of this? Each resource uses exactly the same code irrespective of cluster type, so this feels like it should actually be a separate resource that can be used against any cluster and managed independently of the cluster itself? WDYT?

I'm thinking amount of code that is the same for all 8 resources for a while now. For example, most of the code is the same for accounts, metastores, gateway etc. Especially, it's possible that soon new resource(s) will be added as HDInsight adds cluster(s) based on Apache distribution of Hadoop instead of HDP. https://azure.microsoft.com/en-au/blog/azure-hdinsight-and-azure-database-for-postgresql-news/

For HDInsight to Azure monitor mapping it may make sense to use separate resource and the provider doing it for [azurerm_monitor_diagnostic_setting](https://www.terraform.io/docs/providers/azurerm/r/monitor_diagnostic_setting.html) in generic way and it works with HDI too. On other hand, some resources have additional Azure Monitor/Log Analytics option. I was looking on [AKS](https://www.terraform.io/docs/providers/azurerm/d/kubernetes_cluster.html) and [container group](https://www.terraform.io/docs/providers/azurerm/r/container_group.html) resources which work with Monitor as a part of their definition.

Also, Monitor isn't only one option that can be updated after cluster is ready. For example, now hdinsight resources support resizing and API supports configuring autoscaling without recreating cluster. It was planning to work on that after WIP is done. https://github.com/terraform-providers/terraform-provider-azurerm/issues/6470. Code for autoscaling may look pretty close for the monitor.

I don't have strong preference for the property vs. new resource.  My light preference is to keep monitor as a property, establish a pattern for updating a cluster, and be more aggressive in refactoring of the  hdinsight_* resources to minimize code duplication.

Please, let me know WDYT. Meanwhile, I'll work on other feedback.

azurerm/helpers/azure/hdinsight.go
```diff
@@ -156,6 +156,34 @@ func SchemaHDInsightsExternalMetastore() *schema.Schema {
...
+	return &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"log_analytics_workspace_id": {
+					Type:     schema.TypeString,
+					Required: true,
+					ForceNew: false,
```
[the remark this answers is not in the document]
kosinsky: Not sure sure I understood. It configured as `ForceNew: false,`. In *_updateMonitor tests I'm testing changing workspace_id and doesn't recreate cluster

jackofallops:
> I don't have strong preference for the property vs. new resource. My light preference is to keep monitor as a property, establish a pattern for updating a cluster, and be more aggressive in refactoring of the hdinsight_* resources to minimize code duplication.

I think this will work either way, I tend to lean towards separation where possible for clarity, but that doesn't necessarily make it the _right_ way to go.  I'll review again in light of this.

Also, a quick note on refactoring these resources. We need to be mindful of state change, and of the size of change per PR so any significant rework will likely need to be rolled through one cluster type at a time. 

I may have said it before, but thanks for your continued effort on this service, it's very much appreciated 😄

azurerm/helpers/azure/hdinsight.go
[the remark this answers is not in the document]
jackofallops: Sorry - Just saw `ForceNew`, not the `false`.  We only specify this property when setting it to `true` as it defaults to `false`, so it can be removed here.

azurerm/helpers/azure/hdinsight.go
```diff
@@ -164,6 +164,34 @@ func SchemaHDInsightsExternalMetastore() *schema.Schema {
...
+				},
+				"primary_key": {
+					Type:         schema.TypeString,
+					Required:     true,
+					ForceNew:     false,
```
jackofallops, line 182: Same here, `ForceNew` should be omitted if not being set to `true`.

azurerm/internal/services/hdinsight/common_hdinsight.go
```diff
@@ -106,6 +106,18 @@ func hdinsightClusterUpdate(clusterKind string, readFunc schema.ReadFunc) schema
 			}
 		}
 
+		if d.HasChange("monitor") {
+			log.Printf("[DEBUG] Change Azure Monitor for the HDInsight %q Cluster", clusterKind)
+			extensionsClient := meta.(*clients.Client).HDInsight.ExtensionsClient
```
jackofallops, line 111: Can we bump this client to the top of the func with the rest?

azurerm/internal/services/hdinsight/hdinsight_hbase_cluster_resource.go
```diff
@@ -282,6 +294,13 @@ func resourceArmHDInsightHBaseClusterRead(d *schema.ResourceData, meta interface
 		d.Set("https_endpoint", httpEndpoint)
 		sshEndpoint := azure.FindHDInsightConnectivityEndpoint("SSH", props.ConnectivityEndpoints)
 		d.Set("ssh_endpoint", sshEndpoint)
+
+		monitor, err := extensionsClient.GetMonitoringStatus(ctx, resourceGroup, name)
+		if err != nil {
+			return fmt.Errorf("Error reading monitor configuation for HDInsight Hadoop Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
```
jackofallops, line 300: Cosmetic change, we're removing the duplication of the word "Error" from error messages since these are prefixed with `Error:` anyway.
```suggestion
			return fmt.Errorf("failed reading monitor configuration for HDInsight Hadoop Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
```

azurerm/internal/services/hdinsight/hdinsight_interactive_query_cluster_resource.go
```diff
@@ -282,6 +294,13 @@ func resourceArmHDInsightInteractiveQueryClusterRead(d *schema.ResourceData, met
 		d.Set("https_endpoint", httpEndpoint)
 		sshEndpoint := azure.FindHDInsightConnectivityEndpoint("SSH", props.ConnectivityEndpoints)
 		d.Set("ssh_endpoint", sshEndpoint)
+
+		monitor, err := extensionsClient.GetMonitoringStatus(ctx, resourceGroup, name)
+		if err != nil {
+			return fmt.Errorf("Error reading monitor configuation for HDInsight Hadoop Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
```
jackofallops, line 300: As above:
```suggestion
			return fmt.Errorf("failed reading monitor configuration for HDInsight Hadoop Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
```

azurerm/internal/services/hdinsight/hdinsight_kafka_cluster_resource.go
```diff
@@ -283,6 +295,13 @@ func resourceArmHDInsightKafkaClusterRead(d *schema.ResourceData, meta interface
 		d.Set("https_endpoint", httpEndpoint)
 		sshEndpoint := azure.FindHDInsightConnectivityEndpoint("SSH", props.ConnectivityEndpoints)
 		d.Set("ssh_endpoint", sshEndpoint)
+
+		monitor, err := extensionsClient.GetMonitoringStatus(ctx, resourceGroup, name)
+		if err != nil {
+			return fmt.Errorf("Error reading monitor configuation for HDInsight Hadoop Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
```
jackofallops, line 301:
```suggestion
			return fmt.Errorf("failed reading monitor configuration for HDInsight Hadoop Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
```

azurerm/internal/services/hdinsight/hdinsight_spark_cluster_resource.go
```diff
@@ -282,6 +294,13 @@ func resourceArmHDInsightSparkClusterRead(d *schema.ResourceData, meta interface
 		d.Set("https_endpoint", httpEndpoint)
 		sshEndpoint := azure.FindHDInsightConnectivityEndpoint("SSH", props.ConnectivityEndpoints)
 		d.Set("ssh_endpoint", sshEndpoint)
+
+		monitor, err := extensionsClient.GetMonitoringStatus(ctx, resourceGroup, name)
+		if err != nil {
+			return fmt.Errorf("Error reading monitor configuation for HDInsight Hadoop Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
```
jackofallops, line 300:
```suggestion
			return fmt.Errorf("failed reading monitor configuration for HDInsight Hadoop Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
```

azurerm/internal/services/hdinsight/hdinsight_storm_cluster_resource.go
```diff
@@ -279,6 +291,13 @@ func resourceArmHDInsightStormClusterRead(d *schema.ResourceData, meta interface
 		d.Set("https_endpoint", httpEndpoint)
 		sshEndpoint := azure.FindHDInsightConnectivityEndpoint("SSH", props.ConnectivityEndpoints)
 		d.Set("ssh_endpoint", sshEndpoint)
+
+		monitor, err := extensionsClient.GetMonitoringStatus(ctx, resourceGroup, name)
+		if err != nil {
+			return fmt.Errorf("Error reading monitor configuation for HDInsight Hadoop Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
```
jackofallops, line 297:
```suggestion
			return fmt.Errorf("failed reading monitor configuration for HDInsight Hadoop Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
```

jackofallops:
working through the conflicts after merging 6969...

kosinsky:
> working through the conflicts after merging 6969...
Ooh, I started to work on merge too. @jackofallops, I do the merge if it helps.

jackofallops:
> > working through the conflicts after merging 6969...
> 
> Ooh, I started to work on merge too. @jackofallops, I do the merge if it helps.

I just finished, made a start as I was expecting the history to have been broken due to the fixes we had to do Monday. Turns out, not a problem for this one!  

I'll let travis do it's thing and then trigger the tests on CI.

Thanks @kosinsky 😃

azurerm/helpers/azure/hdinsight.go
```diff
@@ -156,6 +156,34 @@ func SchemaHDInsightsExternalMetastore() *schema.Schema {
...
+		Type:     schema.TypeList,
+		Optional: true,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"log_analytics_workspace_id": {
+					Type:     schema.TypeString,
+					Required: true,
+					ForceNew: false,
+				},
```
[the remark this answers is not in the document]
kosinsky: HDInisght needs `workspace_id` (https://www.terraform.io/docs/providers/azurerm/r/log_analytics_workspace.html#workspace_id) property that just guid instead of `id` that is proper Azure Resource ID that can be validated with `LogAnalyticsWorkspaceID` func

jackofallops:
Test passing (transient failures aside...)
image

Issue #6795, opened 2020-05-06
issues with subscription key and import in API-M since 2.7.0


tnicolaysen:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.12.24
terraform-provider-azurerm 2.8.0 (most likely 2.7.0 also)

### Affected Resource(s)

* `azurerm_api_management_api`
* `azurerm_api_management_subscription`

It's also likely that other API-M resources are affected.

### Terraform Configuration Files

```hcl
provider "azurerm" {
  version = "= 2.8.0"
  skip_provider_registration = true
}

terraform {
  backend "azurerm" {}
  required_version = ">= 0.12.26"
}

data "azurerm_api_management" "apim" {
  name                = var.apim_name
  resource_group_name = var.apim_resource_group
}


resource "azurerm_api_management_user" "example" {
  api_management_name = data.azurerm_api_management.apim.name
  resource_group_name = data.azurerm_api_management.apim.resource_group_name
  email               = "demo@example.org"
  first_name          = "John"
  last_name           = "Doe"
  user_id             = "john-doe"
}


resource "azurerm_api_management_product" "example" {
  api_management_name = data.azurerm_api_management.apim.name
  resource_group_name = data.azurerm_api_management.apim.resource_group_name

  product_id            = "example"
  display_name          = "Example Product"
  published             = true
  subscription_required = true
}


resource "azurerm_api_management_subscription" "example" {
  api_management_name = data.azurerm_api_management.apim.name
  resource_group_name = data.azurerm_api_management.apim.resource_group_name

  display_name = "Example Subscription"
  product_id   = azurerm_api_management_product.example.id
  user_id      = azurerm_api_management_user.example.id
  state        = "active"

  // NOTICE: Primary/Secondary subscription keys not set, supposed to be created automatically
}


resource "azurerm_key_vault_secret" "example_pk" {
  key_vault_id = var.keyvault_id
  name         = "SubscriptionPrimaryKey"
  value        = azurerm_api_management_subscription.example.primary_key

[...]

PR #7076, opened 2020-05-26
azurerm_api_management_api` `wsdl_selector` is required when content_format = "wsdl"

yupwei68:
Fix https://github.com/terraform-providers/terraform-provider-azurerm/issues/6795

### DIFF
--- a/azurerm/internal/services/apimanagement/api_management_api_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_api_resource.go
@@ -261,7 +261,7 @@ func resourceArmApiManagementApiCreateUpdate(d *schema.ResourceData, meta interf
 
 		//`wsdl_selector` is necessary under format `wsdl`
 		if len(wsdlSelectorVs) == 0 && contentFormat == string(apimanagement.Wsdl) {
-			return fmt.Errorf("`wsdl_selector` is necessary when content format is `wsdl` in API Management API %q", name)
+			return fmt.Errorf("`wsdl_selector` is required when content format is `wsdl` in API Management API %q", name)
 		}
 
 		if len(wsdlSelectorVs) > 0 {
--- a/azurerm/internal/services/apimanagement/api_management_api_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_api_resource.go
@@ -258,6 +258,12 @@ func resourceArmApiManagementApiCreateUpdate(d *schema.ResourceData, meta interf
 			},
 		}
 		wsdlSelectorVs := importV["wsdl_selector"].([]interface{})
+
+		//`wsdl_selector` is necessary under format `wsdl`
+		if len(wsdlSelectorVs) == 0 && contentFormat == string(apimanagement.Wsdl) {
+			return fmt.Errorf("`wsdl_selector` is necessary when content format is `wsdl` in API Management API %q", name)
+		}
+
 		if len(wsdlSelectorVs) > 0 {
 			wsdlSelectorV := wsdlSelectorVs[0].(map[string]interface{})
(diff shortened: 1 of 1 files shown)


azurerm/internal/services/apimanagement/api_management_api_resource.go
```diff
@@ -258,6 +258,12 @@ func resourceArmApiManagementApiCreateUpdate(d *schema.ResourceData, meta interf
 			},
 		}
 		wsdlSelectorVs := importV["wsdl_selector"].([]interface{})
+
```
tnicolaysen, line 261: I saw you linked this to #6795, but I don't think that it's the cause for what I reported. It also affected `swagger_json` formats.

I _think_ it might have something to do with the whole if-statement that this code is in. The comment on line 241-242 is:
> 	// If import is used, we need to send properties to Azure API in two operations.
>	// First we execute import and then updated the other props.

This was added when the underlying API was 2018-XX-XX. and I'm guessing that it was a workaround for a bug back then. I could not find anything in the REST API doc. that stated it has to be two requests. 

Could it be an idea to do this in one request instead of two? Should not be very difficult to merge the request into one.

yupwei68: Hi @tnicolaysen ,for issue https://github.com/terraform-providers/terraform-provider-azurerm/issues/6795, I have tried to add `wsdl_selector` in the code, and it's created with success. Do I miss anything?

tnicolaysen: I can try to make a build from this branch and see if it solves my problem. But, it is still weird that the code performs the update in two steps.

tnicolaysen: I have tested this branch with my setup.

* Creation of new OpenAPI APIs work now.
* Update of existing APIs work.
* Switch from swagger_json to openapi on an existing API worked.
* The primary/secondary key issue is confirmed fixed for me. (Not relevant to this PR).

I also got this expected error:
```
Error: `wsdl_selector` is required when content format is `wsdl` in API Management API "xxx-api"

  on main.tf line 163, in resource "azurerm_api_management_api" "xxx":
 163: resource "azurerm_api_management_api" "xxx" {
```

I didn't know that it was required. It's not required when importing a WSDL through Azure Portal. Maybe they just default to the first or something.

yupwei68: hi @tnicolaysen , yes, I have split the two problems in two PRs. Would you like to try if the below code works for you? And on portal, `wsdl_selector` is required as the screenshot below.
```
resource "azurerm_api_management_api" "example" {
  api_management_name = data.azurerm_api_management.apim.name
  resource_group_name = data.azurerm_api_management.apim.resource_group_name

  display_name      = "Example API"
  name              = "example-api"
  path              = "example-api"
  protocols         = ["https"]
  revision          = "1"
  service_url       = "http://example.org"
  soap_pass_through = true

  // NOTICE: It partially works when not having an import, but it's just one of the issues
  import {
    content_format = "wsdl"
    content_value = file("example.wsdl")

 wsdl_selector {
      service_name  = "Calculator"
      endpoint_name = "CalculatorHttpsSoap11Endpoint"
    }
  }
}
```

image

tnicolaysen: Looks like we're seeing different UIs. Might have something to do with the underlying API-M version and when it was provisioned. 

image

I tested with code that was similar to what you showed, and it worked for me.

tnicolaysen:
Any chance to get this one merged for the next release? :)

PR #7079, opened 2020-05-26
azurerm_logic_app_action_http: add `run_after`

magodo:
Fixes terraform-providers/terraform-provider-azurerm#7069

## Test Result

```bash
💤 😡 127 via 🦉 v1.14.3 make testacc TEST=./azurerm/internal/services/logic/tests TESTARGS="-run='TestAccAzureRMLogicAppActionHttp_runAfter'"                       
TF_ACC=1 go test ./azurerm/internal/services/logic/tests -v -run='TestAccAzureRMLogicAppActionHttp_runAfter' -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMLogicAppActionHttp_runAfter
=== PAUSE TestAccAzureRMLogicAppActionHttp_runAfter
=== CONT  TestAccAzureRMLogicAppActionHttp_runAfter
--- PASS: TestAccAzureRMLogicAppActionHttp_runAfter (184.27s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/logic/tests 184.291s
```

### DIFF
--- a/azurerm/internal/services/logic/logic_app_action_http_resource.go
+++ b/azurerm/internal/services/logic/logic_app_action_http_resource.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/services/logic/mgmt/2019-05-01/logic"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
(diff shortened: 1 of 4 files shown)


azurerm/internal/services/logic/logic_app_action_http_resource.go
```diff
@@ -99,6 +123,10 @@ func resourceArmLogicAppActionHTTPCreateUpdate(d *schema.ResourceData, meta inte
 		"type":   "http",
 	}
 
+	if v, ok := d.GetOk("run_after"); ok {
+		action["runAfter"] = expandLogicAppActionRunAfter(v.(*schema.Set).List())
+	}
```
tombuildsstuff, line 128: do we need to set this to an empty array if this is unset?

tombuildsstuff:
Tests pass:

Screenshot 2020-05-26 at 12 20 55

azurerm/internal/services/logic/logic_app_action_http_resource.go
magodo: The test result shows that is not necessary. Also in the previous case where we didn't set it at all, it still works for the first level action (which is right after the trigger).

PR #7087, opened 2020-05-26
azurerm_app_service_certificate - support for hosting_environment_profile_id

wmdave:


azurerm/internal/services/web/resource_arm_app_service_certificate.go
```diff
@@ -130,8 +136,16 @@ func resourceArmAppServiceCertificateCreateUpdate(d *schema.ResourceData, meta i
...
 	keyVaultSecretId := d.Get("key_vault_secret_id").(string)
+	hostingEnvironmentProfileId := d.Get("hosting_environment_profile_id").(string)
 	t := d.Get("tags").(map[string]interface{})
 
+	var hep *web.HostingEnvironmentProfile
+	if len(hostingEnvironmentProfileId) > 0 {
+		hep = &web.HostingEnvironmentProfile{
+			ID: &hostingEnvironmentProfileId,
+		}
+	}
```
katbyte, line 147: if we move this below the creation of the certificate var
```suggestion
	if len(hostingEnvironmentProfileId) > 0 {
		certificate.CertificateProperties = &web.HostingEnvironmentProfile{
			ID: &hostingEnvironmentProfileId,
		}
	}
```

wmdave: I moved the creation, as suggested, and updated the docs, but it'll take me a bit to dig in to your testing to find good tests for this change.

wmdave: On second thought, you sure you want some sort of positive test?  I'm pretty sure that to do so would require spinning up an App Service Environment and those aren't cheap or quick (or easy to tear down).  Example:

Normal Web App Cert:
```
{
...
  "properties": {
  ...
    "hostingEnvironmentProfile": null,
  }
}
```

ASE Web App Cert:
```
{
...
  "properties": {
  ...
    "hostingEnvironmentProfile": {
      "id": "/subscriptions/<subId>/resourceGroups/<rgName>/providers/Microsoft.Web/hostingEnvironments/<ASEName>",
      "name": "<ASEName>",
      "type": "Microsoft.Web/hostingEnvironments"
    },
  }
}
```
If so, I'm game to try, but new to your testing frameworks so any pointers you could provide would be appreciated.

katbyte: @wmdave - you can just take an existing ASE and cert test and then combine them. cost isn't really an issue for us as much as ensuring the functionality works, and doesn't break in the future

dhdanie:
Okay, I added a test for adding a certificate to the ase tests.

dhdanie:
Okay, have an error.  Will have to figure out how to proceed.

azurerm/internal/services/web/resource_arm_app_service_certificate.go
dhdanie: Okay, I think it's ready to go.  Changed the format of the block as described in a subsequent comment, added test, updated site docs and did terrafmt.  Let me know if anything doesn't look right.  Sorry about the delay in getting it all going.

wmdave: Oh yeah, sorry, dhdanie = wmdave.  Didn't realize I was logged in to a different account.

katbyte:
@wmdave - I do, you'd then be able to just us `azurerm_app_service_environment.test.id`, parse it's id with an ASE parse id function and have everything you need

dhdanie:
There may be an easy answer to this, that I'm missing, but new to your test framework.  I was wanting to validate the certificate's `hosting_environment_profile_id` after the apply, in `Check`, but to calculate the expected value, I need to be able to either retrieve the subscription ID that's being used or the resulting ID property of the ASE after it's created.  How can I retrieve either of those within the test case?  I'll keep looking for it, but if you see this, in the meantime, let me know.

dhdanie:
Okay, switched back and I think updated docs and test case.  Passed locally.

PR #7101, opened 2020-05-27
Update the compute service api-version from 2019-07-01 to 2019-12-01

ArcturusZhang:
Some functionalities in the `azurerm_orchestrated_virtual_machine_scale_set` requires 2019-12-01 to use.


azurerm/internal/services/compute/orchestrated_virtual_machine_scale_set_resource.go
```diff
@@ -59,9 +59,10 @@ func resourceArmOrchestratedVirtualMachineScaleSet() *schema.Resource {
 			},
 
 			"single_placement_group": {
-				Type:     schema.TypeBool,
-				Required: true,
-				ForceNew: true,
+				Type:         schema.TypeBool,
+				Optional:     true,
+				Default:      false,
+				ValidateFunc: validateBoolIsFalse,
```
tombuildsstuff, line 65: this is a breaking change - previously this could be set to `true`?

fitzgeraldsteele: Hi Tom, I'm the Microsoft PM behind this feature, working with @ArcturusZhang to expose to Terraform. I can confirm that for this preview feature, we changed from true to false to support an update of the technology stack. We don't have any customers using spg=true for this preview at the moment, but we do have customers waiting for spg=false. 

Would it be ok to allow this value to be true or false, and just pass the value to Azure to do the validation?

katbyte: @fitzgeraldsteele - right now the property is forced to be true in terraform so anyone using the resource would have `spg=true`, and to date the previous API version still works with `spg=true` - are you saying absolutely no one is currently using orchestrated VMSSs via terraform at this time?

katbyte: (its been released so anyone could be using it atm)

fitzgeraldsteele: Checking the usage data now. We are closing/discontinuing the preview that uses spg=true, and are working to unblock customers using spg=false. Our documentation will have this update. If we document that in the changelog here, can we bring this back into the 2.14 release?

Happy to have a non-github discussion...I don't want to post the usage data here. :)  jerry <dot> steele @ microsoft

PR #7106, opened 2020-05-27
r/nat_gateway_public_ip_association: using a custom ID / support for delta updates in `azurerm_nat_gateway`

tombuildsstuff:
This PR fixes some issues spotted in #6450 - namely:

* Using a custom Resource ID for the `azurerm_nat_gateway_public_ip_association` resource - which includes the Public IP Address ID so that we can determine which Public IP Address is being referenced
* Supporting delta updates in the `azurerm_nat_gateway` resource


tombuildsstuff:
Tests pass:

Screenshot 2020-05-27 at 13 16 59

PR #7125, opened 2020-05-28
Suppress diff for tags of app insights

neil-yechenwei:
fixes terraform-providers/terraform-provider-azurerm#7026


tombuildsstuff:
hi @neil-yechenwei 

Thanks for this PR.

The only supported manner to connect an Application Insights instance to a given resource is via the `hidden-link:` tag - as such this is something users use to connect these resources together and isn't something that we should be filtering out. Instead should a user wish to ignore this value they're able to use Terraform's `ignore_changes` functionality to ignore this field specifically.

Since this is the only supported means of integrating resources in the Application Insights API at this time unfortunately this unorthodox approach to linking resources together needs to be exposed to users via the Tags - and as such whilst I'd like to thank you for this contribution I'm going to close this issue for the moment.

Thanks!

Issue #5729, opened 2020-02-13
Backup Policy resources' tags are reported as added, but in fact are not


shybbko:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

+ Terraform v0.11.14
+ provider.azurerm v1.44.0
+ provider.dns v2.2.0
+ provider.template v2.1.2

### Affected Resource(s)

At least the following two:

* `azurerm_backup_policy_vm`
* `azurerm_backup_protected_vm`

### Terraform Configuration Files

```
resource "azurerm_recovery_services_vault" "test-generic-recovery-vault" {
    name                = "${var.cluster_name}-recovery-vault"
    location            = "${var.region}"
    resource_group_name = "${var.test_rg}"
    sku                 = "standard"

    tags {
        deployment = "terraform",
        environment = "${var.environment}",
        type = "${var.type}",
        component = "${var.component}"
        division = "${var.division}"
        creator = "testuser"
    }
}

resource "azurerm_backup_policy_vm" "test-generic-backup-policy" {
  name                = "${var.cluster_name}-backup-policy"
  resource_group_name = "${var.test_rg}"
  recovery_vault_name = "${azurerm_recovery_services_vault.test-generic-recovery-vault.name}"

  timezone = "UTC"

  backup {
    frequency = "Daily"
    time      = "03:00"
  }

  retention_daily {
    count = 7
  }

    tags {
        deployment = "terraform",
        environment = "${var.environment}",
        type = "${var.type}",
        component = "${var.component}"
        division = "${var.division}"
        creator = "testuser"
    }
}
```

### Expected Behavior

Tags are added to the resources.

### Actual Behavior

Tags are not being added to the resources, however Terraform reports they are.

```
azurerm_backup_policy_vm.test-generic-backup-policy: Creating...
	(...)
  tags.%:                  "" => "6"
  tags.component:          "" => "users-test"
  tags.creator:            "" => "testuser"
  tags.deployment:         "" => "terraform"
  tags.division:           "" => "core"
  tags.environment:        "" => "production"
[...]

PR #7127, opened 2020-05-29
 Remove ineffective tags from RecoveryServices.Backup

neil-yechenwei:
fixes terraform-providers/terraform-provider-azurerm#5729

After confirmed with service team, tags is ineffective. (Azure/azure-rest-api-specs#9251)


tombuildsstuff:
hi @neil-yechenwei 

Thanks for this PR.

Taking a look through here whilst it appears the Azure API is broken ([based on this issue](https://github.com/Azure/azure-rest-api-specs/issues/9251) - since this functionality has shipped we can't easily remove this functionality since #5729 has shown that users wish to use this functionality.

As such whilst I'd like to thank you for this contribution I'm going to close this PR for the moment - but I've replied in the upstream API issue to check-in on when the API Team will fix this API bug.

Thanks!

PR #7147, opened 2020-05-31
AuthorizationRules -> authorizationRules

markusleh:
According to the latest API version for Microsoft.EventHub namespaces/authorizationRules templates, the correct path for authorization rules is `authorizationRules` since API version 2017-04-01.
See https://docs.microsoft.com/en-us/azure/templates/Microsoft.EventHub/namespaces/authorizationrules

Currently deployment fails when getting the authorization rule after creating a new authorization rule or fetching old one.

### DIFF
--- a/azurerm/internal/services/eventhub/eventhub_namespace_authorization_rule_resource.go
+++ b/azurerm/internal/services/eventhub/eventhub_namespace_authorization_rule_resource.go
@@ -119,7 +119,7 @@ func resourceArmEventHubNamespaceAuthorizationRuleRead(d *schema.ResourceData, m
 		return err
 	}
 
-	name := id.Path["AuthorizationRules"] // this is different then eventhub where its authorizationRules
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hey @markusleh 

Taking a look through here, whilst this approach will work for newly created resources - unfortunately this approach won't work when an existing resource has been created using an older API version (where this has the older ID format `authorizationRules`.

One option to work around this would be to parse both approaches and hope that one of them is present (as in #7244) - however in this instance we can instead introduce a State Migration to update the ID of the Resource - meaning that all EventHub Namespace Authorization Rules will have their Resource ID updated from the old (`authorizationRules`) to the new (`AuthorizationRules`).

Since State Migrations are a little involved, whilst I'd like to thank you for this contribution - I'm going to close this in favour of #7248 which adds that State Migration so that this can be consistently parsed as `AuthorizationRules` going forward, I hope you don't mind!

Thanks!

PR #7150, opened 2020-05-31
New Resource `azurerm_static_site`

### LINKED ISSUE #7029: Azure Static Web App support
_This issue was originally opened by @andreujuanc as hashicorp/terraform#25000. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._
    
<hr>
    
### Current Terraform Version
```
N/A
```

### Use-cases
Microsoft just announced Static web apps, which is a very interesting offering.
Official doc: https://docs.microsoft.com/en-us/azure/static-web-apps/overview
I have an upcoming project that might use static web apps, and would love to start with IaC as sooner as possible in the design stage.

### Attempted Solutions
N/A

### Proposal
I believe there might be some dependencies that Microsoft hasnt implemented yet, so we can use this issue has a tracker.

### References
Could not find any.

Best

EDIT: Lit issue ID: I've got hashicorp/terraform#25000

aristosvo:
Fixes #7029

## TerraForm Spec

```hcl
resource "azurerm_static_site" "example" {
  name = "example"
  resource_group_name = "rg"
  location = "West Europe"
  
  github_configuration {
    repo_url          = "https://github.com/aristosvo/azure-static-web-app"
    branch            = "master"
    repo_token        = "xxxxxxxx"

    app_location      = "/"
    api_location      = "/api"
    artifact_location = "dist/angular-basic"
  }
}
```

## Acceptance Tests
```bash
--- PASS: TestAccAzureRMStaticSite_basic (168.77s)
--- PASS: TestAccAzureRMStaticSite_requiresImport (171.88s)
```


aristosvo:
Implementation of the initial version is done, but to acctest this properly we would need the binary test functionality to generate the GitHub repo.

azurerm/internal/services/web/tests/helpers_test.go
```diff
@@ -0,0 +1,9 @@
+package tests
+
+import (
+	"os"
+)
+
+func skipStaticSite() bool {
+	return os.Getenv("ARM_TEST_GITHUB_TOKEN") == ""
```
jackofallops, line 8: This should probably have a `TF` prefix, rather than `ARM` as these are for configuring the client itself, rather than test flags.
```suggestion
	return os.Getenv("TF_ACCTEST_GITHUB_TOKEN") == ""
```

azurerm/internal/services/web/parse/static_site.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+)
+
+type StaticSiteResourceID struct {
+	ResourceGroup string
+	Name          string
+}
+
+func StaticSiteID(input string) (*StaticSiteResourceID, error) {
```
jackofallops, line 14: Could we have a test for this function as for other ID type elsewhere in the provider?

jackofallops:
Hi @aristosvo 
I hope you don't mind, I needed to check if this service had been unblocked for us, so I quickly completed the refactor / rework you started to a testable point and fixed the merge conflicts.  Sadly we're still not going to be able to unblock this due to the GH actions behaviours being incompatible with the service being managed by terraform. I've reached out to my contacts at Microsoft to see if there's a route forward for this.  Any meaningful updates I get I'll post here.

payamazadi:
would be really excited to see this come through, any updates please @jackofallops ?

Oceanswave:
> Hi @aristosvo
> I hope you don't mind, I needed to check if this service had been unblocked for us, so I quickly completed the refactor / rework you started to a testable point and fixed the merge conflicts. Sadly we're still not going to be able to unblock this due to the GH actions behaviours being incompatible with the service being managed by terraform. I've reached out to my contacts at Microsoft to see if there's a route forward for this. Any meaningful updates I get I'll post here.

Without any meaningful updates in a while, I'm wondering if there is an alterative path forward - since 

1. This is 'mostly' a blocker only for test code related to the static web app resource creation having a side effect of creating a GH Action
2. The Github Action that gets created by the static web app creation is represented by a .yml file with the associated repository
3. The Github repo, branch and token is included as part of the configuration

could the test be modified so that it includes test tear down behaviour to reach out to the defined GH repo/branch/token and remove the ```.github/workflows/azure-static-web-apps-{{ d.Id() }}.yml``` file - implementation might look [similar to this](https://github.com/terraform-providers/terraform-provider-github/blob/9c532df8193907ff250481be9060290f65bedd66/github/resource_github_repository_file.go#L285)

jackofallops:
Hi @Oceanswave 
Thanks for the comment and the suggestion. It's an approach we have already considered, but involves vendoring a large dependency for the sake of removing a single file from a repository. 
The nightly/regular testing aspect is indeed part of our blocker for proceeding, this makes a _lot_ of noise quite quickly, but this will also translate to similar issue for users employing this resource for non-production type environments (dev / staging / test / other).  We're looking at options, but waiting on a few answers from the service team.  Ultimately, this is still a preview, so could be subject to significant implementation change right up to GA. It may be that after the answers, we go ahead an include it as is, which some warning(s) around this behaviour and some examples/suggestions of how to stop it getting out of hand (since we'd need to employ this for the acceptance tests also).
fwiw - we're chasing this at every opportunity, as we'd like it in too.

payamazadi:
thanks @jackofallops . Any updates?

Meanwhile, I managed to get this patched into my local version of Terraform and it works flawlessly. This was a big service to the community @aristosvo and is much appreciated!

mpetuska:
@payamazadi, could you please briefly outline how you've managed to patch this in locally?

magodo:
Is that possible to ask the service team provide an option to not create the workflow for users, rather return the content of the yaml file in the creation response. Then users can use the [`github_repository_file`](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository_file) to manage that workflow file. 

This way we do not need to involve the dependency on github related packages, and since we recently opt in the binary testing framework, it should also works for acctest.

@jackofallops WDYT?

magodo:
@jackofallops and @aristosvo I've pushed two commits to this PR.

One is to remove the repo-related properties from static site resource (to avoid service manipulate the repo) and adding some comptued properties (to allow users to add the api key and workflow to the repo by their own).
The other one is to merge with the master branch, changing the test style a bit.

I've passed the acctest locally:

```bash
💢 TF_ACC=1 go test -v -timeout=180m -run=TestAccAzureStaticSite_ ./...
2021/03/16 14:46:00 [DEBUG] not using binary driver name, it's no longer needed
2021/03/16 14:46:00 [DEBUG] not using binary driver name, it's no longer needed
=== RUN   TestAccAzureStaticSite_basic
=== PAUSE TestAccAzureStaticSite_basic
=== RUN   TestAccAzureStaticSite_requiresImport
=== PAUSE TestAccAzureStaticSite_requiresImport
=== CONT  TestAccAzureStaticSite_basic
=== CONT  TestAccAzureStaticSite_requiresImport
--- PASS: TestAccAzureStaticSite_basic (221.86s)
--- PASS: TestAccAzureStaticSite_requiresImport (240.79s)
```

@jackofallops Please take another look, thank you!

tombuildsstuff:
@magodo without the ability to connect this resource to a repository, the provisioned Static Site will be non-functional (e.g. no site will be deployed, which is the point of this service) - so this still appears to need service changes to be usable, or is there some other automated way of connecting the repo?

magodo:
@tombuildsstuff You can use the github terraform provider to do that, like this: https://gist.github.com/magodo/287009b96ebfa6b8c173e62d6907fa0f. Note that you'd need to ensure the github access token has the workflow permission, as it will be used to create a workflow file in that repo.

After above resources are provisioned, the github action will be triggered, which will in turn notify the static site to pull the repo and do the deployment.

aristosvo:
Looks good within current limitations from my point of view! I'd probably add your example code to the documentation to have clear instructions how to use this resource, as without the GitHub configuration functionality is indeed limited.

Just thinking out loud, but maybe we'd even go as far as adding that scenario as an acceptance test? 😇

magodo:
@aristosvo Good point! The reason why I didn't add my example to the document is because currently there is a [bug](https://github.com/integrations/terraform-provider-github/issues/733) in the github provider, which will cause the provider panic on file creation. Once that is addressed, we can add it to the doc and acctest.

scuderig:
Any update on this? The service has been announced as GA but we still haven’t got a way to provision using IaC..

dirien:
@scuderig, asking the same question. Looking forward to use this via TF too.

azurerm/internal/services/web/static_site_resource.go
```diff
@@ -0,0 +1,205 @@
...
+		Sku: &web.SkuDescription{
+			Name: utils.String(d.Get("sku_size").(string)),
+			Tier: utils.String(d.Get("sku_tier").(string)),
+		},
+		StaticSite: &web.StaticSite{},
+		Location:   &loc,
+	}
+
+	_, err := client.CreateOrUpdateStaticSite(ctx, id.ResourceGroup, id.Name, siteEnvelope)
+	if err != nil {
```
katbyte, line 116:
```suggestion
	if _, err := client.CreateOrUpdateStaticSite(ctx, id.ResourceGroup, id.Name, siteEnvelope); err != nil {
```

katbyte (approved):
Aside from one comment looks good to me 👍

PR #7155, opened 2020-06-01
`azurerm_cosmosdb_sql_container` `default_ttl` remove computed

yupwei68:
Fix : https://github.com/terraform-providers/terraform-provider-azurerm/issues/7113

=== RUN   TestAccAzureRMCosmosDbSqlContainer_basic
=== PAUSE TestAccAzureRMCosmosDbSqlContainer_basic
=== CONT  TestAccAzureRMCosmosDbSqlContainer_basic
--- PASS: TestAccAzureRMCosmosDbSqlContainer_basic (1424.70s)
=== RUN   TestAccAzureRMCosmosDbSqlContainer_complete
=== PAUSE TestAccAzureRMCosmosDbSqlContainer_complete
=== CONT  TestAccAzureRMCosmosDbSqlContainer_complete
--- PASS: TestAccAzureRMCosmosDbSqlContainer_complete (1414.45s)
=== RUN   TestAccAzureRMCosmosDbSqlContainer_update
=== PAUSE TestAccAzureRMCosmosDbSqlContainer_update
=== CONT  TestAccAzureRMCosmosDbSqlContainer_update
--- PASS: TestAccAzureRMCosmosDbSqlContainer_update (1531.05s)
PASS

### DIFF
--- a/azurerm/internal/services/cosmos/cosmosdb_sql_container_resource.go
+++ b/azurerm/internal/services/cosmos/cosmosdb_sql_container_resource.go
@@ -78,7 +78,6 @@ func resourceArmCosmosDbSQLContainer() *schema.Resource {
 			"default_ttl": {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				Computed:     true,
 				ValidateFunc: validation.IntAtLeast(-1),
 			},
 


azurerm/internal/services/cosmos/cosmosdb_sql_container_resource.go
```diff
@@ -78,7 +78,6 @@ func resourceArmCosmosDbSQLContainer() *schema.Resource {
 			"default_ttl": {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				Computed:     true,
 				ValidateFunc: validation.IntAtLeast(-1),
```
ArcturusZhang, line 81: Could we also add a `Default: -1` here?
Because I suppose if the user does not specify this attribute, he/she would mean absence of this attribute which should be equivalent to `disable`.
In the current implementation, if you leave this empty, terraform would assign `0` for you.

katbyte: We cannot do this as it will break existing users who depend on the computed value in there state.

katbyte:
Hey @yupwei68,

i'm afraid we cannot just remove computed from a property like this as it could break existing users who depend on the default value the API returns. AS such i am going to close this and comment on the issue.

azurerm/internal/services/cosmos/cosmosdb_sql_container_resource.go
yupwei68: @ArcturusZhang Agree. @katbyte But the field is not `computed`. The api will not return a default value if this field is absent in request. The current computed value should be `0` because this field is absent in request. In this case, shall we remove the `computed`?

PR #7157, opened 2020-06-01
fix scope of name availability check in ASE hosted app services

### LINKED ISSUE #3003: AppService name needs to be globally unique even in internal app service environments

Terraform v0.11.11
+ provider.azurerm v1.22.1

### Affected Resource(s)

* resource_arm_app_service

### Expected Behavior

Creation of AppServices in isolated App Service Environements should be possible with names which are only unique in the scope of the ASE. Working well on the Azure Portal (need to set the subscription, resource group and app service plan in order to get the correct dns zone for the name validation)

### Actual Behavior

Error Message "The name [...] used for the App Service needs to be globally unique and isn't available: Hostname [...] already exists. Please select a different name."

### Steps to Reproduce

Create an AppService Resource into an ASEv2 with an Isolated AppServicePlan with a name of an AppService which is already taken [...].azurewebsites.net

### References

Quickly checked the source code. Looks like the name check in resource_arm_app_service.go starting on line 216 does not consider being executed on a isolated ASE.
https://github.com/terraform-providers/terraform-provider-azurerm/blob/a28128133d8c66a03927b81a765a7539e590599b/azurerm/resource_arm_app_service.go#L216-L227

jackofallops:
fixes #3003

### DIFF
--- a/azurerm/internal/services/web/resource_arm_app_service.go
+++ b/azurerm/internal/services/web/resource_arm_app_service.go
@@ -192,6 +192,7 @@ func resourceArmAppService() *schema.Resource {
 
 func resourceArmAppServiceCreate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*clients.Client).Web.AppServicesClient
+	aspClient := meta.(*clients.Client).Web.AppServicePlansClient
 	ctx, cancel := timeouts.ForCreate(meta.(*clients.Client).StopContext, d)
 	defer cancel()
 
@@ -217,6 +218,19 @@ func resourceArmAppServiceCreate(d *schema.ResourceData, meta interface{}) error
 		Name: utils.String(name),
 		Type: web.CheckNameResourceTypesMicrosoftWebsites,
 	}
+
+	appServicePlanId := d.Get("app_service_plan_id").(string)
+	aspID, err := ParseAppServicePlanID(appServicePlanId)
+	if err != nil {
+		return err
+	}
+	// Check if App Service Plan is part of ASE
+	// If so, the name needs updating to <app name>.<ASE name>.appserviceenvironment.net and FQDN setting true for name availability check
+	aspDetails, _ := aspClient.Get(ctx, aspID.ResourceGroup, aspID.Name)
+	if aspDetails.HostingEnvironmentProfile != nil {
+		availabilityRequest.Name = utils.String(fmt.Sprintf("%s.%s.appserviceenvironment.net", name, aspID.Name))
+		availabilityRequest.IsFqdn = utils.Bool(true)
+	}
 	available, err := client.CheckNameAvailability(ctx, availabilityRequest)
 	if err != nil {
 		return fmt.Errorf("Error checking if the name %q was available: %+v", name, err)
@@ -227,7 +241,6 @@ func resourceArmAppServiceCreate(d *schema.ResourceData, meta interface{}) error
 	}
 
 	location := azure.NormalizeLocation(d.Get("location").(string))
-	appServicePlanId := d.Get("app_service_plan_id").(string)
 	enabled := d.Get("enabled").(bool)
 	httpsOnly := d.Get("https_only").(bool)
 	t := d.Get("tags").(map[string]interface{})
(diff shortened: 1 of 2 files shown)


exmnewellwf:
First off thanks to everyone working on this project and forgive my lack of Github etiquette if this is in the wrong place. This change does not resolve the issue for my deployments and appears to have an error in it. Looking at the code in this PR it seems that it never actually does anything with the ASE, only the App Service Plan (ASP) including using the ASP name where the ASE Name is expected to build the FQDN being passed to the availability checker, which of course then fails every time. 

Here is the code in question, "aspID" (App Service Plan ID?) is being created with the details of the ASP and then aspID.Name is being used in place of `<ASE name>` in building of the FQDN string.

https://github.com/terraform-providers/terraform-provider-azurerm/blob/1354593cc47356da33dc0c5f1fd50bdff33996ae/azurerm/internal/services/web/resource_arm_app_service.go#L222-L240

I have verified this is what is occurring by reviewing the trace logs from my most recent attempt to deploy:

#Item# = Redacted
> 2020-06-16T23:59:30.505Z [DEBUG] plugin.terraform-provider-azurerm_v2.13.0_x5: [DEBUG] AzureRM Request: 
> 2020-06-16T23:59:30.505Z [DEBUG] plugin.terraform-provider-azurerm_v2.13.0_x5: POST /subscriptions/#MySubscriptionId#/providers/Microsoft.Web/checknameavailability?api-version=2019-08-01 HTTP/1.1
> ...
> 2020-06-16T23:59:30.506Z [DEBUG] plugin.terraform-provider-azurerm_v2.13.0_x5: {"name":"#MyAppName#.#MyAppServicePlanName#.appserviceenvironment.net","type":"Microsoft.Web/sites","isFqdn":true}
> 2020-06-16T23:59:30.557Z [DEBUG] plugin.terraform-provider-azurerm_v2.13.0_x5: [DEBUG] AzureRM Response for https://management.azure.com/subscriptions/#MySubscriptionId#/providers/Microsoft.Web/checknameavailability?api-version=2019-08-01: 
> 2020-06-16T23:59:30.557Z [DEBUG] plugin.terraform-provider-azurerm_v2.13.0_x5: HTTP/2.0 200 OK
> ...
> 2020-06-16T23:59:30.557Z [DEBUG] plugin.terraform-provider-azurerm_v2.13.0_x5: {"nameAvailable":false,"reason":"Invalid","message":"Hostname must be less than 64 chars and must follow allowed character guidelines at http://tools.ietf.org/html/rfc952."}

FYI - I remain unable to deploy an App Service into an ILB ASE V2 due to this availability check failing.

jackofallops:
Hi @exmnewellwf 
The debug log above appears to be checking the correct scope for an ASP in an ASE (see the call using `appserviceenvironment.net` from which you've redacted the hostname and ASE name)?  The failure appears to be in the hostname itself:

> 2020-06-16T23:59:30.557Z [DEBUG] plugin.terraform-provider-azurerm_v2.13.0_x5: {"nameAvailable":false,"reason":"Invalid","message":"Hostname must be less than 64 chars and must follow allowed character guidelines at http://tools.ietf.org/html/rfc952."}

which is returned by the name check API. Can you check your hostname is compatible with the RFC linked in the error message? (though this should be caught by the schema validation).  If there's still a problem, could you open a new issue with a much information as you can provide and we'll look into it?

Thanks!

exmnewellwf:
jackofallops,

I was able to cobble together a powershell script that calls Microsoft.Web/checknameavailability and did some testing and can now clearly see the issue.  Only RFC valid characters for App Service Names and ASE Names are allowed, but App Service plans do not have that restriction and thus using it to build the FQDN can result in a failure. Consider the following:

App Name: MyApp
ASE Name: MyASE
AppSvcPlan Name: My_AppSvc_Plan

This code in the PR will assemble the name to be sent to checknameavailability using the name of the App Service Plan:

`availabilityRequest.Name = utils.String(fmt.Sprintf("%s.%s.appserviceenvironment.net", name, aspID.Name)) `

Based on the example names above the code would assemble the following name:

> MyApp.My_AppSvc_Plan.appserviceenvironment.net

This name fails the check because the App Service Plan name includes underscore chars, which is the case for my App Service Plan names.

The code should be assembling the following FQDN to pass to the checker:

> MyApp.MyASE.appserviceenvironment.net

As the App and ASE names are limited to the RFC chars this would pass the check.

exmnewellwf:
@jackofallops,
To verify the fault, I destroyed and recreated my App Service Plans with dashes instead of underscores and checknameavailability passed. The App Services were then all created without issue!!!!

PR #7164, opened 2020-06-01
`azurerm_cdn_endpoint` - `origin_host_header` can now be set to empty

neil-yechenwei:
This PR is the fix for property origin_host_header which mentioned in the issue terraform-providers/terraform-provider-azurerm#1595

For updating origin_host_header, terraform should allow empty value for this property since it can be updated to "" after tested. So I submitted this PR to fix it.

For updating origin_path, seems it's an API issue so that I raised an issue on Azure/azure-rest-api-specs#9660. I think we should add tag "upstream-microsoft" for this problem.


mbfrahry (approved):
LGTM

PR #7182, opened 2020-06-03
Azure Function support for Event Grid event subscription

jrauschenbusch:
Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7176

Adds Azure Function support to `azurerm_eventgrid_event_subscription` resource.

### DIFF
--- a/azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
+++ b/azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
@@ -22,6 +22,7 @@ import (
 
 func enpointPropertyNames() []string {
 	return []string{
+		"azure_function_endpoint",
 		"eventhub_endpoint",
 		"eventhub_endpoint_id",
 		"hybrid_connection_endpoint",
@@ -89,6 +90,30 @@ func resourceArmEventGridEventSubscription() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
+			},
+
+			"azure_function_endpoint": {
+				Type:          schema.TypeList,
+				MaxItems:      1,
+				Optional:      true,
+				ConflictsWith: utils.RemoveFromStringArray(enpointPropertyNames(), "azure_function_endpoint"),
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: azure.ValidateResourceID,
+						},
+						"max_events_per_batch": {
+							Type:     schema.TypeInt,
+							Optional: true,
+						},
+						"preferred_batch_size_in_kilobytes": {
+							Type:     schema.TypeInt,
+							Optional: true,
+						},
+					},
+				},
 			},
 
 			"eventhub_endpoint_id": {
@@ -386,6 +411,11 @@ func resourceArmEventGridEventSubscriptionRead(d *schema.ResourceData, meta inte
 			d.Set("topic_name", props.Topic)
 		}
 
+		if azureFunctionEndpoint, ok := props.Destination.AsAzureFunctionEventSubscriptionDestination(); ok {
+			if err := d.Set("azure_function_endpoint", flattenEventGridEventSubscriptionAzureFunctionEndpoint(azureFunctionEndpoint)); err != nil {
+				return fmt.Errorf("Error setting `%q` for EventGrid Event Subscription %q (Scope %q): %s", "azure_function_endpoint", id.Name, id.Scope, err)
+			}
+		}
 		if v, ok := props.Destination.AsEventHubEventSubscriptionDestination(); ok {
 			if err := d.Set("eventhub_endpoint_id", v.ResourceID); err != nil {
 				return fmt.Errorf("Error setting `%q` for EventGrid Event Subscription %q (Scope %q): %s", "eventhub_endpoint_id", id.Name, id.Scope, err)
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -91,6 +92,30 @@ func resourceArmEventGridEventSubscription() *schema.Resource {
...
 			},
 
+			"azure_function_endpoint": {
+				Type:          schema.TypeList,
+				MaxItems:      1,
+				Optional:      true,
+				ConflictsWith: utils.RemoveFromStringArray(enpointPropertyNames(), "azure_function_endpoint"),
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
```
tombuildsstuff, line 102: rather than using `id` (which is vague) can we make this `function_app_id`?

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -593,6 +627,36 @@ func expandEventGridEventSubscriptionHybridConnectionEndpoint(d *schema.Resource
...
+		AzureFunctionEventSubscriptionDestinationProperties: &props,
+	}
+
+	if len(configs) == 0 {
+		return azureFunctionDestination
+	}
+
+	config := configs[0].(map[string]interface{})
+
+	if v, ok := config["id"]; ok {
```
tombuildsstuff, line 645: (as above) can we make this `function_app_id`?

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -699,6 +763,30 @@ func flattenEventGridEventSubscriptionStorageQueueEndpoint(input *eventgrid.Stor
...
+	results := make([]interface{}, 0)
+
+	if input == nil {
+		return results
+	}
+
+	result := make(map[string]interface{})
+
+	if input.ResourceID != nil {
+		result["id"] = *input.ResourceID
```
tombuildsstuff, line 776: (as above) can we make this `function_app_id`?

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -699,6 +763,30 @@ func flattenEventGridEventSubscriptionStorageQueueEndpoint(input *eventgrid.Stor
...
+
+	if input.MaxEventsPerBatch != nil {
+		result["max_events_per_batch"] = *input.MaxEventsPerBatch
+	}
+
+	if input.PreferredBatchSizeInKilobytes != nil {
+		result["preferred_batch_size_in_kilobytes"] = *input.PreferredBatchSizeInKilobytes
+	}
+
+	return append(results, result)
```
tombuildsstuff, line 787: we need to ensure these fields always have a value in the state - as such could we make this:

```suggestion
	functionAppId := ""
	if input.ResourceID != nil {
		functionAppId = *input.ResourceID
	}
	
	maxEventsPerBatch := 0
	if input.MaxEventsPerBatch != nil {
 		maxEventsPerBatch = *input.MaxEventsPerBatch
 	}
 	
	prefferedBatchSize := 0
	if input.PreferredBatchSizeInKilobytes != nil {
 		prefferedBatchSize = *input.PreferredBatchSizeInKilobytes
 	}
	
	return append(results, map[string]interface{}{
		"function_app_id": functionAppId,
		"max_events_per_batch": maxEventsPerBatch,
		"preferred_batch_size_in_kilobytes": preferredBatchSize,
	})
```

website/docs/r/eventgrid_event_subscription.html.markdown
```diff
@@ -97,6 +99,16 @@ A `storage_queue_endpoint` supports the following:
 
 ---
 
+An `azure_function_endpoint` supports the following:
+
+* `id` - (Required) Specifies the Azure Resource ID that represents the endpoint of the Azure function where the Event Subscription will receive events.
```
tombuildsstuff, line 104: we intentionally avoid the phrase `Resource ID` in Terraform since it's ambiguous (and Azure:TF doesn't always map 1:1) - could we make this:

```suggestion
* `id` - (Required) Specifies the ID of the Function App where the Event Subscription will receive events.
```

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
```diff
@@ -91,6 +92,30 @@ func resourceArmEventGridEventSubscription() *schema.Resource {
 				Computed: true,
 			},
 
+			"azure_function_endpoint": {
```
tombuildsstuff, line 95: could we add an acceptance test covering this new block?

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
jrauschenbusch: Will rename it. But it has to be `function_id`, as the id of the function needs to be used here rather than the one of the function app.

website/docs/r/eventgrid_event_subscription.html.markdown
jrauschenbusch: Ok. But with the term `Function` instead of `Function App`, as this must be referenced here.

azurerm/internal/services/eventgrid/eventgrid_event_subscription_resource.go
jrauschenbusch: I would really like to do an acceptance test for it. Unfortunately, it's pretty much the same as the webhook endpoint, which is also hard to test. You would first have to be able to deploy a function with an input binding of type `eventGridBinding` out of Terraform. But that is currently not possible as there is no terraform resource for it. It is only possible to deploy a function app, which hosts multiple functions inside.

I started with an implementation of an `azurerm_function` resource to also be able to write acceptance tests for the function and webhook endpoints of the `eventgrid_event_subscription`, but it's not that easy to find a good schema design for it.

jrauschenbusch:
@tombuildsstuff Changed everything except the missing acc test. Maybe i can add it by introducing another resource `azurerm_function` which would allow us to create acceptance tests for the webhook and azure function endpoint setups.

tombuildsstuff:
Tests pass:

Screenshot 2020-06-05 at 12 43 35

PR #7202, opened 2020-06-04
New Resource: `azurerm_iot_time_series_insights_access_policy`

mbfrahry:
```
--- PASS: TestAccAzureRMIoTTimeSeriesInsightsAccessPolicy_basic (140.50s)
--- PASS: TestAccAzureRMIoTTimeSeriesInsightsAccessPolicy_update (178.49s)
```


azurerm/internal/services/iottimeseriesinsights/resource_iot_time_series_insights_access_policy.go
```diff
@@ -0,0 +1,212 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringMatch(
+					regexp.MustCompile(`^[-\w\._\(\)]+$`),
+					"IoT Time Series Insights Access Policy name must contain only word characters, periods, underscores, and parentheses.",
```
ArcturusZhang, line 45: The regex also says hyphens are allowed, could we please also update the message accordingly? Or please confirm whether the hyphens are allowed?

azurerm/internal/services/iottimeseriesinsights/resource_iot_time_series_insights_access_policy.go
```diff
@@ -0,0 +1,212 @@
...
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"environment_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringMatch(
+					regexp.MustCompile(`^[-\w\._\(\)]+$`),
+					"IoT Time Series Insights Environment name must contain only word characters, periods, underscores, and parentheses.",
```
ArcturusZhang, line 57: Same as above

azurerm/internal/services/iottimeseriesinsights/resource_iot_time_series_insights_access_policy.go
```diff
@@ -0,0 +1,212 @@
...
+				),
+			},
+
+			"principal_object_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
```
ArcturusZhang, line 65: Since this is an object id, could we add a validation of `IsUUID`?
And I suppose maybe `principal_id` is sufficient, since there are a lot of `principal_id`s in the `identity` blocks of other resources.

azurerm/internal/services/iottimeseriesinsights/resource_iot_time_series_insights_access_policy.go
```diff
@@ -0,0 +1,212 @@
...
+	id, err := parse.TimeSeriesInsightsAccessPolicyID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	response, err := client.Delete(ctx, id.ResourceGroup, id.EnvironmentName, id.Name)
+	if err != nil {
+		if !utils.ResponseWasNotFound(response) {
+			return fmt.Errorf("deleting IoT Time Series Insights Access Policy %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
+		}
```
ArcturusZhang, line 185: I think we should not escape `404` errors here. Whether the SDK gives non-nil errors is defined [here](https://github.com/Azure/azure-rest-api-specs/blob/3b552e66a85a3de4cc51de5998989c3db6a63920/specification/timeseriesinsights/resource-manager/Microsoft.TimeSeriesInsights/preview/2018-08-15-preview/timeseriesinsights.json#L1026) in the `response` section of the swagger. For the listed status codes, SDK will not return non-nil errors for them (in this case, they are 200 and 204). And all other status codes, the SDK will return non-nil errors using the error schema defined [here](https://github.com/Azure/azure-rest-api-specs/blob/3b552e66a85a3de4cc51de5998989c3db6a63920/specification/timeseriesinsights/resource-manager/Microsoft.TimeSeriesInsights/preview/2018-08-15-preview/timeseriesinsights.json#L1036).

This means in this case, `404` will always be an error and indicates something is wrong. If this resource does not exist, and we still want to delete this, the service should return `204` for this.

One of the scenario `404` was returned in deletion should be we are deleting this resource, but the resource group does not exist. I believe in this case the service would return `404`. But in this case I believe we should call out and throw errors.
What do you think?

azurerm/internal/services/iottimeseriesinsights/tests/resource_iot_time_series_insights_access_policy_test.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+
+		resp, err := client.Get(ctx, id.ResourceGroup, id.EnvironmentName, id.Name)
+		if err != nil {
+			return fmt.Errorf("Bad: Get on TimeSeriesInsightsAccessPolicyClient: %+v", err)
+		}
+
+		if utils.ResponseWasNotFound(resp.Response) {
+			return fmt.Errorf("Bad: Time Series Insights Access Policy %q (resource group: %q) does not exist", id.Name, id.ResourceGroup)
+		}
```
ArcturusZhang, line 91: Could we please move this if statement should move into the above `if err != nil` statement? Because when the service returns `404`, the `err` will always be non-nil, this means this piece of code is actually unreachable.

azurerm/internal/services/iottimeseriesinsights/tests/resource_iot_time_series_insights_access_policy_test.go
```diff
@@ -0,0 +1,177 @@
...
+		resp, err := client.Get(ctx, id.ResourceGroup, id.EnvironmentName, id.Name)
+
+		if err != nil {
+			return nil
+		}
+
+		if resp.StatusCode != http.StatusNotFound {
+			return fmt.Errorf("time Series Insights Access Policy still exists: %q", id.Name)
+		}
```
ArcturusZhang, line 118: I suppose we should guarantee that this function only returns nil when the response code is `404` and return non-nil errors for all the other cases.

website/docs/r/iot_time_series_insights_access_policy.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+* `update` - (Defaults to 30 minutes) Used when updating the IoT Time Series Insights Access Policy.
+* `read` - (Defaults to 5 minutes) Used when retrieving the IoT Time Series Insights Access Policy.
+* `delete` - (Defaults to 30 minutes) Used when deleting the IoT Time Series Insights Access Policy.
+
+## Import
+
+Azure IoT Time Series Insights Access Policy can be imported using the `resource id`, e.g.
+
+```shell
+terraform import azurerm_iot_time_series_access_policy.example /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/example/providers/Microsoft.TimeSeriesInsights/environments/environment1/accessPolicies/example
```
ArcturusZhang, line 72: In the parsing function, we are poping out the name segment using `accesspolicies`(small p). Therefore could we either modify this example or check if the parsing function is working properly using the small p?

azurerm/internal/services/iottimeseriesinsights/resource_iot_time_series_insights_access_policy.go
mbfrahry: I'm going based off the docs in Azure and the API accepts any string. I'd like to keep it matching with what the API accepts over forcing our own rules.

azurerm/internal/services/iottimeseriesinsights/resource_iot_time_series_insights_access_policy.go
```diff
@@ -0,0 +1,212 @@
...
+				ForceNew: true,
+				ValidateFunc: validation.StringMatch(
+					regexp.MustCompile(`^[-\w\._\(\)]+$`),
+					"IoT Time Series Insights Access Policy name must contain only word characters, periods, underscores, hyphens, and parentheses.",
+				),
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"environment_name": {
```
katbyte, line 51: i think this might be more clear is time_series_environment_name?

azurerm/internal/services/iottimeseriesinsights/resource_iot_time_series_insights_access_policy.go
```diff
@@ -0,0 +1,212 @@
...
+	policy := timeseriesinsights.AccessPolicyCreateOrUpdateParameters{
+		AccessPolicyResourceProperties: &timeseriesinsights.AccessPolicyResourceProperties{
+			Description:       utils.String(d.Get("description").(string)),
+			PrincipalObjectID: utils.String(d.Get("principal_object_id").(string)),
+			Roles:             expandIoTTimeSeriesInsightsAccessPolicyRoles(d.Get("roles").(*schema.Set).List()),
+		},
+	}
+
+	_, err := client.CreateOrUpdate(ctx, resourceGroup, environmentName, name, policy)
+	if err != nil {
```
katbyte, line 120: these lines cna be combines
```suggestion
	if _, err := client.CreateOrUpdate(ctx, resourceGroup, environmentName, name, policy); err != nil {
```

Issue #7184, opened 2020-06-03
Support for Hybrid Connections in App Service


schmid37:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Hybrid Connections are used to access endpoints and resources in an other network, as [described in the documentation here](https://docs.microsoft.com/en-us/azure/app-service/app-service-hybrid-connections).

Currently, azurerm_app_service does not support adding such a hybrid connection.
It is pretty annoying, adding the hybrid connection manualy after a deployment.

Hybrid connections could be created manualy in a seperate resource group and can be referenced by name and/without namespace (depends on implementation) in the app service. 

### New or Affected Resource(s)

azurerm_app_service

### Potential Terraform Configuration

```terraform
resource "azurerm_app_service" "webapp" {
  name                = "name"
  resource_group_name = "rg_name"
  location            = "West US 2"
  app_service_plan_id = azurerm_app_service_plan.sp.id

  .
  .
  .

  networking {
    hybrid_connection {
      name = "hybrid_connection_name_example"
      namespace = "namespace_example"
    }

    hybrid_connection {
      name = "hybrid_connection_name_example2"
      namespace = "namespace_example2"
    }
  }
}
```

schmid37:
@tombuildsstuff will this be ever implemented?

tombuildsstuff:
Fixed via #7224

PR #7224, opened 2020-06-05
New Resource azurerm_app_service_hybrid_connection

jackofallops:
Adds support for App Service Hybrid Connections
supersedes #5330


azurerm/internal/services/web/app_service_hybrid_connection_resource.go
```diff
@@ -0,0 +1,228 @@
...
+			SendKeyValue: utils.String(""), // The service creates this no matter what is sent, but the API requires the field to be set
+		},
+	}
+
+	hybridConnection, err := client.CreateOrUpdateHybridConnection(ctx, resourceGroup, name, namespaceName, relayName, connectionEnvelope)
+	if err != nil {
+		return fmt.Errorf("Error creating App Service Hybrid Connection %q (resource group %q): %s", name, resourceGroup, err)
+	}
+
+	d.SetId(*hybridConnection.ID)
```
mbfrahry, line 152: We should confirm this is nil and not empty before setting it

azurerm/internal/services/web/tests/app_service_hybrid_connection_resource_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+func TestAccAzureRMAppServiceHybridConnection_requiresImport(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_app_service_hybrid_connection", "test")
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMAppServiceHybridConnectionDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMAppService_basic(data),
```
mbfrahry, line 67: We'll want to update this
```suggestion
				Config: testAccAzureRMAppServiceHybridConnection_update(data),
```

aristosvo:
General question: is this also working for Function Apps, similar to the Regional VNet Connection?

azurerm/internal/services/web/tests/app_service_hybrid_connection_resource_test.go
jackofallops: Good spot - ty!

jackofallops:
> General question: is this also working for Function Apps, similar to the Regional VNet Connection?

Hi @aristosvo - afaik the relay doesn't discriminate between app types so should work with Function Apps also, as documented [here](https://docs.microsoft.com/en-us/azure/azure-functions/functions-networking-options#:~:text=Hybrid%20Connections%20is%20a%20feature,all%20but%20the%20Consumption%20plan.)

schmid37:
I think my issue #7184 refer to this. I think we can close my issue then.

PR #7230, opened 2020-06-05
Fix Linux Consumption Function Apps

borancar:
Do not add WEBSITE_CONTENTSHARE and
WEBSITE_CONTENTAZUREFILECONNECTIONSTRING automatically nor ignore it
when comparing resources for Linux Consumption Function Apps. According
to a bug report they should not be there for Linux
https://github.com/Azure/azure-functions-python-worker/issues/598.

Fixes terraform-providers/terraform-provider-azurerm#5209

### DIFF
--- a/azurerm/internal/services/web/resource_arm_function_app.go
+++ b/azurerm/internal/services/web/resource_arm_function_app.go
@@ -664,8 +664,14 @@ func resourceArmFunctionAppRead(d *schema.ResourceData, meta interface{}) error 
 	delete(appSettings, "AzureWebJobsDashboard")
 	delete(appSettings, "AzureWebJobsStorage")
 	delete(appSettings, "FUNCTIONS_EXTENSION_VERSION")
-	delete(appSettings, "WEBSITE_CONTENTSHARE")
-	delete(appSettings, "WEBSITE_CONTENTAZUREFILECONNECTIONSTRING")
+
+	// Let the user have a final say whether he wants to keep these 2 or not on
+	// Linux consumption plans. They shouldn't be there according to a bug
+	// report (see // https://github.com/Azure/azure-functions-python-worker/issues/598)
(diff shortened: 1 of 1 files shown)


jackofallops:
Tests Pass:
image
(One failure, unrelated to this change.)

TSPereira:
Hi,
I have tried using version 2.14.0 to use this fix and I get an odd issue.

I have two azure functions being published in the same stage using
```hcl
publish:
  stage: publish
  script:
    - cd crawler_func; zip -r ../crawler.zip *
    - func azure functionapp publish <func1_name> --build remote
    - cd ..
    - cd extractor_func; zip -r ../extractor.zip *
    - func azure functionapp publish <func2_name> --build remote
    - cd ..
```
And although I can see in the trace the properties being deleted when the azure function starts to be published for some reason it seems to work for function 1 but not for function 2. 
On function 2 eventually it gets a 404 (NotFound) error at different points.

I reverted back to my previous solution (I have a cleanup stage before publishing the functions just to delete these properties as below and provider version = 1.44.0) and everything works fine in it

```hcl
cleanup:
  stage: cleanup
  script:
    - >
      az functionapp config appsettings delete
      --resource-group <resource_group_name_func1>
      --name <func1_name>
      --setting-names {"WEBSITE_CONTENTAZUREFILECONNECTIONSTRING","WEBSITE_CONTENTSHARE"}
    - >
      az functionapp config appsettings delete
      --resource-group <resource_group_name_func2>
      --name <func2_name>
      --setting-names {"WEBSITE_CONTENTAZUREFILECONNECTIONSTRING","WEBSITE_CONTENTSHARE"}
```

It is weird cause it works fine for the first function but not for the second. And it is not function related as I have tried to revert their order and it still works for the function that comes first and still breaks on 2nd one.
By the way, functions are deployed in separate resource groups

PR #7244, opened 2020-06-08
Fix compatibility issue while parsing resource id for event hub auth rule

neil-yechenwei:
fixes terraform-providers/terraform-provider-azurerm#7162


tombuildsstuff:
hi @neil-yechenwei 

Thanks for this PR.

Taking a look through here whilst this is one approach which could work - it'd be better to introduce a State Migration here to update the ID from the old format to the new format. As such I hope you don't mind but I'm going to close this in favour of #7248 which does that - meaning that all existing users will be updated to the latest ID format, so this parser will work consistently.

Thanks!

PR #7259, opened 2020-06-09
Update doc for max_size_gb of sql elastic pool

neil-yechenwei:
fixes terraform-providers/terraform-provider-azurerm#2909

### DIFF
--- a/website/docs/r/mssql_elasticpool.html.markdown
+++ b/website/docs/r/mssql_elasticpool.html.markdown
@@ -67,9 +67,9 @@ The following arguments are supported:
 
 * `max_size_gb` - (Optional) The max data size of the elastic pool in gigabytes. Conflicts with `max_size_bytes`.
 
-~> **Note:** `max_size_gb` is required when `sku` is specified.
-
 * `max_size_bytes` - (Optional) The max data size of the elastic pool in bytes. Conflicts with `max_size_gb`.
+
+-> **Note:** One of either `max_size_gb` or `max_size_bytes` must be specified.
 
 * `tags` - (Optional) A mapping of tags to assign to the resource.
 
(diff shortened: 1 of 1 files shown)


website/docs/r/mssql_elasticpool.html.markdown
```diff
@@ -67,6 +67,8 @@ The following arguments are supported:
 
 * `max_size_gb` - (Optional) The max data size of the elastic pool in gigabytes. Conflicts with `max_size_bytes`.
 
+~> **Note:** `max_size_gb` is required when `sku` is specified.
+
 * `max_size_bytes` - (Optional) The max data size of the elastic pool in bytes. Conflicts with `max_size_gb`.
```
tombuildsstuff, line 72: since one of either `max_size_gb` or `max_size_bytes` must be specified (and they're marked as conflicting) - this probably wants calling out instead:

```suggestion
* `max_size_bytes` - (Optional) The max data size of the elastic pool in bytes. Conflicts with `max_size_gb`.

-> **Note:** One of either `max_size_gb` or `max_size_bytes` must be specified.
```

website/docs/r/mssql_elasticpool.html.markdown
```diff
@@ -67,6 +67,8 @@ The following arguments are supported:
 
 * `max_size_gb` - (Optional) The max data size of the elastic pool in gigabytes. Conflicts with `max_size_bytes`.
 
+~> **Note:** `max_size_gb` is required when `sku` is specified.
```
tombuildsstuff, line 70: the `sku` field is required - however one of either `max_size_gb` or `max_size_bytes` must be configured - so can we remove this:

```suggestion
```

website/docs/r/mssql_elasticpool.html.markdown
neil-yechenwei: Updated

website/docs/r/mssql_elasticpool.html.markdown
neil-yechenwei: updated

neil-yechenwei:
@tombuildsstuff , thanks for your comments. I've updated code.

PR #7267, opened 2020-06-09
Support custom image IDs in DTL VMs

MaDeG:
This PR aims to solve the issue such that it is not possible to create VMs based on custom images in DevTest Labs but only using normal VMs. This because the `custom_image_id` parameter is not exposed by the Terraform resource but is included in the Azure SDK.
In order to do that I implemented the possibility to specify a `custom_image_id` argument for the resource types `azurerm_dev_test_linux_virtual_machine` and `azurerm_dev_test_linux_virtual_machine`.

This allows to specify a DTL VM as follows:
```
resource "azurerm_dev_test_linux_virtual_machine" "example" {
  name                   = "example-vm03"
  lab_name               = azurerm_dev_test_lab.example.name
  resource_group_name    = azurerm_resource_group.example.name
  location               = azurerm_resource_group.example.location
  size                   = "Standard_DS2"
  username               = "exampleuser99"
  ssh_key                = file("~/.ssh/id_rsa.pub")
  lab_virtual_network_id = azurerm_dev_test_virtual_network.example.id
  lab_subnet_name        = azurerm_dev_test_virtual_network.example.subnet[0].name
  storage_type           = "Premium"
  notes                  = "Some notes about this Virtual Machine."

  custom_image_id        = "/subscriptions/{subscription_id}/resourcegroups/{resource_group}/providers/microsoft.devtestlab/labs/{dtl_name}/customimages/{image_name}"
}

```
Where the `custom_image_id` is the same that would be used in an ARM template.

### DIFF
--- a/azurerm/helpers/azure/devtest.go
+++ b/azurerm/helpers/azure/devtest.go
@@ -86,7 +86,7 @@ func ExpandDevTestLabVirtualMachineGalleryImageReference(input []interface{}, os
 func SchemaDevTestVirtualMachineGalleryImageReference() *schema.Schema {
 	return &schema.Schema{
 		Type:     schema.TypeList,
-		Required: true,
+		Optional: true,
 		MaxItems: 1,
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
--- a/azurerm/internal/services/devtestlabs/dev_test_linux_virtual_machine_resource.go
+++ b/azurerm/internal/services/devtestlabs/dev_test_linux_virtual_machine_resource.go
@@ -118,6 +118,12 @@ func resourceArmDevTestLinuxVirtualMachine() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"custom_image_id": {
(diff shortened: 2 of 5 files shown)


MaDeG:
I am trying to understand what could be the best way to implement tests that set `custom_image_id` for both `azurerm_dev_test_windows_virtual_machine` and `azurerm_dev_test_linux_virtual_machine` since it will be necessary to have an available custom image in the DTL in order to do that.
I haven't found any way to create a new custom image in a DTL using Terraform, and I suspect that the only way to do that would be to create a new resource type (probably leveraging on `azure-sdk-for-go/services/devtestlabs/mgmt/2016-05-15/dtl/customimages.go`).
Do you have any other ideas?

katbyte:
@MaDeG - yea, looks like we'll need to implement that to correctly test this resource

katbyte:
Hey @MaDeG - given we are missing the required resources to test this properly i'm going to close it until they have been implemented. At which time we can reopen and re-review this.

PR #7290, opened 2020-06-11
`azurerm_key_vault` allow case-insensitive `sku_name`

magodo:
KeyVault API (2016-10-01) allows case-insensitive `Sku.name`, so we need
to loosen the constraint also in terraform, which should not be a
breaking change.

(fixes: terraform-providers/terraform-provider-azurerm#7254)


katbyte:
@magodo - this sounds like a bug on the API/portal side for allowing the wrong casing to be used? and we should open an issue on the sdk/specs to bring attention to the service team?

katbyte:
I'm going to close this for now as its been a month without a response - can reopen it once we're ready to move forward

magodo:
@katbyte I have submitted the issue and put the link in code. Do you think it is feasible to reopen this PR and label it as `breaking change`?

tombuildsstuff:
@magodo in this instance it sounds like the casing in the Swagger needs updating?

magodo:
@tombuildsstuff The API allows either case. Whilst I feel this PR should be safe to be merged as it just loosens some constraint.

PR #7313, opened 2020-06-12
Implement azurermproviderlint with two rules: AZURERMR001 & AZURERMS001

magodo:
Implement a linter based on tfproviderlint, adding two azurerm provider specific lint rules: AZURERMR001 and AZURERMS001.
In the meanwhile, modify the makefile and corresponding scripts to:
- Add UT for azurermproviderlint itself to the "test" target
- Replace tfproviderlint executable with azurermproviderlint to run the listing rules, while keep the original rule set, plus adding AZURERMS001, which is a reasonable rule to be utilized.

To making the CI pass, also add another commit to modify the violation of rule AZURERMS001 (where some of them are skipping the rule via comment for special reasons).

One **concern** is that the provider SDK is bumped from v1.6.0 to v.1.7.0, because of the transitive dependency of tfproviderlint maybe.


magodo:
The CI failure is unrelated.

tombuildsstuff:
hey @magodo 

As a general rule we're moving towards using static types in the provider where possible, in that vain we've got an [SDK for Typed Resources](https://github.com/terraform-providers/terraform-provider-azurerm/tree/master/azurerm/internal/sdk) we're prototyping at the moment, which allows us to catch more of these items at compile/unit time rather than lint time.

Whilst there's some benefits to introducing a linter to catch some of the issues that we face today, since longer-term Typed Resources remove the majority of these issues (for example, through strong typing) - ultimately Typed Resources ends up solving this in a slightly different way (at compile/unit test time, rather than retroactively).

As such whilst I'd like to thank you for this contribution and apologies for the delayed review here - I'm going to close this PR for the moment - but longer-term the Typed Resources SDK should allow us to catch these errors in another way.

Thanks!

PR #7314, opened 2020-06-12
 azurerm_api_management_subscription specify resource id

Trazarw:
> Distinction between product_id and product resource id
> Distinction between user_id and user resource id

### DIFF
--- a/website/docs/r/api_management_subscription.html.markdown
+++ b/website/docs/r/api_management_subscription.html.markdown
@@ -51,9 +51,9 @@ The following arguments are supported:
 
 * `resource_group_name` - (Required) The name of the Resource Group in which the API Management Service exists. Changing this forces a new resource to be created.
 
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hey @Trazarw 

Thanks for this PR :)

In the Azure Provider we intentionally avoid using the phrase "Resource ID" since it's ambiguous between a Terraform Resource ID and an Azure Resource ID, where the two aren't necessarily the same, instead we use the phrase `The ID of the {thing}` to reference the ID used by Terraform (which in this instance is the same as the Azure Resource ID). By contrast when we're referencing a GUID/UUID (or other field) of a resource we'll call this out as `The GUID/UUID of the {thing}`.

As such whilst I'd like to thank you for this contribution - I'm going to close this PR for the moment since we intentionally avoid using the term "Resource ID" in documentation.

Thanks!

Issue #5699, opened 2020-02-12
Support for flexible Policy enforcement during assignment


bhummerstone:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When assigning a policy in the Portal, you can enable/disable the Enforcement Mode; this is useful for using the same Policy definition (with a built-in Deny) but disabling enforcement in Non-Production environments.

### New or Affected Resource(s)

* azurerm_policy_assignment

### Potential Terraform Configuration

```hcl
resource "azurerm_policy_assignment" "example" {
  name                 = "example-policy-assignment"
  scope                = "${azurerm_resource_group.example.id}"
  policy_definition_id = "${azurerm_policy_definition.example.id}"
  description          = "Policy Assignment created via an Acceptance Test"
  display_name         = "My Example Policy Assignment"
  enforcement_mode = "disabled"

  parameters = <<PARAMETERS
{
  "allowedLocations": {
    "value": [ "West Europe" ]
  }
}
PARAMETERS
}
```

### References

None

ccampo:
I have created a pull request that addresses this issue. Feedback welcome

https://github.com/terraform-providers/terraform-provider-azurerm/pull/7331

bhummerstone:
Perfect! Thank you :)

ArcturusZhang:
Should be resolved in #7331 , closing....

PR #7331, opened 2020-06-15
`azurerm_policy_assignment` - added support for `enforcement_mode` 

ccampo:
I added the parameter enforcement_mode to allow policies that are assigned to a scope to be set to "DoNotEnforce". Default behaviour is unchanged and still "Default".

Added the code of the implementation, update the documentation html and added a testcase. My first pull request for this project, so please if anything is missing.

### DIFF
--- a/azurerm/internal/services/policy/policy_assignment_resource.go
+++ b/azurerm/internal/services/policy/policy_assignment_resource.go
@@ -114,13 +114,9 @@ func resourceArmPolicyAssignment() *schema.Resource {
 			},
 
 			"enforcement_mode": {
-				Type:     schema.TypeString,
+				Type:     schema.TypeBool,
 				Optional: true,
-				ForceNew: true,
-				ValidateFunc: validation.StringInSlice([]string{
-					string(policy.Default),
-					string(policy.DoNotEnforce),
-				}, false),
+				Default:  true,
 			},
 
 			"not_scopes": {
@@ -139,7 +135,7 @@ func resourceArmPolicyAssignmentCreateUpdate(d *schema.ResourceData, meta interf
 
 	name := d.Get("name").(string)
 	scope := d.Get("scope").(string)
-	enforcementMode := policy.EnforcementMode(d.Get("enforcement_mode").(string))
+	enforcementMode := convertEnforcementMode(utils.Bool(d.Get("enforcement_mode").(bool)))
 	policyDefinitionId := d.Get("policy_definition_id").(string)
 	displayName := d.Get("display_name").(string)
 
@@ -262,6 +258,7 @@ func resourceArmPolicyAssignmentRead(d *schema.ResourceData, meta interface{}) e
 		d.Set("policy_definition_id", props.PolicyDefinitionID)
 		d.Set("description", props.Description)
 		d.Set("display_name", props.DisplayName)
+		d.Set("enforcement_mode", props.EnforcementMode)
 
 		if params := props.Parameters; params != nil {
 			json, err := flattenParameterValuesValueToString(params)
@@ -349,4 +346,12 @@ func expandAzureRmPolicyNotScopes(d *schema.ResourceData) *[]string {
 	}
 
 	return &notScopesRes
+}
+
+func convertEnforcementMode(mode *bool) policy.EnforcementMode {
+	if *mode {
+		return policy.Default
+	} else {
+		return policy.DoNotEnforce
+	}
 }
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/policy/policy_assignment_resource.go
```diff
@@ -113,6 +113,16 @@ func resourceArmPolicyAssignment() *schema.Resource {
 				DiffSuppressFunc: structure.SuppressJsonDiff,
 			},
 
+			"enforcement_mode": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
```
mbfrahry, line 119: While testing locally, I was able to update this attribute without recreating the resource. So we should be able to remove this

mbfrahry, line 119: You mentioned that this attribute defaults to `Default` and we can make that change in the schema by adding `Default: string(policy.Default),`

ccampo: you are right. I removed ForceNew

ccampo: I declare a default, however since the type is now boolean I simply put it to 'true'

ccampo:
> Hi @ccampo based on the doc page [here](https://docs.microsoft.com/en-us/azure/governance/policy/concepts/assignment-structure#enforcement-mode), it sounds to me that this `enforcement_mode` is a switch which corresponds to a `Disable` and `Enable` (see the chart in the doc page). Therefore I suppose maybe we could make this attribute a bool with `true` corresponded to `Default` and `false` corresponded to `DoNotEnforce`?
> 
> @mbfrahry what do you think?

that sounds fair. I updated the pull request the make the parameter a boolean

azurerm/internal/services/policy/policy_assignment_resource.go
```diff
@@ -129,7 +135,7 @@ func resourceArmPolicyAssignmentCreateUpdate(d *schema.ResourceData, meta interf
 
 	name := d.Get("name").(string)
 	scope := d.Get("scope").(string)
-
+	enforcementMode := convertEnforcementMode(utils.Bool(d.Get("enforcement_mode").(bool)))
```
mbfrahry, line 138: It doesn't look like we need a pointer to a boolean so we should remove `utils.Bool` 
```suggestion
	enforcementMode := convertEnforcementMode(d.Get("enforcement_mode").(bool))
```

azurerm/internal/services/policy/policy_assignment_resource.go
```diff
@@ -251,6 +258,7 @@ func resourceArmPolicyAssignmentRead(d *schema.ResourceData, meta interface{}) e
 		d.Set("policy_definition_id", props.PolicyDefinitionID)
 		d.Set("description", props.Description)
 		d.Set("display_name", props.DisplayName)
+		d.Set("enforcement_mode", props.EnforcementMode)
```
mbfrahry, line 261: We'll also want to convert this into a boolean with something like:
```suggestion
		d.Set("enforcement_mode", props.EnforcementMode == policy.Default)
```

azurerm/internal/services/policy/policy_assignment_resource.go
```diff
@@ -339,3 +347,11 @@ func expandAzureRmPolicyNotScopes(d *schema.ResourceData) *[]string {
 
 	return &notScopesRes
 }
+
+func convertEnforcementMode(mode *bool) policy.EnforcementMode {
```
mbfrahry, line 351: We'll need to update this too if we remove the pointer up above.

mbfrahry:
> Hi @ccampo based on the doc page [here](https://docs.microsoft.com/en-us/azure/governance/policy/concepts/assignment-structure#enforcement-mode), it sounds to me that this `enforcement_mode` is a switch which corresponds to a `Disable` and `Enable` (see the chart in the doc page). Therefore I suppose maybe we could make this attribute a bool with `true` corresponded to `Default` and `false` corresponded to `DoNotEnforce`?
> 
> @mbfrahry what do you think?

Exactly right! Good catch @ArcturusZhang

azurerm/internal/services/policy/policy_assignment_resource.go
ccampo: thanks thats a left over from a previous testing to find a problem. I removed it see latest commit

azurerm/internal/services/policy/policy_assignment_resource.go
ccampo: I fixed it. Still not sure what this code actually does, or how this fix changes it. The tfstate was already correct and the code was correctly comparing state against definition.

azurerm/internal/services/policy/policy_assignment_resource.go
ccampo: that was clear, I did that

mbfrahry:
Fixes #5699

mbfrahry:
Thanks for taking the time to write this out @ccampo. It'll make it into the next release!

PR #7346, opened 2020-06-16
app_service_environment resource - allow "Web, Publishing" for the internal_load_balancing_mode argument

jbinko:
Currently the resource app_service_environment supports only three modes for InternalLoadBalancingMode argument:

0. None
1. Web
2. Publishing

But ASE supports one more ILB mode. It is combined value: "Web, Publishing"

0. None = public VIP only
1. Web = only ports 80/443 are mapped to ILB VIP
2. Publishing = only FTP ports are mapped to ILB VIP
3. Web, Publishing = both ports 80/443 and FTP ports are mapped to an ILB VIP.

This fix is accepting additional ILB mode number 3.

resource "azurerm_app_service_environment" "ase" {
name = var.aseName
resource_group_name = var.resourceGroupName
subnet_id = data.azurerm_subnet.aseSubnet.id
pricing_tier = "I2"
front_end_scale_factor = 10
internal_load_balancing_mode = "Web, Publishing"
}

### DIFF
--- a/azurerm/internal/services/web/app_service_environment_resource.go
+++ b/azurerm/internal/services/web/app_service_environment_resource.go
@@ -24,6 +24,10 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
+const (
+	InternalLoadBalancingModeWebPublishing web.InternalLoadBalancingMode = "Web, Publishing"
+)
+
 func resourceArmAppServiceEnvironment() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceArmAppServiceEnvironmentCreate,
@@ -66,6 +70,7 @@ func resourceArmAppServiceEnvironment() *schema.Resource {
 					string(web.InternalLoadBalancingModeNone),
 					string(web.InternalLoadBalancingModePublishing),
 					string(web.InternalLoadBalancingModeWeb),
+					string(InternalLoadBalancingModeWebPublishing),
 				}, false),
 			},
 
--- a/azurerm/internal/services/web/app_service_environment_resource.go
+++ b/azurerm/internal/services/web/app_service_environment_resource.go
@@ -65,6 +65,7 @@ func resourceArmAppServiceEnvironment() *schema.Resource {
 			"internal_load_balancing_mode": {
 				Type:     schema.TypeString,
 				Optional: true,
+				ForceNew: true,
 				Default:  string(web.InternalLoadBalancingModeNone),
 				ValidateFunc: validation.StringInSlice([]string{
 					string(web.InternalLoadBalancingModeNone),
--- a/azurerm/internal/services/web/tests/app_service_environment_resource_test.go
+++ b/azurerm/internal/services/web/tests/app_service_environment_resource_test.go
@@ -159,6 +159,26 @@ func TestAccAzureRMAppServiceEnvironment_withCertificatePfx(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMAppServiceEnvironment_internalLoadBalancer(t *testing.T) {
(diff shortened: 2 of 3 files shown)


azurerm/internal/services/web/app_service_environment_resource.go
```diff
@@ -66,6 +70,7 @@ func resourceArmAppServiceEnvironment() *schema.Resource {
 					string(web.InternalLoadBalancingModeNone),
 					string(web.InternalLoadBalancingModePublishing),
 					string(web.InternalLoadBalancingModeWeb),
+					string(InternalLoadBalancingModeWebPublishing),
 				}, false),
 			},
```
jackofallops, line 75: This schema item should also be `ForceNew` as it cannot be changed after deployment, could you update that too while we're here?

website/docs/r/app_service_environment.html.markdown
```diff
@@ -58,7 +59,9 @@ resource "azurerm_app_service_environment" "example" {
 
 ~> **NOTE** a /24 or larger CIDR is required. Once associated with an ASE this size cannot be changed.
 
-* `internal_load_balancing_mode` - (Optional) Specifies which endpoints to serve internally in the Virtual Network for the App Service Environment. Possible values are `None`, `Web` and `Publishing`. Defaults to `None`.
+* `internal_load_balancing_mode` - (Optional) Specifies which endpoints to serve internally in the Virtual Network for the App Service Environment. Possible values are `None`, `Web`, `Publishing` and combined value `"Web, Publishing"`. Defaults to `None`.
+
+~> **NOTE** You should prefer to select value either `None` or combined value `"Web, Publishing"` as the other values might become deprecated in the future.
```
jackofallops, line 64: Is this documented / published somewhere? If not official, it's probably as well to leave this out for the time being.

jbinko: It is info from Microsoft ASE Program Manager - Christina - (my internal communication). No any official statement I'm aware of. Let me know what you prefer I'll change it accordingly.

azurerm/internal/services/web/app_service_environment_resource.go
jbinko: Makes sense, will fix it.

jbinko:
> Hi @jbinko
> Thanks for this PR, just a a couple of comments below to address.
> 
> Additionally, could we have a test that covers this parameter being set too? (something like `TestAccAzureRMAppServiceEnvironment_internalLoadBalancer`)
> 
> Thanks again!

Please, do you have any guidance how to write such tests? Not sure how it is supposed to work and what is needed. Happy to fix and learn it but need to do some readiness here.

azurerm/internal/services/web/tests/app_service_environment_resource_test.go
```diff
@@ -379,3 +379,39 @@ resource "azurerm_subnet" "gateway" {
...
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMAppServiceEnvironmentExists(data.ResourceName),
+					resource.TestCheckResourceAttr(data.ResourceName, "internal_load_balancing_mode", "Web, Publishing"),
+				),
+			},
+			data.ImportStep(),
+		},
+	})
+}
+
```
jackofallops, line 402: Can we have this at the top to match the layout in the other test definitions? The flow usually runs: Test definitions, Test Functions, Test Configs.

jbinko: Thanks! I updated the documentation and code. -> Removal of possible deprecation note, moving test definition to match code layout

jackofallops:
Tests passing:
image

PR #7347, opened 2020-06-16
New resources: azurerm_eventhub_namespace_dedicated and azurerm_eventhub_dedicated

favoretti:
Support of 2 new resources - `azurerm_eventhub_dedicated` and `azurerm_eventhub_namespace_dedicated`. See discussion below.


favoretti:
@mbfrahry Would you be so king to have a look? This one stands in the way of creating larger hubs on EventHub Clusters.

favoretti:
@jackofallops since you've reviewed Matthew's work on eventhub_cluster resource, can you have a look at this as well please?

azurerm/internal/services/eventhub/eventhub_resource.go
```diff
@@ -273,8 +273,8 @@ func resourceArmEventHubDelete(d *schema.ResourceData, meta interface{}) error {
 func ValidateEventHubPartitionCount(v interface{}, _ string) (warnings []string, errors []error) {
 	value := v.(int)
 
-	if !(32 >= value && value >= 1) {
-		errors = append(errors, fmt.Errorf("EventHub Partition Count has to be between 1 and 32"))
+	if !(1024 >= value && value >= 1) {
+		errors = append(errors, fmt.Errorf("EventHub Partition Count has to be between 1 and 32 for standard hubs and 1024 for Event Hubs Clusters"))
```
katbyte, line 277: Given that this would break validation for non is there any way we could validate its <32 for non clustered eventhubs?

favoretti: If someone can suggest a way to check whether namespace is allocated on a cluster or not - happy to implement.

favoretti: Problem with checking this is really the abstraction of the relationship. EventHubs are created in an EventHub Namespace, which resides on the cluster. Namespace is blissfully unaware of cluster type it lives on, so is EventHub itself.. Any suggestions how to validate it are more than welcome.

jackofallops: This won't be possible to do comprehensively in the schema. The two potential sets of boundaries are incompatible, and it's not possible to determine if it's a dedicated cluster before the schema is validated.  The only option would to check against the min-max of the two together and deal with finer grained checks in code, though this is less than ideal.

This may be moot for the moment as the Namespace resource does not yet support the necessary property to create it on a custom cluster afaik, (assuming I've understood `ClusterArmID` correctly in the model).  It may actually be more appropriate to create new, separate resources for `azurerm_eventhub_namespace_dedicated` and `azurerm_eventhub_dedicated` (or similar), since limits and potential behaviours are sufficiently different.  Thoughts?

favoretti: Oh crack, you're completely right, namespace doesn't support this yet. Haven't thought of that, good catch! Eventhubs themselves are working on a namespace that was created manually on dedicated cluster, so the only difference there will be the limits, but namespace is indeed something that can be done. That said, I might just go with an optional clusterID in the current namespace, rather than creating a whole new resource?

jackofallops: Given the challenges of validation already discussed, and I'm reasonably sure there will be others, I think the separate resources my be better suited (since we can't cross-reference other schema items to know which boundaries are appropriate) - This would mean we can pick up errors at plan, rather than having to get part way through an apply and have to throw an error.  

If the service changes significantly (it has been in preview for >2yrs) to bring things closer together they can be combined/deprecated later, but if they diverge further we'll be well placed to absorb any breaking technical / behavioural change that may be introduced.

favoretti: Let's see... taking a stab at dedicated resources, although this would be my first time doing so, so please bear with the newbie while I'm trying to get it to work.

jackofallops: Thanks @favoretti 

I think you'll be able to use the existing resources as a good start.

If you're not already a member, there's a contributor Slack you can join where we try to help out as much as we can, as well as other contributors that can assist too, the link is in the [README.md](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/README.md) in the root of the project.

favoretti:
@jackofallops can you have a look please? If this is ok and merged, I’ll make another PR for data sources.

website/docs/r/eventhub_dedicated.html.markdown
```diff
@@ -0,0 +1,121 @@
...
+  }
+}
+
+resource "azurerm_eventhub" "example" {
+  name                = "acceptanceTestEventHub"
+  namespace_name      = azurerm_eventhub_namespace.example.name
+  resource_group_name = azurerm_resource_group.example.name
+  partition_count     = 2
+  message_retention   = 1
+}
```
jackofallops, line 48: Wrong resource for the example here?
```suggestion
resource "azurerm_eventhub_dedicated" "example" {
  name                = "acceptanceTestEventHub"
  namespace_name      = azurerm_eventhub_namespace_dedicated.example.name
  resource_group_name = azurerm_resource_group.example.name
  partition_count     = 2
  message_retention   = 1
}
```

website/docs/r/eventhub_dedicated.html.markdown
```diff
@@ -0,0 +1,121 @@
...
+
+* `name` - (Required) Specifies the name of the EventHub Namespace resource. Changing this forces a new resource to be created.
+
+* `namespace_name` - (Required) Specifies the name of the EventHub Namespace. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the resource group in which the EventHub's parent Namespace exists. Changing this forces a new resource to be created.
+
+* `partition_count` - (Required) Specifies the current number of shards on the Event Hub. Needs to be between 1 and 1024. Changing this forces a new resource to be created.
+
+* `message_retention` - (Required) Specifies the number of days to retain the events for this Event Hub. Needs to be between 1 and 90 days; or 1 day when using a Basic SKU for the parent EventHub Namespace.
```
jackofallops, line 63:
```suggestion
* `message_retention` - (Required) Specifies the number of days to retain the events for this Event Hub. Needs to be between `1` and `90`. Only a value of `1` is allowed when using a Basic SKU for the parent EventHub Namespace.
```

website/docs/r/eventhub_dedicated.html.markdown
```diff
@@ -0,0 +1,121 @@
...
+* `id` - The ID of the EventHub.
+
+* `partition_ids` - The identifiers for partitions created for Event Hubs.
+
+
+## Timeouts
+
+
+
+The `timeouts` block allows you to specify [timeouts](https://www.terraform.io/docs/configuration/resources.html#timeouts) for certain actions:
```
jackofallops, line 108: additional whitespace
```suggestion
## Timeouts

The `timeouts` block allows you to specify [timeouts](https://www.terraform.io/docs/configuration/resources.html#timeouts) for certain actions:
```

website/docs/r/eventhub_namespace_dedicated.html.markdown
```diff
@@ -0,0 +1,134 @@
...
+* `default_secondary_connection_string_alias` - The alias of the secondary connection string for the
+    authorization rule `RootManageSharedAccessKey`, which is generated when disaster recovery is enabled.
+
+* `default_secondary_key` - The secondary access key for the authorization rule `RootManageSharedAccessKey`.
+
+## Timeouts
+
+
+
+The `timeouts` block allows you to specify [timeouts](https://www.terraform.io/docs/configuration/resources.html#timeouts) for certain actions:
```
jackofallops, line 121: additional whitespace
```suggestion
## Timeouts

The `timeouts` block allows you to specify [timeouts](https://www.terraform.io/docs/configuration/resources.html#timeouts) for certain actions:
```

azurerm/internal/services/eventhub/eventhub_dedicated_resource.go
```diff
@@ -0,0 +1,409 @@
...
+			}
+
+			output["destination"] = []interface{}{destinationOutput}
+		}
+
+		results = append(results, output)
+	}
+
+	return results
+}
```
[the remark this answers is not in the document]
favoretti: Adjusted.

azurerm/internal/services/eventhub/tests/eventhub_dedicated_resource_test.go
```diff
@@ -0,0 +1,625 @@
...
+  resource_group_name = azurerm_resource_group.test.name
+  sku                 = "Standard"
+}
+
+resource "azurerm_eventhub_dedicated" "test" {
+  name                = "acctest-EH-%d"
+  namespace_name      = azurerm_eventhub_namespace_dedicated.test.name
+  resource_group_name = azurerm_resource_group.test.name
+  partition_count     = 2
+  message_retention   = 7
```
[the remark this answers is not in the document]
favoretti: Set both to 50.

azurerm/internal/services/eventhub/eventhub_dedicated_resource.go
```diff
@@ -0,0 +1,409 @@
...
+					ArchiveNameFormat:        utils.String(archiveNameFormat),
+					BlobContainer:            utils.String(blobContainerName),
+					StorageAccountResourceID: utils.String(storageAccountId),
+				},
+			}
+		}
+	}
+
+	return &captureDescription
+}
```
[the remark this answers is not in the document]
favoretti: Removed and adjusted to use the non-dedicated version.

azurerm/internal/services/eventhub/eventhub_dedicated_resource.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	for _, component := range requiredComponents {
+		if !strings.Contains(value, component) {
+			errors = append(errors, fmt.Errorf("%s needs to contain %q", k, component))
+		}
+	}
+
+	return warnings, errors
+}
```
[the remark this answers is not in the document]
favoretti: Removed.

CHANGELOG.md
```diff
@@ -56,7 +60,7 @@ UPGRADE NOTES:
 
 * `azurerm_kubernetes_cluster` - the Azure Policy add-on now only supports `v2` (as per the Azure API) ([#7233](https://github.com/terraform-providers/terraform-provider-azurerm/issues/7233))
 
-DEPENDENCIES: 
+DEPENDENCIES:
```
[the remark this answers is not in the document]
favoretti: Clear, didn't know that :) Reverted.

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,508 @@
...
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+// Default Authorization Rule/Policy created by Azure, used to populate the
+// default connection strings and keys
+var eventHubNamespaceDedicatedDefaultAuthorizationRule = "RootManageSharedAccessKey"
```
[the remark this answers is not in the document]
favoretti: TBH I'm not even sure where it's used. Removed.

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+	client := meta.(*clients.Client).Eventhub.NamespacesClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+	log.Printf("[INFO] preparing arguments for AzureRM EventHub Namespace creation.")
+
+	name := d.Get("name").(string)
+	clusterID := d.Get("cluster_id").(string)
+	resGroup := d.Get("resource_group_name").(string)
+
+	if features.ShouldResourcesBeImported() && d.IsNewResource() {
```
jackofallops, line 218: Sorry, should have called this one out too
```suggestion
	if d.IsNewResource() {
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+
+func resourceArmeventHubNamespaceDedicatedCreateUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Eventhub.NamespacesClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+	log.Printf("[INFO] preparing arguments for AzureRM EventHub Namespace creation.")
+
+	name := d.Get("name").(string)
+	clusterID := d.Get("cluster_id").(string)
+	resGroup := d.Get("resource_group_name").(string)
```
jackofallops, line 216: Could we rename this, we've been doing the same through the rest of the provider where we can?
```suggestion
	resourceGroup := d.Get("resource_group_name").(string)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+
+	if features.ShouldResourcesBeImported() && d.IsNewResource() {
+		existing, err := client.Get(ctx, resGroup, name)
+		if err != nil {
+			if !utils.ResponseWasNotFound(existing.Response) {
+				return fmt.Errorf("Error checking for presence of existing EventHub Namespace %q (Resource Group %q): %s", name, resGroup, err)
```
jackofallops, line 222:
```suggestion
				return fmt.Errorf("Error checking for presence of existing EventHub Namespace %q (Resource Group %q): %s", name, resourceGroup, err)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+			ClusterArmID:         &clusterID,
+		},
+		Tags: tags.Expand(t),
+	}
+
+	if v, ok := d.GetOk("maximum_throughput_units"); ok {
+		parameters.EHNamespaceProperties.MaximumThroughputUnits = utils.Int32(int32(v.(int)))
+	}
+
+	future, err := client.CreateOrUpdate(ctx, resGroup, name, parameters)
```
jackofallops, line 255:
```suggestion
	future, err := client.CreateOrUpdate(ctx, resourceGroup, name, parameters)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+		return fmt.Errorf("Error creating eventhub namespace: %+v", err)
+	}
+
+	read, err := client.Get(ctx, resGroup, name)
+	if err != nil {
+		return err
+	}
+
+	if read.ID == nil {
+		return fmt.Errorf("Cannot read EventHub Namespace %q (resource group %q) ID", name, resGroup)
```
jackofallops, line 270:
```suggestion
		return fmt.Errorf("Cannot read EventHub Namespace %q (resource group %q) ID", name, resourceGroup)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+
+	ruleSets, hasRuleSets := d.GetOk("network_rulesets")
+	if hasRuleSets {
+		rulesets := eventhub.NetworkRuleSet{
+			NetworkRuleSetProperties: expandEventHubNamespaceNetworkRuleset(ruleSets.([]interface{})),
+		}
+
+		// cannot use network rulesets with the basic SKU
+		if parameters.Sku.Name != eventhub.Basic {
+			if _, err := client.CreateOrUpdateNetworkRuleSet(ctx, resGroup, name, rulesets); err != nil {
```
jackofallops, line 283:
```suggestion
			if _, err := client.CreateOrUpdateNetworkRuleSet(ctx, resourceGroup, name, rulesets); err != nil {
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+	future, err := client.CreateOrUpdate(ctx, resGroup, name, parameters)
+	if err != nil {
+		return err
+	}
+
+	if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("Error creating eventhub namespace: %+v", err)
+	}
+
+	read, err := client.Get(ctx, resGroup, name)
```
jackofallops, line 264:
```suggestion
	read, err := client.Get(ctx, resourceGroup, name)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+func resourceArmeventHubNamespaceDedicatedRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Eventhub.NamespacesClient
+	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	id, err := azure.ParseAzureResourceID(d.Id())
+	if err != nil {
+		return err
+	}
+	resGroup := id.ResourceGroup
```
jackofallops, line 308:
```suggestion
	resourceGroup := id.ResourceGroup
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+	name := id.Path["namespaces"]
+
+	resp, err := client.Get(ctx, resGroup, name)
```
jackofallops, line 311:
```suggestion
	resp, err := client.Get(ctx, resourceGroup, name)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error making Read request on EventHub Namespace %q: %+v", name, err)
+	}
+
+	d.Set("name", resp.Name)
+	d.Set("resource_group_name", resGroup)
```
jackofallops, line 321:
```suggestion
	d.Set("resource_group_name", resourceGroup)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+		d.Set("capacity", sku.Capacity)
+	}
+
+	if props := resp.EHNamespaceProperties; props != nil {
+		d.Set("auto_inflate_enabled", props.IsAutoInflateEnabled)
+		d.Set("maximum_throughput_units", int(*props.MaximumThroughputUnits))
+		d.Set("cluster_id", props.ClusterArmID)
+	}
+
+	ruleset, err := client.GetNetworkRuleSet(ctx, resGroup, name)
```
jackofallops, line 337:
```suggestion
	ruleset, err := client.GetNetworkRuleSet(ctx, resourceGroup, name)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+	if err != nil {
+		return fmt.Errorf("Error making Read request on EventHub Namespace %q Network Ruleset: %+v", name, err)
+	}
+
+	if err := d.Set("network_rulesets", flattenEventHubNamespaceNetworkRuleset(ruleset)); err != nil {
+		return fmt.Errorf("Error setting `network_ruleset` for Evenhub Namespace %s: %v", name, err)
+	}
+
+	keys, err := client.ListKeys(ctx, resGroup, name, eventHubNamespaceDefaultAuthorizationRule)
```
jackofallops, line 346:
```suggestion
	keys, err := client.ListKeys(ctx, resourceGroup, name, eventHubNamespaceDefaultAuthorizationRule)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+func resourceArmeventHubNamespaceDedicatedDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Eventhub.NamespacesClient
+	ctx, cancel := timeouts.ForDelete(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	id, err := azure.ParseAzureResourceID(d.Id())
+	if err != nil {
+		return err
+	}
+	resGroup := id.ResourceGroup
```
jackofallops, line 370:
```suggestion
	resourceGroup := id.ResourceGroup
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+	name := id.Path["namespaces"]
+
+	future, err := client.Delete(ctx, resGroup, name)
```
jackofallops, line 373:
```suggestion
	future, err := client.Delete(ctx, resourceGroup, name)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+	if err != nil {
+		if response.WasNotFound(future.Response()) {
+			return nil
+		}
+		return fmt.Errorf("Error issuing delete request of EventHub Namespace %q (Resource Group %q): %+v", name, resGroup, err)
```
jackofallops, line 378:
```suggestion
		return fmt.Errorf("Error issuing delete request of EventHub Namespace %q (Resource Group %q): %+v", name, resourceGroup, err)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	return waitForeventHubNamespaceDedicatedToBeDeleted(ctx, client, resGroup, name, d)
```
jackofallops, line 381:
```suggestion
	return waitForeventHubNamespaceDedicatedToBeDeleted(ctx, client, resourceGroup, name, d)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+		parameters.EHNamespaceProperties.MaximumThroughputUnits = utils.Int32(int32(v.(int)))
+	}
+
+	future, err := client.CreateOrUpdate(ctx, resGroup, name, parameters)
+	if err != nil {
+		return err
+	}
+
+	if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("Error creating eventhub namespace: %+v", err)
```
jackofallops, line 261: again for clarity on resource type, we should specify the dedicated nature in this message:
```suggestion
		return fmt.Errorf("Error creating dedicated eventhub namespace: %+v", err)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+
+	if props := resp.EHNamespaceProperties; props != nil {
+		d.Set("auto_inflate_enabled", props.IsAutoInflateEnabled)
+		d.Set("maximum_throughput_units", int(*props.MaximumThroughputUnits))
+		d.Set("cluster_id", props.ClusterArmID)
+	}
+
+	ruleset, err := client.GetNetworkRuleSet(ctx, resGroup, name)
+	if err != nil {
+		return fmt.Errorf("Error making Read request on EventHub Namespace %q Network Ruleset: %+v", name, err)
```
jackofallops, line 339: as above
```suggestion
		return fmt.Errorf("Error making Read request on Dedicated EventHub Namespace %q Network Ruleset: %+v", name, err)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	if err := d.Set("network_rulesets", flattenEventHubNamespaceNetworkRuleset(ruleset)); err != nil {
+		return fmt.Errorf("Error setting `network_ruleset` for Evenhub Namespace %s: %v", name, err)
```
jackofallops, line 343: as above
```suggestion
		return fmt.Errorf("Error setting `network_ruleset` for Dedicated Eventhub Namespace %s: %v", name, err)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	keys, err := client.ListKeys(ctx, resGroup, name, eventHubNamespaceDefaultAuthorizationRule)
+	if err != nil {
+		log.Printf("[WARN] Unable to List default keys for EventHub Namespace %q: %+v", name, err)
```
jackofallops, line 348:
```suggestion
		log.Printf("[WARN] Unable to List default keys for Dedicated EventHub Namespace %q: %+v", name, err)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+		}
+		return fmt.Errorf("Error issuing delete request of EventHub Namespace %q (Resource Group %q): %+v", name, resGroup, err)
+	}
+
+	return waitForeventHubNamespaceDedicatedToBeDeleted(ctx, client, resGroup, name, d)
+}
+
+func waitForeventHubNamespaceDedicatedToBeDeleted(ctx context.Context, client *eventhub.NamespacesClient, resourceGroup, name string, d *schema.ResourceData) error {
+	// we can't use the Waiter here since the API returns a 200 once it's deleted which is considered a polling status code..
+	log.Printf("[DEBUG] Waiting for EventHub Namespace (%q in Resource Group %q) to be deleted", name, resourceGroup)
```
jackofallops, line 386:
```suggestion
	log.Printf("[DEBUG] Waiting for Dedicated EventHub Namespace (%q in Resource Group %q) to be deleted", name, resourceGroup)
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,416 @@
...
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{"200"},
+		Target:  []string{"404"},
+		Refresh: eventHubNamespaceDedicatedStateStatusCodeRefreshFunc(ctx, client, resourceGroup, name),
+		Timeout: d.Timeout(schema.TimeoutDelete),
+	}
+
+	if _, err := stateConf.WaitForState(); err != nil {
+		return fmt.Errorf("Error waiting for EventHub NameSpace (%q in Resource Group %q) to be deleted: %+v", name, resourceGroup, err)
```
jackofallops, line 395:
```suggestion
		return fmt.Errorf("Error waiting for Dedicated EventHub NameSpace (%q in Resource Group %q) to be deleted: %+v", name, resourceGroup, err)
```

azurerm/internal/services/eventhub/eventhub_authorization_rule_resource.go
```diff
@@ -95,6 +95,12 @@ func resourceArmEventHubAuthorizationRuleCreateUpdate(d *schema.ResourceData, me
 	locks.ByName(namespaceName, eventHubNamespaceResourceName)
 	defer locks.UnlockByName(namespaceName, eventHubNamespaceResourceName)
 
+	locks.ByName(eventHubName, eventHubDedicatedResourceName)
+	defer locks.UnlockByName(eventHubName, eventHubDedicatedResourceName)
+
+	locks.ByName(namespaceName, eventHubNamespaceDedicatedResourceName)
+	defer locks.UnlockByName(namespaceName, eventHubNamespaceDedicatedResourceName)
+
```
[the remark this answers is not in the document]
favoretti: > @favoretti & @jackofallops - ~do you think `azurerm_eventhub_namespace_dedicated` would be more aptly named `azurerm_eventhub_dedicated_namespace`?~ after some internal discussion and that these belong to a cluster i think `azurerm_eventhub_cluster_hub` and `azurerm_eventhub_cluster_[hub_]namespace` would make the most sense?

MSFT isn't super clear on naming either it seems. Lots of documentation calls them dedicated eventhubs, however portal calls them Event Hubs Cluster.

favoretti:
> after some internal discussion and that these belong to a cluster i think azurerm_eventhub_cluster_hub and azurerm_eventhub_cluster_[hub_]namespace would make the most sense?

@katbyte Naming here a bit awkward. Generally relation is Event Hub Cluster -> Namespace(s) -> EventHub(s). For the non-dedicated part it's just Namespace(s) -> EventHub(s).

`azurerm_eventhub_cluster_namespace` and `azurerm_eventhub_cluster_eventhub`? The latter would be quite a repetition, although can't do much about the fact that Event Hubs Cluster is called that way..

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,415 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: azure.ValidateEventHubNamespaceName(),
+			},
+
+			"cluster_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: azure.ValidateResourceID,
```
tombuildsstuff, line 57: can we add a validation function to confirm this is the ID of EventHub Dedicated Cluster? there's examples in the `./azurerm/internal/services/{...}/validate` folders

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,415 @@
...
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+// Default Authorization Rule/Policy created by Azure, used to populate the
+// default connection strings and keys
+var eventHubNamespaceDedicatedResourceName = "azurerm_eventhub_namespace_dedicated"
+
+func resourceArmEventHubNamespaceDedicated() *schema.Resource {
```
tombuildsstuff, line 28: after chatting internally - rather than splitting this out into it's own resource, since the only difference between these two resources is validation, could we add `dedicated_cluster_id` as an optional field to `azurerm_eventhub_namespace`

with regards to the validation differences - since we're already conditionally capturing this logic via the `create` call for network rules - we can catch this in the same manner for non-dedicated clusters during the `create` - so it should be possible to reuse these for the moment

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,415 @@
...
+	parameters := eventhub.EHNamespace{
+		Location: &location,
+		Sku: &eventhub.Sku{
+			Name:     eventhub.SkuName(sku),
+			Tier:     eventhub.SkuTier(sku),
+			Capacity: &capacity,
+		},
+		EHNamespaceProperties: &eventhub.EHNamespaceProperties{
+			IsAutoInflateEnabled: utils.Bool(autoInflateEnabled),
+			ClusterArmID:         &clusterID,
```
tombuildsstuff, line 245: since we'll be inlining this within the main resource, this'll need to be:

```suggestion
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,415 @@
...
+		},
+		Tags: tags.Expand(t),
+	}
```
tombuildsstuff, line 248: .. and then:

```suggestion
	}
	
	if v := d.Get("cluster_id").(string); v != "" {
		parameters.EHNamespaceProperties.ClusterArmID = utils.String(v)
	}
```

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
```diff
@@ -0,0 +1,415 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: azure.ValidateEventHubNamespaceName(),
+			},
+
+			"cluster_id": {
```
tombuildsstuff, line 53: since this is kind of ambiguous, it'd be worth probably calling this out as:

```suggestion
			"dedicated_cluster_id": {
```

(FWIW chatting internally we believe the `azurerm_eventhub_cluster` resource should probably be named `azurerm_eventhub_dedicated_cluster`, but that's a larger discussion)

favoretti:
> hey @favoretti
> 
> Thanks for this PR :)
> 
> I've taken a look through here and we've had a chat about this internally - and ultimately we believe in this instance this'd make sense to reuse the existing `azurerm_eventhub_namespace` and `azurerm_eventhub` resources - rather than duplicating them for this single field - since based on what we can tell in the Azure API docs these should be complementary (ignoring validation/cluster specific configuration) rather than diverging. We've also discussed that the `azurerm_eventhub_cluster` resource probably wants to become `azurerm_eventhub_dedicated_cluster` - but that's a larger issue and probably a 3.0 thing.
> 
> Whilst I appreciate that's quite a substantial change from this PR - in practice these resources are essentially the same and so this likely makes the most sense until they substantially diverge - what do you think? :)
> 
> Thanks

Right... :D Back to where we started. Adding an optional `clusterID` field to existing namespace was my initial suggestion, @jackofallops said duping them would be a better option.  I'm happy to rework this again, but can you folks agree on one way to go? :)

azurerm/internal/services/eventhub/eventhub_namespace_dedicated_resource.go
favoretti: If we're going to ditch dedicated resources, this change will become obsolete then? I'm getting lost in diverging review opinions to be honest :)

favoretti:
Also there's a challenge when we validate the limits for eventHub partition count and retention settings, where the whole discussion actually started. If I add optional field to namespace, this won't address eventhub validation challenge, since that one is unaware of where it's being created.

jackofallops:
> Right... :D @tombuildsstuff Back to where we started. Adding an optional `clusterID` field to existing namespace was my initial suggestion, @jackofallops said duping them would be a better option. I'm happy to rework this again, but can you folks agree on one way to go? :) See [#7347 (comment)](https://github.com/terraform-providers/terraform-provider-azurerm/pull/7347#discussion_r443364641)

Apologies, quite often we find ourselves part way down an implementation path before we find we've gone in the wrong direction. I've been there myself a lot over the years... Making the existing resource fit a preview API felt wrong at the time, but deeper inspection as this has progressed has brought us full circle.

On the validation point, we'll need to tackle it in the manner I would have preferred we could avoid. The schema validation will need to cover both sets of possibilities, and the code will need to react to those values based on the presence of the `dedicated_cluster_id` attribute.

favoretti:
> > Right... :D @tombuildsstuff Back to where we started. Adding an optional `clusterID` field to existing namespace was my initial suggestion, @jackofallops said duping them would be a better option. I'm happy to rework this again, but can you folks agree on one way to go? :) See [#7347 (comment)](https://github.com/terraform-providers/terraform-provider-azurerm/pull/7347#discussion_r443364641)
> 
> Apologies, quite often we find ourselves part way down an implementation path before we find we've gone in the wrong direction. I've been there myself a lot over the years... Making the existing resource fit a preview API felt wrong at the time, but deeper inspection as this has progressed has brought us full circle.
> 
> On the validation point, we'll need to tackle it in the manner I would have preferred we could avoid. The schema validation will need to cover both sets of possibilities, and the code will need to react to those values based on the presence of the `dedicated_cluster_id` attribute.

No worries at all. I'll probably just abandon this and create a new PR with optional `dedicated_cluster_id` attribute. We'll tackle validation from there on?

djsly:
Personally I think it makes more sense. 

Leveraging the existing namespace resources with an optional clusterID

favoretti:
Closing this one in favor of #7548.

PR #7354, opened 2020-06-17
WIP: New Resource: `azurerm_function`

jrauschenbusch:
Add support to deploy functions incl. script files, bindings, etc.

### DIFF
--- a/azurerm/internal/services/web/parse/function.go
+++ b/azurerm/internal/services/web/parse/function.go
@@ -1,0 +1,40 @@
+package parse
+
+import (
+	"fmt"
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+)
+
+type FunctionID struct {
+	ResourceGroup   string
+	FunctionAppName string
+	Name            string
+}
+
+func ParseFunctionID(input string) (*FunctionID, error) {
+	id, err := azure.ParseAzureResourceID(input)
+	if err != nil {
+		return nil, fmt.Errorf("[ERROR] Unable to parse Function ID %q: %+v", input, err)
+	}
+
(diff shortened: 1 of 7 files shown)


jrauschenbusch:
This resource implementation is almost done. The only problem is the state handling, as there is from my point of view an issue with the function host implementation (see https://github.com/Azure/azure-functions-host/issues/6181). I would expect to get a list of created files within the response, but this isn't the case. Hence the import step test fails.

jrauschenbusch:
I think the Azure Function Host issue won't be resolved that quickly (at least no ETA has been communicated yet). Maybe some workaround around the state handling will do the trick. I'll give it a try. Otherwise we just have to wait until a fix is rolled out.

jrauschenbusch:
Update: There is still the open issue with the missing `Files` map in the Azure Function Host response. Hence state handling is still not working properly. Maybe one of the Terraform members (@tombuildsstuff, @mbfrahry, @jackofallops, @tracypholmes) has an idea how to workaround/solve this problem?

Current status: The resource seems to be working properly (creating / updating / deleting the function), but because the "Files" attribute in the status cannot be set correctly (because the current implementation of the Azure Function Host simply does not set it in the response), a `terraform plan` after a `terraform apply` always informs the user that changes have been made, but this is not the case.

jackofallops:
Hi @jrauschenbusch - I'm going to close this WIP PR for the time being pending a response to the issue you raised above, we can re-open as/when appropriate.  

FWIW - I'm planning on giving the App Service as a whole some TLC as part of our next major release, so this is high on the list for inclusion. If there's no progress on the issue above by then I'll be looking at alternative approaches and seeing what additional steps we can take to move it forward.

PR #7356, opened 2020-06-17
azurerm_postgresql_server: Updated "storage_profile" and removed  some args

ckeivin:
Updated the documentation for "azurerm_postgresql_server"
a) Updated "storage_profile" block
b) Removed the following as it seems the azurerm 2.0 doesn't support these arguments
- "public_network_access_enabled" 
- "ssl_minimal_tls_version_enforced"


ckeivin:
documented code is working fine on latest azurerm_v2.14.0

PR #7371, opened 2020-06-17
`azurerm_kusto_cluster` - Support for `optimized_auto_scale`

jrauschenbusch:
Enables the usage of the optimized auto scale feature of a Kusto Cluster.

### DIFF
--- a/azurerm/internal/services/kusto/kusto_cluster_resource.go
+++ b/azurerm/internal/services/kusto/kusto_cluster_resource.go
@@ -88,8 +88,34 @@ func resourceArmKustoCluster() *schema.Resource {
 
 						"capacity": {
 							Type:         schema.TypeInt,
-							Required:     true,
+							Optional:     true,
+							Computed:     true,
 							ValidateFunc: validation.IntBetween(1, 1000),
+						},
+					},
+				},
+			},
+
+			"optimized_auto_scale": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
+						},
+						"minimum_instances": {
+							Type:         schema.TypeInt,
+							Required:     true,
+							ValidateFunc: validation.IntBetween(0, 1000),
+						},
+						"maximum_instances": {
+							Type:         schema.TypeInt,
+							Required:     true,
+							ValidateFunc: validation.IntBetween(0, 1000),
 						},
 					},
 				},
@@ -189,14 +215,33 @@ func resourceArmKustoClusterCreateUpdate(d *schema.ResourceData, meta interface{
 
 	location := azure.NormalizeLocation(d.Get("location").(string))
 
-	sku, err := expandKustoClusterSku(d)
+	sku, err := expandKustoClusterSku(d.Get("sku").([]interface{}))
 	if err != nil {
 		return err
 	}
 
 	zones := azure.ExpandZones(d.Get("zones").([]interface{}))
 
+	optimizedAutoScale := expandOptimizedAutoScale(d.Get("optimized_auto_scale").([]interface{}))
+
+	if optimizedAutoScale != nil && *optimizedAutoScale.IsEnabled {
+		// if Capacity has not been set use min instances
+		if *sku.Capacity == 0 {
+			sku.Capacity = utils.Int32(int32(*optimizedAutoScale.Minimum))
+		}
+
+		// Capacity must be set for the initial creation when using OptimizedAutoScaling but cannot be updated
+		if d.HasChange("sku.0.capacity") && !d.IsNewResource() {
+			return fmt.Errorf("cannot change `sku.capacity` when `optimized_auto_scaling.enabled` is set to `true`")
+		}
+
+		if *optimizedAutoScale.Minimum > *optimizedAutoScale.Maximum {
+			return fmt.Errorf("`optimized_auto_scaling.maximum_instances` must be >= `optimized_auto_scaling.minimum_instances`")
+		}
+	}
+
 	clusterProperties := kusto.ClusterProperties{
+		OptimizedAutoscale:    optimizedAutoScale,
 		EnableDiskEncryption:  utils.Bool(d.Get("enable_disk_encryption").(bool)),
 		EnableStreamingIngest: utils.Bool(d.Get("enable_streaming_ingest").(bool)),
 		EnablePurge:           utils.Bool(d.Get("enable_purge").(bool)),
@@ -325,6 +370,9 @@ func resourceArmKustoClusterRead(d *schema.ResourceData, meta interface{}) error
 	if err := d.Set("zones", azure.FlattenZones(clusterResponse.Zones)); err != nil {
 		return fmt.Errorf("Error setting `zones`: %+v", err)
 	}
+	if err := d.Set("optimized_auto_scale", flattenOptimizedAutoScale(clusterResponse.OptimizedAutoscale)); err != nil {
+		return fmt.Errorf("Error setting `optimized_auto_scale`: %+v", err)
+	}
 
 	if clusterProperties := clusterResponse.ClusterProperties; clusterProperties != nil {
 		d.Set("enable_disk_encryption", clusterProperties.EnableDiskEncryption)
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/kusto/kusto_cluster_resource.go
```diff
@@ -95,6 +95,34 @@ func resourceArmKustoCluster() *schema.Resource {
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"version": {
+							Type:     schema.TypeInt,
+							Computed: true,
+						},
+						"enabled": {
+							Type:     schema.TypeBool,
+							Computed: true,
+						},
```
tombuildsstuff, line 111: Since it appears (below) that these fields are configurable - should we make these configurable (or remove them)? Since this isn't ForceNew - at the moment it appears there's no way to move from enabled -> disabled

azurerm/internal/services/kusto/kusto_cluster_resource.go
```diff
@@ -322,6 +355,44 @@ func validateAzureRMKustoClusterName(v interface{}, k string) (warnings []string
...
+func flattenOptimizedAutoScale(optimizedAutoScale *kusto.OptimizedAutoscale) []interface{} {
+	if optimizedAutoScale == nil {
+		return []interface{}{}
+	}
+
+	oas := map[string]interface{}{
+		"version":           int(*optimizedAutoScale.Version),
+		"enabled":           *optimizedAutoScale.IsEnabled,
+		"minimum_instances": int(*optimizedAutoScale.Minimum),
+		"maximum_instances": int(*optimizedAutoScale.Maximum),
```
tombuildsstuff, line 390: all 4 of these fields are nilable (and could be nil, so crash when accessing due to a bad API response) - as such can we make this:

```
enabled := false
if optimizedAutoScale.IsEnabled != nil {
	enabled = *optimizedAutoScale.IsEnabled
}

maxInstances := 0
if optimizedAutoScale.Maximum != nil {
	maxInstances = *optimizedAutoScale.Maximum
}

minInstances := 0
if optimizedAutoScale.Minimum != nil {
	minInstances = *optimizedAutoScale.Minimum
}

version := 0
if optimizedAutoScale.Version != nil {
	version = int(*optimizedAutoScale.Version)
}

return []interface{}{
	map[string]interface{}{
		"enabled": enabled,
		"maximum_instances": maxInstances,
		"minimum_instances": minInstances,
		"version": version,
	}
}
```

azurerm/internal/services/kusto/kusto_cluster_resource.go
```diff
@@ -322,6 +355,44 @@ func validateAzureRMKustoClusterName(v interface{}, k string) (warnings []string
 	return warnings, errors
 }
 
+func expandOptimizedAutoScale(d *schema.ResourceData) *kusto.OptimizedAutoscale {
```
tombuildsstuff, line 358: rather than passing in `resourceData` here can we pass in the object we expect instead:

```suggestion
func expandOptimizedAutoScale(input []interface{}) *kusto.OptimizedAutoscale {
	if len(input) == 0 {
		return nil
	}
```

azurerm/internal/services/kusto/tests/kusto_cluster_resource_test.go
```diff
@@ -195,6 +195,30 @@ func TestAccAzureRMKustoCluster_vnet(t *testing.T) {
...
+func TestAccAzureRMKustoCluster_optimizedAutoScale(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_kusto_cluster", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMKustoClusterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMKustoCluster_optimizedAutoScale(data),
```
tombuildsstuff, line 207: since it's possible to update these values too, can we add a test disabling/updating/disabling these settings too?

website/docs/r/kusto_cluster.html.markdown
```diff
@@ -92,6 +94,18 @@ An `identity` block supports the following:
...
+
+A `optimized_auto_scale` block supports the following:
+
+* `version` - (Computed) The version of the template defined, for instance 1.
+
+* `enabled` - (Computed) A boolean value that indicate if the optimized autoscale feature is enabled or not.
+
+* `minimum_instances` - (Required) Minimum allowed instances count. Must between 0 and 1000.
+
+* `maximum_instances` - (Required) Maximum allowed instances count. Must between 0 and 1000.
```
tombuildsstuff, line 107: can we quote these:

```suggestion
* `maximum_instances` - (Required) The maximum number of allowed instances. Must between `0` and `1000`.
```

website/docs/r/kusto_cluster.html.markdown
```diff
@@ -92,6 +94,18 @@ An `identity` block supports the following:
...
 
+---
+
+A `optimized_auto_scale` block supports the following:
+
+* `version` - (Computed) The version of the template defined, for instance 1.
+
+* `enabled` - (Computed) A boolean value that indicate if the optimized autoscale feature is enabled or not.
+
+* `minimum_instances` - (Required) Minimum allowed instances count. Must between 0 and 1000.
```
tombuildsstuff, line 105: can we quote these:

```suggestion
* `minimum_instances` - (Required) The minimum number of allowed instances. Must between `0` and `1000`.
```

website/docs/r/kusto_cluster.html.markdown
```diff
@@ -92,6 +94,18 @@ An `identity` block supports the following:
 
 ~> **NOTE:** When `type` is set to `SystemAssigned`, the Principal ID can be retrieved after the cluster has been created. More details are available below. See [documentation](https://docs.microsoft.com/en-us/azure/active-directory/managed-service-identity/overview) for additional information.
 
+---
+
+A `optimized_auto_scale` block supports the following:
+
+* `version` - (Computed) The version of the template defined, for instance 1.
+
+* `enabled` - (Computed) A boolean value that indicate if the optimized autoscale feature is enabled or not.
```
tombuildsstuff, line 103: these are computed - so if they're not being user-configurable then we should remove them (and put them in the "Attributes" section below)

jrauschenbusch: I‘ll change at least the enabled attribute to be configurable. The meaning of the version attribute isn‘t really clear for me. The spec claims that it‘s a template version. Maybe there are several optimized auto scale templates available which can be used instead of the min/max values. 🤷🏼‍♂️

azurerm/internal/services/kusto/kusto_cluster_resource.go
jrauschenbusch: `enabled` will be changed to be configurable. I also have to test whether ForceNew is required if this property is configurable. meaning of version needs also to be evaluated.

tombuildsstuff: if version is unclear let's remove this for now and default it to version 1 - we can always expose it later if it proves to be useful to configure :)

azurerm/internal/services/kusto/kusto_cluster_resource.go
```diff
@@ -87,9 +87,37 @@ func resourceArmKustoCluster() *schema.Resource {
 						},
 
 						"capacity": {
+							Type:          schema.TypeInt,
+							Optional:      true,
```
jrauschenbusch, line 91: Will be changed to Optional and Computed, as the value depends on the optimized auto scale settings if optimized auto scale is enabled.

azurerm/internal/services/kusto/kusto_cluster_resource.go
```diff
@@ -177,14 +205,20 @@ func resourceArmKustoClusterCreateUpdate(d *schema.ResourceData, meta interface{
...
+	sku, err := expandKustoClusterSku(d.Get("sku").([]interface{}))
 	if err != nil {
 		return err
 	}
 
 	zones := azure.ExpandZones(d.Get("zones").([]interface{}))
 
+	optimizedAutoScale := expandOptimizedAutoScale(d.Get("optimized_auto_scale").([]interface{}))
+	if optimizedAutoScale != nil {
+		sku.Capacity = optimizedAutoScale.Minimum
```
jrauschenbusch, line 217: We set the value of capacity to the value of `minimum_instances`. Capacity is a required attribute, although it will be set to the value of `minimum_instances` after creating or updating the resource.

website/docs/r/kusto_cluster.html.markdown
```diff
@@ -66,7 +68,10 @@ A `sku` block supports the following:
 
 * `name` - (Required) The name of the SKU. Valid values are: `Dev(No SLA)_Standard_D11_v2`, `Dev(No SLA)_Standard_E2a_v4`, `Standard_D11_v2`, `Standard_D12_v2`, `Standard_D13_v2`, `Standard_D14_v2`, `Standard_DS13_v2+1TB_PS`, `Standard_DS13_v2+2TB_PS`, `Standard_DS14_v2+3TB_PS`, `Standard_DS14_v2+4TB_PS`, `Standard_E16as_v4+3TB_PS`, `Standard_E16as_v4+4TB_PS`, `Standard_E16a_v4`, `Standard_E2a_v4`, `Standard_E4a_v4`, `Standard_E8as_v4+1TB_PS`, `Standard_E8as_v4+2TB_PS`, `Standard_E8a_v4`, `Standard_L16s`, `Standard_L4s` and `Standard_L8s`
 
-* `capacity` - (Required) Specifies the node count for the cluster. Boundaries depend on the sku name.
+* `capacity` - (Optional) Specifies the node count for the cluster. Boundaries depend on the sku name.
```
jrauschenbusch, line 71: Changed to Optional

jrauschenbusch:
@tombuildsstuff Finally all acc tests are running:

jrauschenbusch:
@tombuildsstuff Finally acceptance tests are running. Please review.

```sh
make acctests SERVICE='kusto' TESTARGS='-run=TestAccAzureRMKustoCluster_optimizedAutoScale' TESTTIMEOUT='60m'
TF_ACC=1 go test -v ./azurerm/internal/services/kusto/tests/ -run=TestAccAzureRMKustoCluster_optimizedAutoScale -timeout 60m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMKustoCluster_optimizedAutoScale
=== PAUSE TestAccAzureRMKustoCluster_optimizedAutoScale
=== CONT  TestAccAzureRMKustoCluster_optimizedAutoScale
--- PASS: TestAccAzureRMKustoCluster_optimizedAutoScale (1997.43s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/kusto/tests 1999.638s
```

PR #7374, opened 2020-06-17
`azurerm_kusto_cluster` - Support `trusted_external_tenants`

jrauschenbusch:
Enables access to Kusto clusters from external tenants.


katbyte (changes requested):
Thanks for allt he PRs @jrauschenbusch! this looks good but i think we'll need to mark it as computed? (and maybe config mode atter so it can be cleared to nothing)

getting a perma diff running the tests
```
trusted_external_tenants.#: "1" => "0"
          trusted_external_tenants.0: "*" => ""
```

jrauschenbusch:
@katbyte Fixed your notes. Please review.

mbfrahry (approved):
LGTM!

PR #7378, opened 2020-06-17
Fix #7376 - azurerm_storage_data_lake_gen2_filesystem crashes during plan if SA deleted

### LINKED ISSUE #7376: azurerm_storage_data_lake_gen2_filesystem crashes during plan if ADLS deleted
### Community Note

`Terraform v0.12.26`

### Affected Resource(s)

* `azurerm_storage_data_lake_gen2_filesystem`

```hcl
resource "azurerm_storage_account" "adlsaccount" {
  name                     = "mytestadls"
  resource_group_name      = "test"
  location                 = "northeurope"
  account_replication_type = "GRS"
  account_kind             = "StorageV2"
  is_hns_enabled           = "true"
}


resource "azurerm_storage_data_lake_gen2_filesystem" "example" {
  name               = "example"
  storage_account_id = azurerm_storage_account.adlsaccount.id
}
```

### Debug Output

`Error: Error retrieving File System "example" in Storage Account "mytestadls": datalakestore.Client#GetProperties: Failure sending request: StatusCode=0 -- Original Error: context deadline exceeded`

### Expected Behavior

The plan should be generated successfully 

### Actual Behavior

Error message as shown above

### Steps to Reproduce

1. `terraform apply`
2. Delete the ADLS storage account manually (example via portal)
3. `terraform plan`

storey247:
added fix to check storage account exists before attempting to read the filesystem
added test
changed filesystem acc tests to use hns

This PR address the issue #7376

### DIFF
--- a/azurerm/internal/services/storage/resource_arm_storage_data_lake_gen2_filesystem.go
+++ b/azurerm/internal/services/storage/resource_arm_storage_data_lake_gen2_filesystem.go
@@ -165,6 +165,7 @@ func resourceArmStorageDataLakeGen2FileSystemUpdate(d *schema.ResourceData, meta
 }
 
 func resourceArmStorageDataLakeGen2FileSystemRead(d *schema.ResourceData, meta interface{}) error {
+	accountsClient := meta.(*clients.Client).Storage.AccountsClient
 	client := meta.(*clients.Client).Storage.FileSystemsClient
 	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
 	defer cancel()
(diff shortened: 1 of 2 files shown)


storey247:
@ArcturusZhang this is an excellent point, and i first started looking into the fix in the giovanni client that gets used. However, looking at the logic there, it seemed that there was a deliberate separation between the concerns of retrieving storage account properties and retrieving filesystem properties.

I therefore took the decision that to fix the issue it was probably best to take the approach in this PR, use the SA client to check if the storage account exists before using the filesystem client to retrieve the filesystem properties.

stuartleeks:
Hi @ArcturusZhang  - if I've understood correctly, I think you are suggesting that `context deadline exceeded` isn't the most helpful response when getting properties for a filesystem where the storage account doesn't exist? If so, I completely agree and would love to see a more helpful error response for this scenario. I'm not sure what the timescales for that would be?

ArcturusZhang:
Hi @storey247 and @stuartleeks My initial though is that this should be a valid work-around rather than a fix, which may hide something wrong behind the curtain. But as a valid work-around, this indeed will fix the corresponding issue. I personally think this should be good to go.

In the meantime, it would be best if we could make a new issue about the error message. The error comes from `giovanni` the storage data plane SDK, and in giovanni it comes from `go-autorest` (according to offline discussion with Tom)

Issue #7368, opened 2020-06-17
Support for Azure Firewall Manager


sarnowski:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The [Azure Firewall Manager](https://azure.microsoft.com/en-us/services/firewall-manager/) is in public preview. It is a new method to manage Azure Firewalls especially in combination with the [Virtual WAN](https://azure.microsoft.com/en-us/services/virtual-wan/). The Virtual WAN has deep integration with the Firewall Manager to provide centralized routing control in the [Hub and Spoke architecture](https://docs.microsoft.com/en-us/azure/architecture/reference-architectures/hybrid-networking/hub-spoke). In order to leverage the new construct, it is also necessary to implement the new Virtual Hub Route Tables that allow dynamic assignments to spokes.

### New or Affected Resource(s)

* **azurerm_firewall**: the resource needs to be enhanced to support the new Hub SKU which comes with the possibility to add new Azure Policies.
* **azurerm_firewall_policy**: is the general new approach to define policies also in hierarchies
* **azurerm_firewall_policy_rule_collection_group**: provides the rule groups, similar to the old collections attached to the new policies
* **azurerm_virtual_hub_route_table**: new routing table construct that allow spokes to be forced to use the firewall

### Potential Terraform Configuration

```hcl
resource "azurerm_firewall" "default" {
  name = "AzureFirewall_vhub-${var.global.location}"
  location = var.location

  resource_group_name = azurerm_resource_group.wan.name

  sku_name = "AZFW_Hub"
  virtual_hub_id = azurerm_virtual_hub.default.id

  firewall_policy_id = azurerm_firewall_policy.central.id
}

resource "azurerm_firewall_policy" "central" {
  name = "Central"
  location = var.location

  resource_group_name = azurerm_resource_group.wan.name

  threat_intel_mode = "Alert"
}

resource "azurerm_firewall_policy_rule_collection_group" "administration" {
  name = "azfwrg-administration"
  firewall_policy_name = azurerm_firewall_policy.central.name
  resource_group_name = azurerm_resource_group.wan.name

  priority = 100

[...]

PR #7379, opened 2020-06-17
First implementation of new Firewall Manager #7368

sarnowski:


magodo:
Hi @sarnowski Many thanks for this PR :+1:

As you know, the firewall policy API has changed a bit in 2020-05-01 version, the PR to upgrade network package version is merged recently. Based on that, I have made several PRs, which essentially supports the same sets of resources as those in your PR.
The [PR](#7390) for `azurerm_firewall_policy` is under review now. There some other PRs for `azurerm_firewall_policy` and modifying existing `azurerm_firewall` to support policy will be submitted once #7390 is merged.

In order not to better track the progress, do you mind me close this PR in favor of #7390 and other following PRs?

BTW, I noticed you implement virtual hub route table in your PR, this seems not so related to the firewall policy, or no?

sarnowski:
Hi, I was some time off. Thank you very much for picking up the topic.

For the virtual hub route table: If I remember correctly, you need to use that construct in order to configure the route table that is pushed to the peered vnets. Only with that construct can you push routes that force internal and Internet traffic to go via the Firewall.

PR #7386, opened 2020-06-18
Fix inconsistent order of property identity_ids for function app

neil-yechenwei:
fixes terraform-providers/terraform-provider-azurerm#7350


katbyte:
Hey @neil-yechenwei,

Thanks for the PR - however as this is a breaking change to anyone who depends on the ability to access the list elements we can't make it until 3.0. As such i am going to close this PR until we get to 3.0 where we will go through the 3.0 milestone and revive any PR that is still applicable.

Issue #7284, opened 2020-06-10
azurerm_virtual_hub_connection resource fails when connection is recreated


jrominesf:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.12.24
provider.azurerm v2.13.0

### Affected Resource(s)

* `azurerm_virtual_hub_connection`

### Terraform Configuration Files

```hcl
variable "subscription_id" {}

provider "azurerm" {
  subscription_id = var.subscription_id
  features {}
}

resource "azurerm_resource_group" "rg" {
  provider = azurerm
  name     = "vwan-issue-rg"
  location = "westus2"
}

resource "azurerm_virtual_wan" "vwan" {
  provider                   = azurerm
  name                       = "vwan-issue-vwan"
  resource_group_name        = azurerm_resource_group.rg.name
  location                   = azurerm_resource_group.rg.location
  allow_vnet_to_vnet_traffic = true
  tags = {
    env = "dev"
  }
}

resource "azurerm_virtual_hub" "vhub" {
  provider            = azurerm
  name                = "vwan-issue-vhub"
  resource_group_name = azurerm_resource_group.rg.name
  location            = azurerm_resource_group.rg.location
  virtual_wan_id      = azurerm_virtual_wan.vwan.id
  address_prefix      = "172.17.2.0/24"
  tags = {
    env = "dev"
  }
}

resource "azurerm_virtual_network" "vnet" {
  provider            = azurerm
  name                = "vwan-issue-vnet"
  resource_group_name = azurerm_resource_group.rg.name
  location            = azurerm_resource_group.rg.location
  address_space       = ["10.127.224.0/19"]
}

resource "azurerm_virtual_hub_connection" "vwanpeering" {
  provider                                       = azurerm
  name                                           = "${azurerm_virtual_network.vnet.name}-conn"
  virtual_hub_id                                 = azurerm_virtual_hub.vhub.id
  remote_virtual_network_id                      = azurerm_virtual_network.vnet.id
  hub_to_vitual_network_traffic_allowed          = false
[...]

PR #7388, opened 2020-06-18
Manage Virtual Hub Connection with new HubVirtualNetworkConnection API

neil-yechenwei:
fixes terraform-providers/terraform-provider-azurerm#7284

Service team suggests to use HubVirtualNetworkConnection API in newer version 2020-05-01 to manage virtual hub connection since the property virtualNetworkConnection as part of virtual hub api has been deprecated and there is no bug which 7284 described in newer version after tested.


neil-yechenwei:
The issue #7284 would be fixed in PR https://github.com/terraform-providers/terraform-provider-azurerm/pull/7601. So I closed this PR.

PR #7407, opened 2020-06-19
Update property name for kusto database principal

neil-yechenwei:
fixes terraform-providers/terraform-provider-azurerm#7383


jrauschenbusch:
@neil-yechenwei @jackofallops Another proposal. We could add a whole new resource `azurerm_database_principal_assignment` and deprecate the current one.

Why? Because at the time of writing the `azurerm_database_principal` resource, there was no dedicated API endpoint for managing principal assignments. But as of 2020-02-15 API there are now additional endpoints to manage cluster and database principals.

API example: https://github.com/Azure/azure-rest-api-specs/blob/master/specification/azure-kusto/resource-manager/Microsoft.Kusto/stable/2020-02-15/examples/KustoDatabasePrincipalAssignmentsCreateOrUpdate.json

Go model: https://github.com/Azure/azure-sdk-for-go/blob/master/services/kusto/mgmt/2020-02-15/kusto/models.go#L1761

Proposal:
```tf
resource "azurerm_database_principal_assignment" "example" {
  "principal_id"   = "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX" // user email, application ID, or security group name.
  "principal_type" = "App" // Possible values include: 'App', 'Group', 'User'
  "role"           = "Viewer" //  Possible values include: 'Admin', 'Ingestor', 'Monitor', 'User', 'UnrestrictedViewers', 'Viewer'
  "tenant_id"      = "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"
}
```

Why not replacing everything within the current implementation?
1. because it would be a breaking change and a lot of users would complain about it. (incompatible state, different attributes)
1. because there are dedicated endpoints and we now have the opportunity to use real IDs (non-artifical IDs) as we had to create before.

The same design could be applied to a new `azurerm_cluster_principal_assignment` resource.

azurerm/internal/services/kusto/kusto_database_principal_resource.go
```diff
@@ -129,16 +138,16 @@ func resourceArmKustoDatabasePrincipalCreate(d *schema.ResourceData, meta interf
...
 	switch principalType {
 	case "User":
-		fqn = fmt.Sprintf("aaduser=%s;%s", objectID, clientID)
+		fqn = fmt.Sprintf("aaduser=%s;%s", clientID, tenantID)
 	case "Group":
-		fqn = fmt.Sprintf("aadgroup=%s;%s", objectID, clientID)
+		fqn = fmt.Sprintf("aadgroup=%s;%s", clientID, tenantID)
 	case "App":
-		fqn = fmt.Sprintf("aadapp=%s;%s", objectID, clientID)
+		fqn = fmt.Sprintf("aadapp=%s;%s", clientID, tenantID)
```
katbyte, line 150: What is the reasoning behind changing the FQDN format here? as a client belongs in a tenant should it be tenant/client not client/tenant? this will change ther esource id and break existing users.

katbyte: i missed the above review from an old tab in changed - i like the proposal for a new resource and deprecating the old one entirely (and maybe adding notes in the docs & resource/property deprecation messages about the confusion

jrauschenbusch:
I've now created a PR for my resource proposal:
https://github.com/terraform-providers/terraform-provider-azurerm/pull/7484

Please take a look and let me know what you think about it.

Deprecation of the current resource is still open, but would be also part of the PR.

jackofallops:
Thanks @jrauschenbusch - That feels like a more comprehensive way to handle the issue; as renaming of the schema here to correct an unfortunate earlier mis-naming feels unwieldy and likely to cause confusion.  I'm going to close this in favour of your new resource on the newer API.

Issue #7336, opened 2020-06-16
Support for threatIntelligenceMode property on azurerm_firewall


rhys96:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
By default, the Azure Firewall is created with a Threat Intelligence mode of 'Alert'. Microsoft now offers "Off, Alert & Deny" modes. The mode selection is offered in the New-AzFirewall Azure CLI command. It would be really useful to be able to configure this attribute in Terraform.

The linked issue states the issue was resolved, however, I can't see it mentioned in any of the Terraform documentation for azurerm-firewall.

Linked issue: [resource_arm_firewall, resource_arm_firewall_application_rule_collection – Mssql and Threat Intelligence #4595](url)

### New or Affected Resource(s)

- azurerm_firewall

### Potential Terraform Configuration

```hcl
resource "azurerm_firewall" "this" {
  name                = local.firewall_name
  location            = data.azurerm_resource_group.this.location
  resource_group_name = data.azurerm_resource_group.this.name
  threat_intelligence_mode = "Alert"

  ip_configuration {
    name                 = format("%s-ipconfig", local.firewall_name)
    subnet_id            = module.firewall_subnet.subnet_ids[0]
    public_ip_address_id = azurerm_public_ip.this.id
  }
}
```

### References

- [https://docs.microsoft.com/en-us/azure/firewall/threat-intel](url)

- [https://docs.microsoft.com/en-us/rest/api/firewall/azurefirewalls/createorupdate#azurefirewallthreatintelmode](url)

magodo:
Hi @rhys96 Thank you for letting us know this :+1:
The linked issue indeed didn't address this gap, I will hopefully submit a PR soon to fix it!

rhys96:
I had already created a PR request [#7414](Adding AzureFirewallThreatIntelMode property to firewall_resource), but I couldn't figure out how to add it to the issue (this is my first contribution to a project).  Thank you for adding the PR :)

magodo:
That is easy, you only need to mention the issue in the description of the PR in format of `#issue_number` (e.g. #7336 in this case).

PR #7414, opened 2020-06-20
Adding AzureFirewallThreatIntelMode property to firewall_resource

rhys96:
Adds feature for #7336.

Allows you to set the AzureThreatIntelMode property when creating an azurerm_firewall resource.


katbyte:
Hi @rhys96,

Thanks for the PR but it seems #7437 was opened and merged already with this functionality, as such i am going to close this.

Issue #4351, opened 2019-09-17
"terraform import azurerm_role_definition" pulls null values for scope


Grant-Rc:
Hello team,

I had a Team member show me a better way to structure my Terraform and pushing state to S3 and getting it into a local gitlab etc.
In doing so I lost the state file for my Azure custom roles, he said I can use the import function to sort this out.

Now this is were my issue comes, the import works fine but when I run Terraform plan it says 
azurerm_role_definition.customrole must be replaced
 scope              = "/subscriptions/00000000-0000-0000-0000-000000000000" # forces replacement
looking at the state file the entries for scope are null 
            "scope": null

and as defined here https://www.terraform.io/docs/providers/azurerm/r/role_definition.html

scope - (Required) The scope at which the Role Definition applies too, such as /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333, /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333/resourceGroups/myGroup, or /subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333/resourceGroups/myGroup/providers/Microsoft.Compute/virtualMachines/myVM. **Changing this forces a new resource to be created.**

So its changing from "null" to a value and as such forcing a the role to be replaced. This would be fine but you cant delete the role because it is in use across multiple subscriptions.

workaround was to manual edit the state file to add the scope but that not ideal.

anyone come across this? googling i dont find direct link to this issue.

tombuildsstuff:
Fixed via #7424

PR #7424, opened 2020-06-22
`azurerm_role_definition` - terraform import now sets scope to prevent a force recreate

logachev:
When running `terraform import` for `azurerm_role_definition` resources, Read operation doesn't expand `Scope`. 

As a result, on `terraform apply` or `terraform plan`, this resource is forced to be recreated which is impossible due to existing role assignments.

This is a simple fix to populate `scope` from `id`.

Also, removed `trim /` from `parseResourceId`, unclear why this is required. `Delete` operation works regardless of `/` symbol + input requires scope to use `/` as a first symbol.


mbfrahry:
Thanks for this @logachev!
Fixes #4351

PR #7426, opened 2020-06-22
azurerm_function_app: optional SyncFunctionTriggers on create\update

logachev:
Azure Functions have a support for `WEBSITE_RUN_FROM_PACKAGE` property which essentially means it is possible to create e2e pipeline to create and deploy Azure Functions purely with Terraform. (Sync Triggers docs: https://docs.microsoft.com/en-us/azure/azure-functions/functions-deployment-technologies#trigger-syncing)

In the pipeline, we prepare Zip file with the function and everything else is handled by Terraform (create infrastructure if missing, upload blob, generate sas, update Azure Function).

Unfortunately, Azure Functions require `SyncFunctionsTrigger` each time there is a change in triggers (which includes new timer functions e.g.). So, in order to make scenario work e2e, we have to use the provisioner to call this API via az cli.

It creates unnecessary complexity (when running via Azure DevOps task, we have to auth az cli..) while Terraform has all tools to support e2e.

This PR adds optional parameter that allows e2e scenario. Another unfortunate part of this API is that it works in `fire and forget` mode, so there is no way to check if it was successful. Azure Core Func Tool apporach is to wait 5 seconds before calling this API: https://bit.ly/312J1Eo so I replicated this in TF code.

Please let me know if you think this is something that can be included in the provider.

Thanks!

### DIFF
--- a/azurerm/internal/services/web/resource_arm_function_app.go
+++ b/azurerm/internal/services/web/resource_arm_function_app.go
@@ -200,6 +200,12 @@ func resourceArmFunctionApp() *schema.Resource {
 				Default:  false,
 			},
 
+			"sync_triggers": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
 			"daily_memory_time_quota": {
 				Type:     schema.TypeInt,
 				Optional: true,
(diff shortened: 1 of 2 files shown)


jackofallops:
Hi @logachev 

Thanks for this PR.

I've done some investigation around this and I'm afraid it doesn't make sense to make the proposed change to the function_app resource in this way.  The sync api is intended to be called conditionally for circumstances that Azure cannot determine a sync is required, rather than any time terraform is run (incidentally, unless there's a change to the resource config in your CI pipeline, terraform won't fire the sync api call in any case).  I appreciate the extra complexity in your pipeline, since you're using ADO have you considered the HTTP POST options in the link you provided, rather than using `az cli`?  

I'm going to close this PR based on the above.

Issue #5102, opened 2019-12-06
Azure Disk Encryption with VM Extensions vs. Using Managed Disks


stefan-rapp:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
```
Terraform v0.12.7
+ provider.azurerm v1.33.1
```

### Affected Resource(s)

* `azurerm_managed_disk`
* `azurerm_virtual_machine_extension`

### Terraform Configuration Files

I created several VMs with managed (data) disks a few month ago according to the following resource :

```hcl
resource "azurerm_managed_disk" "md" {
    count                   = "${var.md_count}"
    name                    = "${var.md_name}-disk${count.index}"
    location                = "${var.md_location}"
    resource_group_name     = "${var.md_rg_name}"
    storage_account_type    = "${var.md_storage_account_type}"
    create_option           = "${var.md_create_option}"
    source_uri              = "${var.md_create_option == "Import" ? var.md_source_uri : null}"
    source_resource_id      = "${var.md_create_option == "Copy" || var.md_create_option == "Restore" ? var.md_source_resource_id : null}"
    image_reference_id      = "${var.md_create_option == "FromImage" ? var.md_image_reference_id : null}"
    os_type                 = "${var.md_create_option == "Copy" || var.md_create_option == "Import" ? var.md_os_type : null}"
    disk_size_gb            = "${var.md_disk_size_gb != "" ? var.md_disk_size_gb : null}"
    tags                    = "${merge(local.default_tags, var.md_tags)}"
}
```

I configured all VMs already (Software, Services, etc) and now the customer has the requirement to encrypt the disks using **BitLocker**. So I created the following resource to encrypt 'All' disk of a VM, and it worked fine so far:

```hcl
resource "azurerm_virtual_machine_extension" "vm_encry_win" {
  count                         = "${var.vm_encry_os_type == "Windows" ? 1 : 0}"
  name                          = "${var.vm_encry_name}"
  location                      = "${var.vm_encry_location}"
  resource_group_name           = "${var.vm_encry_rg_name}"
[...]

PR #7482, opened 2020-06-25
Updated reference Managed disks encryption block

asgr86:
with reference to Issue terraform-providers #5102 updating documentation to include a ignore_changes on encryption block when using Azure Disk Encryption Agents


tombuildsstuff:
hey @asgr86 

Thanks for this PR :)

Taking a look into #5102, Terraform should detect that the `encryption_settings` block is missing from the users Terraform Configuration and prompt them to add this (unless they opt to ignore this, as in #5102) - as such I believe this is working as intended, and whilst I'd like to thank you for this contribution I'm going to close this PR for the moment.

Thanks!

PR #7496, opened 2020-06-25
Replace underscores with hyphens in example service bus namespace names as underscores are not valid

RiverPhillips:
This PR aims to make all examples that create a service bus namespaces use valid names


RiverPhillips:
That's great, thanks for the quick turnaround.

PR #7498, opened 2020-06-25
Fix issues with frontdoor custom https configuration

### LINKED ISSUE #6057: frontdoor custom_https_configuration can't update certificate
### Community Note

### Affected Resource(s)

* `azurerm_frontdoor`

```hcl

  name     = "debug-frontdoor-timj"
  location = "uksouth"
}

data "azurerm_key_vault" "vault" {
  name = "***"
  resource_group_name = "***"
}

resource "azurerm_frontdoor" "example" {
  name                                         = "timj-debug"
  location                                     = "Global"
  resource_group_name                          = azurerm_resource_group.example.name
  enforce_backend_pools_certificate_name_check = false

  routing_rule {
    name               = "exampleRoutingRule1"
    accepted_protocols = ["Http", "Https"]
    patterns_to_match  = ["/*"]
    frontend_endpoints = ["exampleFrontendEndpoint1"]
    forwarding_configuration {
      forwarding_protocol = "HttpOnly"
      backend_pool_name   = "exampleBackendBing"
    }
  }

  backend_pool_load_balancing {
    name = "exampleLoadBalancingSettings1"
  }

  backend_pool_health_probe {
    name = "exampleHealthProbeSetting1"
  }

  backend_pool {
    name = "exampleBackendBing"
    backend {
      host_header = "****"
      address     = "****"
      http_port   = 80
      https_port  = 443
    }

    load_balancing_name = "exampleLoadBalancingSettings1"
    health_probe_name   = "exampleHealthProbeSetting1"
  }

  frontend_endpoint {
    name                              = "exampleFrontendEndpoint1"
    host_name                         = "plum.timja.dev"
    custom_https_provisioning_enabled = true
    custom_https_configuration {
      certificate_source = "AzureKeyVault"
      azure_key_vault_certificate_secret_name = "plum-timja-dev"
      azure_key_vault_certificate_secret_version = "557c2ff331a74e00a413289f01b1b82e"
      azure_key_vault_certificate_vault_id = data.azurerm_key_vault.vault.id
    }
  }

}

```

### Debug Output

The `enableHttps` endpoint isn't called

### Expected Behavior

Certificate should be updated by updating the version field

### Actual Behavior

Plan says it will change the version but it doesn't

### Steps to Reproduce

1. `terraform apply`
2. update certificate version to a new one
3. `terraform apply`
4. observe cert hasn't updated

### Important Factoids

I've taken a look through the code,
this line is the issue:

https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/frontdoor/resource_arm_frontdoor.go#L596

It only updates the https configuration if its state is disabled

If I change it to just:
```
if customHttpsProvisioningEnabled
```

Then it works

I ran it locally and after ~2hr it was updated

I ran it again and its still running after 20m so I don't think that calling this endpoint every time is a great idea,

I guess we need to check each field and see if there's been an update unless anyone has a better idea?	

### References

* #0000

cc @WodansSon

luigibk:
This PR aims to introduce the following changes:

A fix for the below issue which prevents custom https configurations in frontdoor from being updated properly

Separating the custom https configuration resource from the main frontdoor resource. This should allow for parallel update of these resources as currently they are processed serially and that always leads to a timeout given enough frontend endpoints or endless waiting given how long each custom https configuration update can take (e.g. certificate update).

(fixes #6057 )


luigibk:
Hi @WodansSon 
I currently have an issue at apply stage as after the custom https configurations get created (successfully) I get an error:

```
When applying changes to
azurerm_frontdoor_custom_https_configuration.example_custom_https_0, provider
"registry.terraform.io/-/azurerm" produced an unexpected new value for was
present, but now absent.
issue tracker.
```

azurerm/internal/services/frontdoor/frontdoor_resource.go
```diff
@@ -32,6 +33,9 @@ func resourceArmFrontDoor() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 
+		//MigrateState:  resourceAzureRMFrontDoorMigrateState,
```
timja, line 36:
```suggestion
```

azurerm/internal/services/frontdoor/frontdoor_resource.go
```diff
@@ -423,17 +427,21 @@ func resourceArmFrontDoor() *schema.Resource {
...
 						},
 						"custom_https_configuration": {
-							Type:     schema.TypeList,
-							Optional: true,
-							MaxItems: 1,
+							Type:       schema.TypeList,
+							Optional:   true,
+							Computed:   true,
+							MaxItems:   1,
+							Deprecated: "Deprecated in favour of `custom_https_configuration` resource",
```
timja, line 444: shouldn't it use the full name?
```suggestion
							Deprecated: "Deprecated in favour of `azurerm_frontdoor_custom_https_configuration` resource",
```

Looking at the other deprecated config it looks to be all changing attributes inside of the same resource which is why it's not qualified

azurerm/internal/services/frontdoor/frontdoor_resource.go
```diff
@@ -601,40 +609,55 @@ func resourceArmFrontDoorCreateUpdate(d *schema.ResourceData, meta interface{})
...
-					// Enable Custom Domain HTTPS for the Frontend Endpoint
-					if err := resourceArmFrontDoorFrontendEndpointEnableHttpsProvisioning(ctx, true, name, frontendEndpointName, resourceGroup, customHTTPSConfigurationUpdate, meta); err != nil {
-						return fmt.Errorf("Unable enable Custom Domain HTTPS for Frontend Endpoint %q (Resource Group %q): %+v", frontendEndpointName, resourceGroup, err)
-					}
-				} else if !customHttpsProvisioningEnabled && provisioningState == frontdoor.CustomHTTPSProvisioningStateEnabled {
-					// Disable Custom Domain HTTPS for the Frontend Endpoint
-					if err := resourceArmFrontDoorFrontendEndpointEnableHttpsProvisioning(ctx, false, name, frontendEndpointName, resourceGroup, frontdoor.CustomHTTPSConfiguration{}, meta); err != nil {
-						return fmt.Errorf("Unable to disable Custom Domain HTTPS for Frontend Endpoint %q (Resource Group %q): %+v", frontendEndpointName, resourceGroup, err)
-					}
+	return resourceArmFrontDoorRead(d, meta)
```
timja, line 620: what's this for?

timja:
Missing website documentation for new resource and any required updates to the existing documentation

azurerm/internal/services/frontdoor/frontdoor_resource.go
```diff
@@ -423,17 +427,21 @@ func resourceArmFrontDoor() *schema.Resource {
 							Default:  0,
 						},
 						"custom_https_provisioning_enabled": {
-							Type:     schema.TypeBool,
-							Required: true,
+							Type:       schema.TypeBool,
+							Optional:   true,
+							Computed:   true,
+							Deprecated: "Deprecated in favour of `custom_https_configuration` resource",
```
timja, line 433: https://github.com/terraform-providers/terraform-provider-azurerm/pull/7498/files#r448222425
```suggestion
							Deprecated: "Deprecated in favour of `azurerm_frontdoor_custom_https_configuration` resource",
```

azurerm/internal/services/frontdoor/frontdoor_resource.go
```diff
@@ -601,40 +609,55 @@ func resourceArmFrontDoorCreateUpdate(d *schema.ResourceData, meta interface{})
...
+		} else if !customHttpsProvisioningEnabled && provisioningState == frontdoor.CustomHTTPSProvisioningStateEnabled {
+			// Disable Custom Domain HTTPS for the Frontend Endpoint
+			if err := resourceArmFrontDoorFrontendEndpointEnableHttpsProvisioning(ctx, false, frontDoorName, frontendEndpointName, resourceGroup, frontdoor.CustomHTTPSConfiguration{}, meta); err != nil {
+				return fmt.Errorf("Unable to disable Custom Domain HTTPS for Frontend Endpoint %q (Resource Group %q): %+v", frontendEndpointName, resourceGroup, err)
+			}
 		}
 	}
 
-	return resourceArmFrontDoorRead(d, meta)
+	return nil
```
timja, line 660: it's intentional that this doesn't return an updated resource I assume?

timja: I see that it's doing it at the end of the previous function

website/docs/r/frontdoor.html.markdown
```diff
@@ -70,6 +70,89 @@ resource "azurerm_frontdoor" "example" {
 }
 ```
 
+Custom https configurations can also be defined as a separate resource. This allows for parallel creation/update.
```
timja, line 73: shouldn't this be in a separate file?
looking at:

https://www.terraform.io/docs/providers/azurerm/r/key_vault.html
https://www.terraform.io/docs/providers/azurerm/r/key_vault_access_policy.html

luigibk: 👍

website/docs/r/frontdoor_custom_https_configuration.html.markdown
```diff
@@ -0,0 +1,155 @@
...
+page_title: "Azure Resource Manager: azurerm_frontdoor_custom_https_configuration"
+description: |-
+  Manages the Custom Https Configuration for an Azure Front Door Frontend Endpoint.
+---
+
+# azurerm_frontdoor_custom_https_configuration
+
+Manages the Custom Https Configuration for an Azure Front Door Frontend Endpoint..
+
+~> **NOTE:** Custom https configurations for a Front Door Frontened Endpoint be defined both within [the `azurerm_frontdoor` resource](frontdoor.html) via the `custom_https_configuration` block and by using a separate resource, as described in the following sections.
```
timja, line 13:
```suggestion
~> **NOTE:** Custom https configurations for a Front Door Frontened Endpoint can be defined both within [the `azurerm_frontdoor` resource](frontdoor.html) via the `custom_https_configuration` block and by using a separate resource, as described in the following sections.
```

website/docs/r/frontdoor_custom_https_configuration.html.markdown
```diff
@@ -0,0 +1,155 @@
...
+
+-> **NOTE:** Defining custom https configurations using a separate `azurerm_frontdoor_custom_https_configuration` resource allows for parallel creation/update.
```
timja, line 15: shouldn't these disclaimers be added to the main frontdoor resource as well?

luigibk: 👍

luigibk:
Hi @WodansSon 

I have fixed the issue I asked about some time ago + a few others, did some clean up  and have added docs. It would be great if you could have a look.

It's all working as far as I can see apart from one use case: 
- when I delete an endpoint + its custom https configuration, the 2 operations proceed in parallel and the custom https configuration disable returns an error as the endpoint has been (or is being) deleted. Not sure why that is so as a custom_https_configuration depends on its endpoint so my understanding is that the 2 operations should not run in parallel in the first place. Any ideas?

luigibk:
Fixed the issue discussed above related to endpoint delete. I think that should be all.

WodansSon:
@luigibk and @timja, I apologize for the late reply on this (I've been on vacation and waiting to hear back on the service team about this issue) and thank you so much for this PR, it more or less looks really good from a code point of view, but unfortunately due to the architecture of the Front Door API the changes you are proposing will not work as intended. The Front Door API is monolithic, as when you pass your configuration to the top level controller it will then figure out an order of operations, what needs to be done, and then queues the work for the sub systems to pick up. Therefore, the resource has been coded in the way it has for a reason, and as inconvenient as it is, given the current state of the API, it must be done in this order. To do what you would like to do with this PR the following steps should be taken:

1) Disable Custom HTTPS (this may take up to 24 hours to complete)
2) Update the keys in your key vault
3) Enable Custom HTTPS (this may take up to 48 hours to complete)

There is a lot of work the front door API is doing behind the scenes, de-provisioning the cert, updating DNS records, etc... so, I know this is less than ideal, but until the service team updates the API architecture this behavior is currently by design. 

**Per The Front Door Service Team:**
"AFD currently does not support autorotation and does not automatically detect if the customer has changed their cert in their key vault. If custom HTTPS is enabled, there's currently no way to automatically provision the new cert with no customer interaction. If a customer has updated their certificate, they would have to re-enable HTTPS with the new cert version, or as you suggested, disable and re-enable."

I know how much work went into this PR and I am truly sorry, but on advice of the service team I am going to have to reject this PR until the service team updates their API to allow for independent async operation on the resource.

timja:
I replied here as I saw that first:

https://github.com/terraform-providers/terraform-provider-azurerm/issues/6057#issuecomment-654610827

I think the service team misunderstood your question to them, this isn’t about auto rotation it’s about allowing it to be updated, currently in the provider the only way to update it is to disable and enable https which has downtime, and is completely unnecessary

luigibk:
@WodansSon I think there might be a misunderstanding. 
I added a comment in the related issue: 
https://github.com/terraform-providers/terraform-provider-azurerm/issues/6057#issuecomment-654952928

WodansSon:
@luigibk you need to format the HCL in your test code to get rid of the lint error. I tried fixing it for you but I was denied permission to push to your fork.

azurerm/internal/services/frontdoor/tests/frontdoor_custom_https_configuration_resource_test.go
```diff
@@ -0,0 +1,161 @@
...
+%s
+
+resource "azurerm_frontdoor_custom_https_configuration" "test" {
+  front_door_name                   = azurerm_frontdoor.test.name
+  frontend_endpoint_name            = azurerm_frontdoor.test.frontend_endpoint[0].name
+  resource_group_name               = azurerm_resource_group.test.name
+  custom_https_provisioning_enabled = true
+
+  custom_https_configuration {
+    certificate_source                         = "FrontDoor"
```
WodansSon, line 79:
```suggestion
    certificate_source = "FrontDoor"
```

azurerm/internal/services/frontdoor/tests/frontdoor_custom_https_configuration_resource_test.go
```diff
@@ -0,0 +1,161 @@
...
+      https_port  = 443
+    }
+
+    load_balancing_name = local.load_balancing_name
+    health_probe_name   = local.health_probe_name
+  }
+
+  frontend_endpoint {
+    name                              = local.endpoint_name
+    host_name                         = "acctest-FD-%d.azurefd.net"
```
WodansSon, line 157:
```suggestion
    name      = local.endpoint_name
    host_name = "acctest-FD-%d.azurefd.net"
```

azurerm/internal/services/frontdoor/tests/frontdoor_custom_https_configuration_resource_test.go
```diff
@@ -0,0 +1,161 @@
...
+func testAccAzureRMFrontDoorCustomHttpsConfiguration_CustomHttpsDisabled(data acceptance.TestData) string {
+	template := testAccAzureRMFrontDoorCustomHttpsConfiguration_template(data)
+	return fmt.Sprintf(`
+%s
+
+resource "azurerm_frontdoor_custom_https_configuration" "test" {
+ front_door_name                   = azurerm_frontdoor.test.name
+ frontend_endpoint_name            = azurerm_frontdoor.test.frontend_endpoint[0].name
+ resource_group_name               = azurerm_resource_group.test.name
+ custom_https_provisioning_enabled = false
```
WodansSon, line 94:
```suggestion
  front_door_name                   = azurerm_frontdoor.test.name
  frontend_endpoint_name            = azurerm_frontdoor.test.frontend_endpoint[0].name
  resource_group_name               = azurerm_resource_group.test.name
  custom_https_provisioning_enabled = false
```

azurerm/internal/services/frontdoor/frontdoor_custom_https_configuration_resource.go
```diff
@@ -0,0 +1,312 @@
...
+	customHttpsConfigurationNew := make([]interface{}, 0)
+	err = resourceArmFrontDoorFrontendEndpointCustomHttpsConfigurationUpdate(ctx, *resp.ID, false, frontDoorName, frontendEndpointName, resourceGroup, resp.CustomHTTPSProvisioningState, resp.CustomHTTPSConfiguration, customHttpsConfigurationNew, meta)
+	if err != nil {
+		return fmt.Errorf("unable to disable Custom HTTPS configuration for Frontend Endpoint %q (Resource Group %q): %+v", frontendEndpointName, resourceGroup, err)
+	}
+
+	return nil
+}
+
+func frontDoorCustomHttpsConfigurationReadParams(d *schema.ResourceData) (err error, resourceGroup string, frontDoorName string, frontendEndpointName string) {
```
WodansSon, line 188: As stated above:

```suggestion
func frontDoorCustomHttpsConfigurationReadProps(d *schema.ResourceData) (err error, resourceGroup string, frontDoorName string, frontendEndpointName string) {
```

azurerm/internal/services/frontdoor/frontdoor_custom_https_configuration_resource.go
```diff
@@ -0,0 +1,312 @@
...
+			customHTTPSConfiguration := customHttpsConfigurationNew[0].(map[string]interface{})
+			minTLSVersion := frontdoor.OneFullStopTwo // Default to TLS 1.2
+			if httpsConfig := customHTTPSConfigurationCurrent; httpsConfig != nil {
+				minTLSVersion = httpsConfig.MinimumTLSVersion
+			}
+			customHTTPSConfigurationUpdate := makeCustomHttpsConfiguration(customHTTPSConfiguration, minTLSVersion)
+			if provisioningState == frontdoor.CustomHTTPSProvisioningStateDisabled || customHTTPSConfigurationUpdate != *customHTTPSConfigurationCurrent {
+				// Enable Custom Domain HTTPS for the Frontend Endpoint
+				if err := resourceArmFrontDoorFrontendEndpointEnableHttpsProvisioning(ctx, true, frontDoorName, frontendEndpointName, resourceGroup, customHTTPSConfigurationUpdate, meta); err != nil {
+					return fmt.Errorf("Unable to enable/update Custom Domain HTTPS for Frontend Endpoint %q (Resource Group %q): %+v", frontendEndpointName, resourceGroup, err)
```
WodansSon, line 233: Same here with casing...

```suggestion
					return fmt.Errorf("unable to enable/update Custom Domain HTTPS for Frontend Endpoint %q (Resource Group %q): %+v", frontendEndpointName, resourceGroup, err)
```

azurerm/internal/services/frontdoor/frontdoor_custom_https_configuration_resource.go
```diff
@@ -0,0 +1,312 @@
...
+				}
+			}
+		} else if !customHttpsProvisioningEnabled && provisioningState == frontdoor.CustomHTTPSProvisioningStateEnabled {
+			// Disable Custom Domain HTTPS for the Frontend Endpoint
+			if err := resourceArmFrontDoorFrontendEndpointEnableHttpsProvisioning(ctx, false, frontDoorName, frontendEndpointName, resourceGroup, frontdoor.CustomHTTPSConfiguration{}, meta); err != nil {
+				return fmt.Errorf("Unable to disable Custom Domain HTTPS for Frontend Endpoint %q (Resource Group %q): %+v", frontendEndpointName, resourceGroup, err)
```
WodansSon, line 239:
```suggestion
				return fmt.Errorf("unable to disable Custom Domain HTTPS for Frontend Endpoint %q (Resource Group %q): %+v", frontendEndpointName, resourceGroup, err)
```

azurerm/internal/services/frontdoor/frontdoor_resource.go
```diff
@@ -601,68 +571,17 @@ func resourceArmFrontDoorCreateUpdate(d *schema.ResourceData, meta interface{})
...
-				} else if !customHttpsProvisioningEnabled && provisioningState == frontdoor.CustomHTTPSProvisioningStateEnabled {
-					// Disable Custom Domain HTTPS for the Frontend Endpoint
-					if err := resourceArmFrontDoorFrontendEndpointEnableHttpsProvisioning(ctx, false, name, frontendEndpointName, resourceGroup, frontdoor.CustomHTTPSConfiguration{}, meta); err != nil {
-						return fmt.Errorf("Unable to disable Custom Domain HTTPS for Frontend Endpoint %q (Resource Group %q): %+v", frontendEndpointName, resourceGroup, err)
-					}
-				}
+			customHttpsConfigurationNew := frontendEndpoint["custom_https_configuration"].([]interface{})
+			err := resourceArmFrontDoorFrontendEndpointCustomHttpsConfigurationUpdate(ctx, *resp.ID, customHttpsProvisioningEnabled, name, frontendEndpointName, resourceGroup, properties.CustomHTTPSProvisioningState, properties.CustomHTTPSConfiguration, customHttpsConfigurationNew, meta)
+			if err != nil {
+				return fmt.Errorf("Unable to update Custom HTTPS configuration for Frontend Endpoint %q (Resource Group %q): %+v", frontendEndpointName, resourceGroup, err)
```
WodansSon, line 577:
```suggestion
				return fmt.Errorf("unable to update Custom HTTPS configuration for Frontend Endpoint %q (Resource Group %q): %+v", frontendEndpointName, resourceGroup, err)
```

azurerm/internal/services/frontdoor/frontdoor_resource.go
```diff
@@ -1268,87 +1187,72 @@ func flattenArmFrontDoorBackend(input *[]frontdoor.Backend) []interface{} {
...
-						if kvcsp := customHTTPSConfiguration.KeyVaultCertificateSourceParameters; kvcsp != nil {
-							chc["certificate_source"] = string(frontdoor.CertificateSourceAzureKeyVault)
-							chc["azure_key_vault_certificate_vault_id"] = *kvcsp.Vault.ID
-							chc["azure_key_vault_certificate_secret_name"] = *kvcsp.SecretName
-							chc["azure_key_vault_certificate_secret_version"] = *kvcsp.SecretVersion
-						}
-					} else {
-						chc["certificate_source"] = string(frontdoor.CertificateSourceFrontDoor)
-					}
+		if properties := resp.FrontendEndpointProperties; properties != nil {
```
WodansSon, line 1233: Can we use `props` instead of `properties`?

```suggestion
		if props := resp.FrontendEndpointProperties; properties != nil {
```

website/docs/r/frontdoor.html.markdown
```diff
@@ -246,6 +246,8 @@ The following attributes are only valid if `certificate_source` is set to `Azure
 
 ~> **Note:** In order to enable the use of your own custom `HTTPS certificate` you must grant `Azure Front Door Service` access to your key vault. For instuctions on how to configure your `Key Vault` correctly please refer to the [product documentation](https://docs.microsoft.com/en-us/azure/frontdoor/front-door-custom-domain-https#option-2-use-your-own-certificate).
 
+-> **NOTE:** Custom https configurations for a Front Door Frontend Endpoint can be defined both within the `azurerm_frontdoor` resource and by using a separate [`azurerm_frontdoor_custom_https_configuration` resource](frontdoor_custom_https_configuration.html). Defining custom https configurations using a separate resource allows for parallel creation/update.
```
WodansSon, line 249:
```suggestion
-> **NOTE:** Custom https configurations for a Front Door Frontend Endpoint can be defined both within the `azurerm_frontdoor` resource or by using a separate [`azurerm_frontdoor_custom_https_configuration` resource](frontdoor_custom_https_configuration.html). Defining custom https configurations using a separate resource allows for parallel creation/update.
```

timja:
> I tried fixing it for you but I was denied permission to push to your fork.

https://github.community/t/how-can-we-enable-allow-edits-from-maintainers-by-default/2847/7
This was submitted from our org account, which apparently doesn't support the option to allow maintainers to push

We'll keep that in mind in future, for now if you have any follow up changes you want to do yourself feel free to send us a PR and we'll merge it in

azurerm/internal/services/frontdoor/frontdoor_custom_https_configuration_resource.go
```diff
@@ -0,0 +1,312 @@
...
+		Update: resourceArmFrontDoorCustomHttpsConfigurationCreateUpdate,
+		Delete: resourceArmFrontDoorCustomHttpsConfigurationDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(6 * time.Hour),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
```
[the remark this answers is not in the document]
luigibk: This is consistent with the main frontdoor resource:

https://github.com/terraform-providers/terraform-provider-azurerm/commit/a08a2d8b403504f866636f3991d69264b0312764#diff-f40eb1b0d4646b0fed0eab1d3837c42d

I had accepted your change but then I reverted it as I think they should be consistent.

azurerm/internal/services/frontdoor/frontdoor_custom_https_configuration_resource.go
```diff
@@ -0,0 +1,312 @@
...
+			Update: schema.DefaultTimeout(6 * time.Hour),
+			Delete: schema.DefaultTimeout(6 * time.Hour),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"front_door_name": {
```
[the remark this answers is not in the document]
luigibk: I think the best solution is to actually add a `frontend_endpoint_id` and remove both `front_door_name` and `frontend_endpoint_name` as they can be derived from the id. That way it should be easier to manage from a user perspective and the implementation should not be more complex after removing those 2 fields. How does such a solution sound to you?

luigibk:
@WodansSon thanks for your review and comments. I have merged all of them but one and left a comment about the suggestion of adding a front_door_id.

azurerm/internal/services/frontdoor/frontdoor_custom_https_configuration_resource.go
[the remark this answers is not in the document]
WodansSon: That's a good idea, but because of the way Terraform works the core would not be able to build a proper dependency graph during the plan phase, if I am not mistaken, which is why we generally use the parent resources id attribute. You can give it a try and see if it builds a correct provisioning graph, but my gut is telling me it won't be able to connect the two together.

azurerm/internal/services/frontdoor/frontdoor_custom_https_configuration_resource.go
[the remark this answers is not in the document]
WodansSon: Something is amiss, here cause I change front doors to 5 hours last release... can you base master branch be out of sync with the current master branch? 🤔

azurerm/internal/services/frontdoor/frontdoor_custom_https_configuration_resource.go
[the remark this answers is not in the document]
luigibk: In this case the parent resource for the custom https configuration is the frontend endpoint. I've started work on this and it looks good so far. Hopefully won't be long before I commit it.

luigibk:
About the read timeout, it's currently set at 5minutes in frontdoor:

https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/frontdoor/frontdoor_resource.go#L37

and changing it to 5 hours breaks this test:

https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/provider/provider_test.go#L77

so I have left those timeouts at 5 minutes for now

luigibk:
@WodansSon thanks for helping with this. I have just pushed the change related to replacing frontdoor_name and frontend_endpoint_name with the frontend_endpoint_id.

WodansSon:
> About the read timeout, it's currently set at 5minutes in frontdoor:
> 
> https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/frontdoor/frontdoor_resource.go#L37
> 
> and changing it to 5 hours breaks this test:
> 
> https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/provider/provider_test.go#L77
> 
> so I have left those timeouts at 5 minutes for now

It looks like Tom over wrote my change and reverted the value back to 5 minutes, so this is fine. Can we make sure the documentation also reflects this change?

WodansSon:
image

Issue #5730, opened 2020-02-13
Resource "azurerm_kubernetes_cluster" is leaking sensitive data


subesokun:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform: `0.12.20`
AzureRM: `1.41.0`

### Affected Resource(s)

* `azurerm_kubernetes_cluster`

### Debug Output

```
~ resource "azurerm_kubernetes_cluster" "cluster" {
        ...redacted...
        kube_admin_config               = []
        kube_config                     = [
            {
                client_certificate     = "redacted"
                client_key             = "redacted"
                cluster_ca_certificate = "redacted"
                host                   = "redacted"
                password               = "redacted"
                username               = "redacted"
            },
        ]
        kube_config_raw                 = (sensitive value)
        kubernetes_version              = "1.15.7"
        ...redacted...
        service_principal {
            client_id     = "redacted"
            client_secret = (sensitive value)
        }
    }
```

### Expected Behavior

In case of in-place updates, sensitive data like the `kubecfg.client_key` gets redacted and is not visible in the logs.

### Actual Behavior

In case of in-place updates, sensitive data like the `kubecfg.client_key` gets printed in clear text into the logs.

### Steps to Reproduce

Change a property that causes an in-place update of the AKS cluster and run a `terraform plan`

tombuildsstuff:
👋 

Thanks for opening this issue.

This issue needs to be fixed in the Terraform Plugin SDK, as such rather than tracking this issue in multiple places I'm going to close this issue in favour of the upstream issue. Once that's been fixed we'll update the version of the Plugin SDK being used and this should get resolved - as such [please subscribe to the upstream issue for updates](https://github.com/hashicorp/terraform-plugin-sdk/issues/201).

Thanks!

Issue #5083, opened 2019-12-05
Azurem prints secrets during plan phase in plain text..


hajdukd:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

version = "=1.36.1"

### Affected Resource(s)

azurerm_kubernetes_cluster

### Terraform Configuration Files

N/A, any configuration can be an example.

### Debug Output

N/A

### Panic Output

N/A

### Expected Behavior

Every sensitive data will be hidden during plan step. Things like:

- client_certificate
- client_key
- cluster_ca_certificate
- password
- agent_pool_profile.linux_profile.ssh_key.key_data

It works correctly for "kube_config_raw" ( listed as (sensitive value) ).

### Actual Behavior

All sensitive data is printed in plain text.

### Steps to Reproduce

terraform plan

### Important Factoids

N/A

### References

N/A

brennerm:
@hajdukd This issue comes from a [bug in the Terraform Plugin SDK](https://github.com/hashicorp/terraform-plugin-sdk/issues/201). For this reason it is currently not possible to mask parts of nested blocks. One option would be to mark the whole `kube_config` block as sensitive. But to me this sounds like a workaround for the actual problem.

Additionally attributes like `client_certificate`, `cluster_ca_certificate` and `agent_pool_profile.linux_profile.ssh_key.key_data` do not need to be sensitive imo. These are public certificates that shouldn't lead to any damage when being exposed.

ayusmadi:
Quick workaround is to omit kube_config sensitive information using `egrep`:

```
terraform plan | egrep -v 'client_key|password|client_certificate|cluster_ca_certificate'
```

lodejard:
> @hajdukd One option would be to mark the whole `kube_config` block as sensitive. But to me this sounds like a workaround for the actual problem.
> 
> Additionally attributes like `client_certificate`, `cluster_ca_certificate` and `agent_pool_profile.linux_profile.ssh_key.key_data` do not need to be sensitive imo. These are public certificates that shouldn't lead to any damage when being exposed.

[...]

Issue #4621, opened 2019-10-15
[1.35.0] AKS sensitive information in plan output


EPinci:
* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
Running terraform 0.12.10 with AzureRM provider v1.35.0.

### Affected Resource(s)

* `azurerm_kubernetes_cluster`

### Expected Behavior

When running a plan on an AKS cluster that requires an in-place update (such as a kubernetes version upgrade) the plan output shows the following excerpt:

```
        kube_admin_config               = [
            {
                client_certificate     = "[ACTUAL_CERTIFICATE_IS_SHOWN]"
                client_key             = "[ACTUAL_KEY_IS SHOWN]"
                cluster_ca_certificate = "[ACTUAL_CA_CERTIFICATE_IS_SHOWN]"
                host                   = "https://myhost-aaaaaaaa.hcp.westeurope.azmk8s.io:443"
                password               = "[ACTUAL_PASSWORD_IS_SHOWN]"
                username               = "clusterAdmin_myCluster"
            },
        ]
        kube_admin_config_raw           = (sensitive value)
        kube_config                     = [
            {
                client_certificate     = ""
                client_key             = ""
                cluster_ca_certificate = "[ACTUAL_CA_CERTIFICATE_IS_SHOWN]"
                host                   = "https://mycluster-aaaaaaaa.hcp.westeurope.azmk8s.io:443"
                password               = ""
                username               = "clusterUser_myCluster"
            },
        ]
        kube_config_raw                 = (sensitive value)
```

Note that the raw version of the data is considered sensitive and redacted while the parsed data, including keys and password to gain access to the cluster, are actually shown in clear while they should be considered sensitive and redacted as well.

At least the following fields should be marked sensitive:
- kube_admin_config.client_key
- kube_admin_config.password
- kube_config.client_key
- kube_config.password

Also, please evaluata if the following fields can be be somehow hidden from screen dump given long base64 encoded string don't necessarily provide a valuable output for the plan review.
- kube_admin_config.client_certificate
- kube_admin_config.cluster_ca_certificate
[...]

PR #7505, opened 2020-06-25
Hiding sensitive kube_* values

pwiens:
This change will prevent leaking secrets during plan phase. This issue keeps getting punted between terraform core and this provider. This change will hide the entire block which seems safer to me than leaking secrets in deployment logs.

#7238
#5730
#5083
#4621

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_cluster_resource.go
+++ b/azurerm/internal/services/containers/kubernetes_cluster_resource.go
@@ -552,8 +552,9 @@ func resourceArmKubernetesCluster() *schema.Resource {
 			},
 
 			"kube_admin_config": {
-				Type:     schema.TypeList,
-				Computed: true,
+				Type:      schema.TypeList,
+				Computed:  true,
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hi @pwiens 

Thanks for opening this issue.

As mentioned in the other issues - unfortunately this needs to be fixed in the Terraform Plugin SDK, [which is being tracked in this issue](https://github.com/hashicorp/terraform-plugin-sdk/issues/201). Whilst we could look to make the entire block sensitive - some fields within that are not sensitive (e.g. hostname/username) - and so should be shown in the output.

Whilst I'd like to thank you for this contribution - I'm going to close this [in favour of the upstream issue in the Terraform Plugin SDK](https://github.com/hashicorp/terraform-plugin-sdk/issues/201) (which has since been split out from Terraform Core, which is why this is linked in different places) which is tracking this - where once this is fixed this should then be available within the Azure Provider.

Thanks!

PR #7506, opened 2020-06-26
azurerm_template_deployment: Support for subscription deployments

### LINKED ISSUE #6561: Support for subscription scoped deployments in azurerm_template_deployment
### Community Note

### Description

Right now azurerm_template_deployment requires a resource group name.
Arm templates are not necessary deployed at the resource group level, but also at the subscription level:

    budgets
    deployments - for nested templates that deploy to resource groups.
    eventSubscriptions
    peerAsns
    policyAssignments
    policyDefinitions
    policySetDefinitions
    remediations
    resourceGroups
    roleAssignments
    roleDefinitions
    scopeAssignments
    supportPlanTypes
    tags

 Support for this should be added.

* azurerm_template_deployment

### References

* https://docs.microsoft.com/nb-no/azure/azure-resource-manager/templates/deploy-to-subscription?view=azure-cli-latest#supported-resources

tiwood:
This introduces the ability to create subscription ARM deployments.

Changes to the resource:
- `resource_group_name` is now optional. If not provided a subscription deployment is created.
- `location` is a new optional attribute (required for subscription deployments)

This closes #6561

## Example

```hcl
resource "azurerm_template_deployment" "test" {

  // Note that we did not specify 'resource_group_name'
  name                = "testRG"
  location            = "westeurope"

  template_body = <<DEPLOY
  {
    "$schema": "https://schema.management.azure.com/schemas/2018-05-01/subscriptionDeploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
      "resourceGroupName": {
        "type": "string",
        "metadata": {
          "description": "The name of the Resource Group"
        }
      }
    },
    "variables": {},
    "resources": [
      {
        "type": "Microsoft.Resources/resourceGroups",
        "apiVersion": "2019-10-01",
        "name": "[parameters('resourceGroupName')]",
        "location": "westeurope",
        "properties": {}
      }
    ],
    "outputs": {}
  }
DEPLOY

  deployment_mode = "Incremental"

  parameters = {
    "resourceGroupName" = "myTestRG"
  }
}
```


tombuildsstuff:
hey @tiwood 

We've been spending some time thinking about how we currently handle these "generic" resources (which are where the API provisions multiple things [e.g. HDInsight or App Service which provisions different "kinds" of Clusters/Hosted Apps] or supports the same resource at multiple scopes [e.g. Role Assignment, ARM Template Deployments]).

Whilst historically we've tried to ship a single resource to cover these - over time we've found that this ends up causing more issues where these API's diverge in behaviour over time. For example, the recent changes to the Role Assignment resource where we now have to include the Scope in the ID, since this isn't returned from the API when provisioning at a Management Group level.

Instead we're taking the approach to have multiple resources which accurately map these behaviours going forward (so for example, the new Linux & Windows VM resources instead of the combined resource). Whilst there's some downsides to this (implementation time) - there's significant benefits too since we're able to accurately map the behaviour of the API - and so we're taking this approach for new resources (as we did with HDInsight) and gradually updating older resources (such as VM/VMSS, App Service and Template Deployments) to use this new model as time goes on (indeed App Service will likely get reworked as a part of 3.0).

Specifically for the ARM Template Deployment resource is how the API behaviour isn't currently mapped correctly - insofar as Terraform Core (at this time) only supports outputting a Dictionary/Map of a single type (e.g. a Dictionary of Strings or a Dictionary of Ints) - and as such we're unable to output some of the more complex output types. The API also now exposes the resources which have been provisioned by the template - so it's possible for us to clean these up if required during deletion (however that'd be a breaking behavioural change at this time).

In keeping with this aim and in order to be able to correctly map these behaviours, I've spent some time splitting this resource into two new resources which'll ultimately enable this: `azurerm_resource_group_template_deployment` (which'll supersede the existing resource) and a `azurerm_subscription_template_deployment` resource. We'll also look to add a Management Group and Tenant Template Deployment resource in the future too - since these also have different behaviours (for example, 'DeploymentMode' has to be set to 'Incremental' at a Subscription level, and presumably also for Management Groups/Tenants).

As such whilst on the whole this PR looks good - since we're ultimately looking to take a different direction here I'm going to close this PR in favour of #8672, which supports Template Deployments at a Subscription level via a separate resource (and brings these enhancements to the Resource Group resource too).

Thanks!

Issue #7406, opened 2020-06-19
Support for [Azure Synapse Analytics]


njuCZ:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Azure Synapse is an integrated analytics service that accelerates time to insight from all data at any scale, across data warehouses and big data analytics systems. It brings together the best of the SQL technologies used in enterprise data warehousing, Spark technologies used in big data analytics, and Pipelines to orchestrate activities and data movement.

### New or Affected Resource(s)

* azurerm_synapse_workspace
* azurerm_synapse_big_data_pool
* azurerm_synapse_sql_pool

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

* #0000

njuCZ:
I am working on the first resource `azurerm_synapse_workspace`

PR #7517, opened 2020-06-29
new resource and data source for `azurerm_synapse_workspace`

njuCZ:
first resource of #7406

image


azurerm/internal/services/synapse/synapse_workspace_resource.go
```diff
@@ -0,0 +1,349 @@
...
+	id, err := parse.SynapseWorkspaceID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	if _, err := client.Delete(ctx, id.ResourceGroup, id.Name); err != nil {
+		return fmt.Errorf("deleting Synapse Workspace %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
+	}
+	// sometimes the waitforcompletion rest api will return 404, so changed with polling
+	return waitForSynapseWorkspaceToBeDeleted(ctx, client, id.ResourceGroup, id.Name, d)
```
WodansSon, line 268: Can we remove the `waitForSynapseWorkspaceToBeDeleted` function and use the SDKs `WaitForCompletionRef`? Just test the response of the for expected values? Maybe something like this:

```
if err = future.WaitForCompletionRef(ctx, client); err != nil {
	if !response.WasNotFound(future.Response()) {
		return fmt.Errorf("waiting for Synapse Workspace %q (Resource Group %q) to be deleted: %+v", name, resourceGroup, err)
	}
}
```

azurerm/internal/services/synapse/synapse_workspace_resource.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+func waitForSynapseWorkspaceToBeDeleted(ctx context.Context, client *synapse.WorkspacesClient, resourceGroup, name string, d *schema.ResourceData) error {
```
WodansSon, line 271: As stated above, can remove this function and use the standard SDK fuctions?

azurerm/internal/services/synapse/validate/synapse_workspace.go
```diff
@@ -0,0 +1,51 @@
...
+		return
+	}
+
+	// The name attribute rules are :
+	// 1. can contain only lowercase letters or numbers.
+	// 2. must not end with the string 'ondemand'
+	// 3. The value must be between 1 and 45 characters long
+
+	if !regexp.MustCompile(`^[a-z\d]{1,45}$`).MatchString(v) {
+		errors = append(errors, fmt.Errorf("%s can contain only lowercase letters or numbers, must not end with the string 'ondemand', The value must be between 1 and 45 characters long", k))
```
WodansSon, line 22:
```suggestion
		errors = append(errors, fmt.Errorf("%s can contain only lowercase letters or numbers, must not end with the string 'ondemand', and be between 1 and 45 characters long", k))
```

azurerm/internal/services/synapse/validate/synapse_workspace.go
```diff
@@ -0,0 +1,51 @@
+package validate
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+func SynapseWorkspaceName(i interface{}, k string) (warnings []string, errors []error) {
```
WodansSon, line 9: Maybe we should just return a single error instead of slice of errors? Instead of appending the errors maybe expose a boolean like `hasErrors` and then at the end of the function test to see if the bool is true or not, if `true` return the single error string? The reason I am suggesting this is because the first and second error messages have duplicate information about the `ondemand` suffix.

website/allowed-subcategories
```diff
@@ -59,5 +59,6 @@ Service Fabric
 Spring Cloud
 Storage
 Stream Analytics
+Synapse
 Template
 Time Series Insights
```
WodansSon, line 64: Can we add a newline here?

website/docs/r/synapse_workspace.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+subcategory: "Synapse"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_synapse_workspace"
+description: |-
+  Manages a synapse Workspace.
+---
+
+# azurerm_synapse_workspace
+
+Manages a synapse Workspace.
```
WodansSon, line 11:
```suggestion
Manages a synapse workspace.
```

website/docs/r/synapse_workspace.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+  name               = "example"
+  storage_account_id = azurerm_storage_account.example.id
+}
+
+resource "azurerm_synapse_workspace" "example" {
+  name                                 = "example"
+  resource_group_name                  = azurerm_resource_group.example.name
+  location                             = azurerm_resource_group.example.location
+  storage_data_lake_gen2_filesystem_id = azurerm_storage_data_lake_gen2_filesystem.example.id
+  sql_administrator_login              = "sqladminuser"
```
WodansSon, line 41: Can we add the sql admin password to the sample code?

website/docs/r/synapse_workspace.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+
+  identity {
```
WodansSon, line 43: Should we remove the `identity` block from the sample code since it should be pure computed block per comments above?

website/docs/r/synapse_workspace.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+
+* `location` - (Required) Specifies the Azure Region where the synapse Workspace should exist. Changing this forces a new resource to be created.
+
+* `storage_data_lake_gen2_filesystem_id` - (Required) Specifies the ID of storage data lake gen2 filesystem resource. Changing this forces a new resource to be created.
+
+* `identity` - (Required) An `identity` block as defined below.
+
+* `sql_administrator_login` - (Required) Specifies The Login Name of the SQL administrator. Changing this forces a new resource to be created.
+
+* `sql_administrator_login_password` - (Optional) The Password associated with the `sql_administrator_login` for the SQL administrator.
```
WodansSon, line 65:
```suggestion
* `sql_administrator_login_password` - (Required) The Password associated with the `sql_administrator_login` for the SQL administrator.
```

website/docs/r/synapse_workspace.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+
+* `managed_virtual_network_enabled` - (Optional) Is Virtual Network enabled for all computes in this workspace. Changing this forces a new resource to be created.
+
+* `tags` - (Optional) A mapping of tags which should be assigned to the Synapse Workspace.
+
+---
+
+An `identity` block supports the following:
+
+* `type` - (Required) Specifies the identity type of the Synapse Workspace. At this time the allowed value is `SystemAssigned`.
```
WodansSon, line 75: Shouldn't we remove `type` from this block and add it to the `identity` export block?

```suggestion

```

website/docs/r/synapse_workspace.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+* `connectivity_endpoints` - A list of Connectivity endpoints for this Synapse Workspace.
+
+* `managed_resource_group_name` - Workspace managed resource group.
+
+* `identity` - An `identity` block as defined below, which contains the Managed Service Identity information for this Synapse Workspace.
+
+---
+
+The `identity` block exports the following:
+
```
WodansSon, line 92: Can we add `type` as an exported value here?

azurerm/internal/services/synapse/validate/synapse_workspace.go
njuCZ: I have resolved it by removing the duplicate information in the first error message. Because the regex could not detect it

website/allowed-subcategories
njuCZ: this file should be auto-generated ?

njuCZ:
@WodansSon thanks for your suggestions, I have modifed this PR, could you please have a look again?

jackofallops:
Tests pass:
image

PR #7524, opened 2020-06-29
Update `azurerm_linux|windows_virtual_machine_scale_set` - Add support for specialized custom image in vmss

### LINKED ISSUE #7772: azurerm_windows_virtual_machine fails to create when referencing a gallery image.
### Community Note

```
Terraform v0.12.28
+ provider.azuread v0.10.0
+ provider.azurerm v2.18.0
```

### Affected Resource(s)

```hcl
resource "azurerm_windows_virtual_machine" "pool" {
  for_each = {
    for key, i in local.vm_config :
    key => i
  }
  
  name                  = each.value.vm_name
  resource_group_name   = var.resource_group_name
  location              = var.location
  size                  = each.value.vm_size
  admin_username        = each.value.admin_username
  admin_password        = each.value.admin_password
  network_interface_ids = each.value.network_interface_ids
  license_type          = each.value.license_type
  timezone              = each.value.timezone
  
  source_image_id       = each.value.source_image_id
  
  tags                  = merge(var.tags, { purpose = each.value.purpose }, { node = each.value.node_key + 1 })
  
  os_disk {
    caching              = each.value.os_disk_caching
    storage_account_type = each.value.os_disk_storage_account_type
  }
```

### Expected Behavior

Deploy two Windows VMs from "Shared Gallery Image"

### Actual Behavior

A spurious error that does not relate to the issue.
```
module.compute_region_0.module.windows_virtual_machine.azurerm_windows_virtual_machine.pool["1"]: Creating...
Error: creating Windows Virtual Machine "vm2" (Resource Group "rgvmimagedemo"): compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="Parameter 'osProfile' is not allowed." Target="osProfile"

  on ..\azurerm_windows_virtual_machine\main.tf line 82, in resource "azurerm_windows_virtual_machine" "pool":
  82: resource "azurerm_windows_virtual_machine" "pool" {
```
(Line 82 is the being of the resource: `resource "azurerm_windows_virtual_mahine" "pool" {`)

If this isn't a bug, please consider updating the document that would resolve my mistake. Perhaps an example of using `source_image_id`.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

This may be related to #5998 although the error is different.

Plan succeeds.

I did a test by creating the old `os_profile{}` block, adding `admin_username` and `admin_password` which made the plan fail.

If this is the same regression as mentioned in #5998, can you check `azurerm_linux_virtual_machine` too, please?

If I comment out `source_image_id` and use `source_image_reference` instead, `apply` works and the VMs deploy.

T. I. A.

### References

There is an unanswered related question on your discuss site: https://discuss.hashicorp.com/t/unable-to-create-azure-windows-vm/6672

* #5998

ArcturusZhang:
Fixes #7772


azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
```diff
@@ -869,14 +898,20 @@ func resourceArmLinuxVirtualMachineScaleSetRead(d *schema.ResourceData, meta int
 			d.Set("source_image_id", storageImageId)
 		}
 
+		disablePassword := true  // default value of `disable_password_authentication` to avoid unexpected diff
+		provisionVMAgent := true // default value of `provision_vm_agent` to avoid unexpected diff
```
tombuildsstuff, line 902: defaulting this to `true` is an unsafe default - we should default these to `false` unless we can show otherwise - the VM API should be exposing this information either way, if it's not, that's a bug in the Compute API

azurerm/internal/services/compute/windows_virtual_machine_scale_set_resource.go
```diff
@@ -918,6 +955,8 @@ func resourceArmWindowsVirtualMachineScaleSetRead(d *schema.ResourceData, meta i
 			d.Set("source_image_id", storageImageId)
 		}
 
+		enableAutomaticUpdates := true // default value of `enable_automatic_updates` to avoid unexpected diff
+		provisionVMAgent := true // default value of `provision_vm_agent` to avoid unexpected diff
```
tombuildsstuff, line 959: (as above) we cannot assume these are enabled by default - there's false-positives here - we need to default these to false and parse them out of the API

website/docs/r/linux_virtual_machine_scale_set.html.markdown
```diff
@@ -178,6 +176,8 @@ The following arguments are supported:
 
 -> **NOTE:** One of either `source_image_id` or `source_image_reference` must be set.
 
+~> **NOTE:** If `source_image_id` is set to a shared image with `os_state` of `Specialized`, or a shared image version created by a specialized shared image, `admin_username`, `admin_password`, `custom_data`, `admin_ssh_key` and `secret` must not be specified.
```
tombuildsstuff, line 179: we should add a disclaimer about Specialized Images here, since reusing the machine UUID can cause data loss in a clustered scenario:

```suggestion
~> **NOTE:** If `source_image_id` is set to a shared image with `os_state` of `Specialized`, or a shared image version created by a specialized shared image, `admin_username`, `admin_password`, `custom_data`, `admin_ssh_key` and `secret` must not be specified.

!> **Note:** Specialized Images reuse the UUID of the Virtual Machine which can have unexpected consequences, including data corruption (for example, when running in a cluster and two nodes share the same ID) - as such we recommend using a Generalized image where possible.
```

this'll also want to go up the top in the "disclaimers" section

website/docs/r/windows_virtual_machine_scale_set.html.markdown
```diff
@@ -166,6 +164,8 @@ The following arguments are supported:
 
 -> **NOTE:** One of either `source_image_id` or `source_image_reference` must be set.
 
+~> **NOTE:** If `source_image_id` is set to a shared image with `os_state` of `Specialized`, or a shared image version created by a specialized shared image, `admin_username`, `admin_password`, `custom_data`, `additional_unattend_content`, `secret` and `timezone` must not be specified.
```
tombuildsstuff, line 167: (as above)

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
```diff
@@ -377,9 +366,40 @@ func resourceArmLinuxVirtualMachineScaleSetCreate(d *schema.ResourceData, meta i
...
+			_, errs := ValidateLinuxComputerNamePrefix(d.Get("name"), "computer_name_prefix")
+			if len(errs) > 0 {
+				return fmt.Errorf("unable to assume default computer name prefix %s. Please adjust the %q, or specify an explicit %q", errs[0], "name", "computer_name_prefix")
+			}
+			computerNamePrefix = name
+		}
+
+		adminUsername := d.Get("admin_username").(string)
+		if adminUsername == "" {
+			return fmt.Errorf("`admin_username` is required when using platform image, image or generalized shared image")
```
tombuildsstuff, line 398:
```suggestion
			return fmt.Errorf("`admin_username` is required when not using a specialized image")
```

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
ArcturusZhang: In the schema definition, these two attributes are optional and have default value of `true`, and the biggest reason I have set these default values here is that in the SDK, these two attributes are in the `OsProfile` block, which is `nil` in the response when using specialized image (and `OsProfile` has to be nil in the request).
And we can take two different scenarios to view this:
1. The user is not using the specialized image (the use case before this PR). If they are not specifying these two attributes, we will send the default value of these two (which is `true` according to the schema definition). And if the service is not making anything wrong, it should return the same values as the input. Therefore this should be consistency.
2. The user is using the specialized image (the use case that is introducing in this PR). In this case, this indeed is problematic... I am not quite so sure how we could elegantly deal with this. Because these two attributes come from the `OsProfile` block, these two attributes are forbidden to set when using a specialized image. And these two attributes are optional with default values of `true` therefore absence of these two is the same as explicitly assigning those to `true`. And because of the same reason, I just skipped the checking of these two in L420-436. This would look so weird if I check these two are absent (I have to check these two are set to `true`). Similarly, if the user does not set these two in the config, they will have the value of their default (aka `true`). In the response, these two values will not be returned, if we regard this as `false`, in this case we will never get an empty plan after apply.

Do you have any recommendation on this situation?

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
ArcturusZhang: Done

website/docs/r/linux_virtual_machine_scale_set.html.markdown
ArcturusZhang: Done and disclaimers added

website/docs/r/windows_virtual_machine_scale_set.html.markdown
ArcturusZhang: Done and disclaimers added

ArcturusZhang:
Hi @tombuildsstuff I have resolved some of the comments. As for the default values of `true` I have left some explanations, please have a look, thank

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
tombuildsstuff: So, whilst the Compute API can opt to not require this information at Provisioning time - this information should be available from the image - so it should be able to be returned in the Get response once the machine has been provisioned?

From our side, assuming those fields have a default value has security implications - and as such we can't assume they're set to `true`. For example, if a user is using tooling (such as Sentinel or OpenPolicyAgent) to confirm that these flags are set to `true` - we'd assume this value is set to `true` even when it may not be.

As such, I'd suggest reaching out to the Compute Team here since this feels like a bug in the Compute API?

ArcturusZhang: Emmm... I am not fully getting the point. But in this case, if a VM is provisioned using a specialized image, the OSProfile block must be nil, and when you get it, you will also get a nil OSProfile block. These two parameters are in the OSProfile.

I am making these two `true` because when you apply a VM provisioned by a specialized image, the service will not return a valid value for OSProfile and therefore no valid values for `disablePassword` and `provisionVMAgent`. When absence, terraform is assuming those as `false`. And now when you run `terraform plan`, you will get diff (since their default values are true)

Therefore rather than a bug in compute API, I suppose this should be a shortage of terraform that we cannot assign `null` to a bool attribute. And I am considering this as a workaround for the unnecessary diff.

tombuildsstuff: > But in this case, if a VM is provisioned using a specialized image, the OSProfile block must be nil, and when you get it, you will also get a nil OSProfile block. These two parameters are in the OSProfile.

My point here being even if the API doesn't require sending that information during creation - but this information should be returned from the API during the read, that's a bug in the Azure API.

Unfortunately assuming that these two fields are set to `true` is misleading and could lead to a subtle security issue where users believe this functionality is enabled (for example, from the image) but it's not.

Whilst not setting this value in Terraform will make the field `null` - the Azure API itself needs to return this information so that we can confirm to users this security functionality is enabled, else there's a potential security issue here.

jackofallops: I'm in agreement with @tombuildsstuff - assumptions on a security posture must be avoided. For context, the issue isn't just procedural interaction with the API, it's the real-world implication of a key security setting.  I've worked in environments where this kind of potential exposure could result in severe repercussions; both in business reputation/finance as well as potentially legal in extreme cases.

tombuildsstuff:
@ArcturusZhang since [the API doesn't return the data we need here](https://github.com/Azure/azure-rest-api-specs/issues/10166#issuecomment-670134108) unfortunately we're blocked on this one - as such I'm going to close this PR for the moment until there's a means of retrieving this information.

PR #7277, opened 2020-06-10
`azurerm_shared_image` & `azurerm_shared_image_version` - support for specialized images by setting generalized to false

ArcturusZhang:
This PR adds support for the specialized shared image, with the following changes:

* `azurerm_shared_image` - add `os_state` attribute to indicate this image definition is generalized or specialized.
* `azurerm_shared_image_version` - add `os_disk_snapshot_id` attribute to support create new image version using snapshot, which also enables you create new version for a specialized image.

PR #7525, opened 2020-06-30
Update `azurerm_shared_image` - flip `generalized` to `specialized`

ArcturusZhang:
Since the changes on `azurerm_shared_image` has not been released yet, this should not be a breaking change. I suppose it could make more sense if we name this new attribute `specialized` to emphasize when we are assigning a non-zero value to it, the image is specialized. Rather than previous behaviour that we assign `generalized` to `false` to indicate this is specialized.

### DIFF
--- a/website/docs/d/shared_image.html.markdown
+++ b/website/docs/d/shared_image.html.markdown
@@ -43,7 +43,7 @@ The following attributes are exported:
 
 * `location` - The supported Azure location where the Shared Image Gallery exists.
 
-* `generalized` - Is the Operating System present in this Shared Image generalized or not. `false` indicates this image is specialized which for windows images means they have been sysprepped.
+* `specialized` - Is the Operating System present in this Shared Image specialized or not. `true` indicates this image is specialized which for windows images means they have been sysprepped.
 
 * `identifier` - An `identifier` block as defined below.
 
--- a/website/docs/r/shared_image.html.markdown
+++ b/website/docs/r/shared_image.html.markdown
@@ -68,7 +68,7 @@ The following arguments are supported:
 
 * `eula` - (Optional) The End User Licence Agreement for the Shared Image.
 
-* `generalized` - (Optional) Should the Operating System present in this Shared Image be generalized? Defaults to `true`. Setting this to `false` creates a specialized imagine which is the equivalent of sysprepping in windows. Changing this forces a new resource to be created.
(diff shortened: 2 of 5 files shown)


katbyte:
related to #7277

website/docs/r/shared_image.html.markdown
```diff
@@ -68,7 +68,7 @@ The following arguments are supported:
 
 * `eula` - (Optional) The End User Licence Agreement for the Shared Image.
 
-* `generalized` - (Optional) Should the Operating System present in this Shared Image be generalized? Defaults to `true`. Setting this to `false` creates a specialized imagine which is the equivalent of sysprepping in windows. Changing this forces a new resource to be created.
+* `specialized` - (Optional) Should the Operating System present in this Shared Image be specialized? Defaults to `false`. Setting this to `true` creates a specialized imagine which is the equivalent of sysprepping in windows. Changing this forces a new resource to be created.
```
tombuildsstuff, line 71: a Generalized image is one where SysPrep has been applied - Specialised is one where the machine data (including the UUID) is reused - which can cause serious issues (I've personally seen production incidents/data corruption from where sysprep wasn't run) - so we should probably have a warning recommending to use a Generalized image where possible

```suggestion
* `specialized` - (Optional) Specifies that the Operating System used inside this Image has not been Generalized (for example, `sysprep` on Windows has not been run). Defaults to `false`. Changing this forces a new resource to be created.

!> **Note:** It's recommended to Generalize images where possible - Specialized Images reuse the same UUID internally within each Virtual Machine, which can have unintended side-effects.
```

website/docs/d/shared_image.html.markdown
```diff
@@ -43,7 +43,7 @@ The following attributes are exported:
 
 * `location` - The supported Azure location where the Shared Image Gallery exists.
 
-* `generalized` - Is the Operating System present in this Shared Image generalized or not. `false` indicates this image is specialized which for windows images means they have been sysprepped.
+* `specialized` - Is the Operating System present in this Shared Image specialized or not. `true` indicates this image is specialized which for windows images means they have been sysprepped.
```
tombuildsstuff, line 46: (same here)

website/docs/r/shared_image.html.markdown
ArcturusZhang: I understand. I will soon make another PR to update this doc entry

website/docs/d/shared_image.html.markdown
ArcturusZhang: I understand. I will soon make another PR to update this doc entry

PR #7528, opened 2020-06-30
Improve example of the usage of network_profile_id of azurerm_container_group

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/6186

### DIFF
--- a/examples/container-instance/network-profile/main.tf
+++ b/examples/container-instance/network-profile/main.tf
@@ -1,0 +1,79 @@
+provider "azurerm" {
+  features {}
+}
+
+resource "azurerm_resource_group" "example" {
+  name     = "${var.prefix}-resources"
+  location = var.location
+}
+
+resource "azurerm_virtual_network" "example" {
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
hi @neil-yechenwei 

Thanks for this PR.

The example usages within the documentation are intended to be short-examples of how to use a particular resource - rather than covering every use-case. Instead we look to add more complex examples [to the examples directory](https://github.com/terraform-providers/terraform-provider-azurerm/tree/master/examples/container-instance) - and then [link to it from the example usage](https://www.terraform.io/docs/providers/azurerm/r/linux_virtual_machine_scale_set.html#example-usage) - as such would you mind pulling this out to make a complete example here instead?

Thanks!

neil-yechenwei:
@tombuildsstuff , thanks for your suggestion. I moved it into example folder and it has been linked to resource [document](https://www.terraform.io/docs/providers/azurerm/r/container_group.html#example-usage).

PR #7533, opened 2020-06-30
New Resource: `azurerm_kusto_cluster_principal_assignment`

jrauschenbusch:


azurerm/internal/services/kusto/kusto_cluster_principal_assignment_resource.go
```diff
@@ -0,0 +1,235 @@
...
+	props := kusto.ClusterPrincipalProperties{
+		TenantID:      utils.String(tenantID),
+		PrincipalID:   utils.String(principalID),
+		PrincipalType: kusto.PrincipalType(principalType),
+		Role:          kusto.ClusterPrincipalRole(role),
+	}
+
+	principalAssignment := kusto.ClusterPrincipalAssignment{
+		ClusterPrincipalProperties: &props,
+	}
```
mbfrahry, line 140: We could consolidate this quite a bit by assigning everything in `principalAssignment`

azurerm/internal/services/kusto/kusto_cluster_principal_assignment_resource.go
```diff
@@ -0,0 +1,235 @@
...
+		if utils.ResponseWasNotFound(resp.Response) {
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error retrieving Kusto Cluster Principal Assignment %q (Resource Group %q, Cluster %q): %+v", id.Name, id.ResourceGroup, id.Cluster, err)
+	}
+
+	d.Set("resource_group_name", id.ResourceGroup)
+	d.Set("cluster_name", id.Cluster)
+	d.Set("name", id.Name)
```
mbfrahry, line 186: We'll want to nil check all these before setting them

website/docs/r/kusto_cluster_principal_assignment.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+
+* `tenant_id` - (Required) The tenant id in which the principal resides. Changing this forces a new resource to be created.
+
+## Attributes Reference
+
+In addition to the Arguments listed above - the following Attributes are exported: 
+
+* `id` - The ID of the Kusto Cluster Principal Assignment.
+
+* `principal_name` - The computed name of the principal.
```
mbfrahry, line 68:
```suggestion
* `principal_name` - The name of the principal.
```

website/docs/r/kusto_cluster_principal_assignment.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+
+* `tenant_name` - The computed name of the tenant.
```
mbfrahry, line 70:
```suggestion
* `tenant_name` - The name of the tenant.
```

website/docs/r/kusto_cluster_principal_assignment.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+
+## Timeouts
+
+The `timeouts` block allows you to specify [timeouts](https://www.terraform.io/docs/configuration/resources.html#timeouts) for certain actions:
+
+* `create` - (Defaults to 1 hour) Used when creating the Data Explorer.
```
mbfrahry, line 76: We'll want to update the rest of these timeouts with the correct resource
```suggestion
* `create` - (Defaults to 1 hour) Used when creating the Data Explorer Principal Assignment.
```

website/docs/r/kusto_cluster_principal_assignment.html.markdown
jrauschenbusch: Hmm. But one does not need to enter the name, as it is computed by the id. It's only for verification within the state. Maybe it should be moved to the `Attributes Reference` part. But how should it be described there, as it is a nested attribute?

website/docs/r/kusto_cluster_principal_assignment.html.markdown
jrauschenbusch: copy/paste error. I'll fix that.

jrauschenbusch:
@mbfrahry Everything fixed. Ready for a 2nd review.

PR #7566, opened 2020-07-02
Generalized expand/flatten functions

magodo:
## Motivation

The existing helper functions for expanding and flattening have met a big amount of use cases. While there are some other cases users have to write its own expanding/flattening logic.

### Example 1: new type definition against native type

Say we have a SDK type defined as below:

```go
type Foo string

type Bar struct {
  Foos *[]Foo
  //...
}
```

When we are expanding for `Bar.Foos`, we can't use `ExpandStringSlice()` as it only allows the expanded-to type is `*[]string`. In this case, users have to write the almost same logic as it is defined in `ExpandStringSlice()`, just to make the compiler happy.

### Expample 2: new type of struct contains only one member of native type

Say we have a SDK type defined as below:

```go
type Foo struct {
  Bar string
}

type Baz struct {
  Foos *[]Foo
  //...
}
```
As we`Foo` only has one member, so it is likely to flatten `Foo.Bar` out to the `Baz` level in terraform schema. Then, when we are expanding that property, there is no helper function available for now, and users have to write the almost same logic as it is defined in `ExpandStringSlice`, which is also a boilerplate.

## Solution

This PR leverages the `reflect` package in stdlib to handle arbitrary types. The penalty caused by using reflection should be minor and there are already some other usage of reflection used (indirectly) by current provider (e.g. mapstruct, encoding/json). Hence it should be OK to use reflection.

To make it both easier to implement by me and to reason about by users, I have split each scenario into two flavors, for either slice and map.

So with this PR, the expanding and flattening of above code could be simplified as below:

```go

// Example 1

	ExpandSlice(
	        d.Get("foos"),
		Foo(""),
		func(x interface{}) interface{} {
			return Foo(x.(string))
		},
	)

	FlattenSlicePtr(
		resp,
		func(x interface{}) interface{} {
			return string(x.(Foo))
		},
	)

// Example 2

	ExpandSlice(
	        d.Get("foos"),
		Foo{},
		func(x interface{}) interface{} {
			return Foo{Bar: x.(string)}
		},
	)

	FlattenSlicePtr(
		resp,
		func(x interface{}) interface{} {
			return x.(Foo).Bar
		},
	)
```

The unit test includes more other use cases, including some edge cases (e.g. input contains nil, empty input, invalid input type).

### DIFF
--- a/azurerm/utils/common_marshal.go
+++ b/azurerm/utils/common_marshal.go
@@ -90,19 +90,19 @@ func FlattenInt32Slice(input *[]int32) []interface{} {
 func ExpandSlice(input []interface{}, t interface{}, convert func(interface{}) interface{}) interface{} {
 	targetType := reflect.TypeOf(t)
 	result := reflect.MakeSlice(reflect.SliceOf(targetType), 0, 0)
-	for _, item := range input {
-		if item != nil {
-			var vitem reflect.Value
+	for _, v := range input {
+		var vv reflect.Value
+		if v != nil {
 			if convert == nil {
-				vitem = reflect.New(targetType).Elem()
-				vitem.Set(reflect.ValueOf(item))
+				vv = reflect.New(targetType).Elem()
+				vv.Set(reflect.ValueOf(v))
 			} else {
-				vitem = reflect.ValueOf(convert(item))
+				vv = reflect.ValueOf(convert(v))
 			}
-			result = reflect.Append(result, vitem)
 		} else {
-			result = reflect.Append(result, reflect.Zero(targetType))
+			vv = underlyingZeroValue(targetType)
 		}
+		result = reflect.Append(result, vv)
 	}
 
 	resultp := reflect.New(result.Type())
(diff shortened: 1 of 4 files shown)


tombuildsstuff:
hey @magodo 

Thanks for this PR :)

Whilst there's some creative suggestions in here - longer term we're intending to move away from these untyped functions (particularly via `d.Get`) - and towards typed functions (for example `d.GetString`) where we're able to catch type errors via linters.

Some future enhancements to the Terraform Plugin SDK should help to mitigate these functions - by for example allowing serializing/deserializing to native types - which should provide an alternate approach here. Where it appears (from what I've seen so far) that we'll be unable to use Track2 of the Go SDK - it could make sense for the replacement for that to include some extra functionality here too to handle that - that's one which requires investigation.

As we're looking to take a different direction with this - whilst I'd like to thank you for this contribution I'm going to close this PR for the moment - but we'll be making improvements to this area in the future.

Thanks!

Issue #6485, opened 2020-04-15
Azure application Gateway Firewall Policy Associate Listener


HumanJoubert:
Azure has a feature within the application gateway firewall policy where you can associated a specific listener from the APP gateway to the WAF policy. Do we know when this feature will be available in Terraform ?

PR #7580, opened 2020-07-03
Associate WAF policy to Application Gateway listener (fixes 6485)

rikribbers:


rikribbers:
The build is failing on "make depscheck" ; error: gimme: given '1.13.x' but no release for '1.13' found

I don't know why, but does "works on my machine" count as this does not reproduce on my local machine with go 1.13.8

Looking at other build they seem to run on go 1.13 in travis. Could this be a travis hickup or am I missing something.. any feedback is highly appreciated.

website/docs/r/application_gateway.html.markdown
```diff
@@ -302,6 +302,8 @@ A `http_listener` block supports the following:
 
 * `custom_error_configuration` - (Optional) One or more `custom_error_configuration` blocks as defined below.
 
+* `firewall_policy_id` - (Optional) The resource ID of a firewall policy.
```
tombuildsstuff, line 305: we intentionally avoid the phrase `Resource ID` since it's ambiguous between Terraform and Azure Resource ID's - which aren't always the same - as such could we make this:

```suggestion
* `firewall_policy_id` - (Optional) The ID of the Web Application Firewall Policy which should be used as a HTTP Listener.
```

(also worth clarifying this is a "Web Application" Firewall Policy, since Azure Firewall reuses the name too)

azurerm/internal/services/network/application_gateway_resource.go
```diff
@@ -2332,6 +2342,10 @@ func flattenApplicationGatewayHTTPListeners(input *[]network.ApplicationGatewayH
 				output["require_sni"] = *sni
 			}
 
+			if fwp := props.FirewallPolicy; fwp != nil {
```
tombuildsstuff, line 2345: there's a crash here if `id` is nil:

```suggestion
			if fwp := props.FirewallPolicy; fwp != nil && fwp.ID != nil {
```

azurerm/internal/services/network/application_gateway_resource.go
```diff
@@ -2232,6 +2239,9 @@ func expandApplicationGatewayHTTPListeners(d *schema.ResourceData, gatewayID str
 				Protocol:                    network.ApplicationGatewayProtocol(protocol),
 				RequireServerNameIndication: utils.Bool(requireSNI),
 				CustomErrorConfigurations:   customErrorConfigurations,
+				FirewallPolicy: &network.SubResource{
+					ID: utils.String(firewallPolicyID),
+				},
```
tombuildsstuff, line 2244: I think this'll likely need to be conditionally set if `firewallPolicyID` is `!= ""` - or does this accept an empty string?

tombuildsstuff:
@rikribbers taking a look through that appears to be an error within Travis, I've re-triggered this :)

website/docs/r/application_gateway.html.markdown
rikribbers: fixed it in next commit. Found another entry (not on http listener level but at highest level of an app gw) as I copied it from there

manicminer:
Fixes #6485

manicminer:
@rikribbers Thanks for the updates, this LGTM!

Test results:
Screenshot 2020-07-06 15 20 53

HumanJoubert:
Hi, ran into an issue with the associating the listeners. I was able to only associate one WAF policy to an application gateway. Currently i have multiply WAF polices which needs to be associated to different listeners within the same application gateway. I tried this 

image 

Got this error 

Error: Attribute redefined

  on Main.tf line 79, in resource "azurerm_application_gateway" "tenant":
  79:   firewall_policy_id = azurerm_web_application_firewall_policy.externalwafpolicyrdweb.id

The argument "firewall_policy_id" was already set at Main.tf:78,3-21. Each
argument may be set only once.

Any way to get this fixed ?

PR #7585, opened 2020-07-05
Upgrade network to 2020-05-01

magodo:
Upgrade network to 2020-05-01 for all network resources, except **virtualhub** related resources and **virtualwan** related resources, which are still using the legacy version: 2020-03-01, as there is breaking change involved in the new version.
There will be another two separate PRs, one to update the version and fix the breaking changes for **virtualhub** and **virtualwan** resource, the other is to remove the 2020-03-01 version.
(The reason to split that to 2 PRs is to make the merge process easier)


tombuildsstuff:
Tests look good for this 👍

PR #7601, opened 2020-07-07
upgrade network to 2020-05-01 for vhub(conn) & vwan

### LINKED ISSUE #7896: Old Azure vWAN API issues. New vWAN Routing features are not enabled
_This issue was originally opened by @WapoInc as hashicorp/terraform#25672. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._
    
<hr>
    

### Current Terraform Version

```
Terraform v0.12.29
+ provider.azurerm v2.2.0
```

### Use-cases

Azure vWAN has just launched a whole list of new vWAN Routing features. Currently Terraform is using an old API format/version from 2018. 

These new vWAN features are all available using the Azure vWAN 2020-05-01 API. 
Is it possible for you to update your API version to 202-05-01 ?

### Proposal

These new vWAN features are all available using the Azure vWAN 2020-05-01 API. 
Is it possible for you to update your API version to 202-05-01 ?

magodo:
Upgrade network to 2020-05-01 for virtual hub(conn) & virtual wan resources.

Changes include:
- Use the dedicated client for `azurerm_virtualhub_connection`, as the original field in `azurerm_virtualhub` used to create/delete vhub conn is removed from API.
- Deprecate `hub_to_vitual_network_traffic_allowed` and `vitual_network_to_hub_gateways_traffic_allowed` from `azurerm_virtualhub_connection`, as they are deprecated by API.
- Deprecate `allow_vnet_to_vnet_traffic` from `azurerm_virtual_wan` as this is not deprecated by API.

(follow-up PR for #7585 )
(fixes #7896 )


azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -55,16 +54,20 @@ func resourceArmVirtualHubConnection() *schema.Resource {
...
 
+			// TODO 3.0: remove this property
 			"hub_to_vitual_network_traffic_allowed": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				ForceNew: true,
+				Type:       schema.TypeBool,
+				Optional:   true,
+				ForceNew:   true,
+				Deprecated: "this property has been deprecated in the API and will be removed in version 3.0 of the provider",
```
tombuildsstuff, line 62: it'd be worth clarifying this as:

```suggestion
				Deprecated: "Due to a breaking behavioural change in the Azure API this property is no longer functional and will be removed in version 3.0 of the provider",
```

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -55,16 +54,20 @@ func resourceArmVirtualHubConnection() *schema.Resource {
...
 
+			// TODO 3.0: remove this property
 			"vitual_network_to_hub_gateways_traffic_allowed": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				ForceNew: true,
+				Type:       schema.TypeBool,
+				Optional:   true,
+				ForceNew:   true,
+				Deprecated: "this property has been deprecated in the API and will be removed in version 3.0 of the provider",
```
tombuildsstuff, line 70: it'd be worth clarifying this as:

```suggestion
				Deprecated: "Due to a breaking behavioural change in the Azure API this property is no longer functional and will be removed in version 3.0 of the provider",
```

website/docs/r/virtual_hub_connection.html.markdown
```diff
@@ -58,9 +58,9 @@ The following arguments are supported:
 
 ---
 
-* `hub_to_vitual_network_traffic_allowed` - (Optional) Is the Virtual Hub traffic allowed to transit via the Remote Virtual Network? Changing this forces a new resource to be created.
+* `hub_to_vitual_network_traffic_allowed` - (Optional / **Deprecated** ) Is the Virtual Hub traffic allowed to transit via the Remote Virtual Network? Changing this forces a new resource to be created.
 
-* `vitual_network_to_hub_gateways_traffic_allowed` - (Optional) Is Remote Virtual Network traffic allowed to transit the Hub's Virtual Network Gateway's? Changing this forces a new resource to be created.
+* `vitual_network_to_hub_gateways_traffic_allowed` - (Optional / **Deprecated** ) Is Remote Virtual Network traffic allowed to transit the Hub's Virtual Network Gateway's? Changing this forces a new resource to be created.
```
tombuildsstuff, line 63: arguably since this is non-functional we could remove this:

```suggestion
```

website/docs/r/virtual_hub_connection.html.markdown
```diff
@@ -58,9 +58,9 @@ The following arguments are supported:
 
 ---
 
-* `hub_to_vitual_network_traffic_allowed` - (Optional) Is the Virtual Hub traffic allowed to transit via the Remote Virtual Network? Changing this forces a new resource to be created.
+* `hub_to_vitual_network_traffic_allowed` - (Optional / **Deprecated** ) Is the Virtual Hub traffic allowed to transit via the Remote Virtual Network? Changing this forces a new resource to be created.
```
tombuildsstuff, line 61: arguably since this is non-functional we could remove this:

```suggestion
```

website/docs/r/virtual_wan.html.markdown
```diff
@@ -40,7 +40,7 @@ The following arguments are supported:
 
 * `allow_branch_to_branch_traffic` - (Optional) Boolean flag to specify whether branch to branch traffic is allowed. Defaults to `true`.
 
-* `allow_vnet_to_vnet_traffic` - (Optional) Boolean flag to specify whether VNet to VNet traffic is allowed. Defaults to `false`.
+* `allow_vnet_to_vnet_traffic` - (Optional / **Deprecated** ) Boolean flag to specify whether VNet to VNet traffic is allowed. Defaults to `false`.
```
tombuildsstuff, line 43: arguably since this is non-functional we could remove this:

```suggestion
```

azurerm/internal/services/network/tests/virtual_hub_connection_resource_test.go
```diff
@@ -225,94 +211,90 @@ func testAccAzureRMVirtualHubConnection_complete(data acceptance.TestData) strin
...
-  remote_virtual_network_id                      = azurerm_virtual_network.test.id
-  hub_to_vitual_network_traffic_allowed          = true
-  vitual_network_to_hub_gateways_traffic_allowed = false
-  internet_security_enabled                      = false
+  name                      = "acctestvhubconn-%d"
+  virtual_hub_id            = azurerm_virtual_hub.test.id
+  remote_virtual_network_id = azurerm_virtual_network.test.id
+  internet_security_enabled = false
 }
-
```
tombuildsstuff, line 262: leaving the spacing in here does make this easier to read (since we do link to acceptance tests as examples from time to time) - so it'd be nice to leave these in

website/docs/r/virtual_hub_connection.html.markdown
magodo: How about just keep this (as user is still allowed to assign it) but change the document so something like:

`Due to a breaking behavioural change in the Azure API this property is no longer functional and will be removed in version 3.0 of the provider`

azurerm/internal/services/network/tests/virtual_hub_connection_resource_test.go
magodo: This is by accident, I copied it from another PR from github web file view (not raw view).
It behaves strangely that the empty line in the range of backtick is stripped, while the other empty lines are reserved 😂

magodo:
@tombuildsstuff Thank you for the review comments! I have updated accordingly. Please take a look!

magodo:
@WodansSon I have submit another update to ensure the routing state reaches "Provisioned" so as to avoid race condition between vhub internal routing update and hub vnet connection caused routing update.

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -89,32 +92,18 @@ func resourceArmVirtualHubConnectionCreate(d *schema.ResourceData, meta interfac
...
 	name := d.Get("name").(string)
 
-	if features.ShouldResourcesBeImported() {
-		if connection, _ := findVirtualHubConnection(name, virtualHub); connection != nil {
-			return tf.ImportAsExistsError("azurerm_virtual_hub_connection", *connection.ID)
+	if d.IsNewResource() {
+		existing, err := client.Get(ctx, id.ResourceGroup, id.Name, name)
+		if err != nil {
+			if !utils.ResponseWasNotFound(existing.Response) {
+				return fmt.Errorf("checking for present of existing Virtual Hub Connection %q (Resource Group %q): %+v", name, id.ResourceGroup, err)
```
tombuildsstuff, line 101:
```suggestion
				return fmt.Errorf("checking for presence of existing Connection %q (Virtual Hub %q / Resource Group %q): %+v", name, id.Name, id.ResourceGroup, err)
```

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -89,32 +92,18 @@ func resourceArmVirtualHubConnectionCreate(d *schema.ResourceData, meta interfac
...
-	if virtualHub.VirtualHubProperties == nil {
-		return fmt.Errorf("Error retrieving Virtual Hub %q (Resource Group %q): `properties` was nil", id.Name, id.ResourceGroup)
-	}
-
 	name := d.Get("name").(string)
 
-	if features.ShouldResourcesBeImported() {
-		if connection, _ := findVirtualHubConnection(name, virtualHub); connection != nil {
-			return tf.ImportAsExistsError("azurerm_virtual_hub_connection", *connection.ID)
+	if d.IsNewResource() {
```
tombuildsstuff, line 97: this method isn't updating - so we can remove this `if` here

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -123,45 +112,34 @@ func resourceArmVirtualHubConnectionCreate(d *schema.ResourceData, meta interfac
...
-		return fmt.Errorf("Error retrieving Virtual Hub %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
-	}
-	if newConnection == nil {
-		return fmt.Errorf("Connection %q was not found in Virtual Hub %q / Resource Group %q", name, id.Name, id.ResourceGroup)
+		return fmt.Errorf("retrieving Virtual Hub Connection %q (Resource Group %q): %+v", name, id.ResourceGroup, err)
 	}
-	if newConnection.ID == nil && *newConnection.ID == "" {
-		return fmt.Errorf("Error retrieving Connection %q (Virtual Hub %q / Resource Group %q): `id` was nil or empty", name, id.Name, id.ResourceGroup)
+	if resp.ID == nil || *resp.ID == "" {
+		return fmt.Errorf("Cannot read Virtual Hub Connection %q (Resource Group %q) ID", name, id.ResourceGroup)
```
tombuildsstuff, line 133:
```suggestion
		return fmt.Errorf("Cannot read Connection %q (Virtual Hub %q / Resource Group %q) ID", name, id.Name, id.ResourceGroup)
```

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -123,45 +112,34 @@ func resourceArmVirtualHubConnectionCreate(d *schema.ResourceData, meta interfac
...
 	}
 
-	newConnection, err := findVirtualHubConnection(name, virtualHub)
+	resp, err := client.Get(ctx, id.ResourceGroup, id.Name, name)
 	if err != nil {
-		return fmt.Errorf("Error retrieving Virtual Hub %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
-	}
-	if newConnection == nil {
-		return fmt.Errorf("Connection %q was not found in Virtual Hub %q / Resource Group %q", name, id.Name, id.ResourceGroup)
+		return fmt.Errorf("retrieving Virtual Hub Connection %q (Resource Group %q): %+v", name, id.ResourceGroup, err)
```
tombuildsstuff, line 130:
```suggestion
		return fmt.Errorf("retrieving Connection %q (Virtual Hub %q / Resource Group %q): %+v", name, id.Name, id.ResourceGroup, err)
```

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -123,45 +112,34 @@ func resourceArmVirtualHubConnectionCreate(d *schema.ResourceData, meta interfac
...
 	}
 
 	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
-		return fmt.Errorf("Error waiting for addition of Connection %q to Virtual Hub %q (Resource Group %q): %+v", name, id.Name, id.ResourceGroup, err)
-	}
-
-	virtualHub, err = client.Get(ctx, id.ResourceGroup, id.Name)
-	if err != nil {
-		return fmt.Errorf("Error retrieving Virtual Hub %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
+		return fmt.Errorf("waiting for creation of Virtual Hub Connection %q (Resource Group %q): %+v", name, id.ResourceGroup, err)
```
tombuildsstuff, line 125:
```suggestion
		return fmt.Errorf("waiting for creation of Connection %q (Virtual Hub %q / Resource Group %q): %+v", name, id.Name, id.ResourceGroup, err)
```

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -123,45 +112,34 @@ func resourceArmVirtualHubConnectionCreate(d *schema.ResourceData, meta interfac
...
 		},
 	}
-	connections = append(connections, connection)
-	virtualHub.VirtualHubProperties.VirtualNetworkConnections = &connections
 
-	future, err := client.CreateOrUpdate(ctx, id.ResourceGroup, id.Name, virtualHub)
+	future, err := client.CreateOrUpdate(ctx, id.ResourceGroup, id.Name, name, connection)
 	if err != nil {
-		return fmt.Errorf("Error adding Connection %q to Virtual Hub %q (Resource Group %q): %+v", name, id.Name, id.ResourceGroup, err)
+		return fmt.Errorf("creating Virtual Hub Connection %q (Resource Group %q): %+v", name, id.ResourceGroup, err)
```
tombuildsstuff, line 121:
```suggestion
		return fmt.Errorf("creating Connection %q (Virtual Hub %q / Resource Group %q): %+v", name, id.Name, id.ResourceGroup, err)
```

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -170,36 +148,28 @@ func resourceArmVirtualHubConnectionRead(d *schema.ResourceData, meta interface{
 		return err
 	}
 
-	virtualHub, err := client.Get(ctx, id.ResourceGroup, id.VirtualHubName)
+	resp, err := client.Get(ctx, id.ResourceGroup, id.VirtualHubName, id.Name)
 	if err != nil {
-		if utils.ResponseWasNotFound(virtualHub.Response) {
-			log.Printf("[DEBUG] Virtual Hub %q was not found in Resource Group %q - so Connection %q can't exist - removing from state", id.Name, id.VirtualHubName, id.ResourceGroup)
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[INFO] Virtual Hub Connection %q does not exist - removing from state", d.Id())
```
tombuildsstuff, line 154:
```suggestion
			log.Printf("[INFO] Connection %q (Virtual Hub %q / Resource Group %q) does not exist - removing from state", id.Name, id.VirtualHubName, id.ResourceGroup)
```

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -170,36 +148,28 @@ func resourceArmVirtualHubConnectionRead(d *schema.ResourceData, meta interface{
...
 			d.SetId("")
 			return nil
 		}
-
-		return fmt.Errorf("Error retrieving Virtual Hub %q (Resource Group %q): %+v", id.VirtualHubName, id.ResourceGroup, err)
+		return fmt.Errorf("reading Virtual Hub Connection %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
```
tombuildsstuff, line 158:
```suggestion
		return fmt.Errorf("reading Connection %q (Virtual Hub %q / Resource Group %q): %+v", id.Name, id.VirtualHubName, id.ResourceGroup, err)
```

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -170,36 +148,28 @@ func resourceArmVirtualHubConnectionRead(d *schema.ResourceData, meta interface{
...
 	}
 
-	connection, err := findVirtualHubConnection(id.Name, virtualHub)
+	virtualHubResp, err := virtualHubClient.Get(ctx, id.ResourceGroup, id.VirtualHubName)
```
tombuildsstuff, line 161: we shouldn't need to look this up if we use an ID Formatter instead - can we switch this over?

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -224,68 +194,14 @@ func resourceArmVirtualHubConnectionDelete(d *schema.ResourceData, meta interfac
...
-
-		newConnections = append(newConnections, connection)
-	}
-	virtualHub.VirtualHubProperties.VirtualNetworkConnections = &newConnections
-
-	future, err := client.CreateOrUpdate(ctx, id.ResourceGroup, id.VirtualHubName, virtualHub)
+	future, err := client.Delete(ctx, id.ResourceGroup, id.VirtualHubName, id.Name)
 	if err != nil {
-		return fmt.Errorf("Error removing Connection %q to Virtual Hub %q (Resource Group %q): %+v", id.Name, id.VirtualHubName, id.ResourceGroup, err)
+		return fmt.Errorf("deleting Virtual Hub Connection %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
```
tombuildsstuff, line 199:
```suggestion
		return fmt.Errorf("deleting Connection %q (Virtual Hub %q / Resource Group %q): %+v", id.Name, id.VirtualHubName, id.ResourceGroup, err)
```

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -224,68 +194,14 @@ func resourceArmVirtualHubConnectionDelete(d *schema.ResourceData, meta interfac
...
 	}
 
-	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
-		return fmt.Errorf("Error waiting for removal of Connection %q to Virtual Hub %q (Resource Group %q): %+v", id.Name, id.VirtualHubName, id.ResourceGroup, err)
+	if err = future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("waiting for deleting Virtual Hub Connection %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
```
tombuildsstuff, line 203:
```suggestion
		return fmt.Errorf("waiting for deleting Connection %q (Virtual Hub %q / Resource Group %q): %+v", id.Name, id.VirtualHubName, id.ResourceGroup, err)
```

azurerm/internal/services/network/virtual_hub_resource.go
```diff
@@ -142,10 +144,23 @@ func resourceArmVirtualHubCreateUpdate(d *schema.ResourceData, meta interface{})
...
 	}
 
-	resp, err := client.Get(ctx, resourceGroup, name)
+	// Hub returns provisioned while the routing state is still "provisining". This might cause issues with following hubvnet connection operations.
+	// https://github.com/Azure/azure-rest-api-specs/issues/10391
+	// As a workaround, we will poll the routing state and ensure it is "Provisioned".
+	timeout, _ := ctx.Deadline()
+	stateConf := &resource.StateChangeConf{
+		Pending:      []string{"Provisioning"},
+		Target:       []string{"Provisioned", "Failed"},
```
tombuildsstuff, line 153: Failed shouldn't be a success state here - if the connection fails we should raise an error?

azurerm/internal/services/network/virtual_hub_resource.go
```diff
@@ -142,10 +144,23 @@ func resourceArmVirtualHubCreateUpdate(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf("Error waiting for creation of Virtual Hub %q (Resource Group %q): %+v", name, resourceGroup, err)
 	}
 
-	resp, err := client.Get(ctx, resourceGroup, name)
+	// Hub returns provisioned while the routing state is still "provisining". This might cause issues with following hubvnet connection operations.
+	// https://github.com/Azure/azure-rest-api-specs/issues/10391
+	// As a workaround, we will poll the routing state and ensure it is "Provisioned".
+	timeout, _ := ctx.Deadline()
```
tombuildsstuff, line 150: all timeouts should be caught, but we should still raise the error here

azurerm/internal/services/network/virtual_hub_resource.go
```diff
@@ -142,10 +144,23 @@ func resourceArmVirtualHubCreateUpdate(d *schema.ResourceData, meta interface{})
...
+	stateConf := &resource.StateChangeConf{
+		Pending:      []string{"Provisioning"},
+		Target:       []string{"Provisioned", "Failed"},
+		Refresh:      virtualHubCreateRefreshFunc(ctx, client, resourceGroup, name),
+		PollInterval: 5 * time.Second,
```
tombuildsstuff, line 155: this'll cause (rate-limiting) issues for other resources - we likely want this set to `15` at a minimum (although likely 30) with a "Occurences" of 3 to confirm it's stable?

azurerm/internal/services/network/virtual_hub_connection_resource.go
```diff
@@ -170,36 +148,28 @@ func resourceArmVirtualHubConnectionRead(d *schema.ResourceData, meta interface{
...
+		return fmt.Errorf("Cannot read Virtual Hub %q (Resource Group %q) ID", id.VirtualHubName, id.ResourceGroup)
 	}
 
 	d.Set("name", id.Name)
-	d.Set("virtual_hub_id", virtualHub.ID)
+	d.Set("virtual_hub_id", virtualHubResp.ID)
 
-	if props := connection.HubVirtualNetworkConnectionProperties; props != nil {
-		d.Set("hub_to_vitual_network_traffic_allowed", props.AllowHubToRemoteVnetTransit)
-		d.Set("vitual_network_to_hub_gateways_traffic_allowed", props.AllowRemoteVnetToUseHubVnetGateways)
```
tombuildsstuff, line 202: we should probably still set these properties to `false` into the state for now

azurerm/internal/services/network/client/client.go
```diff
@@ -1,7 +1,6 @@
 package client
 
 import (
-	networkLegacy "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-03-01/network"
```
tombuildsstuff, line 4: this is now used for LB's: https://github.com/terraform-providers/terraform-provider-azurerm/blob/8c52648735db0a8b1506178a93687daff75de4fe/azurerm/internal/services/network/client/client.go#L21 - so I think this needs to be left-in & the files re-vendored?

tombuildsstuff (changes requested):
hey @magodo 

Thanks for pushing those changes - I've taken a look through and left some comments inline; if we can fix those up then this should otherwise be good to merge (we'll need to add an "upgrade note" to the changelog, but we'll do that at merge time)

Thanks!

azurerm/internal/services/network/virtual_hub_resource.go
magodo: The second returned value is a `bool` indicating whether the deadline is set on the context, as we have that guaranteed via CI/Makefile, we can simply ignore handling it here?
(did you treat `ctx.Deadline()` as `ctx.Err()` by mistake?)

azurerm/internal/services/network/client/client.go
magodo: Nice catch!

azurerm/internal/services/network/virtual_hub_resource.go
tombuildsstuff: that should be trie - but isn't necessarily if we end up wrapping this in a migration function for whatever reason (where we'd likely use `context.TODO()`, as I believe we do for storage accounts - so there's no harm in checking/raising this, even if it'd be a niche use-case/error

azurerm/internal/services/network/parse/virtual_hub.go
```diff
@@ -1,23 +1,33 @@
...
 }
 
-func ParseVirtualHubID(input string) (*VirtualHubResourceID, error) {
+func NewVirtualHubID(resourceGroup, name string) VirtualHubId {
+	return VirtualHubId{
+		ResourceGroup: resourceGroup,
+		Name:          name,
+	}
+}
+func (id VirtualHubId) ID(subscription string) string {
```
tombuildsstuff, line 20: _minor:_ the other formatters use `subscriptionId` as the argument so we may as well make this consistent:

```suggestion
func (id VirtualHubId) ID(subscriptionId string) string {
```

azurerm/internal/services/network/parse/virtual_hub_connection.go
```diff
@@ -6,20 +6,33 @@ import (
...
-func ParseVirtualHubConnectionID(input string) (*VirtualHubConnectionResourceID, error) {
+func NewVirtualHubConnectionID(id VirtualHubId, name string) VirtualHubConnectionId {
+	return VirtualHubConnectionId{
+		ResourceGroup:  id.ResourceGroup,
+		VirtualHubName: id.Name,
+		Name:           name,
+	}
+}
+
+func (id VirtualHubConnectionId) ID(subscription string) string {
```
tombuildsstuff, line 23: _minor:_ the other formatters use `subscriptionId` as the argument so we may as well make this consistent:

```suggestion
func (id VirtualHubConnectionId) ID(subscriptionId string) string {
```

azurerm/internal/services/network/virtual_hub_connection_resource.go
tombuildsstuff: looks like this is still pending?

azurerm/internal/services/network/virtual_hub_connection_resource.go
tombuildsstuff: this is still pending?

tombuildsstuff (changes requested):
hey @magodo 

Thanks for pushing those changes - I've taken a look through and it appears there's still a couple of pending comments - but if we can fix those up then we should be able to run the tests and get this merged 👍 

Thanks!

azurerm/internal/services/network/virtual_hub_connection_resource.go
magodo: Now that those two deprecated attributes are not `ForceNew()`, I need to extend the create callback to CreateorUpdate, hence keep this one.

azurerm/internal/services/network/virtual_hub_resource.go
magodo: OK, I'll put the check at the very begining of this function to fail fast (if deadline is not set properly).
(does *niche* mean nicht 🇩🇪 here)

magodo:
@tombuildsstuff I have resolved the remaining comments, please take another look! Thanks!

azurerm/internal/services/network/virtual_hub_connection_resource.go
magodo: These properties are always return `true`, as they are deprecated and has no effect whatever value they are set, we can totally ignore them in the diff via introducing a diff suppress function which always returns true?

tombuildsstuff: if someone removes these from their config this'll still be false though (and false for new resources) - so wouldn't that diff suppress func cause the deprecation not to be shown?

magodo: Make sense.

magodo:
@tombuildsstuff I have resolved the remaining comments, especially [this one](https://github.com/terraform-providers/terraform-provider-azurerm/pull/7601#discussion_r477203234). Would you please take another look?

WodansSon:
@magodo, thanks for pushing these changes and fixing the conflict! 🚀

WodansSon:
image

Dilergore:
Hi,

This is very urgent. Using the old API version can clear the routes on vWAN causing major issues. We had a production incident caused by this.

Also mentioned as recommended in the vWAN FAQ:
https://docs.microsoft.com/en-us/azure/virtual-wan/virtual-wan-faq#what-is-the-recommended-api-version-to-be-used-by-scripts-automating-various-virtual-wan-functionalities

tombuildsstuff (approved):
LGTM 👍

PR #7616, opened 2020-07-08
dependencies: updating to v44.0.0 of github.com/Azure/azure-sdk-for-go

ArcturusZhang:
This is required by the customer requirement of BlackRock and another RP onboard of hardware security modules


azurerm/internal/services/blueprints/blueprint_assignment_resource.go
```diff
@@ -311,7 +311,9 @@ func resourceArmBlueprintAssignmentDelete(d *schema.ResourceData, meta interface
 	name := assignmentID.Name
 	targetScope := assignmentID.Scope
 
-	resp, err := client.Delete(ctx, targetScope, name)
+	// We use none here to align the previous behaviour of the blueprint resource
+	// TODO: we could add a features flag for the blueprint to empower terraform when deleting the blueprint to delete all the generated resources as well
```
katbyte, line 315: I think we should remove this as it may or may not actually be a good idea according to @jackofallops 
```suggestion
```

PR #7676, opened 2020-07-10
Update data source `azurerm_managed_disk` - Fix acc tests

ArcturusZhang:
Error message: 
```
panic: Invalid address to set: []string{"image_reference_id"}
```

The data source of `azurerm_managed_disk` is setting a non-existing attribute `image_reference_id` and therefore the test panics

This PR adds the attribute `image_reference_id` to the data source and hence the tests are fixed


tombuildsstuff:
Tests pass:

Screenshot 2020-07-10 at 18 08 51

PR #7680, opened 2020-07-10
data source `azurerm_dns_zone`: make `resource_group_name` required

njuCZ:
according to https://docs.microsoft.com/en-us/azure/dns/dns-zones-records#dns-zones, dns zone name is only unique in resource group. Different resource group could use the same dns zone name. it's not unique global.

So I think it does not make sense to search all resource group and return first found result.

### DIFF
--- a/azurerm/internal/services/dns/dns_zone_data_source.go
+++ b/azurerm/internal/services/dns/dns_zone_data_source.go
@@ -77,14 +77,17 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
 			return fmt.Errorf("Error reading DNS Zone %q (Resource Group %q): %+v", name, resourceGroup, err)
 		}
 	} else {
-		resp, resourceGroup, err = findZone(client, ctx, name)
+		var zone *dns.Zone
+		zone, resourceGroup, err = findZone(client, ctx, name)
 		if err != nil {
 			return err
 		}
 
-		if resp.ID == nil || *resp.ID == "" {
+		if zone == nil {
 			return fmt.Errorf("Error: DNS Zone %q was not found", name)
 		}
+
+		resp = *zone
 	}
 
 	d.SetId(*resp.ID)
@@ -107,10 +110,10 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
 	return tags.FlattenAndSet(d, resp.Tags)
 }
 
-func findZone(client *dns.ZonesClient, ctx context.Context, name string) (dns.Zone, string, error) {
+func findZone(client *dns.ZonesClient, ctx context.Context, name string) (*dns.Zone, string, error) {
 	zonesIterator, err := client.ListComplete(ctx, nil)
 	if err != nil {
-		return dns.Zone{}, "", fmt.Errorf("listing DNS Zones: %+v", err)
+		return nil, "", fmt.Errorf("listing DNS Zones: %+v", err)
 	}
 
 	var found *dns.Zone
@@ -118,22 +121,22 @@ func findZone(client *dns.ZonesClient, ctx context.Context, name string) (dns.Zo
 		zone := zonesIterator.Value()
 		if zone.Name != nil && *zone.Name == name {
 			if found != nil {
-				return dns.Zone{}, "", fmt.Errorf("found multiple DNS zones with name %q, please specify the resource group", name)
+				return nil, "", fmt.Errorf("found multiple DNS zones with name %q, please specify the resource group", name)
 			}
 			found = &zone
 		}
 		if err := zonesIterator.NextWithContext(ctx); err != nil {
-			return dns.Zone{}, "", fmt.Errorf("listing DNS Zones: %+v", err)
+			return nil, "", fmt.Errorf("listing DNS Zones: %+v", err)
 		}
 	}
 
 	if found == nil || found.ID == nil {
-		return dns.Zone{}, "", fmt.Errorf("could not find DNS zone with name: %q", name)
+		return nil, "", fmt.Errorf("could not find DNS zone with name: %q", name)
 	}
 
 	id, err := parse.DnsZoneID(*found.ID)
 	if err != nil {
-		return dns.Zone{}, "", fmt.Errorf("DNS zone id not valid: %+v", err)
+		return nil, "", fmt.Errorf("DNS zone id not valid: %+v", err)
 	}
-	return *found, id.ResourceGroup, nil
+	return found, id.ResourceGroup, nil
 }
--- a/azurerm/internal/services/dns/dns_zone_data_source.go
+++ b/azurerm/internal/services/dns/dns_zone_data_source.go
@@ -5,11 +5,10 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/dns/parse"
-
 	"github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2018-05-01/dns"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
hey @njuCZ 

Thanks for this PR.

Taking a look through here - whilst these can be unique per resource group, Terraform should still support looking these up - however Terraform should raise an error if multiple DNS Zones are found with the same name, would you be able to update this PR to fix this instead?

Thanks!

njuCZ:
@tombuildsstuff I have modified this PR, please have a look again.
1. use `listComplete` rather than list all resource group and then ListByResourceGroup
2. when user does not specify `resource_group_name`, reports error if two dns zone have the same name
3. because resource group of dns zone is always lower case, change the test case

azurerm/internal/services/dns/dns_zone_data_source.go
```diff
@@ -109,26 +108,30 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
...
-			return dns.Zone{}, "", fmt.Errorf("Error listing DNS Zones (Resource Group: %s): %+v", resourceGroup, err)
-		}
-
-		for _, z := range zones.Values() {
-			if *z.Name == name {
-				return z, resourceGroup, nil
+	var found *dns.Zone
+	for zonesIterator.NotDone() {
+		zone := zonesIterator.Value()
+		if *zone.Name == name {
```
tombuildsstuff, line 120: there's a crash here if `zone.Name` is nil

azurerm/internal/services/dns/dns_zone_data_source.go
```diff
@@ -109,26 +108,30 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
...
+			return dns.Zone{}, "", fmt.Errorf("listing DNS Zones: %+v", err)
 		}
 	}
 
-	return dns.Zone{}, "", nil
+	if found == nil {
+		return dns.Zone{}, "", fmt.Errorf("could not find DNS zone with name: %q", name)
+	}
+
+	id, _ := parse.DnsZoneID(*found.ID)
```
tombuildsstuff, line 135: we should handle this error here

njuCZ:
@tombuildsstuff thanks for your suggestions. I have updated this PR, please have a look again

azurerm/internal/services/dns/dns_zone_data_source.go
```diff
@@ -109,26 +107,33 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
 	return tags.FlattenAndSet(d, resp.Tags)
 }
 
-func findZone(client *dns.ZonesClient, rgClient *resources.GroupsClient, ctx context.Context, name string) (dns.Zone, string, error) {
-	groups, err := rgClient.List(ctx, "", nil)
+func findZone(client *dns.ZonesClient, ctx context.Context, name string) (dns.Zone, string, error) {
```
tombuildsstuff, line 110: thinking about it this probably should be a pointer:

```suggestion
func findZone(client *dns.ZonesClient, ctx context.Context, name string) (*dns.Zone, string, error) {
```

azurerm/internal/services/dns/dns_zone_data_source.go
```diff
@@ -109,26 +107,33 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
...
+	zonesIterator, err := client.ListComplete(ctx, nil)
 	if err != nil {
-		return dns.Zone{}, "", fmt.Errorf("Error listing Resource Groups: %+v", err)
+		return dns.Zone{}, "", fmt.Errorf("listing DNS Zones: %+v", err)
```
tombuildsstuff, line 113: that allows this to be nil - ensuring that if we return an error, trying to access the DNS Zone should panic - which'll be caught by tests (and forces us to check errors first):

```suggestion
		return nil, "", fmt.Errorf("listing DNS Zones: %+v", err)
```

azurerm/internal/services/dns/dns_zone_data_source.go
```diff
@@ -109,26 +107,33 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
...
-
-		for _, z := range zones.Values() {
-			if *z.Name == name {
-				return z, resourceGroup, nil
+	var found *dns.Zone
+	for zonesIterator.NotDone() {
+		zone := zonesIterator.Value()
+		if zone.Name != nil && *zone.Name == name {
+			if found != nil {
+				return dns.Zone{}, "", fmt.Errorf("found multiple DNS zones with name %q, please specify the resource group", name)
```
tombuildsstuff, line 121:
```suggestion
				return nil, "", fmt.Errorf("found multiple DNS zones with name %q, please specify the resource group", name)
```

azurerm/internal/services/dns/dns_zone_data_source.go
```diff
@@ -109,26 +107,33 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
...
 			}
+			found = &zone
 		}
+		if err := zonesIterator.NextWithContext(ctx); err != nil {
+			return dns.Zone{}, "", fmt.Errorf("listing DNS Zones: %+v", err)
```
tombuildsstuff, line 126:
```suggestion
			return nil, "", fmt.Errorf("listing DNS Zones: %+v", err)
```

azurerm/internal/services/dns/dns_zone_data_source.go
```diff
@@ -109,26 +107,33 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
...
+		}
+	}
+
+	if found == nil || found.ID == nil {
+		return dns.Zone{}, "", fmt.Errorf("could not find DNS zone with name: %q", name)
```
tombuildsstuff, line 131:
```suggestion
		return nil, "", fmt.Errorf("could not find DNS zone with name: %q", name)
```

azurerm/internal/services/dns/dns_zone_data_source.go
```diff
@@ -109,26 +107,33 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
...
 	}
 
-	return dns.Zone{}, "", nil
+	id, err := parse.DnsZoneID(*found.ID)
+	if err != nil {
+		return dns.Zone{}, "", fmt.Errorf("DNS zone id not valid: %+v", err)
```
tombuildsstuff, line 136:
```suggestion
		return nil, "", fmt.Errorf("DNS zone id not valid: %+v", err)
```

azurerm/internal/services/dns/dns_zone_data_source.go
```diff
@@ -109,26 +107,33 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
...
+	}
+	return *found, id.ResourceGroup, nil
```
tombuildsstuff, line 138: this should be de-referenced anyway - so it should be fine:

```suggestion
	return found, id.ResourceGroup, nil
```

azurerm/internal/services/dns/dns_zone_data_source.go
```diff
@@ -77,16 +77,17 @@ func dataSourceArmDnsZoneRead(d *schema.ResourceData, meta interface{}) error {
...
 			return err
 		}
 
-		if resourceGroup == "" {
+		if zone == nil {
 			return fmt.Errorf("Error: DNS Zone %q was not found", name)
 		}
+
+		resp = *zone
 	}
```
njuCZ, line 91: @tombuildsstuff thanks for your suggestions. I have updated this function to return a pointer. In order to use the same variable `resp` and `resourcegroup`, I made a little update here, please have a check again.

njuCZ:
@tombuildsstuff thanks for your suggestions. I have updated this function to return a pointer. in order to use the same variable `resp` and `resourceGroup`, I made a little change in the read func, please have a check

jackofallops:
Tests passing:
image

PR #7682, opened 2020-07-10
Fix TC for cdn endpoint

neil-yechenwei:
--- PASS: TestAccAzureRMCdnEndpoint_disappears (381.62s)
--- PASS: TestAccAzureRMCdnEndpoint_requiresImport (403.98s)
--- PASS: TestAccAzureRMCdnEndpoint_withGeoFilters (448.88s)
--- PASS: TestAccAzureRMCdnEndpoint_optimized (449.74s)
--- PASS: TestAccAzureRMCdnEndpoint_fullFields (449.88s)
--- PASS: TestAccAzureRMCdnEndpoint_basic (456.86s)
--- PASS: TestAccAzureRMCdnEndpoint_isHttpAndHttpsAllowedUpdate (492.88s)
--- PASS: TestAccAzureRMCdnEndpoint_withTags (504.70s)
--- PASS: TestAccAzureRMCdnEndpoint_updateHostHeader (592.53s)
--- PASS: TestAccAzureRMCdnEndpoint_globalDeliveryRule (594.38s)
--- PASS: TestAccAzureRMCdnEndpoint_deliveryRule (697.89s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/cdn/tests   709.826s

### DIFF
--- a/azurerm/internal/services/cdn/tests/cdn_endpoint_resource_test.go
+++ b/azurerm/internal/services/cdn/tests/cdn_endpoint_resource_test.go
@@ -82,10 +82,10 @@ func TestAccAzureRMCdnEndpoint_updateHostHeader(t *testing.T) {
 		CheckDestroy: testCheckAzureRMCdnEndpointDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccAzureRMCdnEndpoint_hostHeader(data, "www.example.com"),
+				Config: testAccAzureRMCdnEndpoint_hostHeader(data, "www.google.com"),
 				Check: resource.ComposeTestCheckFunc(
 					testCheckAzureRMCdnEndpointExists(data.ResourceName),
-					resource.TestCheckResourceAttr(data.ResourceName, "origin_host_header", "www.example.com"),
+					resource.TestCheckResourceAttr(data.ResourceName, "origin_host_header", "www.google.com"),
 				),
 			},
 			{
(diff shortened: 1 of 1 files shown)


azurerm/internal/services/cdn/tests/cdn_endpoint_resource_test.go
```diff
@@ -82,10 +82,10 @@ func TestAccAzureRMCdnEndpoint_updateHostHeader(t *testing.T) {
 		CheckDestroy: testCheckAzureRMCdnEndpointDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccAzureRMCdnEndpoint_hostHeader(data, "www.example.com"),
+				Config: testAccAzureRMCdnEndpoint_hostHeader(data, "www.google.com"),
```
tombuildsstuff, line 85: [www.example.com](http://www.example.com) is explicitly meant to be used for demonstration purposes - is the API explicitly disallowing this? if so this feels like a bug in the API?

neil-yechenwei: After tested, seems I can create it successfully with same host name in azure portal. After checked api version, I found they are different. Azure portal is using 2020-04-15 but terraform is using 2019-04-15. I assume there is a bug in version 2019-04-15. So I filed an issue on Azure/azure-rest-api-specs#10090

PR #7701, opened 2020-07-11
example for azure monitoring

ishantkumar:


examples/azure-monitoring/variables.tf
```diff
@@ -0,0 +1,22 @@
+variable "cache" {
+	description = "The Azure Redis Cache alerts"
+
+	default = {
```
tombuildsstuff, line 4: for the purposes of the example we could hard-code these values in the resource - in particular it'd be worth provisioning the Redis Cache (and the Action Group) required for this to work end-to-end so that users can run this?

examples/azure-monitoring/variables.tf
```diff
@@ -0,0 +1,22 @@
...
+		
+    }
+
+	}
+
+variable "action_group" {
+  description = "The azure action group"
+  type        = string
+  default = "/subscriptions/<subscription_id>/resourceGroups/<resource_group_name>/providers/microsoft.insights/actionGroups/<action_group_name>"
+}
```
tombuildsstuff, line 22: rather than assuming one exists for demonstration purposes could we provision an action group as a part of the example and thread that through as required?

ishantkumar: Hi @tombuildsstuff , Thanks for pointers. I have committed the changes.

PR #7703, opened 2020-07-12
Property for hbi_enabled with machine learning workspace

damienpontifex:
- Updates ML workspace API version from 2019-11-01 -> 2020-04-01
- Adds `high_business_impact_workspace` property to workspace to indicate high business impact resulting in reduced sensitive data and telemetry collected by Microsoft


website/docs/r/machine_learning_workspace.html.markdown
```diff
@@ -94,6 +94,8 @@ The following arguments are supported:
 
 * `tags` - (Optional) A mapping of tags to assign to the resource. Changing this forces a new resource to be created.
 
+* `hbi_workspace` - (Optional) Flag to signal HBI (High Business Impact) data in the workspace and reduce diagnostic data collected by the service
```
tombuildsstuff, line 97: since we've had to call this out here, it sounds like it'd be worth making the field `high_business_impact_workspace` so this is clearer, what do you think? :)

(also since the rest of these are alphabetically sorted, could we move this up into-place for consistency?)

damienpontifex: 100% agree on the naming/acronym. Originally I had it as the acronym as it aligned with the ARM JSON property. But I like it being easier to understand with this naming.
Also did the alphabetical sorting...let me know if I missed one anywhere with this.

damienpontifex: From your experience @tombuildsstuff is it better to have it as `high_business_impact_workspace = true` vs `high_business_impact = true` considering the property is part of the workspace? i.e. is the workspace suffix on the name redundant? Or equally the extra verbosity is fine and aligns with most other terraform patterns here?

tombuildsstuff: 👍 agreed that the suffix is probably superfluous - so `high_business_impact` makes sense to me 👍

damienpontifex: Updated

tombuildsstuff:
Tests pass:

Screenshot 2020-07-14 at 11 11 17

PR #7704, opened 2020-07-12
Support Azure Data Factory Linked Service OData

pomverte:
Add support for Terraform configuration for Azure Data Factory Linked Service of type OData.

New resource `azurerm_data_factory_linked_service_odata`

```hcl
resource "azurerm_data_factory_linked_service_odata" "example" {
  name                = "example"
  data_factory_name   = azurerm_data_factory.example.name
  authentication_type = "Anonymous"
  url                 = "https://services.odata.org/v4/TripPinServiceRW/People"
}
```

Documentation : https://docs.microsoft.com/en-us/azure/data-factory/connector-odata

### DIFF
--- a/azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
+++ b/azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
@@ -91,15 +91,13 @@ func resourceArmDataFactoryLinkedServiceOData() *schema.Resource {
 			"tenant": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				ValidateFunc: validation.StringIsNotEmpty,
-				// ValidateFunc: validation.StringIsValidRegExp("[a-z 0-9]{8}-[a-z 0-9]{4}-[a-z 0-9]{4}-[a-z 0-9]{4}-[a-z 0-9]{12}"),
+				ValidateFunc: validation.IsUUID,
 			},
 
 			"service_principal_id": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				ValidateFunc: validation.StringIsNotEmpty,
-				// ValidateFunc: validation.StringIsValidRegExp("[a-z 0-9]{8}-[a-z 0-9]{4}-[a-z 0-9]{4}-[a-z 0-9]{4}-[a-z 0-9]{12}"),
+				ValidateFunc: validation.IsUUID,
 			},
 
 			"aad_service_principal_credential_type": {
--- a/azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
+++ b/azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
@@ -82,7 +82,7 @@ func resourceArmDataFactoryLinkedServiceOData() *schema.Resource {
 				ValidateFunc: validation.StringIsNotEmpty,
 			},
 
-			"aad_resource_id": {
+			"azuread_resource_id": {
 				Type:         schema.TypeString,
 				Optional:     true,
 				ValidateFunc: validation.StringIsNotEmpty,
@@ -100,7 +100,7 @@ func resourceArmDataFactoryLinkedServiceOData() *schema.Resource {
 				ValidateFunc: validation.IsUUID,
 			},
 
-			"aad_service_principal_credential_type": {
+			"azuread_service_principal_credential_type": {
 				Type:     schema.TypeString,
 				Required: true,
 				ValidateFunc: validation.StringInSlice([]string{
@@ -184,9 +184,9 @@ func resourceArmDataFactoryLinkedServiceODataCreateUpdate(d *schema.ResourceData
 
 	if authenticationType == string(datafactory.ODataAuthenticationTypeAadServicePrincipal) {
 		servicePrincipalId := d.Get("service_principal_id").(string)
-		aadServicePrincipalCredentialType := d.Get("aad_service_principal_credential_type").(string)
+		aadServicePrincipalCredentialType := d.Get("azuread_service_principal_credential_type").(string)
 		tenant := d.Get("tenant").(string)
-		aadResourceID := d.Get("aad_resource_id").(string)
+		aadResourceID := d.Get("azuread_resource_id").(string)
 
 		if aadServicePrincipalCredentialType == string(datafactory.ServicePrincipalCert) {
 			servicePrincipalEmbeddedCertSecureString := datafactory.SecureString{
@@ -329,10 +329,10 @@ func resourceArmDataFactoryLinkedServiceODataRead(d *schema.ResourceData, meta i
 		d.Set("password", props.Password)
 	}
 	if props.AuthenticationType == datafactory.ODataAuthenticationTypeAadServicePrincipal {
-		d.Set("aad_resource_id", props.AadResourceID)
+		d.Set("azuread_resource_id", props.AadResourceID)
 		d.Set("tenant", props.Tenant)
 		d.Set("service_principal_id", props.ServicePrincipalID)
-		d.Set("aad_service_principal_credential_type", props.AadServicePrincipalCredentialType)
+		d.Set("azuread_service_principal_credential_type", props.AadServicePrincipalCredentialType)
 		switch props.AadServicePrincipalCredentialType {
 		case datafactory.ServicePrincipalCert:
 			d.Set("service_principal_embedded_cert", props.ServicePrincipalEmbeddedCert)
@@ -340,7 +340,7 @@ func resourceArmDataFactoryLinkedServiceODataRead(d *schema.ResourceData, meta i
 		case datafactory.ServicePrincipalKey:
 			d.Set("service_principal_key", props.ServicePrincipalKey)
 		default:
-			return fmt.Errorf("Unsupported `aad_service_principal_credential_type`: %+v", props.AadServicePrincipalCredentialType)
+			return fmt.Errorf("Unsupported `azuread_service_principal_credential_type`: %+v", props.AadServicePrincipalCredentialType)
 		}
 	}
 
--- a/azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
+++ b/azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
@@ -153,8 +153,7 @@ func resourceArmDataFactoryLinkedServiceODataCreateUpdate(d *schema.ResourceData
 		odataLinkedService.ODataLinkedServiceTypeProperties = servicePrincipalAuthProperties
 	}
 
-	if authenticationType == "Anonymous" {
-		// TODO string compare with datafactory.ODataAuthenticationTypeAnonymous
+	if authenticationType == string(datafactory.ODataAuthenticationTypeAnonymous) {
 		anonAuthProperties := &datafactory.ODataLinkedServiceTypeProperties{
 			AuthenticationType: datafactory.ODataAuthenticationType(authenticationType),
 			URL:                utils.String(url),
@@ -162,8 +161,7 @@ func resourceArmDataFactoryLinkedServiceODataCreateUpdate(d *schema.ResourceData
 		odataLinkedService.ODataLinkedServiceTypeProperties = anonAuthProperties
 	}
 
-	if authenticationType == "Basic" || authenticationType == "Windows" {
-		// TODO string compare with datafactory.ODataAuthenticationTypeBasic
+	if authenticationType == string(datafactory.ODataAuthenticationTypeBasic) || authenticationType == string(datafactory.ODataAuthenticationTypeWindows) {
 		username := d.Get("username").(string)
 		password := d.Get("password").(string)
 		passwordSecureString := datafactory.SecureString{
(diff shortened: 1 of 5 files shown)


azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
```diff
@@ -0,0 +1,319 @@
...
+
+	odataLinkedService := &datafactory.ODataLinkedService{
+		Description: &description,
+		Type:        datafactory.TypeOData,
+	}
+
+	url := d.Get("url").(string)
+	authenticationType := d.Get("authentication_type").(string)
+
+	if authenticationType == "AadServicePrincipal" {
```
tombuildsstuff, line 135: we can use the constant here:

```suggestion
	if authenticationType == string(datafactory.ODataAuthenticationTypeAadServicePrincipal) {
```

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
```diff
@@ -0,0 +1,319 @@
...
+		servicePrincipalId := d.Get("service_principal_id").(string)
+		aadServicePrincipalCredentialType := d.Get("aad_service_principal_credential_type").(string)
+		tenant := d.Get("tenant").(string)
+		aadResourceID := d.Get("aad_resource_id").(string)
+		servicePrincipalKeySecureString := datafactory.SecureString{
+			Value: utils.String(d.Get("service_principal_key").(string)),
```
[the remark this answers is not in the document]
pomverte: done

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
```diff
@@ -0,0 +1,319 @@
...
+		Type:        datafactory.TypeOData,
+	}
+
+	url := d.Get("url").(string)
+	authenticationType := d.Get("authentication_type").(string)
+
+	if authenticationType == "AadServicePrincipal" {
+		servicePrincipalId := d.Get("service_principal_id").(string)
+		aadServicePrincipalCredentialType := d.Get("aad_service_principal_credential_type").(string)
+		tenant := d.Get("tenant").(string)
```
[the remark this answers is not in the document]
pomverte: done

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
```diff
@@ -0,0 +1,319 @@
...
+		Description: &description,
+		Type:        datafactory.TypeOData,
+	}
+
+	url := d.Get("url").(string)
+	authenticationType := d.Get("authentication_type").(string)
+
+	if authenticationType == "AadServicePrincipal" {
+		servicePrincipalId := d.Get("service_principal_id").(string)
+		aadServicePrincipalCredentialType := d.Get("aad_service_principal_credential_type").(string)
```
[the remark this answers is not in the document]
pomverte: done

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
```diff
@@ -0,0 +1,319 @@
...
+	odataLinkedService := &datafactory.ODataLinkedService{
+		Description: &description,
+		Type:        datafactory.TypeOData,
+	}
+
+	url := d.Get("url").(string)
+	authenticationType := d.Get("authentication_type").(string)
+
+	if authenticationType == "AadServicePrincipal" {
+		servicePrincipalId := d.Get("service_principal_id").(string)
```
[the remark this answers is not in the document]
pomverte: done

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
```diff
@@ -0,0 +1,319 @@
...
+		aadServicePrincipalCredentialType := d.Get("aad_service_principal_credential_type").(string)
+		tenant := d.Get("tenant").(string)
+		aadResourceID := d.Get("aad_resource_id").(string)
```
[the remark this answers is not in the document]
pomverte: done

pomverte:
Hello @tombuildsstuff , do you have any chance of reviewing this PR ?

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
```
katbyte, line 14: Could we also write a complete test that sets all properties? and then an upgrade test that goes from a basic config -> complete -> basic ?

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
```diff
@@ -0,0 +1,390 @@
...
+			},
+
+			"password": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Sensitive:    true,
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
+
+			"aad_resource_id": {
```
katbyte, line 85: Could we order these optional properties alphabetically? schema should be required -> optional -> computed which each  type also being alphabetically ordered

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
```diff
@@ -0,0 +1,390 @@
...
+				ValidateFunc: validation.IsUUID,
+			},
+
+			"service_principal_id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.IsUUID,
+			},
+
+			"aad_service_principal_credential_type": {
```
katbyte, line 103: Could we expand aad to be clear what the acronym means?
```suggestion
			"azuread_service_principal_credential_type": {
```

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
katbyte, line 85: and make this
```suggestion
			"azuread_resource_id": {
```

website/docs/r/data_factory_linked_service_odata.html.markdown
```diff
@@ -0,0 +1,126 @@
...
+
+* `additional_properties` - (Optional) A map of additional properties to associate with the Data Factory Linked Service.
+
+The following supported arguments are specific to OData Linked Service:
+
+* `authentication_type` - (Required) The type of authentication used to connect to the OData source. Valid options are `AadServicePrincipal`, `Anonymous`, `Basic` and `Windows` (`ManagedServiceIdentity` are is supported at this moment by this provider)
+
+* `url` - (Required) The URL of the OData service endpoint (e.g. https://services.odata.org/v4/TripPinServiceRW/People).
+
+* `username` - (Optional) The username which can be used to authenticate to the OData endpoint.
```
katbyte, line 87: would this make more sense as
```suggestion
* `odata_username` - (Optional) The username which can be used to authenticate to the OData endpoint.
```
or
```suggestion
* `basic_username` - (Optional) The username which can be used to authenticate to the OData endpoint.
```

as well as stating it is only applicable when type is basic?

website/docs/r/data_factory_linked_service_odata.html.markdown
```diff
@@ -0,0 +1,126 @@
...
+
+* `password` - (Optional) The password associated with the username, which can be used to authenticate to the OData endpoint.
```
katbyte, line 89: and
```suggestion
* `odata_password` - (Optional) The password associated with the username, which can be used to authenticate to the OData endpoint.
```
or 
```suggestion
* `basic_password` - (Optional) The password associated with the username, which can be used to authenticate to the OData endpoint.
```

website/docs/r/data_factory_linked_service_odata.html.markdown
```diff
@@ -0,0 +1,126 @@
...
+
+* `service_principal_id` - (Optional) The Azure Active Directory Service Principal ID.
```
katbyte, line 91: given this is only applicable to azuread should this be
```suggestion
* `azuread_service_principal_id` - (Optional) The Azure Active Directory Service Principal ID.
```

alternatively we could drop aad from the blow property, or put them in an azuread block?

katbyte (changes requested):
Thanks for the changes @pomverte, I've given this another review. This needs a complete test and an update test that set all optional properties. Also i think some of the properties could do with some better naming - aad is more clear as azuread, as well as making it clear which properties are required for basic, and which for azuread

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
pomverte: I agree with seperating required -> optional -> computed.

I'm not sure what is the value with alphabetic ordering ?
Properties are ordered in a coherent way with the the service functionalities when exploring the Azure Portal

website/docs/r/data_factory_linked_service_odata.html.markdown
pomverte: Service Principal are always in Azure AD
Looking at the code `principal_id` is always declared this way in other files of the project
I'd rather keep coherence within the project

website/docs/r/data_factory_linked_service_odata.html.markdown
pomverte: IMHO
- `odata_password ` is redundant since this is a OData Linked Service
- `basic_password ` would suggest that is another type of password

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
pomverte: Thanks for the suggestion
It's done ✅

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
pomverte: Thanks for the suggestion
It's done ✅

website/docs/r/data_factory_linked_service_odata.html.markdown
pomverte: IMHO
- `odata_password ` is redundant since this is a OData Linked Service
- `basic_password ` would suggest that is another type of password

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource_test.go
pomverte: How are the integration tests run ? I don't have information concerning an OData endpoint open with othere type of authentication other than anonyme and basic auth

pomverte:
Thank you @katbyte for your review, I've update the PR and some discussions are still open according to me

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
```diff
@@ -0,0 +1,390 @@
...
+				}, false),
+			},
+
+			"url": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
+
+			"username": {
```
mbfrahry, line 72: I think this is going to cause headaches for users when trying to figure which values to pass in based on the authentication type they want to use. I think we can better help users there by doing something like the below for the schema
```
"anonymous_authentication": {
         Type:     schema.TypeList,
         MaxItems: 1,
	Optional: true,
        ExactlyOneOf: []string{"the", "other", "authentication", "methods"},
	Elem: &schema.Resource{
                 // fill out with the other details here
},
// Fill out the other authentication methods 
"service_principal_authentication": {
  ...
```

From that model, a user should be able to quickly see how best to setup authentication for this resource. What do you think?

azurerm/internal/services/datafactory/data_factory_linked_service_odata_resource.go
```diff
@@ -0,0 +1,390 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.DataFactoryName(),
+			},
+
+			// There's a bug in the Azure API where this is returned in lower-case
+			// BUG: https://github.com/Azure/azure-rest-api-specs/issues/5788
+			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
+
+			"authentication_type": {
```
mbfrahry, line 55: If we take the below model, we won't need this attribute as we can just infer the type based on what the user passes in.

mbfrahry (changes requested):
Hey @pomverte. This review is a bit of a doozy because it'll require a significant rewrite but these changes should better help users manage this resource. The main issue I have is that, at the moment, it isn't well known what attributes are needed for what authentication type and if a user submits an attribute not related to the authentication type, we'll have a permanent diff as Terraform tries to make that attribute work. What I propose instead, is that we make specific authentication_type blocks that will house the attributes needed to authenticate correctly whilst making sure users only specify one authentication method. Let me know if that makes sense or if you have other questions/concerns.

katbyte:
Hey @pomverte - as this has been waiting for a response since oct i am going to close this. Happy to continuing reviewing it once mathews comments are addressed!

PR #7713, opened 2020-07-13
azurerm_virtual_hub: fix acc test and make the `address_prefix` as `ForceNew`

magodo:
`address_prefix` is not updatable at least since we do the project layout refactor (as the TC test history shows). So I just mark it as `ForceNew` and remove the update test case for vhub (while there are other separate update test cases).


tombuildsstuff:
Screenshot 2020-07-13 at 17 46 51

PR #7754, opened 2020-07-15
Update validation for timezone of azurerm_automation_schedule

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7735

After checked, seems api behavior has been changed. Currently seems timezone display name isn't supported anymore. We have to use timezone value. Get more available timezone list via https://s2.automation.ext.azure.com/api/Orchestrator/TimeZones?_=1594792230258
So I made a fix for this issue.

image


jackofallops:
Tests passing:
image

PR #7764, opened 2020-07-15
Plugin SDK v2

kmoe:
This is a draft PR for convenience and to trigger TC builds.


tombuildsstuff:
Closing since we should be able to build by branch instead if this is still necessary

PR #7776, opened 2020-07-16
nil check route filter before set in express_route_circuit

jackofallops:


azurerm/internal/services/network/express_route_circuit_peering_resource.go
```diff
@@ -251,7 +251,12 @@ func resourceArmExpressRouteCircuitPeeringRead(d *schema.ResourceData, meta inte
 		d.Set("primary_peer_address_prefix", props.PrimaryPeerAddressPrefix)
 		d.Set("secondary_peer_address_prefix", props.SecondaryPeerAddressPrefix)
 		d.Set("vlan_id", props.VlanID)
-		d.Set("route_filter_id", props.RouteFilter.ID)
+
+		routeFilterId := ""
+		if props.RouteFilter != nil {
```
tombuildsstuff, line 256:
```suggestion
		if props.RouteFilter != nil && props.RouteFilter.ID != nil {
```

jackofallops:
Test passing (1 transient failure)
image

Issue #7530, opened 2020-06-30
Support for update existing AKS to enable uptime SLA


dhananjaya-senanayake:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected Resource(s)

* azurerm_kubernetes_cluster

### Potential Terraform Configuration

```hcl
resource "azurerm_kubernetes_cluster" "example" {
  name                = "example-aks1"
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name
  dns_prefix          = "exampleaks1"
  sku_tier            = "Paid"

  default_node_pool {
    name       = "default"
    node_count = 1
    vm_size    = "Standard_D2_v2"
  }
}
```
default value for sku_tier is `Free`. In terraform changing sku_tier from `Free` to `Paid` forces a new resource to be created. https://www.terraform.io/docs/providers/azurerm/r/kubernetes_cluster.html#sku_tier. Now Azure AKS supports to upgrade existing clusters to Uptime SLA paid SKU.

### References

* https://docs.microsoft.com/en-us/azure/aks/uptime-sla#modify-an-existing-cluster-to-use-uptime-sla
* https://github.com/Azure/AKS/issues/1603#issuecomment-651335745

asubmani:
I am facing the same issue. Using terraform: **0.12.29**  and 1azurerm_provider` : **2.20.0** 

## BELOW WORKAROUND WORKS:

1.  Modify your terraform .tf files to add `sku_tier = paid`
2.  Use `az cli` commands to set `uptime-sla`
2.  Run terraform plan and then apply

## Actual Commands:

```py
az aks update --resource-group $RG --name $AKS --uptime-sla
```
```hcl
terraform plan -var-file="stg-terraform.tfvars" -out='stg.plan' -parallelism=200
terraform apply stg.plan
```

> NOTE: DO NOT run `terraform apply`  if `terraform plan` reports any change.

PR #7790, opened 2020-07-16
Removed ForceNew=true on the sku_tier parameter definition

blame19:
Addresses [#7530](https://github.com/terraform-providers/terraform-provider-azurerm/issues/7530)

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_cluster_resource.go
+++ b/azurerm/internal/services/containers/kubernetes_cluster_resource.go
@@ -509,11 +509,6 @@ func resourceArmKubernetesCluster() *schema.Resource {
 			"sku_tier": {
 				Type:     schema.TypeString,
 				Optional: true,
-				// @tombuildsstuff (2020-05-29) - Preview limitations:
-				//  * Currently, cannot convert as existing cluster to enable the Uptime SLA.
-				//  * Currently, there is no way to remove Uptime SLA from an AKS cluster after creation with it enabled.
-				//  * Private clusters aren't currently supported.
-				ForceNew: true,
 				Default:  string(containerservice.Free),
 				ValidateFunc: validation.StringInSlice([]string{
 					string(containerservice.Free),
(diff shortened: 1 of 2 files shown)


djsly:
@jackofallops is this something we could expect the azurerm folks to tackle if you think its a bit more complex then just removing the forced flag ?

From an SDK point of view, it seems that they are simply accepting update in place.

jackofallops:
> @jackofallops is this something we could expect the azurerm folks to tackle if you think its a bit more complex then just removing the forced flag ?
> 
> From an SDK point of view, it seems that they are simply accepting update in place.

@djsly - I suspect this is probably the case, but don't like to assume.  The flag is quite likely just that, and has little or no impact on the operation of the cluster, but will need working into the `update` func, which is somewhat more involved than most.  I can't commit to when we'd be able to look at this, but we can take it on if @blame19 feels it's more than they're comfortable with.

blame19:
Just a small update on this. 
I went ahead and tried the custom provider without the forceNew tag on the sku_tier. 
The tag per se works as intended, 
image
Yet according to terraform plan the cluster should be recreated:
image
Needless to say, I didn't change my ssh public key between runs. There might be something here I'm not aware of regarding the sku_tier upgrade.

@jackofallops would you happen to have any insight on that?

jackofallops:
Hi @blame19 
The plan doesn't actually make any changes. It just shows what the expected output will be without interacting with the service. Removing the `ForceNew` flag from a parameter means that the plan expects the `update` function to deal with change to the value.  The `update` function does not currently send any value for that parameter, so the `sku_tier` won't actually be changed, and the result will be an inconsistent state after an apply.  Does the above make sense?

Microsoft have requested we take a look at what's involved in supporting this for one of their customers, so I'm investigating it today.

jackofallops:
Update: Since this change is on the `master` branch of your fork I'm not able to safely push changes to your copy so I've created a new branch and PR'd from there. Since you started the work, I'll add you as a co-author when this is merged, but I'll be closing this PR as a result.
Thanks for the contribution.

Issue #6030, opened 2020-03-08
Support for Free Tier on CosmosDB


michaelmsm89:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

CosmosDb now offers a free tier, where they give 400 RU/s and 5 GB of storage for free in an account. You can enable free tier on up to one account per subscription. Is it possible to support this? I am using Pulumi and raised the issue there https://github.com/pulumi/pulumi-azure/issues/480 and was told to raise it here.

I've added the Azure ARM template below as reference to the enableFreeTier property:

```json
{
    "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
        "name": {
            "type": "string"
        },
        "location": {
            "type": "string"
        },
        "locationName": {
            "type": "string"
        },
        "defaultExperience": {
            "type": "string"
        },
        "isZoneRedundant": {
            "type": "string"
        }
    },
    "variables": {},
    "resources": [
        {
            "apiVersion": "2019-12-12",
            "kind": "GlobalDocumentDB",
            "type": "Microsoft.DocumentDb/databaseAccounts",
            "name": "[parameters('name')]",
            "location": "[parameters('location')]",
            "properties": {
                "databaseAccountOfferType": "Standard",
                "locations": [
                    {
                        "id": "[concat(parameters('name'), '-', parameters('location'))]",
                        "failoverPriority": 0,
                        "locationName": "[parameters('locationName')]"
                    }
                ],
                "enableMultipleWriteLocations": false,
                "isVirtualNetworkFilterEnabled": false,
                "virtualNetworkRules": [],
                "ipRangeFilter": "",
                "dependsOn": [],
                "capabilities": [],
                "enableFreeTier": true
            },
            "tags": {
                "defaultExperience": "[parameters('defaultExperience')]",
                "hidden-cosmos-mmspecial": "",
                "CosmosAccountType": "Non-Production"
            }
        }
    ],
    "outputs": {}
}
```

### New or Affected Resource(s)

* azurerm_cosmosdb

[...]

PR #7597, opened 2020-07-06
Update Azure Cosmos DB API to to 2020-04-01 from 2015-04-08

marc-sensenich:
Forked from #6253 and brings the Cosmos DB API up to 2020-04-01

Closes #6253

Acceptance Test Statuses (65/68 passing)
```
--- PASS: TestAccDataSourceAzureRMCosmosDBAccount_basic (1215.08s)
--- PASS: TestAccAzureRMCosmosDbMongoCollection_withIndex (1245.88s)
--- PASS: TestAccAzureRMCosmosDbMongoDatabase_complete (1364.02s)
--- PASS: TestAccAzureRMCosmosDbMongoDatabase_basic (1365.37s)
--- PASS: TestAccAzureRMCosmosDbSqlDatabase_update (1443.23s)
--- PASS: TestAccAzureRMCosmosDbSqlContainer_basic (1549.32s)
--- PASS: TestAccAzureRMCosmosDbMongoCollection_throughput (1585.49s)
--- PASS: TestAccAzureRMCosmosDbMongoCollection_complete (1425.26s)
--- PASS: TestAccAzureRMCosmosDbMongoCollection_update (1493.49s)
--- PASS: TestAccAzureRMCosmosDbMongoCollection_basic (1428.51s)
--- PASS: TestAccAzureRMCosmosDbGremlinGraph_customConflictResolutionPolicy (1346.60s)
--- PASS: TestAccAzureRMCosmosDbGremlinGraph_indexPolicy (1455.86s)
--- PASS: TestAccAzureRMCosmosDbGremlinGraph_update (1754.62s)
--- PASS: TestAccAzureRMCosmosDbGremlinGraph_requiresImport (1580.51s)
--- PASS: TestAccAzureRMCosmosGremlinDatabase_complete (1471.62s)
--- PASS: TestAccAzureRMCosmosDbGremlinGraph_basic (1515.33s)
--- PASS: TestAccAzureRMCosmosGremlinDatabase_requiresImport (1408.04s)
--- PASS: TestAccAzureRMCosmosGremlinDatabase_basic (1452.88s)
--- PASS: TestAccAzureRMCosmosDbCassandraKeyspace_complete (1275.45s)
--- PASS: TestAccAzureRMCosmosDbCassandraKeyspace_update (1591.40s)
--- PASS: TestAccAzureRMCosmosDbCassandraKeyspace_basic (1332.76s)
--- PASS: TestAccAzureRMCosmosDBAccount_vNetFilters (1244.61s)
[...]

PR #7814, opened 2020-07-20
`azurerm_cosmos_db_account` - add support for `enable_free_tier`

giotab:
Azure Cosmos DB API 2020-04-01 allows to set the Free Tier option using EnableFreeTier.

After introducing the new version in #7597, it is possible to solve #6030 and allow Free Cosmos DB pricing option.


website/docs/r/cosmosdb_account.html.markdown
```diff
@@ -73,6 +75,8 @@ The following arguments are supported:
 
 * `ip_range_filter` - (Optional) CosmosDB Firewall Support: This value specifies the set of IP addresses or IP address ranges in CIDR form to be included as the allowed list of client IP's for a given database account. IP addresses/ranges must be comma separated and must not contain any spaces.
 
+* `enable_free_tier` - (Optional) Enable Free Tier pricing option for this Cosmos DB account.
```
mbfrahry, line 78:
```suggestion
* `enable_free_tier` - (Optional) Enable Free Tier pricing option for this Cosmos DB account. Defaults to `false`
```

website/docs/r/cosmosdb_account.html.markdown
```diff
@@ -30,6 +30,8 @@ resource "azurerm_cosmosdb_account" "db" {
   offer_type          = "Standard"
   kind                = "GlobalDocumentDB"
 
+  enable_free_tier    = false
```
mbfrahry, line 33: Because it defaults to `false`, we can remove this

examples/cosmos-db/basic/main.tf
```diff
@@ -13,6 +13,7 @@ resource "azurerm_cosmosdb_account" "example" {
   resource_group_name = "${azurerm_resource_group.example.name}"
   offer_type          = "Standard"
   kind                = "GlobalDocumentDB"
+  enable_free_tier    = false
```
mbfrahry, line 16:
```suggestion
```

azurerm/internal/services/cosmos/tests/cosmosdb_account_resource_test.go
```diff
@@ -786,6 +786,8 @@ resource "azurerm_cosmosdb_account" "test" {
   resource_group_name = azurerm_resource_group.test.name
   offer_type          = "Standard"
   kind                = "GlobalDocumentDB"
+  
+  enable_free_tier    = false
```
mbfrahry, line 790: We can remove this because it defaults to `false`. Instead, we should add a test to enable/disable the free tier. If we can't disable it, we need to add `ForceNew: true` to the schema declaration

azurerm/internal/services/cosmos/cosmosdb_account_resource.go
```diff
@@ -542,6 +552,10 @@ func resourceArmCosmosDbAccountRead(d *schema.ResourceData, meta interface{}) er
 		d.Set("is_virtual_network_filter_enabled", resp.IsVirtualNetworkFilterEnabled)
 	}
 
+	if v := resp.EnableFreeTier; v != nil {
```
mbfrahry, line 555: Because we aren't pointing to the value, we can just set it without doing a nil check
```suggestion
```

azurerm/internal/services/cosmos/tests/cosmosdb_account_resource_test.go
giotab: It is possible to have only one Free Tier account and I do not think it is possible to change it. Hence, I added the ForceNew to the schema and I introduced a new test for this. I appreciate your feedback!

vendor/github.com/Azure/azure-sdk-for-go/services/cosmos-db/mgmt/2020-04-01/documentdb/models.go
```diff
@@ -1312,6 +1312,8 @@ type DatabaseAccountCreateUpdateProperties struct {
 	IPRules *[]IPAddressOrRange `json:"ipRules,omitempty"`
 	// IsVirtualNetworkFilterEnabled - Flag to indicate whether to enable/disable Virtual Network ACL rules.
 	IsVirtualNetworkFilterEnabled *bool `json:"isVirtualNetworkFilterEnabled,omitempty"`
+	// EnableFreeTier - Enables Free Tier pricing option for the Cosmos DB account.
+	EnableFreeTier *bool `json:"enableFreeTier,omitempty"`
```
mbfrahry, line 1316: It looks like these might have been manually added. They already exist in the structs and are causing issues with building the provider. We can run `go mod vendor` and it should fix it up

mbfrahry:
Thanks for this @giotab!

Fixes #6030

PR #7823, opened 2020-07-21
add notes about `skip_credentials_validation`

### LINKED ISSUE #7089: The flag "skip_provider_registration" should also avoid reading the provider list
### Community Note

AzureRM 2.11

### Expected Behavior

I need to use the flag "skip_credentials_validation" to avoid the resource provider list to be read. I should use the flag "skip_provider_registration".

### Actual Behavior
As the doc stated (https://www.terraform.io/docs/providers/azurerm/index.html#skip_provider_registration) this flag is here for environments with restricted permissions. My teams does not any role at the subscription level, so no way for them to read the provider list. Even with the flag "skip_provider_registration", the script failed because the provider list is still read:

> **Error**: Unable to list provider registration status, it is possible that this is due to invalid credentials or the service principal does not have permission to use the Resource Manager API, Azure error: resources.ProvidersClient#List: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="AuthorizationFailed" Message="The client 'XXXXXXXXX' with object id 'YYYYYYYYYYY' does not have authorization to perform action 'Microsoft.Resources/subscriptions/providers/read' over scope '/subscriptions/ZZZZZZZZZZZZZZ' or the scope is invalid. If access was recently granted, please refresh your credentials."

To avoid reading the provider list I need to use the flag "skip_credentials_validation" because it removes the query on the provider list.

### References

Source code : https://github.com/terraform-providers/terraform-provider-azurerm/blob/8174fd51cada9a1fd982d2e7e9fe36a5d70ce2b1/azurerm/internal/provider/provider.go

njuCZ:
fix #7089

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -137,6 +137,8 @@ For some advanced scenarios, such as where more granular permissions are necessa
 
 * `skip_credentials_validation` - (Optional) Should the AzureRM Provider skip verifying the credentials being used are valid? This can also be sourced from the `ARM_SKIP_CREDENTIALS_VALIDATION` Environment Variable. Defaults to `false`.
 
+~> **Note:** if `skip_credentials_validation` is false, AzureRM Provider will send request to list all available providers.
+
 * `skip_provider_registration` - (Optional) Should the AzureRM Provider skip registering the Resource Providers it supports? This can also be sourced from the `ARM_SKIP_PROVIDER_REGISTRATION` Environment Variable. Defaults to `false`.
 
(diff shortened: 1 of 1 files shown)


website/docs/index.html.markdown
```diff
@@ -137,6 +137,8 @@ For some advanced scenarios, such as where more granular permissions are necessa
 
 * `skip_credentials_validation` - (Optional) Should the AzureRM Provider skip verifying the credentials being used are valid? This can also be sourced from the `ARM_SKIP_CREDENTIALS_VALIDATION` Environment Variable. Defaults to `false`.
 
+~> **Note:** if `skip_credentials_validation` is false, AzureRM Provider will send request to list all available providers.
+
```
katbyte, line 141:
```suggestion
* `skip_credentials_validation` - (Optional) Should the AzureRM Provider skip verifying the credentials being used are valid? If `false` the provider will send a request to list all available Resource Providers. This can also be sourced from the `ARM_SKIP_CREDENTIALS_VALIDATION` Environment Variable. Defaults to `false`.

```

Issue #7250, opened 2020-06-08
Replace conflict setting of availability_set and zone on virtual_machine resources


gek0:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected Resource(s)

update of existing resources
* azurerm_linux_virtual_machine
* azurerm_windows_virtual_machine

### Provider code
Linux VM
* https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/compute/linux_virtual_machine_resource.go#L118
* https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/compute/linux_virtual_machine_resource.go#L240

Windows VM
* https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/compute/windows_virtual_machine_resource.go#L120
* https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/compute/windows_virtual_machine_resource.go#L261

### Potential Terraform Configuration

```hcl
# example deployment of VMs in AZ setup - allow setting null/empty AS ID
resource "azurerm_*_virtual_machine" "example" {
    ...
    availability_set_id = null / ""
    zone = "<ZONE_ID>"
    ...
}

# or if current does not support AZ setup - fallback to AS for HA and set AZ to null/blank
resource "azurerm_*_virtual_machine" "example" {
    ...
    availability_set_id = azurerm_availability_set.example.id
    zone = null / ""
    ...
}
```

### Use case
Fallback to Availability Set deployment if Availability Zone redundancy is not available in current location/region.
We are trying to write a module for deploying virtual machines in Azure and to write it as DRY as possible.
With current limitations - `availability_set_id` and `zone` attributes are conflicted and cannot be defined under the same resource type. Neither as dynamic content and neither as null/blank string/boolean (as false) so that only one is used and others ignored.

I can understand that this is a limitation on Azure API, but I would like to be able to define both attributes - with only one of them being set to some value at a time. This way we wouldn't need to define the same `azurerm_*_virtual_machine` resource with repeated attributes

[...]

PR #7826, opened 2020-07-21
Issue 7250

gek0:
Relates to feature request #7250


tombuildsstuff:
hey @gek0 

Thanks for this contribution :)

Whilst the changes proposed in this PR would allow setting these values to empty strings, Terraform instead uses the value `null` to represent an unset value (and an empty string to represent an empty string).

That said it should be possible to achieve this today using a coalesce statement ([as shown in this comment](https://github.com/terraform-providers/terraform-provider-azurerm/issues/7250#issuecomment-669832818)) - to conditionally define both of these fields setting the value to `null` if they're unset, achieving the same thing.

As such whilst I'd like to thank you for this contribution, I'm going to close this behaviour for the moment - since the values for `null` and an empty string `""` intentionally have different semantics.

Thanks!

PR #7841, opened 2020-07-22
update key vault key and secret id when reading

### LINKED ISSUE #6222: The provider does not update Key Vault Secret tags when there are only tag changes/content type
### Community Note

* Started a long time ago and still exist.
Terraform v0.12.23
AzureRM 2.2.0

### Affected Resource(s)

* `azurerm_key_vault_secret`

Terraform Plan Output;
```hcl
  # azurerm_key_vault_secret.my_secret will be updated in-place
  ~ resource "azurerm_key_vault_secret" "my_secret" {
      + content_type = "text/plain"
        id           = "https://***.vault.azure.net/secrets/MySecret/version_number_1_target"
        key_vault_id = "/subscriptions/***/resourceGroups/***/providers/Microsoft.KeyVault/vaults/my_kv"
        name         = "MySecret"
      ~ tags         = {
          + "tag1" = "tag_value_1"
          + "tag2" = "tag_value_2"
          + "tag3" = "tag_value_3"
          + "tag4" = "tag_value_4"
          + "tag5" = "tag_value_5"
          + "tag6" = "tag_value_6"
        }
        value        = (sensitive value)
        version      = "version_number_2_different"
    }
```

### Expected Behavior

The provider should update tags and the content-type without changing the secret's value.

### Actual Behavior

The provider just does not update Key Vault Secret tags or content type without any value changes. It's only applying changes when the value changed. 

### Steps to Reproduce

1. Add a new key vault and a secret with tags.
2. `terraform plan`
3. `terraform apply`
4. Without changing the secret's value, change one of the tag values or tag names or just content type.
5. `terraform plan` 
6. See changes in the plan output.
7. `terraform apply`
8. Nothing changed.

njuCZ:
fix #6222 

the reason of this issue is:
1. user create a new version secret via the portal
2. terraform plan will read the latest secret version and find tags is missing
3. terraform update will only update the tags of `id.version`. 

Because id is not updated, so it's not updating the latest version. The next Time terraform plan it will show diff again.

I think key vault key also have such problem

### DIFF
--- a/azurerm/internal/services/keyvault/key_vault_key_resource.go
+++ b/azurerm/internal/services/keyvault/key_vault_key_resource.go
@@ -362,8 +362,20 @@ func resourceArmKeyVaultKeyRead(d *schema.ResourceData, meta interface{}) error 
 		return err
 	}
 
-	d.Set("name", id.Name)
+	if resp.Key == nil || resp.Key.Kid == nil || *resp.Key.Kid == "" {
+		return fmt.Errorf("empty or nil ID returned for Azure KeyVault Secret %q", id.Name)
+	}
+	// the version may have changed, so parse the updated id
+	respID, err := azure.ParseKeyVaultChildID(*resp.Key.Kid)
+	if err != nil {
+		return err
+	}
+	// the version may have changed, we are comparing diff with the latest version. So set the id
+	d.SetId(*resp.Key.Kid)
 
+	d.Set("name", respID.Name)
+	// Computed
+	d.Set("version", respID.Version)
 	if key := resp.Key; key != nil {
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
hey @njuCZ 

Thanks for this PR.

Updating the ID of a resource during a Read isn't a recommended practice (since this can cause graph changes, which requires multiple Terraform runs) - as such I've taken a look into this and it appears that instead we can update the latest version rather than this version specifically, always pulling the version out during the update - which should achieve the same thing here.

As such whilst I'd like to thank you for opening this PR I'm going to close this in favour of #8304 which solves this problem in a slightly different way.

Thanks!

PR #7847, opened 2020-07-22
Remove unsupported sku from iothub dps

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7242

### DIFF
--- a/website/docs/r/iothub_dps.html.markdown
+++ b/website/docs/r/iothub_dps.html.markdown
@@ -50,7 +50,7 @@ The following arguments are supported:
 
 A `sku` block supports the following:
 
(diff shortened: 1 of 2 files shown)


website/docs/r/iothub_dps.html.markdown
```diff
@@ -50,7 +50,7 @@ The following arguments are supported:
 
 A `sku` block supports the following:
 
-* `name` - (Required) The name of the sku. Possible values are `B1`, `B2`, `B3`, `F1`, `S1`, `S2`, and `S3`.
+* `name` - (Required) The name of the sku. Possible value is `S1`.
```
katbyte, line 53:
```suggestion
* `name` - (Required) The name of the sku. Currently can only be set to `S1`.
```

Issue #7830, opened 2020-07-21
Support for azurerm_mysql_server "Allow access to Azure services"


IanMoroney:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The `azurerm_mysql_server` resource currently does not allow you to set the value "Allow access to Azure services"

### New or Affected Resource(s)

*  `azurerm_mysql_server` 

### Potential Terraform Configuration

```hcl
provider "azurerm" {
  features {}
}

resource "azurerm_resource_group" "example" {
  name     = "example-resources"
  location = "West Europe"
}

resource "azurerm_mysql_server" "example" {
  name                = "example-mysqlserver"
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name

  administrator_login          = "mysqladminun"
  administrator_login_password = "H@Sh1CoR3!"

  sku_name   = "B_Gen5_2"
  storage_mb = 5120
  version    = "5.7"

  auto_grow_enabled                 = true
  backup_retention_days             = 7
  geo_redundant_backup_enabled      = true
  infrastructure_encryption_enabled = true
  public_network_access_enabled     = false
  ssl_enforcement_enabled           = true
  ssl_minimal_tls_version_enforced  = "TLS1_2"
  allow_access_to_azure_services      = true
}
```

### References

* https://docs.microsoft.com/en-us/azure/mysql/howto-connect-webapp#solution-1---allow-azure-services

yupwei68:
Hi @IanMoroney , thanks for opening this issue. Currently it could be implemented by add a firewall ip_address= "0.0.0.0", for example:
```
resource "azurerm_mysql_firewall_rule" "test" {
  name                = "acctestfwrule"
  resource_group_name = azurerm_resource_group.test.name
  server_name         = azurerm_mysql_server.test.name
  start_ip_address    = "0.0.0.0"
  end_ip_address      = "0.0.0.0"
}
```

IanMoroney:
Surely that's the same as having no ip whitelist turned on at all?
This doesn't seem like a very secure solution.

yupwei68:
[...]

PR #7850, opened 2020-07-22
Doc Fix: `azurerm_mysql_firewall_rule` Allow access to Azure services example

IanMoroney:
Updated documentation for azurerm_mysql_firewall_rule to show an example of how to enable the "Allow Access to Azure Services" feature.

This feature was previously undocumented, but has been discussed in #7830 

I have tested this functionality and confirmed that it works as expected.


IanMoroney:
Thanks @mbfrahry , i was just concerned that people would see a 0.0.0.0 rule and automatically think it was an "allow all" rule, which was the reason for the added note.
Hopefully it'll be ok though!

PR #7869, opened 2020-07-23
azurerm_storage_account_network_rules - Note regarding small /31 and /32 IP ranges in ip_rules was added.

arestarh:


website/docs/r/storage_account_network_rules.html.markdown
```diff
@@ -78,6 +78,8 @@ The following arguments are supported:
 
 * `ip_rules` - (Optional) List of public IP or IP ranges in CIDR Format. Only IPV4 addresses are allowed. Private IP address ranges (as defined in [RFC 1918](https://tools.ietf.org/html/rfc1918#section-3)) are not allowed.
 
+-> **NOTE** Small address ranges using "/31" or "/32" prefix sizes are not supported. These ranges should be configured using individual IP address rules without prefix specified.
```
katbyte, line 81:
```suggestion
-> **NOTE** Small address ranges using `/31` or `/32` prefix sizes are not supported. These ranges should be configured using individual IP address rules without prefix specified.
```

PR #7873, opened 2020-07-23
New Resource: `azurerm_api_management_api_diagnostic`

### LINKED ISSUE #4760: Feature request: azurerm_api_management_diagnostic

sirlatrom:
Closes #4760.
Adds support for `azuremonitor` as `identifier` to `azurerm_api_management_diagnostic`

Missing:
  - [x] Tests for the new resource

### DIFF
--- a/azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
@@ -45,6 +45,7 @@ func resourceArmApiManagementApiDiagnostic() *schema.Resource {
 				ForceNew: true,
 				ValidateFunc: validation.StringInSlice([]string{
 					"applicationinsights",
+					"azuremonitor",
 				}, false),
 			},
 
--- a/azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
@@ -111,7 +111,7 @@ func resourceArmApiManagementApiDiagnosticCreateUpdate(d *schema.ResourceData, m
 	}
 	d.SetId(*resp.ID)
 
-	return resourceArmApiManagementDiagnosticRead(d, meta)
+	return resourceArmApiManagementApiDiagnosticRead(d, meta)
 }
 
 func resourceArmApiManagementApiDiagnosticRead(d *schema.ResourceData, meta interface{}) error {
--- a/azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
@@ -138,7 +138,9 @@ func resourceArmApiManagementApiDiagnosticRead(d *schema.ResourceData, meta inte
 	d.Set("identifier", resp.Name)
 	d.Set("resource_group_name", diagnosticId.ResourceGroup)
 	d.Set("api_management_name", diagnosticId.ServiceName)
-	d.Set("api_management_logger_id", resp.LoggerID)
+	if props := resp.DiagnosticContractProperties; props != nil {
+		d.Set("api_management_logger_id", props.LoggerID)
+	}
 
 	return nil
 }
--- a/azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
@@ -135,6 +135,7 @@ func resourceArmApiManagementApiDiagnosticRead(d *schema.ResourceData, meta inte
 		return fmt.Errorf("making Read request for Diagnostic %q (Resource Group %q / API Management Service %q / API %q): %+v", diagnosticId.Name, diagnosticId.ResourceGroup, diagnosticId.ServiceName, diagnosticId.ApiName, err)
 	}
 
+	d.Set("api_name", diagnosticId.ApiName)
 	d.Set("identifier", resp.Name)
 	d.Set("resource_group_name", diagnosticId.ResourceGroup)
 	d.Set("api_management_name", diagnosticId.ServiceName)
--- a/azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
@@ -1,0 +1,168 @@
+package apimanagement
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/apimanagement/mgmt/2019-12-01/apimanagement"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
+	helpersvalidate "github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/apimanagement/parse"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/apimanagement/validate"
+	azSchema "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tf/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func resourceArmApiManagementApiDiagnostic() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceArmApiManagementApiDiagnosticCreateUpdate,
+		Read:   resourceArmApiManagementApiDiagnosticRead,
+		Update: resourceArmApiManagementApiDiagnosticCreateUpdate,
+		Delete: resourceArmApiManagementApiDiagnosticDelete,
+
+		Importer: azSchema.ValidateResourceIDPriorToImport(func(id string) error {
+			_, err := parse.ApiManagementApiDiagnosticID(id)
+			return err
+		}),
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"identifier": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"applicationinsights",
+				}, false),
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"api_management_name": azure.SchemaApiManagementName(),
+
+			"api_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: helpersvalidate.ApiManagementApiName,
+			},
+
+			"api_management_logger_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.ApiManagementLoggerID,
+			},
+
+			"enabled": {
+				Type:       schema.TypeBool,
+				Optional:   true,
+				Deprecated: "this property has been removed from the API and will be removed in version 3.0 of the provider",
+			},
+		},
+	}
+}
+
+func resourceArmApiManagementApiDiagnosticCreateUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).ApiManagement.ApiDiagnosticClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	diagnosticId := d.Get("identifier").(string)
+	resourceGroup := d.Get("resource_group_name").(string)
+	serviceName := d.Get("api_management_name").(string)
+	apiName := d.Get("api_name").(string)
+
+	if d.IsNewResource() {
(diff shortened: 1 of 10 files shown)


hctv19:
Really looking forward to this resource being implemented - we currently have to set this configuration with ARM templates directly and there's a bug with the azurerm_template_deployment resource type that's been causing us some headaches. Thanks for this work @sirlatrom !

jonatasbaldin:
Hi there! Look forward for this implementation too, thanks a lot for the hard work 💪

sirlatrom:
@jackofallops Can this be put on a milestone do you think?

jackofallops:
Hi @sirlatrom 
I started to review this and noticed a bug in the `azurerm_api_management_diagnostic` that I think you may have based this on (since the same bug is present). Specifically the `diagnosticId` and `apiName` parameters of the `CreateOrUpdate` call are in the wrong order.  I'm going need to spend a little time digging into the implications of fixing that before I finish up this review, so apologies in advance if it takes a little while. 

I wasn't involved in the other diagnostic resource (that I remember) - is there a reason that it, and this, doesn't support the other properties of the `DiagnosticContractProperties`? I'd like to understand the intended use-case better, as it currently feels a little lacking?

sirlatrom:
> Hi @sirlatrom
> I started to review this and noticed a bug in the `azurerm_api_management_diagnostic` that I think you may have based this on (since the same bug is present). Specifically the `diagnosticId` and `apiName` parameters of the `CreateOrUpdate` call are in the wrong order. I'm going need to spend a little time digging into the implications of fixing that before I finish up this review, so apologies in advance if it takes a little while.
> 
> I wasn't involved in the other diagnostic resource (that I remember) - is there a reason that it, and this, doesn't support the other properties of the `DiagnosticContractProperties`? I'd like to understand the intended use-case better, as it currently feels a little lacking?

You guessed correctly that I copied and modified the `azurerm_api_management_diagnostic` resource, so I assumed it did things correctly. Since this here is a new resource, I can fix the argument order for that resource in this PR, and `azurerm_api_management_diagnostic` can be fixed in another PR.

I'll get around to making changes and addressing your order concerns on Monday CEST.

jackofallops:
Thanks @sirlatrom - We need to be mindful that fixing the issue doesn't create a breaking change for existing resources, such as by changing ID's, recreating resources unexpectedly etc.  It may be appropriate to have the change still, but knowing the impact in the first instance is important!

Thanks again :)

sirlatrom:
@jackofallops I've now verified that the order of arguments to `apimanagement.APIDiagnosticClient.CreateOrUpdate` is correct after the recently pushed commit c8ab23d.

azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
```diff
@@ -0,0 +1,169 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.ApiManagementLoggerID,
+			},
+
+			"enabled": {
+				Type:       schema.TypeBool,
+				Optional:   true,
+				Deprecated: "this property has been removed from the API and will be removed in version 3.0 of the provider",
+			},
```
jackofallops, line 73: This can be removed. We shouldn't introduce a new resource with an already deprecated / unused property.

azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if _, err := client.CreateOrUpdate(ctx, resourceGroup, serviceName, apiName, diagnosticId, parameters, ""); err != nil {
+		return fmt.Errorf("creating or updating Diagnostic %q (Resource Group %q / API Management Service %q / API %q): %+v", diagnosticId, resourceGroup, serviceName, apiName, err)
+	}
+
+	resp, err := client.Get(ctx, resourceGroup, serviceName, apiName, diagnosticId)
+	if err != nil {
+		return fmt.Errorf("retrieving Diagnostic %q (Resource Group %q / API Management Service %q / API %q): %+v", diagnosticId, resourceGroup, serviceName, apiName, err)
+	}
+	if resp.ID == nil {
```
jackofallops, line 115: Can we also check the ID value is not empty?
```suggestion
	if resp.ID == nil || *resp.ID == "" {
```

website/docs/r/api_management_api_diagnostic.html.markdown
```diff
@@ -0,0 +1,113 @@
...
+* `api_name` - (Required) The name of the API on which to configure the Diagnostics Logs. Changing this forces a new API Management Service API Diagnostics Logs to be created.
+
+* `identifier` - (Required) Identifier of the Diagnostics Logs. Changing this forces a new API Management Service API Diagnostics Logs to be created.
+
+* `resource_group_name` - (Required) The name of the Resource Group where the API Management Service API Diagnostics Logs should exist. Changing this forces a new API Management Service API Diagnostics Logs to be created.
+
+---
+
+* `enabled` - (Optional) Should the Diagnostics Logs be enabled?
+
```
jackofallops, line 91: As above, can be removed.

jackofallops (changes requested):
Hi @sirlatrom 

Thanks for the updates, I've left a few minor comments below, but otherwise LGTM. 

Thanks!

sirlatrom:
@jackofallops I've applied your suggestions, please let me know if any further modifications are needed.

hctv19:
@jackofallops Hi! I was just doing a gentle ping on this one since we are eagerly awaiting this update for our pipeline as well. Thanks!

website/docs/r/api_management_api_diagnostic.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+The following arguments are supported:
+
+* `api_management_logger_id` - (Required) The ID (name) of the Diagnostics Logger.
+
+* `api_management_name` - (Required) The name of the API Management Service instance. Changing this forces a new API Management Service API Diagnostics Logs to be created.
+
+* `api_name` - (Required) The name of the API on which to configure the Diagnostics Logs. Changing this forces a new API Management Service API Diagnostics Logs to be created.
+
+* `identifier` - (Required) Identifier of the Diagnostics Logs. Changing this forces a new API Management Service API Diagnostics Logs to be created.
```
tombuildsstuff, line 84: given there's only two possible values here we should make this:

```suggestion
* `identifier` - (Required) Identifier of the Diagnostics Logs. Possible values are `applicationinsights` and `azuremonitor`. Changing this forces a new API Management Service API Diagnostics Logs to be created.
```

(could we also update the APIM Diagnostic resource too, since that now also supports this value)

Finally - could we also add this to the `example` above, since this is Required?

website/docs/r/api_management_api_diagnostic.html.markdown
```diff
@@ -0,0 +1,109 @@
...
+The `timeouts` block allows you to specify [timeouts](https://www.terraform.io/docs/configuration/resources.html#timeouts) for certain actions:
+
+* `create` - (Defaults to 30 minutes) Used when creating the API Management Service API Diagnostics Logs.
+* `read` - (Defaults to 5 minutes) Used when retrieving the API Management Service API Diagnostics Logs.
+* `update` - (Defaults to 30 minutes) Used when updating the API Management Service API Diagnostics Logs.
+* `delete` - (Defaults to 30 minutes) Used when deleting the API Management Service API Diagnostics Logs.
+
+## Import
+
+API Management Service API Diagnostics Logss can be imported using the `resource id`, e.g.
```
tombuildsstuff, line 105:
```suggestion
API Management Service API Diagnostics Logs can be imported using the `resource id`, e.g.
```

azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go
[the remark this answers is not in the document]
sirlatrom: This was removed in the most recent commit as per @jackofallops's review suggestion.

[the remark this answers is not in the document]
tombuildsstuff: this was from a stale review - sorry 🙃

sirlatrom:
@tombuildsstuff Thank you for your suggestions, I have applied them all.

azurerm/internal/services/apimanagement/client/client.go
```diff
@@ -7,6 +7,7 @@ import (
 
 type Client struct {
 	ApiClient                  *apimanagement.APIClient
+	ApiDiagnosticClient        *apimanagement.APIDiagnosticClient
```
tombuildsstuff, line 10: this client needs to be initialized, at the moment running the tests causes a panic - we can do that below: https://github.com/terraform-providers/terraform-provider-azurerm/pull/7873/files#diff-e334dbeb932a241f1dd03111cd3f4d71R42-R43

Once that's done we should be able to re-run the tests here

tombuildsstuff (dismissed):
hey @sirlatrom 

I've run the tests here but they've crashed since the `ApiDiagnosticClient` isn't being initialized - could we fix this, at which point we should be able to re-run the tests here?

Thanks!

azurerm/internal/services/apimanagement/client/client.go
sirlatrom: Should be done now in e521a8be0.

sirlatrom:
@tombuildsstuff I've added the client initialization in e521a8b.

jackofallops:
Hi @sirlatrom 
I ran the tests, and they all failed with:

```Error: ID contained more segments than required: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/acctestRG-200904094049117016/providers/Microsoft.ApiManagement/service/acctestAM-200904094049117016/apis/acctestAMA-200904094049117016/diagnostics/applicationinsights"```

I couldn't see an immediately obvious problem with your ID parse code, but it's somehow dropping through to the a fail on `ValidateNoEmptySegments`

Could you add a unit test for that parser and investigate?

**Comment by sirlatrom** [comment]
> Hi @sirlatrom
> I ran the tests, and they all failed with:
> 
> `Error: ID contained more segments than required: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/acctestRG-200904094049117016/providers/Microsoft.ApiManagement/service/acctestAM-200904094049117016/apis/acctestAMA-200904094049117016/diagnostics/applicationinsights"`
> 
> I couldn't see an immediately obvious problem with your ID parse code, but it's somehow dropping through to the a fail on `ValidateNoEmptySegments`
> 
> Could you add a unit test for that parser and investigate?

@jackofallops Thanks for finding that. I've added a unit test for that parser in 936bc51, and in 
24f1206 suggested that the `id.Path` variable that is checked in `ValidateNoEmptySegments` is included in the error message, to help any future recipients of that error message.

File: azurerm/internal/services/apimanagement/api_management_api_diagnostic_resource.go [hunk 7]
```diff
@@ -0,0 +1,165 @@
...
+	resp, err := client.Get(ctx, resourceGroup, serviceName, apiName, diagnosticId)
+	if err != nil {
+		return fmt.Errorf("retrieving Diagnostic %q (Resource Group %q / API Management Service %q / API %q): %+v", diagnosticId, resourceGroup, serviceName, apiName, err)
+	}
+	if resp.ID == nil || *resp.ID == "" {
+		return fmt.Errorf("reading ID for Diagnostic %q (Resource Group %q / API Management Service %q / API %q): ID is empty", diagnosticId, resourceGroup, serviceName, apiName)
+	}
+	d.SetId(*resp.ID)
+
+	return resourceArmApiManagementDiagnosticRead(d, meta)
```
jackofallops, line 114:
```suggestion
	return resourceArmApiManagementApiDiagnosticRead(d, meta)
```

azurerm/internal/services/apimanagement/tests/api_management_api_diagnostic_resource_test.go
```diff
@@ -0,0 +1,261 @@
...
+	})
+}
+
+func TestAccAzureRMApiManagementApiDiagnostic_requiresImport(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_api_management_api_diagnostic", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMApiManagementDestroy,
```
jackofallops, line 66:
```suggestion
		CheckDestroy: testCheckAzureRMApiManagementApiDiagnosticDestroy,
```

jackofallops (changes requested):
Hi @sirlatrom 
Thanks for adding the additional unit test, and the additional info to the path error output.  It uncovered the root of the odd issue, which I've commented on below.  (I think they're the only instances, might be worth another quick scan since these resources are so similarly named)

Thanks!

jackofallops (changes requested):
Hi @sirlatrom 
Hope you don't mind, but I pushed a couple fixes this morning to your branch.  The basic and requiresImport are now passing, but on re-review I'm not sure the `update` scenario is valid? Update tests are for in-place changes and there's a change to a `Forcenew` property in that scenario.  The config for that scenario also fails as the `test2` resources have the same names as the `test`, so trigger an import error. Can you review the test and adjust as appropriate?
Thanks.

sirlatrom:
> Hi @sirlatrom
> Hope you don't mind, but I pushed a couple fixes this morning to your branch. The basic and requiresImport are now passing, but on re-review I'm not sure the `update` scenario is valid? Update tests are for in-place changes and there's a change to a `Forcenew` property in that scenario. The config for that scenario also fails as the `test2` resources have the same names as the `test`, so trigger an import error. Can you review the test and adjust as appropriate?
> Thanks.

Hi @jackofallops,
Thanks for doing the work properly when I apparently failed to :smile:
As for the `update` scenario, it _should_ be testing changing the logger ID, but changing the API name should arguably result in a new resource.
I'll try and go through that scenario and make it actually do what it should.

jackofallops (approved):
Thanks @sirlatrom 
This LGTM now, and tests are passing.

image

PR #7897, opened 2020-07-25
Update all deprecated azurerm_subnet address_prefix examples

ibidani:
Updating all references in examples of "azurerm_subnet" from the argument "address_prefix" to "address_prefixes".
per https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/subnet#address_prefix


ibidani:
sorry bad regex

PR #7902, opened 2020-07-26
New Data Source: `azurerm_function_app_host_keys`

### LINKED ISSUE #699: Feature request: export Functions host keys

3mard:
This PR closes #699


azurerm/internal/services/web/data_source_function_app_host_keys.go
```diff
@@ -0,0 +1,86 @@
...
+
+			"function_keys": {
+				Type:     schema.TypeMap,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"system_keys": {
```
3mard, line 44: I am not sure about system_keys .. it exists in the api, however I never used it nor was able to see it being set while I was testing

georghildebrand:
Is there already an information when this feature will be merged?

njuCZ:
@georghildebrand I am sorry that I don't know the detail plan. If you are urgent about this PR, I could make a ping to the reviewers, or you could directly ping the reviewers in the slack channel.

georghildebrand:
@njuCZ , thank you for having a look. Its quite urgent for me. I will also ping the devs in the slack channel

georghildebrand:
@njuCZ , just realized that I have no slack account anymore as i switched jobs. Could you ping the engineers please?

3mard:
@georghildebrand you can use this link to request an invitation : https://terraform-azure.slack.com/

sgettys:
I am also waiting on this!

georghildebrand:
@sgettys , please consider upvoting in slack. actually there is currently now response from the devs. https://app.slack.com/client/TB9RVKMGW/CJD3VK9Q9/thread/CJD3VK9Q9-1596945807.467000

azurerm/internal/services/web/data_source_function_app_host_keys.go
magodo: We can simply just keep it here as long as the test passe :+1:

azurerm/internal/services/web/data_source_function_app_host_keys.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		return fmt.Errorf("Error making Read request on AzureRM Function App Hostkeys %q: %+v", name, err)
+	}
+
+	d.SetId(time.Now().UTC().String())
+	if err := d.Set("master_key", res.MasterKey); err != nil {
+		return err
+	}
+	if err = d.Set("function_keys", res.FunctionKeys); err != nil {
```
magodo, line 80: Can we use `utils.FlattenMapStringPtrString()` here?

```suggestion
	if err = d.Set("function_keys", utils.FlattenMapStringPtrString(res.FunctionKeys)); err != nil {
```

azurerm/internal/services/web/data_source_function_app_host_keys.go
```diff
@@ -0,0 +1,88 @@
...
+		return err
+	}
+	if err = d.Set("system_keys", res.SystemKeys); err != nil {
```
magodo, line 83: Same as above:

```suggestion
	if err = d.Set("system_keys", utils.FlattenMapStringPtrString(res.SystemKeys)); err != nil {
```

azurerm/internal/services/web/data_source_function_app_host_keys.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		if utils.ResponseWasNotFound(res.Response) {
+			return fmt.Errorf("Error: AzureRM Function App %q (Resource Group %q) was not found", name, resourceGroup)
+		}
+
+		return fmt.Errorf("Error making Read request on AzureRM Function App Hostkeys %q: %+v", name, err)
+	}
+
+	d.SetId(time.Now().UTC().String())
+	if err := d.Set("master_key", res.MasterKey); err != nil {
```
magodo, line 77: For primitive type, there is no need to error check for d.Set:

```suggestion
	d.Set("master_key", res.MasterKey)
```

azurerm/internal/services/web/registration.go
```diff
@@ -27,6 +27,7 @@ func (r Registration) SupportedDataSources() map[string]*schema.Resource {
 		"azurerm_app_service_certificate":       dataSourceAppServiceCertificate(),
 		"azurerm_app_service_plan":              dataSourceAppServicePlan(),
 		"azurerm_function_app":                  dataSourceArmFunctionApp(),
+		"azurerm_function_app_host_keys":        dataSourceArmFunctionAppHostKeys(),
```
magodo, line 30: Can we rename this resource as "azurerm_app_service_host_key", since:

1. Making it non-plural as it is only for a certain host
2. Name it as app service instead of function app (though it should works for both as both are using the same API at the end) to make it align with other resource naming conventions (e.g. `azurerm_app_service_slot_virtual_network_swift_connection`)

So you might also need to update the document and rename files accordingly.

magodo (changes requested):
@3mard Thanks for this PR!
I've taken a look through and left some comments inline, but this is mostly looking good to me 👍

azurerm/internal/services/web/data_source_function_app_host_keys.go
jackofallops: If its value is not returned by the API / we're not using it, then it should be removed.

azurerm/internal/services/web/registration.go
tombuildsstuff: this makes more sense as `keys`, since it's returning (presumably) the primary and secondary keys here - so this is fine as plural

azurerm/internal/services/web/data_source_function_app_host_keys.go
```diff
@@ -0,0 +1,88 @@
...
+			},
+
+			"function_keys": {
+				Type:     schema.TypeMap,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Sensitive: true,
+			},
```
tombuildsstuff, line 44: rather than exposing this as a Map - can we lookup the data in a defensive fashion and output those specific fields, so that users don't have to deal with breaking changes in the response format here e.g.

```
defaultFunctionKey := ""
if v, ok := keys["default"]; ok {
  defaultFunctionKey = v
}
d.Set("default_function_key", defaultFunctionKey)
```

azurerm/internal/services/web/data_source_function_app_host_keys.go
```diff
@@ -0,0 +1,88 @@
...
+
+	if err != nil {
+		if utils.ResponseWasNotFound(res.Response) {
+			return fmt.Errorf("Error: AzureRM Function App %q (Resource Group %q) was not found", name, resourceGroup)
+		}
+
+		return fmt.Errorf("Error making Read request on AzureRM Function App Hostkeys %q: %+v", name, err)
+	}
+
+	d.SetId(time.Now().UTC().String())
```
tombuildsstuff, line 76: can we use the ID of the Function App/Key returned here, rather than a date which'll change all the time (and causes issues in Terraform 0.13+)

azurerm/internal/services/web/data_source_function_app_host_keys.go
tombuildsstuff: actually instead can we expose the values from this field in a defensive manner (see above) - exposing a map of values isn't a great UX since these can change in the API response

website/docs/d/function_app_host_keys.html.markdown
```diff
@@ -0,0 +1,40 @@
+---
+subcategory: 'App Service (Web Apps)'
+layout: 'azurerm'
+page_title: 'Azure Resource Manager: azurerm_function_app_host_keys'
+description: |-
+  Gets HostKeys of an existing Function App.
```
tombuildsstuff, line 6:
```suggestion
  Gets the Host Keys of an existing Function App.
```

website/docs/d/function_app_host_keys.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+---
+
+# Data Source: azurerm_function_app_host_keys
+
+Use this data source to fetch the Hostkeys of an existing Function App
```
tombuildsstuff, line 11:
```suggestion
Use this data source to fetch the Host Keys of an existing Function App
```

website/docs/d/function_app_host_keys.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+
+## Example Usage
+
+```hcl
+data "azurerm_function_app_host_keys" "example" {
+  name                = "test-azure-functions"
```
tombuildsstuff, line 17:
```suggestion
  name                = "example-function"
```

website/docs/d/function_app_host_keys.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+}
+```
+
+~> **Note:** The unencrypted value of FunctionKeys, MasterKey and SystemKeys will be stored in the raw state as plain-text.
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `name` - The name of the Function App resource.
```
tombuildsstuff, line 28: can we make this:

```suggestion
- `name` - The name of the Function App.
```

website/docs/d/function_app_host_keys.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+
+- `resource_group_name` - The name of the Resource Group where the Function App exists.
+
+## Attributes Reference
+
+The following arguments are supported:
+
+- `function_keys` - A key-value pair of Host level function keys.
+
+- `system_keys` - A key-value pair of system keys
```
tombuildsstuff, line 38: (as above) can we make these specific fields rather than maps

tombuildsstuff (changes requested):
hey @3mard 

Thanks for this PR :)

I've taken a look through and left some comments inline but if we can fix those up then this otherwise LGTM 👍 

Thanks!

azurerm/internal/services/web/data_source_function_app_host_keys.go
3mard: I agree with @jackofallops, I am going to remove it

azurerm/internal/services/web/data_source_function_app_host_keys.go
3mard: I'd like to go with @tombuildsstuff approach on this if that's okay with @magodo

azurerm/internal/services/web/data_source_function_app_host_keys.go
3mard: @tombuildsstuff  I don't think I have the access to the function Id in the response AFAIK, 

should I do  a 
`	 client.Get(ctx, resourceGroup, name)
` 
just to get the function Id ?

azurerm/internal/services/web/data_source_function_app_host_keys.go
magodo: @3mard Sure, that makes more sense

azurerm/internal/services/web/data_source_function_app_host_keys.go
tombuildsstuff: makes sense for now - we'll switch this out for something else later on 👍

3mard: Oh sorry for the confusion @tombuildsstuff , I went with your suggested solution, (one more API request to retrieve the function Id ), should I keep it like this?

magodo (changes requested):
@3mard 

I've taken another look through and left some comments inline, one thing I'm uncertain is about how should we set the ID for this data source, maybe Tom shall shed some light on it.

Otherwise, this is mostly looking good to me 👍

azurerm/internal/services/web/data_source_function_app_host_keys.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	d.Set("master_key", res.MasterKey)
+	defaultFunctionKey := ""
+
+	if v, ok := res.FunctionKeys["default"]; ok {
+		defaultFunctionKey = *v
+	}
+
+	d.Set("default_function_key", defaultFunctionKey)
```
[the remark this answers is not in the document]
3mard: 🚀

website/docs/d/function_app_host_keys.html.markdown
```diff
@@ -0,0 +1,38 @@
...
+## Example Usage
+
+```hcl
+data "azurerm_function_app_host_keys" "example" {
+  name                = "example-function"
+  resource_group_name = azurerm_resource_group.example.name
+}
+```
+
+~> **Note:** The unencrypted value of FunctionKeys, MasterKey and SystemKeys will be stored in the raw state as plain-text.
```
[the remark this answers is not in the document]
3mard: great catch! I will fix this!

azurerm/internal/services/web/data_source_function_app_host_keys.go
```diff
@@ -0,0 +1,83 @@
...
+
+	functionSettings, err := client.Get(ctx, resourceGroup, name)
+	if err != nil {
+		if utils.ResponseWasNotFound(functionSettings.Response) {
+			return fmt.Errorf("Error: AzureRM Function App %q (Resource Group %q) was not found", name, resourceGroup)
+		}
+		return fmt.Errorf("Error making Read request on AzureRM Function App %q: %+v", name, err)
+	}
+
+	d.SetId(*functionSettings.ID)
```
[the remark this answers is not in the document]
3mard: I had the same concern about this, but then I thought if you are using terraform to setup your infra you are probably sending tones of request already.. and one more request won't be a deal-breaker

website/docs/d/function_app_host_keys.html.markdown
```diff
@@ -19,7 +19,8 @@ data "azurerm_function_app_host_keys" "example" {
 }
 ```
 
-~> **Note:** The unencrypted value of FunctionKeys, MasterKey and SystemKeys will be stored in the raw state as plain-text.
+~> **Note:** All arguments including the secret value will be stored in the raw state as plain-text.
+[Read more about sensitive data in state](/docs/state/sensitive-data.html).asterKey and SystemKeys will be stored in the raw state as plain-text.
```
magodo, line 23: Can we modify the wording here a bit:

```suggestion
~> **Note:** All arguments including the secret value will be stored in the raw state as plain-text, including `default_function_key` and `master_key`. [Read more about sensitive data in state](/docs/state/sensitive-data.html).
```

azurerm/internal/services/web/data_source_function_app_host_keys.go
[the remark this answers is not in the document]
tombuildsstuff: agreed, as above I think this is fine for now; we'll switch this out with the larger app service/function app refactoring come 3.0 :)

magodo (changes requested):
It almost look good to me, except a minor wording change. Otherwise, it should be good to go 🚀

website/docs/d/function_app_host_keys.html.markdown
3mard: image

but I have no problems with changing it if you still wish so

magodo:
image

PR #7918, opened 2020-07-28
New resource: `azurerm_cdn_endpoint_custom_domain`

magodo:
This PR introduces a new resource: `azurerm_cdn_endpoint_custom_domain`, which can optionally support https.

For both cases, the acctest will by default be skipped unless you specify the DNS zone resource group name and the DNS zone name (which should be the same as the DNS domain name) via environment variables: `ARM_TEST_DNS_ZONE_RESOURCE_GROUP_NAME` and `ARM_TEST_DNS_ZONE_NAME`. The reason is that currently there is no support of dns resource (though some users asks for this in terraform-providers/terraform-provider-azurerm#4683). Even that resource is ready, I still feel it is not appropriate to subscribe/unsubscribe a domain name frequently, as the subscription actually happens by delegating to GoDaddy. In which case, it might not be able to unregistered right away, so it is possible to get a conflict. On the other hand, as domain name is limited in the world, the test is likely to fail to create a domain name because the randomly picked name has already been registered. So we shall prepare a domain name and a DNS zone for the test. In each test run, we just create a randomly picked name as the subdomain and make sure to destroy it after the test.

For the non-https case, the test run relatively fast, and test has been passed.

For the https case, there are two cases - CDN managed and user managed. For both cases, the https setting doesn't allow in-place update. That is because when updateing an https setting by disabling and enabling. The enabling request after the endpoint turns into disabled state will result into following error:

> Custom HTTPS cannot be enabled at the moment because it takes 8 hours to
clean up your previous enablement request for the same custom domain.
Please try again later.

So we only allow toggling update, i.e. enable or disable.

Besides, both CDN managed and user managed test case will take a very long time. Only a disable/enable operation will take up to 8 hours to finish (hence I set the timeout accordingly).

Addtionally, for the user managed case, one more environment variable `ARM_TEST_DNS_CERTIFICATE` is needed, which contains the CA signed certificate for that domain.

Implement terraform-providers/terraform-provider-azurerm#398.


azurerm/internal/services/cdn/parse/cdn_endpoint_custom_domain.go
```diff
@@ -0,0 +1,32 @@
...
+type CdnEndpointCustomDomainId struct {
+	ResourceGroup string
+	ProfileName   string
+	EndpointName  string
+	Name          string
+}
+
+func CdnEndpointCustomDomainID(input string) (*CdnEndpointCustomDomainId, error) {
+	// Example ID: /subscriptions/<sub1>/resourcegroups/<grp1>/providers/Microsoft.Cdn/profiles/<profile1>/endpoints/<ep1>/customdomains/<domain1>
+	groups := regexp.MustCompile(`^(.+)/customdomains/([^/]+)$`).FindStringSubmatch(input)
```
tombuildsstuff, line 17: rather than using a regex, can we use `azure.ParseResourceID` here, that provides more granular errors?

magodo: Hi @tombuildsstuff 
I intend to reuse the parse function for cdn endpoint here, otherwise we will have to duplicate the cdn endpoint part parsing code here.

mbok:
Looking forward to have this feature supported in Terraform! Currently it isn't assigned to a milestone. Some ideas? Thanks a lot.

PR #7922, opened 2020-07-28
Remove default value for ip_address_type of azurerm_container_group

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/1697


tombuildsstuff:
👋 

Since this is a breaking change I'm going to temporarily close this PR for the moment until the next major version of the Azure Provider - but we'll re-open this when we're closer to v3 of the Azure Provider.

Thanks!

PR #7949, opened 2020-07-29
Workaround for issue where GET HTTP requests to Azure REST API to get…

amonnphillip:
… the resource ID sometimes does not return at all

Issue 7412

See full issue description at: https://github.com/terraform-providers/terraform-provider-azurerm/issues/7412

### DIFF
--- a/azurerm/internal/requestwrapper/requestwrapper.go
+++ b/azurerm/internal/requestwrapper/requestwrapper.go
@@ -20,10 +20,8 @@ func GetWithTimeoutsAndRetries(
 	requestTimeout int,
 	ctx context.Context,
 	getMethod func(ctx context.Context) (interface{}, error)) (read interface{}, err error) {
-
 	// Retry the request based on <numRetries> should the request return an error or be not reploy afer <requestTimeout> seconds
 	for attempt := 0; attempt < numRetries; attempt++ {
-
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
hey @amonnphillip 

Thanks for this PR :)

Taking a look into this - the Azure SDK [automatically retries failed requests](https://github.com/Azure/go-autorest/blob/master/autorest/azure/rp.go#L30-L74) as [a part of the Request Lifecycle](https://github.com/Azure/azure-sdk-for-go/blob/master/services/compute/mgmt/2020-06-01/compute/virtualmachines.go#L674), as such whilst one option to fix this issue would be to reimplement this behaviour within Terraform, in practice I think this is the wrong place to fix this. Instead if we fix this bug (where requests without responses aren't retried) within [Azure/go-autorest](https://github.com/Azure/go-autorest) which is the framework behind [the Azure SDK for Go](https://github.com/Azure/azure-sdk-for-go) then this fix should be available in all resources.

As such whilst I'd like to thank you for this contribution I'm going to close this PR for the moment - however I'd suggest [opening an issue on Azure/go-autorest](https://github.com/Azure/go-autorest) as a next step here. Once that's available we can vendor those changes in and this should be available within all resources 👍🏻

Thanks!

PR #7953, opened 2020-07-29
`azurerm_data_factory_dataset_delimited_text` - fix issue with property `azure_blob_storage_account`

markti-persistent:
fixes issue #7954

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -98,7 +98,7 @@ testacc: fmtcheck
 	TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
 
 acctests: fmtcheck
-	TF_ACC=1 go test -v ./azurerm/internal/services/$(SERVICE)/tests/ $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
+	TF_ACC=1 go test -v ./azurerm/internal/services/$(SERVICE)/ $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
 
 debugacc: fmtcheck
 	TF_ACC=1 dlv test $(TEST) --headless --listen=:2345 --api-version=2 -- -test.v $(TESTARGS)
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -98,7 +98,7 @@ testacc: fmtcheck
 	TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
 
 acctests: fmtcheck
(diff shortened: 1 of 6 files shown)


markti-persistent:
fixes issue #7954

GNUmakefile
```diff
@@ -98,7 +98,7 @@ testacc: fmtcheck
 	TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
 
 acctests: fmtcheck
-	TF_ACC=1 go test -v ./azurerm/internal/services/$(SERVICE)/tests/ $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
+	TF_ACC=1 go test -v ./azurerm/internal/services/$(SERVICE)/ $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
```
magodo, line 101: Could we keep this as it was? The *tests* subdirectory doesn't exist for datafactory subpacakge now, whilst it should have.
Maybe can we do a folder layout refactor for the datafactory to introduce the *tests* subdirectory, as other services?

azurerm/internal/services/datafactory/data_factory_dataset_json_resource.go
```diff
@@ -92,7 +92,7 @@ func resourceArmDataFactoryDatasetJSON() *schema.Resource {
 				Type:     schema.TypeList,
 				MaxItems: 1,
 				Optional: true,
-				//ConflictsWith: []string{"sftp_server_location", "file_server_location", "s3_location", "blob_storage_location"},
+				//ConflictsWith: []string{"sftp_server_location", "file_server_location", "s3_location", "azure_blob_storage_location"},
```
magodo, line 95: The document of this resource lacks of this `azure_blob_storage_location` field, would you please also complement the document so as user are aware of this feature?

GNUmakefile
markti: Hi @magodo you might want to check with @katbyte . Data factory originally had a tests folder. Someone removed it. I asked the slack channel and was told there is a slow refactor happening to remove the tests folder. I'm fine either way it just needs to be consistent as the acceptance tests are broken when I try to run them.

azurerm/internal/services/datafactory/data_factory_dataset_json_resource.go
markti: You're right. I added to Delimited text but not the json dataset.

GNUmakefile
markti-persistent: @magodo I'ved reverted to included the tests folder path, its just something I'll always have to change until somebody updates all the services to consistently use (or not) a 'tests' subfolder.

azurerm/internal/services/datafactory/data_factory_dataset_json_resource.go
markti-persistent: @magodo I've updated the docs for JSON dataset

GNUmakefile
magodo: You are right (I recall it now), the *tests* subdirectory IS to be deprecated. Whilst that is not happening for all the services, we shall just keep the Makefile as it was for now.

magodo:
image

PR #7956, opened 2020-07-29
azurerm_network_security_rule: adds a feature flag to disable locking

katbyte:


azurerm/internal/provider/features.go
```diff
@@ -135,5 +153,15 @@ func expandFeatures(input []interface{}) features.UserFeatures {
 		}
 	}
 
+	if raw, ok := val["network"]; ok {
+		items := raw.([]interface{})
+		if len(items) > 0 {
+			networkRaw := items[0].(map[string]interface{})
```
mbfrahry, line 159: Just in case, we'll want to nil check this before casting it

PR #7969, opened 2020-07-30
`azurerm_api_management_subscription` - Support `allow_tracing property`

BlakeDickie:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/7968

* Added `allow_tracing` as a property on the `azurerm_api_management_subscription` resource.
  * Defaults to `true` to match the current state of the flag when subscriptions are created before this change.


BlakeDickie:
@tombuildsstuff No worries.  I've added some acceptance tests for ensuring the new property defaults correctly and can be set to both true/false explicitly.  It's my first time in this code base, so let me know if there's any tweaks that are needed.

azurerm/internal/services/apimanagement/api_management_subscription_resource.go
```diff
@@ -184,6 +192,7 @@ func resourceArmApiManagementSubscriptionRead(d *schema.ResourceData, meta inter
 	d.Set("subscription_id", subscriptionId)
 	d.Set("resource_group_name", resourceGroup)
 	d.Set("api_management_name", serviceName)
+	d.Set("allow_tracing", resp.AllowTracing)
```
tombuildsstuff, line 195: rather than using `resp` here, we need to access this from within the `SubscriptionContractProperties` object (`properties` in the response) which could be nil - so can we move this inside the if statement below and make this:

```suggestion
		d.Set("allow_tracing", props.AllowTracing)
```

tombuildsstuff (commented):
hey @BlakeDickie 

Thanks for pushing those changes - I've taken a look through and if we can fix the one comment/issue then this otherwise LGTM 👍 

Thanks!

tombuildsstuff:
Tests pass:

Screenshot 2020-08-28 at 18 14 44

azurerm/internal/services/apimanagement/api_management_subscription_resource.go
BlakeDickie: Okay, I've updated the reading to be in the if statement and use the props object.

tombuildsstuff (approved):
LGTM - thanks for pushing those changes @BlakeDickie  👍

PR #7971, opened 2020-07-31
add check for functionApp kind and dynamic sku tier - "azurerm_app_service_plan"

### LINKED ISSUE #4396: azurerm_app_service_plan replaces Linux App-Plans on every apply
Hi there,

I tried to terraform a Linux FunctionApp using the AzureRM Provider v1.34.0 and terraform 0.12.8.

Here the configuration
```
resource "azurerm_app_service_plan" "myAppPlan" {
  name                = "myAppPlan"
  location            = azurerm_resource_group.rg.location
  resource_group_name = azurerm_resource_group.rg.name
  kind                = "Linux"
  reserved            = true
  sku {
    tier = "Dynamic"
    size = "Y1"
  }
}
```
With this I created the App-Serviceplan and following my App and even though the App is working if I reapply terraform, I got changes for the App-Serviceplan which requires a replacement.
```
 # azurerm_app_service_plan.myAppPlan must be replaced
-/+ resource "azurerm_app_service_plan" "myAppPlan" {
      + app_service_environment_id   = (known after apply)
      ~ id                           = "/subscriptions/.../myAppPlan" -> (known after apply)
      - is_xenon                     = false -> null
      ~ kind                         = "functionapp" -> "Linux" # forces replacement
        location                     = "westeurope"
      ~ maximum_elastic_worker_count = 1 -> (known after apply)
      ~ maximum_number_of_workers    = 0 -> (known after apply)
        name                         = "myAppPlan"
      ~ per_site_scaling             = false -> (known after apply)
        reserved                     = true
        resource_group_name          = "rename"
      ~ tags                         = {} -> (known after apply)

      ~ properties {
          + app_service_environment_id = (known after apply)
          ~ per_site_scaling           = false -> (known after apply)
          ~ reserved                   = true -> (known after apply)
        }

      ~ sku {
          ~ capacity = 0 -> (known after apply)
            size     = "Y1"
            tier     = "Dynamic"
        }
    }
```
And the App itself will be replaced too and here the kind is set to `functionapp,linux` 
```
# azurerm_function_app.myApp must be replaced
-/+ resource "azurerm_function_app" "myApp" {
      ~ app_service_plan_id            = "/subscriptions/.../myAppPlan" -> (known after apply) # forces replacement
      - app_settings                   = {
          - "WEBSITE_RUN_FROM_PACKAGE" = "https://mystorageaccount.blob.core.windows.net/..."
        } -> null
        client_affinity_enabled        = false
      ~ default_hostname               = "myapp.azurewebsites.net" -> (known after apply)
        enable_builtin_logging         = false
        enabled                        = true
        https_only                     = false
      ~ id                             = "/subscriptions/.../myApp" -> (known after apply)
      ~ kind                           = "functionapp,linux" -> (known after apply)
        location                       = "westeurope"
        name                           = "myApp"
...
```
This is a really annoying behavior. Do I miss something or is this a bug?

### LINKED ISSUE #5971: When re-running deployment, Terraform attempts to recreate the function app plan despite already existing
### Community Note

Terraform version: `v0.12.18`
AzureRM provider version: `1.40.0`

### Affected Resource(s)

* `azurerm_app_service_plan`

```hcl
resource "azurerm_app_service_plan" "trafficdata" {
  name                = "plan-${var.appname}-traffic-${var.environment}"
  resource_group_name =  azurerm_resource_group.trafficdata.name
  location            =  var.location
  kind                =  "linux"
  reserved            =  true

  sku {
    tier = "Basic"
    size = "B1"
  }
}
```

### Debug Output

https://gist.github.com/jjgriff93/5836c4a6ea2c4fea69e7cc223b003771

### Expected Behavior

Terraform should have detected existing App Service Plan and not recreated

### Actual Behavior

Despite the configuration not changing, Terraform detects the existing App Service Plan kind as Linux instead of Function App and thus tears down existing environment to replace with the same

### Steps to Reproduce

1. Create a Function App tf similar to the one above for the App Service Plan component and set the `kind` to `functionapp`
2. `terraform apply`
3. `terraform apply` again

* #0000

njuCZ:
fix #4396
fix #5971
1. the backend service will always change the kind to `functionapp` if sku tier is `dynamic`. will change kind to `elastic` if sku tier is `ElasticPremium`. add check to avoid such config.
2. update doc about `reserved`

### DIFF
--- a/azurerm/internal/services/web/resource_arm_app_service_plan.go
+++ b/azurerm/internal/services/web/resource_arm_app_service_plan.go
@@ -201,6 +201,10 @@ func resourceArmAppServicePlanCreateUpdate(d *schema.ResourceData, meta interfac
 		return fmt.Errorf("`kind` has to be set to `FunctionApp` when `sku.tier` is `Dynamic`")
 	}
 
+	if strings.EqualFold(*sku.Tier, "ElasticPremium") && !strings.EqualFold(kind, "elastic") {
+		return fmt.Errorf("`kind` has to be set to `elastic` when `sku.tier` is `ElasticPremium`")
+	}
+
 	if v := d.Get("maximum_elastic_worker_count").(int); v > 0 {
 		appServicePlan.AppServicePlanProperties.MaximumElasticWorkerCount = utils.Int32(int32(v))
 	}
--- a/azurerm/internal/services/web/resource_arm_app_service_plan.go
+++ b/azurerm/internal/services/web/resource_arm_app_service_plan.go
@@ -197,7 +197,7 @@ func resourceArmAppServicePlanCreateUpdate(d *schema.ResourceData, meta interfac
 		return fmt.Errorf("`reserved` has to be set to false when kind is set to `Windows`")
 	}
 
-	if *sku.Tier == "Dynamic" && !strings.EqualFold(kind, "FunctionApp") {
+	if strings.EqualFold(*sku.Tier, "Dynamic") && !strings.EqualFold(kind, "FunctionApp") {
 		return fmt.Errorf("`kind` has to be set to `FunctionApp` when `sku.tier` is `Dynamic`")
 	}
 
--- a/azurerm/internal/services/web/resource_arm_app_service_plan.go
+++ b/azurerm/internal/services/web/resource_arm_app_service_plan.go
@@ -197,6 +197,10 @@ func resourceArmAppServicePlanCreateUpdate(d *schema.ResourceData, meta interfac
 		return fmt.Errorf("`reserved` has to be set to false when kind is set to `Windows`")
 	}
 
+	if *sku.Tier == "Dynamic" && !strings.EqualFold(kind, "FunctionApp") {
+		return fmt.Errorf("`kind` has to be set to `FunctionApp` when `sku.tier` is `Dynamic`")
+	}
+
 	if v := d.Get("maximum_elastic_worker_count").(int); v > 0 {
 		appServicePlan.AppServicePlanProperties.MaximumElasticWorkerCount = utils.Int32(int32(v))
 	}
--- a/website/docs/r/app_service_plan.html.markdown
+++ b/website/docs/r/app_service_plan.html.markdown
@@ -117,7 +117,7 @@ The following arguments are supported:
 
 ~> **NOTE:** Attaching to an App Service Environment requires the App Service Plan use a `Premium` SKU (when using an ASEv1) and the `Isolated` SKU (for an ASEv2).
 
-* `reserved` - (Optional) Is this App Service Plan `Reserved`. Defaults to `false`.
+* `reserved` - (Optional) Specifies whether it's Linux Platform. Should be `true` If Linux app service plan, false otherwise. Defaults to `false`.
 
 * `per_site_scaling` - (Optional) Can Apps assigned to this App Service Plan be scaled independently? If set to `false` apps assigned to this plan will scale to all instances of the plan.  Defaults to `false`.
 


website/docs/r/app_service_plan.html.markdown
```diff
@@ -117,7 +117,7 @@ The following arguments are supported:
 
 ~> **NOTE:** Attaching to an App Service Environment requires the App Service Plan use a `Premium` SKU (when using an ASEv1) and the `Isolated` SKU (for an ASEv2).
 
-* `reserved` - (Optional) Is this App Service Plan `Reserved`. Defaults to `false`.
+* `reserved` - (Optional) Specifies whether it's Linux Platform. Should be `true` If Linux app service plan, false otherwise. Defaults to `false`.
```
tombuildsstuff, line 120: this isn't just for Linux - this specifies if the App Service Plan should be always running or start/stopped on demand

azurerm/internal/services/web/resource_arm_app_service_plan.go
```diff
@@ -197,6 +197,14 @@ func resourceArmAppServicePlanCreateUpdate(d *schema.ResourceData, meta interfac
 		return fmt.Errorf("`reserved` has to be set to false when kind is set to `Windows`")
 	}
 
+	if strings.EqualFold(*sku.Tier, "Dynamic") && !strings.EqualFold(kind, "FunctionApp") {
+		return fmt.Errorf("`kind` has to be set to `FunctionApp` when `sku.tier` is `Dynamic`")
+	}
```
tombuildsstuff, line 202: isn't Dynamic available for Consumption Plan Web Apps too?

azurerm/internal/services/web/resource_arm_app_service_plan.go
```diff
@@ -197,6 +197,14 @@ func resourceArmAppServicePlanCreateUpdate(d *schema.ResourceData, meta interfac
 		return fmt.Errorf("`reserved` has to be set to false when kind is set to `Windows`")
 	}
 
+	if strings.EqualFold(*sku.Tier, "Dynamic") && !strings.EqualFold(kind, "FunctionApp") {
```
tombuildsstuff, line 200: there's a crash here if `sku.Tier` is nil

azurerm/internal/services/web/resource_arm_app_service_plan.go
```diff
@@ -197,6 +197,14 @@ func resourceArmAppServicePlanCreateUpdate(d *schema.ResourceData, meta interfac
...
+		return fmt.Errorf("`kind` has to be set to `FunctionApp` when `sku.tier` is `Dynamic`")
+	}
+
+	if strings.EqualFold(*sku.Tier, "ElasticPremium") && !strings.EqualFold(kind, "elastic") {
```
tombuildsstuff, line 204: there's a crash here if `sku.Tier` is nil

azurerm/internal/services/web/resource_arm_app_service_plan.go
```diff
@@ -197,6 +197,14 @@ func resourceArmAppServicePlanCreateUpdate(d *schema.ResourceData, meta interfac
...
+		return fmt.Errorf("`kind` has to be set to `elastic` when `sku.tier` is `ElasticPremium`")
```
tombuildsstuff, line 205: can we add a test to check this?

azurerm/internal/services/web/resource_arm_app_service_plan.go
```diff
@@ -197,6 +197,14 @@ func resourceArmAppServicePlanCreateUpdate(d *schema.ResourceData, meta interfac
 		return fmt.Errorf("`reserved` has to be set to false when kind is set to `Windows`")
 	}
 
+	if strings.EqualFold(*sku.Tier, "Dynamic") && !strings.EqualFold(kind, "FunctionApp") {
+		return fmt.Errorf("`kind` has to be set to `FunctionApp` when `sku.tier` is `Dynamic`")
```
tombuildsstuff, line 201: can we add a test to check this?

tombuildsstuff (changes requested):
hi @njuCZ 

I've taken a look through and left some comments inline - if we can fix those up then we should be able to take another look

Thanks!

website/docs/r/app_service_plan.html.markdown
njuCZ: This field name is very misleading. According to https://docs.microsoft.com/en-us/rest/api/appservice/appserviceplans/createorupdate#request-body, it says `If Linux app service plan true, false otherwise.`. 

According to the implementation of azure.cli: https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/appservice/custom.py#L1648
users could specify the flag "--is-linux", and azure.cli will convert it to `reserved` field.

I have also emailed the service team, they says "reserved" means the OS type, "kind" is just a ARM level field, it's only used to drive the UX. It's not needed when creating. Azure.cli doesn't have the argument "kind".

Currently I think it's a little messy for resource "azurerm_app_service_plan" to have both field "reserved" and "kind", though it's the fault of service api definition. There seems a issue fired by a member of the service team: https://github.com/terraform-providers/terraform-provider-azurerm/issues/5584

what's more, the OS type of app_service and function_app is defined by the app service plan, but for now both resource have "os_type" field, it should be "computed: true".

I wonder could we fix these in version 3 ?

azurerm/internal/services/web/resource_arm_app_service_plan.go
njuCZ: this PR is just a workaround. `kind` should not be input by user, the backend service will calculate the value for `kind`. But to avoid breaking change, and to make user use it correctly, I added this constraints.

the backend service will always change the kind to `functionapp` if sku tier is `dynamic`. will change kind to `elastic` if sku tier is `ElasticPremium`.

njuCZ:
@tombuildsstuff I have replied some of your questions. After we have draw the conclusion, I would fix the rest suggestions.

Issue #3967, opened 2019-07-30
Can't refresh api management admin user data using azurerm_api_management_user


Sasa33:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
Terraform v0.11.14
provider.azurerm v1.32.0

### Affected Resource(s)

* `azurerm_api_management_user`

### Terraform Configuration Files

```hcl
data "azurerm_api_management_user" "admin" {
  user_id             = "1"
  api_management_name = "${azurerm_api_management.api_management.name}"
  resource_group_name = "${azurerm_resource_group.resource_group.name}"
}
```
I'd like to get the default admin user account  (/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ApiManagement/service/{serviceName}/users/1) of api management
### Expected Behavior
I get the admin user's data

### Actual Behavior
I got the following error:
Error: data.azurerm_api_management_user.admin: "user_id" may only contain alphanumeric characters and dashes up to 80 characters in length (this error comes from the ApiManagementUserName function in api_management validator)

I noticed that other users's id is 24 length digits, if I tried another user, it works as expected.
But I can't get the admin user's data by using "1" as user_id

jhattarki:
I am having the same issue. Is there a solution/workaround.
Thanks.

identifysun:
@jhattarki 

You can try to create a new user(no admin).

EtienneDeneuve:
@Sasa33 do you still have the issue after the @rikribbers modification ?

favoretti:
Since this issue has been reported a long time ago and relates to the version of provider we no longer support - I'm going to close it. Please open a new updated bug report on current versions of terraform and provider if this is still relevant. Thank you.

PR #7975, opened 2020-07-31
d/api_management_user - updating the username validation

rikribbers:
I fixed the retrieval of admin user. 

Rationale for the fix:

I interpreted that the user_id field is the user identifier and not the full Azure resource id. The MS docs state that is the "User identifier" and it is unique within the current APIM instance. The type is string. 

https://docs.microsoft.com/en-us/rest/api/apimanagement/2019-12-01/user/get

Fixes #3967


azurerm/internal/services/apimanagement/api_management_user_data_source.go
```diff
@@ -20,7 +20,10 @@ func dataSourceArmApiManagementUser() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"user_id": azure.SchemaApiManagementUserDataSourceName(),
+			"user_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
```
tombuildsstuff, line 26: the validation function above uses a regex rather than a Resource ID here:

```
// from the portal: `The field may contain only numbers, letters, and dash (-) sign when preceded and followed by number or a letter.`
	if matched := regexp.MustCompile(`(^[a-zA-Z0-9])([a-zA-Z0-9-]{1,78})([a-zA-Z0-9]$)`).Match([]byte(value)); !matched {
		errors = append(errors, fmt.Errorf("%q may only contain alphanumeric characters and dashes up to 80 characters in length", k))
	}
```

would it be possible to give an example of the value that's disallowed? rather than removing this validation it'd be worth updating it to ensure that it's correct

rikribbers: The example is the admin user that is created with user_id  "1" by default. Looking at the API documentation I could not find the restriction on 3-80 characters. Fiddling with the settings I found that the length could be between 1 and 80. 

Will restore the orignal validation and change the regex.

rikribbers:
ok, I understand the tradeoff that was made here.. was thinking about using a regex like this:

^(?!.*-$)^[a-z][a-z-]{0,79}$

However this uses a negative lookup that is not supported by go. So second best is to go for a length of 2:

^[a-zA-Z0-9])([a-zA-Z0-9-]{0,78})([a-zA-Z0-9]$

Or split the check up into 2 steps, one for starting and ending with a '-' character and a allowed character regex : ^[a-zA-Z0-9-]{1,80}

I could use some advice on what to implement. I prefer the last one. as this resembles the actual behaviour from the portal.

katbyte:
Hey @rikribbers - the 2nd option would be fine :)

PR #7978, opened 2020-08-01
r/linux|windows_virtual_machine - support for trying multiple sku's

### LINKED ISSUE #7952: Support for [thing]
### Community Note

We often see issues where deployment of VMs fail due to the selected VM size not currently being available. This is an especially  acute problem with the high performance graphics/GPU type VMs. High performance VMs are expensive so there are many cases where you will want to deploy, configure, perform your task, then destroy the high cost resources as time efficiently as possible. A deployment that breaks half way through because some VMs failed to deploy adds extra cost.

One feature that will help is to add the ability for the user to specify a range of VMs sizes that are acceptable for the deployment. For example:

resource "azurerm_linux_virtual_machine" "example" {
  name                = "example-machine"
// This is the proposed change
  sizes                 = ["Standard_NC6", "Standard_NC12", "Standard_NC24" ]
//
  admin_username      = "adminuser"
  network_interface_ids = [
    azurerm_network_interface.example.id,
  ]

  admin_ssh_key {
    username   = "adminuser"
    public_key = file("~/.ssh/id_rsa.pub")
  }

  os_disk {
  }

  source_image_reference {
    publisher = "Canonical"
    offer     = "UbuntuServer"
    sku       = "16.04-LTS"
    version   = "latest"
  }
}

This change will affect the following resources:

azurerm_linux_virtual_machine
azurerm_windows_virtual_machine

The size property is optional, and can be replaced with a sizes property which takes a list of VM sizes to try. The code will start at the first item, which is assumed to be the first choice, and iterate through the list. The deployment will be considered to be in an erroneous state if no VM size could be found.

### References

* #0000

amonnphillip:
fixes #7952

### DIFF
--- a/azurerm/internal/services/compute/linux_virtual_machine_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_resource.go
@@ -79,9 +79,19 @@ func resourceLinuxVirtualMachine() *schema.Resource {
 			"os_disk": virtualMachineOSDiskSchema(),
 
 			"size": {
-				Type:         schema.TypeString,
-				Required:     true,
-				ValidateFunc: validation.StringIsNotEmpty,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ValidateFunc:  validation.StringIsNotEmpty,
+				ConflictsWith: []string{"sizes"},
+			},
+
+			"sizes": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				ConflictsWith: []string{"size"},
 			},
 
 			// Optional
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -79,9 +79,19 @@ func resourceLinuxVirtualMachine() *schema.Resource {
 			"os_disk": virtualMachineOSDiskSchema(),
 
 			"size": {
-				Type:         schema.TypeString,
-				Required:     true,
-				ValidateFunc: validation.StringIsNotEmpty,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ValidateFunc:  validation.StringIsNotEmpty,
+				ConflictsWith: []string{"sizes"},
```
ArcturusZhang, line 85: `ConflictsWith` means these two attributes cannot co-exist, but it does not prevent these two could be both absent. Therefore could we change this by using `ExactlyOneOf`?
```suggestion
				ExactlyOneOf: []string{"size", "sizes"},
```
And the same for `sizes`

tombuildsstuff:
hey @amonnphillip

Thanks for this contribution - apologies for the delayed response here!

From our side we've been trying to work out how's best to proceed here - whilst this PR does bring some benefits (insofar as the chance of this provisioning being successful _should_ increase) - in a way it goes against Terraform's design, since we can't guarantee in the plan which instance size will be deployed.

One of Terraform's core features is that what you see at Plan time is what you see happen at Apply time, whilst this approach _technically_ doesn't violate that, it instead changes the semantics somewhat to say "one of these values". Whilst Terraform does have that behaviour in the form of Data Sources which can return multiple items, we limit this behaviour to Data Sources to keep the behaviour consistent.

As such whilst I'd like to thank you for this contribution (and apologies for the delayed review whilst we worked out how to proceed here) -  I'm going to close this PR in favour of #7952 which looks to solve this by adding a Data Source for this information.

Thanks!

Issue #7993, opened 2020-08-03
Unable to reference data block for azurerm_private_dns_zone: "Parent resource '...' not found"


fcatacut:
I'm trying to use a data-block to retrieve an existing Private DNS zone so that it can be referenced when creating additional resources, such as `azurerm_private_dns_zone_virtual_network_link` and `azurerm_private_dns_a_record`.

### Community Note

* Please vote on this issue by adding a 👍  [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.12.29
+ provider.azurerm v2.21.0

### Affected Resource(s)

* `azurerm_private_dns_zone`

### Terraform Configuration Files

```hcl
provider "azurerm" {
  features {}
}

data "azurerm_private_dns_zone" "example" {
  name                = "privatelink.documents.azure.com"
  resource_group_name = "<change me>"
}

resource "azurerm_resource_group" "example" {
  name     = "myrg"
  location = "eastus"
}

resource "azurerm_virtual_network" "example" {
  name                = "myvnet"
  resource_group_name = azurerm_resource_group.example.name
  location            = "eastus"
  address_space       = ["10.0.0.0/20"]
}

resource "azurerm_private_dns_zone_virtual_network_link" "example" {
  name                  = "myvnetlink"
  resource_group_name   = azurerm_resource_group.example.name
  private_dns_zone_name = data.azurerm_private_dns_zone.example.name
  virtual_network_id    = azurerm_virtual_network.example.id
}

output "data_private_dns_zone" {
  value = data.azurerm_private_dns_zone.example
}
```

### Debug Output

<https://gist.github.com/fcatacut/59b3005ef022f629502d8f6d6cc4ce41>

### Expected Behavior

`azurerm_private_dns_zone_virtual_network_link` should be successfully created.

### Actual Behavior

```bash
[...]

PR #7998, opened 2020-08-04
Update doc for resource_group_name property of azurerm_private_dns_zone_virtual_network_link

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7993


fcatacut:
> fixes #7993

This does _not_ fix #7993.

PR #8001, opened 2020-08-04
Make update of allow_extensions_operations independent from provision_vm_agent.

### LINKED ISSUE #7985: Fixup pull request #7749 (`azurerm_*_virtual_machine` - `allow_extensions_operations` can now be updated)
### Community Note

### Affected Resource(s)

* `azurerm_linux_virtual_machine`

The code change for https://github.com/terraform-providers/terraform-provider-azurerm/issues/7070 via pull request https://github.com/terraform-providers/terraform-provider-azurerm/pull/7749 introduced a regression. 
I already tried to explain it here:
https://github.com/terraform-providers/terraform-provider-azurerm/pull/7749#discussion_r454835017

This code in azurerm/internal/services/compute/{linux|windows}_virtual_machine_resource.go is wrong:

linux: https://github.com/terraform-providers/terraform-provider-azurerm/blob/55e02f697557906dd4dae43657d5e6381138b491/azurerm/internal/services/compute/linux_virtual_machine_resource.go#L390   
windows: https://github.com/terraform-providers/terraform-provider-azurerm/blob/55e02f697557906dd4dae43657d5e6381138b491/azurerm/internal/services/compute/windows_virtual_machine_resource.go#L414

```
if !provisionVMAgent && allowExtensionOperations {
			return fmt.Errorf("`allow_extension_operations` cannot be set to `true` when `provision_vm_agent` is set to `false`")
}
```
You can change the allow_extension_operations value, even if provision_vm_agent is false. If you did not deploy with provision_vm_agent you can still install it manually. This is explained here: https://docs.microsoft.com/en-us/azure/virtual-machines/troubleshooting/install-vm-agent-offline
After that, you can enable extensions, even with provision_vm_agent set to false. 

### Expected Behavior

Allow to set allow_extension_operations independent of provision_vm_agent. 
The AZURE API should fail if conditions are not fullfilled.

### Actual Behavior

If provision_vm_agent is false, you cannot set allow_extension_operations to true. The provider does not allow it.

### Steps to Reproduce

1. Create an instance with provision_vm_agent = false and allow_extension_operations = false
2. Install vm agent manually as documented by AZURE (see link above)
3. Set allow_extension_operations = true

### References

* #7070 
* #7749

NilsBusche:
This fixes the bug reported in https://github.com/terraform-providers/terraform-provider-azurerm/issues/7985.

There is no API restriction which does not allow enabling
allow_extensions_operations if provision_vm_agent is false.
For example you can install the agent manually and after that enable
allow_extensions_operations.

Fixes #7985

### DIFF
--- a/azurerm/internal/services/compute/linux_virtual_machine_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_resource.go
@@ -816,12 +816,7 @@ func resourceLinuxVirtualMachineUpdate(d *schema.ResourceData, meta interface{})
 	}
 
 	if d.HasChange("allow_extension_operations") {
-		provisionVMAgent := d.Get("provision_vm_agent").(bool)
 		allowExtensionOperations := d.Get("allow_extension_operations").(bool)
-
-		if !provisionVMAgent && allowExtensionOperations {
-			return fmt.Errorf("`allow_extension_operations` cannot be set to `true` when `provision_vm_agent` is set to `false`")
-		}
 
 		shouldUpdate = true
 
--- a/azurerm/internal/services/compute/tests/linux_virtual_machine_resource_other_test.go
+++ b/azurerm/internal/services/compute/tests/linux_virtual_machine_resource_other_test.go
@@ -77,6 +77,34 @@ func TestAccLinuxVirtualMachine_otherAllowExtensionOperationsUpdated(t *testing.
 	})
 }
 
+func TestAccLinuxVirtualMachine_otherAllowExtensionOperationsUpdatedWithoutVmAgent(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_linux_virtual_machine", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: checkLinuxVirtualMachineIsDestroyed,
+		Steps: []resource.TestStep{
+			{
+				Config: testLinuxVirtualMachine_otherAllowExtensionOperationsDisabledWithoutVmAgent(data),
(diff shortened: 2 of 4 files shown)


NilsBusche:
> Hey @NilsBusche, this is just about there but we should keep the check in the create method since we can't create a virtual machine when `allow_extensions_operations` is true but `provision_vm_agent` is false.
> 
> ```
> Original Error: Code="InvalidParameter" Message="The value of parameter allowExtensionOperations is invalid."
> ```
> 
> Removing it from update makes sense to me!

Hi @mbfrahry, you are correct. I changed this.

tombuildsstuff (commented):
hey @NilsBusche 

Thanks for pushing those changes - taking a look through this mostly LGTM - however can we add a test case to confirm that updating this field works (for both the Linux & Windows VM resources)?

Thanks!

NilsBusche:
> hey @NilsBusche
> 
> Thanks for pushing those changes - taking a look through this mostly LGTM - however can we add a test case to confirm that updating this field works (for both the Linux & Windows VM resources)?
> 
> Thanks!

Hi @tombuildsstuff, I think there is already a test for checking if update works:
https://github.com/terraform-providers/terraform-provider-azurerm/blob/8442c30c2d443dcf90cfadbd8b52c4fba16d3189/azurerm/internal/services/compute/tests/linux_virtual_machine_resource_other_test.go#L52

Should we add/change a test to make sure that updating `allow_extension_operations` also works if `provision_vm_agent` was set to false at creation of the virtual machine? Or do you think the existing tests are sufficient?

NilsBusche:
> Let's add those tests for sure just to make sure that it's doing what we expect it to do

Hi @mbfrahry, added tests for updating `allow_extension_operations` if `provision_vm_agent` is false.

tombuildsstuff (approved):
LGTM - thanks for pushing those changes @NilsBusche

Issue #7984, opened 2020-08-03
`azurerm_kubernetes_cluster` addon Profile OMSConfig refresh plan is not empty


yupwei68:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
0.12.26
 provider.azurerm v2.21.0

### Affected Resource(s)
azurerm_kubernetes_cluster

* `azurerm_XXXXX`

### Terraform Configuration Files

https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/containers/tests/kubernetes_cluster_addons_resource_test.go#L156

### Expected Behavior

refresh plan is not empty

### Actual Behavior
~ addon_profile {

          - kube_dashboard {
              - enabled = true -> null
            }

            oms_agent {
                enabled                    = true
                log_analytics_workspace_id = "/subscriptions/.../resourcegroups/ask-resource-group-yup/providers/microsoft.operationalinsights/workspaces/preyup-workspace"
                oms_agent_identity         = []
            }
        }

### Steps to Reproduce

1. Terraform apply
2. Terraform plan

### References

* #0000

favoretti:
Thanks for opening this issue! Since this issue has been reported a long time ago and relates to an older version of provider - I'm going to close it. If this is still relevant and occurring on the latest version of terraform and provider please do open a new issue!

PR #8012, opened 2020-08-05
`azurerm_kubernetes_cluster`  fix acctest `TestAccAzureRMKubernetesCluster_addonProfileOMS` and issue

yupwei68:
Fix https://github.com/terraform-providers/terraform-provider-azurerm/issues/7984

=== RUN   TestAccAzureRMKubernetesCluster_addonProfileOMS
=== PAUSE TestAccAzureRMKubernetesCluster_addonProfileOMS
=== CONT  TestAccAzureRMKubernetesCluster_addonProfileOMS
--- PASS: TestAccAzureRMKubernetesCluster_addonProfileOMS (761.53s)

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_addons.go
+++ b/azurerm/internal/services/containers/kubernetes_addons.go
@@ -86,6 +86,7 @@ func schemaKubernetesAddOnProfiles() *schema.Schema {
 					Type:     schema.TypeList,
 					MaxItems: 1,
 					Optional: true,
+					Computed: true,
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
👋 

The `kube_dashboard` block itself is intentionally not Computed since this isn't available in some regions - and trying to set this block (even to disabled) causes an API error - as such unfortunately this cannot become Computed and so I'm going to close this PR for the moment - however it should be possible to solve #7984 by specifying this block within the config.

Thanks!

PR #8063, opened 2020-08-10
Update api-version of ACR

ArcturusZhang:
This is one of the prerequisites of CMK for ACR.

Also breaking changes have been introduced in this api-version - the `storage_account_id` is no longer updatable, and we have to force replacement when it is changed. I added a note in the documentation of `azurerm_container_registry`.


azurerm/internal/services/containers/client/client.go
```diff
@@ -2,8 +2,8 @@ package client
 
 import (
 	"github.com/Azure/azure-sdk-for-go/services/containerinstance/mgmt/2018-10-01/containerinstance"
-	"github.com/Azure/azure-sdk-for-go/services/containerregistry/mgmt/2018-09-01/containerregistry"
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2020-03-01/containerservice"
+	"github.com/Azure/azure-sdk-for-go/services/preview/containerregistry/mgmt/2019-12-01-preview/containerregistry"
```
tombuildsstuff, line 6: [2019-05-01](https://github.com/Azure/azure-sdk-for-go/tree/master/services/containerregistry/mgmt/2019-05-01/containerregistry) appears to be the latest stable API version - can we use that version (or a newer GA version) instead?

website/docs/r/container_registry.html.markdown
```diff
@@ -14,6 +14,8 @@ Manages an Azure Container Registry.
 ~> **Note:** All arguments including the access key will be stored in the raw state as plain-text.
 [Read more about sensitive data in state](/docs/state/sensitive-data.html).
 
+-> **Note:** The `storage_account_id` is no longer updatable since API version `2019-12-01-preview`. If you are using a container registry with `Classical` sku, changing the `storage_account_id` would result in a replacement of the container registry.
```
tombuildsstuff, line 17: API versions are an implementation detail - so we can remove this:

```suggestion
```

azurerm/internal/services/containers/client/client.go
ArcturusZhang: Since this PR is making preparations of the CMK in ACR, 2019-05-01 does not have CMK related feature. The only api-version that contains CMK is 2019-12-01-preview.

website/docs/r/container_registry.html.markdown
ArcturusZhang: Sure!

ArcturusZhang:
Hi @tombuildsstuff I have remove the notes. As for the preview api-version thing, I will go to the service team to see if they have a plan to release a stable version based on 2019-12-01-preview. 

And after some further investigation, the CMK feature for ACR can only be enabled on newly created container registry and cannot be disabled once enabled. (Refer: https://docs.microsoft.com/en-us/azure/container-registry/container-registry-customer-managed-keys). Therefore I assume the CMK for ACR is not ready to go into terraform yet.

azurerm/internal/services/containers/client/client.go
tombuildsstuff: so there's been no stable release of this API (e.g. guarantees of no breaking changes) for 16 months (tomorrow)? shouldn't ARM Review be catching that? cc @JeffreyRichter

JeffreyRichter: @tombuildsstuff I've raised this internally. The team has promised to reply in a few days.

ArcturusZhang:
Hi @tombuildsstuff since we should get a stable api-version of ACR in the future, and currently I think CMK for ACR is not ready for terraform - it cannot be disabled after enabled, I will close this PR for now. 
When the feature is totally ready, I will open another PR to update the api-version and implement the CMK.

Thanks!

PR #8066, opened 2020-08-10
features/requires-import: gradually removing usage of the old feature-toggle

tombuildsstuff:
The feature toggle `ShouldRequireResourcesToBeImported` has been force-enabled since 2.0 - since we're no longer using it in new PR's, this first PR starts the process of removing it from the existing codebase so that we can gradually deprecate the method and then remove it

Also adds a couple of TODO's for later thought spotted whilst passing through


azurerm/internal/services/batch/batch_application_resource.go
```diff
@@ -84,17 +83,15 @@ func resourceArmBatchApplicationCreate(d *schema.ResourceData, meta interface{})
...
-		if err != nil {
-			if !utils.ResponseWasNotFound(resp.Response) {
-				return fmt.Errorf("Error checking for present of existing Batch Application %q (Account Name %q / Resource Group %q): %+v", name, accountName, resourceGroup, err)
-			}
-		}
+	resp, err := client.Get(ctx, resourceGroup, accountName, name)
+	if err != nil {
 		if !utils.ResponseWasNotFound(resp.Response) {
-			return tf.ImportAsExistsError("azurerm_batch_application", *resp.ID)
+			return fmt.Errorf("Error checking for present of existing Batch Application %q (Account Name %q / Resource Group %q): %+v", name, accountName, resourceGroup, err)
```
jackofallops, line 89: typo
```suggestion
			return fmt.Errorf("checking for presence of existing Batch Application %q (Account Name %q / Resource Group %q): %+v", name, accountName, resourceGroup, err)
```

PR #8067, opened 2020-08-10
new resource `azurerm_spring_cloud_certificate`

njuCZ:
image

### DIFF
--- a/azurerm/internal/services/appplatform/spring_cloud_certificate_resource.go
+++ b/azurerm/internal/services/appplatform/spring_cloud_certificate_resource.go
@@ -1,0 +1,149 @@
+package appplatform
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/preview/appplatform/mgmt/2019-05-01-preview/appplatform"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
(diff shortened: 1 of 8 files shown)


azurerm/internal/services/appplatform/spring_cloud_certificate_resource.go
```diff
@@ -0,0 +1,149 @@
...
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
```
jackofallops, line 43: No length or character constraints on the service here?

jackofallops (changes requested):
Hi @njuCZ 
Thanks for this new resource, just one really minor grammar change for the docs and a query below, other than that I think this is good to merge. 

Thanks!

jackofallops:
Tests pass:
image

azurerm/internal/services/appplatform/spring_cloud_certificate_resource.go
njuCZ: In the portal, the `name` is auto generated, users does not need to input it. I could not get any constraints for this field. So for now, I just set it `NotEmpty`.

PR #8092, opened 2020-08-12
azurerm_storage_account - filter allow_blob_public_access and min_tls_version from Azure US Government

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7812
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/8083

### DIFF
--- a/azurerm/internal/services/storage/client/client.go
+++ b/azurerm/internal/services/storage/client/client.go
@@ -29,7 +29,7 @@ type Client struct {
 	StorageTargetsClient     *storagecache.StorageTargetsClient
 	SubscriptionId           string
 
-	environment   az.Environment
+	Environment   az.Environment
 	storageAdAuth *autorest.Authorizer
 }
 
@@ -62,7 +62,7 @@ func NewClient(options *common.ClientOptions) *Client {
 		CachesClient:             &cachesClient,
 		SubscriptionId:           options.SubscriptionId,
 		StorageTargetsClient:     &storageTargetsClient,
-		environment:              options.Environment,
+		Environment:              options.Environment,
 	}
 
 	if options.StorageUseAzureAD {
@@ -74,7 +74,7 @@ func NewClient(options *common.ClientOptions) *Client {
 
 func (client Client) AccountsDataPlaneClient(ctx context.Context, account accountDetails) (*accounts.Client, error) {
 	if client.storageAdAuth != nil {
-		accountsClient := accounts.NewWithEnvironment(client.environment)
+		accountsClient := accounts.NewWithEnvironment(client.Environment)
 		accountsClient.Client.Authorizer = *client.storageAdAuth
 		return &accountsClient, nil
 	}
@@ -89,14 +89,14 @@ func (client Client) AccountsDataPlaneClient(ctx context.Context, account accoun
 		return nil, fmt.Errorf("Error building Authorizer: %+v", err)
 	}
 
-	accountsClient := accounts.NewWithEnvironment(client.environment)
+	accountsClient := accounts.NewWithEnvironment(client.Environment)
 	accountsClient.Client.Authorizer = storageAuth
 	return &accountsClient, nil
 }
 
 func (client Client) BlobsClient(ctx context.Context, account accountDetails) (*blobs.Client, error) {
 	if client.storageAdAuth != nil {
-		blobsClient := blobs.NewWithEnvironment(client.environment)
+		blobsClient := blobs.NewWithEnvironment(client.Environment)
 		blobsClient.Client.Authorizer = *client.storageAdAuth
 		return &blobsClient, nil
 	}
@@ -111,14 +111,14 @@ func (client Client) BlobsClient(ctx context.Context, account accountDetails) (*
 		return nil, fmt.Errorf("Error building Authorizer: %+v", err)
 	}
 
-	blobsClient := blobs.NewWithEnvironment(client.environment)
+	blobsClient := blobs.NewWithEnvironment(client.Environment)
 	blobsClient.Client.Authorizer = storageAuth
 	return &blobsClient, nil
 }
 
 func (client Client) ContainersClient(ctx context.Context, account accountDetails) (*containers.Client, error) {
 	if client.storageAdAuth != nil {
-		containersClient := containers.NewWithEnvironment(client.environment)
+		containersClient := containers.NewWithEnvironment(client.Environment)
 		containersClient.Client.Authorizer = *client.storageAdAuth
 		return &containersClient, nil
 	}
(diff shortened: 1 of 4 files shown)


azurerm/internal/services/storage/resource_arm_storage_account.go
```diff
@@ -648,13 +650,23 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
...
 	}
 
+	// For US Government Cloud, don't specify "allow_blob_public_access" and "min_tls_version" in request body.
+	// https://github.com/terraform-providers/terraform-provider-azurerm/issues/7812
+	// https://github.com/terraform-providers/terraform-provider-azurerm/issues/8083
+	if env.Name == autorestAzure.USGovernmentCloud.Name {
+		if allowBlobPublicAccess || minimumTLSVersion != string(storage.TLS10) {
+			return fmt.Errorf(`"allow_blob_public_access" and "min_tls_version" are not supported for a Storage Account located in %q`, env.Name)
+		}
+	} else {
```
WodansSon, line 665: I don't think the logic here is quite right:

1. We don't want to return an error when the fields are passed in the USGovernmentCloud or in the ChinaCloud. If we do the customers will not be able to use the resource which defeats the purpose of these fixes.

2. The if statement is only checking name for autorestAzure.USGovernmentCloud.Name, isn't the TLS issue only in the ChinaCloud? Either way we should be checking for both clouds shouldn't we?

azurerm/internal/services/storage/resource_arm_storage_account.go
```diff
@@ -888,28 +901,44 @@ func resourceArmStorageAccountUpdate(d *schema.ResourceData, meta interface{}) e
...
+		if env.Name == autorestAzure.USGovernmentCloud.Name {
+			if minimumTLSVersion != string(storage.TLS10) {
+				return fmt.Errorf(`"min_tls_version" is not supported for a Storage Account located in %q`, env.Name)
+			}
+		} else {
+			opts := storage.AccountUpdateParameters{
+				AccountPropertiesUpdateParameters: &storage.AccountPropertiesUpdateParameters{
+					MinimumTLSVersion: storage.MinimumTLSVersion(minimumTLSVersion),
+				},
+			}
```
WodansSon, line 915: Same here as above.

azurerm/internal/services/storage/resource_arm_storage_account.go
```diff
@@ -888,28 +901,44 @@ func resourceArmStorageAccountUpdate(d *schema.ResourceData, meta interface{}) e
...
+		if env.Name == autorestAzure.USGovernmentCloud.Name {
+			if allowBlobPublicAccess {
+				return fmt.Errorf(`"allow_blob_public_access" is not supported for a Storage Account located in %q`, env.Name)
+			}
+		} else {
+			opts := storage.AccountUpdateParameters{
+				AccountPropertiesUpdateParameters: &storage.AccountPropertiesUpdateParameters{
+					AllowBlobPublicAccess: &allowBlobPublicAccess,
+				},
+			}
```
WodansSon, line 937: Same here as above.

azurerm/internal/services/storage/resource_arm_storage_account.go
```diff
@@ -1066,9 +1096,14 @@ func resourceArmStorageAccountRead(d *schema.ResourceData, meta interface{}) err
...
-		d.Set("min_tls_version", string(props.MinimumTLSVersion))
 		d.Set("is_hns_enabled", props.IsHnsEnabled)
-		d.Set("allow_blob_public_access", props.AllowBlobPublicAccess)
+		// For US Government Cloud, don't specify "allow_blob_public_access" and "min_tls_version" in request body.
+		// https://github.com/terraform-providers/terraform-provider-azurerm/issues/7812
+		// https://github.com/terraform-providers/terraform-provider-azurerm/issues/8083
+		if env.Name != autorestAzure.USGovernmentCloud.Name {
+			d.Set("allow_blob_public_access", props.AllowBlobPublicAccess)
+			d.Set("min_tls_version", string(props.MinimumTLSVersion))
+		}
```
WodansSon, line 1106: Same here, the checks should be split for USGovernmentCloud and ChinaCloud

website/docs/r/storage_account.html.markdown
```diff
@@ -99,8 +99,12 @@ The following arguments are supported:
 
 * `min_tls_version` - (Optional) The minimum supported TLS version for the storage account. Possible values are `TLS1_0`, `TLS1_1`, and `TLS1_2`. Defaults to `TLS1_0` for new storage accounts.
 
+-> **NOTE:** At this time `min_tls_version` is not supported in US Government.
```
WodansSon, line 102:
```suggestion
-> **NOTE:** At this time `min_tls_version` is not supported in China Cloud.
```

azurerm/internal/services/storage/resource_arm_storage_account.go
mybayern1974: @WodansSon , thank you for the comments:
Re # 1: I may be with this solution in this PR. If we do not explicitly throw the error to users, then it means in the Azure.Gov, TF swallows and does not respect/send-to-api whatever users' setting on these fields (take the minimumTLSVer for example), then the next time when users do tf-apply w/o any changes in .tf, he/she will see diff because the .tf contains the user's customized setting on this field but the .tfstate file contains the returned value of this field which is an empty.
And the purpose of this fix is to forbid users from setting these two fields that is by design by the service on Azure.Gov
Re # 2, you are correct but we are lack of testing env of Azure.China. To ensure things are absolutely safe, we may wanna only cover things that are tested

azurerm/internal/services/storage/client/client.go
```diff
@@ -29,7 +29,7 @@ type Client struct {
 	StorageTargetsClient     *storagecache.StorageTargetsClient
 	SubscriptionId           string
 
-	environment   az.Environment
+	Environment   az.Environment
```
katbyte, line 32: Instead of modifying the storage client can we use:
```
 meta.(*clients.Client).Account.Environment.Name
```

azurerm/internal/services/storage/resource_arm_storage_account.go
```diff
@@ -648,13 +650,23 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
...
 			NetworkRuleSet:         expandStorageAccountNetworkRules(d),
 			IsHnsEnabled:           &isHnsEnabled,
-			AllowBlobPublicAccess:  &allowBlobPublicAccess,
 		},
 	}
 
+	// For US Government Cloud, don't specify "allow_blob_public_access" and "min_tls_version" in request body.
+	// https://github.com/terraform-providers/terraform-provider-azurerm/issues/7812
+	// https://github.com/terraform-providers/terraform-provider-azurerm/issues/8083
+	if env.Name == autorestAzure.USGovernmentCloud.Name {
```
katbyte, line 661: we can just do this
```suggestion
	if  meta.(*clients.Client).Account.Environment.Name == autorestAzure.USGovernmentCloud.Name {
```

as the env isn't used anywhere else?

azurerm/internal/services/storage/resource_arm_storage_account.go
WodansSon: @mybayern1974 fair enough... ignore # 1

PR #8095, opened 2020-08-12
new resource `azurerm_synapse_sql_pool` 

### LINKED ISSUE #7406: Support for [Azure Synapse Analytics]

njuCZ:
the third resource of #7406

it needs to take a long time to run test for PointInTime restore (8h according to the doc) and recovery (1 day), so I don't add test case for these two create mode. I have tested it offline and there is no problem

image

(fixes #7406)

### DIFF
--- a/azurerm/internal/services/sql/parse/sql_database.go
+++ b/azurerm/internal/services/sql/parse/sql_database.go
@@ -1,0 +1,41 @@
+package parse
+
+import (
+	"fmt"
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+)
+
+type SqlDatabaseId struct {
+	SubscriptionID string
+	ResourceGroup  string
+	ServerName     string
+	Name           string
+}
+
+func SqlDatabaseID(input string) (*SqlDatabaseId, error) {
+	id, err := azure.ParseAzureResourceID(input)
+	if err != nil {
+		return nil, fmt.Errorf("parsing Synapse Sql Pool ID %q: %+v", input, err)
+	}
+
+	sqlDatabaseId := SqlDatabaseId{
+		SubscriptionID: id.SubscriptionID,
+		ResourceGroup:  id.ResourceGroup,
+	}
+	if sqlDatabaseId.ServerName, err = id.PopSegment("servers"); err != nil {
+		return nil, err
+	}
+	if sqlDatabaseId.Name, err = id.PopSegment("databases"); err != nil {
+		return nil, err
+	}
+	if err := id.ValidateNoEmptySegments(input); err != nil {
+		return nil, err
+	}
+
+	return &sqlDatabaseId, nil
+}
+
+func (id *SqlDatabaseId) String() string {
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Sql/servers/%s/databases/%s", id.SubscriptionID, id.ResourceGroup, id.ServerName, id.Name)
+}
--- a/azurerm/internal/services/sql/parse/sql_database.go
+++ b/azurerm/internal/services/sql/parse/sql_database.go
@@ -7,10 +7,21 @@ import (
 )
 
 type SqlDatabaseId struct {
-	SubscriptionID string
-	ResourceGroup  string
-	ServerName     string
-	Name           string
+	ResourceGroup string
+	ServerName    string
+	Name          string
+}
+
+func NewSqlDatabaseID(resourceGroup, serverName, name string) SqlDatabaseId {
+	return SqlDatabaseId{
+		ResourceGroup: resourceGroup,
+		ServerName:    serverName,
+		Name:          name,
+	}
+}
+
+func (id SqlDatabaseId) ID(subscriptionId string) string {
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Sql/servers/%s/databases/%s", subscriptionId, id.ResourceGroup, id.ServerName, id.Name)
 }
 
 func SqlDatabaseID(input string) (*SqlDatabaseId, error) {
@@ -20,8 +31,7 @@ func SqlDatabaseID(input string) (*SqlDatabaseId, error) {
 	}
 
 	sqlDatabaseId := SqlDatabaseId{
-		SubscriptionID: id.SubscriptionID,
-		ResourceGroup:  id.ResourceGroup,
+		ResourceGroup: id.ResourceGroup,
 	}
 	if sqlDatabaseId.ServerName, err = id.PopSegment("servers"); err != nil {
 		return nil, err
@@ -34,8 +44,4 @@ func SqlDatabaseID(input string) (*SqlDatabaseId, error) {
 	}
 
 	return &sqlDatabaseId, nil
-}
-
-func (id *SqlDatabaseId) String() string {
-	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Sql/servers/%s/databases/%s", id.SubscriptionID, id.ResourceGroup, id.ServerName, id.Name)
 }
--- a/azurerm/internal/services/synapse/synapse_sql_pool_resource.go
+++ b/azurerm/internal/services/synapse/synapse_sql_pool_resource.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	postgresValidate "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/postgres/validate"
 	sqlParse "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/sql/parse"
 	sqlValidate "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/sql/validate"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/synapse/parse"
@@ -20,6 +21,12 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+const (
+	DefaultCreateMode            = "Default"
+	RecoveryCreateMode           = "Recovery"
+	PointInTimeRestoreCreateMode = "PointInTimeRestore"
 )
 
 func resourceArmSynapseSqlPool() *schema.Resource {
@@ -42,7 +49,7 @@ func resourceArmSynapseSqlPool() *schema.Resource {
 					return []*schema.ResourceData{d}, err
 				}
 
-				d.Set("create_mode", "Default")
+				d.Set("create_mode", DefaultCreateMode)
 				if v, ok := d.GetOk("create_mode"); ok && v.(string) != "" {
 					d.Set("create_mode", v)
 				}
@@ -56,7 +63,7 @@ func resourceArmSynapseSqlPool() *schema.Resource {
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     true,
-				ValidateFunc: validate.SynapseSqlPoolName,
+				ValidateFunc: validate.SqlPoolName,
 			},
 
 			"synapse_workspace_id": {
@@ -68,8 +75,7 @@ func resourceArmSynapseSqlPool() *schema.Resource {
 
 			"sku_name": {
 				Type:     schema.TypeString,
-				Optional: true,
-				Default:  "DW1000c",
+				Required: true,
 				ValidateFunc: validation.StringInSlice([]string{
 					"DW100c",
 					"DW200c",
@@ -93,35 +99,60 @@ func resourceArmSynapseSqlPool() *schema.Resource {
 			"create_mode": {
 				Type:     schema.TypeString,
 				Optional: true,
-				Default:  "Default",
+				Default:  DefaultCreateMode,
 				ForceNew: true,
 				ValidateFunc: validation.StringInSlice([]string{
-					"Default",
-					"Recovery",
-					"PointInTimeRestore",
+					DefaultCreateMode,
+					RecoveryCreateMode,
+					PointInTimeRestoreCreateMode,
 				}, false),
 			},
 
 			"collation": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
-				ForceNew: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ForceNew:     true,
+				ValidateFunc: postgresValidate.PostgresDatabaseCollation,
 			},
 
-			"source_database_id": {
-				Type:     schema.TypeString,
-				Optional: true,
+			"recovery_database_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"restore"},
 				ValidateFunc: validation.Any(
-					validate.SynapseSqlPoolID,
+					validate.SqlPoolID,
 					sqlValidate.SqlDatabaseID,
 				),
 			},
 
-			"restore_point_in_time": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				ValidateFunc: validation.IsRFC3339Time,
+			"restore": {
+				Type:          schema.TypeList,
+				ForceNew:      true,
+				Optional:      true,
+				MaxItems:      1,
+				ConflictsWith: []string{"recovery_database_id"},
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"point_in_time": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ForceNew:     true,
+							ValidateFunc: validation.IsRFC3339Time,
+						},
+
+						"source_database_id": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+							ValidateFunc: validation.Any(
+								validate.SqlPoolID,
+								sqlValidate.SqlDatabaseID,
+							),
+						},
+					},
+				},
 			},
 
 			"data_encrypted": {
@@ -138,16 +169,20 @@ func resourceArmSynapseSqlPoolCreate(d *schema.ResourceData, meta interface{}) e
 	sqlClient := meta.(*clients.Client).Synapse.SqlPoolClient
 	sqlPTDEClient := meta.(*clients.Client).Synapse.SqlPoolTransparentDataEncryptionClient
 	workspaceClient := meta.(*clients.Client).Synapse.WorkspaceClient
+	subscriptionId := meta.(*clients.Client).Account.SubscriptionId
 	ctx, cancel := timeouts.ForCreate(meta.(*clients.Client).StopContext, d)
 	defer cancel()
 
 	name := d.Get("name").(string)
-	workspaceId, _ := parse.SynapseWorkspaceID(d.Get("synapse_workspace_id").(string))
+	workspaceId, err := parse.SynapseWorkspaceID(d.Get("synapse_workspace_id").(string))
+	if err != nil {
+		return err
+	}
 
 	existing, err := sqlClient.Get(ctx, workspaceId.ResourceGroup, workspaceId.Name, name)
 	if err != nil {
 		if !utils.ResponseWasNotFound(existing.Response) {
-			return fmt.Errorf("checking for present of existing Synapse SqlPool %q (Resource Group %q / workspaceName %q): %+v", name, workspaceId.ResourceGroup, workspaceId.Name, err)
+			return fmt.Errorf("checking for presence of existing Synapse Sql Pool %q (workspace %q / Resource Group %q): %+v", name, workspaceId.Name, workspaceId.ResourceGroup, err)
 		}
 	}
 	if existing.ID != nil && *existing.ID != "" {
@@ -156,7 +191,7 @@ func resourceArmSynapseSqlPoolCreate(d *schema.ResourceData, meta interface{}) e
 
 	workspace, err := workspaceClient.Get(ctx, workspaceId.ResourceGroup, workspaceId.Name)
 	if err != nil {
-		return fmt.Errorf("reading Synapse workspace %q (Resource Group %q): %+v", workspaceId.Name, workspaceId.ResourceGroup, err)
+		return fmt.Errorf("retrieving Synapse workspace %q (Resource Group %q): %+v", workspaceId.Name, workspaceId.ResourceGroup, err)
 	}
 
 	mode := d.Get("create_mode").(string)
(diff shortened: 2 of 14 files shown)


azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+				Optional: true,
+			},
+
+			"tags": tags.Schema(),
+		},
+	}
+}
+
+func resourceArmSynapseSqlPoolCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Synapse.SqlPoolClient
```
katbyte, line 138: could we use 
```suggestion
	sqlClient := meta.(*clients.Client).Synapse.SqlPoolClient
```

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+	sqlPoolTransparentDataEncryptionClient := meta.(*clients.Client).Synapse.SqlPoolTransparentDataEncryptionClient
```
katbyte, line 139: and 
```suggestion
	sqlPTDEClient := meta.(*clients.Client).Synapse.SqlPoolTransparentDataEncryptionClient
```

website/docs/r/synapse_sql_pool.html.markdown
```diff
@@ -0,0 +1,92 @@
+---
+subcategory: "Synapse"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_synapse_sql_pool"
+description: |-
+  Manages a synapse Sql Pool.
```
katbyte, line 6:
```suggestion
  Manages a Synapse Sql Pool.
```

website/docs/r/synapse_sql_pool.html.markdown
```diff
@@ -0,0 +1,92 @@
...
+---
+
+# azurerm_synapse_sql_pool
+
+Manages a synapse Sql Pool.
```
katbyte, line 11:
```suggestion
Manages a Synapse Sql Pool.
```

website/docs/r/synapse_sql_pool.html.markdown
```diff
@@ -0,0 +1,92 @@
...
+
+* `name` - (Required) The name which should be used for this Synapse Sql Pool. Changing this forces a new synapse SqlPool to be created.
+
+* `synapse_workspace_id` - (Required) The resource ID of Synapse Workspace. Changing this forces a new synapse SqlPool to be created.
+
+* `sku_name` - (Optional) Specifies the SKU Name for this Synapse Sql Pool. Possible values are `DW100c`, `DW200c`, `DW300c`, `DW400c`, `DW500c`, `DW1000c`, `DW1500c`, `DW2000c`, `DW2500c`, `DW3000c`, `DW5000c`, `DW6000c`, `DW7500c`, `DW10000c`, `DW15000c` or `DW30000c`. Defaults to `DW1000c`.
+
+* `create_mode` - (Optional) Specifies how to create the Sql Pool. Valid values are: `Default`, `Recovery` or `PointInTimeRestore`. Must be `Default` to create a new database. Defaults to `Default`.
+
+* `collation` - (Optional) The name of the collation. Applies only if `create_mode` is `Default`.  Azure default is `SQL_LATIN1_GENERAL_CP1_CI_AS`. Changing this forces a new resource to be created.
```
katbyte, line 63:
```suggestion
* `collation` - (Optional) The name of the collation to use with this pool. Applies only if `create_mode` is `Default`.  Azure default is `SQL_LATIN1_GENERAL_CP1_CI_AS`. Changing this forces a new resource to be created.
```

website/docs/r/synapse_sql_pool.html.markdown
```diff
@@ -0,0 +1,92 @@
...
+
+* `source_database_id` - (Optional) The ID of the source_database which is to recovery or back up.
```
katbyte, line 65: Could we clarify this description?

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+			},
+
+			"collation": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"source_database_id": {
```
katbyte, line 112: Would this make more sense as 
```suggestion
			"restore_source_database_id": {
```

njuCZ: according to the service team, sql pool backend service is a proxy to sql database. In the azurerm_sql_database, the field is `source_database_id`, so I think we would better keep them the same ?

njuCZ:
@katbyte I have refined my PR, could you please have a look again?

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.SynapseWorkspaceID,
+			},
+
+			"sku_name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "DW1000c",
```
tombuildsstuff, line 72: this shouldn't be defaulted (and so this field should be Required) - whilst this is defaulted in the Portal this is done for convenience, particularly with a large SKU this could leave users with a large bill (this costs $18/hour / $12,960 a month) - can we update this to be required

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+			"restore_point_in_time": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.IsRFC3339Time,
+			},
+
+			"data_encrypted": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
```
tombuildsstuff, line 130: this looks like a value that'll change in the future:

a) is this defaulted off?
b) is this going to change to `on` in the future?
c) is this not a ForceNew field like in many other API's?

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+
+func resourceArmSynapseSqlPoolCreate(d *schema.ResourceData, meta interface{}) error {
+	sqlClient := meta.(*clients.Client).Synapse.SqlPoolClient
+	sqlPTDEClient := meta.(*clients.Client).Synapse.SqlPoolTransparentDataEncryptionClient
+	workspaceClient := meta.(*clients.Client).Synapse.WorkspaceClient
+	ctx, cancel := timeouts.ForCreate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	name := d.Get("name").(string)
+	workspaceId, _ := parse.SynapseWorkspaceID(d.Get("synapse_workspace_id").(string))
```
tombuildsstuff, line 145: we should still be checking this error - even if it /should/ be caught above

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+
+	existing, err := sqlClient.Get(ctx, workspaceId.ResourceGroup, workspaceId.Name, name)
+	if err != nil {
+		if !utils.ResponseWasNotFound(existing.Response) {
+			return fmt.Errorf("checking for present of existing Synapse SqlPool %q (Resource Group %q / workspaceName %q): %+v", name, workspaceId.ResourceGroup, workspaceId.Name, err)
```
tombuildsstuff, line 150:
```suggestion
			return fmt.Errorf("checking for presence of existing Synapse Sql Pool %q (Workspace %q / Resource Group %q): %+v", name, workspaceId.Name, workspaceId.ResourceGroup, err)
```

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+		}
+	}
+	if existing.ID != nil && *existing.ID != "" {
+		return tf.ImportAsExistsError("azurerm_synapse_sql_pool", *existing.ID)
+	}
+
+	workspace, err := workspaceClient.Get(ctx, workspaceId.ResourceGroup, workspaceId.Name)
+	if err != nil {
+		return fmt.Errorf("reading Synapse workspace %q (Resource Group %q): %+v", workspaceId.Name, workspaceId.ResourceGroup, err)
```
tombuildsstuff, line 159:
```suggestion
		return fmt.Errorf("retrieving Synapse workspace %q (Resource Group %q): %+v", workspaceId.Name, workspaceId.ResourceGroup, err)
```

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+			return fmt.Errorf("`source_database_id` must be set when `create_mode` is `Recovery`")
+		}
+		sqlPoolInfo.SQLPoolResourceProperties.RecoverableDatabaseID = utils.String(sourceDatabaseId)
+	case "PointInTimeRestore":
+		sourceDatabaseId := constructSourceDatabaseId(d.Get("source_database_id").(string))
+		restorePointInTimeStr := d.Get("restore_point_in_time").(string)
+		if sourceDatabaseId == "" || restorePointInTimeStr == "" {
+			return fmt.Errorf("`source_database_id` and `restore_point_in_time` must be set when `create_mode` is `PointInTimeRestore`")
+		}
+		restorePointInTime, _ := time.Parse(time.RFC3339, restorePointInTimeStr)
```
tombuildsstuff, line 189: we should be catching this error

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+	switch mode {
+	case "Default":
+		sqlPoolInfo.SQLPoolResourceProperties.Collation = utils.String(d.Get("collation").(string))
+	case "Recovery":
+		sourceDatabaseId := constructSourceDatabaseId(d.Get("source_database_id").(string))
+		if sourceDatabaseId == "" {
+			return fmt.Errorf("`source_database_id` must be set when `create_mode` is `Recovery`")
+		}
+		sqlPoolInfo.SQLPoolResourceProperties.RecoverableDatabaseID = utils.String(sourceDatabaseId)
+	case "PointInTimeRestore":
```
tombuildsstuff, line 183: rather than using magic strings here - can we define constants for these

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+		return fmt.Errorf("waiting on creating future for Synapse SqlPool %q (Resource Group %q / workspaceName %q): %+v", name, workspaceId.ResourceGroup, workspaceId.Name, err)
+	}
+
+	if d.Get("data_encrypted").(bool) {
+		parameter := synapse.TransparentDataEncryption{
+			TransparentDataEncryptionProperties: &synapse.TransparentDataEncryptionProperties{
+				Status: synapse.TransparentDataEncryptionStatusEnabled,
+			},
+		}
+		if _, err := sqlPTDEClient.CreateOrUpdate(ctx, workspaceId.ResourceGroup, workspaceId.Name, name, parameter); err != nil {
```
tombuildsstuff, line 208: since this is a separate client - this'll likely make sense as a separate resource (since it's likely this resource will support MSI / defining a user-defined key to back this in time?)

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+
+	if d.HasChanges("sku_name", "tags") {
+		sqlPoolInfo := synapse.SQLPoolPatchInfo{
+			Sku: &synapse.Sku{
+				Name: utils.String(d.Get("sku_name").(string)),
+			},
+			Tags: tags.Expand(d.Get("tags").(map[string]interface{})),
+		}
+
+		if _, err := sqlClient.Update(ctx, id.Workspace.ResourceGroup, id.Workspace.Name, id.Name, sqlPoolInfo); err != nil {
```
tombuildsstuff, line 261: this looks like a Swagger/API bug since this should be a Long Running Operation?

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+
+	id, err := parse.SynapseSqlPoolID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	resp, err := sqlClient.Get(ctx, id.Workspace.ResourceGroup, id.Workspace.Name, id.Name)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[INFO] synapse %q does not exist - removing from state", d.Id())
```
tombuildsstuff, line 303:
```suggestion
			log.Printf("[INFO] Synapse SQL Pool %q (Workspace %q / Resource Group %q) does not exist - removing from state", id.Name, id.Workspace.Name, id.Workspace.ResourceGroup)
```

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("retrieving Synapse SqlPool %q (Resource Group %q / workspaceName %q): %+v", id.Name, id.Workspace.ResourceGroup, id.Workspace.Name, err)
```
tombuildsstuff, line 307:
```suggestion
		return fmt.Errorf("retrieving Synapse SqlPool %q (Workspace %q / Resource Group %q): %+v", id.Name, id.Workspace.Name, id.Workspace.ResourceGroup, err)
```

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+	if err != nil {
+		return id
+	}
+	sqlDataBaseId := sqlParse.SqlDatabaseId{
+		SubscriptionID: sqlPoolId.Workspace.SubscriptionID,
+		ResourceGroup:  sqlPoolId.Workspace.ResourceGroup,
+		ServerName:     sqlPoolId.Workspace.Name,
+		Name:           sqlPoolId.Name,
+	}
+	return sqlDataBaseId.String()
```
tombuildsstuff, line 377: rather than doing this - can we switch to using [the ResourceIDFormatter approach](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/frontdoor/parse/health_probe.go#L5-L22) which [ensures they comply with the interface](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/frontdoor/parse/health_probe_test.go#L9)

azurerm/internal/services/synapse/tests/synapse_sql_pool_resource_test.go
```diff
@@ -0,0 +1,223 @@
...
+			if !utils.ResponseWasNotFound(resp.Response) {
+				return fmt.Errorf("bad: Synapse SqlPool %q does not exist", id.Name)
+			}
+			return fmt.Errorf("bad: Get on Synapse.SqlPoolClient: %+v", err)
+		}
+		return nil
+	}
+}
+
+func testCheckAzureRMsynapseSqlPoolDestroy(s *terraform.State) error {
```
tombuildsstuff, line 123:
```suggestion
func testCheckAzureRMSynapseSqlPoolDestroy(s *terraform.State) error {
```

azurerm/internal/services/synapse/tests/synapse_sql_pool_resource_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+func testAccAzureRMSynapseSqlPool_requiresImport(data acceptance.TestData) string {
+	config := testAccAzureRMSynapseSqlPool_basic(data)
+	return fmt.Sprintf(`
+%s
+
+resource "azurerm_synapse_sql_pool" "import" {
+  name                 = azurerm_synapse_sql_pool.test.name
+  synapse_workspace_id = azurerm_synapse_sql_pool.test.synapse_workspace_id
+}
```
tombuildsstuff, line 165: what about the other fields here?

azurerm/internal/services/synapse/tests/synapse_sql_pool_resource_test.go
```diff
@@ -0,0 +1,223 @@
...
+  }
+}
+`, template, data.RandomString)
+}
+
+func testAccAzureRMSynapseSqlPool_template(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
```
tombuildsstuff, line 193: the provider block should be per-test (since this can vary on a per-test-case basis) - not in the template

azurerm/internal/services/synapse/validate/synapse_sql_pool.go
```diff
@@ -0,0 +1,42 @@
+package validate
```
tombuildsstuff, line 1: we can remove the `synapse` prefix here - we're in the `synapse` package

azurerm/internal/services/synapse/validate/synapse_sql_pool.go
```diff
@@ -0,0 +1,42 @@
...
+func SynapseSqlPoolName(i interface{}, k string) (warnings []string, errors []error) {
+	v, ok := i.(string)
+	if !ok {
+		errors = append(errors, fmt.Errorf("expected type of %s to be string", k))
+		return
+	}
+
+	// The name attribute rules are :
+	// 1. can contain only letters, numbers and underscore.
+	// 3. The value must be between 1 and 15 characters long
```
tombuildsstuff, line 19: where's 2? 🙃

website/docs/r/synapse_sql_pool.html.markdown
```diff
@@ -0,0 +1,92 @@
...
+}
+
+resource "azurerm_storage_account" "example" {
+  name                     = "examplestorageacc"
+  resource_group_name      = azurerm_resource_group.example.name
+  location                 = azurerm_resource_group.example.location
+  account_tier             = "Standard"
+  account_replication_type = "LRS"
+  account_kind             = "StorageV2"
+  is_hns_enabled           = "true"
```
tombuildsstuff, line 28: this is a bool

```suggestion
  is_hns_enabled           = true
```

website/docs/r/synapse_sql_pool.html.markdown
```diff
@@ -0,0 +1,92 @@
...
+  resource_group_name                  = azurerm_resource_group.example.name
+  location                             = azurerm_resource_group.example.location
+  storage_data_lake_gen2_filesystem_id = azurerm_storage_data_lake_gen2_filesystem.example.id
+  sql_administrator_login              = "sqladminuser"
+  sql_administrator_login_password     = "H@Sh1CoR3!"
+}
+
+resource "azurerm_synapse_sql_pool" "example" {
+  name                 = "examplesqlpool"
+  synapse_workspace_id = azurerm_synapse_workspace.example.id
```
tombuildsstuff, line 47: what about the other fields here?

website/docs/r/synapse_sql_pool.html.markdown
```diff
@@ -0,0 +1,92 @@
...
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name which should be used for this Synapse Sql Pool. Changing this forces a new synapse SqlPool to be created.
+
+* `synapse_workspace_id` - (Required) The resource ID of Synapse Workspace. Changing this forces a new synapse SqlPool to be created.
```
tombuildsstuff, line 57: we shouldn't be using the term "Resource ID" because it's ambiguous between Terraform and Azure - and they can diverge

```suggestion
* `synapse_workspace_id` - (Required) The ID of Synapse Workspace within which this Sql Pool should be created. Changing this forces a new Synapse Sql Pool to be created.
```

website/docs/r/synapse_sql_pool.html.markdown
```diff
@@ -0,0 +1,92 @@
...
+
+* `sku_name` - (Optional) Specifies the SKU Name for this Synapse Sql Pool. Possible values are `DW100c`, `DW200c`, `DW300c`, `DW400c`, `DW500c`, `DW1000c`, `DW1500c`, `DW2000c`, `DW2500c`, `DW3000c`, `DW5000c`, `DW6000c`, `DW7500c`, `DW10000c`, `DW15000c` or `DW30000c`. Defaults to `DW1000c`.
```
tombuildsstuff, line 59: (as above) we shouldn't default a SKU here - particularly to one which costs $18/hour / ~$13,000 a month

website/docs/r/synapse_sql_pool.html.markdown
```diff
@@ -0,0 +1,92 @@
...
+
+* `create_mode` - (Optional) Specifies how to create the Sql Pool. Valid values are: `Default`, `Recovery` or `PointInTimeRestore`. Must be `Default` to create a new database. Defaults to `Default`.
+
+* `collation` - (Optional) The name of the collation to use with this pool. Applies only if `create_mode` is `Default`.  Azure default is `SQL_LATIN1_GENERAL_CP1_CI_AS`. Changing this forces a new resource to be created.
```
tombuildsstuff, line 63: we should default this within Terraform too:

```suggestion
* `collation` - (Optional) The name of the collation to use with this pool, only applicable when `create_mode` is set to `Default`. Defaults to `SQL_LATIN1_GENERAL_CP1_CI_AS`. Changing this forces a new resource to be created.
```

azurerm/internal/services/sql/parse/sql_database.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	if sqlDatabaseId.ServerName, err = id.PopSegment("servers"); err != nil {
+		return nil, err
+	}
+	if sqlDatabaseId.Name, err = id.PopSegment("databases"); err != nil {
+		return nil, err
+	}
+	if err := id.ValidateNoEmptySegments(input); err != nil {
+		return nil, err
+	}
```
tombuildsstuff, line 34: can we add tests for this method?

azurerm/internal/services/sql/parse/sql_database.go
```diff
@@ -0,0 +1,41 @@
...
+
+	return &sqlDatabaseId, nil
+}
+
+func (id *SqlDatabaseId) String() string {
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Sql/servers/%s/databases/%s", id.SubscriptionID, id.ResourceGroup, id.ServerName, id.Name)
+}
```
tombuildsstuff, line 41: we've got a ResourceID Formatter type which this can comply with now - [example](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/frontdoor/parse/health_probe.go#L5-L22) and [example of the test asserting this conforms to the type](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/frontdoor/parse/health_probe_test.go#L9) - can we update this?

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+					"Recovery",
+					"PointInTimeRestore",
+				}, false),
+			},
+
+			"collation": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
```
tombuildsstuff, line 109: (as below) this should probably be defaulted

I believe we've also got validation for collations somewhere in the codebase we could use here

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -0,0 +1,378 @@
...
+					validate.SynapseSqlPoolID,
+					sqlValidate.SqlDatabaseID,
+				),
+			},
+
+			"restore_point_in_time": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.IsRFC3339Time,
+			},
```
tombuildsstuff, line 125: perhaps this and the "restore id" could be moved into a "restore" block taking a "restore_mode" of "PointInTime" or "Recovery" - making both fields required - rather than having 3 top level fields?

tombuildsstuff (changes requested):
hi @njuCZ 

Thanks for this PR.

I've taken a look through here and left some comments inline - but this is off to a good start. If we can fix up the comments then we should be able to take another look through here.

Thanks!

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
njuCZ: for now, this client could only update `data_encrypted` field, and there is no delete function for it. It's more likely a property of sql pool. So I think  it's better to combine it into this resource ?

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
njuCZ: after testing the api or trying in the portal, I found only changing sku is a long running operation (in the portal, it's called `scale`). Updating other fields is fast. But there are the same rest api. there is not a standalone api for changing sku.

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
njuCZ: a) for now, it's default "off"
b) I am sorry that I don't know whether it will change default value in the future
c) it's a updatable field. I don't see similiar fields in other API?

njuCZ:
@tombuildsstuff I have modifed this PR, could you please have a look again?

jackofallops:
Hi @njuCZ - Is there a reason this is leveraging/updating `sql` rather than the newer `mssql` ?

katbyte (approved):
Thanks @njuCZ  - LGTM 👍

njuCZ:
@jackofallops sorry that I understand the relation between sql and mssql just now. Since this PR has been merged, and only parse function and validation function from sql package is used. I will change to use functions from mssql. This is not breaking change and I will submit follow up PR to do this #8434 . Thanks for your suggestion

PR #8115, opened 2020-08-13
Update private_dns_zone.html.markdown

akingscote:


tombuildsstuff:
hey @akingscote 

Thanks for this PR :) 

Terraform supports both Arguments and Attributes on resources - that is parameters which can be specified (Arguments) and are exported/read-only fields (Attributes). All Arguments which can be provided to a resource are automatically available as Attributes - therefore to avoid duplication in the documentation we intentionally only only Attributes rather than re-listing the Arguments. Since in this case the `name` parameter is an Argument - it'll automatically be available as an Attribute too - and so we intentionally don't document that in the Attributes section.

As such, whilst I'd like to thank you for this contribution - and on the whole it looks good - since this is intentionally not documented within the Attributes block I'm going to close this PR for the moment.

Thanks!

PR #8125, opened 2020-08-14
New resource `azurerm_mysql_server_key` - Add CMK support

ArcturusZhang:
Adding customer managed key support for mySQL server


azurerm/internal/services/mysql/mysql_server_key_resource.go
```diff
@@ -0,0 +1,194 @@
...
+			"key_vault_key_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: azure.ValidateKeyVaultChildId,
+			},
+		},
+	}
+}
+
+func getMySQLServerKeyName(ctx context.Context, vaultsClient *keyvault.VaultsClient, keyVaultKeyURI string) (string, error) {
```
tombuildsstuff, line 60: we use a convention of returning a pointer and an error - thus if someone tries to use the result of an object without checking the error, the object should be nil and the code should crash (which'll get caught during tests) - as such can we update this?

```suggestion
func getMySQLServerKeyName(ctx context.Context, vaultsClient *keyvault.VaultsClient, keyVaultKeyURI string) (*string, error) {
```

azurerm/internal/services/mysql/mysql_server_key_resource.go
```diff
@@ -0,0 +1,194 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.MysqlServerServerID,
+			},
+
+			"key_vault_key_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: azure.ValidateKeyVaultChildId,
```
tombuildsstuff, line 54: does this API support the "versionless" key vault references? if so we'll need to look this version up in the create to find it's ID (but send the versionless uri?)

azurerm/internal/services/mysql/mysql_server_key_resource.go
```diff
@@ -0,0 +1,194 @@
...
+	vaultsClient := meta.(*clients.Client).KeyVault.VaultsClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	serverID, err := parse.MysqlServerServerID(d.Get("server_id").(string))
+	if err != nil {
+		return err
+	}
+	keyVaultKeyURI := d.Get("key_vault_key_id").(string)
+	name, err := getMySQLServerKeyName(ctx, vaultsClient, keyVaultKeyURI)
```
tombuildsstuff, line 87: we could probably simplify this to only use the version of the secret id instead, since we're not using anything else?

azurerm/internal/services/mysql/mysql_server_key_resource.go
```diff
@@ -0,0 +1,194 @@
...
+		Importer: azSchema.ValidateResourceIDPriorToImport(func(id string) error {
+			_, err := parse.MySQLServerKeyID(id)
+			return err
+		}),
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
```
tombuildsstuff, line 40: these timeouts feel optimistic for encrypting/re-encrypting/decrypting larger DB's - do we know how long this is likely to take? generally speaking we use `{expected/average time} * 2` to allow for rate-limiting and slower deployments at peak times

azurerm/internal/services/mysql/mysql_server_key_resource.go
```diff
@@ -0,0 +1,194 @@
...
+	client := meta.(*clients.Client).MySQL.ServerKeysClient
+	ctx, cancel := timeouts.ForDelete(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	id, err := parse.MySQLServerKeyID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	future, err := client.Delete(ctx, id.ServerName, id.Name, id.ResourceGroup)
```
tombuildsstuff, line 185: presumably we'll also need to lock here

azurerm/internal/services/mysql/mysql_server_resource.go
```diff
@@ -24,6 +24,10 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
+const (
+	mySQLServerResourceName = "azurerm_mysql_server"
+)
```
tombuildsstuff, line 29: presumably we'll also need to lock in the create/update/delete here too?

tombuildsstuff (changes requested):
hey @ArcturusZhang 

Thanks for this PR 👍 

I've taken a look through and left some comments inline but this is looking good - if we can work through those then we should be able to take another look 👍 

Thanks!

azurerm/internal/services/mysql/mysql_server_key_resource.go
ArcturusZhang: Done. And I will do the same in postgresql CMK PR

azurerm/internal/services/mysql/mysql_server_key_resource.go
ArcturusZhang: Unfortunately the API requires a key URL with version.

azurerm/internal/services/mysql/mysql_server_key_resource.go
ArcturusZhang: Done. Aligned with MySql Server timeout

azurerm/internal/services/mysql/mysql_server_key_resource.go
ArcturusZhang: I am not quite sure if I get the point - I get the way we are composing the key name here from Azure CLI, they are making the key name in this pattern: vaultname_keyname_keyversion

It is curious that the service does not accept just a random name - internally they must have a pattern for this name. I wrote an email to the service team, asked about exactly what this pattern is. But the contact told me she did not know much about this. Therefore I just took the way in Azure CLI here.

azurerm/internal/services/mysql/mysql_server_key_resource.go
ArcturusZhang: Done

azurerm/internal/services/mysql/mysql_server_resource.go
ArcturusZhang: I assume we do not need locks in postgresql server.

This key resource has a kind of weird behaviour - it is a separated resource from the server by service design, but it is a singleton, you could only have one key at one time. If you are trying to create two keys (with different name) for one server, after the second key is created on the service side, the first key will be automatically removed without any notice.

Therefore we need locks in the server key resource to ensure that if the user is not using this resource correctly (by say create two keys for one server), the order of key creation is predictable - this will always give the user a diff after apply.

The server does not have this kind of constraint, therefore I suppose the server resource does not need locks.

ArcturusZhang:
Hi @tombuildsstuff I have made some changes, please have a look, thanks

azurerm/internal/services/mysql/mysql_server_resource.go
tombuildsstuff: so what happens if you try and decrypt the server and delete it simultaneously? a conflict will likely occur somewhere

ArcturusZhang: By decrypting the server, I assume we are deleting the server key  - since we locked the server in the delete function of server key resource, we could not decrypt the server and delete it simultaneously - during the decrypting, the server is locked and deletion should hold on.

tombuildsstuff: can we test this to confirm?

tombuildsstuff: @ArcturusZhang any update here?

ArcturusZhang: Tests have been done per requested.

katbyte (approved):
Thanks @ArcturusZhang - LGTM 🚀

PR #8126, opened 2020-08-14
New resource `azurerm_postgresql_server_key` - Add CMK support

ArcturusZhang:
Adding customer managed key support for postgreSQL Server
Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7811


azurerm/internal/services/postgres/postgresql_server_key_resource.go
```diff
@@ -0,0 +1,193 @@
...
+			_, err := parse.PostgreSQLServerKeyID(id)
+			return err
+		}),
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
```
tombuildsstuff, line 41: Are these timeouts realistic for encrypting / re-encrypting / decrypting databases - particularly large ones with lots of data?

azurerm/internal/services/postgres/postgresql_server_key_resource.go
```diff
@@ -0,0 +1,193 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.PostgresServerServerID,
+			},
+
+			"key_vault_key_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: azure.ValidateKeyVaultChildId,
```
tombuildsstuff, line 54: Should this also support the "versionless" secrets? If so we'll likely need to look up the current version to be able to use that as a name?

azurerm/internal/services/postgres/postgresql_server_key_resource.go
```diff
@@ -0,0 +1,193 @@
...
+	client := meta.(*clients.Client).Postgres.ServerKeysClient
+	ctx, cancel := timeouts.ForDelete(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	id, err := parse.PostgreSQLServerKeyID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	future, err := client.Delete(ctx, id.ServerName, id.Name, id.ResourceGroup)
```
tombuildsstuff, line 184: Do we also need to lock here?

azurerm/internal/services/postgres/parse/server_key.go
```diff
@@ -0,0 +1,38 @@
...
+type PostgreSQLServerKeyId struct {
+	Name          string
+	ServerName    string
+	ResourceGroup string
+}
+
+func PostgreSQLServerKeyID(input string) (*PostgreSQLServerKeyId, error) {
+	id, err := azure.ParseAzureResourceID(input)
+	if err != nil {
+		return nil, fmt.Errorf("[ERROR] Unable to parse Postgres Server ID %q: %+v", input, err)
```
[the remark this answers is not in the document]
ArcturusZhang: Done

azurerm/internal/services/postgres/postgresql_server_key_resource.go
```diff
@@ -0,0 +1,193 @@
...
+	return fmt.Sprintf("%s_%s_%s", keyVaultID.Name, keyVaultKeyID.Name, keyVaultKeyID.Version), nil
+}
+
+func resourceArmPostgreSQLServerKeyCreateUpdate(d *schema.ResourceData, meta interface{}) error {
+	keysClient := meta.(*clients.Client).Postgres.ServerKeysClient
+	vaultsClient := meta.(*clients.Client).KeyVault.VaultsClient
+	ctx, cancel := timeouts.ForCreate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	serverID, err := parse.PostgresServerServerID(d.Get("server_id").(string))
```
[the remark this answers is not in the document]
ArcturusZhang: Done - and I will do the same for MySQL also

azurerm/internal/services/postgres/postgresql_server_key_resource.go
ArcturusZhang: By saying `key_vault_key_id` here, it means some key URL WITH version like this:
```
https://YourVaultName.vault.azure.net/keys/YourKeyName/01234567890123456789012345678901>
```
Do you suggest we change this schema to align with the CMK for storage? i.e, like this:
```
key_vault_id = 
key_name = 
key_version = 
```
to emphasize that we are requiring a versioned key here?

azurerm/internal/services/postgres/postgresql_server_key_resource.go
ArcturusZhang: Updated the timeout to align with the timeout in postgresql server.

azurerm/internal/services/postgres/postgresql_server_key_resource.go
ArcturusZhang: Added

ArcturusZhang:
Hi @tombuildsstuff I have made some changes, please have a look.

About the version of keys, this resource is requiring a key URL with version. Therefore it is possible that we could change the schema a little bit to align with the CMK in storage account (details in review comments). What do you think?

azurerm/internal/services/postgres/postgresql_server_key_resource.go
tombuildsstuff: right however most Azure API's which support a versioned secret also take a "versionless" version too (since the 'version' gets parsed as empty, which the Key Vault API's return as "latest") - so we should test that

No, this makes sense to leave as `key_vault_key_id` - however we need to confirm if this validation needs to accept both a versioned and versionless Key ID, which we'll identify through testing that?

ArcturusZhang: Sure I will have a try with this

ArcturusZhang: Just have a try if we pass a key URL without version, and get an `internalServerError`.
And per their [document page](https://docs.microsoft.com/en-us/azure/postgresql/howto-data-encryption-cli#set-data-encryption-for-azure-database-for-postgresql-single-server), they should be expecting the key URL with version

katbyte: Then should we be validing that the id has a version then?

ArcturusZhang: This function `ValidateKeyVaultChildId` is expecting this ID to have a version. We have a `ValidateKeyVaultChildIdVersionOptional` for the version is optional scenario

katbyte (changes requested):
Thanks @ArcturusZhang  - however it looks like multiple tests are failing with:
```
 Error: creating PostgreSQL Server "acctest-psql-server-200919181404659776" (Resource Group "acctestRG-psql-200919181404659776"): postgresql.ServersClient#Create: Failure sending request: StatusCode=405 -- Original Error: Code="FeatureSwitchNotEnabled" Message="Requested feature is not enabled"
```

Is there something we need to enable? and how can we add this new feature without breaking existing users?

katbyte (approved):
LGTM as the test failure is unrelated 👍

PR #8134, opened 2020-08-14
compute: add encryption at host

JulienBalestra:
The `EncryptionAtHost` is apparently available since the version of [2020-06-01](https://github.com/Azure/azure-sdk-for-go/tree/master/services/compute/mgmt/2020-06-01/compute).

This PR is upgrading the compute 2019-12-01 to the aforementioned one to be able to specify `encryption_at_host` in all supported places.

I believe this kind of library upgrade is a non trivial task to achieve so I'm open to split this PR in 2 parts as done currently with the 2 commits.

Tested/validated the plugin with some supported and unsupported instance types:

```
Error: Error creating Linux Virtual Machine Scale Set "encrypted-3-20200814131351" (Resource Group "parent1-k8s"): compute.VirtualMachineScaleSetsClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="NotSupported" Message="Encryption at host is not supported for VM size Standard_D4_v3."
```


azurerm/internal/services/compute/dedicated_host_group_data_source.go
```diff
@@ -61,7 +63,7 @@ func dataSourceArmDedicatedHostGroupRead(d *schema.ResourceData, meta interface{
 	name := d.Get("name").(string)
 	resourceGroupName := d.Get("resource_group_name").(string)
 
-	resp, err := client.Get(ctx, resourceGroupName, name)
+	resp, err := client.Get(ctx, resourceGroupName, name, compute.InstanceView)
```
ArcturusZhang, line 66: To keep consistency with the old api-version, could we use an empty string for this new parameter?
```suggestion
	resp, err := client.Get(ctx, resourceGroupName, name, "")
```

azurerm/internal/services/compute/dedicated_host_group_resource.go
```diff
@@ -75,7 +75,7 @@ func resourceArmDedicatedHostGroupCreate(d *schema.ResourceData, meta interface{
 	resourceGroupName := d.Get("resource_group_name").(string)
 
 	if d.IsNewResource() {
-		existing, err := client.Get(ctx, resourceGroupName, name)
+		existing, err := client.Get(ctx, resourceGroupName, name, compute.InstanceView)
```
ArcturusZhang, line 78: Same here
```suggestion
		existing, err := client.Get(ctx, resourceGroupName, name, "")
```

azurerm/internal/services/compute/dedicated_host_group_resource.go
```diff
@@ -105,7 +105,7 @@ func resourceArmDedicatedHostGroupCreate(d *schema.ResourceData, meta interface{
 		return fmt.Errorf("Error creating Dedicated Host Group %q (Resource Group %q): %+v", name, resourceGroupName, err)
 	}
 
-	resp, err := client.Get(ctx, resourceGroupName, name)
+	resp, err := client.Get(ctx, resourceGroupName, name, compute.InstanceView)
```
ArcturusZhang, line 108: Same here
```suggestion
	resp, err := client.Get(ctx, resourceGroupName, name, "")
```

azurerm/internal/services/compute/dedicated_host_group_resource.go
```diff
@@ -129,7 +129,7 @@ func resourceArmDedicatedHostGroupRead(d *schema.ResourceData, meta interface{})
 	resourceGroupName := id.ResourceGroup
 	name := id.Path["hostGroups"]
 
-	resp, err := client.Get(ctx, resourceGroupName, name)
+	resp, err := client.Get(ctx, resourceGroupName, name, compute.InstanceView)
```
ArcturusZhang, line 132: Same here
```suggestion
	resp, err := client.Get(ctx, resourceGroupName, name, "")
```

azurerm/internal/services/compute/dedicated_host_resource.go
```diff
@@ -172,7 +172,7 @@ func resourceArmDedicatedHostRead(d *schema.ResourceData, meta interface{}) erro
 		return err
 	}
 
-	group, err := groupsClient.Get(ctx, id.ResourceGroup, id.HostGroup)
+	group, err := groupsClient.Get(ctx, id.ResourceGroup, id.HostGroup, compute.InstanceView)
```
ArcturusZhang, line 175: Same here
```suggestion
	group, err := groupsClient.Get(ctx, id.ResourceGroup, id.HostGroup, "")
```

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
```diff
@@ -218,9 +218,9 @@ func resourceArmLinuxVirtualMachineScaleSet() *schema.Resource {
 				ForceNew: true,
 				Default:  string(compute.Manual),
 				ValidateFunc: validation.StringInSlice([]string{
-					string(compute.Automatic),
+					string(compute.AutomaticByOS),
 					string(compute.Manual),
-					string(compute.Rolling),
+					string(compute.AutomaticByPlatform),
```
ArcturusZhang, line 223: These changes are not correct - it should be 
``` suggestion
string(compute.UpgradeModeAutomatic),
string(compute.UpgradeModeManual),
string(compute.UpgradeModeRolling),
```

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
```diff
@@ -251,6 +251,13 @@ func resourceArmLinuxVirtualMachineScaleSet() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"encryption_at_host": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: false,
+				Default:  false,
```
ArcturusZhang, line 259: These two flags could be omitted, since these are their default values

azurerm/internal/services/compute/virtual_machine_scale_set_resource.go
```diff
@@ -135,9 +135,9 @@ func resourceArmVirtualMachineScaleSet() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 				ValidateFunc: validation.StringInSlice([]string{
-					string(compute.Automatic),
+					string(compute.AutomaticByOS),
 					string(compute.Manual),
-					string(compute.Rolling),
+					string(compute.AutomaticByPlatform),
```
ArcturusZhang, line 140: Same here, should be `compute.UpgradeModeAutomatic`, `compute,UpgradeModeManual` and `compute.UpgradeModeRolling`

azurerm/internal/services/compute/virtual_machine_scale_set_resource.go
```diff
@@ -778,6 +778,13 @@ func resourceArmVirtualMachineScaleSet() *schema.Resource {
 			},
 
 			"tags": tags.Schema(),
+
+			"encryption_at_host": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: false,
+				Default:  false,
+			},
```
ArcturusZhang, line 787: The old `azurerm_virtual_machine_scale_set` resource is currently in the feature-frozen state, it should not receive any new features. You can find the information [here](https://www.terraform.io/docs/providers/azurerm/r/virtual_machine_scale_set.html).

Therefore could we revert the feature changes in this file (such like this)?

azurerm/internal/services/compute/virtual_machine_scale_set_resource.go
```diff
@@ -858,7 +865,10 @@ func resourceArmVirtualMachineScaleSetCreateUpdate(d *schema.ResourceData, meta
 			RollingUpgradePolicy: expandAzureRmRollingUpgradePolicy(d),
 		},
 		VirtualMachineProfile: &compute.VirtualMachineScaleSetVMProfile{
-			NetworkProfile:   expandAzureRmVirtualMachineScaleSetNetworkProfile(d),
+			NetworkProfile: expandAzureRmVirtualMachineScaleSetNetworkProfile(d),
+			SecurityProfile: &compute.SecurityProfile{
+				EncryptionAtHost: utils.Bool(d.Get("encryption_at_host").(bool)),
+			},
```
ArcturusZhang, line 871: Could we revert this change since this VMSS is feature-frozen?

azurerm/internal/services/compute/virtual_machine_resource.go
```diff
@@ -621,6 +621,13 @@ func resourceArmVirtualMachine() *schema.Resource {
 			},
 
 			"tags": tags.Schema(),
+
+			"encryption_at_host": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: false,
+				Default:  false,
+			},
```
ArcturusZhang, line 630: The `azurerm_virtual_machine` resource is now in feature-frozen state, therefore could we revert the feature changes in this file?

azurerm/internal/services/compute/virtual_machine_resource.go
```diff
@@ -680,6 +687,9 @@ func resourceArmVirtualMachineCreateUpdate(d *schema.ResourceData, meta interfac
 	networkProfile := expandAzureRmVirtualMachineNetworkProfile(d)
 	vmSize := d.Get("vm_size").(string)
 	properties := compute.VirtualMachineProperties{
+		SecurityProfile: &compute.SecurityProfile{
+			EncryptionAtHost: utils.Bool(d.Get("encryption_at_host").(bool)),
+		},
```
ArcturusZhang, line 692: The `azurerm_virtual_machine` resource is now in feature-frozen state, therefore could we revert the feature changes in this file?

azurerm/internal/services/compute/virtual_machine_resource.go
ArcturusZhang: You can find more information about the feature-frozen here - https://www.terraform.io/docs/providers/azurerm/r/virtual_machine.html

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -288,6 +288,13 @@ func resourceWindowsVirtualMachine() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"encryption_at_host": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: false,
+				Default:  false,
```
ArcturusZhang, line 296: These are their default values, could we please omit them?

azurerm/internal/services/compute/windows_virtual_machine_scale_set_resource.go
```diff
@@ -233,9 +233,9 @@ func resourceArmWindowsVirtualMachineScaleSet() *schema.Resource {
 				ForceNew: true,
 				Default:  string(compute.Manual),
 				ValidateFunc: validation.StringInSlice([]string{
-					string(compute.Automatic),
+					string(compute.AutomaticByOS),
 					string(compute.Manual),
-					string(compute.Rolling),
+					string(compute.AutomaticByPlatform),
```
ArcturusZhang, line 238: Should be `compute.UpgradeModeAutomatic`, `compute.UpgradeModeManual` and `compute.UpgradeModeRolling`

JulienBalestra:
@ArcturusZhang thank you very much for the quick review, I hope I addressed all the opened comments 👍

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
```diff
@@ -251,6 +251,11 @@ func resourceArmLinuxVirtualMachineScaleSet() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"encryption_at_host": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
```
tombuildsstuff, line 258: this field isn't computed - can we move it up inline (sorted alphabetically) within the "optional" fields above?

can we also add this to the documentation? (`./website/docs/r/{os}_virtual_machine_scale_set.markdown`)

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
```diff
@@ -378,6 +383,9 @@ func resourceArmLinuxVirtualMachineScaleSetCreate(d *schema.ResourceData, meta i
 
 	virtualMachineProfile := compute.VirtualMachineScaleSetVMProfile{
 		Priority: priority,
+		SecurityProfile: &compute.SecurityProfile{
+			EncryptionAtHost: utils.Bool(d.Get("encryption_at_host").(bool)),
+		},
```
tombuildsstuff, line 388: I believe this'll need to be conditionally set - rather than always set - since this presumably isn't valid when reusing a "specialized" image (e.g. an image without 'sysprep' run on Windows)? the tests should confirm that

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -288,6 +288,11 @@ func resourceWindowsVirtualMachine() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"encryption_at_host": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
```
tombuildsstuff, line 295: (as above)

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -379,6 +384,9 @@ func resourceWindowsVirtualMachineCreate(d *schema.ResourceData, meta interface{
 			HardwareProfile: &compute.HardwareProfile{
 				VMSize: compute.VirtualMachineSizeTypes(size),
 			},
+			SecurityProfile: &compute.SecurityProfile{
+				EncryptionAtHost: utils.Bool(d.Get("encryption_at_host").(bool)),
+			},
```
tombuildsstuff, line 389: (as above)

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
JulienBalestra: In the meantime I fixed the tests and apparently we don't need, from what I understood, to add a condition. Wdyt ?

JulienBalestra:
@tombuildsstuff i addressed all the comments and just rebased 👌

ArcturusZhang:
Also @JulienBalestra the `azurerm_linux|windows_virtual_machine` also supports encryption at host now, would we also support that? (Maybe in another PR? or shipped together within this PR?)

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
```diff
@@ -126,6 +126,11 @@ func resourceArmLinuxVirtualMachineScaleSet() *schema.Resource {
 				Default:  false,
 			},
 
+			"encryption_at_host": {
```
ArcturusZhang, line 129: Also I suppose we could rename this to `encryption_at_host_enabled` given that this is a switch, what do you think?

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -153,6 +153,11 @@ func resourceWindowsVirtualMachine() *schema.Resource {
 				Default:  true,
 			},
 
+			"encryption_at_host": {
```
ArcturusZhang, line 156: Also I suppose we could rename this to `encryption_at_host_enabled` given that this is a switch, what do you think?

ArcturusZhang:
Hi @JulienBalestra just had a closer look - it turns out that we support encryption at host for linux VMSS and windows VM. Could we also add this feature to windows VMSS and linux VM?

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
JulienBalestra: I've just addressed this feedback 👌

JulienBalestra:
@ArcturusZhang I pushed a commit making it available everywhere when possible and to rename the parameter from `encryption_at_host` to `encryption_at_host_enabled`

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
ArcturusZhang: I suppose what we could do it at most only set this when encryption_at_host is true - because we could not tell whether the user is explicitly setting this to false, or omitting this attribute empty

website/docs/r/linux_virtual_machine.html.markdown
```diff
@@ -140,6 +140,8 @@ The following arguments are supported:
 
 -> **NOTE:** When an `admin_password` is specified `disable_password_authentication` must be set to `false`.
 
+* `encryption_at_host_enabled` - (Optional) Should Azure encrypt the data stored on the VM host at rest and flows encrypted to the Storage service.
```
tombuildsstuff, line 143: The Azure API uses slightly different terminology for this:

> This will enable the encryption for all the disks including Resource/Temp disk at host itself

could we make this:

```suggestion
* `encryption_at_host_enabled` - (Optional) Should all of the disks (including the temp disk) attached to this Virtual Machine be encrypted by enabling Encryption at Host?
```

WDYT?

website/docs/r/linux_virtual_machine_scale_set.html.markdown
```diff
@@ -144,6 +144,8 @@ The following arguments are supported:
 
 * `do_not_run_extensions_on_overprovisioned_machines` - (Optional) Should Virtual Machine Extensions be run on Overprovisioned Virtual Machines in the Scale Set? Defaults to `false`.
 
+* `encryption_at_host_enabled` - (Optional) Should Azure encrypt the data stored on the VM host at rest and flows encrypted to the Storage service.
```
tombuildsstuff, line 147: (as above)

website/docs/r/windows_virtual_machine.html.markdown
```diff
@@ -127,6 +127,8 @@ The following arguments are supported:
 
 * `enable_automatic_updates` - (Optional) Specifies if Automatic Updates are Enabled for the Windows Virtual Machine. Changing this forces a new resource to be created.
 
+* `encryption_at_host_enabled` - (Optional) Should Azure encrypt the data stored on the VM host at rest and flows encrypted to the Storage service.
```
tombuildsstuff, line 130: (as above)

website/docs/r/windows_virtual_machine_scale_set.html.markdown
```diff
@@ -130,6 +130,8 @@ The following arguments are supported:
 
 * `enable_automatic_updates` - (Optional) Are automatic updates enabled for this Virtual Machine? Defaults to `true`.
 
+* `encryption_at_host_enabled` - (Optional) Should Azure encrypt the data stored on the VM host at rest and flows encrypted to the Storage service.
```
tombuildsstuff, line 133: (as above)

azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -151,6 +151,11 @@ func resourceLinuxVirtualMachine() *schema.Resource {
 				Default:  true,
 			},
 
+			"encryption_at_host_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
```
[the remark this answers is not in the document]
tombuildsstuff: can we also add an acceptance test covering this? (see `./tests/linux_virtual_machine_{name}_test.go`)

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
```diff
@@ -126,6 +126,11 @@ func resourceArmLinuxVirtualMachineScaleSet() *schema.Resource {
 				Default:  false,
 			},
 
+			"encryption_at_host_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
```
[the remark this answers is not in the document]
tombuildsstuff: can we also read this back / add a test covering setting/updating this field?

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -153,6 +153,11 @@ func resourceWindowsVirtualMachine() *schema.Resource {
 				Default:  true,
 			},
 
+			"encryption_at_host_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
```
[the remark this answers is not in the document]
tombuildsstuff: could we also add a test covering setting/reading this back too?

tombuildsstuff (changes requested):
hey @caius 

Thanks for pushing those changes - I've taken a look through and left some comments inline - but if we can fix those up (and the tests pass) then this is otherwise looking good 👍 

Thanks!

WodansSon:
@caius 

Just curious if you were going to follow up on the suggested changes by @tombuildsstuff ? If not that's fine, but we would like to build on top of what you have provided thus far so we can ship it in the next release of the provider. Do you mind if we step in and finish the resource up? Thanks again for the PR 🚀

tombuildsstuff:
hey @JulienBalestra 

@ArcturusZhang has opened #8322 which includes the commits from this PR plus the changes requested above (and rebases this) - as such I hope you don't mind but since this includes your commits (and so you'll still get authorship for these) - I'm going to close this in favour of #8322 which should allow us to ship this in the next release :)

Thanks!

PR #8152, opened 2020-08-18
azurerm_storage_account: set default for `min_tls_version`

magodo:
For those storage account created using old API version (at least
2018-02-01, before provider v1.29.0), the GET call of new API version
will not return the `minimumTLSVersion` in response body. So those users
will get a diff.

This PR sets the default value for `min_tls_version` in case it is
missing in response, using the same value as the default value defined
in schema.

At least fixes the issue encountered by [this user comment](https://github.com/terraform-providers/terraform-provider-azurerm/pull/7879#issuecomment-671984119).


WodansSon (approved):
@magodo, good catch! This LGTM! 🚀

WodansSon:
Due to the the particular conditions required to reproduce the issue fixed by this PR I am skipping the normally required full test pass of the resource and running only a single test case on the PR to verify the fix does not cause a regression to the resource. I have been assured by the PR author that this has been manually tested and works as expected.

WodansSon:
image

Issue #5976, opened 2020-03-03
Azure RM 2.0 extension approach incompatible with ServiceFabricNode extension requirements of being added at VMSS creation time.


JSchenken:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
Terraform v0.12.21
+ provider.azurerm v2.0.0

### Affected Resource(s)
azurerm_windows_virtual_machine_scale_set
azurerm_virtual_machine_scale_set_extension

### Terraform Configuration Files

```

resource "azurerm_virtual_machine_scale_set_extension" "service_fabric_node" {
    virtual_machine_scale_set_id = azurerm_windows_virtual_machine_scale_set.main.id
    name                       = "${var.node_type}_ServiceFabricNode"
    publisher                  = "Microsoft.Azure.ServiceFabric"
    type                       = "ServiceFabricNode"
    type_handler_version       = "1.1"
    auto_upgrade_minor_version = false
    protected_settings         = <<PROTECTEDSETTINGS
        {
            "StorageAccountKey1": "${var.support_log_storage_account_primary_access_key}",
            "StorageAccountKey2": "${var.support_log_storage_account_secondary_access_key}"
        }
PROTECTEDSETTINGS
    settings                   = <<SETTINGS
        {
            "clusterEndpoint": "${var.cluster_endpoint}",
            "nodeTypeRef": "${var.node_type}",
            "dataPath": ${jsonencode(var.fabric_data_root)},
            "durabilityLevel": "${var.durability_level}",
            "certificate": {
                "commonNames": ["${var.sfserver_cert_common_name}"],
                "x509StoreName": "${var.certificate_store_value}"
            },
            "nicPrefixOverride": ${jsonencode(var.subnetprefix)}
        }
SETTINGS
  }

```

### Expected Behavior

The extension should have been included in the VMSS definition at creation time, and not attempted to be added after the VMSS was already created. 

### Actual Behavior
The VMSS is created and then there is an attempt to add the extension to the VMSS, this is not workable for ServiceFabricNode extension, as seen in error message: 

[...]

PR #8156, opened 2020-08-18
Upgrade Compute to 2020-06-01

rhys96:
I've updated the compute version in all the compute files and refactored the syntax where needed. This update will enable me to submit another Pull Request for #5976.


tombuildsstuff:
hey @rhys96 

Thanks for this PR :)

Taking a look through whilst this is great this is separated out - since there's merge conflicts here (and this SDK upgrade is included as a part of #8134) - whilst I'd like to thank you for this PR I'm going to close this in favour of #8134 for the moment.

With regard to #5976 - this shouldn't require upgrading the API version here, however it's worth noting that @jackofallops is currently working through this (since this looks fairly simple, but unfortunately is a bit of a minefield) - so whilst we'd appreciate the PR here to avoid duplication here would you mind instead detailing your use-case in #5976 so that we can ensure this use-case is covered?

Thanks!

PR #8182, opened 2020-08-20
Remove computed attribute for public_ip_address_id and public_ip_prefix_id of load balancer

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/4648

--- PASS: TestAccAzureRMLoadBalancer_standard (95.51s)
--- PASS: TestAccAzureRMLoadBalancer_frontEndConfigPublicIPPrefix (108.71s)
--- PASS: TestAccAzureRMLoadBalancer_basic (156.25s)
--- PASS: TestAccAzureRMLoadBalancer_emptyPrivateIP (156.70s)
--- PASS: TestAccAzureRMLoadBalancer_requiresImport (164.48s)
--- PASS: TestAccAzureRMLoadBalancer_privateIP (172.30s)
--- PASS: TestAccAzureRMLoadBalancer_tags (179.52s)
--- PASS: TestAccAzureRMLoadBalancer_frontEndConfig (235.25s)
--- PASS: TestAccAzureRMLoadBalancer_removeFirstPublicIPAddress (142.33s)
--- PASS: TestAccAzureRMLoadBalancer_removeFirstPublicIPPrefix (145.89s)

### DIFF
--- a/azurerm/internal/services/network/lb_resource.go
+++ b/azurerm/internal/services/network/lb_resource.go
@@ -102,14 +102,12 @@ func resourceArmLoadBalancer() *schema.Resource {
 						"public_ip_address_id": {
 							Type:         schema.TypeString,
 							Optional:     true,
-							Computed:     true,
 							ValidateFunc: azure.ValidateResourceIDOrEmpty,
 						},
 
 						"public_ip_prefix_id": {
 							Type:         schema.TypeString,
 							Optional:     true,
-							Computed:     true,
 							ValidateFunc: azure.ValidateResourceIDOrEmpty,
 						},
 
--- a/azurerm/internal/services/network/tests/loadbalancer_resource_test.go
+++ b/azurerm/internal/services/network/tests/loadbalancer_resource_test.go
@@ -211,6 +211,60 @@ func TestAccAzureRMLoadBalancer_privateIP(t *testing.T) {
 	})
 }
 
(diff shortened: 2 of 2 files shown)


tombuildsstuff:
hi @neil-yechenwei 

Since the validation functions allow these to be either a Resource ID (for a Public IP Address or a Public IP Prefix) or an Empty String - it's possible to achieve this by assigning an empty string as the value here.

Unfortunately this change would be a breaking change - and since this is already possible, rather than introducing a breaking change here I'm going to close this issue for the moment.

Thanks!

neil-yechenwei:
@tombuildsstuff , after tested, seems there is no diff so that public_ip_address_id/public_ip_prefix_id wouldn't be updated to empty while setting them as empty string value "". So I assume we have to remove the "Computed:true", right?

```
resource "azurerm_lb" "example" {
  name                = "lb-test01"
  sku                 = "Standard"

  frontend_ip_configuration {
    name                 = "PublicIPAddress01"
    public_ip_address_id = ""
  }
  
  frontend_ip_configuration {
    name                 = "PublicIPAddress02"
    public_ip_address_id = azurerm_public_ip.example2.id
  }
  
  frontend_ip_configuration {
    name                = "PublicIPAddress03"
    public_ip_prefix_id = azurerm_public_ip_prefix.example.id
  }
}
```

image

PR #8185, opened 2020-08-20
Update cosmosdb_gremlin_graph.html.markdown

georghildebrand:
With a note on required azurerm_cosmosdb_account.capabilites setup to avoid "not supported error".

### DIFF
--- a/website/docs/r/cosmosdb_gremlin_graph.html.markdown
+++ b/website/docs/r/cosmosdb_gremlin_graph.html.markdown
@@ -50,7 +50,7 @@ resource "azurerm_cosmosdb_gremlin_graph" "example" {
 }
 ```
 
--> **NOTE:** You need to declare `name = "EnableGremlin"` in `azurerm_cosmosdb_account.capabilites`` when using gremlin API. Otherwise the graph creation might end up with error `Requests for API gremlin are not supported for this account`. 
+-> **NOTE:** The CosmosDB Account needs to have the `EnableGremlin` capability enabled to use this resource - which can be done by adding this to the `capabilities` list within the `azurerm_cosmosdb_account` resource.
 
 ## Argument Reference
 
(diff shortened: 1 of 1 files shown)


website/docs/r/cosmosdb_gremlin_graph.html.markdown
```diff
@@ -50,6 +50,8 @@ resource "azurerm_cosmosdb_gremlin_graph" "example" {
 }
 ```
 
+-> **NOTE:** You need to declare `name = "EnableGremlin"` in `azurerm_cosmosdb_account.capabilites`` when using gremlin API. Otherwise the graph creation might end up with error `Requests for API gremlin are not supported for this account`.
```
tombuildsstuff, line 53: could we re-phrase this slightly:

```suggestion
-> **NOTE:** The CosmosDB Account needs to have the `EnableGremlin` capability enabled to use this resource - which can be done by adding this to the `capabilities` list within the `azurerm_cosmosdb_account` resource.
```

tombuildsstuff (commented):
hey @georghildebrand 

Thanks for this PR :)

I've taken a look through and this mostly LGTM - but could we re-word this slightly to explain what needs to be done, rather than highlighting the error message (which unfortunately can change)

Thanks!

georghildebrand:
@tombuildsstuff , sure thing. its a minor note. Accepted your suggestion. 

If you like we could also add an full end to end example of gremlin. Pushed something [here](https://github.com/georghildebrand/cosmosdb_terraform_test/blob/master/main.tf):

tombuildsstuff:
@georghildebrand I think that repository is private (it 404's for me) - but we're definitely interested in another example if you don't mind sending another PR :)

tombuildsstuff (approved):
LGTM - thanks @georghildebrand  👍

PR #8186, opened 2020-08-20
docs: removing ambiguous usages of 'Resource ID'

tombuildsstuff:
Resource ID is ambiguous in these contexts between the Terraform Resource ID and the Azure Resource ID - which are not necessarily the same thing. As such the term "Resource ID" should generally be phrased as "The ID of the {thing}" - rather than "the Resource ID of the {thing}" to ensure the Terraform Resource ID is used here.


jackofallops (approved):
LGTM 👍

Issue #8093, opened 2020-08-12
Support for Azure autoheal


dzmitry-lahoda:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

AppService Web has Autoheal (not docker nor k8s).

### New or Affected Resource(s)

https://docs.microsoft.com/en-us/azure/app-service/overview-diagnostics

https://azure.github.io/AppService/2018/09/10/Announcing-the-New-Auto-Healing-Experience-in-App-Service-Diagnostics.html

https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.management.websites.models.autohealrules?view=azure-dotnet

### Potential Terraform Configuration

auto_heal_rules  and auto_heal_enabled   :
```hcl
resource "azurerm_app_service" "default" {
  for_each            = setintersection(keys(var.service_plan_ids), keys(var.regions))
  name                = "..."
  location            = var.regions[each.key].location
  resource_group_name = var.resource_group.name
  tags                = var.tags
  app_service_plan_id = var.service_plan_ids[each.key]

  client_affinity_enabled = false
  client_cert_enabled     = false
  https_only              = true

  identity {
    type = "SystemAssigned"
  }

  site_config {
    default_documents = ["info/detail"]
    always_on         = true
    http2_enabled     = true
    
    // check that dummy call for now, next iterations will implement dummy queries to relevant data soources (SB, SQL, BLOB, etc) 
    health_check_path   = "/info/detail"
    auto_heal_enabled   = true
    auto_heal_rules {
        triggers {
            // AutoHealTriggers does not have documentation, so resonably assume triggers are glued with OR condition
            slow_requests {
                time_taken = "00:01:00"
                count = 600
                time_interval = "00:10:00"
            }
            status_codes = [
                            {
                                // for not 500, but in future loop over range, like 503 as needed
                                status = 500
                                count  = 100
                                timeInterval = "00:10:00"
                            }
                        ]
        }
        actions {
            action_type = "Recycle"
            min_process_execution_time = "00:15:00"
        }
    }
  }
}
```

Except from ARM:

```json
      {
[...]

PR #8203, opened 2020-08-22
reduced entry barrier for vs csharp people #8093

dzmitry-lahoda:


tombuildsstuff (approved):
LGTM - thanks for this @dzmitry-lahoda

PR #8204, opened 2020-08-22
adding part of auto heal #8093

dzmitry-lahoda:


azurerm/internal/services/web/app_service.go
```diff
@@ -895,6 +907,121 @@ func schemaAppServiceDataSourceIpRestriction() *schema.Schema {
...
+									Schema: map[string]*schema.Schema{
+										"time_taken": {
+											Type:         schema.TypeString,
+											Required:     true,
+											ValidateFunc: validation.StringIsNotWhiteSpace,
+										},
+										"count": {
+											Type:     schema.TypeInt,
+											Required: true,
+										},
```
jackofallops, line 952: Again here, `intAtLeast(1)` should be used to validate.

azurerm/internal/services/web/app_service.go
```diff
@@ -895,6 +907,121 @@ func schemaAppServiceDataSourceIpRestriction() *schema.Schema {
...
+											Required:     true,
+											ValidateFunc: validation.StringIsNotWhiteSpace,
+										},
+									},
+								},
+							},
+							"slow_requests": {
+								Type:     schema.TypeList,
+								Computed: true,
+								Elem: &schema.Resource{
```
jackofallops, line 942: We'll need `Optional` and `MaxItems: 1` here.

azurerm/internal/services/web/app_service.go
```diff
@@ -895,6 +907,121 @@ func schemaAppServiceDataSourceIpRestriction() *schema.Schema {
...
+											Type:         schema.TypeString,
+											Required:     true,
+											ValidateFunc: validation.StringIsNotWhiteSpace,
+										},
+									},
+								},
+							},
+							"status_codes": {
+								Type:     schema.TypeSet,
+								Computed: true,
```
jackofallops, line 963: Again, `Optional` needed here.

azurerm/internal/services/web/app_service.go
```diff
@@ -1722,6 +1868,40 @@ func flattenAppServiceSiteConfig(input *web.SiteConfig) []interface{} {
...
+		vals["time_interval"] = slowRequests.TimeInterval
+		vals["time_taken"] = slowRequests.TimeTaken
+		triggerVals["slow_requests"] = vals
+	}
+
+	actions := input.Actions
+	actionVals := make(map[string]interface{})
+	actionVals["action_type"] = actions.ActionType
+	actionVals["min_process_execution_time"] = actions.MinProcessExecutionTime
+	rule["actions"] = actionVals
```
jackofallops, line 1900: This will need to be reworked a little to check values are not nil before assignment (e.g. input.Triggers is a pointer, that may be nil). We also need to ensure that every expected property has a value set.  You should be able to use `flattenAppServiceLogs` as an example of flattening this type of structure.
We also appear to be missing the status codes section here?

azurerm/internal/services/web/app_service.go
```diff
@@ -1812,6 +1992,55 @@ func flattenAppServiceStorageAccounts(input map[string]*web.AzureStorageInfoValu
 	return results
 }
 
+func expandAppServiceAutoHealRules(input map[string]interface{}) (*web.AutoHealRules, error) {
+	triggerVals := input["triggers"].(map[string]interface{})
+
+	autoHeal := web.AutoHealRules{}
+
+	triggers := web.AutoHealTriggers{}
+	slowRequests := triggerVals["slow_requets"].(map[string]interface{})
```
jackofallops, line 2001: This will crash if this is nil and not map[string]interface{}.  Since this is actually a list that needs to be unpacked, this would be the case. If you use `expandAppServiceLogs` in this file as an example, you can see one way to handle this.

jackofallops (changes requested):
Hi @dzmitry-lahoda 
Thanks for this PR, it's off to a good start.  I've left some comments in line below on the schema, and the flatten and expand elements.
We'd need acceptance tests adding for these new properties on all resources that use the new code, this means that we'll need create and update tests for both `azurerm_app_service` and `azurerm_app_service_slot`, since both leverage that schema. Data Source tests should also be added to ensure they appropriately reflect the new data.
Finally we'll also need all of the new schema items documented for the affected resources and data sources.
Thanks again!

dzmitry-lahoda:
i see, will try these after fix of https://github.com/terraform-providers/terraform-provider-azurerm/issues/8206

azurerm/internal/services/web/app_service.go
```diff
@@ -769,6 +774,13 @@ func schemaAppServiceDataSourceSiteConfig() *schema.Schema {
 					Computed: true,
 				},
 
+				"auto_heal_enabled": {
+					Type:     schema.TypeBool,
+					Optional: true,
+				},
+
+				"auto_heal_rules": schemaAutoHealRules(),
```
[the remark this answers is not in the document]
dzmitry-lahoda: not sure i got this?

dzmitry-lahoda:
some comments on schema and validation structs with examples could help to be more correct with proper schema

azurerm/internal/services/web/app_service.go
```diff
@@ -769,6 +774,13 @@ func schemaAppServiceDataSourceSiteConfig() *schema.Schema {
 					Computed: true,
 				},
 
+				"auto_heal_enabled": {
+					Type:     schema.TypeBool,
+					Optional: true,
+				},
```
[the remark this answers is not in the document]
dzmitry-lahoda: @jackofallops it there any code level comments about behaviour of `Data Source schema items that are not required to get the resource in question should be computed only`, ideally in code?

katbyte:
Hey @dzmitry-lahoda - i'm going to close this as its been sitting with unaddressed feedback for a number of months now. Whenever you do come back to this feel free to reopen this or a new PR so we can continuing working to get this merged 🙂

PR #8205, opened 2020-08-22
azurerm_cdn_endpoint order rules should start from 0

SvanBoxel:
As per Azure's CLI documentation: https://docs.microsoft.com/en-us/cli/azure/cdn/endpoint/rule?view=azure-cli-latest#az-cdn-endpoint-rule-add

> `--order`
> The order of the rule. The order number must start from 0 and consecutive. Rule with a higher order will be applied later.


tombuildsstuff (changes requested):
hey @SvanBoxel 

Thanks for this PR :)

Taking a look through this mostly LGTM - if we can add a new test rather than modifying the existing one (since it appears that starting at 1 is still valid) then this should otherwise be good to merge 👍 

Thanks!

manicminer:
Hi @SvanBoxel, thanks for this PR! I believe the order attribute of `delivery_rule` is indexed from 1 because the zero-th attribute is represented by the `global_delivery_rule` block. In other words, `global_delivery_rule` is the first (zero) rule and cannot have any conditions, and you should use this attribute to set that rule.

Accordingly I'm going to close this PR, as the schema as defined currently should enable you to configure a complete delivery policy. Please feel free to comment if I have misunderstood, and we'll be happy to revisit this.

PR #8216, opened 2020-08-24
Update default value for allow_blob_public_access of azurerm_storage_account

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/8187


waynejnicklin:
This still seems to be an issue. I have allow_blob_public_access set to false in my infrastructure pipelines, and it deploys with "no change" and leaves it set to true when I check the storage account in the portal.

neil-yechenwei:
Maybe there is some latency in Azure Portal. Seems all related test cases passed from my side after updated code.

waynejnicklin:
I don't think it's a latency issue, I deployed weeks ago and it still hasn't changed! But I deployed the change into an existing storage account which had public blob access enabled... and it left it enabled. Maybe when deploying a new clean storage account, it works, but certainly trying to change an existing one doesn't.

neil-yechenwei:
@waynejnicklin , maybe I missed something. Could you have a try with this PR since I just updated this PR? If you still encountered same issue, please let me know. Thanks.

waynejnicklin:
Yes sure thank you :-)

kevinharing:
What is taking so long to complete this?

tombuildsstuff:
Closing in favour of #12689

Issue #8191, opened 2020-08-20
azurerm_dns_a_record failing with weird "from no visitor picked" string in target_resource_id?


dossy:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

```
$ terraform -v
Terraform v0.12.29
+ provider.azurerm v2.24.0
+ provider.null v2.1.2
```

### Affected Resource(s)

* `azurerm_dns_a_record`
* `azurerm_cdn_profile`
* `azurerm_cdn_endpoint`

### Terraform Configuration Files

```hcl
resource "azurerm_cdn_profile" "cdn" {
  resource_group_name = "RG"
  location            = "EastUS2"
  name                = "cdn"
  sku                 = "Standard_Microsoft"
}

resource "azurerm_cdn_endpoint" "endpoint" {
  resource_group_name = "RG"
  location            = "EastUS2"
  profile_name        = azurerm_cdn_profile.cdn.name
  name                = "cdn-endpoint"

  querystring_caching_behaviour   = "UseQueryString"
  origin_host_header              = "www.domain.test"

  origin {
    name      = "www"
    host_name = "www.domain.test"
  }

  global_delivery_rule {
    modify_request_header_action {
      action  = "Delete"
      name    = "Cookie"
    }

    modify_response_header_action {
      action  = "Delete"
      name    = "Set-Cookie"
    }
  }
}

resource "azurerm_dns_a_record" "apex" {
  resource_group_name = "RG"
  zone_name           = "domain.test"
  name                = "@"

  ttl                 = 300
  target_resource_id  = azurerm_cdn_profile.cdn.id
}
```

### Debug Output

<details>
<summary>(I cannot provide the entire log, but I've redacted names and IDs from the relevant portion of the log.  Click to expand.)</summary>

```
2020/08/20 16:34:46 [DEBUG] azurerm_dns_a_record.apex[0]: applying the planned Create change
2020/08/20 16:34:46 [TRACE] GRPCProvider: ApplyResourceChange
[...]

PR #8218, opened 2020-08-24
`azurerm_dns_x_record`: ignore case on `target_resource_id`

magodo:
In some case (at least for the cdn endpoint), the ID of target resource
read from API (which uses the API model of the target resource) doesn't match
the `target_resource_id` if it is imported via Terraform (which actually
uses the API model of the `azurerm_dns_x_record`) in casing.

Fixes: #8191

### DIFF
--- a/azurerm/internal/services/dns/dns_a_record_resource.go
+++ b/azurerm/internal/services/dns/dns_a_record_resource.go
@@ -2,6 +2,7 @@ package dns
 
 import (
 	"fmt"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/suppress"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
@magodo is there an Swagger bug to track this API bug?

ARM API's are supposed to be following Postel's law, where they are liberal in what they accept (eg uri's could be case insensitive) but conservative in what they send (eg case sensitive) - so this is a bug in the API which needs to be fixed

Whilst applying a DiffSuppress does work around this - it's something that we should avoid unless absolutely necessary, so it'd be worth raising the API bug here

dossy:
@tombuildsstuff, @magodo - I'm not familiar enough with TF's code (yet), but I feel like the issue is actually with `azurerm_cdn_endpoint`'s `id` output not properly casing `resourcegroup` as `resourceGroup`.

magodo:
Close in favor of #8237, which modify the resource id of cdn endpoint.

PR #8228, opened 2020-08-24
New resource: `azurerm_api_management_custom_domain`

### LINKED ISSUE #3058: API Management (azurerm_api_management) with custom domain.
I'm trying to script API Management with custom domain where certificate is obtained from Key Vault.

I seem to have a catch-22 where I am defining azurerm_api_management with the Identiity property set to 'SystemAssigned' and with host_configuration > proxy > key_vault_id set to 'azurerm_key_vault_secret.<local>.id'... The problem here is that at this stage the system assigned managed identity has not been given permission to Key Vault (which I have defined in the next step using azurerm_key_vault_access_policy.
rs.html#provider-versions). --->

### Affected Resource(s)

* `azurerm_api_management`
* `azurerm_key_vault_access_policy`
* `azurerm_key_vault_secret`

```hcl
data "azurerm_key_vault_secret" "core_apim" {
  name      = "SslCert"
  vault_uri = "${azurerm_key_vault.core.vault_uri}"
}

resource "azurerm_api_management" "core" {
  name                = "${var.res_prefix}-${var.environment}-apim"
  location            = "${var.primary_location}"
  resource_group_name = "${azurerm_resource_group.core.name}"
  publisher_name      = "HELLO WORLD"
  publisher_email     = "me@mydomain.com"

  sku {
    name     = "${var.apim_tier_name}"
    capacity = "${var.apim_unit_count}"
  }

  identity
  {
    type = "SystemAssigned"
  }

  hostname_configuration {
    proxy {
      host_name = "${replace(replace(replace(var.environment, "/^d.*$/", "dev-api"), "/^t.*$/", "test-api"), "/^p.*$/", "api")}.mydomain.net"
      key_vault_id = "${data.azurerm_key_vault_secret.core_apim.id}"  # TODO: Handle count
    }
  }
}

resource "azurerm_key_vault_access_policy" "core_apim" {
  key_vault_id = "${azurerm_key_vault.core.id}"

  tenant_id = "${azurerm_api_management.core.identity.0.tenant_id}"
  object_id = "${azurerm_api_management.core.identity.0.principal_id}"

  key_permissions = [
    "backup", "create", "decrypt", "delete", "encrypt", "get", "import", "list", "purge", "recover", "restore", "sign", "unwrapKey", "update", "verify", "wrapKey"
  ]

  secret_permissions = [
    "backup", "delete", "get", "list", "purge", "recover", "restore", "set"
  ]

  certificate_permissions = [
    "backup", "create", "delete", "deleteissuers", "get", "getissuers", "import", "list", "listissuers", "managecontacts", "manageissuers", "purge", "recover", "restore", "setissuers", "update"
  ]
}
```

### Expected Behavior

API Management setup with custom domain (using cert from Key Vault).

### Actual Behavior

Error applying plan.... 

azurerm_api_management.core: Error creating/updating API Management Service "rp-dev-apim" (Resource Group "rd-dev-core-rg"): apimanagement.ServiceClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="BadRequest" Message="Request to resource 'https://{subdomain}.vault.azure.net/secrets/RedPlatSslCert/5b_________e4_________fe_____2b?api-version=7.0' failed with StatusCode: Forbidden for RequestId: . Exception message: Operation returned an invalid status code 'Forbidden'"

sirlatrom:
Closes #3058.

The approach used is to modify the API Management service (corresponding to an `azurerm_api_management` resource) by updating an existing instance. A note is added to the documentation in both that and the new resource telling that you should not configure the custom domains in _both_ resources, as that would likely create conflicts.

Semantically, if you delete an `azurerm_api_management_custom_domain`, the `hostname_configurations` field will be set to be empty (null) when the API Management service instance is updated. There should not be more than one `azurerm_api_management_custom_domain` resource for an `azurerm_api_management`, but I have not found a way to enforce this using the Terraform Provider SDK.

---

Includes:
  - [x] Implementation
  - [x] Acceptance tests
  - [x] Website docs

### DIFF
--- a/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
@@ -108,7 +108,7 @@ func apiManagementCustomDomainCreateUpdate(d *schema.ResourceData, meta interfac
 	}
 
 	if d.IsNewResource() {
-		if existing.ServiceProperties != nil && existing.ServiceProperties.HostnameConfigurations != nil {
+		if existing.ServiceProperties != nil && existing.ServiceProperties.HostnameConfigurations != nil && len(*existing.ServiceProperties.HostnameConfigurations) > 0 {
 			return tf.ImportAsExistsError(apiManagementCustomDomainResourceName, *existing.ID)
 		}
 	}
--- a/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
@@ -147,6 +147,12 @@ func apiManagementCustomDomainCreateUpdate(d *schema.ResourceData, meta interfac
 		return fmt.Errorf("cannot read ID for Custom Domain (API Management %q / Resource Group %q)", serviceName, resourceGroup)
 	}
 
+	// Wait for the ProvisioningState to become "Succeeded" before attempting to update
+	log.Printf("[DEBUG] Waiting for API Management Service %q (Resource Group: %q) to become ready", serviceName, resourceGroup)
+	if _, err = stateConf.WaitForState(); err != nil {
+		return fmt.Errorf("waiting for API Management Service %q (Resource Group: %q) to become ready: %+v", serviceName, resourceGroup, err)
+	}
+
 	customDomainsID := fmt.Sprintf("%s/customDomains/default", *read.ID)
 	d.SetId(customDomainsID)
 
@@ -240,15 +246,6 @@ func apiManagementCustomDomainDelete(d *schema.ResourceData, meta interface{}) e
 
 	// Wait for the ProvisioningState to become "Succeeded" before attempting to update
 	log.Printf("[DEBUG] Waiting for API Management Service %q (Resource Group: %q) to become ready", serviceName, resourceGroup)
-	stateConf = &resource.StateChangeConf{
-		Pending:                   []string{"Updating", "Unknown"},
-		Target:                    []string{"Succeeded", "Ready"},
-		Refresh:                   apiManagementRefreshFunc(ctx, client, serviceName, resourceGroup),
-		MinTimeout:                1 * time.Minute,
-		Timeout:                   d.Timeout(schema.TimeoutDelete),
-		ContinuousTargetOccurence: 5,
-	}
-
 	if _, err = stateConf.WaitForState(); err != nil {
 		return fmt.Errorf("waiting for API Management Service %q (Resource Group: %q) to become ready: %+v", serviceName, resourceGroup, err)
 	}
--- a/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
@@ -2,13 +2,14 @@ package apimanagement
 
 import (
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"log"
 	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/apimanagement/mgmt/2019-12-01/apimanagement"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
@@ -123,7 +124,7 @@ func apiManagementCustomDomainCreateUpdate(d *schema.ResourceData, meta interfac
 		Target:                    []string{"Succeeded", "Ready"},
 		Refresh:                   apiManagementRefreshFunc(ctx, client, serviceName, resourceGroup),
 		MinTimeout:                1 * time.Minute,
-		ContinuousTargetOccurence: 5,
+		ContinuousTargetOccurence: 6,
 	}
 	if d.IsNewResource() {
 		stateConf.Timeout = d.Timeout(schema.TimeoutCreate)
@@ -229,7 +230,7 @@ func apiManagementCustomDomainDelete(d *schema.ResourceData, meta interface{}) e
 		Refresh:                   apiManagementRefreshFunc(ctx, client, serviceName, resourceGroup),
 		MinTimeout:                1 * time.Minute,
 		Timeout:                   d.Timeout(schema.TimeoutDelete),
-		ContinuousTargetOccurence: 5,
+		ContinuousTargetOccurence: 6,
 	}
 
 	if _, err = stateConf.WaitForState(); err != nil {
--- a/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
@@ -95,12 +95,12 @@ func apiManagementCustomDomainCreateUpdate(d *schema.ResourceData, meta interfac
 	log.Printf("[INFO] preparing arguments for API Management Custom domain creation.")
 
 	apiManagementID := d.Get("api_management_id").(string)
-	id, err := azure.ParseAzureResourceID(apiManagementID)
+	id, err := parse.ApiManagementID(apiManagementID)
 	if err != nil {
 		return err
 	}
 	resourceGroup := id.ResourceGroup
-	serviceName := id.Path["service"]
+	serviceName := id.ServiceName
 
 	existing, err := client.Get(ctx, resourceGroup, serviceName)
 	if err != nil {
--- a/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
@@ -330,12 +330,7 @@ func flattenApiManagementHostnameConfiguration(input *[]apimanagement.HostnameCo
 			output["key_vault_id"] = *config.KeyVaultID
 		}
 
-		snakeCaseConfigType := azure.ToSnakeCase(string(config.Type))
-		if valsRaw, ok := d.GetOk(snakeCaseConfigType); ok {
-			vals := valsRaw.([]interface{})
-			azure.CopyCertificateAndPassword(vals, *config.HostName, output)
-		}
-
+		var configType string
 		switch strings.ToLower(string(config.Type)) {
 		case strings.ToLower(string(apimanagement.HostnameTypeProxy)):
 			// only set SSL binding for proxy types
@@ -343,18 +338,30 @@ func flattenApiManagementHostnameConfiguration(input *[]apimanagement.HostnameCo
 				output["default_ssl_binding"] = *config.DefaultSslBinding
 			}
 			proxyResults = append(proxyResults, output)
+			configType = "proxy"
 
 		case strings.ToLower(string(apimanagement.HostnameTypeManagement)):
 			managementResults = append(managementResults, output)
+			configType = "management"
 
 		case strings.ToLower(string(apimanagement.HostnameTypePortal)):
 			portalResults = append(portalResults, output)
+			configType = "portal"
 
 		case strings.ToLower(string(apimanagement.HostnameTypeDeveloperPortal)):
 			developerPortalResults = append(developerPortalResults, output)
+			configType = "developer_portal"
 
 		case strings.ToLower(string(apimanagement.HostnameTypeScm)):
 			scmResults = append(scmResults, output)
+			configType = "scm"
+		}
+
+		if configType != "" {
+			if valsRaw, ok := d.GetOk(configType); ok {
+				vals := valsRaw.([]interface{})
+				azure.CopyCertificateAndPassword(vals, *config.HostName, output)
+			}
 		}
 	}
 
--- a/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
@@ -2,6 +2,7 @@ package apimanagement
 
 import (
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"log"
 	"strings"
 	"time"
@@ -115,6 +116,25 @@ func apiManagementCustomDomainCreateUpdate(d *schema.ResourceData, meta interfac
 
 	existing.ServiceProperties.HostnameConfigurations = expandApiManagementCustomDomains(d)
 
+	// Wait for the ProvisioningState to become "Succeeded" before attempting to update
+	log.Printf("[DEBUG] Waiting for API Management Service %q (Resource Group: %q) to become ready", serviceName, resourceGroup)
+	stateConf := &resource.StateChangeConf{
+		Pending:                   []string{"Updating", "Unknown"},
+		Target:                    []string{"Succeeded", "Ready"},
+		Refresh:                   apiManagementRefreshFunc(ctx, client, serviceName, resourceGroup),
+		MinTimeout:                1 * time.Minute,
+		ContinuousTargetOccurence: 5,
+	}
+	if d.IsNewResource() {
+		stateConf.Timeout = d.Timeout(schema.TimeoutCreate)
+	} else {
+		stateConf.Timeout = d.Timeout(schema.TimeoutUpdate)
+	}
+
+	if _, err = stateConf.WaitForState(); err != nil {
+		return fmt.Errorf("waiting for API Management Service %q (Resource Group: %q) to become ready: %+v", serviceName, resourceGroup, err)
+	}
+
 	if _, err := client.CreateOrUpdate(ctx, resourceGroup, serviceName, existing); err != nil {
 		return fmt.Errorf("creating/updating Custom Domain (API Management %q / Resource Group %q): %+v", serviceName, resourceGroup, err)
 	}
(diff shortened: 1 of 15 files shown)


gd-asharov:
Hi,
Do you have a plan when this feature will be completed?

sirlatrom:
> Hi,
> Do you have a plan when this feature will be completed?

You can vote for the issue to be prioritized by giving a :+1: reaction to the top-level issue description that this PR should solve: https://github.com/terraform-providers/terraform-provider-azurerm/issues/3058#issue-421658657.

Whenever a maintainer has time to look through the proposed changes and additions, this PR can move forward.

sirlatrom:
Rebased just in case.

sirlatrom:
@manicminer Is there any chance of this getting reviewed before 2.30.0?

website/docs/r/api_management_custom_domain.html.markdown
```diff
@@ -0,0 +1,152 @@
...
+
+* `host_name` - (Required) The Hostname to use for the SCM domain.
+
+* `certificate` - (Optional) The Base64 Encoded Certificate. (Mutually exlusive with `key_vault_id`.)
+
+* `certificate_password` - (Optional) The password associated with the certificate provided above.
+
+* `key_vault_id` - (Optional) The ID of the Key Vault Secret containing the SSL Certificate, which must be should be of the type application/x-pkcs12.
+
+* `negotiate_client_certificate` - (Optional) Should Client Certificate Negotiation be enabled for this Hostname? Defaults to false.
```
manicminer, line 129: Since these are all the same basic set of attributes, I think we should compress these similarly to how they are in the [api_management resource docs](https://www.terraform.io/docs/providers/azurerm/r/api_management.html#host_name). The description for `host_name` could be something like "The hostname to use for the corresponding API endpoint".

manicminer (changes requested):
Hi @sirlatrom, thanks for this PR! I have a few minor edits and suggestions inline, and I think some more test coverage is needed.

The `testAccAzureRMApiManagementCustomDomain_basic` test looks good, but there is no test that sets any of the custom domains. I'd suggest adding a `testAccAzureRMApiManagementCustomDomain_complete` test having at least two types configured, and then a `testAccAzureRMApiManagementCustomDomain_update` test which first creates a resource with no domains (basic), then adds some domains (complete) and lastly removes then (back to basic), with import steps in between.

If we can expand the tests and fix up the other edits, this looks great to merge!

azurerm/internal/services/apimanagement/tests/api_management_custom_domain_resource_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+		serviceName := rs.Primary.Attributes["api_management_name"]
+		resourceGroup := rs.Primary.Attributes["resource_group_name"]
+
+		resp, err := conn.Get(ctx, resourceGroup, serviceName)
+		if err != nil {
+			if utils.ResponseWasNotFound(resp.Response) {
+				// TODO: Is this an error case? If Custom Domains is destroyed, it should not destroy the API Management Service, so what are we testing for here?
+				// If the entire configuration is torn down, including the API Management Service, then it, too, will not be found, and returning nil is correct.
+				return nil
```
manicminer, line 69: Since the CheckDEstroy func is run after a destroy, the APIM service is probably gone too, so it's fine to return nil here.

manicminer, line 69: Since the CheckDestroy func is run after a destroy, the APIM service is probably gone too, so it's fine to return nil here.

```suggestion
				return nil
```

sirlatrom:
> 
> 
> Hi @sirlatrom, thanks for this PR! I have a few minor edits and suggestions inline, and I think some more test coverage is needed.
> 
> The `testAccAzureRMApiManagementCustomDomain_basic` test looks good, but there is no test that sets any of the custom domains. I'd suggest adding a `testAccAzureRMApiManagementCustomDomain_complete` test having at least two types configured, and then a `testAccAzureRMApiManagementCustomDomain_update` test which first creates a resource with no domains (basic), then adds some domains (complete) and lastly removes then (back to basic), with import steps in between.
> 
> If we can expand the tests and fix up the other edits, this looks great to merge!

@manicminer I've addressed the code comments but have yet to write a `testAccAzureRMApiManagementCustomDomain_update` test. I'm not actually sure if it's actually required to have at least one domain configured - it might be. I've updated the basic test to set two domains of different types. I'm not too familiar with the update type of test.

sirlatrom:
Wow, turns out the success I thought I'd had with my first attempt was a false positive, and it really did nothing. I've added several commits and have now verified that it actually does what it is supposed to now.

manicminer:
@sirlatrom I haven't tested your recent edits, but it makes sense at a glance.

When testing it seemed like having at least one domain wasn't required, so I figure it makes sense to be able to have a resource with no domains configured - thus enabling removal of existing configured domains?

Given that, it's useful to have a test that configures no domains ("basic"), another that configures some ("complete"), and an "update" test that applies a series of configurations ("basic", then "complete", then "basic") to test that moving from no domains, to some domains, and back to no domains, works as expected.

sirlatrom:
> 
> 
> @sirlatrom I haven't tested your recent edits, but it makes sense at a glance.
> 
> When testing it seemed like having at least one domain wasn't required, so I figure it makes sense to be able to have a resource with no domains configured - thus enabling removal of existing configured domains?
> 
> Given that, it's useful to have a test that configures no domains ("basic"), another that configures some ("complete"), and an "update" test that applies a series of configurations ("basic", then "complete", then "basic") to test that moving from no domains, to some domains, and back to no domains, works as expected.

@manicminer Thanks for your feedback and patience with this PR.

The intention (and current implementation in this PR) with regard to being able to remove existing configured domains is that if you've once created an `azurerm_api_management_custom_domain` resource, when you destroy it the custom domains will be removed again. However, I agree it makes sense to validate adding or removing different types of domains in an update test. I'll try and get that done within the next couple of days.

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
```diff
@@ -0,0 +1,325 @@
...
+	existing, err := client.Get(ctx, resourceGroup, name)
+	if err != nil {
+		return fmt.Errorf("finding API Management (API Management %q / Resource Group %q): %s", name, resourceGroup, err)
+	}
+
+	if d.IsNewResource() {
+		if existing.ServiceProperties.HostnameConfigurations != nil {
+			return tf.ImportAsExistsError(apiManagementCustomDomainResourceName, *existing.ID)
+		}
+	}
```
jackofallops, line 105: I think we'll have issue here. If `ServiceProperties` is `nil` we'll crash, and I believe that in a normal response `HostnameConfiguration` isn't nil, so this will always trigger.

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
```diff
@@ -0,0 +1,325 @@
...
+
+	read, err := client.Get(ctx, resourceGroup, name)
+	if err != nil {
+		return fmt.Errorf("retrieving Custom Domain (API Management %q / Resource Group %q): %+v", name, resourceGroup, err)
+	}
+	if read.ID == nil {
+		return fmt.Errorf("cannot read ID for Custom Domain (API Management %q / Resource Group %q)", name, resourceGroup)
+	}
+
+	d.SetId(*read.ID)
```
jackofallops, line 121: I suspect we'll need a custom ID for this resource, since this is just the ID of the api_management resource?

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
```diff
@@ -0,0 +1,325 @@
...
+			Create: schema.DefaultTimeout(30 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"api_management_name": azure.SchemaApiManagementName(),
```
jackofallops, line 40: I think we should use the ID of the api_management resource here, rather than the full schema.  The custom_domain resource is a meta resource that modifies an existing api_management resource, so should be intrinsically linked to it. (we can then get the name and resource group from the ID)

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
```diff
@@ -0,0 +1,325 @@
...
+			vals := valsRaw.([]interface{})
+			for _, val := range vals {
+				oldConfig := val.(map[string]interface{})
+
+				if oldConfig["host_name"] == *config.HostName {
+					output["certificate_password"] = oldConfig["certificate_password"]
+					output["certificate"] = oldConfig["certificate"]
+				}
+			}
+		}
```
jackofallops, line 292: Since we're going to the trouble of switching on `config.Type` I think we should make this a helper func and move its use to the cases. It will involve a few more LoC, but will be more explicit about what we're attempting to get from config/state, and should help guard against changes to the structure in the API responses.

sirlatrom:
> Hi @sirlatrom 
> Thanks for this, I've noted a few changes / issues below.
> Aside from those, was there a reason you chose this as a separate resource over extending the `azrure_api_management` resource with an additional `Optional` schema for this config? (The custom ID note below wouldn't be needed in that approach)

Yes, the sole purpose is to be able to create an `azurerm_key_vault_access_policy` in between the `azurerm_api_management` and this resource. Otherwise, the APIM's System Assigned Identity won't be able to get access to the certificates in the Key Vault.

sirlatrom:
I plan to go through your comments this week, hopefully tomorrow CEST.

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
sirlatrom: Probably, yes. How about appending `/customDomains/default` to the ID of the API Management service?

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
sirlatrom: All the other subresources refer to the APIM by resource_group_name and api_management_name, but if you think direct ID is better (i.e. `api_management_id`) I'll change it to that.

sirlatrom: By looking closer, the SDK client also expects resource group name and name parameters, meaning it would bring additional parsing with it if we switch to `api_management_id`.

sirlatrom:
Would it be better, at least in the website docs, to split out the access policies into two separate `azurerm_key_vault_access_policy` resources? That is how I would use it in practice.

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
sirlatrom: I've gone ahead and appended `/customDomains/default` to the ID, not sure if it needs to be added/assumed any more places.

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
katbyte: @sirlatrom - i agree with @jackofallops, we can use the id and parse out the resource group and name with a custom parse function (which this resource should be switched over to anyways) for example see pardse.CdnEndpointID

sirlatrom: I've done it inside the Read function for now. If you think it warrants its own parse function I can make that during the week.

sirlatrom:
Rebased after conflict arose from main branch.

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
sirlatrom: @katbyte I've made the parse function in e61e67d.

sirlatrom:
@katbyte Is there any chance you'd be able to have a look at this during the week?

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
```diff
@@ -0,0 +1,323 @@
...
+	resourceGroup := id.ResourceGroup
+	serviceName := id.Path["service"]
+
+	existing, err := client.Get(ctx, resourceGroup, serviceName)
+	if err != nil {
+		return fmt.Errorf("finding API Management (API Management %q / Resource Group %q): %s", serviceName, resourceGroup, err)
+	}
+
+	if d.IsNewResource() {
+		if existing.ServiceProperties != nil && existing.ServiceProperties.HostnameConfigurations != nil {
```
manicminer, line 111: As per @jackofallops' comment, we should maybe also check for an empty slice at `existing.ServiceProperties.HostnameConfigurations`

```suggestion
		if existing.ServiceProperties != nil && existing.ServiceProperties.HostnameConfigurations != nil && len(*existing.ServiceProperties.HostnameConfigurations) > 0 {
```

website/docs/r/api_management_custom_domain.html.markdown
```diff
@@ -0,0 +1,174 @@
...
+          "portal.example.com",
+        ]
+      }
+    }
+  }
+}
+
+resource "azurerm_api_management_custom_domain" "example" {
+  resource_group_name = azurerm_resource_group.example.name
+  api_management_name = azurerm_api_management.example.name
```
manicminer, line 89:
```suggestion
  api_management_id = azurerm_api_management.example.id
```

website/docs/r/api_management_custom_domain.html.markdown
```diff
@@ -0,0 +1,174 @@
...
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `api_management_name` - (Required) The name of the API Management service to configure Custom Domain for. Changing this forces a new API Management Custom Domain to be created.
+
+* `resource_group_name` - (Required) The name of the Resource Group where the API Management resource exists. Changing this forces a new API Management Custom Domain to be created.
```
manicminer, line 109:
```suggestion
* `api_management_id` - (Required) The ID of the API Management service for which to configure Custom Domains. Changing this forces a new API Management Custom Domain resource to be created.
```

azurerm/internal/services/apimanagement/tests/api_management_custom_domain_resource_test.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func testAccAzureRMApiManagementCustomDomain_basic(data acceptance.TestData) string {
+	template := testAccAzureRMApiManagementCustomDomain_template(data)
+	return fmt.Sprintf(`
+%s
+
+resource "azurerm_api_management_custom_domain" "test" {
+  api_management_name = azurerm_api_management.test.name
+  resource_group_name = azurerm_resource_group.test.name
```
manicminer, line 121:
```suggestion
  api_management_id = azurerm_api_management.test.id
```

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
```diff
@@ -0,0 +1,323 @@
...
+
+	log.Printf("[INFO] preparing arguments for API Management Custom domain creation.")
+
+	apiManagementID := d.Get("api_management_id").(string)
+	id, err := azure.ParseAzureResourceID(apiManagementID)
+	if err != nil {
+		return err
+	}
+	resourceGroup := id.ResourceGroup
+	serviceName := id.Path["service"]
```
manicminer, line 103: Could we capture this in a custom parse function? e.g. `parse.ApiManagementID`

manicminer (changes requested):
Hi @sirlatrom, sorry for the delay in re-reviewing. Some additional comments inline, and could you add an update test to change/remove the custom domains? Once we have this we should be ok to merge. Thanks!

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
sirlatrom: Done in 7c1e893d967d0f500fc6196e9a5431000efb9212.

sirlatrom (commented):
> Hi @sirlatrom, sorry for the delay in re-reviewing. Some additional comments inline, and could you add an update test to change/remove the custom domains? Once we have this we should be ok to merge. Thanks!

Added update test in fa8a2b2fdf8032f5d087f441ca40b73d50fd321d. Please note that if `hostname_configurations` is specified in the API object, Azure will automatically add an entry for the always-on, built-in proxy domain on `<api_management_service_name>.azure-api.net` - hence the changes to the test configurations.

sirlatrom:
@manicminer Is this good to go?

manicminer:
@sirlatrom I'm still working on the tests, currently trying to track down a spurious import error. Unfortunately it's slow going since each APIM resource takes about 45 mins to stand up, and yesterday the API was timing out for GET requests.

sirlatrom:
> 
> 
> @sirlatrom I'm still working on the tests, currently trying to track down a spurious import error. Unfortunately it's slow going since each APIM resource takes about 45 mins to stand up, and yesterday the API was timing out for GET requests.

APIM is indeed a pain to iterate on... If you need a second pair of eyes on anything, just reach out either here or in Slack.

manicminer:
@sirlatrom Could you fix up the vendoring issue with your recent rebase? Thanks!

sirlatrom:
> 
> 
> @sirlatrom Could you fix up the vendoring issue with your recent rebase? Thanks!

I believe it's in good shape now: https://travis-ci.org/github/terraform-providers/terraform-provider-azurerm/builds/738292335

manicminer:
@sirlatrom Thanks, appreciated. I haven't forgotten about this PR; I've spent several days trying to get a single test to run but being hit by various API error states.

sirlatrom:
@manicminer I appreciate how frustrating APIM is to work with, but is there any chance you have made any progress on this PR since last week, or is there perhaps something I can do to move it forward?

manicminer:
**Test Results**

`TestAccDataSourceAzureRMApiManagement_`
Screenshot 2020-11-05 at 14 23 27

`TestAccAzureRMApiManagementCustomDomain_`
Screenshot 2020-11-05 at 16 05 21

jackofallops (approved):
Thanks @manicminer & @sirlatrom - This LGTM now 👍

manicminer:
Thanks @jackofallops :+1:

Latest test results:

Screenshot 2020-11-11 at 10 13 31

PR #8238, opened 2020-08-25
Add example for storage container

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/8214


examples/storage/storage-container/main.tf
```diff
@@ -0,0 +1,46 @@
...
+resource "azurerm_storage_account" "example2" {
+  name                = "${var.prefix}storageacct2"
+  resource_group_name = azurerm_resource_group.example.name
+  location            = azurerm_resource_group.example.location
+
+  account_tier              = "Standard"
+  account_kind              = "StorageV2"
+  account_replication_type  = "LRS"
+  enable_https_traffic_only = true
+  access_tier               = "Cool"
```
tombuildsstuff, line 38: this should probably be

```suggestion
  access_tier               = "Hot"
```

tombuildsstuff (approved):
one comment to fix but this otherwise LGTM 👍

examples/storage/storage-container/main.tf
neil-yechenwei: updated

neil-yechenwei:
@tombuildsstuff , thanks for your comment. I've updated code.

Issue #8121, opened 2020-08-13
Support for 'SocketAddr' in match_variable for Front Door firewall policy (WAF) custom_rule


darrenk13:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Terraform has support in the Front Door WAF custom policies for a match_variable of `RemoteAddr` for match types (operator) of `IPMatch`.  Azure has an option for another match_variable of `SocketAddr`.  Please add  `SocketAddr` to the list of valid options.

Both options filter on remote IP address but `RemoteAddr` will look at the `X-Forwarded-For` header which is insecure. `SocketAddr` does not have that behaviour.

### New or Affected Resource(s)

Resource:
* azurerm_frontdoor_firewall_policy

### Potential Terraform Configuration

```hcl
resource "azurerm_frontdoor_firewall_policy" "policy" {
    location                          = "global"
    name                              = "example"
    resource_group_name               = "resourceGroupName"

    custom_rule {
        action                         = "Block"
        name                           = "IPmatch"
        priority                       = 50
        type                           = "MatchRule"

        match_condition {
            match_values       = ["10.0.0.0/24"]
            negation_condition = true
            operator           = "IPMatch"

            match_variable     = "SocketAddr"
        }
    }
}
```

### References

* https://docs.microsoft.com/en-us/rest/api/frontdoorservice/webapplicationfirewall/policies/createorupdate#matchvariable

PR #8244, opened 2020-08-25
allow SocketAddr to be used as a match_variable

gchappell99:
This is already supported by the 2020-01-01 Front Door schema, which is an included module in v44.2.0+incompatible of azure-sdk-for-go, so this is a relatively simple change to the validation within the provider only and has a test

Fixes: #8121

### DIFF
--- a/azurerm/internal/services/frontdoor/frontdoor_firewall_policy_resource.go
+++ b/azurerm/internal/services/frontdoor/frontdoor_firewall_policy_resource.go
@@ -167,6 +167,7 @@ func resourceArmFrontDoorFirewallPolicy() *schema.Resource {
 											string(frontdoor.RequestHeader),
 											string(frontdoor.RequestMethod),
 											string(frontdoor.RequestURI),
+											string(frontdoor.SocketAddr),
 										}, false),
 									},
 
--- a/azurerm/internal/services/frontdoor/tests/frontdoor_firewall_policy_resource_test.go
+++ b/azurerm/internal/services/frontdoor/tests/frontdoor_firewall_policy_resource_test.go
@@ -72,6 +72,7 @@ func TestAccAzureRMFrontDoorFirewallPolicy_update(t *testing.T) {
 					resource.TestCheckResourceAttr(data.ResourceName, "name", fmt.Sprintf("testAccFrontDoorWAF%d", data.RandomInteger)),
 					resource.TestCheckResourceAttr(data.ResourceName, "mode", "Prevention"),
 					resource.TestCheckResourceAttr(data.ResourceName, "custom_rule.1.name", "Rule2"),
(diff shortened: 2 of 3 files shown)


azurerm/internal/services/frontdoor/frontdoor_firewall_policy_resource.go
```diff
@@ -167,6 +167,7 @@ func resourceArmFrontDoorFirewallPolicy() *schema.Resource {
 											string(frontdoor.RequestHeader),
 											string(frontdoor.RequestMethod),
 											string(frontdoor.RequestURI),
+											string(frontdoor.SocketAddr),
```
tombuildsstuff, line 170: can we [add this to the documentation too](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/frontdoor_firewall_policy.html.markdown#L170)?

tombuildsstuff (changes requested):
hey @gchappell99 

Thanks for this PR :)

Taking a look through this LGTM - if we can document this new field (and the tests pass) then this should otherwise be good to merge 👍 

Thanks!

azurerm/internal/services/frontdoor/tests/frontdoor_firewall_policy_resource_test.go
```diff
@@ -336,6 +337,32 @@ resource "azurerm_frontdoor_firewall_policy" "test" {
     }
   }
 
+  custom_rule {
+    name                           = "Rule3"
+    enabled                        = true
+    priority                       = 2
```
tombuildsstuff, line 343: looks like the priority needs to be different here:

```
------- Stdout: -------
=== RUN   TestAccAzureRMFrontDoorFirewallPolicy_complete
=== PAUSE TestAccAzureRMFrontDoorFirewallPolicy_complete
=== CONT  TestAccAzureRMFrontDoorFirewallPolicy_complete
TestAccAzureRMFrontDoorFirewallPolicy_complete: testing.go:684: Step 0 error: errors during apply:
Error: creating Front Door Firewall Policy "testAccFrontDoorWAF200826081908186558" (Resource Group "testaccRG-200826081908186558"): frontdoor.PoliciesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="BadRequest" Message="WebApplicationFirewallPolicy validation failed. More information \"Custom rules Rule2 and Rule3 have duplicate priority: 2\"."
```

could we update this so that the tests can pass?

gchappell99: This should now be resolved also, and I think I've found the right values for the environment variables required to run acceptance tests locally in future.

tombuildsstuff:
Tests pass:

Screenshot 2020-08-27 at 08 20 21

tombuildsstuff (approved):
LGTM - thanks for pushing those changes @gchappell99

PR #8256, opened 2020-08-26
refactor: removing usages of the deprecated `features.ShouldResourcesBeImported()`

tombuildsstuff:
Doing this gradually to make this less impactful - this takes us from 203 -> 181 usages (doing most of the smaller/single usage packages)


manicminer (approved):
LGTM

PR #8257, opened 2020-08-26
refactor: updating the networking package to use a parse package like the others

tombuildsstuff:
The ID parsers in this directory existed prior to the `parse` directory convention - so this PR refactors them into that folder

Also renames a few instances of `{Type}ResourceID` to `{Type}Id` to match the convention for types


tombuildsstuff:
closing since half of this has been done in #7601 so these'll conflict

PR #8258, opened 2020-08-26
Fix wrong title for data source azurerm_key_vault_certificate

sebader:
Fixing wrong title.

There might be other things wrong here as well. The sample shows to retrieve thumbprint but the attribute reference further down does not mention this. Not sure what else might be missing.


sebader:
good spot @tombuildsstuff, fixed

tombuildsstuff (approved):
LGTM - thanks for fixing that @sebader

Issue #7557, opened 2020-07-02
Application Gateway, Health Probe does not support custom port.


CarelvanHeerden:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected Resource(s)

* azurerm_application_gateway

### Potential Terraform Configuration

```hcl
resource "azurerm_application_gateway" "application_gw" {
........
  probe {
    host = "127.0.0.1"
    interval = 30
    minimum_servers = 0
    name  = "traefik_dashboard"
    path = "/dashboard/"
    pick_host_name_from_backend_http_settings = false
    protocol = "Http"
    timeout = 30
    unhealthy_threshold = 3
    pick_port_from_backend_HTTP_settings  = false <--This is new
    custom_port = 8081 <--This is new


    match {
      status_code = [
        "200-399",
      ]
    }
  }
.......
}
```

### References

I cannot find a reference to this setting in the Microsoft documentation. Below a screen grab from the azure portal

<img width="608" alt="Screenshot 2020-07-02 at 09 11 56" src="https://user-images.githubusercontent.com/17004772/86328055-80a02b80-bc44-11ea-9908-f722c747ff49.png">

* #0000

PR #8278, opened 2020-08-28
`azurerm_application_gateway ` - Add support for `probe.properties.port`

simnyc:
Fixes [#7557](https://github.com/terraform-providers/terraform-provider-azurerm/issues/7557).

* add support for the property
* create an acceptance test
* add an entry in the documentation

From Azure API [doc](https://docs.microsoft.com/en-us/rest/api/application-gateway/applicationgateways/createorupdate#applicationgatewayprobe):

> Custom port which will be used for probing the backend servers. The valid value ranges from 1 to 65535. In case not set, port from http settings will be used.

### DIFF
--- a/azurerm/internal/services/network/application_gateway_resource.go
+++ b/azurerm/internal/services/network/application_gateway_resource.go
@@ -846,6 +846,12 @@ func resourceArmApplicationGateway() *schema.Resource {
 							Required: true,
 						},
 
+						"port": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validate.PortNumber,
+						},
+
 						"pick_host_name_from_backend_http_settings": {
 							Type:     schema.TypeBool,
 							Optional: true,
@@ -2582,6 +2588,7 @@ func expandApplicationGatewayProbes(d *schema.ResourceData) *[]network.Applicati
 		name := v["name"].(string)
 		probePath := v["path"].(string)
 		protocol := v["protocol"].(string)
+		port := int32(v["port"].(int))
 		timeout := int32(v["timeout"].(int))
 		unhealthyThreshold := int32(v["unhealthy_threshold"].(int))
 		pickHostNameFromBackendHTTPSettings := v["pick_host_name_from_backend_http_settings"].(bool)
@@ -2616,6 +2623,10 @@ func expandApplicationGatewayProbes(d *schema.ResourceData) *[]network.Applicati
 			}
 			outputMatch.Body = utils.String(matchBody)
 			output.ApplicationGatewayProbePropertiesFormat.Match = outputMatch
+		}
+
+		if port != 0 {
+			output.ApplicationGatewayProbePropertiesFormat.Port = utils.Int32(port)
 		}
 
 		results = append(results, output)
@@ -2663,6 +2674,12 @@ func flattenApplicationGatewayProbes(input *[]network.ApplicationGatewayProbe) [
 			if threshold := props.UnhealthyThreshold; threshold != nil {
 				output["unhealthy_threshold"] = int(*threshold)
 			}
+
+			port := 0
+			if props.Port != nil {
+				port = int(*props.Port)
+			}
+			output["port"] = port
 
 			if pickHostNameFromBackendHTTPSettings := props.PickHostNameFromBackendHTTPSettings; pickHostNameFromBackendHTTPSettings != nil {
 				output["pick_host_name_from_backend_http_settings"] = *pickHostNameFromBackendHTTPSettings
--- a/azurerm/internal/services/network/tests/application_gateway_resource_test.go
+++ b/azurerm/internal/services/network/tests/application_gateway_resource_test.go
@@ -502,6 +502,26 @@ func TestAccAzureRMApplicationGateway_probesPickHostNameFromBackendHTTPSettings(
 	})
 }
 
+func TestAccAzureRMApplicationGateway_probesWithPort(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_application_gateway", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMApplicationGatewayDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMApplicationGateway_probesWithPort(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMApplicationGatewayExists(data.ResourceName),
+					resource.TestCheckResourceAttr(data.ResourceName, "probe.0.port", "8082"),
+				),
+			},
+			data.ImportStep(),
+		},
+	})
+}
+
 func TestAccAzureRMApplicationGateway_backendHttpSettingsHostName(t *testing.T) {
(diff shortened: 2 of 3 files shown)


simnyc:
I'm still having some issues with the acceptance test:

The AG is created properly with the `port` set to `8082`:

```
terraform-provider-azurerm git:(simon/app-gw-probe-port) ✗ make acctests SERVICE='network' TESTARGS='-run=TestAccAzureRMApplicationGateway_probesWithPort' TESTTIMEOUT='60m'
...
CREATE: azurerm_application_gateway.test
...
  name:                                                        "" => "acctestag-200827185432169246"
  probe.#:                                                     "" => "1"
  probe.0.host:                                                "" => "azure.com"
  probe.0.id:                                                  "" => "<computed>"
  probe.0.interval:                                            "" => "300"
  probe.0.match:                                               "" => "<computed>"
  probe.0.minimum_servers:                                     "" => "0"
  probe.0.name:                                                "" => "acctest-vnet-200827185432169246-probe1"
  probe.0.path:                                                "" => "/test"
  probe.0.pick_host_name_from_backend_http_settings:           "" => "false"
  probe.0.port:                                                "" => "8082"
```

I can confirm that the port is set properly using the CLI or the Portal:

```
az network application-gateway show --name acctestag-200827185432169246 -g acctestRG-200827185432169246
...
  "probes": [
    {
...
      "name": "acctest-vnet-200827185432169246-probe1",
      "path": "/test",
      "pickHostNameFromBackendHttpSettings": false,
      "port": 8082,
      "protocol": "Http",
      "provisioningState": "Updating",
      "resourceGroup": "acctestRG-200827185432169246",
      "timeout": 120,
      "type": "Microsoft.Network/applicationGateways/probes",
      "unhealthyThreshold": 8
    }
```

But later on the test ignores the value:

```
2020/08/27 18:59:02 [WARN] Provider "azurerm" produced an unexpected new value for azurerm_application_gateway.test, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .enable_http2: was null, but now cty.False
      - .probe[0].port: was cty.NumberIntVal(8082), but now cty.NumberIntVal(0)
      - .backend_http_settings[0].host_name: was null, but now cty.StringVal("")
      - .backend_http_settings[0].affinity_cookie_name: was null, but now cty.StringVal("")
      - .backend_http_settings[0].path: was null, but now cty.StringVal("")
      - .http_listener[0].ssl_certificate_name: was null, but now cty.StringVal("")
      - .http_listener[0].host_name: was null, but now cty.StringVal("")
      - .http_listener[0].require_sni: was null, but now cty.False
      - .http_listener[0].firewall_policy_id: was null, but now cty.StringVal("")
    TestAccAzureRMApplicationGateway_probesWithPort: testing.go:684: Step 0 error: Check failed: Check 2/2 error: azurerm_application_gateway.test: Attribute 'probe.0.port' expected "8082", got "0"
```

Not sure what's going on here, would appreciate some help.

azurerm/internal/services/network/application_gateway_resource.go
```diff
@@ -821,6 +821,12 @@ func resourceArmApplicationGateway() *schema.Resource {
 							}, true),
 						},
 
+						"port": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validate.PortNumber,
+						},
```
tombuildsstuff, line 828: _minor_ this is sorted Required fields then Optional fields (and then alphabetically) - could we move this below so it's consistent?

can we also [add this new field to the documentation](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/application_gateway.html.markdown#L343-L355)?

azurerm/internal/services/network/application_gateway_resource.go
```diff
@@ -2594,6 +2601,7 @@ func expandApplicationGatewayProbes(d *schema.ResourceData) *[]network.Applicati
 				MinServers:                          utils.Int32(minServers),
 				Path:                                utils.String(probePath),
 				Protocol:                            network.ApplicationGatewayProtocol(protocol),
+				Port:                                utils.Int32(port),
```
tombuildsstuff, line 2604: since this is an optional field, presumably a default value of `0` is invalid? if so, we may need to conditionally set this field? e.g.

```
if port != 0 {
  output. ApplicationGatewayProbePropertiesFormat.Port = utils.Int32(port)
}
```

tombuildsstuff (changes requested):
hey @simnyc 

Thanks for this PR :)

Taking a look through this mostly LGTM - if we can fix up the comments then this otherwise LGTM 👍 

Thanks!

azurerm/internal/services/network/application_gateway_resource.go
simnyc: Fixed. Though it seems the Optional fields are not sorted alphabetically.

azurerm/internal/services/network/application_gateway_resource.go
simnyc: Makes sense. Fixed.

tombuildsstuff (approved):
LGTM - thanks for pushing those changes @simnyc

simnyc:
Hey @tombuildsstuff , what's the process now, do you need anything else from me? When can we expect it to be released?
Thanks

tombuildsstuff:
Tests pass:

Screenshot 2020-08-28 at 18 14 44

tombuildsstuff:
@simnyc just waiting for the acctests to pass, we should be good once they're done :)

azurerm/internal/services/network/application_gateway_resource.go
```diff
@@ -846,6 +846,12 @@ func resourceArmApplicationGateway() *schema.Resource {
 							Required: true,
 						},
 
+						"port": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validate.PortNumber,
+						},
```
tombuildsstuff, line 853: I've run the tests and just realised we're not reading this back in in the `flatten` function - can we update this to always export that e.g.:

```
port := 0
if props.Port != nil {
  port = int(*props.Port)
}
output["port"] = port
```

so the tests fail:

```
------- Stdout: -------
=== RUN   TestAccAzureRMApplicationGateway_probesWithPort
=== PAUSE TestAccAzureRMApplicationGateway_probesWithPort
=== CONT  TestAccAzureRMApplicationGateway_probesWithPort
TestAccAzureRMApplicationGateway_probesWithPort: testing.go:684: Step 0 error: Check failed: Check 2/2 error: azurerm_application_gateway.test: Attribute 'probe.0.port' expected "8082", got "0"
--- FAIL: TestAccAzureRMApplicationGateway_probesWithPort (763.77s)
FAIL
```

could we add this to the flatten function?

simnyc: Thanks, added this piece to `flattenApplicationGatewayProbes`.

simnyc:
Not sure what's going on with the CI but the test is finally passing in my subscription:

```
 make acctests SERVICE='network' TESTARGS='-run=TestAccAzureRMApplicationGateway_probesWithPort' TESTTIMEOUT='60m'
--- PASS: TestAccAzureRMApplicationGateway_probesWithPort (772.44s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/tests	773.665s
```

Thanks for the help @tombuildsstuff !

PR #8279, opened 2020-08-28
[Refactor] add ID formatter for `azurerm_availability_set`

ArcturusZhang:


ArcturusZhang:
Question: is it possible to change the signature of `resourceid.Formatter` to `ID() string`? Because by parsing the ID in the parsing function, the subscription ID should be included - `azure.ParseAzureResourceID` has this information, but discarded in the parsing function.

azurerm/internal/services/compute/availability_set_data_source.go
```diff
@@ -74,10 +77,17 @@ func dataSourceArmAvailabilitySetRead(d *schema.ResourceData, meta interface{})
...
 	}
 
-	d.SetId(*resp.ID)
-	if location := resp.Location; location != nil {
-		d.Set("location", azure.NormalizeLocation(*location))
+	if resp.ID == nil || *resp.ID == "" {
+		return fmt.Errorf("nil or empty ID of Availability Set %q (Resource Group %q)", name, resGroup)
 	}
+
+	id, err := parse.AvailabilitySetID(*resp.ID)
```
tombuildsstuff, line 84: rather than parsing this we could new this up and format it here instead?

azurerm/internal/services/compute/migration/availability_set.go
```diff
@@ -0,0 +1,88 @@
...
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"location": azure.SchemaLocation(),
+
+			"platform_update_domain_count": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      5,
+				ForceNew:     true,
+				ValidateFunc: validation.IntBetween(1, 20),
```
tombuildsstuff, line 32: we can remove the validation funcs here:

```suggestion
```

azurerm/internal/services/compute/availability_set_resource.go
```diff
@@ -139,12 +153,11 @@ func resourceArmAvailabilitySetCreateUpdate(d *schema.ResourceData, meta interfa
...
 
-	resp, err := client.CreateOrUpdate(ctx, resGroup, name, availSet)
-	if err != nil {
-		return err
+	if _, err := client.CreateOrUpdate(ctx, resGroup, name, availSet); err != nil {
+		return fmt.Errorf("creating/updating Availability Set %q (Resource Group %q): %+v", name, resGroup, err)
 	}
 
-	d.SetId(*resp.ID)
+	d.SetId(availabilitySetId.ID(subscriptionId))
```
tombuildsstuff, line 160: ordinarily we'd need to ensure this is rewritten in the Read functions in downstream resources, that is:

- azurerm_linux_virtual_machine
- azurerm_site_recovery_replicated_vm
- azurerm_windows_virtual_machine

else users will have a diff here - however in this instance all of these are ignoring the casing on the Resource ID, due to the API returning this incorrectly - so this _appears_ fine

tombuildsstuff (changes requested):
A couple of things to fix but if we can fix those up then this otherwise LGTM 👍 

Out of interest what's the requirement for rewriting these ID's?

Issue #8219, opened 2020-08-24
Support for azurerm_cluster_node_pool with min_count 0


arodriguezdlc:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Support for aks nodepools with min count to 0. I have seen there is a TODO at the code, but I couldn't find any issue to keep track of this feature. 

### New or Affected Resource(s)

* azurerm_kubernetes_cluster_node_pool

### Potential Terraform Configuration

```hcl
resource "azurerm_kubernetes_cluster_node_pool" "user" {
  name                  = "user"
  kubernetes_cluster_id = azurerm_kubernetes_cluster.aks.id
  vm_size               = "Standard_DS2_v2"
  enable_auto_scaling   = true
  min_count             = 0
  max_count             = 20
  max_pods              = 80
}
```

jackofallops:
Closing as resolved by 8300 which will be part of the 2.31.0 release.

PR #8300, opened 2020-08-31
`azurerm_kubernetes_cluster_node_pool` - `node_count`, `min_node`, and `max_node` can now be set to `0`

zeroX-tj:
This updates the validation rules for the new Azure SDK where you can create zero size nodepools

#8219 

If there's anything else i should do, let me know, first PR. I tested it locally by using it.

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_cluster_node_pool_resource.go
+++ b/azurerm/internal/services/containers/kubernetes_cluster_node_pool_resource.go
@@ -352,7 +352,7 @@ func resourceArmKubernetesClusterNodePoolCreate(d *schema.ResourceData, meta int
 			return fmt.Errorf("`min_count` must be configured when `enable_auto_scaling` is set to `true`")
 		}
 
-		if minCount > maxCount {
+		if minCount >= maxCount {
 			return fmt.Errorf("`max_count` must be >= `min_count`")
 		}
 	} else if minCount > 0 || maxCount > 0 {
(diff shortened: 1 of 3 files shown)


gdippolito:
I think this would be really nice to have. @zeroX-tj are you looking into addressing the comment?

zeroX-tj:
> I think this would be really nice to have. @zeroX-tj are you looking into addressing the comment?

I'm really not so familiar with Go so i was planning in looking into it when i had more time. But that didn't happen yet.

gdippolito (approved):

gdippolito (approved):

migolfi:
This looks great and it would be very helpful for some work I am currently on. Is there anything else needed on this? 
It looks like it's waiting for feedback from @mbfrahry

mbfrahry (changes requested):
Hey @zeroX-tj. These changes look good and the test changes works! We are missing two things after further review. `max_count` validation should also start at 0 since it can be greater than or equal to `min_count` and we need to update the docs to reflect the changes here and we'll be good to go. https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/docs/r/kubernetes_cluster_node_pool.html.markdown

zeroX-tj:
Sorry for the reverts, was a bit confused by the failed builds but then i figured out it's because of cancelation :) Monday mornings 

I think everything should be good now.

mbfrahry (approved):
LGTM!

mbfrahry:
Thanks for this @zeroX-tj!

PR #7951, opened 2020-07-29
New resource azurerm_resource_provider

Humoiz:
Implementing feature mentioned in the feature request: #1832

This PR adds following new resource, tests and its documentation:
azurerm_resource_provider_registration

It allows to register specified resource providers.

(fixes #1832)

PR #8309, opened 2020-09-01
refactor: splitting Resource Providers out to their own package

tombuildsstuff:
This is required to unblock #7951, which has a circular reference otherwise


jackofallops (approved):
LGTM 👍

PR #8310, opened 2020-09-01
Remove constraint on network_profile_id of azurerm_container_group

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/8307

--- PASS: TestAccAzureRMContainerGroup_withVirtualNetworkAndIdentity (479.58s)


neil-yechenwei:
After confirmed with service team, the constraint cannot be removed since they didn't support this scenario.

Issue #7396, opened 2020-06-18
Kube-dashboard gets updated on every apply even though it's not configured/used


muellermatthias:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.12.25
provider.azurerm v2.14.0

### Affected Resource(s)

* `azurerm_kubernetes_cluster`

### Terraform Configuration Files

```hcl
resource "azurerm_resource_group" "example" {
  name     = "example"
  location = "West Europe"
}

resource "azurerm_kubernetes_cluster" "example" {
  name                = "example"
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name
  dns_prefix          = "example"

  default_node_pool {
    name       = "default"
    node_count = 1
    vm_size    = "Standard_D2_v2"
  }

  identity {
    type = "SystemAssigned"
  }

  addon_profile {
    oms_agent {
      enabled                    = true
      log_analytics_workspace_id = azurerm_log_analytics_workspace.example.id
    }
  }
}

resource "azurerm_log_analytics_workspace" "example" {
  name                = "example"
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name
  sku                 = "PerGB2018"
  retention_in_days   = 30
}

provider "azurerm" {
  features {}
  version = "2.14.0"
}

```

### Expected Behavior

After provisioning the resources applying the same configuration again should show no changes.

### Actual Behavior

The plan wants to change the kube_dashboard addon:

```
~ addon_profile {
          - kube_dashboard {
              - enabled = true -> null
            }
...
```
Applying the change won't help, because after that the change will be shown again.

I tried the same configuration without addon_profile and it works fine.

### Steps to Reproduce

1. Run `terraform apply` to provision the resources

2. Run `terraform plan/apply` to see the changes

neil-yechenwei:
Thanks for opening this issue. After tested, seems api would return default value "enabled = true" for kube_dashboard. Hopes the fix would be helpful.

neil-yechenwei:
[...]

PR #8320, opened 2020-09-02
Add computed attribute on kube_dashboard of azurerm_kubernetes_cluster

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7396


tombuildsstuff:
👋 

The `kube_dashboard` block itself is intentionally not Computed since this isn't available in some regions - and trying to set this block (even to disabled) causes an API error - as such unfortunately this cannot become Computed and so I'm going to close this PR for the moment - however it should be possible to solve #7396 by specifying this block within the config.

Thanks!

Issue #7594, opened 2020-07-06
New Data Source: `azurerm_mssql_server`


MeNsaaH:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Add mssql server data resource such as this https://www.terraform.io/docs/providers/azurerm/d/sql_server.html but for mssql

### Potential Terraform Configuration

```hcl
data "azurerm_mssql_server" "example" {
  name                = "examplesqlservername"
  resource_group_name = "example-resources"
}

output "mssql_server_id" {
  value = data.azurerm_sql_server.example.id
}

```

### References

* #0000

Issue #7295, opened 2020-06-11
azurerm_sql_firewall_rule: TLS minimal version support for AzureSQL server


Quantalian:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### AzureRM Provider Version

* 2.4.0

### Description

 After Azure latest update for App Registration 'TLS minimal version' is required in Azure SQL server firewall rules if you are using Msal.js <= 1.0 for Single Page Application.

### Expected behavior
ssl_minimal_tls_version  = "TLS1_2"

### Actual behavior
not supported, need to define it from portal after terraform apply

### New or Affected Resource(s)

* azurerm_sql_firewall_rule

### References
* https://azure.microsoft.com/en-au/updates/in-development-build-single-page-application-with-msal-js/

HarleyB123:
Hey 👋 

Whilst this has been solved for azurerm_mssql_server, it would be really cool if this same sort of implementation was available for [azurerm_sql_server](https://www.terraform.io/docs/providers/azurerm/r/sql_server.html)

Will try and make a PR for this myself, though my Go is sketchy and would be copying the great work already done in the mssql PR :)

dhirschfeld:
Hi @HarleyB123 👋 
Just wondering why you wouldn't simply port to `azurerm_mssql_server` since they spin up the same infrastructure?

Even though it's not apparently documented anywhere, from my inquiry `azurerm_sql_server` is effectively a deprecated api which has been superseded by `azurerm_mssql_server`: https://github.com/terraform-providers/terraform-provider-azurerm/issues/8338#issuecomment-689933236

HarleyB123:
Thanks @dhirschfeld , I agree that documentation would be beneficial on this as I thought they were two separate resources too! :) 

Will port over to ```azurerm_mssql_server```

PR #8361, opened 2020-09-05
mssql_server: add minimal_tls_version property

manicminer:
Closes: #7295
Closes: #7594

### DIFF
--- a/azurerm/internal/services/mssql/mssql_server_resource.go
+++ b/azurerm/internal/services/mssql/mssql_server_resource.go
@@ -10,6 +10,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	uuid "github.com/satori/go.uuid"
+
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
@@ -136,6 +137,16 @@ func resourceArmMsSqlServer() *schema.Resource {
 				},
 			},
 
+			"minimal_tls_version": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"1.0",
+					"1.1",
+					"1.2",
+				}, false),
+			},
+
 			"public_network_access_enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
@@ -183,7 +194,7 @@ func resourceArmMsSqlServerCreateUpdate(d *schema.ResourceData, meta interface{}
 		existing, err := client.Get(ctx, resGroup, name)
 		if err != nil {
 			if !utils.ResponseWasNotFound(existing.Response) {
-				return fmt.Errorf("Error checking for presence of existing SQL Server %q (Resource Group %q): %+v", name, resGroup, err)
+				return fmt.Errorf("checking for presence of existing SQL Server %q (Resource Group %q): %+v", name, resGroup, err)
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/mssql/mssql_server_resource.go
```diff
@@ -136,6 +137,16 @@ func resourceArmMsSqlServer() *schema.Resource {
 				},
 			},
 
+			"minimal_tls_version": {
```
katbyte, line 140: This might be better as `minimum_tls_version`? or even `ssl_minimum_tls_version` ?

website/docs/r/mssql_server.html.markdown
```diff
@@ -79,6 +80,8 @@ The following arguments are supported:
 
 * `identity` - (Optional) An `identity` block as defined below.
 
+* `minimal_tls_version` - (Optional) The Minimal TLS Version for all SQL Database and SQL Data Warehouse databases associated with the server. Valid values are: 1.0, 1.1 and 1.2.
```
katbyte, line 83: we should quote the values
```suggestion
* `minimal_tls_version` - (Optional) The Minimal TLS Version for all SQL Database and SQL Data Warehouse databases associated with the server. Valid values are: `1.0`, `1.1` and `1.2`.
```

website/docs/r/mssql_server.html.markdown
```diff
@@ -99,8 +102,6 @@ The following attributes are exported:
 
 * `fully_qualified_domain_name` - The fully qualified domain name of the Azure SQL Server (e.g. myServerName.database.windows.net)
 
-* `restorable_dropped_database_ids` - A list of dropped restorable database IDs on the server.
```
katbyte, line 102: Is there a reason we are removing this?

website/docs/r/mssql_server.html.markdown
```diff
@@ -1,7 +1,7 @@
 ---
 subcategory: "Database"
 layout: "azurerm"
-page_title: "Azure Resource Manager: azurerm_mssql_server"
+page_title: "Azure Resource Manager: azurerm_sql_server"
```
katbyte, line 4: I don't think we should be making this change?

katbyte (changes requested):
Thanks @manicminer - looks good aside from a couple comments. I think we should change the property name to something that reads a bit better and matches redis frontdoor what we'd expect the property to be named.

Looks like postgres and mysql have the poorly named property added already :( might want to rename them for 3.0

azurerm/internal/services/mssql/mssql_server_resource.go
manicminer: Agreed I wouldn't have chosen `minimal` but it seems to have pervaded. It did prompt me to check the dictionary though and _technically_ it seems appropriate, though it feels a bit arcane to me.

website/docs/r/mssql_server.html.markdown
manicminer: Oops, not sure how I did that!

website/docs/r/mssql_server.html.markdown
manicminer: No reason! Fixed :D

manicminer:
@katbyte Thanks for the review and catching my bad edits. Should all be fixed up now.

website/docs/r/mssql_server.html.markdown
```diff
@@ -109,11 +112,11 @@ The following attributes are exported:
 
 * `tenant_id` - The Tenant ID for the Service Principal associated with the Identity of this SQL Server.
 
--> You can access the Principal ID via `${azurerm_sql_server.example.identity.0.principal_id}` and the Tenant ID via `${azurerm_sql_server.example.identity.0.tenant_id}`
+-> You can access the Principal ID via `${azurerm_mssql_server.example.identity.0.principal_id}` and the Tenant ID via `${azurerm_mssql_server.example.identity.0.tenant_id}`
```
katbyte, line 115: Should we change this to .13 syntax?

azurerm/internal/services/mssql/mssql_server_resource.go
katbyte: Yep, regardless of the api/sdk and existing properties i'd prefer we correct it (and the others) to minimum as technically right or not, to match what other resources/apis and the rest of the world uses

katbyte (changes requested):

katbyte (approved):
Thanks @manicminer - LGTM now 🚀

PR #8368, opened 2020-09-07
Set license_type with user specified value for azurerm_mssql_database

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/8346

### DIFF
--- a/azurerm/internal/services/mssql/mssql_database_resource.go
+++ b/azurerm/internal/services/mssql/mssql_database_resource.go
@@ -447,7 +447,7 @@ func resourceArmMsSqlDatabaseRead(d *schema.ResourceData, meta interface{}) erro
 		d.Set("auto_pause_delay_in_minutes", props.AutoPauseDelay)
 		d.Set("collation", props.Collation)
 		d.Set("elastic_pool_id", props.ElasticPoolID)
-		d.Set("license_type", props.LicenseType)
+		d.Set("license_type", d.Get("license_type").(string))
 		if props.MaxSizeBytes != nil {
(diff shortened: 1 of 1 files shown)


azurerm/internal/services/mssql/mssql_database_resource.go
```diff
@@ -447,7 +447,7 @@ func resourceArmMsSqlDatabaseRead(d *schema.ResourceData, meta interface{}) erro
 		d.Set("auto_pause_delay_in_minutes", props.AutoPauseDelay)
 		d.Set("collation", props.Collation)
 		d.Set("elastic_pool_id", props.ElasticPoolID)
-		d.Set("license_type", props.LicenseType)
+		d.Set("license_type", d.Get("license_type").(string))
```
tombuildsstuff, line 450: this doesn't track if the values changed outside of Terraform - is this not being returned from the API? if so that's an API bug?

tombuildsstuff (changes requested):
hey @neil-yechenwei 

Taking a look through this appears to be an API bug - unfortunately the workaround proposed here won't allow us to track changes made externally to this resource - so unfortunately we need to take a different approach here (likely raising an API bug); can you confirm if this information is/is not returned from the API here?

Thanks!

azurerm/internal/services/mssql/mssql_database_resource.go
neil-yechenwei: Yes. API doesn't return the value of LicenseType. OK. I will close this PR and raise an issue on rest api spec.

PR #8372, opened 2020-09-07
azurerm_application_insights_webtests: .Kind turns to null after editing in Portal

### LINKED ISSUE #8356: Azurerm_application_insights_web_test always forces replacement after resource is changed in Portal
### Community Note

Both 2.24.0 and 2.23.0

### Affected Resource(s)
azurerm_application_insights_web_test

```
resource "azurerm_application_insights_web_test" "availability_test" {
...
  kind = "ping"
...
```
### Expected Behavior

There shoudn't be change detected in "kind" property after changing some setting in Azure Portal.

### Actual Behavior

`terraform plan` says that the argument "kind" is added with value of "ping" and needs replacement:

```+ kind = "ping" # forces replacement```

### Steps to Reproduce

1. Create new azurerm_application_insights_web_test with kind "ping"
2. Run `terraform apply`
3. Go to portal and change something - e.g add another region and then remove it (thats what I did)
4. Run `terraform plan`

### PR DESCRIPTION
**Pull request by magodo**
After doing some trivial operation on a terraform managed application insights webtest on Portal, the `kind` will turn into `null`. This seems because the request in Portal doesn't set the `.kind` property but the `.properties.Kind`. To mitigate the impact for those poor souls who want to mix the two, we side pass this problem by keeping the same way as Portal does, in read.

This fixes #8356.

### DIFF
--- a/azurerm/internal/services/applicationinsights/application_insights_webtests_resource.go
+++ b/azurerm/internal/services/applicationinsights/application_insights_webtests_resource.go
@@ -237,13 +237,13 @@ func resourceArmApplicationInsightsWebTestsRead(d *schema.ResourceData, meta int
 	d.Set("application_insights_id", appInsightsId)
 	d.Set("name", resp.Name)
 	d.Set("resource_group_name", resGroup)
-	d.Set("kind", resp.Kind)
 
 	if location := resp.Location; location != nil {
 		d.Set("location", azure.NormalizeLocation(*location))
 	}
 
 	if props := resp.WebTestProperties; props != nil {
+		d.Set("kind", string(props.WebTestKind))
 		d.Set("synthetic_monitor_id", props.SyntheticMonitorID)
 		d.Set("description", props.Description)
 		d.Set("enabled", props.Enabled)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: azurerm/internal/services/applicationinsights/application_insights_webtests_resource.go [hunk 1]
```diff
@@ -237,13 +237,13 @@ func resourceArmApplicationInsightsWebTestsRead(d *schema.ResourceData, meta int
 	d.Set("application_insights_id", appInsightsId)
 	d.Set("name", resp.Name)
 	d.Set("resource_group_name", resGroup)
-	d.Set("kind", resp.Kind)
```
manicminer, line 240: Suggest leaving this as-is for compatibility

azurerm/internal/services/applicationinsights/application_insights_webtests_resource.go
```diff
@@ -237,13 +237,13 @@ func resourceArmApplicationInsightsWebTestsRead(d *schema.ResourceData, meta int
...
 
 	if location := resp.Location; location != nil {
 		d.Set("location", azure.NormalizeLocation(*location))
 	}
 
 	if props := resp.WebTestProperties; props != nil {
+		d.Set("kind", string(props.WebTestKind))
```
manicminer, line 246: This would maybe be safer checking for the top-level field first?

```suggestion
		if resp.Kind == "" {
			d.Set("kind", string(props.WebTestKind))
		}
```

manicminer (changes requested):
@magodo thanks for the fix. My only concern is that we should still be reading from the top-level field if it's present?

manicminer (approved):
Thanks @magodo. LGTM :+1:

Issue #8344, opened 2020-09-03
log_analytics_workspace_id for oms_agent isn't stored in the state


muellermatthias:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

- Provider Version: 2.25.0
- I tested it with Terraform 0.12.25, 0.12.29 and 0.13.0

### Affected Resource(s)

* `azurerm_log_analytics_workspace`
* `azurerm_kubernetes_cluster`

### Terraform Configuration Files

```hcl
resource "azurerm_resource_group" "example" {
  name     = "example"
  location = "West Europe"
}

resource "azurerm_kubernetes_cluster" "example" {
  name                = "example"
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name
  dns_prefix          = "example"

  default_node_pool {
    name       = "default"
    node_count = 1
    vm_size    = "Standard_D2_v2"
  }

  identity {
    type = "SystemAssigned"
  }

  addon_profile {
    oms_agent {
      enabled                    = true
      log_analytics_workspace_id = azurerm_log_analytics_workspace.example.id
    }

    kube_dashboard {
      enabled = true
    }
  }
}

resource "azurerm_log_analytics_workspace" "example" {
  name                = "example"
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name
  sku                 = "PerGB2018"
  retention_in_days   = 30
}

provider "azurerm" {
  features {}
  version = "2.25.0"
}
```
### Steps to Reproduce

1. `terraform apply`
2. Disable OMS agent:
```
...
    oms_agent {
      enabled = false
    }
...
```
3. `terraform apply`
4. Enable OMS agent:
```
...
    oms_agent {
      enabled                    = true
      log_analytics_workspace_id = azurerm_log_analytics_workspace.example.id
    }
...
```
5. `terraform apply`
6. `terraform apply`

### Expected Behavior

The last `terraform apply` in step 6 should show no changes.

### Actual Behavior

Terraform wants to add the LAW id in step 6.
```
...
[...]

PR #8374, opened 2020-09-07
azurerm_kubernetes_cluster - ensure OMS Agent log analytics workspace case is preserved after disabling/enabling

magodo:
AKS service will somehow ignore the config key cases after disable then enable the omsagent. Opened an issue to track this issue: Azure/azure-rest-api-specs#10716.

This PR is a workaround to unblock #8344.


tombuildsstuff (approved):
LGTM 👍

PR #8383, opened 2020-09-08
Bug: `azurerm_windows_virtual_machine` and `azurerm_linux_virtual_machine` - Add WaitForState on VM delete 

WodansSon:
This PR is a fix for the below Issue:
Today customer(s) deletes and then recreates a VM with the same name through Terraform and sometimes the “re-create” PUT gets blocked for a while as ARM hasn’t finished cleaning up the original deleted resources. This comes down to the reality that ARM considers a resource “deleted” not only when the delete job is done but also when a GET returns 404 and if that GET happens after the “re-create” PUT ARM will miss the 404.

### DIFF
--- a/azurerm/internal/services/compute/linux_virtual_machine_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_resource.go
@@ -3,11 +3,13 @@ package compute
 import (
 	"fmt"
 	"log"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
 	"github.com/hashicorp/go-azure-helpers/response"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -1084,5 +1086,39 @@ func resourceLinuxVirtualMachineDelete(d *schema.ResourceData, meta interface{})
 		log.Printf("[DEBUG] Skipping Deleting OS Disk from Linux Virtual Machine %q (Resource Group %q)..", id.Name, id.ResourceGroup)
 	}
 
+	// Need to add a get and a state wait to avoid bug in network API where the attached disk(s) are not actually deleted
+	// Service team indicated that we need to do a get after VM delete call returns to verify that the VM and all attached
+	// disks have actually been deleted.
+
+	log.Printf("[INFO] verifying Linux Virtual Machine %q has been deleted", id.Name)
+	virtualMachine, _ := client.Get(ctx, id.ResourceGroup, id.Name, "")
```
tombuildsstuff, line 1094: we should still check this error here, should network issues/an outage happen or something this could be an error

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -1111,5 +1113,39 @@ func resourceWindowsVirtualMachineDelete(d *schema.ResourceData, meta interface{
 		log.Printf("[DEBUG] Skipping Deleting OS Disk from Windows Virtual Machine %q (Resource Group %q)..", id.Name, id.ResourceGroup)
 	}
 
+	// Need to add a get and a state wait to avoid bug in network API where the attached disk(s) are not actually deleted
+	// Service team indicated that we need to do a get after VM delete call returns to verify that the VM and all attached
+	// disks have actually been deleted.
+
+	log.Printf("[INFO] verifying Windows Virtual Machine %q has been deleted", id.Name)
+	virtualMachine, _ := client.Get(ctx, id.ResourceGroup, id.Name, "")
```
tombuildsstuff, line 1121: we should still check this error here, should network issues/an outage happen or something this could be an error

tombuildsstuff (approved):
a couple of minor points but this otherwise LGTM 👍

azurerm/internal/services/compute/linux_virtual_machine_resource.go
WodansSon: Fixed.

azurerm/internal/services/compute/windows_virtual_machine_resource.go
WodansSon: Fixed.

WodansSon:
Test failures are not related to my change:
image

PR #8387, opened 2020-09-08
fix: change of azure api response

luc1f4:
Azure Response is using lower case instead of camel case, which cause an empty output.


tombuildsstuff:
hey @luc1f4 

Thanks for this PR :)

Taking a look through whilst this'd fix the issue for new deployments - unfortunately this'd break this for existing deployments - and as such this'd be a breaking change. Since #8374 is fixing a similar bug (in a backwards compatible way) - whilst I'd like to thank you for this contribution I hope you don't mind but I'm going to close this PR in favour of that PR which includes these changes made in a backwards-compatible manner (at the same time as the other change).

Thanks!

PR #8388, opened 2020-09-08
azurerm_eventhub_consumer_group - allow the name property to be set to $Default 

### LINKED ISSUE #7720: azurerm_eventhub_consumer_group doesn't support '$Default' consumer group name
### Community Note

+ Terraform v0.12.28
+ provider.azurerm v2.1.0

### Affected Resource(s)

* `azurerm_eventhub_consumer_group`

```hcl
resource "azurerm_eventhub_consumer_group" "example" {
  name                = "$Default"
  namespace_name      = azurerm_eventhub_namespace.example.name
  eventhub_name       = azurerm_eventhub.example.name
  user_metadata       = "some-meta-data"
}
```

### Expected Behavior

Should have accepted the name `$Default` as this is the name that the azure portal assigns by default if one is not specified, when using the `Basic` Tier Event Hub Namespace.

### Actual Behavior

`Error: invalid value for name (The consumer group name can contain only letters, numbers, periods (.), hyphens (-),and underscores (_), up to 50 characters, and it must begin and end with a letter or number.)`

### Steps to Reproduce

As per example above

1. `terraform apply`

### References

* #0000

favoretti:
Fixes #7720

### DIFF
--- a/azurerm/internal/services/eventhub/eventhub_consumer_group_data_source.go
+++ b/azurerm/internal/services/eventhub/eventhub_consumer_group_data_source.go
@@ -22,12 +22,12 @@ func dataSourceEventHubConsumerGroup() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"name": {
-				Type:         schema.TypeString,
-				Required:     true,
+				Type:     schema.TypeString,
+				Required: true,
 				ValidateFunc: validation.Any(
 					azure.ValidateEventHubConsumerName(),
 					validation.StringInSlice([]string{"$Default"}, false),
-					),
+				),
 			},
 
 			"namespace_name": {
--- a/azurerm/internal/services/eventhub/eventhub_consumer_group_data_source.go
+++ b/azurerm/internal/services/eventhub/eventhub_consumer_group_data_source.go
@@ -23,7 +23,7 @@ func dataSourceEventHubConsumerGroup() *schema.Resource {
 			"name": {
 				Type:         schema.TypeString,
 				Required:     true,
-				ValidateFunc: azure.ValidateEventHubConsumerName(),
+				ValidateFunc: azure.ValidateEventHubConsumerDataSourceName(),
 			},
 
 			"namespace_name": {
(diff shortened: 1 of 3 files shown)


favoretti:
@jackofallops would something like this work? Had to split validation for datasource and resource, to make sure people won't mangle `$Default` in resource definition.

azurerm/internal/services/eventhub/eventhub_consumer_group_data_source.go
```diff
@@ -23,7 +23,7 @@ func dataSourceEventHubConsumerGroup() *schema.Resource {
 			"name": {
 				Type:         schema.TypeString,
 				Required:     true,
-				ValidateFunc: azure.ValidateEventHubConsumerName(),
+				ValidateFunc: azure.ValidateEventHubConsumerDataSourceName(),
```
jackofallops, line 26: Since we're catering for one specific string case, it might be better to do:
```suggestion
				ValidateFunc: validation.Any(
					azure.ValidateEventHubConsumerName(),
					validation.StringInSlice([]string{"$Default"},
					)
```
Or modify your new validation function to do similar?

jackofallops (changes requested):
Hi @favoretti 
Thanks for this PR, I've had a quick look over and left one comment to look at initially since the new validation method regex would inadvertently allow invalid strings and not match `$Default` as intended.

Thanks!

azurerm/internal/services/eventhub/eventhub_consumer_group_data_source.go
favoretti: Ah, didn't know it existed. This is indeed much better. I'll commit your suggestion and remove the rest.

neil-yechenwei:
> Address #7720

Please update the comment from "Address #7720" to "fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7720"

azurerm/internal/services/eventhub/eventhub_consumer_group_data_source.go
```diff
@@ -2,6 +2,7 @@ package eventhub
 
 import (
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
```
neil-yechenwei, line 5: Move this line to line 9

favoretti:
@neil-yechenwei @jackofallops Feedback addressed.

katbyte (approved):
Thanks @favoretti - LGTM now! 🚀

PR #8403, opened 2020-09-09
Allow for manually scaled eventhub clusters

favoretti:
In case of scaling eventhub clusters by logging support ticket to MSFT, the `sku_name` of the cluster will change from `Dedicated_1` to `Dedicated_<X>`, where `X` is number of CUs allocated to the cluster. In current mode, TF will try to revert the change, forcing users to use `ignore_changes`. This PR should alleviate that need.

### DIFF
--- a/azurerm/internal/services/eventhub/eventhub_cluster_resource.go
+++ b/azurerm/internal/services/eventhub/eventhub_cluster_resource.go
@@ -3,6 +3,7 @@ package eventhub
 import (
 	"fmt"
 	"log"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jackofallops (commented):
Hi @favoretti 
Thanks for this PR. Sadly since the only valid value for this in the API is `Dedicated_1` this would effectively break the validation for the resource, as anything else is rejected and would cause an error in `apply`.  `ignore_changes` for that property should be used if a cluster is scaled via support request.

I'm going to close this for now, if/when Azure open up the API to accept other values we can revisit this and adjust appropriately. 

Thanks

favoretti:
@jackofallops hmm, I'm not sure I agree here. Here API splits the value to "Dedicated" as sku_name and capacity integer: https://docs.microsoft.com/en-us/rest/api/eventhub/2018-01-01-preview/clusters/createorupdate#clustersku

In SDK for go it's the same:
https://github.com/Azure/azure-sdk-for-go/blob/master/services/preview/eventhub/mgmt/2018-01-01-preview/eventhub/clusters.go#L68

Also in SDK I see no validation for capacity limited to 1. What am I seeing wrong?

favoretti:
Ah, I see, API actually actively refuses anything but 1... Sad story.

PR #8411, opened 2020-09-10
Update azure-sdk-for-go to v46.1.0

ArcturusZhang:


WodansSon (approved):
@ArcturusZhang Thanks for the PR! LGTM! 🚀

azurerm/internal/services/apimanagement/api_management_user_resource.go
```diff
@@ -145,7 +145,7 @@ func resourceArmApiManagementUserCreateUpdate(d *schema.ResourceData, meta inter
 		properties.UserCreateParameterProperties.State = apimanagement.UserState(state)
 	}
 
-	if _, err := client.CreateOrUpdate(ctx, resourceGroup, serviceName, userId, properties, ""); err != nil {
+	if _, err := client.CreateOrUpdate(ctx, resourceGroup, serviceName, userId, properties, nil, ""); err != nil {
```
tombuildsstuff, line 148: what's this nil value here? as per the last PR this is expecting where this resource is being created/updated - so sending the Old/New Portal is misleading - presumably we should be sending `terraform` here?

azurerm/internal/services/apimanagement/api_management_subscription_resource.go
```diff
@@ -150,7 +150,7 @@ func resourceArmApiManagementSubscriptionCreateUpdate(d *schema.ResourceData, me
 	}
 
 	sendEmail := utils.Bool(false)
-	_, err := client.CreateOrUpdate(ctx, resourceGroup, serviceName, subscriptionId, params, sendEmail, "")
+	_, err := client.CreateOrUpdate(ctx, resourceGroup, serviceName, subscriptionId, params, sendEmail, "", "")
```
tombuildsstuff, line 153: what's this empty value here? as per the last PR this is expecting where this resource is being created from, so sending the Old/New Portal is misleading - presumably we should be sending `terraform` here?

tombuildsstuff (changes requested):
hi @ArcturusZhang 

Thanks for this PR - I've taken a look through and it appears there's a couple of changes necessary here to not send a misleading value here?

Thanks!

azurerm/internal/services/apimanagement/api_management_user_resource.go
ArcturusZhang: This new parameter is `notify` - I will try to pass `terraform` to `appType` in other two methods and see if it works

azurerm/internal/services/apimanagement/api_management_subscription_resource.go
ArcturusZhang: I will have a try

WodansSon: I don't believe passing `terraform` is an option, the documentation says: `Default is old publisher portal. Possible values include: 'developerPortal'`

```
UserCreateParameters.AppType Property

Namespace: Microsoft.Azure.Management.ApiManagement.Models
Assembly: Microsoft.Azure.Management.ApiManagement.dll
Package: Microsoft.Azure.Management.ApiManagement v6.0.0-preview

Gets or sets determines the type of application which send the create user request. Default is old publisher portal. Possible values include: 'developerPortal'
```

tombuildsstuff: Yeah that's where I was going with this - it's an implementation detail, passing old or new portal is misleading (since we're not provisioning from either of those, presumably this is used in emails that get sent?) - so the API either needs to support other or this field likely not configurable?

It's not necessarily a blocker, but it's misleading when provisioning resources which would cause confusion?

WodansSon: Totally agree, can we merge this to unblock all the PR's and I can follow up with the service team on Redmond time Monday?

tombuildsstuff: @ArcturusZhang how did you get on with testing this out, did the API accept this value?

WodansSon: @tombuildsstuff @ArcturusZhang I have heard back from the service team:

>From the [release notes](https://azure.microsoft.com/en-us/updates/azure-api-management-update-july-2020/):
>
The PUT Subscription API call now accepts the `appType` query parameter in API versions 2018-06-01-preview or later. The value of the query parameter controls destination of the link included in the subscription confirmation email sent to a user - portal (default) results in a link pointing to the `legacy developer portal`, while `developerPortal` results in a link pointing to the new developer portal.
>
>1.	I suggest that the default in Terraform is `developerPortal` (pointing to new portal)...
>2.	There are no plans to make it a free-form string and there shouldn’t be `other` because we have only two portals – `legacy developer portal` and `new developer portal`. 

So per the service team we should us `developerPortal` instead of the default `portal`, seems rather odd to me that they would make the `legacy portal` the default instead of the `developerPortal`, but I digress... 🙃

tombuildsstuff: 🤷 seems fine for now

azurerm/internal/services/apimanagement/api_management_user_resource.go
tombuildsstuff: @ArcturusZhang can we set this to true/false (matching the default value used in the other resources) here?

tombuildsstuff (approved):
hi @ArcturusZhang 

Taking a look through here, if we can fix up the one comment (to pass the same notify value as the other resources) - and the tests look good then this otherwise seems fine 👍 

Thanks!

azurerm/internal/services/apimanagement/api_management_user_resource.go
ArcturusZhang: No problem. On my way. BTW do we need to revert those changes on the apimanagement identity?

ArcturusZhang:
Maybe we should also remove the `None` from `identity.0.type` candidate list and also change `type` from `Optional` to `Required` to align the document. 
I initially worked around this in a different approach since I assume removing an option / changing an attribute from optional to required might be considered as a breaking change.

PR #8420, opened 2020-09-10
Fix an unspecified parameter to be Unmarshal

416e64726579:
When a deployment uses beta feature support of `extensions` for VMSS there is a bug might occur if no `settings` or `protected_settings` is specified in the terraform module. This results in an Unmarshal error as there is no check for the presence of a key in the `extensionRaw` map. Thus, when not present, the map returns `""`

To test the behavior, it is enough to create a VMSS with only the `settings` block argument, for example:
```
resource "azurerm_linux_virtual_machine_scale_set" "app" {

...omitted

  extension {
      name                 = "extension-app-prod-001"
      publisher            = "Microsoft.Azure.Extensions"
      type                 = "CustomScript"
      type_handler_version = "2.0"

      settings = <<SETTINGS
        {
          "commandToExecute": "echo hi"
        }
SETTINGS
  }
}
```

### DIFF
--- a/azurerm/internal/services/compute/tests/virtual_machine_scale_set_resource_test.go
+++ b/azurerm/internal/services/compute/tests/virtual_machine_scale_set_resource_test.go
@@ -847,6 +847,46 @@ func TestAccAzureRMVirtualMachineScaleSet_multipleExtensions_provision_after_ext
 	})
 }
 
+func TestAccAzureRMVirtualMachineScaleSet_extension_with_settings(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_virtual_machine_scale_set", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMVirtualMachineScaleSetDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMVirtualMachineScaleSetExtensionTemplateOnlySettings(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVirtualMachineScaleSetExists(data.ResourceName),
+					testCheckAzureRMVirtualMachineScaleSetExtension(data.ResourceName),
+				),
+			},
+			data.ImportStep("os_profile.0.admin_password"),
+		},
+	})
+}
+
+func TestAccAzureRMVirtualMachineScaleSet_extension_with_protected_settings(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_virtual_machine_scale_set", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMVirtualMachineScaleSetDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMVirtualMachineScaleSetExtensionTemplateOnlyProtectedSettings(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVirtualMachineScaleSetExists(data.ResourceName),
+					testCheckAzureRMVirtualMachineScaleSetExtension(data.ResourceName),
+				),
+			},
+			data.ImportStep("os_profile.0.admin_password"),
+		},
+	})
+}
+
 func TestAccAzureRMVirtualMachineScaleSet_osDiskTypeConflict(t *testing.T) {
 	data := acceptance.BuildTestData(t, "azurerm_virtual_machine_scale_set", "test")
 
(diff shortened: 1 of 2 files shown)


WodansSon (changes requested):
@416e64726579, thanks for the PR! This mostly LGTM, but could we add two or three test cases for these conditions?

416e64726579:
> @416e64726579, thanks for the PR! This mostly LGTM, but could we add two or three test cases for these conditions?

Ok, will do them shortly.

416e64726579:
I've copied a couple of tests just without specifying each parameter.

jackofallops:
Hi @416e64726579 - Do you have the output from the bug in action?  This wasn't caught by the existing tests that include an extension with only `protected_settings` defined and I want to make sure we're getting the root and addressing gaps in the existing tests.
Thanks!

416e64726579:
I'm sorry guys, I mixed up everything. The talk was about `azurerm_linux_virtual_machine_scale_set` but I added tests for `azurerm_virtual_machine_scale_set`. 
When I use this inside `azurerm_linux_virtual_machine_scale_set`
```
  extension {
      name                 = "extension-test"
      publisher            = "Microsoft.Azure.Extensions"
      type                 = "CustomScript"
      type_handler_version = "2.0"

      settings = <<SETTINGS
        {
          "commandToExecute": "echo $HOSTNAME"
        }
SETTINGS

  }
```
I get 
```
2020/09/14 12:40:39 [DEBUG] azurerm_linux_virtual_machine_scale_set.test: apply errored, but we're indicating that via the Error pointer rather than returning it: failed to parse JSON from `settings`: unexpected end of JSON input
2020/09/14 12:40:39 [ERROR] <root>: eval: *terraform.EvalApplyPost, err: failed to parse JSON from `settings`: unexpected end of JSON input
2020/09/14 12:40:39 [ERROR] <root>: eval: *terraform.EvalSequence, err: failed to parse JSON from `settings`: unexpected end of JSON input

Error: failed to parse JSON from `settings`: unexpected end of JSON input

  on main.tf line 297, in resource "azurerm_linux_virtual_machine_scale_set" "test":
 297: resource "azurerm_linux_virtual_machine_scale_set" "test" {
```

416e64726579:
I might need to dig up a little more as I also thought that the test `TestAccAzureRMLinuxVirtualMachineScaleSet_otherVmExtensionsUpdate` passed ok, and this includes solely `settings` attribute.

416e64726579:
Perhaps, need to add a test to empty string as Terraform initialise unspecified attributes with the default values which for these parameters is `""`.
Somehow like:
```
		if val, ok := extensionRaw["settings"]; ok && val != "" {
			settings, err := structure.ExpandJsonFromString(val.(string))
			if err != nil {
				return nil, fmt.Errorf("failed to parse JSON from `settings`: %+v", err)
			}
			extensionProps.Settings = settings
		}

		if val, ok := extensionRaw["protected_settings"]; ok && val != "" {
			protectedSettings, err := structure.ExpandJsonFromString(val.(string))
			if err != nil {
				return nil, fmt.Errorf("failed to parse JSON from `protected_settings`: %+v", err)
			}
			extensionProps.ProtectedSettings = protectedSettings
		}
```

jackofallops:
Hi @416e64726579 
Your code and error message suggest that the problem is actually with the values you're using, rather than the one's you're missing?
> 2020/09/14 12:40:39 [DEBUG] azurerm_linux_virtual_machine_scale_set.test: apply errored, but we're indicating that via the Error pointer rather than returning it: failed to parse JSON from `settings`: unexpected end of JSON input
Can you try using the built-in terraform function for creating JSON as described in the example and docs, rather than inline, and see if you have the same issue?
e.g.:
```
extension {
      name                 = "extension-test"
      publisher            = "Microsoft.Azure.Extensions"
      type                 = "CustomScript"
      type_handler_version = "2.0"

      settings = jsonencode({
          "commandToExecute" = "echo $HOSTNAME"
      })

  }
    
```

Thanks

416e64726579:
Hi @jackofallops 
Thank you for your response. Yes, the problem actually is with the values I convey but the `jsonencode` function does conversion in this case substituting `"="` to `":"`. 
Yet, I did try the suggested code snippet and it fails too.
Take a look at this https://play.golang.org/p/ZzWSil7oXyX
I guess this is how the procedure looks like when there is only one parameter defined.

Thanks

jackofallops:
Thanks @416e64726579 - I'll need to dig into why the acceptance tests didn't catch this, as we do have test configs in there that appear to cover the condition you're seeing. (for [example](https://github.com/terraform-providers/terraform-provider-azurerm/blob/baa360cfc1c5bfb58e0083de1a66dc5a1b7ddb8f/azurerm/internal/services/compute/tests/linux_virtual_machine_scale_set_other_resource_test.go#L2313))

tombuildsstuff:
hey @416e64726579 

Apologies for the delayed re-review on our side - I've taken a look into rebasing this and where this is ~3000 commits behind that's challenging due to the moving of the test files. Having taken a look through the diff here, the tests which have been added apply to the older `azurerm_virtual_machine_scale_set` resource (which doesn't use this codepath) rather than the newer `azurerm_linux_virtual_machine_scale_set`/`azurerm_windows_virtual_machine_scale_set` - as such whilst they do test this condition they're testing it for the wrong resource unfortunately.

Since rebasing this given the test file move is challenging, and these need to be ignored anyway - I hope you don't mind but I'm going to cherry-pick the first commit here into a new PR, meaning we can ship this. During the time since this was first created and today, there's a test covering these use-cases in the form of the "extension only settings" tests, which I've duplicated from the Linux to the Windows Virtual Machine Scale Set resource.

As such I hope you don't mind but I'm going to close this PR in favour of #10351, which includes your commit/contributions - and should allow us to ship this fix in today's release :)

Thanks!

PR #8434, opened 2020-09-11
`azurerm_synapse_sql_pool` - leverage newer version mssql

njuCZ:
follow up of PR #8095

### DIFF
--- a/azurerm/internal/services/mssql/parse/mssql.go
+++ b/azurerm/internal/services/mssql/parse/mssql.go
@@ -29,6 +29,18 @@ type MsSqlDatabaseExtendedAuditingPolicyId struct {
 	ResourceGroup string
 }
 
+func NewMsSqlDatabaseID(resourceGroup, msSqlServer, name string) MsSqlDatabaseId {
+	return MsSqlDatabaseId{
+		ResourceGroup: resourceGroup,
+		MsSqlServer:   msSqlServer,
+		Name:          name,
+	}
+}
+
+func (id MsSqlDatabaseId) ID(subscriptionId string) string {
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Sql/servers/%s/databases/%s", subscriptionId, id.ResourceGroup, id.MsSqlServer, id.Name)
+}
+
 func MsSqlDatabaseID(input string) (*MsSqlDatabaseId, error) {
 	id, err := azure.ParseAzureResourceID(input)
 	if err != nil {
--- a/azurerm/internal/services/synapse/synapse_sql_pool_resource.go
+++ b/azurerm/internal/services/synapse/synapse_sql_pool_resource.go
@@ -13,9 +13,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
-	postgresValidate "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/postgres/validate"
-	sqlParse "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/sql/parse"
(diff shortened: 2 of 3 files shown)


azurerm/internal/services/mssql/parse/mssql.go
```diff
@@ -29,6 +29,18 @@ type MsSqlDatabaseExtendedAuditingPolicyId struct {
 	ResourceGroup string
 }
 
+func NewMsSqlDatabaseID(resourceGroup, msSqlServer, name string) MsSqlDatabaseId {
```
WodansSon, line 32: Should this be a more descriptive name?

```suggestion
func SynapseSqlPoolDatabaseID(resourceGroup, msSqlServer, name string) MsSqlDatabaseId {
```

azurerm/internal/services/mssql/parse/mssql.go
```diff
@@ -29,6 +29,18 @@ type MsSqlDatabaseExtendedAuditingPolicyId struct {
...
+	return MsSqlDatabaseId{
+		ResourceGroup: resourceGroup,
+		MsSqlServer:   msSqlServer,
+		Name:          name,
+	}
```
WodansSon, line 37: Shouldn't all of these values be checked and validated and return an error if they are not valid?

azurerm/internal/services/mssql/parse/mssql.go
```diff
@@ -29,6 +29,18 @@ type MsSqlDatabaseExtendedAuditingPolicyId struct {
...
+}
+
+func (id MsSqlDatabaseId) ID(subscriptionId string) string {
```
WodansSon, line 40: Same here, shouldn't the subscriptionId input be validated and return an error if it is null, empty, or malformed?

azurerm/internal/services/mssql/parse/mssql.go
WodansSon, line 37: Same here, shouldn't we expose a new `struct` with a more descriptive name?
```suggestion
	return SynapseSqlPollDatabaseId{
		ResourceGroup: resourceGroup,
		SynapseSqlPoolServer:   synapseSqlPoolServer,
		Name:          name,
	}
```

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -113,7 +112,7 @@ func resourceArmSynapseSqlPool() *schema.Resource {
 				Optional:     true,
 				Computed:     true,
 				ForceNew:     true,
-				ValidateFunc: postgresValidate.PostgresDatabaseCollation,
+				ValidateFunc: mssqlValidate.MsSqlDBCollation(),
```
WodansSon, line 115: Is this collation specific to only synapse sql pool's or is it generic for all msSql databases? If it is for only synapse I would update the name to be reflective of that, something like `synapseSqlPoolDBCollation`

azurerm/internal/services/synapse/synapse_sql_pool_resource.go
```diff
@@ -123,7 +122,7 @@ func resourceArmSynapseSqlPool() *schema.Resource {
 				ConflictsWith: []string{"restore"},
 				ValidateFunc: validation.Any(
 					validate.SqlPoolID,
-					sqlValidate.SqlDatabaseID,
+					mssqlValidate.MsSqlDatabaseID,
```
WodansSon, line 125: Same here, is this for all Ms Sql Databases or is it just for synapse sql pool's? If it's just for pool's I would change the name here as well.

WodansSon (changes requested):

njuCZ:
hi @WodansSon, this is a rule that Tom commented: https://github.com/terraform-providers/terraform-provider-azurerm/pull/8095#discussion_r473871779 . This PR is to solve @jackofallops comments in last PR. mssql is a newer version than sql, so this PR change to reference mssql. The source_database_id format is either synapse_sql_pool id or sql database id. But the backend service only supports sql database id format, so need to convert the format

WodansSon:
image

WodansSon (approved):

PR #8444, opened 2020-09-12
correct the doco - the Self-hosted Integration Runtime resource is re…

a30001539:
…-created, not the containing Data Factory.


katbyte (approved):
Thanks for this correction @a30001539 - LGTM 📜

PR #8451, opened 2020-09-14
Update main.tf

cecilozaur:
should be base64encode() not base64_encode() at least in version 1.13 of terraform


katbyte (approved):
Thanks @cecilozaur, LGTM 👍

PR #8461, opened 2020-09-14
azurerm_cosmosdb_sql_container - support for `indexing_policy`

marc-sensenich:
Allows for the setting of indexing policies on Cosmos DB SQL Containers. Completes the work from https://github.com/terraform-providers/terraform-provider-azurerm/pull/5546.

Test status as of https://github.com/terraform-providers/terraform-provider-azurerm/pull/8461/commits/e06ca000a8fbd115b3e057e57956b7fb5fee1f16

```
--- PASS: TestAccAzureRMCosmosDbSqlContainer_basic (1600.44s)
--- PASS: TestAccAzureRMCosmosDbSqlContainer_complete (1713.66s)
--- PASS: TestAccAzureRMCosmosDbSqlContainer_update (1775.82s)
--- PASS: TestAccAzureRMCosmosDbSqlContainer_indexing_policy (1821.13s)
--- PASS: TestAccAzureRMCosmosDbSqlContainer_autoscale (1940.55s)
```

### DIFF
--- a/azurerm/internal/services/cosmos/common/indexing_policy.go
+++ b/azurerm/internal/services/cosmos/common/indexing_policy.go
@@ -47,12 +47,6 @@ func ExpandAzureRmCosmosDbIndexingPolicy(d *schema.ResourceData) *documentdb.Ind
 	policy := &documentdb.IndexingPolicy{}
 
 	if len(i) == 0 || i[0] == nil {
-		policy.IndexingMode = documentdb.Consistent
-		policy.IncludedPaths = &[]documentdb.IncludedPath{
-			{Path: utils.String("/*")},
-		}
-		policy.ExcludedPaths = &[]documentdb.ExcludedPath{}
-
 		return policy
 	}
 
--- a/azurerm/internal/services/cosmos/common/indexing_policy.go
+++ b/azurerm/internal/services/cosmos/common/indexing_policy.go
@@ -1,0 +1,122 @@
+package common
+
+import (
+	"github.com/Azure/azure-sdk-for-go/services/cosmos-db/mgmt/2020-04-01/documentdb"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func expandAzureRmCosmosDBIndexingPolicyIncludedPaths(input []interface{}) *[]documentdb.IncludedPath {
+	if len(input) == 0 {
+		return nil
+	}
+
+	var includedPaths []documentdb.IncludedPath
+
+	for _, v := range input {
+		includedPath := v.(map[string]interface{})
+		path := documentdb.IncludedPath{
+			Path: utils.String(includedPath["path"].(string)),
+		}
+
+		includedPaths = append(includedPaths, path)
+	}
+
+	return &includedPaths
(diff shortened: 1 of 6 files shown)


azurerm/internal/services/cosmos/common/indexing_policy.go
```diff
@@ -0,0 +1,122 @@
+package common
```
marc-sensenich, line 1: Added these to common as the Gremlin Graph resource could migrate over to these in the future

azurerm/internal/services/cosmos/common/schema.go
```diff
@@ -37,3 +40,56 @@ func MongoCollectionAutoscaleSettingsSchema() *schema.Schema {
...
+					}, false),
+				},
+
+				"included_path": {
+					Type:     schema.TypeList,
+					Optional: true,
+					Computed: true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"path": {
```
marc-sensenich, line 70: I only did `path` here as the `indexes` field was accepted by the REST API, but did not return the `indexes` field in response.

azurerm/internal/services/cosmos/common/schema.go
```diff
@@ -37,3 +40,56 @@ func MongoCollectionAutoscaleSettingsSchema() *schema.Schema {
 
 	return autoscaleSettingsDatabaseSchema
 }
+
+func CosmosDbIndexingPolicySchema() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeList,
```
marc-sensenich, line 46: Is there a suggested pattern for validating `TypeList` that does cross field validation? In the indexing policy case there are the validations of

- `included_path` and `excluded_path` can't be set when the `indexing_mode` is `None`
- One of `included_path` or `excluded_path` must contain the path `/*`

favoretti:
It would be great if this would make it in ASAP. We just observed that any terraform changes to a container with custom indexing set  reset the indexing to default. :(

azurerm/internal/services/cosmos/common/schema.go
katbyte: The best solution is to validate it in create and return an error at that point if it fails

katbyte (changes requested):
Thanks for the PR @marc-sensenich - overall looks good to me! if we can add that validation to the create/update function this should be good to merge 🙁

marc-sensenich:
@katbyte validation added in https://github.com/terraform-providers/terraform-provider-azurerm/pull/8461/commits/e06ca000a8fbd115b3e057e57956b7fb5fee1f16

katbyte (changes requested):
thanks @marc-sensenich  - running the tests for the latest changes however and i'm getting:
```
Test Failed

------- Stdout: -------
=== RUN   TestValidateAzureRmCosmosDbIndexingPolicy
    TestValidateAzureRmCosmosDbIndexingPolicy: indexing_policy_test.go:198: Expected an error but didn't get one for "included_path and excluded_path with /*"
--- FAIL: TestValidateAzureRmCosmosDbIndexingPolicy (0.00s)
```

katbyte (approved):
NM @marc-sensenich, ran it again and it passed! thanks for the changes 🙂

PR #8467, opened 2020-09-15
make sku_name suppress case difference - azurerm_key_vault

### LINKED ISSUE #8457: Incorrect casing for SKU name values when importing azurerm_key_vault
Imported a keyvault and issued terraform plan and got:

`Error: expected sku_name to be one of [standard premium], got Standard`

Changed to lowercase standard:

`      ~ sku_name                        = "Standard" -> "standard"`

Should it really trigger a change? I cannot use Standard and standard i triggering a change to the resource. It should import it with lower case or allow me to use upper case. 

According to thread below upper case does not work so it should import with lower case i guess.
https://github.com/terraform-providers/terraform-provider-azurerm/issues/2780

`terraform -v
Terraform v0.13.2
+ provider registry.terraform.io/hashicorp/azurerm v2.26.0`

njuCZ:
fix #8457

now, the portal's default behaviour for "sku" is "Standard". after terraform import there will be diff because only "standard" and "premium" is accepted.

### DIFF
--- a/azurerm/internal/services/keyvault/key_vault_resource.go
+++ b/azurerm/internal/services/keyvault/key_vault_resource.go
@@ -7,9 +7,6 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/keyvault/validate"
-	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network"
-
 	"github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2019-09-01/keyvault"
 	"github.com/hashicorp/go-azure-helpers/response"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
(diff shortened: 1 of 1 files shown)


azurerm/internal/services/keyvault/key_vault_resource.go
```diff
@@ -73,6 +73,7 @@ func resourceArmKeyVault() *schema.Resource {
 					string(keyvault.Standard),
 					string(keyvault.Premium),
 				}, false),
+				DiffSuppressFunc: suppress.CaseDifference,
```
katbyte, line 76: Could we have a link to an bug/issue here that describes why we need to do this?

katbyte (changes requested):

njuCZ:
@katbyte I have added a link to the rest-api-specs issue, could you please have a look again?

njuCZ:
it seems there is some problem with the CI, I will close and reopen it

katbyte (changes requested):
thanks, Instead of just making it fully case insensitive could we just allow `standard` and `premium` in addition to the camel cased ones?

njuCZ:
@katbyte according to service team's reply https://github.com/Azure/azure-rest-api-specs/issues/10943#issuecomment-702984955  sku_name is not case-sensitive, I am afraid that the portal will change again. I think it's better to make it fully case insensitive

scp-mb:
Is there any estimates on when this could be merged and released? We're looking at using Pulumi to manage some key vaults which under the hood uses the terraform azurerm provider, but importing resources will fail due to this case mismatch.

tombuildsstuff:
Closing this in favour of #8983 which instead rewrites the casing to ensure it's consistent with what the SDK defines, meaning the configuration is the source of truth here.

PR #8487, opened 2020-09-16
`data.azurerm_app_service_environment` - Expose vip information of an app service environment 

favoretti:
In order to be able to terraform private DNS entries for `app_service_environment` we need to be able to get to its IP addresses. This patch should expose those properties.


mbfrahry (approved):
LGTM!

Issue #6146, opened 2020-03-17
Support for private_ip_address export from azurerm_private_endpoint resource


johnwildes:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Export of (Computed) value, Private IP address for `azurerm_private_endpoint` resource.

Presently if you set up an `azurerm_private_endpoint` resource, you cannot get the IP address of that resource to feed into something else.  For example, I'd like to update an `A` record within the private dns zone I've configured in azure once the resource is configured.  At this point I must manually create that entry because the `private_ip_address` is not exported from the private endpoint resource.

### New or Affected Resource(s)

`azurerm_private_endpoint`

### Potential Terraform Configuration

```hcl
### KeyVault with Private Link

### Read in the current user context for assignment of access policies
data "azurerm_client_config" "current" {}

resource "azurerm_key_vault" "sandbox" {
  name                        = "${local.prefix}KV"
  location                    = azurerm_resource_group.sandbox.location
  resource_group_name         = azurerm_resource_group.sandbox.name
  enabled_for_disk_encryption = true
  tenant_id                   = data.azurerm_client_config.current.tenant_id
  soft_delete_enabled         = true
  purge_protection_enabled    = false

  sku_name = "standard"

  access_policy {
    tenant_id = data.azurerm_client_config.current.tenant_id
    object_id = data.azurerm_client_config.current.object_id

    key_permissions = [
      "get",
    ]

    secret_permissions = [
      "get",
    ]

    storage_permissions = [
      "get",
    ]
  }

  network_acls {
    default_action = "Deny"
    bypass         = "AzureServices"
  }

  tags = local.common_tags
}


resource "azurerm_private_endpoint" "sandbox_kv" {
  name                = azurerm_key_vault.sandbox.name
  location            = azurerm_resource_group.sandbox.location
  resource_group_name = azurerm_resource_group.sandbox.name
  subnet_id           = azurerm_subnet.sandbox["PrivateLink"].id

  private_service_connection {
    name                           = azurerm_key_vault.sandbox.name
    private_connection_resource_id = azurerm_key_vault.sandbox.id
[...]

PR #8495, opened 2020-09-16
Doc Fix: `azurerm_private_endpoint` Updating documentation for `private_ip_address`

raykao:
Based on the syntax for getting the ```private_ip_address``` for a private endpoint from issue #6146 

Also includes the syntax for ACR which creates multiple private ip address for updating/creating Private DNS Zone A Records


mbfrahry (approved):
Hey @raykao, this PR looks good to me. I saw that we already had `private_ip_address` documented so I went ahead and swapped it down. Thanks for taking the time to write this out!

PR #8497, opened 2020-09-17
Update `azurerm_template_deployment` - Support import

ArcturusZhang:
Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7268

### DIFF
--- a/azurerm/internal/services/resource/parse/template_deployment.go
+++ b/azurerm/internal/services/resource/parse/template_deployment.go
@@ -1,0 +1,48 @@
+package parse
+
+import (
+	"fmt"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+)
+
+type TemplateDeploymentId struct {
+	ResourceGroup string
+	Name string
+}
+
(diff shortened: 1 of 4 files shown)


tombuildsstuff:
hi @ArcturusZhang 

Thanks for this PR.

Taking a look through here, whilst this PR does add support for import to this resource - since the template content isn't being returned from the Azure API here, unfortunately this'd lead to a diff here (`null -> 'template body'`).

Whilst we could look to set this field into the state to allow us to track this - after spending some time with this (and #7506) it likely makes more sense to overhaul these resources to support the updated behaviours now present in this API.

In particular, since the API now returns a list of resources provisioned by this ARM Template - we're able to attempt to clean these up during deletion of the ARM Template Deployment (which can be toggled on/off using the features block).

As we're looking to take a different direction here - whilst I'd like to thank you for this contribution I'm going to close this in favour of #8672 which introduces a replacement for this resource, and the existing `azurerm_template_deployment` resource will be deprecated in version 3.0 of the Azure Provider.

Thanks!

Issue #7540, opened 2020-06-30
azurerm_automation_module creation intermittently fails to create module


ZachTB123:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform version: 0.12.21
AzureRM version: 2.13.0

### Affected Resource(s)

* `azurerm_automation_module`

### Terraform Configuration Files

```hcl
# Resource Group
resource azurerm_resource_group test {
  name     = "test-rg"
  location = "centralus"
}

# Automation account
resource azurerm_automation_account test {
  name                = "test-automation"
  location            = "centralus"
  resource_group_name = azurerm_resource_group.test.name

  sku_name = "Basic"
}

# Powershell module needed for authenticating
resource azurerm_automation_module az_accounts {
  name                    = "Az.Accounts"
  resource_group_name     = azurerm_resource_group.test.name
  automation_account_name = azurerm_automation_account.test.name

  module_link {
    uri = "https://www.powershellgallery.com/api/v2/package/Az.Accounts/1.7.2"
  }
}

resource azurerm_automation_module az_resources {
  name                    = "Az.Resources"
  resource_group_name     = azurerm_resource_group.test.name
  automation_account_name = azurerm_automation_account.test.name

  module_link {
    uri = "https://www.powershellgallery.com/api/v2/package/Az.Resources/1.11.0"
  }

  depends_on = [azurerm_automation_module.az_accounts]
}

# More modules below...
```

### Expected Behavior

The module was created successfully in the automation account.

### Actual Behavior

The first `terraform apply` failed to create the module. It produces the following error message:

"Error waiting for Module "Az.Resources" (Automation Account "test-automation" / Resource Group "test-rg") to finish provisioning: unexpected state 'Failed', wanted target 'Succeeded'. last error: %!s(<nil>)'

Subsequent `terraform apply`'s fail because it says the module already exists and must be imported into state.

When viewing the module in the portal I see the following error:

[...]

PR #8498, opened 2020-09-17
azurerm_automation_module - expose error message 

njuCZ:
related issue: #7540

1. expose the error message, the original error message could not know the detail reason:
```
Error waiting for Module "Az.Resources" (Automation Account "test-automation" / Resource Group "test-rg") to finish provisioning: unexpected state 'Failed', wanted target 'Succeeded'. last error: %!s()'
```
2. delete the resource when waitforstate returned error

for this issue, I think the real problem is why it could succeed sometimes and failed sometimes. I have fired an issue in azure-rest-api-specs


azurerm/internal/services/automation/automation_module_resource.go
```diff
@@ -164,7 +168,13 @@ func resourceArmAutomationModuleCreateUpdate(d *schema.ResourceData, meta interf
...
 
 	if _, err := stateConf.WaitForState(); err != nil {
-		return fmt.Errorf("Error waiting for Module %q (Automation Account %q / Resource Group %q) to finish provisioning: %+v", name, accName, resGroup, err)
+		wrappedErr := fmt.Errorf("waiting for Module %q (Automation Account %q / Resource Group %q) to finish provisioning: %+v", name, accName, resGroup, err)
+		if resp, err := client.Delete(ctx, resGroup, accName, name); err != nil {
+			if !utils.ResponseWasNotFound(resp) {
+				wrappedErr = fmt.Errorf("%s\nattempting to delete failed automation account: %+v", wrappedErr, err)
+			}
+		}
+		return wrappedErr
```
katbyte, line 177: We shouldn't be randomly deleting things during a creation. This is an API bug, could we report it to them and have them clean up the failed create
```suggestion
		return fmt.Errorf("Error waiting for Module %q (Automation Account %q / Resource Group %q) to finish provisioning: %+v", name, accName, resGroup, err)
```

azurerm/internal/services/automation/automation_module_resource.go
```diff
@@ -1,6 +1,7 @@
 package automation
 
 import (
+	"errors"
```
katbyte, line 4:
```suggestion
```

katbyte (approved):
Hey @njuCZ - While i agree we should expose the actual error message, i don't think we should be deleting the broken resource, the service/API should be doing that after a failed create? As such i am going to remove that from this PR and merge.

Thanks

Issue #7677, opened 2020-07-10
API returned extra features I didn't specify in tfconfig for azurerm_signalr_service


neil-yechenwei:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.12.24
+ provider.azurerm (2.18.0)

### Affected Resource(s)

* `azurerm_signalr_service`

### Terraform Configuration Files

```hcl
provider "azurerm" {
  features {}
}

resource "azurerm_resource_group" "test" {
  name     = "acctestRG-ss-test01"
  location = "eastus2"
}

resource "azurerm_signalr_service" "test" {
  name                = "acctestSignalR-test01"
  location            = azurerm_resource_group.test.location
  resource_group_name = azurerm_resource_group.test.name

  sku {
    name     = "Free_F1"
    capacity = 1
  }

  features {
    flag  = "ServiceMode"
    value = "Serverless"
  }
}
```

### Panic Output

```
=== RUN   TestAccAzureRMSignalRService_serviceMode
=== PAUSE TestAccAzureRMSignalRService_serviceMode
=== CONT  TestAccAzureRMSignalRService_serviceMode
    TestAccAzureRMSignalRService_serviceMode: testing.go:684: Step 0 error: After applying this step, the plan was not empty:
        
        DIFF:
        
        UPDATE: azurerm_signalr_service.test
          cors.#:                      "1" => "1"
          cors.0.allowed_origins.#:    "1" => "1"
          cors.0.allowed_origins.0:    "*" => "*"
          features.#:                  "3" => "1"
          features.0.flag:             "ServiceMode" => "ServiceMode"
          features.0.value:            "Serverless" => "Serverless"
          features.1.flag:             "EnableConnectivityLogs" => ""
          features.1.value:            "False" => ""
          features.2.flag:             "EnableMessagingLogs" => ""
          features.2.value:            "False" => ""
          hostname:                    "xxx.service.signalr.net" => "xxx.service.signalr.net"
[...]

PR #8539, opened 2020-09-21
Add diff suppress for features property of azurerm_signalr_service

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7677

--- PASS: TestAccAzureRMSignalRService_serviceMode (236.91s)
--- PASS: TestAccAzureRMSignalRService_cors (236.91s)
--- PASS: TestAccAzureRMSignalRService_standardWithCap2 (294.84s)
--- PASS: TestAccAzureRMSignalRService_standard (294.84s)
--- PASS: TestAccAzureRMSignalRService_capacityUpdate (536.33s)
--- PASS: TestAccAzureRMSignalRService_skuUpdate (601.45s)
--- PASS: TestAccAzureRMSignalRService_skuAndCapacityUpdate (735.92s)
--- PASS: TestAccAzureRMSignalRService_requiresImport (464.32s)
--- PASS: TestAccAzureRMSignalRService_basic (296.97s)

### DIFF
--- a/azurerm/internal/services/signalr/resource_arm_signalr_service.go
+++ b/azurerm/internal/services/signalr/resource_arm_signalr_service.go
@@ -98,6 +98,24 @@ func resourceArmSignalRService() *schema.Resource {
 						},
 					},
 				},
+
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					if k == "features.#" {
+						if (old == "3" && new == "1") || (old == "3" && new == "2") {
+							return true
+						}
+					}
+
+					if old != "" && new == "" {
+						return true
+					}
+
+					if old == new {
+						return true
+					}
+
+					return false
+				},
 			},
 
 			"cors": {


tombuildsstuff:
hi @neil-yechenwei 

Thanks for this PR.

Whilst it's unfortunate the API is designed in such a way that it adds features a user hasn't specified; unfortunately using a diff suppress in this manner will cause other issues, since this is too generic and will other feature flag combinations - instead users will need to update their configuration to match the API.

I've opened #8541 which adds support for delta-updates to this resource - which should allow users to use the `ignore_changes` functionality to ignore changes to this field should they wish - but ultimately users will either need to add these fields to their Terraform Configurations or use the `ignore_changes` functionality to ignore changes to this field until the API supports removing these fields (if users don't wish to enable them).

Ultimately it'd be worth raising a bug on the API Specs repository about the API adding features which are unspecified - since features shouldn't be enabled unless users opt-into them - but this is likely the best we can do at this point in time.

As such whilst I'd like to thank you for this PR I'm going to close this in favour of #8541 which enabled fixing #7677 in a slightly different manner.

Thanks!

Issue #4846, opened 2019-11-09
azurerm_security_center_subscription_pricing does not enable pricing for all resource types


cananac:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Hi,

I'm using azurerm_security_center_subscription_pricing to enable Standard pricing for Azure Security Center. But this does not enable the Standard pricing tier for all the resource types under subscription. Security Center -> Pricing & Settings ->  Pricing Tier, the plan is Disabled for resources apart from virtual machines. 

The case can be reproduced as follows:
- Enable Standard pricing from portal for Security Center and enable plan only for Virtual Machines. Save.
- Then change pricing as Free and save.
- Then set Standard pricing using azurerm_security_center_subscription_pricing

It would also be great to enable/disable resource types individually.

Thanks.

cananac:
Hi,

Any updates on this issue?

jarmbruster74:
Looked into this a bit more.  This actually appears to be a limitation that applies to several clients, when I create a new pricing from the Azure command line I see the same behavior.  I'm unable to find an option at this time other than through the Azure portal to turn on all resources.

jarmbruster74:
From testing this page though it is supported in the underlying API with different pricingName values (e.g. KeyVaults, StorageAccounts, VirtualMachines, etc)

redlineby:
Hi,
Any news with this issue. We have the same problem :\

darren-johnson:
This would be handy for me also!

fireblade95402:
I have a customer with the same issue. It would be great if we could get an update on this?

PR #8549, opened 2020-09-21
Update `azurerm_security_center_subscription_pricing`

beandrad:
Adds `resource_type` to `azurerm_security_center_subscription_pricing` to allow setting pricing tier (Free or Standard) per resource type. At the moment only VirtualMachines are affected by this resource. 

The problem with the current code is that is using the v1 pricing client, which only allows specifying `default` as resource type, which seems to map to Virtual Machines resource type.

Fixes: #4846


Nepomuceno:
If you are changing a resource schema you need to change the schema version and do a schema update script. Instructions on how to do that can be found [here](https://www.terraform.io/docs/extend/resources/state-migration.html) I would also change the way that the `resource_type` is defined to be a map this way there will be no need to define multiple of this to define each toggle.

Nepomuceno:
:shipit: LGTM

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
```diff
@@ -44,17 +58,39 @@ func resourceArmSecurityCenterSubscriptionPricing() *schema.Resource {
 					string(security.Standard),
 				}, false),
 			},
+			"resource_type": {
+				Type:     schema.TypeString,
+				Required: true,
```
jackofallops, line 63: This will need to be `Optional` to avoid the breaking change.  Setting a default value here (of `default` as before, or `VirtualMachines` if that replaces it?) would be appropriate.

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
```diff
@@ -44,17 +58,39 @@ func resourceArmSecurityCenterSubscriptionPricing() *schema.Resource {
...
 
+func resourceArmSecurityCenterSubscriptionPricingParseId(id string) (string, string, error) {
+	parts := strings.SplitN(id, "/", 7)
+
+	if len(parts) != 7 || parts[2] == "" || parts[6] == "" {
+		return "", "", fmt.Errorf("unexpected format of ID (%s)", id)
+	}
+
+	return parts[2], parts[6], nil
+}
```
jackofallops, line 87: This should probably be reworked into a parse helper that enables the use of an ID type for this resource, there are many examples of this pattern in the provider.  (For [example](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/compute/parse/virtual_machine.go))

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
```diff
@@ -86,15 +124,17 @@ func resourceArmSecurityCenterSubscriptionPricingRead(d *schema.ResourceData, me
 	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
 	defer cancel()
 
-	resp, err := client.GetSubscriptionPricing(ctx, securityCenterSubscriptionPricingName)
+	resource_type := d.Get("resource_type").(string)
```
jackofallops, line 127: If the above ID type is used, this can be taken from the that ID, rather than the config to ensure we're reading from the API and not config.

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
```diff
@@ -86,15 +124,17 @@ func resourceArmSecurityCenterSubscriptionPricingRead(d *schema.ResourceData, me
...
 	if err != nil {
 		if utils.ResponseWasNotFound(resp.Response) {
-			log.Printf("[DEBUG] Security Center Subscription was not found: %v", err)
+			log.Printf("[DEBUG] %v Security Center Subscription was not found: %v", resource_type, err)
 			d.SetId("")
 			return nil
 		}
 
-		return fmt.Errorf("Error reading Security Center Subscription pricing: %+v", err)
+		return fmt.Errorf("Error reading %v Security Center Subscription pricing: %+v", resource_type, err)
```
jackofallops, line 137: minor change here for display
```suggestion
		return fmt.Errorf("Error reading %q Security Center Subscription pricing: %+v", resource_type, err)
```

jackofallops (changes requested):
Hi @beandrad 
Thanks for this PR, it's off to a good start. 

We try to avoid using multiple API versions for a service wherever possible. I've had a quick scan over the service and I think this should be possible here without too much difficulty. Since the changes you are targeting in the resource are pretty much required for the API bump could you update to just using v3.0?  I've put some other comments in below that will also need addressing.  We can help you out as needed, and if you're not already joined the community slack, there are instructions for joining in the project [readme](https://github.com/terraform-providers/terraform-provider-azurerm#terraform-provider-for-azure-resource-manager).

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
beandrad: `default` is not allowed in the new version of the API, `VirtualMachines` replaces it. I guess in this case, I can remove the `StateUpgraders`. The reasons I made it a required parameter are (1) in the rest API is required and it's nice to keep a consistent behaviour and (2) not having an explicit resource type can be misleading (when reviewing the code the reader may think that the pricing tier affects all the resources within the subscription).

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
```diff
@@ -25,7 +22,15 @@ func resourceArmSecurityCenterSubscriptionPricing() *schema.Resource {
 		Delete: resourceArmSecurityCenterSubscriptionPricingDelete,
 
 		Importer: &schema.ResourceImporter{
-			State: schema.ImportStatePassthrough,
+			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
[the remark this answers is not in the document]
beandrad: Unfortunately, `tf import` doesn't work if I don't set the resource type in the resource data schema in the func above.

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
beandrad: I'm not sure if I understand; something along the lines of:

```
id := d.GetId()
scPricing := parseId(id)
resp, err := client.Get(ctx, scPricing.resourceType)
```
?

jackofallops: something like:

```
id, err := parse.SecurityPricingID(d.Id())
resp, err := client.Get(ctx, id.resourceType)
```

There's examples in most of the resources of this pattern.  Does that help?

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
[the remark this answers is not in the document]
beandrad: if you don't define the state function, the `Read` func

> Hi @beandrad
> Thanks for this PR, it's off to a good start.
> 
> We try to avoid using multiple API versions for a service wherever possible. I've had a quick scan over the service and I think this should be possible here without too much difficulty. Since the changes you are targeting in the resource are pretty much required for the API bump could you update to just using v3.0? I've put some other comments in below that will also need addressing. We can help you out as needed, and if you're not already joined the community slack, there are instructions for joining in the project [readme](https://github.com/terraform-providers/terraform-provider-azurerm#terraform-provider-for-azure-resource-manager).

Thanks for reviewing! 

You're right, the change to upgrade to v3 is actually trivial; should be fixed now. I created a new commit for this change, do you want me to squash instead.

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
beandrad: @jackofallops do you still think that there should be a default value?

examples/security/securitycenter-pricing/main.tf
```diff
@@ -0,0 +1,8 @@
+provider "azurerm" {
+  features {}
+}
+
+resource "azurerm_security_center_subscription_pricing" "example" {
+    tier = "Standard"
+    resource_type = "VirtualMachines"
```
jackofallops, line 7: Style / fmt change (as per `terraform fmt`)
```suggestion
    tier          = "Standard"
    resource_type = "VirtualMachines"
```

website/docs/r/security_center_subscription_pricing.markdown
```diff
@@ -19,6 +19,7 @@ Manages the Pricing Tier for Azure Security Center in the current subscription.
 ```hcl
 resource "azurerm_security_center_subscription_pricing" "example" {
   tier = "Standard"
+  resource_type = "VirtualMachines"
```
jackofallops, line 22: Style / fmt
```suggestion
  tier          = "Standard"
  resource_type = "VirtualMachines"
```

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
```diff
@@ -44,6 +52,20 @@ func resourceArmSecurityCenterSubscriptionPricing() *schema.Resource {
...
+				ValidateFunc: validation.StringInSlice([]string{
+					"AppServices",
+					"ContainerRegistry",
+					"KeyVaults",
+					"KubernetesService",
+					"SqlServers",
+					"SqlServerVirtualMachines",
+					"StorageAccounts",
+					"VirtualMachines",
+				}, false),
```
jackofallops, line 67: Is there documentation somewhere that exposes these values? Do we know this is an exhaustive list? (The REST docs only indicate a `string` type with no guidance on allowed values)

jackofallops (changes requested):
Hi @beandrad 
Thanks for the updates and changes. I've left a few more comments below, but looking really good so far.

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
beandrad: As discussed, unfortunately, there's no documentation on the values that the resource type can take; I got the list of values using the list endpoint (https://docs.microsoft.com/en-us/rest/api/securitycenter/pricings/list).

The problem seems to be in the Azure API specs, which don't define `PricingName` (resource type) as an enum, see https://github.com/Azure/azure-rest-api-specs/blob/3695f4ef3bf65c253509d2708c31c96b71822613/specification/security/resource-manager/Microsoft.Security/stable/2018-06-01/pricings.json#L223.

Created an issue (https://github.com/Azure/azure-rest-api-specs/issues/11006) so that the enum can be used once it gets resolved.

azurerm/internal/services/securitycenter/resource_arm_security_center_workspace.go
```diff
@@ -77,15 +78,13 @@ func resourceArmSecurityCenterWorkspaceCreateUpdate(d *schema.ResourceData, meta
 
 	// get pricing tier, workspace can only be configured when tier is not Free.
 	// API does not error, it just doesn't set the workspace scope
-	price, err := priceClient.GetSubscriptionPricing(ctx, securityCenterSubscriptionPricingName)
+	isPricingStandard, err := isPricingStandard(ctx, priceClient)
+
 	if err != nil {
-		return fmt.Errorf("Error reading Security Center Subscription pricing: %+v", err)
+		return fmt.Errorf("Error checking Security Center Subscription pricing tier %v", err)
```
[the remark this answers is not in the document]
beandrad: I made the change for all the messages in the .../securitycenter; it's in a different commit.

jackofallops (changes requested):
Hi @beandrad - Thanks for the continued great work here. The changes of API have broken a couple of the acceptnce tests, if you could fix those up also, I think we're good to merge.
For context, `TestAccAzureRMSecurityCenter_pricingAndWorkspace` and `TestAccAzureRMSecurityCenter_contact` are both still using `default`.
Thanks!

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
```diff
@@ -44,6 +52,21 @@ func resourceArmSecurityCenterSubscriptionPricing() *schema.Resource {
 					string(security.Standard),
 				}, false),
 			},
+			"resource_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "VirtualMachines",
+				ValidateFunc: validation.StringInSlice([]string{
+					"AppServices",
+					"ContainerRegistry",
```
tombuildsstuff, line 61: given the others are pluralised - just double-checking this isn't meant to be `ContainerRegistries`?

jackofallops:
Tests passing (failures unrelated to change)
image

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
jackofallops: The list came from the `List` api: https://github.com/terraform-providers/terraform-provider-azurerm/pull/8549#discussion_r498363634
but good spot, we'll need to keep an eye out for this being "brought in line".

jackofallops (approved):

azurerm/internal/services/securitycenter/resource_arm_security_center_subscription_pricing.go
beandrad: Yes, consistency is a problem, it is indeed singular in this case.

Issue #7047, opened 2020-05-22
Support for IP Groups in Azure Firewall


dduleep:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

IP Groups allow to group and manage IP addresses for Azure Firewall rules in the following ways:

- As a source address in DNAT rules
- As a source or destination address in network rules
- As a source address in application rules

An IP Group can have a single IP address, multiple IP addresses, or one or more IP address ranges.
IP Groups could be reused in Azure Firewall DNAT, network, and application rules for multiple firewalls across regions and subscriptions in Azure. Group names must be unique. 

### Affected Resource(s)

* rule{source_addresses} in azurerm_firewall_nat_rule_collection 
* rule{source_addresses} in azurerm_firewall_network_rule_collection
* rule{source_addresses} in azurerm_firewall_application_rule_collection

### New Resource(s)
 * azurerm_ip_groups
* data_ip_groups

### Potential Terraform Configuration

```hcl

azurerm_firewall_nat_rule_collection {
...
   rule{
     source_addresses = data_ip_groups.group-ip.name
  }
}

```

### References

* [Microsoft Doc - IP Groups in Azure Firewall](https://docs.microsoft.com/en-us/azure/firewall/ip-groups)

LucianFrango:
I had a duplicate to this issue as I missed seeing this when i raised #7827 

Any ETA on this? Working with a customer that this functionality is a requirement.

LucianFrango:
Bump...

PR #8556, opened 2020-09-21
New resource: azurerm_ip_groups

manicminer:
- `azurerm_firewall_network_rule_collection`: support destination_ip_groups and source_ip_groups for rules
- `azurerm_firewall_nat_rule_collection`: support source_ip_groups for rules
- `azurerm_firewall_application_rule_collection`: support source_ip_groups for rules
- `azurerm_ip_groups` new resource and data source

Closes: #7047


azurerm/internal/services/network/tests/ip_group_resource_test.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func testAccAzureRMIpGroup_complete(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
+
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-%d"
```
katbyte, line 222: we try to put the service/other identifying name in resource groups

```suggestion
  name     = "acctestRG-network-%d"
```

katbyte (approved):
Aside from one minor comments about resource group names LGTP 👍

azurerm/internal/services/network/tests/ip_group_resource_test.go
manicminer: Great, done!

alexwilcox9:
Hi, I see this PR is adding a data source as well as a resource but the changelog only mentions the data source
Is this an error in the changelog or will the resource be in a different release?

manicminer:
@alexwilcox9 looks like this is an accidental omission. both the resource and data source should make it in the next release. Good catch, thanks!

PR #8557, opened 2020-09-21
`azurerm_search_service` - Add support for `allowed_ips`

mbfrahry:
```
--- PASS: TestAccAzureRMSearchService_basic (84.53s)
--- PASS: TestAccAzureRMSearchService_requiresImport (89.07s)
--- PASS: TestAccAzureRMSearchService_update (1621.21s)
--- PASS: TestAccAzureRMSearchService_ipRules (1168.95s)
```

### DIFF
--- a/azurerm/internal/services/search/resource_arm_search_service.go
+++ b/azurerm/internal/services/search/resource_arm_search_service.go
@@ -109,7 +109,7 @@ func resourceArmSearchService() *schema.Resource {
 				Default:  true,
 			},
 
-			"ip_rules": {
+			"allowed_ips": {
 				Type:     schema.TypeList,
 				Optional: true,
 				Elem: &schema.Schema{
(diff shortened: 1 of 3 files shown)


website/docs/r/search_service.html.markdown
```diff
@@ -52,6 +52,8 @@ The following arguments are supported:
 
 -> **Note:** `partition_count` and `replica_count` can only be configured when using a `standard` sku.
 
+* `ip_rules` - (Optional) A list of IP addresses that are allowed access to the search service endpoint.
```
katbyte, line 55: Should we have an example of whats expected? eg `10.0.0.1` ?

azurerm/internal/services/search/resource_arm_search_service.go
```diff
@@ -108,6 +109,15 @@ func resourceArmSearchService() *schema.Resource {
 				Default:  true,
 			},
 
+			"ip_rules": {
```
katbyte, line 112: would this be better as 
```suggestion
			"allowed_ips": {
```

katbyte (approved):
Aside from a couple comments LGMT 👍

website/docs/r/search_service.html.markdown
mbfrahry: We have some restrictions with loop/local addresses so I'd have to put some random IP there which doesn't feel great. I'll just update that it's a list of ipv4 addresses and we can update it further if there is confusion.

azurerm/internal/services/search/resource_arm_search_service.go
mbfrahry: Done!

Issue #3130, opened 2019-03-27
Allow Keyvault Network ACL to be set independently


simonlegeek:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It would be nice if we could set the Network ACLs (especially for the IP rules) config block from Azure Keyvault independantly from the Keyvault resource itself (like with Azure SQL Server). This would prevent some dependancy cycle issues between resources. (Let's say a WebApp that needs the Keyvault URI in his app settings and the keyvault that needs the oubound IP from the Webapp in his network ACLs.) 

### New or Affected Resource(s)

* azurerm_key_vault

### Potential Terraform Configuration
Examples below of the dependancy cycle I often encounter.

```hcl
resource "azurerm_app_service" "hub_webapp" {
  name                = "hub-webapp-${var.deployment_suffix}"
  location            = "${azurerm_resource_group.main_rg.location}"
  resource_group_name = "${azurerm_resource_group.main_rg.name}"
  app_service_plan_id = "${azurerm_app_service_plan.hub_webapp_plan.id}"

  site_config {
    always_on                = true
    dotnet_framework_version = "v4.0"
  }
  app_settings {
        
    "azure:KeyVaultURL"                         = "${azurerm_key_vault.regional-keyvault.vault_uri}"
  }

resource "azurerm_key_vault" "regional-keyvault" {
    name                            = "hub-keyvault-${var.deployment_suffix}"
    location                        = "${azurerm_resource_group.main_rg.location}"
    resource_group_name             = "${azurerm_resource_group.main_rg.name}"
    enabled_for_disk_encryption     = false
    enabled_for_template_deployment = true
    tenant_id                       = "${var.tenantid}"

    sku {
        name = "standard"
    }

    network_acls {
        default_action             = "Deny"
        bypass                     = "AzureServices"
        ip_rules                   = ["${azurerm_app_service.hub_webapp.possible_outbound_ip addresses}"]
    }
}

```

Shr3ps:
We need this too! For now to break this dep cycle, we need to use a terraform null_resource which execute `az cli` to update those Network ACLs.

[...]

PR #8560, opened 2020-09-22
Separate resource to manage Key vault network ACLs

sureshpatamsetty:


katbyte:
Hi @sureshpatamsetty - thank you for the new resource here. However i am curious as to the justification for moving the ACLs into a separate resource here? What is the use case for this separate resource that cannot be accomplished with them residing in the key vault?

sureshpatamsetty:
Hi @katbyte - The primary use case is when we want to have the network resources in a separate module and have network related configurations along with those network resources.

sureshpatamsetty:
Hi @katbyte, could you please add further comments on the progress/review the PR?

bhavingandhi01:
Hi @katbyte could you please provide updates on when this PR can be merged? The reason for separate resource is to allow network rules configuration separately from resource definition. This is similar to `azurerm_storage_account_network_rules` resource.

bhavingandhi01:
Hello @katbyte, we would like to use this feature in our IAC pipeline (similar to the other PR for ACR), could you please help with the review process?

bhavingandhi01:
Hello @katbyte could you please help approving this PR? It is unclear from the conversation here what is blocking the sign off on this?

m2hd1:
Thanks @sureshpatamsetty for making those changes, everything looks great 

@katbyte we setup our keyvault as well separately and this will be very useful. 
if you could approve it that would be great 🙏

dgcaron:
i was running into an issue when activating the firewall (default action deny) on creation of the keyvault we would get intermittent errors on creating access policies.. this most likely solves the issue

tombuildsstuff:
@dgcaron can you provide a config showing that scenario?

dgcaron:
@tombuildsstuff yes sure.. this is what we have:

```
resource "azurerm_key_vault" "instances" {
  for_each                        = local.all_keyvaults
  name                            = each.value.name
  location                        = var.resource_group_location
  resource_group_name             = each.value.resource_group_name
  enabled_for_deployment          = true
  enabled_for_disk_encryption     = true
  enabled_for_template_deployment = true
  tenant_id                       = var.tenant_id

  sku_name                   = "standard"
  soft_delete_enabled        = true
  soft_delete_retention_days = var.soft_delete_retention_days
  purge_protection_enabled   = true

  dynamic "network_acls" {
    for_each = var.firewall_enabled || var.virtualnetwork_enabled ? [1] : []
    content {
      bypass                     = "AzureServices"
      ip_rules                   = values(var.firewall_rules).*.cidr
      virtual_network_subnet_ids = contains(keys(var.subnets), each.key) ? [var.subnets[each.key]] : []
    }
  }
}

resource "azurerm_key_vault_access_policy" "datafactories" {
  for_each = {
    for kv_key, kv in azurerm_key_vault.instances : kv_key => kv if var.datafactory_enabled
  }

  key_vault_id = each.value.id
  tenant_id    = var.datafactory_instances[each.key].identity.tenant_id
  object_id    = var.datafactory_instances[each.key].identity.principal_id

  key_permissions = [
    "get",
  ]

  secret_permissions = [
    "get",
  ]

  storage_permissions = [
    "get",
  ]

}

resource "azurerm_key_vault_access_policy" "serviceprincipal_id" {
  for_each = azurerm_key_vault.instances

  key_vault_id = each.value.id
  tenant_id    = var.tenant_id
  object_id    = var.terraform_serviceprincipal_id

  secret_permissions = [
    "get",
    "list",
    "set",
    "delete",
    "recover",
    "backup",
    "restore"
  ]
}
```

tombuildsstuff:
hey @sureshpatamsetty @bhavingandhi01 

From our side we've been trying to work out how's best to proceed here - whilst there's benefits in some use-cases for splitting these resources out - we generally only do this when there's circular references involved (for example, API Management and Custom Domains).

Whilst there's some resources which don't fall into that category, as a general rule we're only adding these if there's a common use-case (or a lot of requests for them) - since the conflicting inline and split out blocks can cause confusion for users not using this functionality (for example, when removing the `network_acls` block, the rules should be removed).

Having spent some time mulling this over, whilst adding this new resource could be useful in some circumstances, it appears to be an organisation specific issue that would introduce a behavioural change for the majority of users. Instead we believe that this would be better suited by refactoring the Terraform Configuration, which would allow using the existing `network_acls` block.

As such whilst I'd like to thank you for this contribution (and apologies for the delayed response here), since this isn't something we plan to support I'm going to close this PR for the moment.

Thanks!

mancaus:
@tombuildsstuff - I've followed #3130 here, as this is also an issue for us, but with a use case that I suspect may be more common among enterprise customers, that has no good workaround I'm aware of, and occurs in a typical configuration / usage of Terraform.

The general scenario is:
- We implement private link for access to resources from outside of our VNETs, including from our deployment infrastructure, which is not hosted in Azure.
- We implement network policies that ensure resources can only be accessed by private link, from the VNETs that require access and from any Azure services that access them.
- We try not to arbitrarily split our Terraform into multiple deployments. Each tf deployment corresponds to infrastructure intended to have independent life-cycles.

This means we deploy resources subject to network ACLs, their private link resources and sub-resources (e.g. containers in an SA, certificates in a keyvault) in the same deployment. The sub-resources must be created at a point when there is access to the data plane of the resource. We can't rely on private link for this, as even with a depends_on to the private link resources from the sub-resource,  the DNS propagation delay means the private link is not available when we need it.

For storage accounts however we use the separate network rule resource to manage this. We ensure the network rule only applies after the subresources have been created, via depends_on. This works OK, although it's clearly a workaround rather than a direct solution.

We can't do this with keyvault however. And I'm not aware of any good workaround, other than running the deployment twice (yuck!).

Note that in general I'm a bit worried about the inconsistency between how network rules are managed between different resources in the provider. For example:
 - Storage accounts allow network rules to be managed as a separate resource, and multiple rules to be managed within the a ruleset.
 - Service bus is the same, but as far as I can tell, only allows a single rule in the ruleset.
 - Key vault does not allow separate management.

In my view the tf provider works best when the life-cycle and granularity of resources matches that within Azure itself. For example, IMO each separate resource type returned by "az resource …" should be a separate resource in the provider. Bundling or unbundling resources with separate underlying life-cycles seems to always lead to trouble.

I would prefer network rules be managed similarly to role assignments, where a scope is provided that can be any resource supporting a network rule. Would this be of interest to you? If so, I can raise an enhancement proposing it directly.

tombuildsstuff:
@mancaus unfortunately you're mostly referring to design issues within the Azure API's themselves - that there's no standard behaviour within the API for these is something which would need to be addressed by the ARM Team and subsequently fixed within the specific API's.

Other Cloud Providers use the approach of `association` or `attachment` resources to allow this kind of resource design, which the Azure Provider has emulated in some fashion (such as in the resources you've mentioned above) - but this is ultimately working around the design issues (including circular references) within the API's themselves. Whilst we try our best not to deviate from the Azure API's too much - in practice sometimes it's necessary to work around these fundamental design issues (such as circular references) by introducing these separate resources.

Unfortunately the approach you're proposing (submitting the `network_rules` block off to a given scope) won't work due to the differences between these resources and the fact that there's no SDK available for this (whilst there is an API for a given scope, you'd need to identify the API version for this specific resource which is applicable and ..) so ultimately this ends up becoming more work and edge-cases than supporting this natively within each resource.

Whilst I understand the problem your facing here - unfortunately this is a fundamental design issue within the Azure API's, where the API's are additive (e.g. you update the existing resource with new options) rather than segmented like other cloud providers, so achieving this architecture is challenging to mirror the API's themselves..

As such we appreciate the suggestion, however, since these designs need to be taken into approach when building the services themselves I'd suggest reaching out to the Service Teams in question here - as mentioned other cloud providers use the concept of "attaching" things which works well, which would be my suggestion for fixing that.

mancaus:
Thanks for the further explanation @tombuildsstuff - I'll keep looking for a more elegant solution than what we have today.

PR #8561, opened 2020-09-22
Separate resource to manage ACR network ruleset

sureshpatamsetty:


katbyte:
Hi @sureshpatamsetty - thank you for the new resource here. However i am curious as to the justification for moving the ACLs into a separate resource here? What is the use case for this separate resource that cannot be accomplished with them residing in the ACR resource?

sureshpatamsetty:
Hi @katbyte - The primary use case is when we want to have the network resources in a separate module and have network related configurations along with those network resources.

sureshpatamsetty:
Hi @katbyte, could you please add further comments on the progress/review the PR?

bhavingandhi01:
Hi @katbyte could you please provide updates on when this PR can be merged? The reason for separate resource is to allow network rules configuration separately from resource definition. This is similar to `azurerm_storage_account_network_rules` resource.

bhavingandhi01:
Hello @katbyte, we would like to use this feature in our IAC pipeline, could you please help with the review process?

bhavingandhi01:
Hello @katbyte could you please help approving this PR? It is unclear from the conversation here what is blocking the sign off on this?

tombuildsstuff:
hey @sureshpatamsetty @bhavingandhi01 

From our side we've been trying to work out how's best to proceed here - whilst there's benefits in some use-cases for splitting these resources out - we generally only do this when there's circular references involved (for example, API Management and Custom Domains).

Whilst there's some resources which don't fall into that category, as a general rule we're only adding these if there's a common use-case (or a lot of requests for them) - since the conflicting inline and split out blocks can cause confusion for users not using this functionality (for example, when removing the `network_rule_set` block, the rules should be removed).

Having spent some time mulling this over, whilst adding this new resource could be useful in some circumstances, it appears to be an organisation specific issue that would introduce a behavioural change for the majority of users. Instead we believe that this would be better suited by refactoring the Terraform Configuration, which would allow using the existing `network_rule_set` block.

As such whilst I'd like to thank you for this contribution (and apologies for the delayed response here), since this isn't something we plan to support I'm going to close this PR for the moment.

Thanks!

Issue #7319, opened 2020-06-13
Support for Azure Firewall Manager resources


jturver1:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Support for Azure Firewall Manager Policies

It would be great to see some MVP support for the relative new Azure Firewall Manager constructs:
Azure Firewall Manager constructs, including "Secure Virtual Hub" configurations
Azure Firewall Manager Firewall Policies (parent, child)
Azure Firewall Manager Policy Rule Collections (Network, Application, DNAT)
Azure Firewall Manager Rule Collection Rules

Firewall Manager will likely supersede Azure Firewall (azurerm_firewall) for deploying and managing Azure Firewall in many cases. It would be fantastic to be able to manage these resources via Terraform.

* azurerm_firewallmanager
* azurerm_firewallmanager_securevirtualhub
* azurerm_firewallmanager_securevirtualhub_routesettings
* azurerm_firewallmanager_securevirtualhub_connections
* azurerm_firewallmanager_securevirtualhub_securityproviders
* azurerm_firewallmanager_firewall
* azurerm_firewallmanager_policy (parent and child)
* azurerm_firewallmanager_policy_rulecollectiongroup
* azurerm_firewallmanager_policy_rulecollectiongroup_applicationrule
* azurerm_firewallmanager_policy_rulecollectiongroup_networkrule
* azurerm_firewallmanager_policy_rulecollectiongroup_dnatrule

This provides an example of how Azure Firewall Manager Policy constructs map to existing Azure Firewall policy in terms of some of the resource types required and their properties:
- https://docs.microsoft.com/en-us/azure/firewall-manager/migrate-to-policy

### Potential Terraform Configuration

For MVP, it would be great to have the ability to manage the following resources as a priority/starting point:
1.  Azure Firewall Manager Policies, Rule collections, and Rules
2. "Secure Virtual Hubs" and associated Connections, and Route Settings

### References
https://docs.microsoft.com/en-us/azure/firewall-manager/
https://docs.microsoft.com/en-us/azure/firewall-manager/migrate-to-policy

magodo:
[...]

PR #8603, opened 2020-09-24
new resource: azurerm_firewall_policy_rule_collection_group

magodo:
## Test Result

```bash
💤 make testacc TEST=./azurerm/internal/services/network/tests TESTARGS='-run TestAccAzureRMFirewallPolicyRuleCollectionGroup'

TF_ACC=1 go test ./azurerm/internal/services/network/tests -v -run TestAccAzureRMFirewallPolicyRuleCollectionGroup -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMFirewallPolicyRuleCollectionGroup_basic
=== PAUSE TestAccAzureRMFirewallPolicyRuleCollectionGroup_basic
=== RUN   TestAccAzureRMFirewallPolicyRuleCollectionGroup_complete
=== PAUSE TestAccAzureRMFirewallPolicyRuleCollectionGroup_complete
=== RUN   TestAccAzureRMFirewallPolicyRuleCollectionGroup_update
=== PAUSE TestAccAzureRMFirewallPolicyRuleCollectionGroup_update
=== RUN   TestAccAzureRMFirewallPolicyRuleCollectionGroup_requiresImport
=== PAUSE TestAccAzureRMFirewallPolicyRuleCollectionGroup_requiresImport
=== CONT  TestAccAzureRMFirewallPolicyRuleCollectionGroup_basic
=== CONT  TestAccAzureRMFirewallPolicyRuleCollectionGroup_requiresImport
=== CONT  TestAccAzureRMFirewallPolicyRuleCollectionGroup_update
=== CONT  TestAccAzureRMFirewallPolicyRuleCollectionGroup_complete
--- PASS: TestAccAzureRMFirewallPolicyRuleCollectionGroup_basic (124.12s)
--- PASS: TestAccAzureRMFirewallPolicyRuleCollectionGroup_complete (124.13s)
--- PASS: TestAccAzureRMFirewallPolicyRuleCollectionGroup_requiresImport (134.30s)
--- PASS: TestAccAzureRMFirewallPolicyRuleCollectionGroup_update (216.12s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/tests       216.163s
```

Related issue: #7319


azurerm/internal/services/network/firewall_policy_rule_collection_group_resource.go
```diff
@@ -0,0 +1,816 @@
...
+											Type: schema.TypeString,
+										},
+									},
+									"source_ip_groups": {
+										Type:     schema.TypeSet,
+										Optional: true,
+										Elem: &schema.Schema{
+											Type: schema.TypeString,
+										},
+									},
```
katbyte, line 307: could we validate these elements?

azurerm/internal/services/network/firewall_policy_rule_collection_group_resource.go
```diff
@@ -0,0 +1,816 @@
...
+										Type:         schema.TypeString,
+										Optional:     true,
+										ValidateFunc: validation.IsIPAddress,
+									},
+									"destination_ports": {
+										Type:     schema.TypeSet,
+										Optional: true,
+										Elem: &schema.Schema{
+											Type:         schema.TypeString,
+											ValidateFunc: validate.FirewallPolicyRulePort,
```
katbyte, line 318: could we ue the plugin SDK validater here?

website/docs/r/firewall_policy_rule_collection_group.html.markdown
```diff
@@ -0,0 +1,217 @@
...
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name which should be used for this Firewall Policy Rule Collection Group. Changing this forces a new Firewall Policy Rule Collection Group to be created.
+
+* `firewall_policy_id` - (Required) The ID of the Firewall Policy where the Firewall Policy Rule Collection Group should exist. Changing this forces a new Firewall Policy Rule Collection Group to be created.
+
+* `priority` - (Required) The priority of the Firewall Policy Rule Collection Group. The range is 100-65000.
```
katbyte, line 88:
```suggestion
* `priority` - (Required) The priority of the Firewall Policy Rule Collection Group. The range is `100`-`65000`.
```

website/docs/r/firewall_policy_rule_collection_group.html.markdown
```diff
@@ -0,0 +1,217 @@
...
+
+* `network_rule_collection` - (Optional) One or more `network_rule_collection` blocks as defined below.
+
+---
+
+A `application_rule_collection` block supports the following:
+
+* `name` - (Required) The name which should be used for this application rule collection.
+
+* `action` - (Required) The action type of the application rule collection. Possible values are `Allow` and `Deny`.
```
katbyte, line 104:
```suggestion
* `action` - (Required) The action to take for application rules in this collection. Possible values are `Allow` and `Deny`.
```

website/docs/r/firewall_policy_rule_collection_group.html.markdown
```diff
@@ -0,0 +1,217 @@
...
+
+* `priority` - (Required) The priority of the application rule collection. The range is 100-65000.
```
katbyte, line 106:
```suggestion
* `priority` - (Required) The priority of this application rule collection. The range is `100`-`65000`.
```

website/docs/r/firewall_policy_rule_collection_group.html.markdown
```diff
@@ -0,0 +1,217 @@
...
+
+---
+
+A `rule` (network rule) block supports the following:
+
+* `name` - (Required) The name which should be used for this rule.
+
+* `protocols` - (Required) Specifies a list of network protocols. Possible values are `Any`, `TCP`, `UDP`, `ICMP`.
+
+* `source_addresses` - (Optional) Specifies a list of source ip addresses.
```
katbyte, line 158: IP in IP address should be capitalized
```suggestion
* `source_addresses` - (Optional) Specifies a list of source IP addresses.
```

website/docs/r/firewall_policy_rule_collection_group.html.markdown
```diff
@@ -0,0 +1,217 @@
...
+
+* `destination_fqdn_tags` - (Optional) Specifies a list of destination FQDN tags.
+
+---
+
+A `rule` (network rule) block supports the following:
+
+* `name` - (Required) The name which should be used for this rule.
+
+* `protocols` - (Required) Specifies a list of network protocols. Possible values are `Any`, `TCP`, `UDP`, `ICMP`.
```
katbyte, line 156:
```suggestion
* `protocols` - (Required) Specifies a list of network protocols this rule applies to. Possible values are `Any`, `TCP`, `UDP`, `ICMP`.
```

katbyte (changes requested):
Thanks for the PR @magodo - overall this is looking good and i've left some comments inline.

azurerm/internal/services/network/firewall_policy_rule_collection_group_data_source.go
```diff
@@ -0,0 +1,67 @@
...
+
+			"firewall_policy_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.FirewallPolicyID,
+			},
+		},
+	}
+}
```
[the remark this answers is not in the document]
magodo: I'll remove it as there is no use case for this data source as far as i can tell.

azurerm/internal/services/network/firewall_policy_rule_collection_group_resource.go
```diff
@@ -0,0 +1,816 @@
...
+				ValidateFunc: validate.FirewallPolicyID,
+			},
+
+			"priority": {
+				Type:         schema.TypeInt,
+				Required:     true,
+				ValidateFunc: validation.IntBetween(100, 65000),
+			},
+
+			"application_rule_collection": {
```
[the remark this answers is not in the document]
magodo: This block has an internal block called `rule`, which is a set of rules. Also this block has its own "name", "priority" and "action", which might be better kept as is so that we can reasonably name those properties as "collection name", "collection priority", etc.

Also, this might be a less burden for users who are used to `azurerm_firewall_applicatioon_rule_collection` resource, to migrate to using this new resource.

azurerm/internal/services/network/firewall_policy_rule_collection_group_resource.go
```diff
@@ -0,0 +1,816 @@
...
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+
+			"network_rule_collection": {
```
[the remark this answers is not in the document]
magodo: same as above.

azurerm/internal/services/network/firewall_policy_rule_collection_group_resource.go
```diff
@@ -0,0 +1,816 @@
...
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+
+			"nat_rule_collection": {
```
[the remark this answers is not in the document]
magodo: I propose to keep the naming as it was for the same reason as above.

azurerm/internal/services/network/firewall_policy_rule_collection_group_resource.go
```diff
@@ -0,0 +1,816 @@
...
+										Elem: &schema.Schema{
+											Type: schema.TypeString,
+										},
+									},
+									"destination_ports": {
+										Type:     schema.TypeSet,
+										Optional: true,
+										Elem: &schema.Schema{
+											Type:         schema.TypeString,
+											ValidateFunc: validate.FirewallPolicyRulePort,
```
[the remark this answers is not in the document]
magodo: In fact, the API used here allows not only a simple port number, it also supports a port range (e.g. `1000-2000`). Hence, we need to extend the SDK port validation function a bit here.

azurerm/internal/services/network/firewall_policy_rule_collection_group_resource.go
magodo: In fact, the API used here allows not only a simple port number, it also supports a port range (e.g. `1000-2000`). Hence, we need to extend the SDK port validation function a bit here.

magodo:
@katbyte I have updated the code per your review comments, thank you! Please take another look!

azurerm/internal/services/network/parse/firewall_policy.go
```diff
@@ -31,3 +32,28 @@ func FirewallPolicyID(input string) (*FirewallPolicyId, error) {
...
 	return &policy, nil
 }
+
+type FirewallPolicyRuleCollectionGroupId struct {
+	ResourceGroup string
+	PolicyName    string
+	Name          string
+}
+
+func FirewallPolicyRuleCollectionGroupID(input string) (*FirewallPolicyRuleCollectionGroupId, error) {
```
manicminer, line 42: Could you make this function use `azure.ParseAzureResourceID` rather than depending on `FirewallPolicyID` and a regex match?

azurerm/internal/services/network/firewall_policy_rule_collection_group_resource.go
```diff
@@ -0,0 +1,860 @@
...
+	policyClient := meta.(*clients.Client).Network.FirewallPolicyClient
+	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	id, err := parse.FirewallPolicyRuleCollectionGroupID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	policyResp, err := policyClient.Get(ctx, id.ResourceGroup, id.PolicyName, "")
```
manicminer, line 446: Why do we need to check the existence of the linked policy before retrieving the rule collection group?

azurerm/internal/services/network/firewall_policy_rule_collection_group_resource.go
```diff
@@ -0,0 +1,860 @@
...
+									},
+									"source_ip_groups": {
+										Type:     schema.TypeSet,
+										Optional: true,
+										Elem: &schema.Schema{
+											Type:         schema.TypeString,
+											ValidateFunc: validation.StringIsNotEmpty,
+										},
+									},
+									"destination_address": {
```
manicminer, line 341: Why do we accept a single address/CIDR when the API accepts a list?

manicminer (changes requested):
Hi @magodo, this is looking great. I have one more suggestion and a couple questions below if you could help me understand some of the design decisions. Thanks!

azurerm/internal/services/network/parse/firewall_policy.go
magodo: sure

azurerm/internal/services/network/firewall_policy_rule_collection_group_resource.go
magodo: The reason is that we will need to use the `policyResp` to set the `firewall_policy_id` later. Instead, we should opt in to using the ID Formatter to save this API call.

azurerm/internal/services/network/firewall_policy_rule_collection_group_resource.go
magodo: The API actually doesn't support a list, it will raise error:

> Invalid DNat destination address. Multiple values, ranges and * are not supported.

magodo:
@manicminer I have updated per your commments, please take another look, thank you!

manicminer (changes requested):
@magodo Thanks, there's one other issue I noticed - can we store the subscription ID in the `FirewallPolicyId` and `FirewallPolicyRuleCollectionGroupId` structs, taking it from the original ID string, instead of passing it in afterwards?

magodo:
@manicminer I assume you mean [this part](https://github.com/terraform-providers/terraform-provider-azurerm/pull/8603/commits/35d3cba56ef3210c31f99309c5d5e9bf582d6b5e#diff-c10d33371f7ac7068b69b83cc3954e493c7d43a07946d61a2b5c07a1679172a8R435), this is actually how the [`Formatter`](https://github.com/terraform-providers/terraform-provider-azurerm/blob/2c7ae3cfcc59aad45239d021aa64198dd1701d51/azurerm/internal/resourceid/interface.go) interface defined, and is used prevalently elsewhere.

manicminer:
@magodo Ah ok thanks. I still prefer the subscriptionId be sourced from the resource rather than from the client, but this looks good to merge 👍

manicminer (approved):

manicminer:
Test results:

Screenshot 2020-10-30 at 18 14 52

magodo:
@manicminer TBH, my first adoptions to the id structures imbedded the subscriptionId inside 😅

Issue #8593, opened 2020-09-23
Resource with IDFormatter used for setting ID should also be used all over the place


magodo:
For resources whose ID type implemeted the `resourceid.Formatter` interface, the ID should be used at least following places:

- when setting the resource id in create callback
- when formating the import error message
- other places

Addtionally, we need to ensure the ID Parser conforms to the ID Formatter, in turns of casing.

## Affected Types

- azurerm/internal/services/cdn/parse/cdn_endpoint.go:48:1: CdnEndpointId
- azurerm/internal/services/cdn/parse/cdn_profile.go:42:1: CdnProfileId
- azurerm/internal/services/compute/parse/availability_set.go:21:1: AvailabilitySetId
- azurerm/internal/services/compute/parse/dedicated_host.go:23:1: DedicatedHostId
- azurerm/internal/services/compute/parse/dedicated_host_group.go:21:1: DedicatedHostGroupId
- azurerm/internal/services/compute/parse/disk_encryption_set.go:21:1: DiskEncryptionSetId
- azurerm/internal/services/compute/parse/image.go:21:1: ImageId
- azurerm/internal/services/compute/parse/managed_disk.go:21:1: ManagedDiskId
- azurerm/internal/services/compute/parse/proximity_placement_group.go:21:1: ProximityPlacementGroupId
- azurerm/internal/services/compute/parse/shared_image.go:23:1: SharedImageId
- azurerm/internal/services/compute/parse/shared_image_gallery.go:21:1: SharedImageGalleryId
- azurerm/internal/services/compute/parse/shared_image_version.go:25:1: SharedImageVersionId
- azurerm/internal/services/compute/parse/virtual_machine.go:21:1: VirtualMachineId
- azurerm/internal/services/compute/parse/virtual_machine_extension.go:23:1: VirtualMachineExtensionId
- azurerm/internal/services/compute/parse/virtual_machine_scale_set.go:21:1: VirtualMachineScaleSetId
- azurerm/internal/services/compute/parse/virtual_machine_scale_set_extension.go:23:1: VirtualMachineScaleSetExtensionId
- azurerm/internal/services/frontdoor/parse/backend_pool.go:19:1: BackendPoolId
- azurerm/internal/services/frontdoor/parse/frontdoor.go:56:1: FrontDoorId
- azurerm/internal/services/frontdoor/parse/frontend_endpoint.go:19:1: FrontendEndpointId
[...]

PR #8607, opened 2020-09-24
standardize ID Formatter usage

magodo:
The ID formatter should be used exclusively. If a resource opt in using
the ID formatter, it should use it in below places at the same time:

- when setting the resource id in create callback (resource) or read
  callback (data source)
- when formatting the import error message
- when setting another resource's id as some property

fixes: #8593

### DIFF
--- a/azurerm/internal/services/cdn/cdn_endpoint_resource.go
+++ b/azurerm/internal/services/cdn/cdn_endpoint_resource.go
@@ -236,7 +236,11 @@ func resourceArmCdnEndpointCreate(d *schema.ResourceData, meta interface{}) erro
 	}
 
 	if existing.ID != nil && *existing.ID != "" {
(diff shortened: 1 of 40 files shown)


tombuildsstuff:
hey @magodo 

Taking a look through whilst on the whole this is off to a good start - there's some larger work needed to enable this functionality across all resources. Part of that work is in the form of the recently added Resource ID Formatters, Parsers and Validators - however whilst that provides the baseline, ultimately we need to ensure this is rolled out in a manner that doesn't break existing users, by validating each resource to confirm if a State Migration is necessary or not.

For the moment whilst I'd like to thank you for this contribution I'm going to close this PR for the moment - but ultimately #8593 will be fixed over the coming months in a more gradual fashion, since we're going to need to confirm if each resource requires a state migration or not.

Thanks!

PR #8610, opened 2020-09-24
feat: CDN endpoint disable default value for compression

zekth:
fix: https://github.com/terraform-providers/terraform-provider-azurerm/issues/1109

Regarding this issue, this PR fixes it but it will forces people to set the attribute if they have not set `is_compression_allowed` attribute in their terraform template.

Currently this is a blocking issue for people willing to use Premium Verizon

### DIFF
--- a/azurerm/internal/services/cdn/cdn_endpoint_resource.go
+++ b/azurerm/internal/services/cdn/cdn_endpoint_resource.go
@@ -437,7 +437,6 @@ func resourceArmCdnEndpointRead(d *schema.ResourceData, meta interface{}) error 
 
 	if props := resp.EndpointProperties; props != nil {
 		d.Set("host_name", props.HostName)
-		d.Set("is_compression_enabled", props.IsCompressionEnabled)
 		d.Set("is_http_allowed", props.IsHTTPAllowed)
 		d.Set("is_https_allowed", props.IsHTTPSAllowed)
 		d.Set("querystring_caching_behaviour", props.QueryStringCachingBehavior)
@@ -445,6 +444,10 @@ func resourceArmCdnEndpointRead(d *schema.ResourceData, meta interface{}) error 
 		d.Set("origin_path", props.OriginPath)
 		d.Set("probe_path", props.ProbePath)
 		d.Set("optimization_type", string(props.OptimizationType))
+		
+		if is_compression_enabled := props.IsCompressionEnabled; is_compression_enabled != nil {
+			d.Set("is_compression_enabled", *is_compression_enabled)
+		}
 
 		contentTypes := flattenAzureRMCdnEndpointContentTypes(props.ContentTypesToCompress)
 		if err := d.Set("content_types_to_compress", contentTypes); err != nil {
--- a/azurerm/internal/services/cdn/cdn_endpoint_resource.go
+++ b/azurerm/internal/services/cdn/cdn_endpoint_resource.go
@@ -141,7 +141,6 @@ func resourceArmCdnEndpoint() *schema.Resource {
 			"is_compression_enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
 			},
 
 			"probe_path": {
--- a/azurerm/internal/services/cdn/cdn_endpoint_resource.go
+++ b/azurerm/internal/services/cdn/cdn_endpoint_resource.go
@@ -241,7 +241,6 @@ func resourceArmCdnEndpointCreate(d *schema.ResourceData, meta interface{}) erro
 	location := azure.NormalizeLocation(d.Get("location").(string))
 	httpAllowed := d.Get("is_http_allowed").(bool)
 	httpsAllowed := d.Get("is_https_allowed").(bool)
-	compressionEnabled := d.Get("is_compression_enabled").(bool)
 	cachingBehaviour := d.Get("querystring_caching_behaviour").(string)
 	originHostHeader := d.Get("origin_host_header").(string)
 	originPath := d.Get("origin_path").(string)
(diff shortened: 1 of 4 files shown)


azurerm/internal/services/cdn/cdn_endpoint_resource.go
```diff
@@ -447,6 +445,10 @@ func resourceArmCdnEndpointRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("probe_path", props.ProbePath)
 		d.Set("optimization_type", string(props.OptimizationType))
 
+		if is_compression_enabled := props.IsCompressionEnabled; is_compression_enabled != nil {
+			d.Set("is_compression_enabled", *is_compression_enabled)
+		}
+
```
WodansSon, line 451: In order to get this to work as expected I suspect you are going to have to modify the `resourceArmCdnEndpointCreate` function as well. The reason for this is since the data type of `is_compression_enabled` is `bool` and in the original create function has this line of code:
```
compressionEnabled := d.Get("is_compression_enabled").(bool)
```
So even if the `d.Get` call does not find the `is_compression_enabled` attribute in the configuration file the value of `compressionEnabled` will be `false` because that is the default zero value for `bool`. Which means no matter what the `is_compression_enabled` attribute is, or is not, it will always be set to `false`. To fix this I think you will need to remove this line of code:
```
compressionEnabled := d.Get("is_compression_enabled").(bool)
``` 
and the this line of code from the `cdn.Endpoint` struct definition:
```
IsCompressionEnabled = &compressionEnabled
```
Then switch the logic in the `resourceArmCdnEndpointCreate` function to use the `d.GetOk` instead of just `d.Get` which will only set the value of `compressionEnabled` if the attribute actually exists in the configuration file, something along the lines of:

```
httpAllowed := d.Get("is_http_allowed").(bool)
httpsAllowed := d.Get("is_https_allowed").(bool)
cachingBehaviour := d.Get("querystring_caching_behaviour").(string)
originHostHeader := d.Get("origin_host_header").(string)
originPath := d.Get("origin_path").(string)
probePath := d.Get("probe_path").(string)
optimizationType := d.Get("optimization_type").(string)
contentTypes := expandArmCdnEndpointContentTypesToCompress(d)
geoFilters := expandCdnEndpointGeoFilters(d)

t := d.Get("tags").(map[string]interface{})

endpoint := cdn.Endpoint{
	Location: &location,
	EndpointProperties: &cdn.EndpointProperties{
		ContentTypesToCompress:     &contentTypes,
		GeoFilters:                 geoFilters,
		IsHTTPAllowed:              &httpAllowed,
		IsHTTPSAllowed:             &httpsAllowed,
		QueryStringCachingBehavior: cdn.QueryStringCachingBehavior(cachingBehaviour),
		OriginHostHeader:           utils.String(originHostHeader),
	},
	Tags: tags.Expand(t),
}

if v, ok := d.GetOk("is_compression_enabled"); ok {
	endpoint.EndpointProperties.IsCompressionEnabled = &v.(bool)
}
```
Does that make sense? Also, can you please add a test case for this scenario where the `is_compression_enabled` attribute is not passed in the configuration file?

WodansSon (changes requested):
@zekth, thanks so much for this PR... it's looking good so far, but I left a comment in the PR for a potential issue I noticed while reviewing the code. Get those two things fixed up and I think we would be good to go with merging these changes! 🚀

zekth:
@WodansSon slight changes compared to your review which seems to work. Also added a test that seems to fit the style. Please let me know your thoughts.

WodansSon (approved):
@zekth, thanks for pushing those changes, this LGTM now. If all the tests pass I will go ahead and merge the PR. Thanks again! 🚀

WodansSon:
@zekth, unfortunately the test fails and I cannot merge the code until the test passes:

```
go test -timeout 30000s -v ./azurerm/internal/services/cdn/tests -run TestAccAzureRMCdnEndpoint_withoutCompression

=== RUN   TestAccAzureRMCdnEndpoint_withoutCompression
=== PAUSE TestAccAzureRMCdnEndpoint_withoutCompression
=== CONT  TestAccAzureRMCdnEndpoint_withoutCompression
    TestAccAzureRMCdnEndpoint_withoutCompression: testing.go:684: Step 0 error: Check failed: Check 2/2 error: azurerm_cdn_endpoint.test: Attribute 'is_compression_enabled' found when not expected
--- FAIL: TestAccAzureRMCdnEndpoint_withoutCompression (212.51s)
```  

It appears even though you are attempting to suppress the `is_compression_enabled` attribute from being written to the state file it is still getting persisted some how.

azurerm/internal/services/cdn/cdn_endpoint_resource.go
```diff
@@ -447,6 +448,10 @@ func resourceArmCdnEndpointRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("probe_path", props.ProbePath)
 		d.Set("optimization_type", string(props.OptimizationType))
 
+		if is_compression_enabled := props.IsCompressionEnabled; is_compression_enabled != nil {
+			d.Set("is_compression_enabled", *is_compression_enabled)
+		}
```
WodansSon, line 453: The reason why the test is failing is because the value is always returned by the API, so we need to check the config file again in the `Read` function to determine if we should write the value to the state file or not, something like this:

```suggestion
		if _, ok := d.GetOk("is_compression_enabled"); ok {
			if compressionEnabled := props.IsCompressionEnabled; compressionEnabled != nil {
				d.Set("is_compression_enabled", *compressionEnabled)
			}
		}
```

zekth:
My bad, thanks @WodansSon

zekth:
CI seems to fail on website-lint but doesn't seem to be related

WodansSon:
> CI seems to fail on website-lint but doesn't seem to be related

Yes, it appears a change in the `Desktop Virtualization` was accidentally merged into main which is causing the CI to flag the documentation. I am going to be fixing that issue in another PR, so once that is merged later today you should be able to do a force-push from main and the issue will fix itself. Sorry about that... so close! 🚀

WodansSon:
@zekth, I have merged the PR that will fix your CI lint error, so feel free to do a force-push from main and I can get this one merged too! 🚀

WodansSon:
image

zekth:
Sorry to bring the bad news but it seems there is another issue:

```
cdn.EndpointsClient#Create: Failure sending request: 
StatusCode=400 -- Original Error: Code="BadRequest" 
Message="We couldn’t configure QueryStringCachingBehavior, IsCompressionEnabled, ContentTypesToCompress, GeoFilters and DeliveryPolicy for premium profile endpoints. 
You’ll need to manage these endpoints in the supplemental portal."
```

The terraform provider seems to add the QueryStringCachingBehavior and ContentTypesToCompress by default like IsCompressionEnabled was before the PR.

eg:
```tf
resource "azurerm_cdn_endpoint" "cdn_endpoint" {
  name                          = "cdne${local.formatted_name}"
  profile_name                  = var.cdn_profile_name
  location                      = "global"
  resource_group_name           = var.resource_group_name
  origin_path                   = "/foo"
  origin_host_header            = var.originHostname
  is_https_allowed              = true
  is_http_allowed               = false
  origin {
    name      = "name"
    host_name = var.originHostname
  }
}
```

Seems like the tests of the provider are not really in Sync with API. Especially with the verizon API.

PR #8619, opened 2020-09-25
`azurerm_kubernetes_cluster` - `min_count` can be updated when `enable_auto_scaling` is set to true

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/8576


katbyte (changes requested):
Thanks for this @neil-yechenwei - could we get a test added to validate this?

neil-yechenwei:
@katbyte , I assume this [test case](https://github.com/terraform-providers/terraform-provider-azurerm/blob/756cf9fc8acacc7b988b8ec29284d058494ca5b6/azurerm/internal/services/containers/tests/kubernetes_cluster_node_pool_resource_test.go#L94) has already covered this scenario. So I assume we don't need to add similar test case anymore, right?

neil-yechenwei:
@katbyte , thanks for your comment. Test case is added.

mbfrahry (approved):
LGTM!

PR #8621, opened 2020-09-25
Azurerm role assignment doc

magodo:
The `principal_id` is supposed to be the `object_id` of the `azurerm_client_config` rather than the `client_id`.


katbyte (approved):
Thanks @magodo - LGTM 👍

PR #8632, opened 2020-09-25
Support non-string type for parameters of azurerm_managed_application

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/8608


katbyte (changes requested):
Thanks for the pr @neil-yechenwei - looking at this i'm wondering if it would be better to instead support both and make them conflict? either parameters for those who need a simple `key=value` and then `paramters_as_json` for users who want to just pass in a json block?

neil-yechenwei:
@katbyte , thanks for your comments. I've updated code. Please have an another look. Thanks.

azurerm/internal/services/managedapplications/managed_application_resource.go
```diff
@@ -71,13 +74,25 @@ func resourceManagedApplication() *schema.Resource {
 			},
 
 			"parameters": {
-				Type:     schema.TypeMap,
-				Optional: true,
+				Type:          schema.TypeMap,
+				Optional:      true,
+				Computed:      true,
+				Deprecated:    "Deprecated in favour of `parameter_values`",
```
katbyte, line 80: we can remove this?
```suggestion
```

azurerm/internal/services/managedapplications/managed_application_resource_test.go
```diff
@@ -94,6 +94,21 @@ func TestAccManagedApplication_update(t *testing.T) {
 	})
 }
 
+func TestAccManagedApplication_parameterValues(t *testing.T) {
```
katbyte, line 97: could we add a test that goes string -> non-string -> string and checking to make sure the correct property is used?

katbyte (changes requested):
Thanks @neil-yechenwei - couple more comments to address and this should be good to merge

azurerm/internal/services/managedapplications/managed_application_resource.go
neil-yechenwei: Forget it. Great to catch. Removed.

azurerm/internal/services/managedapplications/managed_application_resource_test.go
neil-yechenwei: Added tc

neil-yechenwei:
@katbyte , thanks for your comments. I've updated code. Please have an another look. Thanks.

katbyte (changes requested):
thanks @neil-yechenwei, looks good. There are two test failures:
```
------- Stdout: -------
=== RUN   TestAccManagedApplication_complete
=== PAUSE TestAccManagedApplication_complete
=== CONT  TestAccManagedApplication_complete
    testing.go:620: Step 1/2 error: Error running apply: exit status 1
        
        Error: A resource with the ID "/subscriptions/*******/providers/Microsoft.MarketplaceOrdering/agreements/cisco/offers/meraki-vmx/plans/meraki-vmx100" already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for "azurerm_marketplace_agreement" for more information.
        
          on terraform_plugin_test.tf line 35, in resource "azurerm_marketplace_agreement" "test":
          35: resource "azurerm_marketplace_agreement" "test" {
        
--- FAIL: TestAccManagedApplication_complete (106.85s)
FAIL

------- Stderr: -------
2021/04/27 04:03:28 [DEBUG] not using binary driver name, it's no longer needed
2021/04/27 04:03:28 [DEBUG] not using binary driver name, it's no longer needed
 ------- Stdout: -------
=== RUN   TestAccManagedApplication_update
=== PAUSE TestAccManagedApplication_update
=== CONT  TestAccManagedApplication_update
    testing.go:620: Step 3/6 error: Error running apply: exit status 1
        
        Error: A resource with the ID "/subscriptions/*******/providers/Microsoft.MarketplaceOrdering/agreements/cisco/offers/meraki-vmx/plans/meraki-vmx100" already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for "azurerm_marketplace_agreement" for more information.
        
          on terraform_plugin_test.tf line 35, in resource "azurerm_marketplace_agreement" "test":
          35: resource "azurerm_marketplace_agreement" "test" {
        
--- FAIL: TestAccManagedApplication_update (437.25s)
FAIL

------- Stderr: -------
2021/04/27 04:03:28 [DEBUG] not using binary driver name, it's no longer needed
2021/04/27 04:03:28 [DEBUG] not using binary driver name, it's no longer needed
```

once fixed this should be good to merge!

neil-yechenwei:
@katbyte , I assume it's your test env issue. Please cancel the terms first in your tes tenv before run test case.

katbyte (approved):
Looks like it's an ongoing problem on master to, these tests always fail in TC - should look into a way to fix them so they don't fail anymore but it doesn't block this pr 👍

PR #8633, opened 2020-09-25
Do not recreate ASE when changing tags

favoretti:
Currently changing ASE tags force creation of a new resource, which is completely uncalled for.


tombuildsstuff:
hey @favoretti 

Thanks for opening this issue.

Unfortunately at this time the App Service Environment API doesn't support updating tags - [the PATCH resource in the SDK doesn't expose that field](https://github.com/Azure/azure-sdk-for-go/blob/master/services/web/mgmt/2019-08-01/web/models.go#L2576-L2587) - which is why this is the case.

From memory we raised an API bug for this - but I can't find it at the moment - however since this field isn't exposed unfortunately making this field `ForceNew: false` won't do anything (since the field can't be updated) - and as such I'm going to close this PR for the moment. Once support for updating tags is available however (by being available in the Patch struct, we should be able to update this however.

Thanks!

PR #8641, opened 2020-09-27
Update `address_space` attribute description

fcatacut:
The output of the `address_space` attribute is a list of strings as documented in its corresponding data block: https://www.terraform.io/docs/providers/azurerm/d/virtual_network.html#address_space.


katbyte (approved):
Thanks for this fix @fcatacut - LGTM 👍

PR #8643, opened 2020-09-27
Add default value for attached property of data source public ips

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/8476

### DIFF
--- a/azurerm/internal/services/network/public_ips_data_source.go
+++ b/azurerm/internal/services/network/public_ips_data_source.go
@@ -33,6 +33,7 @@ func dataSourceArmPublicIPs() *schema.Resource {
 			"attached": {
 				Type:     schema.TypeBool,
 				Optional: true,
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
hi @neil-yechenwei 

The default value for a boolean will be false - so this'll already be the case here - and as such whilst I'd like to thank you for this contribution, since this appears superfluous I'm going to close this PR for the moment.

Thanks!

mloskot:
I think @neil-yechenwei aimed to fix the confusion that I experienced while reading the docs and using the Terraform provider, as I explained in the https://github.com/terraform-providers/terraform-provider-azurerm/issues/8476

> I'd read the docs, but it indicates the attached argument is optional.
> I'd assumed that if I don't specify it, I should get any/all non-empty IPs.

PR #8651, opened 2020-09-28
azurerm_container_group: change `nameservers` type to list to keep order

magodo:
The DNS name server order is honored by Azure, and should be kept from
UI. Since some users might define the name servers order explicitly.


tombuildsstuff (approved):
LGTM 👍

PR #8655, opened 2020-09-28
Update doc for attached property of azurerm_public_ips

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/8476


tombuildsstuff:
hi @neil-yechenwei 

Thanks for this PR.

Taking a look through here since the default value for a boolean would be `false` we intentionally don't document this on every field as it'd be superfluous - as such whilst I'd like to thank you for this contribution I'm going to close this PR for the moment.

Thanks!

PR #8664, opened 2020-09-28
Typo fix

nexxai:
This line is for naming the event hub itself, not the namespace.


katbyte (approved):
Thanks @nexxai - LGTM 👍

PR #8674, opened 2020-09-29
new resource: azurerm_key_vault_certificate_contact

### LINKED ISSUE #1301: Specifying Email Contacts for azurerm_key_vault_certificate
### Community Note

### Description

Specifying email contacts to be updated in Azure Key vault. In the resource 'azurerm_key_vault_certificate', we can only specify action_type as 'EmailContacts' but we cannot specify the email addresses to update email contacts for a specific key vault

* azurerm_key_vault_certificate

njuCZ:
fix #1301

image

### DIFF
--- a/azurerm/internal/services/keyvault/key_vault_certificate_contact_resource.go
+++ b/azurerm/internal/services/keyvault/key_vault_certificate_contact_resource.go
@@ -1,0 +1,253 @@
+package keyvault
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/keyvault/2016-10-01/keyvault"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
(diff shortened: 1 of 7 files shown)


jackofallops:
Hi @njuCZ - Looking at the previous discussion on this issue, I'm not convinced that a separate resource is the correct way to go here. The extra permission in the `access_policy` on the Key Vault is still required in both implementations for the account making the request, so it makes more sense in the Key Vault resource itself since this could be considered optional base configuration for that resource? (And we can add the permission requirement information to the documentation for that config block).

njuCZ:
@jackofallops I have considered placing `contacts` as a property of Key vault. But for the read function, it should invoke listing contacts function, if users don't have manage contact permission this api will throw error. I think it's not easy to check whether current account has this permission for the key vault

tombuildsstuff:
Agreed this doesn't make sense as a separate resource - can we merge this into the other resource?

njuCZ:
@jackofallops @tombuildsstuff sorry for late reply, do you think it's ok to make it as a property of resource `azurerm_key_vault_certificate` rather than `azurerm_key_vault` ? 
In some cases, users would use resource "azurerm_key_vault_access_policy", if "contacts" is a property of "azurerm_key_vault", users could not create them at one time, they have to create "azurerm_key_vault" and "azurerm_key_vault_access_policy" first and then update "azurerm_key_vault"

PR #8682, opened 2020-09-29
[azurerm_function_app] App Settings should be Computed

nexxai:
Function Apps should have computed App Settings just like App Services do, since they are effectively two different offshoots of the same resource type behind the scenes.

Also required as discussed in [this thread](https://github.com/hashicorp/terraform/issues/26401#issuecomment-700847694)


katbyte (approved):
Thanks @nexxai - LGTM 👍

drdamour:
@nexxai  does this fix https://github.com/terraform-providers/terraform-provider-azurerm/issues/5112 too ?

nexxai:
@drdamour No, you will still need to add the fields you wish Terraform to ignore.

PR #8706, opened 2020-10-01
Scaffold azurerm_app_service_environment data docs

hfurubotten:
When looking up the data source for `azurerm_app_service_environment` I noticed that the example contained non-functioning example. Hopefully I can fix the documentation issue with this while I still have it fresh in memory. 

This PR has a new scaffold of the documentation and added back most of the existing documentation for attributes, where those are not autogenerated.


tombuildsstuff (approved):
LGTM - thanks for this @hfurubotten

PR #8713, opened 2020-10-01
Docs: Replace Basic with GeneralPurpose for sku_name

mloskot:
For consistency, since PostgreSQL Virtual Network Rules can only be used
with SKU Tiers of GeneralPurpose or MemoryOptimized.


mbfrahry (approved):
LGTM!

PR #8720, opened 2020-10-02
Network policy calico fix

borkod:
I noticed that the sample for AKS with calico policy was not actually implementing calico network policy.  I fixed the sample.


tombuildsstuff (approved):
LGTM - thanks for this @borkod

PR #8722, opened 2020-10-02
Example of AKS cluster with virtual nodes (Azure Container Instances)

borkod:
This example is used to demonstrate how to provision an AKS cluster that uses virtual nodes (Azure Container Instances).


katbyte (approved):
Thanks @borkod - LGTM 👍

PR #8724, opened 2020-10-02
Example for deploying a Python based Azure Functions App.

borkod:
This example is used to demonstrate how to provision an python functions app. It also contains sample python app that can be deployed.


katbyte (approved):
Thanks @borkod - LGTM 👍

PR #8729, opened 2020-10-02
UPDATE azurerm_subnet documentation example

HappyTobi:
Hi,

just a small PR but if anyone will read the documentation and the examples, they will always use the deprecated property.

All examples should be changed to the new property 
`address_prefix` to `adress_prefixed` 

Br
Tobi


katbyte (approved):
Thanks @HappyTobi! LGTM 🙂

PR #8734, opened 2020-10-03
Documentation updates

nexxai:
Spent some time going through all of the documentation and tried to find as many spelling errors as I could.  I *think* this should be most, if not all.


nexxai:
I don't believe the check is failing due to my changes.

katbyte (approved):
This is awesome @nexxai, Thanks!! LGTM 👍

PR #8747, opened 2020-10-05
sku block has been deprecated

scott1138:
sku block has been deprecated - updated example code


katbyte (approved):
Thanks @scott1138! - LGTM 👍

PR #8759, opened 2020-10-06
Update application_gateway.html.markdown

marcin-dudek:
Add information about Multi site listener type.


katbyte (approved):
Thanks @marcin-dudek  - LGTM 👍

PR #8769, opened 2020-10-06
`azurerm_mssql_server` - update doc and test for sql server min tls

jackofallops:
`minimum_tls_version` setting cannot be removed once set, updates testing config for this and adds note to docs.

### DIFF
--- a/website/docs/r/mssql_server.html.markdown
+++ b/website/docs/r/mssql_server.html.markdown
@@ -82,6 +82,8 @@ The following arguments are supported:
 
 * `minimum_tls_version` - (Optional) The Minimum TLS Version for all SQL Database and SQL Data Warehouse databases associated with the server. Valid values are: `1.0`, `1.1` and `1.2`.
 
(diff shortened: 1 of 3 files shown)


website/docs/r/mssql_server.html.markdown
```diff
@@ -82,6 +82,8 @@ The following arguments are supported:
 
 * `minimum_tls_version` - (Optional) The Minimum TLS Version for all SQL Database and SQL Data Warehouse databases associated with the server. Valid values are: `1.0`, `1.1` and `1.2`.
 
+~> **NOTE:** Once `minimum_tls_version` is set it is not possible to remove this setting and must be given a valid value for any further updates to the resource.
```
mbfrahry, line 85: Is it worth writing a `CustomizeDiff` that checks if the value is going from filled to empty and erroring if so?

mbfrahry (approved):
LGTM with a suggestion should you choose to accept it

website/docs/r/mssql_server.html.markdown
jackofallops: Good call - I'll add that too.

PR #8771, opened 2020-10-06
fix: Change serivce to service in private_endpoint.html.markdown

sdib:
Typo correction in private_endpoint.html.markdown


tombuildsstuff (approved):
LGTM - thanks for this @sdib

PR #8773, opened 2020-10-06
New data source: azurerm_cognitive_account

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/7765

Note that I do not use this resource and as such have not tested this functionality manually.  Instead I am relying on the test coverage, and using the existing resource code, modified for the data source.  If anyone uses this resource and sees anything missing in the data source that they expect to retrieve, let me know.


website/docs/d/cognitive_account.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+The following attributes are exported:
+
+* `location` - The Azure location where the Cognitive Services Account exists
+
+* `kind` - The kind of the Cognitive Services Account
+
+* `sku_name` - The sku name of the Cognitive Services Account
+
+* `endpoint` - The endpoint of the Cognitive Services Account
+
```
katbyte, line 44:
```suggestion
* `endpoint` - The endpoint of the Cognitive Services Account

* `qna_runtime_endpoint` - If `kind` is `QnAMaker` the link to the QNA runtime.
```

katbyte: Missing a property here

katbyte (changes requested):
Hey @Lucretius  - thanks for this pr! overall looks great but there is a failing test due to an invalid tf config.

Once thats fixed up this should be good to go!

Lucretius:
Hi @katbyte thanks for the PR review.  I wanted to ask about the failing test, particularly how I can tell I've got a failing test.  I know on this and another PR I've got open the latest builds that have run both seem to have a passing checkmarks in TravisCI for the `make test` stage (and every other stage) - which is what seems to be running the acceptance test suite.  This makes me think that the tests have all passed, but apparently this is not the case.

I am wondering if there is somewhere else I should be seeing these test failures?  I don't have an Azure account that I am using to run acceptance tests against, so I am depending on the CI tooling to do so.  If there is a place I can see these failing tests, I will make sure to look there to save you and the team some time :)

katbyte:
ahh @Lucretius - you will need an azure account to run things against and setup a service principal. The CI tooling doesn't run any acceptance tests and we don't expose our CI system for them to the public as it costs money to run most tests. Once you have that setup you can run them by `make testacc TEST=./azurerm/internal/services/cognitive/tests TESTARGS="-parallel 11 -test.run=TestAccAzureRMPrefixHere"`

Lucretius:
Oh, that would explain why I keep missing tests.  I've got a personal Azure account, I will have to go and see about setting it up for these tests then.  Thank you for explaining that to me!

katbyte:
no bother 🙂 I have a .sh file thati source to set my env vars for it:
```
       │ File: /Users/kt/hashi/sync/set.az.rm.sh
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ #!/bin/bash
   2   │
   3   │ export ARM_TENANT_ID=1234567
   4   │ export ARM_SUBSCRIPTION_ID=1234567
   5   │ export ARM_CLIENT_ID=1234567
   6   │ export ARM_CLIENT_SECRET=""
   7   │ export ARM_ENVIRONMENT=public
   8   │ export ARM_TEST_LOCATION=westeurope
   9   │ export ARM_TEST_LOCATION_ALT=eastus2
  10   │ export ARM_TEST_LOCATION_ALT2=westus
```
which makes it easy to setup for tests in a new shell window

Lucretius:
I ran the failing tests locally and got things working.  It looks like tflint is now run via Github Actions and a bunch of stuff breaks on that step, but should be unrelated to my changes.

jackofallops:
Hi @Lucretius - there's noise in the linter that will be be resolved by a recent release of the linter, there's a genuine error message in there:
```
==> Checking acceptance test terraform blocks are formatted...
./azurerm/internal/services/cognitive/tests/cognitive_account_data_source_test.go:35
```

Lucretius:
Thanks for hanging with me on that, I think my editor must be auto-formatting the blocks - all I had done was delete a line, did not expect it to cause the linter to fail.

PR #8784, opened 2020-10-07
Update data_factory_dataset_http.html.markdown

burakince:
Fix azurerm_data_factory_dataset_http example resource name and required field key


katbyte (approved):
Thanks @burakince - LGTM 👍

PR #8795, opened 2020-10-08
Validate policy mode value

### LINKED ISSUE #8748: azurerm_policy_definition mode is case sensitive
### Community Note

### Affected Resource(s)

* `azurerm_policy_definition `

```hcl
resource "azurerm_policy_definition" "key_vault_soft_delete" {
  name                  = "Key-Vault-Soft-Delete"
  policy_type           = "Custom"
  mode                  = "indexed"
  display_name          = "Key-Vault-Soft-Delete"
  management_group_name = var.definition_management_group
  policy_rule           = file("${path.module}/policies/keyVaultSoftDelete/policyRule.json")
}
```

### Expected Behavior

When this snippet is applied, a policy definition is created as one would expect. A "terraform plan" after apply should yield that no changes are required.  

### Actual Behavior

However, a "terraform plan" after "terraform apply" yields the need to update the resource.  

The problem is a case sensitive comparison on the 'mode' property and that mode "indexed" is stored as "Indexed" during apply. 

### Steps to Reproduce

1. `terraform apply` with the snippet
2. `terraform plan`

### References

* #0000

beandrad:
Policy creation succeeds even if the policy mode in the request is
not capitalized, however, when fetching the policy mode value from
Azure, the correct capitalized version is returned, which causes
the Terraform state and configuration to be different.

Ideally, there should be an enum with a set of allowed values, but
thi is not the case in the latest version of the Azure Rest API specs.
However, the supported mode values are returned in the create error
response when an invalid mode is requested.

Fixes #8748

### DIFF
--- a/azurerm/internal/services/policy/policy_definition_resource.go
+++ b/azurerm/internal/services/policy/policy_definition_resource.go
@@ -63,6 +63,20 @@ func resourceArmPolicyDefinition() *schema.Resource {
 			"mode": {
 				Type:     schema.TypeString,
 				Required: true,
+				ValidateFunc: validation.StringInSlice(
+					[]string{
+						"All",
+						"Indexed",
+						"Microsoft.ContainerService.Data",
+						"Microsoft.CustomerLockbox.Data",
+						"Microsoft.DataCatalog.Data",
+						"Microsoft.KeyVault.Data",
+						"Microsoft.Kubernetes.Data",
+						"Microsoft.MachineLearningServices.Data",
+						"Microsoft.Network.Data",
+						"Microsoft.Synapse.Data",
+					}, false,
+				),
 			},
 
 			"management_group_id": {
--- a/website/docs/r/policy_definition.html.markdown
+++ b/website/docs/r/policy_definition.html.markdown
@@ -70,8 +70,7 @@ The following arguments are supported:
 * `policy_type` - (Required) The policy type. Possible values are `BuiltIn`, `Custom` and `NotSpecified`. Changing this forces a new resource to be created.
 
 * `mode` - (Required) The policy mode that allows you to specify which resource
-    types will be evaluated.  The value can be "All", "Indexed" or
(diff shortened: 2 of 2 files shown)


beandrad:
Screenshot 2020-10-08 at 12 11 25

azurerm/internal/services/policy/policy_definition_resource.go
```diff
@@ -63,6 +63,20 @@ func resourceArmPolicyDefinition() *schema.Resource {
...
+						"Indexed",
+						"Microsoft.ContainerService.Data",
+						"Microsoft.CustomerLockbox.Data",
+						"Microsoft.DataCatalog.Data",
+						"Microsoft.KeyVault.Data",
+						"Microsoft.Kubernetes.Data",
+						"Microsoft.MachineLearningServices.Data",
+						"Microsoft.Network.Data",
+						"Microsoft.Synapse.Data",
+					}, false,
```
tombuildsstuff, line 78: could we also update the values used in the docs here: https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/policy_definition.html.markdown#L72-L74

```
* `mode` - (Required) The policy mode that allows you to specify which resource types will be evaluated. Possible values are `All`, `Indexed`, `Microsoft.ContainerService.Data`, `Microsoft.CustomerLockbox.Data`, `Microsoft.DataCatalog.Data`, `Microsoft.KeyVault.Data`, `Microsoft.Kubernetes.Data`, `Microsoft.MachineLearningServices.Data`, `Microsoft.Network.Data` and `Microsoft.Synapse.Data`.
```

tombuildsstuff (approved):
hey @beandrad 

Thanks for this PR - if we can also update the docs then this otherwise LGTM 👍 

Thanks!

Nepomuceno:
I do think that a DiffSupress Would be better for this them forcing the values.

PR #8815, opened 2020-10-09
Update the value property in features block as TypeSet for SignalR Service

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7761


katbyte (approved):

PR #8829, opened 2020-10-10
Add timezone logic app recurrence

### LINKED ISSUE #4316: Support for startTime and timeZone in azurerm_logic_app_trigger_recurrence
### Community Note

### Description

In the light of the Azure Scheduler retirement and migration to Logic Apps, it would be great to have support for `startTime` (more importantly) and `timeZone` (not that important, but why not).

* azurerm_logic_app_trigger_recurrence

```hcl
resource "azurerm_logic_app_trigger_recurrence" "test" {
  name         = "run-every-day"
  logic_app_id = "${azurerm_logic_app_workflow.test.id}"
  frequency    = "Day"
  interval     = 1
  start_time   = "..."
  time_zone    = "..."
}
```

Lucretius:
Resolves the other part of https://github.com/terraform-providers/terraform-provider-azurerm/issues/4316

According to [this documentation](https://docs.microsoft.com/en-us/azure/connectors/connectors-native-recurrence) `time_zone` is only applicable during `start_time`.

The list of time zones is not provided exactly.  [This documentation](https://docs.microsoft.com/en-us/azure/logic-apps/workflow-definition-language-functions-reference) seems to indicate its the list I posted in the code - though its also unclear whether or not punctuation needs to be removed.

Fixes #4316


azurerm/internal/services/logic/tests/logic_app_trigger_recurrence_resource_test.go
```diff
@@ -206,6 +206,27 @@ func TestAccAzureRMLogicAppTriggerRecurrence_startTime(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMLogicAppTriggerRecurrence_startTimeWithTimeZone(data, "2020-01-01T01:02:03Z", "U.S. Eastern Standard Time"),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMLogicAppTriggerExists(data.ResourceName),
+					resource.TestCheckResourceAttr(data.ResourceName, "start_time", "2020-01-01T01:02:03Z"),
+					resource.TestCheckResourceAttr(data.ResourceName, "time_zone", "U.S. Eastern Standard Time"),
+				),
+			},
+			data.ImportStep(),
```
tombuildsstuff, line 225: since it's possible to update this field could we also include a second configuration/import step here to ensure this can be updated?

azurerm/internal/services/logic/logic_app_trigger_recurrence_resource.go
```diff
@@ -65,6 +65,12 @@ func resourceArmLogicAppTriggerRecurrence() *schema.Resource {
 				Optional:     true,
 				ValidateFunc: validation.IsRFC3339Time,
 			},
+
+			"time_zone": {
+				Type:         schema.TypeString,
+				Optional:     true,
```
tombuildsstuff, line 71: the update test should confirm this, but I assume this defaults to UTC on the server-side?

tombuildsstuff (dismissed):
hey @Lucretius 

Thanks for this PR :)

Taking a look through this mostly LGTM - if we can update the test to confirm that it's possible to update the timezone here then this otherwise LGTM 👍 

Thanks!

azurerm/internal/services/logic/logic_app_trigger_recurrence_resource.go
Lucretius: The documentation seems to indicate this - I've gone and set the default `time_zone` value to UTC and updated an old test which set just the `start_time` to verify that `UTC` is set as the default.

Lucretius:
Alright, UTC has been set as the default for `time_zone`, and an old test which just sets the `start_time` has been updated to verify.  I also updated the other test to ensure update behavior works.

katbyte (dismissed):
Hey @Lucretius,

Overall this looks good, but there is some failing tests:

```
TestAccAzureRMLogicAppTriggerRecurrence_hour: testing.go:684: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=9) "time_zone": (string) (len=3) "UTC"
        }

```
looks like you'll need to change it from default to computed?

Lucretius:
Swapping the property to computed seemed to fix the test locally.  I also noticed that the timezones did not seem to like punctuation (the docs seemed to hint at this) so the timezone list now does not contain any punctuation (U.S. East -> US East).  All tests were run locally and passed.

jackofallops:
Tests passing:
image

jackofallops (approved):
Thanks @Lucretius - LGMT now the tests are fixed up 👍

PR #8830, opened 2020-10-10
add gateway_resource_id property

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/8537

The resource ID is of an on premises data gateway, which apparently there is no Terraform resource for, so acceptance testing wise, I don't think it's possible to write one for this property.  I'm open to any advice on handling this.

### DIFF
--- a/azurerm/internal/services/analysisservices/analysis_services_server_resource.go
+++ b/azurerm/internal/services/analysisservices/analysis_services_server_resource.go
@@ -72,6 +72,12 @@ func resourceArmAnalysisServicesServer() *schema.Resource {
 				}, false),
 			},
 
+			"gateway_resource_id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: azure.ValidateResourceID,
+			},
+
 			"admin_users": {
 				Type:     schema.TypeSet,
 				Optional: true,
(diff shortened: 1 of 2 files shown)


jackofallops (changes requested):
Hi @Lucretius 
Thanks for this PR. 

I suspect that the new property will not work as expected as there is a separate call to disassociate the Gateway from the server if the value is removed from config. The Update API call is a patch, so will not remove the Gateway if not included. Could you add a test to cover creating a service with a gateway and subsequently removing it to prove this out?

Thanks.

Lucretius:
Hi @jackofallops 

from my understanding, the gateway that is being requested is not an available terraform resource since its an [on-premises gateway](https://docs.microsoft.com/en-us/data-integration/gateway/service-gateway-onprem).  I do see what you mean with the gateway disassociation - perhaps given the limitations regarding testing, maybe we should just set the property to `ForceNew: true` to avoid worrying about the disassociation, and if we get a terraform resource we can modify with some tests to assure functionality?

jackofallops:
That's unfortunate.  I think `ForceNew` is the wrong way to go though, it's an updatable field in a resource that otherwise supports delta changes.  I think it should be implemented using the disassociate API still.  Is this something you are able to manually test in your environment and post the results here? (sensitive values redacted, ofc)

Lucretius:
I don't actually use this resource myself, just seemed like an easy thing to add into the code as it was just a single property but now I see its more nuanced than that.  Best I could do is contact the issue author to see if we could coordinate on testing, or perhaps let them make a PR.  Given that this property is for an on-prem gateway I don't actually see how we can ever automatically test it in Terraform - it sounds like we would need an idle VM in the test environment running the gateway software which must have an ID that then gets linked to the Azure test account which would then reference.

What do you think?

jackofallops:
Hi @Lucretius - If a property/feature isn't testable within the test framework we can sometimes accept the output of manual testing (typically if the property is important to the use of the associated resource if there's no alternative). It's not ideal, but sometimes it is necessary.  If the OP of the issue is able and happy to collaborate with you and together you can cover the update mechanism and provide output (such as `plan`s and `apply`s, and possibly screenshots from the portal, with anything sensitive redacted) it may be we can accept it. I'll check with the rest of the team also, it may be one of them has more experience than I with this service.

PR #8831, opened 2020-10-10
Deprecate unused resource_group_name property for azurerm_lb_*

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/8777

### DIFF
--- a/azurerm/helpers/azure/resource_group.go
+++ b/azurerm/helpers/azure/resource_group.go
@@ -27,6 +27,16 @@ func SchemaResourceGroupNameDeprecated() *schema.Schema {
 	}
 }
 
+func SchemaResourceGroupNameForceNewDeprecated() *schema.Schema {
+	return &schema.Schema{
+		Type:         schema.TypeString,
+		Optional:     true,
+		ForceNew:     true,
+		ValidateFunc: validateResourceGroupName,
+		Deprecated:   "This field is no longer used and will be removed in the next major version of the Azure Provider",
+	}
+}
+
 func SchemaResourceGroupNameDiffSuppress() *schema.Schema {
 	return &schema.Schema{
 		Type:             schema.TypeString,
--- a/website/docs/r/lb_backend_address_pool.html.markdown
+++ b/website/docs/r/lb_backend_address_pool.html.markdown
@@ -51,7 +51,7 @@ The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the Backend Address Pool.
 * `loadbalancer_id` - (Required) The ID of the Load Balancer in which to create the Backend Address Pool.
-* `resource_group_name` - (Optional / **Deprecated**) The name of the resource group in which to create the resource.
+* `resource_group_name` - (Optional / **Deprecated**) The name of the resource group in which to create the resource. Changing this forces a new resource to be created.
 
 ## Attributes Reference
 
--- a/website/docs/r/lb_backend_address_pool.html.markdown
+++ b/website/docs/r/lb_backend_address_pool.html.markdown
@@ -50,8 +50,8 @@ resource "azurerm_lb_backend_address_pool" "example" {
 The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the Backend Address Pool.
-* `resource_group_name` - (Required) The name of the resource group in which to create the resource.
 * `loadbalancer_id` - (Required) The ID of the Load Balancer in which to create the Backend Address Pool.
+* `resource_group_name` - (Optional / **Deprecated**) The name of the resource group in which to create the resource.
 
 ## Attributes Reference
 
--- a/website/docs/r/lb_nat_pool.html.markdown
+++ b/website/docs/r/lb_nat_pool.html.markdown
@@ -57,13 +57,13 @@ resource "azurerm_lb_nat_pool" "example" {
 The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the NAT pool.
-* `resource_group_name` - (Required) The name of the resource group in which to create the resource.
 * `loadbalancer_id` - (Required) The ID of the Load Balancer in which to create the NAT pool.
 * `frontend_ip_configuration_name` - (Required) The name of the frontend IP configuration exposing this rule.
 * `protocol` - (Required) The transport protocol for the external endpoint. Possible values are `Udp` or `Tcp`.
 * `frontend_port_start` - (Required) The first port number in the range of external ports that will be used to provide Inbound Nat to NICs associated with this Load Balancer. Possible values range between 1 and 65534, inclusive.
 * `frontend_port_end` - (Required) The last port number in the range of external ports that will be used to provide Inbound Nat to NICs associated with this Load Balancer. Possible values range between 1 and 65534, inclusive.
 * `backend_port` - (Required) The port used for the internal endpoint. Possible values range between 1 and 65535, inclusive.
+* `resource_group_name` - (Optional / **Deprecated**) The name of the resource group in which to create the resource.
 
 ## Attributes Reference
 
--- a/website/docs/r/lb_nat_rule.html.markdown
+++ b/website/docs/r/lb_nat_rule.html.markdown
@@ -56,7 +56,6 @@ resource "azurerm_lb_nat_rule" "example" {
 The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the NAT Rule.
-* `resource_group_name` - (Required) The name of the resource group in which to create the resource.
 * `loadbalancer_id` - (Required) The ID of the Load Balancer in which to create the NAT Rule.
 * `frontend_ip_configuration_name` - (Required) The name of the frontend IP configuration exposing this rule.
 * `protocol` - (Required) The transport protocol for the external endpoint. Possible values are `Udp`, `Tcp` or `All`.
@@ -65,6 +64,7 @@ The following arguments are supported:
 * `idle_timeout_in_minutes` - (Optional) Specifies the idle timeout in minutes for TCP connections. Valid values are between `4` and `30` minutes. Defaults to `4` minutes.
 * `enable_floating_ip` - (Optional) Are the Floating IPs enabled for this Load Balncer Rule? A "floating” IP is reassigned to a secondary server in case the primary server fails. Required to configure a SQL AlwaysOn Availability Group. Defaults to `false`.
 * `enable_tcp_reset` - (Optional) Is TCP Reset enabled for this Load Balancer Rule? Defaults to `false`.
+* `resource_group_name` - (Optional / **Deprecated**) The name of the resource group in which to create the resource.
 
 ## Attributes Reference
 
--- a/website/docs/r/lb_outbound_rule.html.markdown
+++ b/website/docs/r/lb_outbound_rule.html.markdown
@@ -62,7 +62,6 @@ resource "azurerm_lb_outbound_rule" "example" {
 The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the Outbound Rule. Changing this forces a new resource to be created.
-* `resource_group_name` - (Required) The name of the resource group in which to create the resource. Changing this forces a new resource to be created.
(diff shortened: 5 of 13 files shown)


DenWin:
I admit, my issue pointed to this solution, though in view of the big scheme wouldn't ResGroup + Name more like all other managed resources? Rather than just the ID.

neil-yechenwei:
Currently resource group name would be extracted from load balancer id. So it doesn't need resource_group_name.

DenWin:
I'm aware about this.
As I understand the behavior of all other managed resources, there would not be a resource id, but just ResGroup + LB Name.

neil-yechenwei:
Nope. I assume they're not managed resources. These resources are based on existing resource group and load balancer to perform write operation.

neil-yechenwei:
Nope. I assume it's not managed resource. So it's based on existing resource group and load balancer to perform write/read operation.

website/docs/r/lb_backend_address_pool.html.markdown
```diff
@@ -50,8 +50,8 @@ resource "azurerm_lb_backend_address_pool" "example" {
 The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the Backend Address Pool.
-* `resource_group_name` - (Required) The name of the resource group in which to create the resource.
 * `loadbalancer_id` - (Required) The ID of the Load Balancer in which to create the Backend Address Pool.
+* `resource_group_name` - (Optional / **Deprecated**) The name of the resource group in which to create the resource. Changing this forces a new resource to be created.
```
tombuildsstuff, line 54: which means this is no longer forcenew

website/docs/r/lb_nat_pool.html.markdown
```diff
@@ -57,13 +57,13 @@ resource "azurerm_lb_nat_pool" "example" {
...
 
 * `name` - (Required) Specifies the name of the NAT pool.
-* `resource_group_name` - (Required) The name of the resource group in which to create the resource.
 * `loadbalancer_id` - (Required) The ID of the Load Balancer in which to create the NAT pool.
 * `frontend_ip_configuration_name` - (Required) The name of the frontend IP configuration exposing this rule.
 * `protocol` - (Required) The transport protocol for the external endpoint. Possible values are `Udp` or `Tcp`.
 * `frontend_port_start` - (Required) The first port number in the range of external ports that will be used to provide Inbound Nat to NICs associated with this Load Balancer. Possible values range between 1 and 65534, inclusive.
 * `frontend_port_end` - (Required) The last port number in the range of external ports that will be used to provide Inbound Nat to NICs associated with this Load Balancer. Possible values range between 1 and 65534, inclusive.
 * `backend_port` - (Required) The port used for the internal endpoint. Possible values range between 1 and 65535, inclusive.
+* `resource_group_name` - (Optional / **Deprecated**) The name of the resource group in which to create the resource.
```
tombuildsstuff, line 66: which means this is no longer forcenew

website/docs/r/lb_nat_rule.html.markdown
```diff
@@ -65,6 +64,7 @@ The following arguments are supported:
 * `idle_timeout_in_minutes` - (Optional) Specifies the idle timeout in minutes for TCP connections. Valid values are between `4` and `30` minutes. Defaults to `4` minutes.
 * `enable_floating_ip` - (Optional) Are the Floating IPs enabled for this Load Balncer Rule? A "floating” IP is reassigned to a secondary server in case the primary server fails. Required to configure a SQL AlwaysOn Availability Group. Defaults to `false`.
 * `enable_tcp_reset` - (Optional) Is TCP Reset enabled for this Load Balancer Rule? Defaults to `false`.
+* `resource_group_name` - (Optional / **Deprecated**) The name of the resource group in which to create the resource.
```
tombuildsstuff, line 67: which means this is no longer forcenew

website/docs/r/lb_outbound_rule.html.markdown
```diff
@@ -62,14 +62,14 @@ resource "azurerm_lb_outbound_rule" "example" {
...
-* `resource_group_name` - (Required) The name of the resource group in which to create the resource. Changing this forces a new resource to be created.
 * `loadbalancer_id` - (Required) The ID of the Load Balancer in which to create the Outbound Rule. Changing this forces a new resource to be created.
 * `frontend_ip_configuration` - (Required) One or more `frontend_ip_configuration` blocks as defined below.
 * `backend_address_pool_id` - (Required) The ID of the Backend Address Pool. Outbound traffic is randomly load balanced across IPs in the backend IPs.
 * `protocol` - (Required) The transport protocol for the external endpoint. Possible values are `Udp`, `Tcp` or `All`.
 * `enable_tcp_reset` - (Optional) Receive bidirectional TCP Reset on TCP flow idle timeout or unexpected connection termination. This element is only used when the protocol is set to TCP.
 * `allocated_outbound_ports` - (Optional) The number of outbound ports to be used for NAT.
 * `idle_timeout_in_minutes` - (Optional) The timeout for the TCP idle connection
+* `resource_group_name` - (Optional / **Deprecated**) The name of the resource group in which to create the resource.
```
tombuildsstuff, line 72: which means this is no longer forcenew

website/docs/r/lb_probe.html.markdown
```diff
@@ -51,11 +51,11 @@ resource "azurerm_lb_probe" "example" {
 The following arguments are supported:
 
 * `name` - (Required) Specifies the name of the Probe.
-* `resource_group_name` - (Required) The name of the resource group in which to create the resource.
 * `loadbalancer_id` - (Required) The ID of the LoadBalancer in which to create the NAT Rule.
 * `protocol` - (Optional) Specifies the protocol of the end point. Possible values are `Http`, `Https` or `Tcp`. If Tcp is specified, a received ACK is required for the probe to be successful. If Http is specified, a 200 OK response from the specified URI is required for the probe to be successful.
 * `port` - (Required) Port on which the Probe queries the backend endpoint. Possible values range from 1 to 65535, inclusive.
 * `request_path` - (Optional) The URI used for requesting health status from the backend endpoint. Required if protocol is set to `Http` or `Https`. Otherwise, it is not allowed.
+* `resource_group_name` - (Optional / **Deprecated**) The name of the resource group in which to create the resource.
```
tombuildsstuff, line 58: which means this is no longer forcenew

website/docs/r/lb_rule.html.markdown
```diff
@@ -67,6 +66,7 @@ The following arguments are supported:
 * `load_distribution` - (Optional) Specifies the load balancing distribution type to be used by the Load Balancer. Possible values are: `Default` – The load balancer is configured to use a 5 tuple hash to map traffic to available servers. `SourceIP` – The load balancer is configured to use a 2 tuple hash to map traffic to available servers. `SourceIPProtocol` – The load balancer is configured to use a 3 tuple hash to map traffic to available servers. Also known as Session Persistence, where  the options are called `None`, `Client IP` and `Client IP and Protocol` respectively.
 * `disable_outbound_snat` - (Optional) Is snat enabled for this Load Balancer Rule? Default `false`.
 * `enable_tcp_reset` - (Optional) Is TCP Reset enabled for this Load Balancer Rule? Defaults to `false`.
+* `resource_group_name` - (Optional / **Deprecated**) The name of the resource group in which to create the resource.
```
tombuildsstuff, line 69: which means this is no longer forcenew

tombuildsstuff (changes requested):
hi @neil-yechenwei 

Thanks for this PR.

Taking a look through here it appears we need to make this field both Optional and Computed, rather than ForceNew - else there's no means for users to remove this unnecessary field without recreating the resource.

Thanks!

azurerm/helpers/azure/resource_group.go
```diff
@@ -27,6 +27,16 @@ func SchemaResourceGroupNameDeprecated() *schema.Schema {
 	}
 }
 
+func SchemaResourceGroupNameForceNewDeprecated() *schema.Schema {
+	return &schema.Schema{
+		Type:         schema.TypeString,
+		Optional:     true,
+		ForceNew:     true,
```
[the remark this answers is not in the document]
neil-yechenwei: This function "SchemaResourceGroupNameForceNewDeprecated" is only for lb_backend_address_pool_resource.go since [this resource](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/network/lb_backend_address_pool_resource.go#L23) doesn't include "Update" in schema. So I have to keep forcenew here for this resource.

website/docs/r/lb_backend_address_pool.html.markdown
neil-yechenwei: This function "SchemaResourceGroupNameForceNewDeprecated" is only for lb_backend_address_pool_resource.go since [this resource](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/network/lb_backend_address_pool_resource.go#L23) doesn't include "Update" in schema. So I have to keep forcenew here for this resource.

website/docs/r/lb_nat_pool.html.markdown
neil-yechenwei: Here is using [SchemaResourceGroupNameDeprecated() ](https://github.com/terraform-providers/terraform-provider-azurerm/blob/326ae0569f803b4fbaee7d3a40aaa56ee6568ce8/azurerm/helpers/azure/resource_group.go#L21)which doesn't include forcenew.

website/docs/r/lb_nat_rule.html.markdown
neil-yechenwei: Here is using [SchemaResourceGroupNameDeprecated() ](https://github.com/terraform-providers/terraform-provider-azurerm/blob/326ae0569f803b4fbaee7d3a40aaa56ee6568ce8/azurerm/helpers/azure/resource_group.go#L21)which doesn't include forcenew.

website/docs/r/lb_probe.html.markdown
neil-yechenwei: Here is using [SchemaResourceGroupNameDeprecated() ](https://github.com/terraform-providers/terraform-provider-azurerm/blob/326ae0569f803b4fbaee7d3a40aaa56ee6568ce8/azurerm/helpers/azure/resource_group.go#L21)which doesn't include forcenew.

website/docs/r/lb_rule.html.markdown
neil-yechenwei: Here is using [SchemaResourceGroupNameDeprecated() ](https://github.com/terraform-providers/terraform-provider-azurerm/blob/326ae0569f803b4fbaee7d3a40aaa56ee6568ce8/azurerm/helpers/azure/resource_group.go#L21)which doesn't include forcenew.

website/docs/r/lb_probe.html.markdown
neil-yechenwei: Here is using [SchemaResourceGroupNameDeprecated() ](https://github.com/terraform-providers/terraform-provider-azurerm/blob/326ae0569f803b4fbaee7d3a40aaa56ee6568ce8/azurerm/helpers/azure/resource_group.go#L21)which doesn't include forcenew.

neil-yechenwei:
@tombuildsstuff , Thanks for your comment. Seems we have some communication gap. Could you have a look the comment I left? Thanks.

azurerm/helpers/azure/resource_group.go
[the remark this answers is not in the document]
katbyte: I think you are going to have to add an update function, and then make everything but that property force new then.

katbyte (changes requested):
@neil-yechenwei - see my comment inline, i think we'll have to make this resource updateable and mark everything as force new so we can safely remove that property. Otherwise maybe its best to just leave it in until 3.0 where we can remove it?

neil-yechenwei:
@katbyte , thanks for your suggestion. I prefer to close this PR and expect it to be removed from 3.0.

PR #8846, opened 2020-10-12
`azurerm_postgresql_server` `threat_detection_policy` refresh plan is not emtpy 

yupwei68:
Fix https://github.com/terraform-providers/terraform-provider-azurerm/issues/8738

=== RUN   TestAccAzureRMPostgreSQLServer_threatDetection
=== PAUSE TestAccAzureRMPostgreSQLServer_threatDetection
=== CONT  TestAccAzureRMPostgreSQLServer_threatDetection
--- PASS: TestAccAzureRMPostgreSQLServer_threatDetection (237.72s)

### DIFF
--- a/azurerm/internal/services/postgres/postgresql_server_resource.go
+++ b/azurerm/internal/services/postgres/postgresql_server_resource.go
@@ -898,8 +898,8 @@ func flattenSecurityAlertPolicy(props *postgresql.SecurityAlertPolicyProperties,
 
 	block["enabled"] = props.State == postgresql.ServerSecurityAlertPolicyStateEnabled
 
-	block["disabled_alerts"] = utils.FlattenStringSlice(props.DisabledAlerts)
-	block["email_addresses"] = utils.FlattenStringSlice(props.EmailAddresses)
+	block["disabled_alerts"] = utils.FlattenStringSliceRemoveEmptyString(props.DisabledAlerts)
+	block["email_addresses"] = utils.FlattenStringSliceRemoveEmptyString(props.EmailAddresses)
 
 	if v := props.EmailAccountAdmins; v != nil {
 		block["email_account_admins"] = *v
--- a/azurerm/internal/services/postgres/tests/postgresql_server_resource_test.go
+++ b/azurerm/internal/services/postgres/tests/postgresql_server_resource_test.go
@@ -354,6 +354,24 @@ func TestAccAzureRMPostgreSQLServer_updateSKU(t *testing.T) {
 	})
 }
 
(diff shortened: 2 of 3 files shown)


azurerm/internal/services/postgres/postgresql_server_resource.go
```diff
@@ -898,8 +898,8 @@ func flattenSecurityAlertPolicy(props *postgresql.SecurityAlertPolicyProperties,
 
 	block["enabled"] = props.State == postgresql.ServerSecurityAlertPolicyStateEnabled
 
-	block["disabled_alerts"] = utils.FlattenStringSlice(props.DisabledAlerts)
-	block["email_addresses"] = utils.FlattenStringSlice(props.EmailAddresses)
+	block["disabled_alerts"] = utils.FlattenStringSliceRemoveEmptyString(props.DisabledAlerts)
+	block["email_addresses"] = utils.FlattenStringSliceRemoveEmptyString(props.EmailAddresses)
```
tombuildsstuff, line 902: ultimately it looks like this shouldn't be necessary if we update the Expand to not add the element if it's an empty string: https://github.com/terraform-providers/terraform-provider-azurerm/pull/8846/files#diff-6d8a16299e927bb0011aaa89cbb729f9R6-R10 - can we revert this and fix the Expand method instead?

tombuildsstuff (changes requested):
hi @yupwei68 

Thanks for this PR - taking a look through this appears to be a bug in the Expand rather than the Flatten which is the root-cause here - as such rather than adding a hack here can we instead fix the Expand function to not insert an empty string?

Thanks!

yupwei68:
Hi Tom, on fact, we have not sent the empty string in expand, I think the empty string is a service action:
PUT /subscriptions/.../resourceGroups/.../providers/Microsoft.DBforPostgreSQL/servers/.../securityAlertPolicies/Default?api-version=2017-12-01 HTTP/1.1
`{"properties":{"state":"Enabled","disabledAlerts":[],"emailAddresses":[],"emailAccountAdmins":false,"retentionDays":0}}`

GET /subscriptions/.../resourceGroups/.../providers/Microsoft.DBforPostgreSQL/servers/.../securityAlertPolicies/Default?api-version=2017-12-01 HTTP/1.1
`{"properties":{"state":"Enabled","disabledAlerts":[""],"emailAddresses":[""],"emailAccountAdmins":false,"storageEndpoint":"","storageAccountAccessKey":"","retentionDays":0,"creationTime":"2020-10-13T07:51:56.52Z"},"id":"...","name":"Default","type":"Microsoft.DBforPostgreSQL/servers/securityAlertPolicies"}`

yupwei68:
Upstream microsoft issue raised.

PR #8847, opened 2020-10-12
Update waiting logic for azurerm_subnet

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/3149

### DIFF
--- a/azurerm/internal/services/network/subnet_resource.go
+++ b/azurerm/internal/services/network/subnet_resource.go
@@ -1,12 +1,14 @@
 package network
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-05-01/network"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hi @neil-yechenwei 

Thanks for this PR.

Taking a look at the error message being returned in #3149:

> Error waiting for Azure Storage Account "xxxxx" to be created: Future#WaitForCompletion: the number of retries has been exceeded: StatusCode=400 -- Original Error: Code="NetworkAclsValidationFailure" Message="Validation of network acls failure: SubnetsNotProvisioned:Cannot proceed with operation because subnets xxxxx of the virtual network /subscriptions/xxxx/resourceGroups/xxxx/providers/Microsoft.Network/virtualNetworks/xxx are not provisioned. They are in Updating state.."

Based on this error it appears that the Storage API's validate that the API is in a "Ready" state rather than waiting for it to settle, which it should be doing - as such whilst we could look to fix this in Terraform (by adding a lock into the Network, Subnet and any dependent resources which use a Subnet ID) ultimately this is a bug in the Storage API which needs to be fixed (since it should be validating the specific Subnet being used is Ready - rather than the Network, which is unrelated [and gets marked as "updating" whenever other subnets change).

As such whilst I'd like to thank you for this contribution, ultimately this is a bug in the Storage API - would you mind filing a bug on [the Rest API Specs Repository](https://github.com/Azure/azure-rest-api-specs) and then reaching out to the service team here? In the interim it should be possible to work around this by using a `depends_on` to force the ordering here, as you've mentioned in #3149.

Thanks!

PR #8861, opened 2020-10-13
log_analytics_workspace: support for the `daily_quota_gb` property

Lucretius:
Resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/3288

The daily volume cap is off by default as per [the docs](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/manage-cost-storage#daily-cap), so I've set the default value to -1 which means unlimited (no cap).  Validation is set to either be -1 or some non-negative value.

One new test per resource/data source added, and one old test updated to confirm default value.

### DIFF
--- a/azurerm/internal/services/loganalytics/log_analytics_workspace_resource.go
+++ b/azurerm/internal/services/loganalytics/log_analytics_workspace_resource.go
@@ -137,7 +137,7 @@ func resourceArmLogAnalyticsWorkspaceCreateUpdate(d *schema.ResourceData, meta i
 	}
 
 	retentionInDays := int32(d.Get("retention_in_days").(int))
-	dailyQuotaGb := float64(d.Get("daily_quota_gb").(float64))
+	dailyQuotaGb := d.Get("daily_quota_gb").(float64)
 
 	t := d.Get("tags").(map[string]interface{})
 
--- a/azurerm/internal/services/loganalytics/log_analytics_workspace_resource.go
+++ b/azurerm/internal/services/loganalytics/log_analytics_workspace_resource.go
@@ -73,6 +73,13 @@ func resourceArmLogAnalyticsWorkspace() *schema.Resource {
 				ValidateFunc: validation.Any(validation.IntBetween(30, 730), validation.IntInSlice([]int{7})),
 			},
 
+			"daily_quota_gb": {
+				Type:         schema.TypeFloat,
+				Optional:     true,
+				Default:      float64(-1),
+				ValidateFunc: validation.Any(validation.FloatBetween(-1, -1), validation.FloatAtLeast(0)),
+			},
+
 			"workspace_id": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -130,6 +137,7 @@ func resourceArmLogAnalyticsWorkspaceCreateUpdate(d *schema.ResourceData, meta i
 	}
 
 	retentionInDays := int32(d.Get("retention_in_days").(int))
+	dailyQuotaGb := float64(d.Get("daily_quota_gb").(float64))
 
 	t := d.Get("tags").(map[string]interface{})
 
@@ -140,6 +148,9 @@ func resourceArmLogAnalyticsWorkspaceCreateUpdate(d *schema.ResourceData, meta i
 		WorkspaceProperties: &operationalinsights.WorkspaceProperties{
 			Sku:             sku,
 			RetentionInDays: &retentionInDays,
+			WorkspaceCapping: &operationalinsights.WorkspaceCapping{
+				DailyQuotaGb: &dailyQuotaGb,
+			},
 		},
 	}
 
@@ -197,6 +208,11 @@ func resourceArmLogAnalyticsWorkspaceRead(d *schema.ResourceData, meta interface
 		d.Set("sku", sku.Name)
 	}
 	d.Set("retention_in_days", resp.RetentionInDays)
+	if workspaceCapping := resp.WorkspaceCapping; workspaceCapping != nil {
+		d.Set("daily_quota_gb", resp.WorkspaceCapping.DailyQuotaGb)
+	} else {
+		d.Set("daily_quota_gb", utils.Float(-1))
+	}
 
 	sharedKeys, err := sharedKeysClient.GetSharedKeys(ctx, id.ResourceGroup, id.Name)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


azurerm/internal/services/loganalytics/tests/log_analytics_workspace_resource_test.go
```diff
@@ -146,6 +146,27 @@ func TestAccAzureRMLogAnalyticsWorkspace_withDefaultSku(t *testing.T) {
 				Config: testAccAzureRMLogAnalyticsWorkspace_withDefaultSku(data),
 				Check: resource.ComposeTestCheckFunc(
 					testCheckAzureRMLogAnalyticsWorkspaceExists(data.ResourceName),
+					resource.TestCheckResourceAttr(data.ResourceName, "daily_quota_gb", "-1"),
```
magodo, line 149: This check is redundent here as it will be checked in the `data.ImportStep()`.

azurerm/internal/services/loganalytics/tests/log_analytics_workspace_resource_test.go
```diff
@@ -146,6 +146,27 @@ func TestAccAzureRMLogAnalyticsWorkspace_withDefaultSku(t *testing.T) {
...
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMLogAnalyticsWorkspaceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMLogAnalyticsWorkspace_withVolumeCap(data, 4.5),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMLogAnalyticsWorkspaceExists(data.ResourceName),
+					resource.TestCheckResourceAttr(data.ResourceName, "daily_quota_gb", "4.5"),
```
magodo, line 169: This check is redundent here as it will be checked in the `data.ImportStep()`.

azurerm/internal/services/loganalytics/tests/log_analytics_workspace_resource_test.go
```diff
@@ -307,3 +328,29 @@ resource "azurerm_log_analytics_workspace" "test" {
...
+  location = "%s"
+}
+
+resource "azurerm_log_analytics_workspace" "test" {
+  name                = "acctestLAW-%d"
+  location            = azurerm_resource_group.test.location
+  resource_group_name = azurerm_resource_group.test.name
+  sku                 = "PerGB2018"
+  retention_in_days   = 30
+  daily_volume_cap    = %d
```
magodo, line 349: Is this a typo?

```suggestion
  daily_quota_gb    = %d
```

website/docs/r/log_analytics_workspace.html.markdown
```diff
@@ -43,6 +43,8 @@ The following arguments are supported:
 
 * `retention_in_days` - (Optional) The workspace data retention in days. Possible values are either 7 (Free Tier only) or range between 30 and 730.
 
+* `daily_quota_gb` - (Optional) The workspace of the volume cap in gb.  Defaults to -1 (unlimited).
```
magodo, line 46: Can we adjust the statement a bit?

```suggestion
* `daily_quota_gb` - The workspace daily quota for ingestion in GB.  Defaults to -1 (unlimited).
```

website/docs/d/log_analytics_workspace.html.markdown
```diff
@@ -46,6 +46,8 @@ The following attributes are exported:
 
 * `retention_in_days` - The workspace data retention in days.
 
+* `daily_quota_gb` - The workspace of the volume cap in gb.  Defaults to -1 (unlimited).
```
magodo, line 49: Can we adjust the statement a bit? And no need to mention the default value for data source.

```suggestion
* `daily_quota_gb` - The workspace daily quota for ingestion in GB.
```

magodo (changes requested):
Thanks for this PR!
I have left some additional mostly minor comments that once addressed this should be good to merge 👍

azurerm/internal/services/loganalytics/tests/log_analytics_workspace_resource_test.go
Lucretius: Looks like I was tabbing back and forth between the docs and the code too much :)

website/docs/r/log_analytics_workspace.html.markdown
Lucretius: Do we not want the `(Optional)` in the resource?

azurerm/internal/services/loganalytics/tests/log_analytics_workspace_resource_test.go
```diff
@@ -307,3 +326,29 @@ resource "azurerm_log_analytics_workspace" "test" {
...
+  location = "%s"
+}
+
+resource "azurerm_log_analytics_workspace" "test" {
+  name                = "acctestLAW-%d"
+  location            = azurerm_resource_group.test.location
+  resource_group_name = azurerm_resource_group.test.name
+  sku                 = "PerGB2018"
+  retention_in_days   = 30
+  daily_quota_gb      = %d
```
jackofallops, line 347: Incorrect type here for float:
```suggestion
  daily_quota_gb      = %f
```

azurerm/internal/services/loganalytics/log_analytics_workspace_resource.go
```diff
@@ -130,6 +137,7 @@ func resourceArmLogAnalyticsWorkspaceCreateUpdate(d *schema.ResourceData, meta i
 	}
 
 	retentionInDays := int32(d.Get("retention_in_days").(int))
+	dailyQuotaGb := float64(d.Get("daily_quota_gb").(float64))
```
jackofallops, line 140: I don't think you need the cast here as you're asserting a `float64`, rather than converting between types?

jackofallops (changes requested):
Hi @Lucretius - I was running through a review, but I see you're updating while I'm part done, so I'll leave this partial review here and come back later :)

website/docs/r/log_analytics_workspace.html.markdown
katbyte: we do

magodo:
@Lucretius iirc the only moments you willl need to use `TestCheckResourceAttr`/`TestCheckResourceAttrSet` is when the attributes you are checking are computed, e.g. when checking against a data source.

Lucretius:
I think I just need to kick the tests off for this one again, as it appears one of the linter tools had an error on downloading as opposed to breaking on an individual step.  @jackofallops you are free to continue your review, I fixed the two things you pointed out.  Thanks for bearing with me as I refresh my Go :)

katbyte (approved):
Thanks @Lucretius - this LGTM 👍

PR #8863, opened 2020-10-13
new resource for `azurerm_synapse_role_assignment`

njuCZ:
synapse data plane resource

image

points that needs to pay attention:
currently the id format is `{synapseWorkspaceId}|{synapseRoleAssignmentId}`, please confirm it's ok.
or this id format `{synapseWorkspaceId}/roleAssignments/{synapseRoleAssignmentId}` is better


azurerm/internal/services/synapse/tests/synapse_role_assignment_resource_test.go
```diff
@@ -0,0 +1,188 @@
...
+			{
+				Config: testAccAzureRMSynapseRoleAssignment_template(data),
+			},
+			// create synapse role assignment
+			{
+				Config: testAccAzureRMSynapseRoleAssignment_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMSynapseRoleAssignmentExists(data.ResourceName),
+				),
+			},
```
katbyte, line 55: how come this is a two step process? this means users won't be able to chain these togeather?

azurerm/internal/services/synapse/tests/synapse_role_assignment_resource_test.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func testAccAzureRMSynapseRoleAssignment_template(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
+
+resource "azurerm_resource_group" "test" {
+  name     = "acctest-Synapse-%d"
```
katbyte, line 152: this should be 
`acctestRG-synapse-%d`

azurerm/internal/services/synapse/tests/synapse_role_assignment_resource_test.go
```diff
@@ -0,0 +1,188 @@
...
+  account_replication_type = "LRS"
+}
+
+resource "azurerm_storage_data_lake_gen2_filesystem" "test" {
+  name               = "acctest-%d"
+  storage_account_id = azurerm_storage_account.test.id
+}
+
+resource "azurerm_synapse_workspace" "test" {
+  name                                 = "acctestsw%d"
```
katbyte, line 171: could we use capitals hereif allowed to ensure the API handles them correctly?

azurerm/internal/services/synapse/tests/synapse_role_assignment_resource_test.go
```diff
@@ -0,0 +1,188 @@
...
+  name                     = "acctestacc%s"
+  resource_group_name      = azurerm_resource_group.test.name
+  location                 = azurerm_resource_group.test.location
+  account_kind             = "BlobStorage"
+  account_tier             = "Standard"
+  account_replication_type = "LRS"
+}
+
+resource "azurerm_storage_data_lake_gen2_filesystem" "test" {
+  name               = "acctest-%d"
```
katbyte, line 166: could we use some capitals here?

website/docs/r/synapse_role_assignment.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+```
+## Argument Reference
+
+The following arguments are supported:
+
+* `synapse_workspace_id` - (Required) The ID of the Synapse Workspace on which to create the Role Assignment. Changing this forces a new resource to be created.
+
+* `role_name` - (Required) The Role Name of the Synapse Built-In Role. Changing this forces a new resource to be created.
+
+-> **NOTE:** Currently, the Synapse built-in roles are `Workspace Admin`, `Apache Spark Admin` and `Sql Admin`.
```
katbyte, line 68: could we validate against these values?

katbyte (changes requested):
Thanks @njuCZ - i've given this a review and left some comments inline

azurerm/internal/services/synapse/tests/synapse_role_assignment_resource_test.go
njuCZ: user must have network access permission first and then they could assign synapse role. users could use depends_on to  chain these togeather

azurerm/internal/services/synapse/tests/synapse_role_assignment_resource_test.go
njuCZ: synapse workspace name could only contain lowercase letters or numbers. We already have this validation rule

azurerm/internal/services/synapse/tests/synapse_role_assignment_resource_test.go
njuCZ: the name of storage data_lake_gen2_filesystem could only contain lowercase alphanumeric characters and hyphens, we already have this validation rule in this resource

website/docs/r/synapse_role_assignment.html.markdown
njuCZ: I have added dynamic validation in the codes. If the backend service add more built-in rules, we don't need to change the code and could give user a good hint. Do you think it's better to add these three value in the field's validation rules? If so, I will do so

njuCZ:
@katbyte Thanks for your time to review this PR,  could you please have a look again?

azurerm/internal/services/synapse/synapse_role_assignment_resource.go
```diff
@@ -0,0 +1,191 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.IsUUID,
+			},
+
+			"role_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
```
katbyte, line 58: Again, could we please validate these against the list of valid roles?

azurerm/internal/services/synapse/tests/synapse_role_assignment_resource_test.go
katbyte: Could we use depends on in the tf configs instead of 2 tests steps?

website/docs/r/synapse_role_assignment.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+  end_ip_address       = "255.255.255.255"
+}
+
+data "azurerm_client_config" "current" {}
+
+resource "azurerm_synapse_role_assignment" "test" {
+  synapse_workspace_id = azurerm_synapse_workspace.test.id
+  role_name            = "Sql Admin"
+  principal_id         = data.azurerm_client_config.current.object_id
+}
```
katbyte, line 58: could we put a depends on here so users know what to do? as otherwise this config doesn't work?

website/docs/r/synapse_role_assignment.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+```
+## Argument Reference
+
+The following arguments are supported:
+
+* `synapse_workspace_id` - (Required) The ID of the Synapse Workspace on which to create the Role Assignment. Changing this forces a new resource to be created.
```
katbyte, line 64: could we also add a note about needing to use a firewall rule to allow access?

katbyte (changes requested):
Thanks @njuCZ, just a few more comments to address and this should be good to merge

njuCZ:
Hi @katbyte, I have updated this PR, could you have a look again when free?

katbyte (approved):
Thanks @njuCZ - LGTM 👍

vsabella:
@katbyte @njuCZ 

This regresses the provider for any Azure environment which does not have Synapse available - which is basically China, USGov, and German clouds. The provider is no longer usable in those environments because the call to create a synapse token fails. See https://github.com/terraform-providers/terraform-provider-azurerm/issues/9099

https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go

You shouldn't add Synapse credentials into the main provider here: https://github.com/terraform-providers/terraform-provider-azurerm/blob/2c7ae3cfcc59aad45239d021aa64198dd1701d51/azurerm/internal/clients/builder.go#L100

I imagine you would want to use the Client model using Autorest to create a new client - similar to how DataLake provisions their client.

Do we not run automated tests against environments than public cloud (like say German cloud) to prevent regression?

PR #8878, opened 2020-10-14
azurerm_firewall: supports `dns_setting`

magodo:
`azurerm_firewall` supports `dns_setting` and `azurerm_firewall_network_rule_collection` supports `destination_fqdns`.

## Test Result

```bash
💢 make testacc TEST=./azurerm/internal/services/network/tests TESTARGS='-run "TestAccAzureRMFirewallNetworkRuleCollection_fqdns|TestAccAzureRMFirewallNetworkRuleCollection_noDestination|TestAccAzureRMFirewall_enableDNS"'

==> Checking that code complies with gofmt requirements...
==> Checking that Custom Timeouts are used...
==> Checking that acceptance test packages are used...
TF_ACC=1 go test ./azurerm/internal/services/network/tests -v -run "TestAccAzureRMFirewallNetworkRuleCollection_fqdns|TestAccAzureRMFirewallNetworkRuleCollection_noDestination|TestAccAzureRMFirewall_enableDNS" -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMFirewallNetworkRuleCollection_fqdns
=== PAUSE TestAccAzureRMFirewallNetworkRuleCollection_fqdns
=== RUN   TestAccAzureRMFirewallNetworkRuleCollection_noDestination
=== PAUSE TestAccAzureRMFirewallNetworkRuleCollection_noDestination
=== RUN   TestAccAzureRMFirewall_enableDNS
=== PAUSE TestAccAzureRMFirewall_enableDNS
=== CONT  TestAccAzureRMFirewallNetworkRuleCollection_fqdns
=== CONT  TestAccAzureRMFirewall_enableDNS
=== CONT  TestAccAzureRMFirewallNetworkRuleCollection_noDestination
--- PASS: TestAccAzureRMFirewallNetworkRuleCollection_fqdns (2213.69s)
--- PASS: TestAccAzureRMFirewall_enableDNS (2222.99s)
--- PASS: TestAccAzureRMFirewallNetworkRuleCollection_noDestination (2234.54s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/tests       2234.580s
```

Fixes #8312, fixes #7743.

PR #8879, opened 2020-10-14
azurerm_firewall: support for firewall manager (firewall policy)

magodo:
This PR adds the required properties to `azurerm_firewall` for make use of `azurerm_firewall_policy`, under both [secured virtual hub cases and hub virtual network cases](https://docs.microsoft.com/en-us/azure/firewall-manager/deployment-overview).

This implements part of feature request in #7319.

## Test Result

```bash
💤 make testacc TEST=./azurerm/internal/services/network/tests TESTARGS='-run "TestAccAzureRMFirewall_basic|TestAccAzureRMFirewall_inVirtualHub|TestAccAzureRMFirewall_withFirewallPolicy"'
                                                                      
TF_ACC=1 go test ./azurerm/internal/services/network/tests -v -run "TestAccAzureRMFirewall_basic|TestAccAzureRMFirewall_inVirtualHub|TestAccAzureRMFirewall_withFirewallPolicy" -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersi
on=acc"
=== RUN   TestAccAzureRMFirewall_basic
=== PAUSE TestAccAzureRMFirewall_basic                          
=== RUN   TestAccAzureRMFirewall_withFirewallPolicy
=== PAUSE TestAccAzureRMFirewall_withFirewallPolicy
=== RUN   TestAccAzureRMFirewall_inVirtualHub      
=== PAUSE TestAccAzureRMFirewall_inVirtualHub                                                                                                
=== CONT  TestAccAzureRMFirewall_basic                                                                                                       
=== CONT  TestAccAzureRMFirewall_inVirtualHub            
=== CONT  TestAccAzureRMFirewall_withFirewallPolicy
--- PASS: TestAccAzureRMFirewall_basic (1139.61s)
--- PASS: TestAccAzureRMFirewall_withFirewallPolicy (1616.77s)                                                                               
--- PASS: TestAccAzureRMFirewall_inVirtualHub (2658.22s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/tests       2658.259s
```

### DIFF
--- a/azurerm/internal/services/network/firewall_resource.go
+++ b/azurerm/internal/services/network/firewall_resource.go
@@ -183,15 +183,9 @@ func resourceArmFirewall() *schema.Resource {
 							Default:      1,
 						},
 						"public_ip_addresses": {
-							Type: schema.TypeList,
-							// Here O+C is because in most cases "public_ip_addresses" can be the Computed attributes,
-							// but still allow users to specify certain IPs to retain when scale down.
-							Optional: true,
+							Type:     schema.TypeList,
 							Computed: true,
-							Elem: &schema.Schema{
-								Type:         schema.TypeString,
-								ValidateFunc: validation.IsIPAddress,
-							},
+							Elem:     &schema.Schema{Type: schema.TypeString},
 						},
 						"private_ip_address": {
 							Type:     schema.TypeString,
@@ -282,7 +276,7 @@ func resourceArmFirewallCreateUpdate(d *schema.ResourceData, meta interface{}) e
 		parameters.AzureFirewallPropertiesFormat.FirewallPolicy = &network.SubResource{ID: &policyId}
 	}
 
-	vhub, hubIpAddresses, ok := expandArmFirewallVirtualHubSetting(existing, d.Get("virtual_hub").([]interface{}))
+	vhub, hubIpAddresses, ok := expandArmFirewallVirtualHubSetting(d.Get("virtual_hub").([]interface{}))
 	if ok {
 		parameters.AzureFirewallPropertiesFormat.VirtualHub = vhub
 		parameters.AzureFirewallPropertiesFormat.HubIPAddresses = hubIpAddresses
@@ -635,7 +629,7 @@ func flattenArmFirewallDNSServers(input map[string]*string) []interface{} {
 	return utils.FlattenStringSlice(&servers)
 }
 
-func expandArmFirewallVirtualHubSetting(existing network.AzureFirewall, input []interface{}) (vhub *network.SubResource, ipAddresses *network.HubIPAddresses, ok bool) {
+func expandArmFirewallVirtualHubSetting(input []interface{}) (vhub *network.SubResource, ipAddresses *network.HubIPAddresses, ok bool) {
 	if len(input) == 0 {
 		return nil, nil, false
 	}
@@ -644,35 +638,9 @@ func expandArmFirewallVirtualHubSetting(existing network.AzureFirewall, input []
 
 	vhub = &network.SubResource{ID: utils.String(b["virtual_hub_id"].(string))}
 
-	// If scaling down the hub public IPs, API requires users to specify the certain existing public IPs to retain.
-	// In case users do not specify the "public_ip_addresses" explicitly, the computed list of last provision will
-	// be used, in which case the amount of IPs (oldCount) is greater than the target (newCount).
-	// In this case, we will keep the first newCount public IPs.
-	oldCount := 0
-	if prop := existing.AzureFirewallPropertiesFormat; prop != nil {
-		if ipaddress := prop.HubIPAddresses; ipaddress != nil {
-			if pips := ipaddress.PublicIPs; pips != nil {
-				if count := pips.Count; count != nil {
-					oldCount = int(*count)
-				}
-			}
-		}
-	}
-	newCount := b["public_ip_count"].(int)
-
-	var addresses []network.AzureFirewallPublicIPAddress
-	for _, addr := range b["public_ip_addresses"].([]interface{}) {
-		addr := addr.(string)
-		addresses = append(addresses, network.AzureFirewallPublicIPAddress{Address: &addr})
-	}
-	if newCount < oldCount {
-		addresses = addresses[:newCount]
-	}
-
 	ipAddresses = &network.HubIPAddresses{
 		PublicIPs: &network.HubPublicIPAddresses{
-			Addresses: &addresses,
-			Count:     utils.Int32(int32(newCount)),
+			Count: utils.Int32(int32(b["public_ip_count"].(int))),
 		},
 	}
 
--- a/azurerm/internal/services/network/firewall_resource.go
+++ b/azurerm/internal/services/network/firewall_resource.go
@@ -51,9 +51,39 @@ func resourceArmFirewall() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupName(),
 
+			// TODO 3.0: change this to required
+			"sku_name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(network.AZFWHub),
+					string(network.AZFWVNet),
+				}, false),
+			},
+
+			// TODO 3.0: change this to required
+			"sku_tier": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(network.Premium),
+					string(network.Standard),
+				}, false),
+			},
+
+			"firewall_policy_id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validate.FirewallPolicyID,
+			},
+
 			"ip_configuration": {
 				Type:     schema.TypeList,
-				Required: true,
+				Optional: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"name": {
@@ -116,6 +146,9 @@ func resourceArmFirewall() *schema.Resource {
 				Optional: true,
 				Default:  string(network.AzureFirewallThreatIntelModeAlert),
 				ValidateFunc: validation.StringInSlice([]string{
+					// TODO 3.0: remove the default value and the `""` below. So if it is not specified
+					// in config, it will not be send in request, which is required in case of vhub.
+					"",
 					string(network.AzureFirewallThreatIntelModeOff),
 					string(network.AzureFirewallThreatIntelModeAlert),
 					string(network.AzureFirewallThreatIntelModeDeny),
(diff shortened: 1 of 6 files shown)


azurerm/internal/services/network/firewall_resource.go
```diff
@@ -51,9 +51,39 @@ func resourceArmFirewall() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupName(),
 
+			// TODO 3.0: change this to required
```
katbyte, line 54: Does this default to a value right now? if so can we just defaut to it?

azurerm/internal/services/network/firewall_resource.go
```diff
@@ -51,9 +51,39 @@ func resourceArmFirewall() *schema.Resource {
...
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(network.AZFWHub),
+					string(network.AZFWVNet),
+				}, false),
+			},
+
+			// TODO 3.0: change this to required
```
katbyte, line 66: Does this default to a value right now? if so can we just defaut to it?

katbyte (changes requested):
Thanks for the pr @magodo - left some comments inline to address

website/docs/r/firewall.html.markdown
```diff
@@ -106,6 +116,18 @@ A `management_ip_configuration` block supports the following:
...
+
+A `virtual_hub_setting` block supports the following:
+
+* `virtual_hub_id` - (Required) Specifies the ID of the Virtual Hub where the Firewall resides in.
+
+* `public_ip_count` - (Optional) Specifies the amount of public IP to assign for the Firewall. Defaults to `1`.
+
+* `public_ip_addresses` - (Optional) A list of public IP addresses associated with the Firewall.
+
+-> **NOTE** In most cases, users do not need to specify this argument, except when scaling down the public IPs (by changing the `public_ip_count`), this argument allows users to specify the existing IPs to be retained. Otherwise, the first existing `public_ip_count` public ip addresses will be retained.
```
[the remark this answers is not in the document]
magodo: The last change here will change the meaning a bit, it will affect the first existing N public ip addresses actually, where N equals to `public_ip_count`.

azurerm/internal/services/network/firewall_resource.go
magodo: Yes, the default sku name is: `AZFW_VNet`.

But persumably it will introduce a breaking change. Given a user has an existing FW with another sku name (because service is using another sku name at the time the FW is created), then the plan will mark the resource to be replaced.

WDYT?

azurerm/internal/services/network/firewall_resource.go
magodo: Yes, currently the default sku name is: `Standard`.

But same question as above.

website/docs/r/firewall.html.markdown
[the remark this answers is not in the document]
katbyte: I'm not sure how better to word it as i still don't fully understand what it is doing :/

azurerm/internal/services/network/firewall_resource.go
katbyte: I think it would only affect users who have modified this resource outside of terraform or imported one that was created outside of terraform? but fair point - easy enough to mark it as computed till 3.0!

website/docs/r/firewall.html.markdown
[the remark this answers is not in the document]
magodo: This means when you scale down the public IPs of the firewall, say from (pip1, pip2, pip3) to only one pip. If you simply change the `public_ip_count` from 3 to 1. Then it ends up with the pip1. However, if you intend to keep the pip3, then you can specify it using this property, say `public_ip_addresses = [pip3]`.

magodo:
Hi @katbyte Thank you for the review comments! I have updated the code accordingly, and fix one coflict setting introduced in #8878 after rebasing to the master.

Both the test introduced in this PR and #8878 are passed now:

```shell
terraform-provider-azurerm on  new_firewall_support_policy [$] via 🐹 v1.15.3 
💢 make testacc TEST=./azurerm/internal/services/network/tests TESTARGS='-run "TestAccAzureRMFirewall_inVirtualHub"'
TF_ACC=1 go test ./azurerm/internal/services/network/tests -v -run "TestAccAzureRMFirewall_inVirtualHub" -timeout 180m -ldflags="-X=github.com/terraform-provi
ders/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMFirewall_inVirtualHub                                                                                                                 === PAUSE TestAccAzureRMFirewall_inVirtualHub
=== CONT  TestAccAzureRMFirewall_inVirtualHub
--- PASS: TestAccAzureRMFirewall_inVirtualHub (2742.87s)
PASS                                 
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/tests       2742.907s

💤 make testacc TEST=./azurerm/internal/services/network/tests TESTARGS='-run "TestAccAzureRMFirewall_withFirewallPolicy|TestAccAzureRMFirewall_enableDNS"'
TF_ACC=1 go test ./azurerm/internal/services/network/tests -v -run "TestAccAzureRMFirewall_withFirewallPolicy|TestAccAzureRMFirewall_enableDNS" -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMFirewall_enableDNS
=== PAUSE TestAccAzureRMFirewall_enableDNS
=== RUN   TestAccAzureRMFirewall_withFirewallPolicy
=== PAUSE TestAccAzureRMFirewall_withFirewallPolicy
=== CONT  TestAccAzureRMFirewall_enableDNS
=== CONT  TestAccAzureRMFirewall_withFirewallPolicy
--- PASS: TestAccAzureRMFirewall_withFirewallPolicy (2013.98s)
--- PASS: TestAccAzureRMFirewall_enableDNS (3747.25s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/tests       3747.293s
```

azurerm/internal/services/network/firewall_resource.go
```diff
@@ -132,6 +165,42 @@ func resourceArmFirewall() *schema.Resource {
...
+							ValidateFunc: validate.ValidateVirtualHubID,
+						},
+						"public_ip_count": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validation.IntAtLeast(1),
+							Default:      1,
+						},
+						"public_ip_addresses": {
+							Type: schema.TypeList,
```
manicminer, line 186: @magodo Is it possible to test scenarios involving the `public_ip_addresses` attribute? I'm curious what happens if you scale out the number of public IPs and only specify a subset of them. I also suspect if there are ordering issues, this attribute may need to be a set.

magodo: @manicminer The reason why this is a list is because it is a Computed attribute, a list is more user friendly. Regarding to test for `public_ip_addresses`, I guess we need to enhance the data source of the `firewall` resource so that we can retrieve the Computed `public_ip_addresses` and feed a subset of them to the resource for another apply?

magodo: @manicminer I added the test accordingly, please take another look!

manicminer: @magodo thanks, that's good for the scaling in case, but can we also add a test for the scaling out case?

magodo: @manicminer I meant that I've added the scaling out case in the last commit [here](https://github.com/terraform-providers/terraform-provider-azurerm/pull/8879/commits/b98806e5fad991c30e01c69c88bfb0c21cbd89d7#diff-dcc9ede2bfb4a29b38331b171874632050bfa7d2551d63fa98c04f971049b5c0R347).

manicminer: @magodo Thanks, I saw that. I'd describe that case as scaling in / down. The use case I'm interested in is scaling up / out and how that'll affect the `public_ip_addresses` attribute.

For example, say I have `public_ip_count = 2` and `public_ip_addresses = ["1.2.3.4", "1.2.3.5"]`, and then I increase `public_ip_count = 4` and apply. Are the newly provisioned public IP addresses returned in the `AzureFirewallPropertiesFormat.HubIPAddresses.PublicIPs.Addresses` field? If so, what happens to the Terraform plan?

magodo: @manicminer Thank you for clarify that (i mistakely regard scaling out as scaling down...).

In this case, the first apply will take effect and then set the `public_ip_addresses` with 4 elements (adding two new ip addresses on top of the other two ips you've specified). The second terraform plan will show diff if the user is literally setting two elements in the config. However, in the scaling out/up use case, the user should not set that property at all, as I mentioned in the [document](https://github.com/terraform-providers/terraform-provider-azurerm/pull/8879/files?file-filters%5B%5D=.markdown#diff-50fbb9bccec1371a75f26bb6531f3339aaf13f78b4d8321fbc3a06284410e7a1R131). This property is mostly meant to be computed, and only useful in scale in/down use case.

manicminer: @magodo Thanks for clarifying, that was my interpretation. In this case, the API is treating the field differently depending on how the value of `public_ip_count` is changing - essentially this should be two separate fields in the API. However, with it being one field having multiple behaviors, we can't really support this. Attributes that are only set for the duration of a single apply are going to produce inconsistent plans in pretty much any change scenario.

I suggest we simply make this field Computed (read-only), which will expose the allocated IPs for reference by the operator in other resources and I think is more important to enable.

magodo: @manicminer This makes total sense to me! I have made the changes to make `public_ip_addresses` to be computed only in the latest commit, please take another look!

katbyte:
I believe this is blocked by a breaking change in the service? see #9312

tests are failing with:
```
        Error: creating Firewall Policy "acctest-networkfw-Policy-201114233159948085" (Resource Group "acctestRG-networkfw-201114233159948085"): network.FirewallPoliciesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="FirewallPolicyDeprecatedRequireProxyForNetworkRules" Message="RequireProxyForNetworkRules ( DNS Settings' Property) in Firewall Policy /subscriptions/*******/resourceGroups/acctestRG-networkfw-201114233159948085/providers/Microsoft.Network/firewallPolicies/acctest-networkfw-Policy-201114233159948085 is deprecated for all the API Versions. Please remove the property and try again."
```

magodo:
@katbyte The test failure is not related to the issue #9312 (though I created another PR to fix this issue), since the firewall test cases do not touch dns setting for firewall policy at all.

I submit another commit to fix the test failures.

```bash
💢 make testacc TEST=./azurerm/internal/services/network/tests TESTARGS='-run "TestAccAzureRMFirewall_withFirewallPolicy|TestAccAzureRMFirewall_inVirtualHub"'
TF_ACC=1 go test ./azurerm/internal/services/network/tests -v -run "TestAccAzureRMFirewall_withFirewallPolicy|TestAccAzureRMFirewall_inVirtualHub" -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMFirewall_withFirewallPolicy
=== PAUSE TestAccAzureRMFirewall_withFirewallPolicy
=== RUN   TestAccAzureRMFirewall_inVirtualHub
=== PAUSE TestAccAzureRMFirewall_inVirtualHub
=== CONT  TestAccAzureRMFirewall_withFirewallPolicy
=== CONT  TestAccAzureRMFirewall_inVirtualHub
--- PASS: TestAccAzureRMFirewall_withFirewallPolicy (1090.44s)
--- PASS: TestAccAzureRMFirewall_inVirtualHub (2831.62s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/tests       2831.662s
```

katbyte (approved):
Thanks @magodo - was able to get the tests to pass! LGTM 👍

woo-Erupt:
Hi .. can you tell me when the "Add security to virtual networks" feature will be added for the Firewall Manager service?
The features required would be:
feature #1 "Create a new hub virtual network"
feature #2: "Convert virtual network"

PR #8888, opened 2020-10-14
[WIP] Fix example where infrastructure_encryption_enabled is not supported

Miouge1:
I tried the example provided in the docs and got an error that:

- `infrastructure_encryption_enabled` is not supported for sku Tier `Basic`
- Geo-Redundant Backup is not supported

Therefore I propose to disable it in the example.


katbyte (approved):
Thanks @Miouge1! LGTM 👍

PR #8895, opened 2020-10-15
new resource "azurerm_key_vault_key_encrypt" and data source "azurerm_key_vault_key_decrypt"

### LINKED ISSUE #7856: Support for decrypt secrets from data encrypted with Azure keyvault key
### Community Note

### Description

AWS provider has this data source which decrypts cipherblob and then stores it in plain text within secrets manager. Need similar functionality for azurerm, which will decrypt the cipherblob (using keyvault key) and then store it in plain text in keyvault secrets.

https://www.terraform.io/docs/providers/aws/d/kms_secrets.html

njuCZ:
fix #7856

resource "azurerm_key_vault_key_encrypt" is similar with "aws_kms_ciphertext"
data source "azurerm_key_vault_key_decrypt" is similar with "aws_kms_secret"

to keep a stable encrypted value, make "azurerm_key_vault_key_encrypt" be a resource but not a data source.
there is no need to support "import" operation for resource "azurerm_key_vault_key_encrypt"

image


tombuildsstuff:
dependent on https://github.com/terraform-providers/terraform-provider-azurerm/issues/7856#issuecomment-711619498

saranshlamba:
@tombuildsstuff Good morning. Any updates on this and when it will be merged?

azurerm/internal/services/keyvault/key_vault_key_encrypt_resource.go
```diff
@@ -0,0 +1,94 @@
+package keyvault
```
IanMoroney, line 1: If the intention is to encrypt secrets, might i suggest that the resource instead be called:
`azurerm_key_vault_secret_encrypt`

azurerm/internal/services/keyvault/key_vault_key_decrypt_data_source.go
```diff
@@ -0,0 +1,83 @@
+package keyvault
```
IanMoroney, line 1: If the intention is to decrypt secrets, might i suggest that the resource instead be called:
`azurerm_key_vault_secret_decrypt`

azurerm/internal/services/keyvault/key_vault_key_encrypt_resource.go
```diff
@@ -0,0 +1,94 @@
...
+
```
IanMoroney, line 2: I can see there being a value in having a `resource` block for encrypting secrets with a static encrypted blob, and a `data` block for encrypting secrets with a value that changes on every apply, but I only see an `encrypt` resource. 

Could you create an equivalent `encrypt` data block, as if we're modelling it after aws_kms_ciphertext, we should have the ability for users to choose whether to change the encrypted blob frequently or infrequently.

azurerm/internal/services/keyvault/key_vault_key_decrypt_data_source.go
```diff
@@ -0,0 +1,83 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"key_vault_key_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: azure.ValidateKeyVaultChildId,
+			},
+
+			"payload": {
```
katbyte, line 31: what format are we expecting here? base64?
```suggestion
			"encrypted_base64_data": {
```

azurerm/internal/services/keyvault/key_vault_key_encrypt_resource.go
```diff
@@ -0,0 +1,94 @@
...
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(keyvault.RSA15),
+					string(keyvault.RSAOAEP),
+					string(keyvault.RSAOAEP256),
+				}, false),
+			},
+
+			"cipher_text": {
```
katbyte, line 60: should this also then be
```suggestion
			"encrypted_data_in_base64": {
```

azurerm/internal/services/keyvault/key_vault_key_encrypt_resource.go
```diff
@@ -0,0 +1,94 @@
...
+		}),
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(5 * time.Minute),
+			// avoid test fail, add timeout config for read and delete
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
```
katbyte, line 34: does it make sense to add a name property so we can identify this and use it in the ID?

azurerm/internal/services/keyvault/tests/key_vault_key_decrypt_data_source_test.go
```diff
@@ -0,0 +1,39 @@
...
+		},
+	})
+}
+
+func testAccDataSourceKeyVaultKeyDecrypt_basic(data acceptance.TestData, plaintext string) string {
+	t := testAccAzureRMKeyVaultKeyEncrypt_basic(data, plaintext)
+	return fmt.Sprintf(`
+%s
+
+data "azurerm_key_vault_key_decrypt" "test" {
```
katbyte, line 33: could we use the encrypt resource and decrypt to check that the workflow works?

website/docs/d/key_vault_key_decrypt.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_key_vault_key_decrypt"
+description: |-
+  Decrypt data encrypted with the Key Vault Key.
+
+---
+
+# Data Source: azurerm_key_vault_key_decrypt
+
+Use this data source to decrypt data encrypted with the Key Vault Key.
```
katbyte, line 12: grammer
```suggestion
Use this data source to decrypt data encrypted with a Key Vault Key.
```

website/docs/r/key_vault_key_encrypt.html.markdown
```diff
@@ -0,0 +1,56 @@
...
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_key_vault_key_encrypt"
+description: |-
+  Encrypt data with the Key Vault Key.
+
+---
+
+# azurerm_key_vault_key_encrypt
+
+Use this resource to encrypt data with the Key Vault Key.
```
katbyte, line 12: grammer

katbyte (changes requested):
Thanks @njuCZ - left some comments inline

azurerm/internal/services/keyvault/tests/key_vault_key_decrypt_data_source_test.go
njuCZ: yes, in this data source, we assign the `cipher_text` of `azurerm_key_vault_key_encrypt` to `payload`, and in the check functions, we checked whether the descrypt value is the inital input plaintext.
In this way, we checked the whole encrypt and decrypt workflow

azurerm/internal/services/keyvault/key_vault_key_encrypt_resource.go
njuCZ: the result format is `base64url`, which is slightly different from `base64`. please refer to RFC: https://tools.ietf.org/html/rfc4648#section-5. I have noted it in the doc. Should I still rename this property?

azurerm/internal/services/keyvault/key_vault_key_decrypt_data_source.go
njuCZ: as below

njuCZ:
Thanks @katbyte , I have rebased and refine this PR

azurerm/internal/services/keyvault/key_vault_key_encrypt_resource.go
katbyte: yes lets rename it to match what is expected so people don't have to look at docs and can tell from the property name. as for url in there is this different then any other base64 property for azure?

azurerm/internal/services/keyvault/key_vault_key_decrypt_data_source.go
```diff
@@ -0,0 +1,84 @@
...
+			"key_vault_key_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NestedItemId,
+			},
+
+			"payload": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
```
katbyte, line 35: lets just use base64 without any url in the property name and then validate it?

azurerm/internal/services/keyvault/key_vault_key_encrypt_resource.go
katbyte: lets go with `encrypted_data_in_base64` and then add a validation function to ensure its the right base64 format

katbyte (changes requested):

njuCZ:
@katbyte Thanks for your opinion, I have renamed the field and added validation

IanMoroney:
Any thoughts on my comments @njucz?

azurerm/internal/services/keyvault/key_vault_key_encrypt_resource.go
njuCZ: becasue key vault key and key vault secret are different resource, and the rest api is using key vault key to encrypt and decrypt data. So personally I think `azurerm_key_vault_secret_encrypt` maybe a little misleading?

azurerm/internal/services/keyvault/key_vault_key_decrypt_data_source.go
njuCZ: as above

azurerm/internal/services/keyvault/key_vault_key_encrypt_resource.go
njuCZ: is there any scenario that users want to change the encrypted data every time? In the `resource`, we could also change the encrypted data by changing the property `name`

njuCZ:
@IanMoroney sorry for late reply. I thought this PR was blocked, but it seems things become better now. I have commented inline. Thanks for your review

gro1m:
@njuCZ @IanMoroney What is the status here, actually?

tombuildsstuff:
hi @njuCZ 

Apologies for the delayed re-review of this PR - taking a look through here whilst this is functional these both want to be data sources and since the value can be converted bi-directionally we can make these a singular data source for "encrypted value".

Since there's some larger changes to get this PR in, I hope you don't mind but I've opened #15873 which supersedes this PR and will allow this to ship via the `azurerm_key_vault_encrypted_value` data source, which as mentioned above allows a bi-directional encryption/decryption of a plain-text/encrypted value as necessary - as such whilst I'd like to thank you for this contribution I'm going to close this in favour of #15873

Thanks!

PR #8901, opened 2020-10-15
Add support for SecurityProviderName property of Virtual Hub

neil-yechenwei:
Although the [SecurityProviderName](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/Microsoft.Network/stable/2020-05-01/virtualWan.json#L4850) property belongs to Virtual Hub. But this property cannot be implemented at azurerm_virtual_hub since this property only can be set after vpn gateway is attached to virtual hub according by the instruction of below api returned error message. However, vpn gateway (azurerm_vpn_gateway) is a separate resource and it depends on virtual hub, which means azurerm_virtual_hub would be created before azurerm_vpn_gateway is created. So this property has to been implemented at azurerm_vpn_gateway.

Error message:
```
network.VirtualHubsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="VirtualHubSecurityProviderNotSupportedWithoutVPNGw" Message="SecurityProviderName 'xxxxxx' is not supported without VPN Gateway in the Hub." Details=[]
```

### DIFF
--- a/azurerm/internal/services/network/tests/vpn_gateway_resource_test.go
+++ b/azurerm/internal/services/network/tests/vpn_gateway_resource_test.go
@@ -130,6 +130,25 @@ func TestAccAzureRMVPNGateway_tags(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMVPNGateway_withSecurityProviderName(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_vpn_gateway", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMVPNGatewayDestroy,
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
hi @neil-yechenwei 

Thanks for this PR.

Taking a look through here whilst this property requires that a VPN Gateway exists - in practice this likely makes more sense as a separate resource rather than being bundled together within the VPN Gateway resource. As such can we update this to be a separate resource, rather than bundled within the VPN Gateway resource?

Thanks!

neil-yechenwei:
After double confirmed with service team, we should use [SecurityPartnerProvider API](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/securitypartnerproviders/createorupdate) to associate security provider to virtual hub not using securityProviderName. So I submit a [PR](https://github.com/terraform-providers/terraform-provider-azurerm/pull/8978) for SecurityPartnerProvider and I will close this PR.

PR #8903, opened 2020-10-15
Adding support for os_disk_image_size to get the disk size of the image

harshavmb:
Adding support for os_disk_image_size to get the disk size of the image…


katbyte:
Hey @harshavmb - thanks for this PR, but as it appears to be a duplicate of #8904 i'm going to close it in favour of the newer PR.

PR #8907, opened 2020-10-15
`azurerm_search_service` - add support for `identity`

### LINKED ISSUE #7657: Support for service identity on azurerm_search_service
### Community Note

### Description

Azure Cognitive Search supports enabling a service identity in order to authenticate with Key Vault to obtain customer-managed keys, as documented here: https://docs.microsoft.com/en-us/azure/search/search-security-manage-encryption-keys#3---create-a-service-identity

It should be possible to enable this service identity when creating an Azure Search service using Terraform.

The object_id of the identity should be exported as an output, so that it can be used to configure an Access Policy on a Key Vault in order to grant the Search Service identity access to the key, as explained here: https://docs.microsoft.com/en-us/azure/search/search-security-manage-encryption-keys#4---grant-key-access-permissions

This appears to be available in the SDK here: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/vendor/github.com/Azure/azure-sdk-for-go/services/search/mgmt/2015-08-19/search/models.go#L68

* azurerm_search_service

Probably this should work the same way as the identity block for a virtual machine, and exported in the same way.

### References

* #0000

mbfrahry:
Fixes #7657

### DIFF
--- a/azurerm/internal/services/search/resource_arm_search_service.go
+++ b/azurerm/internal/services/search/resource_arm_search_service.go
@@ -120,6 +120,33 @@ func resourceArmSearchService() *schema.Resource {
 				},
 			},
 
+			"identity": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


katbyte (approved):
Thanks @mbfrahry - LGTM 👍

PR #8006, opened 2020-08-04
Use API version 2020-03-01 for Loadbalancer

marc-sensenich:
Closes #7691 by using the 2020-03-01 network API version. Based off a comment by @tombuildsstuff from #7741 as this issue is present on more than just `azurerm_lb_rule` deletion 

The reason for that issue is that service can't tolerate the client sending properties.backendAddressPools.n.properties.backendIPConfigurations in request body for load balancer resource since 2020-05-01, as it is a READ-ONLY attribute (this only occurs when azurerm_lb's sku is Standard).
Currently, Azure Go SDK can trim the READ-ONLY attributes during marshaling, but only for the root level attributes, so, in this case, Go SDK will still marshal the attribute into the request body. This is tracked in issue: Azure/autorest.go#438.

**Acceptance Tests**

```
--- PASS: TestAccAzureRMLoadBalancer_standard (74.84s)
--- PASS: TestAccAzureRMLoadBalancer_requiresImport (126.75s)
--- PASS: TestAccAzureRMLoadBalancerRule_disappears (136.04s)
--- PASS: TestAccAzureRMLoadBalancerBackEndAddressPool_basic (140.11s)
--- PASS: TestAccAzureRMLoadBalancerRule_updateMultipleRules (211.43s)
--- PASS: TestAccAzureRMLoadBalancerRule_removal (211.44s)
--- PASS: TestAccAzureRMLoadBalancer_basic (136.61s)
--- PASS: TestAccAzureRMLoadBalancerRule_inconsistentReads (211.45s)
--- PASS: TestAccAzureRMLoadBalancerProbe_basic (321.01s)
--- PASS: TestAccAzureRMLoadBalancer_frontEndConfigPublicIPPrefix (132.72s)
--- PASS: TestAccAzureRMLoadBalancerRule_complete (147.61s)
--- PASS: TestAccAzureRMLoadBalancerProbe_update (341.34s)
--- PASS: TestAccAzureRMLoadBalancerRule_basic (137.91s)
--- PASS: TestAccAzureRMLoadBalancerProbe_requiresImport (169.84s)
--- PASS: TestAccAzureRMLoadBalancerProbe_disappears (392.64s)
--- PASS: TestAccAzureRMLoadBalancerProbe_removal (202.78s)
--- PASS: TestAccAzureRMLoadBalancerProbe_updateProtocol (389.76s)
--- PASS: TestAccAzureRMLoadBalancerRule_update (1464.92s)
[...]

Issue #7691, opened 2020-07-10
2.18.0 - Error removing/updating load balancer rules


hbuckle:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.12.9
+ provider.azurerm v2.18.0

### Affected Resource(s)

* `azurerm_lb_rule`

### Expected Behavior

Something seems to have gone terribly wrong with deleting load balancer rules in 2.18.0 (I guess due to the new network SDK?), I'm seeing the error below.
It is fine with 2.17.0

### Actual Behavior

[...]

PR #8910, opened 2020-10-16
update api version of LB to 2020-05-01

magodo:
The LB api version is rolled back to 2020-03-01 in #8006 because of issue #7691. Now that the issue has been addressed in Azure/azure-rest-api-specs#10131 and the change has been involved in the Go SDK currently under used. So we shall upgrade the API back to 2020-05-01 so that all network resources are using the same one.


tombuildsstuff (approved):
LGTM - thanks for this @magodo

tombuildsstuff:
@magodo how does https://github.com/Azure/azure-rest-api-specs/pull/10131 fix this? That Swagger fix appears to be just re-ordering two fields in the Swagger, rather than doing anything else - IIRC this was an API bug where these fields were being incorrectly mapped?

magodo:
@tombuildsstuff The reason is that the service somehow can't tolerate the first properties to be read only, after swapping the order, the service can correctly decode the request.

Quote from service team:

> We took a look at the issue. The problem is that payload does not match what service expects:
In json payload we see both backendIPConfigurations and loadBalancerBackendAddressesProperties in backendAddressPool child resource.
It is not expected in service side to get the backendIPConfigurations property located first in JSON. That is causing the error on server side when deserialization happens.

tombuildsstuff:
@magodo given the JSON spec states that the keys inside a map are can be in any order (and I can't think of a JSON Serialisation library which supports that?) - that sounds like an API bug which still needs to be resolved? Unfortunately we'd be unable to ship this until this is fixed, due to the design of/monthly churn from the Networking API's

magodo:
Then it is possible that the Go SDK has changed the marshalling code generator to not include readonly properties in the request body (Azure/autorest.go#446)?

magodo:
@tombuildsstuff Yes, you are right. So the reason why it now works seems because [we do not send the other properties in the API request](https://github.com/magodo/terraform-provider-azurerm/blob/de7d098fb7c8b2548e59622356f927a6f5456224/vendor/github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-05-01/network/models.go#L6838-L6838) in Go SDK any more due to Azure/autorest.go#446.

tombuildsstuff:
@magdoo unfortunately that's not something we can depend/reply on, due to the large churn within the Azure SDK - so it's possible for this to subtly break at any time, in a way that's incredibly hard to diagnose?

As such I think this PR is blocked for the moment - can we open a Swagger issue and reach out to the Service Team to get this bug fixed?

PR #8920, opened 2020-10-16
Lb backend address pool more attributes

magodo:
Based on #8910

- Support `outbound_rules` for azurerm_lb_backend_address_pool (both data source and resource)
- Support `backend_address` for azurerm_lb_backend_address_pool (data source only)
- Use the backend address pool dedicated endpoint in data source to make the code easier to maintain

Additionally, the backend address pool dedicated endpoint supports PUT operation for Standard sku LB only. So I didn't use it for the resource code. In the meanwhile, I submit an API issue for that: Azure/azure-rest-api-specs#11234


tombuildsstuff:
@magodo since this is blocked on https://github.com/Azure/azure-rest-api-specs/issues/11234 (and #8910) - rather than leaving this open blocked on that, I'm going to close this until #8910 is unblocked, at which point we can reopen this and take another look here.

PR #8938, opened 2020-10-19
`azurerm_data_factory_linked_service_data_lake_storage_gen2` - Supports managed identity auth through `use_managed_identity `

### LINKED ISSUE #6501: Support for adding Managed Identity to Linked Services to ADLS Gen 2 for Azure Data Factory.
## Managed Identity for Linked Service to ADLS Gen 2 for Azure Data Factory

As of January 2020, Azure Data Factory (ADF) now supports Managed Identity (formerly known as  Managed Service Identity - MSI) to connect to other Azure resources like Azure Data Lake Storage (ADLS).  Link to the update can be found [here](https://docs.microsoft.com/en-us/azure/data-factory/data-factory-service-identity)

It would be super nice, if we can perform this function in Terraform and add the corresponding role to the resource as a one step process.

### Community Note

* azurerm_data_factory_linked_service_data_lake_storage_gen2

```hcl
```

### References

* #0000

dnlbunting:
Closes  #6501

Adds a use_managed_identity flag to azurerm_data_factory_linked_service_data_lake_storage_gen2.

If both use_managed_identity=true and service_principal_key or service_principal_id are specified raises an error.

First contribution so feel free to add feedback

### DIFF
--- a/azurerm/internal/services/datafactory/data_factory_linked_service_data_lake_storage_gen2_resource.go
+++ b/azurerm/internal/services/datafactory/data_factory_linked_service_data_lake_storage_gen2_resource.go
@@ -59,21 +59,29 @@ func resourceArmDataFactoryLinkedServiceDataLakeStorageGen2() *schema.Resource {
 			},
 
 			"use_managed_identity": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				Default:  false,
+				Type:          schema.TypeBool,
+				Optional:      true,
+				Default:       false,
+				ConflictsWith: []string{"service_principal_key", "service_principal_id"},
+				AtLeastOneOf:  []string{"service_principal_key", "service_principal_id", "use_managed_identity"},
 			},
 
 			"service_principal_id": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				ValidateFunc: validation.IsUUID,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ValidateFunc:  validation.IsUUID,
+				RequiredWith:  []string{"service_principal_key"},
+				ConflictsWith: []string{"use_managed_identity"},
+				AtLeastOneOf:  []string{"service_principal_key", "service_principal_id", "use_managed_identity"},
 			},
 
 			"service_principal_key": {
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/datafactory/data_factory_linked_service_data_lake_storage_gen2_resource.go
```diff
@@ -58,21 +58,27 @@ func resourceArmDataFactoryLinkedServiceDataLakeStorageGen2() *schema.Resource {
 				ValidateFunc: validation.IsURLWithHTTPS,
 			},
 
+			"use_managed_identity": {
+				Type:     schema.TypeBool,
+				Optional: true,
```
mbfrahry, line 63: We'll want to add `ConflictsWith` here too to make sure we don't specify it with the other service principal attributes.

mbfrahry: And to round it all out we should throw in `AtLeastOneOf` with all of these attributes to make sure at least one of them gets specified when creating this resource.

mbfrahry (changes requested):
Hey @dnlbunting, thanks for opening this PR! Things look nearly there but some of the validation you wrote can be better surfaced at the plan level so we error before Terraform even applies. I've detailed some of what you need below and feel free to reach out if you have any questions around it.

mbfrahry (commented):
Actually, would it make sense to just infer that we should use managed identity if we don't specify the service principal? Or is it better for the user to purposefully specify that they want managed identity?

dnlbunting:
@mbfrahry Thanks for the pointers on the schema validation, I'll make those changes. IMO making the managed identity opt-in rather the the default was clearer, since the managed identity has to be explicitly enabled in the parent ADF resource too.

mbfrahry:
That works for me! I'll look forward to the schema validation changes

mbfrahry (approved):
LGTM!

mbfrahry:
Thanks for this @dnlbunting. It should make it into the next release!

PR #8946, opened 2020-10-19
New resources: `azurerm_log_analytics_cluster` and `azurerm_log_analytics_cluster_customer_managed_key`

WodansSon:


azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
```diff
@@ -189,6 +189,7 @@ func apiManagementCustomDomainRead(d *schema.ResourceData, meta interface{}) err
 		configs := flattenApiManagementHostnameConfiguration(resp.ServiceProperties.HostnameConfigurations, d)
 		for _, config := range configs {
 			for key, v := range config.(map[string]interface{}) {
+				// nolint R001: ResourceData.Set() key argument should be string literal
```
jackofallops, line 192: Could you revert this one please? It's in master on a service unrelated to this PR, so should be fixed separately.

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
```diff
@@ -0,0 +1,397 @@
...
+
+						"tenant_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+
+			"key_vault_property": {
```
jackofallops, line 85: Should this not be replaced with just `key_vault_key_id` that provides us with all this information?

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
```diff
@@ -0,0 +1,397 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.LogAnalyticsClustersName,
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"location": azure.SchemaLocation(),
+
+			"identity": {
```
jackofallops, line 56: This feels redundant here.  There's only one valid value for the type, and the other values are all computed.  Perhaps better to simply have `principal_id` and `tenant_id` as top level computed fields and send the singular mandatory type property without user input here?

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
```diff
@@ -0,0 +1,397 @@
...
+	resp, err := client.Get(ctx, resourceGroup, name)
+	if err != nil {
+		return fmt.Errorf("retrieving Log Analytics Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	if resp.ID == nil || *resp.ID == "" {
+		return fmt.Errorf("empty or nil ID returned for Log Analytics Cluster %q (Resource Group %q) ID", name, resourceGroup)
+	}
+
+	d.SetId(*resp.ID)
```
jackofallops, line 186: Can we replace this with the newer ID generation implementation? (see `log_analytics_workspace` for an example)

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+	if existing.ID != nil && *existing.ID != "" {
+		return tf.ImportAsExistsError("azurerm_log_analytics_cluster", *existing.ID)
+	}
+
+	keyVaultProps := expandArmLogAnalyticsClusterKeyVaultProperties(d.Get("key_vault_property").([]interface{}))
+
+	if d.IsNewResource() && keyVaultProps != nil {
+		return fmt.Errorf("the Log Analytics Cluster %q (Resource Group %q) must be successfully provisioned before it can be configured to support customer managed keys", name, resourceGroup)
+	}
```
jackofallops, line 154: The `d.IsNewResource` test is only needed when the create and update funcs are combined.  In this instance, it's not necessary here. Since the CMK element cannot be applied on create, it should probably be omitted entirely, and return the `Update` func instead of `Read` below.

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
```diff
@@ -0,0 +1,397 @@
...
+			Capacity: utils.Int64(int64(d.Get("size_gb").(int))),
+			Name:     operationalinsights.CapacityReservation,
+		}
+	}
+
+	if d.HasChange("tags") {
+		parameters.Tags = tags.Expand(d.Get("tags").(map[string]interface{}))
+	}
+
+	if _, err := client.Update(ctx, id.ResourceGroup, id.Name, parameters); err != nil {
```
jackofallops, line 255: Given the note below, can we raise an issue on this as it sounds like it should be an LRO?

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
```diff
@@ -0,0 +1,397 @@
...
+		Timeout:    d.Timeout(schema.TimeoutUpdate),
+		Refresh: func() (interface{}, string, error) {
+			log.Printf("[INFO] checking on state of Log Analytics Cluster %q", id.Name)
+
+			resp, err := client.Get(ctx, id.ResourceGroup, id.Name)
+			if err != nil {
+				return nil, "nil", fmt.Errorf("polling for the status of Log Analytics Cluster %q (Resource Group %q): %v", id.Name, id.ResourceGroup, err)
+			}
+
+			if resp.ClusterProperties.ProvisioningState != operationalinsights.Updating && resp.ClusterProperties.ProvisioningState != operationalinsights.Succeeded {
```
jackofallops, line 276: This will need nil checking as `ClusterProperties` is a pointer.

azurerm/internal/services/loganalytics/tests/log_analytics_cluster_resource_test.go
```diff
@@ -0,0 +1,282 @@
...
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMLogAnalyticsClusterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMLogAnalyticsCluster_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMLogAnalyticsClusterExists(data.ResourceName),
+				),
+			},
+			data.ImportStep("size_gb"), // not returned by the API
```
jackofallops, line 28: This feels like a bug, it should be present in the Sku?

jackofallops (changes requested):
Hi @WodansSon 
Thanks for this PR for a new resource.  I've left some comments below, but it's off to a good start.  I'll follow up with support, as noted in the new doc, to get this enabled for our test subscription asap.

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
WodansSon: I am fine with that, but it was causing a lint error for this PR. If you can grant me an exception for passing all lint tests for the PR I would be happy to do so. 🙂

WodansSon:
> I'll follow up with support, as noted in the new doc, to get this enabled for our test subscription asap.

@jackofallops, no need to do this part as I have already done it for you. 🙂

azurerm/internal/services/loganalytics/tests/log_analytics_cluster_resource_test.go
WodansSon: It is in the `Sku` I collapsed the `Sku` as it has only one possible value. The only value that is possible for the end user to change is `capacity` value in the `Sku` so I pulled that out to a top level attribute and by doing so that causes the `ImportStep` to fail because the state and `GET` do not match.

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
WodansSon: I can open an issue just to cross our T's and dot our I's, but I have already discussed this with the service team and they have stated that this behavior is currently `By Design` to allow for concurrent operations to be performed on the resource. 🤷‍♂️

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
WodansSon: I thought that too at first, but if we only allow for the `key_vault_id` we would still need to expose a way for the end user to define what `key` and `version` they would like to use with the resource. Is there a better way to do that?

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
WodansSon: I have just opened PR #9298 to fix this issue as suggested above. Thanks!

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
WodansSon: Yeah, that makes sense, however I thought the purpose of exposing this was to explicitly note in the resource `HCL` that the resource has it's own principal identity in `AAD`? I am fine with changing it, but I was just following the pattern that has already been established in other resources, LMK what you would like me to do here and I would be happy to change it. Thanks. 🙂

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
WodansSon: Fixed.

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
WodansSon: Fixed.

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
WodansSon: Fixed.

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
WodansSon: I have determined through testing that this check needs to exist. Why? Because you cannot create the access policy in the key vault granting access to the cluster before the cluster resource has first been successfully provisioned. If you skip straight to the update function from the create function to apply the `key_vault_properties` the cluster will not have the correct permissions to access the key vault and will always receive the following error: 

```
Error: updating Log Analytics Cluster "acctest-LA-201112132727764751" (Resource Group "acctestRG-la-201112132727764751"): operationalinsights.ClustersClient#Update: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="BadRequest" Message="Failed to get access to encryption key https://vault201112132727764751.vault.azure.net:443/keys/key-98scj/***** due to the next error: Forbidden"
```

Is there a particular why you would prefer me to implement this check differently than I had originally implemented it?

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
jackofallops: Nope, that's fair, I'd forgotten the precedent and intent, yesterday was a long day!

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
jackofallops: Apologies, I was suggesting the ID of the key (in `azure/helpers/key_vault_child.go`), rather than the Key Vault, as that contains all 3 components noted here.  There's support for specifying a key without version also (which specifies the version as `""`, interpreted by Key Vault as `latest`, so that may need validating against what this service expects and the behaviour it creates.)

azurerm/internal/services/loganalytics/log_analytics_clusters_resource.go
jackofallops: I think we should, at least that can then be added to the comment below for the official answer and saves us asking the question again at a later date?

azurerm/internal/services/loganalytics/log_analytics_cluster_customer_managed_key_resource.go
```diff
@@ -0,0 +1,205 @@
...
+			KeyVaultProperties: &operationalinsights.KeyVaultProperties{
+				KeyVaultURI: nil,
+				KeyName:     nil,
+				KeyVersion:  nil,
+			},
+		},
+	}
+
+	_, err = client.Update(ctx, clusterId.ResourceGroup, clusterId.Name, clusterPatch)
+	if err != nil {
```
katbyte, line 194: these two lines could be merged

azurerm/internal/services/loganalytics/log_analytics_cluster_resource.go
```diff
@@ -0,0 +1,300 @@
...
+	if err != nil {
+		return fmt.Errorf("creating Log Analytics Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("waiting on creating future for Log Analytics Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	_, err = client.Get(ctx, resourceGroup, name)
+	if err != nil {
```
katbyte, line 149: these two lines could be merged

azurerm/internal/services/loganalytics/log_analytics_cluster_resource.go
```diff
@@ -0,0 +1,300 @@
...
+		Capacity: utils.Int64(int64(d.Get("size_gb").(int))),
+		Name:     operationalinsights.CapacityReservation,
+	}
+
+	parameters := operationalinsights.Cluster{
+		Location: utils.String(location.Normalize(d.Get("location").(string))),
+		Identity: expandArmLogAnalyticsClusterIdentity(d.Get("identity").([]interface{})),
+		Sku:      sku,
+		Tags:     tags.Expand(d.Get("tags").(map[string]interface{})),
+	}
```
katbyte, line 137: this could be
```suggestion
	parameters := operationalinsights.Cluster{
		Location: utils.String(location.Normalize(d.Get("location").(string))),
		Identity: expandArmLogAnalyticsClusterIdentity(d.Get("identity").([]interface{})),
		Sku:      sku := &operationalinsights.ClusterSku{
			Capacity: utils.Int64(int64(d.Get("size_gb").(int))),
			Name:     operationalinsights.CapacityReservation,
		},
		Tags:     tags.Expand(d.Get("tags").(map[string]interface{})),
	}
```

katbyte (approved):
Aside from a few comments this LGTM 👍

jackofallops:
Tests passing:
image

PR #8952, opened 2020-10-20
Dash shell compliant check-test-package.sh

richeney:
Dash is the default SHELL for the Travis-CI jobs. 

Many make commands run fmtcheck that runs ./scripts/check-test-package.sh. This had a Bash/POSIX style `[[ textexpr ]]` that caused the following errors.

```text
./scripts/check-test-package.sh: 11: [[: not found
./scripts/check-test-package.sh: 11: [[: not found
./scripts/check-test-package.sh: 11: [[: not found
./scripts/check-test-package.sh: 11: [[: not found
 <repeated>
```

The script was therefore not successfully checking for acceptance test files using the same package and could have missed circular dependencies.

New version is Dash shell compliant.

### DIFF
--- a/scripts/check-test-package.sh
+++ b/scripts/check-test-package.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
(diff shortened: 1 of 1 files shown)


tombuildsstuff (approved):
LGTM - thanks for this @richeney

PR #8978, opened 2020-10-22
New resource: `azurerm_virtual_hub_security_partner_provider`

neil-yechenwei:
API spec: https://docs.microsoft.com/en-us/rest/api/virtualnetwork/securitypartnerproviders/createorupdate


website/docs/r/security_partner_provider.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) The name which should be used for this Security Partner Provider. Changing this forces a new resource to be created.
+
+* `resource_group_name` - (Required) The name of the Resource Group where the Security Partner Provider should exist. Changing this forces a new resource to be created.
+
+* `location` - (Required) The Azure Region where the Security Partner Provider should exist. Changing this forces a new resource to be created.
+
+* `security_provider_type` - (Required) The security provider name. Possible values are `ZScaler`, `IBoss` and `Checkpoint` is allowed. Changing this forces a new resource to be created.
```
katbyte, line 67: Given the docs say name should this be
```suggestion
* `security_provider_name` - (Required) The security provider name. Possible values are `ZScaler`, `IBoss` and `Checkpoint` is allowed. Changing this forces a new resource to be created.
```

website/docs/r/security_partner_provider.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+
+* `virtual_hub_id` - (Optional) The ID of the Virtual Hub within which this Security Partner Provider should be created. Changing this forces a new resource to be created.
```
katbyte, line 69: What happens when no vhub is specified?

azurerm/internal/services/network/tests/security_partner_provider_resource_test.go
```diff
@@ -0,0 +1,229 @@
...
+`, config)
+}
+
+func testAccAzureRMSecurityPartnerProvider_complete(data acceptance.TestData) string {
+	template := testAccAzureRMSecurityPartnerProvider_template(data)
+	return fmt.Sprintf(`
+%s
+
+resource "azurerm_security_partner_provider" "test" {
+  name                   = "acctest-spp-%d"
```
katbyte, line 216: Could we ensure capitals work
```suggestion
  name                   = "acctest-SPP-%d"
```

azurerm/internal/services/network/tests/security_partner_provider_resource_test.go
```diff
@@ -0,0 +1,229 @@
...
+  resource_group_name    = azurerm_resource_group.test.name
+  location               = azurerm_resource_group.test.location
+  virtual_hub_id         = azurerm_virtual_hub.test.id
+  security_provider_type = "ZScaler"
+
+  tags = {
+    ENV = "test"
```
katbyte, line 223: and here
```suggestion
    ENv = "Test"
```

azurerm/internal/services/network/registration.go
```diff
@@ -102,6 +102,7 @@ func (r Registration) SupportedResources() map[string]*schema.Resource {
 		"azurerm_route_filter":                                                           resourceArmRouteFilter(),
 		"azurerm_route_table":                                                            resourceArmRouteTable(),
 		"azurerm_route":                                                                  resourceArmRoute(),
+		"azurerm_security_partner_provider":                                              resourceArmSecurityPartnerProvider(),
```
katbyte, line 105: I think we should have this be a little more specific? what is this security partner applied to? vhubs? then should this be:
```suggestion
		"azurerm_virtual_hub_security_partner_provider":                                              resourceArmSecurityPartnerProvider(),
```

katbyte (changes requested):
Thanks for the pr @neil-yechenwei  - overall looks good but i have some questions about resource and schema naming. please see comments left inline

website/docs/r/security_partner_provider.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+subcategory: "Network"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_security_partner_provider"
+description: |-
+  Manages a Security Partner Provider.
+---
+
+# azurerm_security_partner_provider
+
+Manages a Security Partner Provider.
```
katbyte, line 11: Could we elaborate on what a SPP is and what it is applied to?

website/docs/r/security_partner_provider.html.markdown
neil-yechenwei: updated

website/docs/r/security_partner_provider.html.markdown
neil-yechenwei: The independent security partner provider would be created and it wouldn't be associated to virtual hub.

azurerm/internal/services/network/tests/security_partner_provider_resource_test.go
neil-yechenwei: updated

azurerm/internal/services/network/tests/security_partner_provider_resource_test.go
neil-yechenwei: updated

azurerm/internal/services/network/registration.go
neil-yechenwei: updated

website/docs/r/security_partner_provider.html.markdown
neil-yechenwei: updated

neil-yechenwei:
@katbyte , thanks for your comments. I've updated code. Please have a look. Thanks.

katbyte (approved):

PR #8979, opened 2020-10-22
azurerm_virtual_network: support for `bgp_community` and `vnet_protection_enabled`

magodo:
Support for `bgp_community` and `vnet_protection_enabled` properties for vnet.

### DIFF
--- a/azurerm/internal/services/network/tests/virtual_network_resource_test.go
+++ b/azurerm/internal/services/network/tests/virtual_network_resource_test.go
@@ -203,6 +203,79 @@ func TestAccAzureRMVirtualNetwork_deleteSubnet(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMVirtualNetwork_bgpCommunity(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_virtual_network", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMVirtualNetworkDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMVirtualNetwork_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVirtualNetworkExists(data.ResourceName),
+				),
+			},
+			data.ImportStep(),
+			{
+				Config: testAccAzureRMVirtualNetwork_bgpCommunity(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVirtualNetworkExists(data.ResourceName),
+				),
+			},
+			data.ImportStep(),
+			data.ImportStep(),
+			{
+				Config: testAccAzureRMVirtualNetwork_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVirtualNetworkExists(data.ResourceName),
+				),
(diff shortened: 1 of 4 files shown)


azurerm/internal/services/network/virtual_network_resource.go
```diff
@@ -63,6 +65,14 @@ func resourceArmVirtualNetwork() *schema.Resource {
 				},
 			},
 
+			"bgp_community": {
```
katbyte, line 68: should this be
```suggestion
			"bgp_community_asn": {
```

and as the first is always the microsoft one, should we just take a single int and then silently send the microsoft one? if it does change in the future to allow multiple we can always add a new property

katbyte (changes requested):
thanks @magodo - just have a question about schema design but otherwise this looks pretty good

azurerm/internal/services/network/virtual_network_resource.go
magodo: I'm not quite sure whether this is a good idea, as the asn format of asn (defined by the rfc1997) is to use this 32bit as a whole and split into first two bytes and final two bytes only:

> The rest of the community attribute values shall be encoded using an
   autonomous system number in the first two octets.  The semantics of
   the final two octets may be defined by the autonomous system.

In other word, I think if we only export the final two bytes to users, it doesn't make sense to name that property as `bgp_community_asn`. However, I can't come up with a better name 🤷‍♂️

azurerm/internal/services/network/tests/virtual_network_resource_test.go
```diff
@@ -203,6 +203,79 @@ func TestAccAzureRMVirtualNetwork_deleteSubnet(t *testing.T) {
...
+			},
+			data.ImportStep(),
+			{
+				Config: testAccAzureRMVirtualNetwork_bgpCommunity(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVirtualNetworkExists(data.ResourceName),
+				),
+			},
+			data.ImportStep(),
+			data.ImportStep(),
```
katbyte, line 228: why do we have two import steps in a row?

azurerm/internal/services/network/virtual_network_resource.go
katbyte: I've come back to this multiple times over the last week and really can't think of a good way to express this.. right now you have:
```
bgp_community = [12076, 20000]
```
we could also expose what your doing under the hood and the API expects:
```
 bgp_community = "12076:20000"
```

and then we can validate it always starts with the correct as-number.. i'm leaning towards the 2nd option even if the code would be a little messier to validate it as it matches documentation i've found online eg: https://www.juniper.net/documentation/en_US/junos/topics/concept/policy-bgp-communities-extended-communities-match-conditions-overview.html

katbyte (changes requested):
Thanks @magodo - i see yourj point and have addressed it inline

azurerm/internal/services/network/virtual_network_resource.go
magodo: Yeah, that seems better than a two-length list :) I'll update.

azurerm/internal/services/network/tests/virtual_network_resource_test.go
magodo: My bad, thank you...

magodo:
@katbyte Thank you for the comments! I have updated the bgp community format per your suggestion, please take another look!

katbyte (approved):
Thanks @magodo  - LGTM 👍

Issue #8657, opened 2020-09-28
aks: support for ephemeral disks on default & custom node pools


4c74356b41:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Add an ability to create node pools with ephemeral disks

### New or Affected Resource(s)

* azurerm_kubernetes_cluster
* azurerm_kubernetes_cluster_node_pool

### Potential Terraform Configuration
```hcl
    default_node_pool {
        xxx
        os_disk_type = "Ephemeral" # or "Managed"
        xxx
    }
```

### References
https://github.com/Azure/AKS/issues/1370
https://github.com/Azure/azure-sdk-for-go/blob/master/services/containerservice/mgmt/2020-09-01/containerservice/models.go#L1379
https://docs.microsoft.com/en-us/rest/api/aks/agentpools/createorupdate#osdisktype

davidack:
A big thumbs-up on this one, but please add `kubernetes_cluster_node_pool` to the affected resources.

For anyone not already aware, this problem can be worked around by creating a node pool with `az aks nodepool add EnableEphemeralOSDisk=true` then importing the newly created node pool into the Terraform state, but this is of course non-optimal.

4c74356b41:
@davidack yeah, sorry, my bad, forgot about it

davidack:
Thanks all for getting this done, you are awesome!

blafry:
As I can see support for `os_disk_type` was added only to `kubernetes_cluster_node_pool` resource. Thanks a lot for that!

But what about `azurerm_kubernetes_cluster` mentioned in first comment? `default_node_pool` should also support `os_disk_type`.

mrowken:
According to [AKS release notes](https://github.com/Azure/AKS/releases/tag/2020-11-16) Ephemeral OS is now GA and used as default.
Current behavior of `azurerm_kubernetes_cluster` makes it impossible to use this feature, because it always sets `default_node_pool->os_disk_type = "Managed"` and doesn't expose parameter to override.

Could you please prioritize this change or is there any particular reason for being missing ?

PR #8982, opened 2020-10-22
Upgrade containerservice API to 2020-09-01

favoretti:
In prep for #8657


favoretti:
Fixing acceptance tests atm, will take a few more days I think...

favoretti:
@tombuildsstuff I think now we should be ready to review this.

azurerm/internal/services/containers/kubernetes_nodepool.go
```diff
@@ -102,15 +102,6 @@ func SchemaDefaultNodePool() *schema.Schema {
...
 				},
 
-				"node_taints": {
-					Type:     schema.TypeList,
-					ForceNew: true,
-					Optional: true,
-					Elem: &schema.Schema{
-						Type: schema.TypeString,
-					},
-				},
```
tombuildsstuff, line 112: this'd be a breaking change - can we revert this and deprecate this field (with a message) and error if this is set during creation?

azurerm/internal/services/containers/tests/kubernetes_cluster_data_source_test.go
```diff
@@ -500,7 +499,11 @@ func testAccDataSourceAzureRMKubernetesCluster_addOnProfileAzurePolicy(t *testin
 					testCheckAzureRMKubernetesClusterExists(data.ResourceName),
 					resource.TestCheckResourceAttr(data.ResourceName, "addon_profile.0.azure_policy.#", "1"),
 					resource.TestCheckResourceAttr(data.ResourceName, "addon_profile.0.azure_policy.0.enabled", "true"),
-					resource.TestCheckResourceAttr(data.ResourceName, "addon_profile.0.azure_policy.0.version", "v2"),
+					// === RUN   TestAccAzureRMKubernetes_all/datasource/addOnProfileAzurePolicy
+					//    testing.go:684: Step 0 error: Check failed: Check 4/4 error: azurerm_kubernetes_cluster.test: Attribute 'addon_profile.0.azure_policy.0.version' not found
+					// It seems like it doesn't return version property anymore?
```
tombuildsstuff, line 504: entirely possibly, the transition period wasn't that long so maybe we need to default this in the response now?

tombuildsstuff (changes requested):
hey @favoretti 

As discussed offline, I was hoping to dig in deeper into these comments today to be able to confirm the migration path (e.g. if we should rip these out, since they're no longer usable due to the breaking change on the AKS API side - or if we should keep them supported and deprecated for now, raising an error if the fields specified).

Unfortunately I didn't get a chance to do that today - but generally speaking we'd tend to deprecate this field, and then remove it in 3.0 - rather than ripping this out and introducing a breaking config change - to follow the semantic versioning rules here (the best we can, anyway). What that means is introducing a "DeprecationMessage" to the items in the schema to raise the deprecation alert in the Terraform CLI - and an error during creation for the Pod Security Policy which can no longer be configured, if that's set to true - what do you think?

Thanks!

azurerm/internal/services/containers/kubernetes_cluster_node_pool_resource.go
```diff
@@ -76,6 +76,7 @@ func resourceArmKubernetesClusterNodePool() *schema.Resource {
 			"availability_zones": {
 				Type:     schema.TypeList,
 				Optional: true,
+				Computed: true,
```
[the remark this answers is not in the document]
favoretti: Because this property is always returned apparently. I wasn't too sure how to go about this to be honest.

favoretti:
Hey @tombuildsstuff, thank you for the review! Committed your suggestions, will work on deprecation tomorrow/monday.

favoretti:
@tombuildsstuff resolved everything but 2 things, will try to catch you on slack tomorrow to talk about those, not really sure what to do about them.

azurerm/internal/services/containers/kubernetes_nodepool.go
```diff
@@ -368,7 +367,6 @@ func FlattenDefaultNodePool(input *[]containerservice.ManagedClusterAgentPoolPro
 			"name":                  name,
 			"node_count":            count,
 			"node_labels":           nodeLabels,
-			"node_taints":           nodeTaints,
```
tombuildsstuff, line 371: so this isn't a breaking change, for now we'll need to set an empty list into the state:

```suggestion
			"node_taints":           []string{},
```

(we can remove this once we remove the field in 3.0)

tombuildsstuff (approved):
hey @favoretti 

Thanks for pushing those changes - taking a look through, aside from one minor comment this otherwise LGTM 👍 

Thanks!

tombuildsstuff:
As discussed offline - the "addOnProfile" blocks are intentionally failing (both in master and this branch) due to a behavioural difference between Azure Public and other Azure Environments - where the Kubernetes Dashboard is enabled by default in Azure Public up until Kubernetes 1.18/1.19 (at which point it's no longer available), but isn't available (at all) in some other Azure Environments (China, USGovernment). As such whilst it's unfortunate these tests fail, it's required to be able to covers these scenarios when using different versions of AKS in different Azure Environments - but the resource itself works fine 👍

tombuildsstuff (approved):
LGTM 👍

tombuildsstuff:
Ignoring some intentionally failing tests for the "AddOnProfile", due to the behaviour of AKS in Azure Public - and one temporary AKS failure the tests otherwise look good:

Screenshot 2020-11-05 at 11 08 40

PR #8983, opened 2020-10-22
r/key_vault: ensuring the casing on the sku is consistent

### LINKED ISSUE #8457: Incorrect casing for SKU name values when importing azurerm_key_vault

tombuildsstuff:
Matching the SKU Casing with what the SDK expects.

Supersedes #8467
Fixes #8457

### DIFF
--- a/azurerm/internal/services/keyvault/key_vault_resource.go
+++ b/azurerm/internal/services/keyvault/key_vault_resource.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/keyvault/validate"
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
Tests pass:

Screenshot 2020-10-22 at 14 18 25

manicminer (approved):
LGTM :+1:

PR #8991, opened 2020-10-22
generator: running make generate

tombuildsstuff:
I ran `go install` rather than `make` locally, so these didn't get generated 🤦


manicminer (approved):
LGTM!

PR #9004, opened 2020-10-23
Storage Account Minimum TLS defaults to 1.2, rather than 1.0

HarleyB123:
I noticed that the default for the minimum TLS version for Storage Accounts is 1.0, which isn't inline with the TLS version defaults we see in [function apps](https://www.terraform.io/docs/providers/azurerm/r/function_app.html#min_tls_version) or [app services](https://www.terraform.io/docs/providers/azurerm/r/app_service.html#min_tls_version).

Have updated the code and the docs so that Storage Account TLS is now 1.2 by default

### DIFF
--- a/.github/workflows/thirty-two-bit.yaml
+++ b/.github/workflows/thirty-two-bit.yaml
@@ -9,7 +9,7 @@ on:
       - '.github/workflows/**'
 
 jobs:
-  website-test:
+  compatability-32bit-test:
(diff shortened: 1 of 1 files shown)


HarleyB123:
Hey 👋

Is anything extra needed to merge this? 😄

HarleyB123:
Apologies!!! Will reopen this

PR #9019, opened 2020-10-26
azurerm_key_vault - support export `private_endpoint_connection`

njuCZ:


azurerm/internal/services/keyvault/key_vault_resource.go
```diff
@@ -195,6 +195,42 @@ func resourceArmKeyVault() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"private_endpoint_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"connection_state": {
```
tombuildsstuff, line 209: this doesn't make sense to expose in Terraform - it's not a field that users can do anything with - can we remove this?

azurerm/internal/services/keyvault/key_vault_resource.go
```diff
@@ -195,6 +195,42 @@ func resourceArmKeyVault() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"private_endpoint_connection": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"private_endpoint_id": {
```
tombuildsstuff, line 204: this can be moved to a top-level field, I guess, but it's not immediately useful here since this won't be available until a Private Endpoint is connected via a separate resource - so ultimately this doesn't appear to warrant exposing on this resource?

tombuildsstuff (changes requested):
hi @njuCZ 

Thanks for this PR.

Taking a look through here it appears this information isn't usable to expose from Terraform, either because this information isn't available at the time the resource is created (for both items) and (for the `connection_state` block) isn't something users can action in Terraform.

Whilst it could make sense to expose this information on the Data Source (since you could chain them together in such an order that this Private Endpoint ID is available) - ultimately this isn't data that's usable in Terraform, and is something that folks would query with the Azure CLI instead (since you can't action any of the fields in the `connection_state` here).

Since these fields aren't usable at this scope whilst I'd like to thank you for this PR, I'm going to close this for the moment - however it could potentially make use to expose the Private Endpoint ID on the Data Source?

Thanks!

tombuildsstuff:
Closing as discussed above.

PR #9024, opened 2020-10-26
Fix App Service documentation for Linux plans

danner-swap:
When creating a Linux-based app service plan it was previously documented that "kind" should equal "FunctionApp" and "reserved" should equal true. This is no longer true and returns the following error: "Error: `reserved` has to be set to false when kind isn't set to `Linux`". Updated documentation to match intended usage of setting "kind" to "Linux" for linux-baseed app service plans.


favoretti:
Aside of documentation fix, wouldn't it be better to also add this to schema validation to error out on plan phase, rather than on apply?

katbyte:
@favoretti - agreed, I think it would be

katbyte (approved):
Thanks @danner-swap - going to merge this without validate for now, that can be added later

PR #9034, opened 2020-10-27
azurerm_log_analytics_saved_search - support for `tags`

dw511214992:


azurerm/internal/services/loganalytics/log_analytics_saved_search_resource.go
```diff
@@ -94,6 +94,26 @@ func resourceArmLogAnalyticsSavedSearch() *schema.Resource {
 					),
 				},
 			},
+
+			// tag is defined as an object set and cannot be updated in service side
+			"tag": {
+				Type:     schema.TypeSet,
```
tombuildsstuff, line 100: since this contains a key and a value we should be able to use the regular `tags.ForceNewSchema()` here instead

ultimately this likely wants to be named `tags` too?

azurerm/internal/services/loganalytics/tests/log_analytics_saved_search_resource_test.go
```diff
@@ -205,3 +224,41 @@ resource "azurerm_log_analytics_saved_search" "test" {
...
+  display_name = "Create or Update Saved Search Test"
+  query        = "Heartbeat | summarize Count() by Computer | take a"
+  tag {
+    name  = "Environment"
+    value = "test"
+  }
+
+  tag {
+    name  = ""
+    value = ""
```
tombuildsstuff, line 260: this configuration shouldn't be valid - can we add better test data here?

tombuildsstuff (changes requested):
hi @dw511214992 

Thanks for this PR.

Taking a look through here it appears that we could use a regular Tags Schema here instead of this custom object?

Thanks!

dw511214992:
Thanks @tombuildsstuff I have updated the pr. please help re-review it.

azurerm/internal/services/loganalytics/tests/log_analytics_saved_search_resource_test.go
```diff
@@ -205,3 +224,35 @@ resource "azurerm_log_analytics_saved_search" "test" {
...
+
+resource "azurerm_log_analytics_saved_search" "test" {
+  name                       = "acctestLASS-%d"
+  log_analytics_workspace_id = azurerm_log_analytics_workspace.test.id
+
+  category     = "Saved Search Test Category"
+  display_name = "Create or Update Saved Search Test"
+  query        = "Heartbeat | summarize Count() by Computer | take a"
+  tags = {
+    "Environment" = "test"
```
katbyte, line 254: we should always ensure capitals work for both key and value
```suggestion
    "Environment" = "Test"
```

azurerm/internal/services/loganalytics/tests/log_analytics_saved_search_resource_test.go
```diff
@@ -205,3 +224,35 @@ resource "azurerm_log_analytics_saved_search" "test" {
...
+}
+
+resource "azurerm_log_analytics_saved_search" "test" {
+  name                       = "acctestLASS-%d"
+  log_analytics_workspace_id = azurerm_log_analytics_workspace.test.id
+
+  category     = "Saved Search Test Category"
+  display_name = "Create or Update Saved Search Test"
+  query        = "Heartbeat | summarize Count() by Computer | take a"
+  tags = {
```
katbyte, line 253:
```suggestion

  tags = {
```

website/docs/r/log_analytics_saved_search.html.markdown
```diff
@@ -54,6 +54,8 @@ The following arguments are supported:
 
 * `function_parameters` - (Optional) The function parameters if the query serves as a function. Changing this forces a new resource to be created.
 
+* `tag` - (Optional) The tag of Saved Search. Changing this forces a new resource to be created.
```
katbyte, line 57: we need to update this to match schema

```suggestion
* `tags` - (Optional) A mapping of tags which should be assigned to the Logs Analytics Saved Search.
```

katbyte (approved):
Thanks @dw511214992 - i've given this a review and left some comments inline which i will address now. LGTM 👍

PR #9035, opened 2020-10-27
azurerm_vpn_gateway - support block "bgp_peering_addresses"

njuCZ:
After confirmed with service team, this field "bgp_peering_addresses" could not be set when creating, otherwise the rest api will report error. It could only be updated.

image

### DIFF
--- a/azurerm/internal/services/network/tests/vpn_gateway_resource_test.go
+++ b/azurerm/internal/services/network/tests/vpn_gateway_resource_test.go
@@ -225,14 +225,12 @@ resource "azurerm_vpn_gateway" "test" {
     asn         = 65515
     peer_weight = 0
 
-    bgp_peering_addresses {
-      ip_configuration_id     = "Instance0"
-      custom_bgp_ip_addresses = ["169.254.21.5"]
+    instance_bgp_peering_address {
+      custom_ips = ["169.254.21.5"]
     }
 
-    bgp_peering_addresses {
-      ip_configuration_id     = "Instance1"
-      custom_bgp_ip_addresses = ["169.254.21.10"]
+    instance_bgp_peering_address {
+      custom_ips = ["169.254.21.10"]
     }
   }
 }
--- a/azurerm/internal/services/network/tests/vpn_gateway_resource_test.go
+++ b/azurerm/internal/services/network/tests/vpn_gateway_resource_test.go
@@ -225,11 +225,11 @@ resource "azurerm_vpn_gateway" "test" {
     asn         = 65515
     peer_weight = 0
 
-    instance_bgp_peering_address {
+    instance_0_bgp_peering_address {
       custom_ips = ["169.254.21.5"]
     }
 
-    instance_bgp_peering_address {
+    instance_1_bgp_peering_address {
       custom_ips = ["169.254.21.10"]
     }
   }
--- a/azurerm/internal/services/network/tests/vpn_gateway_resource_test.go
+++ b/azurerm/internal/services/network/tests/vpn_gateway_resource_test.go
@@ -224,6 +224,16 @@ resource "azurerm_vpn_gateway" "test" {
   bgp_settings {
     asn         = 65515
     peer_weight = 0
+
+    bgp_peering_addresses {
+      ip_configuration_id     = "Instance0"
+      custom_bgp_ip_addresses = ["169.254.21.5"]
+    }
+
+    bgp_peering_addresses {
+      ip_configuration_id     = "Instance1"
+      custom_bgp_ip_addresses = ["169.254.21.10"]
+    }
   }
 }
 `, template, data.RandomInteger)
--- a/azurerm/internal/services/network/vpn_gateway_resource.go
+++ b/azurerm/internal/services/network/vpn_gateway_resource.go
@@ -6,8 +6,6 @@ import (
 	"log"
 	"time"
 
-	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/validate"
-
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-05-01/network"
 	"github.com/hashicorp/go-azure-helpers/response"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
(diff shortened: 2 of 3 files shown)


azurerm/internal/services/network/vpn_gateway_resource.go
```diff
@@ -78,6 +79,45 @@ func resourceArmVPNGateway() *schema.Resource {
...
+						"bgp_peering_addresses": {
+							Type:     schema.TypeSet,
+							Optional: true,
+							Computed: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"ip_configuration_id": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
```
katbyte, line 92: could we validate this?

website/docs/r/vpn_gateway.html.markdown
```diff
@@ -75,6 +75,16 @@ A `bgp_settings` block supports the following:
 
 * `peer_weight` - (Required) The weight added to Routes learned from this BGP Speaker. Changing this forces a new resource to be created.
 
+* `bgp_peering_addresses` - (Optional) One or more `bgp_peering_addresses` block as defined below.
+
+---
+
+A `bgp_peering_addresses` block supports the following:
+
+* `ip_configuration_id` - (Required) The ID of IP configuration which belongs to gateway.
```
katbyte, line 84: Could we explain what this property is more? configuration is very vauge

website/docs/r/vpn_gateway.html.markdown
```diff
@@ -75,6 +75,16 @@ A `bgp_settings` block supports the following:
...
+
+* `custom_bgp_ip_addresses` - (Required) The list of custom BGP peering addresses which belong to IP configuration.
```
katbyte, line 86: grammer
```suggestion
* `custom_bgp_ip_addresses` - (Required) The list of custom BGP peering addresses which belong to the above IP configuration.
```

azurerm/internal/services/network/vpn_gateway_resource.go
```diff
@@ -78,6 +79,45 @@ func resourceArmVPNGateway() *schema.Resource {
...
 							Computed: true,
 						},
+
+						"bgp_peering_addresses": {
+							Type:     schema.TypeSet,
+							Optional: true,
+							Computed: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"ip_configuration_id": {
```
katbyte, line 89: Should this be called something else? ip configuration id implied an azure resource, when its a text string? 

what does this value represent and is there a more descriptive/accurate name?

azurerm/internal/services/network/vpn_gateway_resource.go
```diff
@@ -78,6 +79,45 @@ func resourceArmVPNGateway() *schema.Resource {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
+
+						"bgp_peering_addresses": {
```
katbyte, line 83: should this be 
```suggestion
						"bgp_peering_ip_addresses": {
```

or
```suggestion
						"ip_bgp_peering_addresses": {
						"instance_bgp_peering_addresses": {
```
given that ip_configuration_id seems to tie it into something that is not an ip resource id?

also the properties within could be
ip_???
custom_ips
default_ips
tunnel_ips

givenits within a block `bgp_peering_addresses` i think we can drop it from all the contained properties

website/docs/r/vpn_gateway.html.markdown
```diff
@@ -89,6 +99,16 @@ A `bgp_settings` block exports the following:
...
 
+* `bgp_peering_addresses` - One or more `bgp_peering_addresses` block as defined below.
+
+---
+
+A `bgp_peering_addresses` block exports the following:
+
+* `default_bgp_ip_addresses` - The list of default BGP peering addresses which belong to IP configuration.
+
+* `tunnel_ip_addresses` - The list of tunnel public IP addresses which belong to IP configuration.
```
katbyte, line 110: grammer and could we expand on this

website/docs/r/vpn_gateway.html.markdown
```diff
@@ -89,6 +99,16 @@ A `bgp_settings` block exports the following:
 
 * `bgp_peering_address` - The Address which should be used for the BGP Peering.
 
+* `bgp_peering_addresses` - One or more `bgp_peering_addresses` block as defined below.
+
+---
+
+A `bgp_peering_addresses` block exports the following:
+
+* `default_bgp_ip_addresses` - The list of default BGP peering addresses which belong to IP configuration.
```
katbyte, line 108: grammer and could we expand on this

katbyte (changes requested):
Thanks for the PR @njuCZ - however i think the block/properties are poorly named and confusing.

it seems like the block is tied to an "instance" and then these are the addresses for it? we should name the block something that is a bit more explicit, like `bgp_ip|instance_peering_addresses` and then inside `ip|instance_id` ?

njuCZ:
Hi @katbyte thanks for your review. After consulting the service team further, they say `IpconfigurationID` is fixed to `Instance0` and `Instance1`. Users could not change it.

The portal's experience is: creating VPN Gateway only needs parameters of `AS number` and `gateway scale units`. and updating could only update custom bgp ip address, the picture is shown blow
image

So my sugggestion is to make `IpconfigurationID` computed, and limit the size of `instance_bgp_peering_address` block to 2. Do you think it is ok? I have modified this PR, could you have a look again?

azurerm/internal/services/network/tests/vpn_gateway_resource_test.go
```diff
@@ -224,6 +224,14 @@ resource "azurerm_vpn_gateway" "test" {
   bgp_settings {
     asn         = 65515
     peer_weight = 0
+
+    instance_bgp_peering_address {
```
katbyte, line 228: This isn't exactly clear, maybe it would be best to create two named blocks?
```suggestion
    instance_0_bgp_peering_address {
```

azurerm/internal/services/network/tests/vpn_gateway_resource_test.go
```diff
@@ -224,6 +224,14 @@ resource "azurerm_vpn_gateway" "test" {
...
+      custom_ips = ["169.254.21.5"]
+    }
+
+    instance_bgp_peering_address {
```
katbyte, line 232: and then
```suggestion
    instance_1_bgp_peering_address {
```

azurerm/internal/services/network/vpn_gateway_resource.go
```diff
@@ -78,6 +80,47 @@ func resourceArmVPNGateway() *schema.Resource {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
+
+						"instance_bgp_peering_address": {
```
katbyte, line 84: Maybe its best to split this into two named blocks to match the portal?

katbyte (changes requested):
Hey @njuCZ - i've been thinking about this on and off for a week and i dont' this the current schema is very clear or good ux. I've left some comments inline but basically i think creating two named blocks might be the best and most clear solution here? WDYT?

njuCZ:
@katbyte thanks for your review, I have modifed this PR, please have a look when free

katbyte (approved):
Thanks @njuCZ - LGTM 👍

PR #9049, opened 2020-10-28
`azurerm_virtual_network` - make `address_space` to be `ForceNew`

magodo:
This PR consists of two commits:

1. Makes the `address_space` to be `ForceNew` for `azurerm_virtual_network` and deprecate the `virtual_network_name` and `resource_group_name` in favor of `virtual_network_id` for `azurerm_subnet`. 
2. Flush all the TF snippet in acctests and document for the azurerm_subnet to using the `virtual_network_id`.

This fixex #5235


magodo:
This PR might not work as a Virtual Network can contain multiple Address Prefixes (CIDR blocks), so if you have Subnets/NICs in one subnet and you want to add a second unused CIDR block, that’s not possible without recreating all of the other infrastructure (e.g. subnets), which is mostly unexpected.

Another possible option is to suggest users to using the [`cidrsubnet`](https://www.terraform.io/docs/configuration/functions/cidrsubnet.html) function in the subnet so as to add a relationship between the vnet and the subnet. And once the affected vnet address space got updated, we shall only recreate the affected subnets. However, for the same reason that the cidr in subnet is (maybe) also updatable, users should not anticipate to seeing all the infrastructures in the subnet?

favoretti:
Also if your VNET has peerings one can't update CIDR blocks either, which is... Less than optimal. Maybe we should log a bug/feature req for MSFT instead?

tombuildsstuff:
As discussed offline, as there's multiple use-cases when changing this field this is unfortunately a complex tradeoff - whilst this PR does fix the scenario where the CIDR Block changes between two incompatible versions (which can be achieved by tainting/destroying the resources) - it's more common to see a second address prefix added, which this would break (and require destroying all infrastructure for a currently non-breaking change) and as such I'm going to close this PR for the moment.

PR #9050, opened 2020-10-28
Remove cosmosdb autoscale upper cap

favoretti:
MSFT support can scale accounts and containers to the skies (largest
we've heard was around 30M RUs), so capping this in terraform breaks
custom accounts and containers scaled to large ~~wallets~~ volumes.

### DIFF
--- a/azurerm/helpers/validate/cosmos_test.go
+++ b/azurerm/helpers/validate/cosmos_test.go
@@ -133,6 +133,10 @@ func TestCosmosMaxThroughput(t *testing.T) {
 			Errors: 0,
 		},
 		{
+			Value:  1100000,
+			Errors: 0,
+		},
+		{
 			Value:  "400",
 			Errors: 1,
 		},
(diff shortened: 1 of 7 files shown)


WodansSon (approved):
@favoretti thanks for submitting this PR, these changes LGTM. I think it is safe to ignore the lint errors as they are not related to your changes. 🚀

azurerm/helpers/validate/cosmos_test.go
```diff
@@ -132,10 +132,6 @@ func TestCosmosMaxThroughput(t *testing.T) {
 			Value:  1000000,
 			Errors: 0,
 		},
-		{
-			Value:  1100000,
-			Errors: 1,
-		},
```
WodansSon, line 138: Personally I would leave this test in the test suite but change Errors to expect 0 to verify that the upper limit on throughput has been removed. WDYT?

favoretti:
Thanks for fixing gents!

WodansSon:
@jackofallops thanks for pushing those changes, this looks perfect now! 🚀

PR #9051, opened 2020-10-28
azurerm_private_endpoint subresource_names parameter should be marked as required instead of optional

schugDev:
I tried to create a private endpoint and get the following message:
Error: Error creating Private Endpoint "xxxxxxxt"  due to missing 'group Id', ensure that the 'subresource_names' type is populated: network.PrivateEndpointsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="MissingParameterOnPrivateLinkServiceConnection" Message="Private link service connection /subscriptions/xxxxxxx/resourceGroups/xxxxxxxxxxxxxxx/providers/Microsoft.Network/privateEndpoints/xxxxxxxxxxxxxxxx/privateLinkServiceConnections/xxxxx-privateserviceconnection is missing required parameter 'group Id'." Details=[]

So the subresource_names parameter should be marked as required instead of optional

### DIFF
--- a/website/docs/r/private_endpoint.html.markdown
+++ b/website/docs/r/private_endpoint.html.markdown
@@ -135,7 +135,7 @@ A `private_service_connection` supports the following:
 
 * `private_connection_resource_id` - (Required) The ID of the Private Link Enabled Remote Resource which this Private Endpoint should be connected to. Changing this forces a new resource to be created.
 
-* `subresource_names` - (Optional) A list of subresource names which the Private Endpoint is able to connect to. `subresource_names` corresponds to `group_id`. Changing this forces a new resource to be created.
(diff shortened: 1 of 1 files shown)


jackofallops:
Hi @schugDev - Thanks for this PR.

To the best of my knowledge the attribute in question is not `Required` in every circumstance, so is marked in the [schema](https://github.com/terraform-providers/terraform-provider-azurerm/blob/b2bcde9229236680b09506b2985c79f00f05d6e1/azurerm/internal/services/network/private_endpoint_resource.go#L110) and the documentation as `Optional` purposefully. If you check some of the configs in the tests folder for that service you can see [examples](https://github.com/terraform-providers/terraform-provider-azurerm/blob/b2bcde9229236680b09506b2985c79f00f05d6e1/azurerm/internal/services/network/tests/private_endpoint_resource_test.go#L383) where that value needs to be omitted. Unfortunately we have to pass that item of complexity onto users in this case. 

I hope that clarifies the situation.

Based on the above, I'm going to close this.

Issue #9052, opened 2020-10-28
Terraform crash when applying source_image_reference with empty values


jaksonwkr:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.13.5
provider registry.terraform.io/hashicorp/azurerm v2.33.0

### Affected Resource(s)

* `azurerm_linux_virtual_machine`
* `azurerm_windows_virtual_machine`

### Terraform Configuration Files

```hcl
terraform {
  required_version = "~>0.13"
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "~>2"
    }
  }
}

provider "azurerm" {
  features {}
}

resource "azurerm_resource_group" "rg" {
  name     = "jaksonwkr"
  location = "SouthCentralUS"
}

resource "azurerm_virtual_network" "vnet" {
  name                = "jaksonwkr"
  address_space       = ["192.168.0.0/16"]
  location            = azurerm_resource_group.rg.location
  resource_group_name = azurerm_resource_group.rg.name
}

resource "azurerm_subnet" "subnet" {
  name                 = "internal"
  resource_group_name  = azurerm_resource_group.rg.name
  virtual_network_name = azurerm_virtual_network.vnet.name
  address_prefixes     = ["192.168.0.0/24"]
}

resource "azurerm_network_interface" "nic" {
  name                = "jaksonwkr-nic"
  location            = azurerm_resource_group.rg.location
  resource_group_name = azurerm_resource_group.rg.name

  ip_configuration {
    name                          = "internal"
    subnet_id                     = azurerm_subnet.subnet.id
    private_ip_address_allocation = "Dynamic"
    primary                       = true
  }
}

resource "azurerm_linux_virtual_machine" "vm" {
  name                = "jaksonwkr-vm"
  resource_group_name = azurerm_resource_group.rg.name
  location            = azurerm_resource_group.rg.location
  size                = "Standard_B1s"


  admin_username                  = "jaksonwkr"
  admin_password                  = "Jakson@123"
  disable_password_authentication = false

[...]

PR #9055, opened 2020-10-28
Validate source_image_reference values

jaksonwkr:
Validate if options on source_image_reference if not an empty string
Address the issue #9052


tombuildsstuff (approved):
LGTM - thanks for this @jaksonwkr

PR #9069, opened 2020-10-28
azurerm_function_app - make `pre_warmed_instance_count` computed to use azure's default

### LINKED ISSUE #9066: Terraform Azure Function Default Prewarmed Instances Does Not Match Azure Default
### Community Note

Terarform Version: 0.13.3
Azurerm Version: Confirmed on 2.28.0 and 2.33.0

### Affected Resource(s)

* `azurerm_function_app`

```hcl
  name     = "azure-functions-test-rg"
  location = "westus2"
}

resource "azurerm_storage_account" "example" {
  name                     = "functionsapptestsa"
  resource_group_name      = azurerm_resource_group.example.name
  location                 = azurerm_resource_group.example.location
}

resource "azurerm_app_service_plan" "example" {
  name                = "azure-functions-test-service-plan"

  sku {
    tier = "ElasticPremium"
    size = "EP1"
  }
}

resource "azurerm_function_app" "example" {
  name                       = "test-azure-functions"
  location                   = azurerm_resource_group.example.location
  resource_group_name        = azurerm_resource_group.example.name
  app_service_plan_id        = azurerm_app_service_plan.example.id
  storage_account_name       = azurerm_storage_account.example.name
  storage_account_access_key = azurerm_storage_account.example.primary_access_key
}
```

### Expected Behavior

Without explicity setting the prewarmed instances I would expect it to default to 1 as detailed by Microsoft [here](https://docs.microsoft.com/en-us/azure/azure-functions/functions-premium-plan)

> The default pre-warmed instance count is 1, and for most scenarios should remain as 1.

### Actual Behavior

Prewarmed instances set to 0
### Steps to Reproduce

1. `terraform apply`

### References

* #0000

RiverPhillips:
This resolves  #9066 by setting the default prewarmed pre_warmed_instance_count to 1


jackofallops:
Hi @RiverPhillips - Thanks for this PR. I think this was left without a default due to non-premium SKU's discarding this value, so would create a Diff as the value will be `0` when read back from the API.  As such, this would probably need to be validated / adjusted in code, or possibly just updated in the documentation for appropriate usage guidance? WDYT?

RiverPhillips:
HI @jackofallops thanks for your feedback. I think it would be sensible if we defaulted to 1 for all plans except for the consumption plan. I think that should solve the issue you raised surrounding the diff. Or we could require it be set for non-consumption SKUs.

katbyte (changes requested):
@RiverPhillips - instead of adding a default here maybe we could change it to `computed` to let azure pick the default?

RiverPhillips:
@katbyte - I hadn't considered that approach, I think it's the most practical and future proof so I have made that change

katbyte (approved):
Thanks @RiverPhillips - this LGTM 👍

PR #9081, opened 2020-10-29
Add tflint ignores

favoretti:


katbyte (approved):
Thanks @favoretti - LGTM 👍

PR #9109, opened 2020-10-30
remove throttling from go linting

jackofallops:


.github/workflows/lintrest.yaml
```diff
@@ -18,6 +18,6 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '1.14.5'
-      - run: bash scripts/gogetcookie.sh
-      - run: make tools
-      - run: GOGC=5 make lintrest
+      - uses: golangci/golangci-lint-action@v2
```
katbyte, line 21: does this assume `.golangci.ym` ? we can probably rename this step to golint rather then lint rest if it actually manages to run the full linting suite

katbyte (approved):
aside from one comment LGTM 👍

PR #9111, opened 2020-10-30
Key Vault Certificate Issuer resource doc update

kevinmarx:
The docs showed org_id as optional but it is required by the resource.

```terraform
Error: Missing required argument

  on key_vault/main.tf line 14, in resource "azurerm_key_vault_certificate_issuer" "onecert_v2":
  14: resource "azurerm_key_vault_certificate_issuer" "onecert_v2" {

The argument "org_id" is required, but no definition was found.
```


kevinmarx:
Was on outdated version of the provider

Issue #9120, opened 2020-11-02
The acceptance test for devtest policy did not flag the issue fixed in #9077


mattcowen:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

v0.13.5 and v2.34

### Affected Resource(s)

azurerm_dev_test_policy

### Expected Behavior

The acceptance test should have flagged the delete problem fixed in #9077 (for v2.33). I expected the testCheckDevTestPolicyDestroy in  /terraform-provider-azurerm/azurerm/internal/services/resource/tests/resource_arm_resource_group_test.go to have discovered the policy still existed after delete but it doesn't seem possible.

I also think that if the policy isn't found when trying to delete then it should error since if this scenario happens it causes subsequent problems with synchronicity of the state with Azure i.e. Terraform thinks there is no policy when in fact there is.

### Actual Behavior

The acceptance test passed.

rcskosir:
Thanks for opening this issue. This was a problem in the 2.x version of the provider which is no longer actively maintained.
If this is still an issue with the 4.x version of the provider please do let us know by opening a new issue, thanks!

PR #9121, opened 2020-11-02
Allow devtest policy delete to error if not found as per #9120

mattcowen:
Removing the "not found" check when deleting a policy in a devtest lab. It should error IMO. 

See #9120


manicminer:
@mattcowen Thanks for this PR. Generally speaking, a resource should handle manual deletion of a resource and so returning is the appropriate course here. If there are consistency issues with a resource not being available, that should rather be fixed at the source of the error.

PR #9131, opened 2020-11-03
Update role definition docs to include `role_definition_resource_id`

notchairmk:
Reference to `role_definition_resource_id` was previously removed from the website documentation. This updates the `role_definition_id` summary to match the import note.

There's a disconnect for `role_definition_id` as an argument vs. as an attribute, which is potentially a bug. As an argument `role_definition_id` takes a UUID compared to the computed attribute which concatenates `roleDefinitionId|scope`.


katbyte (approved):
Thanks @notchairmk - LGTM 👍

PR #9134, opened 2020-11-03
New resource: `azurerm_hci_cluster`

neil-yechenwei:
Currently, terraform doesn't support to provision Azure Stack HCI Cluster resource. So I submit this PR to support it. 

The overview of Azure Stack HCI: https://docs.microsoft.com/en-us/azure-stack/hci/
API spec: https://github.com/Azure/azure-rest-api-specs/blob/master/specification/azurestackhci/resource-manager/Microsoft.AzureStackHCI/stable/2020-10-01/azurestackhci.json

--- PASS: TestAccHCICluster_complete (126.11s)
--- PASS: TestAccHCICluster_basic (126.12s)
--- PASS: TestAccHCICluster_update (168.28s)
--- PASS: TestAccHCICluster_requiresImport (202.93s)


azurerm/internal/services/azurestackhci/tests/hyper_converged_cluster_resource_test.go
```diff
@@ -0,0 +1,209 @@
...
+  name                = azurerm_hyper_converged_cluster.test.name
+  resource_group_name = azurerm_hyper_converged_cluster.test.resource_group_name
+  location            = azurerm_hyper_converged_cluster.test.location
+  client_id           = azurerm_hyper_converged_cluster.test.client_id
+  tenant_id           = azurerm_hyper_converged_cluster.test.tenant_id
+}
+`, config)
+}
+
+func testAccAzureRMHyperConvergedCluster_complete(data acceptance.TestData) string {
```
ArcturusZhang, line 192: Could we add another test case to test the update on `tags`?

azurerm/internal/services/azurestackhci/validate/hyper_converged_cluster.go
```diff
@@ -0,0 +1,37 @@
...
+func HyperConvergedClusterName(i interface{}, k string) (warnings []string, errors []error) {
+	v, ok := i.(string)
+	if !ok {
+		errors = append(errors, fmt.Errorf("expected type of %q to be string", k))
+		return
+	}
+
+	if matched := regexp.MustCompile(`^.{1,260}$`).Match([]byte(v)); !matched {
+		errors = append(errors, fmt.Errorf("%q may only be up to 260 characters in length", k))
+	}
```
ArcturusZhang, line 34: It is hard to believe that the name of this resource has only one constraint on the length but does not have any constraint on the characters allowed.
If this is the only constraint we know for now, could we switch the regular expression to directly check the length for better readability? 
Also, it is the best if we could have the full and accurate constraint for this name - including the length constraint and any invalid characters - for instance, I believe this name that only have several spaces is clearly illegal

ArcturusZhang (changes requested):
Hi @neil-yechenwei thanks for this PR and this mostly looks good aside from some minor comment, please have a look, thanks

azurerm/internal/services/azurestackhci/tests/hyper_converged_cluster_resource_test.go
neil-yechenwei: updated

azurerm/internal/services/azurestackhci/validate/hyper_converged_cluster.go
neil-yechenwei: Sorry. I didn't find any constraint rule from azure doc and error message except length constraint. So I have to use length validation.

neil-yechenwei:
@ArcturusZhang , thanks for your comments. I've updated code.

website/docs/r/hyper_converged_cluster.html.markdown
```diff
@@ -0,0 +1,73 @@
+---
+subcategory: "AzureStackHCI"
```
katbyte, line 2:
```suggestion
subcategory: "Azure Stack HCI"
```

website/docs/r/hyper_converged_cluster.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_hyper_converged_cluster"
```
katbyte, line 4: this is referred to as azure stack HCI everywhere, so why are we using this resource name?
```suggestion
page_title: "Azure Resource Manager: azurerm_azure_stack_hci_cluster"
```
or
```suggestion
page_title: "Azure Resource Manager: azurerm_hci_cluster"
```

website/docs/r/hyper_converged_cluster.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+description: |-
+  Manages a Hyper Converged Cluster.
```
katbyte, line 6: this name doens't match the docs

website/docs/r/hyper_converged_cluster.html.markdown
neil-yechenwei: I remember you suggest to not define the name with abbreviation in previous PR review. So I updated the name with full name. Based on above options, I suggest to use `azurerm_hci_cluster` since the folder name includes "azurestackhci".

image

website/docs/r/hyper_converged_cluster.html.markdown
neil-yechenwei: updated

website/docs/r/hyper_converged_cluster.html.markdown
neil-yechenwei: updated

website/docs/r/hyper_converged_cluster.html.markdown
neil-yechenwei: Updated

neil-yechenwei:
@katbyte , thanks for your comments. I've updated code. Please have an another look. Thanks.

website/docs/r/hyper_converged_cluster.html.markdown
katbyte: Generally yes, but given all the docs use HCI i think it is the more user friendly option, hyper converged cluster is much less likely to be recognizable to a user/not show up in a search

azurerm/internal/services/azurestackhci/tests/hci_cluster_resource_test.go
```diff
@@ -0,0 +1,228 @@
...
+func testAccAzureRMHCICluster_template(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
+
+data "azurerm_client_config" "current" {}
+
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-AzureStackHCI-%d"
```
katbyte, line 156:
```suggestion
  name     = "acctestRG-hci-%d"
```

azurerm/internal/services/azurestackhci/tests/hci_cluster_resource_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/azurestackhci/parse"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func TestAccAzureRMHCICluster_basic(t *testing.T) {
```
katbyte, line 15:
```suggestion
func TestAccHCICluster_basic(t *testing.T) {
```

Lets remove AzureRM from the test names

katbyte (approved):
LGTM aside from a couple comments

azurerm/internal/services/azurestackhci/tests/hci_cluster_resource_test.go
neil-yechenwei: updated

azurerm/internal/services/azurestackhci/tests/hci_cluster_resource_test.go
neil-yechenwei: updated

neil-yechenwei:
@katbyte , thanks for your comments. I've updated code. Please have a look. Thanks.

ArcturusZhang (approved):
LGTM 👍

azurerm/internal/services/azurestackhci/hci_cluster_resource.go
```diff
@@ -0,0 +1,189 @@
...
+	resp, err := client.Get(ctx, resourceGroup, name)
+	if err != nil {
+		return fmt.Errorf("retrieving HCI Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	if resp.ID == nil || *resp.ID == "" {
+		return fmt.Errorf("empty or nil ID returned for HCI Cluster %q (Resource Group %q) ID", name, resourceGroup)
+	}
+
+	d.SetId(*resp.ID)
```
magodo, line 113: Can we construct an ID here and set the result of `id.ID()` back as the argument of `d.SetId()`?

azurerm/internal/services/azurestackhci/parse/hci_cluster.go
```diff
@@ -0,0 +1,33 @@
+package parse
```
magodo, line 1: Can we generate this file by creating a *resourceids.go* under the *azurestackhci* directory with following content:

```
package azurestackhci
 
//go:generate go run ../../tools/generator-resource-id/main.go -path=./ -name=HCIClusterId -id=/subscriptions/12345678-1234-9876-4563-123456789012/resourceGroups/resGroup1/providers/Microsoft.AzureStackHCI/Clusters/cluster1
```

azurerm/internal/services/azurestackhci/parse/hci_cluster_test.go
```diff
@@ -0,0 +1,72 @@
+package parse
```
magodo, line 1: This one should also be replaced by the generated one.

azurerm/internal/services/azurestackhci/tests/hci_cluster_resource_test.go
```diff
@@ -0,0 +1,228 @@
+package tests
```
magodo, line 1: Can we convert this to the latest test code style?

azurerm/internal/services/azurestackhci/validate/hci_cluster.go
```diff
@@ -0,0 +1,37 @@
+package validate
```
magodo, line 1: This also wants to be auto generated.
Once you opt in the codegen for ID validate, you will need to move the `HCIClusterName()` to a new file named *hci_cluster_name.go*.

magodo (changes requested):
Thanks for this PR!
I've taken a look through and mostly LGTM! Whilst since this is a new RP, can we opt-in the codegen for the ID parser (and validator, together with their tests), and also opt-in the new test style?

azurerm/internal/services/azurestackhci/tests/hci_cluster_resource_test.go
neil-yechenwei: Could you share me an example? Thanks.

azurerm/internal/services/azurestackhci/validate/hci_cluster.go
neil-yechenwei: Could you share me an example? Thanks.

azurerm/internal/services/azurestackhci/parse/hci_cluster_test.go
neil-yechenwei: Could you share me an example? Thanks.

azurerm/internal/services/azurestackhci/parse/hci_cluster.go
neil-yechenwei: Could you share me an example PR? Thanks.

azurerm/internal/services/azurestackhci/hci_cluster_resource.go
neil-yechenwei: Could you share me an example PR? Thanks.

azurerm/internal/services/azurestackhci/parse/hci_cluster.go
magodo: https://github.com/magodo/terraform-provider-azurerm/blob/524170e685640a10101b6e0ac787ad06f8924cdc/azurerm/internal/services/network/resourceids.go#L2-L2

azurerm/internal/services/azurestackhci/parse/hci_cluster.go
```diff
@@ -0,0 +1,45 @@
...
+
+type HCIClusterId struct {
+	ResourceGroup string
+	Name          string
+}
+
+func (id HCIClusterId) ID(subscriptionId string) string {
+	fmtString := "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.AzureStackHCI/clusters/%s"
+	return fmt.Sprintf(fmtString, subscriptionId, id.ResourceGroup, id.Name)
+}
```
ArcturusZhang, line 17: We need to include the subscriptionId into the struct and do not use the subscriptionId parameter (which is being deprecated)
```suggestion
type HCIClusterId struct {
SubscriptionID string
	ResourceGroup string
	Name          string
}

func (id HCIClusterId) ID(_string) string {
	fmtString := "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.AzureStackHCI/clusters/%s"
	return fmt.Sprintf(fmtString, id.SubscriptionID, id.ResourceGroup, id.Name)
}
```

azurerm/internal/services/azurestackhci/hci_cluster_resource.go
```diff
@@ -0,0 +1,183 @@
...
+			AadTenantID: utils.String(d.Get("tenant_id").(string)),
+		},
+		Tags: tags.Expand(d.Get("tags").(map[string]interface{})),
+	}
+
+	if _, err := client.Create(ctx, resourceGroup, name, cluster); err != nil {
+		return fmt.Errorf("creating HCI Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	d.SetId(id.ID(subscriptionId))
```
ArcturusZhang, line 107:
```suggestion
	d.SetId(id.ID(""))
```

ArcturusZhang (changes requested):

azurerm/internal/services/azurestackhci/validate/hci_cluster.go
neil-yechenwei: updated

azurerm/internal/services/azurestackhci/parse/hci_cluster_test.go
neil-yechenwei: updated

azurerm/internal/services/azurestackhci/parse/hci_cluster.go
neil-yechenwei: updated

azurerm/internal/services/azurestackhci/hci_cluster_resource.go
neil-yechenwei: updated

azurerm/internal/services/azurestackhci/parse/hci_cluster.go
neil-yechenwei: updated

azurerm/internal/services/azurestackhci/parse/hci_cluster.go
magodo: I've submit a PR #9748 for this.

tombuildsstuff: To be honest this likely wants to be named `Cluster` since that's what this is - HCI is the product name so it's unnecessary in this context, but agreed we should update the generator to handle this case

neil-yechenwei: Once PR 9748 is merged, I will update this PR for this case.

tombuildsstuff: @neil-yechenwei this PR is not dependent on 9748, we can call this `Cluster` as mentioned above.

neil-yechenwei: updated. thx

azurerm/internal/services/azurestackhci/validate/cluster_name.go
```diff
@@ -0,0 +1,18 @@
+package validate
+
+import "fmt"
+
+func ClusterName(i interface{}, k string) (warnings []string, errors []error) {
+	v, ok := i.(string)
+	if !ok {
+		errors = append(errors, fmt.Errorf("expected type of %q to be string", k))
+		return
```
magodo, line 9: Can we be consistent here to either implicit return or explicit?

magodo (changes requested):
Thanks for this PR!
I have left some additional mostly minor comments that once addressed this should be good to merge 👍

azurerm/internal/services/azurestackhci/validate/cluster_name.go
neil-yechenwei: Good catch. Updated

magodo:
LGTM :+1:

image

azurerm/internal/services/azurestackhci/hci_cluster_resource.go
```diff
@@ -0,0 +1,183 @@
...
+
+	existing, err := client.Get(ctx, resourceGroup, name)
+	if err != nil {
+		if !utils.ResponseWasNotFound(existing.Response) {
+			return fmt.Errorf("checking for present of existing HCI Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
+		}
+	}
+
+	if existing.ID != nil && *existing.ID != "" {
+		return tf.ImportAsExistsError("azurerm_hci_cluster", *existing.ID)
```
tombuildsstuff, line 91: this should be using the Resource ID formatter's ID here: `id.ID("")`

azurerm/internal/services/azurestackhci/hci_cluster_resource.go
```diff
@@ -0,0 +1,183 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.IsUUID,
+			},
+
+			"tenant_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
```
tombuildsstuff, line 64: isn't this statistically likely to be the current Tenant ID, so shouldn't this be Optional?

azurerm/internal/services/azurestackhci/hci_cluster_resource_test.go
```diff
@@ -0,0 +1,186 @@
...
+	template := r.template(data)
+	return fmt.Sprintf(`
+%s
+
+resource "azurerm_hci_cluster" "test" {
+  name                = "acctest-HCICluster-%d"
+  resource_group_name = azurerm_resource_group.test.name
+  location            = azurerm_resource_group.test.location
+  client_id           = data.azurerm_client_config.current.client_id
+  tenant_id           = data.azurerm_client_config.current.tenant_id
```
tombuildsstuff, line 115: rather than using the current service principal, since this is expecting an AAD application can we spin one up and use that instead?

azurerm/internal/services/azurestackhci/registration.go
```diff
@@ -0,0 +1,29 @@
...
+
+// SupportedDataSources returns the supported Data Sources supported by this Service
+func (r Registration) SupportedDataSources() map[string]*schema.Resource {
+	return map[string]*schema.Resource{}
+}
+
+// SupportedResources returns the supported Resources supported by this Service
+func (r Registration) SupportedResources() map[string]*schema.Resource {
+	return map[string]*schema.Resource{
+		"azurerm_hci_cluster": resourceArmHCICluster(),
```
tombuildsstuff, line 27: this name isn't clear this is for Azure Stack HCI directly, should this become `azurerm_stack_hci_cluster`?

website/docs/r/hci_cluster.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+---
+
+# azurerm_hci_cluster
+
+Manages an Azure Stack HCI Cluster.
+
+## Example Usage
+
+```hcl
+data "azurerm_client_config" "current" {}
```
tombuildsstuff, line 16: this should likely be using the AAD Application Data Source to look up an existing application, rather than the service principal, which it's highly unlikely is being reused for AAD?

website/docs/r/hci_cluster.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+resource "azurerm_hci_cluster" "example" {
+  name                = "example-cluster"
+  resource_group_name = azurerm_resource_group.example.name
+  location            = azurerm_resource_group.example.location
+  client_id           = data.azurerm_client_config.current.client_id
+  tenant_id           = data.azurerm_client_config.current.tenant_id
+
+  tags = {
+    ENV = "Prod"
+  }
```
tombuildsstuff, line 32: tags aren't a required field, can we remove these from the example?

website/docs/r/hci_cluster.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+
+* `resource_group_name` - (Required) The name of the Resource Group where the HCI Cluster should exist. Changing this forces a new resource to be created.
+
+* `location` - (Required) The Azure Region where the HCI Cluster should exist. Changing this forces a new resource to be created.
+
+* `client_id` - (Required) The Client ID of the Azure Active Directory which is used by the HCI Cluster. Changing this forces a new resource to be created.
+
+* `tenant_id` - (Required) The Tenant ID of the Azure Active Directory which is used by the HCI Cluster. Changing this forces a new resource to be created.
+
+* `tags` - (Optional) A mapping of tags which should be assigned to the HCI Cluster.
```
tombuildsstuff, line 50: for all of these, can we use `Azure Stack HCI Cluster` rather than `HCI Cluster`

website/docs/r/hci_cluster.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+* `id` - The ID of the HCI Cluster.
+
+## Timeouts
+
+The `timeouts` block allows you to specify [timeouts](https://www.terraform.io/docs/configuration/resources.html#timeouts) for certain actions:
+
+* `create` - (Defaults to 30 minutes) Used when creating the HCI Cluster.
+* `read` - (Defaults to 5 minutes) Used when retrieving the HCI Cluster.
+* `update` - (Defaults to 30 minutes) Used when updating the HCI Cluster.
+* `delete` - (Defaults to 30 minutes) Used when deleting the HCI Cluster.
```
tombuildsstuff, line 65: (same here)

website/docs/r/hci_cluster.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+
+* `tenant_id` - (Required) The Tenant ID of the Azure Active Directory which is used by the HCI Cluster. Changing this forces a new resource to be created.
+
+* `tags` - (Optional) A mapping of tags which should be assigned to the HCI Cluster.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The ID of the HCI Cluster.
```
tombuildsstuff, line 56: (same here)

tombuildsstuff (commented):
hey @neil-yechenwei @magodo 

I've taken a look through and left some comments inline - can we fix these up so that this resource makes matches the way users are likely to use it?

Thanks!

azurerm/internal/services/azurestackhci/hci_cluster_resource.go
neil-yechenwei: updated

azurerm/internal/services/azurestackhci/registration.go
neil-yechenwei: updated

website/docs/r/hci_cluster.html.markdown
neil-yechenwei: updated

website/docs/r/hci_cluster.html.markdown
neil-yechenwei: updated

website/docs/r/hci_cluster.html.markdown
neil-yechenwei: updated

website/docs/r/hci_cluster.html.markdown
neil-yechenwei: Updated to `Azure Stack HCI Cluster`

azurerm/internal/services/azurestackhci/hci_cluster_resource.go
tombuildsstuff: Optional (and Computed*) and ForceNew, actually - since this could be defaulted and overwritten

neil-yechenwei: updated

azurerm/internal/services/azurestackhci/hci_cluster_resource_test.go
neil-yechenwei: Sorry.  For the other comments you mentioned, I've submitted PR. But for this comment, seems I have no permission to create an AAD application via "azuread_application". So I haven't updated test code against this case.

website/docs/r/hci_cluster.html.markdown
neil-yechenwei: Sorry.  For the other comments you mentioned, I've submitted PR. But for this comment, seems I have no permission to create an AAD application via "azuread_application". So I haven't updated test code against this case.

azurerm/internal/services/azurestackhci/hci_cluster_resource_test.go
tombuildsstuff: you'll need to grant permissions to use [the Azure AD API](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/guides/service_principal_configuration) - unfortunately this has to provision an AD Application as a part of the test, reusing the SP would cause issues

website/docs/r/hci_cluster.html.markdown
neil-yechenwei: OK. Updated test code.

azurerm/internal/services/azurestackhci/hci_cluster_resource_test.go
neil-yechenwei: OK. Updated test code.

PR #9158, opened 2020-11-05
azurerm_point_to_site_vpn_gateway - support `route` block

njuCZ:
image

### DIFF
--- a/azurerm/internal/services/network/point_to_site_vpn_gateway_resource.go
+++ b/azurerm/internal/services/network/point_to_site_vpn_gateway_resource.go
@@ -231,9 +231,12 @@ func resourceArmPointToSiteVPNGatewayCreateUpdate(d *schema.ResourceData, meta i
 				ID: utils.String(virtualHubId),
 			},
 			VpnGatewayScaleUnit: utils.Int32(int32(scaleUnit)),
-			CustomDNSServers:    utils.ExpandStringSlice(d.Get("custom_dns_servers").([]interface{})),
 		},
 		Tags: tags.Expand(t),
+	}
+	customDNSServers := utils.ExpandStringSlice(d.Get("custom_dns_servers").([]interface{}))
+	if len(*customDNSServers) != 0 {
+		parameters.P2SVpnGatewayProperties.CustomDNSServers = customDNSServers
 	}
 
 	future, err := client.CreateOrUpdate(ctx, resourceGroup, name, parameters)
--- a/azurerm/internal/services/network/point_to_site_vpn_gateway_resource.go
+++ b/azurerm/internal/services/network/point_to_site_vpn_gateway_resource.go
@@ -93,6 +93,48 @@ func resourceArmPointToSiteVPNGateway() *schema.Resource {
 								},
 							},
 						},
+
+						"route_config": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Computed: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"associated_route_table_id": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validate2.VirtualHubRouteTableID,
+									},
+
+									"propagated_route_table": {
+										Type:     schema.TypeList,
+										Optional: true,
+										MaxItems: 1,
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/network/point_to_site_vpn_gateway_resource.go
```diff
@@ -93,6 +93,48 @@ func resourceArmPointToSiteVPNGateway() *schema.Resource {
 								},
 							},
 						},
+
+						"route_config": {
```
katbyte, line 97: can we drop config from here? 
```suggestion
						"route": {
```

azurerm/internal/services/network/point_to_site_vpn_gateway_resource.go
```diff
@@ -93,6 +93,48 @@ func resourceArmPointToSiteVPNGateway() *schema.Resource {
...
+														ValidateFunc: validate2.VirtualHubRouteTableID,
+													},
+												},
+
+												"labels": {
+													Type:     schema.TypeList,
+													Optional: true,
+													Elem: &schema.Schema{
+														Type: schema.TypeString,
+													},
```
katbyte, line 130: could we validate this?

azurerm/internal/services/network/point_to_site_vpn_gateway_resource.go
```diff
@@ -93,6 +93,48 @@ func resourceArmPointToSiteVPNGateway() *schema.Resource {
...
+													Type:     schema.TypeList,
+													Required: true,
+													Elem: &schema.Schema{
+														Type:         schema.TypeString,
+														ValidateFunc: validate2.VirtualHubRouteTableID,
+													},
+												},
+
+												"labels": {
+													Type:     schema.TypeList,
```
katbyte, line 126: shouldn't this be a set?

azurerm/internal/services/network/point_to_site_vpn_gateway_resource.go
```diff
@@ -103,6 +145,46 @@ func resourceArmPointToSiteVPNGateway() *schema.Resource {
 				ValidateFunc: validation.IntAtLeast(0),
 			},
 
+			"custom_dns_servers": {
```
katbyte, line 148: i'm not sure if custom is really necessary here as its implied?

website/docs/r/point_to_site_vpn_gateway.html.markdown
```diff
@@ -52,18 +54,50 @@ A `connection_configuration` block supports the following:
...
+
+* `propagated_route_table` - (Optional) A `propagated_route_table` block as defined below.
+
+---
+
+A `propagated_route_table` block supports the following:
+
+* `ids` - (Required) The list of Virtual Hub Route Table resource id which the routes will be propagated to.
+
+* `labels` - (Optional) The list of labels to logically group VWan Route Tables.
```
katbyte, line 79: Could we fix this up?

azurerm/internal/services/network/point_to_site_vpn_gateway_resource.go
```diff
@@ -103,6 +145,46 @@ func resourceArmPointToSiteVPNGateway() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.IsIPv4Address,
+				},
+			},
+
+			"connection_health": {
```
katbyte, line 157: i don't think any of this makes sense in terraform so could we remove them? they will constantly be changed

katbyte (changes requested):
Thanks @njuCZ - i've left some comments inline that need to be addressed

azurerm/internal/services/network/point_to_site_vpn_gateway_resource.go
njuCZ: I am sorry I could not try out the validation rule for this field, just add a basic validation: `ValidateFunc: validation.StringIsNotEmpty`

njuCZ:
@katbyte thanks for your review, I have updated this PR, could you have a look again?

katbyte (approved):
Thanks @njuCZ - LGTM 👍

PR #9160, opened 2020-11-05
New resource: azurerm_vpn_gateway_connection

magodo:
## Background

This resource helps users to create s2s connection under virtual wan. Essentially, it connects between a VPN Site (represents an on-premise network) and a vpn gateway in the virtual hub.

See the [official tutorial](https://docs.microsoft.com/en-us/azure/virtual-wan/virtual-wan-site-to-site-portal) for more info.

## Test Result

```bash
💢 make testacc TEST=./azurerm/internal/services/network/tests TESTARGS='-run TestAccAzureRMVpnGatewayConnection_'
TF_ACC=1 go test ./azurerm/internal/services/network/tests -v -run TestAccAzureRMVpnGatewayConnection_ -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMVpnGatewayConnection_basic
=== PAUSE TestAccAzureRMVpnGatewayConnection_basic
=== RUN   TestAccAzureRMVpnGatewayConnection_complete
=== PAUSE TestAccAzureRMVpnGatewayConnection_complete
=== RUN   TestAccAzureRMVpnGatewayConnection_update
=== PAUSE TestAccAzureRMVpnGatewayConnection_update
=== RUN   TestAccAzureRMVpnGatewayConnection_customRouteTable
=== PAUSE TestAccAzureRMVpnGatewayConnection_customRouteTable
=== RUN   TestAccAzureRMVpnGatewayConnection_requiresImport
=== PAUSE TestAccAzureRMVpnGatewayConnection_requiresImport
=== CONT  TestAccAzureRMVpnGatewayConnection_basic
=== CONT  TestAccAzureRMVpnGatewayConnection_customRouteTable
=== CONT  TestAccAzureRMVpnGatewayConnection_requiresImport
=== CONT  TestAccAzureRMVpnGatewayConnection_complete
=== CONT  TestAccAzureRMVpnGatewayConnection_update
--- PASS: TestAccAzureRMVpnGatewayConnection_requiresImport (4144.28s)
--- PASS: TestAccAzureRMVpnGatewayConnection_complete (4187.42s)
--- PASS: TestAccAzureRMVpnGatewayConnection_basic (4214.95s)
--- PASS: TestAccAzureRMVpnGatewayConnection_customRouteTable (4400.18s)
--- PASS: TestAccAzureRMVpnGatewayConnection_update (4593.23s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/tests       4593.302s
```

## Design Concerns

One concern is that for [`routing_configuration` property](https://github.com/terraform-providers/terraform-provider-azurerm/compare/master...magodo:azurerm_vpn_gateway_connection?expand=1#diff-9777c73cb5e15e7b95697a513b33c184522b80d08b2643c3a7033cf885c0e6b1R75), if the user doesn't specify the route table, service will implicitly create one. This kind of behavior seems a leak of management from Terraform. So one alternative might be to set that property as `Required` and always force users to specify the route tables.


azurerm/internal/services/network/tests/vpn_gateway_connection_resource_test.go
```diff
@@ -0,0 +1,396 @@
...
+`, template)
+}
+
+func testAccAzureRMVpnGatewayConnection_template(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
+resource "azurerm_resource_group" "test" {
+  name     = "acctest-rg-%[1]d"
```
katbyte, line 355: we always use acctestRG to ensure capitals work correctly and include the service
```suggestion
  name     = "acctestRG-vpn-%[1]d"
```

azurerm/internal/services/network/vpn_gateway_connection_resource.go
```diff
@@ -0,0 +1,643 @@
...
+			},
+
+			"internet_security_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
+			// Service will create a route table for the user if this is not specified.
+			"routing_configuration": {
```
katbyte, line 76: i think we can drop configuration
```suggestion
			"routing": {
```

azurerm/internal/services/network/vpn_gateway_connection_resource.go
```diff
@@ -0,0 +1,643 @@
...
+						},
+
+						"bgp_enabled": {
+							Type:     schema.TypeBool,
+							ForceNew: true,
+							Optional: true,
+							Default:  false,
+						},
+
+						"use_policy_based_traffic_selector": {
```
katbyte, line 156:
```suggestion
						"policy_based_traffic_selector_enabled": {
```

azurerm/internal/services/network/vpn_gateway_connection_resource.go
```diff
@@ -0,0 +1,643 @@
...
+							},
+						},
+
+						"ratelimit_enabled": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
+
+						"use_local_azure_ip_address": {
```
katbyte, line 273:
```suggestion
						"local_azure_ip_address_enabled": {
```

azurerm/internal/services/network/vpn_gateway_connection_resource.go
```diff
@@ -0,0 +1,643 @@
...
+										Type:         schema.TypeInt,
+										Required:     true,
+										ValidateFunc: validation.IntBetween(300, 172799),
+									},
+									"sa_data_size_kb": {
+										Type:         schema.TypeInt,
+										Required:     true,
+										ValidateFunc: validation.IntBetween(1024, 2147483647),
+									},
+									"ipsec_encryption_algorithm": {
```
katbyte, line 178: I think we can drop ipsec as its in the block name?
```suggestion
									"encryption_algorithm": {
```

azurerm/internal/services/network/vpn_gateway_connection_resource.go
```diff
@@ -0,0 +1,643 @@
...
+											string(network.IpsecEncryptionAES256),
+											string(network.IpsecEncryptionDES),
+											string(network.IpsecEncryptionDES3),
+											string(network.IpsecEncryptionGCMAES128),
+											string(network.IpsecEncryptionGCMAES192),
+											string(network.IpsecEncryptionGCMAES256),
+											string(network.IpsecEncryptionNone),
+										}, false),
+									},
+									"ipsec_integrity_algorithm": {
```
katbyte, line 193:
```suggestion
									"integrity_algorithm": {
```

website/docs/r/vpn_gateway_connection.html.markdown
```diff
@@ -0,0 +1,165 @@
...
+
+# azurerm_vpn_gateway_connection
+
+Manages a VPN Gateway Connection.
+
+## Example Usage
+
+```hcl
+resource "azurerm_resource_group" "example" {
+  name     = "example-resourcestest1234"
```
katbyte, line 17: could we remove test from here?
```suggestion
  name     = "example-resources1234"
```

website/docs/r/vpn_gateway_connection.html.markdown
```diff
@@ -0,0 +1,165 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) The name which should be used for this VPN Gateway Connection. Changing this forces a new VPN Gateway Connection to be created.
+
+* `remote_vpn_site_id` - (Required) The ID of the remote VPN Site, which will connect to the VPN Gateway. Changing this forces a new VPN Gateway Connection to be created.
+
+* `vpn_gateway_id` - (Required) The ID of the VPN Gateway that this VPN Gateway Connection belongs to. Changing this forces a new VPN Gateway Connection to be created.
+
+* `vpn_link_connection` - (Required) One or more `vpn_link_connection` blocks as defined below.
```
katbyte, line 84: given connection is in the resource name i think we can remove it from here?
```suggestion
* `vpn_link` - (Required) One or more `vpn_link` blocks as defined below.
```

katbyte (changes requested):
Thanks @magodo, overall this looks good and i've left some comments inline to address. Is the reason we need to preserve the connections on the vpn gateway that these can be set via the new client & the existing vpn connection client?

azurerm/internal/services/network/vpn_gateway_connection_resource.go
magodo: Instead, I think we shall rename the `ipsec_policy` block to `policy` since it actually define both ipsec and ike policy.

azurerm/internal/services/network/vpn_gateway_connection_resource.go
magodo: Instead, I think we shall rename the `ipsec_policy` block to `policy` since it actually define both ipsec and ike policy.

magodo:
@katbyte Thank you for the review, I have resolved the comments, please take another look!

Regarding the changes to vpn gateway, yes, it is for reserving the connections purpose. Whilst after resolving the conflicts after merging to master branch, it appears that currently the update of vpn gateway will only do "patch" on top of the GET on the existing resource, so it means the connection will not be cleaned. So the only guarantee we need to keep is that the read-then-write operation happens as atomic, so the lock is needed only in update.

magodo:
The CI failure is unrelated to this PR...

azurerm/internal/services/network/vpn_gateway_connection_resource.go
katbyte: I don't think that makes sense? isn't ike policy part of ipsec? from google

```
IKE is a key management protocol used with IPsec protocol to establish a secure communication channel. IKE provides additional feature, flexibility, and ease of configuration for IPsec standard. policy selections, along with any preshared key, must be reflected in the VPN.
```

azurerm/internal/services/network/vpn_gateway_connection_resource.go
katbyte: see above comment

azurerm/internal/services/network/vpn_gateway_connection_resource.go
magodo: Make sense! Will do the change suggested originally.

azurerm/internal/services/network/vpn_gateway_connection_resource.go
magodo: Make sense! Will do the change suggested originally.

azurerm/internal/services/network/vpn_site_resource.go
```diff
@@ -123,7 +123,7 @@ func resourceArmVpnSite() *schema.Resource {
 									"asn": {
 										Type:         schema.TypeInt,
 										Required:     true,
-										ValidateFunc: validation.IntAtLeast(1),
+										ValidateFunc: validation.IntBetween(1, 4294967295),
```
jackofallops, line 126: @magodo - This upper boundary will break the 32-bit builds, I suspect that this change will need to be reverted.  Is this field expecting asplain for the value, or can it take asdot+ also?

katbyte:
@magodo - the CI failure is related to this PR, you have a 32bit int overflow

katbyte (approved):
Thanks @magodo - LGTM 👍

Issue #6236, opened 2020-03-24
Feature request: Set cache control for Storage Account Blob 


rjkokko:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Support for cache control headers for Storage Account blobs.

### New or Affected Resource(s)

* azurerm_storage_blob

### References

* [How to set cacheControl using Azrure REST API](https://docs.microsoft.com/en-us/azure/cdn/cdn-manage-expiration-of-blob-content#azure-storage-services-rest-api)

rjkokko:
A workaround for this issue is to use a provisioner and azure-cli.
```hcl
resource "azurerm_storage_blob" "foo" {
  name                   = "foo"
  storage_account_name   = "mystorageaccount"
  storage_container_name = "mycontainer"
  type                   = "Block"
  source                 = "foo.json"
  provisioner "local-exec" {
    command = "az storage blob update --account-name ${self.storage_account_name} --container-name ${self.storage_container_name} --name ${self.name} --content-cache-control 'no-store'"
  }
}
```

tombuildsstuff:
Giovanni supports this ([via the SetProperties method](https://github.com/tombuildsstuff/giovanni/blob/master/storage/2018-11-09/blob/blobs/properties_set.go#L15-L35)) - so this should be possible to add 👍

NT-D:
Hi @tombuildsstuff , nice to meet you. Can I handle this issue if no one drive this now ?

jmostella:
Are there any plans to fix this issue from the Terraform side? Possibly someone else has decided to work on this?

SymbioticKilla:
Any Updates?

PR #9199, opened 2020-11-06
Add basic Blob content metadata and cache_control

jmostella:
cache_control
content_type
content_disposition
content_encoding
content_language

* go.mod upgraded
* go.sum upgraded
* blob.go Added suppport for new fields
* resource_arm_storage_blob.go Added suppport for new fields
* storage_blob.html.markdown updated docs


jmostella:
Closes #6236

PR #9207, opened 2020-11-09
Policy dependency update - update api-version to 2020-03-01-preview

ArcturusZhang:
Prerequisite of https://github.com/terraform-providers/terraform-provider-azurerm/issues/9197


azurerm/internal/services/policy/client/client.go
```diff
@@ -2,7 +2,7 @@ package client
 
 import (
 	"github.com/Azure/azure-sdk-for-go/services/preview/policyinsights/mgmt/2019-10-01-preview/policyinsights"
-	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-09-01/policy"
+	"github.com/Azure/azure-sdk-for-go/services/preview/resources/mgmt/2020-03-01-preview/policy"
```
tombuildsstuff, line 5: why are we moving to a Preview API, it's been 8 months since then, is there not a stable release available?

tombuildsstuff (changes requested):
hi @ArcturusZhang 

Thanks for this PR.

Taking a look through whilst this seems ok (albeit there appears to be some internal changes due to issues with the Swagger definition?) - we're upgrading to an 8-month old Preview API here but we're not using it in any way? Given this is now 8 months old, is there not a stable API version available? However since we're not using any of the new Preview functionality, I'm wondering if we need to do this?

Thanks!

azurerm/internal/services/policy/client/client.go
ArcturusZhang: This is quite awkward that part in this package is stable and other part in this package is preview.
Policy assignment, definition and set-definition in this package are using stable api-version 2020-03-01, and policy exemption is using api-version 2020-07-01-preview and this resource currently does not have stable api-versions yet.
The "displayed" api-version in a go SDK package is usually decided by the service team (how they name their readme tag), therefore this package is named as 2020-03-01-preview.
Also I have changed the description, this upgrade is prerequisite to support the policy exemption.

ArcturusZhang:
Hi @tombuildsstuff please have a look at the explanation in the comment. thanks.

tombuildsstuff:
hey @ArcturusZhang 

Since we only need to import this for the Preview package itself I'm going to close this issue for the moment - but we should be able to use the Preview API for this specific resource as a part of the Policy Exceptions PR.

Thanks!

PR #9215, opened 2020-11-09
New resource: `azurerm_api_management_policy`

### LINKED ISSUE #9214: Support for separate resource for API Management service All APIs Policy
### Community Note

### Description

If references to named values are used in the API Management service (henceforth APIM) policy for All APIs, the named values must exist first. However, for named values to be created, they need the APIM to be created first, hence creating a catch-22.

* azurerm_api_management_policy

```hcl
  location = "West US"
}

resource "azurerm_api_management" "example" {
  name                = "example-apim"
  publisher_name      = "pub1"
  publisher_email     = "pub1@email.com"

  sku_name = "Developer_1"
}

resource "azurerm_api_management_named_value" "example" {
  name                = "example-apimg"
  api_management_name = azurerm_api_management.example.name
  display_name        = "ExampleProperty"
  value               = "Example Value"
}

resource "azurerm_api_management_policy" "example" {
  apim_management_id = azurerm_api_management.example.id
  xml_content        = file("example.xml")
}
```

### References

N/A.

sirlatrom:
New resource: `azurerm_api_management_policy`

Closes #9214

Note: Includes a custom parser for API Management service in 42cee46e3c45c5595649a3d9af914a9e1d12f4c1, which is adapted from #8228's 9e279c092a59716f15caf4b6bb40dec4d76f7ab0.


sirlatrom:
(Rebased on master for selfish purposes - easier building on top of the merged #8228.)

knutwannheden:
Question: Wouldn't it make sense to also deprecate the existing `azurerm_api_management.policy` in favor of this new resource?

azurerm/internal/services/apimanagement/tests/api_management_policy_resource_test.go
```diff
@@ -0,0 +1,254 @@
...
+  location            = azurerm_resource_group.test.location
+  resource_group_name = azurerm_resource_group.test.name
+  publisher_name      = "pub1"
+  publisher_email     = "pub1@email.com"
+  sku_name            = "Developer_1"
+}
+
+resource "azurerm_api_management_policy" "test" {
+  api_management_id = azurerm_api_management.test.id
+  xml_link          = "https://gist.githubusercontent.com/riordanp/ca22f8113afae0eb38cc12d718fd048d/raw/d6ac89a2f35a6881a7729f8cb4883179dc88eea1/example.xml"
```
manicminer, line 195: I've committed this file to master, can we download it from the repo to avoid relying on an external gist?

```suggestion
  xml_link          = "https://raw.githubusercontent.com/terraform-providers/terraform-provider-azurerm/master/azurerm/internal/services/apimanagement/tests/testdata/api_management_policy_test.xml"
```

azurerm/internal/services/apimanagement/tests/api_management_policy_resource_test.go
```diff
@@ -0,0 +1,254 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMApiManagementPolicyExists(data.ResourceName),
+				),
+			},
+			{
+				ResourceName:            data.ResourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"xml_link"},
+			},
```
manicminer, line 35: We can use the helper func for these

```suggestion
			data.ImportStep("xml_link"),
```

azurerm/internal/services/apimanagement/tests/api_management_policy_resource_test.go
```diff
@@ -0,0 +1,254 @@
...
```
manicminer, line 84: Same here

```suggestion
			data.ImportStep("xml_link"),
```

azurerm/internal/services/apimanagement/tests/api_management_policy_resource_test.go
```diff
@@ -0,0 +1,254 @@
...
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMApiManagementPolicyDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMApiManagementPolicy_basic(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMApiManagementPolicyExists(data.ResourceName),
+				),
+			},
+			{
```
manicminer, line 73: Can we add an import step here to aid debugging in case the azurerm_api_management resource breaks?

```suggestion
			data.ImportStep("xml_link"),
			{
```

azurerm/internal/services/apimanagement/tests/api_management_policy_resource_test.go
```diff
@@ -0,0 +1,254 @@
...
+				ResourceName:            data.ResourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"xml_link"},
+			},
```
manicminer, line 108: and here

```suggestion
			data.ImportStep("xml_link"),
```

azurerm/internal/services/apimanagement/api_management_policy_resource.go
```diff
@@ -0,0 +1,202 @@
...
+				Computed:         true,
+				ConflictsWith:    []string{"xml_link"},
+				DiffSuppressFunc: XmlWithDotNetInterpolationsDiffSuppress,
+			},
+
+			"xml_link": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"xml_content"},
+			},
```
manicminer, line 56: Should these have `ExactlyOneOf` (or `AtLeastOneOf`) set?

manicminer (changes requested):
@sirlatrom Thanks for this PR. This is looking great, just a few minor comments and then, pending tests, this looks good to merge.

azurerm/internal/services/apimanagement/api_management_policy_resource.go
sirlatrom: I wasn't aware of that option, but yes, that would be the right thing to do.

sirlatrom: Done in 69bf29a7e9f7e6f06b3a8a21132fbd68caee9c42.

manicminer:
@sirlatrom Testing this surfaced an issue with import detection for this new resource.

New API Management resources get the following default policy XML, which is just an empty structure but is nonetheless returned by `GET /subscriptions/1a6092a6-137e-4025-9a7c-ef77f76f2c02/resourceGroups/acctestRG-201123211614252791/providers/Microsoft.ApiManagement/service/acctestAM-201123211614252791/policies/policy`

```xml
<!--
    IMPORTANT:
    - Policy elements can appear only within the <inbound>, <outbound>, <backend> section elements.
    - Only the <forward-request> policy element can appear within the <backend> section element.
    - To apply a policy to the incoming request (before it is forwarded to the backend service), place a corresponding policy element within the <inbound> section element.
    - To apply a policy to the outgoing response (before it is sent back to the caller), place a corresponding policy element within the <outbound> section element.
    - To add a policy position the cursor at the desired insertion point and click on the round button associated with the policy.
    - To remove a policy, delete the corresponding policy statement from the policy document.
    - Policies are applied in the order of their appearance, from the top down.
-->
<policies>
    <inbound />
    <backend>
        <forward-request />
    </backend>
    <outbound />
    <on-error />
</policies>
```

sirlatrom:
> 
> 
> @sirlatrom Testing this surfaced an issue with import detection for this new resource.
> 
> New API Management resources get the following default policy XML, which is just an empty structure but is nonetheless returned by `GET /subscriptions/1a6092a6-137e-4025-9a7c-ef77f76f2c02/resourceGroups/acctestRG-201123211614252791/providers/Microsoft.ApiManagement/service/acctestAM-201123211614252791/policies/policy`
> 
> ```
> 
> <policies>
>     <inbound />
>     <backend>
>         <forward-request />
>     </backend>
>     <outbound />
>     <on-error />
> </policies>
> ```

That's an interesting challenge. Given that a default policy is installed and returned, how does it affect the implementation and/or acceptance tests, and how do you suggest they are changed to accommodate that?

manicminer:
Given that the policy must be a valid XML document and cannot be completely empty, I think the only way is to omit the import check here. I doubt we want to be delving into the XML schema so from the provider's perspective, it probably wants to ignore any existing policy. The documentation should also reflect this and include a warning note.

sirlatrom:
> Given that the policy must be a valid XML document and cannot be completely empty, I think the only way is to omit the import check here. I doubt we want to be delving into the XML schema so from the provider's perspective, it probably wants to ignore any existing policy. The documentation should also reflect this and include a warning note.

Alright, so to summarize:
* I'll remove the import steps in the test cases
* I'll add a warning note at the argument level along the lines of "Beware that an API Management service resource always has a policy, even if none is specified. Hence, before creating this resource, and after destroying it, a default policy will be configured in place of the one configured with this resource."

manicminer:
> > Given that the policy must be a valid XML document and cannot be completely empty, I think the only way is to omit the import check here. I doubt we want to be delving into the XML schema so from the provider's perspective, it probably wants to ignore any existing policy. The documentation should also reflect this and include a warning note.
> 
> Alright, so to summarize:
> 
>     * I'll remove the import steps in the test cases
> 
>     * I'll add a warning note at the argument level along the lines of "Beware that an API Management service resource always has a policy, even if none is specified. Hence, before creating this resource, and after destroying it, a default policy will be configured in place of the one configured with this resource."

@sirlatrom Sounds good! I'd also add a note that this resource will overwrite any existing policy you might have set, without prompting.

sirlatrom:
> 
> 
> > > Given that the policy must be a valid XML document and cannot be completely empty, I think the only way is to omit the import check here. I doubt we want to be delving into the XML schema so from the provider's perspective, it probably wants to ignore any existing policy. The documentation should also reflect this and include a warning note.
> > 
> > 
> > Alright, so to summarize:
> > ```
> > * I'll remove the import steps in the test cases
> > 
> > * I'll add a warning note at the argument level along the lines of "Beware that an API Management service resource always has a policy, even if none is specified. Hence, before creating this resource, and after destroying it, a default policy will be configured in place of the one configured with this resource."
> > ```
> 
> @sirlatrom Sounds good! I'd also add a note that this resource will overwrite any existing policy you might have set, without prompting. And that you should not use it at the same time as the `policy` block in the `azurerm_api_management` resource.

I'll use this phrasing then:

> ~> **NOTE:** This resource will, upon creation, **overwrite any existing policy in the API Management service**, as there is no feasible way to test whether the policy has been modified from the deault. Similarly, when this resource is destroyed, the API Management service will revert to its default policy.

As for the import step in the test cases, shouldn't I be able to just add the `xml_content` attribute to the `ignore` argument? Or does that effectively make it redundant?

What happens when you execute the acceptance tests for these lines?
```go
	if d.IsNewResource() {
		existing, err := client.Get(ctx, resourceGroup, serviceName, apimanagement.PolicyExportFormatXML)
		if err != nil {
			if !utils.ResponseWasNotFound(existing.Response) {
				return fmt.Errorf("checking for presence of existing Policy (API Management Service %q / Resource Group %q): %s", serviceName, resourceGroup, err)
			}
		}


		if existing.ID != nil && *existing.ID != "" {
			return tf.ImportAsExistsError("azurerm_api_management_policy", *existing.ID)
		}
	}
```
https://github.com/sirlatrom/terraform-provider-azurerm/blob/a5ce42fa8b9dc88ea46c6a2a23690b1d2e350758/azurerm/internal/services/apimanagement/api_management_policy_resource.go#L74-L87

If there is always a policy, shouldn't that return with `tf.ImportAsExistsError`?

manicminer:
> I'll use this phrasing then:
> 
> > ~> **NOTE:** This resource will, upon creation, **overwrite any existing policy in the API Management service**, as there is no feasible way to test whether the policy has been modified from the deault. Similarly, when this resource is destroyed, the API Management service will revert to its default policy.

LGTM 👍 

> 
> As for the import step in the test cases, shouldn't I be able to just add the `xml_content` attribute to the `ignore` argument? Or does that effectively make it redundant?

We can retain the test import step(s), as they serve to verify the resource was created and read back correctly. It's the existing/import check in the CreateUpdate function that poses the problem.

> 
> What happens when you execute the acceptance tests for these lines?
> 
> ```go
> 	if d.IsNewResource() {
> 		existing, err := client.Get(ctx, resourceGroup, serviceName, apimanagement.PolicyExportFormatXML)
> 		if err != nil {
> 			if !utils.ResponseWasNotFound(existing.Response) {
> 				return fmt.Errorf("checking for presence of existing Policy (API Management Service %q / Resource Group %q): %s", serviceName, resourceGroup, err)
> 			}
> 		}
> 
> 
> 		if existing.ID != nil && *existing.ID != "" {
> 			return tf.ImportAsExistsError("azurerm_api_management_policy", *existing.ID)
> 		}
> 	}
> ```
> 
> https://github.com/sirlatrom/terraform-provider-azurerm/blob/a5ce42fa8b9dc88ea46c6a2a23690b1d2e350758/azurerm/internal/services/apimanagement/api_management_policy_resource.go#L74-L87
> 
> If there is always a policy, shouldn't that return with `tf.ImportAsExistsError`?

When creating, this always returns tf.ImportAsExistsError() since it picks up the default policy document - this is the check we'll need to drop (replacing with a comment explaining why - as we may in future throw a warning message here when we move to v2 of the TF Plugin SDK)

sirlatrom:
@manicminer Should be ready now

manicminer (approved):
LGTM  🚀 

Test results:
Screenshot 2020-12-01 at 12 39 53

PR #9261, opened 2020-11-11
"azurerm_spring_cloud_service" - exporting `outbound_public_ip_addresses`

### LINKED ISSUE #9346: spring_cloud_service
_This issue was originally opened by @waynejnicklin as hashicorp/terraform#26933. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._
    
<hr>
    
  # module.spring_cloud.azurerm_spring_cloud_service.spring_cloud must be replaced
-/+ resource "azurerm_spring_cloud_service" "spring_cloud" {

...
      - network { # forces replacement
          - cidr_ranges = [] -> null
        }

...
    }

The above is appearing in our plan files for spring cloud, which it wasn't doing previously. Nothing has changed in the Terraform code, and it looks like it's not a real change (just changing [] to null). We don't use networking.

The end result is that it will destroy our entire spring cloud service because of this.

Is this a known issue? I'm using Terraform 0.13.2.

Thanks,
Wayne

njuCZ:
this PR serves two purpose:
1. fix #9346 
2. exports "outbound_ip" block

the reason of issue 9346:
image
"outbound_ip" is a new field, and there is always a value for this field no matter spring cloud instance is in a virtual network.
so we could make "network" block computed to solve this issue.

image

### DIFF
--- a/azurerm/internal/services/appplatform/spring_cloud_service_resource.go
+++ b/azurerm/internal/services/appplatform/spring_cloud_service_resource.go
@@ -111,22 +111,6 @@ func resourceArmSpringCloudService() *schema.Resource {
 							Computed: true,
 							ForceNew: true,
 						},
-
-						"outbound_ip": {
-							Type:     schema.TypeList,
-							Computed: true,
-							Elem: &schema.Resource{
-								Schema: map[string]*schema.Schema{
-									"public_ips": {
-										Type:     schema.TypeList,
-										Computed: true,
-										Elem: &schema.Schema{
-											Type: schema.TypeString,
-										},
-									},
-								},
-							},
-						},
 					},
 				},
 			},
@@ -220,6 +204,22 @@ func resourceArmSpringCloudService() *schema.Resource {
 						"instrumentation_key": {
 							Type:     schema.TypeString,
 							Required: true,
+						},
+					},
+				},
+			},
+
+			"outbound_ip": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"public_ips": {
+							Type:     schema.TypeList,
+							Computed: true,
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
 						},
 					},
 				},
@@ -395,6 +395,11 @@ func resourceArmSpringCloudServiceRead(d *schema.ResourceData, meta interface{})
 		}
 		if err := d.Set("network", flattenArmSpringCloudNetwork(resp.Properties.NetworkProfile)); err != nil {
 			return fmt.Errorf("setting `network`: %+v", err)
+		}
+		if resp.Properties.NetworkProfile != nil {
+			if err := d.Set("outbound_ip", flattenArmSpringCloudNetworkOutboundIPs(resp.Properties.NetworkProfile.OutboundIPs)); err != nil {
+				return fmt.Errorf("setting `outbound_ip`: %+v", err)
+			}
 		}
 		if resp.Properties.ConfigServerProperties != nil && resp.Properties.ConfigServerProperties.ConfigServer != nil {
 			if props := resp.Properties.ConfigServerProperties.ConfigServer.GitProperty; props != nil {
@@ -821,7 +826,6 @@ func flattenArmSpringCloudNetwork(input *appplatform.NetworkProfile) []interface
 			"cidr_ranges":                            cidrRanges,
 			"app_network_resource_group":             appNetworkResourceGroup,
 			"service_runtime_network_resource_group": serviceRuntimeNetworkResourceGroup,
-			"outbound_ip":                            flattenArmSpringCloudNetworkOutboundIPs(input.OutboundIPs),
 		},
 	}
 }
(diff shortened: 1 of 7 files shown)


azurerm/internal/services/appplatform/spring_cloud_service_resource.go
```diff
@@ -69,6 +69,7 @@ func resourceArmSpringCloudService() *schema.Resource {
 			"network": {
 				Type:     schema.TypeList,
 				Optional: true,
+				Computed: true,
```
tombuildsstuff, line 72: why is this now Computed, this means it's not possible to remove it, which would be a breaking change?

azurerm/internal/services/appplatform/spring_cloud_service_resource.go
```diff
@@ -110,6 +111,22 @@ func resourceArmSpringCloudService() *schema.Resource {
...
+							Computed: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"public_ips": {
+										Type:     schema.TypeList,
+										Computed: true,
+										Elem: &schema.Schema{
+											Type: schema.TypeString,
+										},
+									},
```
tombuildsstuff, line 126: there's no benefit in this value being nested, so can we move this to a top level field?

tombuildsstuff (changes requested):
hi @njuCZ 

Thanks for this PR - I've taken a look through and left some comments inline but if we can clear those up then we should be good to go 👍 

Thanks!

azurerm/internal/services/appplatform/spring_cloud_service_resource.go
tombuildsstuff: digging into the specifics for this, the larger question from #9346 is why only the `cidr_ranges` element is being set given all of the fields are set into the state: https://github.com/terraform-providers/terraform-provider-azurerm/blob/23539e6b36c719931db16f554c20ec85134d8226/azurerm/internal/services/appplatform/spring_cloud_service_resource.go#L801-L807

as such, rather than making this Computed I believe this may want the block omitting if all of these fields are empty, as this appears to be a bug in the API? We should also raise a bug in the API, since this is an API bug from what I can tell, unless there's other fields returned in this block by default?

njuCZ: @tombuildsstuff 
first of all, this is a sample config which will cause the diff:
```
resource "azurerm_spring_cloud_service" "test" {
  name                = "sc-example"
}
```
Before the backend service supports "outbound_ip" block, in the response of read rest api, the `networkProfile` is empty, so in the flatten function, we check it's nil and return `[]interface{}{}`, everything is right.

Recently, "outbound_ip" block is returned in the rest api, the response is like the following:
```
{
  "properties": {
    ...
    "networkProfile": {
      "outboundIPs": {
        "publicIPs": [
          "20.49.108.33",
          "20.49.108.45"
        ]
      }
    }
  },
  ...
}
```
because the `networkProfile` is not nil, in the flatten func, we will end up return a struct like this:
```
[]interface{}{
		map[string]interface{}{
			"app_subnet_id":                          "",
			"service_runtime_subnet_id":              "",
			"cidr_ranges":                            [],
			"app_network_resource_group":             "",
			"service_runtime_network_resource_group": "",
		},
	}
```

It seems this is a bug in the terraform-plugin-sdk. If user does not specify the whole `network` block, it could regard the string "" as empty, but thinks [] is different

azurerm/internal/services/appplatform/spring_cloud_service_resource.go
njuCZ: I will do this after we made a conclusion of the above issue

njuCZ: have moved to a top level field

azurerm/internal/services/appplatform/spring_cloud_service_resource.go
njuCZ: there is an alternative way to fix this issue: before the return statement in the flatten function, we check the length of all fields. if all the length is 0, we directly return `[]interface{}{}`. In this way, we could avoid adding "computed: true"

njuCZ:
@tombuildsstuff I have moved the field to a top level field. Could you have a look again and check which way is better to solve the diff issue ?

azurerm/internal/services/appplatform/spring_cloud_service_resource.go
katbyte: that is what we should do @njuCZ - detect the empty value and return nothing

njuCZ: done

njuCZ:
@katbyte Thanks for your suggestions. I have updated this PR and all acctest could pass. Please have a look again when free.

tombuildsstuff:
@njuCZ hope you don't mind but I've pushed a commit to rebase this and move this field to the top level as discussed above

jackofallops (approved):
Thanks @njuCZ & @tombuildsstuff - This LGTM now 👍

tombuildsstuff:
Acceptance Tests look good 👍

PR #9262, opened 2020-11-11
Update `azurerm_linux|windows_virtual_machine_scale_set` - Support `platform_fault_domain_count`, `disk_iops_read_write` and `disk_mbps_read_write`

ArcturusZhang:


website/docs/r/windows_virtual_machine_scale_set.html.markdown
```diff
@@ -264,6 +266,10 @@ A `data_disk` block supports the following:
 
 ~> **NOTE:** Disk Encryption Sets are in Public Preview in a limited set of regions
 
+* `disk_iops_read_write` - (Optional) Specifies the Read-Write IOPS for this Data Disk. Only settable for UltraSSD disks.
```
katbyte, line 269: Would this make more sense as
```suggestion
* `ultra_ssd_disk_iops_read_write` - (Optional) Specifies the Read-Write IOPS for this Data Disk. Only settable for UltraSSD disks.
```

to make it clear it only applies to ultra ssds?

azurerm/internal/services/compute/virtual_machine_scale_set.go
```diff
@@ -807,6 +819,14 @@ func ExpandVirtualMachineScaleSetDataDisk(input []interface{}) *[]compute.Virtua
 			}
 		}
 
+		if iops := raw["disk_iops_read_write"].(int); iops != 0 {
```
katbyte, line 822: should we be validating this isn't set when the disk type is set to something other then an ultra SSD?

azurerm/internal/services/compute/virtual_machine_scale_set.go
```diff
@@ -779,6 +779,18 @@ func VirtualMachineScaleSetDataDiskSchema() *schema.Schema {
...
+				"disk_iops_read_write": {
+					Type:     schema.TypeInt,
+					Optional: true,
+					Computed: true,
+				},
+
+				"disk_mbps_read_write": {
+					Type:     schema.TypeInt,
+					Optional: true,
+					Computed: true,
```
katbyte, line 792: what are these set to when the disk is not an ultra ssd?

katbyte (changes requested):
Thanks @ArcturusZhang - have a could questions i've left inline

azurerm/internal/services/compute/virtual_machine_scale_set.go
ArcturusZhang: In this case, the service does not return a value for this. In the terraform state, this will be set to 0

website/docs/r/windows_virtual_machine_scale_set.html.markdown
ArcturusZhang: Well, I would love to do this change, but these two attributes already exist in the `azurerm_managed_disk` resource, would this cause any incosistency?

azurerm/internal/services/compute/virtual_machine_scale_set.go
ArcturusZhang: Will do

website/docs/r/windows_virtual_machine_scale_set.html.markdown
jackofallops: We should leave these consistent with elsewhere in the provider, even though they only have meaning in one circumstance.

ArcturusZhang:
Hi @katbyte I have resolved the comments, please have a look, thanks

azurerm/internal/services/compute/virtual_machine_scale_set.go
```diff
@@ -779,12 +779,24 @@ func VirtualMachineScaleSetDataDiskSchema() *schema.Schema {
 					Optional: true,
 					Default:  false,
 				},
+
+				"disk_iops_read_write": {
```
katbyte, line 783:
```suggestion

				// TODO 3.0 - change this to ultra_ssd_disk_iops_read_write
				"disk_iops_read_write": {
```

azurerm/internal/services/compute/virtual_machine_scale_set.go
```diff
@@ -779,12 +779,24 @@ func VirtualMachineScaleSetDataDiskSchema() *schema.Schema {
...
+					Type:     schema.TypeInt,
+					Optional: true,
+					Computed: true,
+				},
+
+				"disk_mbps_read_write": {
+					Type:     schema.TypeInt,
```
katbyte, line 790:
```suggestion

				// TODO 3.0 - change this to ultra_ssd_disk_iops_read_write
				"disk_mbps_read_write": {
					Type:     schema.TypeInt,
```

katbyte (approved):
Thanks @ArcturusZhang - we should eventually change these properties to have ultra_ssd in them but that can be a 3.0 task - LGTM 👍

website/docs/r/windows_virtual_machine_scale_set.html.markdown
katbyte: Should eventually change both to `ultra_ssd` then - i'll just add a comment

PR #9264, opened 2020-11-11
Update `azurerm_linux|windows_virtual_machine` - make `dedicated_host_id` updatable

ArcturusZhang:


azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -741,6 +740,13 @@ func resourceLinuxVirtualMachineUpdate(d *schema.ResourceData, meta interface{})
 		update.Identity = identity
 	}
 
+	if d.HasChange("dedicated_host_id") {
+		shouldUpdate = true
```
tombuildsstuff, line 744: presumably this also needs to be shut-down and de-allocated - or does the API support live-migration?

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -786,6 +785,18 @@ func resourceWindowsVirtualMachineUpdate(d *schema.ResourceData, meta interface{
 		update.Identity = identity
 	}
 
+	if d.HasChange("dedicated_host_id") {
+		shouldUpdate = true
+		shouldDeallocate = true
```
tombuildsstuff, line 790: presumably this also needs to be shut-down too?

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -786,6 +785,18 @@ func resourceWindowsVirtualMachineUpdate(d *schema.ResourceData, meta interface{
...
+		if v, ok := d.GetOk("dedicated_host_id"); ok {
+			update.Host = &compute.SubResource{
+				ID: utils.String(v.(string)),
+			}
+		} else {
+			update.Host = &compute.SubResource{}
+		}
```
tombuildsstuff, line 797: can we add this to the Linux one too?

azurerm/internal/services/compute/tests/windows_virtual_machine_resource_scaling_test.go
```diff
@@ -72,6 +72,36 @@ func TestAccWindowsVirtualMachine_scalingDedicatedHost(t *testing.T) {
...
+			),
+			{
+				Config: testWindowsVirtualMachine_scalingDedicatedHostUpdate(data),
+				Check: resource.ComposeTestCheckFunc(
+					checkWindowsVirtualMachineExists(data.ResourceName),
+				),
+			},
+			data.ImportStep(
+				"admin_password",
+			),
```
tombuildsstuff, line 100: since we have support for removing the Dedicated Host ID, can we add a test for starting without a dedicated host, then using one and then not using a dedicated host?

azurerm/internal/services/compute/tests/linux_virtual_machine_resource_scaling_test.go
```diff
@@ -66,6 +66,31 @@ func TestAccLinuxVirtualMachine_scalingDedicatedHost(t *testing.T) {
...
+					checkLinuxVirtualMachineExists(data.ResourceName),
+				),
+			},
+			data.ImportStep(),
+			{
+				Config: testLinuxVirtualMachine_scalingDedicatedHostUpdate(data),
+				Check: resource.ComposeTestCheckFunc(
+					checkLinuxVirtualMachineExists(data.ResourceName),
+				),
+			},
```
tombuildsstuff, line 89: since we have support for removing the Dedicated Host ID, can we add a test for starting without a dedicated host, then using one and then not using a dedicated host?

we're also missing an import test at the end here?

tombuildsstuff (changes requested):
hi @ArcturusZhang 

Thanks for this PR.

I've taken a look through and left some comments inline but if we can fix those up then we can run the tests and see how this looks 👍 

Thanks!

ArcturusZhang:
Hi @tombuildsstuff thank you for review. 
I have updated the tests to make the dedicated host id could be updated from 
1. no dedicated host associated
1. associated to one
1. associated to another one
1. association removed.
But I have to write the config with a lot of redundant resources to avoid the terraform bug (I am not sure whether it is resolved).
When we have a state of a dedicated host and a VM associated with that, in the config, we removed dedicated host and the association, in the next apply, terraform will try to first delete the dedicated host and service will return an error complaining the dedicated host cannot be deleted since it is in-use in the VM. To avoid this issue, I have to remain the dedicated host unreferenced in the last step's config.

ArcturusZhang:
And @tombuildsstuff amazingly actually this update of dedicated host id does not requires a shutdown of the VM - at least my tests could work properly.

azurerm/internal/services/compute/tests/linux_virtual_machine_resource_scaling_test.go
ArcturusZhang: Test added and fixed

azurerm/internal/services/compute/tests/windows_virtual_machine_resource_scaling_test.go
ArcturusZhang: Test added

azurerm/internal/services/compute/windows_virtual_machine_resource.go
ArcturusZhang: Done

azurerm/internal/services/compute/linux_virtual_machine_resource.go
ArcturusZhang: Yeah, it needs to be deallocated. But shutdown might not be required, the API does not return any error on that and the update was successful

katbyte (approved):
Thanks @ArcturusZhang - this LGTM now 👍

PR #9285, opened 2020-11-12
Update dns_servers of azurerm_virtual_network from TypeList to TypeSet

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/9063


katbyte:
Hi @neil-yechenwei - thanks for this PR but i believe this is the wrong approach. DNS server order matters, or at least should matter, wrt to the order of name resolution so we cannot change this to a set. This seems like an API/service bug so could we open an issue there - as for a potential fix if there is a change to the DNS servers we could remove all of them and apply the new order?

As this is not a valid fix i am going to close this for the moment.

PR #9296, opened 2020-11-12
Use Rawxml format for global API policies string

### LINKED ISSUE #9223: Error when API Management policy C# expression contains double quotes
### Community Note

```
Terraform v0.13.5
+ provider registry.terraform.io/hashicorp/azurerm v1.43.0
```

### Affected Resource(s)

* `azurerm_api_management`

```hcl
terraform {
  backend "azurerm" {}
}

provider "azurerm" {
  version         = "~> 1.43.0"
}

resource "azurerm_api_management" "example" {
    location                  = "westeurope"
    name                      = "iis-sandbox-apimgmt"
    policy                    {
            xml_content = file("sandbox-policy.cshtml")
        }
    # ....
}
```
The referenced `sandbox-policy.cshtml` looks something like this:
```xml
<policies>
        <inbound>
                <set-variable name="token" value="@(context.Request.Headers.GetValueOrDefault("field"))" />
        </inbound>
        <backend />
        <outbound />
        <on-error />
</policies>
```
### Debug Output

https://gist.github.com/knutwannheden/380ed0fc78f0fc023cd831b2c9c789d6

### Expected Behavior

It should be possible to use the same pseudo-XML content with Terraform as with the online Azure Portal editor.

### Actual Behavior

Currently `terraform apply` appears to choke whenever a policy has an XML attribute value quoted using `"` which itself also contains double quotes, but without encoding them using XML entities. So this does not work:

```xml
<set-variable name="token" value="@(context.Request.Headers.GetValueOrDefault("field"))" />
```

It reports an error like the following:
```
Error: Error setting Policies for API Management Service "iis-sandbox-apimgmt" (Resource Group "iis-apimgmt-sandbox-rg"): apimanagement.PolicyClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="ValidationError" Message="One or more fields contain incorrect values:" Details=[{"code":"ValidationError","message":"'field' is an unexpected token. Expecting white space. Line 3, position 96.","target":"representation"}]
```

### Steps to Reproduce

1. Create files as given
1. `terraform apply`

### Important Factoids
According to issue #3918 and the associated PR #4140 (see e.g. https://github.com/terraform-providers/terraform-provider-azurerm/pull/4140#issuecomment-552490298) I got the impression that the Terraform provider should be able to handle "XML" where the attribute values and content where XML characters like `<`, `>`, and `"` are not encoded using XML entities, just like also the editor in Azure Portal accepts it.

To be honest I don't really understands how the Azure Portal does its parsing (well it isn't legal XML anyway), but it seems to work somehow and for the user experience I think it would be great it the same policy content could be provided via Terraform (i.e. without having to use XML entities).

As a workaround I was able to use single quotes to enclose the attribute value. But note that the Azure Portal online editor automatically converts them back to double quotes, so if that is copied back to the file used by Terraform it will break again.

### References

* #3918

knutwannheden:
Fixes #9223

### DIFF
--- a/azurerm/internal/services/apimanagement/api_management_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_resource.go
@@ -20,7 +20,6 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/apimanagement/parse"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
-	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tf/suppress"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
@@ -357,7 +356,7 @@ func resourceArmApiManagementService() *schema.Resource {
 							Optional:         true,
 							Computed:         true,
 							ConflictsWith:    []string{"policy.0.xml_link"},
-							DiffSuppressFunc: suppress.XmlDiff,
+							DiffSuppressFunc: XmlWithDotNetInterpolationsDiffSuppress,
 						},
 
 						"xml_link": {
@@ -1343,7 +1342,7 @@ func expandApiManagementPolicies(input []interface{}) (*apimanagement.PolicyCont
 	if xmlContent != "" {
 		return &apimanagement.PolicyContract{
 			PolicyContractProperties: &apimanagement.PolicyContractProperties{
-				Format: apimanagement.XML,
+				Format: apimanagement.Rawxml,
 				Value:  utils.String(xmlContent),
 			},
 		}, nil
--- a/azurerm/internal/services/apimanagement/tests/api_management_resource_test.go
+++ b/azurerm/internal/services/apimanagement/tests/api_management_resource_test.go
@@ -626,6 +626,7 @@ resource "azurerm_api_management" "test" {
     xml_content = <<XML
 <policies>
   <inbound>
+    <set-variable name="abc" value="@(context.Request.Headers.GetValueOrDefault("X-Header-Name", ""))" />
     <find-and-replace from="xyz" to="abc" />
   </inbound>
 </policies>


knutwannheden:
Hmm, I assume this is obsoleted by #9215. Feel free to close again.

tombuildsstuff (approved):
LGTM - thanks for this @knutwannheden 👍

azurerm/internal/services/apimanagement/api_management_resource.go
```diff
@@ -1343,7 +1343,7 @@ func expandApiManagementPolicies(input []interface{}) (*apimanagement.PolicyCont
 	if xmlContent != "" {
 		return &apimanagement.PolicyContract{
 			PolicyContractProperties: &apimanagement.PolicyContractProperties{
-				Format: apimanagement.XML,
+				Format: apimanagement.Rawxml,
```
tombuildsstuff, line 1346: Running the tests for this we get the following diff:

```
  <policies>
    <inbound>
      <set-variable name=\"abc\" value=\"@(context.Request.Headers.GetValueOrDefault(&quot;X-Header-Name&quot;, &quot;&quot;))\" />
      <find-and-replace from=\"xyz\" to=\"abc\" />
    </inbound>
  </policies>
  " => "
  <policies>
    <inbound>
      <set-variable name=\"abc\" value=\"@(context.Request.Headers.GetValueOrDefault(\"X-Header-Name\", \"\"))\" />
      <find-and-replace from=\"xyz\" to=\"abc\" />
    </inbound>
  </policies>
```

To fix that we'll also need to update this line: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/apimanagement/api_management_resource.go#L362

to ignore the differences when comparing the diff - since these get returned encoded from the API, we can do this using this function: https://github.com/terraform-providers/terraform-provider-azurerm/blob/841a9c9f7da454581b6fa761754909e90d3f8b7e/azurerm/internal/services/apimanagement/api_management_product_policy_resource.go#L47

As such could we update this so that this test passes?

knutwannheden: Thanks for the review and the guidance. I will update the code accordingly.

knutwannheden: How come the acceptance test didn't fail in GitHub? Are only unit tests performed there?

knutwannheden: I was unfortunately not able to get the acceptance test to run in my VS Code devcontainer. I think I would have to set a whole bunch of `ARM_*` environment variables somehow.

tombuildsstuff: Yeah Github Actions only runs the unit tests - the acceptance tests take longer (and can cost a lot) - which is why we run these manually using another tool

tombuildsstuff:
Tests look good 👍

PR #9298, opened 2020-11-12
`azurerm_api_management_custom_domain` - Add exception for R001 lint rule

WodansSon:
Adding a nolint declaration for lint rule `R001: ResourceData.Set() key argument should be string literal` as this is causing the lint checks to fail for all open PR's.

Related: #8946


katbyte (approved):
Thanks @WodansSon - LGTM 👍

PR #9308, opened 2020-11-13
support multiple Radius Server for vpn_server_configuration_resource

haiyuazhang:
Spec:
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/Microsoft.Network/stable/2020-05-01/virtualWan.json#L5895

ACC Tests result:
image

### DIFF
--- a/azurerm/internal/services/network/tests/vpn_server_configuration_resource_test.go
+++ b/azurerm/internal/services/network/tests/vpn_server_configuration_resource_test.go
@@ -382,18 +382,18 @@ resource "azurerm_vpn_server_configuration" "test" {
   vpn_authentication_types = ["Radius"]
 
   radius_server {
- 
-  	server {
-			address = "10.104.1.1"
-			secret  = "vindicators-the-return-of-worldender"
-			score = 1
-  	}
 
-  	server {
-			address = "10.104.1.2"
-			secret  = "vindicators-the-return-of-worldender"
-			score = 30
-  	}
+    server {
+      address = "10.104.1.1"
+      secret  = "vindicators-the-return-of-worldender"
+      score   = 1
+    }
+
+    server {
+      address = "10.104.1.2"
+      secret  = "vindicators-the-return-of-worldender"
+      score   = 30
+    }
 
     server_root_certificate {
       name             = "DigiCert-Federated-ID-Root-CA"
--- a/azurerm/internal/services/network/tests/vpn_server_configuration_resource_test.go
+++ b/azurerm/internal/services/network/tests/vpn_server_configuration_resource_test.go
@@ -98,6 +98,25 @@ func TestAccAzureRMVPNServerConfiguration_radius(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMVPNServerConfiguration_multipleRadius(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_vpn_server_configuration", "test")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.PreCheck(t) },
+		Providers:    acceptance.SupportedProviders,
+		CheckDestroy: testCheckAzureRMVPNServerConfigurationDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAzureRMAzureRMVPNServerConfiguration_multipleRadius(data),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckAzureRMVPNServerConfigurationExists(data.ResourceName),
+				),
+			},
+			data.ImportStep(),
+		},
+	})
+}
+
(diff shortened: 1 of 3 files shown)


website/docs/r/vpn_server_configuration.html.markdown
```diff
@@ -152,16 +152,28 @@ A `ipsec_policy` block supports the following:
...
 
 ---
 
+A `server` block supports the following:
+
+* `address` - (Required) The Address of the Radius Server.
+
+* `secret` - (Required) The Secret used to communicate with the Radius Server.
+
+* `score` - (Required) The score of the Radius Server determines the priority of the server. Ranges from 1 to 30.
```
mybayern1974, line 173:
```suggestion
* `score` - (Required) The Score of the Radius Server determines the priority of the server. Ranges from 1 to 30.
```
Making it upper case to serve Case consistence as descriptions of `address` and `secret`.

azurerm/internal/services/network/vpn_server_configuration_resource.go
```diff
@@ -237,14 +237,43 @@ func resourceArmVPNServerConfiguration() *schema.Resource {
...
 
 						"secret": {
-							Type:      schema.TypeString,
-							Required:  true,
-							Sensitive: true,
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validation.StringIsNotEmpty,
+							Sensitive:    true,
+						},
```
jackofallops, line 250: Your update to the docs suggests these are being replace/deprecated, is this the case?

haiyuazhang: Azure service team said  (address, secret) can still be used to config single Radius Server. These two will still be in specification for backward compatibility. Along side, new "RadiusServers" field has been introduced into the Spec, which can
cover both single and multiple Radius Server configuration setting.  So The two fields are not officially deprecated but will, I assume, be deprecated in future.

azurerm/internal/services/network/vpn_server_configuration_resource.go
```diff
@@ -237,14 +237,43 @@ func resourceArmVPNServerConfiguration() *schema.Resource {
...
+
+						"server": {
```
katbyte, line 252: Instead of adding a sub block here, can we just remove the max items on radius server, add score, and then if score is set, us the new property, and if score is not set & there is only 1 radius server use the old property?

katbyte (changes requested):

azurerm/internal/services/network/vpn_server_configuration_resource.go
haiyuazhang: Agree, that's a better solution. I'll do the change.

haiyuazhang: I forget to mention that after removing the constrain of max items on radius server, we need to lift client_root_certificate
and server_root_certificate in the radius server block to the outer scope, since these two settings are shared by all radius servers. If so, it will introduce a break-change. Now we're in the situation either introduce a break change or compromise a little a bit as what I did in the current PR. So before proceed,  I'd like to know your opinion on it.

katbyte: oh dang, didn't notice those! hmm

What we could do is keep `radius_server` as is and deprecated it then add a new block `radius` which would have those two properties and have a `servers` list inside. Then we can make them conflict so the user has to pick one or the other and we just use the one that is configured?

haiyuazhang: I've done the change based on your suggestion.

haiyuazhang: @katbyte  Since the waiting-response label still there, just ping you again in case you didn't notice my above reply. thanks.

azurerm/internal/services/network/tests/vpn_server_configuration_resource_test.go
```diff
@@ -98,6 +98,25 @@ func TestAccAzureRMVPNServerConfiguration_radius(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMVPNServerConfiguration_multipleRadius(t *testing.T) {
```
katbyte, line 101: Could we add some more tests?

Going from 1 server with the old block -> 1 server with the new block, updating the new block from 1 -> 2 -> 1 -> none?

katbyte (changes requested):
Thanks, if we can add some more tests i think this'll be good to merge

azurerm/internal/services/network/tests/vpn_server_configuration_resource_test.go
haiyuazhang: @katbyte  tests has been updated.

katbyte (approved):
Thanks @haiyuazhang - LGTM 👍

PR #9314, opened 2020-11-13
storage: adding a shim layer to enable toggling between the Data Plane and Resource Manager API

tombuildsstuff:
This PR introduces a shim layer to the Storage resources - which will enable (in the future) users to toggle between the Data Plane API and the Resource Manager API for provisioning certain resources nested within a Storage Account.

Notably at this time, whilst this shim layer is implemented - only one of the Resource Manager wrappers is functionally equivalent (Containers) - as such @yupwei68 will need to liaise with the Storage Team to surface this missing functionality in the Resource Manager API's.

Due to this not being available, at this time I've removed the WIP Resource Manager shims in `35f85897a14fe71c65f8bd751920b6fd274541dd` - but once the Storage Team exposes this functionality we can look to re-add this in the future.

Whilst this PR doesn't fix it, this provides a path to solving #2977, once the Storage API's support all of the necessary functionality.


tombuildsstuff:
Tests look good 👍

aristosvo:
Thanks for your amazing work @tombuildsstuff !!

Is it possible to enable the functionality only for the parts which already have an API implemented?

tombuildsstuff:
@aristosvo we're reaching out to the service team about the missing functionality, bear with us - if there's a fix planned in the not-too-distant future we may as well make this consistent across all resources to ensure this behaviours consistent; if it's a while off we can look to do that, however :)

katbyte (approved):
LGTM 👍

Issue #9309, opened 2020-11-13
vmss resource not in state file when error occurs during deployment also leading to destroy errors 


obourdon:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

- Terraform 0.12.29
- AzureRM 2.36.0

### Affected Resource(s)

* `azurerm_linux_virtual_machine_scale_set`

### Terraform Configuration Files

```hcl
# No need for HCL code, failing test case included in AzureRM provider code
```

### Debug Output

If you run the following command in AzureRM provider code cloned from [this repository](https://github.com/squarescale/terraform-provider-azurerm)

```
TF_ACC=1 go test github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/compute/tests -v -run=TestAccAzureRMLinuxVirtualMachineScaleSet_imagesNoPlan  -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
```

You get the following:

```
=== RUN   TestAccAzureRMLinuxVirtualMachineScaleSet_imagesNoPlan
=== PAUSE TestAccAzureRMLinuxVirtualMachineScaleSet_imagesNoPlan
=== CONT  TestAccAzureRMLinuxVirtualMachineScaleSet_imagesNoPlan
--- FAIL: TestAccAzureRMLinuxVirtualMachineScaleSet_imagesNoPlan (96.27s)
    testing.go:684: Step 1 error: Resource specified by ResourceName couldn't be found: azurerm_linux_virtual_machine_scale_set.test
    testing.go:745: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.

[...]

PR #9316, opened 2020-11-13
Fix for issue #9309

obourdon:
Subject says it all

### DIFF
--- a/azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
@@ -505,10 +505,11 @@ func resourceArmLinuxVirtualMachineScaleSetCreate(d *schema.ResourceData, meta i
 	}
 
 	log.Printf("[DEBUG] Waiting for Linux Virtual Machine Scale Set %q (Resource Group %q) to be created..", name, resourceGroup)
+	var finalErr error
 	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
-		return fmt.Errorf("Error waiting for creation of Linux Virtual Machine Scale Set %q (Resource Group %q): %+v", name, resourceGroup, err)
+		finalErr = fmt.Errorf("Error waiting for creation of Linux Virtual Machine Scale Set %q (Resource Group %q): %+v", name, resourceGroup, err)
 	}
-	log.Printf("[DEBUG] Virtual Machine Scale Set %q (Resource Group %q) was created", name, resourceGroup)
+	log.Printf("[DEBUG] Virtual Machine Scale Set %q (Resource Group %q) was created (%v)", name, resourceGroup, finalErr)
 
 	log.Printf("[DEBUG] Retrieving Virtual Machine Scale Set %q (Resource Group %q)..", name, resourceGroup)
 	resp, err := client.Get(ctx, resourceGroup, name)
@@ -521,7 +522,11 @@ func resourceArmLinuxVirtualMachineScaleSetCreate(d *schema.ResourceData, meta i
 	}
 	d.SetId(*resp.ID)
 
-	return resourceArmLinuxVirtualMachineScaleSetRead(d, meta)
+	res := resourceArmLinuxVirtualMachineScaleSetRead(d, meta)
+	if res == nil {
+		return finalErr
+	}
+	return res
 }
 
 func resourceArmLinuxVirtualMachineScaleSetUpdate(d *schema.ResourceData, meta interface{}) error {
--- a/azurerm/internal/services/compute/tests/linux_virtual_machine_scale_set_images_resource_test.go
+++ b/azurerm/internal/services/compute/tests/linux_virtual_machine_scale_set_images_resource_test.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"fmt"
+	"regexp"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


tombuildsstuff:
hey @obourdon 

Thanks for this PR.

The Azure API uses a `provisioningState` to return the status of each API mutation (e.g. Create/Update/Delete) - when things go smoothly this should be "Successful" (or some variant on that depending on the API).

By contrast, this field being returned as "Failed" can mean several different things, the first being "this resource isn't usable" - but can also be used to indicate that a resource requires further attention, or may eventually become usable. To give a good example here, when provisioning a Virtual Machine a certain amount of time is allowed by the hypervisor for the operating system to boot; if a virtual machine doesn't return as "booted" within this fixed period (3 hours, which is possible in some scenarios) then a `provisioningState` of `Failed` is returned, even if it's still booting (and will boot).

There's plenty of other examples across other Azure API's, as such whilst we could try and infer the behaviour that most users would want - for example tracking and tainting the resource, or deleting it - in practice this varies on a per API basis, and the behaviour users desire would change on a per-operator basis.

Since the meaning of "Failed" varies on a per-API basis and within those API's the meaning can vary based on the payload - unfortunately it's impractical for Terraform to decide which way to proceed here (particularly for updates).

As such to ensure this behaviour is consistent across the Provider - we've opted to require that the operator decides how to proceed, which is a trade-off. This approach gives operators the flexibility to determine that these resources are healthy or not - meaning that Terraform is less likely to cause further  damage when the Azure API returns an error. Which is also the down-side, if this is a legitimate error then we can't determine that - so we intentionally don't taint these resources, which isn't ideal.

Unfortunately this is a case of picking the behaviour that makes sense for all users - and to be honest there isn't a good default here, which is why we raise the error and don't track these resources (since users know how to identify them, as they've defined the name/resource group they live in).

As the error behaviour is intentionally consistent across resources for the reasons outlined above - and we don't plan to change this to track every error condition of every resource (since that information isn't practical to obtain) - whilst I'd like to thank you for this contribution I'm going to close this PR for the moment.

Thanks!

obourdon:
@tombuildsstuff needless to say that I am pretty disappointed by this explanation and the quick closing of this PR.

Furthermore I can not understand what you mean by `we've opted to require that the operator decides how to proceed,` as terraform is an automation framework and there should be nothing for the operator to decide besides proper coding of the deployment files.

Moreover the error is well propagated, the PR code  does not seem to break the acceptance tests runs and furthermore, if you read the associated issue, without this fix, you endup in a situation where apply failed and destroy can not be applied which is, to my opinion, way against terraform way of functioning

Sorry for insisting

jeromedoucet:
@tombuildsstuff I agree with @obourdon. In this case, there is a resource that has been created by terraform (201 returned by the API), then something failed after, on "wait for resource state".

It is reasonable to considers that Terraform cannot know the real state of this resource in this case, and then update on this resource may be impossible.

But it isn't arguable to make the destroy impossible. The resources has been successfully created by Terraform, so it must be at least destroyable by terraform.

tombuildsstuff:
@obourdon as mentioned above unfortunately the Azure API returning "Failed" doesn't necessarily mean it's Failed (it could later recover, or it could be unusable from the start as in this case) - so whilst adding this to the state and tainting this would be useful in some scenarios, in others this could be the opposite of what other users expect (where this could be "Failed" but actually end up "Successful" several hours later, due to the behaviour of the Azure Hypervisor).

> Furthermore I can not understand what you mean by we've opted to require that the operator decides how to proceed, as terraform is an automation framework and there should be nothing for the operator to decide besides proper coding of the deployment files.

Unfortunately the Azure API doesn't have a definitive list of the failure modes for every resource, as such these must be worked out via trial and error - as such whilst we'd love to be able to conditionally determine this, this isn't practical.

Terraform is a provisioning tool rather than a deployment tool - whilst that's perhaps a subtle difference, Terraform doesn't (for example) handle rollbacks and requires the operator to decide how to proceed when there's an error. Due to the nature of the Azure API - where a result of "Failed" doesn't necessarily mean "Unusable" - we have to push this back to operators to decide how's best to proceed, since whilst auto-deleting a failed deployment could work great for some operators/environments - for others that could cause larger issues.

Given the Azure API is the limiting factor here - whilst we'd like to determine the correct behaviour for all use-cases here, unfortunately the Azure API doesn't expose this information for each resource - and so this isn't possible to do this consistently across all resources at this time. Whilst we appreciate the contribution - unfortunately at this time this behaviour is the "least bad" thing to do - pushing the question of "how do we recover" on to the operators, who can confirm that for their specific scenario/environment.

---

@jeromedoucet 

> But it isn't arguable to make the destroy impossible. The resources has been successfully created by Terraform, so it must be at least destroyable by terraform.

Unfortunately there's a number of resources in the Azure API which can be created but can't be deleted - for example where they're provisioned in an Azure-internal Subscription and then made available, or can't be deleted until ~12 hours later - or soft delete which exacerbates this issue. In practice whilst we'd like to be able to do more here, but we're limited by the Azure API - whilst always tracking the resource if it fails to provision sounds great, the Azure API doesn't expose enough context to do so in a consistent manner.

In the instance of the "201" being returned from the API as opposed the 200 which is defined in the API Definitions - unfortunately that's a breaking behavioural change in the Azure API - whilst we're working on something to make that less of an issue going forward - at this point in time this relies on the Azure API returning only status codes that they've defined, which is why the Azure SDK is raising this error.

Thanks!

PR #9324, opened 2020-11-13
Updates time zone description

asgr86:
Updates time zone description with possible values reference


katbyte (approved):
THanks @asgr86,LGTM 👍

PR #9325, opened 2020-11-13
updated with possible values for time zone

asgr86:
updated with possible values for time zone for backup policy


katbyte (approved):
Thanks @caius, LGTM :

PR #9333, opened 2020-11-16
Storage Account Minimum TLS defaults to 1.2, rather than 1.0

HarleyB123:
Supersedes #9004 - Which was accidentally deleted.

"I noticed that the default for the minimum TLS version for Storage Accounts is 1.0, which isn't inline with the TLS version defaults we see in function apps or app services.

Have updated the code and the docs so that Storage Account TLS is now 1.2 by default"


katbyte:
Hey @HarleyB123, 

Thank you for this PR, however this is a breaking change as it would update any storage accounts without a specified value to 1.2 which could potentially break deployments - thus something we would do for our next major release: 3.0. As such I'm going to mark this as 3.0 and close it to be revisited at such time.

PR #9345, opened 2020-11-16
This line can be removed now that the service is GA as of 14/02/20

danielfears:
Removed line about the service being in public preview as it went GA earlier in the year.

Source: https://azure.microsoft.com/en-gb/updates/private-link-now-available-in-ga/


tombuildsstuff (approved):
LGTM - thanks for this @caius

PR #9350, opened 2020-11-17
Refactor: Shim acc test for policy

ArcturusZhang:
I did not migrate the tests for policy set definition since it will cause conflicts with PR https://github.com/terraform-providers/terraform-provider-azurerm/pull/9259


tombuildsstuff (approved):
LGTM - thanks for this @ArcturusZhang

PR #9372, opened 2020-11-18
Added custom_domain_verification_id to azurerm_app_service

### LINKED ISSUE #7537: azurerm_app_service - way to get 'Custom Domain Verification ID'
_This issue was originally opened by @jonsamwell as hashicorp/terraform#25432. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._
    
<hr>
    
I need a way to get the `Custom Domain Verification ID` of an azure web app so that I can automate binding a custom host name.

image

I've looked through all the exported attributes when using `azurerm_app_service` but I am unable to find a way to get the verification id which I can use to add a TXT record to an Azure DNS zone then bind a custom host name without performing the verification step manually.

This is a big hole in our environment creation process we would like to automate.

Sorry if I missed anything!

Thanks!

AdamCoulterOz:
Added custom_domain_verification_id to azurerm_app_service data and resource types and added documentation and test updates. This fixes #7537 

This depends on azure-sdk-for-go v46.1.0 and the updated `azure-sdk-for-go/services/web/mgmt/2020-06-01/web` api, which I've had a go at updating too.


AdamCoulterOz:
have built and run locally, works perfectly for what I tested, can we regression test this somehow?

jackofallops:
Hi @AdamCoulterOz - I've tested and merged the `2020-06-01` API upgrade now so I'm going to fold the additional computed property from this PR into your #9378 and close this one.

PR #9374, opened 2020-11-18
kubernetes: running tests individually by default

tombuildsstuff:
The "combined" test is no longer necessary, since we now limit concurrency for this service package at the CI level (and users will do the same when running this locally - so this shouldn't be an issue).

This means we no longer need to set the env var when running Kubernetes tests to get these results split.


manicminer (approved):
LGTM

jackofallops (approved):

PR #9378, opened 2020-11-18
New Resource: azurerm_app_service_managed_certificate

### LINKED ISSUE #7537: azurerm_app_service - way to get 'Custom Domain Verification ID'

AdamCoulterOz:
forked from @tiwood, updated with the latest master, and fixed the password issue you were having previously... it just needed to be set as an empty string, and it worked. This addresses part of #4824

```go
	certificate := web.Certificate{
		CertificateProperties: &web.CertificateProperties{
			CanonicalName: utils.String(name),
			ServerFarmID:  utils.String(appServicePlanID),
			Password:      new(string),
		},
		Location: utils.String(location),
		Tags:     tags.Expand(t),
	}
```

I then needed to handle the 202 accepted "error" and wait 30 seconds for it to be created before trying to read it again. 

I've tested this locally with my own Azure subscription.

This still needs to have tests added and documentation written. Its also probably worth creating the data resource at the same time.

fixes #7537


AdamCoulterOz:
This is a problem though....

https://github.com/Azure/azure-sdk-for-go/issues/13665

jackofallops:
Thanks for getting this started @AdamCoulterOz - I've done some additional work on this over the weekend (I'll push it later today) and hit a possible 2nd issue with app service that I'm looking into (It may need us to bump the API to `2020-06-01`). I'll update here as soon as I have progress.

AdamCoulterOz:
@jackofallops:
> Thanks for getting this started @AdamCoulterOz - I've done some additional work on this over the weekend (I'll push it later today) and hit a possible 2nd issue with app service that I'm looking into (It may need us to bump the API to `2020-06-01`). I'll update here as soon as I have progress.

Thanks Steve... 
Yep... for this to work, my other PR (https://github.com/terraform-providers/terraform-provider-azurerm/pull/9372) which exposes the CustomDomainVerificationID needs to be included which also has a dependency on 2020-06-01, which I actually included already in that last week. I've noticed you've also now done it in a seperate PR too :-)

jackofallops:
@AdamCoulterOz - Apologies, I missed that when I looked for an SDK bump in existing PR's.  I think we should probably split it out of yours since we know there's at least one breaking change if that's OK, and include the additional property after to help keep things clean?  I'll comment there also for completeness.

AdamCoulterOz:
@jackofallops yep that's cool, I figured you guys would want it that way too, happy to wait, as long as it can be in the next or so 😄 a bit stuck on a project atm.

jackofallops:
> @jackofallops yep that's cool, I figured you guys would want it that way too, happy to wait, as long as it can be in the next or so 😄 a bit stuck on a project atm.

It's my main focus this week, so 🤞

AdamCoulterOz:
@jackofallops I've been working on this and I've realised there is no point creating this resource to be separate from `azurerm_app_service_custom_hostname_binding` and then creating a `azurerm_app_service_managed_certificate` when it can never be created separately from a `hostname_binding`, only to then create another `azurerm_app_service_custom_hostname_certificate_binding` to modify the existing `hostname_binding`.

We could do this much easier by adding a single attribute `use_managed_certificate` to the `azurerm_app_service_custom_hostname_binding` and create a managed_certificate inside that lifecycle. I can put together as an example, will be much less code to manage and much simpler to use.

jackofallops:
Thanks @AdamCoulterOz - I think, given the current state of the service, it would be better to keep them as separate resources.  I've got this all working locally, but hit something of a road-block on setting up something for the acceptance tests (something that will be a problem no matter the implementation).  I'm going to get some good manual test coverage done, and discuss with the team about taking this forward on that basis (there's precedent for this approach in App Service, so shouldn't be a problem).  I'll push after I revert my manual hacks to the acctests later today so another member of the team can review.

```
=== RUN   TestAccAzureRMAppServiceManagedCertificate_basicLinux
=== PAUSE TestAccAzureRMAppServiceManagedCertificate_basicLinux
=== CONT  TestAccAzureRMAppServiceManagedCertificate_basicLinux
--- PASS: TestAccAzureRMAppServiceManagedCertificate_basicLinux (274.42s)
PASS
```

AdamCoulterOz:
@jackofallops if you're going to stick with `azurerm_app_service_managed_certificate` then I would change the inputs to be just this single attribute:

- app_service_custom_hostname_binding_id

This is because in order to create a managed certificate you must already have an `azurerm_app_service_hostname_binding` and can only use the other derived attributes as the allowed values.

For example, the current other input attributes:

- name - (name of the app_service) must be the same as the binding_id app service name
- canonical_name - must be identical to the hostname in the binding_id
- location - must be the location of the underlying app service plan
- resource_group_name - must be the same as the underlying app service plan
- app_service_plan_id - must be the id of the underlying app service plan

So there is no point allowing the user to specify any of them.

jackofallops:
> @jackofallops if you're going to stick with `azurerm_app_service_managed_certificate` then I would change the inputs to be just this single attribute:
> 
> * app_service_custom_hostname_binding_id
> 
> This is because in order to create a managed certificate you must already have an `azurerm_app_service_hostname_binding` and can only use the other derived attributes as the allowed values.

Very good shout 👍 - I'll refactor now that I have working tests.

jackofallops:
CI not ready/able to run these tests yet, but passing locally:
```
➜  terraform-provider-azurerm git:(adamcoulteroz-app_service_cert) make testacc TESTARGS='-run TestAccAzureRMAppServiceManagedCertificate_' TEST=./azurerm/internal/services/web/tests/
TF_ACC=1 go test ./azurerm/internal/services/web/tests/ -v -run TestAccAzureRMAppServiceManagedCertificate_ -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMAppServiceManagedCertificate_basicLinux
=== PAUSE TestAccAzureRMAppServiceManagedCertificate_basicLinux
=== RUN   TestAccAzureRMAppServiceManagedCertificate_basicImport
=== PAUSE TestAccAzureRMAppServiceManagedCertificate_basicImport
=== RUN   TestAccAzureRMAppServiceManagedCertificate_basicWindows
=== PAUSE TestAccAzureRMAppServiceManagedCertificate_basicWindows
=== CONT  TestAccAzureRMAppServiceManagedCertificate_basicLinux
=== CONT  TestAccAzureRMAppServiceManagedCertificate_basicWindows
=== CONT  TestAccAzureRMAppServiceManagedCertificate_basicImport
--- PASS: TestAccAzureRMAppServiceManagedCertificate_basicWindows (239.06s)
--- PASS: TestAccAzureRMAppServiceManagedCertificate_basicLinux (246.19s)
--- PASS: TestAccAzureRMAppServiceManagedCertificate_basicImport (258.05s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/web/tests	259.985s
```

azurerm/internal/services/web/app_service_managed_certificate_resource.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	if props := resp.CertificateProperties; props != nil {
+		d.Set("canonical_name", props.CanonicalName)
+		d.Set("friendly_name", props.FriendlyName)
+		d.Set("subject_name", props.SubjectName)
+		d.Set("host_names", props.HostNames)
+		d.Set("issuer", props.Issuer)
+		d.Set("issue_date", props.IssueDate.Format(time.RFC3339))
+		d.Set("expiration_date", props.ExpirationDate.Format(time.RFC3339))
```
tombuildsstuff, line 219: there's two potential crashes here if `IssueDate` or `ExpirationDate` aren't returned for whatever reason, which is unlikely but we should still catch:

```suggestion
		expirationDate := ""
		if props.ExpirationDate != nil {
			expirationDate = props.ExpirationDate.Format(time.RFC3339)
		}
		d.Set("expiration_date", expirationDate)
```

azurerm/helpers/validate/domain_name.go
```diff
@@ -0,0 +1,23 @@
...
+// e.g. portal.azure.com
+func DomainName(i interface{}, k string) (warnings []string, errors []error) {
+	//
+	v, ok := i.(string)
+	if !ok {
+		errors = append(errors, fmt.Errorf("expected %q to be string", k))
+		return
+	}
+	// The following regexp is a good example of why pre-processing support would be nice in GoLang
+	if matched := regexp.MustCompile(`^(?:[_a-z0-9](?:[_a-z0-9-]{0,61}[a-z0-9]\.)|(?:[0-9]+/[0-9]{2})\.)+(?:[a-z](?:[a-z0-9-]{0,61}[a-z0-9])?)?$`).Match([]byte(v)); !matched || len(v) > 253 {
```
tombuildsstuff, line 18: could we add some tests for this? I have a feeling we may be able to simplify this to:

```
if _, err := uri.Parse(fmt.Sprintf("http://%s/", v)); err != nil {
  ...
}
```

which should handle this for us?

azurerm/internal/services/web/resourceids.go
```diff
@@ -0,0 +1,3 @@
+package web
+
+//go:generate go run ../../tools/generator-resource-id/main.go -path=./ -name=ManagedCertificate -id=/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resGroup1/providers/Microsoft.Web/certificates/customhost.contoso.com
```
tombuildsstuff, line 3: _minor_ with the other PR's I've intentionally avoided using `0`'s for this - since this could indicate a parsing bug (whilst we don't check it at this point, these are UUID's and this is the default value for those) - it may be worth adding one in here:

```suggestion
//go:generate go run ../../tools/generator-resource-id/main.go -path=./ -name=ManagedCertificate -id=/subscriptions/12345678-1234-9876-4563-123456789012/resourceGroups/resGroup1/providers/Microsoft.Web/certificates/customhost.contoso.com
```

website/docs/d/app_service.html.markdown
```diff
@@ -43,6 +43,8 @@ output "app_service_id" {
 
 * `client_affinity_enabled` - Does the App Service send session affinity cookies, which route client requests in the same session to the same instance?
 
+* `custom_domain_verification_id` - An identifier used by App Service to perform domain ownership verification via TXT record.
```
tombuildsstuff, line 46: is it worth being explicit here?

```suggestion
* `custom_domain_verification_id` - An identifier used by App Service to perform domain ownership verification via a DNS TXT record.
```

website/docs/d/function_app.html.markdown
```diff
@@ -40,6 +40,8 @@ The following attributes are exported:
 
 * `connection_string` - An `connection_string` block as defined below.
 
+* `custom_domain_verification_id` - An identifier used by App Service to perform domain ownership verification via TXT record.
```
tombuildsstuff, line 43: is it worth being explicit here?

```suggestion
* `custom_domain_verification_id` - An identifier used by App Service to perform domain ownership verification via a DNS TXT record.
```

website/docs/r/app_service.html.markdown
```diff
@@ -405,6 +405,8 @@ The following attributes are exported:
 
 * `id` - The ID of the App Service.
 
+* `custom_domain_verification_id` - An identifier used by App Service to perform domain ownership verification via TXT record.
```
tombuildsstuff, line 408: is it worth being explicit here?

```suggestion
* `custom_domain_verification_id` - An identifier used by App Service to perform domain ownership verification via a DNS TXT record.
```

website/docs/r/app_service_managed_certificate.html.markdown
```diff
@@ -0,0 +1,114 @@
...
+
+resource "azurerm_app_service_custom_hostname_binding" "example" {
+  hostname            = "mycustomhost.contoso.com"
+  app_service_name    = azurerm_app_service.example.name
+  resource_group_name = azurerm_resource_group.example.name
+}
+
+
+resource "azurerm_app_service_managed_certificate" "example" {
+  custom_hostname_binding_id = "TODO"
```
tombuildsstuff, line 60: TODO?

website/docs/r/app_service_managed_certificate.html.markdown
```diff
@@ -0,0 +1,114 @@
...
+
+# azurerm_app_service_managed_certificate
+
+Manages a App Service Managed Certificate.
+
+## Example Usage
+
+```hcl
+resource "azurerm_resource_group" "example" {
+  name     = "exampleRG"
```
tombuildsstuff, line 17: to match the other examples could we make this:

```suggestion
  name     = "example-resources"
```

website/docs/r/app_service_managed_certificate.html.markdown
```diff
@@ -0,0 +1,114 @@
...
+  location = "West Europe"
+}
+
+resource "azurerm_app_service_plan" "example" {
+  name                = "examplePlan"
```
tombuildsstuff, line 22: to match the other examples could we make this:

```suggestion
  name                = "example-plan"
```

website/docs/r/app_service_managed_certificate.html.markdown
```diff
@@ -0,0 +1,114 @@
...
+  reserved            = true
+
+  sku {
+    tier = "Basic"
+    size = "B1"
+  }
+}
+
+resource "azurerm_app_service" "example" {
+  name                = "exampleApp"
```
tombuildsstuff, line 35: to match the other examples could we make this:

```suggestion
  name                = "example-app"
```

website/docs/r/app_service_managed_certificate.html.markdown
```diff
@@ -0,0 +1,114 @@
...
+  location            = azurerm_resource_group.example.location
+  resource_group_name = azurerm_resource_group.example.name
+  app_service_plan_id = azurerm_app_service_plan.example.id
+}
+
+resource "azurerm_dns_txt_record" "test" {
+  name                = join(".", ["asuid", "%s"])
```
tombuildsstuff, line 42: I presume this wants to be:

```suggestion
  name                = "mycustomhost.contoso.com"
```

website/docs/r/app_service_managed_certificate.html.markdown
```diff
@@ -0,0 +1,114 @@
...
+  resource_group_name = data.azurerm_dns_zone.test.resource_group_name
+  ttl                 = 300
+
+  record {
+    value = azurerm_app_service.test.custom_domain_verification_id
+  }
+}
+
+resource "azurerm_app_service_custom_hostname_binding" "example" {
+  hostname            = "mycustomhost.contoso.com"
```
tombuildsstuff, line 53: this may as well become a reference to the one above to enforce the dependency, I think?

```suggestion
  hostname            = azurerm_dns_txt_record.test.name
```

website/docs/r/app_service_managed_certificate.html.markdown
```diff
@@ -0,0 +1,114 @@
...
+  app_service_name    = azurerm_app_service.example.name
+  resource_group_name = azurerm_resource_group.example.name
+}
+
+
```
tombuildsstuff, line 58: minor, but if we're passing through here anyway we could likely remove this one

```suggestion
```

tombuildsstuff (approved):
hey @AdamCoulterOz / @jackofallops 

Thanks for pushing those changes - I've taken a look through and whilst this is mostly looking good I've left a few comments inline, if we can fix those up then this should otherwise be good to merge 👍 

Thanks!

website/docs/r/app_service_managed_certificate.html.markdown
jackofallops: Would appear I missed it from the docs generator 🙈

website/docs/r/app_service_managed_certificate.html.markdown
jackofallops: Not quite - both records are required but the TXT is for auth, the CNAME + subdomain is required here, so perhaps: 
```suggestion
  hostname            = join(".", [azurerm_dns_cname_record.example.name, azurerm_dns_cname_record.example.zone_name])
```
to show the dependency on both?

azurerm/helpers/validate/domain_name.go
jackofallops: As discussed offline, that will unfortunately accept any string that meets allowed character requirements, rather than qualifying complete hostname.

website/docs/r/app_service_managed_certificate.html.markdown
jackofallops: This needs to be created as the hostname to be used, prefixed with `asuid.`

jackofallops:
Acceptance tests passing (failures unrelated) in addition to local tests above:
image

imod:
@AdamCoulterOz sorry if i ask a dump question about the this solution: does it handle the case where the cert has to be renewed? is it enough to just trigger terraform regularly? Is this documented?

AdamCoulterOz:
@imod as per this documentation from Microsoft the certificate is automatically renewed every 5 months (and remains valid for 6 months).

https://azure.microsoft.com/en-au/updates/secure-your-custom-domains-at-no-cost-with-app-service-managed-certificates-preview/

tombuildsstuff:
@AdamCoulterOz given App Service will auto-rotate this for you behind the scenes, do you think it's worth adding an info box to the resource page to explain that behaviour?

AdamCoulterOz:
@tombuildsstuff Sure, I'll put in a PR for it.

tombuildsstuff:
👍 awesome, thanks @AdamCoulterOz :)

PR #9382, opened 2020-11-18
New Data Source azurerm_virtual_wan

EtienneDeneuve:
That's my first try in Go, so please be kind and helpful :)


website/docs/d/virtual_wan.html.markdown
```diff
@@ -0,0 +1,97 @@
...
+
+- `disable_vpn_encryption` - (Optional) True, if vpn encryption is disabled.
+
+- `location` - The Azure Region where the Virtual Wan exists.
+
+- `office365_local_breakout_category` - (Optional) Possible value are :
+    - `All`
+    - `None`
+    - `Optimize`
+    - `OptimizeAndAllow`.
```
tombuildsstuff, line 83: (same here) could we make this:

- `office365_local_breakout_category` - The Office365 Local Breakout Category.

website/docs/d/virtual_wan.html.markdown
```diff
@@ -0,0 +1,97 @@
...
+
+- `sku` - Type of Virtual Wan (Basic or Standard).
+
+- `tags` - A mapping of tags assigned to the Virtual Wan.
+
+- `virtual_hubs` - (Optional) list of virtual hubs attached.
+
+- `vpn_sites` - (Optional) list of vpn sites attached.
```
tombuildsstuff, line 91: since this is now a list of ID's can we make this:

```suggestion
- `vpn_site_ids` - A list of VPN Site ID's attached to this Virtual WAN.
```

website/docs/d/virtual_wan.html.markdown
```diff
@@ -0,0 +1,97 @@
...
+    - `All`
+    - `None`
+    - `Optimize`
+    - `OptimizeAndAllow`.
+
+- `sku` - Type of Virtual Wan (Basic or Standard).
+
+- `tags` - A mapping of tags assigned to the Virtual Wan.
+
+- `virtual_hubs` - (Optional) list of virtual hubs attached.
```
tombuildsstuff, line 89: since this is now a list of ID's can we make this:

```suggestion
- `virtual_hub_ids` - A list of Virtual Hubs ID's attached to this Virtual WAN.
```

azurerm/internal/services/network/virtual_wan_data_source.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmVirtualWan() *schema.Resource {
```
tombuildsstuff, line 19: in order to confirm this Data Source works going forward (for example when we upgrade the Azure SDK/an API Version), we need to add an Acceptance Test to provision a Virtual WAN and confirm that this Data Source works as expected - as such could we add a test to check that? Here's an example of how we do that for `azurerm_subnet`: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/network/tests/subnet_data_source_test.go

tombuildsstuff (changes requested):
hey @EtienneDeneuve

Thanks for this PR :)

Taking a look through this is off to a good start - if we can fix up the comments and add an acceptance test to confirm this works going forward then we should be able to run those tests and get this merged 👍 

Thanks!

website/docs/d/virtual_wan.html.markdown
EtienneDeneuve: I’ll change that :)

azurerm/internal/services/network/virtual_wan_data_source.go
EtienneDeneuve: done :)

EtienneDeneuve (commented):
good :)

EtienneDeneuve (commented):
Good To Go

katbyte (approved):
Thanks @EtienneDeneuve - this LGTM 👍

PR #9396, opened 2020-11-19
Docs: Use proper resource names in import section

piontas:
Documentation for `virtual_desktop_application_group` and `virtual_desktop_host_pool` has a wrong import example - missing `azurerm_` prefix.


tombuildsstuff (approved):
LGTM - thanks for this @piontas

PR #9421, opened 2020-11-23
SDK: Strongly-Typed Data Sources & Resources

tombuildsstuff:
This PR introduces an SDK for strongly-typed Data Sources and Resources - allowing the use of native Go types for creating Terraform Resources.

---

A quick note to passers-by: this is more of a prototype than ready for production usage, but will likely form the foundation for Terraform Resources going forward.

---

There's several components to this Pull Request:

1. Introduction of "Typed" Data Sources and Resources
	* Introduces the "DataSource" and "Resource" interfaces which remove the ambiguous use of `interface{}` in favour of typed objects - meaning we can rely on higher-level types rather than casting in every resource.
	* A shim layer exists between the Terraform Plugin SDK and these resources, meaning we can lean on each as it's useful (for example, Tooling can build on top of the "DataSource" and "Resource" interfaces, rather than assuming the implementation)
	* This allows a significant amount of code reduction - introducing some Conventions to do so (for example, automatically calling Read after Create/Update, ensuring that Contexts are automatically configured wrapped with a deadline)

2. Introduction of Encode and Decode functions - to enable encoding/decoding values from the Terraform State from/to a native Go struct
	* Ultimately this allows for `hcl` tags to be defined twice (in the schema and the Go struct) - rather than 5/6 times at present (one time per schema, and once per CRUD method)
	* This will enable us (in a follow up PR/commit) to enable unit testing to catch type/serialization errors by confirming that the `hcl` values in the struct tags exist in the schema
	* This also allows us to lean more heavily on the compiler and tooling (such as refactoring tooling) rather than find+replacing magic-strings.

3. Fixing up the Provider Registration and associated Tooling to support both Typed (new) and Untyped (existing) resources

Notably the Typed Resources which are being introduced enable falling back to mutating `ResourceData` directly (for example to call `HasChanges` in an update function) - so this is a wrapper rather than an outright replacement.

For the purposes of completeness I've reimplemented the Resource Group resource using this new SDK (you can find it in the Readme, I've pulled it out of the history) and the tests for that resource pass 👍🏻

---

As mentioned above, at this point in time this is still a prototype, so whilst there's a bunch of tests there's almost certainly some bugs too - as such new functionality should still be built using the Terraform Plugin SDK directly at this time.

As such whilst we don't plan to accept external PR's using this new SDK at this time - we're hoping to enable doing so in the coming weeks - at the same time we'll look to add more documentation to solve #2688 too.


azurerm/internal/sdk/logger_console.go
```diff
@@ -0,0 +1,33 @@
...
+// Warn prints out a message prefixed with `[WARN]` formatted verbatim
+func (ConsoleLogger) Warn(message string) {
+	log.Print(message)
+}
+
+// Warnf prints out a message prefixed with `[WARN]` formatted
+// with the specified arguments
+func (ConsoleLogger) Warnf(format string, args ...interface{}) {
+	log.Print(fmt.Sprintf(format, args...))
+}
```
tombuildsstuff, line 33: actually these all need to be fixed

azurerm/internal/services/resource/parse/group.go
```diff
@@ -0,0 +1,45 @@
...
+	return ResourceGroupId{
+		SubscriptionId: subscriptionId,
+		Name:           name,
+	}
+}
+
+func ResourceGroupID(input string) (*ResourceGroupId, error) {
+	id, err := azure.ParseAzureResourceID(input)
+	if err != nil {
+		return nil, fmt.Errorf("[ERROR] Unable to parse Resource Group ID %q: %+v", input, err)
```
jackofallops, line 29:
```suggestion
		return nil, fmt.Errorf("unable to parse Resource Group ID %q: %+v", input, err)
```

azurerm/internal/sdk/resource_encode.go
```diff
@@ -0,0 +1,142 @@
...
+				iv := fieldVal.Int()
+				debugLogger.Infof("Setting %q to %d", tfschemaTag, iv)
+
+				output[tfschemaTag] = iv
+
+			case reflect.Float32, reflect.Float64:
+				fv := fieldVal.Float()
+				debugLogger.Infof("Setting %q to %f", tfschemaTag, fv)
+
+				output[tfschemaTag] = fv
```
jackofallops, line 62:
```suggestion
				debugLogger.Infof("Setting %q to %d", tfschemaTag, iv)
				output[tfschemaTag] = iv

			case reflect.Float32, reflect.Float64:
				fv := fieldVal.Float()
				debugLogger.Infof("Setting %q to %f", tfschemaTag, fv)
				output[tfschemaTag] = fv
```

jackofallops (approved):
Hi @tombuildsstuff / @mbfrahry - Couple really minor comments, but otherwise LGTM 👍

azurerm/internal/services/resource/parse/group.go
tombuildsstuff: this'll get replaced with a generated version shortly, so I'm going to suggest we wait until then for that

Issue #9428, opened 2020-11-23
Support for overriding the subscriptionID on individual resources/data sources


BrendanThompson:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

As the provider stands today it is not possible to override the Subscription in which the resource(s) are being provisioned or referenced. Terraform Core offers the functionality of utilising Provider Aliases, this is capable of solving a lot of issues around this, however, not all.

There are many instances within large organisations where the executing identity has permissions scoped to a management group and is required to perform actions across numerous subscriptions. In this scenario using the functionality provided currently by Terraform Core an alias **must** be provisioned for each subscription, and then that alias must be applied to an additional resource/data source declaration. This means that there will inevitably be a lot of code duplication.

What I am proposing is allowing the ability to pass in a `subscription_id` onto resources/data sources that override what is configured within the provider block. As long as the executing identity has the required permissions to perform the desired actions this will work perfectly. An example of this is within the Google provider it allows you to configure the `project` on the provider block and then override that within any calls to resources/data sources.

Allowing this `subscription_id` to be passed into resources/data sources means that consumers would be able to use `for_each` loops on a resource declaration and have those resources deployed into seperate subscriptions. 

One place I see this as incredibly valuable is within the network stack. Adhering to Microsoft's recommended Hub & Spoke approach, I have single TFE workspace that controls the spoke networks for a given environment, this environment spans approximately 20 subscriptions. Currently this means I need to re-declare the same things numerous times, the override solves this.

The above is just one example of where this is useful for large enterprise clients.

I have created an implementation of this how I think it could work, it has been done only to the data source for `azurerm_storage_account`. I would really appreciate some feedback on this as an idea and the implementation.

### New or Affected Resource(s)

* azurerm_*

### Potential Terraform Configuration

```hcl
provider "azurerm" {
  subscription_id = "00000000-0000-0000-0000-000000000000"

  features = {}
}

data "azurerm_storage_account" "this" {
  name                = "blt0"
  resource_group_name = "RG-MEL-BLT-0"
}

data "azurerm_storage_account" "that" {
  name                = "blt1"
[...]

PR #9429, opened 2020-11-23
feature: add subscription override for storage account ds

BrendanThompson:
This is the implementation to support #9428 in the hopes of adding the ability to override the `subscription_id` for individual resources and data sources.

### DIFF
--- a/azurerm/internal/services/storage/client/client.go
+++ b/azurerm/internal/services/storage/client/client.go
@@ -36,6 +36,7 @@ type Client struct {
 
 	resourceManagerAuthorizer autorest.Authorizer
 	storageAdAuth             *autorest.Authorizer
+	options *common.ClientOptions
 }
 
 func NewClient(options *common.ClientOptions) *Client {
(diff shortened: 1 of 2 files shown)


BrendanThompson:
@tombuildsstuff @katbyte @ArcturusZhang - would absolutely love some feedback here from you guys 😄

tombuildsstuff:
hi @BrendanThompson 

Thanks for this contribution - apologies for the delayed review here!

As mentioned in #9428, ultimately this PR looks to workaround a limitation of Terraform Core not supporting `for_each` for Providers by spinning up a Client for a given Subscription ID at usage time.

Whilst this could help in the short-term in some specific cases, unfortunately users tend to expect this functionality across all resources - meaning that if we implement this on a single resource we'd need to implement this across all of them. Since this is ultimately looking to workaround missing functionality in Terraform Core, which would likely become available in the future - whilst we appreciate the contribution we believe this is ultimately needs to be fixed in Terraform Core, rather than hacked around in each Provider.

As such, whilst I'd like to thank you again for this contribution, I'm going to close this issue for the moment - but I'd encourage you to [open a Feature Enhancement on the Terraform Core repository](https://github.com/hashicorp/terraform/issues) where somebody from that team should be able to take a look at supporting this functionality within Terraform Core.

Thanks!

PR #9433, opened 2020-11-24
Refactor: Migrate cdn tests

ArcturusZhang:
Note: since our import step will check the attributes, therefore I removed a lot of manual checking in this PR - except for those computed attributes

It might be hard to review the diffs in this PR, but you could only select the second and the fourth commit to show the actual diffs.


katbyte (approved):
Thanks! LGTM 👍

katbyte:
once the merge conflicts are fixed up we'll get thismerged

jackofallops:
Tests looking good:
image

PR #9434, opened 2020-11-24
refactor: migrate authorization test

ArcturusZhang:
I did not migrate the test cases for role assignment, since we currently do not have a function for non-paralleled tests


tombuildsstuff (approved):
LGTM 👍

tombuildsstuff:
After moving the Role Assignment up a level too (since the CI system assumes all tests are either in the `./tests` folder or in the root, not spread across both) - tests look good 👍

PR #9446, opened 2020-11-24
appinsights: switching to use generated ID Formatters/Parsers

tombuildsstuff:
Also added enhanced import validation for App Insights Components


mbfrahry (approved):
LGTM!

PR #9459, opened 2020-11-25
generator/resource-id: removing the type name from the generated name

tombuildsstuff:
This PR/the first commit changes the Resource ID generator slightly so that should the singular segment key match the type name, we'll switch it out for `Name` - meaning that (for example) `CertificateName` for a Batch Certificate becomes `Name` - which better matches the existing convention

The follow up commits then fix up these changes in the existing generated resources


jackofallops (approved):

PR #9466, opened 2020-11-25
customproviders: switching to use a generated Resource ID Formatter/Parser

tombuildsstuff:
Renamed the type to better match the Azure naming here too


manicminer (approved):
LGTM

PR #9467, opened 2020-11-25
databasemigration: switching to use a generated Resource ID Formatter/Parser

tombuildsstuff:
Ditched the `DatabaseMigration` prefix since we're within that package


manicminer (approved):
LGTM

PR #9469, opened 2020-11-25
datafactory: updating to use a generated Resource ID Formatter & Parser

tombuildsstuff:
This also fixed a bug in the generator for keys suffixed with `ies` now there's multiple use-cases - so I've re-generated the `SharedImage` parsers too


katbyte (approved):
LGTM 👍

PR #9474, opened 2020-11-25
devtestlabs: switching to use the Generated Resource ID Formatters/Parsers

tombuildsstuff:
Notably this PR uses the "Parse VM ID" from Compute since this was doing exactly that


katbyte (approved):
LGTM 👍

PR #9490, opened 2020-11-26
Raise limit of destination_ports of network_rule_collection for azurerm_firewall_policy_rule_collection_group

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/9304

--- PASS: TestAccAzureRMFirewallPolicyRuleCollectionGroup_requiresImport (153.90s)
--- PASS: TestAccAzureRMFirewallPolicyRuleCollectionGroup_basic (194.10s)
--- PASS: TestAccAzureRMFirewallPolicyRuleCollectionGroup_complete (202.27s)
--- PASS: TestAccAzureRMFirewallPolicyRuleCollectionGroup_update (269.51s)


azurerm/internal/services/firewall/validate/firewall_policy_network_rule_port.go
```diff
@@ -0,0 +1,53 @@
+package validate
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+)
+
+func FirewallPolicyNetworkRulePort(i interface{}, k string) (warnings []string, errors []error) {
```
katbyte, line 9: Instead of creating an entire new function could we add this to the validate package as `validate.PortOrPortRangeWithin(1,65535)` and then both properties can share it?

katbyte (changes requested):
Thanks for the PR @neil-yechenwei - I think we can simplify this and share the functino between the properties with a small change/ See my comment inline

azurerm/internal/services/firewall/validate/firewall_policy_network_rule_port.go
neil-yechenwei: Updated.

neil-yechenwei:
@katbyte , thanks for your comments. I've updated code. Please have an another look.

katbyte (approved):

PR #9498, opened 2020-11-26
iot: switching to use generated Resource ID Formatters/Parsers

tombuildsstuff:
Grouped these together since they're relatively small but related


jackofallops (approved):
LGTM 👍

PR #9527, opened 2020-11-27
network & privatedns: switching to use a Generated Resource ID Formatter/Parser

tombuildsstuff:
This PR ends up doing a little refactoring to be able to generate the Resource ID's for both Network and Private DNS, which was nested within the Network package.


katbyte (approved):
LGTM 👍

PR #9530, opened 2020-11-27
portal: switching to use the Generated Resource ID Formatters/Parsers

tombuildsstuff:
Also adds enhanced import validation to the `azurerm_dashboard` resource


katbyte (approved):
LGTM 👍

PR #9535, opened 2020-11-27
parsers: switching some instances over to use the parser directly

tombuildsstuff:
This removes the direct dependency on `azure.ParseResourceID` so that we can centralise this parsing logic


katbyte (approved):
LGTM 👍

Issue #9544, opened 2020-11-29
Unable to build master following instructions in README.md


sharebear:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

master branch

### Affected Resource(s)

N/A

### Terraform Configuration Files

N/A

### Debug Output

N/A

### Panic Output

N/A

### Expected Behaviour

Following the instructions from README.md for building should work, allowing new contributors a stable base to build on. Currently instructions are basically `make tools && make build`

### Actual Behaviour

`make build` fails.

Initially it fails with lots of error messages of the form

```
$ make build
==> Checking that code complies with gofmt requirements...
==> Checking that Custom Timeouts are used...
==> Checking that acceptance test packages are used...
go generate ./azurerm/internal/services/...
panic: writing contents to "/tmptemp-1606686499.go": open /tmptemp-1606686499.go: permission denied

goroutine 1 [running]:
main.main()
	/home/jshare/go/src/github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tools/generator-resource-id/main.go:29 +0x294
exit status 2
azurerm/internal/services/analysisservices/resourceids.go:3: running "go": exit status 1

# panic repeated 30-40 times
```

Note the lack of a `/` between `/tmp` and `temp-xxxx.go`. I've temporarily hacked this locally with the following change

```diff
 func (f GolangCodeFormatter) randomFilePath() string {
        time := time.Now().Unix()
-       return fmt.Sprintf("%stemp-%d.go", os.TempDir(), time)
+       return fmt.Sprintf("%s/temp-%d.go", os.TempDir(), time)
 }
```

Although a better solution would probably use `filepath.Join` or `ioutil.TempFile`.

Running `make build` with this change I then stumble on the following error

```
$ make build
==> Checking that code complies with gofmt requirements...
==> Checking that Custom Timeouts are used...
==> Checking that acceptance test packages are used...
go generate ./azurerm/internal/services/...
go generate ./azurerm/internal/provider/
[...]

PR #9545, opened 2020-11-29
generator/resource-id: refactor to use ioutil.TempFile instead of cus…

sharebear:
…tom file name generation

Fixes first complaint in #9544

I'm new to go, this was basically copy-pasted from the documentation with the necessary adjustments to make the code compile.


tombuildsstuff (approved):
LGTM - thanks for this @sharebear

PR #9546, opened 2020-11-29
generator/resource-id: Import strings module when necessary

sharebear:
Multiple generated files are missing the strings import, (see #9544). This might be a little ugly but it does the job.

If there's a more idiomatic way to do this in generate code let me know and I can rewrite the PR.

### DIFF
--- a/azurerm/internal/tools/generator-resource-id/main.go
+++ b/azurerm/internal/tools/generator-resource-id/main.go
@@ -215,13 +215,18 @@ type ResourceIdGenerator struct {
 }
 
 func (id ResourceIdGenerator) Code() string {
+	insensitive := id.codeForParserInsensitive()
+	stringsImport := ""
+	if strings.Contains(insensitive, "strings") {
+		stringsImport = "\n\t\"strings\""
+	}
 	return fmt.Sprintf(`
 package parse
 
 // NOTE: this file is generated via 'go:generate' - manual changes will be overwritten
 
 import (
-	"fmt"
+	"fmt"%s
 
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 )
@@ -231,7 +236,7 @@ import (
 %s
 %s
 %s
-`, id.codeForType(), id.codeForConstructor(), id.codeForFormatter(), id.codeForParser(), id.codeForParserInsensitive())
+`, stringsImport, id.codeForType(), id.codeForConstructor(), id.codeForFormatter(), id.codeForParser(), insensitive)
 }
 
 func (id ResourceIdGenerator) codeForType() string {


tombuildsstuff:
hey @sharebear 

Thanks for this PR.

Taking a look through here, whilst this PR does look to solve the issue of the missing `strings` import - the root-cause for this looks to be slightly different.

When generating the Resource ID Formatters/Parsers we first generate the code and then run each file through `go fmt -w {file}` (to format the code) and `goimports` (to ensure the imports are correct).

Both `go fmt` and `goimports` return exit codes indicating if the files have been updated or not - however since it's possible the source file could have no changes necessary (for example, if all of the required imports are included) - the generator doesn't check/handle those exit codes (since we're not particularly concerned with if there's been changes or not, as the compiler will ultimately tell us if the files are valid or not).

Whilst that approach works when the tooling is available - it'll fail when it's not, which appears to be the root cause here. As mentioned above, within the generator [we call both `go fmt` and `goimports`](https://github.com/terraform-providers/terraform-provider-azurerm/blob/962e31258685e4c9c5288617c5ca5b081bd7e766/azurerm/internal/tools/generator-resource-id/main.go#L675-L676) - whilst `go fmt` ships as a part of the Go tooling, `goimports` isn't shipped along with Go and appears to be [missing from `make tools`](https://github.com/terraform-providers/terraform-provider-azurerm/blob/962e31258685e4c9c5288617c5ca5b081bd7e766/GNUmakefile#L13-L21).

Removing `goimports` from my system, I'm able to confirm that by re-generating these files causes the diff:

```
$ git diff
diff --git a/azurerm/internal/services/desktopvirtualization/parse/application_group.go b/azurerm/internal/services/desktopvirtualization/parse/application_group.go
index 1469e4692..789433d66 100644
--- a/azurerm/internal/services/desktopvirtualization/parse/application_group.go
+++ b/azurerm/internal/services/desktopvirtualization/parse/application_group.go
@@ -4,7 +4,6 @@ package parse

 import (
        "fmt"
-       "strings"

        "github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 )
diff --git a/azurerm/internal/services/desktopvirtualization/parse/host_pool.go b/azurerm/internal/services/desktopvirtualization/parse/host_pool.go
index 7fa736677..3fd339092 100644
--- a/azurerm/internal/services/desktopvirtualization/parse/host_pool.go
+++ b/azurerm/internal/services/desktopvirtualization/parse/host_pool.go
@@ -4,7 +4,6 @@ package parse

 import (
        "fmt"
-       "strings"

        "github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 )
```
 
After adding `goimports` to `make tools` and then running `make tools` and re-generating I'm getting a clean-diff - as such I believe this is the root-cause here and I've opened #9551 to update `make tools` here.

In the interim you should be able to workaround this by running:

```sh
GO111MODULE=off go get -u golang.org/x/tools/cmd/goimports
```

As such whilst I'd like to thank you for this PR - I'm going to close this in favour of #9551 which adds that line to `make tools` which should fix the root cause here.

Thanks!

sharebear:
Thanks for the detailed explanation, I agree your solution is much better, still finding my feet in the golang ecosystem.

PR #9548, opened 2020-11-30
azurerm_storage_account  - allow hns when `account_tier` is `Premium`

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/9528


website/docs/r/storage_account.html.markdown
```diff
@@ -107,7 +107,7 @@ The following arguments are supported:
 
 * `is_hns_enabled` - (Optional) Is Hierarchical Namespace enabled? This can be used with Azure Data Lake Storage Gen 2 ([see here for more information](https://docs.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-quickstart-create-account/)). Changing this forces a new resource to be created.
 
--> **NOTE:** When this is set to `true` the `account_tier` argument must be set to `Standard`
+-> **NOTE:** When this is set to `true` the `account_tier` argument could be set to `Standard` or `Premium`.
```
katbyte, line 110:
```suggestion
-> **NOTE:** this can only be `true` when `account_tier` is `Standard` or `Premium`.
```

azurerm/internal/services/storage/tests/resource_arm_storage_account_test.go
```diff
@@ -819,6 +819,25 @@ func TestAccAzureRMStorageAccount_largeFileShare(t *testing.T) {
 	})
 }
 
+func TestAccAzureRMStorageAccount_withPremiumBlockBlobStorageAndEnabledHns(t *testing.T) {
```
katbyte, line 822: Could we update this to the new test framework?
```suggestion
func TestAccStorageAccount_hnsWithPremiumStorage(t *testing.T) {
```

katbyte (changes requested):
Thanks @neil-yechenwei - left some comments inline to address before merge

website/docs/r/storage_account.html.markdown
neil-yechenwei: updated

azurerm/internal/services/storage/tests/resource_arm_storage_account_test.go
neil-yechenwei: updated

neil-yechenwei:
@katbyte , thanks for your comments. I've updated code. Please have an another look.

katbyte (approved):
Thanks @neil-yechenwei - this LGTM 👍

PR #8850, opened 2020-10-12
`azurerm_cdn_endpoint` - `delivery_rule` condition optional `match_values`

magodo:
When the `delivery_rule` condition operator is set to `Any`, we should
not pass any `match_values` at all.

Fixes #8770

PR #9560, opened 2020-11-30
match_values not required when the operator is any

uolter:


katbyte (approved):
Thanks for the PR @uolter, could we update the documentation for this and add a check to error out if match value is missing when required?

thanks!

uolter:
> Thanks for the PR @uolter, could we update the documentation for this and add a check to error out if match value is missing when required?
> 
> thanks!

Yes, sure! Unfortunately I do not have a deep knowledge with the provider. However in many cases the match_values attribute is not required when the operator is cdn.*Any, while is required with different operator vaules. 
I "fixed" just the one useful to me, but some other match_values need to be fixed and a better validation to be introduces.
Thank you,
Walter (@uolter)

katbyte:
Hey @uolter - this is included in #8850 which i will hopefully merge for the next release, as such i'm going to close this. THanks again for the PR!

uolter:
> Hey @uolter - this is included in #8850 which i will hopefully merge for the next release, as such i'm going to close this. THanks again for the PR!

Yep, thanks a lot!

Issue #9557, opened 2020-11-30
azurerm_storage_data_lake_gen2_path: Error message has incorrect parameters in formatting


ross-p-smith:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.13.5
+ provider registry.terraform.io/hashicorp/azurerm v2.37.0

### Affected Resource(s)

azurerm_storage_data_lake_gen2_path

### Expected Behaviour

return fmt.Errorf("Error setting access control for Path %q in File System %q in Storage Account %q: %s", path, id.FileSystemName, id.AccountName, err)

### Actual Behaviour

return fmt.Errorf("Error setting access control for Path %q in File System %q in Storage Account %q: %s", id.FileSystemName, path, id.AccountName, err)

https://github.com/terraform-providers/terraform-provider-azurerm/blob/aaa9beffe58ea21187b0e4b2d9ec4279b77a4aff/azurerm/internal/services/storage/resource_arm_storage_data_lake_gen2_path.go#L290

PR #9561, opened 2020-11-30
azurerm_storage_data_lake_gen2_path: Fix error message

ross-p-smith:
This fixes the error message when ACL's cannot be applied. The Path and FileSystem name are in the incorrect position in the error message. #9557


katbyte (approved):
Thanks @ross-p-smith - LGTM 👍

PR #9567, opened 2020-11-30
generator/resource-id: support for generating Resource ID Validator functions

tombuildsstuff:
This PR introduces support for generating Validate functions for Resource ID's - when the other Resource ID's are generated.

**At this time this functionality is disabled**, but ultimately generates a Validate Func:

```go
package validate

// NOTE: this file is generated via 'go:generate' - manual changes will be overwritten

import (
	"fmt"

	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/analysisservices/parse"
)

func ServerID(input interface{}, key string) (warnings []string, errors []error) {
	v, ok := input.(string)
	if !ok {
		errors = append(errors, fmt.Errorf("expected %q to be a string", key))
		return
	}

	if _, err := parse.ServerID(v); err != nil {
		errors = append(errors, err)
	}

	return
}
```

and associated Unit Test:

```go
package validate

// NOTE: this file is generated via 'go:generate' - manual changes will be overwritten

import "testing"

func TestServerID(t *testing.T) {
	cases := []struct {
		Input string
		Valid bool
	}{

		{
			// empty
			Input: "",
			Valid: false,
		},

		{
			// missing SubscriptionId
			Input: "/",
			Valid: false,
		},

		{
			// missing value for SubscriptionId
			Input: "/subscriptions/",
			Valid: false,
		},

		{
			// missing ResourceGroup
			Input: "/subscriptions/12345678-1234-9876-4563-123456789012/",
			Valid: false,
		},

		{
			// missing value for ResourceGroup
			Input: "/subscriptions/12345678-1234-9876-4563-123456789012/resourceGroups/",
			Valid: false,
		},

		{
			// missing Name
			Input: "/subscriptions/12345678-1234-9876-4563-123456789012/resourceGroups/resGroup1/providers/Microsoft.AnalysisServices/",
			Valid: false,
		},

		{
			// missing value for Name
			Input: "/subscriptions/12345678-1234-9876-4563-123456789012/resourceGroups/resGroup1/providers/Microsoft.AnalysisServices/servers/",
			Valid: false,
		},

		{
			// valid
			Input: "/subscriptions/12345678-1234-9876-4563-123456789012/resourceGroups/resGroup1/providers/Microsoft.AnalysisServices/servers/Server1",
			Valid: true,
		},

		{
			// upper-cased
			Input: "/SUBSCRIPTIONS/12345678-1234-9876-4563-123456789012/RESOURCEGROUPS/RESGROUP1/PROVIDERS/MICROSOFT.ANALYSISSERVICES/SERVERS/SERVER1",
			Valid: false,
		},
	}
	for _, tc := range cases {
		t.Logf("[DEBUG] Testing Value %s", tc.Input)
		_, errors := ServerID(tc.Input, "test")
		valid := len(errors) == 0

		if tc.Valid != valid {
			t.Fatalf("Expected %t but got %t", tc.Valid, valid)
		}
	}
}
```

for each Resource ID that gets generated.

This ends up in `./validate/{resource}_id.go` - once the existing ID Validators are renamed and split out, it should be possible to generate these - but at this time there's a combination of Name and ID Parsers sharing the same files, so these need to be split prior to generation, presumably it'd be worth adding `./validate/{resource}_name.go` for the `name` validators.

### DIFF
--- a/.github/workflows/golint.yaml
+++ b/.github/workflows/golint.yaml
@@ -21,3 +21,4 @@ jobs:
       - uses: golangci/golangci-lint-action@v2
         with:
           version: 'v1.32'
+          args: --timeout=30m0s
--- a/azurerm/internal/tools/generator-resource-id/main.go
+++ b/azurerm/internal/tools/generator-resource-id/main.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"strings"
 	"unicode"
 )
@@ -17,6 +18,7 @@ func main() {
 	id := flag.String("id", "", "An example of this Resource ID")
 	rewrite := flag.Bool("rewrite", false, "Should this Resource ID be parsed insensitively, to workaround an API bug?")
 	showHelp := flag.Bool("help", false, "Display this message")
+	shouldValidate := false // TODO: enable once the existing files are renamed
 
 	flag.Parse()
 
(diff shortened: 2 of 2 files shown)


katbyte (approved):
LGTM 👍

PR #9583, opened 2020-12-01
Support for new property managed_by in azurerm_resource_group

neil-yechenwei:
API Spec: https://github.com/Azure/azure-rest-api-specs/blob/285beb67bcb1ff2035983cada86fadb12074750b/specification/resources/resource-manager/Microsoft.Resources/stable/2020-10-01/resources.json#L5557

### DIFF
--- a/azurerm/internal/services/resource/resource_arm_resource_group.go
+++ b/azurerm/internal/services/resource/resource_arm_resource_group.go
@@ -42,6 +42,13 @@ func resourceArmResourceGroup() *schema.Resource {
 
 			"location": azure.SchemaLocation(),
 
+			"managed_by": {
+				Type:         schema.TypeString,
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
hi @neil-yechenwei 

Thanks for opening this PR.

Taking a look through here unfortunately linking to the Swagger file doesn't explain how this functionality is intended to work, can you list the use-cases your looking to solve here? As it stands a Resource Group managing another Resource Group could use more restrictive validation, but it appears this functionality is incomplete in a number of ways, including testing?

Thanks!

tombuildsstuff:
@neil-yechenwei since we've not heard back here I'm going to close this PR for the moment - but if you can confirm the use-case this is looking to solve then we can re-open this and take another look.

PR #9606, opened 2020-12-01
I -> M: splitting the validate functions to allow generation

tombuildsstuff:
Once all of these files are renamed & split we can generate these and remove all of this human-maintained code.


jackofallops (approved):
LGTM 👍

PR #9609, opened 2020-12-01
Add support for service tags to ip_restriction for function_app and app_service

### LINKED ISSUE #9380: ip_restriction block in App_service missing "Service Tag" functionality
For the app service IP restriction, Microsoft supports the use of service Tags.  This is particularly useful for whitelisting the app service against something like Azure Front Door.  Terraform module for app service will only support an ipaddress in the ipaddress field and not a service tag:  https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/app_service#ip_restriction
I attach example screenshots of this property in the Azure environment.  Is this functionality getting added?  I tried entering the service tag in the ipaddress field, but it wouldn't take.
example
example2

sharebear:
Service tags are now allowed in the ip restrictions. In order to support this I've weakened the field validation on ip_address field. Behaviour is that if the field does not start with a number we assume you were attempting to set a service tag, otherwise default to assuming the field contains a CIDR.

I've intentionally not updated documentation yet until I've received feedback on whether this is an acceptable solution to the issue.

I chose to reuse the ip_address field for the service tag instead of adding a new field as this is consistent with the network_security_group resource, where the user can input either an ip or a service tag for source and destination.

Unfortunately this means that I had to remove the IsCIDR validation on the field meaning that validation will not catch errors that it may have previously caught, reducing the user experience a little.

Thinking further on this, would it be possible to implement and extended service tag validation similar to the location validation (the `ServiceTagsClient` can be used to retrieve a list of all service tags)? If so that should probably be a separate PR to enable it for all places a service tag would be valid input.

Last questions, what is the best strategy for testing this change? Is it possible to unit test? How much coverage should I add in the acceptance tests (e.g. one happy day test for both function_app and app_service or do I need to duplicate the tests that remove rules like `TestAccAzureRMFunctionApp_ipRestrictionRemoved`). Is there an ELI5 guide for setting up the prerequisites for running the acceptance tests against my own subscription?

Currently I've just tested manually with the following config

```terraform
resource "azurerm_function_app" "main" {
  name                       = "${var.prefix}-function"
  resource_group_name        = azurerm_resource_group.main.name
  location                   = azurerm_resource_group.main.location
  app_service_plan_id        = azurerm_app_service_plan.main.id
  storage_account_name       = azurerm_storage_account.main.name
  storage_account_access_key = azurerm_storage_account.main.primary_access_key


  site_config {
    ip_restriction {
      ip_address = "AzureEventGrid"
      priority   = 1
      name       = "My Filter"
    }
  }
}
```

Resolves #9380

### DIFF
--- a/azurerm/internal/services/web/app_service.go
+++ b/azurerm/internal/services/web/app_service.go
@@ -1745,12 +1745,14 @@ func flattenAppServiceIpRestriction(input *[]web.IPSecurityRestriction) []interf
 			if *ip == "Any" {
 				continue
 			} else {
-				if v.Tag == web.Default {
+				switch v.Tag {
+				case web.Default:
 					restriction["ip_address"] = *ip
-				} else if v.Tag == web.ServiceTag {
+				case web.ServiceTag:
 					restriction["service_tag"] = *ip
-				} else {
-					// FIXME: I want to error here if we got XffProxy. I don't remember being able to set this in the portal
+				default:
+					// FIXME: I want to error here if we got XffProxy. It doesn't seem possible to set in the portal
+					// but can theoretically come back from the API as defined
 				}
 			}
 		}
--- a/azurerm/internal/services/web/app_service.go
+++ b/azurerm/internal/services/web/app_service.go
@@ -812,7 +812,7 @@ func schemaAppServiceIpRestriction() *schema.Schema {
 				"ip_address": {
 					Type:         schema.TypeString,
 					Optional:     true,
-					ValidateFunc: validation.IsCIDR,
+					ValidateFunc: validation.StringIsNotEmpty,
 				},
 
 				"subnet_id": {
@@ -1848,6 +1848,12 @@ func expandAppServiceIpRestriction(input interface{}) ([]web.IPSecurityRestricti
 		ipSecurityRestriction := web.IPSecurityRestriction{}
 		if ipAddress == "Any" {
 			continue
+		}
+
+		// if ipAddress doesn't look like an ipAddress, set Tag to serviceTag
+		firstChar := ipAddress[0]
+		if !('0' <= firstChar && firstChar <= '9') {
+			ipSecurityRestriction.Tag = web.ServiceTag
 		}
 
 		if ipAddress != "" {
--- a/azurerm/internal/services/web/app_service.go
+++ b/azurerm/internal/services/web/app_service.go
@@ -812,7 +812,7 @@ func schemaAppServiceIpRestriction() *schema.Schema {
 				"ip_address": {
 					Type:         schema.TypeString,
 					Optional:     true,
-					ValidateFunc: validation.StringIsNotEmpty,
+					ValidateFunc: validation.IsCIDR,
 				},
 
 				"service_tag": {
--- a/azurerm/internal/services/web/app_service.go
+++ b/azurerm/internal/services/web/app_service.go
@@ -1746,13 +1746,10 @@ func flattenAppServiceIpRestriction(input *[]web.IPSecurityRestriction) []interf
 				continue
 			} else {
 				switch v.Tag {
-				case web.Default:
-					restriction["ip_address"] = *ip
 				case web.ServiceTag:
 					restriction["service_tag"] = *ip
 				default:
-					// FIXME: I want to error here if we got XffProxy. It doesn't seem possible to set in the portal
-					// but can theoretically come back from the API as defined
+					restriction["ip_address"] = *ip
 				}
 			}
 		}
--- a/azurerm/internal/services/web/app_service.go
+++ b/azurerm/internal/services/web/app_service.go
@@ -815,6 +815,12 @@ func schemaAppServiceIpRestriction() *schema.Schema {
 					ValidateFunc: validation.StringIsNotEmpty,
 				},
 
+				"service_tag": {
+					Type:         schema.TypeString,
+					Optional:     true,
+					ValidateFunc: validation.StringIsNotEmpty,
+				},
+
 				"subnet_id": {
 					// TODO - Remove in 3.0
 					Type:         schema.TypeString,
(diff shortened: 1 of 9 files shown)


sharebear:
There's no CONTRIBUTING file or obvious place to look for support, so apologies if it's seen as bad ettiquette to tag people directly without previous interaction. I also am consious that it appears there's a lot of energy going into a refactoring right now, but a quick nudge in the right way forward on this PR would be appreciated /cc @tombuildsstuff @ArcturusZhang

katbyte (changes requested):
Hi @sharebear - thanks for this PR. While i can appreciate the node to consistency i think it would be better to break out this into a separate `service_tag` property that conflicts with ip_address. Overloading `ip_address` is honestly is pretty confusing and not great UX.

azurerm/internal/services/web/app_service.go
```diff
@@ -1735,7 +1745,13 @@ func flattenAppServiceIpRestriction(input *[]web.IPSecurityRestriction) []interf
 			if *ip == "Any" {
 				continue
 			} else {
-				restriction["ip_address"] = *ip
+				if v.Tag == web.Default {
+					restriction["ip_address"] = *ip
+				} else if v.Tag == web.ServiceTag {
+					restriction["service_tag"] = *ip
+				} else {
+					// FIXME: I want to error here if we got XffProxy. I don't remember being able to set this in the portal
```
sharebear, line 1753: The flatten method here doesn't currently allow an error to be returned, should I refactor to allow it or handle this case silently?

sharebear:
@katbyte I just added a commit to expand the schema with a `service_tag` property as suggested. It appears to work based on some quick local testing. If you're happy with this I'll start looking at the integration tests and documentation

azurerm/internal/services/web/app_service.go
```diff
@@ -812,7 +812,13 @@ func schemaAppServiceIpRestriction() *schema.Schema {
 				"ip_address": {
 					Type:         schema.TypeString,
 					Optional:     true,
-					ValidateFunc: validation.IsCIDR,
+					ValidateFunc: validation.StringIsNotEmpty,
```
katbyte, line 815: we should revert this?

azurerm/internal/services/web/app_service.go
```diff
@@ -1735,7 +1745,15 @@ func flattenAppServiceIpRestriction(input *[]web.IPSecurityRestriction) []interf
...
 			} else {
-				restriction["ip_address"] = *ip
+				switch v.Tag {
+				case web.Default:
+					restriction["ip_address"] = *ip
+				case web.ServiceTag:
+					restriction["service_tag"] = *ip
+				default:
+					// FIXME: I want to error here if we got XffProxy. It doesn't seem possible to set in the portal
+					// but can theoretically come back from the API as defined
```
katbyte, line 1755: I think we can safely ignore it?

katbyte (commented):
Thanks @sharebear - looking good

azurerm/internal/services/web/app_service.go
sharebear: Ok, what I did for now was to set the `ip_address` property in the default case, this might be incorrect in the long term if this is to be supported, but doesn't change the behaviour from before.

azurerm/internal/services/web/app_service.go
sharebear: Of course, I was intending to do this but forgot, reverted now.

sharebear:
@katbyte Any guidance on what kind of test coverage I should add here? I think I've got the existing tests running locally now, even if they are very slow, not sure if this is rate limiting or limits of the region I'm running in (westeurope). I'm open for advice on how to make the experience a little more pleasant here...

katbyte:
@sharebear - adding a new ip restriction to the complete test config using service tags would probably be good enough

sharebear:
As far as I can tell, the linting error is not in code touched by this PR.... please let me know if I've missed something here.

CHANGELOG.md
```diff
@@ -9,12 +9,16 @@ IMPROVEMENTS:
...
+* `azurerm_app_service` - support `service_tag` in `ip_restriction` [GH-9609]
+* `azurerm_app_service_slot` - support `service_tag` in `ip_restriction` [GH-9609]
 * `azurerm_cdn_endpoint` - only send `content_types_to_compress` and `geo_filter` to the API when actually set [GH-9902]
 * `azurerm_cosmosdb_mongo_collection` - correctly read back the `_id` index when mongo 3.6 [GH-8690]
 * `azurerm_container_group` - support for the `volume.empty_dir` property [GH-9836]
 * `azurerm_data_factory_linked_service_azure_file_storage` - support for the `file_share` property [GH-9934]
 * `azurerm_dedicated_host` - support for addtional `sku_name` values [GH-9951]
 * `azurerm_function_app` - make `pre_warmed_instance_count` computed to use azure's default [GH-9069]
+* `azurerm_function_app` - support `service_tag` in `ip_restriction` [GH-9609]
+* `azurerm_function_app_slot` - support `service_tag` in `ip_restriction` [GH-9609]
```
jackofallops, line 21: Could you revert changes to CHANGELOG.md please? This is updated by the maintainers via another process when a PR is accepted, tested and merged.

sharebear: Done, I just rebased away the commit. I'd love to see a `CONTRIBUTING` file that clearly communicated the expectations for Pull Requests to be expected.

jackofallops: We're working on that, hopefully not too far out after we catch up from the holidays 👍

jackofallops (approved):
Thanks @sharebear - This LGTM now, I'll get the tests run asap and assuming all looks good there we'll be good to merge. 👍

jackofallops:
Tests looking OK, failures are unrelated / transient:
image

PR #9653, opened 2020-12-03
web: splitting the validate funcs to enable generation

tombuildsstuff:
Once this is in we can generate the ID validation functions


manicminer (approved):
LGTM

PR #9654, opened 2020-12-03
Generating Resource ID Validator functions

tombuildsstuff:
This PR generates Validate functions for Resource ID's at the same time as generating the Resource ID Formatters/Parsers, using the same information.

Whilst these may be superfluous at this time - we'll work through adding these into resources over time, as such it makes sense to generate these so that we can take advantage of them.

I'd strongly suggest reviewing this commit by commit - since each commit contains a particular area of focus (generator, generating the existing, generating the new ones) so this should be easier to review rather than in one go.

Once this PR's in, and this is ready to use - I'll send a PR documenting this in the readme, and adding a Github Action to ensure the generated files match what's checked in.

Dependent on #9653


manicminer (approved):
LGTM :+1:

PR #9666, opened 2020-12-03
relay/web: consolidating tests

tombuildsstuff:
For relay this moves the Parsers up a level and then generates those
For web this moves those tests up a level so they're all run in the same place

Due to a limitation of the test runner to work around a limitation in Go - all acctests must either be in the tests folder or in the root - not shared across them


jackofallops (approved):
LGTM 👍

PR #9672, opened 2020-12-04
Function App resource doc - Document default value for `ftps_state` argument

fcatacut:
Documents the default value of `site_config.ftps_state` when it's not specified: <https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/function_app#ftps_state>.


tombuildsstuff (approved):
LGTM 👍

PR #9674, opened 2020-12-04
Fix resource id generator on windows

ArcturusZhang:
The resource id generator now raises errors on windows because the absolute path on windows will be using `\` as path separator instead of `/`. 
In this PR, I just simply replace all of the `\` before split the path into segments by `/` to avoid this panic


tombuildsstuff (approved):
LGTM 👍

PR #9675, opened 2020-12-04
Support new properties for azurerm_subscription_template_deployment

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/9721

--- PASS: TestAccSubscriptionTemplateDeployment_withOutputs (157.31s)
--- PASS: TestAccSubscriptionTemplateDeployment_singleItemUpdatingParams (163.57s)
--- PASS: TestAccSubscriptionTemplateDeployment_empty (282.25s)
--- PASS: TestAccSubscriptionTemplateDeployment_singleItemUpdatingTemplate (284.62s)
--- PASS: TestAccSubscriptionTemplateDeployment_switchBetweenTemplateLinkAndTemplateContent (409.12s)


tombuildsstuff:
@neil-yechenwei what's the use-case adding these fields is looking to solve? is there a Github issue for this?

neil-yechenwei:
@tombuildsstuff , attached the github issue link https://github.com/terraform-providers/terraform-provider-azurerm/issues/9721 here,

tombuildsstuff:
hi @neil-yechenwei 

Similar to #9719 - after chatting about this internally we believe there's more downsides than upsides to supporting sourcing content from a URI at this time, since these variables are unlikely to be reused across deployments and may contain sensitive variables (which would have to be available at a publically available URI).

As such whilst I'd like to thank you for opening this issue - at this point in time we don't plan to support the `parameters_link` field within the `azurerm_subscription_template_deployment` resource and as such I'm going to close this PR for the moment - however should a compelling use-case become available we can take another look.

Thanks!

PR #9698, opened 2020-12-05
role_assignment: fix immediate read after write issue

dlamotte:
Azure AD API requests provide inconsistent results.  The `Create` API
call returns enough information to be used in future read calls and
should be used instead of requiring a `Get` API call immediately after.
Once we have the ID, we can wait for the state change later and poll
until the AD API becomes consistent for the resource just created.


gandelman-a (approved):
This fixes the race we've been hitting. Thank you.

katbyte (approved):
Thanks @dlamotte - this LGTM 👍

dlamotte:
@katbyte 👍  I just merged master in as there was conflicts.

jackofallops:
Hi @dlamotte - Thanks for this PR.  I've had to take a look at this issue for another reason and needed to take a different approach to avoid introducing other subtle bugs / issues, as such I'll need to close your PR (since it's part of an Org, not a user, I can't push over the top) and open a new one with my changes. 
Thanks again for your contribution, we appreciate it!

dlamotte:
@jackofallops I took a look at your PR and it looks good to me.  Excited to have it merged.

PR #9716, opened 2020-12-07
New Github Action: checking the Generator output matches what's committed

tombuildsstuff:
This PR introduces a new Github Action and Makefile step to confirm that the generated code matches what's checked in - I'll send a separate PR to add documentation for how these Resource ID's can be generated

This also fixes a bug in the generator for running tests by service packages, since Service Registrations are now reused across Typed and Untyped resources and this wasn't taken into account


jackofallops (approved):
LGTM 👍

PR #9736, opened 2020-12-07
refactor: removing unnecessary arguments with subscriptionId for auto-generated Resource ID Formatters/Parsers

tombuildsstuff:
This cleans up some now unnecessary code for the generated Resource ID Formatters/Parsers - there's still some manual ones which require further work - but that's one for another PR.


katbyte (approved):
LGTM 👍

PR #9748, opened 2020-12-08
generator-resource-id: filename support cascaded uppercase runes

magodo:
Support the use case as mentioned in: https://github.com/terraform-providers/terraform-provider-azurerm/pull/9134#discussion_r538127206.


tombuildsstuff (approved):
LGTM 👍

PR #9753, opened 2020-12-08
azurerm_storage_account_network_rules Fix misleading error

lawrencegripper:
In the case that you attempt to use this resource on a storage account which already has network rules attached with either `networkAcls.ipRules` or `networkAcls.virtualNetworkRules` non-empty arrays the error returned suggests that you use `terraform import` to fix it.

That won't fix it and is misleading, I've updated the error to indicate that the user should instead remove any existing network rules on the storage account.


tombuildsstuff:
@lawrencegripper importing this resource into the state and then running an update should update those rules - what behaviour are you seeing here?

lawrencegripper:
My bad @tombuildsstuff, really sorry for wasting time, we have two resources one creating the account and this one managing the rules. 

I read the error and assumed it was incorrect as the storage account it suggested importing was already managed by TF (by the `azurerm_storage_account`).

tombuildsstuff:
@lawrencegripper not a problem at all - just double checking I understood what we were looking to fix, glad that's working for you 👍

PR #9754, opened 2020-12-08
servicebus: generating & switching to use the remaining Resource ID Formatters/Parsers

tombuildsstuff:
This PR generates the remaining ID Formatters/Parsers for this package, switches over to using them, and adds enhanced import validation for the following resources:

* `azurerm_servicebus_queue`
* `azurerm_servicebus_queue_authorization_rule`
* `azurerm_servicebus_subscription`
* `azurerm_servicebus_namespace`
* `azurerm_servicebus_namespace_authorization_rule`
* `azurerm_servicebus_subscription_rule`
* `azurerm_servicebus_topic_authorization_rule`
* `azurerm_servicebus_topic`

In addition this PR also fixes an issue in the Resource ID generator when multiple `subscriptions` elements are present (for example in the ServiceBus Subscription ID) - ands moves the validation logic for ServiceBus into this package


azurerm/internal/services/servicebus/internal.go
```diff
@@ -0,0 +1,112 @@
...
+func authorizationRuleCustomizeDiff(d *schema.ResourceDiff, _ interface{}) error {
+	listen, hasListen := d.GetOk("listen")
+	send, hasSend := d.GetOk("send")
+	manage, hasManage := d.GetOk("manage")
+
+	if !hasListen && !hasSend && !hasManage {
+		return fmt.Errorf("One of the `listen`, `send` or `manage` properties needs to be set")
+	}
+
+	if manage.(bool) && !listen.(bool) && !send.(bool) {
```
jackofallops, line 107: Based on the error message, should this not be?
```suggestion
	if manage.(bool) && (!listen.(bool) || !send.(bool)) {
```

azurerm/internal/services/servicebus/validate/subscription_name.go
```diff
@@ -0,0 +1,15 @@
...
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+)
+
+func SubscriptionName() schema.SchemaValidateFunc {
+	return validation.StringMatch(
+		regexp.MustCompile("^[a-zA-Z0-9][-._a-zA-Z0-9]{0,48}([a-zA-Z0-9])?$"),
+		"The name can contain only letters, numbers, periods, hyphens and underscores. The name must start and end with a letter or number and be less then 50 characters long.",
```
jackofallops, line 13: This regexp will allow up to 50 chars, but the error message specifies less than 50?

jackofallops (approved):
Thanks @tombuildsstuff - couple of questions below, but otherwise LGTM 👍

azurerm/internal/services/servicebus/validate/subscription_name.go
tombuildsstuff: 48 + the start and end, so this is 50 :)

PR #9755, opened 2020-12-08
securitycenter: refactoring to use an ID Formatter

tombuildsstuff:
This is hand-maintained due to the nature of these ID's - but this updates this service package to match the others


jackofallops (approved):
LGTM 👍

PR #9770, opened 2020-12-09
resource_id_gen: always set the field/argument to `name` for last segment

magodo:
The original code will make the last field of the ID structure
superflurous when its last segment name is not the same as the type name
specified in the commandline. This is possible if the resource name used
in swagger is different than it in Terraform.

### DIFF
--- a/azurerm/internal/tools/generator-resource-id/main.go
+++ b/azurerm/internal/tools/generator-resource-id/main.go
@@ -197,7 +197,7 @@ func NewResourceID(typeName, servicePackageName, resourceId string) (*ResourceId
 			continue
 		}
 
-		var segmentBuilder = func(key, value string) ResourceIdSegment {
+		var segmentBuilder = func(key, value string, last bool) ResourceIdSegment {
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hey @magodo

Thanks for this PR.

Whilst we looked at taking this approach whilst implementing the generator, unfortunately "name" is used as a segment in some Resource ID's (for example the upcoming App Configuration Data Plane ones). As such unfortunately defaulting this last segment to name would break parsing those ID's - meaning that we're unable to take this approach.

The Generator intentionally doesn't infer the name of the ID Parser from the last segment, pulling this from a name which makes more sense in Terraform. On a related note, we intentionally aren't expose a "replace" argument as a command line argument in the generator to avoid some of the larger maintainability issues with the Swagger/autorest markdown files.

As such whilst I'd like to thank you for this PR I'm going to close this PR for the moment - but this is something we're thinking about and will likely become less of an issue with some of the changes currently in flight.

Thanks!

PR #9772, opened 2020-12-09
network: splitting Load Balancers out into their own package

tombuildsstuff:
This is part of some larger work in the networking package to make this more maintainable and limits the blast radius of Networking SDK upgrades - Load Balancer consumes an older version of the networking package at this time due to upstream API issues

This PR's mostly just a lift & shift - nothing particularly exciting to note


jackofallops (approved):

PR #9778, opened 2020-12-09
generator: outputting the type name and sorting the segments to match the ID

tombuildsstuff:
This better matches the existing output - example:

```go
type ApiDiagnosticId struct {
	SubscriptionId string
	ResourceGroup  string
	ServiceName    string
	ApiName        string
	DiagnosticName string
}

func (id ApiDiagnosticId) String() string {
	segments := []string{
		fmt.Sprintf("Diagnostic Name %q", id.DiagnosticName),
		fmt.Sprintf("Api Name %q", id.ApiName),
		fmt.Sprintf("Service Name %q", id.ServiceName),
		fmt.Sprintf("Resource Group %q", id.ResourceGroup),
	}
	segmentsStr := strings.Join(segments, " / ")
	return fmt.Sprintf("%s: (%s)", "Api Diagnostic", segmentsStr)
}
```

Ultimately this makes this easier to consume in error messages directly e.g.:

``` go
return fmt.Errorf("checking for presence of existing %s: %+v", id, err)
return fmt.Errorf("retrieving %s: %+v", id, err)
```

Prior to this being merged the generator will need re-running against everything - but I've intentionally not done that due to open PR's

### DIFF
--- a/azurerm/internal/services/analysisservices/parse/server.go
+++ b/azurerm/internal/services/analysisservices/parse/server.go
@@ -25,10 +25,11 @@ func NewServerID(subscriptionId, resourceGroup, name string) ServerId {
 
 func (id ServerId) String() string {
 	segments := []string{
-		fmt.Sprintf("Resource Group %q", id.ResourceGroup),
 		fmt.Sprintf("Name %q", id.Name),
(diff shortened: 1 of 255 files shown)


jackofallops (approved):
LGTM 👍

PR #9783, opened 2020-12-09
dependencies: upgrading github.com/Azure/go-autorest/validation to v0.3.1

tombuildsstuff:
This allows us to opt out of the Azure SDK's validation - since this isn't particularly helpful for Terraform's problem space since this only takes affect half way through an apply (when the API call is made), when this'll already be caught by Terraform's validation.


jackofallops (approved):
LGTM 👍

tombuildsstuff:
Tests look good:

Screenshot 2020-12-09 at 17 22 24

PR #9793, opened 2020-12-10
network/subnet: correct logic for policy bools

dlamotte:
The Azure CLI is inverted from the API.  This corrects the logic.

### DIFF
--- a/azurerm/internal/services/network/subnet_resource.go
+++ b/azurerm/internal/services/network/subnet_resource.go
@@ -504,26 +504,18 @@ func flattenSubnetDelegation(delegations *[]network.Delegation) []interface{} {
 	return retDeles
 }
 
-// TODO: confirm this logic below
-
 func expandSubnetPrivateLinkNetworkPolicy(enabled bool) *string {
-	// This is strange logic, but to get the schema to make sense for the end user
-	// I exposed it with the same name that the Azure CLI does to be consistent
-	// between the tool sets, which means true == Disabled.
 	if enabled {
-		return utils.String("Disabled")
+		return utils.String("Enabled")
 	}
 
-	return utils.String("Enabled")
+	return utils.String("Disabled")
 }
 
 func flattenSubnetPrivateLinkNetworkPolicy(input *string) bool {
-	// This is strange logic, but to get the schema to make sense for the end user
-	// I exposed it with the same name that the Azure CLI does to be consistent
-	// between the tool sets, which means true == Disabled.
 	if input == nil {
-		return false
+		return true
 	}
 
-	return strings.EqualFold(*input, "Disabled")
+	return strings.EqualFold(*input, "Enabled")
 }


aleksa (approved):

WodansSon (changes requested):
@dlamotte, thank you so much for the PR. The PR looks good, but it would be a breaking change and would flip the logic thus breaking the implementation of the resource. I am not adverse to changing it, but might I suggest exposing two new attributes to follow what the API service team has done. It appears that the CLI team has have changed the name of the argument from:
Old | New
-- | --
`enforce-private-endpoint-network-policies` | `disable-private-endpoint-network-policies` 
`enforce-private-link-service-network-policies` | `disable_private_link_service_network_policies`

Yet in the actual API they are implemented as:
* PrivateEndpointNetworkPolicies = [Enabled | Disabled]
* PrivateLinkServiceNetworkPolicies = [Enabled | Disabled]

With both of the CLI command line argument the issue is that they added the verbs enforce. enable, and disable which flips the meaning of the action when compared to the API implementation, which is why this logic was originally implemented this way(e.g. Enabled == False and Disabled == True). 

Which means when you define the attribute in the Terraform config file:
enforce_private_link_endpoint_network_policies = true 

The API receives:
PrivateEndpointNetworkPolicies: Disabled 

Because the policy you are **enforcing** is to **disable** the the private endpoint policies to allow for a private endpoint to be connected to the subnet, I know this is very confusing but that was the reason behind this.

So to correct this confusion I would suggest that we break the dependency on matching the CLI implementation and instead adopt the logic used in the API directly. I would mark the two existing attributes as **deprecated** and expose the following two new fields in the resource:

* PrivateEndpointNetworkPolicies
* PrivateLinkServiceNetworkPolicies

For the new attributes they would follow exactly what you have in your PR (e.g. Enabled == True and Disabled == False). By doing this way it solves the breaking changes, as the original fields can still be used with the existing logic and the new fields can also be used with the modified logic. Does that make sense? WDYT?

katbyte:
@WodansSon & @dlamotte - may i suggest putting enabled on the end as that would be consistent with all other boolean properties in the provider, ie `private_link_service_network_policies_enabled`

WodansSon:
@katbyte - To bring this into alignment we would have to use `private_link_service_network_policies_disabled` if you are ok with that I am fine with the name change. 🙂

dlamotte:
In the end, I don't have too much of a preference.  With my experience with terraform, I feel like resources are closer to the APIs than any relevant CLI, so I'm not sure why we would care what the CLI is doing.  That said, I'm not clear what we're agreeing to here.  Are you saying we should have two new attributes that don't have inverted logic (matching the API)?

- `private_endpoint_network_policies` `true == Enabled`
- `private_link_service_network_policies` `true == Enabled`

WodansSon:
@dlamotte - Yes, that is what I was suggesting, but after discussing the PR with other engineers we have come to the conclusion that it would be best to keep the attributes and logic as is (see the original [PR](https://github.com/terraform-providers/terraform-provider-azurerm/pull/4493) for full discussion about the naming of these attributes). So with that said, if it's ok with you I am going to close this PR as it is currently by design. Thanks again for submitting this tho' 🚀

dlamotte:
Honestly, I think this is wrong and very confusing.  That said, as long as it remains consistent and this is intentional, I can at least live with the current behavior.

Anyone reading Azure docs https://docs.microsoft.com/en-us/azure/private-link/disable-private-endpoint-network-policy will think the correct setting for this is `false` in terraform.

PR #9794, opened 2020-12-10
network/private-endpoint: allow alias as well

dlamotte:
The input Resource ID may be a resource id or service alias.  This
validation is preventing the service alias from being accepted.  The
service alias has a standard suffix that can be used to validate the
input.

https://docs.microsoft.com/en-us/rest/api/virtualnetwork/privatelinkservices/get#privatelinkservice
https://docs.microsoft.com/en-us/rest/api/virtualnetwork/privateendpoints/createorupdate#privatelinkserviceconnection

It's not clear from the documentation, but I've verified that using the
Alias is functional this way.

### DIFF
--- a/azurerm/internal/services/network/private_endpoint_resource.go
+++ b/azurerm/internal/services/network/private_endpoint_resource.go
@@ -106,10 +106,22 @@ func resourceArmPrivateEndpoint() *schema.Resource {
 							ForceNew: true,
 						},
 						"private_connection_resource_id": {
-							Type:         schema.TypeString,
-							Required:     true,
-							ForceNew:     true,
-							ValidateFunc: azure.ValidateResourceID,
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+							ValidateFunc: func(i interface{}, k string) (warnings []string, errors []error) {
+								v, ok := i.(string)
+								if !ok {
+									errors = append(errors, fmt.Errorf("expected type of %q to be string", k))
+									return
+								}
+
+								if strings.HasSuffix(v, ".azure.privatelinkservice") {
+									return
+								}
+
+								return azure.ValidateResourceID(i, k)
+							},
 						},
 						"subresource_names": {
 							Type:     schema.TypeList,
--- a/website/docs/r/private_endpoint.html.markdown
+++ b/website/docs/r/private_endpoint.html.markdown
@@ -95,6 +95,39 @@ resource "azurerm_private_endpoint" "example" {
 }
 ```
 
+Using a Private Link Service Alias with existing resources:
+
+```hcl
+data "azurerm_resource_group" "rg" {
+  name = "example-resources"
+}
+
+data "azurerm_virtual_network" "vnet" {
+  name                = "example-network"
+  resource_group_name = data.azurerm_resource_group.rg.name
+}
+
+data "azurerm_subnet" "subnet" {
+  name                 = "default"
+  virtual_network_name = data.azurerm_virtual_network.vnet.name
+  resource_group_name  = data.azurerm_resource_group.rg.name
+}
+
+resource "azurerm_private_endpoint" "example" {
+  name                = "example-endpoint"
+  location            = data.azurerm_resource_group.rg.location
+  resource_group_name = data.azurerm_resource_group.rg.name
+  subnet_id           = data.azurerm_subnet.subnet.id
+
+  private_service_connection {
+    name                           = "example-privateserviceconnection"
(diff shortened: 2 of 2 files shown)


aleksa (approved):

azurerm/internal/services/network/private_endpoint_resource.go
```diff
@@ -106,10 +106,22 @@ func resourceArmPrivateEndpoint() *schema.Resource {
 							ForceNew: true,
 						},
 						"private_connection_resource_id": {
```
katbyte, line 108: WOuld it make sense to add a new property `private_connection_service_alias` rather then overload this one? - as resource_id is fairly explicit in what values it is expecting.

katbyte (changes requested):
Thanks for this fix @dlamotte - i've left one question about the schema design inline to answer - also docs need to be updated.

azurerm/internal/services/network/private_endpoint_resource.go
dlamotte: This is how the Azure API works (says one thing, accepts a few things).  Does this need to be more strict than that?

dlamotte: https://docs.microsoft.com/en-us/rest/api/virtualnetwork/privateendpoints/createorupdate#request-body

tombuildsstuff: hey @dlamotte

Apologies for the delayed reply here.

Chatting about this internally whilst the Azure API exposes a single field for multiple purposes within Terraform we're tending to split these out into separate fields so this is clearer to users and then set the appropriate field in the request/parsing the response back out to two fields here - and we can ensure that users only set a single field by using an `ExactlyOneOf` in the Schema too. As Katie has mentioned the benefit to doing this is primarily advanced validation - as such could we split this out here?

Thanks!

dlamotte: Sure.  Sounds good.

katbyte:
Hey @dlamotte - just wondering if your still working on this?

dlamotte:
@katbyte sorry, just haven't had time to come back around to it.  Happy to fix this up though but if someone else takes it and fixes it, fine by me.

mbfrahry:
Hey @dlamotte! I've got the changes written the way we'd like them but I can't push to your branch. Can you make sure that `Allow edits by maintainers` is checked for this PR? It should be near where you can open a PR 

image

Also, I was unable to get a prviate endpoint created through an alias. I kept getting the below error

```
Expect fully qualified resource Id that start with '/subscriptions/{subscriptionId}' or '/providers/{resourceProviderNamespace}/'.
```

Can you confirm that you were able to get alias working when submitting it through the API?

dlamotte:
@mbfrahry sorry, finally getting back to you on this.  I asked our IT if this was possible and they said it was not due to this being an Organization repo.  Let me know how'd you like to proceed.  I'm perfectly happy if you prefer to open a new PR submit fixes on that.

katbyte:
@dlamotte - i think @mbfrahry was looking to confirm that you could get the service alias to work before he opened a new PR as it didn't work for him

dlamotte:
@katbyte it definitely worked once I made the tweak in this PR.  So I can confirm that bit at least.  The back and forth here was more about ensuring that terraform hard distinguished between the variants even though the API doesn't (ie: Service Alias and Resource ID are accepted as the ID parameter in the API IIRC).

mbfrahry:
Hey @dlamotte, I wrote out a test for your changes but the API rejected the alias because it wasn't in the right format.

```
Expect fully qualified resource Id that start with '/subscriptions/{subscriptionId}' or '/providers/{resourceProviderNamespace}/'.
```

Since I can't push to this branch, would you be open to me pushing out a new PR that you can play with and see if you can get these changes working?

dlamotte:
@mbfrahry please do open a new PR.  Please note that Service Alias format is like: `name.00000000-0000-0000-0000-000000000000.centralus.azure.privatelinkservice` and won't have the `/subscriptions` or `/providers` prefix.

mbfrahry:
Just opened the new PR #10779 @dlamotte. If you note [this line](https://github.com/terraform-providers/terraform-provider-azurerm/pull/10779/files#diff-1c8fa442341f21fcdac8d8169e570adb1d76915f70e7d82b6798eaab4b19100dR739) where we use alias but the API returns a 400 and the above error

dlamotte:
Replaced by https://github.com/terraform-providers/terraform-provider-azurerm/pull/10779

PR #9797, opened 2020-12-10
Support new properties `alert_rule_template_guid` and `display_name_exclude_filter` for azurerm_sentinel_alert_rule_ms_security_incident

neil-yechenwei:
The purpose of this PR is to support below two new properties for `azurerm_sentinel_alert_rule_ms_security_incident`.

The first property is `alert_rule_template_guid`. It is the id of the alert rule template which is used to create this Sentinel Scheduled Alert Rule.

The second property is `display_name_exclude_filter`. It is used to create incidents when the alert display name doesn't contain text from the specified list.

--- PASS: TestAccAzureRMSentinelAlertRuleMsSecurityIncident_basic (165.11s)
--- PASS: TestAccAzureRMSentinelAlertRuleMsSecurityIncident_complete (165.12s)
--- PASS: TestAccAzureRMSentinelAlertRuleMsSecurityIncident_withAlertRuleTemplateName (165.74s)
--- PASS: TestAccAzureRMSentinelAlertRuleMsSecurityIncident_requiresImport (178.59s)
--- PASS: TestAccAzureRMSentinelAlertRuleMsSecurityIncident_withDisplayNameExcludeFilter (300.20s)
--- PASS: TestAccAzureRMSentinelAlertRuleMsSecurityIncident_update (331.34s)

API doc for these two properties:
https://docs.microsoft.com/en-us/rest/api/securityinsights/alertrules/createorupdate#microsoftsecurityincidentcreationalertrule


website/docs/r/sentinel_alert_rule_ms_security_incident.html.markdown
```diff
@@ -56,12 +56,18 @@ The following arguments are supported:
 
 ---
 
+* `alert_rule_template_name` - (Optional) The name of the alert rule template which is used to create this Sentinel Scheduled Alert Rule. Changing this forces a new Sentinel MS Security Incident Alert Rule to be created.
```
katbyte, line 59: If this can only be a guid, wouldn't';t this make more sense as
```suggestion
* `alert_rule_template_guid` - (Optional) The name of the alert rule template which is used to create this Sentinel Scheduled Alert Rule. Changing this forces a new Sentinel MS Security Incident Alert Rule to be created.
```

katbyte (changes requested):

website/docs/r/sentinel_alert_rule_ms_security_incident.html.markdown
neil-yechenwei: updated

neil-yechenwei:
@katbyte , thanks for your comments. I've updated code per your suggestion.

katbyte (approved):
Thanks @neil-yechenwei - this LGTM 👍

PR #9801, opened 2020-12-10
loganalytics: removing the superflurous `subscriptionId` argument from the ID func

tombuildsstuff:
This is the last remnant, at which point we can remove this unnecessary argument from the formatter and re-generate


jackofallops (approved):
LGTM 👍

tombuildsstuff:
Generation failure has been fixed via #9802 - have re-run this locally and it's fine, but this isn't worth another full test suite run in GHA

PR #9802, opened 2020-12-10
media: switching to use a Resource ID Formatter for `azurerm_media_transform`

tombuildsstuff:
Also fixing some tests and making things case-sensitive whilst passing through, and adding Requires Import to both Media Account and Media Transform


tombuildsstuff:
Tests pass:

Screenshot 2020-12-10 at 12 48 41

jackofallops (approved):

PR #9804, opened 2020-12-10
kusto: switching the delete function to use the same Resource ID Parser

tombuildsstuff:
Noticed whilst passing through the Read is using this but the Delete isn't


jackofallops (approved):
LGTM 👍

PR #9811, opened 2020-12-10
r/media_asset: switching to use a Resource ID Formatter

tombuildsstuff:
This is a new resource so this should be switched over too, I'd missed this earlier on.


manicminer (approved):
LGTM 👍

PR #9819, opened 2020-12-10
test shim: security centre

katbyte:


azurerm/internal/services/securitycenter/security_center_contact_resource_test.go
```diff
@@ -1,188 +1,130 @@
...
+			Config: r.template("require@example.com", "+1-555-555-5555", true, true),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+				check.That(data.ResourceName).Key("email").HasValue("require@example.com"),
+				check.That(data.ResourceName).Key("phone").HasValue("+1-555-555-5555"),
+				check.That(data.ResourceName).Key("alert_notifications").HasValue("true"),
+				check.That(data.ResourceName).Key("alerts_to_admins").HasValue("true"),
+			),
 		},
+		data.RequiresImportErrorStepWithConfig(r.requiresImportCfg("email1@example.com", "+1-555-555-5555", true, true)),
```
tombuildsstuff, line 64: rather than doing this, we can make this:

```suggestion
		data.RequiresImportStep(func(data acceptance.TestData) string {
			return r.requiresImportCfg("email1@example.com", "+1-555-555-5555", true, true)
		}),
```

which means we can remove this overload, since everything should be using TestData here

azurerm/internal/acceptance/steps.go
```diff
@@ -92,3 +92,10 @@ func (td TestData) RequiresImportErrorStep(configBuilder func(data TestData) str
 		ExpectError: RequiresImportError(td.ResourceType),
 	}
 }
+
+func (td TestData) RequiresImportErrorStepWithConfig(config string) resource.TestStep {
+	return resource.TestStep{
+		Config:      config,
+		ExpectError: RequiresImportError(td.ResourceType),
+	}
+}
```
tombuildsstuff, line 101: which means we should be able to remove this - since we should be encouraging leaning on `data` where possible (that means we could replay those tests easier in the future)

tombuildsstuff (approved):
a couple of comments but this otherwise LGTM 👍

PR #9834, opened 2020-12-11
resource-ids: removing the unused `subscriptionId` argument

tombuildsstuff:
This PR removes the unused `subscriptionId` argument from the ID Formatters, now that these are (almost) all generated

Unfortunately this'll conflict with everything else, but there's not really another way of doing this right now


manicminer (approved):
LGTM  🚀

tombuildsstuff:
Ran the resources tests and they look good too 👍

PR #9836, opened 2020-12-11
azurerm_container_group - support `volume.empty_dir`

afdezl:
This PR adds the `emptyDir` volume type for ACI.
Note that because Azure requires this to be empty, it is passed as a bool and then turned into an empty object.


afdezl:
References: #1860

katbyte (changes requested):
Thanks for the PR @afdezl - overall this looks great but we will need to set the value for empty dir on read to prevent this a failing test:

`container.0.volume.0.empty_dir:                      "false" => "true" (forces new resource)`

afdezl:
> Thanks for the PR @afdezl - overall this looks great but we will need to set the value for empty dir on read to prevent this a failing test:
> 
> `container.0.volume.0.empty_dir: "false" => "true" (forces new resource)`

Thanks @katbyte. Just implemented.

katbyte (approved):
Thanks @afdezl - this LGTM now 🚀

PR #9842, opened 2020-12-12
New resource/Data Source: azurerm_ssh_public_key

### LINKED ISSUE #9822: Support for SSH Keys
### Community Note

### Description

There is no current resource / data source for SSH Keys

https://docs.microsoft.com/en-us/rest/api/compute/sshpublickeys

### References

* https://docs.microsoft.com/en-us/rest/api/compute/sshpublickeys

AliAllomani:
Added new resource and data source `azurerm_ssh_public_key`  to support creation/retrieving SSH Public Keys

Fixes #9822

* https://docs.microsoft.com/en-us/rest/api/compute/sshpublickeys

Documentation Updated + Acceptance test added and passed.


katbyte (changes requested):
Thanks for this @AliAllomani - overall this looks good, but i have two questions:
- is this only for VMs? if so would it make sense to call it `azurerm_virtual_machine_ssh_public_key`?
- should be be normalizing the ssh key at all?

thanks

AliAllomani:
> Thanks for this @AliAllomani - overall this looks good, but i have two questions:
> 
> * is this only for VMs? if so would it make sense to call it `azurerm_virtual_machine_ssh_public_key`?
> * should be be normalizing the ssh key at all?
> 
> thanks

Hello @katbyte,

This can be considered as ssh keys store which can be used with any resource that accepts ssh key as input, no only virtual machines, as there are several azure services need ssh keys.

The service also support generating ssh keys but due to security considerations and that the api returns the private key only one time on generation i think it wouldn't be possible to add this feature to the terraform resource, if we compare it with aws it would be corresponding to `aws_key_pair`.

Thanks

katbyte (approved):
Thanks @AliAllomani - this LGTM 👍

PR #9853, opened 2020-12-14
r/log_analytics_workspace: force-upgrading the resource id's

### LINKED ISSUE #9790: azurerm_log_analytics_workspace WorkspaceMigrateState looking up resource_group rather than resource_group_name
I'm running the latest version and am trying to redeploy a log analytics workspace, and for some reason it is triggering this WorkspaceMigrateState call, and this code is trying to lookup an attribute (resource_group) that isn't part of the azurerm_log_analytics_workspace definition, and as far as I can tell going back through releases back to 1.24 was not part of it then.  I believe it should be looking up resource_group_name.

Anyhow, it is then triggering an error as it gets a null resource group for use.

I can't see why this code is actually being called either.
```

func WorkspaceMigrateState(v int, is *terraform.InstanceState, meta interface{}) (*terraform.InstanceState, error) {
	switch v {
	case 0:
		log.Println("[INFO] Migrating Log Analytics Workspace State V0 to V1")
		return workspaceStateMigrateV0toV1(is, meta)
	default:
		return is, fmt.Errorf("Unexpected schema version: %d", v)
	}
}

func workspaceStateMigrateV0toV1(is *terraform.InstanceState, meta interface{}) (*terraform.InstanceState, error) {
	if is.Empty() {
		log.Println("[DEBUG] Empty InstanceState; nothing to migrate.")
		return is, nil
	}

	log.Printf("[DEBUG] Migrating IDs to correct casing for Log Analytics Workspace")
	name := is.Attributes["name"]
	resourceGroup := is.Attributes["resource_group"]
	subscriptionId := meta.(*clients.Client).Account.SubscriptionId
	id := parse.NewLogAnalyticsWorkspaceID(name, resourceGroup)
	is.ID = id.ID(subscriptionId)

	return is, nil
}
```

tombuildsstuff:
This works around an issue in the legacy state migration where the field
`resource_group` was used in place of `resource_group_name`. Whilst we
can't re-run the state migration from V0 -> V1 since those resources have
been upgraded - we can duplicate it for V1 -> V2, and patch it for new users
meaning this should work in both scenarios.

Fixes #9790


azurerm/internal/services/loganalytics/migration/workspace_v1_to_v2.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/loganalytics/parse"
+)
+
+func WorkspaceV1ToV2() schema.StateUpgrader {
+	// V1 to V2 is the same as v1 to v0 - to workaround a historical issue where `resource_group` was
```
jackofallops, line 12: super minor typo in comment
```suggestion
	// V1 to V2 is the same as v0 to v1 - to workaround a historical issue where `resource_group` was
```

jackofallops (approved):
Just one trivial observation, but otherwise LGTM 👍

tombuildsstuff:
Running a subset of the tests these look good:

Screenshot 2020-12-16 at 11 28 30

PR #9855, opened 2020-12-14
tooling: adding a gradually deprecated check for relying on the Azure Resource ID's

tombuildsstuff:
To workaround issues within the Azure API - Terraform now manages the Resource ID's for these resources.

Resource ID Formatters, Parsers and Validators are now generate-able - so these should be generated where possible (albeit hand-written where necessary) - and the Azure Resource ID should no longer be used

The existing Resource ID's will be converted over time, but this should mean new resources use these new Formatters/Parsers by default.


jackofallops (approved):
LGTM 👍

manicminer:
> LGTM 👍

I'll echo that 🥁

Issue #9714, opened 2020-12-07
Add property to document required providers when using azurerm_resource_group_template_deployment


sharebear:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When using `azurerm_resource_group_template_deployment` there is no way for terraform to know which providers are necessary to successfully apply the deployment. This means that when running terraform in a locked-down production environment with the `skip_provider_registration` feature enabled terraform has no way to identify which providers could be needed within the template, potentially resulting in an error message such as the following.

```
Error: waiting for creation of Template Deployment "eventgrid-deployment" (Resource Group "<redacted>"): Code="DeploymentFailed" Message="At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details." Details=[{"code":"Conflict","message":"{\r\n  \"error\": {\r\n    \"code\": \"MissingSubscriptionRegistration\",\r\n    \"message\": \"The subscription is not registered to use namespace 'Microsoft.EventGrid'. See https://aka.ms/rps-not-found for how to register subscriptions.\",\r\n    \"details\": [\r\n      {\r\n        \"code\": \"MissingSubscriptionRegistration\",\r\n        \"target\": \"Microsoft.EventGrid\",\r\n        \"message\": \"The subscription is not registered to use namespace 'Microsoft.EventGrid'. See https://aka.ms/rps-not-found for how to register subscriptions.\"\r\n      }\r\n    ]\r\n  }\r\n}"}]
```

Would it be an idea to add a `required_providers` list to the resource that the azurerm provider could use to identify dependent providers?

Alternatively could this be an explicit resource type?

### New or Affected Resource(s)

* azurerm_resource_group_template_deployment
* azurerm_subscription_template_deployment

### Potential Terraform Configuration

```hcl
resource "azurerm_resource_group_template_deployment" "main" {
  name                = "eventgrid-deployment"
  resource_group_name = var.resource_group_name
  deployment_mode     = "Incremental"
  required_providers  = ["Microsoft.EventGrid"]   # <--- NEW ARGUMENT HERE
[...]

PR #9864, opened 2020-12-15
Refactor: Migrate storage tests

ArcturusZhang:
Reopen of #9714


azurerm/internal/services/storage/storage_account_resource_test.go
```diff
@@ -36,864 +39,694 @@ func TestValidateArmStorageAccountName(t *testing.T) {
...
+				check.That(data.ResourceName).Key("account_tier").HasValue("Standard"),
+				check.That(data.ResourceName).Key("account_replication_type").HasValue("LRS"),
+				check.That(data.ResourceName).Key("tags.%").HasValue("1"),
+				check.That(data.ResourceName).Key("tags.environment").HasValue("production"),
+				resource.TestCheckResourceAttr(data.ResourceName, "primary_connection_string", ""),
+				resource.TestCheckResourceAttr(data.ResourceName, "secondary_connection_string", ""),
+				resource.TestCheckResourceAttr(data.ResourceName, "primary_blob_connection_string", ""),
+				resource.TestCheckResourceAttr(data.ResourceName, "secondary_blob_connection_string", ""),
+				resource.TestCheckResourceAttr(data.ResourceName, "primary_access_key", ""),
+				resource.TestCheckResourceAttr(data.ResourceName, "secondary_access_key", ""),
```
katbyte, line 125: COuld we use the check.That functions here?

azurerm/internal/services/storage/storage_account_resource_test.go
```diff
@@ -36,864 +39,694 @@ func TestValidateArmStorageAccountName(t *testing.T) {
...
-
-		resp, err := conn.GetProperties(ctx, resourceGroup, name, "")
-		if err != nil {
-			return nil
-		}
-
-		if resp.StatusCode != http.StatusNotFound {
-			return fmt.Errorf("Storage Account still exists:\n%#v", resp.AccountProperties)
-		}
+func (r StorageAccountResource) Destroy(ctx context.Context, client *clients.Client, state *terraform.InstanceState) (*bool, error) {
```
katbyte, line 718: I don't think we need destroy functions? the new test shim framework uses the exists function to check that

katbyte (changes requested):
Thanks @ArcturusZhang - just a couple comments inline. IU don't think we need to use the destroy function anymore?

azurerm/internal/services/storage/storage_account_resource_test.go
ArcturusZhang: Oops, I missed this one

azurerm/internal/services/storage/storage_account_resource_test.go
ArcturusZhang: I implement this function because the storage_account resource has a `TestAccStorageAccount_disappears` test which requires a Destroy function

ArcturusZhang:
Hi @katbyte I have resolved the comments, please have a look, thanks

katbyte (approved):
Thanks @ArcturusZhang - this LGMT 👍

PR #9866, opened 2020-12-15
Update tc for Virtual Hub IP and Virtual Hub Bgp Connection

neil-yechenwei:
After confirmed with service team, currently it requires one GatewaySubnet and one default subnet under VNet while creating Virtual Hub IP. I assume it's a test issue. So I updated test cases for this problem. Now all related test cases are passed.

--- PASS: TestAccAzureRMVirtualHubIP_basic (1502.76s)
--- PASS: TestAccAzureRMVirtualHubIP_complete (1507.61s)
--- PASS: TestAccAzureRMVirtualHubIP_requiresImport (1518.83s)
--- PASS: TestAccAzureRMVirtualHubIP_update (1555.54s)

--- PASS: TestAccAzureRMVirtualHubBgpConnection_basic (1579.16s)
--- PASS: TestAccAzureRMVirtualHubBgpConnection_requiresImport (1579.16s)


jackofallops (changes requested):
Hi @neil-yechenwei
Thanks for this PR.  Since that subnet is a mandatory requirement, and is a special case Subnet (there can be only one) I think we should validate in code that it's there before attempting to create this resource. WDYT?

neil-yechenwei:
@jackofallops , Maybe I cannot do this since service team says having GatewaySubnet created in the Vnet requirement would be removed in near future.

jackofallops:
> @jackofallops , Maybe I cannot do this since service team says having GatewaySubnet created in the Vnet requirement would be removed in near future.

Hi @neil-yechenwei - Can you get an estimate on when that is likely to be deployed so we can make a decision on what to do with the service?

neil-yechenwei:
March 2021.

neil-yechenwei:
@jackofallops , After tested, seems service team has removed the requirement involving GatewaySubnet created in the Vne since all related test cases are passed now. So I removed "t.skip()" to enabled the test cases. Please have an another look. Thanks.

katbyte (approved):
Thanks @neil-yechenwei - tests are passing

katbyte:
good to merge once the conflicts are sorted @neil-yechenwei

neil-yechenwei:
@katbyte , sorry. I am on holiday a few days ago. I resolved conflict. Please have an another look. Thanks.

PR #9878, opened 2020-12-15
Upgrading to Go 1.15.5

tombuildsstuff:


manicminer (commented):
LGTM, does the version also need updating in go.mod?

tombuildsstuff:
@manicminer nope, that's updated as/when dependencies change - so not a big deal

manicminer (approved):

PR #9881, opened 2020-12-15
Doc Fix: app_service_managed_certificate.html.markdown

DavidChristiansen:
- Corrected code sample (reference to test instead of correct value - example)
- Inclusion of missing `azurerm_dns_zone`  data source


mbfrahry (approved):
LGTM! Thanks for this @DavidChristiansen

PR #9884, opened 2020-12-16
Increase limit from 1 to 10 for `application_security_group` in `network_security_rule`

mybayern1974:
The Azure service team has increased the number of allowed ASGs in an NSR from 1 to 10. I've tested it locally and it worked: the service could accept up to 10 and reported errors when accepting 10+.

Below is the corresponding test case pass:
```
=== RUN   TestAccAzureRMNetworkSecurityRule_applicationSecurityGroups
=== PAUSE TestAccAzureRMNetworkSecurityRule_applicationSecurityGroups
=== CONT  TestAccAzureRMNetworkSecurityRule_applicationSecurityGroups
--- PASS: TestAccAzureRMNetworkSecurityRule_applicationSecurityGroups (173.47s)
PASS
```


katbyte (approved):
Thanks, LGTM 👍

PR #9887, opened 2020-12-16
Acceptance tests: fix CheckDestroyedFunc

manicminer:
- state.RootModule().Resources uses full notation for key values
- Using Exists() for destroy check should fail when err == nil


tombuildsstuff (approved):
LGTM 👍

PR #9920, opened 2020-12-17
web/app-service: fixing the Exists function

tombuildsstuff:
The App Service API has "404" as an "ok" status code.. so this doesn't return an error. This PR fixing the Exists function used in the tests to account for this - these provision fine otherwise.


katbyte (approved):
LGTM

PR #9929, opened 2020-12-18
refactor apimanagement: generators (part 1)

favoretti:


azurerm/internal/services/apimanagement/api_management_product_api_resource.go
```diff
@@ -43,13 +45,15 @@ func resourceArmApiManagementProductApi() *schema.Resource {
...
 	client := meta.(*clients.Client).ApiManagement.ProductApisClient
+	subscriptionId := meta.(*clients.Client).Account.SubscriptionId
 	ctx, cancel := timeouts.ForCreate(meta.(*clients.Client).StopContext, d)
 	defer cancel()
 
 	resourceGroup := d.Get("resource_group_name").(string)
 	serviceName := d.Get("api_management_name").(string)
 	apiName := d.Get("api_name").(string)
 	productId := d.Get("product_id").(string)
+	id := parse.NewProductApiID(subscriptionId, resourceGroup, serviceName, productId, apiName)
```
katbyte, line 56: I don't think we can do this in the create function at this time, as if there is a change to the case (correcting it) it would change the ID and require a state migration function?

favoretti: What do you want me to do, remove this from create or add a state migration?

favoretti: OTOH why would it matter on create? I set the resource ID below from the generator, so kinda ignoring what API returns back..

katbyte: If another resource is depending on the ID being a certain/existing casing this could break resources that consume it. For now we are updating read/delete/update (but not resource ID) and tests to use the new parsers. Once every resource everywhere has been updated we can then move onto having the create functions use them.

favoretti: Resolved.

katbyte (changes requested):
Thanks @favoretti - looks good but i think there may be an issue with one of the parsers:

```
------- Stdout: -------
=== RUN   TestAccApiManagementSubscription_basic
=== PAUSE TestAccApiManagementSubscription_basic
=== CONT  TestAccApiManagementSubscription_basic
        
        When applying changes to azurerm_api_management_subscription.test, provider
        "azurerm" produced an unexpected new value for was present, but now absent.
        
        issue tracker.
        
--- FAIL: TestAccApiManagementSubscription_basic (2896.99s)
FAIL
```

favoretti:
> Thanks @favoretti - looks good but i think there may be an issue with one of the parsers:
> 
> ```
> ------- Stdout: -------
> === RUN   TestAccApiManagementSubscription_basic
> === PAUSE TestAccApiManagementSubscription_basic
> === CONT  TestAccApiManagementSubscription_basic
>     testing.go:684: Step 0 error: errors during apply:
>         
>         Error: Provider produced inconsistent result after apply
>         
>         When applying changes to azurerm_api_management_subscription.test, provider
>         "azurerm" produced an unexpected new value for was present, but now absent.
>         
>         This is a bug in the provider, which should be reported in the provider's own
>         issue tracker.
>         
> --- FAIL: TestAccApiManagementSubscription_basic (2896.99s)
> FAIL
> ```

@katbyte I'll try to catch what's going on tomorrow.

favoretti:
@katbyte Fixed.

```
$ TF_ACC=1 go test -v ./azurerm/internal/services/apimanagement -timeout=1000m -run TestAccApiManagementSubscription_basic
=== RUN   TestAccApiManagementSubscription_basic
=== PAUSE TestAccApiManagementSubscription_basic
=== CONT  TestAccApiManagementSubscription_basic
--- PASS: TestAccApiManagementSubscription_basic (2717.17s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/apimanagement	2718.325s
```

katbyte (approved):
LGTM 👍

PR #9931, opened 2020-12-18
Fix LAW capitalization in security_center_workspace (again)

favoretti:


azurerm/internal/services/securitycenter/security_center_workspace_resource.go
```diff
@@ -187,7 +187,13 @@ func resourceSecurityCenterWorkspaceRead(d *schema.ResourceData, meta interface{
 
 	if properties := resp.WorkspaceSettingProperties; properties != nil {
 		d.Set("scope", properties.Scope)
-		d.Set("workspace_id", properties.WorkspaceID)
+		workspaceId := ""
+		id, err := parse.LogAnalyticsWorkspaceID(*properties.WorkspaceID)
```
katbyte, line 191: We should probably check properties.WorkspaceID is not nil

katbyte (changes requested):
Aside from one comment this looks good! thanks @favoretti

katbyte (approved):
LGTM @favoretti 👍

PR #9932, opened 2020-12-18
`azurerm_log_analytics_linked_service` - Add validation for resource ID type

WodansSon:
Adding validation for **Resource ID type** based on **field**:

Field Name | Required Resource ID type
-- | --
resource_id | Microsoft.Automation/automationAccounts
read_access_id | Microsoft.Automation/automationAccounts
write_access_id | Microsoft.OperationalInsights/clusters


azurerm/internal/services/loganalytics/log_analytics_linked_service_resource.go
```diff
@@ -167,6 +167,31 @@ func resourceArmLogAnalyticsLinkedService() *schema.Resource {
...
+				if readAccessID := d.Get("read_access_id").(string); readAccessID != "" {
+					if !validate.IsAutomationAccountID(readAccessID) {
+						return fmt.Errorf("'read_access_id' must be an Automation Account resource ID, got %q", readAccessID)
+					}
+				}
+			}
+
+			if d.HasChange("write_access_id") {
+				if writeAccessID := d.Get("write_access_id").(string); writeAccessID != "" {
+					if !validate.IsLogAnalyticsClusterID(writeAccessID) {
```
favoretti, line 189: There's a generated validator for this one in loganalytics validate, maybe use that one instead?

WodansSon: Fair point... I'll update my logic to use the generated validator rather than my own.

katbyte (approved):
LGTM 👍

WodansSon:
Hit Cluster quota limit for the subscription, all other tests passed:
 
image

azurerm/internal/services/loganalytics/validate/log_analytics_linked_service_automation_id.go
```diff
@@ -0,0 +1,17 @@
...
+	id, err := azure.ParseAzureResourceID(input)
+	if err != nil {
+		return false
+	}
+
+	if _, err = id.PopSegment("automationAccounts"); err != nil {
+		return false
+	}
+
+	return true
```
tombuildsstuff, line 16: @WodansSon rather than including an ID from outside of it's package, can we generate the "AutomationAccountId" in the automation package and reference that's validation function here, rather than adding this in? Else we'll end up with duplicates all over the place

WodansSon: I'll open another PR to fix that bit...  thanks...

PR #9950, opened 2020-12-19
`azurerm_log_analytics_linked_service` - Update validation to use generated validate functions

WodansSon:
Addressing comment by @tombuildsstuff on PR #9932: 

> @WodansSon rather than including an ID from outside of it's package, can we generate the "AutomationAccountId" in the automation package and reference that's validation function here, rather than adding this in? Else we'll end up with duplicates all over the place

* Added generated code to Automation resource (parse, validate)
* Update logic in log analytics linked service to use generated validation code
* Removed manually written validation code from log analytics linked service


WodansSon:
image

katbyte (approved):
LGTM 👍

PR #9957, opened 2020-12-21
"azurerm_spring_cloud_app" - supports property `https_only`, `is_public`, `persistent_disk`

njuCZ:
spring cloud app supports more properties: `https_only`, `is_public`, `persistent_disk`

`https_only` and `persistent_disk` could only be set by update, creat rest api will ignore those values

image


katbyte (approved):
Thanks @njuCZ - this LGTM 👍

PR #9967, opened 2020-12-21
#9151: Change loganalytics resourceId format to match api response fo…

### LINKED ISSUE #9151: azurerm_kubernetes_cluster incorrect diff due to casing on addon_profile.oms_agent.log_analytics_workspace_id
The following update on property addon_profile.oms_agent.log_analytics_workspace_id for a azurerm_kubernetes_cluster resource is unneccessary because only the casing is different:

```
~ log_analytics_workspace_id = "/subscriptions/b55149b2-2597-4455-b224-5159c38e1a2f/resourceGroups/abc-shared/providers/Microsoft.OperationalInsights/workspaces/abc-workspace-f17310012a04c444" -> "/subscriptions/b55149b2-2597-4455-b224-5159c38e1a2f/resourcegroups/abc-shared/providers/microsoft.operationalinsights/workspaces/abc-workspace-f17310012a04c444"
```

Using terraform 0.13.5 and azurerm provider 2.34.0.

blaw2422:
Fixes #9151


blaw2422:
I'm not sure why I have so many files with a single space but I'll do whatever necessary to remove those from the diff.  Not sure if its a configuration issue or what.  If its better for me to manually remove those from the diff I can.

blaw2422:
It appears this commit was recently introduced, and there was either a difference in what an api returns, or a wrong assumption.  Either way - i believe this is the issue.
https://github.com/terraform-providers/terraform-provider-azurerm/pull/9764/files#diff-dae653c88d478c57db3808e62fa794a966046932980dc67fc48f04a4a5b67942R15

favoretti:
Breaking change? I just switched log analytics to use generated parsers. What broke?

blaw2422:
> Breaking change? I just switched log analytics to use generated parsers. What broke?

It looks like this issue wants to replace resources because the casing is different in the resource id: https://github.com/terraform-providers/terraform-provider-azurerm/issues/9151

I believe the 2nd link i posted is where the resource id changes its casing

favoretti:
Your proposed fix changes generated files, so this won't work unfortunately. I'll try to have a look at oms agent addon and monitor diagnostic setting tomorrow my morning. But casing in code is correct. API gives inconsistent results, hence the generators.

blaw2422:
> Your proposed fix changes generated files, so this won't work unfortunately. I'll try to have a look at oms agent addon and monitor diagnostic setting tomorrow my morning. But casing in code is correct. API gives inconsistent results, hence the generators.

Here is the file i changed, others were changed based on some of the `make` commands.
https://github.com/terraform-providers/terraform-provider-azurerm/pull/9967/files#diff-b791bc580b92d22c33a27eb19811a15c8296b5f0a1011803026f805be31b4017

favoretti:
Ah, sorry, missed that. Looking from the phone... well, if you look at SDK, the casing I initially put in was correct, it's the API that's giving incorrect results. We just need to parse those IDs wherever they are used I think. @katbyte any hints here?

WodansSon:
@blaw2422 @favoretti - Thanks for this PR, but instead of changing the casing in all of these source files would it be better to implement a function to compare the resource id in a case insensitive way? Kind of like what we did in [Frontdoor](https://github.com/terraform-providers/terraform-provider-azurerm/blob/6d5c6c8285bc6176ec34c36b78f9615eee1560b3/azurerm/internal/services/frontdoor/parse/front_door.go#L77) when the casing switched between different versions of the API?

WodansSon (changes requested):
@blaw2422 I am blocking the PR to avoid this PR from accidentally being merged.

favoretti:
@WodansSon I thought that was the point of moving it all to generators, to not to care much about the casing of API. LogAnalytics itself works just fine, there are places where that ID is passed around as ID and it needs to be wrapped into LA's parser to conform with whatever we store in the state. I'll check the use-case of @blaw2422 to see which part needs to be fixed.

favoretti:
@WodansSon here's what I propose: #9976

blaw2422:
> @WodansSon here's what I propose: #9976

I'll test this branch and report back

blaw2422:
closing this PR as #9976 fixes the problem

PR #9985, opened 2020-12-22
Minor change in servicebus_subscription.html.markdown

shantanu-parate:


katbyte (approved):
Thanks @shantanu-parate - LGTM 🚀

PR #10005, opened 2020-12-26
`azurerm_traffic_manager_profile` - add traffic view support

notchairmk:
adds `traffic_view_enabled` to both traffic manager profile resource and data source.


katbyte (approved):
Thanks for this @notchairmk - LGTM 👍

Issue #9988, opened 2020-12-23
Azure Key Vault now requires Soft Deleted to be Enabled


ruandersMSFT:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

AzureRM v2.41.0

### Affected Resource(s)

* azurerm_key_vault

### References

Creating an Azure Key Vault with Soft Delete disabled is no longer permitted.

[Dec 15 2020: Soft-delete will be enabled on all key vaults](https://docs.microsoft.com/en-us/azure/key-vault/general/soft-delete-change)

* #0000

PR #10007, opened 2020-12-28
azurerm_key_vault - soft_delete_enabled now required

ruandersMSFT:
#9988 

Azure Key Vault now requires Soft Delete to be enabled, and is being retroactively enabled on all existing key vault resources where it was previously disabled.


tombuildsstuff:
hey @ruandersMSFT

I've spent some time digging into the Key Vault Soft Delete changes as a part of some broader work here, and whilst defaulting the value of the field `soft_delete_enabled` to `true` does partially fix this - through testing this I discovered that there's more work necessary for this migration path to succeed. In particular that requires:

* Defaulting the value of the `soft_delete_retention_days` field to `90` if its unset - this is awkward since the API only allows this to be configured a single time, but doesn't return this field unless it's explicitly set. As such this requires a State Migration to implement since this requires updating existing values in the state to match this default, since this was previously set to `0` to match the API response (which when omitted, set `0`, the default value for an integer).
* Changing the migration path so that it's possible to omit this field from the configuration, so that this field is non-functional. Whilst this PR went some way to fixing this, ultimately this since this can no longer be configured (in any Azure Environment) - in practice there's not much point having this be user-configurable, since there's only a single value. After testing the migration path it appeared that making this field do nothing seemed like the path of least hassle for users, since this'll show up a diff which can be resolved - rather than raising an error here.

As such whilst I'd like to thank you for this contribution, I'm going to close this PR in favour of #XXX which makes these changes in a slightly different manner, allowing Terraform to handle this behavioural change in today's release.

Thanks!

PR #10028, opened 2020-12-31
Adding description to ADF managed IR docs

allantargino:
The field `description` of `azurerm_data_factory_integration_runtime_managed` is already implemented but was missing on the resource docs:

https://github.com/terraform-providers/terraform-provider-azurerm/blob/50bf34b6e5e2cdb72727cce32c92ed6e0afa6791/azurerm/internal/services/datafactory/data_factory_integration_runtime_managed_resource.go#L47-L50


katbyte (approved):
Thanks for this @allantargino! LGTM 👍

PR #10030, opened 2021-01-01
Server vulnerability assessment resource

martenbohlin:
Adds a new resource azurerm_server_vulnerability_assessment that installes the vulnerability assessment on a VM


azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
```diff
@@ -0,0 +1,167 @@
...
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(5 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"virtual_machine_id": {
```
neil-yechenwei, line 36: Seems ServerVulnerabilityAssessment supports servers, which can be of type “Microsoft.Compute/virtualMachines”, But also “[Microsoft.HybridCompute/machines](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/hybridcompute/resource-manager/Microsoft.HybridCompute/stable/2020-08-02/HybridCompute.json)”.

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
```diff
@@ -0,0 +1,167 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
```
neil-yechenwei, line 39: Seems ServerVulnerabilityAssessment can be updated from server A to server B. So ForceNew isn't required. Could you double check whether it could be updated in-place?

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
martenbohlin: I can not find any information on how to create or work with Microsoft.HybridCompute resources from terraform. @neil-yechenwei do you have any information on how to work with HybridCompute in terraform? 
I will need some help in order to know how to setup something where I could test and verify HybridCompute support.

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
martenbohlin: I will take a look at this.

website/docs/r/azurerm_server_vulnerability_assessment.markdown
```diff
@@ -0,0 +1,112 @@
...
+  resource_group_name = azurerm_resource_group.example.name
+
+  ip_configuration {
+    name                          = "vm-example"
+    subnet_id                     = azurerm_subnet.example.id
+    private_ip_address_allocation = "Dynamic"
+  }
+}
+
+resource "azurerm_virtual_machine" "example" {
```
sharebear, line 47: The documentation for `azurerm_virtual_machine` states

> The azurerm_virtual_machine resource has been superseded by the azurerm_linux_virtual_machine and azurerm_windows_virtual_machine resources.

Should examples in documentation be updated to one of these?

sharebear:
Fun, I was just looking at the API for this yesterday to see how much work it would be to implement :)

This is getting towards a functional question rather than a PR comment, but have I understood correctly that you need to have activated Azure Defender for Servers enabled on the Subscription in order to use Qualys;

1. Should there be a note about this in the documentation?
2. How does the resource behave if the Subscription does _not_ have Azure Defender enabled?
3. Do the acceptance tests run in a Subscription that has Azure Defender enabled?

martenbohlin:
> This is getting towards a functional question rather than a PR comment, but have I understood correctly that you need to have activated Azure Defender for Servers enabled on the Subscription in order to use Qualys;
> 
> 1. Should there be a note about this in the documentation?
> 2. How does the resource behave if the Subscription does _not_ have Azure Defender enabled?
> 3. Do the acceptance tests run in a Subscription that has Azure Defender enabled?

Good  questions. It would be interesting to know what would happen if Azure Defender is disabled. If I could get help to test that it would be great since the company I work for require defender to be enabled on all subscriptions. (Should be enough to run the test TestAccServerVulnerabilityAssessment_basic)

website/docs/r/azurerm_server_vulnerability_assessment.markdown
martenbohlin: Thanks, I have updated the documentation

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
martenbohlin: @neil-yechenwei I can not see how I would do a in-place update with the API.The API has the method **CreateOrUpdate**, but that method takes the parameters resourceGroupName, resourceNamespace, resourceType, and resourceName to identify the **server** to create or update the vulnerability assessment on. So I do not think it is possible to specify a specific vulnerability assessment and move that. Instead you need to delete the old and install a new.
Do you agree, or am I missing something?

neil-yechenwei: Yes. I assume you're right.

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
neil-yechenwei: Terraform doesn't support Microsoft.HybridCompute/machines for now. I assume you can create/test this resource with [rest api](https://docs.microsoft.com/en-us/rest/api/hybridcompute/machines/createorupdate). I assume the resource `azurerm_server_vulnerability_assessment` could be made more generic to support both of them if ServerVulnerabilityAssessment API also supports Microsoft.HybridCompute/machines?

martenbohlin: I have added support for Microsoft.HybridCompute/machines (Azure ARC). But I have kept the name virtual_machine_id for the property since that seems to be most consistent with other resources. Or does anybody have any suggestions on a better name?

martenbohlin:
> This is getting towards a functional question rather than a PR comment, but have I understood correctly that you need to have activated Azure Defender for Servers enabled on the Subscription in order to use Qualys;
> 
> 1. Should there be a note about this in the documentation?
> 2. How does the resource behave if the Subscription does _not_ have Azure Defender enabled?
> 3. Do the acceptance tests run in a Subscription that has Azure Defender enabled?

@sharebear 
I was able to test this my self. So here are the answers:
1. I added a note in the documentation that Azure Defender must be enabled.
2. You get the following error message back from the API that is show:`Subscription is not in the Standard or Standard Trial subscription plan. Please upgrade to use this feature.`
3. I am guessing here, but there is tests for the resource azurerm_mssql_server_vulnerability_assessment so I belive Azure Defender is enabled where the integration tests run.

sharebear:
> @sharebear
> I was able to test this my self. So here are the answers:
> 
> 1. I added a note in the documentation that Azure Defender must be enabled.
> 2. You get the following error message back from the API that is show:`Subscription is not in the Standard or Standard Trial subscription plan. Please upgrade to use this feature.`
> 3. I am guessing here, but there is tests for the resource azurerm_mssql_server_vulnerability_assessment so I belive Azure Defender is enabled where the integration tests run.

Awesome, I think that's the best possible answer for these questions.

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
```diff
@@ -0,0 +1,176 @@
...
+			return nil
+		} else {
+			return fmt.Errorf("removing Server Vulnerability Assessment for %q/%q: %+v\n", resourceGroup, name, err)
+		}
+	} else {
+		return nil
+	}
+}
+
+func parseVirtualMacineId(stringId string) (string, string, string, string, *azure.ResourceID, error) {
```
sharebear, line 155: Typo here `parseVirtualMachineId` however.... There already exists a parser for virtualMachines `azurerm/internal/services/compute/parse/virtual_machine.go:41` you should reuse that. This however does not support Arc.

If you want to support Arc then I would look into using the generator for resource id parsers/validators `azurerm/internal/tools/generator-resource-id/README.md`. Then I would split the input parameters into conflicting `virtual_machine_id` and `hybrid_machine_id` allowing you to use the generated validators for each id type. Although _personally_ I would probably drop the Arc support from this PR to keep it small, replacing this parser with the existing parser. Then if Arc support is desired, implement that as a new PR on top of these changes.

martenbohlin: @neil-yechenwei is Arc support something you need, or can we wait with that?

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
```diff
@@ -0,0 +1,176 @@
...
+		response, err := client.Get(ctx, resourceGroup, provider, resourceType, name)
+		if err != nil || response.ProvisioningState != security.ProvisioningState1Provisioning {
+			break
+		}
+		time.Sleep(time.Duration(retry) * time.Second)
+	}
+
+	response, err := client.Delete(ctx, resourceGroup, provider, resourceType, name)
+	if err != nil {
+		if utils.ResponseWasStatusCode(response, 202) {
```
sharebear, line 145: I would file this as a bug against the upstream SDK. If the API gives a 202 on "success" then it should probably be added to the list of accepted codes in DeleteResponder.

martenbohlin: [Bug added to the SDK](https://github.com/Azure/azure-sdk-for-go/issues/14120)

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
martenbohlin: @sharebear I would appreciate if you took a look at my changes in commit 2ffd690 and let me know if you think they look good.

sharebear: It's probably best to wait and see what the maintainers say, and I'm still very new to golang and writing providers. That said, I have a gut feeling that the implementation feels a little complicated compared to other resources I've looked at.... what if read looked like the following

```
targetId = ""
virtualMachineId, err := computeParse.VirtualMachineID(d.Id())
if virtualMachineId != nil {
	targetId = virtualMachineId.ID()
	d.Set("virtual_machine_id", targetId)
}

if targetId == "" {
	hybridMachineId, err := computeParse.HybridMachineID(d.Id())
	if hybridMachineId != nil {
		targetId = hybridMachineId.ID()
		d.Set("hybrid_machine_id", targetId)
	}
}

// TODO: Return err if targetId still emtpy

// do client.Get
```

I think create would be even easier to simplify, then you could drop the whole TargetId struct.... disclaimer, I'm shooting from the hip here, no idea if this would work in practice.

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
neil-yechenwei: How about `target_machine_id`?

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
```diff
@@ -0,0 +1,233 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"virtual_machine_id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: computeValidate.VirtualMachineID,
+			},
+			"hybrid_machine_id": {
```
neil-yechenwei, line 49: Is it possible to merge "virtual_machine_id" and "hybrid_machine_id" to one property "target_machine_id"?

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
```diff
@@ -0,0 +1,233 @@
...
+	SubscriptionId string
+	ResourceGroup  string
+	Provider       string
+	ResourceType   string
+	Name           string
+}
+
+func parseTargetId(d *schema.ResourceData) (*TargetId, error) {
+	rawVirtualMachineId := d.Get("virtual_machine_id").(string)
+	rawHybridMachineId := d.Get("hybrid_machine_id").(string)
```
neil-yechenwei, line 177: Does it make sense to parse it by judging whether the "target_machine_id" exists the resource id segment "Microsoft.HybridCompute/machines" after merged these two properties to one property "target_machine_id"?

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+
+Adds Azure Server Vulnerability Assessment (Qualys) to a VM.
+
+-> **NOTE** Azure Defender has to be enabled on the subscription in order for this resource to work.
+
+## Example Usage
+
+```hcl
+resource "azurerm_resource_group" "example" {
+  name     = "example-resource-group"
```
neil-yechenwei, line 19: I assume here should be "example-resources".

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+  location = "West US"
+}
+
+resource "azurerm_virtual_network" "example" {
+  name                = "net-example"
```
neil-yechenwei, line 24: I assume here should be "example-vnet".

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+  resource_group_name = azurerm_resource_group.example.name
+  address_space       = ["192.168.1.0/24"]
+  location            = azurerm_resource_group.example.location
+}
+
+resource "azurerm_subnet" "example" {
+  name                 = "snet-example"
```
neil-yechenwei, line 31: I assume here should be "example-subnet".

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+  resource_group_name  = azurerm_resource_group.example.name
+  virtual_network_name = azurerm_virtual_network.example.name
+  address_prefix       = "192.168.1.0/24"
+}
+
+resource "azurerm_network_interface" "example" {
+  name                = "vm-example"
```
neil-yechenwei, line 38: I assume here should be "example-nic".

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+
+  ip_configuration {
+    name                          = "vm-example"
+    subnet_id                     = azurerm_subnet.example.id
+    private_ip_address_allocation = "Dynamic"
+  }
+}
+
+resource "azurerm_linux_virtual_machine" "example" {
+  name                = "vm-example"
```
neil-yechenwei, line 50: I assume here should be "example-vm".

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+* `read` - (Defaults to 5 minutes) Used when retrieving the Advanced Threat Protection.
+* `delete` - (Defaults to 30 minutes) Used when deleting the Advanced Threat Protection.
+
+## Import
+
+Server Vulnerability Assessment can be imported using the `resource id`, e.g.
+
+```shell
+terraform import azurerm_server_vulnerability_assessment.example /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resource-group-name/providers/Microsoft.Compute/virtualMachines/vm-name/providers/Microsoft.Security/serverVulnerabilityAssessments/Default
+```
```
neil-yechenwei, line 111: Suggest to also add an example of Microsoft.HybridMachine.

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+
+The `timeouts` block allows you to specify [timeouts](https://www.terraform.io/docs/configuration/resources.html#timeouts) for certain actions:
+
+* `create` - (Defaults to 30 minutes) Used when creating the Advanced Threat Protection.
+* `read` - (Defaults to 5 minutes) Used when retrieving the Advanced Threat Protection.
+* `delete` - (Defaults to 30 minutes) Used when deleting the Advanced Threat Protection.
+
+## Import
+
+Server Vulnerability Assessment can be imported using the `resource id`, e.g.
```
neil-yechenwei, line 107: I assume it should be ”Server Vulnerability Assessments“

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+Adds Azure Server Vulnerability Assessment (Qualys) to a VM.
+
+-> **NOTE** Azure Defender has to be enabled on the subscription in order for this resource to work.
+
+## Example Usage
+
+```hcl
+resource "azurerm_resource_group" "example" {
+  name     = "example-resource-group"
+  location = "West US"
```
neil-yechenwei, line 20: I assume here should be "West Europe"

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+  ip_configuration {
+    name                          = "vm-%[1]d"
+    subnet_id                     = azurerm_subnet.test.id
+    private_ip_address_allocation = "Dynamic"
+  }
+}
+
+resource "azurerm_linux_virtual_machine" "test" {
+  name                = "vm-%[1]d"
```
neil-yechenwei, line 91: I assume here should be "acctestVM-%[1]d"

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+resource "azurerm_subnet" "test" {
+  name                 = "snet-%[1]d"
+  resource_group_name  = azurerm_resource_group.test.name
+  virtual_network_name = azurerm_virtual_network.test.name
+  address_prefixes     = ["192.168.1.0/24"]
+}
+
+resource "azurerm_network_interface" "test" {
+  name                = "vm-%[1]d"
```
neil-yechenwei, line 79: I assume here should be "acctestNIC-%[1]d"

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+resource "azurerm_virtual_network" "test" {
+  name                = "net-%[1]d"
+  resource_group_name = azurerm_resource_group.test.name
+  address_space       = ["192.168.1.0/24"]
+  location            = azurerm_resource_group.test.location
+}
+
+resource "azurerm_subnet" "test" {
+  name                 = "snet-%[1]d"
```
neil-yechenwei, line 72: I assume here should be "acctestSubnet-%[1]d"

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
```diff
@@ -0,0 +1,119 @@
...
+  features {}
+}
+
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-sc-%[1]d"
+  location = "%[2]s"
+}
+
+resource "azurerm_virtual_network" "test" {
+  name                = "net-%[1]d"
```
neil-yechenwei, line 65: I assume here should be "acctestVNet-%[1]d"

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (ServerVulnerabilityAssessmentResource) basicCfg(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
+
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-sc-%[1]d"
```
neil-yechenwei, line 60: I assume here should be "acctestRG-sva-%[1]d"

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	resp, err := clients.SecurityCenter.ServerVulnerabilityAssessmentClient.Get(ctx, resourceGroup, "Microsoft.Compute", "virtualMachines", name)
+	if err != nil {
+		return nil, fmt.Errorf("reading Server Vulnerability Assessment (%s/%s): %+v", resourceGroup, name, err)
+	}
+
+	return utils.Bool(resp.ID != nil), nil
+}
+
+func (ServerVulnerabilityAssessmentResource) basicCfg(data acceptance.TestData) string {
```
neil-yechenwei, line 53: I assume the function name should be "template" not "basicCfg"

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
```diff
@@ -0,0 +1,119 @@
...
+			Config: r.basicCfg(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+			),
+		},
+		data.ImportStep(),
+	})
+}
+
+func (ServerVulnerabilityAssessmentResource) Exists(ctx context.Context, clients *clients.Client, state *terraform.InstanceState) (*bool, error) {
```
neil-yechenwei, line 36: Seems this function also should be compatible with Hybrid Machine, right?

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance/check"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+type ServerVulnerabilityAssessmentResource struct {
```
neil-yechenwei, line 18: I assume it's necessary to add a test case for requiresImport, right?

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
+---
+subcategory: "Security Center"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_server_vulnerability_assessment"
+description: |-
+  Adds Azure Vulnerability Assessment (Qualys) to a VM or Azure ARC server.
```
neil-yechenwei, line 6: It could be updated to "Manages an Azure Vulnerability Assessment (Qualys) to a VM or Azure ARC server" to align the style with other existing resources. Does it make sense?

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+---
+
+# azurerm_server_vulnerability_assessment
+
+Adds Azure Server Vulnerability Assessment (Qualys) to a VM.
```
neil-yechenwei, line 11: It could be updated to "Manages an Azure Server Vulnerability Assessment (Qualys) to a VM." to align the style with other existing resources. Does it make sense?

neil-yechenwei:
@martenbohlin , Thanks for the update. I have left a couple of comments, please have a look, thanks

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
martenbohlin: Agree, my problem is that the API needs a couple specific parts of the id and not the entire id. Will keep like this and see what the maintainers say as you suggested.

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
martenbohlin: Yes I had it like that fist. But then got the suggestion to split into separate properties. Do you have any strong arguments to just have one property?
I like the name virtual_machine_id since that is the most commonly used when you reference a VM.

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
martenbohlin: Yes, depending on if we merge the properties or not.

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
martenbohlin: I do not think it is necessary since it is just to copy paste whatever ID your resource has.

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
martenbohlin: Any reason you think we should use Europe instead of US?

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
martenbohlin: I have no way to automatically create a Hybrid Machine. I have only been able to verify that manually.

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
martenbohlin: On line 32 I have data.ImportStep() that will verify that import works as part of the TestAccServerVulnerabilityAssessment_basic test case.

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
neil-yechenwei: I saw an example that it lists all possible samples. It's fine to keep it for now.

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
neil-yechenwei: I assume it should be the the code style of terraform. You can reference other [resources](https://raw.githubusercontent.com/terraform-providers/terraform-provider-azurerm/master/website/docs/r/stack_hci_cluster.html.markdown) as sample.

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
neil-yechenwei: Maybe you can manually create it first via [azure rest api](https://docs.microsoft.com/en-us/rest/api/hybridcompute/machines/createorupdate) and configure the resource id of this resource in test case. Then you can verify it in test case to complete the Exists function. Does it make sense?

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
neil-yechenwei: I assume it's not enough. Suggest to add "requiresImport" tc like other [resources](https://github.com/terraform-providers/terraform-provider-azurerm/blob/59066bd464d5f5cce8eddc3e4d5e4f2fb0842bb4/azurerm/internal/services/azurestackhci/stack_hci_cluster_resource_test.go#L34).

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
martenbohlin: That will be a bit to complicated for me. But if you want to give it a go I think we can include it.

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
martenbohlin: There is a wide range of locations are used in the documentation, West US is used in 123 other files. So keeping it for now.

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
martenbohlin: That example imports on something else then the ID full id, so I think that is different since this only supports importing on the ID.

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource_test.go
martenbohlin: I will tak a look at this

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
katbyte: this is fine, thou fwiw we do typically default to west europe

azurerm/internal/services/securitycenter/server_vulnerability_assessment_resource.go
katbyte: I think its better to have it split as it allows for better validation and becomes more obvious, alternatively it might make sense to split this into two resources..

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
katbyte: I'd like us to show both please

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
+---
+subcategory: "Security Center"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_server_vulnerability_assessment"
```
katbyte, line 4: as this is in the security center service could we rename this to
```suggestion
page_title: "Azure Resource Manager: azurerm_security_centre_server_vulnerability_assessment"
```

website/docs/r/azurerm_server_vulnerability_assessment.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+description: |-
+  Manages an Azure Vulnerability Assessment (Qualys) to a VM or Azure ARC server.
+---
+
+# azurerm_server_vulnerability_assessment
+
+Manages an Azure Server Vulnerability Assessment (Qualys) to a VM.
+
+-> **NOTE** Azure Defender has to be enabled on the subscription in order for this resource to work.
```
katbyte, line 13: could we include a link here to the docs on how to get started with this?

katbyte (changes requested):
Thanks @martenbohlin - i've given this a review and overall while its good i've let some comments inline to address before merge

martenbohlin:
Thanks for the feedback @katbyte ! I think all of your suggestions were great and have implemented them.

manicminer:
@martenbohlin I've tested this locally and it looks good. Could you resolve the merge conflict and I will retest? Thanks!

manicminer:
My bad, there wasn't a conflict after all, but thanks for merging in master.

manicminer (approved):
Tests are passing, this looks good to merge

Screenshot 2021-03-09 at 20 13 54

Issue #10029, opened 2020-12-31
azurerm_application_gateway identity_ids are case-sensitive, and shouldn't be


JohnDelisle:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

```
2020-12-31T22:01:19.0919305Z [command]C:\hostedtoolcache\windows\terraform\0.14.3\x64\terraform.exe version
2020-12-31T22:01:19.2411671Z Terraform v0.14.3
2020-12-31T22:01:19.2416643Z + provider registry.terraform.io/hashicorp/azurerm v2.41.0
2020-12-31T22:01:19.2417148Z + provider registry.terraform.io/hashicorp/random v3.0.0
```

### Affected Resource(s)

* `azurerm_application_gateway`

### Terraform Configuration Files

NA - see output, it's an obvious issue

### Debug Output

Observe the difference in case of "resourcegroups" vs. "resourceGroups" below in the identity_ids.

```
2020-12-31T22:01:36.2104268Z An execution plan has been generated and is shown below.
2020-12-31T22:01:36.2105043Z Resource actions are indicated with the following symbols:
2020-12-31T22:01:36.2105494Z   ~ update in-place
2020-12-31T22:01:36.2105690Z 
2020-12-31T22:01:36.2106027Z Terraform will perform the following actions:
2020-12-31T22:01:36.2106725Z 
2020-12-31T22:01:36.2396259Z   # module.main.module.app_role.module.appgw["centralus"].azurerm_application_gateway.appgw will be updated in-place
2020-12-31T22:01:36.2397258Z   ~ resource "azurerm_application_gateway" "appgw" {
2020-12-31T22:01:36.2398070Z         id                  = "/subscriptions/xxxx-xxxxx-xxxx-xxxx/resourceGroups/my-rg-name/providers/Microsoft.Network/applicationGateways/my-rg-name-app-appgw"
2020-12-31T22:01:36.2398843Z         name                = "my-rg-name-app-appgw"
2020-12-31T22:01:36.2399274Z         tags                = {}
[...]

PR #10031, opened 2021-01-01
r/application_gateway: ensuring the casing on `identity_ids` within the `identity` block

JohnDelisle:
Fixes issue with `identity_ids` being case-sensitive, which results in terraform forcing an update on each apply due to case differences in e.g. `.../resourceGroups/...` vs. `.../resourcegroups/...`.  See issue #10029

Tested by building azurerm provider 2.4.0 locally.


favoretti:
Convert it to generator instead maybe?

katbyte (dismissed):
I'm going to second what @favoretti said, instead of making this property case insensitive we should use the new Parse functions when reading them in from azure to store them in the correct case.

JohnDelisle:
> I'm going to second what @favoretti said, instead of making this property case insensitive we should use the new Parse functions when reading them in from azure to store them in the correct case.

Thanks for the feedback.  Would your suggested approach still achieve the objective of being case-insensitive?

The `identity_ids` could be getting pulled programatically (e.g. as a data source or as a resource elsewhere in the module), or they could be added by humans who expect them to be case-insensitive.

katbyte:
@JohnDelisle - it will treat parts of the resource id as case insensitive (the non user specified bits like resourceGroup), you can see how to generate the parse functions in the resourceid.go files in many of the services now (not sure if they exist for ap gateware, i'm on mobile right now)

JohnDelisle:
I've followed @sharebear's lead in #10105 and using `azurerm/internal/services/msi/parse`.  Tested working with `/ResourceGroup/` and `/resourceGroup/`.

tombuildsstuff (approved):
LGTM - thanks for this @JohnDelisle

tombuildsstuff:
Tests look good 👍

PR #10049, opened 2021-01-05
r/devspace_controller: disabling the tests & highlighting this'll be removed in 3.0

tombuildsstuff:
There's a breaking behavioural change in the API where provisioning new instances returns:

> Azure Dev Spaces Controller creation is blocked because Azure Dev Spaces is being retired. Bridge to Kubernetes is an improved alternative to many of the Azure Dev Spaces development scenarios. Learn more here: https://aka.ms/migrate-to-bridge.

As such, there's no point running these tests since they're going to fail - but this resource will continue to be usable for existing users, albeit we can no longer test it.

This PR introduces a deprecation message for this resource - and prepares it to be removed in 3.0.


jackofallops (approved):
LGTM 👍

PR #10052, opened 2021-01-05
batch: fixing broken tests

tombuildsstuff:
This also ensures that all fields are set during the `flattenBatchPoolNetworkConfiguration` function


tombuildsstuff:
Ignoring one failing test (which is due to a behavioural change in the newer version of the Batch API, which'll need updating/new validation/documenting separately) - the tests pass:

Screenshot 2021-01-05 at 14 13 31

jackofallops (approved):
LGTM 👍

PR #10053, opened 2021-01-05
r/analysisservices_server: fixing the broken Suspend test

tombuildsstuff:
This uses a different resource group name to the Basic test meaning that when we scale this, this ultimately ends up destroying and recreating the Server (and the RG) rather than pausing and suspending it.


jackofallops (approved):
LGTM 👍

tombuildsstuff:
Tests pass:

Screenshot 2021-01-06 at 09 26 45

PR #10060, opened 2021-01-06
key vault related test fix-ups - part 1

jackofallops:
This PR attempts to resolve _most_ of the test issues introduced by the soft delete change on the service at the end of 2020.  Several test scenarios are not currently possible to fix at present, but these should be resolvable when binary testing is available.


tombuildsstuff (approved):
LGTM 👍

PR #10062, opened 2021-01-06
Fixed resource provider names for Blueprint and FHIR

shibayan:
Fix wrong resource provider names for Azure Blueprints and Azure API for FHIR. Deployment of Blueprint and FHIR by Terraform may fail if the provider is not registered.

Reference:
https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-services-resource-providers


katbyte (approved):
Thanks @shibayan - LGTM 👍

PR #10074, opened 2021-01-07
New resource: azurerm_express_route_port

magodo:
New resource: azurerm_express_route_port

## Reference

- https://docs.microsoft.com/en-us/azure/expressroute/expressroute-erdirect-about

## Test Result

```bash
💤 make testacc TEST=./azurerm/internal/services/network/tests TESTARGS='-run=TestAccAzureRMExpressRoutePort_'
TF_ACC=1 go test ./azurerm/internal/services/network/tests -v -run=TestAccAzureRMExpressRoutePort_ -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMExpressRoutePort_basic
=== PAUSE TestAccAzureRMExpressRoutePort_basic
=== RUN   TestAccAzureRMExpressRoutePort_complete
=== PAUSE TestAccAzureRMExpressRoutePort_complete
=== RUN   TestAccAzureRMExpressRoutePort_update
=== PAUSE TestAccAzureRMExpressRoutePort_update
=== RUN   TestAccAzureRMExpressRoutePort_requiresImport
=== PAUSE TestAccAzureRMExpressRoutePort_requiresImport
=== RUN   TestAccAzureRMExpressRoutePort_userAssignedIdentity
=== PAUSE TestAccAzureRMExpressRoutePort_userAssignedIdentity
=== CONT  TestAccAzureRMExpressRoutePort_basic
=== CONT  TestAccAzureRMExpressRoutePort_requiresImport
=== CONT  TestAccAzureRMExpressRoutePort_update
=== CONT  TestAccAzureRMExpressRoutePort_complete
=== CONT  TestAccAzureRMExpressRoutePort_userAssignedIdentity
--- PASS: TestAccAzureRMExpressRoutePort_basic (166.46s)
--- PASS: TestAccAzureRMExpressRoutePort_requiresImport (177.45s)
--- PASS: TestAccAzureRMExpressRoutePort_userAssignedIdentity (189.52s)
--- PASS: TestAccAzureRMExpressRoutePort_complete (227.78s)
--- PASS: TestAccAzureRMExpressRoutePort_update (289.52s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/tests       289.569s
```

## Cost

This resource takes a high cost on creation. Please refer to its [pricing page](https://azure.microsoft.com/en-us/pricing/details/expressroute/) for details.


azurerm/internal/services/network/express_route_port_resource.go
```diff
@@ -0,0 +1,514 @@
...
+					// string(network.GcmAes128),
+					// string(network.GcmAes256),
+				}, false),
+			},
+			"macsec_ckn_keyvault_identifier": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
+			"macsec_cak_keyvault_identifier": {
```
katbyte, line 68: would it make sense to group these into a macsec block?

azurerm/internal/services/network/express_route_port_resource.go
```diff
@@ -0,0 +1,514 @@
...
+							}, true),
+							DiffSuppressFunc: suppress.CaseDifference,
+						},
+					},
+				},
+			},
+
+			"link1": expressRoutePortSchema,
+
+			"link2": expressRoutePortSchema,
```
katbyte, line 190: could we use a list here so link.0/link.1 works instead?

azurerm/internal/services/network/express_route_port_resource.go
```diff
@@ -0,0 +1,514 @@
...
+
+			"ether_type": {
```
katbyte, line 192: is this short for ethernet_type?
```suggestion
			"ethernet_type": {
```

website/docs/r/express_route_port.html.markdown
```diff
@@ -0,0 +1,126 @@
...
+
+# azurerm_express_route_port
+
+Manages a Express Route Port.
+
+## Example Usage
+
+```hcl
+resource "azurerm_resource_group" "example" {
+  name     = "group1"
```
katbyte, line 17:
```suggestion
  name     = "example-resources"
```

azurerm/internal/services/network/express_route_port_resource.go
```diff
@@ -0,0 +1,514 @@
...
+				// TODO: The following hardcode can be replaced by SDK types once following is merged:
+				// 	https://github.com/Azure/azure-rest-api-specs/pull/12329
+				ValidateFunc: validation.StringInSlice([]string{
+					"GcmAes128",
+					"GcmAes256",
+					// string(network.GcmAes128),
+					// string(network.GcmAes256),
+				}, false),
+			},
+			"macsec_ckn_keyvault_identifier": {
```
katbyte, line 63: is there a reason not to just use id here?
```suggestion
			"macsec_ckn_keyvault_id": {
```

katbyte (approved):
Aside from a few minor comments about schema this looks good @magodo

azurerm/internal/services/network/express_route_port_resource.go
magodo: It groups things better, whilst we have to make the container block as required. Otherwise, we can allow uses to leave these macsec properties empty and use its default values. This is kind of a trade off, I'm not sure which one is better.

azurerm/internal/services/network/express_route_port_resource.go
magodo: We can make it to be a list with exactly two items, whilst it would require users to set the first link in order to update anything on the second link. Otherwise, we can allow users to modify each link independently.
Also confirmed from service team that there will be no more links added for the express route port resource, only twos for high availability.

azurerm/internal/services/network/express_route_port_resource.go
magodo: EtherType is a term it self, see: https://en.wikipedia.org/wiki/EtherType and https://docs.microsoft.com/en-us/azure/expressroute/expressroute-erdirect-about#technical-requirements, so we shall keep it as this.

azurerm/internal/services/network/express_route_port_resource.go
magodo: Change this to `macsec_ckn_keyvault_secret_id`.

magodo:
@katbyte Thank you for the review comments! I've resolved some of them and leaving the others for your further inputs. Please take another look :)

azurerm/internal/services/network/express_route_port_resource.go
katbyte: huh well TIL! https://en.wikipedia.org/wiki/EtherType

given the wiki and there being no space between the words i think it should be ethertype?
```suggestion
			"ethertype": {
```

azurerm/internal/services/network/express_route_port_resource.go
katbyte: thats a good point, lets keep it like this then

azurerm/internal/services/network/express_route_port_resource.go
katbyte: i'll leave it up to you, so whichever you think is best

magodo: I feel it is more easy for users to leave all this empty, rather than setting an empty `macsec {}` block in order to use the default 😅

azurerm/internal/services/network/express_route_port_resource.go
katbyte: normally yes, but i think in this instance it makes more sense to not add the _ as its not two words but one

azurerm/internal/services/network/express_route_port_resource.go
katbyte: 👍

website/docs/r/express_route_port.html.markdown
```diff
@@ -0,0 +1,126 @@
...
+
+* `id` - The ID of the Express Route Port.
+
+* `identity` - A `identity` block as defined below.
+  
+* `link` - A list of `link` block as defined below.
+
+* `guid` - The resource GUID of the Express Route Port.
+  
+* `ethertype` - The ether type of the Express Route Port.
```
katbyte, line 91:
```suggestion
* `ethertype` - The EtherType of the Express Route Port.
```

katbyte (approved):
Thanks @magodo aside from one comment this LGTM 👍

PR #10078, opened 2021-01-07
Add support for event_grouping_setting of azurerm_sentinel_alert_rule_scheduled

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/9725

Currently azurerm_sentinel_alert_rule_scheduled doesn't support grouping the events. So this PR is to implement it.

The EventGroupingSettings in rest api spec:
https://github.com/Azure/azure-rest-api-specs/blob/96ba256c1f61d4dc2a7928d7658e024b9b1615a6/specification/securityinsights/resource-manager/Microsoft.SecurityInsights/preview/2019-01-01-preview/SecurityInsights.json#L9262

--- PASS: TestAccSentinelAlertRuleScheduled_basic (160.00s)
--- PASS: TestAccSentinelAlertRuleScheduled_complete (160.04s)
--- PASS: TestAccSentinelAlertRuleScheduled_withAlertRuleTemplateGuid (160.07s)
--- PASS: TestAccSentinelAlertRuleScheduled_requiresImport (192.13s)
--- PASS: TestAccSentinelAlertRuleScheduled_updateEventGroupingSetting (222.86s)
--- PASS: TestAccSentinelAlertRuleScheduled_update (290.24s)


azurerm/internal/services/sentinel/sentinel_alert_rule_scheduled_resource.go
```diff
@@ -73,6 +73,24 @@ func resourceSentinelAlertRuleScheduled() *schema.Resource {
 				ValidateFunc: validation.StringIsNotEmpty,
 			},
 
+			"event_grouping_setting": {
```
katbyte, line 76: I think we can remove setting
```suggestion
			"event_grouping": {
```

azurerm/internal/services/sentinel/sentinel_alert_rule_scheduled_resource.go
```diff
@@ -73,6 +73,24 @@ func resourceSentinelAlertRuleScheduled() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"aggregation_kind": {
```
katbyte, line 82: given there is only one property here? maybe we should pull this out and call it
```suggestion
						"Event_grouping_aggregation_methold": {
```

(kind isn't great, methold is more descriptive)

katbyte (changes requested):
Thanks @neil-yechenwei - overall this looks good but i have one comment about the property name i've left inline

azurerm/internal/services/sentinel/sentinel_alert_rule_scheduled_resource.go
neil-yechenwei: I am not sure whether api would introduce other types in the future. If I put this out, I still need to update the property schema like this and it would become breaking change. Does it make sense?

neil-yechenwei: It's "method" not "methold", right?

azurerm/internal/services/sentinel/sentinel_alert_rule_scheduled_resource.go
neil-yechenwei: updated

neil-yechenwei:
@katbyte , thanks for your comments. I've updated code and answered your questions. Please have an another look.

azurerm/internal/services/sentinel/sentinel_alert_rule_scheduled_resource.go
katbyte: Yes method, and its up to you, honestly its not a big deal to deprecate a string property if they add one

neil-yechenwei: OK. Thanks. I prefer to keep it.

katbyte (approved):

PR #10088, opened 2021-01-07
r/key_vault: defaulting `soft_delete_enabled` on to match the updated API behaviour

### LINKED ISSUE #9988: Azure Key Vault now requires Soft Deleted to be Enabled

tombuildsstuff:
This PR updates the `azurerm_key_vault` resource to match the updated API behaviour rolled out over late last month / the new year. Whilst #10007 provides one solution to this, after extensive testing there was ultimately more necessary to make this work, as such this PR includes:

* Data Source: `azurerm_key_vault` - deprecating the field `soft_delete_enabled` (and conditionally removing this should 3.0 mode be enabled)
* Resource: `azurerm_key_vault` - deprecating the field `soft_delete_enabled` and defaulting this to true (conditionally removing this should 3.0 mode be enabled). Notably this PR removes the error when attempting to disable/enable this - but since this is hard-coded to `true` in the Read function, operators can update their configurations (or remove the field) to clear this diff.
* * Resource: `azurerm_key_vault` - defaulting the field `soft_delete_retention_days` to `90` days, to match the API behaviour. Notably it's only possible to set this value a single time (either during create, or one-time during update) at which point it's immutable - so a state migration was necessary to default this if not explicitly set - as it's defaulted in the API but not returned unless explicitly configured.

As such this PR supersedes #10007 but fixes #9988 (albeit in a slightly different manner).

### DIFF
--- a/azurerm/internal/services/keyvault/key_vault_data_source.go
+++ b/azurerm/internal/services/keyvault/key_vault_data_source.go
@@ -8,6 +8,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/keyvault/validate"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tf/set"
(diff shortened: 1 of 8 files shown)


tombuildsstuff:
For the changelog:

```
BREAKING CHANGES

* `azurerm_key_vault` - the field `soft_delete_enabled` is now defaulted to `true` to match the breaking change in the Azure API where Key Vaults now have Soft Delete enabled by default, which cannot be disabled. This property is now non-functional, defaults to `true` and will be removed in version 3.0 of the Azure Provider. [GH-10088]
* `azurerm_key_vault` - the field `soft_delete_retention_days` is now defaulted to `90` days to match the Azure API behaviour, as the Azure API does not return a value for this field when not explicitly configured, so defaulting this removes a diff with `0`. [GH-10088]

IMPROVEMENTS

* `azurerm_key_vault` - the field `soft_delete_enabled` is now defaulted to `true` to match the Azure API behaviour where Soft Delete is force-enabled and can no longer be disabled. This field is deprecated, can be safely removed from your Terraform Configuration, and will be removed in version 3.0 of the Azure Provider. [GH-10088]

BUG FIXES

* `azurerm_key_vault` - the field `soft_delete_retention_days` is now defaulted to `90` days to match the Azure API behaviour. [GH-10088]
```

tombuildsstuff:
Ignoring one failing test in a destroy function due to some recent refactoring (which'll be fixed separately) - the tests pass:

Screenshot 2021-01-07 at 21 37 27

katbyte (approved):
LGTM 👍

PR #10090, opened 2021-01-08
Enhancement on MSI auth guide

magodo:
Add more less known facts about MSI auth for Terraform.

### DIFF
--- a/website/docs/guides/managed_service_identity.html.markdown
+++ b/website/docs/guides/managed_service_identity.html.markdown
@@ -3,17 +3,16 @@ layout: "azurerm"
 page_title: "Azure Provider: Authenticating via Managed Identity"
 description: |-
   This guide will cover how to use managed identity for Azure resources as authentication for the Azure Provider.
-
 ---
 
 # Azure Provider: Authenticating using managed identities for Azure resources
 
 Terraform supports a number of different methods for authenticating to Azure:
 
-* [Authenticating to Azure using the Azure CLI](azure_cli.html)
-* Authenticating to Azure using Managed Identity (covered in this guide)
-* [Authenticating to Azure using a Service Principal and a Client Certificate](service_principal_client_certificate.html)
-* [Authenticating to Azure using a Service Principal and a Client Secret](service_principal_client_secret.html)
+- [Authenticating to Azure using the Azure CLI](azure_cli.html)
+- Authenticating to Azure using Managed Identity (covered in this guide)
+- [Authenticating to Azure using a Service Principal and a Client Certificate](service_principal_client_certificate.html)
+- [Authenticating to Azure using a Service Principal and a Client Secret](service_principal_client_secret.html)
 
 ---
 
@@ -71,11 +70,13 @@ Terraform can be configured to use managed identity for authentication in one of
 
 ### Configuring with environment variables
 
-Setting the`ARM_USE_MSI` (equivalent to provider block argument [`use_msi`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#use_msi)) to `true` tells Terraform to use a managed identity.
+Setting the`ARM_USE_MSI` environment variable (equivalent to provider block argument [`use_msi`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#use_msi)) to `true` tells Terraform to use a managed identity.
 
-By default, Terraform will use the system assigned identity for authentication. To use a user assigned identity instead, you will need to specify the `ARM_CLIENT_ID` (equivalent to provider block argument [`client_id`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#client_id)) to the [client id](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/user_assigned_identity#client_id) of the identity.
+By default, Terraform will use the system assigned identity for authentication. To use a user assigned identity instead, you will need to specify the `ARM_CLIENT_ID` environment variable (equivalent to provider block argument [`client_id`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#client_id)) to the [client id](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/user_assigned_identity#client_id) of the identity.
 
-By default, Terraform will use a well-known MSI endpoint to get the authentication token, which covers the most use cases. Whilst the endpoint might be different (e.g. Azure Function App) in other cases, where users need to explicitly specify the endpoint via `ARM_MSI_ENDPOINT` (equivalent to provider block argument [`msi_endpoint`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#msi_endpoint)).
+By default, Terraform will use a well-known MSI endpoint to get the authentication token, which covers most use cases. In other cases where the endpoint is different (e.g. when running as an Azure Function App), you must explicitly specify the endpoint using the `ARM_MSI_ENDPOINT` environment variable (equivalent to provider block argument [`msi_endpoint`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#msi_endpoint)).
+
+!> **Note:** we recommend against running Terraform inside of a Function App as the low memory ceiling can lead to Terraform being terminated and data (including the State File) being lost. Instead we’d recommend considering triggering an external process, such as Terraform Cloud or a CI System to run these longer-running more intensive processes - see [Terraform in Automation](https://learn.hashicorp.com/tutorials/terraform/automate-terraform) for more details.
 
 In addition to a properly-configured management identity, Terraform needs to know the subscription ID and tenant ID to identify the full context for the Azure provider.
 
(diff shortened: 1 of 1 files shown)


website/docs/guides/managed_service_identity.html.markdown
```diff
@@ -71,12 +71,20 @@ Terraform can be configured to use managed identity for authentication in one of
 
 ### Configuring with environment variables
 
-Setting the `ARM_USE_MSI` environment variable to `true` tells Terraform to use a managed identity. In addition to a properly-configured management identity, Terraform needs to know the subscription ID and tenant ID to identify the full context for the Azure provider.
+Setting the `ARM_USE_MSI` environment variable to `true` tells Terraform to use a managed identity.
+
+By default, Terraform will use the system assigned identity for authentication. If users want to use user assigned identity instead, the `ARM_CLIENT_ID` have to be specified to the [client id](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/user_assigned_identity#client_id) of the identity.
+
+By default, Terraform will use the MSI endpoint provided by MSI VM Extension to get the authentication token, which covers the most use cases. Whilst the endpoint might be different (e.g. Azure Function App) in other cases, where users need to explicitly specify the endpoint via `ARM_MSI_ENDPOINT`.
```
tombuildsstuff, line 78: this is no longer a VM Extension as far as I'm aware, more something the hypervisor exposes? but same here, could we highlight the provider block field (as above)

website/docs/guides/managed_service_identity.html.markdown
```diff
@@ -71,12 +71,20 @@ Terraform can be configured to use managed identity for authentication in one of
 
 ### Configuring with environment variables
 
-Setting the `ARM_USE_MSI` environment variable to `true` tells Terraform to use a managed identity. In addition to a properly-configured management identity, Terraform needs to know the subscription ID and tenant ID to identify the full context for the Azure provider.
+Setting the `ARM_USE_MSI` environment variable to `true` tells Terraform to use a managed identity.
```
tombuildsstuff, line 74: we should also highlight the provider block field here, as we're doing it below

website/docs/guides/managed_service_identity.html.markdown
```diff
@@ -71,12 +71,20 @@ Terraform can be configured to use managed identity for authentication in one of
...
+
+By default, Terraform will use the system assigned identity for authentication. If users want to use user assigned identity instead, the `ARM_CLIENT_ID` have to be specified to the [client id](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/user_assigned_identity#client_id) of the identity.
```
tombuildsstuff, line 76:
```suggestion
By default, Terraform will use the system assigned identity for authentication. To use a user assigned identity instead, you will need to specify the  [`client_id`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/user_assigned_identity#client_id) of the identity (which can also be sourced from the Environment Variable `ARM_CLIENT_ID`).
```

tombuildsstuff (commented):
A few minor points but this otherwise LGTM 👍

website/docs/guides/managed_service_identity.html.markdown
magodo: I changed it to "well-known" endpoint.

website/docs/guides/managed_service_identity.html.markdown
magodo: Given this section talks about configuring with environment variables, I still mainly talks about env var, but added the provider block argument alternatives after each env var.

website/docs/guides/managed_service_identity.html.markdown
magodo: The link of "client id" here redirects to the document of `azurerm_user_assigned_identity.client_id` actually. Although I added the provider block argument alternatives after each env var, I still want to keep above link there to guide users to use the right attribute if they provision user assigned identity via Terraform.

magodo:
@tombuildsstuff To be able to use MSI for other endpoint other than the well-known one, we probably need to merge hashicorp/go-azure-helpers#69 and also update the azure-go-helper in this repo.

website/docs/guides/managed_service_identity.html.markdown
```diff
@@ -71,12 +71,20 @@ Terraform can be configured to use managed identity for authentication in one of
 
 ### Configuring with environment variables
 
-Setting the `ARM_USE_MSI` environment variable to `true` tells Terraform to use a managed identity. In addition to a properly-configured management identity, Terraform needs to know the subscription ID and tenant ID to identify the full context for the Azure provider.
+Setting the`ARM_USE_MSI` (equivalent to provider block argument [`use_msi`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#use_msi)) to `true` tells Terraform to use a managed identity.
```
manicminer, line 74:
```suggestion
Setting the`ARM_USE_MSI` environment variable (equivalent to provider block argument [`use_msi`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#use_msi)) to `true` tells Terraform to use a managed identity.
```

website/docs/guides/managed_service_identity.html.markdown
```diff
@@ -71,12 +71,20 @@ Terraform can be configured to use managed identity for authentication in one of
...
+
+By default, Terraform will use the system assigned identity for authentication. To use a user assigned identity instead, you will need to specify the `ARM_CLIENT_ID` (equivalent to provider block argument [`client_id`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#client_id)) to the [client id](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/user_assigned_identity#client_id) of the identity.
+
+By default, Terraform will use a well-known MSI endpoint to get the authentication token, which covers the most use cases. Whilst the endpoint might be different (e.g. Azure Function App) in other cases, where users need to explicitly specify the endpoint via `ARM_MSI_ENDPOINT` (equivalent to provider block argument [`msi_endpoint`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#msi_endpoint)).
```
manicminer, line 78:
```suggestion
By default, Terraform will use a well-known MSI endpoint to get the authentication token, which covers most use cases. In other cases where the endpoint is different (e.g. when running as an Azure Function App), you must explicitly specify the endpoint using the `ARM_MSI_ENDPOINT` environment variable (equivalent to provider block argument [`msi_endpoint`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#msi_endpoint)).
```

website/docs/guides/managed_service_identity.html.markdown
```diff
@@ -71,12 +71,20 @@ Terraform can be configured to use managed identity for authentication in one of
...
+
+In addition to a properly-configured management identity, Terraform needs to know the subscription ID and tenant ID to identify the full context for the Azure provider.
 
 ```shell
 $ export ARM_USE_MSI=true
 $ export ARM_SUBSCRIPTION_ID=159f2485-xxxx-xxxx-xxxx-xxxxxxxxxxxx
 $ export ARM_TENANT_ID=72f988bf-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+$ # export ARM_CLIENT_ID=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx # only necessary for user assigned identity
+$ # export ARM_MSI_ENDPOINT=$MSI_ENDPOINT # only necessary when the msi endpoint is different than the well-known one
```
manicminer, line 87: I think it's less confusing without commenting these out? It also highlights them properly.

```suggestion
$ export ARM_CLIENT_ID=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx # only necessary for user assigned identity
$ export ARM_MSI_ENDPOINT=$MSI_ENDPOINT # only necessary when the msi endpoint is different than the well-known one
```

manicminer (changes requested):
@magodo I made a few wording suggestions to clarify the intent a little.

website/docs/guides/managed_service_identity.html.markdown
tombuildsstuff: We've previously seen customers encounter Terraform crashing (and the state file being lost) when running Terraform inside of a Function App/Container with a low hard-memory ceiling. After chatting about this internally - we believe it's worth adding a big red box here recommending against this:

```suggestion
By default, Terraform will use a well-known MSI endpoint to get the authentication token, which covers the most use cases. Whilst the endpoint might be different (e.g. Azure Function App) in other cases, where users need to explicitly specify the endpoint via `ARM_MSI_ENDPOINT` (equivalent to provider block argument [`msi_endpoint`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#msi_endpoint)).

!> **Note:** we recommend against running Terraform inside of a Function App as the low memory ceiling can lead to Terraform being terminated and data (including the State File) being lost. Instead we’d recommend considering triggering an external process, such as Terraform Cloud or a CI System to run these longer-running more intensive processes - see [Terraform in Automation](https://learn.hashicorp.com/tutorials/terraform/automate-terraform) for more details.
```

magodo:
Thank you @manicminer and @tombuildsstuff for your second round of review. I've made another commit, please take a look.

tombuildsstuff (approved):
LGTM - thanks for those changes @magodo

Issue #10097, opened 2021-01-08
monitor_diagnostic_setting_resource crashes with Error: ID was missing the `authorizationRules` element


andrey-dubnik:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

azurerm 2.42.0
terraform 0.12.30

### Affected Resource(s)

* `monitor_diagnostic_setting_resource`

### Issue Description

Over the last few days we start having issues with the infrastructure builds and modifications while using the diagnostics resource.

The cause seem to be the diagnostic receives AuthorizationRules where Azure SDK change now requires authorizationRules.

Following is the current state representation

```
{
      "module": "module.bastionhost.module.bastion_diagnostics",
      "mode": "managed",
      "type": "azurerm_monitor_diagnostic_setting",
      "name": "logs",
      "each": "list",
      "provider": "provider.azurerm",
      "instances": [
        {
          "index_key": 0,
          "schema_version": 0,
          "attributes": {
            "eventhub_authorization_rule_id": "/subscriptions/****/resourceGroups/rgpazewp-udm-cms/providers/Microsoft.EventHub/namespaces/udmcmsuf-prod-ehnm/AuthorizationRules/DataWriter",
```

Accordingly to the code event rule gets parsed when received from the diagnostics resource

https://github.com/terraform-providers/terraform-provider-azurerm/blob/31341dda6bf1c8771b2f57e54522961872dfcddd/azurerm/internal/services/monitor/monitor_diagnostic_setting_resource.go#L321

Result of the parse ends up being "Error: ID was missing the `authorizationRules` element" cause Parse logic expects authorizationRules in the path...

The obvious fix would be to alter the retrieved ID and replace AuthorizationRules with authorizationRules but I suspect there may be a better option?

andrey-dubnik:
P.S. the issue here is not the state but the actual ID resource returns - e.g. AuthorizationRules VS authorizationRules hence the parse fails

ecarlson94:
This is a breaking change between 2.41 and 2.42 that has cause my company a massive headache.

[...]

PR #10104, opened 2021-01-08
r/monitor_diagnostic_setting: handling a difference in casing the EHNamespace Authorization Rule

andrey-dubnik:
Handle Azure read response with AuthorizationRules (legacy) and authorizationRules

#10097


azurerm/internal/services/monitor/monitor_diagnostic_setting_resource.go
```diff
@@ -318,7 +318,7 @@ func resourceMonitorDiagnosticSettingRead(d *schema.ResourceData, meta interface
 	d.Set("eventhub_name", resp.EventHubName)
 	eventhubAuthorizationRuleId := ""
 	if resp.EventHubAuthorizationRuleID != nil && *resp.EventHubAuthorizationRuleID != "" {
-		parsedId, err := eventhubParse.NamespaceAuthorizationRuleID(*resp.EventHubAuthorizationRuleID)
+		parsedId, err := eventhubParse.NamespaceAuthorizationRuleID(strings.Replace(*resp.EventHubAuthorizationRuleID, "/AuthorizationRules/", "/authorizationRules/", 1))
```
tombuildsstuff, line 321: rather than mutating the ID here, we can use an insensitive segment parser instead, by adding a `-rewrite=true` argument to this line: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/eventhub/resourceids.go#L6 and then running `make generate` (or `go generate ./azurerm/internal/services/eventhub`, which is likely quicker here)

This then introduces a secondary `{Name}Insensitively` method which can be used for parsing the segments insensitively, meaning that we can make this:

```suggestion
		parsedId, err := eventhubParse.NamespaceAuthorizationRuleIDInsensitively(*resp.EventHubAuthorizationRuleID)
```

tombuildsstuff (changes requested):
hey @andrey-dubnik 

Thanks for this PR - taking a look through whilst this does solve the issue by manipulating the ID, we should be able to use a segment insensitive parser here during rewriting by generating that instead - as such if we can update this to use that then this otherwise LGTM 👍 

Thanks!

tombuildsstuff (approved):
LGTM - thanks for pushing those changes @andrey-dubnik 👍

andrey-dubnik:
@tombuildsstuff I've also changed a code gen as otherwise generated code was failing on lint during the CI tests. All passes now :)

tombuildsstuff (approved):
LGTM  👍

tombuildsstuff:
Tests look good 👍

PR #10106, opened 2021-01-08
r/security_center_automation: switching to check the state which is sufficient

tombuildsstuff:
This removes usages on the `acceptance.AzureProvider` since these checks can be done by checking the state


katbyte (approved):
LGTM 👍

PR #10125, opened 2021-01-11
azurerm_mssql_virtual_machine: fix tests and persistent diff, use relative expiry for service principal password

manicminer:


jackofallops (approved):
Thanks @manicminer - LGTM 👍

manicminer:
The test still doesn't pass, but the failure is due to a persistent diff on another block which will be fixed in another PR.

```
    === CONT  TestAccMsSqlVirtualMachine_withKeyVault
        testing.go:684: Step 0 error: After applying this step, the plan was not empty:
            
            DIFF:
            
            UPDATE: azurerm_mssql_virtual_machine.test
              auto_patching.#:                                 "0" => "0"
              id:                                              "/subscriptions/*******/resourceGroups/acctestRG-mssql-210111071526852118/providers/Microsoft.SqlVirtualMachine/sqlVirtualMachines/acctest-VM-210111071526852118" => "/subscriptions/*******/resourceGroups/acctestRG-mssql-210111071526852118/providers/Microsoft.SqlVirtualMachine/sqlVirtualMachines/acctest-VM-210111071526852118"
              key_vault_credential.#:                          "1" => "1"
              key_vault_credential.0.key_vault_url:            "https://acckv-210111071526852118.vault.azure.net/keys/key-210111071526852118/dd21fd1eba9f42e091936a9fb93494cb" => "https://acckv-210111071526852118.vault.azure.net/keys/key-210111071526852118/dd21fd1eba9f42e091936a9fb93494cb"
              key_vault_credential.0.name:                     "acckv-210111071526852118:acctestkv" => "acckv-210111071526852118:acctestkv"
              key_vault_credential.0.service_principal_name:   "acctestspa210111071526852118" => "acctestspa210111071526852118"
              key_vault_credential.0.service_principal_secret: "d78b538b-c58e-46ac-57d4-247a1afd04a8" => "d78b538b-c58e-46ac-57d4-247a1afd04a8"
              r_services_enabled:                              "false" => "false"
              sql_connectivity_port:                           "1433" => "1433"
              sql_connectivity_type:                           "PRIVATE" => "PRIVATE"
              sql_license_type:                                "PAYG" => "PAYG"
              storage_configuration.#:                         "1" => "0"
              storage_configuration.0.data_settings.#:         "0" => ""
              storage_configuration.0.disk_type:               "" => ""
              storage_configuration.0.log_settings.#:          "0" => ""
              storage_configuration.0.storage_workload_type:   "" => ""
              storage_configuration.0.temp_db_settings.#:      "0" => ""
              virtual_machine_id:                              "/subscriptions/*******/resourceGroups/acctestRG-mssql-210111071526852118/providers/Microsoft.Compute/virtualMachines/acctest-VM-210111071526852118" => "/subscriptions/*******/resourceGroups/acctestRG-mssql-210111071526852118/providers/Microsoft.Compute/virtualMachines/acctest-VM-210111071526852118"
```

manicminer:
Tests pass with one unrelated destroy failure:

Screenshot 2021-01-11 at 23 34 55

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
```diff
@@ -510,10 +528,10 @@ func expandSqlVirtualMachineDataStorageSettings(input []interface{}) *sqlvirtual
 }
 
 func expandSqlVirtualMachineStorageSettingsLuns(input []interface{}) *[]int32 {
-	expandedLuns := make([]int32, len(input))
+	expandedLuns := make([]int32, 0, len(input))
```
jackofallops, line 531: Not sure we need the capacity arg here since we're ranging over `input`?

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
```diff
@@ -486,15 +486,33 @@ func flattenSqlVirtualMachineStorageConfigurationSettings(input *sqlvirtualmachi
...
-		},
+	output := make(map[string]interface{})
+
+	if v := string(input.DiskConfigurationType); v != "" {
+		output["disk_type"] = v
+	}
+
+	if storageWorkloadType != "" {
+		output["storage_workload_type"] = storageWorkloadType
+	}
```
jackofallops, line 497: this possibly wants setting as an empty string so there's always a value?

jackofallops (approved):
Thanks @manicminer - Just a couple questions / comments in-line, otherwise LGTM 👍

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
manicminer: OK cool, yeah it works without it too.

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
manicminer: Unfortunately the problem is that Terraform shows state drift if the containing list has any elements when there are none in the config. It looks like the API has started returning an array having zero values instead of omitting it from the response, which is why I've made it try and detect that. Is there is a better way to avoid state drift in such circumstances?

tombuildsstuff: agreed we should track an empty string in the state in all cases - but if the nested object is returned as an empty object (e.g. all fields are empty) then we'd need to do conditionally return an empty object rather than the nested object - and presumably raise an API bug for that?

manicminer: @tombuildsstuff Yep an empty object is exactly what's being returned. I'll raise an API bug today and adjust the logic so that an empty string is saved if the other attributes are present and non-zero.

tombuildsstuff: given this affects the entire block - presuming we're using intermediate variables, we should be able to conditionally check the set and conditionally return the nested object, e.g.:

```
foo := ""
if props.Foo != nil { foo = *props.Foo }

bar := ""
if props.Bar != nil { bar = *props.Bar }

if foo == "" && bar == "" {  return []interface{}{} }

return []interface{}{
  map[string]interface{}{
    "foo": foo,
    "bar": bar,
  },
}
```

(we've a few instances of this of late)

manicminer: Cool, makes sense. Have adjusted it to always set all attributes [in the block], but return an empty slice if they all have zero values.

katbyte (approved):
Thanks @manicminer - LGTM 👍

PR #10131, opened 2021-01-11
r/hdinsight_kafka_cluster: switching the AAD Client to come from Authorization

tombuildsstuff:
This PR moves the usage of the AAD Client out into the Authorization package - where the other AAD items are stored

Supersedes #10130


manicminer (approved):
LGTM :+1:

PR #10144, opened 2021-01-12
sentinel: generating & switching to use the resource id parsers

magodo:
Sentinel resources opt-in resourceid.


azurerm/internal/services/sentinel/sentinel_alert_rule_data_source.go
```diff
@@ -51,17 +53,22 @@ func dataSourceSentinelAlertRuleRead(d *schema.ResourceData, meta interface{}) e
...
 		if utils.ResponseWasNotFound(resp.Response) {
-			return fmt.Errorf("Sentinel Alert Rule %q (Resource Group %q / Workspace: %q) was not found", name, workspaceID.ResourceGroup, workspaceID.WorkspaceName)
+			return fmt.Errorf("Sentinel Alert Rule %q (Workspace: %q) was not found", name, workspaceID)
 		}
 
-		return fmt.Errorf("retrieving Sentinel Alert Rule %q (Resource Group %q / Workspace: %q): %+v", name, workspaceID.ResourceGroup, workspaceID.WorkspaceName, err)
+		return fmt.Errorf("retrieving Sentinel Alert Rule %q (Workspace: %q): %+v", name, workspaceID, err)
 	}
 
 	id := alertRuleID(resp.Value)
```
tombuildsstuff, line 62: since we're managing our own ID and are only concerned that this exists within the API - we don't need to rely on the one returned from the API here - so this can become:

```suggestion
	id :=  parse.NewAlertRuleID(subscriptionId, resourceGroup, workspaceName, name)
	d.SetID(id.ID())
```

azurerm/internal/services/sentinel/sentinel_alert_rule_fusion_resource.go
```diff
@@ -106,28 +106,35 @@ func resourceSentinelAlertRuleFusionCreateUpdate(d *schema.ResourceData, meta in
...
+		return fmt.Errorf("retrieving Sentinel Alert Rule Fusion %q (Workspace: %q): %+v", name, workspaceID, err)
 	}
+
 	id := alertRuleID(resp.Value)
 	if id == nil || *id == "" {
-		return fmt.Errorf("empty or nil ID returned for Sentinel Alert Rule Fusion %q (Resource Group %q / Workspace: %q) ID", name, workspaceID.ResourceGroup, workspaceID.WorkspaceName)
+		return fmt.Errorf("empty or nil ID returned for Sentinel Alert Rule Fusion %q (Workspace: %q) ID", name, workspaceID)
+	}
+
+	resId, err := parse.AlertRuleID(*id)
```
tombuildsstuff, line 132: same here, we should be able to remove the line 126/127 down to here in favour of building this up using the  NewAlertRuleID func instead?

azurerm/internal/services/sentinel/sentinel_alert_rule_ms_security_incident_resource.go
```diff
@@ -196,28 +196,35 @@ func resourceSentinelAlertRuleMsSecurityIncidentCreateUpdate(d *schema.ResourceD
 	if !d.IsNewResource() {
 		resp, err := client.Get(ctx, workspaceID.ResourceGroup, "Microsoft.OperationalInsights", workspaceID.WorkspaceName, name)
 		if err != nil {
-			return fmt.Errorf("retrieving Sentinel Alert Rule Ms Security Incident %q (Resource Group %q / Workspace: %q): %+v", name, workspaceID.ResourceGroup, workspaceID.WorkspaceName, err)
+			return fmt.Errorf("retrieving Sentinel Alert Rule Ms Security Incident %q (Workspace: %q): %+v", name, workspaceID, err)
```
tombuildsstuff, line 199: heads up in this case this'll be rendered as:

```
(Workspace: "Workspace "name" (Resource Group %q)")
```

however if we need to build up the ID below anyway - we could likely reduce this down to:

```suggestion
			return fmt.Errorf("retrieving Ms Security Incident %s: %+v", id, err)
```

(same through the other errors here)

azurerm/internal/services/sentinel/sentinel_alert_rule_ms_security_incident_resource.go
```diff
@@ -196,28 +196,35 @@ func resourceSentinelAlertRuleMsSecurityIncidentCreateUpdate(d *schema.ResourceD
...
+		return fmt.Errorf("retrieving Sentinel Alert Rule Ms Security Incident %q (Workspace: %q): %+v", name, workspaceID, err)
 	}
+
 	id := alertRuleID(resp.Value)
 	if id == nil || *id == "" {
-		return fmt.Errorf("empty or nil ID returned for Sentinel Alert Rule Ms Security Incident %q (Resource Group %q / Workspace: %q) ID", name, workspaceID.ResourceGroup, workspaceID.WorkspaceName)
+		return fmt.Errorf("empty or nil ID returned for Sentinel Alert Rule Ms Security Incident %q (Workspace: %q) ID", name, workspaceID)
+	}
+
+	resId, err := parse.AlertRuleID(*id)
```
tombuildsstuff, line 222: (as above) we should be able to compose this information based off user input, rather than making this additional Get request and parsing this from the API

azurerm/internal/services/sentinel/sentinel_alert_rule_ms_security_incident_resource_test.go
```diff
@@ -137,19 +137,19 @@ func TestAccSentinelAlertRuleMsSecurityIncident_withDisplayNameExcludeFilter(t *
...
+	resp, err := clients.Sentinel.AlertRulesClient.Get(ctx, id.ResourceGroup, "Microsoft.OperationalInsights", id.WorkspaceName, id.Name)
 	if err != nil {
-		return nil, fmt.Errorf("reading Sentinel Alert Rule Ms Security Incident %q (Resource Group %q): %v", id.Name, id.ResourceGroup, err)
+		return nil, fmt.Errorf("reading Sentinel Alert Rule Ms Security Incident %q: %v", id, err)
 	}
 
 	rule, ok := resp.Value.(securityinsight.MicrosoftSecurityIncidentCreationAlertRule)
 	if !ok {
-		return nil, fmt.Errorf("reading Sentinel Alert Rule Ms Security Incident %q (Resource Group %q) is not of MicrosoftSecurityIncidentCreationAlertRule kind", id.Name, id.ResourceGroup)
+		return nil, fmt.Errorf("reading Sentinel Alert Rule Ms Security Incident %q is not of MicrosoftSecurityIncidentCreationAlertRule kind", id)
```
tombuildsstuff, line 152: on these messages, the phrasing `kind` is unclear - could we make this clearer to users:

```suggestion
		return nil, fmt.Errorf("the Alert Rule %q is not a MS Security Incident Alert Rule", id)
```

azurerm/internal/services/sentinel/sentinel_alert_rule_scheduled_resource.go
```diff
@@ -305,29 +305,34 @@ func resourceSentinelAlertRuleScheduledCreateUpdate(d *schema.ResourceData, meta
...
-		return fmt.Errorf("empty or nil ID returned for Sentinel Alert Rule Scheduled %q (Resource Group %q / Workspace %q) ID", name, workspaceID.ResourceGroup, workspaceID.WorkspaceName)
+		return fmt.Errorf("empty or nil ID returned for Sentinel Alert Rule Scheduled %q (Workspace: %q) ID", name, workspaceID)
 	}
-	d.SetId(*id)
+
+	resId, err := parse.AlertRuleID(*id)
+	if err != nil {
+		return err
+	}
+	d.SetId(resId.ID())
```
tombuildsstuff, line 335: (as above)

tombuildsstuff (changes requested):
hey @magodo 

Thanks for this PR - taking a look through on the whole this is looking good, but as a general rule we should be able to lean on the "New" function here rather than parsing and reformatting the ID here instead, that said if we can fix up the comments this should be otherwise good to go 👍 

Thanks!

magodo:
@tombuildsstuff Thank you for the comments 👍  I've updated the code accordingly, please take another look!

tombuildsstuff (approved):
LGTM 👍

tombuildsstuff:
@magodo the tests for this are failing with:

```
------- Stdout: -------
=== RUN   TestAccSentinelAlertRuleMsSecurityIncident_basic
=== PAUSE TestAccSentinelAlertRuleMsSecurityIncident_basic
=== CONT  TestAccSentinelAlertRuleMsSecurityIncident_basic
When applying changes to
azurerm_sentinel_alert_rule_ms_security_incident.test, provider "azurerm"
produced an unexpected new value for was present, but now absent.
issue tracker.
--- FAIL: TestAccSentinelAlertRuleMsSecurityIncident_basic (160.41s)
FAIL
```

```
------- Stdout: -------
=== RUN   TestAccSentinelAlertRuleFusion_update
=== PAUSE TestAccSentinelAlertRuleFusion_update
=== CONT  TestAccSentinelAlertRuleFusion_update
When applying changes to azurerm_sentinel_alert_rule_fusion.test, provider
"azurerm" produced an unexpected new value for was present, but now absent.
issue tracker.
--- FAIL: TestAccSentinelAlertRuleFusion_update (161.47s)
FAIL
```

It appears the ID being set in the Create/Update is returning a not found when it's then being looked up in the Read?

magodo:
@tombuildsstuff Ah... that's a typo.. Have been fixed and passed test locally.

tombuildsstuff:
Rebased this on top of master

tombuildsstuff:
Tests pass:

Screenshot 2021-01-14 at 10 52 00

Issue #7516, opened 2020-06-29
azure_linux_virtual_machine - Allocation failure of VM requires deleting or importing resource


KimShepherdVibrato:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.12.24
+ provider.azurerm v2.3.0

### Affected Resource(s)

* `azurerm_linux_virtual_machine`

This likely affects `azurerm_windows_virtual_machine` and possibly `azurerm_virtual_machine` as well.

### Terraform Configuration Files

```hcl
resource "azurerm_resource_group" "example" {
  name     = "example-resources"
  location = "West Europe"
}

resource "azurerm_virtual_network" "example" {
  name                = "example-network"
  address_space       = ["10.0.0.0/16"]
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name
}

resource "azurerm_subnet" "example" {
  name                 = "internal"
  resource_group_name  = azurerm_resource_group.example.name
  virtual_network_name = azurerm_virtual_network.example.name
  address_prefix       = "10.0.2.0/24"
}

resource "azurerm_network_interface" "example" {
  name                = "example-nic"
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name

  ip_configuration {
    name                          = "internal"
    subnet_id                     = azurerm_subnet.example.id
    private_ip_address_allocation = "Dynamic"
  }
}

resource "azurerm_linux_virtual_machine" "example" {
  name                = "example-machine"
  resource_group_name = azurerm_resource_group.example.name
  location            = azurerm_resource_group.example.location
  size                = "Standard_F2"
  admin_username      = "adminuser"
  network_interface_ids = [
    azurerm_network_interface.example.id,
  ]

  admin_ssh_key {
    username   = "adminuser"
    public_key = file("~/.ssh/id_rsa.pub")
  }

  os_disk {
    caching              = "ReadWrite"
    storage_account_type = "Standard_LRS"
  }

  source_image_reference {
    publisher = "Canonical"
[...]

PR #10150, opened 2021-01-12
[FIX]  azure_linux_virtual_machine - Allocation failure

HappyTobi:
Hi all,

there was just a small fix for the issue #7516.

When you create a VM in a region where you couldn't get the requested resource you ran in that issue mentioned in #7516.

To create a Linux Resource we are using the async [Azure api](https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/createorupdate) to create or update the VM.
The Azure api normally returns an **20x** status code that will give you the garantie that the VM will be created. (also wenn it's later running into a **failed** state).
So it was not enough to handle an error (Http status 4XX/5XX ) because the VM will be created on Azure ... also when it runs into a failed state later.
To fix that issue,  we have to add the ID of the VM into the state.

With the updated state / error handling of deploying an VW it's now possible to handle all errors that could be happen after creating an VM.

A `terraform destroy` was also working for that issue.

**Current state:**
```
No virtual machine informations saved

"resources": [
    {
       ...
    }
}
```
Azure view:
Screen Shot 2021-01-12 at 10 25 08

**State after the FIX:**
```
"resources": [
    {
        "mode": "managed",
        "type": "azurerm_linux_virtual_machine",
        "name": "example",
        "provider": "provider[\"registry.terraform.io/hashicorp/azurerm\"]",
        "instances": [
          {
            ...,
            "id": "/subscriptions/XXXXXXX-XXXX-XXXX-XXXXXXX-XXXXXXXXX/resourceGroups/test/providers/Microsoft.Compute/virtualMachines/example-machine",
            ...
          }
        ]
    }
}
```
Azure view:
Screen Shot 2021-01-12 at 10 25 08

Bg Tobi

### DIFF
--- a/azurerm/internal/services/compute/linux_virtual_machine_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_resource.go
@@ -490,7 +490,7 @@ func resourceLinuxVirtualMachineCreate(d *schema.ResourceData, meta interface{})
 	}
 
 	var vm compute.VirtualMachine
-	json.Unmarshal(vmResp, &vm)
+	_ = json.Unmarshal(vmResp, &vm)
 
 	if vm.ID == nil {
 		return fmt.Errorf("retrieving Linux Virtual Machine %q (Resource Group %q): `id` was nil", name, resourceGroup)
--- a/azurerm/internal/services/compute/linux_virtual_machine_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_resource.go
@@ -1,7 +1,9 @@
 package compute
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hey @HappyTobi 

Thanks for this PR.

Unfortunately the Azure API (and in particular the Compute API) misuses the "Failed" state to mean several things, including "this won't recover" and "this took longer than the hard-coded hypervisor timeout to boot, but may recover", "this has failed and will be automatically cleaned up" and "this is fine, but the internal healthcheck failed".

Whilst it appears that the "Failed" state should be Terminal - unfortunately since this state isn't reliable Terraform isn't able to guarantee if a particular state is recoverable or not (or if it'll recover, or if it's safe to apply changes) - as such we intentionally require the operator to decide how to proceed here to prevent causing further issues.

As such whilst I'd like to thank you for this contribution, since we're unable to guarantee the state is actually Failed (and so to prevent knock-on issues) unfortunately we're intentionally not tracking the Failed state at this time - and as such I'm going to close this PR for the moment. However as mentioned in #7516, I believe the root-cause for this will ultimately be solved by #7952, so would you mind subscribing to that issue for updates?

Thanks!

MSSedusch:
@tombuildsstuff this PR fixes the following situation:
1. Create a new VM with TF
2. The VM deployment fails for whatever reason but the VM resource was created
3. The VM is not in the TF state and a destroy would not delete which breaks a retry (that would include a TF destroy)

From what I understand, the Azure API guaranteees that the VM resource is created if the first HTTP PUT call returns 200, 201 or 202. Even if the long running operation is failing, the VM resource would still exist and IMHO TF would be responsible for the resource e.g. it should destroy the VM during a TF destroy.

PR #10163, opened 2021-01-13
New Resource: azurerm_template_spec

neil-yechenwei:
After investigated, I think there are two ways to submit arm template with template file.
One is using arm template file to implement it. Another one is using template spec to implement it. The first one has been implemented in `azurerm_resource_group_template_deployment`. The second one hasn’t been implemented. So before supporting Template Spec in `azurerm_resource_group_template_deployment`, I assume we have to implement `Template Spec API` in terraform first. 

The overview of Template spec:
https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-specs?tabs=azure-powershell

The rest api spec of Template spec:
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/resources/resource-manager/Microsoft.Resources/preview/2019-06-01-preview/templateSpecs.json


tombuildsstuff:
hey @neil-yechenwei 

Thanks for this PR.

As discussed offline @jackofallops has spent some time looking into this and realised that we need to take a different approach here and splitting this out into separate resources - as such whilst I'd like to thank you for this contribution, I'm going to close this in favour of #10163 which should allow this to ship.

Thanks!

PR #10164, opened 2021-01-13
dependencies: updating to v0.11.16 of github.com/Azure/go-autorest

tombuildsstuff:
This includes a fix for Log Analytics Clusters which returns a literal `null` as the response which caused the existing version of go-autorest to panic.


jackofallops (approved):
LGTM 👍

PR #10171, opened 2021-01-13
dependencies: updating to `2020-12-01` of `containerservice`

tombuildsstuff:
This PR updates the ContainerService package to API version  `2020-12-01` - but also includes an older API version (`2019-08-01`) where the `ListOrchestrator` call is coming from, which was historically present in the later swaggers as a backported call but isn't anymore.


tombuildsstuff:
Tests look good 👍

katbyte (approved):
LGTM 👍

PR #10181, opened 2021-01-13
New data source `azurerm_search_service`

### LINKED ISSUE #10153: Support for Azure Search Service Data Source
### Community Note

### Description

Currently you cannot add a Data Source of type azurerm_search_service. 

* azurerm_search_service

```hcl
data "azurerm_search_service" "ss_some_search_service" {
  name                = "ss-some-search-service"
  resource_group_name = "rg-myResourceGroup"
}
```

brunhil:
Fixes #10153

### DIFF
--- a/azurerm/internal/services/search/search_service_data_source.go
+++ b/azurerm/internal/services/search/search_service_data_source.go
@@ -2,7 +2,6 @@ package search
 
 import (
 	"fmt"
-	"log"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
@@ -25,7 +24,6 @@ func dataSourceSearchService() *schema.Resource {
 			"name": {
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: true,
 			},
 
 			"resource_group_name": azure.SchemaResourceGroupName(),
@@ -78,7 +76,6 @@ func dataSourceSearchService() *schema.Resource {
 			"identity": {
 				Type:     schema.TypeList,
 				Computed: true,
-				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"type": {
@@ -104,27 +101,22 @@ func dataSourceSearchService() *schema.Resource {
 
 func dataSourceSearchServiceRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*clients.Client).Search.ServicesClient
+	subscriptionID := meta.(*clients.Client).Account.SubscriptionId
 	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
 	defer cancel()
 
-	id, err := parse.SearchServiceID(d.Id())
-	if err != nil {
-		return err
-	}
+	id := parse.NewSearchServiceID(subscriptionID, d.Get("resource_group_name").(string), d.Get("name").(string))
 
 	resp, err := client.Get(ctx, id.ResourceGroup, id.Name, nil)
 	if err != nil {
 		if utils.ResponseWasNotFound(resp.Response) {
-			log.Printf("[INFO] Error reading Search Service %q - removing from state", d.Id())
-			d.SetId("")
 			return fmt.Errorf("Search Service %q (Resource Group %q) was not found", id.Name, id.ResourceGroup)
 		}
 
 		return fmt.Errorf("Error reading Search Service: %+v", err)
 	}
 
-	d.Set("name", id.Name)
-	d.Set("resource_group_name", id.ResourceGroup)
+	d.SetId(id.ID())
 
 	if props := resp.ServiceProperties; props != nil {
 		if count := props.PartitionCount; count != nil {
--- a/azurerm/internal/services/search/search_service_data_source.go
+++ b/azurerm/internal/services/search/search_service_data_source.go
@@ -117,7 +117,7 @@ func dataSourceSearchServiceRead(d *schema.ResourceData, meta interface{}) error
 		if utils.ResponseWasNotFound(resp.Response) {
 			log.Printf("[INFO] Error reading Search Service %q - removing from state", d.Id())
 			d.SetId("")
-			return nil
+			return fmt.Errorf("Search Service %q (Resource Group %q) was not found", id.Name, id.ResourceGroup)
 		}
 
 		return fmt.Errorf("Error reading Search Service: %+v", err)
--- a/azurerm/internal/services/search/search_service_data_source.go
+++ b/azurerm/internal/services/search/search_service_data_source.go
@@ -1,0 +1,159 @@
+package search
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/search/parse"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceSearchService() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceSearchServiceRead,
+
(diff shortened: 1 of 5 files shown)


azurerm/internal/services/search/search_service_data_source.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return err
+	}
+
+	resp, err := client.Get(ctx, id.ResourceGroup, id.Name, nil)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[INFO] Error reading Search Service %q - removing from state", d.Id())
+			d.SetId("")
+			return nil
```
tombuildsstuff, line 120: since this is a Data Source we should raise an error if this is not found rather than returning `nil` here - so could we make this:

```suggestion
			return fmt.Errorf("%s was not found", id)
```

(the format string for "id" will be "Search Service %q (Resource Group %q)")

azurerm/internal/services/search/search_service_data_source_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func TestAccDataSourceSearchService_basic(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_search_service", "test")
+	r := SearchServiceDataSource{}
+
+	data.DataSourceTest(t, []resource.TestStep{
+		{
+			Config: r.basic(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).Key("id").Exists(),
```
tombuildsstuff, line 23: we can remove this, since otherwise the resource won't be in the state/will raise an error (which'll be hit before this point)

```suggestion
```

website/docs/d/search_service.html.markdown
```diff
@@ -0,0 +1,85 @@
...
+
+* `query_keys` - A `query_keys` block as defined below.
+
+* `public_network_access_enabled` - Whether or not public network access is allowed for this resource. Defaults to `true`.
+
+* `partition_count` - The number of partitions which have been created.
+
+* `replica_count` - The number of replica's which have been created.
+
+-> **Note:** `partition_count` and `replica_count` can only be configured when using a `standard` sku.
```
tombuildsstuff, line 53: since this is a data source, this isn't configurable - so we can remove this (since the fields should get a default value either way)

```suggestion
```

website/docs/d/search_service.html.markdown
```diff
@@ -0,0 +1,85 @@
...
+
+* `id` - The ID of the Search Service.
+
+* `primary_key` - The Primary Key used for Search Service Administration.
+
+* `secondary_key` - The Secondary Key used for Search Service Administration.
+
+* `query_keys` - A `query_keys` block as defined below.
+
+* `public_network_access_enabled` - Whether or not public network access is allowed for this resource. Defaults to `true`.
```
tombuildsstuff, line 47: this is a Data Source which is read-only, so the default value doesn't apply here - could we also update the messaging to highlight this isn't configurable:

 ```suggestion
* `public_network_access_enabled` - Whether or not public network access is enabled for this Search Service.
```

File: website/docs/d/search_service.html.markdown [hunk 5]
```diff
@@ -0,0 +1,85 @@
...
+
+* `partition_count` - The number of partitions which have been created.
+
+* `replica_count` - The number of replica's which have been created.
+
+-> **Note:** `partition_count` and `replica_count` can only be configured when using a `standard` sku.
+
+* `identity` - An `identity` block as defined below.
+
+* `tags` - (Optional) A mapping of tags which should be assigned to the Search Service.
```
tombuildsstuff, line 57: (as above) since this isn't configurable can we update the wording to highlight that this is read only:

```suggestion
* `tags` - A mapping of tags which are assigned to the Search Service.
```

website/docs/d/search_service.html.markdown
```diff
@@ -0,0 +1,85 @@
...
+
+---
+
+A `identity` block supports the following:
+
+* `type` - The Type of Identity which should be used for the Search Service. At this time the only possible value is `SystemAssigned`.
```
tombuildsstuff, line 63: (as above)

```suggestion
* `type` - The type of Managed Identity which used for the Search Service.
```

tombuildsstuff (changes requested):
hey @brunhil 

Thanks for this PR :)

I've taken a look through and this is looking good, if we can fix up the comments and linting issues then this should otherwise be good to go 👍 

Thanks!

website/docs/d/search_service.html.markdown
brunhil: We don't actually set the tags on this data source so I'll just remove this line 😄

brunhil:
Thanks for the review @tombuildsstuff ! Let me know if there are any other changes you see

tombuildsstuff (changes requested):
hey @brunhil 

Thanks for pushing those changes - I've taken a look through and left some comments inline but if we can fix those up then this should be good to go 👍 

Thanks!

katbyte:
Test is also failing with 
```
------- Stdout: -------
=== RUN   TestAccDataSourceSearchService_basic
=== PAUSE TestAccDataSourceSearchService_basic
=== CONT  TestAccDataSourceSearchService_basic
        
        Error: Cannot parse Azure ID: parse "": empty url
        
          on /opt/teamcity-agent/temp/buildTmp/tf-test424173270/main.tf line 24:
          (source code not available)
        
        
--- FAIL: TestAccDataSourceSearchService_basic (136.98s)
FAIL
```

tombuildsstuff (approved):
LGTM - thanks for pushing those changes @brunhil

azurerm/internal/services/search/search_service_data_source_test.go
```diff
@@ -0,0 +1,46 @@
...
+	data.DataSourceTest(t, []resource.TestStep{
+		{
+			Config: r.basic(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).Key("replica_count").Exists(),
+				check.That(data.ResourceName).Key("partition_count").Exists(),
+				check.That(data.ResourceName).Key("primary_key").Exists(),
+				check.That(data.ResourceName).Key("secondary_key").Exists(),
+				check.That(data.ResourceName).Key("public_network_access_enabled").Exists(),
+				check.That(data.ResourceName).Key("identity.type").Exists(),
```
tombuildsstuff, line 28: this is a list so this needs to be:

```suggestion
				check.That(data.ResourceName).Key("identity.0.type").Exists(),
```

to access the element - else the test fails:

```
------- Stdout: -------
=== RUN   TestAccDataSourceSearchService_basic
=== PAUSE TestAccDataSourceSearchService_basic
=== CONT  TestAccDataSourceSearchService_basic
testing.go:684: Step 0 error: Check failed: Check 6/8 error: data.azurerm_search_service.test: Attribute 'identity.type' expected to be set
--- FAIL: TestAccDataSourceSearchService_basic (108.32s)
FAIL
```

but if we can fix that up then this should otherwise be good to go 👍

katbyte (changes requested):
Even after committing that change it the test is still failing:
```
------- Stdout: -------
=== RUN   TestAccDataSourceSearchService_basic
=== PAUSE TestAccDataSourceSearchService_basic
=== CONT  TestAccDataSourceSearchService_basic
    testing.go:684: Step 0 error: Check failed: Check 6/8 error: data.azurerm_search_service.test: Attribute 'identity.0.type' expected to be set
--- FAIL: TestAccDataSourceSearchService_basic (104.58s)
FAIL
```

brunhil:
Turns out I was missing the `identity` block which I added in the most recent commit. Hopefully things are good now @katbyte ! 🤞

azurerm/internal/services/search/search_service_data_source_test.go
```diff
@@ -0,0 +1,64 @@
...
+				check.That(data.ResourceName).Key("partition_count").Exists(),
+				check.That(data.ResourceName).Key("primary_key").Exists(),
+				check.That(data.ResourceName).Key("secondary_key").Exists(),
+				check.That(data.ResourceName).Key("public_network_access_enabled").Exists(),
+				check.That(data.ResourceName).Key("identity.0.type").Exists(),
+				check.That(data.ResourceName).Key("identity.0.principal_id").Exists(),
+				check.That(data.ResourceName).Key("identity.0.tenant_id").Exists(),
+			),
+		},
+		data.ImportStep(),
```
katbyte, line 33: We don't need this here
```suggestion
```

katbyte (approved):
Aside from one comment this LGTM 👍

PR #10189, opened 2021-01-14
UPDATE delete VMs that are in failed state

HappyTobi:
Hi @tombuildsstuff,

here is just a fix to delete VM's that are in failed state.

I also deleted the comments about "power off was saving costs, thats wrong see also the azure sdk comment: [power off](https://github.com/Azure/azure-sdk-for-go/blob/4a06a7dc9c3c720194d8967fbe1eb585246fb709/services/compute/mgmt/2020-06-30/compute/virtualmachines.go#L1409))

The scenario for that was that you import you VM's into your TF state and some of them are at a "Failed" state. Atm the VM's couldn't be delete with TF because the "power off" will fail.

With the small fix it was possible to clean the VM's.

Bg 
Tobi

### DIFF
--- a/azurerm/internal/services/compute/linux_virtual_machine_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_resource.go
@@ -1089,10 +1089,14 @@ func resourceLinuxVirtualMachineDelete(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf("retrieving Linux Virtual Machine %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
 	}
 
-	// If the VM was in a Failed state we can skip porwering off
+	// If the VM was in a Failed state we can skip powering off, since that'll fail
 	if strings.EqualFold(*existing.ProvisioningState, "failed") {
 		log.Printf("[DEBUG] Powering Off Linux Virtual Machine was skipped because the VM was in %q state %q (Resource Group %q).", *existing.ProvisioningState, id.Name, id.ResourceGroup)
 	} else {
+		//ISSUE: 4920
+		// shutting down the Virtual Machine prior to removing it means users are no longer charged for some Azure resources
+		// thus this can be a large cost-saving when deleting larger instances
+		// https://docs.microsoft.com/en-us/azure/virtual-machines/states-lifecycle
 		log.Printf("[DEBUG] Powering Off Linux Virtual Machine %q (Resource Group %q)..", id.Name, id.ResourceGroup)
 		skipShutdown := !meta.(*clients.Client).Features.VirtualMachine.GracefulShutdown
 		powerOffFuture, err := client.PowerOff(ctx, id.ResourceGroup, id.Name, utils.Bool(skipShutdown))
--- a/azurerm/internal/services/compute/linux_virtual_machine_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_resource.go
@@ -1089,19 +1089,21 @@ func resourceLinuxVirtualMachineDelete(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf("retrieving Linux Virtual Machine %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
 	}
 
-	// ISSUE: XXX
-	// shutting down the Virtual Machine prior to removing it means users are no longer charged for the compute
-	// thus this can be a large cost-saving when deleting larger instances
-	log.Printf("[DEBUG] Powering Off Linux Virtual Machine %q (Resource Group %q)..", id.Name, id.ResourceGroup)
-	skipShutdown := !meta.(*clients.Client).Features.VirtualMachine.GracefulShutdown
-	powerOffFuture, err := client.PowerOff(ctx, id.ResourceGroup, id.Name, utils.Bool(skipShutdown))
-	if err != nil {
-		return fmt.Errorf("powering off Linux Virtual Machine %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
+	// If the VM was in a Failed state we can skip porwering off
+	if strings.EqualFold(*existing.ProvisioningState, "failed") {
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -1089,19 +1089,21 @@ func resourceLinuxVirtualMachineDelete(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf("retrieving Linux Virtual Machine %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
 	}
 
-	// ISSUE: XXX
-	// shutting down the Virtual Machine prior to removing it means users are no longer charged for the compute
-	// thus this can be a large cost-saving when deleting larger instances
```
tombuildsstuff, line 1094: can we add this comment back in (but move it below) - details here: https://github.com/terraform-providers/terraform-provider-azurerm/issues/4920

azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -1089,19 +1089,21 @@ func resourceLinuxVirtualMachineDelete(d *schema.ResourceData, meta interface{})
...
-	log.Printf("[DEBUG] Powering Off Linux Virtual Machine %q (Resource Group %q)..", id.Name, id.ResourceGroup)
-	skipShutdown := !meta.(*clients.Client).Features.VirtualMachine.GracefulShutdown
-	powerOffFuture, err := client.PowerOff(ctx, id.ResourceGroup, id.Name, utils.Bool(skipShutdown))
-	if err != nil {
-		return fmt.Errorf("powering off Linux Virtual Machine %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
-	}
-	if err := powerOffFuture.WaitForCompletionRef(ctx, client.Client); err != nil {
-		return fmt.Errorf("waiting for power off of Linux Virtual Machine %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
+	// If the VM was in a Failed state we can skip porwering off
```
tombuildsstuff, line 1092:
```suggestion
	// If the VM was in a Failed state we can skip powering off, since that'll fail
```

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -1175,19 +1175,20 @@ func resourceWindowsVirtualMachineDelete(d *schema.ResourceData, meta interface{
 		return fmt.Errorf("retrieving Windows Virtual Machine %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
 	}
 
-	// ISSUE: XXX
-	// shutting down the Virtual Machine prior to removing it means users are no longer charged for the compute
-	// thus this can be a large cost-saving when deleting larger instances
```
tombuildsstuff, line 1180: (as above) can we add this back in - details are in https://github.com/terraform-providers/terraform-provider-azurerm/issues/4920

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -1175,19 +1175,20 @@ func resourceWindowsVirtualMachineDelete(d *schema.ResourceData, meta interface{
...
-	log.Printf("[DEBUG] Powering Off Windows Virtual Machine %q (Resource Group %q)..", id.Name, id.ResourceGroup)
-	skipShutdown := !meta.(*clients.Client).Features.VirtualMachine.GracefulShutdown
-	powerOffFuture, err := client.PowerOff(ctx, id.ResourceGroup, id.Name, utils.Bool(skipShutdown))
-	if err != nil {
-		return fmt.Errorf("powering off Windows Virtual Machine %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
-	}
-	if err := powerOffFuture.WaitForCompletionRef(ctx, client.Client); err != nil {
-		return fmt.Errorf("waiting for power off of Windows Virtual Machine %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
+	if strings.EqualFold(*existing.ProvisioningState, "failed") {
```
tombuildsstuff, line 1178: can we add a matching comment here:

```suggestion
	// If the VM was in a Failed state we can skip powering off, since that'll fail
	if strings.EqualFold(*existing.ProvisioningState, "failed") {
```

tombuildsstuff (changes requested):
hi @HappyTobi 

Thanks for this PR - I've taken a look through and left some comments inline but if we can fix those up then we should be able to run the tests and get this merged 👍 

Thanks!

azurerm/internal/services/compute/linux_virtual_machine_resource.go
HappyTobi: Fine for me to add a comment but I try to rewrite the comment because you don't save money by the vm itself only for the disk etc. (https://docs.microsoft.com/en-us/azure/virtual-machines/states-lifecycle)

azurerm/internal/services/compute/linux_virtual_machine_resource.go
HappyTobi: Thx I will change that

azurerm/internal/services/compute/windows_virtual_machine_resource.go
HappyTobi: will add the same comment on both vm files

azurerm/internal/services/compute/windows_virtual_machine_resource.go
HappyTobi: Yes I will do that, thx

HappyTobi:
Hi @tombuildsstuff,

I changed the comments hope you like that.

Bg

tombuildsstuff (approved):
LGTM - thanks for pushing those changes @HappyTobi

tombuildsstuff:
Tests look good 👍

PR #10196, opened 2021-01-15
user_assigned_identity: add state migration for ID casing

favoretti:
Old `user_assigned_identity` are saved in the state with `resourcegroups` all in lowercase, forcing all resource that use it's `.id` property reference do a lowercasing dance every plan/apply.
This should fix the problem.

### DIFF
--- a/azurerm/internal/services/msi/migration/user_assigned_identity_V0_to_V1.go
+++ b/azurerm/internal/services/msi/migration/user_assigned_identity_V0_to_V1.go
@@ -8,7 +8,6 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
-	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/msi/parse"
 )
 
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/msi/migration/user_assigned_identity_V0_to_V1.go
```diff
@@ -0,0 +1,62 @@
...
+			"client_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func userAssignedIdentityUpgradeV0ToV1(rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
+	subscriptionId := meta.(*clients.Client).Account.SubscriptionId
```
tombuildsstuff, line 53: there's one issue here where we assume the subscription being used is the same one used to provision the resource, which is the case 99% of the time, but may not be during state migrations if someone's using the Azure CLI and has updated their default) - as such instead it'd be worth parsing the existing ID which'll pull out the SubscriptionID being used:

```suggestion
	oldId := rawState["id"].(string)
	id, err := parse.UserAssignedIdentityID(oldId)
	if err != nil {
		return rawState, err
	}
	
	newId := id.ID()
	log.Printf("Updating `id` from %q to %q", oldId, newId)
	rawState["id"] = newId
	return rawState, nil
```

arguably there's a separate issue here insofar as we're not updating the Subscription used for the clients on demand to account for this, but that's a much larger issue we're not likely to fix anytime soon

tombuildsstuff (changes requested):
hey @favoretti 

Thanks for this PR :)

Taking a look through this mostly LGTM - if we can update the State Migration to parse the existing ID to account for one subtle bug then this should otherwise LGTM 👍 

Thanks!

tombuildsstuff (approved):
LGTM 👍

tombuildsstuff:
Tested locally looks good 👍

Issue #10202, opened 2021-01-15
Support for Visual Studio Code DevContainer


ross-p-smith:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Provide a .devcontainer folder in the source repository so that it makes it easier for contributors to get a working environment on their development machine. This would be a docker file with all the pre-requisites defined in it that could just be spun up when required. This would mean that the version of Go and other libraries could be kept in step with the repo.

### New or Affected Resource(s)

No new resources

### References

https://code.visualstudio.com/docs/remote/containers

* #0000

tombuildsstuff:
hey @ross-p-smith 

Thanks for opening this issue.

Whilst we could look to add integrations and support for specific IDE's - we've intentionally avoided doing so in favour of more generalized tooling which can be integrated into anything.

Where DevContainers are a feature only supported by VSCode (and not by other IDE's such as [Goland](https://www.jetbrains.com/go/) or [Chime](https://www.chimehq.com)) - since nobody on the team is using VSCode we're concerned that were we to add support for this, this'd fall out of sync pretty quickly and become a hinderance rather than a help. Notably when contributing on Windows there's also some additional challenges for users using WSL, WSL2 and Git Bash, where the setup can be quite different - whilst using Docker can help unfortunately it's not a panacea.

For what it's worth most of the team are using [Goland](https://www.jetbrains.com/go/) and we're also taking the same approach here, by [intentionally ignoring the `.idea` directory in the `.gitignore`](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/.gitignore#L24) (we're also [doing the same for the `.vscode` folder for the same reason](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/.gitignore#L31)).

As mentioned above since we're trying to use more generalized tooling where possible, in this instance we're using a Makefile (and optionally Docker) - which should work cross-platform (via `make tools` to get the tools, `make build` or `make docker-build` to build locally/in docker and `make fmt` to run the formatters) - and documenting these commands in the README.

[...]

PR #10204, opened 2021-01-15
Added devcontainer

ross-p-smith:
This builds a golang docker image 1.15 and installs the necessary tools you need to debug and run the azurerm provider inside a remote container using Visual Studio Code - https://code.visualstudio.com/docs/remote/containers.

1. Install the Remote Development extension in vscode
2. Clone the azurerm provider from GitHub
3. Open in Visual Studio Code
4. ReOpen in a container when prompted - vscode will build the docker image for you and mount your source code inside the container
5. DevContainer will also run "make tools" for you after the container is created.

Issue: #10202

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,67 @@
+FROM golang:1.15-buster
+
+# Avoid warnings by switching to noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
+
+# This Dockerfile adds a non-root 'vscode' user with sudo access. However, for Linux,
+# this user's GID/UID must match your local user UID/GID to avoid permission issues
+# with bind mounts. Update USER_UID / USER_GID if yours is not 1000. See
+# https://aka.ms/vscode-remote/containers/non-root-user for details.
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+# Terraform and tflint versions
+ARG TERRAFORM_VERSION=0.14.4
+
+ENV GO111MODULE=on
+
+# Configure apt, install packages and tools
+RUN apt-get update \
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
hey @ross-p-smith 

Thanks for this PR.

Whilst we could look to add integrations and support for specific IDE's - we've intentionally avoided doing so in favour of more generalized tooling which can be integrated into anything.

Where DevContainers are a feature only supported by VSCode (and not by other IDE's such as [Goland](https://www.jetbrains.com/go/) or [Chime](https://www.chimehq.com)) - since nobody on the team is using VSCode we're concerned that were we to add support for this, this'd fall out of sync pretty quickly and become a hinderance rather than a help. Notably when contributing on Windows there's also some additional challenges for users using WSL, WSL2 and Git Bash, where the setup can be quite different - whilst using Docker can help unfortunately it's not a panacea.

For what it's worth most of the team are using [Goland](https://www.jetbrains.com/go/) and we're also taking the same approach here, by [intentionally ignoring the `.idea` directory in the `.gitignore`](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/.gitignore#L24) (we're also [doing the same for the `.vscode` folder for the same reason](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/.gitignore#L31)).

As mentioned above since we're trying to use more generalized tooling where possible, in this instance we're using a Makefile (and optionally Docker) - which should work cross-platform (via `make tools` to get the tools, `make build` or `make docker-build` to build locally/in docker and `make fmt` to run the formatters) - and documenting these commands in the README.

We do have plans to add more specific documentation here (partly tracked in #2688 - and others via some internal things) - since we're not planning on adding integrations with specific IDE's whilst I'd like to thank you for this contribution, I'm going to close this PR for the moment.

Thanks!

PR #10206, opened 2021-01-15
azurerm_kubernetes_cluster: support aad-pod-identity

### LINKED ISSUE #9885: Support for Pod Identity Add On for AKS
### Community Note

### Description

Please support the Pod Identity Add-On for AKS

* azurerm_kubernetes_cluster

```hcl
resource "azurerm_kubernetes_cluster" "aks" {

  addon_profile {
    kube_dashboard {
      enabled = true
    }
    azure_policy {
      enabled = true
    }

    pod_identity {
      enabled = true
      principal_id = azurerm_user_assigned_identity.podid.principal_id
    }

  }

}
```

### References

* https://docs.microsoft.com/en-us/azure/aks/use-azure-ad-pod-identity

feliperoveran:
Support enabling ad-pod-identity, creating pod identities, and pod exceptions.

Fixes #9885

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_cluster_resource.go
+++ b/azurerm/internal/services/containers/kubernetes_cluster_resource.go
@@ -215,6 +215,88 @@ func resourceArmKubernetesCluster() *schema.Resource {
 				},
 			},
 
+			"pod_identity_profile": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
+						"user_assigned_identities": {
+							Type:     schema.TypeList,
+							Optional: true,
+							// AKS limitation
+							MaxItems: 50,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"name": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+									"namespace": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+									"identity": {
+										Type:     schema.TypeList,
+										Optional: true,
+										MaxItems: 1,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"resource_id": {
+													Type:         schema.TypeString,
+													Required:     true,
+													ValidateFunc: azure.ValidateResourceID,
+												},
+												"client_id": {
+													Type:         schema.TypeString,
+													Required:     true,
+													ValidateFunc: validation.IsUUID,
+												},
+												"object_id": {
+													Type:         schema.TypeString,
+													Required:     true,
+													ValidateFunc: validation.IsUUID,
+												},
+											},
+										},
+									},
+								},
+							},
+						},
+						"user_assigned_identity_exceptions": {
+							Type:     schema.TypeList,
+							Optional: true,
+							// AKS limitation
+							MaxItems: 50,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"name": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+									"namespace": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+									"pod_labels": {
+										Type:     schema.TypeMap,
+										Optional: true,
+										Elem: &schema.Schema{
+											Type: schema.TypeString,
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+
 			"linux_profile": {
 				Type:     schema.TypeList,
 				Optional: true,
@@ -690,6 +772,9 @@ func resourceArmKubernetesClusterCreate(d *schema.ResourceData, meta interface{}
 	dnsPrefix := d.Get("dns_prefix").(string)
 	kubernetesVersion := d.Get("kubernetes_version").(string)
 
+	podIdentityProfileRaw := d.Get("pod_identity_profile").([]interface{})
+	podIdentityProfile := expandManagedClusterPodIdentityProfile(podIdentityProfileRaw)
+
 	linuxProfileRaw := d.Get("linux_profile").([]interface{})
 	linuxProfile := expandKubernetesClusterLinuxProfile(linuxProfileRaw)
 
@@ -767,6 +852,7 @@ func resourceArmKubernetesClusterCreate(d *schema.ResourceData, meta interface{}
 			LinuxProfile:           linuxProfile,
 			WindowsProfile:         windowsProfile,
 			NetworkProfile:         networkProfile,
+			PodIdentityProfile:     podIdentityProfile,
 			NodeResourceGroup:      utils.String(nodeResourceGroup),
 		},
 		Tags: tags.Expand(t),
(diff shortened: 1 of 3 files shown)


feliperoveran:
@tombuildsstuff this should be good to go :)

the only part that I wasn't sure about was since enabling ad-pod-identity on cluster creation is not supported, should we fail or just filter those fields? I've added that limitation to the docs, but cluster creation will fail if `pod_identity_profile` is configured (updates are good, of course).

how does the provider handle such scenarios? please let me know and I'd be glad to change anything :)

dhirschfeld:
> cluster creation will fail if `pod_identity_profile` is configured

That seems pretty problematic for automation scenarios 🙁 Is that just a limitation of the *"preview"* feature?

feliperoveran:
@dhirschfeld yea, it's an Azure limitation :/ the API errors out saying that ad pod identity cannot be enabled on cluster creation

Tenseiga:
Interesting i dont see any where in az docs that mentions abt kubenet cni not supported.

jrauschenbusch:
@feliperoveran It should also work directly during the cluster creation, as otherwise the azure cli would also just support it within an update operation, which is not the case. If it really just works within an update operation, it would be required to update it within the terraform create operation. Otherwise it would not be useable from an automation context.

@Tenseiga By trying to create an AKS cluster with Pod Identities enabled, i had to also set azure cni for the network component. This is also mentioned in the blog post here: https://blog.baeke.info/2020/12/09/azure-ad-pod-managed-identities-in-aks-revisited/. Also the az cli command within the Azure docs sets the network plugin to azure cni.
https://docs.microsoft.com/en-us/azure/aks/use-azure-ad-pod-identity#create-an-aks-cluster-with-managed-identities

feliperoveran:
@jrauschenbusch it might be due to the fact it's a preview feature.

enabling the extension is supported on cluster creation, but creating the identities/exceptions is not.

jrauschenbusch:
@feliperoveran Ok. Then i got you wrong, that enabling the pod identity feature during cluster creation is completely impossible. But now it sounds like that just the binding of the identities is not possible during the cluster creation. If i'm not totally wrong, this should also be possible using the added Kubernetes CRDs. I'm not sure if MS will change the API behavior if the feature will get GA.

ekhaydarov:
Apologies for not adding to the discussion but just wondering if someone would not mind letting me know a very loose timeline on when this can be expected to make it into a provider release?

djsly:
should we move the adding policies to a cluster as a separate resource ? a bit like the agent_pools are managed ?

this way we have a dedicate resource for managing the AKS pod identities assignments outside of the kubenertes_cluster resource, the kubernetes_resource would be managing the add-on toggle only.

jrauschenbusch:
@djsly If the internal AKS ARM implementation remains as it is this should be the way to go.

djsly:
@jrauschenbusch  I will try to get the information from MSFT to see what is the expect behaviour after the feature reaches GA. more to come.

feliperoveran:
FYI I'll work on moving the identities to their own TF resource in the next couple of days

azurerm/internal/services/containers/kubernetes_cluster_resource.go
```diff
@@ -215,6 +215,88 @@ func resourceArmKubernetesCluster() *schema.Resource {
...
+									},
+									"identity": {
+										Type:     schema.TypeList,
+										Optional: true,
+										MaxItems: 1,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"resource_id": {
+													Type:         schema.TypeString,
+													Required:     true,
```
whiskeysierra, line 252: All of those three ids are optional in Azure. I guess one of them is enough.

whiskeysierra:
> If i'm not totally wrong, this should also be possible using the added Kubernetes CRDs.

Has anyone tried this?

jrauschenbusch:
@whiskeysierra Not yet. But since you can also install everything via a simple HELM deployment (https://github.com/Azure/aad-pod-identity), I strongly believe that this must be possible.

But unfortunately i have bad news for the release of the current version of this feature. I've asked the team when they will GA the feature. The answer is [here](https://github.com/Azure/AKS/issues/1480#issuecomment-780620631).

TL;DR: V1 will not go GA. V2 will go GA at the end of the year.

whiskeysierra:
Thanks for the link and the follow up.

Actually, there are three versions of AAD Pod Identity then:
1. Open-Source version (https://github.com/Azure/aad-pod-identity)
2. V1, official Azure product, currently in preview
3. V2, upcoming version, as described in the answer from Microsoft

What does that mean for the terraform integration? Will it support v1 until June or will this wait until V2 goes GA?

ekhaydarov:
What does this mean for this PR? its taking over a year for azure to release a stable pod-identity solution. especially with an awkward workaround to pod identity as well as killing off the pod security policy, all of which is not even reflected in terraform provider documentation

Understand the provider team is powerless to what the AKS team is doing but its just frustrating trying to get AKS to work like GKE or EKS does security wise.

Seems like the forced practical solution here is to use the OS helm chart and not even bother with whatever AKS team is doing for over year

whiskeysierra:
> Seems like the forced practical solution here is to use the OS helm chart and not even bother with whatever AKS team is doing for over year

The problem with the open source version (https://github.com/Azure/aad-pod-identity) is that you have to choice between to bad options:

1. MIC (managed identity controller) which takes up to 60 seconds to dynamically assign an identity to the VMSS.
2. Managed mode which breaks whenever you auto scale the node pool or do node updates.

I'm using an ARM template now instead of the azurerm_kubernetes_cluster resource. But it's definitely not fun.

mazen423:
Do we have updates regarding this request ?

jozamp:
We're just starting to move our IaC to terraform and will also will want to use pod identity. Our current structure would actually enjoy having the identity binding as separate configurable resources dependent on the aks resource. Our micro-services are expressed as modules and it would be nice to encapsulate the services identity in its own module as opposed to the "main" aks provider config.

jozamp:
> > If i'm not totally wrong, this should also be possible using the added Kubernetes CRDs.
> 
> Has anyone tried this?

I just tried enabling pod identity with azcli aks-preview --enable-pod-identity, then using kubernetes-alpha/manifest and got  "No valid OpenAPI definition" trying to plan.

I'm going go with the oss aad pod identity chart until this settles out.

njuCZ:
@feliperoveran Hi, May I ask are you working on moving identities into separate TF resources? I am glad if there is anything I could help

njuCZ:
@feliperoveran this feature is needed a little urgent. Since no further response, would you mind that I will take over this PR and finish the rest part (a seperate resource) ?

djsly:
@njuCZ  this won't be possible, we had a discussion with the terraform provider owner and MSFT and they are already deprecating the v1 api, and the design doesnt fit with terraform. 

CC: @feliperoveran  / @tombuildsstuff

ekhaydarov:
OK people will just manually use this feature in the UI? Some comms from MS would be nice about what their plan is for this or some kind of roadmap. I guess we shot ourselves in the foot by choosing to use pod identity in the first place which is an alpha feature. i guess its a lesson not to use MS feature until they are GA. Even then we are filing bugs quite often

feliperoveran:
sorry, I missed all the notifications :(

Summary of the discussion on Slack:

image

Slack thread URL: https://terraform-azure.slack.com/archives/CJD3VK9Q9/p1613082979127400

In the end, there is no good solution and we opted to wait for the next version of AAD Pod Identity.

For those trying to use the CRDs and CRs, it doesn't work unless you are not using the managed version of AAD Pod Identity. Otherwise, the identities are never picked up by AKS.

feliperoveran:
@ekhaydarov you can use the az cli to assign the identities (and maybe automate that), but the whole managed identity setup is super clunky

jackofallops:
I believe this has been superseded by #11492, released in `v2.64.0` so I'm going to close it for now.

njuCZ:
@jackofallops just point out that this feature is not included in terraform because the current swagger design about pod identity will be deprecated soon and new design will be drafted soon, so Tom has reverted that RP by [#12225](https://github.com/terraform-providers/terraform-provider-azurerm/pull/12225).

jackofallops:
Thanks @njucz - Given that there will need to be a new PR on that new design, I'm going to leave this one closed. Will you be picking up that work when the service team make the new one available?

PR #10241, opened 2021-01-19
test fix: securitycenter contact

njuCZ:
in the func `RunTestsInSequence`, it just iterate the map and run the testcase. But for each test case, it will invoke `data.ResourceTest`, then invoke `td.runAcceptanceTest(t, testCase)`, then invoke `resource.ParallelTest(t, testCase)`, it will set the `Parallel` flag to `true`.

in go testing framework, it will use goroutine to run test case

So the func `RunTestsInSequence` could not work properly.

A simple test case to confirm:
```
func TestParallel(t *testing.T) {
	tests := map[string]map[string]func(t *testing.T){
		"test": {
			"test1": func(t *testing.T) {
				t.Parallel()
				log.Println("test1 start")
				time.Sleep(10 * time.Second)
				log.Println("test1 end")
			},
			"test2": func(t *testing.T) {
				t.Parallel()
				log.Println("test2 start")
				time.Sleep(10 * time.Second)
				log.Println("test2 end")
			},
		},
	}
	for group, m := range tests {
		m := m
		t.Run(group, func(t *testing.T) {
			for name, tc := range m {
				tc := tc
				t.Run(name, func(t *testing.T) {
					tc(t)
				})
			}
		})
	}
}
```

-----------------------------------------------------------------

have already been fixed in PR #10924 
update the PR to fix the remaining test case that needs to be run in sequence

### DIFF
--- a/azurerm/internal/acceptance/testcase.go
+++ b/azurerm/internal/acceptance/testcase.go
@@ -4,10 +4,9 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/terraform-providers/terraform-provider-azuread/azuread"
-
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/terraform-providers/terraform-provider-azuread/azuread"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance/helpers"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance/testclient"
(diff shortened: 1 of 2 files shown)


katbyte (changes requested):
@njuCZ - it seems like we've added bits of this in #10924 - could we refactor this to use the new bits and just have the `DataSourceTestInSequence()` call the new td.runSequentialTest function? thanks

njuCZ:
@katbyte have updated this PR. only the security center contact test case needs to be updated. I have modified the PR description and title

image

katbyte (approved):
LGTM 👍

PR #10264, opened 2021-01-21
`azurerm_kusto_cluster` - supports double_encryption_enabled, systemAssignedIdentity and systemAssignedUserAssignedIdentity

njuCZ:
support new fields: "double_encryption_enabled"
enhancement for `identity` block: support systemAssignedIdentity and systemAssignedUserAssignedIdentity
once more validation item for sku name

the acctest "TestAccKustoCluster_zones" failed either in master branch, I have emailed the service team

acording to https://docs.microsoft.com/en-us/azure/data-explorer/double-encryption?tabs=portal
> Enabling double encryption is only possible during cluster creation.
> Once infrastructure encryption is enabled on your cluster, you can't disable it.

We could infer that it's a forcenew field

### DIFF
--- a/azurerm/internal/services/kusto/identity.go
+++ b/azurerm/internal/services/kusto/identity.go
@@ -1,6 +1,8 @@
 package kusto
 
 import (
+	"fmt"
+
 	"github.com/Azure/azure-sdk-for-go/services/kusto/mgmt/2020-09-18/kusto"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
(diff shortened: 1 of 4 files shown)


njuCZ:
the website-test CI has been removed. It seems reopen the PR could not trigger the CI

azurerm/internal/services/kusto/identity.go
```diff
@@ -22,44 +24,63 @@ func schemaIdentity() *schema.Schema {
...
+						string(kusto.IdentityTypeUserAssigned),
+						string(kusto.IdentityTypeSystemAssignedUserAssigned),
 					}, true),
 					DiffSuppressFunc: suppress.CaseDifference,
 				},
+
+				"identity_ids": {
+					Type:     schema.TypeSet,
+					Optional: true,
+					Elem: &schema.Schema{
```
katbyte, line 37: should this still not also be computed?

katbyte (changes requested):

azurerm/internal/services/kusto/identity.go
njuCZ: @katbyte this is not a computed field. when the type is `SystemAssigned, UserAssigned`, users could set this property to pass the user managed identity. In this PR, I also add such an example

an example in VM: https://github.com/terraform-providers/terraform-provider-azurerm/blob/2598a62911c599278f06cf3fe9b57224e6bd3339/azurerm/internal/services/compute/virtual_machine.go#L70

katbyte: 👍

katbyte (approved):
Thanks @njuCZ - LGTM 👍

PR #10265, opened 2021-01-21
identity: adding a helper package for Managed Identities

tombuildsstuff:
The intention here is to make this code and convention more reusable, since the `identity` block should be treated in a common manner across different resources.

That is, the presence of an `identity` block should indicate that an identity is attached - and the absence should indicate that no identity should be assigned.

This commit introduces the foundation for this - and an implementation for System Assigned Identities - which maps to the following schema:

```
resource "azurerm_example" "example" {
  identity {
    type = "SystemAssigned"
  }
}
```

With documentation showing how to use this to map to and from the Azure SDK types - since unfortunately the Azure SDK isn't using common base types here.

This PR also switches the `azurerm_spring_cloud_app` resource over to using this new helper function.

### DIFF
--- a/azurerm/internal/identity/README.md
+++ b/azurerm/internal/identity/README.md
@@ -1,0 +1,63 @@
+# Identity
+
+This package contains helpers for working with Managed Identities.
+
+Azure supports up to 4 different combinations of Managed Identities:
+
+* `None` - where no Managed Identity is available/configured for this Azure Resource.
+* `SystemAssigned` - where Azure will generate a Managed Identity (Service Principal) for this Azure Resource.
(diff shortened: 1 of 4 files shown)


tombuildsstuff:
```
$ TF_ACC=1 go test -v ./azurerm/internal/services/springcloud/... -run=TestAccSpringCloudService_complete -timeout=60m         (r/identity)
=== RUN   TestAccSpringCloudService_complete
=== PAUSE TestAccSpringCloudService_complete
=== CONT  TestAccSpringCloudService_complete
--- PASS: TestAccSpringCloudService_complete (456.89s)
```

manicminer (approved):
Awesome, LGTM 👍

tombuildsstuff:
Tests pass:

Screenshot 2021-01-22 at 10 19 36

PR #10268, opened 2021-01-21
Adding new `application_gateway` data_source

petems:
Closes https://github.com/terraform-providers/terraform-provider-azurerm/issues/5030

Draft for now, as I want to make sure I'm making the data_source correctly, after that I'll add the rest of the attributes

### DIFF
--- a/azurerm/internal/services/network/application_gateway_data_source.go
+++ b/azurerm/internal/services/network/application_gateway_data_source.go
@@ -2,16 +2,21 @@ package network
 
 import (
 	"fmt"
-	"log"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-05-01/network"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/identity"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/location"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/parse"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
+
+type applicationGatewayDataSourceIdentity = identity.UserAssigned
 
 func dataSourceApplicationGateway() *schema.Resource {
 	return &schema.Resource{
@@ -31,6 +36,8 @@ func dataSourceApplicationGateway() *schema.Resource {
 
 			"resource_group_name": azure.SchemaResourceGroupNameForDataSource(),
 
+			"identity": applicationGatewayDataSourceIdentity{}.SchemaDataSource(),
+
 			"tags": tags.SchemaDataSource(),
 		},
 	}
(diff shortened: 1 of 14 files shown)


tombuildsstuff (changes requested):
hey @petems

Thanks for opening this PR.

I've taken a look through and this is off to a good start - I've taken a look through and left some comments inline, as mentioned offline this'll need to be registered here to be usable in the provider: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/network/registration.go#L24

Thanks!

azurerm/internal/services/network/registration.go
```diff
@@ -21,6 +21,7 @@ func (r Registration) WebsiteCategories() []string {
 // SupportedDataSources returns the supported Data Sources supported by this Service
 func (r Registration) SupportedDataSources() map[string]*schema.Resource {
 	return map[string]*schema.Resource{
+		"azurerm_application_gateway":                       dataSourceApplicationGateway(),
```
petems, line 24: @tombuildsstuff Is this correct?

azurerm/internal/services/network/application_gateway_data_source_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (AppGatewayDataSource) complete(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+%s
+
+data "azurerm_application_gateway" "test" {
+  resource_group_name = azurerm_application_gateway.test.resource_group_name
+  name                = azurerm_application_gateway.test.name
+}
+`, AppGatewayDataSource{}.complete(data))
```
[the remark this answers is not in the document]
petems: Done! 👍

azurerm/internal/services/network/application_gateway_data_source.go
```diff
@@ -0,0 +1,68 @@
...
+			return fmt.Errorf("Error: Application Security Group %q (Resource Group %q) was not found", name, resourceGroup)
+		}
+
+		return fmt.Errorf("Error making Read request on Application Security Group %q (Resource Group %q): %+v", name, resourceGroup, err)
+	}
+
+	d.SetId(*resp.ID)
+
+	d.Set("name", resp.Name)
+	d.Set("resource_group_name", resourceGroup)
```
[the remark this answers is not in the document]
petems: Good point, fixed

petems:
I've pulled most of the logic from the resource, so there's a lot of duplicted code here, is there value in refactoring this into a shared librariy? If so, is there an example out there?

azurerm/internal/services/network/registration.go
katbyte: Yep this is!

katbyte:
@petems - we're going to need to flip all the properties to computed as this is a data source

azurerm/internal/services/network/application_gateway_data_source.go
```diff
@@ -0,0 +1,181 @@
...
+			return nil
+		}
+
+		return fmt.Errorf("Error making Read request on Application Gateway %s: %+v", name, err)
+	}
+
+	if location := applicationGateway.Location; location != nil {
+		d.Set("location", azure.NormalizeLocation(*location))
+	}
+	d.Set("zones", applicationGateway.Zones)
```
tombuildsstuff, line 65: this data source only defines `name, location, resource_group_name, tags` above, these other fields don't exist (so will raise an error) - can we remove them, or add the additional fields to the schema?

azurerm/internal/services/network/application_gateway_data_source_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (AppGatewayDataSource) basic(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+%s
+
+data "azurerm_application_gateway" "test" {
+  resource_group_name = azurerm_application_gateway.test.resource_group_name
+  name                = azurerm_application_gateway.test.name
+}
+`, AppGatewayResource{}.basic(data))
```
tombuildsstuff, line 37: sorry, my bad - the type name for this is

```suggestion
`, ApplicationGatewayResource{}.basic(data))
```

tombuildsstuff (changes requested):
hey @petems 

Thanks for pushing those changes - I've taken a look through and left some comments inline, if we can get those sorted (and subsequently add/generate some documentation, there's details on how to generate that in the readme) - then we should be able to run the tests here 👍 

Thanks!

tombuildsstuff:
Tests look good 👍

PR #10272, opened 2021-01-21
Adding note about databricks public/private subnets

allantargino:
When integrating databricks into an existing VNET, both required subnets (public and private) must be delegated to `Microsoft.Databricks/workspaces`


katbyte (approved):
Thanks @allantargino - LGTM 👍

PR #10280, opened 2021-01-22
datafactory & keyvault: removing direct usages on `acceptance.AzureProvider`

tombuildsstuff:
This PR removes direct assertions on `acceptance.AzureProvider` by switching to use Check funcs.

This also introduces a new assertion function to validate upon a json value stored in the state, rather than checking the value in the API (since this is being read back)


jackofallops (approved):

PR #10283, opened 2021-01-22
azurerm_redis_cache - support multiple zone availability 

michaelelleby:
Azure Cache for Redis supports deploying to multiple availability zones. The Terraform provider should support this as well.


katbyte (approved):
Thanks @michaelelleby - this LGTM 👍

PR #8801, opened 2020-10-08
[WIP] Added JitNetworkAccessPolicy basic implementation

PaulDFoster:
Adds the Azure just-in-time network access policy setting capability for virtual machine ports documented in issue [#3661 ](https://github.com/terraform-providers/terraform-provider-azurerm/issues/3661)

Implemented as part of the Security Center resources. The acceptance test demonstrates the implementation of the JIT network access policies on the required Network Security Group (NSG) associated with a subnet used by the target virtual machine.  The JIT Network Access Policy REST API creates a new security rule on the NSG which currently trips up Terraform state management. 

Expert guidance would be appreciated on how to best resolve this. E,G, Make functionality part of the NSG resource provider

Implemented in this draft PR for the JIT Network Access Policy:
1. Custom client to accommodate the passing of a location parameter (default client is hardcoded to Global)
2. Parse functionality
3. Basic acceptance test
4. Example Terraform including the *required* virtual machine, subnet, network security group dependency
5. Resource implementation using simple schema:

```
resource "azurerm_jit_network_access_policies" "test-jit-network-access" {
	name = "default-test"
	asc_location = "northeurope"
	resource_group_name = "test-jit"
	kind = "Basic"
	virtual_machines {
		name = "test-jit-vm-1"
		ports {
			port = 22
			protocol = "*"
			allowed_source_address_prefix = "*"
			max_request_access_duration = "PT3H"
		}	
	}
  }

PR #10326, opened 2021-01-27
new resource "azurerm_jit_network_access_policy"

### LINKED ISSUE #3661: Enable just-in-time network access policies for VMs
### Community Note

### Description

Just in time access is available to enable time-based access to dedicated virtual machines on configurable ports. 

When just-in-time is enabled, Security Center locks down inbound traffic to your Azure VMs by creating an NSG rule. You select the ports on the VM to which inbound traffic will be locked down. These ports are controlled by the just-in-time solution. See https://docs.microsoft.com/en-us/azure/security-center/security-center-just-in-time. 

I am not sure if this should be part of network security rules (or network security groups) or the virtual machine as this is deeply linked to both resources. As `Set-AzJitNetworkAccessPolicy` is attached to the VM this might be more suitable as JIT policy attached directly to a virtual machine instead of network group.

See `Set-AzJitNetworkAccessPolicy` https://docs.microsoft.com/en-us/powershell/module/az.security/set-azjitnetworkaccesspolicy?view=azps-2.2.0.

* azurerm_virtual_machine
* azurerm_network_security_group
* azurerm_network_security_rule

```hcl
resource "azurerm_virtual_machine" "test" {
  name                  = "${var.prefix}-vm"
  location              = "${azurerm_resource_group.main.location}"
  resource_group_name   = "${azurerm_resource_group.main.name}"
  network_interface_ids = ["${azurerm_network_interface.main.id}"]
  vm_size               = "Standard_DS1_v2"
}


resource "azurerm_virtual_machine_jit_network_access_policy" "test" {
  name                 = "default"
  location             = azurerm_resource_group.main.location
  resource_group_name  = azurerm_resource_group.main.name
  virtual_machine_name = azurerm_virtual_machine.test.name
  kind                 = "Basic"

  jit_policy {
    port                        = 22
    protocol                    = "*"
    source_address_prefix       = "*"
    max_request_access_duration = "PT3H"
  }

  jit_policy {
    port                        = 3389
    protocol                    = "*"
    source_address_prefix       = "*"
    max_request_access_duration = "PT3H"
  }
}
```

### References

* https://docs.microsoft.com/en-us/powershell/module/az.security/set-azjitnetworkaccesspolicy?view=azps-2.2.0

njuCZ:
fix #3661

thanks @PaulDFoster for the initial contribution, this PR is on top of his PR: #8801

doc: https://docs.microsoft.com/en-us/azure/security-center/just-in-time-explained

two points that needs to pay attention
1. this resource needs a `location` property which needs to be same with vm, but in the go sdk, we could not pass the `location` parameter to any CURD method. The go sdk just use the `location` property from the client struct. So in every time, we should initialize a new jitNetworkAccessPolicyClient
2. the property `allowedSourceAddressPrefixes` and `allowedSourceAddressPrefix` are mutual exclusive. `allowedSourceAddressPrefixes` allows multiple, and `allowedSourceAddressPrefix` only allows one. In the implementation, I just expose "allowed_source_address_prefixes" in the schema, and convert based on the element array size


mark-strasser-nnl:
Awesome work!  Thanks!

I'll file a separate issue for this, but note that if a JIT access request is in effect, re-running `terraform apply` wipes out the JIT security rules added to the NSG.   Otherwise, Terraform (as expected) updates the NSG to remove the rules added by the JIT VM request.

njuCZ:
@tombuildsstuff thanks for your suggestion. That totally make sense. I will try to reach the service team to see if there is any solutions

njuCZ:
@tombuildsstuff I have another thought, is there any possiblity to deprecate `security_rule` for resource `azurerm_network_security_group` and encourage users to use resource `azurerm_network_security_rule`. In this way, this resource `jitNetworkAccessPolicy` will have no side effect

mark-strasser-nnl:
@njuCZ and @tombuildsstuff - I started work yesterday to add a provider feature to ignore JIT rules, as they have a well-defined naming convention.

Input would look like:

```terraform
provider "azurerm" {
  features {
    network { 
      ignore_jit_rules = true
    } 
  }
}
```

By default, this option would be disabled.  If enabled, any rules prefixed with `"SecurityCenter-JITRule"` would be disregarded, and therefore not constitute a change to the network security group.

I don't want to obviate the great work that @njuCZ has done here, but if desired I can follow up with the service team.  Let me know either way.

tombuildsstuff:
@mark-strasser-nnl to be honest this is more something that needs to be fixed in the Networking API - that the JIT Policies use NSG Rules is an implementation detail - so whilst we could look to fix this from within Terraform, this isn't something we can rely on. (since as an implementation detail it could change).

Whilst we could look to filter these out from the diff, we'd need to also handle this inside the Update method when NSG rules are defined inline within the NSG resource to ensure we don't remove those there too - so whilst it's possible to fix/filter these out from within Terraform - ultimately I think this needs to be fixed in the Networking API unfortunately.

mark-strasser-nnl:
> this needs to be fixed in the Networking API

This sounds complicated.  Since NSGs are assigned to a subnet, the Networking API would have to look at (a) if a JIT policy is defined, (b) if the JIT policy is assigned to a VM, and (c) if that VM has a NIC in any subnet to which the NSG is assigned to properly understand if a JIT policy is in effect.  Is that what you are thinking?

With the number of Terraform Azurerm users who employ JIT being apparently rather small, this seems unlikely to be ever implemented.  Just trying to be realistic.

Thanks for your followup.

tombuildsstuff:
@mark-strasser-nnl I'm referring to the same fix you are (e.g. filtering out the NSG Rules within that API) but within the API rather than within external tooling - we'll reach out to the Service Team here, but unfortunately since this is an implementation detail of the service, we can't assume this behaviour wouldn't change (or potentially be abused by someone disguising a custom NSG rule as a JIT rule, I guess) - so the API should have the context to filter out this implementation detail.

PR #10336, opened 2021-01-27
tc: updating the default version of tf to 0.14.4

tombuildsstuff:
This won't take effect until binary testing is enabled but there's no harm in staying up to date here


katbyte (approved):
LGTM 👍

PR #10337, opened 2021-01-27
r/azurerm_servicebus_namespace - Allowed setting messaging units capacity to 16 for Premium SKU

leewilson86:
Allowed the `azurerm_servicebus_namespace` resource to support the capacity of 16 for the number of messaging units to be set when using the Premium SKU.

Screenshot 2021-01-27 at 17 41 03

Microsoft documentation (https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-premium-messaging) only shows support for a capacity of up to 8, but we have been able to request 16 due to high levels of throttling at 8. Microsoft has confirmed going up to 16 is not possible in the portal, but customers can request it via a support ticket. This change is essential for us as Terraform wants to keep reverting our capacity back to 8:

```
  ~ update in-place
  # azurerm_servicebus_namespace.servicebusNameSpace[0] will be updated in-place
  ~ resource "azurerm_servicebus_namespace" "servicebusNameSpace" {
      ~ capacity                            = 16 -> 8
        default_primary_connection_string   = (sensitive value)
        default_primary_key                 = (sensitive value)
        default_secondary_connection_string = (sensitive value)
        default_secondary_key               = (sensitive value)
        id                                  = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/prdsbrg/providers/Microsoft.ServiceBus/namespaces/prdsb"
        location                            = "canadaeast"
        name                                = "prdsb"
        resource_group_name                 = "prdsbrg"
        sku                                 = "premium"
        tags                                = {}
        zone_redundant                      = false
    }
Plan: 0 to add, 1 to change, 0 to destroy.
```

But when we cannot enforce 16 in config due to validation rules within the provider, which this PR now addresses:

```
Error: expected capacity to be one of [0 1 2 4 8], got 16
  on ../../modules/resources/paas/serviceBus/nameSpace.tf line 1, in resource "azurerm_servicebus_namespace" "servicebusNameSpace":
   1: resource "azurerm_servicebus_namespace" "servicebusNameSpace" {
```

### DIFF
--- a/azurerm/internal/services/servicebus/servicebus_namespace_resource.go
+++ b/azurerm/internal/services/servicebus/servicebus_namespace_resource.go
@@ -78,7 +78,7 @@ func resourceServiceBusNamespace() *schema.Resource {
 				Type:         schema.TypeInt,
 				Optional:     true,
 				Default:      0,
-				ValidateFunc: validation.IntInSlice([]int{0, 1, 2, 4, 8}),
+				ValidateFunc: validation.IntInSlice([]int{0, 1, 2, 4, 8, 16}),
 			},
 
 			"default_primary_connection_string": {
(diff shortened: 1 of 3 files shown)


tombuildsstuff (approved):
LGTM - thanks for this @leewilson86

PR #10349, opened 2021-01-28
testing: data sources should fail using import steps

tombuildsstuff:
This commit forces a test failure when using an ImportStep within a Data Source, since the output is misleading:

```
testing.go:684: Step 1 error: 7 problems:
This is a bug in the provider, which should be reported in the provider's own issue tracker.
- Provider produced invalid object: Provider "azurerm" planned an invalid value for data.azurerm_something.test during refresh: .timeouts: unsupported attribute "update".
This is a bug in the provider, which should be reported in the provider's own issue tracker.
- Provider produced invalid object: Provider "azurerm" planned an invalid value for data.azurerm_something.test during refresh: .timeouts: unsupported attribute "create".
This is a bug in the provider, which should be reported in the provider's own issue tracker.
- Provider produced invalid object: Provider "azurerm" planned an invalid value for data.azurerm_something.test during refresh: .timeouts: unsupported attribute "delete".
```

### DIFF
--- a/azurerm/internal/acceptance/steps.go
+++ b/azurerm/internal/acceptance/steps.go
@@ -3,6 +3,7 @@ package acceptance
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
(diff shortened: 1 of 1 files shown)


katbyte (approved):
LGTM 👍

PR #10352, opened 2021-01-28
r/app_configuration: updating the requiresImport test to use a standard tier

tombuildsstuff:
There can only be a single "free" instance in a subscription at any one time, as there's a free test checking that sku can be provisioned - updating the requiresImport test to use a standard sku makes this pass consistently.


katbyte (approved):
LGTM 👍

tombuildsstuff:
Tests pass:

Screenshot 2021-01-28 at 09 52 48

PR #10353, opened 2021-01-28
new resource "azurerm_iot_security_device_group"

njuCZ:
rest api doc: https://docs.microsoft.com/en-us/rest/api/securitycenter/devicesecuritygroups/createorupdate

points needs attention:  
- the property "ThresholdRules" and "DenylistRules" are not supported by the service tam  
- the property "AllowlistRules" and "TimeWindowRules" supports different rule type, but they are defined together. I have asked service team about the mapping relation.
- the swagger is not good, in this PR there are some duplicate code to new different kinds of CustomAlertRule and flatten.

image

---

some context about device security group:
doc: 
https://docs.microsoft.com/en-us/azure/defender-for-iot/concept-customizable-security-alerts
https://docs.microsoft.com/en-us/azure/defender-for-iot/quickstart-create-custom-alerts

in short:
the deviceSecurityGroups resources allows the user the define customized alert rules for groups of IoT devices (connected to a specific IotSecuritySolution, meaning, connected to a specific IoT Hub).
The resource itself contains the custom rules and a group name used to correlate these rules to a group of IoT devices.


azurerm/internal/services/securitycenter/device_security_group_resource.go
```diff
@@ -0,0 +1,571 @@
...
+			},
+
+			"target_resource_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: azure.ValidateResourceID,
+			},
+
+			"allow_list_rule": {
```
katbyte, line 55: i think list is redundant here?
```suggestion
			"allow_rule": {
```

azurerm/internal/services/securitycenter/device_security_group_resource.go
```diff
@@ -0,0 +1,571 @@
...
+								string(security.RuleTypeProcessNotAllowed),
+							}, false),
+						},
+
+						"values": {
+							Type:     schema.TypeSet,
+							Required: true,
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
```
katbyte, line 75: can we validate this?

azurerm/internal/services/securitycenter/device_security_group_resource.go
```diff
@@ -0,0 +1,571 @@
...
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								string(security.RuleTypeConnectionToIPNotAllowed),
+								string(security.RuleTypeLocalUserNotAllowed),
+								string(security.RuleTypeProcessNotAllowed),
+							}, false),
+						},
+
+						"values": {
```
katbyte, line 70: what sort of values are allowed here?

azurerm/internal/services/securitycenter/device_security_group_resource.go
```diff
@@ -0,0 +1,571 @@
...
+							ValidateFunc: validation.IntAtLeast(0),
+						},
+
+						"min_threshold": {
+							Type:         schema.TypeInt,
+							Required:     true,
+							ValidateFunc: validation.IntAtLeast(0),
+						},
+
+						"time_window_size": {
```
katbyte, line 121: given the block already has time_window in it, should this just be window_size? and/or should this block be time_rule?

azurerm/internal/services/securitycenter/device_security_group_resource.go
```diff
@@ -0,0 +1,571 @@
...
+								string(security.RuleTypeDirectMethodInvokesNotInAllowedRange),
+								string(security.RuleTypeFailedLocalLoginsNotInAllowedRange),
+								string(security.RuleTypeFileUploadsNotInAllowedRange),
+								string(security.RuleTypeQueuePurgesNotInAllowedRange),
+								string(security.RuleTypeTwinUpdatesNotInAllowedRange),
+								string(security.RuleTypeUnauthorizedOperationsNotInAllowedRange),
+							}, false),
+						},
+
+						"max_threshold": {
```
katbyte, line 109: what is this the max of?

azurerm/internal/services/securitycenter/device_security_group_resource.go
```diff
@@ -0,0 +1,571 @@
...
+			timeWindowSizePointer = v.TimeWindowSize
+			minThresholdPointer = v.MinThreshold
+			maxThresholdPointer = v.MaxThreshold
+		case security.AmqpC2DMessagesNotInAllowedRange:
+			isEnabled = v.IsEnabled
+			t = string(v.RuleType)
+			timeWindowSizePointer = v.TimeWindowSize
+			minThresholdPointer = v.MinThreshold
+			maxThresholdPointer = v.MaxThreshold
+		case security.MqttC2DMessagesNotInAllowedRange:
```
katbyte, line 461: I think we can do something better here? there is so much repetition here... all these case statements are exactly the same?

website/docs/r/device_security_group.html.markdown
```diff
@@ -0,0 +1,112 @@
+---
+subcategory: "Security Center"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_device_security_group"
+description: |-
+  Manages a Device Security Group.
+---
+
+# azurerm_device_security_group
```
katbyte, line 9: given this works with iot security solutions this makes more sense as:
```suggestion
# azurerm_iot_security_device_group
```

website/docs/r/device_security_group.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) Specifies the name of the Device Security Group. Changing this forces a new resource to be created.
+
+* `target_resource_id` - (Required) The ID of the Azure Resource which to create the Device Security Group. Changing this forces a new resource to be created.
```
katbyte, line 65: what sort of target resources are valid here? only iot hubs? in which case
```suggestion
* `target_iot_hub_id` - (Required) The ID of the IoT Hub which to link the Device Security Group to. Changing this forces a new resource to be created.
```
 (also grammer)

website/docs/r/device_security_group.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+
+* `time_window_rule` - (Optional) One or more `time_window_rule` blocks as defined below.
+
+---
+
+An `allow_list_rule` block supports the following:
+
+* `type` - (Required) The type of supported rule type. Possible Values are `ConnectionToIpNotAllowed`, `LocalUserNotAllowed` and `LocalUserNotAllowed`.
+
+* `values` - (Required) The values to allow.
```
katbyte, line 77: looking at this now would it make sense to add different allow rule type blocks? block_ip_rule? allow_local_user_rule? 

also duplicate value above

website/docs/r/device_security_group.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+
+---
+
+An `time_window_rule` block supports the following:
+
+* `type` - (Required) The type of supported rule type. Possible Values are `ActiveConnectionsNotInAllowedRange`, `AmqpC2DMessagesNotInAllowedRange`, `MqttC2DMessagesNotInAllowedRange`, `HttpC2DMessagesNotInAllowedRange`, `AmqpC2DRejectedMessagesNotInAllowedRange`, `MqttC2DRejectedMessagesNotInAllowedRange`, `HttpC2DRejectedMessagesNotInAllowedRange`, `AmqpD2CMessagesNotInAllowedRange`, `MqttD2CMessagesNotInAllowedRange`, `HttpD2CMessagesNotInAllowedRange`, `DirectMethodInvokesNotInAllowedRange`, `FailedLocalLoginsNotInAllowedRange`, `FileUploadsNotInAllowedRange`, `QueuePurgesNotInAllowedRange`, `TwinUpdatesNotInAllowedRange` and `UnauthorizedOperationsNotInAllowedRange`.
```
katbyte, line 83: given this is a "time window rule" what does ActiveConnectionsNotInAllowedRange? or is this really a "range" or "window" rule rahter then anything to do with time?

katbyte (changes requested):
Thanks, i've given this a review and have some questions about the schema

azurerm/internal/services/securitycenter/device_security_group_resource.go
njuCZ: for different `type`, the validation rule is different.

There is another way: make each type a standalone schema

which way do you prefer?

azurerm/internal/services/securitycenter/device_security_group_resource.go
njuCZ: above

azurerm/internal/services/securitycenter/device_security_group_resource.go
njuCZ: personally I prefer "window_size", have updated

azurerm/internal/services/securitycenter/device_security_group_resource.go
njuCZ: yes, I think the swagger is not good, which caused all structs have the same property. In fact, they should promote those common property to parent struct.

Fow now, the only way I could think of is to use reflect, but that may make the codes not easy to read and maintain.

katbyte: i mean you could have a map of all these values and then if ok := map[key], ok, then do the 5 lines of assignment, or use case fallthrough - anything but this

website/docs/r/device_security_group.html.markdown
njuCZ: have made each type a different schema

azurerm/internal/services/securitycenter/device_security_group_resource.go
njuCZ: it seems both way could not work
1. in golang, map could not store the struct type as value
2. fallthrough could not be used in type switch

website/docs/r/device_security_group.html.markdown
njuCZ: this means:
Get an alert when the number of active connections of a device in the time window is not in the allowed range

azurerm/internal/services/securitycenter/device_security_group_resource.go
njuCZ: Users will get alert when the data in the time window is not in the threashold [min, max]

njuCZ:
@katbyte thanks for your suggestions. Could you have a look again?

website/docs/r/device_security_group.html.markdown
katbyte: So why is this a "time" window rule, when it has nothing to do with time? i think range_rule makes more sense?

azurerm/internal/services/securitycenter/device_security_group_resource.go
katbyte: sounds like this is a range rule for a certain duration, so it would be clear if the schema was:
```
range_rule {
  type
  max
  min
  duration
}
```

i get what time winder is trying to say, but its rather.. unclear (in/outside this range for this duration) so the terms range, max, min, and duration make more sense?

azurerm/internal/services/securitycenter/device_security_group_resource.go
katbyte: I'm not sure i follow? you have a type that you are testing on, you can fallthrough each statement till the last one does the assingents as every case is the same? or you can have a map of valid types:

```
	switch 1 {
	case 1:
		fallthrough
	case 2:
		fallthrough
	case 3:
		fmt.Println("case 1-3")
	default:
		fmt.Println("no match paic")
	}

	valid := map[string]bool{
		"AmqpC2DMessagesNotInAllowedRange": true,
		"AmqpC2DMessagesInAllowedRange":    true,
	}
	
	if valid["AmqpC2DMessagesInAllowedRange"] {
		fmt.Println("valid - do stuff")
	} else {
		//not valid panic?
	}
```

would not either of these work?

(https://play.golang.org/p/k5Uhdml9fU7)

njuCZ: for those case, it could work.
But here is a special case, we are switch the type of a struct. In type switch, fallthrough could not be used, go build will report error. Here is a simple example: https://play.golang.org/p/kc0flfjcrXg

as for using a map, we must first know the concrete type of the struct and then we could get the field ruleType. In go, it seems the only way to do dynamic type casting is to use type switch

katbyte: *sigh* ok i see it now, security.RuleTypeMqttC2DRejectedMessagesNotInAllowedRange is an actual type not a constant assigned to a type property 😪 - so absurd it didn't sink in that was the case.

Could we open an issue on the swagger/SDK and link to it both here in the expand and flatten functions? your right i don't think there is anything we can do to make this better without fixing the swagger and i'm going to guess that will take to long if it ever gets fixed?

azurerm/internal/services/securitycenter/device_security_group_resource.go
njuCZ: have updated

azurerm/internal/services/securitycenter/device_security_group_resource.go
njuCZ: yes, have created an issue to track it: https://github.com/Azure/azure-sdk-for-go/issues/14282

katbyte (approved):
Thanks @njuCZ - LGTM 👍

PR #10354, opened 2021-01-28
media: removing dependency on the azuread helpers

tombuildsstuff:
These are importing the AzureAD provider rather than AzureRM (or in this case, which should be using methods in the Plugin SDK)


katbyte (approved):
LGTM 👍

PR #9850, opened 2020-12-14
r/role_definition: fixing a crash when `permissions` is nil & making permissions optional

tombuildsstuff:
In testing it appears this block can be left empty (e.g. `[]`) - since the API will transform an empty permissions block into `[]`

Fixes #9815
Fixes #440
Fixes #3732

PR #10377, opened 2021-01-29
r/role_definition: switching to use the test shim

tombuildsstuff:
This is probably going to conflict with #9850, but I'll resolve that later.


manicminer (approved):
LGTM

PR #10390, opened 2021-01-30
WIP: `azurerm_kubernetes_cluster` optional kubeconfig

artburkart:
### Community Note

### Description

Because there are alternative means to obtain the kubeconfig file via
the Azure API, including the kube_config and kube_admin_config values as
attributes on the `azurerm_kubernetes_cluster` resource and data source
is unnecessary. Additionally, we don't need the values written to our
Terraform state files. This pull request introduces two new arguments
to the resource and data source, `exclude_kube_admin_config` and
`exclude_kube_config`, which—when set to `true`—omit the kubeconfig
values from the attributes of the objects.

For folks who generate a kubeconfig file for each user of a Kubernetes
cluster, this change enables those users to still leverage a
`terraform_remote_state` data source to access the attributes of an AKS
cluster created via Terraform without acquiring the ability to escalate
their own permissions to administrator level.

* azurerm_kubernetes_cluster

```hcl
provider "azurerm" {
  features {}
}

  name     = "test"
  location = "centralus"
}

resource "azurerm_kubernetes_cluster" "test" {
  name                = "test"
  dns_prefix          = "test"

  default_node_pool {
    name       = "default"
    node_count = 1
    vm_size    = "Standard_D2_v2"
  }

  identity {
    type = "SystemAssigned"
  }

  tags = {
    Environment = "test"
  }

  exclude_kube_admin_config = true
  exclude_kube_config       = true
}

data "azurerm_kubernetes_cluster" "test" {
  name                      = "test"
  resource_group_name       = azurerm_resource_group.test.name
  exclude_kube_admin_config = true
  exclude_kube_config       = true
}

```

### References

* https://github.com/hashicorp/terraform-plugin-sdk/issues/201

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_cluster_data_source.go
+++ b/azurerm/internal/services/containers/kubernetes_cluster_data_source.go
@@ -25,6 +25,16 @@ func dataSourceKubernetesCluster() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
+			"exclude_kube_admin_config": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
+			"exclude_kube_config": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
 			"name": {
 				Type:     schema.TypeString,
 				Required: true,
@@ -636,7 +646,7 @@ func dataSourceKubernetesClusterRead(d *schema.ResourceData, meta interface{}) e
 		}
 
 		// adminProfile is only available for RBAC enabled clusters with AAD
-		if props.AadProfile != nil {
+		if props.AadProfile != nil && !d.Get("exclude_kube_admin_config").(bool) {
 			adminProfile, err := client.GetAccessProfile(ctx, resourceGroup, name, "clusterAdmin")
 			if err != nil {
 				return fmt.Errorf("Error retrieving Admin Access Profile for Managed Kubernetes Cluster %q (Resource Group %q): %+v", name, resourceGroup, err)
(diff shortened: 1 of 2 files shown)


artburkart:
I committed this PR as WIP, because I didn't want to commit the time to working out the tests unless we determine this implementation has viability. 👍

artburkart:
Is there anything I can do to help move this forward?

manicminer:
Hi @artburkart, thanks for this PR. I'm confused as to why it would be desirable to exclude this from state? If you want to prevent downstream users from consuming it, you can wrap your azurerm_kubernetes_cluster resource / DS in a module and exclude it from the module's output variables? If the intent is to prevent this being saved to state for security reasons, the state should always be considered sensitive and protected accordingly, and there are a plethora of other sensitive data to be found there. Could you clarify the use case a little, I'd like to better understand the intent here? Thanks!

artburkart:
It's pulling in credentials unnecessarily and surprisingly. I certainly did not expect admin credentials to pop up in my Terraform state, and as it is today, yes, I treat my state like it's sensitive, but this decision to store admin credentials in the state defeats any attempt to set up an auditable system where users interact with the system using their own credentials. Anyone can escalate their own permissions. From this perspective, this implementation is broken.

I'm open to the possibility that I'm totally missing something here, but if not, I'd expect these credentials can be stored optionally, especially since thia implementation isn't aligned with the current API implementation.

-------- Original Message --------

artburkart:
Oh, the piece I forgot to mention is that I'd like to reference my Terraform state as a data source, and this prevents me from doing so.

-------- Original Message --------

manicminer:
Thanks @artburkart, that makes sense and I understand your perspective that saving admin credentials in state has some drawbacks. However this is a generally expected pattern and you'll find that many resources across different providers store their admin-equivalent credentials in state, to the extent that your statefile may contain most or all of the secrets for your managed infrastructure. For this reason you should protect your statefile accordingly.

If you make your Terraform state available via the `terraform_remote_state` data source, only the output variables specifically configured will be available for use by Terraform. It is true that the entire statefile is exposed on the wire, but [as detailed in the documentation](https://www.terraform.io/docs/language/state/remote-state-data.html) it's recommended to not expose your state to untrusted/external users and publish the relevant data to an external location.

Since this is by design across the Terraform ecosystem, I don't believe that non-exporting a particular set of credentials, such as proposed here, would have a net positive impact on Terraform security and so whilst I'd like to thank you for this proposal, I'm going to close this PR. If you have specific governance needs, or other concerns about exposing credentials to untrusted consumers of your Terraform state, as mentioned in the linked article I recommend you look into publishing select data to an alternate location.

PR #10396, opened 2021-01-31
Update to docs for setting provider version

krowlandson:
This PR provides a quick update to the docs to ensure they reflect the latest recommended approach for setting provider versions in Terraform.

The update is supported by Terraform 0.13.x and newer, and generates the following deprecation warning in Terraform 0.14.x

```bash
Warning: Version constraints inside provider configuration blocks are deprecated

  on main.tf line 2, in provider "azurerm":
   2:   version = "2.42.0"

Terraform 0.13 and earlier allowed provider version constraints inside the
provider configuration block, but that is now deprecated and will be removed
in a future version of Terraform. To silence this warning, move the provider
version constraint into the required_providers block.
```

Some examples in the repository may also need updating, but for now I wanted to focus on the docs before going further with this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,13 +9,22 @@ Version 2.x of the AzureRM Provider requires Terraform 0.12.x and later.
 
 ## Usage Example
 
-```
+> When using the AzureRM Provider with Terraform 0.13 and later, the recommended approach is to declare Provider versions in the Terraform configuration for the root module, using a `required_providers` block as per the following example. For previous versions, please continue to pin the version within the provider block.
+
+```hcl
+# We strongly recommend using the required_providers block to set the
+# Azure Provider source and version being used
+terraform {
+  required_providers {
+    azurerm = {
+      source = "hashicorp/azurerm"
+      version = "=2.40.1"
+    }
+  }
+}
+
 # Configure the Microsoft Azure Provider
 provider "azurerm" {
-  # We recommend pinning to the specific version of the Azure Provider you're using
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -9,13 +9,22 @@ Version 2.x of the AzureRM Provider requires Terraform 0.12.x and later.
...
+> When using the AzureRM Provider with Terraform 0.13 and later, the recommended approach is to declare Provider versions in the root module Terraform configuration, using a `required_providers` block as per the following example. For previous versions, please continue to pin the version within the provider block.
+
+```hcl
+# We strongly recommend using the required_providers block to set the
+# Azure Provider source and version being used
+terraform {
+  required_providers {
+    azurerm = {
+      source = "hashicorp/azurerm"
+      version = "=2.40.1"
```
tombuildsstuff, line 21: we don't have a v2.40.1 of the Provider - so for all of these could we make these:

```suggestion
      version = "=2.46.0"
```

(since this'll go out with v2.46 of the Provider later in the week)

tombuildsstuff (changes requested):
hey @krowlandson 

Thanks for this PR :)

Taking a look through on the whole this looks pretty good - but version `2.40.1` of the Azure Provider doesn't exist - so can we update this to be `2.46.0` (in all of the examples) since that's where this'll be released 👍 

Thanks!

katbyte (approved):
LGTM 👍

PR #10378, opened 2021-01-29
network: removing the direct dependency on `acceptance.AzureProvider`

tombuildsstuff:
Continuing the ongoing split to enable binary testing

PR #10403, opened 2021-02-01
network: switching the hand-written ID parsers for generated ones

tombuildsstuff:
Noticed as a part of fixing up #10378 - these can now be replaced by generated versions


katbyte (approved):
LGTM 👍

PR #10404, opened 2021-02-01
new resource "azurerm_spring_cloud_custom_domain"

njuCZ:
- this resource needs two computed property from app and certificate, so this PR also contains small changes of app and certificate resource
- to run the acctest, you should have a domain and delegate to azure dns zone, and then set the environment


azurerm/internal/services/springcloud/spring_cloud_custom_domain_resource.go
```diff
@@ -0,0 +1,151 @@
...
+			},
+
+			"spring_cloud_app_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validate.SpringCloudAppID,
+			},
+
+			"cert_name": {
```
katbyte, line 54: Could we not shorten certificate?
```suggestion
			"certificate_name": {
```

azurerm/internal/services/springcloud/spring_cloud_custom_domain_resource_test.go
```diff
@@ -0,0 +1,289 @@
...
+		},
+		data.ImportStep(),
+	})
+}
+
+func TestAccSpringCloudCustomDomain_requiresImport(t *testing.T) {
+	if os.Getenv("ARM_TEST_DNS_ZONE") == "" || os.Getenv("ARM_TEST_DATA_RESOURCE_GROUP") == "" {
+		t.Skip("Skipping as ARM_TEST_DNS_ZONE and/or ARM_TEST_DATA_RESOURCE_GROUP are not specified")
+		return
+	}
```
katbyte, line 45: Is there no way to provision this in the config to remove this external dep?

katbyte (dismissed):
Thanks @njuCZ - just a couple comments but otherwise looks good

azurerm/internal/services/springcloud/spring_cloud_custom_domain_resource_test.go
njuCZ: Steve suggests this way, which is the same with https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/web/app_service_managed_certificate_resource_test.go

azurerm/internal/services/springcloud/spring_cloud_custom_domain_resource_test.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func TestAccSpringCloudCustomDomain_update(t *testing.T) {
+	if os.Getenv("ARM_TEST_DNS_ZONE") == "" || os.Getenv("ARM_TEST_DATA_RESOURCE_GROUP") == "" {
+		t.Skip("Skipping as ARM_TEST_DNS_ZONE and/or ARM_TEST_DATA_RESOURCE_GROUP are not specified")
+		return
+	}
+
+	data := acceptance.BuildTestData(t, "azurerm_spring_cloud_custom_domain", "test")
+	r := SpringCloudJavaDeploymentResource{}
```
jackofallops, line 88: Wrong type here?

jackofallops (changes requested):
Hi @njuCZ - Thanks for this PR.  There's a couple of comments to address below and then I should be able to pull and test locally.

azurerm/internal/services/springcloud/spring_cloud_custom_domain_resource_test.go
njuCZ: sorry for the typo, have fixed it

azurerm/internal/services/springcloud/validate/spring_cloud_custom_domain_name.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func SpringCloudCustomDomainName(input interface{}, key string) (warnings []string, errors []error) {
+	v, ok := input.(string)
+	if !ok {
+		errors = append(errors, fmt.Errorf("expected %q to be a string", key))
+		return
+	}
+
+	if !regexp.MustCompile(`^(([a-zA-Z])|([a-zA-Z][a-zA-Z])|([a-zA-Z][0-9])|([0-9][a-zA-Z])|([a-zA-Z0-9][a-zA-Z0-9-_]{1,61}[a-zA-Z0-9]))\.([a-zA-Z]{2,6}|[a-zA-Z0-9-]{2,30}\.[a-zA-Z]{2,3})$`).MatchString(v) {
```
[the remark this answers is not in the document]
njuCZ: have updated the regex according to https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch08s15.html

added more test case

njuCZ:
@jackofallops thanks for your suggestions, I have updated this PR, I reruned the test in my local and it could succeed

website/docs/r/spring_cloud_app.html.markdown
```diff
@@ -77,6 +77,8 @@ The following attributes are exported:
 
 * `id` - The ID of the Spring Cloud Application.
 
+* `fqdn` - The Fully Qualified DNS Name.
```
jackofallops, line 80: Can we be  a little more verbose here? This is `*.azuremicroservices.io` on Azure public, which may not be obvious to users. I'm not sure if something else is used on sovereign / gov clouds?
Perhaps:
```suggestion
* `fqdn` - The Fully Qualified DNS Name of the Spring Application in the service.
```

jackofallops (changes requested):
Hi @njuCZ - Thanks for this PR.  I've left some comments and questions. If you can address those I'll continue testing/reviewing.
Thanks

azurerm/internal/services/springcloud/spring_cloud_custom_domain_resource_test.go
```diff
@@ -0,0 +1,289 @@
...
+      key_usage = [
+        "cRLSign",
+        "dataEncipherment",
+        "digitalSignature",
+        "keyAgreement",
+        "keyCertSign",
+        "keyEncipherment",
+      ]
+
+      subject = join("", ["CN=*.", data.azurerm_dns_zone.test.name])
```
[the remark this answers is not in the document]
njuCZ: have changed to use the specific name

website/docs/r/spring_cloud_app.html.markdown
njuCZ: this is a `computed` field, which is used in the `azurerm_dns_cname_record` resource. I think there is no need in terraform side to concern the suffix, and actually the suffix is different for different clouds.

azurerm/internal/services/springcloud/spring_cloud_custom_domain_resource_test.go
```diff
@@ -0,0 +1,289 @@
...
+			Config: r.complete(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+			),
+		},
+		data.ImportStep(),
+	})
+}
+
+func TestAccSpringCloudCustomDomain_update(t *testing.T) {
```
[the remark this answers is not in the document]
njuCZ: if `thumbprint` or `certificate_name` is forcenew, it seems no need to add the update acctest.

azurerm/internal/services/springcloud/spring_cloud_custom_domain_resource.go
```diff
@@ -0,0 +1,151 @@
...
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
+
+			"thumbprint": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
```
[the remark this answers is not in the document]
njuCZ: I could changed it to `forcenew`, but I could not understand the reason. I have tried to modify the name of `azurerm_spring_cloud_certificate` to let it recreate, it actually failed and reoprts: `The cert xxx is used by domain xx.xx.xx, can not be deleted`. This behaviour is controlled by the backend service, I think change it to `forcenew` does not help.

what's more, if this field is `forcenew`, I think maybe field `certificate_name` should also be `forcenew` ?

another thought, since `certificate_name` and `thumbprint` both comes from `azurerm_spring_cloud_certificate`, maybe I could directly reference the resource id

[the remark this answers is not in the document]
jackofallops: > The cert xxx is used by domain xx.xx.xx, can not be deleted
Correct, this is exactly the behaviour I was suggesting we avoid. If a user updates / replaces the referenced certificate, Terraform will be unable to delete that referenced certificate and will fail apply, since it will be in use.  By having `ForceNew` on this field (which will have a different value for a replace certificate, where name won't necessarily), we recreate the Custom Domain to release the association and allow the underlying certificate to be replaced. Does that make sense?

> maybe I could directly reference the resource id
I think this would be too restrictive?  The name may not be the only one on the certificate and another name on the same certificate (and subsequently same thumbprint) could be changed without needing to recreate this virtual resource? If it's a difference certificate, then the thumbprint will change and trigger the `ForceNew` in any case.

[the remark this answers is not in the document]
njuCZ: thanks you for your explanation. I have made it `forcenew`

njuCZ:
@jackofallops thanks for your carefully review. I have refined this PR.

jackofallops (approved):
Thanks for the updates @njuCZ - This LGTM now 👍  Assuming the test results look good, I'll get this merged later today.

jackofallops:
Tests are looking good 👍

PR #10446, opened 2021-02-03
Recoveryservices - support for recovery_public_ip

MitchDrage:
Resolving issue [8837](https://github.com/terraform-providers/terraform-provider-azurerm/issues/8837).

Depends on pull [10373](https://github.com/terraform-providers/terraform-provider-azurerm/pull/10373)

Changes:
- Created "recovery_public_ip" field within "network_interface" of an "azurerm_site_recovery_replicated_vm" resource.
- Added recovery public IP to existing test with the [attached results](https://gist.github.com/MitchDrage/0a0a380d2ab78fa7b872a48f13551240).
- Adjusted tests to use list of IP ranges in subnet due to the previous method being deprecated.
- Set the recovery target network settings to ForceNew: False since they can be changed after deployment without re-running the protection of the VM (reprotection can take an hour or more during an apply).
- Updated webpage to reference the new field.
- Remove resource "azurerm_subnet" "test2_3" from test as it was previously unused.


katbyte (approved):
Thanks @MitchDrage - LGTM 👍

PR #10451, opened 2021-02-03
Remove failing `kubernetes_cluster` test for deprecated bridge network mode

aristosvo:
While running the acctests for AKS I bumped into this one.

According to [AKS PM comments](https://github.com/Azure/AKS/issues/1954) `transparent` is now the only and default `network_mode` for Azure CNI.

When running the test this is confirmed: `Code="NetworkModeInvalid" Message="Network mode bridge is not valid, the valid values are 'transparent' or ''. We are looking to deprecate network mode"`

### DIFF
--- a/website/docs/r/kubernetes_cluster.html.markdown
+++ b/website/docs/r/kubernetes_cluster.html.markdown
@@ -304,11 +304,9 @@ A `network_profile` block supports the following:
 
 -> **NOTE:** When `network_plugin` is set to `azure` - the `vnet_subnet_id` field in the `default_node_pool` block must be set and `pod_cidr` must not be set.
 
-* `network_mode` - (Optional) Network mode to be used with Azure CNI. Possible values are `bridge` or `transparent`. Changing this forces a new resource to be created.
+* `network_mode` - (Optional) Network mode to be used with Azure CNI. Possible value is `transparent`, which is also the AKS default for Azure CNI. Changing this forces a new resource to be created. `network_mode` will probably be deprecated by Azure.
 
 ~> **NOTE:** This property can only be set when `network_plugin` is set to `azure`.
-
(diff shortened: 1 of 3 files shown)


website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -304,12 +304,10 @@ A `network_profile` block supports the following:
 
 -> **NOTE:** When `network_plugin` is set to `azure` - the `vnet_subnet_id` field in the `default_node_pool` block must be set and `pod_cidr` must not be set.
 
-* `network_mode` - (Optional) Network mode to be used with Azure CNI. Possible values are `bridge` or `transparent`. Changing this forces a new resource to be created.
+* `network_mode` - (Optional) Network mode to be used with Azure CNI. Possible value is `transparent`, which is also the AKS default for Azure CNI. Changing this forces a new resource to be created. `network_mode` will probably be deprecated by Azure.
```
tombuildsstuff, line 307: as a general rule we're tending to phrase this as:

```suggestion
* `network_mode` - (Optional) Network mode to be used with Azure CNI. Possible values are `bridge` and `transparent`. Changing this forces a new resource to be created.

~> **Note:** `network_mode` can only be set to `bridge` for existing Kubernetes Clusters and cannot be used to provision new Clusters - this will be removed by Azure in the future.
```

tombuildsstuff (commented):
hey @aristosvo 

Thanks for this PR - taking a look through this is looking good, if we can update the documentation comment then this should otherwise be good to go 👍 

Thanks!

website/docs/r/kubernetes_cluster.html.markdown
aristosvo: Thanks, makes sense!

tombuildsstuff (approved):
LGTM 👍

PR #10107, opened 2021-01-08
mssql: support auto backup for azurerm_mssql_virtual_machine

manicminer:

PR #10460, opened 2021-02-04
azurerm_mssql_virtual_machine: support auto_backup

manicminer:
Replaces #10107

Support configuring auto backup for azurerm_mssql_virtual_machine

- Enabling/disabling is determined by presence of `auto_backup` block
- WaitForState func is needed as auto backup settings take awhile to take effect, whether adding/removing or updating

### DIFF
--- a/azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
+++ b/azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
@@ -79,9 +79,10 @@ func resourceMsSqlVirtualMachine() *schema.Resource {
 						},
 
 						"encryption_password": {
-							Type:      schema.TypeString,
-							Optional:  true,
-							Sensitive: true,
+							Type:         schema.TypeString,
+							Optional:     true,
+							Sensitive:    true,
+							ValidateFunc: validation.StringIsNotEmpty,
 						},
 
 						"manual_schedule": {
--- a/azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
+++ b/azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
@@ -297,6 +297,13 @@ func resourceMsSqlVirtualMachine() *schema.Resource {
 }
 
 func resourceMsSqlVirtualMachineCustomDiff(d *schema.ResourceDiff, _ interface{}) error {
+	// ForceNew when removing the auto_backup block.
+	// See https://github.com/Azure/azure-rest-api-specs/issues/12818#issuecomment-773727756
+	old, new := d.GetChange("auto_backup")
+	if len(old.([]interface{})) == 1 && len(new.([]interface{})) == 0 {
+		return d.ForceNew("auto_backup")
+	}
+
 	encryptionEnabled := d.Get("auto_backup.0.encryption_enabled")
 	v, ok := d.GetOk("auto_backup.0.encryption_password")
 
--- a/azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
+++ b/azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
@@ -1,6 +1,7 @@
 package mssql
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"strings"
@@ -8,6 +9,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/services/preview/sqlvirtualmachine/mgmt/2017-03-01-preview/sqlvirtualmachine"
 	"github.com/hashicorp/go-azure-helpers/response"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
```diff
@@ -61,6 +66,84 @@ func resourceMsSqlVirtualMachine() *schema.Resource {
 				}, false),
 			},
 
+			"auto_backup": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"backup_system_databases": {
```
katbyte, line 75: should we put enabled on this?
```suggestion
						"backup_system_databases_enabled": {
```

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
```diff
@@ -61,6 +66,84 @@ func resourceMsSqlVirtualMachine() *schema.Resource {
...
+							}, false),
+						},
+
+						"full_backup_start_hour": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validation.IntBetween(0, 23),
+						},
+
+						"full_backup_window_hours": {
```
katbyte, line 114: would it make sense to put these into a block?

katbyte: and full_backup_window_hours coul be `full_backup_window_in_hours`

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
```diff
@@ -61,6 +66,84 @@ func resourceMsSqlVirtualMachine() *schema.Resource {
...
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
+
+						"encryption_password": {
+							Type:      schema.TypeString,
+							Optional:  true,
+							Sensitive: true,
+						},
```
katbyte, line 96: should this be validate as NoEmptyString - to prevent empty values?

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
```diff
@@ -61,6 +66,84 @@ func resourceMsSqlVirtualMachine() *schema.Resource {
...
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"backup_system_databases": {
+							Type:     schema.TypeBool,
+							Optional: true,
+						},
+
+						"backup_schedule_automated": {
```
katbyte, line 80: if we are not going to have enabled here this should be
```suggestion
						"automated_backup_schedule": {
```

and does this preclude setting the vull backup schedule properties

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
```diff
@@ -61,6 +66,84 @@ func resourceMsSqlVirtualMachine() *schema.Resource {
...
+							ValidateFunc: validation.IntBetween(1, 23),
+						},
+
+						"log_backup_frequency": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validation.IntBetween(5, 60),
+						},
+
+						"retention_period": {
```
katbyte, line 126: period in..? days?

katbyte (changes requested):
Have a couple questions about schema names but otherwise is looking good

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
manicminer: Sure, how about `system_databases_backup_enabled`?

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
manicminer: Will change to `retention_period_days` 👍

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
manicminer: 💡 yes indeed!

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
manicminer: Agreed, a block for these makes total sense and a more descriptive property name is better

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
manicminer: It does preclude setting the scheduling properties, putting those in a block will help tidy that up.

azurerm/internal/services/mssql/mssql_virtual_machine_resource.go
manicminer: Or rather, since it defaults true, perhaps reversing it to `manual_backup_schedule_enabled` would be more fitting

manicminer:
@katbyte I've dropped the toggle attribute for automatic/manual backup scheduling, and used the presence of the `manual_schedule` block instead, which has simplified the custom diff function. Also, due to a service error, `auto_backup` cannot be removed without causing the resource to fail.

Still pending test results due to the memory issue on the CI server, but it's passing locally for me with these test configs.

manicminer:
Test results
Screenshot 2021-02-05 at 15 16 32

katbyte (approved):
Thanks @manicminer - LGTM now!

PR #10464, opened 2021-02-04
provider: deprecate `skip_credentials_validation`

magodo:
This field got added to allow people to input fake data to get `terraform validate` working - which is fine but some people then complained that plan showed invalid data because of it - in short it shouldn’t have been added in the first.

### DIFF
--- a/azurerm/internal/provider/provider.go
+++ b/azurerm/internal/provider/provider.go
@@ -212,14 +212,6 @@ func azureProvider(supportLegacyTestSuite bool) terraform.ResourceProvider {
 
 			"features": schemaFeatures(supportLegacyTestSuite),
 
-			// Advanced feature flags
-			"skip_credentials_validation": {
-				Type:        schema.TypeBool,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("ARM_SKIP_CREDENTIALS_VALIDATION", false),
-				Description: "This will cause the AzureRM Provider to skip verifying the credentials being used are valid.",
-			},
-
 			"skip_provider_registration": {
 				Type:        schema.TypeBool,
 				Optional:    true,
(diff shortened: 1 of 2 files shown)


azurerm/internal/provider/provider.go
```diff
@@ -212,14 +212,6 @@ func azureProvider(supportLegacyTestSuite bool) terraform.ResourceProvider {
 
 			"features": schemaFeatures(supportLegacyTestSuite),
 
-			// Advanced feature flags
-			"skip_credentials_validation": {
-				Type:        schema.TypeBool,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("ARM_SKIP_CREDENTIALS_VALIDATION", false),
-				Description: "This will cause the AzureRM Provider to skip verifying the credentials being used are valid.",
-			},
```
tombuildsstuff, line 221: this'd be a breaking change rather than a deprecation (although ignoring it's value is otherwise fine) - but we can deprecate this by using a feature toggle below and adding a "deprecation" message:

```
if !features.ThreePointOh() {
		p.Schema["skip_credentials_validation"] = &schema.Schema{
			Type:        schema.TypeBool,
			Optional:    true,
			DefaultFunc: schema.EnvDefaultFunc("ARM_SKIP_CREDENTIALS_VALIDATION", false),
			Description: "[DEPRECATED] This will cause the AzureRM Provider to skip verifying the credentials being used are valid.",
			Deprecated:  "This field is deprecated and will be removed in version 3.0 of the Azure Provider",
		}
	}
```

tombuildsstuff (changes requested):
hey @magodo 

Thanks for this PR - taking a look through this mostly LGTM - but we'll need to add a deprecation message to the existing field rather than outright removing it for the moment (although the behaviour can be removed, since this hasn't worked since ~2.0 anyway).

Thanks!

tombuildsstuff (approved):
LGTM 👍

schuettecarsten:
Now I always get a warning that the block will be deprecated - even if I do not use this block....?

NewGyu:
Hi @magodo . Should this warning be shown evenif the field is not specified explicitly?

My main.tf is as below. The `skip_credentials_validation` is not specified explicitly.
```
provider "azurerm" {
  subscription_id = "6fd834f7-50a1-4d6a-b4f5-735ed47cbe9f"
  features {}
}
```

But the message below is shown when `terraform plan` is running.

```
Warning: "skip_credentials_validation": [DEPRECATED] This field is deprecated and will be removed in version 3.0 of the Azure Provider.
```

Is this expected?

NewGyu:
Sorry, #10560 means that. (I didn't read this thread enough.)

PR #10466, opened 2021-02-04
`azurerm_[linux|windows]_virtual_machine` - add Data Disks support

### LINKED ISSUE #6117: Support for disk attachment to VMs at creation time
### Community Note

### Description

Azure [allows VMs to be booted with managed data disks pre-attached](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/tutorial-manage-disks#create-and-attach-disks)/attached-on-boot. This enables use cases where `cloud-init` and/or other "on-launch" configuration management tooling is able to prepare them for use as part of the initialisation process.

This provider currently only supports this case for individual VMs with the older, deprecated `azurerm_virtual_machine` resource. The new `azurerm_linux_virtual_machine` and `azurerm_windows_virtual_machine` resources instead opt to push users towards the separate `azurerm_virtual_machine_data_disk_attachment` which only attaches data disks to an existing VM post-boot, which fails to service the use case laid out above.

This is in contrast to the respective `*_scale_set` providers which (albeit out of necessity) support this behaviour.

Please could a repeatable `data_disk` block be added to the new VM resources (analogous to the same block in their scale_set counterparts) in order to allow VMs to be started with managed data disks pre-attached.

Thanks! 😁 

* `azurerm_linux_virtual_machine`

```hcl
resource "azurerm_linux_virtual_machine" "example" {
  [...]

  os_disk {
    name                 = "example-os"
    storage_account_type = "StandardSSD_LRS"
  }

  data_disk {
    name                 = "example-data"
    disk_size_gb         = 4096
    lun                  = 0
    storage_account_type = "StandardSSD_LRS"
  }

  [...]
}
```

### References

* https://docs.microsoft.com/en-us/azure/virtual-machines/linux/tutorial-manage-disks#create-and-attach-disks
* https://github.com/terraform-providers/terraform-provider-azurerm/issues/6074

### LINKED ISSUE #8794: Support for Importing OSDisk of azurerm_windows_virtual_machine/azurerm_linux_virtual_machine
### Community Note

### Description

Not being able to import the OSDisk has deadlocked me, since a machine that I am using uses a disk encryption set, I cannot fall back to azurerm_virtual_machine, since it does not support disk encryption sets.

image

* azurerm_windows_virtual_machine
* azurerm_linux_virtual_machine

```hcl
resource "azurerm_windows_virtual_machine" "dc1" {
  name                = "machinename"
  resource_group_name = azurerm_resource_group.rgdc.name
  location            = var.location
  size                = "Standard_D2s_v3"
  admin_username      = "azureuser"
  admin_password      = random_password.admin_password.result
  network_interface_ids = [
    azurerm_network_interface.vm1nic0.id,
  ]

  provision_vm_agent  = true
  availability_set_id = azurerm_availability_set.asdc0.id

  os_disk {
    name                      = "diskname"
    caching                   = "None"
    storage_account_type      = "Premium_LRS"
    write_accelerator_enabled = "false"
    disk_encryption_set_id    = azurerm_disk_encryption_set.des0.id
  }

  source_image_reference {
    publisher = "MicrosoftWindowsServer"
    offer     = "WindowsServer"
    sku       = "2019-Datacenter"
    version   = "latest"
  }
}
```
```powershell
terraform1228.exe import azurerm_windows_virtual_machine.dc1 /subscriptions/{GUID}/resourceGroups/rgname/providers/Microsoft.Compute/virtualMachines/machinename
```

### References

* #0000

jackofallops:
This PR creates an Opt-in beta for allowing the use of Data Disks in-line with the Linux and Windows Virtual Machine resources. Support for local disk creation and attachment of existing disks is included. 

`create` disks are new disks created with the VM and can be partially managed (e.g. can be grown)
`attach` disks are exiting managed disks, which cannot be managed via the VM configuration.

By default, `create` data disks are deleted on VM deletion, which is controlled by the features block in the provider configuration. `attach` disks are always left intact by the VM.

resolves #6117 
Fixes #8794

### DIFF
--- a/azurerm/internal/services/compute/linux_virtual_machine_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_resource.go
@@ -16,6 +16,7 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
 	azValidate "github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/validate"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/features"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/locks"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/compute/parse"
 	computeValidate "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/compute/validate"
@@ -31,7 +32,7 @@ import (
 // TODO: confirm locking as appropriate
 
 func resourceLinuxVirtualMachine() *schema.Resource {
-	return &schema.Resource{
+	linuxVirtualMachineSchema := &schema.Resource{
 		Create: resourceLinuxVirtualMachineCreate,
 		Read:   resourceLinuxVirtualMachineRead,
 		Update: resourceLinuxVirtualMachineUpdate,
@@ -300,6 +301,11 @@ func resourceLinuxVirtualMachine() *schema.Resource {
 			},
 		},
 	}
+	if features.VMDataDiskBeta() {
+		linuxVirtualMachineSchema.Schema["data_disks"] = virtualMachineDataDiskSchema()
+	}
+
+	return linuxVirtualMachineSchema
 }
 
 func resourceLinuxVirtualMachineCreate(d *schema.ResourceData, meta interface{}) error {
@@ -363,6 +369,15 @@ func resourceLinuxVirtualMachineCreate(d *schema.ResourceData, meta interface{})
 	osDiskRaw := d.Get("os_disk").([]interface{})
 	osDisk := expandVirtualMachineOSDisk(osDiskRaw, compute.Linux)
 
+	dataDisks := &[]compute.DataDisk{}
+
+	if features.VMDataDiskBeta() {
+		dataDisks, err = expandVirtualMachineDataDisks(d, meta)
+		if err != nil {
+			return err
+		}
+	}
+
 	secretsRaw := d.Get("secret").([]interface{})
 	secrets := expandLinuxSecrets(secretsRaw)
 
@@ -405,10 +420,7 @@ func resourceLinuxVirtualMachineCreate(d *schema.ResourceData, meta interface{})
 			StorageProfile: &compute.StorageProfile{
 				ImageReference: sourceImageReference,
 				OsDisk:         osDisk,
-
-				// Data Disks are instead handled via the Association resource - as such we can send an empty value here
-				// but for Updates this'll need to be nil, else any associations will be overwritten
-				DataDisks: &[]compute.DataDisk{},
+				DataDisks:      dataDisks,
 			},
 
 			// Optional
@@ -688,6 +700,16 @@ func resourceLinuxVirtualMachineRead(d *schema.ResourceData, meta interface{}) e
 		if err := d.Set("source_image_reference", flattenSourceImageReference(profile.ImageReference)); err != nil {
 			return fmt.Errorf("setting `source_image_reference`: %+v", err)
 		}
+
+		if features.VMDataDiskBeta() {
+			if profile.DataDisks != nil {
+				dataDisks, err := flattenVirtualMachineDataDisks(profile.DataDisks)
+				if err != nil {
+					return err
+				}
+				d.Set("data_disks", dataDisks)
+			}
+		}
 	}
 
 	encryptionAtHostEnabled := false
@@ -719,6 +741,7 @@ func resourceLinuxVirtualMachineRead(d *schema.ResourceData, meta interface{}) e
 
 func resourceLinuxVirtualMachineUpdate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*clients.Client).Compute.VMClient
+	disksClient := meta.(*clients.Client).Compute.DisksClient
 	ctx, cancel := timeouts.ForUpdate(meta.(*clients.Client).StopContext, d)
 	defer cancel()
 
@@ -871,6 +894,55 @@ func resourceLinuxVirtualMachineUpdate(d *schema.ResourceData, meta interface{})
 		osDisk := expandVirtualMachineOSDisk(osDiskRaw, compute.Linux)
 		update.VirtualMachineProperties.StorageProfile = &compute.StorageProfile{
 			OsDisk: osDisk,
+		}
+	}
+
+	// list of disks for expand operations
+	dataDisksForGrow := make([]map[string]interface{}, 0)
+	// list of disks with encryption set changes
+	dataDisksForEncrypt := make([]map[string]interface{}, 0)
+	if features.VMDataDiskBeta() && d.HasChange("data_disks") {
+		shouldUpdate = true
+
+		oldRaw, newRaw := d.GetChange("data_disks.0.local")
+		oldDisks := oldRaw.(*schema.Set).List()
+		newDisks := newRaw.(*schema.Set).List()
+		for _, o := range oldDisks {
+			oldDisk := o.(map[string]interface{})
+			if oldDiskName, ok := oldDisk["name"]; ok {
+				for _, n := range newDisks {
+					newDisk := n.(map[string]interface{})
+					if newDiskName, ok := newDisk["name"]; ok && oldDiskName.(string) == newDiskName.(string) {
+						if newDisk["disk_size_gb"].(int) < oldDisk["disk_size_gb"].(int) {
+							return fmt.Errorf("new disk size cannot be smaller than existing for %q, in Virtual Machine %q (resource group %q)", oldDisk["name"], id.Name, id.ResourceGroup)
+						} else if newDisk["disk_size_gb"].(int) > oldDisk["disk_size_gb"].(int) {
+							shouldShutDown = true
+							shouldDeallocate = true
+							dataDisksForGrow = append(dataDisksForGrow, newDisk)
+						}
+						if newDisk["disk_encryption_set_id"].(string) != oldDisk["disk_encryption_set_id"].(string) {
+							dataDisksForEncrypt = append(dataDisksForEncrypt, newDisk)
+						}
+					}
+				}
+			}
+		}
+		// EncryptionSet changes for "existing" disks
+		oldExistingRaw, newExistingRaw := d.GetChange("data_disks.0.existing")
+		oldExisting := oldExistingRaw.(*schema.Set).List()
+		newExisting := newExistingRaw.(*schema.Set).List()
+		for _, o := range oldExisting {
+			oldDisk := o.(map[string]interface{})
+			if oldDiskID, ok := oldDisk["managed_disk_id"]; ok {
+				for _, n := range newExisting {
+					newDisk := n.(map[string]interface{})
+					if newDiskID, ok := newDisk["managed_disk_id"]; ok && oldDiskID.(string) == newDiskID.(string) {
+						if newDisk["disk_encryption_set_id"].(string) != oldDisk["disk_encryption_set_id"].(string) {
+							dataDisksForEncrypt = append(dataDisksForEncrypt, newDisk)
+						}
+					}
+				}
+			}
 		}
 	}
 
@@ -1029,8 +1101,6 @@ func resourceLinuxVirtualMachineUpdate(d *schema.ResourceData, meta interface{})
 		newSize := d.Get("os_disk.0.disk_size_gb").(int)
 		log.Printf("[DEBUG] Resizing OS Disk %q for Linux Virtual Machine %q (Resource Group %q) to %dGB..", diskName, id.Name, id.ResourceGroup, newSize)
 
-		disksClient := meta.(*clients.Client).Compute.DisksClient
-
 		update := compute.DiskUpdate{
 			DiskUpdateProperties: &compute.DiskUpdateProperties{
 				DiskSizeGB: utils.Int32(int32(newSize)),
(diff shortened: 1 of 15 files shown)


katbyte (approved):
LGTM 👍

azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -1030,7 +1101,76 @@ func resourceLinuxVirtualMachineUpdate(d *schema.ResourceData, meta interface{})
...
+
+		// Do we have disks to resize or change encryption set?
+		for _, v := range dataDisksForEncrypt {
+			diskName, ok := v["name"].(string)
+			if !ok {
+				return fmt.Errorf("could not resize data disk, empty value for `name`")
+			}
+			diskEncryptionSetID, ok := v["disk_encryption_set_id"].(string)
+			if !ok {
+				return fmt.Errorf("failed to read new disk Encryption Eet ID for Data Disk %q (resource group %q)", diskName, id.ResourceGroup)
```
tombuildsstuff, line 1123:
```suggestion
				return fmt.Errorf("failed to read new Disk Encryption Set ID for Data Disk %q (resource group %q)", diskName, id.ResourceGroup)
```

azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -1030,7 +1101,76 @@ func resourceLinuxVirtualMachineUpdate(d *schema.ResourceData, meta interface{})
...
+			}
+			if err = resizeFuture.WaitForCompletionRef(ctx, client.Client); err != nil {
+				return fmt.Errorf("failed waiting for resize of Data Disk %q for Virtual Machine %q (resource group %q): %+v", diskName, id.Name, id.ResourceGroup, err)
+			}
+		}
+
+		// now disks are resized, we can put the new list into the VMUpdate
+		if update.VirtualMachineProperties.StorageProfile == nil {
+			update.VirtualMachineProperties.StorageProfile = &compute.StorageProfile{
+				DataDisks: dataDisks,
```
tombuildsstuff, line 1170: this should never be nil, since the OS Disk is captured within it - so I think this'd be invalid?

azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -1160,6 +1300,29 @@ func resourceLinuxVirtualMachineDelete(d *schema.ResourceData, meta interface{})
...
+				dataDisks := *props.StorageProfile.DataDisks
+				for _, v := range dataDisks {
+					if v.CreateOption == compute.DiskCreateOptionTypesEmpty && v.Name != nil {
+						deleteFuture, err := disksClient.Delete(ctx, id.ResourceGroup, *v.Name)
+						if err != nil {
+							return fmt.Errorf("failure deleting Data Disk %q (Virtual Machine %q / resource group %q): %+v", *v.Name, id.Name, id.ResourceGroup, err)
+						}
+
+						if err = deleteFuture.WaitForCompletionRef(ctx, disksClient.Client); err != nil {
+							return fmt.Errorf("failure waiting for deletion of Data Disk%q (Virtual Machine %q / resource group %q): %+v", *v.Name, id.Name, id.ResourceGroup, err)
```
tombuildsstuff, line 1318:
```suggestion
							return fmt.Errorf("failure waiting for deletion of Data Disk %q (Virtual Machine %q / resource group %q): %+v", *v.Name, id.Name, id.ResourceGroup, err)
```

azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -1160,6 +1300,29 @@ func resourceLinuxVirtualMachineDelete(d *schema.ResourceData, meta interface{})
...
+	if features.VMDataDiskBeta() {
+		deleteDataDiskOnDeletion := meta.(*clients.Client).Features.VirtualMachine.DeleteDataDiskOnDeletion
+
+		// delete any disks created with the VM if feature toggled to do so. "existing" disks are not affected.
+		if deleteDataDiskOnDeletion {
+			if props := existing.VirtualMachineProperties; props != nil && props.StorageProfile != nil && props.StorageProfile.DataDisks != nil {
+				dataDisks := *props.StorageProfile.DataDisks
+				for _, v := range dataDisks {
+					if v.CreateOption == compute.DiskCreateOptionTypesEmpty && v.Name != nil {
+						deleteFuture, err := disksClient.Delete(ctx, id.ResourceGroup, *v.Name)
```
tombuildsstuff, line 1312: we don't need to for this initial version, but it'd be worth making this easier to do this in parallel in the future - perhaps it's worth pulling these into a list, then calling a separate method to make this easier to parallelise in the future and also reusable?

azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -828,6 +852,55 @@ func resourceLinuxVirtualMachineUpdate(d *schema.ResourceData, meta interface{})
...
+					newDisk := n.(map[string]interface{})
+					if newDiskID, ok := newDisk["managed_disk_id"]; ok && oldDiskID.(string) == newDiskID.(string) {
+						if newDisk["disk_encryption_set_id"].(string) != oldDisk["disk_encryption_set_id"].(string) {
+							dataDisksForEncrypt = append(dataDisksForEncrypt, newDisk)
+						}
+					}
+				}
+			}
+		}
+	}
```
tombuildsstuff, line 902: (same here) should we pull this out into a common helper method/type here? seems worthwhile to expose a method to resize the disks / encryption update which could potentially be done in parallel in the future - and shared across Linux & Windows in the interim?

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -706,6 +722,16 @@ func resourceWindowsVirtualMachineRead(d *schema.ResourceData, meta interface{})
 		if err := d.Set("source_image_reference", flattenSourceImageReference(profile.ImageReference)); err != nil {
 			return fmt.Errorf("setting `source_image_reference`: %+v", err)
 		}
+
+		if features.VMDataDiskBeta() {
+			if props.StorageProfile.DataDisks != nil {
```
tombuildsstuff, line 727: (same here) we should remove this `if` statement

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -912,6 +939,54 @@ func resourceWindowsVirtualMachineUpdate(d *schema.ResourceData, meta interface{
 		}
 	}
 
+	dataDisksForGrow := make([]map[string]interface{}, 0)
+	// list of disks with encryption set changes
+	dataDisksForEncrypt := make([]map[string]interface{}, 0)
+	if features.VMDataDiskBeta() && d.HasChange("data_disks") {
```
tombuildsstuff, line 945: (same here) we should pull this out into it's own method

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -1119,6 +1190,75 @@ func resourceWindowsVirtualMachineUpdate(d *schema.ResourceData, meta interface{
...
+				return fmt.Errorf("failed resizing Data Disk %q for Virtual Machine %q (resource group %q): %+v", diskName, id.Name, id.ResourceGroup, err)
+			}
+			if err = resizeFuture.WaitForCompletionRef(ctx, client.Client); err != nil {
+				return fmt.Errorf("failed waiting for resize of Data Disk %q for Virtual Machine %q (resource group %q): %+v", diskName, id.Name, id.ResourceGroup, err)
+			}
+		}
+
+		// now disks are resized, we can put the new list into the VMUpdate
+		if update.VirtualMachineProperties.StorageProfile == nil {
+			update.VirtualMachineProperties.StorageProfile = &compute.StorageProfile{
```
tombuildsstuff, line 1254: (same here)

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -1245,6 +1385,29 @@ func resourceWindowsVirtualMachineDelete(d *schema.ResourceData, meta interface{
 		log.Printf("[DEBUG] Skipping Deleting OS Disk from Windows Virtual Machine %q (Resource Group %q)..", id.Name, id.ResourceGroup)
 	}
 
+	if features.VMDataDiskBeta() {
+		deleteDataDiskOnDeletion := meta.(*clients.Client).Features.VirtualMachine.DeleteDataDiskOnDeletion
+
+		// delete any disks created with the VM if feature toggled to do so. "existing" disks are not affected.
+		if deleteDataDiskOnDeletion {
```
tombuildsstuff, line 1392: (same here)

azurerm/internal/services/compute/windows_virtual_machine_resource.go
```diff
@@ -1245,6 +1385,29 @@ func resourceWindowsVirtualMachineDelete(d *schema.ResourceData, meta interface{
...
+			if props := existing.VirtualMachineProperties; props != nil && props.StorageProfile != nil && props.StorageProfile.DataDisks != nil {
+				dataDisks := *props.StorageProfile.DataDisks
+				for _, v := range dataDisks {
+					if v.CreateOption == compute.DiskCreateOptionTypesEmpty && v.Name != nil {
+						deleteFuture, err := disksClient.Delete(ctx, id.ResourceGroup, *v.Name)
```
tombuildsstuff, line 1397: this is slightly presumptuous - disks _should_ be in the same RG but don't have to be - perhaps we should parse the Disk ID and use those details?

azurerm/internal/services/compute/linux_virtual_machine_resource_data_disk_test.go
```diff
@@ -0,0 +1,803 @@
...
+}
+
+func TestAccLinuxVirtualMachine_dataDisksComplete(t *testing.T) {
+	if !features.VMDataDiskBeta() {
+		t.Skip("skipping as In-line Data Disk beta is not enabled")
+	}
+
+	if !features.VMDataDiskBeta() {
+		t.Skip("skipping as In-line Data Disk beta is not enabled")
+	}
```
tombuildsstuff, line 38: dupe

azurerm/internal/services/compute/windows_virtual_machine_resource_data_disk_test.go
```diff
@@ -0,0 +1,770 @@
...
+  features {}
+}
+
+resource "azurerm_managed_disk" "test" {
+  name                 = "acctested-%d"
+  location             = azurerm_resource_group.test.location
+  resource_group_name  = azurerm_resource_group.test.name
+  storage_account_type = "Standard_LRS"
+  create_option        = "Empty"
+  disk_size_gb         = "1"
```
tombuildsstuff, line 727: we should add a test (to both Linux and Windows) which increases the size of the external disk too - whilst there's a test for that in the Managed Disk resource tests, perhaps it's worth updating an internal and external at the same time to validate the locking works (since that stops/starts the VM)?

azurerm/internal/services/compute/virtual_machine.go
```diff
@@ -411,3 +416,407 @@ func flattenVirtualMachineOSDisk(ctx context.Context, disksClient *compute.Disks
...
+					Type:         schema.TypeString,
+					Required:     true,
+					ValidateFunc: validation.StringIsNotEmpty,
+				},
+
+				"storage_account_type": {
+					Type:     schema.TypeString,
+					Required: true,
+					// whilst this appears in the Update block the API returns this when changing:
+					// Changing property 'osDisk.managedDisk.storageAccountType' is not allowed
```
tombuildsstuff, line 468: this comment doesn't appear to be applicable?

azurerm/internal/services/compute/virtual_machine.go
```diff
@@ -411,3 +416,407 @@ func flattenVirtualMachineOSDisk(ctx context.Context, disksClient *compute.Disks
...
+					ValidateFunc: validation.StringInSlice([]string{
+						// note: OS Disks don't support Ultra SSDs
```
tombuildsstuff, line 470: this is a Data Disk, which do?

azurerm/internal/services/compute/virtual_machine.go
```diff
@@ -411,3 +416,407 @@ func flattenVirtualMachineOSDisk(ctx context.Context, disksClient *compute.Disks
...
 }
+
+func virtualMachineDataDiskSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeList,
+		MaxItems: 1,
+		Optional: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"local": virtualMachineLocalDataDiskSchema(),
```
tombuildsstuff, line 427: part of me thinks this should be called `new` rather than `local`?

azurerm/internal/services/compute/virtual_machine.go
```diff
@@ -411,3 +416,407 @@ func flattenVirtualMachineOSDisk(ctx context.Context, disksClient *compute.Disks
...
+				"lun": {
+					Type:         schema.TypeInt,
+					Required:     true,
+					ForceNew:     true,
+					ValidateFunc: validation.IntAtLeast(0),
+				},
+
+				"name": {
+					Type:         schema.TypeString,
+					Required:     true,
```
tombuildsstuff, line 460: we don't support renaming disks, so should this be ForceNew?

azurerm/internal/services/compute/virtual_machine.go
```diff
@@ -411,3 +416,407 @@ func flattenVirtualMachineOSDisk(ctx context.Context, disksClient *compute.Disks
...
+
+func virtualMachineExistingDataDiskSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeSet,
+		Optional: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"managed_disk_id": {
+					Type:         schema.TypeString,
+					Required:     true,
```
tombuildsstuff, line 511: shouldn't this be ForceNew?

azurerm/internal/services/compute/virtual_machine.go
```diff
@@ -411,3 +416,407 @@ func flattenVirtualMachineOSDisk(ctx context.Context, disksClient *compute.Disks
...
+					}, false),
+				},
+
+				"storage_account_type": {
+					Type:     schema.TypeString,
+					Required: true,
+					// whilst this appears in the Update block the API returns this when changing:
+					// Changing property 'osDisk.managedDisk.storageAccountType' is not allowed
+					ValidateFunc: validation.StringInSlice([]string{
+						// note: OS Disks don't support Ultra SSDs
```
tombuildsstuff, line 537: (as above) the comment isn't valid for Data Disks and I think we should support Ultra disks here?

azurerm/internal/services/compute/virtual_machine.go
```diff
@@ -411,3 +416,407 @@ func flattenVirtualMachineOSDisk(ctx context.Context, disksClient *compute.Disks
...
+}
+
+func expandVirtualMachineLocalDataDisksForUpdate(input interface{}, d *schema.ResourceData, meta interface{}) ([]compute.DataDisk, error) {
+	if input == nil || len(input.(*schema.Set).List()) == 0 {
+		return []compute.DataDisk{}, nil
+	}
+
+	diskClient := meta.(*clients.Client).Compute.DisksClient
+	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+	defer cancel()
```
tombuildsstuff, line 627: we should pass the context (and client) into this method - else we'll have 2 contexts firing (the Update, which is valid) and a Read (5m) at staggered times

azurerm/internal/services/compute/virtual_machine.go
```diff
@@ -411,3 +416,407 @@ func flattenVirtualMachineOSDisk(ctx context.Context, disksClient *compute.Disks
...
+}
+
+func expandVirtualMachineExistingDataDisksForCreate(input interface{}, d *schema.ResourceData, meta interface{}) ([]compute.DataDisk, error) {
+	if input == nil || len(input.(*schema.Set).List()) == 0 {
+		return []compute.DataDisk{}, nil
+	}
+
+	disksClient := meta.(*clients.Client).Compute.DisksClient
+	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+	defer cancel()
```
tombuildsstuff, line 682: (same here)

azurerm/internal/services/compute/virtual_machine.go
```diff
@@ -411,3 +416,407 @@ func flattenVirtualMachineOSDisk(ctx context.Context, disksClient *compute.Disks
...
+			existingDisks = append(existingDisks, dataDisk)
+
+		default:
+			return nil, fmt.Errorf("unsupported `createOption` type while flattening: %s", string(createOption))
+		}
+	}
+	return []interface{}{
+		map[string]interface{}{
+			"local":    schema.NewSet(resourceArmVirtualMachineNewDataDiskHash, newDisks),
+			"existing": existingDisks,
```
tombuildsstuff, line 802: just to confirm: this is also a set - but using the default hash func

tombuildsstuff (changes requested):
Taking a look through on the whole this looks pretty good - I've left some comments inline, mostly fairly minor around naming - but main issue here is around the context, where we're firing both the Update and Read contexts off in the Update, where the Read has a smaller timeout (and these are staggered) - instead we should pass through the Update context through the entire Update method, so that we're using that timeout for this entire process. Also worth noting there's no docs for this at this time?

azurerm/internal/services/compute/linux_virtual_machine_resource.go
jackofallops: Only if `os_disk` has a change also, since it's a Patch the StorageProfile will be nil here if only the Data Disk(s) have change?

azurerm/internal/services/compute/linux_virtual_machine_resource.go
```diff
@@ -650,6 +663,16 @@ func resourceLinuxVirtualMachineRead(d *schema.ResourceData, meta interface{}) e
 		if err := d.Set("source_image_reference", flattenSourceImageReference(profile.ImageReference)); err != nil {
 			return fmt.Errorf("setting `source_image_reference`: %+v", err)
 		}
+
+		if features.VMDataDiskBeta() {
+			if profile.DataDisks != nil {
```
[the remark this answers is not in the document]
jackofallops: the presence of `data_disk` in the schema is toggled on the same func, so I think this is correct as is?  Or we'll be setting a value for a property that does not exist?

azurerm/internal/services/compute/windows_virtual_machine_resource.go
jackofallops: If the disk is created with the VM, it will always be in the same RG since there's no option to create it elsewhere? We don't touch `existing` disks that have been attached as part of the `deleteDataDiskOnDeletion` process.

azurerm/internal/services/compute/linux_virtual_machine_resource.go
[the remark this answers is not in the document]
jackofallops: Understood now, updated.

azurerm/internal/services/compute/virtual_machine.go
jackofallops: As discussed offline, these have been updated to `create` and `attach` to better describe their relationship to the VM.

jackofallops:
Closing this for the time being due to technical limitations on implementation that are going to be resolved in future, but no set timeframe is currently available. Will re-open / start afresh as soon as possible.

PR #10470, opened 2021-02-04
azurerm_storage_data_lake_gen2_filesystem_resource: do not set/retrieve ACLs when HNS is not Enabled

njuCZ:
all synapse related test case are failed because of getting acl for storage data lake gen2 resource.

according to https://docs.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-access-control#do-i-have-to-enable-support-for-acls

acl is enabled when Hierarchical Namespace (HNS) feature in storage account is turned ON.

After testing, when Hierarchical Namespace feature is disabled in storage account, the getAcl rest api will report error, which cause all tc in synapse fails

### DIFF
--- a/azurerm/internal/services/storage/storage_data_lake_gen2_filesystem_resource.go
+++ b/azurerm/internal/services/storage/storage_data_lake_gen2_filesystem_resource.go
@@ -139,6 +139,12 @@ func resourceStorageDataLakeGen2FileSystemCreate(d *schema.ResourceData, meta in
 		return fmt.Errorf("Error checking for existence of Storage Account %q (Resource Group %q): %+v", storageID.Name, storageID.ResourceGroup, err)
 	}
 
+	if acl != nil && (storageAccount.AccountProperties == nil ||
+		storageAccount.AccountProperties.IsHnsEnabled == nil ||
+		!*storageAccount.AccountProperties.IsHnsEnabled) {
+		return fmt.Errorf("ACL is enabled only when the Hierarchical Namespace (HNS) feature is turned ON")
+	}
+
 	fileSystemName := d.Get("name").(string)
 	propertiesRaw := d.Get("properties").(map[string]interface{})
 	properties := ExpandMetaData(propertiesRaw)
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/storage/storage_data_lake_gen2_filesystem_resource.go
```diff
@@ -292,24 +296,19 @@ func resourceStorageDataLakeGen2FileSystemRead(d *schema.ResourceData, meta inte
...
+	if storageAccount.IsHnsEnabled != nil && *storageAccount.IsHnsEnabled {
+		// The above `getStatus` API request doesn't return the ACLs
+		// Have to make a `getAccessControl` request, but that doesn't return all fields either!
+		pathResponse, err := pathClient.GetProperties(ctx, id.AccountName, id.DirectoryName, "/", paths.GetPropertiesActionGetAccessControl)
+		if err == nil {
+			acl, err := accesscontrol.ParseACL(pathResponse.ACL)
+			if err != nil {
+				return fmt.Errorf("Error parsing response ACL %q: %s", pathResponse.ACL, err)
+			}
+			d.Set("ace", FlattenDataLakeGen2AceList(acl))
```
tombuildsstuff, line 309: "ace" should always be set to a value, so we should assign `FlattenDataLakeGen2AceList` to an `[]interface{}` variable and then always set `ace` so that we can display drift here

azurerm/internal/services/storage/storage_data_lake_gen2_filesystem_resource_test.go
```diff
@@ -144,15 +144,30 @@ func (r StorageDataLakeGen2FileSystemResource) Destroy(ctx context.Context, clie
 }
 
 func (r StorageDataLakeGen2FileSystemResource) basic(data acceptance.TestData) string {
-	template := r.template(data)
 	return fmt.Sprintf(`
-%s
```
tombuildsstuff, line 149: this should be using the template instead, can we revert this?

azurerm/internal/services/storage/storage_data_lake_gen2_filesystem_resource_test.go
```diff
@@ -144,15 +144,30 @@ func (r StorageDataLakeGen2FileSystemResource) Destroy(ctx context.Context, clie
...
+  account_tier             = "Standard"
+  account_replication_type = "LRS"
+}
 
 resource "azurerm_storage_data_lake_gen2_filesystem" "test" {
   name               = "acctest-%d"
   storage_account_id = azurerm_storage_account.test.id
 }
-`, template, data.RandomInteger)
+`, data.RandomInteger, data.Locations.Primary, data.RandomString, data.RandomInteger)
```
tombuildsstuff, line 170: (as above)

tombuildsstuff (changes requested):
hi @njuCZ

Thanks for this PR - taking a look through I've left some comments inline, but if we can fix those up then we should be able to take another look.

Thanks!

jackofallops:
Tests passing:
image

katbyte (approved):
Thanks @njuCZ - LGTM 👍

PR #10471, opened 2021-02-04
refactor: removing the `tests `folder for authorization and datafactory

tombuildsstuff:
If a tests folder exists only a subset of the acctests get run - so these need to be moved to the root so all tests are run


katbyte (approved):
LGTM 👍

PR #10474, opened 2021-02-04
Support AzureAD Auth for azurerm_data_factory_integration_runtime_managed

neil-yechenwei:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/8994

--- PASS: TestAccDataFactoryIntegrationRuntimeManagedSsis_basic (155.25s)
--- PASS: TestAccDataFactoryIntegrationRuntimeManagedSsis_complete (226.08s)
--- PASS: TestAccDataFactoryIntegrationRuntimeManagedSsis_aadAuth (280.40s)


katbyte (changes requested):
Thanks @neil-yechenwei - this looks good but we've deprecated this resources in #10236 - could we update the PR to add this fix/tests on the new resource? thanks!

neil-yechenwei:
@katbyte , thanks for your comment. I've updated code. Please have an another look. Thanks.

katbyte (approved):
Thanks @neil-yechenwei - this LGTM now, might be a good idea to update the old resource too but this fixes it in the new one so gonna merge it

PR #10291, opened 2021-01-23
azurerm_lb_backend_address_pool: support for `backend_address`

magodo:
- Support `outbound_rules` for both ds and resource
- Support `backend_address` for both ds and resource
- Use the backend address pool dedicated endpoint in data source and resource Read callback to make the code easier to maintain

Note that the backend address pool API is kind of confusing, for which I've submit Azure/azure-rest-api-specs#11234. However, after offline ping with service team, it turns out they implement *Standard* sku LB and *Basic* sku LB in quite different way. They are not probably gonna change current API behavior in turns of backend address pool. See my [comment](https://github.com/Azure/azure-rest-api-specs/issues/11234#issuecomment-765871828) for more details.

Because of this fact, the implementation in this PR (mostly for CreateUpdate/Delete) will first check the LB sku, then call towards different endpoints accordingly.

Supersedes #8920

## Test Result

```shell
💢 make testacc TEST=./azurerm/internal/services/loadbalancer TESTARGS='-run TestAccAzureRMLoadBalancerBackEndAddressPool_'
==> Checking that code complies with gofmt requirements...
==> Checking that Custom Timeouts are used...
==> Checking that acceptance test packages are used...
TF_ACC=1 go test ./azurerm/internal/services/loadbalancer -v -run TestAccAzureRMLoadBalancerBackEndAddressPool_ -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccAzureRMLoadBalancerBackEndAddressPool_basicSkuBasic
=== PAUSE TestAccAzureRMLoadBalancerBackEndAddressPool_basicSkuBasic
=== RUN   TestAccAzureRMLoadBalancerBackEndAddressPool_standardSkuNIC
=== PAUSE TestAccAzureRMLoadBalancerBackEndAddressPool_standardSkuNIC
=== RUN   TestAccAzureRMLoadBalancerBackEndAddressPool_standardSkuIPBasic
=== PAUSE TestAccAzureRMLoadBalancerBackEndAddressPool_standardSkuIPBasic
=== RUN   TestAccAzureRMLoadBalancerBackEndAddressPool_standardSkuIPUpdate
=== PAUSE TestAccAzureRMLoadBalancerBackEndAddressPool_standardSkuIPUpdate
=== RUN   TestAccAzureRMLoadBalancerBackEndAddressPool_requiresImport
[...]

PR #10481, opened 2021-02-05
azurerm_lb_backend_address_pool: mark `backend_address` as computed 

magodo:
Fix the regression caused by #10291, that [the associated NIC will be cleared by LB BAP](https://github.com/terraform-providers/terraform-provider-azurerm/pull/10291#issuecomment-773691693)


jackofallops:
Tests passing:
image

PR #10488, opened 2021-02-05
New Resource: `azurerm_lb_backend_address_pool_address`

tombuildsstuff:
This PR introduces the new resource `azurerm_lb_backend_address_pool_address` - and force-deprecates the field `backend_addresses` within the `azurerm_backend_address_pool` resource, since this is only applicable when using a Standard Load Balancer.

Whilst nesting this within the `azurerm_backend_address_pool` does make sense in some circumstances - since these are additive (when a VM/VMSS is appended to the Backend Addresses list, this likely makes more sense as a separate resource, rather than making this Optional & Computed (as done in #10481) which, whilst works around this issue for Basic SKU's makes this harder to clear.

This PR force-deprecates the `backend_addresses` field added to the `azurerm_backend_address_pool` resource in #10291 - by ignoring the expanded value and setting the flattened value to an empty list to deprecate this field - to migrate people across.


tombuildsstuff:
Tests pass:

Screenshot 2021-02-05 at 18 02 12

azurerm/internal/services/loadbalancer/backend_address_pool_address_resource.go
```diff
@@ -0,0 +1,335 @@
...
+						break
+					}
+				}
+			}
+			if backendAddress == nil {
+				metadata.Logger.Warnf("%s was not found - removing from state!")
+				return metadata.MarkAsGone()
+			}
+
+			backendAddressPoolId := parse.NewLoadBalancerBackendAddressPoolID(id.SubscriptionId, id.ResourceGroup, id.LoadBalancerName, id.BackendAddressPoolName)
```
magodo, line 192: We might need further ensure this backend_address_pool_address resource is the IP based address, rather than the NIC based address. For the NIC based addresses, they only have `name` and users are not able to set the other arguments.

The check might be appropriated to do in the `CustomImporter()`?

azurerm/internal/services/loadbalancer/backend_address_pool_address_resource_test.go
```diff
@@ -0,0 +1,275 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("updating %s: %+v", *id, err)
+	}
+	if err := future.WaitForCompletionRef(ctx, client.LoadBalancers.LoadBalancerBackendAddressPoolsClient.Client); err != nil {
+		return nil, fmt.Errorf("waiting for update of %s: %+v", *id, err)
+	}
+	return utils.Bool(true), nil
+}
+
+func (BackendAddressPoolAddressResourceTests) backendAddressPoolHasAddresses(expected int) acceptance.ClientCheckFunc {
```
magodo, line 151: Is this method not used?

magodo (dismissed):
@tombuildsstuff I've walked through this PR and left some comments, otherwise it LGTM 👍

etaham:
Would it be possible for this resource to accept a list of IP addresses? Without that, it will require more instances of this resource to be created and managed.

etaham:
> Would it be possible for this resource to accept a list of IP addresses? Without that, it will require more instances of this resource to be created and managed.

I guess there would be complexity around mapping an IP to the correct instance of the resource - but I don't know the internals to know if this is an issue.

Thank you!

jackofallops:
Tests Passing:
image

katbyte (approved):
LGTM 👍

PR #10493, opened 2021-02-07
Minor tweak for SDK: return nil instead of return err

ArcturusZhang:
We do not have a `err` variable in this scope, therefore we are actually returning the `err` on line 29


tombuildsstuff (approved):
LGTM 👍

PR #10502, opened 2021-02-08
azurerm_data_factory - support for the customer_managed_key_id property

### LINKED ISSUE #10417: Support for ADF Encryption (in API v2) Settings
### Community Note

### Description

In the azurerm Azure Data Factory (v2):
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/data_factory
does not support the capabilities for passing the properties in Azure ADF v2 API, namely, the encryption object.

* azurerm_data_factory

```hcl
```

### References

* https://docs.microsoft.com/en-us/rest/api/datafactory/v1
* https://docs.microsoft.com/en-us/rest/api/datafactory/v2

* #0000

allantargino:
fix #10417 
This PR adds both [CMK (Customer Managed Keys) ](https://docs.microsoft.com/en-us/azure/data-factory/enable-customer-managed-key) and user assigned identities to ADF.
> Azure Data Factory currently only allows using CMK encryption at creation time when using user assigned identities.


allantargino:
I need some help with this PR 😢 
`TestAccDataFactory_userAssignedIdentity` is failing and I can't figure it out why the plan is different:
```
Running tool: /usr/local/go/bin/go test -timeout 3600s -run ^TestAccDataFactory_userAssignedIdentity$ github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/datafactory -v

=== RUN   TestAccDataFactory_userAssignedIdentity
=== PAUSE TestAccDataFactory_userAssignedIdentity
=== CONT  TestAccDataFactory_userAssignedIdentity
    /home/allan/repos/terraform/azurerm/azurerm/internal/services/datafactory/testing.go:684: Step 0 error: After applying this step and refreshing, the plan was not empty:

        DIFF:

        UPDATE: azurerm_data_factory.test
          github_configuration.#:         "0" => "0"
          id:                             "/subscriptions/x/resourceGroups/acctestrg-df-210208104113152502/providers/Microsoft.DataFactory/factories/acctest210208104113152502" => "/subscriptions/x/resourceGroups/acctestrg-df-210208104113152502/providers/Microsoft.DataFactory/factories/acctest210208104113152502"
          identity.#:                     "1" => "1"
          identity.0.principal_id:        "" => ""
          identity.0.tenant_id:           "" => ""
          identity.0.type:                "UserAssigned" => "UserAssigned"
          identity.0.user_identity_ids.#: "1" => "1"
          identity.0.user_identity_ids.0: "/subscriptions/x/resourcegroups/acctestRG-df-210208104113152502/providers/Microsoft.ManagedIdentity/userAssignedIdentities/acctest210208104113152502" => "/subscriptions/x/resourceGroups/acctestRG-df-210208104113152502/providers/Microsoft.ManagedIdentity/userAssignedIdentities/acctest210208104113152502"
          location:                       "brazilsouth" => "brazilsouth"
          name:                           "acctest210208104113152502" => "acctest210208104113152502"
          public_network_enabled:         "true" => "true"
          resource_group_name:            "acctestrg-df-210208104113152502" => "acctestrg-df-210208104113152502"
          vsts_configuration.#:           "0" => "0"
```

azurerm/internal/services/datafactory/data_factory_resource.go
```diff
@@ -70,12 +73,24 @@ func resourceDataFactory() *schema.Resource {
 							Required: true,
 							ValidateFunc: validation.StringInSlice([]string{
 								"SystemAssigned",
+								"UserAssigned",
 							}, false),
 						},
+
+						"user_identity_ids": {
```
katbyte, line 80: could we make this 
```suggestion
						"identity_ids": {
```

to be consistent with the rest of the provider?

azurerm/internal/services/datafactory/data_factory_resource.go
```diff
@@ -70,12 +73,24 @@ func resourceDataFactory() *schema.Resource {
...
+							Type:     schema.TypeSet,
```
ArcturusZhang, line 81: Also we might need this to be `TypeList` to be consistent with the rest of the provider. Also this might be why we are getting a non-empty plan during the acc test

ArcturusZhang:
And here is why we are getting the non-empty plan:
- /subscriptions/x/**resourcegroups**/acctestRG-df-210208104113152502/providers/Microsoft.ManagedIdentity/userAssignedIdentities/acctest210208104113152502
- /subscriptions/x/**resourceGroups**/acctestRG-df-210208104113152502/providers/Microsoft.ManagedIdentity/userAssignedIdentities/acctest210208104113152502

we could fix this by introducing a custom ID parser

ArcturusZhang:
Well it turns out that our user_assigned_identity resource already implemented the ID Formatters - therefore the root cause of the non-empty diff has been fixed. @allantargino could you please update your PR, resolve the merge conflict and try the test again?

azurerm/internal/services/datafactory/data_factory_resource.go
allantargino: done :)

azurerm/internal/services/datafactory/data_factory_resource.go
allantargino: done!

allantargino:
I renamed `key_vault_key_id` to `customer_managed_key_id` so it is aligned to #11328

allantargino:
@ArcturusZhang even after merging `main` into this branch, I am getting the same `resourcegroups` problem :(

```
              ~ identity {
                  ~ identity_ids = [
                      - "/subscriptions/x/resourcegroups/acctestRG-df-210426143744464185/providers/Microsoft.ManagedIdentity/userAssignedIdentities/acctest210426143744464185",
                      + "/subscriptions/x/resourceGroups/acctestRG-df-210426143744464185/providers/Microsoft.ManagedIdentity/userAssignedIdentities/acctest210426143744464185",
                    ]
                    # (1 unchanged attribute hidden)
                }
            }

```

katbyte:
@allantargino - the new parser will need to be used in "case insensitive mode" to parse and the pop out the correctly cased id

azurerm/internal/services/datafactory/data_factory_resource.go
```diff
@@ -421,12 +484,21 @@ func flattenDataFactoryIdentity(identity *datafactory.FactoryIdentity) interface
...
 		tenantId = identity.TenantID.String()
 	}
+	var identityIds []string
+	if identity.UserAssignedIdentities != nil {
+		userIdentities := make([]string, 0)
+		for key := range identity.UserAssignedIdentities {
+			userIdentities = append(userIdentities, key)
+		}
+		identityIds = userIdentities
+	}
```
ArcturusZhang, line 494: First we might need to simplify this part - we do not need two slices of strings here. And before we setting those IDs, we need to first parse it and then format it.
```suggestion
	var identityIds []string
	if identity.UserAssignedIdentities != nil {
		for key := range identity.UserAssignedIdentities {
		        id, err := msiparse.UserAssignedIdentityID(key)
		        if err != nil {
		              return nil, err
		        }
			identityIds = append(identityIds , id.ID())
		}
	}
```
In this way we can keep the casing of user assigned identity consistent

ArcturusZhang (changes requested):

azurerm/internal/services/datafactory/data_factory_resource.go
ArcturusZhang: also since the parsing function will return an error, you might need to change your function signature to make sure this function could return an error

allantargino:
Thank you @ArcturusZhang and @katbyte ! Super clear now!
I refactored the user assigned identity assignment part during the create/update since the API was returning a 400 when using the array empty (instead of nil).
I also included the user identity property on the data source, since it used the same flattening function as the resource.

katbyte (approved):
Thanks @allantargino - this LGTM 👍

PR #10511, opened 2021-02-09
testing: groundwork for Binary Testing

tombuildsstuff:
This PR adds the necessary groundwork for Binary Testing to work in a feature-toggled manner. There's a few components necessary for this to work:

1. Removing the remaining dependencies on `acceptance.AzureProvider` (in flight)
2. Upgrading the Plugin SDK to use `v1.16.0`
3. Turning on the feature toggle

As such whilst this work isn't immediately usable, after rebasing this on top of the changes from 1, including 2 and turning on 3 - both Binary Testing and the current testing framework both work.

Note to reviewers: the `features` block defined below is intentionally using the default feature-set - since the Exists functions within the test framework don't (and shouldn't) make use of them (but they are available during Terraform runtime)


jackofallops (approved):

PR #10512, opened 2021-02-09
monitor diagnostic: allow updating workspace ID in place

notchairmk:
Azure API allows for updating log analytics workspace ID in place


katbyte (approved):
Thanks @notchairmk - LGTM 👍

PR #10526, opened 2021-02-10
tests: adding `azuread` provider blocks

tombuildsstuff:
This becomes necessary for Binary Testing to be able to download the AzureAD Provider on demand rather than having this vendored into the codebase. Whilst the AzureAD Provider is available for the moment, it won't be going forwards - as such these provider blocks need to be defined.


jackofallops (approved):
LGTM 👍

Issue #10545, opened 2021-02-11
Terraform for azurerm provider: Event_Grid name limitation less than AZURE (50 char)


sti291271:
**Community Note**

- Please vote on this issue by adding a 👍 reaction to the original issue to help the community and maintainers prioritize this request

- Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request

- If you are interested in working on this issue or have submitted a pull request, please leave a comment

**Terraform (and AzureRM Provider) Version**
Terraform version: 0.14.5
AzureRM version : 2.45.1

**Affected Resource(s)**
azurerm_eventgrid_event_subscription

 I have to create an event grid subscription with a 59 characters name, but i receive an error:

Error: invalid value for name (EventGrid subscription name must be 3 - 50 characters long, contain only letters, numbers and hyphens.)

 I try to create it directly in the portal and it works,, i saw that we can use name with 64 characters long but terraform is limited to 50.  Is it possible to have the same limitation than the Microsoft AZURE one?

resource "azurerm_eventgrid_domain" "evg1" {
    name                = "${local.type["event_grid"]}-${var.appname}-${local.env_code}${local.regionality["northeurope"]}-01"
    location            = local.secondary_geo
    resource_group_name = azurerm_resource_group.rgp_evg.name
    input_schema        = "CloudEventSchemaV1_0"

    tags = {
      "environment" = lower(terraform.workspace)
      "application" = var.appname
      "deployed-by" = "iac"
    }
}

...

resource "azurerm_eventgrid_event_subscription" "evg1_evs" {
    name                = "${local.type["event_grid_sub"]}-${var.appname}-${local.env_code}${local.regionality["northeurope"]}-${var.evsname}"
    scope               = azurerm_eventgrid_domain_topic.omc_ne_evt.id
    event_delivery_schema         = "CloudEventSchemaV1_0"
    included_event_types          = ["${var.included_ev_typ}"]
    service_bus_topic_endpoint_id = azurerm_servicebus_topic.sbu_topic.id
    subject_filter {
      subject_begins_with         = "${var.subject}"
    }
}

**References:**
https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules

 Thanks for your help on this subject.

jackofallops:
Closed by 10583

PR #10583, opened 2021-02-14
FIX: r/azurerm_eventgrid_event_subscription - limitation less than AZURE (50 char)

jbinko:
Fix to address the bug described here: #10545
Small change in validation -> from 50 to 64 characters.

Tested with 
azurerm_eventgrid_event_subscription
and
azurerm_eventgrid_system_topic_event_subscription

both can support up to 64 characters in name now.

For len 64+ error message bellow will be shown:

Error: invalid value for name (EventGrid subscription name must be 3 - 64 characters long, contain only letters, numbers and hyphens.)
  on main.tf line 60, in resource "azurerm_eventgrid_event_subscription" "default":
  60: resource "azurerm_eventgrid_event_subscription" "default" {

Error: invalid value for name (EventGrid subscription name must be 3 - 64 characters long, contain only letters, numbers and hyphens.)
  on main.tf line 70, in resource "azurerm_eventgrid_system_topic_event_subscription" "example":
  70: resource "azurerm_eventgrid_system_topic_event_subscription" "example" {


katbyte (approved):
Thanks @jbinko - LGTM 👍

PR #10585, opened 2021-02-15
virtual_machine_scale_set datasource - support network_interfaces

shurikk:
Knowing scale set `network_profile` is important when creating load balancers as a separate process/state because the scale set and the load balancer must be on the same virtual network.


katbyte (changes requested):
Thanks for the pr @shurikk - could we update the docs to reflect this new property? thanks

shurikk:
yes, will do, was planning to but got distracted

azurerm/internal/services/compute/virtual_machine_scale_set_data_source.go
```diff
@@ -12,6 +12,14 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
 
+func schemaNetworkProfileForDataSource() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeSet,
+		Computed: true,
+		Elem:     resourceVirtualMachineScaleSet().Schema["network_profile"].Elem,
```
katbyte, line 19: We should pull that out into a shared schema function instead of doing this

katbyte (changes requested):
Thanks @shurikk - overall this looks good but instead of referring to the schema in another resource could we pull that out? I think the new windows|linux virtual resources already have done that so you should be able to use `"network_interface": VirtualMachineScaleSetNetworkInterfaceSchema(),`

azurerm/internal/services/compute/virtual_machine_scale_set_data_source.go
```diff
@@ -31,6 +39,8 @@ func dataSourceVirtualMachineScaleSet() *schema.Resource {
 
 			"location": azure.SchemaLocationForDataSource(),
 
+			"network_profile": schemaNetworkProfileForDataSource(),
```
katbyte, line 42: which would make this
```suggestion
			"network_interface": VirtualMachineScaleSetNetworkInterfaceSchema(),
```

shurikk:
@katbyte just to be on the same page, we want this change to be aligned with `linux_virtual_machine_scale_set` and `windows_virtual_machine_scale_set` rather than `virtual_machine_scale_set`? If yes, then the docs should be updated as well

katbyte:
@shurikk - correct, those are the new resources that will replace the old `virtual_machine_scale_set`. they share the `VirtualMachineScaleSetNetworkInterfaceSchema` so it should be fine to add this to the `virtual_machine_scale_set` data source

shurikk:
@katbyte updated, still need `schemaNetworkInterfaceForDataSource` to set `Computed: true`, cannot use `VirtualMachineScaleSetNetworkInterfaceSchema()` directly without that

katbyte:
oh right 🤦‍♀️🤦‍♀️ sorry i totally blanked on this, this is a datasource so we need to have everything be computed, which means copying that function's schema and changing everything to be computed

shurikk:
@katbyte how should we proceed from here? I've updated docs and the code to be aligned with `linux` and `windows` scale sets.

shurikk:
config

```
data azurerm_virtual_machine_scale_set vmss {
  name                = "test-vmss"
  resource_group_name = "test"
}

output x { 
  value = data.azurerm_virtual_machine_scale_set.vmss
}
```

terraform output

```
x = {
  "id" = "/subscriptions/.../.../Microsoft.Compute/virtualMachineScaleSets/test-vmss"
  "identity" = tolist([])
  "location" = tostring(null)
  "name" = "test-vmss"
  "network_interface" = tolist([
    {
      "dns_servers" = tolist([])
      "enable_accelerated_networking" = true
      "enable_ip_forwarding" = false
      "ip_configuration" = tolist([
        {
          "application_gateway_backend_address_pool_ids" = toset([])
          "application_security_group_ids" = toset([])
          "load_balancer_backend_address_pool_ids" = toset([
            "/subscriptions/.../.../backendAddressPools/test-lb",
          ])
          "load_balancer_inbound_nat_rules_ids" = toset([])
          "name" = "test-vmss"
          "primary" = true
          "public_ip_address" = tolist([])
          "subnet_id" = "/subscriptions/.../.../subnets/subnet12345"
          "version" = "IPv4"
        },
      ])
      "name" = "test-vmss"
      "network_security_group_id" = ""
      "primary" = true
    },
  ])
  "resource_group_name" = "test"
  "timeouts" = null /* object */
}
```

katbyte:
@shurikk - i hope you don't mind but I just pushed the changes required to get this merged, i'll have someone else review it and get it into this weeks release 🙂

shurikk:
> @shurikk - i hope you don't mind but I just pushed the changes required to get this merged, i'll have someone else review it and get it into this weeks release 🙂

thanks, works, it's not super urgent. also noticed some doc updates, still work in progress?

azurerm/internal/services/compute/virtual_machine_scale_set_data_source.go
```diff
@@ -82,12 +84,21 @@ func dataSourceVirtualMachineScaleSetRead(d *schema.ResourceData, meta interface
...
+		}
+	}
+
 	identity, err := FlattenVirtualMachineScaleSetIdentity(resp.Identity)
 	if err != nil {
 		return err
 	}
 	if err := d.Set("identity", identity); err != nil {
-		return fmt.Errorf("setting `identity`: %+v", err)
+		return fmt.Errorf("[DEBUG] Error setting `identity`: %+v", err)
```
WodansSon, line 101:
```suggestion
		return fmt.Errorf("setting `identity`: %+v", err)
```

WodansSon (approved):
@katbyte, thank you for this PR... I left one super minor comment and there are a few lint errors. Other than that this LGTM! 🚀

manicminer (approved):
LGTM! 🚀

azurerm/internal/services/compute/virtual_machine_scale_set.go
```diff
@@ -280,6 +321,70 @@ func virtualMachineScaleSetIPConfigurationSchema() *schema.Schema {
 	}
 }
 
+func virtualMachineScaleSetIPConfigurationSchemaForDataSource() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeList,
+		Required: true,
```
magodo, line 327: Should we change this to `Computed: true`?

azurerm/internal/services/compute/virtual_machine_scale_set.go
```diff
@@ -340,6 +445,53 @@ func virtualMachineScaleSetPublicIPAddressSchema() *schema.Schema {
 	}
 }
 
+func virtualMachineScaleSetPublicIPAddressSchemaForDataSource() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
```
magodo, line 451: Should we change this to `Computed: true`?

magodo (commented):
Shall we change the schemas above to be `Computed: true`?

PR #10598, opened 2021-02-15
subscription: exposing the `id` field

stawii:
Fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/10093

### DIFF
--- a/azurerm/internal/services/subscription/data_source_subscriptions.go
+++ b/azurerm/internal/services/subscription/data_source_subscriptions.go
@@ -34,6 +34,11 @@ func dataSourceSubscriptions() *schema.Resource {
 				Computed: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"resource_manager_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
 						"subscription_id": {
 							Type:     schema.TypeString,
 							Computed: true,
@@ -99,6 +104,9 @@ func dataSourceSubscriptionsRead(d *schema.ResourceData, meta interface{}) error
 
 		s := make(map[string]interface{})
 
+		if v := val.ID; v != nil {
+			s["resource_manager_id"] = *v
+		}
 		if v := val.SubscriptionID; v != nil {
 			s["subscription_id"] = *v
 		}
--- a/azurerm/internal/services/subscription/data_source_subscriptions.go
+++ b/azurerm/internal/services/subscription/data_source_subscriptions.go
@@ -34,7 +34,7 @@ func dataSourceSubscriptions() *schema.Resource {
 				Computed: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"resource_manager_id": {
+						"id": {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/subscription/data_source_subscriptions.go
```diff
@@ -34,6 +34,11 @@ func dataSourceSubscriptions() *schema.Resource {
 				Computed: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"resource_manager_id": {
```
tombuildsstuff, line 37: can we make this just `id` to match other resources:

```suggestion
						"id": {
```

azurerm/internal/services/subscription/data_source_subscriptions.go
```diff
@@ -99,6 +104,9 @@ func dataSourceSubscriptionsRead(d *schema.ResourceData, meta interface{}) error
 
 		s := make(map[string]interface{})
 
+		if v := val.ID; v != nil {
+			s["resource_manager_id"] = *v
```
tombuildsstuff, line 108: can we make this just `id` to match other resources:

```suggestion
			s["id"] = *v
```

azurerm/internal/services/subscription/data_source_subscriptions_test.go
```diff
@@ -17,6 +17,7 @@ func TestAccDataSourceSubscriptions_basic(t *testing.T) {
 		{
 			Config: SubscriptionsDataSource{}.basic(),
 			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).Key("subscriptions.0.resource_manager_id").Exists(),
```
tombuildsstuff, line 20: can we make this just `id` to match other resources:

```suggestion
				check.That(data.ResourceName).Key("subscriptions.0.id").Exists(),
```

website/docs/d/subscriptions.html.markdown
```diff
@@ -36,6 +36,7 @@ output "first_available_subscription_display_name" {
 
 The `subscription` block contains:
 
+* `resource_manager_id` - The Resource Manager ID of this subscription.
```
tombuildsstuff, line 39: can we make this just `id` to match other resources:

```suggestion
* `id` - The ID of this Subscription.
```

tombuildsstuff (changes requested):
hey @stawii 

Thanks for this PR.

Taking a look through this mostly LGTM - however can we rename this nested field to `id` rather than `resource_manager_id`?

Thanks!

website/docs/d/subscriptions.html.markdown
stawii: I was a bit confused if I should use `id` or `resource_manager_id` ;) It fits "better" and it compatible with azurerm_subscription, but I've seen `resource_manager_id` in a lot of different places too. Seems I've picked wrong.
I'll change it soon.

tombuildsstuff: no worries - we use `resource_manager_id` only where a secondary is present (e.g. data plane id) - otherwise this should be `id`

tombuildsstuff (approved):
LGTM - thanks for this @stawii

tombuildsstuff:
Tests pass:

Screenshot 2021-02-16 at 10 56 42

PR #10605, opened 2021-02-16
azurerm_virtual_network_gateway_connection: Prevent simultaneous activities

### LINKED ISSUE #10582: unable to create multiple azurerm_virtual_network_gateway_connection resources
### Community Note

### Terraform (and AzureRM Provider) Version - 

Terraform 0,14.5
AzureRM - v2.46.1

### Affected Resource(s)

* 'azurerm_virtual_network_gateway_connection'

```hcl
```
```hcl
resource "azurerm_virtual_network_gateway_connection" "expressroute_to_core" {
  count                      = length(var.gateway_core_snet)
  name                       = "${var.name}-${format(1 + count.index)}"
  resource_group_name        = var.resource_group_name
  location                   = var.az_region
  type                       = "ExpressRoute"
  express_route_circuit_id   = var.ecr_circuit
  virtual_network_gateway_id = var.gateway_core_snet[count.index]
}
```

### Debug Output

the first connection created fine, however, the next connections get the bellow error

```
Error: Error waiting for completion of Virtual Network Gateway Connection "conenction-2" (Resource Group "rg"): Code="ConflictError" Message="The current operation could not be executed because it is already in progress." Details=[]
```

### Expected Behaviour

The connection should be created one after another, without error

### Actual Behaviour

first connection in the count gets created fine, the next ones get above error and show as failed in the Azure portal. They are not imported in the state either. the only way to fix it is to manually remove the network connection from the Azure Portal and re-run the terraform again

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

this is running multiple connections one after another to the same ECR 

### References

* #0000

favoretti:
Prevent simultaneous activities activities

Fixes #10582

### DIFF
--- a/azurerm/internal/services/network/virtual_network_gateway_connection_resource.go
+++ b/azurerm/internal/services/network/virtual_network_gateway_connection_resource.go
@@ -2,6 +2,7 @@ package network
 
 import (
 	"fmt"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/locks"
 	"log"
 	"time"
 
@@ -16,6 +17,8 @@ import (
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
 )
+
+var virtualNetworkGatewayConnectionResourceName = "azurerm_virtual_network_gateway_connection"
 
 func resourceVirtualNetworkGatewayConnection() *schema.Resource {
 	return &schema.Resource{
@@ -298,6 +301,9 @@ func resourceVirtualNetworkGatewayConnectionCreateUpdate(d *schema.ResourceData,
 	name := d.Get("name").(string)
 	resGroup := d.Get("resource_group_name").(string)
 
+	locks.ByName(name, virtualNetworkGatewayConnectionResourceName)
+	defer locks.UnlockByName(name, virtualNetworkGatewayConnectionResourceName)
+
 	if d.IsNewResource() {
 		existing, err := client.Get(ctx, resGroup, name)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


azurerm/internal/services/network/virtual_network_gateway_connection_resource.go
```diff
@@ -455,6 +461,10 @@ func resourceVirtualNetworkGatewayConnectionDelete(d *schema.ResourceData, meta
 		return err
 	}
 
+	// We do not know the gateway name here, so have to lock on a fake name
```
katbyte, line 464: could we not lock on the ID or part of it instead?

favoretti: I think we could, but what would be the difference/benefit here?

azurerm/internal/services/network/virtual_network_gateway_connection_resource.go
```diff
@@ -326,6 +329,9 @@ func resourceVirtualNetworkGatewayConnectionCreateUpdate(d *schema.ResourceData,
 		VirtualNetworkGatewayConnectionPropertiesFormat: properties,
 	}
 
+	locks.ByName(*properties.VirtualNetworkGateway1.Name, virtualNetworkGatewayConnectionResourceName)
```
magodo, line 332: Shall we use the `locks.ByID` to lock against the ID of the including vnet gw instead? Otherwise, it might unecessarily locks two same named connections in different resource groups?

magodo (changes requested):
@favoretti Thank you for this fix!

I'm wondering whether we should use the `locks.ByID` rather than the `locks.ByName` to avoid unncessarily locks in some cases (e.g locks two same named connections in different resource groups).
Besides, it would be appreciated if you can add some tests for the use case as described in #10582.

favoretti:
Holding off on this for the moment as I've misunderstood the use-case initially and want to gather some intel from product team for a proper fix.

katbyte:
Hey @favoretti - any word on this or should we close it till your ready to move forwad?

favoretti:
There’s word, just no time atm to flip this around. Feel free to close,
I’ll reopen once I implemented it.

katbyte:
thx, will do @favoretti

PR #10608, opened 2021-02-16
Docs: AAAA record - add example and fix typo

danmichaelo:
Since the second example is using alias, I feel like `records` should be included in the first.

Fixed IPv4 → IPv6 typo


tombuildsstuff (approved):
LGTM - thanks for this @danmichaelo

PR #10610, opened 2021-02-16
azurerm_netapp_volume - support the `data_protection_replication` block

paulomarquesc:
This PR implements support for Cross-Region Replication (CRR) feature of ANF Volumes. 

Worth it to call out that since CRR is extremely sensitive on resources being complete, we extended usage  of stateCOnf.WaitForState for all steps of volume creation/deletion.


azurerm/internal/services/netapp/netapp_volume_data_source.go
```diff
@@ -75,6 +75,32 @@ func dataSourceNetAppVolume() *schema.Resource {
...
+							Required: true,
+						},
+
+						"replication_schedule": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+					},
+				},
+			},
```
katbyte, line 103: as these are in a data sould this should beb computed?

azurerm/internal/services/netapp/netapp_volume_resource.go
```diff
@@ -194,6 +195,41 @@ func resourceNetAppVolume() *schema.Resource {
...
+
+			"data_protection_replication": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"endpoint_type": {
+							Type:     schema.TypeString,
+							Required: true,
```
katbyte, line 207: if there is only one possible value should be default it to "dst" and mark it optional?

or even hardcode this and remove it for now?

azurerm/internal/services/netapp/netapp_volume_resource.go
```diff
@@ -194,6 +195,41 @@ func resourceNetAppVolume() *schema.Resource {
...
+								"dst",
+							}, true),
+						},
+
+						"remote_volume_location": azure.SchemaLocation(),
+
+						"remote_volume_resource_id": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
```
katbyte, line 218: could we validate this?

azurerm/internal/services/netapp/netapp_volume_resource.go
```diff
@@ -194,6 +195,41 @@ func resourceNetAppVolume() *schema.Resource {
...
+
+						"replication_schedule": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"_10minutely",
+								"daily",
+								"hourly",
+							}, true),
```
katbyte, line 227: could we make this case sensitive
```suggestion
							}, false),
```

azurerm/internal/services/netapp/netapp_volume_resource_test.go
```diff
@@ -248,6 +264,62 @@ resource "azurerm_netapp_volume" "test" {
 `, template, data.RandomInteger, data.RandomInteger)
 }
 
+func (NetAppVolumeResource) crr(data acceptance.TestData) string {
```
katbyte, line 267: could we expand crr to be unabbreviated

azurerm/internal/services/netapp/netapp_volume_resource_test.go
```diff
@@ -51,6 +51,22 @@ func TestAccNetAppVolume_nfsv41(t *testing.T) {
 	})
 }
 
+func TestAccNetAppVolume_crr(t *testing.T) {
```
katbyte, line 54: and here

azurerm/internal/services/netapp/netapp_volume_resource.go
```diff
@@ -496,3 +794,19 @@ func flattenNetAppVolumeMountIPAddresses(input *[]netapp.MountTargetProperties)
...
 }
+
+func flattenNetAppVolumeDataProtectionReplication(input *netapp.VolumePropertiesDataProtection) []interface{} {
+	if input == nil || input.Replication == nil || strings.ToLower(string(input.Replication.EndpointType)) != "dst" {
+		return []interface{}{}
+	}
+
+	return []interface{}{
+		map[string]interface{}{
+			"endpoint_type":             string(input.Replication.EndpointType),
```
katbyte, line 805: we will need to nil check Replication here

azurerm/internal/services/netapp/netapp_volume_resource.go
```diff
@@ -194,6 +195,41 @@ func resourceNetAppVolume() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"endpoint_type": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"dst",
```
katbyte, line 209: could we use the enum constants in the SDK here?

katbyte (changes requested):
Thanks for the PR @paulomarquesc - overall this looks good, but i've left a couple comments inline that once addressed this should be good to merge 🚀

azurerm/internal/services/netapp/netapp_volume_resource.go
paulomarquesc: The following check is done in the beginning of the function and I think it covers null check:

````go
	if input == nil || input.Replication == nil || strings.ToLower(string(input.Replication.EndpointType)) != "dst" {
		return []interface{}{}
	}
```

it checks the possible path, input then if replication is nil then if endpoint type is != "dst" wich nil would be caught as well.

azurerm/internal/services/netapp/netapp_volume_data_source.go
paulomarquesc: Yes, just fixed it.

azurerm/internal/services/netapp/netapp_volume_resource.go
paulomarquesc: I made it optional defaulting to "dst".

azurerm/internal/services/netapp/netapp_volume_resource.go
paulomarquesc: Yes, thanks, I included `ValidateFunc: azure.ValidateResourceID`

azurerm/internal/services/netapp/netapp_volume_resource.go
paulomarquesc: Changed to case sensitive.

azurerm/internal/services/netapp/netapp_volume_resource_test.go
paulomarquesc: Sure, replaced crr with crossRegionReplication

azurerm/internal/services/netapp/netapp_volume_resource_test.go
paulomarquesc: Done.

azurerm/internal/services/netapp/netapp_volume_resource.go
paulomarquesc: I would leave as is since the sdk provides "src" as well which is automatically included on primary volumes and we're hidding that operation on primaries since it is all automatically and using the constants will pass validation for "src".

paulomarquesc:
Hi @katbyte , just pushed new commit to fix your comments and CI linting issues.

azurerm/internal/services/netapp/netapp_snapshot_resource.go
```diff
@@ -71,6 +71,8 @@ func resourceNetAppSnapshot() *schema.Resource {
 				ValidateFunc: ValidateNetAppVolumeName,
 			},
 
+			// TODO: remove this in a next breaking changes release since tags are
+			// not supported anymore on Snapshots
```
katbyte, line 75: and we should add the 3.0 tag here

```suggestion
			// not supported anymore on Snapshots (todo 3.0)
```

azurerm/internal/services/netapp/netapp_volume_resource.go
katbyte: can we we still not use the enu. constant for dst by its self?
```suggestion
								string(sdk.dstvalue),
```

azurerm/internal/services/netapp/netapp_volume_resource.go
```diff
@@ -194,6 +194,43 @@ func resourceNetAppVolume() *schema.Resource {
...
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"endpoint_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "dst",
+							ValidateFunc: validation.StringInSlice([]string{
+								"dst",
+							}, true),
```
katbyte, line 211: could we not be case insensitive here
```suggestion
							}, falsee),
```

azurerm/internal/services/netapp/netapp_volume_resource.go
```diff
@@ -194,6 +194,43 @@ func resourceNetAppVolume() *schema.Resource {
...
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: azure.ValidateResourceID,
+						},
+
+						"replication_schedule": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"_10minutely",
```
katbyte, line 226: this is a fun value, maybe we should change it to
```suggestion
								"10minutes",
```

and inject the odd SDK value behind the scenes?

azurerm/internal/services/netapp/netapp_volume_resource.go
```diff
@@ -194,6 +194,43 @@ func resourceNetAppVolume() *schema.Resource {
...
+
+						"remote_volume_location": azure.SchemaLocation(),
+
+						"remote_volume_resource_id": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: azure.ValidateResourceID,
+						},
+
+						"replication_schedule": {
```
katbyte, line 222: given this is not so much a schedule as a frequency 
```suggestion
						"replication_frequency": {
```

azurerm/internal/services/netapp/netapp_volume_data_source.go
```diff
@@ -75,6 +75,32 @@ func dataSourceNetAppVolume() *schema.Resource {
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+
+			"data_protection_replication": {
+				Type:     schema.TypeList,
+				Computed: true,
+				MaxItems: 1,
```
katbyte, line 82:
```suggestion
```

azurerm/internal/services/netapp/netapp_volume_resource.go
```diff
@@ -194,6 +194,43 @@ func resourceNetAppVolume() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+
+			"data_protection_replication": {
```
katbyte, line 198: is there a reason this is called `data_protection_replication` when the sdk property is `data_protection` (thou i'd argue data_replication or even replication might suffice)

katbyte (changes requested):
Thanks @paulomarquesc - i've left some more comments inline

azurerm/internal/services/netapp/netapp_volume_resource.go
paulomarquesc: Completely agree with the oddness of the value, I'll work on the translation, trying to be consistent with our tooling where customers will still see _10minutelly

azurerm/internal/services/netapp/netapp_volume_resource.go
paulomarquesc: This one will need to stay as is since our documentation, portal, sdk and other tools references this object by that name.

azurerm/internal/services/netapp/netapp_volume_data_source.go
paulomarquesc: If I remove MaxItems: 1, will this enforce the same rule of not allowing more than one "data_protection_replication" ?

azurerm/internal/services/netapp/netapp_volume_resource.go
paulomarquesc: There will be other types of data_protection coming up, e.g. snapshotpolicy, backup, etc.. So keeping this naming will make it consistent with that customers see in the portal and other tools and will make sense to them.

azurerm/internal/services/netapp/netapp_volume_resource.go
paulomarquesc: I thought I changed it.

azurerm/internal/services/netapp/netapp_volume_data_source.go
paulomarquesc: I looked at options like AtLeastOneOf and ExactOneOf by they will make this attribute required and we can't have it, maybe I'm thinking if we can extend the class that implements these and include a third option, e.g. WhenInUseExactOneOf. Thoughts on that?

paulomarquesc: If that is ok, I forked the Terraform SDK and committed what I think it is the implementation for that, let me know if this is a good idea and I can move on with adding the testcases and place a PR on the SDK. Is there a way to test this on my provider repo? Wanted to run some tests before the PR on SDK. I changed my local  vendor\github.com\hashicorp\terraform-plugin-sdk\helper\schema\schema.go but this attribute still cannot be recognized.

image

paulomarquesc: Just realized I didn't paste the sdk commit for you to check: https://github.com/paulomarquesc/terraform-plugin-sdk/commit/73b5f2b1a37d2e49f23c3738ea01ddc16c12bf4a

azurerm/internal/services/netapp/netapp_snapshot_resource.go
```diff
@@ -101,9 +110,14 @@ func resourceNetAppSnapshotCreate(d *schema.ResourceData, meta interface{}) erro
 
 	location := azure.NormalizeLocation(d.Get("location").(string))
 
+	// Snapshot resource in Azure changed its type to proxied resource, therefore
+	// tags are not supported anymore, ignoring any tags.
```
katbyte, line 114:
```suggestion
```

katbyte (approved):
Thanks @paulomarquesc - aside from a few minor comments this LGTM now ! 👍

paulomarquesc:
Thanks Katie.

PR #10636, opened 2021-02-18
azurerm_api_management_custom_domain: resource gets updated every time

patst:
A possible solution for https://github.com/terraform-providers/terraform-provider-azurerm/issues/10253 .

As described in the issue it is unclear if the default hostname for a proxy (`<apim-name>.azure-api.net`) needs to be included in the `azurerm_api_management_custom_domain` resource. The test cases include the default hostname, the example in the docs don't.

In both cases there are problems using the `azurerm_api_management_custom_domain` resource at the moment because `tf plan` will try to apply changes on subsequent runs (see issue).

My solution filters out the default hostname (as the `azurerm_api_management` resource does as well, see https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/apimanagement/api_management_resource.go#L986 ).

The test cases are succeeding and my local tests are successful with this behaviour:

```
make testacc TEST='./azurerm/internal/services/apimanagement/api_management_custom_domain_resource_test.go' TESTARGS='-run=TestAccApiManagementCustomDomain_update' TESTTIMEOUT='120m'
TF_ACC=1 go test ./azurerm/internal/services/apimanagement/api_management_custom_domain_resource_test.go -v -run=TestAccApiManagementCustomDomain_update -timeout 120m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccApiManagementCustomDomain_update
=== PAUSE TestAccApiManagementCustomDomain_update
=== CONT  TestAccApiManagementCustomDomain_update
--- PASS: TestAccApiManagementCustomDomain_update (6269.11s)
PASS
ok      command-line-arguments  6271.038s


make testacc TEST='./azurerm/internal/services/apimanagement/api_management_custom_domain_resource_test.go' TESTARGS='-run=TestAccApiManagementCustomDomain_basic' TESTTIMEOUT='120m'
TF_ACC=1 go test ./azurerm/internal/services/apimanagement/api_management_custom_domain_resource_test.go -v -run=TestAccApiManagementCustomDomain_basic -timeout 120m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
=== RUN   TestAccApiManagementCustomDomain_basic
=== PAUSE TestAccApiManagementCustomDomain_basic
=== CONT  TestAccApiManagementCustomDomain_basic
--- PASS: TestAccApiManagementCustomDomain_basic (4115.71s)
PASS
ok      command-line-arguments  4118.349s
```

### DIFF
--- a/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
+++ b/azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
@@ -163,6 +163,7 @@ func apiManagementCustomDomainCreateUpdate(d *schema.ResourceData, meta interfac
 
 func apiManagementCustomDomainRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*clients.Client).ApiManagement.ServiceClient
+	environment := meta.(*clients.Client).Account.Environment
 	ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
 	defer cancel()
 
@@ -187,7 +188,8 @@ func apiManagementCustomDomainRead(d *schema.ResourceData, meta interface{}) err
 	d.Set("api_management_id", resp.ID)
 
 	if resp.ServiceProperties != nil && resp.ServiceProperties.HostnameConfigurations != nil {
-		configs := flattenApiManagementHostnameConfiguration(resp.ServiceProperties.HostnameConfigurations, d)
+		apimHostNameSuffix := environment.APIManagementHostNameSuffix
+		configs := flattenApiManagementHostnameConfiguration(resp.ServiceProperties.HostnameConfigurations, d, *resp.Name, apimHostNameSuffix)
 		for _, config := range configs {
 			for key, v := range config.(map[string]interface{}) {
 				// lintignore:R001
@@ -305,7 +307,7 @@ func expandApiManagementCustomDomains(input *schema.ResourceData) *[]apimanageme
 	return &results
 }
 
-func flattenApiManagementHostnameConfiguration(input *[]apimanagement.HostnameConfiguration, d *schema.ResourceData) []interface{} {
+func flattenApiManagementHostnameConfiguration(input *[]apimanagement.HostnameConfiguration, d *schema.ResourceData, name, apimHostNameSuffix string) []interface{} {
 	results := make([]interface{}, 0)
 	if input == nil {
 		return results
@@ -319,6 +321,11 @@ func flattenApiManagementHostnameConfiguration(input *[]apimanagement.HostnameCo
 
 	for _, config := range *input {
 		output := make(map[string]interface{})
+
+		// There'll always be a default custom domain with hostName "apim_name.azure-api.net" and Type "Proxy", which should be ignored
+		if *config.HostName == strings.ToLower(name)+"."+apimHostNameSuffix && config.Type == apimanagement.HostnameTypeProxy {
+			continue
+		}
 
 		if config.HostName != nil {
 			output["host_name"] = *config.HostName
--- a/website/docs/r/api_management_custom_domain.html.markdown
+++ b/website/docs/r/api_management_custom_domain.html.markdown
@@ -135,7 +135,7 @@ A `developer_portal`, `management`, `portal` or `scm` block supports the followi
 
 A `proxy` block supports the following:
 
--> **Tip:** The default hostname ending with `.azure-api.net` must not be added as it will be included by default and it is filtered out, so it would always show in `tf plan` but never have any effect.
+-> **Tip:** The default proxy hostname ending with `.azure-api.net` must not be added as it will be automatically created by Azure and ignored by Terraform.
 
 * `host_name` - (Required) The Hostname to use for the API Proxy Endpoint.
 
--- a/website/docs/r/api_management_custom_domain.html.markdown
+++ b/website/docs/r/api_management_custom_domain.html.markdown
@@ -135,6 +135,8 @@ A `developer_portal`, `management`, `portal` or `scm` block supports the followi
 
 A `proxy` block supports the following:
 
+-> **info:** The default hostname ending with `.azure-api.net` must not be added as it will be included by default.
+
 * `host_name` - (Required) The Hostname to use for the API Proxy Endpoint.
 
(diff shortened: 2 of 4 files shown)


sirlatrom:
As the original author of the resource, I'm pleased to see you found a solution to this quirk 😄 How will existing configurations be affected? If they include the default hostname today, will it appear twice in the config (thus becoming invalid) or will it be filtered out then, too?

patst:
@sirlatrom I just tried that. It will just work. You won't get a diff in `tf plan` if you have specified the default hostname at the moment.
If you delete it there will be no change in `tf plan`  as well.

sirlatrom:
> It will just work.

@patst That is excellent! Thanks again for the improvement!

patst:
@sirlatrom  sorry my test had a little error. I updated my post above.

sirlatrom:
> @sirlatrom sorry my test had a little error. I updated my post above.

@patst Well that's the second-best possible outcome and is still fine, as it will help current users see they don't need to add it anymore. Maybe it should be added to the website docs for the resource, in order to help people understand what is happening if they have been copying the current example from the dosc?

patst:
@sirlatrom thanks for the feedback, good point. I added a note in the resource documentation to not include the default hostname

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
```diff
@@ -305,7 +307,7 @@ func expandApiManagementCustomDomains(input *schema.ResourceData) *[]apimanageme
 	return &results
 }
 
-func flattenApiManagementHostnameConfiguration(input *[]apimanagement.HostnameConfiguration, d *schema.ResourceData) []interface{} {
+func flattenApiManagementHostnameConfiguration(input *[]apimanagement.HostnameConfiguration, d *schema.ResourceData, name, apimHostNameSuffix string) []interface{} {
```
manicminer, line 310: I think it would be cleaner to pass in the full hostname to be ignored, and determine that hostname before calling `flattenApiManagementHostnameConfiguration()`. If it accepts a slice that effectively makes it optional, and allows for potential future additions relatively easily.

```suggestion
func flattenApiManagementHostnameConfiguration(input *[]apimanagement.HostnameConfiguration, ignoreHostname []string) []interface{} {
```

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
```diff
@@ -320,6 +322,11 @@ func flattenApiManagementHostnameConfiguration(input *[]apimanagement.HostnameCo
 	for _, config := range *input {
 		output := make(map[string]interface{})
 
+		// There'll always be a default custom domain with hostName "apim_name.azure-api.net" and Type "Proxy", which should be ignored
+		if *config.HostName == strings.ToLower(name)+"."+apimHostNameSuffix && config.Type == apimanagement.HostnameTypeProxy {
```
manicminer, line 326: As per above, can we move this logic outside the flatten function?

website/docs/r/api_management_custom_domain.html.markdown
```diff
@@ -135,6 +135,8 @@ A `developer_portal`, `management`, `portal` or `scm` block supports the followi
 
 A `proxy` block supports the following:
 
+-> **info:** The default hostname ending with `.azure-api.net` must not be added as it will be included by default.
```
manicminer, line 138:
```suggestion
-> **Tip:** The default hostname ending with `.azure-api.net` must not be added as it will be included by default.
```

manicminer (changes requested):
Thanks for this @patst, this is looking good and I think it's a reasonable compromise to have a persistent plan in pursuit of removing this boilerplate, as long as its nondestructive. I still need to test this myself, but I have a suggestion about simplifying the flatten function in the meantime.

website/docs/r/api_management_custom_domain.html.markdown
sirlatrom: Maybe append to
>as it will be included by default **and it is filtered out, so it would always show in `tf plan` but never have any effect**.

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
patst: @manicminer thanks for the idea. 

I could move the whole filtering logic on layer up into the `apiManagementCustomDomainRead` function then it wouldn't be required to pass the `ignoreHostname[]` into the flatten  function at all.

One thing to consider: Right now it is exactly the same implementation as in the `api_management_resource` (see: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/apimanagement/api_management_resource.go#L1019 ). My original idea was to keep the logic and code similar to that.

What do you think?

website/docs/r/api_management_custom_domain.html.markdown
patst: Changed to `Tip:` and added the information why

azurerm/internal/services/apimanagement/api_management_custom_domain_resource.go
manicminer: @patst I'd forgotten that the parent resource had the same logic - in that case let's leave your original implementation as-is for now and disregard my suggestion :)

website/docs/r/api_management_custom_domain.html.markdown
```diff
@@ -135,6 +135,8 @@ A `developer_portal`, `management`, `portal` or `scm` block supports the followi
 
 A `proxy` block supports the following:
 
+-> **Tip:** The default hostname ending with `.azure-api.net` must not be added as it will be included by default and it is filtered out, so it would always show in `tf plan` but never have any effect.
```
manicminer, line 138:
```suggestion
-> **Tip:** The default proxy hostname ending with `.azure-api.net` must not be added as it will be automatically created by Azure and ignored by Terraform.
```

manicminer:
Awaiting acceptance test results

katbyte (approved):
tests passed, LGTM 👍

PR #10642, opened 2021-02-18
`azurerm_batch_certificate`: Allow empty `password` when format is pfx

austin-ek:
When creating a batch certificate, the password is not required by the API and should not be required by the terraform provider. It is optional in other interfaces, such as [ARM templates](https://docs.microsoft.com/en-us/azure/templates/microsoft.batch/batchaccounts/certificates#certificatecreateorupdateproperties-object), [REST API](https://docs.microsoft.com/en-us/rest/api/batchmanagement/certificate/create#request-body), and [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/batch/certificate?view=azure-cli-latest#az_batch_certificate_create-optional-parameters)


mbfrahry (approved):
LGTM!

PR #10646, opened 2021-02-19
"azurerm_monitor_smart_detector_alert_rule" - supports "tags"

njuCZ:
supports missing property "tags"

image


katbyte (approved):
Thanks @njuCZ - LGTM 👍

PR #10647, opened 2021-02-19
vm feature: `power_off_on_deletion`

magodo:
Introduce a new feature flag for vm `power_off_on_deletion` which allows users to speed up destroying vm instance in case it is a relatively small ones or those who don't care too much about chargings during deletion.

This flag defaults to `true`, which aligns with current behavior.

This PR also modifies the wordings for the `graceful_shutdown` a bit to emphasize to avoid confusion like "if it is set to false, there will be no poweroff happens during destroy".

### DIFF
--- a/azurerm/internal/features/defaults.go
+++ b/azurerm/internal/features/defaults.go
@@ -15,6 +15,7 @@ func Default() UserFeatures {
 		},
 		VirtualMachine: VirtualMachineFeatures{
 			DeleteOSDiskOnDeletion: true,
+			PowerOffOnDeletion:     true,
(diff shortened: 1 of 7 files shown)


tombuildsstuff:
👋 

As discussed offline, unfortunately this isn't something we plan to support as a feature-toggle at this time, since there's reliability benefits further down the dependency chain (e.g. working around issues in other API's) by ensuring that the virtual machine is shut-down and no longer using resources; as such whilst I'd like to thank you for this contribution I'm going to close this PR for the moment.

Thanks!

Issue #10634, opened 2021-02-18
Unnecessary restriction on number of match_condition elements within azurerm_frontdoor_firewall_policy


thecodetinker:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.13.6
azurerm v2.47.0

### Affected Resource(s)

* `azurerm_frontdoor_firewall_policy`

### Terraform Configuration Files

```hcl
locals {
    ip_list = [
        "192.168.0.1",
        "192.168.0.2",
        "192.168.0.3",
        # etc... this is just for test and could be done with a CIDR format, but in our case these IPs are not contiguous
        "192.168.0.101",
        "192.168.0.102"
    ]
}

resource "azurerm_frontdoor_firewall_policy" "waf-policy" {
  name                = "WafPolicy"
  resource_group_name = "Main"
  enabled             = true

  custom_rule {
    name     = "IPAllowList"
    enabled  = true
    priority = 1
    type     = "MatchRule"
    action   = "Block"

    match_condition {
      match_variable     = "RemoteAddr"
      operator           = "IPMatch"
      negation_condition = true
      match_values       = local.ip_list
    }
  }
}
```

### Expected Behaviour

All the match conditions should be added to the WAF Policy custom rule.

### Actual Behaviour

Terraform plan reported an error:

```
Error: custom_rule.0.match_condition.0.match_values: attribute supports 100 item maximum, config has 109 declared

  on front-door-waf-policy.tf line 1, in resource "azurerm_frontdoor_firewall_policy" "waf-policy":
   1: resource "azurerm_frontdoor_firewall_policy" "waf-policy" {

```

### Steps to Reproduce

We wish to restrict access to a website to only known IP ranges via the Front Door WAF policy. Being a large organisation, this list is quite long, with various global express routes, VPN egress points, access via firewalls etc. Therefore we need a large number of match conditions on the custom rule. 

[...]

PR #10659, opened 2021-02-19
FIX: r/frontdoor_firewall_policy - Restriction on number of match_condition elements

jbinko:
Fix for: #10634

Based on this:
https://github.com/MicrosoftDocs/azure-docs/blob/master/includes/front-door-limits.md

match_condition can support only up to 10 items
Verified in Azure it can support only up to 10 items – not more.
Otherwise error:
Error: creating Front Door Firewall Policy "examplefdwafpolicy" (Resource Group "example-rg"): frontdoor.PoliciesClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="BadRequest" Message="WebApplicationFirewallPolicy validation failed. More information \"Rule Rule1 exceeds 10 match conditions\"."
Because of that I reduced limit from 100 to 10 for match_condition

match_values can support up to 600 items
Verified in Azure it can support up to 600 items for IP Addresses. (operator= "IPMatch")
Otherwise error:
Error: creating Front Door Firewall Policy "examplefdwafpolicy" (Resource Group "example-rg"): frontdoor.PoliciesClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="BadRequest" Message="WebApplicationFirewallPolicy validation failed. More information \"Policy includes a rule which exceeds 600 IPMatch values\"."

match_values can support up to 10 items
Verified in Azure it can support up to 10 items for string match values. (operator= "stringOperator")
Otherwise error:
Error: creating Front Door Firewall Policy "examplefdwafpolicy" (Resource Group "example-rg"): frontdoor.PoliciesClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="BadRequest" Message="WebApplicationFirewallPolicy validation failed. More information \"Match condition in Rule1 rule has more than 10 match values.\"."

Because of that I increased limit from 100 to 600 for match_values

Match string value can be up to 256 characters.
Verified in Azure - it can support up to 256 characters in string items
Otherwise error:
Error: creating Front Door Firewall Policy "examplefdwafpolicy" (Resource Group "example-rg"): frontdoor.PoliciesClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="BadRequest" Message="WebApplicationFirewallPolicy validation failed. More information \"Match condition in Rule1 rule has match value with more than 256 characters\"."

I changed documentation and also changed the example name which was not valid for this resource.

More comments:
Ideally this requires conditional validation. 
For case IPMatch it can support up to 600 items. This is total number across of all match_values in all match_condition in the one policy
For case string operation it can support up to 10 items only. This is number in match_values only so total number can be higher across of all match_values in all match_condition in the one policy.
Not sure about what is the total number if mixing IPMatch and string types in multiple match_condition blocks.

Do we have any example where we use such sophisticated conditional validator? Any guidance for this?

This initial implementation of fix is conceptually the same as original implementation. It is just limiting to the highest number, lower limits will be handled by Azure as before.

### DIFF
--- a/azurerm/internal/services/frontdoor/frontdoor_firewall_policy_resource.go
+++ b/azurerm/internal/services/frontdoor/frontdoor_firewall_policy_resource.go
@@ -152,7 +152,7 @@ func resourceFrontDoorFirewallPolicy() *schema.Resource {
 						"match_condition": {
 							Type:     schema.TypeList,
 							Optional: true,
-							MaxItems: 100,
+							MaxItems: 10,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"match_variable": {
@@ -174,10 +174,10 @@ func resourceFrontDoorFirewallPolicy() *schema.Resource {
 									"match_values": {
 										Type:     schema.TypeList,
 										Required: true,
-										MaxItems: 100,
+										MaxItems: 600,
 										Elem: &schema.Schema{
 											Type:         schema.TypeString,
-											ValidateFunc: validation.StringIsNotEmpty,
+											ValidateFunc: validation.StringLenBetween(1, 256),
 										},
 									},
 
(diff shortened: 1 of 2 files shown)


katbyte (approved):
Thanks @jbinko - this LGTM 👍

PR #10665, opened 2021-02-20
new resource: `azurerm_sentinel_data_connector_azure_active_directory`

magodo:
new resource: `azurerm_sentinel_data_connector_azure_active_directory`


azurerm/internal/services/sentinel/sentinel_data_connector_azure_active_directory.go
```diff
@@ -0,0 +1,165 @@
...
+
+	if d.IsNewResource() {
+		resp, err := client.Get(ctx, id.ResourceGroup, OperationalInsightsResourceProvider, id.WorkspaceName, name)
+		if err != nil {
+			if !utils.ResponseWasNotFound(resp.Response) {
+				return fmt.Errorf("checking for existing %s: %+v", id, err)
+			}
+		}
+
+		return tf.ImportAsExistsError("azurerm_sentinel_data_connector_azure_active_directory", id.ID())
```
tombuildsstuff, line 84: this'll need to be wrapped:

```suggestion
		if !utils.ResponseWasNotFound(resp.Response) {
			return tf.ImportAsExistsError("azurerm_sentinel_data_connector_azure_active_directory", id.ID())
		}
```

azurerm/internal/services/sentinel/sentinel_data_connector_azure_active_directory.go
```diff
@@ -0,0 +1,165 @@
...
+			return nil
+		}
+
+		return fmt.Errorf("retrieving %s: %+v", id, err)
+	}
+
+	if err := assertDataConnectorKind(resp.Value, securityinsight.DataConnectorKindAzureActiveDirectory); err != nil {
+		return fmt.Errorf("asserting %s: %+v", id, err)
+	}
+	dc := resp.Value.(securityinsight.AADDataConnector)
```
tombuildsstuff, line 140: we can remove the `assert` func in favour of:

```suggestion
	dc, ok := resp.Value.(securityinsight.AADDataConnector)
	if !ok {
		return fmt.Errorf("%s was not an Azure Active Directory Data Connector", id)
	}
```

tombuildsstuff (changes requested):
hey @magodo 

Thanks for rebasing this PR - I've taken a look through and left a couple of comments inline but this otherwise LGTM 👍 

Thanks!

azurerm/internal/services/sentinel/sentinel_data_connector_azure_active_directory.go
magodo: In case the the status code is 404, the `err` is always non-nil. In other words, if the `err` is nil, then this condition (`!utils.ResponseWasNotFound(resp.Response)`) is always true.

tombuildsstuff (approved):
LGTM 👍

PR #10694, opened 2021-02-23
new resource "azurerm_security_center_assessment" and rename "azurerm_security_center_assessment_metadata" to "azurerm_security_center_assessment_policy"

njuCZ:
api doc: https://docs.microsoft.com/en-us/rest/api/securitycenter/assessments/createorupdate

according to the doc: an assessment metadata must be predefined and the name should be the same, so in the schema design, I add a property "assessment_metadata_id"

image

business description:
Assessment metadata represents static data on an assessment type. Meaning it is a resource you create and usually doesn’t change over time.
It contains much static data such as description, assessment severity, threats and more.
Assessment metadata can only be created at subscription level since it only represents metadata of assessments, therefore doesn’t apply to specific resources.

Assessments are specific instances of the corresponding Assessment Metadata.
Assessment is an extension resource over an existing resource. 
For example- you can have Assessment Metadata which describes a VM has an open port, and a corresponding Assessment can be created on a specific VM which specifies that VM has an open port.


katbyte (changes requested):
Thanks @njuCZ - i've left one comment inline, in addition to that could you explain how this resource/service work so i can better review the PR?

azurerm/internal/services/securitycenter/parse/assessment.go
```diff
@@ -0,0 +1,42 @@
+package parse
```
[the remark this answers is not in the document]
njuCZ: It could not be generated,  because the resource id starts from another resource

njuCZ:
Hi @katbyte, I have added business description between assessment and assessment metadata. Could you have a review again?

website/docs/r/security_center_assessment.html.markdown
```diff
@@ -0,0 +1,131 @@
...
+
+    ip_configuration {
+      name      = "internal"
+      primary   = true
+      subnet_id = azurerm_subnet.internal.id
+    }
+  }
+}
+
+resource "azurerm_security_center_assessment_metadata" "example" {
```
katbyte, line 72: Sadly i think this should have be called `azurerm_security_center_assessment_policy` to be explicit and clear what its doing. could we copy that resource to `azurerm_security_center_assessment_policy` and deprecate the old one?

website/docs/r/security_center_assessment.html.markdown
```diff
@@ -0,0 +1,131 @@
...
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `assessment_metadata_id` - (Required) The ID of the security Assessment Metadata resource. Changing this forces a new security Assessment to be created.
+
+* `target_resource_id` - (Required) The ID of the target resource. Changing this forces a new security Assessment to be created.
```
katbyte, line 94: what type of resources does this apply to? just VMs?

website/docs/r/security_center_assessment.html.markdown
```diff
@@ -0,0 +1,131 @@
...
+* `read` - (Defaults to 5 minutes) Used when retrieving the Security Center Assessment.
+* `update` - (Defaults to 30 minutes) Used when updating the Security Center Assessment.
+* `delete` - (Defaults to 30 minutes) Used when deleting the Security Center Assessment.
+
+## Import
+
+Security Assessment can be imported using the `resource id`, e.g.
+
+```shell
+terraform import azurerm_security_center_assessment.example /subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/resGroup1/providers/Microsoft.Compute/virtualMachineScaleSets/vmss1/providers/Microsoft.Security/assessments/00000000-0000-0000-0000-000000000000
```
katbyte, line 130: will this not change depending on the type of resource?

website/docs/r/security_center_assessment.html.markdown
```diff
@@ -0,0 +1,131 @@
...
+    code = "Healthy"
+  }
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `assessment_metadata_id` - (Required) The ID of the security Assessment Metadata resource. Changing this forces a new security Assessment to be created.
```
katbyte, line 92: this would then be
```suggestion
* `assessment_policy_id` - (Required) The ID of the security Assessment policy to apply to this resource. Changing this forces a new security Assessment to be created.
```

katbyte (changes requested):
Thanks @njuCZ - i've left some comments inline, and after figuring out how this service works and whats it does i think `azurerm_security_center_assessment_metadata` should have been named `azurerm_security_center_assessment_policy` - WDYT?

website/docs/r/security_center_assessment.html.markdown
njuCZ: It could be many kinds of resource. Not only VMs

website/docs/r/security_center_assessment.html.markdown
njuCZ: it will change. So should I replace with the following?
```
terraform import azurerm_security_center_assessment.example {scope}/providers/Microsoft.Security/assessments/00000000-0000-0000-0000-000000000000
```

website/docs/r/security_center_assessment.html.markdown
njuCZ: done

njuCZ:
@katbyte Have updated this PR. added new resource "azurerm_security_center_assessment_policy" and deprecated "azurerm_security_center_assessment_metadata"

website/docs/r/security_center_assessment.html.markdown
katbyte: i'd just have two examples

katbyte (approved):
Thanks @njuCZ - LGTM 👍

PR #10711, opened 2021-02-24
Update redis patch schedule documentation

michaelelleby:
Scheduled updates is now available for all Azure Cache for Redis SKUs.


manicminer (approved):
Hi @michaelelleby, thanks for this patch. I've tested with a Basic SKU and scheduled updates work for me!

Issue #1608, opened 2018-07-19
r/resource_group: feature flag for only deleting when empty / no nested resources


StevenHessing:
When deleting resource groups created using the azurerm provider, any Azure RDFE resources created previously or afterwards under the same resource group using the Azure portal or other tooling are also deleted.

This unfortunately appears to be by-design by Azure: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-deployment-model

This causes Terraform to delete RDFE resources it never created and which it does not have in its state file. 

As solution, Terraform should check whether there are RDFE resources in the resource group it is about to delete. If there are, Terraform should reject executing its plan until the user has rectified the situation.

Terraform v0.11.7
terraform-provider-azurerm_v1.7.0_x4

tombuildsstuff:
hi @StevenHessing 

Thanks for opening this issue :)

> This unfortunately appears to be by-design by Azure: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-deployment-model

Unfortunately as you've mentioned this is by design within Azure - whilst Terraform could check for the presence of items in a Resource Group and fail to delete it, I believe users would expect Terraform to mirror the behaviour they're seeing elsewhere. That said, we could potentially add an override flag to this (in a similar way we have for the `azurerm_virtual_machine` resource to delete the OS/Data Disks) to prevent deletion if child-resources still exist, which would allow both the existing behaviour (and this request to work simultaneously) - what do you think?

Thanks!

dmedeiros204:
Might have saved this guy from making a "big mistake" https://www.craigstuntz.com/posts/2021-07-08-how-i-made-a-giant-mistake-with-terraform.html

PR #10718, opened 2021-02-24
**New Resource:** `azurerm_subscription`

### LINKED ISSUE #1216: Create subscriptions programmatically using Azure Resource Manager
### Community Note

### Description

Create subscriptions programmatically using Azure Resource Manager.  Here are the details about the API being released from Azure https://docs.microsoft.com/en-gb/azure/azure-resource-manager/programmatically-create-subscription

Subscriptions

* azurerm_XXXXX

```hcl
```

### References

* #0000

jackofallops:
This PR introduces a new resource to manage Subscriptions using the new SubscriptionAlias API.

**Note:** Creating Subscriptions requires a high level of access to the Azure Billing account in which the Subscription is being created/managed.  The level of access is dependent on the account type with Azure (e.g. Enterprise Account, MCA, or Partner Account). 

**Note:** Using this resource will create a Subscription Alias meta resource. Azure allows multiple Aliases per subscription, however, for Terraform to correctly manage this the Provider limits this to a 1:1 relationship. The Alias Meta resource has been introduced to give a common interface for creating Subscriptions across the various account types.

~**Warning:** Destroying an `azurerm_subscription` resource will deactivate/cancel your subscription, including all resources contained within, through this resource. See [here](https://docs.microsoft.com/en-us/azure/cost-management-billing/manage/cancel-azure-subscription#what-happens-after-subscription-cancellation) for what happens when this occurs.~

Subscriptions containing resources _*not*_ managed by Terraform will prevent that Subscription from being deleted/destroyed and return an error.
```Error: autorest/azure: Service returned an error. Status=400 Code="ResourcesExistingOnSubscription" Message="There are existing resources associated with subscription. Delete them before cancelling the subscription"```

closes #1216

### CODE REVIEW

File: azurerm/internal/services/subscription/subscription_resource.go [hunk 1]
```diff
@@ -0,0 +1,505 @@
...
+		return fmt.Errorf("creating new Subscription (Alias %q): %+v", aliasName, err)
+	}
+
+	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("waiting for creation of Subscription with Alias %q: %+v", id.Name, err)
+	}
+
+	if reactivate {
+		_, err := subscriptionClient.Enable(ctx, subscriptionId)
+		if err != nil {
```
katbyte, line 256: think thes elines can be combined?

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,505 @@
...
+	subscriptionId := resp.Properties.SubscriptionID
+	if subscriptionId == nil || *subscriptionId == "" {
+		return fmt.Errorf("could not read Subscription ID from Alias")
+	}
+	if d.HasChange("subscription_name") {
+		displayName := subscription.Name{
+			SubscriptionName: utils.String(d.Get("subscription_name").(string)),
+		}
+		_, err := subscriptionClient.Rename(ctx, *subscriptionId, displayName)
+		if err != nil {
```
katbyte, line 300: can cobine these lines

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,505 @@
...
+			return fmt.Errorf("could not update Display Name of Subscription %q: %+v", *subscriptionId, err)
+		}
+	}
+
+	if d.HasChange("state") {
+		newState := d.Get("state").(string)
+		switch newState {
+		case "Active":
+			_, err := subscriptionClient.Enable(ctx, *subscriptionId)
+			if err != nil {
```
katbyte, line 310: and these

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,505 @@
...
+				return fmt.Errorf("failed to Enable Subscription %q: %+v", *subscriptionId, err)
+			}
+		case "Cancelled":
+			_, err := subscriptionClient.Cancel(ctx, *subscriptionId)
+			if err != nil {
```
katbyte, line 315: and these

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,505 @@
...
+	resp, err := client.Delete(ctx, id.Name)
+	if err != nil {
+		if !utils.ResponseWasNotFound(resp) {
+			return fmt.Errorf("could not delete Alias %q for Subscription %q (ID: %q): %+v", id.Name, subscriptionName, subscriptionId, err)
+		}
+	}
+
+	// Cancel the Subscription
+	_, err = subscriptionClient.Cancel(ctx, subscriptionId)
+	if err != nil {
```
katbyte, line 437: and these

azurerm/internal/services/subscription/subscription_resource_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	data.ResourceTest(t, r, []resource.TestStep{
+		{
+			Config: r.requiresImport(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r)),
+		},
+	})
+}
+
```
katbyte, line 53: should we have an update test?

katbyte (changes requested):
Thanks @jackofallops - overall looks good and i have soe minor comments i've left inline

azurerm/internal/services/subscription/client/client.go
```diff
@@ -2,17 +2,32 @@ package client
 
 import (
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-11-01/subscriptions"
+	"github.com/Azure/azure-sdk-for-go/services/subscription/mgmt/2020-09-01/subscription"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/common"
 )
 
+// (@jackofallops) - This RP straddles old and new RPs for subscription management as they differ in available
+// functionality. Be mindful of the pluralisation on clients, they have different purposes.
```
tombuildsstuff, line 10: you can likely import alias to make this clearer?

website/docs/r/subscription.html.markdown
```diff
@@ -0,0 +1,93 @@
...
+page_title: "Azure Resource Manager: azurerm_subscription"
+description: |-
+  Manages a Subscription.
+---
+
+# azurerm_subscription
+
+Manages a Subscription.
+
+~> **NOTE:** Destroying a Subscription controlled by this resource will place the Subscription into a cancelled state. It is possible to re-activate a subscription within 90-days of cancellation, after which time the Subscription is irrevocably deleted and the Subscription ID cannot be re-used. For further information see [here](https://docs.microsoft.com/en-us/azure/cost-management-billing/manage/cancel-azure-subscription#what-happens-after-subscription-cancellation)
```
tombuildsstuff, line 13: also worth mentioning it's possible to actually "Delete" a subscription 3 days after it's cancelled, but TF can't do this for users

website/docs/r/subscription.html.markdown
```diff
@@ -0,0 +1,93 @@
...
+
+## Example Usage
+
+```hcl
+resource "azurerm_subscription" "example" {
+  alias              = "examplesub"
```
tombuildsstuff, line 19: since this is a optional field, should we remove this:

```suggestion
```

website/docs/r/subscription.html.markdown
```diff
@@ -0,0 +1,93 @@
...
+## Import
+
+Subscriptions can be imported using the `resource id`, e.g.
+
+```shell
+terraform import azurerm_subscription.example "/providers/Microsoft.Subscription/aliases/subscription1"
+```
+
+!> **NOTE:** When importing a Subscription that was not created programmatically, either by this Terraform resource or using the Alias API, it will have no Alias ID to import via `terraform import`.  
+In this scenario, the `subscription_id` property can be completed and Terraform will assume control of the existing subscription by creating an Alias. e.g.
```
tombuildsstuff, line 84: presumably this wants to be:

```suggestion
In this scenario, the `subscription_id` property can be completed and Terraform will assume control of the existing subscription by updating the Subscription to add an Alias. e.g.
```

website/docs/r/subscription.html.markdown
```diff
@@ -0,0 +1,93 @@
...
+
+* `workload` - (Optional) The workload type of the Subscription.  Possible values are `Production` (default) and `DevTest`. Changing this forces a new Subscription to be created.
+
+## Attributes Reference
+
+In addition to the Arguments listed above - the following Attributes are exported: 
+
+* `id` - The ID of the Subscription.
+
+* `tenant` - The ID of the Tenant to which the subscription belongs.
```
tombuildsstuff, line 64: can we make this `tenant_id`

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,500 @@
...
+
+	id, err := parse.SubscriptionAliasID(d.Id())
+	if err != nil {
+		return err
+	}
+	d.Set("alias", id.Name)
+
+	alias, err := client.Get(ctx, id.Name)
+	if err != nil {
+		return fmt.Errorf("reading Subscription Aliss %q: %+v", id.Name, err)
```
tombuildsstuff, line 340:
```suggestion
		return fmt.Errorf("reading Subscription Alias %q: %+v", id.Name, err)
```

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,500 @@
...
+	alias, err := client.Get(ctx, id.Name)
+	if err != nil || alias.Properties == nil {
+		return fmt.Errorf("could not read Alias %q for Subscription: %+v", id.Name, err)
+	}
+	subscriptionId := ""
+	if subscriptionIdRaw := alias.Properties.SubscriptionID; subscriptionIdRaw != nil {
+		subscriptionId = *subscriptionIdRaw
+	}
+
+	sub, err := subscriptionsClient.Get(ctx, subscriptionId)
```
tombuildsstuff, line 410: similar to #1608 - should we conditionally (feature-flag, defaulted off) check for the presence of existing resources prior to cancelling this?

tombuildsstuff (changes requested):
hey @jackofallops 

On the whole this looks pretty good - I've taken a look through and left some comments inline, but if we can fix those up then this should be otherwise good to go 👍 

Thanks!

website/docs/r/subscription.html.markdown
jackofallops: It's a one-way reference. An Alias is a separate meta-resource that has a subscription ID reference, where the Subscription resource has no notion of Aliases, so I think this change might be misleading?

website/docs/r/subscription.html.markdown
jackofallops: It is in the resource, this is a typo 🙈

azurerm/internal/services/subscription/subscription_resource.go
jackofallops: As discussed offline, despite the docs suggesting that Subscriptions can be cancelled regardless of resources, the API rejects cancellations where resources are present in the Subscription. PR description and docs updated to reflect this.

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+)
+
+func resourceSubscription() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceSubscriptionCreate,
+		Update: resourceSubscriptionUpdate,
+		Read:   resourceSubscriptionRead,
+		Delete: resourceSubscriptionDelete,
+
+		Importer: azSchema.ValidateResourceIDPriorToImport(func(id string) error {
```
tombuildsstuff, line 37: can we make this:

```suggestion
		Importer: azSchema.ValidateResourceIDPriorToImportThen(func(id string) error {
```

and then validate the subscription isn't cancelled during creation

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+
+			"state": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "Active",
+				ValidateFunc: validation.StringInSlice([]string{
+					"Active",
+					"Cancelled",
+				}, false),
+			},
```
tombuildsstuff, line 74: can we remove this?

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+			return fmt.Errorf("enabling Subscription %q: %+v", subscriptionId, err)
+		}
+	}
+
+	alias, err := aliasClient.Get(ctx, id.Name)
+	if err != nil || alias.Properties == nil || alias.Properties.SubscriptionID == nil {
+		return fmt.Errorf("failed reading subscription details for Alias %q: %+v", id.Name, err)
+	}
+
+	if err := waitForSubscriptionStateToSettle(ctx, meta.(*clients.Client), *alias.Properties.SubscriptionID, targetState, d.Timeout(schema.TimeoutCreate)); err != nil {
```
tombuildsstuff, line 264: this timeout will want to be:

```suggestion
	timeout, err := timeout.Until(ctx.Deadline())
	if err != nil { return err }
	if err := waitForSubscriptionStateToSettle(ctx, meta.(*clients.Client), *alias.Properties.SubscriptionID, targetState, timeout); err != nil {
```

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+	return nil
+}
+
+func checkExistingAliases(ctx context.Context, client subscriptionAlias.AliasClient, subscriptionId string) (*string, error) {
+	aliasList, err := client.List(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("could not List existing Subscription Aliases")
+	}
+
+	for _, v := range *aliasList.Value {
```
tombuildsstuff, line 493: there's a panic here:

```suggestion
	if aliasList.Value == nil { return nil, err }
	for _, v := range *aliasList.Value {
```

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+				ForceNew:    true,
+				Optional:    true,
+				Computed:    true,
+			},
+
+			"tenant_id": {
+				Type:        schema.TypeString,
+				Description: "The Tenant ID to which the subscription belongs",
+				Computed:    true,
+			},
```
tombuildsstuff, line 160: we should probably also consider adding tags to these resources too, since they can be set at the subscription level

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+	// remove the alias
+	resp, err := aliasClient.Delete(ctx, id.Name)
+	if err != nil {
+		if !utils.ResponseWasNotFound(resp) {
+			return fmt.Errorf("could not delete Alias %q for Subscription %q (ID: %q): %+v", id.Name, subscriptionName, subscriptionId, err)
+		}
+	}
+
+	// Cancel the Subscription
+	if _, err := subscriptionClient.Cancel(ctx, subscriptionId); err != nil {
```
tombuildsstuff, line 431: we should likely do this conditionally - checking if this is the last alias for the subscription and locking on the subscription id during create and delete (when re-enabling and cancelling)

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/subscription/parse"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/subscription/validate"
+	azSchema "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tf/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+var (
+	billingScopeEnrollmentFmt = "/providers/Microsoft.Billing/billingAccounts/%s/enrollmentAccounts/%s"
+	billingScopeMCAFmt        = "/providers/Microsoft.Billing/billingAccounts/%s/billingProfiles/%s/invoiceSections/%s"
```
tombuildsstuff, line 27: as discussed offline - we should make data sources for these (and make the fields `{thing}_id`), even if initially these are just templating the ID's whilst the SDK is fixed - since this means if other formats are added later we can more easily validate and differentiate these, these presumably should work:

https://docs.microsoft.com/en-us/rest/api/billing/2019-10-01-preview/invoicesections/get
https://docs.microsoft.com/en-us/rest/api/billing/2019-10-01-preview/enrollmentaccounts

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+				}, false),
+			},
+
+			"billing_account": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				Sensitive:    true,
+				Description:  "The name of the billing account under which the Subscription will be created.",
+				ValidateFunc: azValidate.NoEmptyStrings, // TODO - Put some actual protection around this?
```
tombuildsstuff, line 82: we should be able to validate this ID with the data source mentioned above

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+				Optional:    true,
+				ForceNew:    true,
+				Sensitive:   true,
+				Description: "The name of the enrollment account in which to create the subscription. Used for EA accounts.",
+				ConflictsWith: []string{
+					"invoice_section",
+					"billing_profile",
+					"subscription_id",
+				},
+				ValidateFunc: azValidate.NoEmptyStrings, // TODO - Put some actual protection around this?
```
tombuildsstuff, line 96: we should be able to validate this ID with the data source mentioned above

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+				Sensitive:   true,
+				Description: "The name of the Billing Profile under which the Subscription should be created. Used for MCA and Partner Agreements.",
+				ConflictsWith: []string{
+					"enrollment_account",
+					"subscription_id",
+				},
+				RequiredWith: []string{
+					"invoice_section",
+				},
+				ValidateFunc: azValidate.NoEmptyStrings, // TODO - Put some actual protection around this?
```
tombuildsstuff, line 112: we should be able to validate this ID with the data source mentioned above

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+				Description: "The Invoice Section of the Billing Profile which will be used for the subscription. Used for MCA and Partner Agreements.",
+				ConflictsWith: []string{
+					"enrollment_account",
+					"subscription_id",
+				},
+				RequiredWith: []string{
+					"billing_profile",
+				},
+				ValidateFunc: azValidate.NoEmptyStrings, // TODO - Put some actual protection around this?
+			},
```
tombuildsstuff, line 129: we should be able to validate this ID with the data source mentioned above

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+	}
+
+	targetState := d.Get("state").(string)
+	reactivate := false
+	subscriptionId := ""
+
+	// Check if we're adding alias management for an existing subscription
+	if subscriptionIdRaw, ok := d.GetOk("subscription_id"); ok {
+		subscriptionId = subscriptionIdRaw.(string)
+		exists, err := checkExistingAliases(ctx, *aliasClient, subscriptionId)
```
tombuildsstuff, line 211:
```suggestion
		// Terraform assumes a 1:1 mapping between a Subscription and an Alias - first check if there's an existing alias
		exists, err := checkExistingAliases(ctx, *aliasClient, subscriptionId)
```

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+	}
+
+	if err := future.WaitForCompletionRef(ctx, aliasClient.Client); err != nil {
+		return fmt.Errorf("waiting for creation of Subscription with Alias %q: %+v", id.Name, err)
+	}
+
+	if reactivate {
+		if _, err := subscriptionClient.Enable(ctx, subscriptionId); err != nil {
+			return fmt.Errorf("enabling Subscription %q: %+v", subscriptionId, err)
+		}
```
tombuildsstuff, line 256: do we need to wait for this to settle first?

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,499 @@
...
+	"log"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-11-01/subscriptions"
+	subscriptionAlias "github.com/Azure/azure-sdk-for-go/services/subscription/mgmt/2020-09-01/subscription"
+	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	azValidate "github.com/terraform-providers/terraform-provider-azuread/azuread/helpers/validate"
```
tombuildsstuff, line 15: this shouldn't be importing azuread here

website/docs/r/subscription.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+  Manages a Subscription.
+---
+
+# azurerm_subscription
+
+Manages a Subscription.
+
+~> **NOTE:** Destroying a Subscription controlled by this resource will place the Subscription into a cancelled state. It is possible to re-activate a subscription within 90-days of cancellation, after which time the Subscription is irrevocably deleted, and the Subscription ID cannot be re-used. For further information see [here](https://docs.microsoft.com/en-us/azure/cost-management-billing/manage/cancel-azure-subscription#what-happens-after-subscription-cancellation). Users can optionally delete a Subscription once 72 hours have passed, however, this functionality is not suitable for Terraform. A `Deleted` subscription cannot be reactivated.
+
+~> **NOTE:** It is not possible to destroy (cancel) a subscription if it contains resources. If resources are present that are not managed by Terraform then these will need to be removed before the Subscription can be destroyed.
```
tombuildsstuff, line 15: should we also add a note here about assuming a 1:1 between an alias and a subscription

website/docs/r/subscription.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+subcategory: "Base"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_subscription"
+description: |-
+  Manages a Subscription.
+---
+
+# azurerm_subscription
+
+Manages a Subscription.
```
tombuildsstuff, line 11:
```suggestion
Manages an Alias for a Subscription - which adds an Alias to an existing Subscription, allowing it to be managed in Terraform - or create a new Subscription with a new Alias.
```

website/docs/r/subscription.html.markdown
```diff
@@ -0,0 +1,94 @@
...
+In this scenario, the `subscription_id` property can be completed and Terraform will assume control of the existing subscription by creating an Alias. e.g.
+
+```hcl
+// importing existing Manually Created Subscription with no Alias
+resource "azurerm_subscription" "example" {
+  alias             = "examplesub"
+  subscription_name = "My Example Subscription"
+  subscription_id   = "12345678-12234-5678-9012-123456789012"
+}
+```
```
tombuildsstuff, line 94: maybe we should also move this into the example usage above - it's a single (small) resource so having multiple example usages should be fine?

tombuildsstuff (changes requested):
Taking a look through on the whole this looks pretty good - but we should update this to use the ID's of these items rather than the Name's, since there's different scopes/options for Billings Accounts etc, which'd help us differentiate these in the future

azurerm/internal/services/subscription/subscription_resource.go
jackofallops: I can't find API / SDK support for managing tags, only reading back. I'll add as `computed` for now.

azurerm/internal/services/subscription/subscription_resource.go
jackofallops: yes, and this has also moved as a result of removing the `state` property noted above.

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,472 @@
...
+	ctx, cancel := timeouts.ForUpdate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	id, err := parse.SubscriptionAliasID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	locks.ByName(id.Name, SubscriptionResourceName)
+	defer locks.UnlockByName(id.Name, SubscriptionResourceName)
```
tombuildsstuff, line 254: do we want to lock on the Subscription ID here instead?

azurerm/internal/services/subscription/subscription_resource.go
```diff
@@ -0,0 +1,472 @@
...
+	ctx, cancel := timeouts.ForDelete(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	id, err := parse.SubscriptionAliasID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	locks.ByName(id.Name, SubscriptionResourceName)
+	defer locks.ByName(id.Name, SubscriptionResourceName)
```
tombuildsstuff, line 354: (same here) do we want to be locking on the Subscription ID?

tombuildsstuff (approved):
hey @jackofallops 

Thanks for pushing those changes - I've left some more comments inline but this otherwise LGTM 👍

azurerm/internal/services/subscription/subscription_resource.go
jackofallops: There are effectively 2 resources at play here, so I've (defensively) opted to try and guard against change to either, since we need to maintain the 1:1 relation.

azurerm/internal/services/subscription/subscription_resource.go
jackofallops: As above.

PR #10729, opened 2021-02-25
not set `source` property when update Spring cloud deployment

njuCZ:
split createUpdate func.

fix bug: 
terraform does not support upload jar, users would do this through other tools, such as azure.cli.
when create in terraform, we would set the `source` to a default value, and for update fucntion, we should not change the `source` property

image


tombuildsstuff (approved):
LGTM 👍

tombuildsstuff:
Tests pass:

PR #10746, opened 2021-02-26
Fix sql auditing to Azure Monitor

yupwei68:


katbyte (approved):
Thanks @yupwei68 - LGMT 👍

PR #10760, opened 2021-03-01
Fix acc test case failure of `TestAccLinuxVirtualMachine_imageFromImage`

ArcturusZhang:
and `TestAccLinuxVirtualMachine_imageFromSharedImageGallery`

We are invoking an SSH login in `generalizeVirtualMachine` function with a certain pattern of admin username and password, but those were assigned differently in the configuration of these two test cases


tombuildsstuff (approved):
LGTM - thanks for this @ArcturusZhang

PR #10764, opened 2021-03-01
Documentation: Change default value for policy_settings.enabled argument

madsaune:
Hi,

This PR changes the documentation for `azurerm_web_application_firewall_policy`. More specific the `enabled` argument on the `policy_settings` block from the string version "Enabled" to a bool type "true".

According to the source code the correct type is bool and not string as implied by the documentation.

```hcl
# ./azurerm/internal/services/network/web_application_firewall_policy_resource.go

# ...

"policy_settings": {
  Type:     schema.TypeList,
  Optional: true,
  MaxItems: 1,
  Elem: &schema.Resource{
      "enabled": {
        Type:     schema.TypeBool,
        Optional: true,
        Default:  true,
      },

      # ...
```

### DIFF
--- a/website/docs/r/web_application_firewall_policy.html.markdown
+++ b/website/docs/r/web_application_firewall_policy.html.markdown
@@ -164,7 +164,7 @@ The `match_variables` block supports the following:
 
 The `policy_settings` block supports the following:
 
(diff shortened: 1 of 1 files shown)


tombuildsstuff (approved):
LGTM - thanks for this @madsaune

PR #10768, opened 2021-03-01
Fix file_upload_limit_in_mb attribute name in docs

sirlatrom:
It was previously listed as `file_upload_limit_mb` whereas the correct name was used in the example.


katbyte (approved):
LGTM 👍

PR #10770, opened 2021-03-01
`sentinel_alert_rule_xxx` acctests add dependency on log analytics solution

magodo:
Since service team starts enforcing the official requirement that a Sentinel customer (workspace) must have “SecurityInsights” solution installed. This PR added this dependency in the acctests, which should fix the test failures of sentinel alert rules.

Note that there is a change in `log_analytics_solution_resource.go`, this is to leverage the id parser so as to avoid the plan diff later observed in the sentinel alert rule resources (the case of `resourceG/group`).


katbyte (approved):
Thanks @magodo - LGTM 👍

PR #10779, opened 2021-03-01
network/private-endpoint: allow alias as well

mbfrahry:
Supersedes #9794


dlamotte:
@mbfrahry replying to https://github.com/terraform-providers/terraform-provider-azurerm/pull/9794#issuecomment-788314267

That error is foreign to me.  FWIW, I validated that my previous code was functioning (I used it end to end and saw it work).  I don't think the Azure API has probably changed to require something like this.  Seems like something else is going on.  These Service Alias' are meant to be used like this I believe.  Otherwise, I'm not sure how these Service Alias would work.

mbfrahry:
@dlamotte can you write a test to replicate how you got it working end to end and add it to this PR?

katbyte (approved):
LGTM 👍

dlamotte:
@mbfrahry I can try.  It will probably be some time before I can though.  Still circling around other priorities atm.

kenzwang:
@mbfrahry hi, can I know the current status of this PR? I only see one linting error in the CI checks. Is there a reason this is not merged yet? Are you still waiting on something?

katbyte:
@kenzwang - if you can see from the other PR @mbfrahry couldn't get it to work, so while approved in its current state it doesn't appear the tests work so some additional changes will probably be required.

@dlamotte - any eta on when you can help get this working or shall i close this until you have time?

kenzwang:
@katbyte I can continue @dlamotte 's work in this PR.

@mbfrahry could you give more info of what couldn't work and/or where I can start?

kenzwang:
@mbfrahry and @katbyte 

I was able to use the [REST API portal](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/privateendpoints/createorupdate) to reproduce the issue mentioned earlier here (https://github.com/terraform-providers/terraform-provider-azurerm/pull/9794#issuecomment-788203567).

The issue I found is, the request will fail when using `privateLinkServiceConnections`, but it will succeed when using `manualPrivateLinkServiceConnections`. 

Here's an example of a successful request of mine:
Screen Shot 2021-04-01 at 2 24 44 PM

So when we set `is_manual_connection` to `true` this should work.

Could you guys verify it works this way?

mbfrahry:
That worked! Thanks @kenzwang!

PR #8887, opened 2020-10-14
WIP Support for Active Directory Domain Services

manicminer:
- New resource: `azurerm_active_directory_domain_service`
- New data source: `azurerm_active_directory_domain_service`
- Remove azuread dependency in cosmos package

PR #10782, opened 2021-03-02
Active Directory Domain Services

manicminer:
- New data source: `azurerm_active_directory_domain_service`
- New resource: `azurerm_active_directory_domain_service`
- New resource: `azurerm_active_directory_domain_service_replica_set`
- New resource: `azurerm_virtual_network_dns_servers`

## Background

AADDS is a necessarily tightly orchestrated service, with some fragility. To begin, each tenant (or subscription) can only have a single AADDS resource. Scaling is provided for by "replica sets", which are pairs of managed domain controllers and are part of the AADDS resource. When creating an AADDS resource, it must have one replica set in the same location as the parent resource. Post-creation, additional replica sets can be added or removed as needed, however each additional replica set has specific constraints, namely:

1. Each one requires a vnet and subnet allocated with sufficient address space, which must be peered with the vnet containing the initial replica set
2. The DNS servers for both vnets must be set to use the IP addresses of the domain controllers belonging to the initial replica set
3. Network security group with rules sufficient to enable intra-DC traffic
4. Following each write operation to a replica set, no further writes are accepted until all replica sets have fully provisioned, _and_ both domain controllers in each replica set are available, i.e. they have published their internal IPv4 addresses

Failure to meet these requirements causes a replica set to enter a Failed state.

## Implementation

Based on these constraints, this proposed implementation contains two related resources, `azurerm_active_directory_domain_service` and `azurerm_active_directory_domain_service_replica_set`, which can be used together to enable orchestration of multiple replica sets. To enable the vnet DNS server component, also proposed is an `azurerm_virtual_network_dns_servers` resource, which can join the resource graph in between the domain service resource and it's related replica set resources. The `dns_servers` property of the `azurerm_virtual_network` resource also has the Computed field set to help reduce config boilerplate via ignore_changes.

Example configuration is given in the acceptance test, which, due to the aforementioned constraints cannot be parallelised and therefore runs as a single monolithic test (including the data source).

Other notes:
- AADDS API doesn't support PATCH, only PUT
- When orchestrating replica sets, the ID of each replica set is only returned once it has been provisioned
- When first adding a replica set, and we don't yet know its ID, we identify it in the response by its subnet ID, assuming that replica sets cannot coexist in a subnet
- The ID of the "initial" replica set forms part of the resource ID for the `azurerm_active_directory_domain_service` resource in order to facilitate importing
- Median provisioning times are approximately 1-2 hours for each write operation that involves a replica set, sometimes longer

Checklist:
- [x] Create a working domain service resource
- [x] Add replica sets to the domain service
- [x] Remove replica sets from the domain service
- [x] Delete the domain service
- [x] Configure [secure LDAP](https://docs.microsoft.com/en-us/azure/active-directory-domain-services/tutorial-configure-ldaps) (LDAP over TLS)
- [x] Deploy in [User Forest mode](https://docs.microsoft.com/en-us/azure/active-directory-domain-services/concepts-resource-forest) (passwords are synced from AAD)
- [ ] ~Deploy in [Resource Forest mode](https://docs.microsoft.com/en-us/azure/active-directory-domain-services/concepts-forest-trust) (no sync, accounts are sourced from on-premise AD) with trust relationships~  _to follow in another PR_

## Initial test result

Screenshot 2021-06-16 at 23 41 31

## Future Considerations

- Maybe try to recover Failed replica sets by deleting them and starting again

Replaces: #8887
Closes: #8970
Related: https://github.com/Azure/azure-rest-api-specs/issues/11200, https://github.com/Azure/azure-sdk-for-go/issues/14747


azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,863 @@
...
+							Required:  true,
+							Sensitive: true,
+						},
+					},
+				},
+			},
+
+			"location": azure.SchemaLocation(),
+
+			"name": {
```
katbyte, line 117: could we put required attributes at the top?

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,863 @@
...
+						"notify_global_admins": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
+					},
+				},
+			},
+
+			"replica_set": {
```
katbyte, line 155: what does set refer to here? and should this be
```suggestion
			"replication_set": {
```

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,863 @@
...
+						"location": {
+							Type:             schema.TypeString,
+							Required:         true,
+							ForceNew:         true, // TODO: figure out if this is needed
+							ValidateFunc:     location.EnhancedValidate,
+							StateFunc:        location.StateFunc,
+							DiffSuppressFunc: location.DiffSuppressFunc,
+						},
+
+						"replica_set_id": {
```
katbyte, line 185: i think id works as the block as the same prefix?
```suggestion
						"id": {
```

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,863 @@
...
+						},
+
+						"subnet_id": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ForceNew:     true, // TODO: figure out if this is needed
+							ValidateFunc: azure.ValidateResourceID,
+						},
+
+						"vnet_site_id": {
```
katbyte, line 202: should this be
```suggestion
						"site_vnet_id": {
```

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,863 @@
...
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+
+			"resource_forest": {
```
katbyte, line 210: is this different from a regular ad forest?
```suggestion
			"forest": {
```

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,863 @@
...
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"resource_forest": {
```
katbyte, line 216: and this is the same name as the block?

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,863 @@
...
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validation.StringIsNotEmpty,
+						},
+
+						"forest_trust": {
```
katbyte, line 222: can drop the block name?
```suggestion
						"trust": {
```

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,863 @@
...
+										Elem: &schema.Schema{
+											Type:         schema.TypeString,
+											ValidateFunc: validation.StringIsNotEmpty,
+										},
+									},
+
+									"trust_direction": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringIsNotEmpty,
```
katbyte, line 247: feel like this is one of two values so we should beable to validate it?

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,863 @@
...
+			},
+
+			"domain_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringIsNotEmpty, // TODO: proper validation, first prefix must be 15 chars or less
+			},
+
+			"filtered_sync_enabled": {
```
katbyte, line 70: this sounds weird to me.. maybe?
```suggestion
			"filter_sync_results": {
```

website/docs/r/active_directory_domain_service.html.markdown
```diff
@@ -0,0 +1,292 @@
...
+Manages an Active Directory Domain Service.
+
+~> Implementation Note: Before using this resource, there must exist in your tenant a service principal for the Domain Services published application. This service principal cannot be easily managed by Terraform and it's recommended to create this manually, as it does not exist by default. See [official documentation](https://docs.microsoft.com/en-us/azure/active-directory-domain-services/powershell-create-instance#create-required-azure-ad-resources) for details.
+
+## Example Usage
+
+```hcl
+resource "azurerm_resource_group" "deploy" {
+  name     = "deploy-rg"
+  location = "westeurope"
```
katbyte, line 20:
```suggestion
  name     = "example-resources"
  location = "West Europe"
```

website/docs/r/active_directory_domain_service.html.markdown
```diff
@@ -0,0 +1,292 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `domain_configuration_type` - (Optional) The forest type to use when creating the managed domain. Set to `ResourceTrusting` to create a _Resource Forest_, omit this property to create a _User Forest_. Changing this forces a new resource to be created.
+  
+* `domain_name` - (Required) The Active Directory domain to use. See [official documentation](https://docs.microsoft.com/en-us/azure/active-directory-domain-services/tutorial-create-instance#create-a-managed-domain) for constraints and recommendations.
+
+* `filtered_sync_enabled` - (Optional) Whether to enable filtered sync for users and groups in Azure Active Directory. Defaults to `false`.
```
katbyte, line 163: might be nice to expand on what filtered sync does

katbyte (changes requested):
Thanks @manicminer - given this a quick look. Skipped over the datasource and focused on the resource for now

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
manicminer: I'm not actually sure since this attribute has yet to be populated by the API, I'm half expecting this to be a GUID but if it doesn't make an appearance I'll drop it.

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
manicminer: It is - the terminology is specific to AADDS. It comes in two flavors, a "user forest" and a "resource forest".

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
manicminer: Agreed, this is some really bad field naming in the API and will rename this once I figure out what it does :D

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
manicminer: Agreed, I haven't gotten to successfully using this setting yet but will add validation once I figure out the (undocumented) values

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
manicminer: It's the product term used here, a 'replica set' is a pair of domain controllers in a given region.

Screenshot 2021-03-04 at 02 40 10

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,863 @@
...
+			Update: schema.DefaultTimeout(120 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{ // TODO: add computed attributes: deployment_id, sync_owner
+			"deployment_id": {
```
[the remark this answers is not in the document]
manicminer: Sure thing, I had them in alphabetical order indiscriminately whilst I figured out which ones were read-only.

tombuildsstuff (changes requested):
I only got half way through this and haven't had time to circle back yet, but here's comments from a review this morning

azurerm/internal/services/domainservices/active_directory_domain_service_data_source.go
```diff
@@ -0,0 +1,305 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/azure"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tags"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func dataSourceArmActiveDirectoryDomainService() *schema.Resource {
```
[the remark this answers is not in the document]
manicminer: Some that I can think of in the context of configuring other services (e.g. VMSS/VM resources) are maybe

- ldaps.0.external_access_ip_address
- replica_sets.x.external_access_ip_address (along with location for selecting one)
- replica_sets.x.domain_controller_ip_addresses (with subnet_id)

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,875 @@
...
+		}
+		if input.Ldaps == aad.LdapsEnabled {
+			result["enabled"] = true
+		}
+		if input.PfxCertificate != nil {
+			result["pfx_certificate"] = *input.PfxCertificate
+		}
+		if input.PfxCertificatePassword != nil {
+			result["pfx_certificate_password"] = *input.PfxCertificatePassword
+		}
```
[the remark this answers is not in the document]
manicminer: These are set around line 700-704?

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,875 @@
...
+	}
+	if input.AdditionalRecipients != nil {
+		result["additional_recipients"] = *input.AdditionalRecipients
+	}
+	if input.NotifyDcAdmins == aad.NotifyDcAdminsEnabled {
+		result["notify_dc_admins"] = true
+	}
+	if input.NotifyGlobalAdmins == aad.NotifyGlobalAdminsEnabled {
+		result["notify_global_admins"] = true
+	}
```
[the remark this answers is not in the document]
manicminer: These are also set on lines 730-734

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,875 @@
...
+			}
+			if rf.TrustDirection != nil {
+				ft["trust_direction"] = *rf.TrustDirection
+			}
+			if rf.TrustPassword != nil {
+				ft["trust_password"] = rf.TrustPassword
+			}
+			if rf.TrustedDomainFqdn != nil {
+				ft["trusted_domain_fqdn"] = *rf.TrustedDomainFqdn
+			}
```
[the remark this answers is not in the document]
manicminer: As above, on lines 798-804 for the inner block and lines 831-834 for the outer block

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
```diff
@@ -0,0 +1,875 @@
...
+					},
+				},
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
+
+			"security": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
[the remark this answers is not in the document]
manicminer: The `ldaps` (to be renamed `secure_ldap`) block, and it's attribute `enabled` is computed to keep compatibility with the data source - we could drop that if we don't include the DS (or write a flatten function just for the DS).

`notifications` and `security` probably don't need to be, will fix those 👍

azurerm/internal/acceptance/testclient/client.go
```diff
@@ -17,7 +17,7 @@ var (
 	// @tombuildsstuff: this is left in as a compatibility layer for the moment
 	// in the near future we'll remove this, and remove the "azuread" provider below
 	// but for the moment there's no need to remove the vendor here imminantly afaict
-	EnableBinaryTesting = true
+	EnableBinaryTesting = false
```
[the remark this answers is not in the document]
manicminer: Binary testing wasn't happy with the azuread usage - this is fixed in master, will rebase

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
manicminer: Double checked and it hasn't once been returned, so dropping it.

manicminer:
Newer API versions are available, awaiting SDK support

https://github.com/Azure/azure-sdk-for-go/issues/14747

azurerm/helpers/validate/strings.go
```diff
@@ -1,10 +1,29 @@
 package validate
 
 import (
+	"encoding/base64"
 	"fmt"
 	"strings"
 )
 
+// Base64EncodedString validates that the string is base64 encoded
+func Base64EncodedString(i interface{}, k string) ([]string, []error) {
```
katbyte, line 10: looks like we already have this function:
image

lets maybe unify the two?

also could we place this in internal/tf/validate ?

azurerm/internal/services/domainservices/active_directory_domain_service_data_source.go
```diff
@@ -0,0 +1,338 @@
...
+					Schema: dataSourceActiveDirectoryDomainServiceReplicaSetSchema(),
+				},
+			},
+
+			"resource_forest": {
+				Type:     pluginsdk.TypeList,
+				Computed: true,
+				Elem: &pluginsdk.Resource{
+					Schema: map[string]*pluginsdk.Schema{
+						"resource_forest": {
```
katbyte, line 94: this is the same name as the block name?

azurerm/internal/services/domainservices/active_directory_domain_service_data_source.go
```diff
@@ -0,0 +1,338 @@
...
+
+			"replica_sets": {
+				Type:     pluginsdk.TypeList,
+				Computed: true,
+				Elem: &pluginsdk.Resource{
+					Schema: dataSourceActiveDirectoryDomainServiceReplicaSetSchema(),
+				},
+			},
+
+			"resource_forest": {
```
katbyte, line 89: would this be better as
```suggestion
			"forest": {
```

azurerm/internal/services/domainservices/active_directory_domain_service_data_source.go
```diff
@@ -0,0 +1,338 @@
...
+				Type:     pluginsdk.TypeList,
+				Computed: true,
+				Elem: &pluginsdk.Resource{
+					Schema: map[string]*pluginsdk.Schema{
+						"resource_forest": {
+							Type:     pluginsdk.TypeString,
+							Computed: true,
+						},
+
+						"forest_trust": {
```
katbyte, line 99: forest is in the block name?
```suggestion
						"trust": {
```

azurerm/internal/services/domainservices/active_directory_domain_service_data_source.go
```diff
@@ -0,0 +1,338 @@
...
+
+									"remote_dns_ips": {
+										Type:     pluginsdk.TypeList,
+										Computed: true,
+										Elem: &pluginsdk.Schema{
+											Type: pluginsdk.TypeString,
+										},
+									},
+
+									"trust_direction": {
```
katbyte, line 117:
```suggestion
									"direction": {
```

azurerm/internal/services/domainservices/active_directory_domain_service_data_source.go
```diff
@@ -0,0 +1,338 @@
...
+										Type:     pluginsdk.TypeString,
+										Computed: true,
+									},
+
+									"trust_password": {
```
katbyte, line 122:
```suggestion
									"password": {
```

azurerm/internal/services/domainservices/active_directory_domain_service_test.go
```diff
@@ -0,0 +1,531 @@
...
+type ActiveDirectoryDomainServiceResource struct {
+	adminPassword string
+}
+
+type ActiveDirectoryDomainServiceReplicaSetResource struct {
+}
+
+// Running all tests sequentially here in the same acctest (including the data source) since you can create only one
+// AADDS resource per tenant, or per location, or per subscription, making parallel testing infeasible.
+func TestAccActiveDirectoryDomainService(t *testing.T) {
```
katbyte, line 48: could we break this up into test functions and then sequentially run them as we do elsewhere in the provider so its a bit mroe clear what test is what?

katbyte (changes requested):

azurerm/helpers/validate/strings.go
manicminer: 💯  have updated the other resource to use the common function

azurerm/internal/services/domainservices/active_directory_domain_service_data_source.go
manicminer: Good call, have renamed those child attributes

azurerm/internal/services/domainservices/active_directory_domain_service_data_source.go
manicminer: I'm unsure about this one because the API response is all I have to go on - the docs do not elaborate. That said, the mode of operation that populates this block is specifically called "Resource Forest mode"? As for the child attribute of the same name, I'm really not sure. Perhaps a better route would be to remove these for now and not support them until we know more?

azurerm/internal/services/domainservices/active_directory_domain_service_test.go
manicminer: This could definitely be explained better; I've added a more thorough comment rationalising the test, and a TODO that we should revisit this in future when the service is better optimised

azurerm/internal/services/domainservices/active_directory_domain_service_data_source.go
manicminer: Per my other comment, maybe on balance we should not support this block until we have a better understanding of what will be populated here?

manicminer: Now removed until we know more about it and can test

azurerm/internal/services/domainservices/active_directory_domain_service_data_source.go
manicminer: Now removed as per above

azurerm/internal/services/domainservices/active_directory_domain_service_resource.go
katbyte: any comments on this?

katbyte (approved):
aside from one minor comment this LGTM now 🌈🍰

PR #10784, opened 2021-03-02
testing: removing the old test framework

tombuildsstuff:
This PR removes the feature-toggle between the new/old test framework - and fixes an issue where the AAD Provider wasn't available at test time


manicminer (approved):
LGTM!

PR #10795, opened 2021-03-03
Fixes bug #10773

### LINKED ISSUE #10773: azurerm_netapp_volume broken in 2.49
### Community Note

### Affected Resource(s)

* `azurerm_netapp_volume`

```hcl
resource "azurerm_netapp_volume" "netapp-volume-1" {
  lifecycle {
    prevent_destroy = false
  }

  name                  = var.volume_name_1
  account_name          = azurerm_netapp_account.netapp.name
  location              = "westus"
  resource_group_name   = var.resource_group_name
  pool_name             = azurerm_netapp_pool.netapp.name
  volume_path           = var.volume_path_1
  subnet_id             = var.subnet_id
  service_level         = var.service_level
  storage_quota_in_gb   = var.storage_quota_in_gb_1
}
```

### Debug Output

https://gist.github.com/nshores/5ecd47ddc351fff1030256e557596850

### Expected Behaviour
azurerm_netapp_volume should refresh state of current volumes.

### Actual Behaviour

Crash when refreshing state of volume. 

### Steps to Reproduce

1. `terraform plan`

### Important Factoids

Rolling back to 2.46.1 fixes the issue.

### References

#10610  appears to be the breaking change. @paulomarquesc

paulomarquesc:
* Fixing null check for issue related to executing plan in state generated by previous azurerm version (e.g. 2.48.0) where data_protection_replication was not part of the schema yet.
* Improvement on null and endpoint checks on flattenNetAppVolumeDataProtectionReplication function.

Note: fixes #10773

### DIFF
--- a/azurerm/internal/services/netapp/netapp_volume_data_source.go
+++ b/azurerm/internal/services/netapp/netapp_volume_data_source.go
@@ -155,7 +155,7 @@ func dataSourceNetAppVolumeRead(d *schema.ResourceData, meta interface{}) error 
 			return fmt.Errorf("setting `mount_ip_addresses`: %+v", err)
 		}
 
-		if props.DataProtection.Replication != nil {
+		if props.DataProtection != nil && props.DataProtection.Replication != nil {
 			if err := d.Set("data_protection_replication", flattenNetAppVolumeDataProtectionReplication(props.DataProtection)); err != nil {
 				return fmt.Errorf("setting `data_protection_replication`: %+v", err)
 			}
--- a/azurerm/internal/services/netapp/netapp_volume_data_source.go
+++ b/azurerm/internal/services/netapp/netapp_volume_data_source.go
@@ -154,11 +154,8 @@ func dataSourceNetAppVolumeRead(d *schema.ResourceData, meta interface{}) error 
 		if err := d.Set("mount_ip_addresses", flattenNetAppVolumeMountIPAddresses(props.MountTargets)); err != nil {
 			return fmt.Errorf("setting `mount_ip_addresses`: %+v", err)
 		}
-
-		if props.DataProtection != nil && props.DataProtection.Replication != nil {
-			if err := d.Set("data_protection_replication", flattenNetAppVolumeDataProtectionReplication(props.DataProtection)); err != nil {
-				return fmt.Errorf("setting `data_protection_replication`: %+v", err)
-			}
+		if err := d.Set("data_protection_replication", flattenNetAppVolumeDataProtectionReplication(props.DataProtection)); err != nil {
+			return fmt.Errorf("setting `data_protection_replication`: %+v", err)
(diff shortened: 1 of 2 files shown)


katbyte (approved):
LGTM 👍

azurerm/internal/services/netapp/netapp_volume_resource.go
```diff
@@ -390,7 +390,7 @@ func resourceNetAppVolumeRead(d *schema.ResourceData, meta interface{}) error {
 		if err := d.Set("mount_ip_addresses", flattenNetAppVolumeMountIPAddresses(props.MountTargets)); err != nil {
 			return fmt.Errorf("setting `mount_ip_addresses`: %+v", err)
 		}
-		if props.DataProtection.Replication != nil {
+		if props.DataProtection != nil && props.DataProtection.Replication != nil {
 			if err := d.Set("data_protection_replication", flattenNetAppVolumeDataProtectionReplication(props.DataProtection)); err != nil {
```
tombuildsstuff, line 394: from Terraform's side we should be setting this field in all cases, since the flatten function returns an empty slice when the `if` conditions above aren't met - so could we remove the if statement:

```suggestion
		if err := d.Set("data_protection_replication", flattenNetAppVolumeDataProtectionReplication(props.DataProtection)); err != nil {
```

azurerm/internal/services/netapp/netapp_volume_data_source.go
```diff
@@ -155,7 +155,7 @@ func dataSourceNetAppVolumeRead(d *schema.ResourceData, meta interface{}) error
 			return fmt.Errorf("setting `mount_ip_addresses`: %+v", err)
 		}
 
-		if props.DataProtection.Replication != nil {
+		if props.DataProtection != nil && props.DataProtection.Replication != nil {
```
tombuildsstuff, line 158: (same here) could we remove this `if` statement?

tombuildsstuff (changes requested):
hey @paulomarquesc 

Thanks for this PR.

Taking a look through on the whole this looks pretty good - however we'll want to remove the `if` statements around the Set calls to ensure we track when these fields are removed/emptied out - but if we can fix those up this should otherwise be good to go 👍 

Thanks!

paulomarquesc:
Hi @tombuildsstuff, the problem is that our api does not even return the property if nothing is set there, I can set  data_protection_replication as empty no problem but if customer didn't set anything related to data protection the API does not return the field within VolumeProperties, that's why I had to include the first test to `props.DataProtection != nil` otherwise it will fail that's why this bug was raised.

paulomarquesc:
@tombuildsstuff, changes are now applied, they make sense, thanks for spotting it.

tombuildsstuff (approved):
LGTM 👍

PR #10796, opened 2021-03-03
documentation: align all examples to a single location

katbyte:


website/docs/r/virtual_network_gateway_connection.html.markdown
```diff
@@ -92,7 +92,7 @@ in different locations/regions.
 ```hcl
 resource "azurerm_resource_group" "us" {
   name     = "us"
-  location = "East US"
+  location = "West Europe"
```
favoretti, line 95: These kind of examples will be tricky to understand...

favoretti: Namely, this demonstrates peering/gateways between regions and now they are both referring to the same region.

tombuildsstuff (approved):
LGTM 👍

website/docs/r/virtual_network_gateway_connection.html.markdown
tombuildsstuff: looks like this has been resolved

PR #10799, opened 2021-03-03
new resource "azurerm_spring_cloud_config_server"

njuCZ:
split config server from `azurerm_spring_cloud_service`.

changes in this PR:
- make `config_server_git_setting` deprecated in favor of this new resource.
- modify flatten function in `azurerm_spring_cloud_service`  so that `azurerm_spring_cloud_config_server` could reuse.

image

### DIFF
--- a/azurerm/internal/services/springcloud/spring_cloud_config_server_resource_test.go
+++ b/azurerm/internal/services/springcloud/spring_cloud_config_server_resource_test.go
@@ -1,0 +1,202 @@
+package springcloud_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/springcloud/parse"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance/check"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+type SpringCloudConfigServerResource struct {
+}
+
+func TestAccSpringCloudConfigServer_basic(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_spring_cloud_config_server", "test")
+	r := SpringCloudConfigServerResource{}
+
+	data.ResourceTest(t, r, []resource.TestStep{
+		{
+			Config: r.basic(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+			),
+		},
+		data.ImportStep(),
+	})
+}
+
+func TestAccSpringCloudConfigServer_complete(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_spring_cloud_config_server", "test")
+	r := SpringCloudConfigServerResource{}
+
+	data.ResourceTest(t, r, []resource.TestStep{
+		{
+			Config: r.complete(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+			),
+		},
+		data.ImportStep(
+			// those field returned by api are "*"
+			// import state verify ignore those fields
+			"ssh_auth.0.private_key",
+			"ssh_auth.0.host_key",
+			"ssh_auth.0.host_key_algorithm",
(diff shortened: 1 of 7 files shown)


tombuildsstuff:
@njuCZ changing this to `schema.SchemaConfigModeAttr` is a breaking change unfortunately

njuCZ:
@tombuildsstuff May I ask what's the correct process to split a field to a standalone resource?
to change it to `schema.SchemaConfigModeAttr` serves the purpose that user could correctly set it to empty using `azurerm_spring_cloud_service`, so exsiting users would not be blocked

njuCZ:
@tombuildsstuff have updated and remove the schemaModeAttr, could you have a look again?

njuCZ:
fix the conflicts

azurerm/internal/services/springcloud/spring_cloud_config_server_resource_test.go
```diff
@@ -0,0 +1,207 @@
...
+resource "azurerm_spring_cloud_service" "test" {
+  name                = "acctest-sc-%d"
+  location            = azurerm_resource_group.test.location
+  resource_group_name = azurerm_resource_group.test.name
+
+  lifecycle {
+    ignore_changes = [
+      config_server_git_setting,
+    ]
+  }
```
jackofallops, line 204: This suggests that we still have a breaking change here.  Using a new meta resource shouldn't break the parent, which it appears this does in all cases. I'd recommend taking a look at this config block (`config_server_git_setting` that is) again on the service resource to see if we can handle the deprecation, and the "external" change without this.

njuCZ: @jackofallops according to former experience, when split a field into a standalone resource, we would make that field Optional + Computed.  Then encountered a bug that it could not be set to empty, we would set the configMode to `schema.SchemaConfigModeAttr`. There are some examples like this in the codebase. But Tom suggest this will also have a break change when using `dynamic` keyword, and says just left it uncomputed and use ignore_changes. please refer to this link: https://hashicorp.slack.com/archives/G012LHU9XDJ/p1614842151096000

from the api side, I am sorry to say I could not find a way to avoid it

tombuildsstuff: @njuCZ ultimately this is a breaking change either way - what use-case is this PR looking to solve exactly?

yucwan: I'm the PM from service team and the use-case is as below in brief:

1. Our enterprise customers use Terraform in both their infra team and apps team.
2. Typically, Infra team mange all the service level configurations like vnet using azurerm_spring_cloud_service, and apps team manage app level configurations using azurerm_spring_cloud_app.
3. Config server is on service level,  but it contains remote settings for all apps so in many cases apps team need to manage configurations here.
4. As infra team and apps team have separate TF repo and pipelines, it's extremely difficult for apps team to manage config server with current service + app resources. A separate resource for config server will perfectly solve this.

Our top customers requested for this change and they are waiting for the fix. Thanks a lot if you could help make that happen.

jackofallops: Hi @yucwan - As I understand it, the base config server configuration is common for all apps deployed on the service and _should_ be configured with it. If any infra property is updated, as the whole thing is ForceNew, it will certainly break any deployed application(s) without a co-ordinated effort to apply both team's Terraform runs in the correct order. In the use case you describe, it sounds like the App team should be deploying the `azurerm_spring_cloud_service` using the infrastructure values provided by the infra team, or from data sources, e.g. `location`, `network.*` ? WDYT?

yucwan: @jackofallops "config server configuration is common for all apps" True and that's why we put config server on the service level in our resource model. However the solution does not work for our customers. Infra team take cares of fundamentals like number of service instance, pricing tier and network connectivity/security, thus infra team must control/monitor the `azurerm_spring_cloud_service`. Asking customers changing their workflow for this does not sounds reasonable.

tombuildsstuff: @yucwan

Whilst other tooling such as PowerShell handles these delta/patch deployments by updating the field, they don't track changes to the values of these fields - Terraform by contrast tracks the values of these fields in order to determine when there's a diff between the value in Azure and the local value.

Since we already expose the ability to configure the settings at the top level (the service) unfortunately there's no way for us to split these out without making the use-case for the 80% of users harder - or without it being a breaking change.

Based on the information above, it sounds like the teams themselves should be deploying the Service and the Application given the way the service is designed, using Data Sources to look up the relevant information needed to provision the Service if necessary.

Given it doesn't make sense to split this resource out in Terraform (since it'd make it harder for the majority of use-cases here) - whilst I'd like to thank @njucz for the contribution I'm going to close this PR for the moment.

Thanks!

PR #10801, opened 2021-03-03
`azurerm_frontdoor`: Expose feature flag to override `backend_pool` MaxItem quota

### LINKED ISSUE #10800: `azurerm_frontdoor`: Expose a way to bypass `backend_pool` max `backend_pool` count.
Some customers have gotten exceptions from Microsoft Support to increase the number of supported `backend_pool`'s allowed for their Front Door resource instance(s). However, the Terraform providers implementation of the resource hardcodes this limit to the default limit per subscription of 50 `backend_pool`'s. This issue has been escalated and a fix has been requested.

WodansSon:
(fixes #10800)

### DIFF
--- a/azurerm/internal/features/defaults.go
+++ b/azurerm/internal/features/defaults.go
@@ -3,6 +3,9 @@ package features
 func Default() UserFeatures {
 	return UserFeatures{
 		// NOTE: ensure all nested objects are fully populated
+		Frontdoor: FrontdoorFeatures{
+			IgnoreBackendPoolLimit: false,
+		},
 		KeyVault: KeyVaultFeatures{
 			PurgeSoftDeleteOnDestroy:    true,
 			RecoverSoftDeletedKeyVaults: true,
--- a/azurerm/internal/features/user_flags.go
+++ b/azurerm/internal/features/user_flags.go
@@ -1,11 +1,16 @@
 package features
 
 type UserFeatures struct {
+	Frontdoor              FrontdoorFeatures
 	VirtualMachine         VirtualMachineFeatures
 	VirtualMachineScaleSet VirtualMachineScaleSetFeatures
 	KeyVault               KeyVaultFeatures
 	Network                NetworkFeatures
 	TemplateDeployment     TemplateDeploymentFeatures
+}
+
+type FrontdoorFeatures struct {
+	IgnoreBackendPoolLimit bool
 }
 
 type VirtualMachineFeatures struct {
--- a/azurerm/internal/provider/features.go
+++ b/azurerm/internal/provider/features.go
@@ -9,6 +9,20 @@ func schemaFeatures(supportLegacyTestSuite bool) *schema.Schema {
 	// NOTE: if there's only one nested field these want to be Required (since there's no point
 	//       specifying the block otherwise) - however for 2+ they should be optional
 	features := map[string]*schema.Schema{
+		"frontdoor": {
+			Type:     schema.TypeList,
+			Optional: true,
+			MaxItems: 1,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"ignore_backend_pool_limit": {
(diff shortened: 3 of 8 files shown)


tombuildsstuff:
@WodansSon there's no need for a feature-flag for this - can we increase the validation limit instead?

WodansSon:
@tombuildsstuff, Yes we could do that, but that is not inclusive and we have no idea what the correct upper limit would be for a given subscription would be. Removing the limitation completely also feels wrong to me, since for most instances of the resource the limitation is still valid. By using a feature flag we are forcing the end user to state that they have received an exemption from Microsoft support and they are declaring that the default limits do not apply to them, effectively opting-out of the check so at that point the results of the attempted provisioning of the resource is on them not us or Azure. I feel this is the best way to go about it, that said I am open to other ideas, suggestions, and input. 🙂

tombuildsstuff:
@WodansSon the Service Team should be able to confirm the maximum limit here, those limits should be published _somewhere_ even if they're (as you mentioned) request-only - if there's no fixed limit, but a quota we should be able to use "at least X" or "at least X, a multiple of Y" as a validation method for these.

From Terraform's side we've a convention of exposing these values and a yellow info box explaining the default maximum value (or that this is an opt-in thing), as shown here: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/search_service#sku - so whilst we could make this a feature-flag - the `features` block is more intended for behavioural features than for validation.

WodansSon:
@tombuildsstuff, from my current understanding there is no real upper limit here(per the service team), it is just what is pressed on the subscription as a `default`(e.g. 50). I agree that Terraform exposes the default values in a info box, however Terraform does not expose an easy way for customers to exempt themselves from this limit once they have had an exemption from Microsoft support. Our options here are to make the max count the absolute maximum limit that the service team will support, which most users will fail because the default is way below that number on their subscription by default and the will get an error when they apply (which is hella annoying), or allow them to say, via a feature flag I have an exemption from Microsoft support and I can have more than the default. It is a grey area here, and the answer, I feel is a fine line, a finessed line, force the user to say they can have more than the default and then let the RP decide if they are true or have Terraform's validation code say, Whoa that is way to many of X, however most of the rules in Terraform I've seen have been in the form of Whoa that is way to many of X, so if we want to go the other way we need to unify this behavior across all resources else it is really confusing for our customers. This really feels like a double edged sword at this point, Terraform, from my understanding, is supposed to give the end user the highest probability of success, but by artificially inflating the max number of instances of a resource to the absolute maximum allowed outside of the normal default for all accounts without exceptions from Microsoft Support, only sets them up for frustration and failure... I am happy to go either way this, it is your product and I will follow whatever your rules are, however I am having an hard time understanding what those rules are as they seem to be a bit arbitrary at times per resource....

tombuildsstuff:
@WodansSon ultimately it's on the service team to publish their maximum values, either "these are the hard limits" or "above X can be requested", from Terraform's side we'll either set an upper threshold of the hard limit (e.g. "maximum of 50") - or a minimum of X (potentially with a "but also must be dividable by Y").

Since in this case there isn't an upper limit we can explicitly lean on here, the validation for this resource will have to be best effort (e.g. "a minimum of 1" with a note in the docs explaining that going above 50 requires a quota increase) - but from Terraform's side we're trying to ensure there's _some_ validation for the field - that there's no upper limit is fine (although not ideal). Ultimately Terraform's validation should be sufficient of saying whether something will pass or fail - so this is ultimately best effort - but notably we shouldn't lean on the RP for validation since it's too late, at that point we're half way through an apply (and potentially provisioning other resources in parallel).

For the moment we should be able to remove the "maximum" validation here and add a note to the docs - but should there be a published request-only maximum value, we can look to add that in the future 👍

tombuildsstuff:
Closing this in favour of #10828 which achieves the same thing but without the `features` flag (since this is validation rather than a behavioural feature)

PR #10802, opened 2021-03-03
Fix for Data source name and example usage

rkaev:
Fix data source name to match to title 
Fix example usage in output 
From
data.azurerm_key_vault_certificate.example.pem
To
data.azurerm_key_vault_certificate_data.example.pem


tombuildsstuff (approved):
LGTM - thanks for this @rkaev

PR #10803, opened 2021-03-03
Update `azurerm_linux|windows_virtual_machine` - support `platform_fault_domain` for VMSS Flex support

ArcturusZhang:


jackofallops (dismissed):
Hi @ArcturusZhang - Thanks for this PR.

I have concerns around the behaviour of this property.  The validation suggests that `-1` is not a valid value, but this is the default and the docs suggest the user can specify it (which they will not currently be able to). Can you please check which is correct and I'll continue review?

Thanks

ArcturusZhang:
> Hi @ArcturusZhang - Thanks for this PR.
> 
> I have concerns around the behaviour of this property. The validation suggests that `-1` is not a valid value, but this is the default and the docs suggest the user can specify it (which they will not currently be able to). Can you please check which is correct and I'll continue review?
> 
> Thanks

Oh, sorry I missed that! I will change the validation to `AtLeast(-1)` to allow the user to explicit set `-1` to it.

The logic in this attribute is like this:
We could set a `platform_fault_domain_count` in the associated `azurerm_orchestrated_virtual_machine_scale_set` resource to define the count of the platform fault domains. For instance, it is assigned by `2`. Then all the VMs added to that `azurerm_orchestrated_virtual_machine_scale_set` could assign a platform_fault_domain like an index, from `0` to `1` in my example. In the REST API, you could leave this property to `null` to let the API choose a `platform_fault_domain` for you.

To support the `null` behaviour, I make the default of this attribute to `-1` because `0` has specific meaning other than absence.

Also when we are not setting the `platform_fault_domain` (or setting it to -1 in terraform), the API will not return the automatically-chosen value in the GET response (the platform fault domain will still be `null` when you send `null`). We have another API the `instanceView` to check the actual platform_fault_domain value.

ArcturusZhang:
Hi @jackofallops I just updated the validation function, please take a look, thanks

website/docs/r/linux_virtual_machine.html.markdown
```diff
@@ -156,6 +156,8 @@ The following arguments are supported:
 
 * `plan` - (Optional) A `plan` block as defined below. Changing this forces a new resource to be created.
 
+* `platform_fault_domain` - (Optional) Specifies the virtual machine scale set logical fault domain into which this Linux Virtual Machine will be created. Defaults to `-1`. By specifying the default value, this Linux Virtual Machine will be automatically assigned to a fault domain that best maintains balance across available fault domains.
```
tombuildsstuff, line 159: we can simplify this to (and mark this as ForceNew):

```suggestion
* `platform_fault_domain` - (Optional) Specifies the Platform Fault Domain in which this Linux Virtual Machine should be created. Defaults to `-1`, which means this will be automatically assigned to a fault domain that best maintains balance across the available fault domains. Changing this forces a new Linux Virtual Machine to be created.
```

website/docs/r/windows_virtual_machine.html.markdown
```diff
@@ -149,6 +149,8 @@ The following arguments are supported:
 
 * `plan` - (Optional) A `plan` block as defined below. Changing this forces a new resource to be created.
 
+* `platform_fault_domain` - (Optional) Specifies the virtual machine scale set logical fault domain into which this Windows Virtual Machine will be created. Defaults to `-1`. By specifying the default value, this Windows Virtual Machine will be automatically assigned to a fault domain that best maintains balance across available fault domains.
```
tombuildsstuff, line 152: we can simplify this to (and mark this as ForceNew):

```suggestion
* `platform_fault_domain` - (Optional) Specifies the Platform Fault Domain in which this Windows Virtual Machine should be created. Defaults to `-1`, which means this will be automatically assigned to a fault domain that best maintains balance across the available fault domains. Changing this forces a new Windows Virtual Machine to be created.
```

tombuildsstuff (approved):
A few comments inline but this otherwise LGTM 👍

tombuildsstuff:
Merging the suggestions so we can get this into this release

tombuildsstuff:
Tests look good 👍

PR #10807, opened 2021-03-03
Fix case sensitivity issue for axurerm_monitor_metric_alert import command

ashsmith:
The import example resource id contained a lowercase letter for monitorAlerts. This is actually case-sensitive and terraform will attempt to force-replace the resource if it's incorrect.


favoretti (approved):
Hi @ashsmith, thank you for this PR, looks good to me! 👍🏻

katbyte (approved):

PR #10810, opened 2021-03-03
add HPCC override, rename appplatorm override to springcloud

jackofallops:
Upcoming quota limit for HPC Caches, throttling test parallelism to limit -1.

Updated override for `springcloud` that was renamed from `appplatform`


mbfrahry (approved):
LGTM

PR #10816, opened 2021-03-03
kubernetes_cluster_resource: lower length validation on windows password

### LINKED ISSUE #10815: Validation of windowsProfile password breaks on existing clusters with shorter passwords
The validation introduced by https://github.com/terraform-providers/terraform-provider-azurerm/pull/10452 breaks updates on existing clusters.

Also, changing the windows password is not supported by AKS, so clusters created with shorter password cannot be updated with when using `azurerm > 2.46`

Error message when running `terraform plan` on an existing cluster:
```
expected length of windows_profile.0.admin_password to be in the range (14 - 123), got <the-actual-password>
```

When trying to update the password:
```
Error: updating Managed Kubernetes Cluster "felipe-k8s" (Resource Group "felipe-aks-rg"): containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="PropertyChangeNotAllowed" Message="Changing property 'windowsProfile.adminPassword' is not allowed." Target="windowsProfile.adminPassword"
```

feliperoveran:
Fixes #10815

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_cluster_resource.go
+++ b/azurerm/internal/services/containers/kubernetes_cluster_resource.go
@@ -599,7 +599,7 @@ func resourceKubernetesCluster() *schema.Resource {
 							Type:         schema.TypeString,
 							Optional:     true,
 							Sensitive:    true,
-							ValidateFunc: validation.StringLenBetween(14, 123),
+							ValidateFunc: validation.StringLenBetween(8, 123),
 						},
 					},
 				},


azurerm/internal/services/containers/kubernetes_cluster_resource.go
```diff
@@ -596,10 +596,12 @@ func resourceKubernetesCluster() *schema.Resource {
...
-							Type:         schema.TypeString,
-							Optional:     true,
-							Sensitive:    true,
-							ValidateFunc: validation.StringLenBetween(14, 123),
+							Type:      schema.TypeString,
+							Optional:  true,
+							Sensitive: true,
+							// not doing length validation here otherwise it breaks on existing
+							// clusters with shorter passwords
+							ValidateFunc: validation.StringIsNotEmpty,
```
favoretti, line 604:
```suggestion
							ValidateFunc: validation.StringLenBetween(8, 123),
```

Hey @feliperoveran, thank you for the PR! Looking good, but could we leave the validation in form of lowering the lower boundary, instead of removing it completely? This way we'll still be able to safeguard from passwords that are too long. I'm not sure whether 8 is low enough, but should be acceptable?

favoretti (changes requested):

azurerm/internal/services/containers/kubernetes_cluster_resource.go
feliperoveran: 8 should be good enough as that used to be the minimum a few months ago :)

djsly: https://docs.microsoft.com/en-us/rest/api/compute/virtualmachinescalesets/createorupdate#virtualmachinescalesetosprofile
```
Minimum-length (Windows): 8 characters 

Minimum-length (Linux): 6 characters 
```

favoretti: @djsly Thank you so much for finding this. @feliperoveran my apologies, would you mind lowering this to 6?

djsly: @favoretti  6 is for Linux, this var is for Windows.

favoretti: @djsly Oh, you're completely correct.

favoretti (approved):
Thank you @feliperoveran, LGTM 👍🏻

katbyte (approved):

PR #10828, opened 2021-03-04
r/frontdoor: removing the MaxItems validation from the Backend Pools

### LINKED ISSUE #10800: `azurerm_frontdoor`: Expose a way to bypass `backend_pool` max `backend_pool` count.

tombuildsstuff:
Azure Support allows overriding the maximum value, so this is best-effort

Supersedes #10801
Fixes #10800

### DIFF
--- a/azurerm/internal/services/frontdoor/frontdoor_resource.go
+++ b/azurerm/internal/services/frontdoor/frontdoor_resource.go
@@ -347,7 +347,6 @@ func resourceFrontDoor() *schema.Resource {
 
 			"backend_pool": {
 				Type:     schema.TypeList,
-				MaxItems: 50,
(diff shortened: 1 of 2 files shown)


WodansSon (approved):
I still believe we should still have a check for the upper bound limit, however unblocking our customers is the most important thing here.

WodansSon:
Tests passed with one expected failure due to known order issue...
image

PR #10831, opened 2021-03-04
r/kubernetes_cluster: recommending using an up to date version of the provider when using AKS

tombuildsstuff:
The behaviour of the AKS API can change over time, in practice using the latest version of the Provider ensures we use the latest API version which matches the latest behaviour of the Azure API


manicminer (approved):
LGTM 👍

PR #10841, opened 2021-03-04
website: removing the old sidebar which is now unused

tombuildsstuff:
I'd been meaning to remove this for a while, but it keeps appearing in spurious PR's - since this is no longer needed we can remove it


katbyte (approved):
LGTM 👍

Issue #8768, opened 2020-10-06
azurerm_app_service ip_restriction doesn't remove subnet_id when unset


martinjt:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
* Terraform v0.12.28
* provider.azurerm v2.30.0

### Affected Resource(s)

* `azurerm_app_service`

### Terraform Configuration Files

before
```hcl

resource "azurerm_app_service" "app" {
  ...
  site_config {
    
    ip_restriction {
      ip_address                = null
      priority                  = 1000
      name                      = "InternalAppSubnet"
      action                    = "Allow"
      virtual_network_subnet_id = azurerm_subnet.app.id
      subnet_id                 = azurerm_subnet.app.id
    }
}

```
After
```hcl

resource "azurerm_app_service" "app" {
  ...
  site_config {
    
    ip_restriction {
      ip_address                = 10.0.0.0/8
      priority                  = 1000
      name                      = "InternalAppSubnet"
      action                    = "Allow"
      virtual_network_subnet_id = null
      subnet_id                 = null
    }
}

```

### Expected Behavior

When switching from using subnet's to IP addresses, the subnet should be blanked.  This is more of an issue if the `ip_restriction` block is part of a dynamic nested block, as removing an item in the middle of the list will cause the plan to think it's changing them in sequence.

### Actual Behavior

The plan tries to add the IP address is added to the existing restriction, but the subnet is not removed.  This causes a failure as only 1 of the 2 can be set at a time.

### Steps to Reproduce

1. Plan and Apply the "before" block into Azure, so you have a subnet added as an IP restriction.
1. Change the block to the "after" block
1. Plan and Apply again.

### Important Factoids

The above usecase is a minimal replication, the most common usecase will be using a dynamic block for the IP restrictions which is far more likely to cause the issue.

### References

* #0000

w0ut0:
[...]

PR #10843, opened 2021-03-04
Fix 8768 corrupt ip_restrictions in App Service and add tests

sdebruyn:
- Added acceptance tests to cover the error from #8768 (the second one didn't succeed at first)
- Fixed an if condition which should have alerted about the error
- Fixed the issue itself by setting the other two of subnet/IP/tag to empty when a restriction is being reused for a new subnet/IP/tag is used

The root cause was that IP restriction lines are being reused. When a reused one had a subnet and it would be changed to an IP, then the subnet was not removed. Thus, to know what has changed, I had to pass ResourceData down the tree.

Any kind of feedback is welcome, I don't regularly use Go.

### DIFF
--- a/azurerm/internal/services/web/app_service.go
+++ b/azurerm/internal/services/web/app_service.go
@@ -1909,7 +1909,7 @@ func expandAppServiceIpRestriction(input interface{}) ([]web.IPSecurityRestricti
 		priority := restriction["priority"].(int)
 		action := restriction["action"].(string)
 
-		if vNetSubnetID != "" && ipAddress != "" && serviceTag != "" {
+		if (vNetSubnetID != "" && (ipAddress != "" || serviceTag != "")) || (ipAddress != "" && (vNetSubnetID != "" || serviceTag != "")) {
 			return nil, fmt.Errorf("only one of `ip_address`, `service_tag` or `virtual_network_subnet_id` can be set for an IP restriction")
 		}
 
--- a/azurerm/internal/services/web/app_service.go
+++ b/azurerm/internal/services/web/app_service.go
@@ -1532,7 +1532,8 @@ func flattenAppServiceIdentity(identity *web.ManagedServiceIdentity) ([]interfac
 	}, nil
 }
 
-func expandAppServiceSiteConfig(input interface{}) (*web.SiteConfig, error) {
+func expandAppServiceSiteConfig(d *schema.ResourceData) (*web.SiteConfig, error) {
+	input := d.Get("site_config")
 	configs := input.([]interface{})
 	siteConfig := &web.SiteConfig{}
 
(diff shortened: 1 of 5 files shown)


azurerm/internal/services/web/app_service.go
```diff
@@ -1882,10 +1881,15 @@ func flattenAppServiceStorageAccounts(input map[string]*web.AzureStorageInfoValu
 	return results
 }
 
-func expandAppServiceIpRestriction(input interface{}) ([]web.IPSecurityRestriction, error) {
+func expandAppServiceIpRestriction(d *schema.ResourceData, schemaKey string) ([]web.IPSecurityRestriction, error) {
 	restrictions := make([]web.IPSecurityRestriction, 0)
 
-	for _, r := range input.([]interface{}) {
+	count := d.Get(schemaKey + ".#").(int)
```
DylanMeeus, line 1887: unsafe caste here ;-)

azurerm/internal/services/web/app_service.go
```diff
@@ -1909,7 +1913,28 @@ func expandAppServiceIpRestriction(input interface{}) ([]web.IPSecurityRestricti
...
+		}
+
+		if d.HasChanges(iterKey+".subnet_id", iterKey+".virtual_network_subnet_id") {
+			if vNetSubnetID != "" {
+				ipAddress = ""
+				serviceTag = ""
+			}
+		}
+
+		if (vNetSubnetID != "" && (ipAddress != "" || serviceTag != "")) || (ipAddress != "" && (vNetSubnetID != "" || serviceTag != "")) {
```
DylanMeeus, line 1937: Personally I'd extract this if-statement to a variable named after what it's checking. 

e.g

```
validAddress := vNetSubnetID != "" && ..

if validAddress {
    // code here 
}
```

jackofallops:
Hi @sdebruyn - I've taken a quick look over this and I think we may be addressing the wrong problem here? It appears that when the `subnet_id` property was deprecated, it was done so with a bug in there that forced users to also specify `virtual_network_subnet_id` to use it, which was not the intent as this property was meant to replace it.  (this can be seen in the example config in the linked issue). I'm going to do a little more investigation, but I think it can be resolved relatively simply. If that's the case, I will open a new PR to supersede this one (with some more complex test cases), if that's OK with you?

sdebruyn:
Hi @jackofallops, certainly! Whatever gets the issue fixed! It's a real dealbreaker for us.

In my findings the extra subnet property was not the cause of this issue (although it was my first thought too).

I would merge the tests anyway so that this issue can be tracked.

jackofallops:
Thanks @sdebruyn - You are correct, there's an issue with the way the flatten and expand functions are working there too... I've got a branch running through the full regression suite at the moment that appears to resolve the initial issue, I'll add some more complex scenarios too for extra confidence.  Typically we phase out deprecated properties from tests, but I think in this case it'll be worth having some coverage there. As soon as I'm confident I'll open the PR and close here with references.

katbyte:
I'm going to close this in favour of #11170 - thanks for starting this @sdebruyn

PR #10866, opened 2021-03-06
`azurerm_postgresql_server` - fix for `TestAccPostgreSQLServer_updated` test case

WodansSon:
Fix `TestAccPostgreSQLServer_updated` test case. The configuration was bad for this test case, so I fixed the issues and the test passes now.

Issues with Test Case:
* The `complete2` test configuration disabled `ssl_enforcement_enabled` but still sent the `ssl_minimal_tls_version_enforced` field causing the error:
```
Code="InvalidParameterValue" Message="Invalid value given for parameter 'MinimalTlsVersion'. Specify a valid parameter value."
```

* The `complete2` test configuration attempts to assign the `storage_endpoint` and `storage_account_access_key` fields after the initial creation of the `threat_detection_policy` causing the error:
```
    TestAccPostgreSQLServer_updated: testing_new_import_state.go:249: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

        (map[string]string) (len=1) {
         (string) (len=52) "threat_detection_policy.0.storage_account_access_key": (string) ""
        }


        (map[string]string) (len=1) {
         (string) (len=52) "threat_detection_policy.0.storage_account_access_key": (string) (len=88) "9eRwWdJ+XyIlp5lXmiBLPm1b7j1cLvyje/Lx4b9AvXDNP4xC/3EXux0wwaUYebrZ7Exg0+7BuAQSgRJdLzCCJw=="
        }
```

* The final step of the `TestAccPostgreSQLServer_updated` test case attempts to revert to the original `gp` sku which has a `storage_profile` defined with a `storage_mb` value of `51200` MB, however the configuration `complete` scales the `storage_profile`'s `storage_mb` up to `640000` MB. So when the test case attempt to revert to the original `gp` sku you recieve the below error. Due to this constraint the last test step can only be revert to the previous sku `complete` which has the same `storage_profile` configuration:
```
Code="UnsupportedUpdate" Message="Cannot update StorageMB Downscaling for server"
```

### DIFF
--- a/azurerm/internal/services/postgres/postgresql_server_resource.go
+++ b/azurerm/internal/services/postgres/postgresql_server_resource.go
@@ -1048,6 +1048,10 @@ func postgreSqlStateRefreshFunc(ctx context.Context, client *postgresql.ServersC
 			return res, string(postgresql.ServerStateInaccessible), nil
 		}
 
-		return res, string(res.ServerProperties.UserVisibleState), nil
+		if res.ServerProperties != nil && res.ServerProperties.UserVisibleState != "" {
+			return res, string(res.ServerProperties.UserVisibleState), nil
+		}
+
+		return res, string(postgresql.ServerStateInaccessible), nil
 	}
 }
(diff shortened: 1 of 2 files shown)


WodansSon:
image

katbyte (approved):
LGTM 👍

PR #10867, opened 2021-03-07
azurerm_key_vault_certificate: Support curve property for EC keys

simonbrady:
The `azurerm_key_vault_certificate` resource allows for ECDSA certificates with a key type of `EC`, but unlike `azurerm_key_vault_key` it doesn't let the user specify which elliptic curve to use.

The [v7.1 Create Certificate API](https://docs.microsoft.com/en-us/rest/api/keyvault/createcertificate/createcertificate) adds a `crv` attribute to the [KeyProperties](https://docs.microsoft.com/en-us/rest/api/keyvault/createcertificate/createcertificate#keyproperties) type, so this PR updates the `keyvault` service to the v7.1 API and adds `curve` to the allowed key properties, matching the behaviour of `azurerm_key_vault_key` wherever possible.

All the `keyvault` acceptance tests pass (including new coverage for `curve`), but I'm raising this as a draft PR to get feedback on some implications of the change.

### DIFF
--- a/azurerm/internal/services/keyvault/key_vault_certificate_resource.go
+++ b/azurerm/internal/services/keyvault/key_vault_certificate_resource.go
@@ -16,6 +16,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/keyvault/parse"
@@ -148,7 +149,7 @@ func resourceKeyVaultCertificate() *schema.Resource {
 											string(keyvault.RSA),
 											string(keyvault.RSAHSM),
 											string(keyvault.Oct),
-										}, false),
+										}, true),
 									},
 									"reuse_key": {
 										Type:     schema.TypeBool,
--- a/azurerm/internal/services/keyvault/key_vault_certificate_resource.go
+++ b/azurerm/internal/services/keyvault/key_vault_certificate_resource.go
@@ -108,6 +108,18 @@ func resourceKeyVaultCertificate() *schema.Resource {
 							MaxItems: 1,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
+									"curve": {
+										Type:     schema.TypeString,
+										Optional: true,
+										Computed: true,
+										ForceNew: true,
+										ValidateFunc: validation.StringInSlice([]string{
+											string(keyvault.P256),
+											string(keyvault.P256K),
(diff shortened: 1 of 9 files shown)


simonbrady:
Minor merge conflicts with #10873 which updates the keyvault/mgmt API, but I can easily rebase this if that one's merged first.

simonbrady:
Rebased off master, questions in #10868 still to be resolved.

azurerm/internal/services/keyvault/key_vault_key_resource.go
```diff
@@ -102,9 +102,9 @@ func resourceKeyVaultKey() *schema.Resource {
 				ForceNew: true,
 				ValidateFunc: validation.StringInSlice([]string{
 					string(keyvault.P256),
+					string(keyvault.P256K),
 					string(keyvault.P384),
 					string(keyvault.P521),
```
manicminer, line 107: Can we add the old value as a hardcoded string, and translate it so that existing configurations continue to work? We can add a TODO to remove it in v3.0.

```suggestion
					string(keyvault.P521),
					"SECP256K1", // TODO: remove this in v3.0 as it was renamed to keyvault.P256K
```

azurerm/internal/services/keyvault/key_vault_certificate_resource.go
```diff
@@ -127,6 +142,12 @@ func resourceKeyVaultCertificate() *schema.Resource {
 										Type:     schema.TypeString,
 										Required: true,
 										ForceNew: true,
+										ValidateFunc: validation.StringInSlice([]string{
+											string(keyvault.EC),
+											string(keyvault.ECHSM),
+											string(keyvault.RSA),
+											string(keyvault.RSAHSM),
+										}, false),
```
manicminer, line 150: We should be case-insensitive here until v3.0 to avoid breaking existing configs

```suggestion
										}, true), // TODO: make this case-sensitive in v3.0
```

manicminer (changes requested):
Hi @simonbrady, thanks for this PR! Overall this looks good and the tests are passing, but where possible we need to ensure that existing configurations are not broken. I have some suggestions inline.

If you want to compute the `key_size` property, you should be able to do this if you move the logic in [your gist](https://gist.github.com/simonbrady/82c0805acc8a32665fcc6e4bfea54d9c) into the `expandKeyVaultCertificatePolicy()` function - just be sure to prefer the value set in configuration to prevent unwanted diffs. Also add the Computed flag to the key_size property to let Terraform know.

manicminer:
@simonbrady I hope you don't mind, I've pushed the changes detailed in my earlier review including making the `key_size` property optional in `azurerm_key_vault_certificate`. Let me know if this doesn't capture your original intent, I'm going to open this for further review in the meantime. Thanks!

katbyte (approved):
LGTM 👍

manicminer:
Rebased, now re-running tests

manicminer:
Tests passing (service flapping, passed on second run)

Screenshot 2021-04-29 at 21 48 15
Screenshot 2021-04-29 at 21 48 28

simonbrady:
Thanks for all the assistance @manicminer - I hadn't had time to get to this, but I'm going to claim that's exactly what I would've done :)

PR #10874, opened 2021-03-08
Update azurerm_subscription Docs Page 

jtracey93:
This PR updates the output explanation for ID, as currently it states the subscription ID is exported but actually the ARM resource ID for the alias is exported.


katbyte (approved):
LGTM 👍

Issue #9869, opened 2020-12-15
azurerm_function_app_host_keys is highly unstable


ChristianSauer:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform 0.14
azurerm: 2.40.0
codecentric/leanix: 1.0.0-rc4

### Affected Resource(s)

* `data azurerm_function_app_host_keys`

### Terraform Configuration Files

```hcl
assume we have an app service plan, resource group etc.

data "archive_file" "functiondata" {
  type        = "zip"
  source_dir = abspath("${path.module}/../../functioncode")
  output_path = "${path.module}/files/function.zip"
}

resource "azurerm_storage_blob" "storageblob-function" {
  name                   = "deployments/${data.archive_file.functiondata.output_md5}.zip"
  storage_account_name   = azurerm_storage_account.storageacc.name
  storage_container_name = azurerm_storage_container.storagecontainer.name
  type                   = "Block"
  source                 = data.archive_file.functiondata.output_path
}

resource "azurerm_function_app" "func" {
  name                       = "func-${var.project_name}-${var.region}"
  location                   = azurerm_resource_group.rg.location
app_settings = {
        https_only = true
        FUNCTIONS_WORKER_RUNTIME = "node"
        WEBSITE_NODE_DEFAULT_VERSION = "~12"
        FUNCTION_APP_EDIT_MODE = "readonly"
        WEBSITE_RUN_FROM_PACKAGE = "https://${azurerm_storage_account.storageacc.name}.blob.core.windows.net/${azurerm_storage_container.storagecontainer.name}/${azurerm_storage_blob.storageblob-function.name}${data.azurerm_storage_account_sas.sas-function.sas}"
        
    }
}

data "azurerm_function_app_host_keys" "host_keys" {
  depends_on = [  azurerm_function_app.func]
  name                =  azurerm_function_app.func.name
  resource_group_name =  azurerm_resource_group.rg.name
}

resource "leanix_webhook_subscription" "events_azure" {
...
[...]

Issue #9854, opened 2020-12-14
Return the app function keys from the azurerm_function_app without relying on azurerm_function_app_host_keys data source


sonic1981:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

This is closely related to this issue [Feature request: export Functions host keys](https://github.com/terraform-providers/terraform-provider-azurerm/issues/699). Unfortunately the solution here seems to have been to create a datasource that returns these values but to obtain that data the function app must exist. So the flow of 

- Create function
- Reference keys in another resource

doesn't work. You would need to (unless I'm mistaken) need to run these two scripts separately and the downsteam dependencies don't work.

### New or Affected Resource(s)

* `azurerm_function_app`
* `azurerm_function_app_host_keys`

### Potential Terraform Configuration

The following: produces the error
```hcl
resource "azurerm_function_app" "sche" {
  name               = ""
  location          = ""
 ...etc
}


data "azurerm_function_app_host_keys" "dsche" {
  name                = azurerm_function_app.sche.name
  resource_group_name = var.rg.name
}

resource "azurerm_logic_app_action_http" "webhook" {
  name         = "webhook"
  logic_app_id = azurerm_logic_app_workflow.schew.id
  method       = "GET"
  uri          = "https://${azurerm_function_app.sche.default_hostname}/api/RunSchedule?code=${data.azurerm_function_app_host_keys.dsche.default_function_key }"
}
```

```
Error: cannot read ID for AzureRM Function App "..." (Resource Group "...")

  on modules\schedule\main.tf line 71, in data "azurerm_function_app_host_keys" "dsche":
  71: data "azurerm_function_app_host_keys" "dsche" {
```

Can we get `azurerm_function_app` to return the keys as an attribute so that we could then reference them like below without the need for a data source?

```hcl
resource "azurerm_function_app" "sche" {
  name               = ""
  location          = ""
 ...etc
}


resource "azurerm_logic_app_action_http" "webhook" {
  name         = "webhook"
  logic_app_id = azurerm_logic_app_workflow.schew.id
  method       = "GET"
[...]

PR #10894, opened 2021-03-09
azurerm_function_app_host_keys: Retry when reading function key

lawrencegripper:
This looks to address the issue seen in  #9854 and #9869. 

Specifically there is a race condition between the function app being created and the keys being available to read. This appears more in premium funcs on vnets than on normal functions. 

This PR adds retry to the request to get keys, this should resolve the issue. I tested this approach with the following script to validate it, [see blog here](https://blog.gripdev.xyz/2021/03/09/azure-functions-get-key-from-terraform-without-internalservererror/)

Integration tests passing 
image

### DIFF
--- a/azurerm/internal/services/web/function_app_host_keys_data_source.go
+++ b/azurerm/internal/services/web/function_app_host_keys_data_source.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tombuildsstuff (approved):
LGTM 👍

PR #10896, opened 2021-03-09
move subscriptionless ID parsing to helper func

jackofallops:


manicminer (changes requested):
Hi @jackofallops, this looks good, I would suggest adding a godoc comment for `ParseAzureResourceIDWithoutSubscription()` and also a unit test.

manicminer (approved):
LGTM 🚀

PR #10908, opened 2021-03-10
Customize ARM correlation request id via `ARM_CORRELATION_REQUEST_ID`

magodo:
Currently, everytime Terraform initialized, it will generate a random UUID
to be used as the `x-ms-correlation-request-id` in the HTTP request sent via Go SDK.

In some cases, users might want to get that correlation request id so that
it can be used to track the E2E traces. This can be done by looking into the debug
log or using some http proxy. However, this is not an effective way in case Terraform
is running in a large scale automatically, where the debug log is disabled (for some
performance reason), also the proxy is not a good choice there. In this case, allowing
users to specify a correlation request id to be used by Terraform would be a good alternative.

This PR allows users to specify an environment variable `ARM_CORRELATION_REQUEST_ID`, which
takes higher priority than auto-generating a UUID.

### DIFF
--- a/azurerm/internal/common/correlation_id.go
+++ b/azurerm/internal/common/correlation_id.go
@@ -24,7 +24,7 @@ func withCorrelationRequestID(uuid string) autorest.PrepareDecorator {
 	return autorest.WithHeader(HeaderCorrelationRequestID, uuid)
 }
 
-// correlationRequestID either uses a user provided correlation request id or generates a UUID to pass through `x-ms-correlation-request-id` header.
+// correlationRequestID generates an UUID to pass through `x-ms-correlation-request-id` header.
 func correlationRequestID() string {
 	msCorrelationRequestIDOnce.Do(func() {
 		var err error
@@ -32,7 +32,6 @@ func correlationRequestID() string {
 		if err != nil {
 			log.Printf("[WARN] Failed to generate uuid for msCorrelationRequestID: %+v", err)
 		}
-
 		log.Printf("[DEBUG] Genereated Provider Correlation Request Id: %s", msCorrelationRequestID)
 	})
 
--- a/azurerm/internal/common/correlation_id.go
+++ b/azurerm/internal/common/correlation_id.go
@@ -2,7 +2,6 @@ package common
 
 import (
 	"log"
-	"os"
 	"sync"
 
 	"github.com/Azure/go-autorest/autorest"
@@ -12,7 +11,6 @@ import (
 const (
 	// HeaderCorrelationRequestID is the Azure extension header to set a user-specified correlation request ID.
 	HeaderCorrelationRequestID = "x-ms-correlation-request-id"
-	envArmCorrelationRequestID = "ARM_CORRELATION_REQUEST_ID"
 )
 
 var (
@@ -29,12 +27,6 @@ func withCorrelationRequestID(uuid string) autorest.PrepareDecorator {
 // correlationRequestID either uses a user provided correlation request id or generates a UUID to pass through `x-ms-correlation-request-id` header.
 func correlationRequestID() string {
 	msCorrelationRequestIDOnce.Do(func() {
-		msCorrelationRequestID = os.Getenv(envArmCorrelationRequestID)
-		if msCorrelationRequestID != "" {
-			log.Printf("[DEBUG] User Specified Provider Correlation Request Id: %s", msCorrelationRequestID)
-			return
-		}
-
 		var err error
 		msCorrelationRequestID, err = uuid.GenerateUUID()
 		if err != nil {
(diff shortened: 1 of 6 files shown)


azurerm/internal/common/correlation_id.go
```diff
@@ -24,12 +26,17 @@ func withCorrelationRequestID(uuid string) autorest.PrepareDecorator {
 	return autorest.WithHeader(HeaderCorrelationRequestID, uuid)
 }
 
-// correlationRequestID generates an UUID to pass through `x-ms-correlation-request-id` header.
+// correlationRequestID either uses a user provided correlation request id or generates a UUID to pass through `x-ms-correlation-request-id` header.
 func correlationRequestID() string {
 	msCorrelationRequestIDOnce.Do(func() {
+		msCorrelationRequestID = os.Getenv(envArmCorrelationRequestID)
```
tombuildsstuff, line 32: rather than pulling this from an env var - we should instead make this configurable via the provider block, so let's remove this:

```suggestion
```

meaning we could instead add something like this:

```
	if !o.DisableCorrelationRequestID {
		if o.CustomCorrelationRequestID != nil {
			useCustomCorrelationRequestID(*o.CustomCorrelationRequestID)
		}
		c.RequestInspector = withCorrelationRequestID(correlationRequestID())
	}
```

here: https://github.com/terraform-providers/terraform-provider-azurerm/blob/5facd461e7075074c8270d68d884692c201f7c45/azurerm/internal/common/client_options.go#L44-L46

and presumably add a setter method in the form of `useCustomCorrelationRequestID` to this file, which'd allow this to be configured in the provider block and via an env var - WDYT?

tombuildsstuff (changes requested):
hey @magodo 

Thanks for this PR - taking a look through this conceptually looks good, but the implementation wants tweaking slightly, since we shouldn't be reading environment variables at this layer - instead can we thread this through from the config object?

Thanks!

tombuildsstuff:
@magodo also worth noting some API's reject sending the same correlation ID's to these services - so this won't work for those (albeit most of those disable sending the correlation ID at this point)

azurerm/internal/common/correlation_id.go
tombuildsstuff: yeah sorry I'd meant to edit this but forgot 🤦 - we can add this to the `ClientBuilder` struct but still feed this from an env var, it just means it's pulled from a single place, which achieves the same thing (e.g. having all of those in the same place, so this isn't hidden) but means this isn't exposed in that block

magodo:
@tombuildsstuff I've now lift the `custom_correlation_request_id` to the provider block, I guess this is what you intended to be? Please take another look :)

azurerm/internal/provider/provider.go
```diff
@@ -228,6 +228,13 @@ func azureProvider(supportLegacyTestSuite bool) terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("ARM_STORAGE_USE_AZUREAD", false),
 				Description: "Should the AzureRM Provider use AzureAD to access the Storage Data Plane API's?",
 			},
+
+			"custom_correlation_request_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     schema.EnvDefaultFunc("ARM_CORRELATION_REQUEST_ID", ""),
+				Description: "The value of the x-ms-correlation-request-id header (otherwise an auto-generated UUID will be used).",
+			},
```
tombuildsstuff, line 237: we can remove this and instead use `os.GetEnv` below since this isn't an intentionally user exposed value

website/docs/index.html.markdown
```diff
@@ -136,6 +136,8 @@ More information on [how to configure a Service Principal using Managed Service
 
 For some advanced scenarios, such as where more granular permissions are necessary - the following properties can be set:
 
+* `custom_correlation_request_id` - (Optional) The value of the `x-ms-correlation-request-id` header, otherwise an auto-generated UUID will be used.
+
```
tombuildsstuff, line 140: we can remove this, since this isn't directly exposed here:

```suggestion
```

azurerm/internal/provider/provider.go
```diff
@@ -312,6 +319,7 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
 			TerraformVersion:            terraformVersion,
 			PartnerId:                   d.Get("partner_id").(string),
 			DisableCorrelationRequestID: d.Get("disable_correlation_request_id").(bool),
+			CustomCorrelationRequestID:  d.Get("custom_correlation_request_id").(string),
```
tombuildsstuff, line 322: since we're not exposing this in the Provider block, we can pull from the env var here directly, e.g.

```suggestion
			CustomCorrelationRequestID:  os.GetEnv("ARM_CORRELATION_REQUEST_ID"),
```

tombuildsstuff (approved):
hey @magodo 

Thanks for pushing those changes - whilst this mostly LGTM I'm going to push a commit to resolve the final comments but this should otherwise be good to go 👍 

Thanks!

PR #10914, opened 2021-03-10
New Api Management Email Template Resource

jan-mrm:
Possible solution/implementation that resolves https://github.com/terraform-providers/terraform-provider-azurerm/issues/10875 - new resource azurerm_api_management_email_template

Two things to point out (both have a comment in the code):

**1.**: The generated enum values for the template name all start with a lower case letter. That is what the Management API documents (https://docs.microsoft.com/en-us/rest/api/apimanagement/2019-12-01/emailtemplate/createorupdate#templatename) but it is not what the API accepts. It does want the first letter to be upper case.
```
GET /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ApiManagement/service/{serviceName}/templates/confirmSignUpIdentityDefault?api-version=2019-12-01
-> 404

GET /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ApiManagement/service/{serviceName}/templates/ConfirmSignUpIdentityDefault?api-version=2019-12-01
-> 200
```
A issue about that was just opened: https://github.com/Azure/azure-rest-api-specs/issues/13341
I'm guessing the enums are generated from the Management API.

For now: I capitalized the first letter in the validation of the template_name definition.

---

**2.**: The Management API only allows to update templates and not to create them. A PUT/PATCH updates/customizes the template and a DELETE resets it. I wanted to avoid having to `terraform import` each template that is customized through this terraform resource. 

So when creating the terraform resource (case IsNewResource) and the template is still default (so not yet edited, isDefault flag) then the ImportAsExistsError is not issued.

---

My local manual testing worked, and the implemented acc tests passed.

I'm new to Golang so any feedback is much appreciated.


website/docs/r/api_management_email_template.html.markdown
```diff
@@ -0,0 +1,104 @@
...
+
+---
+
+A `parameters` block exports the following:
+
+* `name` - The name of this parameter.
+
+* `title` - The title of this parameter.
+
+* `description` - The description of this parameter.
```
tombuildsstuff, line 87: are these parameters the ones listed above? if so, maybe we don't need to export these if they're already known?

tombuildsstuff (changes requested):
hey @jan-mrm 

Thanks for this PR :)

I've taken a look through and this is off to a good start, if we can fix up the comments then we should be able to take another look and run the tests 👍 

Thanks!

azurerm/internal/services/apimanagement/api_management_email_template_resource.go
```diff
@@ -0,0 +1,248 @@
...
+					},
+				},
+			},
+		},
+	}
+}
+
+func resourceApiManagementEmailTemplateCreateUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).ApiManagement.EmailTemplateClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
```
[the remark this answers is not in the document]
jan-mrm: fixed

azurerm/internal/services/apimanagement/api_management_email_template_resource.go
```diff
@@ -0,0 +1,248 @@
...
+	defer cancel()
+
+	resourceGroup := d.Get("resource_group_name").(string)
+	serviceName := d.Get("api_management_name").(string)
+	templateName := apimanagement.TemplateName(d.Get("template_name").(string))
+
+	if d.IsNewResource() {
```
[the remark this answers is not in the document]
jan-mrm: good idea, fixed

website/docs/r/api_management_email_template.html.markdown
jan-mrm: Yes, those are the `$DevFirstName` parameters. After you brought my attention to this, I double checked and realized that the available parameters depend on the template. So I updated the *Note* above.

My question is how to proceed with those parameters?
It is not plausible to reference the parameters in e.g. the `subject` in the tf resource because you would have a self-reference. Also you would need to know the parameter in advance.
We could just not export them (remove them from the resource schema definition) and put a reference in the *Note* that one can look them up in the Azure Portal - Notification Template section of the ApiManagement Service instance or Management REST API.
Or export them and hint that they also can be seen in the tf state?
Or even list all available parameters for each of the possible templates?

What do you think? :)

jan-mrm (commented):
Thanks @tombuildsstuff for the review and comments/suggestions!

website/docs/r/api_management_email_template.html.markdown
jan-mrm: I just pushed my proposal: I removed the parameters and just mentioned in the *Note* of the documentation where one can find the available values.

jan-mrm:
Hi @tombuildsstuff, 
I implemented your review findings and proposed a solution for the parameters. I ran the 3 implemented acc tests and they pass.
If you want you can have another look at this PR :)

manicminer (approved):
@jan-mrm Thanks for making the changes. This LGTM and the tests are passing.

Screenshot 2021-04-07 at 01 16 34

PR #10919, opened 2021-03-10
Include time in start and expiry arguments

jogleasonjr:
I could not generate a valid SAS token using the provided example, unless an ISO-8601 compliant time format was appended to the `start` and `expiry` arguments. Azure would reject the URLs with the SAS token appended.

The [tests](https://github.com/terraform-providers/terraform-provider-azurerm/blob/c39a047aef64cd045b3c1003a49c3ff5e61519b3/azurerm/internal/services/storage/storage_account_sas_data_source_test.go#L18) use this:

```golang
utcNow := time.Now().UTC()
startDate := utcNow.Format(time.RFC3339)
endDate := utcNow.Add(time.Hour * 24).Format(time.RFC3339)
```

Which result in this:
```
2021-03-10T17:21:19Z
2021-03-11T17:21:19Z
```

So I think the time should be included in the example docs.

### DIFF
--- a/website/docs/d/storage_account_sas.html.markdown
+++ b/website/docs/d/storage_account_sas.html.markdown
@@ -54,8 +54,8 @@ data "azurerm_storage_account_sas" "example" {
     file  = false
   }
 
-  start  = "2018-03-21"
-  expiry = "2020-03-21"
+  start  = "2018-03-21T00:00:00Z"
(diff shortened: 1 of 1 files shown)


katbyte (approved):
THanks @jogleasonjr - LGTM 👍 (an aside it might make sense to add validation to the resource to catch this at plan)

PR #10920, opened 2021-03-10
key_vault_secret: Set value of recoverd key vault secret

dvob:
Currently if you create a secret for which already a soft-deleted secret exists the soft-deleted secret is only recoverd but the value is not set. This change also sets the value after the secret is recovered.


tombuildsstuff (approved):
LGTM - thanks for this @dvob  👍

tombuildsstuff:
Tests pass:

Screenshot 2021-03-11 at 11 48 28

PR #10926, opened 2021-03-11
Upgrade Cosmos version from `2020-04-01-preview` to `2021-01-15`

yupwei68:
All tests have passed

changes:
1. variable `indexing_mode` and `order`  in resources `azurerm_cosmosdb_sql_container` and `azurerm_cosmosdb_gremlin_graph` is defined and returned in lower case instead of hybrid case. We keep the original hybrid cases instead of breaking changes. 

breaking changes:
1. cosmos mongodb account default to "v3.6" instead of "v3.2". The changes occur in `azurerm_cosmosdb_account` tests. Thus, `_id`  index is required in `azurerm_cosmosdb_mongo_collection` .
2. "patition_key_path" is "required" instead of "optional" in `azurerm_cosmosdb_gremlin_graph` and `azurerm_cosmosdb_sql_container` .


yupwei68:
image

azurerm/internal/services/cosmos/common/schema.go
```diff
@@ -123,7 +123,7 @@ func CosmosDbIndexingPolicySchema() *schema.Schema {
 					ValidateFunc: validation.StringInSlice([]string{
 						string(documentdb.Consistent),
 						string(documentdb.None),
-					}, false),
+					}, true),
```
katbyte, line 126: Why are we making this case insensitive? the SDK values should match what we get from the API?

azurerm/internal/services/cosmos/common/schema.go
```diff
@@ -179,7 +179,7 @@ func CosmosDbIndexingPolicySchema() *schema.Schema {
 												[]string{
 													string(documentdb.Ascending),
 													string(documentdb.Descending),
-												}, false),
+												}, true),
```
katbyte, line 182: same here?

azurerm/internal/services/cosmos/cosmosdb_account_data_source_test.go
```diff
@@ -54,6 +54,8 @@ func (CosmosDBAccountDataSourceResource) basic(data acceptance.TestData) string
 data "azurerm_cosmosdb_account" "test" {
   name                = azurerm_cosmosdb_account.test.name
   resource_group_name = azurerm_resource_group.test.name
+
+  depends_on = [azurerm_cosmosdb_account.test]
```
katbyte, line 58: this is.. odd - why is this required?

azurerm/internal/services/cosmos/cosmosdb_gremlin_graph_resource.go
```diff
@@ -114,7 +114,7 @@ func resourceCosmosDbGremlinGraph() *schema.Resource {
 								string(documentdb.Consistent),
 								string(documentdb.Lazy),
 								string(documentdb.None),
-							}, false),
+							}, true),
```
katbyte, line 117: same here, why is this now case insensitive?

katbyte (commented):
Thanks @yupwei68 - i've left some comments inline. Also we have a test failure:
```
------- Stdout: -------
=== RUN   TestAccCosmosDbMongoCollection_autoscale
=== PAUSE TestAccCosmosDbMongoCollection_autoscale
=== CONT  TestAccCosmosDbMongoCollection_autoscale
    testing_new_import_state.go:249: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) (len=1) {
         (string) (len=10) "throughput": (string) (len=3) "500"
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=10) "throughput": (string) (len=3) "400"
        }
--- FAIL: TestAccCosmosDbMongoCollection_autoscale (1235.42s)
FAIL

```

katbyte (changes requested):
.

azurerm/internal/services/cosmos/common/schema.go
yupwei68: In preview version, the value is "consistent" and "none", but in current version, the value is "Consistent" and "None". To resolve the breaking change, I set it to be case insensitive?

azurerm/internal/services/cosmos/common/schema.go
yupwei68: same here.

azurerm/internal/services/cosmos/cosmosdb_account_data_source_test.go
yupwei68: I have upgraded my local Terraform to 0.14. And the data source acctest runs fail for lack of the declare of the dependency. But I think it should work in TC. I'll remove it.

azurerm/internal/services/cosmos/cosmosdb_gremlin_graph_resource.go
yupwei68: same here. To solve the breaking change.

yupwei68:
Hi @katbyte  Thanks for comments. I've run this test for three times locally. It all passed. I think it's a random hiccup.
=== RUN   TestAccCosmosDbMongoCollection_autoscale
=== PAUSE TestAccCosmosDbMongoCollection_autoscale
=== CONT  TestAccCosmosDbMongoCollection_autoscale
--- PASS: TestAccCosmosDbMongoCollection_autoscale (5617.59s)

yupwei68:
Hi kt, I've solved the breaking change. And all tests have passed.
image

azurerm/internal/services/cosmos/common/schema.go
katbyte: instead of making this case insensitive can we just correct the casing to the SDK values on read?

also - could we open an issue on the SDK or somewhere to track this discrepancy?

azurerm/internal/services/cosmos/cosmosdb_gremlin_graph_resource.go
```diff
@@ -92,6 +92,7 @@ func resourceCosmosDbGremlinGraph() *schema.Resource {
 				Optional:     true,
 				ForceNew:     true,
 				ValidateFunc: validation.StringIsNotEmpty,
+				Default:      "/_partitionKey",
```
katbyte, line 95: talked this over internal and instead of a default lets just leave this as required so the user can decide what to do rather then have magic upgrade behaviour

azurerm/internal/services/cosmos/cosmosdb_gremlin_graph_resource.go
katbyte: same as other 2 - lets correc the case on read and add a comment to an open issue tracking this

katbyte (changes requested):
Thanks, just a few small things left to address and this'll be good to merge

azurerm/internal/services/cosmos/common/schema.go
yupwei68: The case has been change when the service has released 2020-04-01 stable version. The old 2020-04-01-preview version has been deleted. 
Since preview version permits breaking changes, shall I not open the issue?

yupwei68:
Thanks kt for your comments. I've changed all these. Please continue reviewing.

azurerm/internal/services/cosmos/cosmosdb_gremlin_graph_resource.go
```diff
@@ -102,14 +103,15 @@ func resourceCosmosDbGremlinGraph() *schema.Resource {
 							Default:  true,
 						},
 
+						// case change in 2021-01-15, issue https://github.com/Azure/azure-rest-api-specs/issues/14051
 						"indexing_mode": {
```
katbyte, line 107: could we add a comment 
```suggestion
						// todo: change to SDK constants and remove translation code in 3.0
						"indexing_mode": {
```

for the constants we've done this with?

katbyte (approved):
Thanks @yupwei68 - aside from one comment inline (add todo 3.0 to those constants) and these two test failures this looks ready to merge!

```
------- Stdout: -------
=== RUN   TestAccCosmosDbSqlStoredProcedure_basic
=== PAUSE TestAccCosmosDbSqlStoredProcedure_basic
=== CONT  TestAccCosmosDbSqlStoredProcedure_basic
    testing.go:620: Step 1/2 error: Error running pre-apply refresh: 
        Error: Missing required argument
        
          on config293530581/terraform_plugin_test.tf line 35, in resource "azurerm_cosmosdb_sql_container" "test":
          35: resource "azurerm_cosmosdb_sql_container" "test" {
        
        The argument "partition_key_path" is required, but no definition was found.
        
--- FAIL: TestAccCosmosDbSqlStoredProcedure_basic (6.26s)
FAIL

------- Stderr: -------
2021/04/22 04:53:16 [DEBUG] not using binary driver name, it's no longer needed
2021/04/22 04:53:16 [DEBUG] not using binary driver name, it's no longer needed
------- Stdout: -------
=== RUN   TestAccCosmosDbSqlStoredProcedure_update
=== PAUSE TestAccCosmosDbSqlStoredProcedure_update
=== CONT  TestAccCosmosDbSqlStoredProcedure_update
    testing.go:620: Step 1/4 error: Error running pre-apply refresh: 
        Error: Missing required argument
        
          on config415328271/terraform_plugin_test.tf line 35, in resource "azurerm_cosmosdb_sql_container" "test":
          35: resource "azurerm_cosmosdb_sql_container" "test" {
        
        The argument "partition_key_path" is required, but no definition was found.
        
--- FAIL: TestAccCosmosDbSqlStoredProcedure_update (6.02s)
FAIL

------- Stderr: -------
2021/04/22 04:53:21 [DEBUG] not using binary driver name, it's no longer needed
2021/04/22 04:53:21 [DEBUG] not using binary driver name, it's no longer needed
```

yupwei68:
image
Hi kt, tests have passed.

yupwei68:
Hi kt, I've updated the description of the PR to list all changes and breaking changes.

PR #10939, opened 2021-03-11
Docs: azurerm_resource_group_template_deployment show passing variables

lawrencegripper:
Tweak to the docs to demonstrate how to pass TF vars to the ARM template.

### DIFF
--- a/website/docs/r/resource_group_template_deployment.html.markdown
+++ b/website/docs/r/resource_group_template_deployment.html.markdown
@@ -45,7 +45,7 @@ resource "azurerm_resource_group_template_deployment" "example" {
         {
             "type": "Microsoft.Network/virtualNetworks",
             "apiVersion": "2020-05-01",
-            "name": "acctest-network",
+            "name": "[variables("vnetName")]",
             "location": "[resourceGroup().location]",
             "properties": {
                 "addressSpace": {
--- a/website/docs/r/resource_group_template_deployment.html.markdown
+++ b/website/docs/r/resource_group_template_deployment.html.markdown
@@ -82,7 +82,9 @@ The following arguments are supported:
 
 * `debug_level` - (Optional) The Debug Level which should be used for this Resource Group Template Deployment. Possible values are `none`, `requestContent`, `responseContent` and `requestContent, responseContent`.
 
-* `parameters_content` - (Optional) The contents of the ARM Template parameters file - containing a JSON list of parameters. See example for how to pass in Terraform Variable to an ARM parameter.
+* `parameters_content` - (Optional) The contents of the ARM Template parameters file - containing a JSON list of parameters.
+
+-> An example of how to pass Terraform variables into an ARM Template can be seen in the example.
 
 * `tags` - (Optional) A mapping of tags which should be assigned to the Resource Group Template Deployment.
 
--- a/website/docs/r/resource_group_template_deployment.html.markdown
+++ b/website/docs/r/resource_group_template_deployment.html.markdown
@@ -15,31 +15,47 @@ Manages a Resource Group Template Deployment.
 ## Example Usage
 
 ```hcl
+locals {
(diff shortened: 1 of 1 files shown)


website/docs/r/resource_group_template_deployment.html.markdown
```diff
@@ -15,31 +15,47 @@ Manages a Resource Group Template Deployment.
 ## Example Usage
 
 ```hcl
+locals {
+  vnet_name = "example-vnet"
+}
+
 resource "azurerm_resource_group_template_deployment" "example" {
-  name                = "example-deploy"
+  name                = "lgtest1"
```
katbyte, line 23: could we revert this
```suggestion
  name                = "example-deploy"
```

katbyte (approved):
aside from 1 comment this LGTM 👍  htanks @lawrencegripper

website/docs/r/resource_group_template_deployment.html.markdown
```diff
@@ -15,31 +15,47 @@ Manages a Resource Group Template Deployment.
...
 resource "azurerm_resource_group_template_deployment" "example" {
   name                = "example-deploy"
   resource_group_name = "example-group"
   deployment_mode     = "Complete"
+  parameters_content  = jsonencode({
+    "vnetName" = {
+      value = local.vnet_name
+    }
+  })
   template_content    = <<TEMPLATE
```
katbyte, line 31:
```suggestion
  template_content = <<TEMPLATE
```

website/docs/r/resource_group_template_deployment.html.markdown
```diff
@@ -15,31 +15,47 @@ Manages a Resource Group Template Deployment.
...
-      }
-    }
-  ]
+    },
+    "variables": {},
+    "resources": [
+        {
+            "type": "Microsoft.Network/virtualNetworks",
+            "apiVersion": "2020-05-01",
+            "name": "acctest-network",
```
tombuildsstuff, line 48: for the purposes of this example we should probably use this too, e.g.

```suggestion
            "name": "[variables("vnetName")]",
```

website/docs/r/resource_group_template_deployment.html.markdown
```diff
@@ -66,7 +82,7 @@ The following arguments are supported:
 
 * `debug_level` - (Optional) The Debug Level which should be used for this Resource Group Template Deployment. Possible values are `none`, `requestContent`, `responseContent` and `requestContent, responseContent`.
 
-* `parameters_content` - (Optional) The contents of the ARM Template parameters file - containing a JSON list of parameters.
+* `parameters_content` - (Optional) The contents of the ARM Template parameters file - containing a JSON list of parameters. See example for how to pass in Terraform Variable to an ARM parameter.
```
tombuildsstuff, line 85: Let's make this a blue box:

```suggestion
* `parameters_content` - (Optional) The contents of the ARM Template parameters file - containing a JSON list of parameters.

-> An example of how to pass Terraform variables into an ARM Template can be seen in the example.
```

tombuildsstuff (changes requested):
hey @lawrencegripper 

Thanks for pushing those changes

Taking a look through I think it'd be good to demonstrate using the new variable in this example, but if we can fix that up then this should otherwise be good to go 👍 

Thanks!

website/docs/r/resource_group_template_deployment.html.markdown
lawrencegripper: Agree, good call.

lawrencegripper:
Thanks for taking a look, suggestions look good to me. Have accepted them all.

lawrencegripper:
I've added in additional example of how to use ARM outputs from the template in TF too.

katbyte (approved):
LGTM 👍

PR #10956, opened 2021-03-13
Additional Note for Allow Azure Service - Synapse Firewall

Lachlan-White:
The `Allow access to Azure services` feature requires the `name` of the Synapse Firewall rule to be set as `AllowAllWindowsAzureIps`.

This not helps to prevent this error


katbyte (approved):
LGTM 👍

Issue #10077, opened 2021-01-07
Case sensitive resource_group_name


Tbohunek:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

```
Terraform v0.14.3
+ provider registry.terraform.io/hashicorp/azuread v1.1.1
+ provider registry.terraform.io/hashicorp/azurerm v2.41.0
```

### Affected Resource(s)

* `azurerm_windows_virtual_machine`
* `azurerm_linux_virtual_machine` (possibly)

Surprisingly, the issue does not seem to impact `azurerm_network_interface`

### Terraform Configuration Files

```hcl
resource azurerm_resource_group "rg" {
  name     = var.name
  location = var.location
}

resource "azurerm_windows_virtual_machine" "vm" {
  name                  = var.name
  resource_group_name   = azurerm_resource_group.rg.name
  ...
}
```

### Expected Behaviour

Terraform would ignore the fact that the RG name is uppercase, not replacing the VM.

### Actual Behaviour

Terraform wants to replace the VM, but it does not plan to replace the RG because that is already case-insensitive.
The result is a never-functional configuration

```
  # azurerm_resource_group.rg will be updated in-place
  ~ resource "azurerm_resource_group" "rg" {
        id       = "/subscriptions/xxx/resourceGroups/TESTRG"
        name     = "testrg"
      ~ tags     = {
          + "Owner"          = "me"
        }
        # (1 unchanged attribute hidden)
        # (1 unchanged block hidden)
    }

  # azurerm_windows_virtual_machine.vm must be replaced
  -/+ resource "azurerm_windows_virtual_machine" "vm" {
      ~ id                         = "/subscriptions/xxx/resourceGroups/TESTRG/providers/Microsoft.Compute/virtualMachines/testvm" -> (known after apply)
      ~ resource_group_name        = "TESTRG" -> "testrg" # forces replacement
      ~ tags                       = {
          - "Owner"          = "me"
        }
        ...
    }
```

### Steps to Reproduce

1. Manually create a Resource group with UPPERCASE name (I don't know how that happened but it did)
[...]

PR #10959, opened 2021-03-13
case-insensitive resource_group_name schema by default

AliAllomani:
As the resource group name is case-insensitive in Azure,

it might be a good idea to use `DiffSuppressFunc: suppress.CaseDifference` in the default `SchemaResourceGroupName` schema instead of explicitly using `SchemaResourceGroupNameDiffSuppress` schema for specific resources.

related to #10077

### DIFF
--- a/azurerm/helpers/azure/resource_group.go
+++ b/azurerm/helpers/azure/resource_group.go
@@ -11,10 +11,11 @@ import (
 
 func SchemaResourceGroupName() *schema.Schema {
 	return &schema.Schema{
-		Type:         schema.TypeString,
-		Required:     true,
-		ForceNew:     true,
-		ValidateFunc: validateResourceGroupName,
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
hey @AliAllomani 

Thanks for this PR.

Whilst Resource Group names are treated as case-insensitive when searching, there's certain API's (and any externally-configured integrations) which do require the casings to match - as such whilst this PR looks to make these case insensitive by default, we've noticed this can subtle issues when integrating with other API's.

From Terraform's side we're moving away from relying on the ID returned from the Azure API due to the casing issues identified in #10077, having chatted with the ARM Team - Resource ID's are supposed to be case-insensitive during Requests, but returned case-sensitively in responses - but unfortunately different service teams have interpreted this differently, meaning these are ultimately bugs in the API.

Since we're looking to workaround these API issues in a different manner, whilst I'd like to thank you for this contribution I'm going to close this PR for the moment - but we'll continue patching these API's up as we see them (and all new Resources are now using Terraform-managed ID's, rather than relying on this Azure API field which works around this going forwards).

Thanks!

PR #10971, opened 2021-03-15
Docs: fix valid expander

yteraoka:
`max-pods` is not valid expander.

max-pods -> most-pods

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders

https://github.com/terraform-providers/terraform-provider-azurerm/blob/1d934f7476d2b580c181955a041f28f3097f9ef1/azurerm/internal/services/containers/kubernetes_cluster_resource.go#L111-L121


tombuildsstuff (approved):
LGTM - thanks for this @yteraoka

PR #10972, opened 2021-03-15
upgrade aks api version

njuCZ:
this PR contains changes:
1. fix aks test case
1.1 for acctest `TestAccKubernetesCluster_diskEncryption`, it fails because purging key vault key failed. since using `azurerm_disk_encryption_set` needs to enable `purge_protection`, the fix way is to not purge key vault key when delete
2. upgrade api version to 2021-02-01

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_addons.go
+++ b/azurerm/internal/services/containers/kubernetes_addons.go
@@ -88,7 +88,6 @@ func schemaKubernetesAddOnProfiles() *schema.Schema {
 					Type:     schema.TypeList,
 					MaxItems: 1,
 					Optional: true,
-					Computed: true,
 					Elem: &schema.Resource{
 						Schema: map[string]*schema.Schema{
 							"enabled": {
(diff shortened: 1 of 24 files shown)


azurerm/internal/services/containers/kubernetes_addons.go
```diff
@@ -88,6 +88,7 @@ func schemaKubernetesAddOnProfiles() *schema.Schema {
 					Type:     schema.TypeList,
 					MaxItems: 1,
 					Optional: true,
+					Computed: true,
```
tombuildsstuff, line 91: we need to revert this - Azure Public is the only Environment where this is enabled, and that's being EOL in the near future

njuCZ: from the http response, it seems the value is default `false`. I think there is no harm after adding `Computed: true`?

tombuildsstuff: that's only true for certain Kubernetes versions in certain Azure Environments - this cannot be specified in other Environments, so there's an explicit difference between `false` and `null` - unfortunately this is the design of the API.

njuCZ: Got it. has reverted

tombuildsstuff (approved):
LGTM 👍

PR #10987, opened 2021-03-16
mysql: switching to use ID Formatters

tombuildsstuff:
This PR switches the `mysql` resources over to using ID Formatters, as a part of the ongoing refactor


jackofallops (approved):

PR #10992, opened 2021-03-17
scripts/gradually-deprecated: exiting with the right status code

tombuildsstuff:
Whilst the Azure API's _should_ be consistent insofar as they're supposed to follow Postel's Law (in that they're lossy in what they accept but strict in what they send) - in practice this isn't the case. To work around this we're moving towards using the (generated) Terraform-maintained ID Formatters/Parsers across the provider - ultimately allowing us to remove casing issues within the Azure API (which cause consistency issues within Terraform Core).

As such this PR adds a new "gradually deprecated" check to look for usages of `d.SetID(*` (as in `d.SetId(*resp.ID)`) in new changes, meaning that new PR's going forward will be using ID Formatters - back porting the ID Formatters to existing resources is a separate/longer job, but this prevents new resources from using the Azure Resource ID in the response.

### DIFF
--- a/.github/workflows/gradually-deprecated.yaml
+++ b/.github/workflows/gradually-deprecated.yaml
@@ -17,6 +17,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
+          fetch-depth: 0
           go-version: '1.16.0'
-      - run: git fetch origin --unshallow
       - run: ./scripts/run-gradually-deprecated.sh
--- a/.github/workflows/gradually-deprecated.yaml
+++ b/.github/workflows/gradually-deprecated.yaml
@@ -1,0 +1,22 @@
+---
+name: Check for new usages of deprecated functionality
+on:
+  pull_request:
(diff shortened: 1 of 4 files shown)


.github/workflows/gradually-deprecated.yaml
```diff
@@ -0,0 +1,21 @@
...
+  pull_request:
+    types: ['opened', 'synchronize']
+    paths:
+      - '**.go'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
```
jackofallops, line 13: I've learned that since we're not using matrix settings, we can remove this from our actions. It does nothing currently, so we can follow up to remove them all at once later.

jackofallops (approved):

.github/workflows/gradually-deprecated.yaml
tombuildsstuff: makes sense, but gonna leave that for another PR since this needs to be done everywhere at once

PR #11007, opened 2021-03-17
Test Fix: Mariadb

mbfrahry:
```
--- PASS: TestAccMariaDbServer_createReplica (834.93s)
--- PASS: TestAccMariaDbConfiguration_logSlowAdminStatements (265.00s)
```

### DIFF
--- a/azurerm/internal/services/mariadb/mariadb_configuration_resource_test.go
+++ b/azurerm/internal/services/mariadb/mariadb_configuration_resource_test.go
@@ -72,7 +72,7 @@ func TestAccMariaDbConfiguration_logSlowAdminStatements(t *testing.T) {
 		{
 			Config: r.logSlowAdminStatements(data),
 			Check: resource.ComposeTestCheckFunc(
-				data.CheckWithClient(checkValueIs("on")),
+				data.CheckWithClient(checkValueIs("On")),
 			),
 		},
 		data.ImportStep(),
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/mariadb/mariadb_configuration_resource_test.go
```diff
@@ -72,7 +72,7 @@ func TestAccMariaDbConfiguration_logSlowAdminStatements(t *testing.T) {
 		{
 			Config: r.logSlowAdminStatements(data),
 			Check: resource.ComposeTestCheckFunc(
-				data.CheckWithClient(checkValueIs("On")),
+				data.CheckWithClient(checkValueIs("on")),
```
katbyte, line 75: Should we change the config to `On` to ensure this API respects case instead?

azurerm/internal/services/mariadb/mariadb_server_resource_test.go
```diff
@@ -237,7 +237,6 @@ func TestAccMariaDbServer_createReplica(t *testing.T) {
 			Config: r.createReplica(data, version),
 			Check: resource.ComposeTestCheckFunc(
 				check.That(data.ResourceName).ExistsInAzure(r),
-				check.That("azurerm_mariadb_server").ExistsInAzure(r),
```
katbyte, line 240: Should we instead check that the replica exists? `check.That("azurerm_mariadb_server.replica").ExistsInAzure(r),`

katbyte (changes requested):
two comments left inline @mbfrahry

azurerm/internal/services/mariadb/mariadb_server_resource_test.go
mbfrahry: Ooo yes. Good call

katbyte (approved):
LGTM 👍

PR #11011, opened 2021-03-18
azurerm_container_registry - add quarantine policy property to container registry

ms-henglu:
Below paste the run pass result for the test cases

```
=== RUN   TestAccContainerRegistry_policies
=== PAUSE TestAccContainerRegistry_policies
=== CONT  TestAccContainerRegistry_policies
--- PASS: TestAccContainerRegistry_policies (294.62s)
PASS

```


azurerm/internal/services/containers/container_registry_resource.go
```diff
@@ -170,6 +170,23 @@ func resourceContainerRegistry() *schema.Resource {
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
+					},
+				},
+			},
```
tombuildsstuff, line 188: this doesn't make sense as a block - we can reduce this down to a boolean flag `quarantine_enabled` and achieve the same behaviour (erroring if this is enabled for a non-premium sku) - can we update this?

azurerm/internal/services/containers/container_registry_resource.go
```diff
@@ -804,6 +851,18 @@ func flattenNetworkRuleSet(networkRuleSet *containerregistry.NetworkRuleSet) []i
...
 }
 
+func flattenQuarantinePolicy(p *containerregistry.Policies) []interface{} {
+	if p == nil || p.QuarantinePolicy == nil {
+		return nil
+	}
+
+	t := *p.QuarantinePolicy
+	quarantinePolicy := make(map[string]interface{})
+	enabled := strings.EqualFold(string(t.Status), string(containerregistry.Enabled))
```
tombuildsstuff, line 861: this shouldn't be doing a case insensitive comparison - that would hide an API bug which needs to be fixed, so we can reduce this to:

```suggestion
	enabled := t.Status == containerregistry.Enabled
```

tombuildsstuff (changes requested):
hey @ms-henglu 

Thanks for this PR.

Taking a look through here whilst this looks to support this functionality, we can reduce this to a top-level boolean rather than having this as a nested block - as such can we update this?

Thanks!

azurerm/internal/services/containers/container_registry_resource.go
ms-henglu: I think it's better to keep consistent schema with API. For example the `trust_policy` below, it will add a property called `type` and this way is easier to add a new property without breaking changes.


refs:
https://docs.microsoft.com/en-us/azure/templates/microsoft.containerregistry/registries?tabs=json#TrustPolicy

tombuildsstuff: There's no need for this to be a nested block, so this can be condensed - API compatibility is helpful in some cases but in this instance doesn't make sense

ms-henglu: Okay, I'll update it. Thanks!

ms-henglu:
@tombuildsstuff I have added a commit, please review again. Thanks!

website/docs/r/container_registry.html.markdown
```diff
@@ -60,11 +60,13 @@ The following arguments are supported:
 
 * `network_rule_set` - (Optional) A `network_rule_set` block as documented below.
 
+* `quarantine_policy_enabled` - (Optional) Boolean value that indicates whether the policy is enabled. Defaults to `false`.
```
katbyte, line 63:
```suggestion
* `quarantine_policy_enabled` - (Optional) Boolean value that indicates whether quarantine policy is enabled. Defaults to `false`.
```

katbyte (approved):
aside from one comment this LGTM 👍

ms-henglu:
Thanks for code review!

ms-henglu:
It seems there's a conflict after https://github.com/terraform-providers/terraform-provider-azurerm/pull/10969 merged, I'll fix it.

ms-henglu:
@katbyte , I have fixed the conflicts, please check. Thanks!

PR #11012, opened 2021-03-18
make rg test names consistant

katbyte:


jackofallops (approved):
Thanks @katbyte - LGTM 👍

Issue #10500, opened 2021-02-08
azurerm_kubernetes_cluster - api_server_authorized_ip_ranges keeps re-applying


T1loc:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.14.5
hashicorp/azurerm v2.46.1

### Affected Resource(s)

* `azurerm_kubernetes_cluster`

### Terraform Configuration Files

```hcl
# Add api_server_authorized_ip_ranges without mask
api_server_authorized_ip_ranges = ['xx.xx.xx.xx', 'yy.yy.yy.yy']

```

### Expected Behaviour

No changes

### Actual Behaviour

Diff add /32 all the time

```hcl
 resource "azurerm_kubernetes_cluster" "main" {
      ~ api_server_authorized_ip_ranges = [
          + "xx.xx.xx.xx",
          - "xx.xx.xx.xx/32",
          + "yy.yy.yy.yy",
          - "yy.yy.yy.yy/32",
        ]
```

### Steps to Reproduce

1. `terraform apply`

### References

* https://github.com/terraform-providers/terraform-provider-azurerm/issues/8701

T1loc:
I'm not sure if that problem is related to the upstream bug.
They expect a ranges, I'm giving a range and the change remove the range mask.

:-/

katbyte:
Thank you for taking the time to report this issue, and apologies that it has sat without a resolution for so long.

This bug was reported against an early major version of the AzureRM provider. The provider has since moved on to v5.x, and the affected functionality has often been significantly reworked across those major releases, so bug reports from legacy major versions prior to v4.0 are generally no longer actionable as written.

As part of a cleanup of historical issues we are closing bug reports that target legacy provider versions. If you are still running into this behaviour on the current major version, please open a new issue with up-to-date reproduction details (feel free to reference this one) so it can be triaged with current context.

Thank you again for the report, and for your patience.

PR #11014, opened 2021-03-18
avoid diff for property "api_server_authorized_ip_ranges"

njuCZ:
fix issue #10500

image

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_cluster_auth_resource_test.go
+++ b/azurerm/internal/services/containers/kubernetes_cluster_auth_resource_test.go
@@ -429,7 +429,7 @@ resource "azurerm_kubernetes_cluster" "test" {
   }
 
   api_server_authorized_ip_ranges = [
-    "8.8.8.8/32",
+    "8.8.8.8",
     "8.8.4.4/32",
     "8.8.2.0/24",
   ]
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
hey @njuCZ 

Thanks for this PR.

Taking a look through here this appears to be an API bug rather than something which Terraform should support - the description for this field in the SDK Is:

> // AuthorizedIPRanges - Authorized IP Ranges to kubernetes API server.

Since this is expecting an IP Range rather than a single IP address, it's likely this could change in the future to restrict this to an IP Range instead - so whilst we could look to add support for these, since the field itself is specifically mentioning an IP Range rather than an IP and/or an IP Range for the moment we need to assume this is a range.

Would you be able to reach out to the service team to confirm if this behaviour is intentional or a bug so that we can work out how to proceed here?

Thanks!

katbyte:
@njuCZ - any updates here?

njuCZ:
hi @katbyte, it actually should be IP ranges from service team. So should we use more specific validation rule?

katbyte:
@njuCZ - it appears the service does accept single IP addresses, is this something they will fix/we shouldn't allow?

njuCZ:
sorry, still get no official reply from aks team. I will close it for now. If there is any further information, I will reopen it. Thanks

Issue #7034, opened 2020-05-21
Support for [thing]


dpbustos:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Add ability to ignore tags not set via the various `azurerm` resource blocks.

At the least, add the ability to be able to specify tag key's which ARE under management and to ignore changes on any other tag key that may be encountered. (This would be the opposite of the `ignore_tags` block functionality for the AWS provider, but having both an ignore and manage option like a white/black list working together might give the most flexibility.

### New or Affected Resource(s)

All resources with the `tags` property.

### Potential Terraform Configuration

*Option 1 - Add a boolean flag to each resource to be able to ignore changes to tags not deliberately set by the resource, e.g.:*

```hcl
resource "azurerm_resource_group" "example" {
  name     = "example"
  location = "West Europe"
  tags = {
    foo = "bar"
  }
  ignore_unmanaged_tags = true
}
```

*Option 2 - Add ability to add set of managed tag keys on the provider block (and to ignore changes made on any other keys), e.g.:*

```hcl
provider "azurerm" {
  features = {}

  managed_tags {
    key_prefixes = ["foo:"]
  }
}
```

### References

Related Issues:
- https://github.com/hashicorp/terraform/issues/24908
- https://github.com/hashicorp/terraform/issues/6632

* #0000

markslater:
Perhaps it could work like [this similar change](https://github.com/terraform-providers/terraform-provider-aws/issues/10689) that was implemented recently for the AWS provider?

As it stands, thanks to my organisation's decision to add several tags at the subscription level, every one of my Terraform resources has a lengthy list of fake values for tags, plus an equivalent list in the `ignore_changes`.  For many resources, the actual definition is swamped by the tag ignoring boilerplate, which is unfortunate (maybe there's a simpler way of doing this I've missed).

sennerholm:
A simular use case that would be good to support is the possibility to:

```
resource "azurerm_resource_group" "example" {
  name     = "example"
  location = "West Europe"
  tags = {
    foo = "bar"
  }
  lifecycle {
[...]

PR #11020, opened 2021-03-18
Ignore Tags Feature block

etiennejournet:
Hi !

Following the example provided by aws provider about ignore tags and requested [here ](https://github.com/terraform-providers/terraform-provider-azurerm/issues/7034) I'm submitting this Pull Request for validation.

This is essentially a work in progress as i wanted to get review/validation from maintainers before going ahead with the process of implementing the "IgnoreTags" function in every resources/datasources.
Also I'm not an expert in Go so I thought you guys would probably have ideas/ways for improvements in my code.

WIP :  (waiting for validation of the whole process)
1. Idempotency:
The function basically compares a list of keys (and key_prefixes) with the keys of the tagMap retrieved when reading the resource. This creates an idempotency problem when a tag is at the same time in "ignore_tags" features AND defined in a resource block. thought i could maybe use the CustomizeDiff hook for warning/erroring on this.
Another might be to duplicate the "HasChange" into an "HasChangeTag" method but not sure whether it's something you guys would accept or not.
2. Tests
3. Doc 

Thanks

### DIFF
--- a/azurerm/internal/features/user_flags.go
+++ b/azurerm/internal/features/user_flags.go
@@ -7,6 +7,7 @@ type UserFeatures struct {
 	Network                NetworkFeatures
 	TemplateDeployment     TemplateDeploymentFeatures
 	LogAnalyticsWorkspace  LogAnalyticsWorkspaceFeatures
+	IgnoreTags             IgnoreTagsFeatures
 }
 
 type VirtualMachineFeatures struct {
@@ -33,4 +34,9 @@ type TemplateDeploymentFeatures struct {
 
 type LogAnalyticsWorkspaceFeatures struct {
 	PermanentlyDeleteOnDestroy bool
+}
+
+type IgnoreTagsFeatures struct {
+	Keys        []string
+	KeyPrefixes []string
 }
--- a/azurerm/internal/provider/features.go
+++ b/azurerm/internal/provider/features.go
@@ -99,6 +99,31 @@ func schemaFeatures(supportLegacyTestSuite bool) *schema.Schema {
 				},
 			},
 		},
+
+		"ignore_tags": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			MaxItems:    1,
+			Description: "Configuration block with settings to ignore resource tags across all resources.",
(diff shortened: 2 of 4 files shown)


tombuildsstuff:
hey @etiennejournet 

Thanks for this PR.

Whilst the AWS and Azure Providers behave fairly consistently from a users perspective - the API's backing these two cloud providers are substantially different. The AWS API's expose one HTTP Method for each type of CRUD (Create-Read-Update-Delete) operation - that is a PUT for a Create and a PATCH for an Update, notably this allows only the delta of changes to be passed over the wire. By contrast ~95%+ of the Azure API's combine the Create and Update calls together into a single HTTP Method, where all fields have to be specified during both a Create and an Update.

Whilst this may sound like a subtle difference, it means within the Azure Provider that most resources don't handle `ignore_changes` on all fields at this time due to the nature of the Azure API (although this is something we're looking to address over time).

Where this is relevant for Tags is that today these values get pulled from the Terraform Configuration during an update at any point - so whilst we could potentially filter them out from the diff, with the way things work at this point in time Terraform would ultimately overwrite these ignored values during an update, which doesn't seem ideal.

Ultimately to be able to correctly support #7034 we'd need to first split all resources Create and Update functions to be able to support Delta updates, and then conditionally load in the ignored values from the Azure API during an update, to allow us to ensure we don't remove any tags which are ignored in the provider block.

As such whilst we appreciate this contribution - unfortunately removing these fields from the diff isn't sufficient given the way that Azure works (including when taking into account Azure Policy). From Terraform's side this is something we want to support, but for this to be usable given the constraints of the Azure API, we ultimately need to implement the larger changes described above to make that possible.

Since this is ultimately a considerably larger task affecting most resources in the codebase, whilst I'd like to thank you for this contribution I'm going to close this PR for the moment. That said, this is something we're working to enable over time - but since this will involve splitting the Create and Update methods within Terraform, which is going to be a longer-term task, rather than keeping this PR open indefinitely I'm going to close this in favour of #7034 for the moment, where once this is resolved we can take another look.

Thanks!

PR #11035, opened 2021-03-19
Upgrade golang to 1.16 and bring in devcontainer

LauraDamianTNA:
Updating the environment building tools to use golang 1.16

Devcontainer was copied from github.com/microsoft/vscode-dev-containers/ to keep golang versions in sync between GNUmakefile and devcontainer.json.

Will remove the devcontainer configuration from vscode-dev-containers when this is merged.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,69 @@
+FROM golang:1.16-buster
+
+# Avoid warnings by switching to noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
+
+# This Dockerfile adds a non-root 'vscode' user with sudo access. However, for Linux,
+# this user's GID/UID must match your local user UID/GID to avoid permission issues
+# with bind mounts. Update USER_UID / USER_GID if yours is not 1000. See
+# https://aka.ms/vscode-remote/containers/non-root-user for details.
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+# Terraform and tflint versions
+ARG TERRAFORM_VERSION=0.14.4
+
+ENV GO111MODULE=on
+
+# Configure apt, install packages and tools
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends curl unzip apt-utils dialog \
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
hi @LauraDamianTNA 


Thanks for this PR.

Whilst we could look to add integrations and support for specific IDE's - we've intentionally avoided doing so in favour of more generalized tooling which can be integrated into anything.

Where DevContainers are a feature only supported by VSCode (and not by other IDE's such as [Goland](https://www.jetbrains.com/go/) or [Chime](https://www.chimehq.com)) - since nobody on the team is using VSCode we're concerned that were we to add support for this, this'd fall out of sync pretty quickly and become a hinderance rather than a help. Notably when contributing on Windows there's also some additional challenges for users using WSL, WSL2 and Git Bash, where the setup can be quite different - whilst using Docker can help unfortunately it's not a panacea.

For what it's worth most of the team are using [Goland](https://www.jetbrains.com/go/) and we're also taking the same approach here, by [intentionally ignoring the `.idea` directory in the `.gitignore`](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/.gitignore#L24) (we're also [doing the same for the `.vscode` folder for the same reason](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/.gitignore#L31)).

As mentioned above since we're trying to use more generalized tooling where possible, in this instance we're using a Makefile (and optionally Docker) - which should work cross-platform (via `make tools` to get the tools, `make build` or `make docker-build` to build locally/in docker and `make fmt` to run the formatters) - and documenting these commands in the README.

We do have plans to add more specific documentation here (partly tracked in #2688 - and others via some internal things) - since we're not planning on adding integrations with specific IDE's whilst I'd like to thank you for this contribution, I'm going to close this PR for the moment.

Thanks!

LauraDamianTNA:
Thanks for the detailed explanation. I have created a separate PR in the github.com/microsoft/vscode-dev-containers and one in the current project to allow me to use Docker.
 I added a thumb up for the documentation as well.

PR #11057, opened 2021-03-20
Fix for azurerm_resource_group_template_deployment example

Lachlan-White:
Documentation makes reference to a non-existent variable, when it should be a parameter


tombuildsstuff (approved):
LGTM - thanks for this @Lachlan-White

PR #11060, opened 2021-03-21
azurerm_kubernetes_cluster: Support for max_empty_bulk_delete in auto_scaler_profile block

liammoat:
This PR adds support for the `max_empty_bulk_delete` property in the `auto_scaler_profile` block for `azurerm_kubernetes_cluster`.

As per https://docs.microsoft.com/en-us/azure/aks/cluster-autoscaler#using-the-autoscaler-profile

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_cluster_resource.go
+++ b/azurerm/internal/services/containers/kubernetes_cluster_resource.go
@@ -171,7 +171,7 @@ func resourceKubernetesCluster() *schema.Resource {
 							Optional: true,
 							Computed: true,
 						},
-						"max_empty_bulk_delete": {
+						"empty_bulk_delete_max": {
 							Type:     schema.TypeString,
 							Optional: true,
 							Computed: true,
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/containers/kubernetes_cluster_resource.go
```diff
@@ -171,6 +171,11 @@ func resourceKubernetesCluster() *schema.Resource {
 							Optional: true,
 							Computed: true,
 						},
+						"max_empty_bulk_delete": {
```
katbyte, line 174: could we call this `empty_bulk_delete_max`?

katbyte (changes requested):
Aside from one comment this looks good @liammoat - once address this should be good to merge!

azurerm/internal/services/containers/kubernetes_cluster_resource.go
liammoat: @katbyte thanks! Updated as per your suggestion

njuCZ:
@liammoat the tflint reports error because of test file is not formatted, yuo could install this tool https://github.com/katbyte/terrafmt and execute `terrafmt fmt -f path/to/file`.

katbyte (approved):
Thanks @liammoat - LGTM 👍

PR #11064, opened 2021-03-22
`azurerm_spring_cloud_app` - supports property `tls_enabled`

njuCZ:
upgrade api version to `2020-11-01-preview`
support new property: `tls_enabled`

all test in spring cloud RP could pass. Following snippet is a part of the result
image


katbyte (approved):
LGTM 👍

PR #11074, opened 2021-03-22
fix Route Table routes to be Set instead of List

thllxb:
this commit fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/7489

### DIFF
--- a/azurerm/internal/services/network/route_table_resource.go
+++ b/azurerm/internal/services/network/route_table_resource.go
@@ -54,7 +54,7 @@ func resourceRouteTable() *schema.Resource {
 			"resource_group_name": azure.SchemaResourceGroupName(),
 
 			"route": {
-				Type:       schema.TypeList,
+				Type:       schema.TypeSet,
 				ConfigMode: schema.SchemaConfigModeAttr,
 				Optional:   true,
 				Computed:   true,


thllxb:
@tombuildsstuff I am not sure why this would be a breaking change. I have tested the fix with resources created with v2.x and didn't break anything for create, update, and delete.
Can you explain a bit here?

katbyte:
@thllxb - given it seems that azure orders them in a predictable way? i think what we should do is sort them so after 1 diff whats in state and returned from azure will match. This will also not be a breaking change as people can still refer to routes (even if it might be in an unexpected order)

WDYT?

thllxb:
> @thllxb - given it seems that azure orders them in a predictable way? i think what we should do is sort them so after 1 diff whats in state and returned from azure will match. This will also not be a breaking change as people can still refer to routes (even if it might be in an unexpected order)
> 
> WDYT?

@katbyte  Indeed, the list returned from azure is sorted in a particular, but not obvious, order. Matching the sort method would a possible way to fix it if we can find it. However, it would still be a guess which might fail once the sample size gets larger, unless Azure can confirm it and promise not to change it.
Another way is to customize the compare logic to first sort both the local list and returned list in a particular order. I would be happy to see if it gets implemented.

katbyte:
@thllxb - if we can sort this in a predictable way i think we can move forward without a breaking change (as actual order doesn't matter) and merge it in - but until we can, changing from a list to a set is a breaking change for existing tf configs that reference it and we cannot accept breaking changes till 3.0 without a really compelling reason. So unless there is a non breaking change path forward for < 3.0 i'll mark it for that major release and close it till we start work on it (which fwiw is 3-6 months away)

katbyte:
Hey @thllxb - as this has sat dormant for a number of months now i am going to mark it 3.0 and close it - we'll revisit this when we are closer to that that major release, thanks!

PR #11083, opened 2021-03-23
Upgrade hpc cache API version to 2021-03-01

magodo:
During the update, there are some breaking changes involved in turns of API model and behavior, whilst there is no actual change happens on the existing resources. Let me explain these one by one.

## Breaking Changes in API Model and Behavior

In previous version (2020-03-01), the cache's [`securitySettings`](https://github.com/Azure/azure-rest-api-specs/blob/e05b5a2b68745a5850ba64c04e158483aa4d626e/specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2020-03-01/storagecache.json#L1407-L1407) property only contains one sub-property: `rootSquash`.

image

Whilst in this latest version, the cache's [`securitySettings`](https://github.com/Azure/azure-rest-api-specs/blob/e05b5a2b68745a5850ba64c04e158483aa4d626e/specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2021-03-01/storagecache.json#L1419-L1419) contains an array of `NfsAccessPolicy` models, which in turns contain a `name` property and an array of `NfsAccessRule`, which define different access level for up to three scopes (`default`, `network`, `host`).

image

Note that in the new version, the `NfsAccessRule` also contains a property called `rootSquash`.

## Semantic Changes

In previous version, when the `rootSqush` is set, it enables root squash for all the HPC cache target junctions that are mounted to this cache.

In new version, each cache will have a default access policy with name `default`, which users can't delete. Every newly created HPC cache target junction, by default use this `default` policy. However, users can also create customized policies, and assign those policy to cache target junctions during creation or update afterwards.

Under the covers every junction in the cache is associated with an access policy, no matter old or new API. In the old API, there were two predefined access policies, named `policySquash` and `policyNoSquash`. All the junctions would be assigned one of those policies, based on the global `rootSquash` setting. The existence of these policies was hidden under the old API version. In the new API version, the definitions of the policies become visible at the cache level, and which policy is assigned to a junction is visible at the storage target level. The `policySquash` and `policyNoSquash` policies are automatically created when using the old api to create a cache or change the global squash setting.

Regarding the compatibility factor: caches created under the old API version can be examined with the new api and you’ll see the predefined `policySquash`, `policyNoSquash`, and `default` policies in the cache. The junctions will **all** be assigned to `policySquash` or `policyNoSquash`. You can modify individual junctions and create/modify/delete policies using the new api version. . If you change the global `rootSquash` value using the old api, all the junctions will be updated to the corresponding `policySquash`/`policyNoSquash` (not that in new API version, it only create a policy but doesn't change the junction's setting). One major issue occurs when you look at a cache created with the new api using the old api. In that case, the policies on individual junctions won’t be visible, and the global `rootSquash` value will be the last value set using the old api version, regardless of whether all the junctions are still using the corresponding `policySquash`/`policyNoSquash`.

## Implementation Details

The goal here is to seek for backward compatibility in best effort.

The first change is to make the `root_squash_enabled` non-functional and make it as `Computed` so that existing config will not see diff. In the meanwhile, I'm adding new property `access_policy_name` to `azurerm_hpc_cache_blob_target`, and also set it as `O+C` so that existing target will have either `policyNoSquash` or `policySquash` returned. However, I didn't do the same thing for `azurerm_hpc_cache_nfs_target` as the junction block is inside a `TypeSet` and it can't work by simply setting the `access_policy_name` as `O+C` (which will cause the whole Set's hash change). I guess we can further find some other solution to workaround this, e.g. making the junction containing block to be List instead.

Another change is that I split the `securitySettings` of cache into `default_access_policy` inside resource `azurerm_hpc_cache`, and create separate resource `azurerm_hpc_cache_access_policy` to allow useres to create custom policies. If we do not do this split, imagin users have a junction associate with a policy `p1`, then wants to create a new one `p2` and use it, also delete the `p1` (as no one is using it). This will fail if users edit the policy in the `azurerm_hpc_cache` and `azurerm_hpc_cache_blob_target`, then apply. Because the apply order will be first update the cache (this will fail, complaining the `p1` is still under used) and then update the blob target. While if we have a separate `azurerm_hpc_cache_access_policy` resource, we can add `lifecycle = {create_before_destroy=true}` on it to revert the apply order.

## Test Result

```bash
💤 make testacc TEST=./azurerm/internal/services/hpccache TESTARGS='-run TestAccHPCCacheAccessPolicy_'
TF_ACC=1 go test ./azurerm/internal/services/hpccache -v -run TestAccHPCCacheAccessPolicy_ -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
2021/03/23 12:52:29 [DEBUG] not using binary driver name, it's no longer needed
2021/03/23 12:52:29 [DEBUG] not using binary driver name, it's no longer needed
=== RUN   TestAccHPCCacheAccessPolicy_basic
=== PAUSE TestAccHPCCacheAccessPolicy_basic
=== RUN   TestAccHPCCacheAccessPolicy_complete
=== PAUSE TestAccHPCCacheAccessPolicy_complete
=== RUN   TestAccHPCCacheAccessPolicy_update
=== PAUSE TestAccHPCCacheAccessPolicy_update
=== RUN   TestAccHPCCacheAccessPolicy_requiresImport
=== PAUSE TestAccHPCCacheAccessPolicy_requiresImport
=== CONT  TestAccHPCCacheAccessPolicy_basic
=== CONT  TestAccHPCCacheAccessPolicy_requiresImport
=== CONT  TestAccHPCCacheAccessPolicy_update
=== CONT  TestAccHPCCacheAccessPolicy_complete

--- PASS: TestAccHPCCacheAccessPolicy_requiresImport (1642.84s)
--- PASS: TestAccHPCCacheAccessPolicy_complete (1763.22s)
--- PASS: TestAccHPCCacheAccessPolicy_basic (1776.94s)
--- PASS: TestAccHPCCacheAccessPolicy_update (2191.30s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/hpccache    2192.066s


💢 make testacc TEST=./azurerm/internal/services/hpccache TESTARGS='-run TestAccHPCCacheBlobTarget_accessPolicy'
TF_ACC=1 go test ./azurerm/internal/services/hpccache -v -run TestAccHPCCacheBlobTarget_accessPolicy -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
2021/03/23 13:12:45 [DEBUG] not using binary driver name, it's no longer needed
2021/03/23 13:12:45 [DEBUG] not using binary driver name, it's no longer needed
=== RUN   TestAccHPCCacheBlobTarget_accessPolicy
=== PAUSE TestAccHPCCacheBlobTarget_accessPolicy
=== CONT  TestAccHPCCacheBlobTarget_accessPolicy
--- PASS: TestAccHPCCacheBlobTarget_accessPolicy (2982.35s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/hpccache    2982.536s

💢 make testacc TEST=./azurerm/internal/services/hpccache TESTARGS='-run "TestAccHPCCache_defaultAccessPolicy|TestAccHPCCache_rootSquashDeprecated"'
TF_ACC=1 go test ./azurerm/internal/services/hpccache -v -run "TestAccHPCCache_defaultAccessPolicy|TestAccHPCCache_rootSquashDeprecated" -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
2021/03/23 16:17:57 [DEBUG] not using binary driver name, it's no longer needed
2021/03/23 16:17:57 [DEBUG] not using binary driver name, it's no longer needed
=== RUN   TestAccHPCCache_rootSquashDeprecated
=== PAUSE TestAccHPCCache_rootSquashDeprecated
=== RUN   TestAccHPCCache_defaultAccessPolicy
=== PAUSE TestAccHPCCache_defaultAccessPolicy
=== CONT  TestAccHPCCache_rootSquashDeprecated
=== CONT  TestAccHPCCache_defaultAccessPolicy
--- PASS: TestAccHPCCache_rootSquashDeprecated (1382.09s)
--- PASS: TestAccHPCCache_defaultAccessPolicy (2144.48s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/hpccache    2144.611s
```


katbyte (changes requested):
have some test failures:
```

TestAccHPCCacheBlobTarget_accessPolicy
--
TestAccHPCCacheBlobTarget_update
TestAccHPCCacheBlobTarget_basic


```

azurerm/internal/services/hpccache/hpc_cache_access_policy_resource_test.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+func (r HPCCacheAccessPolicyResource) template(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
+
+resource "azurerm_resource_group" "test" {
+  name     = "acctestRG-storage-%d"
```
katbyte, line 202:
```suggestion
  name     = "acctestRG-hpcc-%d"
```

PR #11106, opened 2021-03-25
"azurerm_spring_cloud_service" - supports property `sample_rate`

njuCZ:
supports property `sample_rate`.

if not set this property, the default value from backend is 10.0.

image


katbyte (approved):
LGTM 👍

PR #11112, opened 2021-03-25
Documentation update for `monitor_scheduled_query_rules_alert`

ilchebedelovski:
The `action` block should be started with the name of the block, the `action` block name should not be part of the bullets because it is the name of the block.


mbfrahry (approved):
LGTM!

PR #11135, opened 2021-03-28
azurerm_security_center_workspace - Allow free SKU

MattiasAng:
This will allow using resource without having to set security center subscription pricing to standard.
Moved combined tests from security_center_test.go in to their seperate resource tests.

Had to add case insensitivity DiffSuppressFunc on workspace_id due to API reporting back lowercase workspace id, example:

````
PUT /subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Security/workspaceSettings/default?api-version=2017-08-01-preview HTTP/1.1
Host: management.azure.com
Authorization: Bearer redacted
Content-Type: application/json
Content-Length: 471

{
    "id": "/subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Security/workspaceSettings/default",
    "name": "default",
    "type": "Microsoft.Security/workspaceSettings",
    "properties": {
        "workspaceId": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/Testing-RG/providers/Microsoft.OperationalInsights/workspaces/Test-LA",
        "scope": "/subscriptions/00000000-0000-0000-0000-000000000000"
    }
}
````
And response when doing a GET:
````
GET /subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Security/workspaceSettings?api-version=2017-08-01-preview HTTP/1.1
Host: management.azure.com
Authorization: Bearer redacted

Response:
{
    "value": [
        {
            "id": "/subscriptions/00000000-0000-0000-0000-000000000000c/providers/Microsoft.Security/workspaceSettings/default",
            "name": "default",
            "type": "Microsoft.Security/workspaceSettings",
            "properties": {
                "workspaceId": "/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/testing-rg/providers/microsoft.operationalinsights/workspaces/test-la",
                "scope": "/subscriptions/00000000-0000-0000-0000-000000000000"
            }
        }
    ]
}
````

### CODE REVIEW

File: azurerm/internal/services/securitycenter/security_center_workspace_resource.go [hunk 1]
```diff
@@ -49,20 +49,19 @@ func resourceSecurityCenterWorkspace() *schema.Resource {
 			},
 
 			"workspace_id": {
-				Type:         schema.TypeString,
-				Required:     true,
-				ValidateFunc: azure.ValidateResourceID,
+				Type:             schema.TypeString,
+				Required:         true,
+				ValidateFunc:     validate.LogAnalyticsWorkspaceID,
+				DiffSuppressFunc: suppress.CaseDifference,
```
katbyte, line 55: whats the reason for making the case insensitive?

katbyte (changes requested):
Thanks for the PR @MattiasAng - i'm wondering why you made the tests run at the same time instead of in sequence? i'm getting test failures when running these tests

azurerm/internal/services/securitycenter/security_center_workspace_resource.go
MattiasAng: As said in PR description, Azure API will always return lower case workspace id, regardless of creation input. So without suppression it will always find differences. Current tests in master are failing, so this must have been an issue for a while.

If you have a better solution I am open to it.

MattiasAng:
> Thanks for the PR @MattiasAng - i'm wondering why you made the tests run at the same time instead of in sequence? i'm getting test failures when running these tests

Ok, so I am somewhat new to Golang, but from what I can see in master they are running in parallel in master but pricing test runs first (which is not required after this), so I am seeing the same failures there. 
Anyway, yes, they need to run in sequence like you said, which is my mistake for not testing all at once.

I have comitted a combined test that makes them run in sequence, but I am not sure if that is the best way to write it so feel free to give me feedback or PR to my branch.

Unfortunately this API seems to have issues with new Log Analytics Workspaces (existing ones are instant) so tests might timeout from time to time due to missing workspaceId attribute.

manicminer:
@MattiasAng I've been looking at this today but the API is omitting the workspaceId property every time for me, even after the resource is deleted and recreated 😔

This may block this PR, even though the tests in master suffer the same issue.

manicminer:
Raised upstream as https://github.com/Azure/azure-rest-api-specs/issues/14138

MattiasAng:
> @MattiasAng I've been looking at this today but the API is omitting the workspaceId property every time for me, even after the resource is deleted and recreated 😔
> 
> This may unfortunately block this PR, even though the tests in master suffer the same issue.

Yeah, seems like it has gotten even worse. A couple of weeks ago I could successfully test it 20% of the time before timing out.
Thanks for creating upstream issue.

katbyte:
Hey @MattiasAng - given it seems this is blocked for now do you mind if i close it out until the upstream issue is fixed?

MattiasAng:
@katbyte Sure, no problem.

PR #11139, opened 2021-03-28
azurerm_api_management: make certificate_password optional

MattiasAng:
Couldn't find an issue for this when looking, but my guess is that a lot (including me) uses a workaround and setting password to empty string when importing certificates that doesn't have a password.

Certificate added for testdata is just a base64 encoded x509 from the pfx that is used.


favoretti (approved):
Hi @MattiasAng thank you for the PR. LGTM 👍🏻

katbyte (approved):
LGTM too! 👍

PR #11145, opened 2021-03-29
`azurerm_spring_cloud_custom_domain`: `thumbprint` is required when specifying `certificate_name`

favoretti:
`azurerm_spring_cloud_custom_domain`: `thumbprint` is required when specifying `certificate_name`

### DIFF
--- a/azurerm/internal/services/springcloud/spring_cloud_custom_domain_resource.go
+++ b/azurerm/internal/services/springcloud/spring_cloud_custom_domain_resource.go
@@ -62,6 +62,7 @@ func resourceSpringCloudCustomDomain() *schema.Resource {
 				Optional:     true,
 				ForceNew:     true,
 				ValidateFunc: validation.StringIsNotEmpty,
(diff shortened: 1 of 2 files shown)


mbfrahry (commented):
Hey @favoretti! This change looks good but I'm wondering if we need that `RequiredWith` on `certificate_name` as well? Or can we specify `certificate_name` without a `thumbprint`?

favoretti:
@mbfrahry i thought that's exactly what I was addressing. We can't have certificate_name without a thumbprint, although I haven't checked other way around. I'll try to test today.

favoretti:
@mbfrahry You're right. Funny thing - API doesn't give any errors if one of the 2 is specified, it just never adds a certificate to a custom domain. So marking the combo as required with each other.

mbfrahry (approved):
LGTM!

mbfrahry:
That is weird for sure but luckily these changes should help! Thanks again @favoretti!

PR #11148, opened 2021-03-29
Implementation of TDE for MS SQL Server

### LINKED ISSUE #87: SQL Server: support for Transparent Data Encryption

MoAhuja:
Fixes #87 

TF_ACC=1 go test -v ./azurerm/internal/services/mssql -run=TestAccMsSqlServerTransparentDataEncryption -timeout 60m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
2021/03/29 14:47:59 [DEBUG] not using binary driver name, it's no longer needed
2021/03/29 14:47:59 [DEBUG] not using binary driver name, it's no longer needed
=== RUN   TestAccMsSqlServerTransparentDataEncryption_byok
=== PAUSE TestAccMsSqlServerTransparentDataEncryption_byok
=== RUN   TestAccMsSqlServerTransparentDataEncryption_systemManaged
=== PAUSE TestAccMsSqlServerTransparentDataEncryption_systemManaged
=== RUN   TestAccMsSqlServerTransparentDataEncryption_update
=== PAUSE TestAccMsSqlServerTransparentDataEncryption_update
=== CONT  TestAccMsSqlServerTransparentDataEncryption_byok
=== CONT  TestAccMsSqlServerTransparentDataEncryption_update
=== CONT  TestAccMsSqlServerTransparentDataEncryption_systemManaged
2021/03/29 14:48:04 [DEBUG] TF_ACCTEST_REATTACH set to 1 and acctest.TestHelper is not nil, using reattach-based testing
--- PASS: TestAccMsSqlServerTransparentDataEncryption_systemManaged (246.96s)
--- PASS: TestAccMsSqlServerTransparentDataEncryption_byok (491.30s)
--- PASS: TestAccMsSqlServerTransparentDataEncryption_update (523.55s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/mssql       523.850s


azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,213 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"server_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
```
jackofallops, line 50: Could we use the server ID here in place of the name and resource group?  This can then be parsed / validated using the service packages?

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,213 @@
...
+	encryptionProtectorClient := meta.(*clients.Client).MSSQL.EncryptionProtectorClient
+	serverKeysClient := meta.(*clients.Client).MSSQL.ServerKeysClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	log.Printf("[INFO] preparing arguments for MSSQL Encrypted Protector creation.")
+
+	// encryptedProtectorName := d.Get("name").(string)
+	serverName := d.Get("server_name").(string)
+	resGroup := d.Get("resource_group_name").(string)
```
jackofallops, line 70: Using the ID means we could have:
```suggestion
      serverId, err := parse.ServerId(d.Get("server_id").(string))
      if err != nil {
        return err
      }
```

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,213 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"server_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
+			"key_vault_uri": {
```
jackofallops, line 51: I think this should be `key_vault_key_id`?

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,213 @@
...
+	if err = futureEncryptionProtector.WaitForCompletionRef(ctx, encryptionProtectorClient.Client); err != nil {
+		return fmt.Errorf("waiting on create/update future for Encryption Protector on server %q (Resource Group %q): %+v", serverName, resGroup, err)
+	}
+
+	resp, err := encryptionProtectorClient.Get(ctx, resGroup, serverName)
+	if err != nil {
+		return fmt.Errorf("issuing get request for Encryption protector on server %q (Resource Group %q): %+v", serverName, resGroup, err)
+	}
+
+	d.SetId(*resp.ID)
```
jackofallops, line 167: Can we make use of the generated ID functions here?  You should have enough information to use 
```
id := NewEncryptionProtectorID(serverId.Subscription, serverId.ResourceGroup, serverId.Name, name)
```

```suggestion
	d.SetId(id.ID())
```

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	data.ResourceTest(t, r, []resource.TestStep{
+		{
+			Config: r.byok(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+				check.That(data.ResourceName).Key("key_vault_uri").Exists(),
+			),
+		},
+		data.ImportStep(""),
```
jackofallops, line 32: This can be simplified a little
```suggestion
			),
		},
		data.ImportStep(),
```

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	data.ResourceTest(t, r, []resource.TestStep{
+		{
+			Config: r.systemManaged(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+				check.That(data.ResourceName).Key("key_vault_uri").DoesNotExist(),
+			),
+		},
+		data.ImportStep(""),
```
jackofallops, line 48: and here
```suggestion
			),
		},
		data.ImportStep(),
```

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource_test.go
```diff
@@ -0,0 +1,206 @@
...
+	// Test going from systemManaged to byok and back
+	data.ResourceTest(t, r, []resource.TestStep{
+		{
+			Config: r.byok(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+				check.That(data.ResourceName).Key("key_vault_uri").Exists(),
+			),
+		},
+		data.ImportStep(""),
```
jackofallops, line 65: and here
```suggestion
			),
		},
		data.ImportStep(),
```

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource_test.go
```diff
@@ -0,0 +1,206 @@
...
+		{
+			Config: r.systemManaged(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+				check.That(data.ResourceName).Key("key_vault_uri").IsEmpty(),
+			),
+		},
+
+		data.ImportStep(""),
```
jackofallops, line 74: and here
```suggestion
			),
		},
		data.ImportStep(),
```

jackofallops (changes requested):
Hi @MoAhuja - Thanks for this, it's off to a good start.  I've left some comments below that I think need to be addressed before continuing to review.  If you can fix those up, I'll continue asap.
Thanks!

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
MoAhuja: Agreed and changed.

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
MoAhuja: Updated to use the generated server id

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
MoAhuja: Updated

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
MoAhuja: Updated

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource_test.go
MoAhuja: Updated

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource_test.go
MoAhuja: Updated

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource_test.go
MoAhuja: Updated

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource_test.go
MoAhuja: Updated

MoAhuja:
@jackofallops thanks for the review! I've made the changes you suggested and it's ready for another review. Thanks.

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// Only set the key type if it's an AKV key. For service managed, we can omit the setting the key_vault_key_id
+	if resp.EncryptionProtectorProperties.ServerKeyType == sql.AzureKeyVault {
+		log.Printf("[INFO] Setting Key Vault URI to %s", *resp.EncryptionProtectorProperties.URI)
+
+		if err := d.Set("key_vault_key_id", resp.EncryptionProtectorProperties.URI); err != nil {
+			return fmt.Errorf("setting key_vault_key_id`: %+v", err)
+		}
+	}
+
```
jackofallops, line 205: I suspect we'll need to set this in all cases to correctly identify diffs, and we'll need to nil check the `EncryptionProtectorProperties` to make sure we don't get a panic if it's missing for any reason.
```suggestion

	keyVaultKeyId := ""
	if resp.EncryptionProtectorProperties != nil && resp.EncryptionProtectorProperties.ServerKeyType  == sql.AzureKeyVault {
		if resp.EncryptionProtectorProperties.URL != nil {
			log.Printf("[INFO] Setting Key Vault URI to %s", *resp.EncryptionProtectorProperties.URI)
			keyVaultKeyId = *resp.EncryptionProtectorProperties.URI
		}
	}
	
	if err := d.Set("key_vault_key_id", keyVaultKeyId); err != nil {
		return fmt.Errorf("setting key_vault_key_id`: %+v", err)
	}

```

jackofallops (changes requested):
Hi @MoAhuja - Thanks for the quick turn-around on the changes.  There's a couple more below, in particular the behaviour of the resource Delete. Again, I'll loop back as quickly as I can once these are addressed.

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
MoAhuja: @jackofallops on this one, does this now mean that key_vault_key_id would be empty string (that can be the default value) when the user wants to specify Service Managed and it would have a valid key vault key id when they want to use AKV? If so, I'm guessing that also means I'd have to remove the validationFunc that ensures key_vault_key_id is a valid KeyVault NestedItemId?

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,213 @@
...
+		keyDetails, err := keyVaultParser.ParseNestedItemID(keyUri)
+
+		if err != nil {
+			return fmt.Errorf("Unable to parse key: %q: %+v", keyUri, err)
+		}
+
+		// Make sure it's a key, if not, throw an error
+		if keyDetails.NestedItemType == "keys" {
+			keyName := keyDetails.Name
+			keyVersion := keyDetails.Version
```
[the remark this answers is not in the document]
MoAhuja: Updated. Will be fixed in next push.

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,213 @@
...
+
+		// Set the SQL Server Key properties
+		serverKeyProperties := sql.ServerKeyProperties{
+			ServerKeyType: sql.AzureKeyVault,
+			URI:           &keyUri,
+		}
+		serverKey.ServerKeyProperties = &serverKeyProperties
+
+		// Set the encryption protector properties
+		keyDetails, err := keyVaultParser.ParseNestedItemID(keyUri)
```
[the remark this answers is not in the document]
MoAhuja: Updated. Will be fixed in next push.

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource_test.go
```diff
@@ -0,0 +1,202 @@
...
+func TestAccMsSqlServerTransparentDataEncryption_byok(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_mssql_server_transparent_data_encryption", "test")
+	r := MsSqlServerTransparentDataEncryptionResource{}
+
+	data.ResourceTest(t, r, []resource.TestStep{
+		{
+			Config: r.byok(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+				check.That(data.ResourceName).Key("key_vault_key_id").Exists(),
```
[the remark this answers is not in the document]
MoAhuja: Updated

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	return nil
+}
+
+func resourceMsSqlTransparentDataEncryptionDelete(d *schema.ResourceData, meta interface{}) error {
+	// Note that encryption protector cannot be deleted. It can only be updated between AzureKeyVault
+	// and SystemManaged.
+	return nil
+}
```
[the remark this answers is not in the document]
MoAhuja: Latest commit has this change in place

MoAhuja:
@jackofallops Just an FYI, I rebased on top of the latest changes from master as there were some conflicts in the mssql clients.go file.

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,244 @@
...
+		ServerKeyType: serverKeyType,
+		ServerKeyName: &serverKeyName,
+	}
+
+	// Only create a server key if the properties have been set
+	if serverKey.ServerKeyProperties != nil {
+		// Create a key on the server
+		futureServers, err := serverKeysClient.CreateOrUpdate(ctx, serverId.ResourceGroup, serverId.Name, serverKeyName, serverKey)
+		if err != nil {
+			return err
```
manicminer, line 140: We should wrap this error

```suggestion
			return fmt.Errorf("creating/updating server key for %s: %+v", serverId, err)
```

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+
+		if err = futureServers.WaitForCompletionRef(ctx, serverKeysClient.Client); err != nil {
+			return err
```
manicminer, line 144: This one too

```suggestion
			return fmt.Errorf("waiting on update of %s: %+v", serverId, err)
```

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err = futureEncryptionProtector.WaitForCompletionRef(ctx, encryptionProtectorClient.Client); err != nil {
+		return fmt.Errorf("waiting on create/update future for Encryption Protector on server %q (Resource Group %q): %+v", serverId.Name, serverId.ResourceGroup, err)
+	}
+
+	// Encryption protector always uses "current" for the name
+	id := parse.NewEncryptionProtectorID(serverId.SubscriptionId, serverId.ResourceGroup, serverId.Name, "current")
```
manicminer, line 162: Also here, and we can use the string representation of the ID for convenience/consistency :)

```suggestion
	// Encryption protector always uses "current" for the name
	id := parse.NewEncryptionProtectorID(serverId.SubscriptionId, serverId.ResourceGroup, serverId.Name, "current")

	futureEncryptionProtector, err := encryptionProtectorClient.CreateOrUpdate(ctx, id.ResourceGroup, id.ServerName, encryptionProtectorObject)
	if err != nil {
		return fmt.Errorf("creating/updating %s: %+v", id, err)
	}

	if err = futureEncryptionProtector.WaitForCompletionRef(ctx, encryptionProtectorClient.Client); err != nil {
		return fmt.Errorf("waiting on create/update future for %s: %+v", id, err)
	}

```

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,244 @@
...
+		return err
+	}
+
+	resp, err := encryptionProtectorClient.Get(ctx, id.ResourceGroup, id.ServerName)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error making Read request for Encryption Protector on server %s: %v", id.ServerName, err)
```
manicminer, line 185: Same here, we can use the id string representation

```suggestion
		return fmt.Errorf("making Read request for %s: %v", id, err)
```

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,244 @@
...
+func resourceMsSqlTransparentDataEncryptionDelete(d *schema.ResourceData, meta interface{}) error {
+	// Note that encryption protector cannot be deleted. It can only be updated between AzureKeyVault
+	// and SystemManaged. For safety, when this resource is deleted, we're resetting the key type
+	// to service managed to prevent accidental lockout if someone were to delete the keys from key vault
+
+	encryptionProtectorClient := meta.(*clients.Client).MSSQL.EncryptionProtectorClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	serverId, err := parse.ServerID(d.Get("server_id").(string))
```
manicminer, line 218: Although using a ServerID works here, could we use an EncryptionProtectorID instead?

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
```diff
@@ -0,0 +1,244 @@
...
+		},
+	}
+
+	futureEncryptionProtector, err := encryptionProtectorClient.CreateOrUpdate(ctx, serverId.ResourceGroup, serverId.Name, encryptionProtector)
+	if err != nil {
+		return err
+	}
+
+	if err = futureEncryptionProtector.WaitForCompletionRef(ctx, encryptionProtectorClient.Client); err != nil {
+		return fmt.Errorf("waiting on create/update future for Encryption Protector on server %q (Resource Group %q): %+v", serverId.Name, serverId.ResourceGroup, err)
```
manicminer, line 240: (assuming we swap the ServerID for an EncryptionProtectorID)

```suggestion
		return fmt.Errorf("waiting on create/update future for %s: %+v", id, err)
```

website/docs/r/mssql_server_transparent_data_encryption.html.markdown
```diff
@@ -0,0 +1,178 @@
...
+
+## Timeouts
+
+The `timeouts` block allows you to specify [timeouts](https://www.terraform.io/docs/configuration/resources.html#timeouts) for certain actions:
+
+* `create` - (Defaults to 30 minutes) Used when creating the MSSQL.
+* `read` - (Defaults to 5 minutes) Used when retrieving the MSSQL.
+* `update` - (Defaults to 30 minutes) Used when updating the MSSQL.
+* `delete` - (Defaults to 30 minutes) Used when deleting the MSSQL.
+
```
manicminer, line 178: Can we add an Import section to the end here?

````suggestion

## Import

SQL Server Transparent Data Encryption can be imported using the resource id, e.g.

```shell
terraform import azurerm_mssql_server_transparent_data_encryption.example /subscriptions/12345678-1234-9876-4563-123456789012/resourceGroups/group1/providers/Microsoft.Sql/servers/server1/encryptionProtector/current
```

````

manicminer (changes requested):
Hi @MoAhuja, thanks for addressing @jackofallops' earlier comments. I've rebased and pushed the branch to resolve the merge conflict, and the acceptance tests look good and are passing 🎉 

Screenshot 2021-04-20 at 12 18 07


I have a few additional comments inline - if you can take a look at these then this should be good to merge. Thanks!

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
MoAhuja: I assume you mean serverId.ID() instead of id?

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
MoAhuja: id is a struct here - Did you mean id.ID() or id.String()?

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
MoAhuja: I could be wrong here, but I don't think it'll work in this case. The schema input being defined is the server_id (which takes the format: /sub/rg/serverName) not an encryption protector ID. As a result, when doing the delete it doesn't have enough information to parse an encryption protector id (which takes the format /sub/rg/serverName/encryptionProtectorName). 

The reason we don't need the encryption protector ID and instead rely on the server ID is because the EP is effectively a singleton, and there can only be one assigned to a server and the name is fixed (API sets it to "current").

MoAhuja: If you meant, `id, err := parse.EncryptionProtectorID(d.Id())` then yes I can make the adjustment. I forgot about the d.Id() property :)

website/docs/r/mssql_server_transparent_data_encryption.html.markdown
MoAhuja: I'm curious about this one. Encryption Protector is automatically created as part of SQL Server. The intent was to disable importing and instead allow this resource to overwrite the existing EP. You'll notice the typical ImportAsExists error has been removed. If we force importing the resource, it means you won't be able to create SQL Server & create the EP in the same configuration since that error will be thrown as a result of MS SQL server creating it under the covers. To support importing, you'd have to:

1. Create SQL Server
2. Import Transparent Data Encryption resource
3. Update TDE (Encryption Protector) to your desired settings

To me it made more sense to disable importing of this resource. If so, should the Importer property on line 30 `Importer: azSchema.ValidateResourceIDPriorToImport(func(id string) error` be set to nil instead?

azurerm/internal/services/mssql/mssql_server_transparent_data_encryption_resource.go
manicminer: Sorry for the ambiguity - that is indeed what I was trying to say :D

website/docs/r/mssql_server_transparent_data_encryption.html.markdown
manicminer: You're absolutely right, I missed that detail when considering the import of an already managed SQL Server instance - and that's not a viable workflow that we'd want to inflict. Perhaps we should add a short note in that section stating "Import is not required for this resource"?

I think we can leave the existing Importer func in place, as there's no reason _not_ to allow import if someone tries to do that?

manicminer (approved):

manicminer:
Test passing

Screenshot 2021-04-20 at 21 25 27

PR #11165, opened 2021-03-31
`azurerm_container_registry` - deprecated `classic` `sku` and `storage_account_id`

magodo:
The `classic` sku is deprecated by service. The
[document](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-skus) even doesn't show it up. Meanwhile, the service reject this sku for (at least) `PUT` request, (at least) since 2017-10-01 (used by provider v2.0.0).

Since `storage_account_id` is used together with `classic` sku, this proeprty is also useless. Furthermore, this property has been removed in [Swagger](https://github.com/Azure/azure-rest-api-specs/pull/13001).

Besides, confirmed with ACR team that all the classic instances should have been migrated today. So it should be safe to deprecate this sku.

### DIFF
--- a/azurerm/internal/services/containers/container_registry_data_source.go
+++ b/azurerm/internal/services/containers/container_registry_data_source.go
@@ -56,11 +56,6 @@ func dataSourceContainerRegistry() *schema.Resource {
 				Computed: true,
 			},
 
-			"storage_account_id": {
-				Type:     schema.TypeString,
-				Computed: true,
-			},
-
 			"tags": tags.SchemaDataSource(),
 		},
 	}
@@ -94,10 +89,6 @@ func dataSourceContainerRegistryRead(d *schema.ResourceData, meta interface{}) e
 
 	if sku := resp.Sku; sku != nil {
 		d.Set("sku", string(sku.Tier))
-	}
-
-	if account := resp.StorageAccount; account != nil {
-		d.Set("storage_account_id", account.ID)
 	}
 
 	if *resp.AdminUserEnabled {
--- a/azurerm/internal/services/containers/container_registry_resource.go
+++ b/azurerm/internal/services/containers/container_registry_resource.go
@@ -56,10 +56,9 @@ func resourceContainerRegistry() *schema.Resource {
 			"sku": {
 				Type:             schema.TypeString,
 				Optional:         true,
-				Default:          string(containerregistry.Classic),
+				Default:          string(containerregistry.Basic),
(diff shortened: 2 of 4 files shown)


azurerm/internal/services/containers/container_registry_data_source.go
```diff
@@ -56,11 +56,6 @@ func dataSourceContainerRegistry() *schema.Resource {
 				Computed: true,
 			},
 
-			"storage_account_id": {
```
katbyte, line 59: someone might be using this in a config so can we just deprecate it for now?

azurerm/internal/services/containers/container_registry_data_source.go
```diff
@@ -96,10 +91,6 @@ func dataSourceContainerRegistryRead(d *schema.ResourceData, meta interface{}) e
 		d.Set("sku", string(sku.Tier))
 	}
 
-	if account := resp.StorageAccount; account != nil {
```
katbyte, line 99: and it could be an old account so maybe we should leave this till 3.0 - should be migrated isn't a 100% yes they are all migrated

azurerm/internal/services/containers/container_registry_resource.go
```diff
@@ -56,10 +56,9 @@ func resourceContainerRegistry() *schema.Resource {
 			"sku": {
 				Type:             schema.TypeString,
 				Optional:         true,
-				Default:          string(containerregistry.Classic),
+				Default:          string(containerregistry.Basic),
 				DiffSuppressFunc: suppress.CaseDifference,
 				ValidateFunc: validation.StringInSlice([]string{
-					string(containerregistry.Classic),
```
katbyte, line 62: this is probably fine if puts where failing, but what if someone had an old one that never udpated?

azurerm/internal/services/containers/container_registry_resource.go
```diff
@@ -89,12 +88,6 @@ func resourceContainerRegistry() *schema.Resource {
 				Default:  true,
 			},
 
-			"storage_account_id": {
```
katbyte, line 92: same here, i don't see any reason to remove it now instead of deprecating it and removing in 3.0

katbyte (changes requested):
Thanks @magodo - comments have been left inline

magodo:
@katbyte I've label this as a breaking-change now, let's wait until 3.0.

katbyte:
kk @magodo  - going to close this and we can revisit/reopen when we start 3.0 work

PR #11177, opened 2021-03-31
Manage Geo backup policy for Data warehouse DBs

MoAhuja:
=== RUN   TestAccMsSqlDatabase_geoBackupPolicy
=== PAUSE TestAccMsSqlDatabase_geoBackupPolicy
=== CONT  TestAccMsSqlDatabase_geoBackupPolicy
--- PASS: TestAccMsSqlDatabase_geoBackupPolicy (550.61s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/mssql       551.681s

### DIFF
--- a/azurerm/internal/services/mssql/mssql_database_resource.go
+++ b/azurerm/internal/services/mssql/mssql_database_resource.go
@@ -292,6 +292,12 @@ func resourceMsSqlDatabase() *schema.Resource {
 				},
 			},
 
+			"geo_backup_policy": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
 			"tags": tags.Schema(),
 		},
 
@@ -311,6 +317,8 @@ func resourceMsSqlDatabaseCreateUpdate(d *schema.ResourceData, meta interface{})
 	threatClient := meta.(*clients.Client).MSSQL.DatabaseThreatDetectionPoliciesClient
 	longTermRetentionClient := meta.(*clients.Client).MSSQL.BackupLongTermRetentionPoliciesClient
 	shortTermRetentionClient := meta.(*clients.Client).MSSQL.BackupShortTermRetentionPoliciesClient
+	geoBackupPoliciesClient := meta.(*clients.Client).MSSQL.GeoBackupPoliciesClient
+
 	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
 	defer cancel()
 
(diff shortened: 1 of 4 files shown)


azurerm/internal/services/mssql/mssql_database_resource.go
```diff
@@ -292,6 +292,12 @@ func resourceMsSqlDatabase() *schema.Resource {
 				},
 			},
 
+			"geo_backup_policy": {
```
katbyte, line 295: Could we rename this to 
```suggestion
			"geo_backup_enabled": {
```
to match other boolean properties

azurerm/internal/services/mssql/mssql_database_resource.go
```diff
@@ -444,6 +452,30 @@ func resourceMsSqlDatabaseCreateUpdate(d *schema.ResourceData, meta interface{})
...
+		}
+
+		geoBackupPolicy := sql.GeoBackupPolicy{
+			GeoBackupPolicyProperties: &sql.GeoBackupPolicyProperties{
+				State: geoBackupPolicyState,
+			},
+		}
+
+		_, err := geoBackupPoliciesClient.CreateOrUpdate(ctx, serverId.ResourceGroup, serverId.Name, name, geoBackupPolicy)
+		if err != nil {
```
katbyte, line 474: these lines can be merged

azurerm/internal/services/mssql/mssql_database_resource.go
```diff
@@ -444,6 +452,30 @@ func resourceMsSqlDatabaseCreateUpdate(d *schema.ResourceData, meta interface{})
 
 	d.SetId(*read.ID)
 
+	// For datawarehouse SKUs, set the geo-backup policy
+	if d.HasChange("geo_backup_policy") && strings.HasPrefix(skuName.(string), "DW") {
+		v := d.Get("geo_backup_policy").(bool)
```
katbyte, line 457: should we also only call this API on creation when its set to true (or has a change on update?)

katbyte (changes requested):
Thanks for the pr @MoAhuja - overall this looks good and i've left a couple comments inline to address

azurerm/internal/services/mssql/mssql_database_resource.go
MoAhuja: No problem. Will change in next commit.

azurerm/internal/services/mssql/mssql_database_resource.go
MoAhuja: Changed in next commit

azurerm/internal/services/mssql/mssql_database_resource.go
MoAhuja: The default for Geo backup policy is true, so make the logic so that it will set the property:

- on changes
- on new resource creation where the geo backup enabled property is false.

MoAhuja:
@katbyte  Thanks for the review. I've made the suggested adjustments.

katbyte (approved):
Thanks @MoAhuja - LGTM 👍

PR #11189, opened 2021-04-01
add open service mesh option for kuberentes service (public preview)

yershalom:
https://techcommunity.microsoft.com/t5/apps-on-azure/announcing-public-preview-of-the-open-service-mesh-osm-aks-add/ba-p/2247361

https://docs.microsoft.com/en-us/azure/aks/servicemesh-osm-about?pivots=client-operating-system-linux

### DIFF
--- a/azurerm/internal/services/containers/kubernetes_addons.go
+++ b/azurerm/internal/services/containers/kubernetes_addons.go
@@ -21,6 +21,7 @@ const (
 	kubernetesDashboardKey    = "kubeDashboard"
 	httpApplicationRoutingKey = "httpApplicationRouting"
 	omsAgentKey               = "omsagent"
+	openServiceMeshKey        = "openServiceMesh"
 )
 
 // The AKS API hard-codes which add-ons are supported in which environment
@@ -34,11 +35,13 @@ var unsupportedAddonsForEnvironment = map[string][]string{
 		azurePolicyKey,            // https://github.com/terraform-providers/terraform-provider-azurerm/issues/6462
 		httpApplicationRoutingKey, // https://github.com/terraform-providers/terraform-provider-azurerm/issues/5960
 		kubernetesDashboardKey,    // https://github.com/terraform-providers/terraform-provider-azurerm/issues/7487
+		openServiceMeshKey,        // Preview features are not supported in Azure China
 	},
 	azure.USGovernmentCloud.Name: {
 		azurePolicyKey,            // https://github.com/terraform-providers/terraform-provider-azurerm/issues/6702
 		httpApplicationRoutingKey, // https://github.com/terraform-providers/terraform-provider-azurerm/issues/5960
 		kubernetesDashboardKey,    // https://github.com/terraform-providers/terraform-provider-azurerm/issues/7136
+		openServiceMeshKey,        // Preview features are not supported in Azure Government
 	},
 }
 
@@ -154,6 +157,20 @@ func schemaKubernetesAddOnProfiles() *schema.Schema {
 						},
 					},
 				},
+
+				"open_service_mesh": {
+					Type:     schema.TypeList,
+					MaxItems: 1,
+					Optional: true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"enabled": {
+								Type:     schema.TypeBool,
+								Required: true,
+							},
+						},
+					},
+				},
 			},
 		},
 	}
@@ -170,6 +187,7 @@ func expandKubernetesAddOnProfiles(input []interface{}, env azure.Environment) (
 		kubernetesDashboardKey:    &disabled,
 		httpApplicationRoutingKey: &disabled,
 		omsAgentKey:               &disabled,
+		openServiceMeshKey:        &disabled,
 	}
 
 	if len(input) == 0 {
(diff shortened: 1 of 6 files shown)


azurerm/internal/services/containers/kubernetes_addons.go
```diff
@@ -154,6 +157,20 @@ func schemaKubernetesAddOnProfiles() *schema.Schema {
 						},
 					},
 				},
+
+				"open_service_mesh": {
```
tombuildsstuff, line 161: Can we document this new block [both in the `addon_profile` block](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/kubernetes_cluster.html.markdown#L198-L216) and the specific field within it ([example](https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/r/kubernetes_cluster.html.markdown#L268-L274))? Notably since this is in Preview we'll also want to add this below both of those:

```
~> **Note:** Open Service Mesh is available on an opt-in Preview basis - to use this the Preview Feature `AKS-OpenServiceMesh` will need to be registered, [see the AKS Documentation for more information](https://docs.microsoft.com/en-gb/azure/aks/servicemesh-osm-about)
```

azurerm/internal/services/containers/kubernetes_cluster_data_source.go
```diff
@@ -118,6 +118,19 @@ func dataSourceKubernetesCluster() *schema.Resource {
 								},
 							},
 						},
+
+						"open_service_mesh": {
```
tombuildsstuff, line 122: same for the data source - can we add this to the `addon_profile` block: https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/d/kubernetes_cluster.html.markdown#L91-L101

and the specific fields below that as here: https://github.com/terraform-providers/terraform-provider-azurerm/blame/master/website/docs/d/kubernetes_cluster.html.markdown#L255-L257

tombuildsstuff (changes requested):
hey @yershalom 

Thanks for this PR - taking a look through on the whole this looks pretty good to me, but can we add some documentation for these new fields so that users are able to find them?

Thanks!

azurerm/internal/services/containers/kubernetes_addons.go
yershalom: @tombuildsstuff Added, thanks

azurerm/internal/services/containers/kubernetes_cluster_data_source.go
yershalom: @tombuildsstuff Added, thanks

azurerm/internal/services/containers/kubernetes_addons.go
```diff
@@ -248,6 +266,17 @@ func expandKubernetesAddOnProfiles(input []interface{}, env azure.Environment) (
 		}
 	}
 
+	openServiceMesh := profile[""].([]interface{})
```
tombuildsstuff, line 269: presumably this wants to be:

```suggestion
	openServiceMesh := profile["open_service_mesh"].([]interface{})
```

azurerm/internal/services/containers/kubernetes_addons.go
```diff
@@ -154,6 +157,20 @@ func schemaKubernetesAddOnProfiles() *schema.Schema {
...
+				"open_service_mesh": {
+					Type:     schema.TypeList,
+					MaxItems: 1,
+					Optional: true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"enabled": {
+								Type:     schema.TypeBool,
+								Required: true,
+							},
```
tombuildsstuff, line 170: can we ensure these are set back into the state too? this is done in the flatten function below: https://github.com/terraform-providers/terraform-provider-azurerm/pull/11189/files#diff-c28ad69551916efc44930e639bf392636923aa37f4a68e17cb874edf2c34b101R312-R341

azurerm/internal/services/containers/kubernetes_cluster_data_source.go
```diff
@@ -118,6 +118,19 @@ func dataSourceKubernetesCluster() *schema.Resource {
...
+							Computed: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:     schema.TypeBool,
+										Computed: true,
+									},
+								},
+							},
+						},
```
tombuildsstuff, line 133: we'll also need to flatten this value for the Data Source, this is done in `flattenKubernetesClusterDataSourceAddonProfiles` (by parsing the add-on profile from the response and reading the enabled state)

tombuildsstuff (changes requested):
hey @yershalom 

Thanks for pushing those changes, I've taken a look through and left a few more comments inline, but if we can fix those up (and the tests pass) then this should be good to go 👍 

Thanks!

katbyte:
Hey @yershalom - just wondering if your still working on this 🙂

yershalom:
Hi @katbyte , yes, I have a lot of pressure now but I will fix the comment and conflicts when I'll have time :-)

katbyte:
hey @yershalom - i'm going to close this for now as its been a month and the branch is getting out of date/conflicted - please do reopen this once you've had time to address the reviews - thanks again!

PR #11196, opened 2021-04-01
`azurerm_web_application_firewall_policy` - `http_listener_ids` and `path_based_rule_ids` are now Computed only

mbfrahry:
Unfortunately, `http_listener_ids` and `path_based_rule_ids` are listed as READONLY in all versions of the azure-sdk. This PR sets those fields as Computed and removes the failing test.

```
// HTTPListeners - READ-ONLY; A collection of references to application gateway http listeners.
HTTPListeners *[]SubResource `json:"httpListeners,omitempty"`
// PathBasedRules - READ-ONLY; A collection of references to application gateway path rules.
PathBasedRules *[]SubResource `json:"pathBasedRules,omitempty"`
```


katbyte (approved):
LGTM 👍

aberva:
Hi,

The import for Web Application Firewall Policy does not include http_listener_ids and path_based_rule_ids.

Additionally, could you provide an example http_listener_ids or path_based_rule_ids declaration?
I did not find any related examples.
I was hoping to be able to base myself on the azurerm_application_gateway resource (in the same configuration file) but in the documentation it does not seem possible to export the "http_listener".

Thank you for your help.

mbfrahry:
Hi @aberva. These changes will go out in the next version of the Azure provider. Also, `http_listener_ids` and `path_based_rule_ids` are not declarable because the API treats these attributes as read only. 

If you have any other questions, please open a new issue and we can continue our conversation there.

sirlatrom:
I'm surprised to learn this, but then I didn't know where to check for read only status.

It is possible to set the attributes through the Azure Portal, so when I added them to the provider, it was with the impression it was possible. I tested it in a sandbox subscription, too, so I think it might be the docs on these attributes that are wrong...

aberva:
I just opened an issue on the import of the zurerm_web_application_firewall_policy (version v.2.55.0) which is closely related to the subject.

https://github.com/terraform-providers/terraform-provider-azurerm/issues/11273#issue-854379807

sirlatrom:
> I'm surprised to learn this, but then I didn't know where to check for read only status.
> 
> It is possible to set the attributes through the Azure Portal, so when I added them to the provider, it was with the impression it was possible. I tested it in a sandbox subscription, too, so I think it might be the docs on these attributes that are wrong...

Turns out this is fixed in #11239

PR #11236, opened 2021-04-07
`azurerm_hpc_cache` - support for `ntp_server` and `dns`

magodo:
Support `ntp_server` and `dns_setting` for `azurerm_hpc_cache`.

### DIFF
--- a/azurerm/internal/services/hpccache/hpc_cache_resource.go
+++ b/azurerm/internal/services/hpccache/hpc_cache_resource.go
@@ -533,7 +533,6 @@ func flattenStorageCacheNetworkSettings(settings *storagecache.CacheNetworkSetti
 				"search_domain": searchDomain,
 			},
 		}
-
 	}
 	return
 }
--- a/azurerm/internal/services/hpccache/hpc_cache_resource.go
+++ b/azurerm/internal/services/hpccache/hpc_cache_resource.go
@@ -93,7 +93,7 @@ func resourceHPCCache() *schema.Resource {
 				ValidateFunc: validation.StringIsNotEmpty,
 			},
 
-			"dns_setting": {
+			"dns": {
 				Type:     schema.TypeList,
 				Optional: true,
 				MaxItems: 1,
@@ -333,8 +333,8 @@ func resourceHPCCacheRead(d *schema.ResourceData, meta interface{}) error {
 		mtu, ntpServer, dnsSetting := flattenStorageCacheNetworkSettings(props.NetworkSettings)
 		d.Set("mtu", mtu)
 		d.Set("ntp_server", ntpServer)
-		if err := d.Set("dns_setting", dnsSetting); err != nil {
-			return fmt.Errorf("setting `dns_setting`: %v", err)
+		if err := d.Set("dns", dnsSetting); err != nil {
+			return fmt.Errorf("setting `dns`: %v", err)
 		}
 
 		if securitySettings := props.SecuritySettings; securitySettings != nil {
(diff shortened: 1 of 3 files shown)


azurerm/internal/services/hpccache/hpc_cache_resource.go
```diff
@@ -86,6 +86,38 @@ func resourceHPCCache() *schema.Resource {
 				ValidateFunc: validation.IntBetween(576, 1500),
 			},
 
+			"ntp_server": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "time.windows.com",
```
tombuildsstuff, line 92: does this change depending on the Azure Environment being used? I'm not sure if this is what's used in Azure China/Germany/US Gov?

azurerm/internal/services/hpccache/hpc_cache_resource.go
```diff
@@ -454,3 +483,50 @@ func flattenStorageCacheNfsAccessRules(input *[]storagecache.NfsAccessRule) ([]i
...
+	}
+	return out
+}
+
+func flattenStorageCacheNetworkSettings(settings *storagecache.CacheNetworkSettings, d *schema.ResourceData) error {
+	if settings == nil {
+		return nil
+	}
+
+	d.Set("mtu", settings.Mtu)
```
tombuildsstuff, line 509: this is a `*int32` so this'll want casting to an `int` for `d.Set`?

azurerm/internal/services/hpccache/hpc_cache_resource.go
```diff
@@ -454,3 +483,50 @@ func flattenStorageCacheNfsAccessRules(input *[]storagecache.NfsAccessRule) ([]i
...
+		}
+
+		dnsSetting := []interface{}{
+			map[string]interface{}{
+				"servers":       dnsServers,
+				"search_domain": searchDomain,
+			},
+		}
+
+		if err := d.Set("dns_setting", dnsSetting); err != nil {
```
tombuildsstuff, line 527: this shouldn't be settings things within this function, can we return an empty array (and above where this is nil - to track changes for the nil case) and have this in the read?

tombuildsstuff (changes requested):
hey @magodo 

Thanks for this PR.

I've taken a look through and left some comments inline but on the whole this looks pretty good - if we can fix those comments up then we should be able to take another look.

Thanks!

azurerm/internal/services/hpccache/hpc_cache_resource.go
magodo: Confirmed with service team that "it appears" the default value is used among all envs.

azurerm/internal/services/hpccache/hpc_cache_resource.go
```diff
@@ -86,6 +86,38 @@ func resourceHPCCache() *schema.Resource {
...
 			},
 
+			"ntp_server": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "time.windows.com",
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
+
+			"dns_setting": {
```
katbyte, line 96: i think setting is redundant here?
```suggestion
			"dns": {
```

magodo:
I've updated per comment, please take another look.

## Test Result

```bash
💤 TF_ACC=1 go test ./azurerm/internal/services/hpccache -v -run="TestAccHPCCache_ntpServer|TestAccHPCCache_dnsSetting" -timeout=1h
2021/04/08 10:01:03 [DEBUG] not using binary driver name, it's no longer needed
2021/04/08 10:01:04 [DEBUG] not using binary driver name, it's no longer needed
=== RUN   TestAccHPCCache_ntpServer
=== PAUSE TestAccHPCCache_ntpServer
=== RUN   TestAccHPCCache_dnsSetting
=== PAUSE TestAccHPCCache_dnsSetting
=== CONT  TestAccHPCCache_ntpServer
=== CONT  TestAccHPCCache_dnsSetting
--- PASS: TestAccHPCCache_ntpServer (1760.56s)
--- PASS: TestAccHPCCache_dnsSetting (1930.41s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/hpccache    1932.010s
```

tombuildsstuff (approved):
LGTM - thanks for pushing those changes @magodo

Issue #11231, opened 2021-04-07
Upcoming Breaking Changes to Front Door / CDN's


sebader:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

A few days ago I got his mail from Azure:

> Before you delete Azure Front Door or Azure Content Delivery Network resources, remove their endpoint CNAME records from DNS
You're receiving this email because you use Azure Front Door or Content Delivery Network. 
On 9 April 2021, we're updating Azure Front Door and Content Delivery Network to help prevent dangling DNS entries and the security risks they create. At that time, we'll start requiring the removal of canonical name (CNAME) records for Azure Front Door and Content Delivery Network resource endpoints from DNS before the resources can be deleted. 
**To delete Azure Front Door or Content Delivery Network resources, you must first remove the resource endpoint CNAME records from DNS starting on 9 April 2021.** 

I see a potential issue that could affect the terraform provider with the snippet below that I have been using so far (see below)

### New or Affected Resource(s)

* azurerm_dns_cname_record
* azurerm_frontdoor

### Potential Terraform Configuration

```hcl
data "azurerm_dns_zone" "customdomain" {
  name                = var.custom_domain_name
  resource_group_name = var.dns_zone_resourcegroup_name
}

resource "azurerm_dns_cname_record" "app_subdomain" {
  name                = "${var.prefix}-app"
  zone_name           = data.azurerm_dns_zone.customdomain.name
  resource_group_name = var.dns_zone_resourcegroup_name
  ttl                 = 3600
  record              = "${var.prefix}frontdoor.azurefd.net"
}

resource "azurerm_frontdoor" "main" {
  name                             = "${var.prefix}frontdoor"
  resource_group_name   = azurerm_resource_group.main.name

  frontend_endpoint {
    name                     = "CustomDomainFrontendEndpoint"
    host_name                = azurerm_dns_cname_record.app_subdomain.fqdn
  }
}
```

This will obviously create a dependency graph from FrontDoor on the CNAME when you create the resources. But destroy would usually work the other way round - which will probably break things with the newly announced behavior.

[...]

PR #11237, opened 2021-04-07
docs: adding a heads up about the breaking api change rolling out 2021-04-09

tombuildsstuff:
More details are in #11231, although it's hard for us to do much about this ahead of time unfortunately due to the inability to test these changes (and confirm which breaking changes are necessary), as such giving some notice this is happening is the most helpful thing we can do at the moment.


katbyte (approved):
LGTM 👍

PR #11240, opened 2021-04-07
Make KeyVault work in Private Networks

ericfode:
Currently, if you disable public network access when a keyvault is created (with the intent of setting up a private dns endpoint) you put it in a state that terraform cannot proceed from. This is caused by it waiting for the actual http url to become available. This pattern is a contrast to other resources which separate out accessing (and controlling) the data plane of the provisioned resource from executing actions against azures control plane. 

The proposed solution adds a flag that tells terraform to skip checking for liveness of the vault url, and only check for contacts if they are actually used in the resource or said flag is not activated. 

The (arguably more correct) alternative would be to separate key vault contacts out into it's own resource.

### DIFF
--- a/azurerm/internal/features/defaults.go
+++ b/azurerm/internal/features/defaults.go
@@ -4,6 +4,7 @@ func Default() UserFeatures {
 	return UserFeatures{
 		// NOTE: ensure all nested objects are fully populated
 		KeyVault: KeyVaultFeatures{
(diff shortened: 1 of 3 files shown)


manicminer:
@ericfode Thanks for this suggestion. I've replied in more detail on the related issue #11245, but as I'm unable to reproduce I'm going to close this for now. Please feel free to follow up in that issue if you are still encountering this. Thanks!

PR #11247, opened 2021-04-08
`azurerm_hpc_cache_nfs_target` - support of more `usage_model` enums

magodo:
`azurerm_hpc_cache_nfs_target` - support of more `usage_model` enums

### DIFF
--- a/azurerm/internal/services/hpccache/hpc_cache_nfs_target_resource.go
+++ b/azurerm/internal/services/hpccache/hpc_cache_nfs_target_resource.go
@@ -101,8 +101,12 @@ func resourceHPCCacheNFSTarget() *schema.Resource {
 				Required: true,
 				ValidateFunc: validation.StringInSlice([]string{
 					"READ_HEAVY_INFREQ",
+					"READ_HEAVY_CHECK_180",
 					"WRITE_WORKLOAD_15",
 					"WRITE_AROUND",
+					"WRITE_WORKLOAD_CHECK_30",
+					"WRITE_WORKLOAD_CHECK_60",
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/hpccache/hpc_cache_nfs_target_resource.go
```diff
@@ -101,8 +101,12 @@ func resourceHPCCacheNFSTarget() *schema.Resource {
 				Required: true,
 				ValidateFunc: validation.StringInSlice([]string{
 					"READ_HEAVY_INFREQ",
+					"READ_HEAVY_CHECK_180",
 					"WRITE_WORKLOAD_15",
 					"WRITE_AROUND",
+					"WRITE_WORKLOAD_CHECK_30",
+					"WRITE_WORKLOAD_CHECK_60",
+					"WRITE_WORKLOAD_CLOUDWS",
```
tombuildsstuff, line 109: @magodo is there a source for these constants? appears these aren't in the SDK but perhaps they should be?

tombuildsstuff (approved):
one follow up item but this otherwise LGTM 👍

azurerm/internal/services/hpccache/hpc_cache_nfs_target_resource.go
magodo: Indeed, I'll submit a github issue for this. Currently, these are from [this API](https://github.com/Azure/azure-rest-api-specs/blob/7e3bf6306c5ffd00b97a93870945fa5debd873af/specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2021-03-01/storagecache.json#L119).

PR #11259, opened 2021-04-08
Remove obsolete reference to virtual_network_name.

benwarfield-usds:
As far as I can tell, this argument was removed from app_service_slot when it was removed from app_service and function_app, but the docs were not updated.


katbyte (approved):
Thanks @benwarfield-usds - LGTM 👍

PR #11267, opened 2021-04-09
`azurerm_app_service_slot` - fix crash bug when given empty `http_logs`

phekmat:
This fixes a provider crash when the `http_logs` block is empty for the app service slot (and likely the parent app service):
```hcl
  logs {
    http_logs {
    }
  }
```

IIRC, that configuration is currently the only way to disable the app service HTTP logs.

### DIFF
--- a/azurerm/internal/services/web/app_service.go
+++ b/azurerm/internal/services/web/app_service.go
@@ -1394,9 +1394,12 @@ func expandAppServiceLogs(input interface{}) web.SiteLogsConfigProperties {
 		appLogsConfigs := v.([]interface{})
 
 		for _, config := range appLogsConfigs {
-			appLogsConfig := config.(map[string]interface{})
-
 			logs.ApplicationLogs = &web.ApplicationLogsConfig{}
+
+			if config == nil {
+				continue
+			}
+			appLogsConfig := config.(map[string]interface{})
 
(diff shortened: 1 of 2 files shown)


azurerm/internal/services/web/app_service.go
```diff
@@ -1424,10 +1424,13 @@ func expandAppServiceLogs(input interface{}) web.SiteLogsConfigProperties {
 		httpLogsConfigs := v.([]interface{})
 
 		for _, config := range httpLogsConfigs {
-			httpLogsConfig := config.(map[string]interface{})
-
 			logs.HTTPLogs = &web.HTTPLogsConfig{}
 
+			if config == nil {
```
magodo, line 1429: Since the `http_logs` is a list with maximum length set to `1`, the for loop here is a bit of weird. However, given this is the existing code, let's keep it as it was for now. 

Can we add the same logic for `application_logs`?

magodo (approved):
@phekmat Thank you for this PR 👍 It looks good to me. A little suggestion is that maybe we shall also do the same nil check for `application_logs`, WDYT?

azurerm/internal/services/web/app_service.go
phekmat: It's less likely to occur there since you can turn those off via the log levels, but it wouldn't hurt. I'll update in the morning

PR #11275, opened 2021-04-09
Stop force recreate on Key Vault Certificate tags

jvassbo:
Changing Key Vault Certificate tags should not trigger a recreate of resource.


jvassbo:
I see updating Key Vault Certificate resource is not implemented

PR #11278, opened 2021-04-09
r/subscription: support for tags

harshavmb:
Adding tags to subscription resource.. will add logic for update & write documentation as & when this PR works


tombuildsstuff:
Dependent on #11277

jackofallops:
Hi @harshavmb - Thanks for this PR. I'm afraid there is, currently, no way to set tags on subscriptions via the SDK or API, so I'm going to close this for now.  As soon as this feature is supported at the API, we can add this in.

PR #11301, opened 2021-04-13
`azurerm_storage_account` `blob_properties` enhancement 

yupwei68:
Extends https://github.com/terraform-providers/terraform-provider-azurerm/pull/9277
Fix:
https://github.com/terraform-providers/terraform-provider-azurerm/issues/8268
https://github.com/terraform-providers/terraform-provider-azurerm/issues/7722
https://github.com/terraform-providers/terraform-provider-azurerm/issues/6627

=== RUN   TestAccStorageAccount_blobProperties
=== PAUSE TestAccStorageAccount_blobProperties
=== CONT  TestAccStorageAccount_blobProperties
--- PASS: TestAccStorageAccount_blobProperties (274.03s)

Since `restore_policy` has some update problem, it'll be enhanced when the fix is released.

`retentionDays` in `ChangeFeed` is only supported in canary regions now, all regions will be supported soon.


azurerm/internal/services/storage/validate/storage_blob_properties_default_service_version.go
```diff
@@ -0,0 +1,48 @@
...
+		"2017-07-29",
+		"2017-11-09",
+		"2018-03-28",
+		"2018-11-09",
+		"2019-02-02",
+		"2019-07-07",
+		"2019-12-12",
+		"2020-02-10",
+		"2020-04-08",
+		"2020-06-12",
```
tombuildsstuff, line 38: looks like at least one of these doesn't exist? https://docs.microsoft.com/en-us/rest/api/storageservices/previous-azure-storage-service-versions

website/docs/r/storage_account.html.markdown
```diff
@@ -137,6 +137,14 @@ A `blob_properties` block supports the following:
 
 * `delete_retention_policy` - (Optional) A `delete_retention_policy` block as defined below.
 
+* `versioning_enabled` - (Optional) Is versioning enabled? Default to `false`.
+
+* `change_feed` - (Optional) A `change_feed` block as defined below.
+
+* `default_service_version` - (Optional) The default version to use for requests to the Blob service if an incoming request’s version is not specified. Default to `2020-06-12`.
```
tombuildsstuff, line 144: as is the case for most of the Swagger descriptions, this isn't clear:

```suggestion
* `default_service_version` - (Optional) The API Version which should be used by default for requests to the Data Plane API if an incoming request doesn't specify an API Version. Defaults to `2020-06-12`.
```

website/docs/r/storage_account.html.markdown
```diff
@@ -173,6 +181,23 @@ A `container_delete_retention_policy` block supports the following:
...
+
+---
+
+`last_access_time_tracking_policy` exports the following:
+
+* `name` - (Optional) The name of the policy. Default to `AccessTimeTracking`. Current possible value is `AccessTimeTracking`.
+
+* `granularity_in_days` - (Optional) The blob object tracking granularity in days, typically how often the blob object should be tracked. Default to `1`. Current possible value is `1`.
+
+* `blob_type` - (Optional) An array of predefined supported blob types. Default to `blockBlob`. Current possible value is `blockBlob`.
```
tombuildsstuff, line 198: A few things here:

1. This is an exported block, so needs to be moved below
2. Terraform doesn't support arrays, it's lists
3. The descriptions for `name`, `granularity_in_days` and `blob_type` aren't clear what this does - can we improve this?
4. If `blockBlob` is the only supported value, there's no point exposing this field at this point in time?

azurerm/internal/services/storage/storage_account_resource.go
```diff
@@ -1851,6 +1992,44 @@ func flattenBlobPropertiesDeleteRetentionPolicy(input *storage.DeleteRetentionPo
...
+	}
+
+	days := 0
+	if input.RetentionInDays != nil {
+		days = int(*input.RetentionInDays)
+	}
+
+	return []interface{}{map[string]interface{}{
+		"retention_in_days": days,
+	},
```
tombuildsstuff, line 2009: _minor_ but I'm surprised the linter isn't catching this:

```suggestion
	return []interface{}{
		map[string]interface{}{
			"retention_in_days": days,
		},
```

azurerm/internal/services/storage/storage_account_resource.go
```diff
@@ -267,6 +267,66 @@ func resourceStorageAccount() *schema.Resource {
...
+										ValidateFunc: validation.IntInSlice([]int{1}),
+									},
+									"blob_type": {
+										Type:     schema.TypeList,
+										Optional: true,
+										Computed: true,
+										Elem: &schema.Schema{
+											Type:         schema.TypeString,
+											ValidateFunc: validation.StringInSlice([]string{"blockBlob"}, false),
+										},
```
tombuildsstuff, line 324: a few things here:

1. `granularity_in_days` and `blob_type` can't be set, so should be computed-only
2. there's a single possible value for `name` and `blob_type` - so these likely shouldn't be exposed

as such this block could likely be removed in place of a single field for `last_access_time_tracking_policy_enabled` - unless there's plans by the service team to add other functionality here?

azurerm/internal/services/storage/storage_account_resource.go
```diff
@@ -1556,6 +1644,45 @@ func expandBlobPropertiesCors(input []interface{}) *storage.CorsRules {
...
 
+func expandBlobPropertiesChangeFeed(inputs []interface{}, isHnsEnabled bool, accountKind string) (*storage.ChangeFeed, error) {
+	if len(inputs) == 0 {
+		return &storage.ChangeFeed{
+			Enabled: utils.Bool(false),
+		}, nil
+	}
+
+	if isHnsEnabled {
+		return nil, fmt.Errorf("`change_feed` in `blob_properties` could not be enabled when `is_hns_enabled` is enabled")
```
tombuildsstuff, line 1655:
```suggestion
		return nil, fmt.Errorf("`change_feed` in `blob_properties` cannot be enabled when `is_hns_enabled` is enabled")
```

azurerm/internal/services/storage/storage_account_resource.go
```diff
@@ -1556,6 +1644,45 @@ func expandBlobPropertiesCors(input []interface{}) *storage.CorsRules {
...
+	}
+
+	if accountKind != string(storage.BlobStorage) && accountKind != string(storage.StorageV2) {
+		return nil, fmt.Errorf("`change_feed` in `blob_properties` is only support `account_kind` is `BlobStorage` or `StorageV2`")
```
tombuildsstuff, line 1659:
```suggestion
		return nil, fmt.Errorf("`change_feed` can only be configured when `account_kind` is set to `BlobStorage` or `StorageV2`")
```

azurerm/internal/services/storage/storage_account_resource.go
```diff
@@ -267,6 +267,66 @@ func resourceStorageAccount() *schema.Resource {
...
+						"change_feed": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"retention_in_days": {
+										Type:         schema.TypeInt,
+										Optional:     true,
+										ValidateFunc: validation.IntBetween(1, 146000),
```
tombuildsstuff, line 280: > 146000/365 = 400

This is 400 years worth of changes.. is there a realistic maximum value from the service team here?

azurerm/internal/services/storage/storage_account_resource.go
```diff
@@ -1851,6 +1992,44 @@ func flattenBlobPropertiesDeleteRetentionPolicy(input *storage.DeleteRetentionPo
...
+	if input.TrackingGranularityInDays != nil {
+		days = int(*input.TrackingGranularityInDays)
+	}
+
+	return []interface{}{map[string]interface{}{
+		"name":                input.Name,
+		"granularity_in_days": days,
+		"blob_type":           utils.FlattenStringSlice(input.BlobType),
+	},
+	}
```
tombuildsstuff, line 2030: _minor_ I'm surprised the linter didn't complain about this but:

```suggestion
	return []interface{}{
		map[string]interface{}{
			"name":                input.Name,
			"granularity_in_days": days,
			"blob_type":           utils.FlattenStringSlice(input.BlobType),
		},
	}
```

tombuildsstuff (changes requested):
hi @yupwei68 

Thanks for this PR.

I've taken a look through here and left some comments inline, if we can fix those up then we should be able to take another look - but in particular if we can fix up the descriptions to be clearer that'd be great.

Thanks!

azurerm/internal/services/storage/validate/storage_blob_properties_default_service_version.go
yupwei68: The "2008-10-27" and "2020-06-12" are supported, but not list in the list. I've tested these two.

azurerm/internal/services/storage/storage_account_resource.go
yupwei68: Yes. The service plans to support the `granularity_in_days` and `blob_type` in the future. I just thought that it may be hard to covert `last_access_time_tracking_policy` from bool to list in the future.
Currently `azurerm_management_policy` is dependent on the `last_access_time_tracking_policy`.
But I will change it into bool now.

azurerm/internal/services/storage/storage_account_resource.go
yupwei68: Yes. It's 400 years.

yupwei68: > Minimum value is 1 day and maximum value is 146000 days (400 years).

yupwei68:
Hi Tom, thanks for your comments. Changes have been pushed. Tests have all passed. Please continue reviewing.

azurerm/internal/services/storage/storage_account_resource.go
```diff
@@ -267,6 +267,41 @@ func resourceStorageAccount() *schema.Resource {
...
+						},
+
+						"default_service_version": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							Default:      "2020-06-12",
+							ValidateFunc: validate.BlobPropertiesDefaultServiceVersion,
+						},
+
+						"last_access_time_tracking_policy_enabled": {
```
katbyte, line 299: i think we can shorten this to
```suggestion
						"last_access_time_enabled": {
```

tracking and policy are redundant i think?

katbyte (changes requested):
Thanks @yupwei68 - aside from one comment on a property name this looks good to me

yupwei68:
Hi kt, thanks for your comments. Tests have passed. The CI has failed for linter broken. Please continue reviewing.
=== RUN   TestAccStorageAccount_blobProperties
=== PAUSE TestAccStorageAccount_blobProperties
=== CONT  TestAccStorageAccount_blobProperties
--- PASS: TestAccStorageAccount_blobProperties (687.90s)

katbyte (approved):
@yupwei68 - ci needs to be fix but otherwise looks goof

PR #11303, opened 2021-04-13
New resource azurerm_express_route_circuit_connection 

xuzhang3:
`Express Route Circuit Connection` is a special resource,  also know as `Global Reach` in portal. 
Test constraints: 
1.  Two existed Express Route Circuits with `Provider provisioned` , the provision status is controlled by [Service provider](https://docs.microsoft.com/en-us/azure/expressroute/expressroute-locations-providers#partners), can be provisioned by [send the service key with provider](https://docs.microsoft.com/en-us/azure/expressroute/expressroute-howto-circuit-portal-resource-manager#send-the-service-key-to-your-connectivity-provider-for-provisioning).  This will cost money, the cost depends on the provider.
2. The [provider location](https://docs.microsoft.com/en-us/azure/expressroute/expressroute-global-reach#availability) should support global reach.

Acceptance tests depends on environment variables:
 `ARM_TEST_DATA_RESOURCE_GROUP` : resource group name,
`ARM_TEST_CIRCUIT_NAME_FIRST`:  The first Express Route Circuit  name
`ARM_TEST_CIRCUIT_NAME_SECOND`: The  second Express Route Circuit  name

AccTests result:
```log
TF_ACC=1 go test -v ./azurerm/internal/services/network -run=TestAccExpressRouteCircuitConnection -timeout 60m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
2021/04/13 15:03:31 [DEBUG] not using binary driver name, it's no longer needed
2021/04/13 15:03:32 [WARN] could not locate a terraform executable on system path; continuing
2021/04/13 15:03:53 [DEBUG] not using binary driver name, it's no longer needed
2021/04/13 15:03:53 [WARN] could not locate a terraform executable on system path; continuing
=== RUN   TestAccExpressRouteCircuitConnectionDataSource_basic
=== PAUSE TestAccExpressRouteCircuitConnectionDataSource_basic
=== RUN   TestAccExpressRouteCircuitConnection_basic
--- PASS: TestAccExpressRouteCircuitConnection_basic (185.05s)
=== RUN   TestAccExpressRouteCircuitConnection_requiresImport
--- PASS: TestAccExpressRouteCircuitConnection_requiresImport (215.39s)
=== RUN   TestAccExpressRouteCircuitConnection_complete
--- PASS: TestAccExpressRouteCircuitConnection_complete (198.20s)
=== RUN   TestAccExpressRouteCircuitConnection_update
--- PASS: TestAccExpressRouteCircuitConnection_update (369.41s)
=== RUN   TestAccExpressRouteCircuitConnection_updateIpv6CircuitConnectionConfig
--- PASS: TestAccExpressRouteCircuitConnection_updateIpv6CircuitConnectionConfig (314.78s)
=== CONT  TestAccExpressRouteCircuitConnectionDataSource_basic
--- PASS: TestAccExpressRouteCircuitConnectionDataSource_basic (189.36s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network     1514.873s

```


azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+	azSchema "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tf/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func resourceExpressRouteCircuitConnection() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceExpressRouteCircuitConnectionCreateUpdate,
+		Read:   resourceExpressRouteCircuitConnectionRead,
+		Update: resourceExpressRouteCircuitConnectionCreateUpdate,
```
tombuildsstuff, line 24: newer resources shouldn't be reusing the Create/Update functions as it means that it's not possible to `ignore_changes` on fields - can we split the Create and Update methods to support Delta Updates (e.g. d.HasChanges) here

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"address_prefix": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validation.IsCIDR,
+						},
```
tombuildsstuff, line 92: given there's 2 address prefixes here, this'd likely make sense as two top level fields: `address_prefix_ipv4` and `address_prefix_ipv6`, WDYT?

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+
+	id, err := parse.ExpressRouteCircuitConnectionID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	resp, err := client.Get(ctx, id.ResourceGroup, id.CircuitName, id.PeeringName, id.Name)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[INFO] express route circuit connection %q does not exist - removing from state", d.Id())
```
tombuildsstuff, line 167: we can make this:

```suggestion
			log.Printf("[INFO] %s does not exist - removing from state", *id)
```

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("retrieving ExpressRouteCircuitConnection %q (Resource Group %q / circuitName %q / peeringName %q): %+v", id.Name, id.ResourceGroup, id.CircuitName, id.PeeringName, err)
```
tombuildsstuff, line 171: we can make this:

```suggestion
		return fmt.Errorf("retrieving %s: %+v", *id, err)
```

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	d.Set("name", id.Name)
+	d.Set("resource_group_name", id.ResourceGroup)
+	d.Set("circuit_name", id.CircuitName)
+	d.Set("peering_id", resp.ExpressRouteCircuitPeering.ID)
+	d.Set("peer_peering_id", resp.PeerExpressRouteCircuitPeering.ID)
```
tombuildsstuff, line 177: both of these are nested within the properties block, (that these are at the top level is a legacy decision in the Track1 SDK to work around API design bugs) - so can we move these into the properties block, and also nil-check these

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	future, err := client.Delete(ctx, id.ResourceGroup, id.CircuitName, id.PeeringName, id.Name)
+	if err != nil {
+		return fmt.Errorf("deleting ExpressRouteCircuitConnection %q (Resource Group %q / circuitName %q / peeringName %q): %+v", id.Name, id.ResourceGroup, id.CircuitName, id.PeeringName, err)
+	}
+
+	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("waiting for deletion of the ExpressRouteCircuitConnection %q (Resource Group %q / circuitName %q / peeringName %q): %+v", id.Name, id.ResourceGroup, id.CircuitName, id.PeeringName, err)
+	}
```
tombuildsstuff, line 205: can we make these:

```suggestion
	if err != nil {
		return fmt.Errorf("deleting %s: %+v", *id, err)
	}

	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
		return fmt.Errorf("waiting for deletion of %s: %+v", *id, err)
	}
```

azurerm/internal/services/network/express_route_circuit_connections_resource_test.go
```diff
@@ -0,0 +1,265 @@
...
+	id, err := parse.ExpressRouteCircuitConnectionID(state.ID)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := client.Network.ExpressRouteCircuitConnectionClient.Get(ctx, id.ResourceGroup, id.CircuitName, id.PeeringName, id.Name)
+	if err != nil {
+		if utils.ResponseWasNotFound(resp.Response) {
+			return utils.Bool(false), nil
+		}
+		return nil, fmt.Errorf("retrieving ExpressRouteCircuitConnection %q (Resource Group %q / circuitName %q / peeringName %q): %+v", id.Name, id.ResourceGroup, id.CircuitName, id.PeeringName, err)
```
tombuildsstuff, line 151:
```suggestion
		return nil, fmt.Errorf("retrieving %s: %+v", *id, err)
```

azurerm/internal/services/network/express_route_circuit_connections_resource_test.go
```diff
@@ -0,0 +1,265 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/acceptance/check"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/parse"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+type ExpressRouteCircuitConnectionResource struct{}
```
tombuildsstuff, line 18: can we add a Requires Import test here?

azurerm/internal/services/network/parse/express_route_circuit_connection.go
```diff
@@ -0,0 +1,56 @@
...
+		SubscriptionId: subscriptionId,
+		ResourceGroup:  resourcegroup,
+		CircuitName:    circuitname,
+		PeeringName:    peeringname,
+		Name:           name,
+	}
+}
+
+func (id NetworkExpressRouteCircuitConnectionId) ID() string {
+	fmtString := "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/expressRouteCircuits/%s/peerings/%s/connections/%s"
```
tombuildsstuff, line 28: these should be generated using `go:generate`: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/network/resourceids.go#L4

azurerm/internal/services/network/parse/express_route_circuit_connection_test.go
```diff
@@ -0,0 +1,129 @@
+package parse
```
tombuildsstuff, line 1: (as above) this should be generated

website/docs/d/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,47 @@
+---
```
tombuildsstuff, line 1: this file name is wrong?

website/docs/d/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+subcategory: "Network"
+layout: "azurerm"
+page_title: "Azure Resource Manager: Data Source: azurerm_express_route_circuit_connection"
+description: |-
+  Gets information about an existing network ExpressRouteCircuitConnection.
```
tombuildsstuff, line 6: can we add a proper description here

website/docs/d/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+---
+
+# Data Source: azurerm_express_route_circuit_connection
+
+Use this data source to access information about an existing network ExpressRouteCircuitConnection.
```
tombuildsstuff, line 11: can we add a proper description here

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
+---
+subcategory: "Network"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_express_route_circuit_connection"
+description: |-
+  Manages a network ExpressRouteCircuitConnection.
```
tombuildsstuff, line 6: can we add a proper description here

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+---
+
+# azurerm_express_route_circuit_connection
+
+Manages a network ExpressRouteCircuitConnection.
```
tombuildsstuff, line 11: can we add a proper description here

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
+---
```
tombuildsstuff, line 1: this filename is named incorrectly

website/docs/d/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+In addition to the Arguments listed above - the following Attributes are exported:
+
+* `id` - The ID of the network ExpressRouteCircuitConnection.
+
+## Timeouts
+
+The `timeouts` block allows you to specify [timeouts](https://www.terraform.io/docs/configuration/resources.html#timeouts) for certain actions:
+
+* `read` - (Defaults to 5 minutes) Used when retrieving the network ExpressRouteCircuitConnection.
```
tombuildsstuff, line 47: can we add a proper description here

website/docs/d/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+* `resource_group_name` - (Required) The name of the Resource Group where the network ExpressRouteCircuitConnection exists.
+
+* `circuit_name` - (Required) The name of the express route circuit.
+
+## Attributes Reference
+
+In addition to the Arguments listed above - the following Attributes are exported:
+
+* `id` - The ID of the network ExpressRouteCircuitConnection.
```
tombuildsstuff, line 41: can we add a proper description here

website/docs/d/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of this network ExpressRouteCircuitConnection.
+
+* `resource_group_name` - (Required) The name of the Resource Group where the network ExpressRouteCircuitConnection exists.
```
tombuildsstuff, line 33: can we add a proper description here

website/docs/d/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+output "id" {
+  value = data.azurerm_express_route_circuit_connection.example.id
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of this network ExpressRouteCircuitConnection.
```
tombuildsstuff, line 31: can we add a proper description here

website/docs/d/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+* `resource_group_name` - (Required) The name of the Resource Group where the network ExpressRouteCircuitConnection exists.
+
+* `circuit_name` - (Required) The name of the express route circuit.
```
tombuildsstuff, line 35: can we add a proper description here

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+    address_prefix = "2002:db01::/125"
+  }
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name which should be used for this network ExpressRouteCircuitConnection. Changing this forces a new network ExpressRouteCircuitConnection to be created.
```
tombuildsstuff, line 98: can we add a proper description here

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+
+* `resource_group_name` - (Required) The name of the Resource Group where the network ExpressRouteCircuitConnection should exist. Changing this forces a new network ExpressRouteCircuitConnection to be created.
```
tombuildsstuff, line 100: can we add a proper description here

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+
+* `circuit_name` - (Required) The name of the express route circuit. Changing this forces a new network ExpressRouteCircuitConnection to be created.
+  
+* `peering_id` - (Required) The ID of the Express Route Circuit Private Peering Resource of the circuit initiating connection.
```
tombuildsstuff, line 104: can we add a proper description here - in addition there's no need to reference `Resource` here

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+
+---
+
+* `authorization_key` - (Optional) The authorization key.
+
+* `ipv6circuit_connection_config` - (Optional) A `ipv6circuit_connection_config` block as defined below.
+
+---
+
+An `ipv6circuit_connection_config` block exports the following:
```
tombuildsstuff, line 118: this isn't exported, but user-configurable

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+
+* `address_prefix` - (Optional) /125 IP address space to carve out customer addresses for global reach.
+
+## Attributes Reference
+
+In addition to the Arguments listed above - the following Attributes are exported:
+
+* `id` - The ID of the network ExpressRouteCircuitConnection.
```
tombuildsstuff, line 126: can we add a proper description here?

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+
+* `authorization_key` - (Optional) The authorization key.
+
+* `ipv6circuit_connection_config` - (Optional) A `ipv6circuit_connection_config` block as defined below.
+
+---
+
+An `ipv6circuit_connection_config` block exports the following:
+
+* `address_prefix` - (Optional) /125 IP address space to carve out customer addresses for global reach.
```
tombuildsstuff, line 120: can we add a proper description here?

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+
+In addition to the Arguments listed above - the following Attributes are exported:
+
+* `id` - The ID of the network ExpressRouteCircuitConnection.
+
+## Timeouts
+
+The `timeouts` block allows you to specify [timeouts](https://www.terraform.io/docs/configuration/resources.html#timeouts) for certain actions:
+
+* `create` - (Defaults to 30 minutes) Used when creating the network ExpressRouteCircuitConnection.
```
tombuildsstuff, line 132: can we add a proper description here?

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+* `read` - (Defaults to 5 minutes) Used when retrieving the network ExpressRouteCircuitConnection.
```
tombuildsstuff, line 133: can we add a proper description here?

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+* `delete` - (Defaults to 30 minutes) Used when deleting the network ExpressRouteCircuitConnection.
```
tombuildsstuff, line 134: can we add a proper description here?

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+
+## Import
+
+network ExpressRouteCircuitConnections can be imported using the `resource id`, e.g.
```
tombuildsstuff, line 138: can we add a proper description here?

tombuildsstuff (changes requested):
hi @xuzhang3 

Thanks for this PR.

I've taken a look through and left some comments inline, but this is off to a good start - unfortunately it appears there's several design changes needed and the documentation needs to be clearer.

Thanks!

azurerm/internal/services/network/express_route_circuit_connections_data_source.go
```diff
@@ -0,0 +1,66 @@
...
+		if utils.ResponseWasNotFound(resp.Response) {
+			return fmt.Errorf(" ExpressRouteCircuitConnection %q (Resource Group %q / circuitName %q) does not exist", name, resourceGroup, circuitName)
+		}
+		return fmt.Errorf("retrieving ExpressRouteCircuitConnection %q (Resource Group %q / circuitName %q): %+v", name, resourceGroup, circuitName, err)
+	}
+	if resp.ID == nil || *resp.ID == "" {
+		return fmt.Errorf("empty or nil ID returned for ExpressRouteCircuitConnection %q (Resource Group %q / circuitName %q) ID", name, resourceGroup, circuitName)
+	}
+
+	d.SetId(*resp.ID)
```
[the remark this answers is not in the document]
neil-yechenwei: I think we have no particular reason for adding data source. So I removed this data source file.

azurerm/internal/services/network/parse/express_route_circuit_connection.go
neil-yechenwei: Updated

azurerm/internal/services/network/parse/express_route_circuit_connection_test.go
neil-yechenwei: Updated

website/docs/d/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+			},
+
+			"resource_group_name": azure.SchemaResourceGroupName(),
+
+			"circuit_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
```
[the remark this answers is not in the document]
neil-yechenwei: `circuit_name` and `peering_id` are based on the same express route circuit. I assume `circuit_name` could be extracted from `express_route_circuit_peering_id`. So I removed `circuit_name` and updated the logic to extract circuit name from `express_route_circuit_peering_id`.

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+
+			"peering_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: azure.ValidateResourceID,
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+			},
+
+			"peer_peering_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: azure.ValidateResourceID,
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+	subscriptionId := meta.(*clients.Client).Account.SubscriptionId
+	client := meta.(*clients.Client).Network.ExpressRouteCircuitConnectionClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	name := d.Get("name").(string)
+	resourceGroup := d.Get("resource_group_name").(string)
+	circuitName := d.Get("circuit_name").(string)
+
+	id := parse.NewExpressRouteCircuitConnectionID(subscriptionId, resourceGroup, circuitName, "AzurePrivatePeering", name).ID()
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+	if d.IsNewResource() {
+		existing, err := client.Get(ctx, resourceGroup, circuitName, "AzurePrivatePeering", name)
+		if err != nil {
+			if !utils.ResponseWasNotFound(existing.Response) {
+				return fmt.Errorf("checking for existing ExpressRouteCircuitConnection %q (Resource Group %q / circuitName %q): %+v", name, resourceGroup, circuitName, err)
+			}
+		}
+		if !utils.ResponseWasNotFound(existing.Response) {
+			return tf.ImportAsExistsError("azurerm_express_route_circuit_connection", id)
+		}
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	expressRouteCircuitConnectionParameters := network.ExpressRouteCircuitConnection{
+		Name: utils.String(d.Get("name").(string)),
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	future, err := client.CreateOrUpdate(ctx, resourceGroup, circuitName, "AzurePrivatePeering", name, expressRouteCircuitConnectionParameters)
+	if err != nil {
+		return fmt.Errorf("creating/updating ExpressRouteCircuitConnection %q (Resource Group %q / circuitName %q): %+v", name, resourceGroup, circuitName, err)
+	}
+
+	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("waiting for creation/update of the ExpressRouteCircuitConnection %q (Resource Group %q / circuitName %q): %+v", name, resourceGroup, circuitName, err)
+	}
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource.go
neil-yechenwei: There is no Update operation for this RP. So seems we cannot split the Create and Update methods.

azurerm/internal/services/network/express_route_circuit_connections_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource_test.go
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/d/express_route_circuit_connections.html.markdown
neil-yechenwei: I think we don't need data source for this RP. So I removed this file.

website/docs/d/express_route_circuit_connections.html.markdown
neil-yechenwei: I think we don't need data source for this RP. So I removed this file.

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/d/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/d/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/d/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/d/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/d/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name which should be used for this network ExpressRouteCircuitConnection. Changing this forces a new network ExpressRouteCircuitConnection to be created.
+
+* `resource_group_name` - (Required) The name of the Resource Group where the network ExpressRouteCircuitConnection should exist. Changing this forces a new network ExpressRouteCircuitConnection to be created.
+
+* `circuit_name` - (Required) The name of the express route circuit. Changing this forces a new network ExpressRouteCircuitConnection to be created.
```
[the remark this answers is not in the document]
neil-yechenwei: Removed this property.

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+  
+* `peering_id` - (Required) The ID of the Express Route Circuit Private Peering Resource of the circuit initiating connection.
+  
+* `peer_peering_id` - (Required) The ID of the Express Route Circuit Private Peering Resource of the peered circuit.
+
+* `address_prefix` - (Required) /29 IP address space to carve out Customer addresses for tunnels.
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
```diff
@@ -0,0 +1,142 @@
...
+
+* `name` - (Required) The name which should be used for this network ExpressRouteCircuitConnection. Changing this forces a new network ExpressRouteCircuitConnection to be created.
+
+* `resource_group_name` - (Required) The name of the Resource Group where the network ExpressRouteCircuitConnection should exist. Changing this forces a new network ExpressRouteCircuitConnection to be created.
+
+* `circuit_name` - (Required) The name of the express route circuit. Changing this forces a new network ExpressRouteCircuitConnection to be created.
+  
+* `peering_id` - (Required) The ID of the Express Route Circuit Private Peering Resource of the circuit initiating connection.
+  
+* `peer_peering_id` - (Required) The ID of the Express Route Circuit Private Peering Resource of the peered circuit.
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connections_resource_test.go
neil-yechenwei: Seems already included _requiresImport tc.

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Removed it since it has been updated to "address_prefex_ipv6"

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connections.html.markdown
neil-yechenwei: Updated

neil-yechenwei:
@tombuildsstuff , thanks for your comments. I've updated code. Please have an another look. Thanks.

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+		return fmt.Errorf("retrieving %s: %+v", *id, err)
+	}
+
+	d.Set("name", id.ConnectionName)
+	d.Set("peering_id", parse.NewExpressRouteCircuitPeeringID(id.SubscriptionId, id.ResourceGroup, id.ExpressRouteCircuitName, "AzurePrivatePeering").ID())
+
+	if resp.PeerExpressRouteCircuitPeering != nil && resp.PeerExpressRouteCircuitPeering.ID != nil {
+		d.Set("peer_peering_id", resp.PeerExpressRouteCircuitPeering.ID)
+	}
```
jackofallops, line 170: This should be moved inside the nil checked `props` below as it's a child object of props also?

jackofallops (changes requested):
hi @neil-yechenwei / @xuzhang3 - Thanks for the continued work on this. Following up on previous reviews I've made some additional comments below.

website/docs/r/express_route_circuit_connection.html.markdown
```diff
@@ -0,0 +1,122 @@
...
+  
+* `peer_peering_id` - (Required) The ID of the peered Express Route Circuit Private Peering. Changing this forces a new Express Route Circuit Connection to be created.
+  
+* `address_prefix_ipv4` - (Required) The IPv4 address space to carve out customer addresses for tunnels.
+
+---
+
+* `authorization_key` - (Optional) The authorization key which is associated with the Express Route Circuit Connection.
+
+* `address_prefix_ipv6` - (Required) The IPv6 address space to carve out customer addresses for global reach.
```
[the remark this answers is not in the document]
neil-yechenwei: It should be optional property.

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,211 @@
...
+	client := meta.(*clients.Client).Network.ExpressRouteCircuitConnectionClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	expressRouteCircuitPeeringId, err := parse.ExpressRouteCircuitPeeringID(d.Get("peering_id").(string))
+	if err != nil {
+		return err
+	}
+
+	id := parse.NewExpressRouteCircuitConnectionID(expressRouteCircuitPeeringId.SubscriptionId, expressRouteCircuitPeeringId.ResourceGroup, expressRouteCircuitPeeringId.ExpressRouteCircuitName, "AzurePrivatePeering", d.Get("name").(string))
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,211 @@
...
+		if err != nil {
+			if !utils.ResponseWasNotFound(existing.Response) {
+				return fmt.Errorf("checking for existing %s: %+v", id, err)
+			}
+		}
+
+		if !utils.ResponseWasNotFound(existing.Response) {
+			return tf.ImportAsExistsError("azurerm_express_route_circuit_connection", id.ID())
+		}
+	}
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

neil-yechenwei:
@jackofallops , thanks for your comment. I've updated test code. Please have an another look. Thanks.

website/docs/r/express_route_circuit_connection.html.markdown
```diff
@@ -0,0 +1,122 @@
...
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name which should be used for this Express Route Circuit Connection. Changing this forces a new Express Route Circuit Connection to be created.
+
+* `peering_id` - (Required) The ID of the Express Route Circuit Private Peering within which this Express Route Circuit Connection should be created. Changing this forces a new Express Route Circuit Connection to be created.
```
magodo, line 89:
```suggestion
* `peering_id` - (Required) The ID of the Express Route Circuit Private Peering that this Express Route Circuit Connection connects with. Changing this forces a new Express Route Circuit Connection to be created.
```

website/docs/r/express_route_circuit_connection.html.markdown
```diff
@@ -0,0 +1,122 @@
...
+  
+* `peer_peering_id` - (Required) The ID of the peered Express Route Circuit Private Peering. Changing this forces a new Express Route Circuit Connection to be created.
+  
+* `address_prefix_ipv4` - (Required) The IPv4 address space to carve out customer addresses for tunnels.
```
magodo, line 93: Can we use the same tone as `address_prefix_ipv6`?

```suggestion
* `address_prefix_ipv4` - (Required) The IPv4 address space from which to allocate customer address for global reach.
```

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,254 @@
...
+	client := meta.(*clients.Client).Network.ExpressRouteCircuitConnectionClient
+	ctx, cancel := timeouts.ForCreate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	expressRouteCircuitPeeringId, err := parse.ExpressRouteCircuitPeeringID(d.Get("peering_id").(string))
+	if err != nil {
+		return err
+	}
+
+	id := parse.NewExpressRouteCircuitConnectionID(expressRouteCircuitPeeringId.SubscriptionId, expressRouteCircuitPeeringId.ResourceGroup, expressRouteCircuitPeeringId.ExpressRouteCircuitName, string(network.ExpressRoutePeeringTypeAzurePrivatePeering), d.Get("name").(string))
```
magodo, line 93: Is there any reason to use the enum here rather than using the `expressRouteCircuitPeeringId.Name`?

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,254 @@
...
+
+	existing, err := client.Get(ctx, id.ResourceGroup, id.ExpressRouteCircuitName, string(network.ExpressRoutePeeringTypeAzurePrivatePeering), id.ConnectionName)
```
magodo, line 95: Is there any reason not to use the `id.PeeringName` here?

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,254 @@
...
+
+	expressRouteCircuitConnectionParameters := network.ExpressRouteCircuitConnection{
+		Name: utils.String(id.ConnectionName),
+		ExpressRouteCircuitConnectionPropertiesFormat: &network.ExpressRouteCircuitConnectionPropertiesFormat{
+			AddressPrefix: utils.String(d.Get("address_prefix_ipv4").(string)),
+			ExpressRouteCircuitPeering: &network.SubResource{
+				ID: utils.String(expressRouteCircuitPeeringId.ID()),
+			},
+			PeerExpressRouteCircuitPeering: &network.SubResource{
+				ID: utils.String(d.Get("peer_peering_id").(string)),
```
magodo, line 114: Can we normalize the `peer_peering_id` here as is done for the `peering_id`?

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,254 @@
...
+		expressRouteCircuitConnectionParameters.ExpressRouteCircuitConnectionPropertiesFormat.AuthorizationKey = utils.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("address_prefix_ipv6"); ok {
+		expressRouteCircuitConnectionParameters.ExpressRouteCircuitConnectionPropertiesFormat.Ipv6CircuitConnectionConfig = &network.Ipv6CircuitConnectionConfig{
+			AddressPrefix: utils.String(v.(string)),
+		}
+	}
+
+	future, err := client.CreateOrUpdate(ctx, id.ResourceGroup, id.ExpressRouteCircuitName, string(network.ExpressRoutePeeringTypeAzurePrivatePeering), id.ConnectionName, expressRouteCircuitConnectionParameters)
```
magodo, line 129: Can we use the `id.PeeringName` instead here?

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,254 @@
...
+		if utils.ResponseWasNotFound(resp.Response) {
+			log.Printf("[INFO] %s does not exist - removing from state", *id)
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("retrieving %s: %+v", *id, err)
+	}
+
+	d.Set("name", id.ConnectionName)
+	d.Set("peering_id", parse.NewExpressRouteCircuitPeeringID(id.SubscriptionId, id.ResourceGroup, id.ExpressRouteCircuitName, string(network.ExpressRoutePeeringTypeAzurePrivatePeering)).ID())
```
magodo, line 164: Can we use the id.PeeringName instead here?

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+
+		if props.Ipv6CircuitConnectionConfig != nil && props.Ipv6CircuitConnectionConfig.AddressPrefix != nil {
+			d.Set("address_prefix_ipv6", props.Ipv6CircuitConnectionConfig.AddressPrefix)
+		} else {
+			d.Set("address_prefix_ipv6", "")
+		}
+
+		if props.PeerExpressRouteCircuitPeering != nil && props.PeerExpressRouteCircuitPeering.ID != nil {
+			d.Set("peer_peering_id", resp.PeerExpressRouteCircuitPeering.ID)
```
magodo, line 182: Can we normalize the id here?

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,254 @@
...
+	d.Set("peering_id", parse.NewExpressRouteCircuitPeeringID(id.SubscriptionId, id.ResourceGroup, id.ExpressRouteCircuitName, string(network.ExpressRoutePeeringTypeAzurePrivatePeering)).ID())
+
+	if props := resp.ExpressRouteCircuitConnectionPropertiesFormat; props != nil {
+		d.Set("address_prefix_ipv4", props.AddressPrefix)
+
+		if props.AuthorizationKey != nil {
+			d.Set("authorization_key", props.AuthorizationKey)
+		} else {
+			d.Set("authorization_key", "")
+		}
```
magodo, line 173: This can be simplified to be:

```suggestion
                d.Set("authorization_key", props.AuthorizationKey)
```

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,254 @@
...
+
+		if props.Ipv6CircuitConnectionConfig != nil && props.Ipv6CircuitConnectionConfig.AddressPrefix != nil {
+			d.Set("address_prefix_ipv6", props.Ipv6CircuitConnectionConfig.AddressPrefix)
+		} else {
+			d.Set("address_prefix_ipv6", "")
+		}
```
magodo, line 179: Can be simplified to be:

```suggestion
		if props.Ipv6CircuitConnectionConfig != nil {
			d.Set("address_prefix_ipv6", props.Ipv6CircuitConnectionConfig.AddressPrefix)
		}
```

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,254 @@
...
+
+	expressRouteCircuitConnectionParameters := network.ExpressRouteCircuitConnection{
+		Name: utils.String(id.ConnectionName),
+		ExpressRouteCircuitConnectionPropertiesFormat: &network.ExpressRouteCircuitConnectionPropertiesFormat{
+			AddressPrefix: utils.String(d.Get("address_prefix_ipv4").(string)),
+			ExpressRouteCircuitPeering: &network.SubResource{
+				ID: utils.String(d.Get("peering_id").(string)),
+			},
+			PeerExpressRouteCircuitPeering: &network.SubResource{
+				ID: utils.String(d.Get("peer_peering_id").(string)),
```
magodo, line 207: Can we normalize the ID here?

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,254 @@
...
+		expressRouteCircuitConnectionParameters.ExpressRouteCircuitConnectionPropertiesFormat.AuthorizationKey = utils.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("address_prefix_ipv6"); ok {
+		expressRouteCircuitConnectionParameters.ExpressRouteCircuitConnectionPropertiesFormat.Ipv6CircuitConnectionConfig = &network.Ipv6CircuitConnectionConfig{
+			AddressPrefix: utils.String(v.(string)),
+		}
+	}
+
+	future, err := client.CreateOrUpdate(ctx, id.ResourceGroup, id.ExpressRouteCircuitName, string(network.ExpressRoutePeeringTypeAzurePrivatePeering), id.ConnectionName, expressRouteCircuitConnectionParameters)
```
magodo, line 222: Can we use `id.PeeringName` here?

azurerm/internal/services/network/express_route_circuit_connection_resource_test.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+type ExpressRouteCircuitConnectionResource struct{}
+
+func TestAccExpressRouteCircuitConnection_basic(t *testing.T) {
+	if os.Getenv("ARM_TEST_DATA_RESOURCE_GROUP") == "" || os.Getenv("ARM_TEST_CIRCUIT_NAME_FIRST") == "" || os.Getenv("ARM_TEST_CIRCUIT_NAME_SECOND") == "" {
+		t.Skip("Skipping as ARM_TEST_DATA_RESOURCE_GROUP and/or ARM_TEST_CIRCUIT_NAME_FIRST and/or ARM_TEST_CIRCUIT_NAME_SECOND are not specified")
+		return
+	}
```
magodo, line 24: Can we abstract this into a function and reuse it in each place?

azurerm/internal/services/network/express_route_circuit_connection_resource_test.go
```diff
@@ -0,0 +1,244 @@
...
+	return fmt.Sprintf(`
+provider "azurerm" {
+  features {}
+}
+
+resource "azurerm_express_route_circuit_peering" "test" {
+  peering_type                  = "AzurePrivatePeering"
+  express_route_circuit_name    = "%[1]s"
+  resource_group_name           = "%[2]s"
+  shared_key                    = "SSSSsssssshhhhhItsASecret"
```
magodo, line 226: Can we pick another less offensive key?

azurerm/internal/services/network/express_route_circuit_connection_resource_test.go
```diff
@@ -0,0 +1,244 @@
...
+  primary_peer_address_prefix   = "192.168.1.0/30"
+  secondary_peer_address_prefix = "192.168.1.0/30"
+  vlan_id                       = 100
+}
+
+resource "azurerm_express_route_circuit_peering" "peer_test" {
+  peering_type                  = "AzurePrivatePeering"
+  express_route_circuit_name    = "%[3]s"
+  resource_group_name           = "%[2]s"
+  shared_key                    = "SSSSsssssshhhhhItsASecret"
```
magodo, line 237: Can we pick another less offensive key?

magodo (changes requested):
@neil-yechenwei I've taken a look through and left some comments inline, but this is mostly looking good to me 👍

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,254 @@
+package network
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-11-01/network"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connection.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_circuit_connection.html.markdown
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource_test.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource_test.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource_test.go
neil-yechenwei: Updated

neil-yechenwei:
@magodo , thanks for your comments. I've updated code. Please have an another look. Thanks.

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,272 @@
+package network
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-11-01/network"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
```
magodo, line 9: This is wants to be also replaced by pluginsdk

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,272 @@
...
+	d.Set("peering_id", parse.NewExpressRouteCircuitPeeringID(id.SubscriptionId, id.ResourceGroup, id.ExpressRouteCircuitName, id.PeeringName).ID())
+
+	if props := resp.ExpressRouteCircuitConnectionPropertiesFormat; props != nil {
+		d.Set("address_prefix_ipv4", props.AddressPrefix)
+
+		authorizationKey := ""
+		if props.AuthorizationKey != nil {
+			authorizationKey = *props.AuthorizationKey
+		}
+		d.Set("authorization_key", authorizationKey)
```
magodo, line 177: Can be simplified to:

```suggestion
		d.Set("authorization_key", props.AuthorizationKey)
```

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -0,0 +1,272 @@
...
+
+		addressPrefixIPv6 := ""
+		if props.Ipv6CircuitConnectionConfig != nil && props.Ipv6CircuitConnectionConfig.AddressPrefix != nil {
+			addressPrefixIPv6 = *props.Ipv6CircuitConnectionConfig.AddressPrefix
+		}
+		d.Set("address_prefix_ipv6", addressPrefixIPv6)
```
magodo, line 183: Can be simplified to:

```suggestion
		if props.Ipv6CircuitConnectionConfig != nil  {
			d.Set("address_prefix_ipv6", props.Ipv6CircuitConnectionConfig.AddressPrefix )
		}
```

azurerm/internal/services/network/express_route_circuit_connection_resource_test.go
```diff
@@ -0,0 +1,245 @@
...
+  shared_key                    = "ItsASecret"
+  peer_asn                      = 100
+  primary_peer_address_prefix   = "192.168.1.0/30"
+  secondary_peer_address_prefix = "192.168.1.0/30"
+  vlan_id                       = 100
+}
+`, circuitName1, rg, circuitName2)
+}
+
+func skipExpressRouteCircuitConnection() bool {
```
magodo, line 239: Can we pass in the `t` and call t.Skip() inside this function rather than checking outside?

magodo (changes requested):
Thank you for updating this, I've added some other comments, desipite which, it LGTM!

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: @magodo , we cannot do this because we cannot set it as empty string when Ipv6CircuitConnectionConfig  is nil.

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource_test.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
magodo: That's true. Thank you!

magodo: Then can we also ensure to set the peering id in the following code?

neil-yechenwei: Sure. Updated

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -177,13 +177,11 @@ func resourceExpressRouteCircuitConnectionRead(d *pluginsdk.ResourceData, meta i
 		}
 		d.Set("address_prefix_ipv6", addressPrefixIPv6)
 
-		if props.PeerExpressRouteCircuitPeering != nil && props.PeerExpressRouteCircuitPeering.ID != nil {
-			circuitPeerPeeringId, err := parse.ExpressRouteCircuitPeeringID(*props.PeerExpressRouteCircuitPeering.ID)
-			if err != nil {
-				return err
-			}
-			d.Set("peer_peering_id", circuitPeerPeeringId.ID())
+		circuitPeerPeeringId, err := parse.ExpressRouteCircuitPeeringID(*props.PeerExpressRouteCircuitPeering.ID)
```
magodo, line 180: Wondering why the omit the nil check here?

neil-yechenwei: Sorry. I didn't get your point. https://github.com/terraform-providers/terraform-provider-azurerm/pull/11303#discussion_r654242616

magodo:
@neil-yechenwei The test on TC failed with:

image

Is there anything else needed to be setup in order to create the ERP in the specified location?

neil-yechenwei:
@magodo , did you see this error before? I think you have knowledge about creating express route port.

neil-yechenwei:
@magodo , seems it's subscription limitation. So I've updated test code to run tc in parallel. Now all TCs are passed.
image

azurerm/internal/services/network/express_route_circuit_connection_resource.go
```diff
@@ -186,8 +186,10 @@ func resourceExpressRouteCircuitConnectionRead(d *pluginsdk.ResourceData, meta i
 	if props := resp.ExpressRouteCircuitConnectionPropertiesFormat; props != nil {
 		d.Set("address_prefix_ipv4", props.AddressPrefix)
 
+		// The ExpressRoute Circuit Connection API returns "*****************" for AuthorizationKey when it's changed from a valid value to `nil`
+		// See more details from https://github.com/Azure/azure-rest-api-specs/issues/15030
 		authorizationKey := ""
-		if props.AuthorizationKey != nil {
+		if props.AuthorizationKey != nil && *props.AuthorizationKey != "*****************" {
```
magodo, line 189: Why don't we ignore those invalid value to be specified at the first place?

neil-yechenwei: We cannot ignore it at the beginning. Because "*****************" is returned when authorizationKey is changed from "xxxxx" to nil.

Scenario #1:
when authorizationKey isn't specified in tf, api returns nil

Scenario #2:
When authorizationKey is specified in tf, api returns "xxxxx"

Scenario #3:
When authorizationKey is changed from "xxxx" to nil in tf, api returns "*****************"

PR #11307, opened 2021-04-13
 new resource "azurerm_spring_cloud_app_cosmosdb_association"

njuCZ:
cosmosdb supports 5 different sub type association. I combined them in one resource, please check whether it's ok or do I need to split into different sub resource.

image

### DIFF
--- a/website/docs/r/spring_cloud_app_cosmosdb_association.html.markdown
+++ b/website/docs/r/spring_cloud_app_cosmosdb_association.html.markdown
@@ -1,0 +1,104 @@
+---
+subcategory: "Spring Cloud"
+layout: "azurerm"
+page_title: "Azure Resource Manager: azurerm_spring_cloud_app_cosmosdb_association"
+description: |-
+  Associates a [Spring Cloud Application](spring_cloud_app.html) with a [CosmosDB Account](cosmosdb_account.html).
+---
+
+# azurerm_spring_cloud_app_cosmosdb_association
+
+Associates a [Spring Cloud Application](spring_cloud_app.html) with a [CosmosDB Account](cosmosdb_account.html).
+
+## Example Usage
+
+```hcl
+resource "azurerm_resource_group" "example" {
+  name     = "example-resources"
+  location = "West Europe"
+}
+
+resource "azurerm_spring_cloud_service" "example" {
+  name                = "example-springcloud"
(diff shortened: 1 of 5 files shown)


website/docs/r/spring_cloud_app_cosmosdb_association.html.markdown
```diff
@@ -0,0 +1,104 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) Specifies the name of the Spring Cloud Application Association. Changing this forces a new resource to be created.
+
+* `spring_cloud_app_id` - (Required) Specifies the ID of the Spring Cloud Application where this Association is created. Changing this forces a new resource to be created.
+
+* `cosmosdb_account_id` - (Required) Specifies the ID of the CosmosDB Account. Changing this forces a new resource to be created.
+
+* `api_type` - (Required) Specifies the api type which should be used when connecting to the CosmosDB Account. Possible values are `cassandra`, `gremlin`, `mongo`, `sql` or `table`. Changing this forces a new resource to be created.
```
katbyte, line 69: couldn't we determine this from the supplied `cosmosdb_x` property? and then remove this

katbyte (changes requested):
Thanks @njuCZ - aside from one comment about better ux this is looking good

website/docs/r/spring_cloud_app_cosmosdb_association.html.markdown
njuCZ: yes, we could do that. I am just afraid that `table` type does not need to supply those optional property. It may be not easy for users to understand the difference

katbyte: does it make sense to even allow specifying type without the cosmosdb property? and the api doesn't error when they are all left empty?

njuCZ: for all api type, cosmosDB account is required. For api type `table`, no more property is needed. For the rest api type, Additional property needs to be provided

katbyte: ahh that makes sense - thanks

katbyte (approved):
LGTM 👍

PR #11317, opened 2021-04-13
Ensures storage profile is included with storage_image_id or storage_image_reference updates

### LINKED ISSUE #8273: VMSS scale set inconsistent storage_account_type and non-idempotent behavior
We are using TF azure provider and tested with versions 2.17 to create VMSS sets used to back a kubernetes deployment. After bootstrapping the VMSS's , and without any changes to the TF code, using plan sometimes result in the following proposed changed:
~ storage_account_type      = "Premium_LRS" -> "Standard_LRS" # forces replacement
For the os disk , a slew of other related changes follow.
When examining this further, it appears that the initial apply can result in an inconsistent configuration of the storage_account_type where some VMSS's are created with Premium_LRS while others adhere to the specified Standard_LRS in the code.

The following snippet is creates the VMSS sets:
```
resource "azurerm_linux_virtual_machine_scale_set" "dp_master_scale" {
  count = length(var.zones)

  name                = "${var.cluster_name}-master-${var.zones[count.index]}-scale"
  location            = azurerm_resource_group.k8s.location
  resource_group_name = azurerm_resource_group.k8s.name
  zones               = ["${var.zones[count.index]}"]

  sku                 = var.master_instance_type
  instances           = 0

  identity {
    type         = "SystemAssigned, UserAssigned"
    identity_ids = [azurerm_user_assigned_identity.master_id.id]
  }

  lifecycle {
    ignore_changes = [
      instances,
    ]
  }

  source_image_id = var.here_dp_image_id

  os_disk {
  }

  data_disk {
    lun                  = 0
    disk_size_gb         = 50
  }
```
Additionally, we have seen this appear with 'null image reference' , as well as proposing the data disks as new '+' additions. So it appears that there is a partial application or failure when creating the vmss and configuring its storage profile.

Azure template exports for affected VMSS shows the following:
`
"storageProfile": {
                        "osDisk": {
                            "createOption": "FromImage",
                            "caching": "ReadWrite",
                            "managedDisk": {
                                "storageAccountType": "Premium_LRS"
                            },
                            "diskSizeGB": 30
                        },
                        "imageReference": {
                            "id": "[concat(parameters('galleries_dpimages_externalid'), '/images/privatename/versions/privateversion')]"
                        }
                    },
`
as well as
`
osDisk": {
                        "osType": "Linux",
                        "name": "some-private-prefix-eOS__1_dasdf43488347239289284",
                        "createOption": "FromImage",
                        "caching": "ReadWrite",
                        "writeAcceleratorEnabled": false,
                        "managedDisk": {
                            "storageAccountType": "Standard_LRS",
                            "id": "[parameters('disks_thesameprivateprefix_eOS__1_thesameuuidhashlikeasabove_externalid')]"
                        },
                        "diskSizeGB": 30
                    }, 
`
Note how the same disk appears with different storage types in profile vs osdisk serializations

I think this is similar/related to the following issues:
#https://github.com/Azure/azure-rest-api-specs/issues/10373
#https://github.com/terraform-providers/terraform-provider-azurerm/issues/6902

The impact is severe:

- We can't rely on TF to create our infrastructure correctly on bootstrap
- Updates may introduce catastrophic destructions/changes and will require human intervention and further scrutiny.

Please let me know if you have any further questions or any suggestions to collect more information or try work arounds.

torrick:
fixes #8273

Also fixes some terraform in the acceptance tests.

### DIFF
--- a/azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
@@ -702,13 +702,23 @@ func resourceLinuxVirtualMachineScaleSetUpdate(d *schema.ResourceData, meta inte
 		}
 
 		if d.HasChange("source_image_id") || d.HasChange("source_image_reference") {
+			// Must include all storage profile properties when updating disk image.  See: https://github.com/terraform-providers/terraform-provider-azurerm/issues/8273
+			ultraSSDEnabled := d.Get("additional_capabilities.0.ultra_ssd_enabled").(bool)
+			dataDisks, _ := ExpandVirtualMachineScaleSetDataDisk(d.Get("data_disk").([]interface{}), ultraSSDEnabled)
+			if err != nil {
+				return fmt.Errorf("expanding `data_disk`: %+v", err)
+			}
+
+			osDiskRaw := d.Get("os_disk").([]interface{})
+
 			sourceImageReferenceRaw := d.Get("source_image_reference").([]interface{})
 			sourceImageId := d.Get("source_image_id").(string)
 			sourceImageReference, err := expandSourceImageReference(sourceImageReferenceRaw, sourceImageId)
 			if err != nil {
 				return err
 			}
-
+			updateProps.VirtualMachineProfile.StorageProfile.DataDisks = dataDisks
+			updateProps.VirtualMachineProfile.StorageProfile.OsDisk = ExpandVirtualMachineScaleSetOSDiskUpdate(osDiskRaw)
 			updateProps.VirtualMachineProfile.StorageProfile.ImageReference = sourceImageReference
 		}
 	}
--- a/azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
+++ b/azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
@@ -702,24 +702,24 @@ func resourceLinuxVirtualMachineScaleSetUpdate(d *schema.ResourceData, meta inte
 		}
 
 		if d.HasChange("source_image_id") || d.HasChange("source_image_reference") {
-			// Must include all storage profile properties when updating disk image.  See: https://github.com/terraform-providers/terraform-provider-azurerm/issues/8273
-			ultraSSDEnabled := d.Get("additional_capabilities.0.ultra_ssd_enabled").(bool)
-			dataDisks, _ := ExpandVirtualMachineScaleSetDataDisk(d.Get("data_disk").([]interface{}), ultraSSDEnabled)
-			if err != nil {
-				return fmt.Errorf("expanding `data_disk`: %+v", err)
-			}
-
-			osDiskRaw := d.Get("os_disk").([]interface{})
-
 			sourceImageReferenceRaw := d.Get("source_image_reference").([]interface{})
 			sourceImageId := d.Get("source_image_id").(string)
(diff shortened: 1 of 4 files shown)


azurerm/internal/services/compute/windows_virtual_machine_scale_set_resource.go
```diff
@@ -732,13 +732,23 @@ func resourceWindowsVirtualMachineScaleSetUpdate(d *schema.ResourceData, meta in
...
+
 			sourceImageReferenceRaw := d.Get("source_image_reference").([]interface{})
 			sourceImageId := d.Get("source_image_id").(string)
 			sourceImageReference, err := expandSourceImageReference(sourceImageReferenceRaw, sourceImageId)
 			if err != nil {
 				return err
 			}
-
+			updateProps.VirtualMachineProfile.StorageProfile.DataDisks = dataDisks
+			updateProps.VirtualMachineProfile.StorageProfile.OsDisk = ExpandVirtualMachineScaleSetOSDiskUpdate(osDiskRaw)
```
tombuildsstuff, line 751: rather than pulling this from the config, since it's possible `ignore_changes` could be being used here, we should likely use the value from the API here instead, which is available via:

> existing.VirtualMachineScaleSetProperties.VirtualMachineProfile.StorageProfile.OSDisk
> existing.VirtualMachineScaleSetProperties.VirtualMachineProfile.StorageProfile.DataDisks

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
```diff
@@ -702,13 +702,23 @@ func resourceLinuxVirtualMachineScaleSetUpdate(d *schema.ResourceData, meta inte
...
 			sourceImageReferenceRaw := d.Get("source_image_reference").([]interface{})
 			sourceImageId := d.Get("source_image_id").(string)
 			sourceImageReference, err := expandSourceImageReference(sourceImageReferenceRaw, sourceImageId)
 			if err != nil {
 				return err
 			}
-
+			updateProps.VirtualMachineProfile.StorageProfile.DataDisks = dataDisks
+			updateProps.VirtualMachineProfile.StorageProfile.OsDisk = ExpandVirtualMachineScaleSetOSDiskUpdate(osDiskRaw)
 			updateProps.VirtualMachineProfile.StorageProfile.ImageReference = sourceImageReference
```
tombuildsstuff, line 722: rather than pulling this from the config, since it's possible `ignore_changes` could be being used here, we should likely use the value from the API here instead, which is available via:

> existing.VirtualMachineScaleSetProperties.VirtualMachineProfile.StorageProfile.OSDisk
> existing.VirtualMachineScaleSetProperties.VirtualMachineProfile.StorageProfile.DataDisks

tombuildsstuff (changes requested):
hey @torrick 

Thanks for this PR.

Taking a look through here whilst this looks good, since we're it's possible these fields are being ignored via external updates, rather than pulling these from the configuration - can we instead pull these values from the current state of the virtual machine scale set. That said if we can fix that (and the tests pass) then this should otherwise be good to go 👍 

Thanks!

azurerm/internal/services/compute/linux_virtual_machine_scale_set_resource.go
torrick: Thanks @tombuildsstuff -- I've incorporated the changes you requested.  Let me know if there's anything else I can add here. 👍

katbyte (approved):
Thanks @torrick - LGTM 👍

PR #11320, opened 2021-04-14
New resource: azurerm_express_route_connection

neil-yechenwei:
Currently there is no way to create express route connection. So it has to submit this PR to implement it.

Test constraints:
The existed Express Route Circuit with Provider provisioned , the provision status is controlled by [Service provider](https://docs.microsoft.com/en-us/azure/expressroute/expressroute-locations-providers#partners), can be provisioned by [send the service key with provider](https://docs.microsoft.com/en-us/azure/expressroute/expressroute-howto-circuit-portal-resource-manager#send-the-service-key-to-your-connectivity-provider-for-provisioning). This will cost money, the cost depends on the provider.

AccTests result:
```
--- PASS: TestAccExpressRouteConnection (9370.04s)
    --- PASS: TestAccExpressRouteConnection/Resource (9370.04s)
        --- PASS: TestAccExpressRouteConnection/Resource/basic (4830.20s)
        --- PASS: TestAccExpressRouteConnection/Resource/requiresImport (4539.84s)
        --- PASS: TestAccExpressRouteConnection/Resource/complete (5094.51s)
        --- PASS: TestAccExpressRouteConnection/Resource/update (4948.59s)
```

References:
1. [API Spec](https://github.com/Azure/azure-rest-api-specs/blob/ea5bc27ee9cadeb67767d774c82095be2420bcad/specification/network/resource-manager/Microsoft.Network/stable/2020-05-01/virtualWan.json#L3544)
2. [Overview](https://docs.microsoft.com/en-us/azure/expressroute/expressroute-faqs#what-are-the-benefits-of-using-expressroute-and-private-network-connections)


azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,340 @@
...
+
+									"route_table_ids": {
+										Type:     schema.TypeList,
+										Optional: true,
+										Computed: true,
+										Elem: &schema.Schema{
+											Type:         schema.TypeString,
+											ValidateFunc: validate.HubRouteTableID,
+										},
+									},
```
ArcturusZhang, line 110: We are having `route_table_ids` as a list and `labels` as a set, are there some particular reason that we treat these two differently? These two look like similar to each other - distinct and sequence irrelevant - if possible it is better to unify these two.

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,340 @@
...
+						},
+					},
+				},
+			},
+
+			"routing_weight": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				ValidateFunc: validation.IntBetween(0, 32000),
+			},
```
ArcturusZhang, line 122: I would like to ask on this attribute: is it the same in the request when this is `null` and when this is explicitly set to `0`?

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,340 @@
...
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/helpers/tf"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/clients"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/parse"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/validate"
+	azSchema "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tf/schema"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func resourceArmExpressRouteConnection() *schema.Resource {
```
ArcturusZhang, line 19: Could we remove the `Arm` in the name of this function?
```suggestion
func resourceExpressRouteConnection() *schema.Resource {
```

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,340 @@
...
+	return &schema.Resource{
+		Create: resourceArmExpressRouteConnectionCreateUpdate,
+		Read:   resourceArmExpressRouteConnectionRead,
+		Update: resourceArmExpressRouteConnectionCreateUpdate,
+		Delete: resourceArmExpressRouteConnectionDelete,
```
ArcturusZhang, line 24: Also these four and their definition
```suggestion
		Create: resourceExpressRouteConnectionCreateUpdate,
		Read:   resourceExpressRouteConnectionRead,
		Update: resourceExpressRouteConnectionCreateUpdate,
		Delete: resourceExpressRouteConnectionDelete,
```

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("waiting for deletion of %s: %+v", *id, err)
+	}
+
+	return nil
+}
+
+func expandArmExpressRouteConnectionRouting(input []interface{}) *network.RoutingConfiguration {
```
ArcturusZhang, line 260:
```suggestion
func expandExpressRouteConnectionRouting(input []interface{}) *network.RoutingConfiguration {
```

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	if propagatedRouteTable := v["propagated_route_table"].([]interface{}); len(propagatedRouteTable) != 0 {
+		result.PropagatedRouteTables = expandArmExpressRouteConnectionPropagatedRouteTable(propagatedRouteTable)
+	}
+
+	return &result
+}
+
+func expandArmExpressRouteConnectionPropagatedRouteTable(input []interface{}) *network.PropagatedRouteTable {
```
ArcturusZhang, line 281:
```suggestion
func expandExpressRouteConnectionPropagatedRouteTable(input []interface{}) *network.PropagatedRouteTable {
```

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	if routeTableIds := v["route_table_ids"].([]interface{}); len(routeTableIds) != 0 {
+		result.Ids = expandIDsToSubResources(routeTableIds)
+	}
+
+	return &result
+}
+
+func flattenArmExpressRouteConnectionRouting(input *network.RoutingConfiguration) []interface{} {
```
ArcturusZhang, line 301:
```suggestion
func flattenExpressRouteConnectionRouting(input *network.RoutingConfiguration) []interface{} {
```

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,340 @@
...
+
+	return []interface{}{
+		map[string]interface{}{
+			"associated_route_table_id": associatedRouteTableId,
+			"propagated_route_table":    flattenArmExpressRouteConnectionPropagatedRouteTable(input.PropagatedRouteTables),
+		},
+	}
+}
+
+func flattenArmExpressRouteConnectionPropagatedRouteTable(input *network.PropagatedRouteTable) []interface{} {
```
ArcturusZhang, line 319:
```suggestion
func flattenExpressRouteConnectionPropagatedRouteTable(input *network.PropagatedRouteTable) []interface{} {
```

azurerm/internal/services/network/express_route_connection_resource_test.go
```diff
@@ -0,0 +1,226 @@
...
+  name           = "acctest-VHUBRT-%d"
+  virtual_hub_id = azurerm_virtual_hub.test.id
+}
+
+resource "azurerm_express_route_connection" "test" {
+  name                             = "acctest-ExpressRouteConnection-%d"
+  express_route_gateway_id         = azurerm_express_route_gateway.test.id
+  express_route_circuit_peering_id = azurerm_express_route_circuit_peering.test.id
+  routing_weight                   = 2
+  authorization_key                = "90f8db47-e25b-4b65-a68b-7743ced2a16b"
```
ArcturusZhang, line 164: Do we have a better way than hard-code this key here? And should we regard this as a credential?

website/docs/r/express_route_connection.html.markdown
```diff
@@ -0,0 +1,134 @@
...
+page_title: "Azure Resource Manager: azurerm_express_route_connection"
+description: |-
+  Manages an Express Route Connection.
+---
+
+# azurerm_express_route_connection
+
+Manages an Express Route Connection.
+
+~> **NOTE:** The provider status of the Express Route Circuit must be set as provisioned while creating the Express Route Connection. See more details from https://docs.microsoft.com/en-us/azure/expressroute/expressroute-howto-circuit-portal-resource-manager#send-the-service-key-to-your-connectivity-provider-for-provisioning.
```
ArcturusZhang, line 13:
```suggestion
~> **NOTE:** The provider status of the Express Route Circuit must be set as provisioned while creating the Express Route Connection. See more details [here](https://docs.microsoft.com/en-us/azure/expressroute/expressroute-howto-circuit-portal-resource-manager#send-the-service-key-to-your-connectivity-provider-for-provisioning).
```

website/docs/r/express_route_connection.html.markdown
```diff
@@ -0,0 +1,134 @@
...
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name which should be used for this Express Route Connection. Changing this forces a new resource to be created.
+
+* `express_route_circuit_peering_id` - (Required) The ID of the Express Route Circuit Peering within this Express Route Connection should be created. Changing this forces a new resource to be created.
```
ArcturusZhang, line 85:
```suggestion
* `express_route_circuit_peering_id` - (Required) The ID of the Express Route Circuit Peering within which this Express Route Connection should be created. Changing this forces a new resource to be created.
```

website/docs/r/express_route_connection.html.markdown
```diff
@@ -0,0 +1,134 @@
...
+
+* `express_route_gateway_id` - (Required) The ID of the Express Route Gateway within this Express Route Connection should be created. Changing this forces a new resource to be created.
```
ArcturusZhang, line 87:
```suggestion
* `express_route_gateway_id` - (Required) The ID of the Express Route Gateway within which this Express Route Connection should be created. Changing this forces a new resource to be created.
```

website/docs/r/express_route_connection.html.markdown
```diff
@@ -0,0 +1,134 @@
...
+
+* `authorization_key` - (Optional) The authorization key to establish the Express Route Connection.
+
+* `enable_internet_security` - (Optional) Is internet security enabled for this Express Route Connection?
+
+* `routing` - (Optional)  A `routing` block as defined below.
+
+* `routing_weight` - (Optional) The routing weight associated to the Express Route Connection. Valid values are between `0` and `32000`.
```
ArcturusZhang, line 95: We could only assign one value for this, right?
```suggestion
* `routing_weight` - (Optional) The routing weight associated to the Express Route Connection. Valid value is between `0` and `32000`.
```

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,340 @@
...
+
+			"enable_internet_security": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
+			"routing": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
ArcturusZhang, line 74: Is this one (and everything it contains) computed because if this is absent and the service will create one for us?

ArcturusZhang (changes requested):
Hi @neil-yechenwei thanks for opening this PR!

Generally this looks good, but I have left some questions and comments inline. Please take a look.

website/docs/r/express_route_connection.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_connection.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_connection.html.markdown
neil-yechenwei: Updated

website/docs/r/express_route_connection.html.markdown
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Yes.

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Yes. You're right. I've unified them to TypeSet.

azurerm/internal/services/network/express_route_connection_resource_test.go
neil-yechenwei: Here is the fake test data. So I think we don't care about it.

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Yes. So I added Computed attribute.

neil-yechenwei:
@ArcturusZhang , thanks for your comments. I've updated code. Please have an another look. Thanks.

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,337 @@
...
+										Optional: true,
+										Computed: true,
+										Elem: &schema.Schema{
+											Type:         schema.TypeString,
+											ValidateFunc: validate.HubRouteTableID,
+										},
+									},
+								},
+							},
+						},
```
ArcturusZhang, line 113: It occurs to me that is this block too much "nested"? Usually we might keep the levels of nesting blocks as low as possible. Therefore could we flatten this block, and renaming them like `propagated_route_table_labels` and `propagated_route_table_ids`?

ArcturusZhang (commented):
Hi @neil-yechenwei thanks for the revisions!

I have a suggestion on the schema - could you please take a look?

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Yes. I understand. But I think three levels is not very deep. After checked other resources, seems three levels is pretty common. So I suggest to keep it.

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,337 @@
...
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"associated_route_table_id": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							Computed:     true,
```
ArcturusZhang, line 81: Well, similar question here: if we set this block but omit this `associated_route_table_id`, will the service return a default one for us?
If so, we might need a test case for this

azurerm/internal/services/network/express_route_connection_resource.go
ArcturusZhang, line 113: Similar question on this block

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: @ArcturusZhang , here "this block" means "routing"?

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: @ArcturusZhang , you mean setting this block "propagated_route_table" but omit "route_table_ids", right?

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,336 @@
...
+			d.Set("express_route_circuit_peering_id", v.ID)
+		}
+
+		if v := props.AuthorizationKey; v != nil {
+			d.Set("authorization_key", v)
+		}
+
+		if v := props.EnableInternetSecurity; v != nil {
+			d.Set("enable_internet_security", v)
+		}
```
ArcturusZhang, line 224: We need to always set these properties, such like 
```suggestion
                peeringID := ""
		if v := props.ExpressRouteCircuitPeering; v != nil {
		        peeringID = *v.ID
		}
			d.Set("express_route_circuit_peering_id", peeringID)

			d.Set("authorization_key", v)

			d.Set("enable_internet_security", v)
```

neil-yechenwei: Updated

neil-yechenwei:
@ArcturusZhang , thanks for your comments. I've updated code. Please have an another look. Thanks.

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,333 @@
...
+	if props := resp.ExpressRouteConnectionProperties; props != nil {
+		d.Set("routing_weight", props.RoutingWeight)
+		d.Set("authorization_key", props.AuthorizationKey)
+		d.Set("enable_internet_security", props.EnableInternetSecurity)
+
+		circuitPeeringID := ""
+		if v := props.ExpressRouteCircuitPeering; v != nil {
+			circuitPeeringID = *v.ID
+		}
+		d.Set("express_route_circuit_peering_id", circuitPeeringID)
```
ArcturusZhang, line 221: could we also parse the peering ID here, format it and set it here? Similar to line 210
```suggestion
		circuitPeeringID := ""
		if v := props.ExpressRouteCircuitPeering; v != nil {
			circuitPeeringID = *v.ID
		}
		peeringId, err := parse.ExpressRouteCircuitPeeringID(circuitPeeringID)
		if err != nil {
		     return fmt.Errorf(...)
		}
		d.Set("express_route_circuit_peering_id", peeringId.ID())
```

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	associatedRouteTableId := ""
+	if input.AssociatedRouteTable != nil && input.AssociatedRouteTable.ID != nil {
+		associatedRouteTableId = *input.AssociatedRouteTable.ID
+	}
+
+	return []interface{}{
+		map[string]interface{}{
+			"associated_route_table_id": associatedRouteTableId,
```
ArcturusZhang, line 306: Similar in this ID, could we also format this one using its IDFormatter?

ArcturusZhang (commented):
Hi @neil-yechenwei thanks a lot for the revisions!

Here should be the final round of comments, I suppose we should introduce the IDFormatter for all of the IDs that we know which resource it belongs to, to keep those IDs consistent. Could you please take a look?

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

neil-yechenwei:
@ArcturusZhang , thanks for your comments. I've updated code. Please have an another look. Thanks.

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,349 @@
...
+				MaxItems:   1,
+				ConfigMode: schema.SchemaConfigModeAttr,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						// Note: When the `associated_route_table_id` property isn't specified, it has default value. So it has to add the `Computed` attribute to ignore the diff.
+						// But if the `Computed` attribute is added, it cannot be rolled back to default value after this property is specified. So the `Computed` attribute has to be removed from schema.
+						"associated_route_table_id": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validation.Any(validate.HubRouteTableID, validation.StringIsEmpty),
```
ArcturusZhang, line 86: I personally prefer adding the `Computed` back here.

neil-yechenwei: So you mean we have to abandon the ability to set the default value back?

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,349 @@
...
+			},
+			RoutingWeight: utils.Int32(int32(d.Get("routing_weight").(int))),
+		},
+	}
+
+	if v, ok := d.GetOk("authorization_key"); ok {
+		expressRouteConnectionParameters.ExpressRouteConnectionProperties.AuthorizationKey = utils.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("enable_internet_security"); ok {
```
magodo, line 172: As this always has a value, can we inline this when initializing the `expressRouteConnectionParameters`?

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,349 @@
...
+		expressRouteConnectionParameters.ExpressRouteConnectionProperties.EnableInternetSecurity = utils.Bool(v.(bool))
+	}
+
+	if v, ok := d.GetOk("routing"); ok {
+		expressRouteConnectionParameters.ExpressRouteConnectionProperties.RoutingConfiguration = expandExpressRouteConnectionRouting(v.([]interface{}))
```
magodo, line 177: Can we inline this when initializing the `expressRouteConnectionParameters`?

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,349 @@
...
+
+	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
+		return fmt.Errorf("waiting for deletion of %s: %+v", *id, err)
+	}
+
+	return nil
+}
+
+func expandExpressRouteConnectionRouting(input []interface{}) *network.RoutingConfiguration {
+	if len(input) == 0 {
```
magodo, line 266: We should also nil check here as both fields are optional:

```suggestion
	if len(input) == 0 || input[0] == nil {
```

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,349 @@
...
+
+	if propagatedRouteTable := v["propagated_route_table"].([]interface{}); len(propagatedRouteTable) != 0 {
+		result.PropagatedRouteTables = expandExpressRouteConnectionPropagatedRouteTable(propagatedRouteTable)
+	}
+
+	return &result
+}
+
+func expandExpressRouteConnectionPropagatedRouteTable(input []interface{}) *network.PropagatedRouteTable {
+	if len(input) == 0 {
```
magodo, line 287: Same here:

```suggestion
	if len(input) == 0 || input[0] == nil {
```

magodo (changes requested):
@neil-yechenwei I've looked through the PR and put some comments, whilst overall it LGTM. Thank you!

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,349 @@
...
+		},
+	}
+}
+
+func resourceExpressRouteConnectionCreateUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*clients.Client).Network.ExpressRouteConnectionsClient
+	ctx, cancel := timeouts.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
+	defer cancel()
+
+	expressRouteGatewayId, err := parse.ExpressRouteGatewayID(d.Get("express_route_gateway_id").(string))
```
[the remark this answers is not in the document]
magodo: From the ID, it appears this piece of code should be right.

website/docs/r/express_route_connection.html.markdown
```diff
@@ -0,0 +1,134 @@
...
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name which should be used for this Express Route Connection. Changing this forces a new resource to be created.
+
+* `express_route_circuit_peering_id` - (Required) The ID of the Express Route Circuit Peering within which this Express Route Connection should be created. Changing this forces a new resource to be created.
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

website/docs/r/express_route_connection.html.markdown
```diff
@@ -0,0 +1,134 @@
...
+
+* `express_route_gateway_id` - (Required) The ID of the Express Route Gateway within which this Express Route Connection should be created. Changing this forces a new resource to be created.
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

website/docs/r/express_route_connection.html.markdown
```diff
@@ -0,0 +1,134 @@
...
+
+* `authorization_key` - (Optional) The authorization key to establish the Express Route Connection.
+
+* `enable_internet_security` - (Optional) Is internet security enabled for this Express Route Connection?
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

website/docs/r/express_route_connection.html.markdown
```diff
@@ -0,0 +1,134 @@
...
+
+* `routing` - (Optional)  A `routing` block as defined below.
+
+* `routing_weight` - (Optional) The routing weight associated to the Express Route Connection. Valid value is between `0` and `32000`. Defaults to `0`.
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource_test.go
```diff
@@ -0,0 +1,228 @@
...
+			),
+		},
+		data.ImportStep(),
+		{
+			Config: r.complete(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+			),
+		},
+		data.ImportStep(),
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,349 @@
+package network
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-05-01/network"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
neil-yechenwei: Updated

magodo:
@neil-yechenwei Would you please rerun the test on your environment after the changes (especially for the acctest), and provide the result here?

neil-yechenwei:
@magodo , here is the test result after rerun:
--- PASS: TestAccExpressRouteConnection/Resource/basic (798.93s)
--- PASS: TestAccExpressRouteConnection/Resource/requiresImport (809.84s)
--- PASS: TestAccExpressRouteConnection/Resource/complete (784.32s)
--- PASS: TestAccExpressRouteConnection/Resource/update (1479.08s)

magodo (changes requested):
Can we also add a separate Update for this resouce as is done in #11303?

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -163,11 +160,14 @@ func resourceExpressRouteConnectionCreateUpdate(d *pluginsdk.ResourceData, meta
...
 			EnableInternetSecurity: utils.Bool(d.Get("enable_internet_security").(bool)),
-			RoutingConfiguration:   expandExpressRouteConnectionRouting(d.Get("routing").([]interface{})),
 			RoutingWeight:          utils.Int32(int32(d.Get("routing_weight").(int))),
 		},
 	}
 
+	if v, ok := d.GetOk("routing"); ok {
+		parameters.ExpressRouteConnectionProperties.RoutingConfiguration = expandExpressRouteConnectionRouting(v.([]interface{}))
+	}
+
```
magodo, line 170: Wondering why this check is needed here?

neil-yechenwei: It's optional property. So added a check.

azurerm/internal/services/network/express_route_connection_resource.go
```diff
@@ -0,0 +1,344 @@
...
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/tf/pluginsdk"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/timeouts"
+	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/utils"
+)
+
+func resourceExpressRouteConnection() *pluginsdk.Resource {
+	return &pluginsdk.Resource{
+		Create: resourceExpressRouteConnectionCreateUpdate,
+		Read:   resourceExpressRouteConnectionRead,
+		Update: resourceExpressRouteConnectionCreateUpdate,
```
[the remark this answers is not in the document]
neil-yechenwei: Updated

azurerm/internal/services/network/express_route_connection_resource.go
magodo: Isn't this already checked in the expand function?

neil-yechenwei: Updated

neil-yechenwei:
--- PASS: TestAccExpressRouteConnection/Resource/basic (768.89s)
--- PASS: TestAccExpressRouteConnection/Resource/requiresImport (802.97s)
--- PASS: TestAccExpressRouteConnection/Resource/complete (820.47s)
--- PASS: TestAccExpressRouteConnection/Resource/update (1329.18s)

magodo:
@neil-yechenwei The test on TC failed with:

image

Is there anything else needed to be setup in order to create the ERP in the specified location?

neil-yechenwei:
@magodo , did you see this error before? I think you have knowledge about creating express route port.

neil-yechenwei:
@magodo , I updated code. Now all test cases are passed on TeamCity.
image

katbyte (approved):
LGTM 🍰

PR #11321, opened 2021-04-14
"azurerm_kubernetes_cluster" - supports property "dns_prefix_private_cluster"

njuCZ:
this property in swagger is "fqdnSubdomain", it is used in private cluster to set custom dns subdomain, so I renamed it to "private_cluster_custom_dns_prefix"

Doc refer to https://docs.microsoft.com/en-us/azure/aks/private-clusters#configure-private-dns-zone

difference between property `dns_prefix` and `private_cluster_custom_dns_prefix`:
1. use "dns_prefix", the fqdn format is like `<dns-prefix>-<random-hash>.xxx.<region>.azmk8s.io`
2. use "private_cluster_custom_dns_prefix", the fqdn format is like `<fqdnsubdomain>.privatelink.<region>.azmk8s.io`

there should be exactly one of `dns_prefix` and `private_cluster_custom_dns_prefix`, otherwise the rest api will report error.

image


website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -66,7 +66,9 @@ The following arguments are supported:
 
 * `default_node_pool` - (Required) A `default_node_pool` block as defined below.
 
-* `dns_prefix` - (Required) DNS prefix specified when creating the managed cluster. Changing this forces a new resource to be created.
+* `dns_prefix` - (Optional) DNS prefix specified when creating the managed cluster. Changing this forces a new resource to be created.
+
+* `private_cluster_custom_dns_prefix` - (Optional) Specifies the Custom DNS prefix when creating the private cluster. Changing this forces a new resource to be created.
```
katbyte, line 71: grammar and i think custom is redundant?

```suggestion
* `private_cluster_dns_prefix` - (Optional) Specifies the DNS prefix to use with private clusters. Changing this forces a new resource to be created.
```

also given there is a dns_prefix property, maybe this makes sense to keep them ordered/auto completed at the same time

```suggestion
* `dns_prefix_private_cluster` - (Optional) Specifies the DNS prefix to use with private clusters. Changing this forces a new resource to be created.
```

katbyte (changes requested):
aside from one comment about this property name this is looking good

njuCZ:
@katbyte Thanks for the suggestion. I have renamed this property

katbyte (approved):
Thanks @njuCZ - LGTM 👍

PR #11328, opened 2021-04-14
Adding CMK to Synapse

allantargino:
This PR adds support for [CMK](https://docs.microsoft.com/en-us/azure/synapse-analytics/security/workspaces-encryption#workspace-encryption-configuration) on Synapse.


azurerm/internal/services/synapse/synapse_workspace_resource.go
```diff
@@ -228,6 +229,12 @@ func resourceSynapseWorkspace() *schema.Resource {
 				Optional: true,
 			},
 
+			"customer_managed_key": {
```
WodansSon, line 232: The name of this attribute appears to be a bit ambiguous, it should be named something that would be a bit more intuitive as to what data the end user should be expected to put here, maybe something like `customer_managed_key_versionless_id`?

azurerm/internal/services/synapse/synapse_workspace_resource.go
```diff
@@ -642,3 +660,14 @@ func flattenIdentityControlSQLSettings(settings synapse.ManagedIdentitySQLContro
 
 	return false
 }
+
+func flattenEncryptionDetails(encryption *synapse.EncryptionDetails) string {
```
WodansSon, line 664: Maybe have this return a pointer instead of a string so you can check for `nil`?
```suggestion
func flattenEncryptionDetails(encryption *synapse.EncryptionDetails) *string {
```

azurerm/internal/services/synapse/synapse_workspace_resource.go
```diff
@@ -642,3 +660,14 @@ func flattenIdentityControlSQLSettings(settings synapse.ManagedIdentitySQLContro
...
+	if cmk := encryption.Cmk; cmk != nil {
+		if key := cmk.Key; key != nil {
+			if keyVaultUrl := key.KeyVaultURL; keyVaultUrl != nil {
+				return *keyVaultUrl
+			}
+		}
+	}
+	return ""
```
WodansSon, line 672: If the return value is a pointer this can change to `nil` instead of an empty string.

azurerm/internal/services/synapse/synapse_workspace_resource_test.go
```diff
@@ -280,10 +296,64 @@ resource "azurerm_synapse_workspace" "test" {
...
+}
+
 func (r SynapseWorkspaceResource) template(data acceptance.TestData) string {
 	return fmt.Sprintf(`
 provider "azurerm" {
-  features {}
+  features {
+    key_vault {
+      purge_soft_delete_on_destroy = false
+    }
```
WodansSon, line 355: Is there a reason for disabling the `purge_soft_delete_on_destroy` feature?

website/docs/r/synapse_workspace.html.markdown
```diff
@@ -81,6 +81,8 @@ The following arguments are supported:
 
 * `github_repo` - (Optional) A `github_repo` block as defined below.
 
+* `customer_managed_key` - (Optional) An Azure Key Vault key ID used as a Customer Managed Key (CMK) for double encryption.
```
WodansSon, line 84: Maybe give an example of expected input here?

```suggestion
* `customer_managed_key_versionless_id` - (Optional) The Azure Key Vault Key Versionless ID to be used as the Customer Managed Key (CMK) for double encryption (e.g. `https://example-keyvault.vault.azure.net/type/cmk/`). 
```

WodansSon (changes requested):
@allantargino, thank you so much for the PR. It's looking good so far but I left a few minor comments and questions. Once we get those addressed we can get this merged and shipped in the next release! 🚀 

**Requested Change:**
Can we actually pull this out of the `azurerm_synapse_workspace` resource and make it its own separate `azurerm_synapse_workspace_customer_managed_key` resource to be more consistent with the already existing implementations of this functionality? (e.g. `azurerm_storage_account_customer_managed_key`, `azurerm_log_analytics_cluster_customer_managed_key`, etc.)

WodansSon:
@allantargino, to get rid of the CI error you may need to pull the origin/master branch into your fork.

azurerm/internal/services/synapse/synapse_workspace_resource.go
allantargino: Agreed! I will change it.

allantargino:
Hi @WodansSon, thanks for reviewing the PR!

I first was trying to implement `azurerm_synapse_workspace_customer_managed_key` to be consistent with the other implementations as you said, but seems Synapse has a different behavior than the other resources when it comes to double encryption.

You can only specify if you need double encryption or not when you are creating the resource. If you opted for it, you need to pass a key vault key on the body of the request. Otherwise, you won't be able to update Synapse to use an existing one after the workspace has been created. So that's why I chose to embed it :(

Please check [this documentation](https://docs.microsoft.com/en-us/azure/synapse-analytics/security/workspaces-encryption#workspace-encryption-configuration).
> Workspaces can be configured to enable double encryption with a customer-managed key at the time of workspace creation.
> [...]
> Important: The configuration setting for double encryption cannot be changed after the workspace is created.

azurerm/internal/services/synapse/synapse_workspace_resource_test.go
allantargino: I added it because I keep receive the following error message:
> The user, group or application 'appid=...;oid=...;iss=https://sts.windows.net/.../' does not have keys purge permission on key vault 'acckv210422144256592239;location=...'

katbyte (approved):
Thanks @allantargino - LGTM 👍

sehgalnamit:
It doe snot work as it needs grant to be done to managed identity to the AKV

2021-05-05 18:25:10  
2021-05-05 18:25:10  Error: updating Synapse Workspace "syn-sgrtss-dev-az1-mwrgy7" Sql Admin (Resource Group "rsg-sgrtss-dev-az1-mwrgy7"): synapse.WorkspaceAadAdminsClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="WorkspaceActivationRequired" Message="Workspace needs to be activated, by adding the managed identity in the KeyVault containing the customer managed key and activating the workspace through the keys subresource."

Azure Synapse Encryption can only be enabled during creation time for some unknown reason.
 During creation of Azure Synapse it fails as managed identity of Synapse getting created is not granted access to AKV.

MS document says activation can only be done manually:-
https://docs.microsoft.com/en-us/azure/synapse-analytics/security/workspaces-encryption

So this feature doe snot work with terraform and need to be relloked.

I can grant MSI access in my module 

resource "azurerm_key_vault_access_policy" "instance" {
  
  key_vault_id        = var.keyvault_id
  tenant_id          = data.azurerm_client_config.current.tenant_id
  object_id          = azurerm_synapse_workspace.synapse.identity[0].principal_id
  key_permissions    = ["Get", "List", "Delete", "Create", "Encrypt", "Recover", "Purge", "Verify", "UnwrapKey", "WrapKey"]
  secret_permissions = ["Get", "List", "Set", "Delete"]
depends_on = [
    azurerm_synapse_workspace.synapse,
  ]
}

But issue is Synapse is still getting created and it just fails and there is not az command to activate it.

Unlike other service Azure SQL, Postgres etc.. which can enable TDE after instance creation, Azure Synapse doe snot allow this.

sehgalnamit:
This is the Bug which MS has to fix in its ARM template and also in terraform
  #aad_admin {
  #  login     = var.settings["aad_admin"]
  #  object_id = element(concat(data.azuread_user.synapse_aadadmin.*.id, data.azuread_group.synapse_aadadmin.*.id), 0)
  #  tenant_id = data.azurerm_client_config.current.tenant_id
  #}


•	Create workspace, with CMK set
•	Grant workspace identity permission to the key vault key
•	Active the workspace
•	Create the workspace's AAD admin ( this one not sure how to dit as we can just add roles for synapse admin, sql admin and spark pool admin).

DesaCh01:
Hello, 

I too am facing the same issue as @sehgalnamit. 

Error received while trying to create a synapse workspace using customer_managed_key_versionless_id property.
 
**_Error: updating Synapse Workspace "rsgue2fidodasa02" Sql Admin (Resource Group "shared-fido-dev-eastus2"): synapse.WorkspaceAadAdminsClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="WorkspaceActivationRequired" Message="Workspace needs to be activated, by adding the managed identity in the KeyVault containing the customer managed key and activating the workspace through the keys subresource."

  on azure-synapse.tf line 6, in resource "azurerm_synapse_workspace" "synapse":
   6: resource "azurerm_synapse_workspace" "synapse"_** 
   
 
 I can assign the key permissions to Managed Identity only once the workspace is deployed and I get the identity's principal ID. But, currently it erroring out while synapse workspace creation.

PR #11335, opened 2021-04-15
Fix gradual deprecate script

magodo:
The `grep -R` will cause `grep` not reading from pipe, but from the cwd recursively. This causes CI breaks. Meanwhile, the error message doesn't provide enough information to identify where the deprecated usage appears. This is due to the script was grepping from `git diff`.

Rather than restrict the change scope into added lines, it might be reasonable to extend the scope a bit to the affected files (despite the removed ones), and only focus on their latest content. This way we can provide the filename and the affected line#.

I can tell the good intent here to not bring noise to people who didn't added deprecated lines. However, this still can't avoid the cases that some users move the existing deprecated lines around, which might also ends up to show up as newly added lines in the `git diff`. So I think it makes sense to extent the scope to file based (rather than line based).


scripts/run-gradually-deprecated.sh
```diff
@@ -3,59 +3,60 @@
...
+      echo "  >   return err"
+      echo "  > }"
+      echo ""
+      echo "New Resources should be using Resource ID Formatters/Parsers by default"
+      echo "however existing (unmodified) resources can continue to use the Azure ID"
+      echo "for the moment - but over time these will be switched across."
+      exit 1
+  }
+
+done
```
tombuildsstuff, line 59: do we need the done here?

tombuildsstuff (approved):
one comment but this otherwise LGTM - thanks for fixing this @magodo

PR #8253, opened 2020-08-26
New Resource: `azurerm_log_analytics_saved_search`

mbfrahry:
Dependent on #8234 

```
--- PASS: TestAccAzureRMLogAnalyticsSavedSearch_complete (139.56s)
--- PASS: TestAccAzureRMLogAnalyticsSavedSearch_basic (140.36s)
--- PASS: TestAccAzureRMLogAnalyticsSavedSearch_requiresImport (144.51s)
```

PR #11341, opened 2021-04-15
azurerm_log_analytics_saved_search : Fix Regex within function_parame…

kaovd:
…ters validation

The regex requirement in the given resource works for integers, as given in the example in the original PR, however it does not for strings. If a string has a default value supplied it must be in quotes (i.e in powershell New-AzOperationalInsightsSavedSearch, it would be -FunctionParameter "ID:string='g'"). While you can still make a successful apply doing var:string=abc, it will break when using the function as it is essentially going to run (e.g) | where Computer contains abc. abc does not exist. It needs to be "abc" if you where doing a query in LA (just an example). 

Example Error:
image

I have changed two things about the regex, one is to allow single quotes after the ='s if a user wants to set a default string, I have also put an OR so its optional to just not have any default values at all, which are both valid options using powershell / raw api. 

One option is just to get rid of this regex check completely and give responsibility to the user rather than having an extensive regex check, it seems this is more just so the users knows how to use the resource.

@mbfrahry

#8253

### DIFF
--- a/azurerm/internal/services/loganalytics/log_analytics_saved_search_resource.go
+++ b/azurerm/internal/services/loganalytics/log_analytics_saved_search_resource.go
@@ -90,8 +90,8 @@ func resourceLogAnalyticsSavedSearch() *schema.Resource {
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 					ValidateFunc: validation.StringMatch(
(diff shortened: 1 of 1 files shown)


katbyte (approved):
Thanks @kaovd - LGTM 👍

PR #11343, opened 2021-04-15
Downgrading the Azure SDK to v52.6.0

tombuildsstuff:
This PR downgrades the Azure SDK for Go to v52.6.0

This is required due to the latest changes to the Azure SDK's Future implementation, which polls for the completion of long running resources within Azure - is fundamentally broken when a Future fails. This means that no HTTP Response is returned and so attempts to access `future.Response()` panic.

[There's an upstream PR fixing this in the Azure code generator](https://github.com/Azure/autorest.go/pull/605), but that's waiting on Microsoft to merge/re-generate - whilst I did investigate regenerating the SDK in the interim, the documentation for doing so is out of date - as such the quickest thing to unblock us here is to downgrade.


katbyte (approved):
LGTM 👍

Issue #11103, opened 2021-03-24
Support for tags on azurerm_subscription resource


ondrejvaclavu:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Tags property in azurerm_subscription resource is read only. Should be writable, so that tags can be assigned in Terraform code, as for all other Azure resources.

### New or Affected Resource(s)

* azurerm_subscription

### Potential Terraform Configuration

```hcl

resource "azurerm_subscription" "example" {
  subscription_name = "My Example EA Subscription"
  billing_scope_id  = data.azurerm_billing_enrollment_account_scope.example.id

  tags = {
    "CostCenter" = "021 - Finance"
    "OwnerEmail" = "finance@company.xyz"
  }
}

```

### References

* https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/subscription#example-usage---creating-a-new-alias-and-subscription-for-an-enrollment-account
* https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/tag-resources

Burbert:
Is there a timeline for this feature? Maybe in the near future? :)

jhisc:
> Is there a timeline for this feature? Maybe in the near future? :)

This PR has a little more information about progress:
https://github.com/hashicorp/terraform-provider-azurerm/pull/12287

artur-carvalho:
Any idea of when this will be available?

oswalya:
This is also a feature I am looking forward to have. Would be great to have this in the provider soon!

jackofallops:
This was closed by #14445 and has been released in provider version 2.88.0.

PR #11347, opened 2021-04-15
Support for tags on azurerm_subscription resource #11103

crkg:
- The code is tested and works as expected.

### DIFF
--- a/azurerm/internal/services/subscription/client/client.go
+++ b/azurerm/internal/services/subscription/client/client.go
@@ -4,12 +4,14 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-11-01/subscriptions"
 	subscriptionAlias "github.com/Azure/azure-sdk-for-go/services/subscription/mgmt/2020-09-01/subscription"
 	"github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/common"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2020-06-01/resources"
 )
 
 type Client struct {
 	Client             *subscriptions.Client
 	AliasClient        *subscriptionAlias.AliasClient
 	SubscriptionClient *subscriptionAlias.Client
+	TagsClient		   *resources.TagsClient
 }
 
 func NewClient(o *common.ClientOptions) *Client {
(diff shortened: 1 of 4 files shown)


jackofallops (changes requested):
Hi @crkg - Many thanks for this PR.  If this does work as described, I think this should be added as an enhancement to the existing `azurerm_subscription` resource, rather than a separate resource for users to manage. I noticed this is your first contribution, is this something that you feel comfortable to do?

crkg:
@Steve, thanks for your feedback, let me contribute to the subscription resource.

katbyte:
I think we support it in the resource currently (https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/subscription/subscription_resource.go#L114) but i don't think that would work with existing subscriptions which this would @jackofallops ?

crkg:
The tags from this resource never works for a new subscription or update the existing subscription. the error received is field cannot be set. hence i created a separate resource with createorupdateatscope . the supported API i found for scope specific  - https://docs.microsoft.com/en-us/rest/api/resources/tags/createorupdateatscope

jackofallops:
Hi @katbyte - @crkg is correct, the Subscriptions APIs (in both the Subscription and Resource RP's) do not support writing tags, and they are currently read-only on this resource.  I was previously unaware that the API noted in this PR could be leveraged to perform the necessary update to the tags on Subscriptions (I've not had chance to test it does actually successfully do so, however).

crkg:
Hi @jackofallops  ( Steve), all the checks have passed, would it be possible for you to review and merge the code to master? for community availability.

jackofallops:
> Hi @jackofallops ( Steve), all the checks have passed, would it be possible for you to review and merge the code to master? for community availability.

Hi @crkg - My earlier feedback was to include this in the existing resource as an enhancement, and this PR is still introducing a new one? We shouldn't have discrete resources for tagging as users would expect to tag the resource at creation time. Since we can leverage the tagsClient to this end, it should be possible to create this experience in-line with the `azurern_subscription` resource. Apologies if I was unclear previously.

favoretti:
Also when looking through the code I see you're attempting to reactivate disabled/deleted subs on tags update - I'm not sure we want this implicit behavior. Enabling/Disabling subscriptions is a very lengthy process and I wouldn't want mine to be re-activated by accident if I made a coding error in TF, I'd rather have TF fail.

crkg:
> Also when looking through the code I see you're attempting to reactivate disabled/deleted subs on tags update - I'm not sure we want this implicit behavior. Enabling/Disabling subscriptions is a very lengthy process and I wouldn't want mine to be re-activated by accident if I made a coding error in TF, I'd rather have TF fail.

I am working on adding the tags to azurerm_subscription so we dont have discrete resource. point noted, did you test it if it enables? i am not sure as i did not had a disabled subscription to test. also why would one make such human error.

jackofallops:
Hi @crkg - Just quickly checking in to see if you are still looking to implement this feature? No worries if not, I can make some time in the next couple weeks to add it if that would be your preference?
Thanks!

crkg:
> Hi @crkg - Just quickly checking in to see if you are still looking to implement this feature? No worries if not, I can make some time in the next couple weeks to add it if that would be your preference?
> Thanks!

Hi Steve,  i am middle of completion since the helper functions changed. i got stuck in an error where delete tags wasnt working as expected.. i will let you know if fixing the error takes little longer should not take longer then a day or two.

katbyte:
@crkg to going to close this in favour of the new PR #11347

PR #11351, opened 2021-04-16
azurerm_databricks_workspace - allow for workspace names up to 64 characters

pmackles:
Limiting workspace names to 30 characters because of the validation rules enforced by the UI prevents importing workspaces that were previously created through the API (which allows up to 64 characters). This change increases the limit to 64. Note that workspace names >30 characters work fine in the portal. I believe this change is backwards compatible.

Note that I had some trouble running the unit tests due to the `runGraduallyDeprecatedFunctions` checks. Being completely unfamiliar with golang, I didn't feel comfortable making the suggested changes (which had nothing to do with my change) so I am hoping someone here can look at that.


katbyte (approved):
Thanks @pmackles - this LGTM 👍

PR #11352, opened 2021-04-16
`azurerm_eventgrid_system_topic` - remove strict validation on `topic_type`

### LINKED ISSUE #11308: azurerm_eventgrid_system_topic to support topic_type "Microsoft.PolicyInsights.PolicyStates"
I am trying to create an eventgrid system topic according to MSFT [tutorial ](https://docs.microsoft.com/en-us/azure/governance/policy/tutorials/route-state-change-events#create-an-event-grid-system-topic) using terraform but I found the given topic_type is not yet supported.

Location should be "Global"

MattiasAng:
Fixes #11308

Instead of doing validation of every topic type I think this should be changed to letting API handle it as the error message is very clear. This would not require changes to any code when new topic types are released.

Error message on a system topic that is not available:
` Original Error: Code="InvalidRequest" Message="Unrecognized topic type justtesting in system topic properties."`

### DIFF
--- a/azurerm/internal/services/eventgrid/eventgrid_system_topic_resource.go
+++ b/azurerm/internal/services/eventgrid/eventgrid_system_topic_resource.go
@@ -65,29 +65,10 @@ func resourceEventGridSystemTopic() *schema.Resource {
 			},
 
 			"topic_type": {
-				Type:     schema.TypeString,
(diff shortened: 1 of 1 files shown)


mbfrahry (approved):
LGTM!

PR #11378, opened 2021-04-19
refactor: splitting the validators out into their own packages

tombuildsstuff:
This PR splits the validators out from the Service Packages into the `validate` package, which is necessary to avoid a circular reference with the upcoming work to enable the Plugin SDK upgrade.


azurerm/internal/services/synapse/synapse_managed_private_endpoint_resource.go
```diff
@@ -5,13 +5,14 @@ import (
 	"log"
 	"time"
 
+	validate2 "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/network/validate"
```
manicminer, line 8: I think `networkValidate` for this alias would be a little clearer

manicminer (approved):
LGTM! I left a few comments on the import blocks of a few files but those aren't blocking.

azurerm/internal/services/synapse/synapse_managed_private_endpoint_resource.go
tombuildsstuff: agreed, I should file a Goland bug about these actually 🙃

Issue #11379, opened 2021-04-19
azurerm_subscription should be able to search by display_name


donwlewis:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It would be great to be able to find an exact subscription via display name.  I know that the `azurerm_subscriptions` research is able to do it, but it would be great to be able to pin-point an exact subscription.

### New or Affected Resource(s)

`azurerm_subscription`

### Potential Terraform Configuration

```hcl
data azurerm_subscription "main" {
  display_name = "mysubscription"
}
```

### References

* #0000

donwlewis:
After doing a little research on this, I think the best approach would be the leverage the `azurerm_subscriptions` resource, and then add an `exact_match` property.  I have something working in testing.  I can try and submit a PR for this.

magodo:
Does a [for expression](https://www.terraform.io/docs/language/expressions/for.html) like below do the same thing:

```
> [for elem in [{name= "foo", id=123}, {name="foo", id=456}, {name="bar", id=789}]: elem if elem.name=="foo"]
[
  {
    "id" = 123
    "name" = "foo"
  },
  {
    "id" = 456
    "name" = "foo"
  },
]
```

PR #11380, opened 2021-04-19
Added ability to look for exact matches for subscriptions.

donwlewis:
Pull request for feature requested in #11379.

The idea is to use the `exact_match` field in the `azurerm_subscriptions` resource to get all subscriptions that exactly match the given display name.  This could still result in multiples as it seems like you can create subscriptions in Azure with the same name.  Also, if the `exact_match` property is used, the `display_name_prefix` and `display_name_contains` fields are ignored.

### DIFF
--- a/azurerm/internal/services/subscription/data_source_subscriptions.go
+++ b/azurerm/internal/services/subscription/data_source_subscriptions.go
@@ -29,6 +29,10 @@ func dataSourceSubscriptions() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"exact_match": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
 			"subscriptions": {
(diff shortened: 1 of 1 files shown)


donwlewis:
Honestly, that probably makes a lot more sense.  Yea that would do it.

manicminer:
Thanks for the PR @donwlewis! Whilst I appreciate the benefit of being able to identify a single subscription, at the same time subscription display names are not enforced to be unique, so even adding this functionality does not necessarily solve the problem as it's still down to the practitioner to enforce a naming convention within their account. And given that prerequisite, I'm not sure matching the name _exactly_ provides much benefit over the existing prefix-matching functionality. Am I missing something?

donwlewis:
Yup!  I agree.  appreciate the input and consideration.  I will close this one out.

Issue #11371, opened 2021-04-19
netapp dual protocol volume export policy does not allow to use of allow_root_access


upgradeservices:
using
tf 14.10
hashicorp/azurerm 2.55.0

Create a dual protocol netapp volume.
Use example: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/examples/netapp/volume_dual_protocol/main.tf .
Add allow_root_access = yes to export policy block per documentation https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/netapp_volume#allow_root_access:
  export_policy_rule {
    rule_index = 1
    allowed_clients = ["0.0.0.0/0"]
    cifs_enabled = true
    nfsv3_enabled = true
    nfsv4_enabled = false
    unix_read_write = true
    allow_root_access = true
  }

Error result is:
Error: Unsupported argument

  on deploy.tf line 80, in resource "azurerm_netapp_volume" "ultra-aaa-volume021":
  80:     allow_root_access = true

An argument named "allow_root_access" is not expected here.

BrendanThompson:
This is actually just a mistake in the documentation. I will put up a PR momentarily, however if you use the field `root_access_enabled` the code should work as expected.

PR #11386, opened 2021-04-20
fix: update docs to represent struct

BrendanThompson:
Resolves: #11371


magodo:
@BrendanThompson Thank you, this LGTM 👍

Issue #10341, opened 2021-01-27
IotHub IP filter rule ordering is unpredictable


abilous-ti:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

* `azurerm_iothub`

### Terraform Configuration Files

```hcl
resource "azurerm_iothub" "iothubname" {
  name                = "somename"
  resource_group_name = azurerm_resource_group.someresourcegroup
  location            = azurerm_resource_group.somelocation

  sku {
    name     = "B2"
    capacity = "2"
  }

  fallback_route {
    enabled = true
  }



ip_filter_rule {
    action = "Accept"
    ip_mask ="some_ip_range_1"
    name = "some_name_1"
}



 ip_filter_rule {
      action = "Accept"
        ip_mask ="some_ip_range_2"
        name = "some_name_2"   }



ip_filter_rule {
      action = "Accept"
        ip_mask ="some_ip_range_3"
        name = "some_name_3"
  }



 ip_filter_rule {
      action = "Reject"
        ip_mask ="0.0.0.0/0"
        name = "everything_else"
  }
}
```

### Expected Behaviour

Order of IP filter rule in IoTHub
some_name_1
some_name_2
some_name_3

### Actual Behaviour

Order of IP filter rule in IoTHub is unpredictable and could be
some_name_2
some_name_1
some_name_3

### Steps to Reproduce

1. `terraform apply`

### References

* #0000

abilous-ti:
@lrxtom2 @katbyte
I see you are involved in IotHub issues.
Could you please help with this issue?

PR #11390, opened 2021-04-20
azurerm_iothub - change `filter_rule` from TypeSet to TypeList

magodo:
This is to reserve the order of the rules as defined in the TF config.

Fixes: #10341


manicminer:
Thanks @magodo, I've done some testing and it looks like the API ordering is stable for this property. I've expanded the test case a little to try and catch any regressions and re-running now.

manicminer (approved):
Thanks @magodo, I've done some testing and it looks like the API ordering is stable for this property. I've expanded the test case a little to try and catch any regressions and re-running now.

manicminer:
Test result:

Screenshot 2021-04-20 at 17 48 02

PR #11394, opened 2021-04-20
refactor: moving state migrations into their own package / migrating to State Upgraders

tombuildsstuff:
This PR builds on the work of #11378 by refactoring the state migrations into their own package, and migrates all of the legacy MigrateState functions into StateUpgraders. Whilst this may seem like an insignificant change, ultimately this allows us to avoid a circular reference and makes the Plugin SDK changes easier to implement.


jackofallops (approved):
Thanks @tombuildsstuff - There's a small number of observations, and a couple of Actions failures, but otherwise LGTM 👍

PR #11286, opened 2021-04-11
`mssql_database` Update documentation and print warning for secondary db with `max_size_gb`

aristosvo:
Fixes #11282

I don't know how to signal this warning more explicitly than proposed, without breaking anything.

PR #11401, opened 2021-04-20
`azurerm_mssql_database ` - error when secondary database uses `max_size_gb`

aristosvo:
@katbyte I introduced a bug in #11286 :(

Because `max_size_gb` is `Computed`, `GetOk("max_size_gb")` is not empty in a second run. I added `HasChange("max_size_gb")` as an extra check before I throw an error to fix it.

### DIFF
--- a/azurerm/internal/services/mssql/mssql_database_resource.go
+++ b/azurerm/internal/services/mssql/mssql_database_resource.go
@@ -396,7 +396,7 @@ func resourceMsSqlDatabaseCreateUpdate(d *schema.ResourceData, meta interface{})
 	}
 
 	if v, ok := d.GetOk("max_size_gb"); ok {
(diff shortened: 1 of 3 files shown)


mbfrahry (changes requested):
Hey @aristosvo! Can we confirm this fix through a test? We want to make sure that this change is broken in the current version of the provider and that it'll be fixed in the next version of the provider.

aristosvo:
Sure! I had a hard time reproducing the issue in the first place, but I'll try!!

aristosvo:
@mbfrahry how can I make sure a certain version in the test? When I do a `make build` the latest version from master is used (it seems at least that way)

mbfrahry (approved):
LGTM! Thanks for getting this written out into a test!

PR #11404, opened 2021-04-21
Fix markdown heading not rendering

mboutet:
The heading was not rendering in the Terraform registry doc. It looks like that the space between `##` and `Disclaimers` was not a space.


katbyte (approved):
Thanks @mboutet! LGTM 👍

PR #11406, opened 2021-04-21
"azurerm_kubernetes_cluster" - supports "max_node_provisioning_time","max_unready_percentage" and "max_unready_nodes"

njuCZ:
This PR adds support for the `max_node_provision_time`, `max_total_unready_percentage` and `ok_total_unready_count` property in the `auto_scaler_profile block` for `azurerm_kubernetes_cluster`.

As per https://docs.microsoft.com/en-us/azure/aks/cluster-autoscaler#using-the-autoscaler-profile

the original acctest `TestAccKubernetesCluster_autoScalingProfile` fails because the version "1.19.6" is not available in the westeurope
image


azurerm/internal/services/containers/kubernetes_cluster_resource.go
```diff
@@ -124,12 +124,28 @@ func resourceKubernetesCluster() *schema.Resource {
...
+							Type:         schema.TypeString,
+							Optional:     true,
+							Computed:     true,
+							ValidateFunc: containerValidate.Duration,
+						},
+						"max_total_unready_percentage": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+						},
```
magodo, line 137: Can we add a validation here?

azurerm/internal/services/containers/kubernetes_cluster_resource.go
```diff
@@ -124,12 +124,28 @@ func resourceKubernetesCluster() *schema.Resource {
...
 							Type:         schema.TypeString,
 							Optional:     true,
 							Computed:     true,
 							ValidateFunc: containerValidate.Duration,
 						},
+						"ok_total_unready_count": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+						},
```
magodo, line 148: Can we add a validation here?

azurerm/internal/services/containers/kubernetes_cluster_resource.go
```diff
@@ -124,12 +124,28 @@ func resourceKubernetesCluster() *schema.Resource {
...
 							Computed: true,
 						},
+						"max_node_provision_time": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							Computed:     true,
+							ValidateFunc: containerValidate.Duration,
+						},
+						"max_total_unready_percentage": {
+							Type:     schema.TypeString,
```
magodo, line 134: Does it make sense to change this to `TypeInt`?

azurerm/internal/services/containers/kubernetes_cluster_resource.go
```diff
@@ -124,12 +124,28 @@ func resourceKubernetesCluster() *schema.Resource {
...
+							Computed: true,
+						},
 						"new_pod_scale_up_delay": {
 							Type:         schema.TypeString,
 							Optional:     true,
 							Computed:     true,
 							ValidateFunc: containerValidate.Duration,
 						},
+						"ok_total_unready_count": {
+							Type:     schema.TypeString,
```
magodo, line 145: Does it make sense to change this to `TypeInt`?

magodo (changes requested):
Thanks for this PR!
I've taken a look through and left some comments inline, but this is mostly looking good to me 👍

magodo:
LGTM 👍
image

magodo:
@njuCZ About the default values, I'm a bit concerned if some existing resource has different default values being set when they created the resource with old provider, then they'll get a diff in the new provider. Though this is not likely to happen if the service keeps the behavior consistent.

njuCZ:
@magodo the simplest way is to make them typeString, empty string will make the backend service return the  default value. This is a little common for aks. But I also agrees that exlipcit type is more helpful. To make it TypeInt,  we should set a default value, because 0 is also a valid value.

magodo:
If they are O+C, I assume the default is not necessary?

njuCZ:
If it's TypeString, O+C has no problem. but if we set it as TypeInt, it will have a default value 0, 0 is a valid value for this property, and the backend service will think client want to set it as 0 and does not return the default value

magodo:
Ah.. Because those properties are inside a nested block, which means there is no way to tell whether they are omitted in the config or not..

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -217,16 +217,22 @@ A `addon_profile` block supports the following:
...
 
 * `max_graceful_termination_sec` - Maximum number of seconds the cluster autoscaler waits for pod termination when trying to scale down a node. Defaults to `600`.
 
+* `max_node_provision_time` - Maximum time the autoscaler waits for a node to be provisioned. Defaults to `15m`.
+
+* `max_total_unready_percentage` - Maximum percentage of unready nodes the cluster autoscaler will stop if the percentage is exceeded. Defaults to `45`.
+
 * `new_pod_scale_up_delay` - For scenarios like burst/batch scale where you don't want CA to act before the kubernetes scheduler could schedule all the pods, you can tell CA to ignore unscheduled pods before they're a certain age. Defaults to `10s`.
 
+* `ok_total_unready_count` - Number of allowed unready nodes. Defaults to `3`.
```
katbyte, line 234: ok does not make sense here
```suggestion
* `unready_nodes_allowed` - Number of allowed unready nodes. Defaults to `3`.
```

or for consistancy

```suggestion
* `max_unready_nodes` - Number of allowed unready nodes. Defaults to `3`.
```

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -217,16 +217,22 @@ A `addon_profile` block supports the following:
...
-A `auto_scaler_profile` block supports the following:
+An `auto_scaler_profile` block supports the following:
 
 * `balance_similar_node_groups` - Detect similar node groups and balance the number of nodes between them. Defaults to `false`.
 
 * `expander` - Expander to use. Possible values are `least-waste`, `priority`, `most-pods` and `random`. Defaults to `random`.
 
 * `max_graceful_termination_sec` - Maximum number of seconds the cluster autoscaler waits for pod termination when trying to scale down a node. Defaults to `600`.
 
+* `max_node_provision_time` - Maximum time the autoscaler waits for a node to be provisioned. Defaults to `15m`.
```
katbyte, line 228: grammer
```suggestion
* `max_node_provisioning_time` - Maximum time the autoscaler waits for a node to be provisioned. Defaults to `15m`.
```

website/docs/r/kubernetes_cluster.html.markdown
```diff
@@ -217,16 +217,22 @@ A `addon_profile` block supports the following:
...
+
+* `max_total_unready_percentage` - Maximum percentage of unready nodes the cluster autoscaler will stop if the percentage is exceeded. Defaults to `45`.
```
katbyte, line 230:
```suggestion
* `max_unready_percentage` - Maximum percentage of unready nodes the cluster autoscaler will stop if the percentage is exceeded. Defaults to `45`.
```

katbyte (changes requested):
Thanks @njuCZ - just a couple comments

njuCZ:
@katbyte thanks for your opinion, I have renamed those fields.

katbyte (approved):
Thanks @njuCZ - LGTM 👍

PR #11410, opened 2021-04-21
fix(docs): d/shared_image_version: description error in the gallery_name

aslafy-z:
Fixes an error in the documentation. `The name of the Shared Image **Gallery** in which the Shared Image exists.`  instead of `The name of the Shared Image in which the Shared Image exists.`


tombuildsstuff (approved):
LGTM - thanks for this @aslafy-z

PR #11415, opened 2021-04-21
refactor: renaming a few fields to avoid conflicts with the Plugin SDK script

tombuildsstuff:
Whilst we could make the (simple) Plugin SDK script complex and handle all of these scenarios, it's considerably easier to rename these fields and avoid conflicts


katbyte (approved):
LGTM 👍

PR #11416, opened 2021-04-21
plugin-sdk: introducing a wrapper for the Plugin SDK

tombuildsstuff:
Whilst this PR may initially seem odd, there's a specific problem we're trying to solve - since the Plugin SDK is a major Go version (with breaking changes to the constructs) - rather than doing that in one go and breaking any and all PR's, this PR introduces a wrapper layer allowing us to do this more gradually.

Ultimately upgrading the Plugin SDK to Plugin SDKv2 gradually involves several steps:

1. Moving the State Migrations into `./migration` folders within each service package (#11394)
2. Moving the Validation functions from each package into `./validate (#11378)
3. Introduce this wrapper (done)
4. Updating the Customize Diff functions to use the new wrapper
5. Updating the Import functions to use the new wrappers
6. Updating the Migration wrapper and functions to take `ctx` as the first param
7. Updating the Customize Diff wrapper and functions to take `ctx` as the first param
8. Updating the Import wrapper and functions to take `ctx` as the first param
9. Update the `parse` and `validate` package tests to include `_test` suffix for the package name
10. Gradually updating each service package to use the new wrappers using the upgrade script
11. Upgrade to Plugin SDKv2
12. Update the CRUD|Import methods from `Create` to `CreateContext`

Whilst a little verbose, this means we can do this gradually without causing (m)any conflicts given the size of these changes.

This PR also updates the `advisor` and `analysisservices` packages over to using these aliases using this approach, to show how this works.


aristosvo:
@tombuildsstuff Thanks for not only doing all this amazing work, but also explaining the foreseeable changes for an endeavour like this!

katbyte (approved):
Great work @tombuildsstuff! - LGTM 👍

azurerm/internal/tf/validation/pluginsdk.go
```diff
@@ -0,0 +1,208 @@
+package validation
```
magodo, line 1: Wondering why this file is not located in the `pluginsdk` folder?

azurerm/internal/acceptance/plugin_sdk_aliases.go
```diff
@@ -0,0 +1,59 @@
+package acceptance
```
magodo, line 1: Wondering why this file is not placed to the `tfpluginsdk` folder as others.

magodo (dismissed):
@tombuildsstuff Thank you for the great work 👍

While I have some small questions mainly about the layout of the shim files and the operation feasibility. I'm wondering why most of the shims are located at the package `pluginsdk`, whilst some of them are not, like the `acceptance` and the `validation`. For `validation`, I guess it is to match the same pacakge name as is in the v2 plugin sdk, so that when we have done all the migration, it is easy to remove the wrapper packages. Whilst I'm not sure why `acceptance` here has its own place, since the functions got shimed here ends up in the `resource` package in the v2 sdk. Similarily, the package `tfpluginsdk` seems shim most of the stuff, which in the v2 plugin sdk correspond to multiple packages. It cause the work to remove the shim hardly to be able to do automatically, as you will need to figure out which function/type in the `tfpluginsdk` corresponds to which package in the v2 sdk.

I'm not sure if it is a good idea to follow what is done for the `validation` package, so that it is less work to remove the shim.

Another point is that, maybe we can make use of the AST/DST to do the package replacement, rather than using the `sed`?

azurerm/internal/acceptance/plugin_sdk_aliases.go
tombuildsstuff: these are purely test related (and in this instance a tie-over) - so whilst we _could_ put these in the `pluginsdk` folder, having all of the test items in a single place is itself helpful for discovery (even if the methods in this file are mostly temporary, as a general rule this means we can document the `acceptance` package and how to use it)

azurerm/internal/tf/validation/pluginsdk.go
tombuildsstuff: so `validation` made sense as a separate package to me since this is what I think you'd expect to search for? arguably we could put this `validation` package within the `pluginsdk` folder, but whilst there's some benefits to having the wrapper in a single package - ultimately I don't think that'd be as discoverable tbh?

magodo: I agree that it make sense to put these code to a package with some reasonable name, which helps it easier to find. Whilst it makes it a bit harder to tear down the wrappers later, once we've got migrated all the code to v2. This is some kind of a tradeoff.

tombuildsstuff: FWIW there's some larger questions we're trying to answer regarding the Typed vs Untyped resources - ultimately we're intending on migrating over to Typed resources at some point (so migrating off the wrapper becomes less useful in favour of ditching it in favour of the Typed SDK), but there's some upstream questions to resolve first (which is why this is in limbo at the moment)

tombuildsstuff:
@magodo regarding the AST thing, I'd spent some time looking into that but ultimately this is a short-term script which'll be human-ran/checked, so I don't think there's that much to be gained from switching that at this point? That said, this type of thing could be useful going forwards on a per-package basis with the Untyped resources - but longer-term the Typed resources make this type of refactoring easier (since there's an internal-interface these implement, so it's easier to do this gradually if necessary), so this should be less of an issue?

azurerm/internal/acceptance/plugin_sdk_aliases.go
tombuildsstuff: for now I'm going to merge this since it's blocking several other things - I don't think we should move the test related items out (since they're explicitly test related and will be deprecated in favour of the other test bits) - but the other ones I'm open to splitting out into nested packages presuming we can group these adequately

tombuildsstuff:
Tests look good 👍

PR #11425, opened 2021-04-22
Update `azurerm_linux|windows_virtual_machine_scale_set` - change extension block from List to Set

ArcturusZhang:
fixes https://github.com/terraform-providers/terraform-provider-azurerm/issues/11367

<del>I am only adding the note comment here since changing this type would be a provider breaking change.
Please let me know if we could make the actual changes in this PR.</del>


hbuckle:
Just to note I feel like the current behaviour is a bug, as changing the order of the extensions doesn't change it on the Azure side, so you get a perpetual diff. So it would be great if it could be fixed here, otherwise I will have to keep using the deprecated scale set resource, which has its own bugs to deal with.

tombuildsstuff:
@ArcturusZhang @hbuckle 

Whilst this would technically be a breaking change - tbh I'd think this is probably fine to change prior to 3.0, since it's not a field which users are likely to interpolate from.

That said, switching this from a List to a Set will make this considerably harder to read in the plan, so ultimately this probably needs the new "Dictionary" type rather than being a Set outright - so in practice this'd likely need to change twice, once to a Set and then again to the new Dictionary type once it's available. In either case both of those are going to require migration testing, since the trade-off here is the Plan which will be unclear (particularly if new fields are added to this block over time) - but I don't think there's much we can do there but test this and see how the UX works

WDYT?

hbuckle:
Sounds good to me, the new dictionary type sounds interesting as well.
For our use case we aren't going to be adding and removing extensions very often, we just have lots of existing scale sets where the extensions are stored in a different order on Azure's side, so having a set will allow us to upgrade to the new resource

ArcturusZhang:
Hi @tombuildsstuff if we all agree to migrate it now, I will then make some more commit to make the changes in this PR.

katbyte:
@ArcturusZhang - i think we're good to migrate now, and then again in the future to unblock people. just need to make sure we test old -> new versions of the resource

ArcturusZhang:
> @ArcturusZhang - i think we're good to migrate now, and then again in the future to unblock people. just need to make sure we test old -> new versions of the resource

I will let you know when I have done the testing on this

azurerm/internal/services/compute/virtual_machine_scale_set.go
```diff
@@ -1342,7 +1342,7 @@ func FlattenVirtualMachineScaleSetAutomaticRepairsPolicy(input *compute.Automati
 
 func VirtualMachineScaleSetExtensionsSchema() *schema.Schema {
 	return &schema.Schema{
-		Type:     schema.TypeList,
+		Type:     schema.TypeSet,
```
tombuildsstuff, line 1345: This'll also need the expand functions changed from `.([]interface{})` to `.(*schema.Set).List()`?

ArcturusZhang:
Hi @tombuildsstuff I had some updates, please take a look, thanks

katbyte (approved):
LGTM 👍

PR #11427, opened 2021-04-22
resources: removing some unnecessary state functions on computed fields

tombuildsstuff:
Plugin SDKv2 highlights these can be removed, since they're superflurous:

```
* resource azurerm_tenant_template_deployment: output_content: StateFunc is
extraneous, value should just be changed before setting on computed-only field
* resource azurerm_subscription_template_deployment: output_content:
StateFunc is extraneous, value should just be changed before setting on
computed-only field
* resource azurerm_resource_group_template_deployment: output_content:
StateFunc is extraneous, value should just be changed before setting on
computed-only field
* resource azurerm_management_group_template_deployment: output_content:
StateFunc is extraneous, value should just be changed before setting on
computed-only field
* data source azurerm_template_spec_version: template_body: StateFunc is
extraneous, value should just be changed before setting on computed-only field
```

### DIFF
--- a/azurerm/internal/services/resource/management_group_template_deployment_resource.go
+++ b/azurerm/internal/services/resource/management_group_template_deployment_resource.go
@@ -99,9 +99,8 @@ func managementGroupTemplateDeploymentResource() *schema.Resource {
 
 			// Computed
 			"output_content": {
-				Type:      schema.TypeString,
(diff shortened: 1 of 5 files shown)


manicminer (approved):
LGTM 👍

PR #11428, opened 2021-04-22
Update `azurerm_dedicated_host_group` - support automatic placement

ArcturusZhang:
This PR adds the `automatic_placement_enabled` attribute for `azurerm_dedicated_host_group`, which gives the ability of VMs and VMSSs to directly scales depending a dedicated host group instead of a specific dedicated host.

I will open another PR for VM and VMSS to support dedicated host group after this PR merges.


katbyte (approved):
LGTM 👍

PR #11440, opened 2021-04-22
migrations: switching to use the plugin-sdk wrapper

tombuildsstuff:
This PR switches the migrations over to using the new Plugin SDK wrapper as detailed in #11416 - meaning that each State Migration now takes a `context` which will be populated when we switch over to using Plugin SDKv2

This also fixes a bug I spotted/introduced when refactoring these yesterday regarding the key vault schema (although this was a historical port, so nobody should be using that at this point in time).


katbyte (approved):
LGTM 👍

PR #11447, opened 2021-04-23
plugin-sdk: updating the custom imports to use the Plugin SDK wrapper

tombuildsstuff:
This PR updates the custom importers (that is, `ValidateResourceIDPriorToImport` and `ValidateResourceIDPriorToImportThen`) to use the Plugin SDK Wrapper

This means that these functions now take a context, matching the behaviour in Plugin SDKv2, albeit that's a hard-coded one for now - we'll need to do another sweep of these in a month or so to confirm there's none left after PR's have been merged, but that's a future problem

### DIFF
--- a/azurerm/internal/sdk/wrapper_resource.go
+++ b/azurerm/internal/sdk/wrapper_resource.go
@@ -1,6 +1,7 @@
 package sdk
 
 import (
+	"context"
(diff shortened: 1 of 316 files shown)


manicminer (approved):
LGTM! 🚂

PR #11449, opened 2021-04-23
refactor: updating the default importers to use the plugin-sdk wrapper

tombuildsstuff:
This PR is based on top of #11447 - ultimately this extends that PR to change the Default Importers for resources over to using the Plugin SDK wrapper.

At the same time this adds a TODO to this to add an ID validator to this importer, since this is something which needs to be done at some point.

Depends on #11447


manicminer (approved):
LGTM! 🪐

Issue #11422, opened 2021-04-21
Error: The terraform-provider-azurerm_v2.45.0_x5.exe plugin crashed!


allinonedashboard:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

* `azurerm_api_management`

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "=2.46.0"
    }
  }
}

# Configure the Microsoft Azure Provider
provider "azurerm" {
  features {}
}

data "azurerm_key_vault" "keyvault" {
  name                = "xxxxxxxxxxx"
  resource_group_name = "xxxx-xxxxxx"
}

data "azurerm_key_vault_certificate" "my_key_vault_ssl" {
  name         = "xxxxxxxx"
  key_vault_id = data.azurerm_key_vault.keyvault.id
}

resource "azurerm_api_management" "apimtest" {
  name                = var.name
  location            = var.location
  resource_group_name = var.resource_group_name
  publisher_name      = "Terraform"
  publisher_email     = var.publisher_email
  tags                = var.tags
  sku_name            = var.sku_name
  identity {
    type = "SystemAssigned"
  }
  certificate {
    encoded_certificate  = data.azurerm_key_vault_certificate.my_key_vault_ssl.certificate_data_base64
    store_name           = "Root"
    certificate_password = ""
  }
}

```

### Panic Output

 
panic: interface conversion: interface {} is nil, not map[string]interface {}

goroutine 63 [running]:
github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/apimanagement.expandAzureRmApiManagementHostnameConfigurations(0xc000e96bd0, 0x59d2191)
        /opt/teamcity-agent/work/5d79fe75d4460a2f/src/github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/apimanagement/api_management_resource.go:891 +0xee7
[...]

PR #11452, opened 2021-04-23
Fix potential panics due to nested block contains all optional sub-properties

magodo:
For schemas where a non-computed nested block property of `TypeList` that only contains optional-only and non-default child properties, which has no `AtLeastOneOf`/`ExactlyOneOf` constraint. Then after getting this parent schema from `schema.ResourceData`, it might be a `[]interface{}` with one `nil` element, which corresponds to the case that users has specify an empty block, like:

```hcl
foo {}
```

I've created a linter rule to scan all the schemas that is a composite literal, and found ~100 schemas that need special attention. Otherwise, the use case above will cause panic as reported in #11422.

The fix in this PR follows following rules:

1. If the children properties are addressable (i.e. all its parent properties (if any) are set `MaxItem` to 1), then these `Optional` properties are set the `AtLeastOneOf` constraint to invalidate the empty block use case.
2. Otherwise, either if there is only one child property, or the child properties are not addressable: I will add a lint rule comment and ensure that the expand code for the parent schema does `nil` check. For this case, especially when there are multiple child properties, I'll further ensure there is no pre-size for the expanded arraies. Because we will `continue` the loop if the element under iteration is nil.

Note that there are also some data source which has set the property as `Optional` by mistake, which should be `Computed` actually. I've fixed those, too.

The lint rule comment is used for my local run, to ensure there are no more warnings. Once bflad/tfproviderlint#236 and bflad/tfproviderlint#237 are merged, then these comments can also benefit our code base, if we enable the rule XS003.

The last thing worth mentioning is that even we resolved all the lint warnings from the XS003, there are still cases that missed, including those schema definitions which are not pure composite literal.


katbyte (approved):
Awesome @magodo! LGTM 👍

katbyte:
good to merge once test failure is fixed:
```
=== RUN   TestProvider
    provider_test.go:13: err: 3 errors occurred:
        	* AtLeastOneOf: delete_os_disk_on_deletion configuration block reference (features.0.virtual_machine.0.delete_os_disk_on_deletion) can only be used with TypeList and MaxItems: 1 configuration blocks
        	* resource azurerm_kubernetes_cluster: AtLeastOneOf: admin_group_object_ids references unknown attribute (role_based_access_control.0.azure_activity_directory.0.client_app_id) at part (azure_activity_directory)
        	* resource azurerm_application_gateway: AtLeastOneOf: policy_name configuration block reference (ssl_policy.0.disabled_protocols) can only be used with TypeList and MaxItems: 1 configuration blocks
        
--- FAIL: TestProvider (0.22s)
```

magodo:
@katbyte I've resolved the UT failures. Additionally, I've removed the constraints in the migration files, as they are not necessary to be defined in the migration schemas.

PR #11459, opened 2021-04-25
GNUMakefile: use `go install` to install tools since Go v1.16

magodo:
As we have declared `go 1.16` in `go.mod` and also updated the Github Action workflow files to using Go v1.16, it should make sense to use the now recommanded way to install binaries since v1.16.


katbyte (approved):
LGTM 👍

PR #11461, opened 2021-04-25
 GNUMakefile: Remove `GO111MODULE=on` and `GOFLAGS=-mod=vendor` env vars

magodo:
The two environment variables populated in the GNUMakefile: `GO111MODULE=on` and `GOFLAGS=-mod=vendor` need to remove, since:

- Module mode is now on by default since v1.16
- Setting `GOFLAGS=-mod=vendor` causes `go install` failed with: *cannot query module due to -mod=vendor*


katbyte (approved):
LGTM 👍

PR #11464, opened 2021-04-25
Revert the `AtLeastOneOf` added in #11452, rather work around the panic

magodo:
This PR reverts some of the `AtLeastOneOf` added in #11452, for which properties that are valid to be left as an empty block.

These properties have special meanings when being an empty block. Hence, rather than adding the constraints, make sure there will be no panic when being empty block and add the lint comment instead.


katbyte (approved):
👍

PR #11467, opened 2021-04-25
`postgresql_server`: Update a PostgreSQL Replica to `create_mode` = "Default"

### LINKED ISSUE #11330: azurerm_postgresql_server doesn't disable replication when create_mode is changed.
### Community Note

```
Terraform v0.14.9
+ provider registry.terraform.io/hashicorp/azurerm v2.55.0

Your version of Terraform is out of date! The latest version
is 0.14.10. You can update by downloading from https://www.terraform.io/downloads.html
```
### Affected Resource(s)

* `azurerm_postgresql_server`

Two files:

**postgres.tf**
```hcl
resource "azurerm_resource_group" "rg" {
  name     = var.resource_group
  location = var.main_location
}


resource "azurerm_postgresql_server" "postgres_master" {
  name                             = var.main_db_name
  location                         = var.main_location
  resource_group_name              = azurerm_resource_group.rg.name
  sku_name                         = var.sku_name
  administrator_login              = var.administrator_login
  administrator_login_password     = var.administrator_login_password
  version                          = var.postgres_version
  ssl_enforcement_enabled          = true
  ssl_minimal_tls_version_enforced = "TLS1_2"
  backup_retention_days            = var.retention_days
  geo_redundant_backup_enabled     = var.replicas_count != 0 ? true : false
  storage_mb                       = var.storage_mb
  public_network_access_enabled    = var.public_access_enabled
  auto_grow_enabled                = true

  lifecycle {
    ignore_changes = [
      # Autogrow is enabled
      storage_mb,
    ]
  }

}

resource "azurerm_postgresql_server" "postgres_standby" {
  count                            = var.replicas_count
  name                             = "${azurerm_postgresql_server.postgres_master.name}-r-${var.replicas_count}"
  location                         = var.replicas_location
  resource_group_name              = var.resource_group
  sku_name                         = var.sku_name
  version                          = var.postgres_version
  ssl_enforcement_enabled          = true
  ssl_minimal_tls_version_enforced = "TLS1_2"
  storage_mb                       = var.storage_mb
  public_network_access_enabled    = var.public_access_enabled
  create_mode                      = "Replica"
  creation_source_server_id        = azurerm_postgresql_server.postgres_master.id
  auto_grow_enabled                = true

  lifecycle {
    ignore_changes = [
      # Autogrow is enabled
      storage_mb,
    ]
  }

}
```
**var.tf**
```hcl
provider "azurerm" {
  features {}
}

variable "main_db_name" {
    default = "nfpgsame"
}

variable "main_location" {
    default = "westus2"
}
variable "resource_group" {
    default = "pg-test-same-reg"
}
variable "sku_name" {
    default = "GP_Gen5_2"
}
variable "administrator_login" {
    default = "nilfranadmin"
}
variable "administrator_login_password" {
    default = "superSecure123$"
}
variable "postgres_version" {
    default = 9.6
}
variable "retention_days" {
    default = 7
}
variable "replicas_count" {
    default = 1
}
variable "storage_mb" {
    default = 5120
}
variable "replicas_location" {
    default = "westus2"
}
variable "public_access_enabled" {
    default = false
    }


```
### Debug Output
https://gist.github.com/NillsF/6a3681370e9de26e9a00963164ae5d03

### Expected Behaviour
- When disabling replication, replication should be disabled. 
### Actual Behaviour
- `terraform plan` and `terraform apply` show that TF is trying to make the changes.
- Changes aren't applied on Azure side. 

### Steps to Reproduce
(issue exists if replica is in same of different region)

1. `terraform apply` files as provided
3. In `postgres.tf` remove following lines:
```hcl
  create_mode                      = "Replica"
  creation_source_server_id        = azurerm_postgresql_server.postgres_master.id
```
4. `terraform plan` will show plan output to change the replica:
```

  # azurerm_postgresql_server.postgres_standby[0] will be updated in-place
  ~ resource "azurerm_postgresql_server" "postgres_standby" {
      ~ create_mode                       = "Replica" -> "Default"
      - creation_source_server_id         = "/subscriptions/d19dddf3-9520-4226-a313-ae8ee08675e5/resourceGroups/pg-test-same-reg/providers/Microsoft.DBforPostgreSQL/servers/nfpgsame" -> null
        id                                = "/subscriptions/d19dddf3-9520-4226-a313-ae8ee08675e5/resourceGroups/pg-test-same-reg/providers/Microsoft.DBforPostgreSQL/servers/nfpgsame-r-1"
        name                              = "nfpgsame-r-1"
        tags                              = {}
        # (15 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```
5. `terraform apply` will _seemingly_ apply the changes. 
6. Checking the resource in Azure, the create_mode is unchanged:
```
az postgres server show --name nfpgsame-r-1 --resource-group pg-test-same-reg
```
results in:
```
{
  "administratorLogin": "nilfranadmin",
  "byokEnforcement": "Disabled",
  "earliestRestoreDate": "2021-04-14T21:54:01.920000+00:00",
  "fullyQualifiedDomainName": "nfpgsame-r-1.postgres.database.azure.com",
  "id": "/subscriptions/d19dddf3-9520-4226-a313-ae8ee08675e5/resourceGroups/pg-test-same-reg/providers/Microsoft.DBforPostgreSQL/servers/nfpgsame-r-1",
  "identity": null,
  "infrastructureEncryption": "Disabled",
  "location": "westus2",
  "masterServerId": "/subscriptions/d19dddf3-9520-4226-a313-ae8ee08675e5/resourceGroups/pg-test-same-reg/providers/Microsoft.DBforPostgreSQL/servers/nfpgsame",
  "minimalTlsVersion": "TLS1_2",
  "name": "nfpgsame-r-1",
  "privateEndpointConnections": [],
  "publicNetworkAccess": "Disabled",
  "replicaCapacity": 0,
  "replicationRole": "Replica",
  "resourceGroup": "pg-test-same-reg",
  "sku": {
    "capacity": 2,
    "family": "Gen5",
    "name": "GP_Gen5_2",
    "size": null,
    "tier": "GeneralPurpose"
  },
  "sslEnforcement": "Enabled",
  "storageProfile": {
    "backupRetentionDays": 7,
    "geoRedundantBackup": "Enabled",
    "storageAutogrow": "Enabled",
    "storageMb": 5120
  },
  "tags": {},
  "type": "Microsoft.DBforPostgreSQL/servers",
  "userVisibleState": "Ready",
  "version": "9.6"
}
```

### References

No immediate references, but I have another replica related issue open:
- #11318

aristosvo:
Fixes #11330

## Acceptance tests
- [x] New acceptance test for new functionality: `TestAccPostgreSQLServer_updateReplicaToDefault`
   ```bash
   ❯ make acctests SERVICE='postgres' TESTARGS='-run=updateReplicaToDefault'
  TF_ACC=1 go test -v ./azurerm/internal/services/postgres -run=updateReplicaToDefault -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
  2021/04/25 23:41:55 [DEBUG] not using binary driver name, it's no longer needed
  2021/04/25 23:41:56 [DEBUG] not using binary driver name, it's no longer needed
  === RUN   TestAccPostgreSQLServer_updateReplicaToDefault
  === PAUSE TestAccPostgreSQLServer_updateReplicaToDefault
  === CONT  TestAccPostgreSQLServer_updateReplicaToDefault
  --- PASS: TestAccPostgreSQLServer_updateReplicaToDefault (884.38s)
  PASS
  ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/postgres    887.796s
   ```

### DIFF
--- a/azurerm/internal/services/postgres/postgresql_server_resource.go
+++ b/azurerm/internal/services/postgres/postgresql_server_resource.go
@@ -64,13 +64,30 @@ func resourcePostgreSQLServer() *schema.Resource {
 
 		Importer: &schema.ResourceImporter{
 			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-				if _, err := parse.ServerID(d.Id()); err != nil {
+				client := meta.(*clients.Client).Postgres.ServersClient
+				ctx, cancel := timeouts.ForRead(meta.(*clients.Client).StopContext, d)
+				defer cancel()
+
+				id, err := parse.ServerID(d.Id())
+				if err != nil {
 					return []*schema.ResourceData{d}, err
 				}
 
+				resp, err := client.Get(ctx, id.ResourceGroup, id.Name)
+				if err != nil {
+					return []*schema.ResourceData{d}, fmt.Errorf("reading PostgreSQL Server %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
+				}
+
 				d.Set("create_mode", "Default")
-				if v, ok := d.GetOk("create_mode"); ok && v.(string) != "" {
-					d.Set("create_mode", v)
+				d.Set("creation_source_server_id", "")
+				if *resp.ReplicationRole != "Master" && *resp.ReplicationRole != "None" {
+					d.Set("create_mode", *resp.ReplicationRole)
+
+					masterServerId, err := parse.ServerID(*resp.MasterServerID)
+					if err != nil {
+						return []*schema.ResourceData{d}, fmt.Errorf("parsing Postgres Master Server ID : %v", err)
+					}
+					d.Set("creation_source_server_id", masterServerId.ID())
 				}
 
 				return []*schema.ResourceData{d}, nil
@@ -730,6 +747,9 @@ func resourcePostgreSQLServerUpdate(d *schema.ResourceData, meta interface{}) er
 		},
 		Sku:  sku,
 		Tags: tags.Expand(d.Get("tags").(map[string]interface{})),
+	}
+	if old, new := d.GetChange("create_mode"); postgresql.CreateMode(old.(string)) == postgresql.CreateModeReplica && postgresql.CreateMode(new.(string)) == postgresql.CreateModeDefault {
+		properties.ServerUpdateParametersProperties.ReplicationRole = utils.String("None")
 	}
 
 	future, err := client.Update(ctx, id.ResourceGroup, id.Name, properties)
--- a/azurerm/internal/services/postgres/postgresql_server_resource.go
+++ b/azurerm/internal/services/postgres/postgresql_server_resource.go
@@ -79,8 +79,7 @@ func resourcePostgreSQLServer() *schema.Resource {
 				}
 
 				d.Set("create_mode", "Default")
-				d.Set("creation_source_server_id", "")
-				if *resp.ReplicationRole != "Master" && *resp.ReplicationRole != "None" {
+				if resp.ReplicationRole != nil && *resp.ReplicationRole != "Master" && *resp.ReplicationRole != "None" {
 					d.Set("create_mode", *resp.ReplicationRole)
 
 					masterServerId, err := parse.ServerID(*resp.MasterServerID)
--- a/azurerm/internal/services/postgres/postgresql_server_resource_test.go
+++ b/azurerm/internal/services/postgres/postgresql_server_resource_test.go
@@ -338,7 +338,7 @@ func TestAccPostgreSQLServer_updateReplicaToDefault(t *testing.T) {
 				check.That(data.ResourceName).ExistsInAzure(r),
 			),
 		},
-		data.ImportStep(),
+		data.ImportStep("creation_source_server_id"),
 	})
 }
 
--- a/azurerm/internal/services/postgres/postgresql_server_resource_test.go
+++ b/azurerm/internal/services/postgres/postgresql_server_resource_test.go
@@ -321,6 +321,27 @@ func TestAccPostgreSQLServer_createReplica(t *testing.T) {
 	})
 }
 
+func TestAccPostgreSQLServer_updateReplicaToDefault(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azurerm_postgresql_server", "replica")
+	r := PostgreSQLServerResource{}
+	data.ResourceTest(t, r, []resource.TestStep{
+		{
+			Config: r.createReplica(data, "GP_Gen5_2"),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+			),
+		},
+		data.ImportStep(),
+		{
+			Config: r.updateReplicaToDefault(data, "GP_Gen5_2"),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+			),
+		},
+		data.ImportStep(),
+	})
+}
+
 func TestAccPostgreSQLServer_scaleReplicas(t *testing.T) {
 	data := acceptance.BuildTestData(t, "azurerm_postgresql_server", "test")
 	r := PostgreSQLServerResource{}
@@ -767,6 +788,30 @@ resource "azurerm_postgresql_server" "replica" {
 
   create_mode               = "Replica"
   creation_source_server_id = azurerm_postgresql_server.test.id
+
+  public_network_access_enabled = false
+  ssl_enforcement_enabled       = true
+}
+`, r.template(data, sku, "11"), data.RandomInteger, data.Locations.Secondary, sku)
+}
+
+func (r PostgreSQLServerResource) updateReplicaToDefault(data acceptance.TestData, sku string) string {
+	return fmt.Sprintf(`
+%[1]s
+
+resource "azurerm_resource_group" "replica" {
+  name     = "acctestRG-psql-%[2]d-replica"
+  location = "%[3]s"
+}
+
+resource "azurerm_postgresql_server" "replica" {
+  name                = "acctest-psql-server-%[2]d-replica"
+  location            = "%[3]s"
+  resource_group_name = azurerm_resource_group.replica.name
+
+  sku_name    = "%[4]s"
+  version     = "11"
+  create_mode = "Default"
 
   public_network_access_enabled = false
   ssl_enforcement_enabled       = true


azurerm/internal/services/postgres/postgresql_server_resource.go
```diff
@@ -64,13 +64,30 @@ func resourcePostgreSQLServer() *schema.Resource {
...
+				resp, err := client.Get(ctx, id.ResourceGroup, id.Name)
+				if err != nil {
+					return []*schema.ResourceData{d}, fmt.Errorf("reading PostgreSQL Server %q (Resource Group %q): %+v", id.Name, id.ResourceGroup, err)
+				}
+
 				d.Set("create_mode", "Default")
-				if v, ok := d.GetOk("create_mode"); ok && v.(string) != "" {
-					d.Set("create_mode", v)
+				d.Set("creation_source_server_id", "")
+				if *resp.ReplicationRole != "Master" && *resp.ReplicationRole != "None" {
```
katbyte, line 83: we'll need to nil check ReplicationRole to prevent panics

katbyte (changes requested):
Aside from one comment this looks good 🚀  thanks @aristosvo !

katbyte:
Also most tests are failing with:
```
------- Stdout: -------
=== RUN   TestAccPostgreSQLServer_basicEleven
=== PAUSE TestAccPostgreSQLServer_basicEleven
=== CONT  TestAccPostgreSQLServer_basicEleven
    testing_new_import_state.go:249: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) (len=1) {
         (string) (len=25) "creation_source_server_id": (string) ""
        }
        
        
        (map[string]string) {
        }
--- FAIL: TestAccPostgreSQLServer_basicEleven (393.28s)
FAIL

------- Stderr: -------
2021/04/26 01:01:46 [DEBUG] not using binary driver name, it's no longer needed
2021/04/26 01:01:47 [DEBUG] not using binary driver name, it's no longer needed
```

azurerm/internal/services/postgres/postgresql_server_resource.go
aristosvo: Thanks, done!

aristosvo:
> Also most tests are failing with:
> 
> ```
> ------- Stdout: -------
> === RUN   TestAccPostgreSQLServer_basicEleven
> === PAUSE TestAccPostgreSQLServer_basicEleven
> === CONT  TestAccPostgreSQLServer_basicEleven
>     testing_new_import_state.go:249: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
>         
>         (map[string]string) (len=1) {
>          (string) (len=25) "creation_source_server_id": (string) ""
>         }
>         
>         
>         (map[string]string) {
>         }
> --- FAIL: TestAccPostgreSQLServer_basicEleven (393.28s)
> FAIL
> 
> ------- Stderr: -------
> 2021/04/26 01:01:46 [DEBUG] not using binary driver name, it's no longer needed
> 2021/04/26 01:01:47 [DEBUG] not using binary driver name, it's no longer needed
> ```

Oops, my mistake. I took the gamble as I had no credits left in my Azure Visual Studio Enterprise Subscription to role the whole suite.

Setting `creation_source_server_id = null` on the new test to should fix it instead of `d.Set("creation_source_server_id", "")` on import

mybayern1974:
Hi @aristosvo ,

Thank you for raising this PR! Given I’m also subscribing the issue [11330](https://github.com/terraform-providers/terraform-provider-azurerm/issues/11330), I’d like to share my 2 cents though I believe this fix is able to solve this issue.
1.	At the service backend, there are two parameters `CreateMode` and `ReplicationRole`. The former only serves CREATE/PUT and the latter only serves UPDATE/PATCH. Given what the issue wants is the latter, will it be more clear to directly expose the latter property rather than making the magic happen through making the former as a wrapper?
2.	The purpose of the issue is to make both the `Primary` and `Replica` server [become `Standalone` server](https://docs.microsoft.com/en-us/azure/postgresql/concepts-read-replicas#stop-replication--promote-replica) by passing `ReplicationRole=None` to the backend. Directly exposing the `ReplicationRole` may help make users clear on what they are doing, while exposing `CreateMode=Replica ==> CreateMode=Default` as this PR does may bring the unclear to some users: What does `Default` mode mean?
3.	Despite of which implementation solution would finally be, suggest to add below changes in doc and codes:
- Tell users that passing `ReplicationRole=None` is irreversible.
- In the code, add check that if in-update() && `ReplicationRole=None` && original-mode-is-replica, users must erase `creation_source_server_id` in their tf config, because at the service backend, when `ReplicationRole=None` gets passed to the backend, the service will erase the `creation_source_server_id` in the backend which will bring state inconsistency between client and service. This concern indeed has been detected by the regression testing result provided by @ Kt. Besides adding this detection in code, suggest also mention it in the doc.

As for follow ups, I would appreciate if the PR could be tuned per above suggest or I would be happy to submit a PR embodying the above if we can reach agreement here. I do not object to the current solution, either, if we agree on it, which could also solve the issue, as long as necessary in-code check and doc gets enriched.

Lastly, I cannot tell for now whether there would be other problems before we running full regression testing.

aristosvo:
Hi @mybayern1974!

Thanks for your comments, like it! Have thought about exposing `replication_role`, but it didn't fit in my Desired State-thinking to have a separate flag for creation and updates 🤣 I believe it may be better to expose that extra variable, but it will be confusing as well, I'd prefer a better API from Azure side :)

I'll for sure add some extra docs! Thanks for the pointers in that regard.

> In the code, add check that if in-update() && ReplicationRole=None && original-mode-is-replica, users must erase creation_source_server_id in their tf config, because at the service backend, when ReplicationRole=None gets passed to the backend, the service will erase the creation_source_server_id in the backend which will bring state inconsistency between client and service. This concern indeed has been detected by the regression testing result provided by @ Kt. Besides adding this detection in code, suggest also mention it in the doc.
With regard to your last bullet point, can you explain what you mean? `creation_source_server_id` is only used on creation time, so there is always state inconsistency if this flag is filled after creation.  It also looks like this state variable cannot be removed completely (causing me trouble in the diff between import and state), but that is a minor problem in my opinion.

katbyte (approved):
thanks @aristosvo - tests pass now so i'm going to approve and merge this so the fix makes it into this weeks release and any additional fixes can be done in a subsequent pr. I think it's reasonable to expect going to expect that diff there and  while not ideal agree its minor

aristosvo:
@mybayern1974 Feel free to create another PR to expose `replication_role` as this one is merged.

Can you also pick up the docs from there, as it is double the effort if I do it for current workings as well!?

PR #11468, opened 2021-04-26
fix TC for containers

njuCZ:
list all available kubernetes version: `az aks get-versions --location westus`
`1.19.6` is not supported, that's why a lot of acctests fail.


katbyte (approved):
Thanks @njuCZ - LGTM 👍

PR #11470, opened 2021-04-26
fix TC for lighthouse

njuCZ:
1. if env is not set, skip the acctest
2. `scope` is a required property in resource `azurerm_lighthouse_definition`

the acctest result when providing the required env
image


katbyte (approved):
Thanks @njuCZ - LGTM 👍

Issue #10632, opened 2021-02-18
Support for customer managed key setting for azurerm_healthcare_service


sven-hoffmann:
## Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Azure API for FHIR now supports the usage of customer managed keys for encyrption at rest in its configuration. This new setting is currently not available in terraform. The concrete example for ARM templates (or other ways) can be found at the official documentation site: https://docs.microsoft.com/en-in/azure/healthcare-apis/customer-managed-key#using-azure-resource-manager-template

Concretely, here's a reduced example:

```
{
            "type": "Microsoft.HealthcareApis/services",
            "apiVersion": "2020-03-30",
            <snip>
            "properties": {
                <snip>
                "cosmosDbConfiguration": {
                    "offerThroughput": 400,
                    "keyVaultKeyUri": "https://<my-vault>.vault.azure.net/keys/<my-key>"
                },
               <snip>
            }
        }
```

### New or Affected Resource(s)

* azurerm_healthcare_service

### Potential Terraform Configuration

I could imagine either:

```hcl
resource "azurerm_healthcare_service" "example" {
  name                = "uniquefhirname"
  resource_group_name = "sample-resource-group"
  location            = "westus2"
  kind                = "fhir-R4"
  cosmosdb_throughput = "2000"
  cosmosdb_keyVaultKeyUri = "https://<my-vault>.vault.azure.net/keys/<my-key>"
}
```

Or this:

```hcl
resource "azurerm_healthcare_service" "example" {
  name                = "uniquefhirname"
  resource_group_name = "sample-resource-group"
  location            = "westus2"
  kind                = "fhir-R4"
  cosmosdb_configuration = {
    throughput = "2000"
    keyVaultKeyUri = "https://<my-vault>.vault.azure.net/keys/<my-key>"
  }
}
```

I'm leaning towards the second option but the first one is backwards compatible and a non-breaking change.

### References

* https://docs.microsoft.com/en-in/azure/healthcare-apis/customer-managed-key

PR #11481, opened 2021-04-26
azurerm_healthcare_service: support for customer managed key

b4rd:
This PR adds CMK support for `azurerm_healthcare_service`.

Reference document: https://docs.microsoft.com/en-in/azure/healthcare-apis/fhir/customer-managed-key

Fixes: [#10632](https://github.com/terraform-providers/terraform-provider-azurerm/issues/10632)

### Test Results
```
TF_ACC=1 go test -v ./azurerm/internal/services/healthcare  -timeout 60m -ldflags="-X=github.com/terraform-providers/terraform-provider-azurerm/version.ProviderVersion=acc"
2021/04/27 12:58:43 [DEBUG] not using binary driver name, it's no longer needed
2021/04/27 12:58:43 [DEBUG] not using binary driver name, it's no longer needed
=== RUN   TestAccHealthCareServiceDataSource_basic
=== PAUSE TestAccHealthCareServiceDataSource_basic
=== RUN   TestAccHealthCareServiceDataSource_complete
=== PAUSE TestAccHealthCareServiceDataSource_complete
=== RUN   TestAccHealthCareService_basic
=== PAUSE TestAccHealthCareService_basic
=== RUN   TestAccHealthCareService_requiresImport
=== PAUSE TestAccHealthCareService_requiresImport
=== RUN   TestAccHealthCareService_complete
=== PAUSE TestAccHealthCareService_complete
=== CONT  TestAccHealthCareServiceDataSource_basic
=== CONT  TestAccHealthCareService_requiresImport
=== CONT  TestAccHealthCareService_complete
=== CONT  TestAccHealthCareService_basic
=== CONT  TestAccHealthCareServiceDataSource_complete
--- PASS: TestAccHealthCareServiceDataSource_basic (1112.86s)
--- PASS: TestAccHealthCareService_basic (1115.46s)
--- PASS: TestAccHealthCareService_requiresImport (1189.59s)
--- PASS: TestAccHealthCareServiceDataSource_complete (1219.64s)
--- PASS: TestAccHealthCareService_complete (1223.75s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/healthcare  1224.166s
```


azurerm/internal/services/healthcare/healthcare_service_resource.go
```diff
@@ -60,11 +62,27 @@ func resourceHealthcareService() *schema.Resource {
...
+						},
+						"key_vault_key_id": {
+							Type:             schema.TypeString,
+							Optional:         true,
+							ForceNew:         true,
+							DiffSuppressFunc: diffSuppressIgnoreKeyVaultKeyVersion,
+							ValidateFunc:     keyVaultValidate.VersionlessNestedItemId,
+						},
+					},
+				},
```
manicminer, line 85: Unfortunately we can't remove properties without first deprecating them as this would be a breaking change for users with existing configurations. Can we restore the original `cosmosdb_throughput` property and rather add a toplevel `keyvault_key_id` property? We can look at adjusting the schema for v3.0 of the provider.

azurerm/internal/services/healthcare/healthcare_service_data_source.go
```diff
@@ -37,9 +37,21 @@ func dataSourceHealthcareService() *schema.Resource {
...
+						"throughput": {
+							Type:     schema.TypeInt,
+							Computed: true,
+						},
+						"key_vault_key_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
```
manicminer, line 54: Can we amend the data source to have properties matching the resource?

azurerm/internal/services/healthcare/healthcare_service_resource.go
```diff
@@ -2,10 +2,12 @@ package healthcare
 
 import (
 	"fmt"
+	keyVaultParse "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/keyvault/parse"
+	keyVaultValidate "github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/keyvault/validate"
```
manicminer, line 6: Can we move these imports into the next section, with the other imports from the same module?

manicminer (changes requested):
Hi @b4rd, thanks for the PR! This is off to a great start, the main issue with the proposed implementation is that it removes an existing property which is considered a breaking change. Whilst we can look to adjust the schema in such a way in a major version bump, such changes can't be included in a regular (minor) release. I've made a suggestion inline, if you can restore the original property and add the new one alongside it then I'll be happy to review again.

azurerm/internal/services/healthcare/healthcare_service_resource.go
b4rd: I decided to use `cosmosdb_key_vault_key_id` for the attribute name, I think it better describes its purpose. If you strongly believe that `keyvault_key_id` is more suitable, I'm opt for changing it.

azurerm/internal/services/healthcare/healthcare_service_data_source.go
b4rd: Is this still relevant?

manicminer: We should maintain the existing `cosmosdb_throughput` attribute at least until v3.0, and I think it makes sense to have the same schema in the data source as in the corresponding resource.

azurerm/internal/services/healthcare/healthcare_service_resource.go
b4rd: Fixed

azurerm/internal/services/healthcare/healthcare_service_data_source.go
b4rd: This should be fixed now. I might be wrong, but I think that the data source attributes should have their `Computed` properties set to `true`, and they don't necessarily need to be the same as their resource counterparts.

manicminer: Sorry that is correct of course, they should be Computed where they are not used as input properties. In terms of matching the names / nesting with the resource, it's not _mandatory_ but it's good to be consistent :)

azurerm/internal/services/healthcare/healthcare_service_data_source.go
```diff
@@ -42,6 +42,11 @@ func dataSourceHealthcareService() *schema.Resource {
 				Computed: true,
 			},
 
+			"cosmosdb_key_vault_key_id": {
```
katbyte, line 45: could we include that it needs to be versionless here?
```suggestion
			"cosmosdb_key_vault_key_versionless_id": {
```

katbyte (changes requested):

azurerm/internal/services/healthcare/healthcare_service_data_source.go
b4rd: I was inspired by the cosmosdb attribute [here](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/cosmosdb_account#key_vault_key_id). If you don't mind this inconsistency I'm fine with the suggestion.

b4rd: Do you still think I should make this change @katbyte?

katbyte: Not at all, we're going to move toward including versionless going forward to indicate what type of key to use

b4rd: I updated the name of this attribute.

azurerm/internal/services/healthcare/healthcare_service_data_source.go
```diff
@@ -140,11 +145,16 @@ func dataSourceHealthcareServiceRead(d *schema.ResourceData, meta interface{}) e
...
-			cosmosThroughput = int(*props.CosmosDbConfiguration.OfferThroughput)
+		cosmosDbThroughput := 0
+		if cosmos := props.CosmosDbConfiguration; cosmos != nil {
+			if v := cosmos.OfferThroughput; v != nil {
+				cosmosDbThroughput = int(*v)
+			}
+			if v := cosmos.KeyVaultKeyURI; v != nil {
+				d.Set("cosmosdb_key_vault_key_versionless_id", v)
+			}
 		}
```
manicminer, line 156: We should set the attribute regardless of API response (as the `cosmosdb_throughput` attribute is).

```suggestion
		cosmodDbKeyVaultKeyVersionlessId := ""
		cosmosDbThroughput := 0
		if cosmos := props.CosmosDbConfiguration; cosmos != nil {
			if v := cosmos.OfferThroughput; v != nil {
				cosmosDbThroughput = int(*v)
			}
			if v := cosmos.KeyVaultKeyURI; v != nil {
				cosmodDbKeyVaultKeyVersionlessId = *v
			}
		}
		d.Set("cosmosdb_key_vault_key_versionless_id", cosmodDbKeyVaultKeyVersionlessId)
```

azurerm/internal/services/healthcare/healthcare_service_data_source_test.go
```diff
@@ -24,12 +24,26 @@ func TestAccHealthCareServiceDataSource_basic(t *testing.T) {
 				check.That(data.ResourceName).Key("resource_group_name").Exists(),
 				check.That(data.ResourceName).Key("kind").Exists(),
 				check.That(data.ResourceName).Key("cosmosdb_throughput").Exists(),
+				check.That(data.ResourceName).Key("cosmosdb_key_vault_key_versionless_id").DoesNotExist(),
```
manicminer, line 27: With the above, I think we need to check for a zero value here

```suggestion
				check.That(data.ResourceName).Key("cosmosdb_key_vault_key_versionless_id").HasValue(""),
```

azurerm/internal/services/healthcare/healthcare_service_resource.go
```diff
@@ -282,11 +295,16 @@ func resourceHealthcareServiceRead(d *schema.ResourceData, meta interface{}) err
...
-			cosmosThroughput = int(*props.CosmosDbConfiguration.OfferThroughput)
+		cosmosDbThroughput := 0
+		if cosmos := props.CosmosDbConfiguration; cosmos != nil {
+			if v := cosmos.OfferThroughput; v != nil {
+				cosmosDbThroughput = int(*v)
+			}
+			if v := cosmos.KeyVaultKeyURI; v != nil {
+				d.Set("cosmosdb_key_vault_key_versionless_id", v)
+			}
 		}
```
manicminer, line 306: As above, this should be set every time, using a zero value default

```suggestion
		cosmodDbKeyVaultKeyVersionlessId := ""
		cosmosDbThroughput := 0
		if cosmos := props.CosmosDbConfiguration; cosmos != nil {
			if v := cosmos.OfferThroughput; v != nil {
				cosmosDbThroughput = int(*v)
			}
			if v := cosmos.KeyVaultKeyURI; v != nil {
				cosmodDbKeyVaultKeyVersionlessId = *v
			}
		}
		d.Set("cosmosdb_key_vault_key_versionless_id", cosmodDbKeyVaultKeyVersionlessId)
```

manicminer (changes requested):
@b4rd Thanks for making the changes! This is looking great, I have made a couple of suggestions inline below and committed them. Pending test results, this should be good to merge 👍

manicminer (approved):
Test results

Screenshot 2021-04-29 at 13 47 54

PR #11486, opened 2021-04-27
Support for `access_key_metadata_writes_enabled`, `mongo_server_version` and `network_acl_bypass` in `azurerm_cosmosdb_account`

yupwei68:
Fix https://github.com/terraform-providers/terraform-provider-azurerm/issues/11487

### DIFF
--- a/azurerm/internal/services/cosmos/cosmosdb_account_resource.go
+++ b/azurerm/internal/services/cosmos/cosmosdb_account_resource.go
@@ -109,12 +109,14 @@ func resourceCosmosDbAccount() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Default:  false,
+				ForceNew: true,
 			},
 
 			"analytical_storage_enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Default:  false,
+				ForceNew: true,
 			},
 
 			"public_network_access_enabled": {
@@ -275,7 +277,7 @@ func resourceCosmosDbAccount() *schema.Resource {
 				Default:  false,
 			},
 
-			"key_based_meta_write_access_enabled": {
+			"access_key_metadata_writes_enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Default:  true,
@@ -293,14 +295,10 @@ func resourceCosmosDbAccount() *schema.Resource {
 				}, false),
 			},
 
-			"network_acl_bypass": {
-				Type:     schema.TypeString,
+			"network_acl_bypass_for_azure_servers": {
+				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  string(documentdb.NetworkACLBypassNone),
-				ValidateFunc: validation.StringInSlice([]string{
-					string(documentdb.NetworkACLBypassNone),
-					string(documentdb.NetworkACLBypassAzureServices),
-				}, false),
+				Default:  false,
 			},
 
 			"network_acl_bypass_ids": {
@@ -454,6 +452,11 @@ func resourceCosmosDbAccountCreate(d *schema.ResourceData, meta interface{}) err
 		publicNetworkAccess = documentdb.Disabled
 	}
 
+	networkByPass := documentdb.NetworkACLBypassNone
+	if d.Get("network_acl_bypass_for_azure_servers").(bool) {
+		networkByPass = documentdb.NetworkACLBypassAzureServices
+	}
+
 	account := documentdb.DatabaseAccountCreateUpdateParameters{
 		Location: utils.String(location),
 		Kind:     documentdb.DatabaseAccountKind(kind),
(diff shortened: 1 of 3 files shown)


website/docs/r/cosmosdb_account.html.markdown
```diff
@@ -106,6 +106,16 @@ The following arguments are supported:
 
 * `enable_multiple_write_locations` - (Optional) Enable multi-master support for this Cosmos DB account.
 
+* `key_based_meta_write_access_enabled` - (Optional) Is write operations on metadata resources (databases, containers, throughput) via account keys enabled? Defaults to `true`.
+
+* `mongo_server_version` - (Optional) The Server Version of a MongoDB account. Possible values are `4.0`, `3.6`, and `3.2`. Changing this forces a new resource to be created.
+
+* `network_acl_bypass` - (Optional) The services which are allowed to bypass firewall checks. Possible values are `None` and `AzureServices`. Defaults to `None`.
+
+* `network_acl_bypass_ids` - (Optional) A list of resource Ids for Network Acl Bypass for this Cosmos DB account.
```
katbyte, line 115: grammer

website/docs/r/cosmosdb_account.html.markdown
```diff
@@ -106,6 +106,16 @@ The following arguments are supported:
 
 * `enable_multiple_write_locations` - (Optional) Enable multi-master support for this Cosmos DB account.
 
+* `key_based_meta_write_access_enabled` - (Optional) Is write operations on metadata resources (databases, containers, throughput) via account keys enabled? Defaults to `true`.
+
+* `mongo_server_version` - (Optional) The Server Version of a MongoDB account. Possible values are `4.0`, `3.6`, and `3.2`. Changing this forces a new resource to be created.
+
+* `network_acl_bypass` - (Optional) The services which are allowed to bypass firewall checks. Possible values are `None` and `AzureServices`. Defaults to `None`.
```
katbyte, line 113: will there be other vcalues here?  this may be better as
```suggestion
* `network_acl_bypass_for_azure_servers` - (Optional) If azure services can bypass ACLs. Defaults to `false`.
```

website/docs/r/cosmosdb_account.html.markdown
```diff
@@ -106,6 +106,16 @@ The following arguments are supported:
 
 * `enable_multiple_write_locations` - (Optional) Enable multi-master support for this Cosmos DB account.
 
+* `key_based_meta_write_access_enabled` - (Optional) Is write operations on metadata resources (databases, containers, throughput) via account keys enabled? Defaults to `true`.
+
+* `mongo_server_version` - (Optional) The Server Version of a MongoDB account. Possible values are `4.0`, `3.6`, and `3.2`. Changing this forces a new resource to be created.
```
katbyte, line 111: since the default version was 3.2 before this, would it make sense to default it to 3.2 to be consistent with the previous version and remove that breaking change? and then in 3.0 we can change it to 3.6?

also would that mean the mogo index _id is no longer required if we use 3.2 here? also removing that breaking change?

website/docs/r/cosmosdb_account.html.markdown
```diff
@@ -106,6 +106,16 @@ The following arguments are supported:
 
 * `enable_multiple_write_locations` - (Optional) Enable multi-master support for this Cosmos DB account.
 
+* `key_based_meta_write_access_enabled` - (Optional) Is write operations on metadata resources (databases, containers, throughput) via account keys enabled? Defaults to `true`.
```
katbyte, line 109: this makes little sense, by account keys does it mean the access keys? are there are ways you can write?
```suggestion
* `access_key_metadata_writes_enabled` - (Optional) Is write operations on metadata resources (databases, containers, throughput) via account keys enabled? Defaults to `true`.
```

katbyte (changes requested):
Thanks @yupwei68 - have some comment about schema and if we can remove some of the breaking changes in this release

website/docs/r/cosmosdb_account.html.markdown
yupwei68: Since the breaking change has happened for several months, I think the users have all changed their scripts. (Since it's a breaking change from service side). Do we need to break that back?

katbyte: didn't we only update the API now, so 2.57 is when we would start defaulting to 3.6, however if we default this to 3.2 it means there effectively would be no change?

yupwei68: No. We only update the tests in 2.57 to adapt to the service changes. The changes have happened for several months.

azurerm/internal/services/cosmos/cosmosdb_account_resource.go
```diff
@@ -277,6 +277,38 @@ func resourceCosmosDbAccount() *schema.Resource {
...
+				ForceNew: true,
+				Computed: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					string(documentdb.ThreeFullStopTwo),
+					string(documentdb.ThreeFullStopSix),
+					string(documentdb.FourFullStopZero),
+				}, false),
+			},
+
+			"network_acl_bypass_for_azure_servers": {
```
katbyte, line 298: shouldn't this be
```suggestion
			"network_acl_bypass_for_azure_services": {
```
?

katbyte (changes requested):

yupwei68:
Hi kt, tests have passed. Please continue reviewing.
image

katbyte (approved):
Thanks @yupwei68 - LGTM 👍

PR #11490, opened 2021-04-27
docs: add missing underscore in http_get reference

wernerfred:
add missing underscore in http_get block reference


manicminer (approved):
Hi @wernerfred, many thanks for this documentation fix, this LGTM!

PR #11500, opened 2021-04-28
generate id parser for `azurerm_public_ip_prefix`

njuCZ:
through the comment of the https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/network/public_ip_prefix_resource_test.go#L91-L92, it seems it's by design that acctest `TestAccPublicIpPrefix_prefixLength24` would fail in my sub. Others could pass.

image


katbyte (approved):
LGTM 👍

PR #11509, opened 2021-04-28
Acceptance Tests bug fixes

paulomarquesc:
Fixing the following issues with some of Azure NetApp Files acceptance tests:

* Usage of hash strings to test for values which is not valid anymore
* Removal of tag related tests for volume snapshots since it is not supported anymore by this type of resource.


katbyte (approved):
Thanks @paulomarquesc - LGTM 👍

Issue #11511, opened 2021-04-28
Empty resource group name gives a misleading error messages


paulroub:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version

Terraform v0.15.1 on darwin_amd64

### Affected Resource(s)

* `azurerm_2.52.0`

### Terraform Configuration Files

```hcl
resource "azurerm_application_insights" "OurApp" {
  name                = var.APP_NAME
  location            = var.resource_location
  resource_group_name = var.APP_RESOURCE_GROUP_NAME
  application_type    = "web"
}
```

Where `APP_RESOURCE_GROUP_NAME` is empty.

### Expected Behaviour

An error message indicating that resource group names can't be empty.

### Actual Behaviour

An error message indicating that resource group names "may only contain alphanumeric characters, dash, underscores, parentheses and periods", which obscured the actual cause (an unset environment variable, in this case).

### Steps to Reproduce

1. `terraform apply`

favoretti:
Closed by #11512

favoretti:
Thanks @paulroub!

PR #11512, opened 2021-04-28
Empty resource group name gives less-misleading error

paulroub:
Addresses issue #11511.

We now return a specific error message when a resource group name is empty, rather than a list of disallowed characters that an empty string -- by definition -- can't possibly contain.


katbyte (approved):
Thanks @paulroub - LGTM 👍

PR #11514, opened 2021-04-28
Remove default value for `default_service_version` in `blob_properties` in `azurerm_storage_account`

yupwei68:
The `default_service_version` may change over time. So it may cause breaking changes in the future.


katbyte (approved):
LGTM 👍

yupwei68:
Thanks kt

yupwei68:
Hi kt, tests have passed. NFSv3 is failed for a known configuration problem only in west europe and france central. The fix deployment is in process. 
image

Issue #11292, opened 2021-04-12
Error: setting `data_protection_replication`: Invalid address to set: []string{"data_protection_replication", "0", "replication_frequency"}


BartoszDopke:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureRM Provider) Version
Terraform v0.14.4
azurerm v2.55.0

### Affected Resource(s)

* `azurerm_netapp_volume`

### Terraform Configuration Files

```hcl
data "azurerm_netapp_volume" "volume" {
  account_name        = "account_name"
  name                = "name"
  pool_name           = "pool_name"
  resource_group_name = "resource_group_name"
}
```

### Debug Output

https://gist.github.com/BartoszDopke/9569a3df08063cea886bf683729e3d69

### Expected Behaviour

Terraform should read existing NetApp Volume data but it doesn't.

### Actual Behaviour
```
Error: setting `data_protection_replication`: Invalid address to set: []string{"data_protection_replication", "0", "replication_frequency"}
```
### Steps to Reproduce

1. `terraform init`
2. `terraform plan`

### References

* #0000

PR #11530, opened 2021-04-29
Renamed replication_schedule to replication_frequence on Azure NetApp Files volume data source

paulomarquesc:
Fixes bug #11292, where the data source resource didn't match the correct name.


manicminer (approved):
Hi @paulomarquesc, thanks for this bugfix! I've pushed a documentation fix to add the missing attributes and this LGTM 👍

PR #11534, opened 2021-04-29
Adding replication_schedule back as deprecated property

paulomarquesc:
Adding replication_schedule back as deprecated property


katbyte (approved):
LGTM 👍

PR #11560, opened 2021-05-03
azurerm_key_vault_certificate: Compute key_size when key_type is lowercase

simonbrady:
#10867 computes `key_properties.key_size` when `key_properties.key_type` is `EC`, but not when it's `ec` (which is also allowed by that PR). Let's fix that.

### DIFF
--- a/azurerm/internal/services/keyvault/key_vault_certificate_resource.go
+++ b/azurerm/internal/services/keyvault/key_vault_certificate_resource.go
@@ -144,6 +144,9 @@ func resourceKeyVaultCertificate() *schema.Resource {
 										Type:     schema.TypeString,
 										Required: true,
 										ForceNew: true,
+										DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+											return old == strings.ToUpper(new)
+										},
 										ValidateFunc: validation.StringInSlice([]string{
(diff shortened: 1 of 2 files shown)


katbyte:
Hey @simonbrady - Thanks for the PR but i'm not sure what is the motivation to support `ec` is - terraform and the provider is generally case sensitive unless working around an API bug and `EC` is the casing used by the azure SDK.

simonbrady:
Thanks for the feedback @katbyte - this came up running a local testcase I'd written for #10867 before @manicminer kindly finished off that PR for me. I was thinking we'd missed an edge case, but now that I think about it more that's not really true. The support for lowercase `key_type` was added for existing `rsa` (not `RSA`) certs, and computing `key_size` is only for `EC` certs which didn't work before the PR so there can't be existing `ec` configurations that we need to support.

tl;dr yes the original PR has a small hole in it, but I'm solving a problem here that doesn't matter so this one can probably be closed.

manicminer:
Just noting that the case insensitivity is for compatibility to avoid breaking existing configurations since the validation was recently added, and the auto key size functionality is new.

simonbrady:
Agreed, that's what I should have thought about before rushing to "fix" this, so happy to close this PR as unnecessary. Thanks again for the feedback everyone.

PR #11552, opened 2021-05-01
New Resource - sentinel_alert_rule_machine_learning_behavior_analytics

kaovd:
PR implements Azure Sentinel ML Behavior Analytics rules as per #11551
This quite literally just copies from Fusion rules as it has the exact same properties. 

# Reference

- Properties/Spec Ref: https://github.com/Azure/bicep-types-az/blob/9d7f9f5caab4cf0b9645aee3dd6328406f0904ab/generated/securityinsights/microsoft.securityinsights/2019-01-01-preview/types.md#mlbehavioranalyticsalertruleproperties
- API: https://github.com/AzureSDKAutomation/azure-resource-manager-schemas/pull/1487/commits/e31fdb6c651295486de179225dae79bd8a02aa7c

# Test Results
```
2021/05/01 22:51:52 [DEBUG] not using binary driver name, it's no longer needed
2021/05/01 22:51:52 [DEBUG] not using binary driver name, it's no longer needed
=== RUN   TestAccSentinelAlertRuleMLBehaviorAnalytics_basic
=== PAUSE TestAccSentinelAlertRuleMLBehaviorAnalytics_basic
=== RUN   TestAccSentinelAlertRuleMLBehaviorAnalytics_complete
=== PAUSE TestAccSentinelAlertRuleMLBehaviorAnalytics_complete
=== RUN   TestAccSentinelAlertRuleMLBehaviorAnalytics_update
=== PAUSE TestAccSentinelAlertRuleMLBehaviorAnalytics_update
=== RUN   TestAccSentinelAlertRuleMLBehaviorAnalytics_requiresImport
=== PAUSE TestAccSentinelAlertRuleMLBehaviorAnalytics_requiresImport
=== CONT  TestAccSentinelAlertRuleMLBehaviorAnalytics_basic
=== CONT  TestAccSentinelAlertRuleMLBehaviorAnalytics_update
=== CONT  TestAccSentinelAlertRuleMLBehaviorAnalytics_complete
=== CONT  TestAccSentinelAlertRuleMLBehaviorAnalytics_requiresImport
--- PASS: TestAccSentinelAlertRuleMLBehaviorAnalytics_basic (253.75s)
--- PASS: TestAccSentinelAlertRuleMLBehaviorAnalytics_complete (255.03s)
--- PASS: TestAccSentinelAlertRuleMLBehaviorAnalytics_requiresImport (363.84s)
--- PASS: TestAccSentinelAlertRuleMLBehaviorAnalytics_update (563.17s)
[...]

PR #11593, opened 2021-05-05
deps: Update Go SDK to 54.0.0

favoretti:


azurerm/internal/services/resource/resource_group_resource.go
```diff
@@ -124,7 +124,8 @@ func resourceResourceGroupDelete(d *schema.ResourceData, meta interface{}) error
 		return err
 	}
 
-	deleteFuture, err := client.Delete(ctx, id.ResourceGroup)
+	// @favoretti: This is currently the only supported added value, do we want to add it as an optional parameter to the schema?
+	deleteFuture, err := client.Delete(ctx, id.ResourceGroup, "Microsoft.Compute/virtualMachines,Microsoft.Compute/virtualMachineScaleSets")
```
favoretti, line 128: @katbyte @tombuildsstuff Opinions on how to go about this one?

jackofallops: Just taken a quick look and for maintaining existing behaviour I think this should to be:
```suggestion
	deleteFuture, err := client.Delete(ctx, id.ResourceGroup, "")
```

favoretti: @jackofallops thank you, commited suggestion. Wasn't sure whether we want it as an empty string or not.
I think we can extrapolate that into a separate argument once this gets in?

favoretti:
Rebase, rename file introduced in #11552

favoretti:
@jackofallops this should now be good to go, if you could take another look please?

katbyte (approved):
LGTM 👍

Issue #10217, opened 2021-01-17
Deprecated sql_server but no note on top


ghost:
Could you add a 'this resource is deprecated' on top? because people land on the deprecated resource a lot and like me spend hours searching on what the differences are between sql_server and mssql_server 
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/sql_server

tombuildsstuff:
hi @Dakpan94 

Thanks for opening this issue.

This resource isn't strictly deprecated - there's two usage models (sku vs vcores) available in the two different resources, which is why these are separate, however (as I understand it) Microsoft are no longer adding new functionality to the older `sku` based model, rather recommending folks switch to using the `vcores` instead - but as far as I understand it this hasn't been officially deprecated at this point in time.

Since this resource isn't strictly deprecated, at this point in time deprecating this resource is likely unnecessary - but I could see an argument for cross-linking between the two resources to better highlight this, something along the lines of:

```
~> **Note:** The `azurerm_mssql_server` resource provides support for `vcores` and is recommended going forwards - the `azurerm_sql_server` resource (this resource) provides support for the older `sku` based model.
```

What do you think?

Thanks!

Dakpan94:
Hi @tombuildsstuff ,

This would be a great addition IMHO and could some time for future developers!

Kind regards :)

ruandersMSFT:
@tombuildsstuff Here are some thoughts:

- It's a fair statement that both DTU and vCore models are supported [Choose between the vCore and DTU purchasing models](https://docs.microsoft.com/en-gb/azure/azure-sql/database/purchasing-models)
- There is **one** [API to create an Azure SQL Database](https://docs.microsoft.com/en-us/rest/api/sql/databases/createorupdate); yet, Terraform still has **two separate** object implementations ([azurerm_sql_server](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/sql_server) vs. [azurerm_mssql_server](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/mssql_server)) and ([azurerm_sql_database](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/sql_database) vs. [azurerm_mssql_database](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/mssql_database)).
[...]

Issue #6502, opened 2020-04-16
Merge azurerm_sql_database and azurerm_mssql_database


frugecn:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Can we merge azurerm_sql_database and azurerm_mssql_database so that we get the new features on both resources.  The main difference is mssql uses SKU now.  We're getting changes to one resource but not the other and getting some drift.

### New or Affected Resource(s)

azurerm_sql_database
azurerm_mssql_database

yupwei68:
hi @frugecn ,thanks for opening this issue. We are in process of eliminating our old sql database `azurerm_sql_database` and we'll not update new feature on our old sql database.

katbyte:
Hey @frugecn,

To expand on @yupwei68 - the `azurerm_mssql_database` resource is the new version of `azurerm_sql_database` that is using a newer version of the API that supports more features. As such we have been migrating all the supported features of the old resource to the new and it should support everything that `azurerm_sql_database` does. 

May i ask what functionality is missing from the new resource?

frugecn:
What caught my attention was in v2.6.0, connection policy was added to azurerm_sql_server, but I just realized that is the server resource and not the database resource.  I thought it was added to the sql_database resource and not the mssql_resource, but since it is sql_server resource, it was a mistake on my part.  Sorry.

Looking at the documentation between the two:

import from like a bacpac
extended_auditing_policy


Is the new resource supposed to be backward compatible with the old service model?  Or will we keep both resources depending on which service model you choose?  As long as Azure is offering both service models (which they currently are), we'll need both models.  I see smaller databases using the old service model for a while and larger databases using the new model.

The extended_auditing_policy is now on the server resource, so that one may not be a big deal, unless you need a different policy than the server level.

yupwei68:
[...]

PR #11601, opened 2021-05-06
Add note to SQL server doc to use mssql resource instead

gkman:
As per issues #10217 and #6502 it seems clear that the SQL server API usage using `sku`s has not been as of yet deprecated, but a warning or note notifying new users of the situation seems appropriate. 

While I was here also removed trailing spaces on the files I opened. 

Am happy to make any adjustments, but as a new user to SQL on Azure it took quite some digging and understand as to why the `sql_server` and `mssql_server` existed, and I feel that some note at the top would have been very helpful.

### DIFF
--- a/website/docs/d/sql_database.html.markdown
+++ b/website/docs/d/sql_database.html.markdown
@@ -36,8 +36,8 @@ output "sql_database_id" {
 
 * `id` - The SQL Database ID.
 
-* `collation` - The name of the collation. 
-    
+* `collation` - The name of the collation.
+
 * `creation_date` - The creation date of the SQL Database.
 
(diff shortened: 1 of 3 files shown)


katbyte (approved):
Thanks @gkman - this LGTM ✅  it may be time to deprecate the old resources and if sku support is needed add it to the newer resource, i believe there was some incompatibilities at the time we required a newer API version and thus a new resource at the time

gkman:
@katbyte should all the `sql` based items just be marked as deprecated if there is already a `mssql` counterpart available then? It would encourage any missing features to be pushed as issues on the newer API capabilities too.

PR #11608, opened 2021-05-07
fix failure TC for maintenance RP

njuCZ:
the old id format is case insensitive, and in `maintenance_assignment_xxx` resource there is a validation method checking the `maintenance_configuration` id, so acctest failed.

using the generated id

### DIFF
--- a/azurerm/internal/services/maintenance/maintenance_configuration_resource.go
+++ b/azurerm/internal/services/maintenance/maintenance_configuration_resource.go
@@ -49,9 +49,7 @@ func resourceArmMaintenanceConfiguration() *schema.Resource {
 
 			"location": azure.SchemaLocation(),
 
-			// There's a bug in the Azure API where this is returned in lower-case
-			// BUG: https://github.com/Azure/azure-rest-api-specs/issues/8653
-			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
+			"resource_group_name": azure.SchemaResourceGroupName(),
 
 			"scope": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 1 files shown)


azurerm/internal/services/maintenance/maintenance_configuration_resource.go
```diff
@@ -49,9 +49,7 @@ func resourceArmMaintenanceConfiguration() *schema.Resource {
 
 			"location": azure.SchemaLocation(),
 
-			// There's a bug in the Azure API where this is returned in lower-case
-			// BUG: https://github.com/Azure/azure-rest-api-specs/issues/8653
-			"resource_group_name": azure.SchemaResourceGroupNameDiffSuppress(),
+			"resource_group_name": azure.SchemaResourceGroupName(),
```
tombuildsstuff, line 52: This'll only work for new resources and not existing ones - which'll need a state migration to correct the casing of the resource group name (by queuing that API if that's the case?

tombuildsstuff (changes requested):
Hi @njuCZ 

Thanks for this PR.

Taking a look through whilst on the whole this looks good, this'll only work for new resources and not for ones with the incorrect ID - so we'll need to add a stage migration to correct the casing on the Resource Group name (from the resources API) if we want to make this case sensitive here

Thanks

azurerm/internal/services/maintenance/maintenance_configuration_resource.go
njuCZ: I thought there is no need to change the `resource_group_name`, so I just revert it

katbyte: should we mark this as a todo for 3.0 and make that breaking change then?

njuCZ:
added migration to migrate the `id`, as for `resourcegroup`, we could not correct it to case sensitive string, because in the state it's already in lower case and we could not get the config value. I have marked it as todo 3.0

katbyte (approved):
LGTM 👍

PR #11625, opened 2021-05-08
Support for `spatial_index` in `azurerm_cosmosdb_sql_container`

yupwei68:
Fix https://github.com/terraform-providers/terraform-provider-azurerm/issues/8817

--- PASS: TestAccCosmosDbSqlContainer_indexing_policy (1559.79s)


azurerm/internal/services/cosmos/common/indexing_policy.go
```diff
@@ -66,6 +66,29 @@ func expandAzureRmCosmosDBIndexingPolicyCompositeIndexes(input []interface{}) *[
...
+		documentdb.SpatialTypePolygon,
+	}
+
+	for _, i := range input {
+		indexPair := i.(map[string]interface{})
+		indexes = append(indexes, documentdb.SpatialSpec{
+			Types: &spatialTypes,
+			Path:  utils.String(indexPair["path"].(string)),
+		})
+	}
```
katbyte, line 87: should we expose this in schema instead of applying each spatial index to all types?

azurerm/internal/services/cosmos/common/schema.go
```diff
@@ -185,6 +185,20 @@ func CosmosDbIndexingPolicySchema() *schema.Schema {
 						},
 					},
 				},
+
+				"spatial_index": {
+					Type:     schema.TypeList,
+					Optional: true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"path": {
```
katbyte, line 194: so we'd have both types and path here?

katbyte (changes requested):

azurerm/internal/services/cosmos/common/schema.go
yupwei68: same above. I'll make types computed.

yupwei68:
Thanks kt for your comments. Please continue reviewing.
=== RUN   TestAccCosmosDbSqlContainer_indexing_policy
=== PAUSE TestAccCosmosDbSqlContainer_indexing_policy
=== CONT  TestAccCosmosDbSqlContainer_indexing_policy
--- PASS: TestAccCosmosDbSqlContainer_indexing_policy (1618.76s)

azurerm/internal/services/cosmos/common/indexing_policy.go
katbyte: i'm not sure why that doesn't mean we shouldn't expose the four types for people to individually set?

katbyte: sorry but i still don't follow, why are there four "types" to set an index path on if they will all be the same regardless of what we set?

yupwei68: I think it's by design by service. Here is the service's reply:
https://gist.github.com/yupwei68/6c3ecffe6aeec33d945210e70bb3513a#gistcomment-3735507

katbyte: my point is would someone ever want to set a different spatial index path for different types? it type1 is path1 and type2 is path2 ect or will there always only ever be one index path for all types?

yupwei68: There will always be index paths for all types, according to the service team. The user can't specify the specific types for their paths.

katbyte: ok so to be clear, there is a single path that is shared by all types regardless and its not possible to assign different paths to different types? i assume the portal only has a "path" property then and hides away types from the user? if this is all the case maybe it make sense to document htis in the property so users know the path is applied to all types?

yupwei68:
Thanks kt for your comment. I've added a comment and some description in the documents for this problem. Please continue reviewing.

katbyte:
@yupwei68 - looking for confirmation about this: https://github.com/terraform-providers/terraform-provider-azurerm/pull/11625#discussion_r633211967

azurerm/internal/services/cosmos/common/indexing_policy.go
yupwei68: Yes. The portal has not supported this yet. I have updated the document and I have marked the types computed.

yupwei68:
Thanks kt. Sorry for the missing. Please continue.

katbyte (approved):
Thanks @yupwei68 - LGTM 👍 just needt he merge conflict sorted

yupwei68:
Thanks kt, I've resolved the conflicts.

PR #11630, opened 2021-05-08
Fix `backup` in `azurerm_cosmosdb_account` when `type` is `Continuous`

yupwei68:
Sorry that I have set the default value for `interval_in_minutes` and `retention_in_hours` to cause the test fails when `type` in `backup` is `Continuous`.

Error: expanding `backup`: `interval_in_minutes` can not be set when `type` in`backup` is `Continuous` 
        
          with azurerm_cosmosdb_account.test,
          on terraform_plugin_test.tf line 11, in resource "azurerm_cosmosdb_account" "test":
          11: resource "azurerm_cosmosdb_account" "test" {
        
--- FAIL: TestAccCosmosDBAccount_backupContinuous (82.66s)

I think it doesn't break users. Because by default, no matter we update 0 or default value for the two fields, the service will always return default values.


katbyte (approved):
LGTM 👍

PR #11633, opened 2021-05-08
"azurerm_spring_cloud_service" - supports property "required_network_traffic_rules"

njuCZ:
this is a readonly property. 

It lists the required traffic in virtual network mode.

some users want to limit spring cloud app's traffic, the "required_traffic" tells users which traffic must be enabled.

image


website/docs/r/spring_cloud_service.html.markdown
```diff
@@ -158,6 +158,22 @@ The following attributes are exported:
 
 * `outbound_public_ip_addresses` - A list of the outbound Public IP Addresses used by this Spring Cloud Service.
 
+* `required_traffic` - A list of `required_traffic` blocks as defined below.
```
katbyte, line 161: what does required traffic mean? is there a better more clear name for this property?

katbyte (changes requested):

website/docs/r/spring_cloud_service.html.markdown
njuCZ: according to the service team's explanation, some users want to limit the traffic of spring cloud app, but when spring cloud service is integrated with virtual network, there are some essential network trafiic that must be enabled. Users could put these into white list or firewall rules.

bowen5: Hi @katbyte , the reason why we don't want to put "firewall" keyword in this property, is that Azure offers another service "network security group" which basically could do the similar job of allowing/denying traffic, and many customers are using it to secure their network. We don't want to have the "firewall" keyword in property name, which might be potentially misleading. Thanks.

yucwan: In that case I think the current name `required_traffic` is a better fit than `required_access` since:
1. those are network configuration, not authorization
2. the name aligns our API/SDK and will reduce confusion.
@katbyte what do you think?

katbyte: `required_traffic` is rather vague and is confusing, if it can be used elsewhere then maybe it should be `required_network_access` or even `required_network` so it reads like `required_network.port`

yucwan: `required_network.direction` looked strange to me so how about `required_traffic_rules`?

katbyte: your not wrong, i think its traffic without a qualifier that stands out to me, split the difference with `required_network_traffic_rules`? 🙃

yucwan: ok let move forward with that name

njuCZ:
@katbyte thanks for your suggestion, I have renamed to `required_network_traffic_rules`. please have a look again

katbyte (approved):
Thaks @njuCZ - LGTM 👍

PR #11641, opened 2021-05-10
fix `stateUpgrades` func

njuCZ:


katbyte (approved):
Good catch @njuCZ! LGTM

Issue #11565, opened 2021-05-03
Support for 'AzureActiveDirectoryDiagnostics' kind in azurerm_sentinel_data_connector_*


madsaune:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Today I tried to deploy "Azure Active Directory" connector in Azure Sentinel with Terraform and discovered that the resource "azurerm_sentinel_data_connector_azure_active_directory" actually deploys the "Azure Active Directory Identity Protection" connector in Sentinel instead which is not the same as "Azure Active Directory" connector.

In code, the "Azure Active Directory Identity Protection" connector is refered to as `kind: AzureActiveDirectory` which is a little confusing, since they also have the "Azure Active Directory" connector which is not related to this.

I'm not very experienced with any Azure SDK, Azure REST API or the Azure Powershell module, but I have not found anything related to the "Azure Active Directory" connector when looking. I only found one reference in another repository and there it's referenced as `kind: AzureActiveDirectoryDiagnostics`.

[Link 1](https://github.com/javiersoriano/sentinel-all-in-one/blob/a79b593e954d3a59b5916fa5436fc9c5fe697e56/connectors.json#L17-L30)
[Link 2](https://github.com/javiersoriano/sentinel-all-in-one/blob/a79b593e954d3a59b5916fa5436fc9c5fe697e56/SentinelallInOne.ps1#L345-L376)

As you might notice, in link 2, they use the `/providers/microsoft.aadiam/diagnosticSettings/AzureSentinel_${Workspace}?api-version=2017-04-01` resource instead of `securityInsights`. This indicates that maybe it belongs in the `azuread` provider, but since the Sentinel resource belongs to `azurerm` I decided to try here first.

Maybe I'm looking in the wrong place, or maybe Microsoft has not implemented this in their public API/SDKs yet, but if anyone knows how to implement this, I would be very happy!

### New or Affected Resource(s)

* azurerm_sentinel_data_connector_azure_active_directory_diagnostics

### Potential Terraform Configuration

[...]

PR #11660, opened 2021-05-11
New Resource - `azurerm_monitor_aad_diagnostic_setting`

magodo:
This PR implements a new resource: `azurerm_monitor_aad_diagnostic_setting`.

This is the same as the `azurerm_monitor_diagnostic_setting` in most parts, except it is using a different API endpoint: `/providers/microsoft.aadiam/diagnosticSettings/{name}`. I've verified this endpoint is not a wrapper API (at least from the API level) for the monitor one - `/providers/microsoft.aadiam/diagnosticSettings/providers/microsoft.insights/diagnosticSettings/{name}`, hence decided to create this separate resource.

Another point worth mentioning is that the data source to query the available log categories for AAD (which is analogous to the `azuerm_monitor_diagnostic_catagories`) is currently blocked due to Azure/azure-rest-api-specs#14355.

Fixes: #11565 

## Test

The test below was running via the user account in my local. This is because the SP seems has no permission to read the AAD's diagnostic setting. Please let me know whether it makes sense to skip the acctest for now until there is a way to run these tests via the SP.

```bash
💤 TF_ACC=1 go test -v -timeout=2h ./azurerm/internal/services/monitor -run="TestAccMonitorAADDiagnosticSetting_"
2021/05/11 21:08:13 [DEBUG] not using binary driver name, it's no longer needed
2021/05/11 21:08:14 [DEBUG] not using binary driver name, it's no longer needed
=== RUN   TestAccMonitorAADDiagnosticSetting_eventhubDefault
=== PAUSE TestAccMonitorAADDiagnosticSetting_eventhubDefault
=== RUN   TestAccMonitorAADDiagnosticSetting_eventhub
=== PAUSE TestAccMonitorAADDiagnosticSetting_eventhub
=== RUN   TestAccMonitorAADDiagnosticSetting_requiresImport
=== PAUSE TestAccMonitorAADDiagnosticSetting_requiresImport
=== RUN   TestAccMonitorAADDiagnosticSetting_logAnalyticsWorkspace
=== PAUSE TestAccMonitorAADDiagnosticSetting_logAnalyticsWorkspace
=== RUN   TestAccMonitorAADDiagnosticSetting_storageAccount
=== PAUSE TestAccMonitorAADDiagnosticSetting_storageAccount
=== CONT  TestAccMonitorAADDiagnosticSetting_eventhubDefault
=== CONT  TestAccMonitorAADDiagnosticSetting_logAnalyticsWorkspace
=== CONT  TestAccMonitorAADDiagnosticSetting_requiresImport
=== CONT  TestAccMonitorAADDiagnosticSetting_storageAccount
=== CONT  TestAccMonitorAADDiagnosticSetting_eventhub
--- PASS: TestAccMonitorAADDiagnosticSetting_storageAccount (366.39s)
--- PASS: TestAccMonitorAADDiagnosticSetting_logAnalyticsWorkspace (367.52s)
--- PASS: TestAccMonitorAADDiagnosticSetting_eventhub (455.61s)
--- PASS: TestAccMonitorAADDiagnosticSetting_requiresImport (482.31s)
--- PASS: TestAccMonitorAADDiagnosticSetting_eventhubDefault (484.99s)
PASS
ok      github.com/terraform-providers/terraform-provider-azurerm/azurerm/internal/services/monitor     485.162s
```


azurerm/internal/services/monitor/monitor_aad_diagnostic_setting_resource_test.go
```diff
@@ -0,0 +1,369 @@
...
+    for_each = local.disabled_log_categories
+    content {
+      category = log.value
+      enabled  = false
+      retention_policy {
+        enabled = false
+        days    = 0
+      }
+    }
+  }
```
manicminer, line 238: As above, I think we should avoid dynamic blocks unless necessary.

azurerm/internal/services/monitor/monitor_aad_diagnostic_setting_resource_test.go
```diff
@@ -0,0 +1,369 @@
...
```
manicminer, line 311: As above

azurerm/internal/services/monitor/monitor_aad_diagnostic_setting_resource_test.go
```diff
@@ -0,0 +1,369 @@
...
```
manicminer, line 366: As above

website/docs/r/monitor_aad_diagnostic_setting.html.markdown
```diff
@@ -0,0 +1,137 @@
...
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name which should be used for this Monitor Azure Active Directory Diagnostic Setting. Changing this forces a new Monitor Azure Active Directory Diagnostic Setting to be created.
+  
+* `log` - (Required) One or more `log` blocks as defined below.
+
+~> **Note:** At least one of the `log` block should have the `enabled` set to `true`.
```
manicminer, line 76:
```suggestion
~> **Note:** At least one of the `log` blocks must have the `enabled` property set to `true`.
```

website/docs/r/monitor_aad_diagnostic_setting.html.markdown
```diff
@@ -0,0 +1,137 @@
...
+
+---
+
+* `eventhub_authorization_rule_id` - (Optional) Specifies the ID of an Event Hub Namespace Authorization Rule used to send Diagnostics Data. Changing this forces a new resource to be created.
+
+-> **NOTE:** This can be sourced from [the `azurerm_eventhub_namespace_authorization_rule` resource](eventhub_namespace_authorization_rule.html) and is different from [a `azurerm_eventhub_authorization_rule` resource](eventhub_authorization_rule.html).
+
+-> **NOTE:** One of `eventhub_authorization_rule_id`, `log_analytics_workspace_id` and `storage_account_id` must be specified.
```
manicminer, line 84: I think it's fine to note this once after the list of properties

```suggestion
```

website/docs/r/monitor_aad_diagnostic_setting.html.markdown
```diff
@@ -0,0 +1,137 @@
...
+
+* `eventhub_name` - (Optional) Specifies the name of the Event Hub where Diagnostics Data should be sent. Changing this forces a new resource to be created.
+
+-> **NOTE:** If this isn't specified then the default Event Hub will be used.
+
+* `log_analytics_workspace_id` - (Optional) Specifies the ID of a Log Analytics Workspace where Diagnostics Data should be sent.
+
+-> **NOTE:** One of `eventhub_authorization_rule_id`, `log_analytics_workspace_id` and `storage_account_id` must be specified.
```
manicminer, line 92: As above

```suggestion
```

website/docs/r/monitor_aad_diagnostic_setting.html.markdown
```diff
@@ -0,0 +1,137 @@
...
+
+* `eventhub_authorization_rule_id` - (Optional) Specifies the ID of an Event Hub Namespace Authorization Rule used to send Diagnostics Data. Changing this forces a new resource to be created.
+
+-> **NOTE:** This can be sourced from [the `azurerm_eventhub_namespace_authorization_rule` resource](eventhub_namespace_authorization_rule.html) and is different from [a `azurerm_eventhub_authorization_rule` resource](eventhub_authorization_rule.html).
+
+-> **NOTE:** One of `eventhub_authorization_rule_id`, `log_analytics_workspace_id` and `storage_account_id` must be specified.
+
+* `eventhub_name` - (Optional) Specifies the name of the Event Hub where Diagnostics Data should be sent. Changing this forces a new resource to be created.
+
+-> **NOTE:** If this isn't specified then the default Event Hub will be used.
```
manicminer, line 88: This probably doesn't warrant a boxed note

```suggestion
* `eventhub_name` - (Optional) Specifies the name of the Event Hub where Diagnostics Data should be sent. If not specified, the default Event Hub will be used. Changing this forces a new resource to be created.
```

azurerm/internal/services/monitor/monitor_aad_diagnostic_setting_resource.go
```diff
@@ -0,0 +1,388 @@
...
+		if err != nil {
+			return err
+		}
+
+		storageAccountId = parsedId.ID()
+	}
+	d.Set("storage_account_id", storageAccountId)
+
+	if err := d.Set("log", flattenMonitorAADDiagnosticLogs(resp.Logs)); err != nil {
+		return fmt.Errorf("Error setting `log`: %+v", err)
```
manicminer, line 260:
```suggestion
		return fmt.Errorf("setting `log`: %+v", err)
```

azurerm/internal/services/monitor/monitor_aad_diagnostic_setting_resource.go
```diff
@@ -0,0 +1,388 @@
...
+		}
+	}
+
+	logs := expandMonitorAADDiagnosticsSettingsLogs(d.Get("log").(*schema.Set).List())
+
+	// If there is no `enabled` log entry, the PUT will succeed while the next GET will return a 404.
+	// Therefore, ensure users has at least one enabled log entry.
+	valid := false
+	for _, log := range logs {
+		if *log.Enabled {
```
manicminer, line 153:
```suggestion
		if log.Enabled != nil && *log.Enabled {
```

manicminer (changes requested):
Hi @magodo, thanks for this PR! I've taken a look through and tried it out, it mostly LGTM and the tests are passing.

I noticed one filename `azurerm/internal/services/monitor/parse/monitor_aad_diagnositc_setting_test.go` has a typo in it. Are you able to generate ID parsers/formatters for this resource?

I've also left a few suggestions inline, if you can take a look at those then this should be good to merge. Thanks!

azurerm/internal/services/monitor/monitor_aad_diagnostic_setting_resource.go
```diff
@@ -0,0 +1,388 @@
...
+
+						"retention_policy": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:     schema.TypeBool,
+										Required: true,
```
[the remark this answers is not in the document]
magodo: What is the reason to make it `Optional` and `Default` to `true? I saw other azurerm resources with `retention_policy` are defining this as `Required`?

azurerm/internal/services/monitor/monitor_aad_diagnostic_setting_resource.go
```diff
@@ -0,0 +1,388 @@
...
+				ForceNew:     true,
+				ValidateFunc: validate.MonitorDiagnosticSettingName,
+			},
+
+			"eventhub_name": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: eventhubValidate.ValidateEventHubName(),
+			},
```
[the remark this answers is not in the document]
magodo: In case the `eventhub_name` is an empty string in the request, it is still an empty string back in response (incl. `GET`). Therefore seems not necessary to make it computed?

azurerm/internal/services/monitor/monitor_aad_diagnostic_setting_resource.go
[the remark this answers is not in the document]
magodo: I'm making the `Default` to be `false` , and also making the `Default` of the `days` to `0`. I assume the intent here is to allow users to only specify an empty `retention_policy{}` in the configuration to enable.

However, if the `log` is disabled, then the response of a `GET` will **always** return the `retention_policy.enabled` as `false` and `retention_policy.days` as `0`, regardless what is specified in the prior request. Since the `log` is a `Set` here, otherwise we can suppress the diff in case the log is disabled as is done [here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/346b2312f3177f8104dd0cfb06afd4b6029951f7/azurerm/internal/services/network/network_watcher_flow_log_resource.go#L95).

Due to this limitation, I feel it makes more sense to make the defaults to be the same as when the `log` is disabled. This allows the users to specify an empty `retention_policy{}` no matter `log` is enabled/disabled. Only when users want to enable `retention_policy`, then the user need to set the `enabled` and `days` explicitly.

WDYT?

magodo:
@manicminer Thank you for the review! I've updated most of the comments, but leave some for open discussion. Please take another look.

azurerm/internal/services/monitor/monitor_aad_diagnostic_setting_resource.go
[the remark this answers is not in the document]
manicminer: @magodo SGTM 👍

azurerm/internal/services/monitor/monitor_aad_diagnostic_setting_resource.go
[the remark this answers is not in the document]
manicminer: Ah ok, then yeah there's no need for Computed 👍

manicminer (approved):
Thanks @magodo! This LGTM and tests still passing

Screenshot 2021-05-13 at 10 26 59

Issue #8150, opened 2020-08-17
Support for container_registry customer managed keys


pkelly808:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Please add support to use customer managed keys when provisioning ACR.

### New or Affected Resource(s)

* azurerm_container_registry

### References

* https://docs.microsoft.com/en-us/azure/container-registry/container-registry-customer-managed-keys

raypettersen:
API now supports CMK, so this issue should be a low-hanging fruit.

FrankFong:
> API now supports CMK, so this issue should be a low-hanging fruit.

Just wondering if we have an ETA for this feature to be added into the ACR resource. We are looking to deploy private ACR resources soon.

PR #11661, opened 2021-05-11
updating to allow encryption on azure container registry

implodingduck:
* updating to allow encryption on azure container registry
* Hopefully helps issue see #8150


katbyte (approved):
Thanks @implodingduck - LGTM 👍

PR #11684, opened 2021-05-12
azurerm_netapp_volume - support for the `security_style` property - Fixes #11370

### LINKED ISSUE #11370: netapp dual protocol volume created with wrong order of protocols and security style
using
tf 14.10
hashicorp/azurerm 2.55.0

Create a dual protocol netapp volume.
Use example: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/examples/netapp/volume_dual_protocol/main.tf
modify line 45 to: protocols           = ["NFSv3","CIFS"]

created volume has protocols created in reverse "smb, nfsv3"
image

As a result of this security style is set to NTFS, which makes the volume inaccessible from nfs mounts.
image

This happens because in https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/netapp/netapp_volume_resource.go on line 277 protocols is retrieved like this:
"protocols := d.Get("protocols").(*schema.Set).List()" ; default sort applies hence cifs is inserted ahead of nfsv3.

Correct behavior should be creation of volume with correct protocol order and security style with order defined in protocols:
image

paulomarquesc:
This PR implements `security_style` volume property. It Fixes #11370.


azurerm/internal/services/netapp/netapp_volume_resource.go
```diff
@@ -118,6 +118,17 @@ func resourceNetAppVolume() *schema.Resource {
...
 			},
 
+			"security_style": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"Unix",
+					"Ntfs",
```
katbyte, line 128: could we use the SDK build in constants here?

website/docs/r/netapp_volume.html.markdown
```diff
@@ -106,6 +107,8 @@ The following arguments are supported:
 
 * `protocols` - (Optional) The target volume protocol expressed as a list. Supported single value include `CIFS`, `NFSv3`, or `NFSv4.1`. If argument is not defined it will default to `NFSv3`. Changing this forces a new resource to be created and data will be lost. Dual protocol scenario is supported for CIFS and NFSv3, for more information, please refer to [Create a dual-protocol volume for Azure NetApp Files](https://docs.microsoft.com/en-us/azure/azure-netapp-files/create-volumes-dual-protocol) document.
 
+* `security_style` - (Optional) Volume security style, accepted values are `Unix` or `Ntfs`. If not provided, single-protocol volume is created defaulting to 'Unix' if it is `NFSv3` or `NFSv4.1` volume, if `CIFS`, it will default to `Ntfs`. In a dual-protocol volume, if not provided, its value will be `Unix`.
```
katbyte, line 110:
```suggestion
* `security_style` - (Optional) Volume security style, accepted values are `Unix` or `Ntfs`. If not provided, single-protocol volume is created defaulting to `Unix` if it is `NFSv3` or `NFSv4.1` volume, if `CIFS`, it will default to `Ntfs`. In a dual-protocol volume, if not provided, its value will be `Unix`.
```

katbyte (changes requested):
Thanks @paulomarquesc - just a couple small comments

azurerm/internal/services/netapp/netapp_volume_resource.go
paulomarquesc: We need to leave as is since the backend is changing the values to pascal case which differs from the sdk.

katbyte (approved):
Thanks @paulomarquesc - LGTM 👍

PR #11686, opened 2021-05-12
Support Redis Replicas configuration

omgjlk:
A Redis cluster can be configured with more than one replica per master.
See
https://docs.microsoft.com/en-us/azure/azure-cache-for-redis/cache-how-to-multi-replicas
for documentation. ReplicasPerMaster is supported in the Go SDK.

Without the ability to set replicas we are unable to set up Redis to be
fully zone redundant. One must have as many nodes aS there are zones.
THe default replica count is 1, for a total of two nodes, which would
only support two zones of redundancy. Many regions have three zones.

### DIFF
--- a/azurerm/internal/services/redis/redis_cache_resource.go
+++ b/azurerm/internal/services/redis/redis_cache_resource.go
@@ -100,6 +100,11 @@ func resourceRedisCache() *schema.Resource {
 				}, false),
 			},
 
+			"replicas_per_master": {
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
+
 			"shard_count": {
 				Type:     schema.TypeInt,
 				Optional: true,
@@ -340,6 +345,11 @@ func resourceRedisCacheCreate(d *schema.ResourceData, meta interface{}) error {
 		Tags: expandedTags,
 	}
 
+	if v, ok := d.GetOk("replicas_per_master"); ok {
+		replicasPerMaster := int32(v.(int))
+		parameters.ReplicasPerMaster = &replicasPerMaster
+	}
+
 	if v, ok := d.GetOk("shard_count"); ok {
 		shardCount := int32(v.(int))
 		parameters.ShardCount = &shardCount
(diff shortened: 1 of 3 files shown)


website/docs/r/redis_cache.html.markdown
```diff
@@ -69,6 +69,8 @@ The following arguments are supported:
 
 * `redis_configuration` - (Optional) A `redis_configuration` as defined below - with some limitations by SKU - defaults/details are shown below.
 
+* `replicas_per_master` - (Optional) *Only available when using the Premium SKU* The number of Replicas to create per master on the Redis Cluster. *Cannot be used in conjunction with shards*
+
 * `shard_count` - (Optional) *Only available when using the Premium SKU* The number of Shards to create on the Redis Cluster.
```
katbyte, line 74: we should use a note
```suggestion
* `replicas_per_master` - (Optional) The number of Replicas to create per master on the Redis Cluster.

~> **Note:** Configuring the number of replicas per faster is only available when using the Premium SKU and cannot be used in conjunction with shards.

* `shard_count` - (Optional) *Only available when using the Premium SKU* The number of Shards to create on the Redis Cluster.
```

azurerm/internal/services/redis/redis_cache_resource.go
```diff
@@ -100,6 +100,11 @@ func resourceRedisCache() *schema.Resource {
 				}, false),
 			},
 
+			"replicas_per_master": {
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
```
katbyte, line 106: what is the allowable range here? i presume >= 1, but is there a maximum? could we validate this?

katbyte (changes requested):
THanks @omgjlk - overall this looks good and is just missing some validation. once thats been added this should be good to merge!

website/docs/r/redis_cache.html.markdown
katbyte: can we also include the default?

azurerm/internal/services/redis/redis_cache_resource.go
omgjlk: Unfortunately there is no limit, other than what appears to be int32 in the Go SDK. The Resource Manager [template](https://docs.microsoft.com/en-us/azure/templates/microsoft.cache/2020-06-01/redis?tabs=json#rediscreateproperties-object) does not list a limit for this either.

The portal stops me from making more than 3 replicas, but I don't know where the limit comes from.

katbyte: if the portal stops at three lets do the same, could we add a validation function that insure the value is 1-3 (unless 0 is valid then 0-3)?

katbyte:
I'm going to close this in favour of #11714 as its closer, with your doc note so i can get it merged today. Thanks @omgjlk !

PR #11690, opened 2021-05-13
migrate `azurerm_synapse_role_assignment` to support new roles and scopes

### LINKED ISSUE #10141: Support for Synapse Workspace new roles for Role Assignment
### Community Note

### Description

New role assignments for Azure Synapse Workspace have been released with GA, replacing the old role definitions.
- Existing role definitions
  - Workspace Admin
  - Sql Admin
  - Apache Spark Admin
- New role definitions
  - **Synapse Administrator**
  - **Synapse SQL Administrator**
  - **Synapse Spark Administrator**
  - **Synapse Contributor (preview)**
  - **Synapse Artifact Publisher (preview)**
  - **Synapse Artifact User (preview)**
  - **Synapse Compute Operator (preview)**
  - **Synapse Credential User (preview)**

* azurerm_synapse_role_assignment

```hcl
resource "azurerm_synapse_role_assignment" "example" {
  synapse_workspace_id = azurerm_synapse_workspace.example.id
  role_name            = "Synapse Administrator"
  principal_id         = data.azurerm_client_config.current.object_id

  depends_on = [azurerm_synapse_firewall_rule.example]
}
```

### References

* https://docs.microsoft.com/en-us/azure/synapse-analytics/overview-faq
* https://docs.microsoft.com/en-us/azure/synapse-analytics/security/synapse-workspace-synapse-rbac-roles

njuCZ:
fix #10141 

in old api version, the only supported role is `Workspace Admin`, `Sql Admin` and `Apache Spark Admin`. The scope is workspace.

in new api version, exsiting roles are renamed and new roles are added, Users could also specify different scope: workspace, spark pool or others (not suported in terraform, so not added in this PR) .

image


njuCZ:
this PR contains breaking changes, is it possible to be merged in the next major version 3.0 ?

katbyte (changes requested):
@njuCZ - is there a reason we can't map the old roles to the new ones so this isn't a breaking change? (at least for upgrading?)

njuCZ:
@katbyte we could map the old roles to the new, in this PR I also added a stateUpgrade func to make it. But in the old api, the scope of role assignment is always workspace level, while the new api could be workspace level, sparkpool level and others. So I removed property `synapse_workspace_id` and add property `synapse_scope`, I think this is a breaking change for users.

chrsundermann:
Does it make sense not to hardcode possible roles? As soon as new roles will be integrated they can't be picked and code has to be updated.

katbyte:
> But in the old api, the scope of role assignment is always workspace level, while the new api could be workspace level, sparkpool level and others

if we default to workspace level and mark that as optional it's not a breaking change?

njuCZ:
@katbyte yes, in the state upgrade func I did that. I thought changing the schema name (synapse_workspace_id -> syanpse_scope) belongs to breaking change because we need users to modify the property name in TF config file. If this change is OK, I think we could remove the breaking change label.

njuCZ:
another way to make this PR: we could mark `synapse_workspace_id` optional, and add other schema for scope, for example: spark_pool_id. In this way, we would have multiple specific schema to represent scope and only one of them should be specified.

katbyte:
> another way to make this PR: we could mark synapse_workspace_id optional, and add other schema for scope, for example: spark_pool_id. In this way, we would have multiple specific schema to represent scope and only one of them should be specified.

I like this id! and then we get proper validation on the id of the scope and no breaking change, this works for me!

njuCZ:
@katbyte I have refined this PR, now there is no breaking change. All acctest could pass.
image

njuCZ:
@chrsundermann sorry for late reply. Currently there is not many roles, so we could hardcode in the validation function and provide more friendly user experience . If there are many roles added in future, we could add another data source for role definition. 

As for synapse scope, since integrationRuntimes, credentials, linkedservice are not supported in terraform yet, so in this PR, I don't support them. I think it's easy to add support for those scope in future.

njuCZ:
@katbyte the acctest in teamcity could pass now.

azurerm/internal/services/synapse/synapse_role_assignment_resource.go
```diff
@@ -55,9 +72,16 @@ func resourceSynapseRoleAssignment() *pluginsdk.Resource {
 				Required: true,
 				ForceNew: true,
 				ValidateFunc: validation.StringInSlice([]string{
-					"Workspace Admin",
-					"Apache Spark Admin",
-					"Sql Admin",
```
katbyte, line 60: i think we'll still need to keep these around until 3.0 so the user doesn't have to change their config (prevent breaking change(

katbyte (changes requested):
one last comment i think

azurerm/internal/services/synapse/synapse_role_assignment_resource.go
katbyte: and could we add a comment to be removed

njuCZ:
@katbyte thanks for your suggestions. I have added support for old roles and add a suppressFunc to avoid diff between old roles and new roles.

image

azurerm/internal/services/synapse/client/client.go
```diff
@@ -26,7 +26,7 @@ func NewClient(o *common.ClientOptions) *Client {
 	firewallRuleClient := synapse.NewIPFirewallRulesClientWithBaseURI(o.ResourceManagerEndpoint, o.SubscriptionId)
 	o.ConfigureClient(&firewallRuleClient.Client, o.ResourceManagerAuthorizer)
 
-	// the service team hopes to rename it to sparkPool, so rename the sdk here
+	// the service team hopes to rename it to sparkPool, mso rename the sdk here
```
katbyte, line 29:
```suggestion
	// the service team hopes to rename it to sparkPool, so rename the sdk here
```

katbyte (approved):
Thanks @njuCZ - LGTM aside from one comment which i'll fix now 🍰

PR #11694, opened 2021-05-13
fix doc of `azurerm_policy_set_definition`

### LINKED ISSUE #11666: `azurerm_policy_set_definition` requires `additional_metadata_resource_id` while documentation says `additional_metadata_id`
Could you please either fix the documentation or the source code:

The documentation says it is **additional_metadata_id**:
See https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/policy_set_definition#additional_metadata_id

Actual implementation defines and uses **additional_metadata_resource_id**:
See https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/policy/policy_set_definition_resource.go#L198

It looks like the issue is with the source code since Microsoft uses `additionalMetadataId` field in ARM templates, thus it should be **additional_metadata_id** without `_resource`.

Please note, now you need to support both improper and proper fields, similarly to `management_group_id` and `management_group_name`

Thank you!

njuCZ:
fix #11666

### DIFF
--- a/website/docs/d/policy_set_definition.html.markdown
+++ b/website/docs/d/policy_set_definition.html.markdown
@@ -80,7 +80,7 @@ An `policy_definition_group` block exports the following:
 
 * `description` - The description of this policy definition group.
 
-* `additional_metadata_id` - The ID of a resource that contains additional metadata about this policy definition group.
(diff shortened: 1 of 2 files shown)


katbyte (approved):
LGTM 👍

